PR #93108, opened 2026-04-21
Suspend dynamic route params in dev instant shell

acdlite:
In dev with cacheComponents, the instant-nav prefetch for a dynamic route resolved the slug into the cached shell instead of returning the generic fallback, so `params` never suspended under a locked instant scope.

Existing e2e coverage missed this because the fixture's dynamic-params page had `generateStaticParams` defined. That populates the dev prerender manifest and masks the bug — the broken path is the one where no `generateStaticParams` is defined. The regression test uses a new `ungenerated-params/[slug]` fixture that leaves it out.

### DIFF
--- a/test/development/app-dir/instant-navs-devtools/instant-navs-devtools.test.ts
+++ b/test/development/app-dir/instant-navs-devtools/instant-navs-devtools.test.ts
@@ -136,7 +136,7 @@ describe('instant-nav-panel', () => {
     await clearInstantModeCookie(browser)
   })
 
-  it('should show loading skeleton during SPA navigation after clicking Start', async () => {
+  it('should show loading skeletons during SPA navigation after clicking Start', async () => {
     const targetPage = '/target-page/my-post?search=foo'
     if (isNextDev && !isTurbopack) {
       // warmup target page compilation before clicking Start, to avoid extra flakiness.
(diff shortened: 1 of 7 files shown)


test/development/app-dir/instant-navs-devtools/instant-navs-devtools.test.ts
```diff
@@ -173,6 +173,49 @@ describe('instant-nav-panel', () => {
...
+
+    // The route param (slug) should suspend during the lock — the skeleton
+    // should be visible and the resolved value should not have leaked
+    // through.
+    await retry(
+      async () => {
+        const paramSkeleton = await browser.hasElementByCss(
+          '[data-testid="param-skeleton"]'
+        )
+        expect(paramSkeleton).toBe(true)
```
samselikoff: Should we move this assertion up to the previous test, alongside the check for `data-skeleton`? And also add a check for `search-param-skeleton` there too?

acdlite: What's the difference? I did it as a separate test so if it regresses the test failure is more likely to point to the specific cause

acdlite: Oh I see what you mean, there's no coverage of the search params suspending at all. Ok sure I'll combine the tests.

samselikoff (approved):

PR #93114, opened 2026-04-22
[turbopack] Enforce `root` attribute for strongly consistent reads and collectibles

sokra:
### What?

Replaces the silent "make root node" promotion in the turbo-tasks backend with a panic that enforces tasks to already have the `root` attribute before performing strongly consistent reads, reading task collectibles, or removing collectibles.

### Why?

Previously, the backend would silently promote any non-root task to a root node (aggregation number `u32::MAX`) when these operations were requested. This masked incorrect task configuration — tasks that needed root-level aggregation weren't explicitly declared as such, making it harder to reason about the aggregation graph and hiding potential performance issues from implicit promotions.

### How?

**Backend enforcement (3 locations):**
- **Strongly consistent read** (`backend/mod.rs`): Checks `NativeFunction.is_root` on the target task. If it's a persistent task without `root`, panics with both the target and reader task descriptions.
- **Read task collectibles** (`backend/mod.rs`): Same check when reading collectibles from a task.
- **Remove collectibles** (`operation/update_collectible.rs`): Same check when removing collectibles (count < 0).

All three locations drop the task guard before panicking to avoid deadlocking with `debug_get_task_description`.

**Macro support for `root` on methods:**
- `value_impl_macro.rs`: Now reads the `root` attribute from `#[turbo_tasks::function(root)]` on inherent impl and trait impl methods (previously hardcoded to `false`).
- `value_trait_macro.rs`: Same for trait default methods.

**`root` attribute additions:**
- All `#[turbo_tasks::function(operation)]` in test files, benchmarks, and fuzz code
- Production operations in `crates/next-api/` that are read with strong consistency
- Regular functions and methods in tests that use `.strongly_consistent()`
- Trait methods in `value_impl` and `value_trait` blocks used with strong consistency

**New test:**
- `non_root_task_panic.rs`: Verifies the panic fires when attempting a strongly consistent read on a non-root operation task. Captures the panic from the worker thread via a panic hook since the panic propagates as a channel error to the test thread.

### DIFF
--- a/turbopack/crates/turbo-tasks-macros/src/value_trait_macro.rs
+++ b/turbopack/crates/turbo-tasks-macros/src/value_trait_macro.rs
@@ -233,6 +233,7 @@ pub fn value_trait(args: TokenStream, input: TokenStream) -> TokenStream {
                     method_name: #method_name_str,
                     default_method: Some(&#native_function_ident),
                     index: #index,
+                    is_root: #is_root,
                 },
             });
             default_methods.push(quote! { Some(&#native_function_ident) });
@@ -270,6 +271,7 @@ pub fn value_trait(args: TokenStream, input: TokenStream) -> TokenStream {
                     method_name: #method_name_str,
                     default_method: None,
                     index: #index,
+                    is_root: #is_root,
                 },
             });
             default_methods.push(quote! { None });
--- a/turbopack/crates/turbo-tasks-macros/src/value_trait_macro.rs
+++ b/turbopack/crates/turbo-tasks-macros/src/value_trait_macro.rs
@@ -181,6 +181,7 @@ pub fn value_trait(args: TokenStream, input: TokenStream) -> TokenStream {
         }
 
         let is_self_used = default.as_ref().map(is_self_used).unwrap_or(false);
+        let is_root = func_args.root.is_some();
         let Some(turbo_fn) = TurboFn::new(
             sig,
             DefinitionContext::ValueTrait,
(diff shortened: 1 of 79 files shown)


turbopack/crates/turbo-tasks-macros/src/value_trait_macro.rs
```diff
@@ -177,6 +177,7 @@ pub fn value_trait(args: TokenStream, input: TokenStream) -> TokenStream {
         }
 
         let is_self_used = default.as_ref().map(is_self_used).unwrap_or(false);
+        let is_root = func_args.root.is_some();
```
lukesandberg, line 184: For traits we should require matching root params on the trait and the impl. Or at least that if root is in the trait it's it must also be on the impls

Or maybe we should just disallow root on trait items? Do we have a use case?

sokra: :robot: Added bidirectional matching enforcement: `is_root` is now stored on `TraitMethod` and a runtime check in `register_all_trait_methods` panics if the trait method's `root` attribute doesn't match an impl method's. Already caught one pre-existing inconsistency (`AssetHashesManifestAsset::Asset::content` had `root` on the impl but not the trait — removed from the impl since adding it to the trait would force all 49 `Asset` implementors to be root).

lukesandberg (dismissed):

mischnic (approved):

PR #93125, opened 2026-04-22
feat(metadata): support non-standard directives in robots.ts via `other`

### LINKED ISSUE #89521: Support non-standard properties in `robots.ts` file

https://github.com/Joozty/robots-ts-example

### To Reproduce

1. Create robots.ts file
2. Put there this context and open localhost:3000/robots.txt you will see request rate directive is ignored

```ts
export default function robots(): MetadataRoute.Robots {
	const rules: MetadataRoute.Robots['rules'] = [
		{
			allow: '/',
			userAgent: '*',
		},
		{
			'allow': '/',
			// See: https://o-seznam.cz/napoveda/vyhledavani/en/crawling-control/#request-rate-directive
			// @ts-expect-error not a standard property
			'Request-Rate': '10/1m', // ~10k requests/day
			'userAgent': 'SeznamBot',
		},
	]

	return {
		rules,
	}
}

```

I would expect this non-standard directive to still be propagated. Obviously, it does not make sense to include it in the types, so I’m fine with suppressing the error, but it should still propagate to robots.txt. It is very useful when Seznam bots crawl the site too aggressively.

At this point, I’m forced to create a static robots.txt file without using the TS version.

See: https://o-seznam.cz/napoveda/vyhledavani/en/crawling-control/#request-rate-directive

```bash
N/A
```

Not sure

next build (local)

### Additional context

_No response_

timneutkens:
### What?

Adds an `other` field to each rule in `MetadataRoute.Robots` so `app/robots.ts` can emit non-standard per-user-agent directives (e.g. Seznam `Request-Rate`, Yandex `Clean-param`) alongside the standard `allow`/`disallow`/`crawlDelay` fields.

```ts
// app/robots.ts
export default function robots(): MetadataRoute.Robots {
  return {
    rules: [
      { userAgent: '*', allow: '/' },
      {
        userAgent: 'SeznamBot',
        allow: '/',
        other: { 'Request-Rate': '10/1m' },
      },
    ],
  }
}
```

emits:

```
User-Agent: *
Allow: /

User-Agent: SeznamBot
Allow: /
Request-Rate: 10/1m
```

### Why?

The current `resolveRobots` serializer only writes the four standard rule fields and silently drops everything else, so users who need non-standard directives (common for Seznam, Yandex, and other crawlers) have to delete `robots.ts` entirely and hand-maintain a static `robots.txt`, losing the type-safe, code-driven workflow.

### How?

- **Type** (`packages/next/src/lib/metadata/types/metadata-interface.ts`): factored a shared `RobotsRuleBase` containing the existing optional fields plus `other?: Record<string, string | number | Array<string | number>>`. `RobotsFile.rules` still enforces `userAgent` as required on the array form.
- **Serializer** (`packages/next/src/build/webpack/loaders/metadata/resolve-route-data.ts`): after `Crawl-delay` and before the trailing blank line, walks `rule.other` and emits `Key: value` lines preserving casing. Array values expand into repeated lines; `null`/`undefined` entries are skipped. Keeps non-standard directives scoped to their `User-Agent` block.
- **Tests**: added two `resolveRobots` cases covering the `Request-Rate` reproduction and an array-valued `Clean-param`, plus null/undefined handling. Existing snapshots are unchanged, proving the change is additive.
- **Docs**: new "Non-standard directives" section in `docs/01-app/03-api-reference/03-file-conventions/01-metadata/robots.mdx` with TS/JS examples and updated `Robots` type reference.

Fixes #89521

### DIFF
--- a/docs/01-app/03-api-reference/03-file-conventions/01-metadata/robots.mdx
+++ b/docs/01-app/03-api-reference/03-file-conventions/01-metadata/robots.mdx
@@ -119,6 +119,59 @@ Disallow: /
 Sitemap: https://acme.com/sitemap.xml
 ```
 
+### Non-standard directives
+
+Some search engines support directives that aren't part of the [Robots Exclusion Standard](https://en.wikipedia.org/wiki/Robots.txt#Standard), such as `Request-Rate` (Seznam) or `Clean-param` (Yandex). Pass these through the `other` field on a rule. Keys preserve their casing and array values emit one line per entry, scoped to the rule's `User-Agent` block.
+
+```ts filename="app/robots.ts" switcher
+import type { MetadataRoute } from 'next'
+
+export default function robots(): MetadataRoute.Robots {
+  return {
+    rules: [
+      { userAgent: '*', allow: '/' },
+      {
+        userAgent: 'SeznamBot',
+        allow: '/',
+        other: {
+          'Request-Rate': '10/1m',
+        },
+      },
+    ],
+  }
+}
+```
+
+```js filename="app/robots.js" switcher
+export default function robots() {
+  return {
+    rules: [
+      { userAgent: '*', allow: '/' },
+      {
+        userAgent: 'SeznamBot',
+        allow: '/',
+        other: {
+          'Request-Rate': '10/1m',
+        },
+      },
+    ],
+  }
+}
+```
+
+Output:
+
+```txt
+User-Agent: *
+Allow: /
+
+User-Agent: SeznamBot
+Allow: /
+Request-Rate: 10/1m
+```
+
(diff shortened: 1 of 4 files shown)


timneutkens:
* **#93125** <a href="https://app.graphite.com/github/pr/vercel/next.js/93125?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93125?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

timneutkens:
Re-opened here: #93206

PR #93134, opened 2026-04-22
fix: handling of falsey values in error boundaries

lubieowoce:
our error boundaries had a bunch of logic that set `state.error = error` and then checked `if (state.error)`, which only works correctly if thrown value is truthy. it breaks if something does e.g. `throw undefined`. in this case, we would incorrectly think that no error occurred and render children again (instead of a fallback), which can then lead to an infinite loop if the children throw again.

the fix is to wrap the thrown value, so `state.error` is either `null` (initial/reset) or `{ thrownValue: ... }` if something errored. i initially considered using a separate `state.hasError` boolean, but that's a bit annoying to type, and really we want to model this as a discriminated union, so using a pseudo-Optional thing is nicer.

### DIFF
--- a/packages/next/src/client/components/catch-error.tsx
+++ b/packages/next/src/client/components/catch-error.tsx
@@ -30,7 +30,7 @@ type CatchErrorProps<P extends UserProps> = {
 }
 
 type CatchErrorState = {
-  error: Error | null
+  error: null | { thrownValue: unknown }
   previousPathname: string | null
 }
 
@@ -38,7 +38,7 @@ type CatchErrorState = {
 // TODO: Extend it instead of forking to easily sync the behavior?
 class CatchError<P extends UserProps> extends React.Component<
   CatchErrorProps<P>,
-  { error: Error | null; previousPathname: string | null }
+  CatchErrorState
 > {
   declare context: AppRouterInstance | null
   static contextType = AppRouterContext
@@ -54,14 +54,16 @@ class CatchError<P extends UserProps> extends React.Component<
     }
   }
 
-  static getDerivedStateFromError(error: Error) {
-    if (isNextRouterError(error)) {
+  static getDerivedStateFromError(
+    thrownValue: unknown
+  ): Partial<CatchErrorState> {
+    if (isNextRouterError(thrownValue)) {
       // Re-throw if an expected internal Next.js router error occurs
       // this means it should be handled by a different boundary (such as a NotFound boundary in a parent segment)
-      throw error
+      throw thrownValue
     }
 
-    return { error }
+    return { error: { thrownValue } }
   }
 
   static getDerivedStateFromProps(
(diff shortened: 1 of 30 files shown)


packages/next/src/next-devtools/userspace/pages/pages-dev-overlay-error-boundary.tsx
lubieowoce, line 1: here, the thrown value itself was unused, so i switched it to a boolean instead.

packages/next/src/client/components/catch-error.tsx
```diff
@@ -30,15 +30,15 @@ type CatchErrorProps<P extends UserProps> = {
 }
 
 type CatchErrorState = {
-  error: Error | null
+  error: null | { thrownValue: unknown }
```
eps1lon, line 33: Why not 
```suggestion
  error: unknown
  hasError: boolean
```
?
That's what we originally discussed in the meeting.

packages/next/src/client/components/catch-error.tsx
```diff
@@ -124,13 +126,14 @@ class CatchError<P extends UserProps> extends React.Component<
...
-      handleISRError({ error: this.state.error })
+      const thrownValue = this.state.error.thrownValue
+      handleISRError({ error: thrownValue })
 
       return (
         <this.props.fallback
           props={this.props.props}
           errorInfo={{
-            error: this.state.error,
+            error: thrownValue as Error,
```
eps1lon: The Component type of `this.props.fallback` should accept unknown in `error` instead.

packages/next/src/client/components/catch-error.tsx
lubieowoce: harder to make it typesafe, `error` should only be accessed when `hasError` is true, this is an easy way to guarantee that imho

eps1lon (dismissed):

packages/next/src/client/components/catch-error.tsx
eps1lon: yeah, that's fair

packages/next/src/client/components/catch-error.tsx
lubieowoce: unfortunately i'm not sure if we can do that w/ o a breaking change -- this `Error` comes from the `ErrorInfo` type (also used for `error.js` fallbacks)
https://github.com/vercel/next.js/blob/7b01c98a4a94f8cbb3a3cd02b599490410c5042c/packages/next/src/client/components/error-boundary.tsx#L17-L21
and we explicitly document the fallback components as receiving an `Error` object (even though this is a lie, we don't ensure that it's an `Error` at all) here: https://nextjs.org/docs/app/api-reference/file-conventions/error#error

eps1lon: It's not a breaking change if it reflects the actual runtime value. Otherwise you would consider every type bugfix a breaking change.

We could make a judgement call that this is too disruptive and keep the public types until v17. But then we should split public types and actual types. This is the kind of unsound usage that caused the original bug so it seems worth it to at least fix it in the internal representation.

lubieowoce: well, we could also `getProperError` the thrown value so that it actually is actually always an error, and keep the types as they were

eps1lon: That feels more like an actual breaking change since we're now changing what you receive instead of what we're claiming you receive.

lubieowoce: yeah okay, agreed. switched the internal types to `unknown`, will update docs to match reality separately https://linear.app/vercel/issue/NAR-804

eps1lon (approved):

PR #93150, opened 2026-04-23
Support template literal expressions in middleware config matcher

### LINKED ISSUE #56398: middleware matcher should support template literals

https://github.com/stefanprobst/issue-next-middleware-matcher

### To Reproduce

1. clone https://github.com/stefanprobst/issue-next-middleware-matcher
2. `pnpm install && pnpm run dev`
3. see error message in console: 'Next.js can't recognize the exported `config` field in route "/middleware": Unsupported template literal with expressions at "config.matcher[1]".'

i would want to be able to do this:

```ts
const locales = ["de", "en"] as const;

export const config = {
  matcher: ["/", `/(${locales.join("|")})/:path*`],
};
```

### Verify canary release

- [X] I verified that the issue exists in the latest Next.js canary release

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #26~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jul 13 16:27:29 UTC 2
Binaries:
  Node: 18.18.0
  npm: 9.8.1
  Yarn: 1.22.19
  pnpm: 8.8.0
Relevant Packages:
  next: 13.5.4
  eslint-config-next: N/A
  react: 18.2.0
  react-dom: 18.2.0
  typescript: 5.2.2
Next.js Config:
  output: N/A
```

Middleware / Edge (API routes, runtime)

### Additional context

_No response_

ChihebBENCHEIKH1:
## Change Summary

Resolves #56398.

The static config extractor (`extractExportedConstValue`) now supports:

- **Template literals with expressions** — evaluates each expression and concatenates with quasis
- **Module-level const identifier resolution** — resolves references to `const` declarations in the same file
- **Safe method calls** — supports `Array.prototype.join()` on resolved array values

This enables the common i18n pattern that was previously blocked:

```typescript
const locales = ['en', 'de', 'fr'] as const

export const config = {
  matcher: ['/', `/(${locales.join('|')})/:path*`],
}
// Previously: ❌ "Unsupported template literal with expressions"
// Now:        ✅ Resolves to ['/', '/(en|de|fr)/:path*']
```

### Safety guarantees

- Only `const` declarations are resolved (no `let`/`var` since they could be reassigned)
- Circular references are detected and reported as errors
- Only a whitelist of side-effect-free methods is supported (`Array.join`)
- Truly unresolvable expressions (function calls, imports, binary expressions) still produce clear error messages

### Turbopack

This PR updates the webpack/SWC config extraction path only. Turbopack has its own config analysis in Rust — happy to follow up with a separate PR for Turbopack parity if needed, or coordinate with the team on the best approach.

## Related issue number

Fixes #56398

## Checklist

- [x] PR title is a good summary (used in changelog)
- [x] Unit tests exist for changes
- [x] E2e test verifies middleware matcher with template literals works end-to-end
- [x] Documentation reflects changes

### DIFF
--- a/packages/next/src/build/analysis/extract-const-value.test.ts
+++ b/packages/next/src/build/analysis/extract-const-value.test.ts
@@ -315,6 +315,96 @@ describe('extractExportedConstValue', () => {
     })
   })
 
+  describe('edge cases', () => {
+    it('evaluates numeric expressions in template literals', () => {
+      // const version = 2
+      // export const config = { matcher: `/api/v${version}` }
+      const module = createModule([
+        constDeclaration('version', numericLiteral(2)),
+        constDeclaration(
+          'config',
+          objectExpression({
+            matcher: templateLiteral(['/api/v', ''], [identifier('version')]),
+          }),
+          true
+        ),
+      ])
+
+      const result = extractExportedConstValue(module, 'config')
+      expect(result).toEqual({ value: { matcher: '/api/v2' } })
+    })
+
+    it('evaluates Array.join() on an empty array', () => {
+      const module = createModule([
+        constDeclaration('empty', arrayExpression([])),
+        constDeclaration(
+          'config',
+          objectExpression({
+            value: callExpression(identifier('empty'), 'join', [stringLiteral('|')]),
+          }),
+          true
+        ),
+      ])
+
+      const result = extractExportedConstValue(module, 'config')
+      expect(result).toEqual({ value: { value: '' } })
+    })
+
+    it('resolves identifiers through multiple levels', () => {
+      // const a = b; const b = '/path'
+      // export const config = { matcher: a }
+      const module = createModule([
+        constDeclaration('a', identifier('b')),
+        constDeclaration('b', stringLiteral('/path')),
+        constDeclaration(
+          'config',
+          objectExpression({ matcher: identifier('a') }),
+          true
+        ),
+      ])
+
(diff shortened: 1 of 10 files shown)


ChihebBENCHEIKH1:
Fixed the Vercel bot finding in 1c838d1. The refactored code now evaluates each template literal expression individually rather than rejecting all expressions upfront. So for `something-${'real' + 1 + 'y odd'}`, the error is now the more specific "Unsupported node type BinaryExpression" instead of the blanket "Unsupported template literal with expressions". Updated the test expectation accordingly.

mischnic:
What if you do this
```
const segments = ['dashboard', 'settings']
segments.push("bar")
export const config = {
  matcher: ['/', `/(${segments.join('|')})/:path*`],
}
```
Does it still compute the correct value?

ChihebBENCHEIKH1:
Good question. Yes, it would compute the value based on the initial declaration only, so it would resolve to \`"dashboard|settings"\` and miss \`"bar"\`.

This is a limitation of static analysis: we can resolve the initial \`const\` binding but we cannot track runtime mutations like \`push()\`. The same limitation exists in the current code for any other resolved value (e.g. if someone did \`const x = "a"; x += "b"\` it would also be wrong, but that is prevented by \`const\`).

Since arrays are mutable even with \`const\`, one option would be to restrict resolution to arrays declared with \`as const\` (which are readonly at the type level and signal immutability intent). Would you prefer that approach?

PR #93151, opened 2026-04-23
fix(i18n): use correct defaultLocale when domains array is present on default host

### LINKED ISSUE #90131: Locale detection isn't working correctly after the update from 15.3 to 15.4/15.5 if a domains array is present in next.config

https://codesandbox.io/p/devbox/determined-zhukovsky-vzgqvh?workspaceId=ws_XMt216w8w9LTvcfWfEEkxN

### To Reproduce

1. Open codesandbox url
2. check that the version of next is 15.5.12
3. run pnpm install
4. run pnpm dev
5. you can also open the dev tools in codesandbox to see more information
6. click on different links in the demo nextjs app to open different locales
7. the locales listed in the "domains" array will end up taking the defaultLocale of the "domain" and not the defaultLocale of the main domain. The locales not listed in the "domains" array still work correctly
8. as you can see, the correct value is still present in the req[Symbol.for("NextInternalRequestMeta")]; but it doesn't affect anything.

v15.5 detects the "defaultLocale" of the "domain" in the "domains" array, although being ran on the defaultDomain.

Also SSR adds the domain to the href of the next/links (and because of this also hydration errors)

Image

in v15.3 no issues were present - both the defaultDomain and the domains from the domains array had always the correct defaultLocale. Also the next/links had the correct links without adding the extra "domain".

For even more information pls see "Additional context"

```bash
This happens on every environment - local, docker, codesandbox.

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:48:41 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T6041
  Available memory (MB): 24576
  Available CPU cores: 14
Binaries:
  Node: 22.22.0
  npm: 10.9.4
  Yarn: 4.12.0
  pnpm: N/A
Relevant Packages:
  next: 15.5.12
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.7.2
Next.js Config:
  output: N/A
```

Pages Router, Internationalization (i18n)

### Additional context

Our "default" domain supports **all** the listed locales.
Our other domains (e.g. "example.fr" has only "fr" and "fr-en", "example.nl" has "nl" and "nl-en"), but the same locales are also supported by the default domain.

in the version 15.3 the locale detection worked perfectly for all these cases.
On the "real" domains, listed in the "domains" array it worked and still works correctly, but the problem is with the "default" domain, that also supports the same locales.

v15.5
open the preview page in codesandbox
default path **"/"** detects the correct values

> locale from GSSP args: de
> defaultLocale from GSSP args: de
> locale from NextInternalRequestMeta: de
> defaultLocale from NextInternalRequestMeta: de
> router.locale: de
> router.defaultLocale: de

If you click on a NextLink (**CSR**) to **"/fr"** - that is listed in the "domains" array - the router still has the correct value for "router.defaultLocale: de", but the defaultLocale from the args is now wrong (should be "de")

> locale from GSSP args: fr
> defaultLocale from GSSP args: fr
> locale from NextInternalRequestMeta: fr
> defaultLocale from NextInternalRequestMeta: de
> router.locale: fr
> router.defaultLocale: de

but if you **reload** (**SSR**) the same page **"/fr"** - router.defaultLocale changes to fr and the defaultLocale from the args is also wrong (both should be "de")

> locale from GSSP args: fr
> defaultLocale from GSSP args: fr
> locale from NextInternalRequestMeta: fr
> defaultLocale from NextInternalRequestMeta: de
> router.locale: fr
> router.defaultLocale: fr

and a hydration error comes up in the console:

Image

so the next/link is adding the domain to the href now.

The locales, that are not listed in the domains, like "/it" - are being detected correctly. Also the "complex" locales for the listed domains - "fr-en" or "nl-en" are working absolute fine.

hammadxcm:
### What?

Fixes the `defaultLocale` resolution in `RouteModule.prepare()` when a `domains` array is configured in the i18n config and the request is served from the default domain (not a configured domain hostname).

### Why?

When visiting a locale path like `/fr` on the default domain (e.g., `localhost`), `detectDomainLocale()` matches a configured domain entry (e.g., `example.fr`) based on the detected locale — not hostname. The matched domain's `defaultLocale` was then incorrectly used as the request's `defaultLocale`, causing:

- Wrong `defaultLocale` in `getServerSideProps` / `getStaticProps` args
- Wrong `router.defaultLocale` on SSR (sent via `__NEXT_DATA__`)
- Hydration errors from `<Link>` components rendering domain-prefixed hrefs

This is a regression introduced between v15.3 and v15.4 when the Pages Router handler was refactored to use `RouteModule.prepare()`.

### How?

Removed `domainLocale?.defaultLocale` from the `defaultLocale` fallback chain in `RouteModule.prepare()`. The correct `defaultLocale` is already set in request metadata by `base-server.ts` using hostname-only domain matching, so the locale-matched `domainLocale` value is not needed here.

The `domainLocale` variable (with locale-based matching) is still used for `isLocaleDomain` — which is needed for domain-prefixed link generation.

**Before:**
```typescript
const defaultLocale =
  getRequestMeta(req, 'defaultLocale') ||
  domainLocale?.defaultLocale ||
  i18n?.defaultLocale
```

**After:**
```typescript
const defaultLocale =
  getRequestMeta(req, 'defaultLocale') || i18n?.defaultLocale
```

### Test plan

- [x] Added regression test: visits `/go/another` and `/do/another` on localhost (no `host` header) and verifies `defaultLocale` is `en-US` (main config) in both GSSP props and `router.defaultLocale`
- [x] Full i18n integration test suite passes (189/193, 4 skipped)
- [x] Existing domain locale tests still pass (link domain prefixing, correct default locale on configured domains)

Fixes #90131


hammadxcm:
Rebased onto the latest `canary` to resolve the merge conflict — the PR is now mergeable.

A couple of notes for reviewers:

- In the meantime the i18n test suite was migrated from `test/integration/i18n-support/test/shared.ts` to `test/e2e/i18n-support/shared.ts`. The regression test moved with it (no behavioral change) and still passes:
  `pnpm test-start-turbo test/e2e/i18n-support/i18n-support.test.ts -t "should not use domain defaultLocale when visiting domain locale on default host"` → ✅
- The fix itself is unchanged and the regression is still present on current `canary` (`route-module.ts`, the `domainLocale?.defaultLocale` fallback in the `defaultLocale` chain), so this is still applicable.

The commit is signed/verified.

PR #93154, opened 2026-04-23
fix(router): preserve multi-value search params in segment cache key

### LINKED ISSUE #93104: multi→single search-param unset does not commit RSC

https://github.com/dave-wwg/next-multi-value-searchparam-repro

### To Reproduce

I've generated a test suite which demonstrates this bug

```bash
pnpm install
pnpm dev            # starts on :3100
# in another shell:
pnpm exec playwright install chromium
pnpm exec playwright test
```

You can also re-produce manually by: 
1. loading `localhost:3100`
2. toggling each of the 2 filters on, incrementing the count by 1 each time
3. toggling one of the filters off, which does not decrement the count as expected

**Current**

When router.replace(url, { scroll: false }) changes a repeated search-param from multiple values (?k=A&k=B) to a single value (?k=B), the new RSC payload is fetched and the server renders with the correct searchParams, but the client tree never commits the update. The server-rendered JSX and the client component's server-passed props stay stale. Other transitions on the same param (empty→single, single→multi) commit correctly — only multi→single is broken. It also seems to work if the toggle is done a second time?

here are the logs pulled from my browser console. 

```
Server  [server] f.status = [] count = 0
toggles.tsx:20 [client] props.count = 0 props.values = [] url.values = []
installHook.js:1 [client] props.count = 0 props.values = [] url.values = []
host-console-events.js:1 [client] props.count = 0 props.values = [] url.values = []
host-console-events.js:1 [client] props.count = 0 props.values = [] url.values = []
host-console-events.js:1  Server  [server] f.status = ["A"] count = 1
host-console-events.js:1 [client] props.count = 1 props.values = ["A"] url.values = ["A"]
host-console-events.js:1 [client] props.count = 1 props.values = ["A"] url.values = ["A"]
host-console-events.js:1 [client] props.count = 1 props.values = ["A"] url.values = ["A"]
host-console-events.js:1 [client] props.count = 1 props.values = ["A"] url.values = ["A"]
host-console-events.js:1  Server  [server] f.status = ["A","B"] count = 2
host-console-events.js:1 [client] props.count = 2 props.values = ["A","B"] url.values = ["A","B"]
host-console-events.js:1 [client] props.count = 2 props.values = ["A","B"] url.values = ["A","B"]
host-console-events.js:1 [client] props.count = 2 props.values = ["A","B"] url.values = ["A","B"]
host-console-events.js:1 [client] props.count = 2 props.values = ["A","B"] url.values = ["A","B"]
host-console-events.js:1 [client] props.count = 2 props.values = ["A","B"] url.values = ["B"]
host-console-events.js:1 [client] props.count = 2 props.values = ["A","B"] url.values = ["B"]
host-console-events.js:1  Server  [server] f.status = ["B"] count = 1
```

they are missing the client logs which should say: 
```
[client] props.count = 1 props.values = ["B"] url.values = ["B"]
host-console-events.js:1 [client] props.count = 1 props.values = ["B"] url.values = ["B"]
```

**Expected**

The client tree always responds to data updates 

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:53:31 PST 2026; root:xnu-12377.81.4~5/RELEASE_ARM64_T8122
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: 10.29.2
Relevant Packages:
  eslint-config-next: N/A
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

React

### Additional context

full disclosure, this re-production was generated by Claude Code! But spun out of a real bug i found in my code. I took a look at the minimal re-pro and it seems like it should be working

NiteshCodes7:
Closes #93104

## What's happening

When navigating from `?k=A&k=B` (multi-value) to `?k=B` (single-value) via `router.replace()`, the client tree never commits the update — props stay stale even though the server renders correctly.

## Root cause

In `createSegmentFromRouteTree` inside `ppr-navigations.ts`, the client builds a segment cache key from the current `renderedSearch` string:

```ts
// Before (buggy)
const stringifiedQuery = JSON.stringify(
  Object.fromEntries(new URLSearchParams(renderedSearch))
)
```

`Object.fromEntries` on a `URLSearchParams` with repeated keys silently drops all but the last value. This means:

- `?k=A&k=B` → `{ k: "B" }` → `__PAGE__?{"k":"B"}`
- `?k=B` → `{ k: "B" }` → `__PAGE__?{"k":"B"}`

Both produce the **same segment key**, so `matchSegment` treats them as identical, the stale cache node is reused, and the client never re-renders with the new props.

This explains the asymmetric behavior reported in the issue:
- `empty → single` ✅ (keys are genuinely different)
- `single → multi` ✅ (keys are genuinely different)
- `multi → single` ❌ (keys wrongly collide due to `Object.fromEntries`)
- "works on second toggle" ✅ (by then `currentRenderedSearch` has caught up so the comparison path differs)

## Fix

Use `getAll()` to preserve repeated values, matching exactly what the server does in `addSearchParamsIfPageSegment`:

```ts
// After (fixed)
const urlSearchParams = new URLSearchParams(renderedSearch)
const queryObj: Record<string, string | string[]> = {}
for (const key of new Set(urlSearchParams.keys())) {
  const vals = urlSearchParams.getAll(key)
  queryObj[key] = vals.length === 1 ? vals[0] : vals
}
const stringifiedQuery = JSON.stringify(queryObj)
```

Now:
- `?k=A&k=B` → `{ k: ["A", "B"] }` → `__PAGE__?{"k":["A","B"]}`
- `?k=B` → `{ k: "B" }` → `__PAGE__?{"k":"B"}`

These are distinct keys, the cache node is correctly replaced, and the client commits the updated props.

## Testing

Reproduction repo from the issue: https://github.com/dave-wwg/next-multi-value-searchparam-repro

The existing playwright test suite in that repo passes after this fix (it was designed to fail when the bug is present).


NiteshCodes7:
Updated this PR to address the additional occurrences flagged by the bot.

I extracted the search param conversion into a shared helper that preserves repeated keys (e.g. `?k=A&k=B` -> `{ k: ["A", "B"] }`) instead of collapsing them via `Object.fromEntries(...)`.

The helper is now reused in the related cache key / segment generation paths to keep behavior consistent and avoid collisions caused by multi-value params.

gadkadosh:
I can confirm that this PR solves this problem for me as well and the implementation seems sound to me. 
Before finding this I made a minimal repo of my own. I'm posting it here in case that helps speed up the review and merge process:
https://github.com/gadkadosh/next-rsc-searchparams-repro

NiteshCodes7:
Hey team, Just a quick reminder — there's an open PR for our Next.js Vercel deployment that still needs a review. Would appreciate it if someone could take a look when they get a chance. Thanks!

PR #93164, opened 2026-04-23
[ci]: switch to GitHub runners

ztanner:
## Summary

Switch CI from self-hosted runners to GitHub-hosted runners and fix the follow-up issues that showed up once the hosted jobs were exercised.

## Changes

- Move Linux workflows onto `ubuntu-latest-16-core-oss`, Windows workflows onto `windows-latest-8-core-oss`, and the native mac release lane onto `macos-15-intel`.
- Update the reusable build workflow to work in hosted environments by relying on `runner.os` instead of label string matching, and by removing the hardcoded `/home/runner` `fnm` path.
- Make Turbo/sccache configuration explicit for hosted runners:
  - unify on `vtest314-next-adapter-e2e-tests`
  - pass `TURBO_TOKEN` via workflow env/secrets instead of assuming runner-level env
  - switch Turbo cache mode to `local:rw,remote:rw` so jobs still have local cache behavior when remote cache is unavailable
- Update the `sccache` action defaults/docs to use the hosted-runner secret setup (`TURBO_TOKEN`) instead of the old self-hosted token wiring.
- Make Datadog reporting fail open so missing `DATA_DOG_API_KEY` does not block CI.
- Fix the CLI reserved-port test so it asserts Next’s own `-p 1` validation instead of failing earlier on hosted Linux due to privileged-port binding.
- Replace the live `musl.cc` dependency in `native-builder.Dockerfile` with musl sysroots imported from GHCR-hosted `rust-musl-cross` images, keeping the existing `/opt/*-cross` layout used by the Linux native build scripts.
- Fix the macOS native build workaround s so it applies to all *-apple-darwin targets based on the Cargo target OS rather than the host architecture, which avoids napi-build injecting the unsupported -Wl linker arg when @next/swc is linked with rust-lld on hosted mac runners.

Test Plan:
- This PR's CI
- This [build_and_deploy](https://github.com/vercel/next.js/actions/runs/24865715983/job/72801225722) job running on every arch

### DIFF
--- a/scripts/native-builder.Dockerfile
+++ b/scripts/native-builder.Dockerfile
@@ -10,10 +10,13 @@
 #   - Ubuntu 20.04 (glibc 2.31 — broad compatibility baseline)
 #   - Clang/LLD for all compilation and linking via --target
 #   - GNU cross-sysroots via crossbuild-essential (Ubuntu multiarch)
-#   - musl sysroots from musl.cc (headers + libs only; clang/lld do the work)
+#   - musl sysroots from GHCR-hosted rust-musl-cross images
 #   - Node.js 20 (glibc-linked, used as build tool for all targets)
 #   - Rust nightly toolchain (pinned to match rust-toolchain.toml)
 #   - @napi-rs/cli for building native Node.js addons
+
+FROM ghcr.io/rust-cross/rust-musl-cross:x86_64-musl AS musl_x86_64
+FROM ghcr.io/rust-cross/rust-musl-cross:aarch64-musl AS musl_aarch64
 
 FROM ubuntu:20.04 AS builder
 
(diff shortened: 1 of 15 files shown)


feedthejim (approved):

bgw (approved):

scripts/native-builder.Dockerfile
```diff
@@ -10,11 +10,14 @@
...
 #   - Clang/LLD for all compilation and linking via --target
 #   - GNU cross-sysroots via crossbuild-essential (Ubuntu multiarch)
-#   - musl sysroots from musl.cc (headers + libs only; clang/lld do the work)
+#   - musl sysroots from GHCR-hosted rust-musl-cross images
 #   - Node.js 20 (glibc-linked, used as build tool for all targets)
 #   - Rust nightly toolchain (pinned to match rust-toolchain.toml)
 #   - @napi-rs/cli for building native Node.js addons
 
+FROM ghcr.io/rust-cross/rust-musl-cross:x86_64-musl AS musl_x86_64
+FROM ghcr.io/rust-cross/rust-musl-cross:aarch64-musl AS musl_aarch64
```
bgw, line 19: We should pin these to specific hashes

PR #92636, opened 2026-04-10
fix(use-cache): remove awaiting revalidation

awo00:
### What?
fixes #74882

The `use cache` directive does not revalidate in the background while serving (not expired) stale data.

For the static pages, the whole page is blocked during revalidation, and fresh data is fetched served. While I don't agree with this, as it is just a duplication of the expire behaviour, this is documented as intended in the comments in the code.

For the dynamic pages, when the data is flagged for revalidation, the stale data is blocked while fresh data is fetched. Once the fresh data is fetched, the function returns the stale data. On refreshing, the fresh data is served. This is the worst of both worlds, as we are still waiting for fresh data, and then seeing the stale data.

In both cases there is no background revalidation of data, leading to slow, unresponsive pages.

I have an example app showing this behaviour here: https://github.com/awo00/nextjs-use-cache-revalidate

### Why?

In the `use-cache-wrapper`, the `generateCacheEntry()` function is awaited during revalidation which is blocking the stale data being returned.

### How?

Removing the await allows the stale data to be returned faster.

PR #74164, opened 2024-12-20
Use provided waitUntil for pending revalidates

ijjk:
Currently we are using `pendingWaitUntil` on `renderOpts` to handle pending revalidations (fetch updates and revalidate tag calls). This uses our old strategy of `waitUntil` inside of `sendResponse` which just keeps the stream open until the promise resolves. This isn't ideal if we have a proper waitUntil strategy provided so this updates to ensure we use that instead if available. 

Also adds some debug logs so we can track when this pending revalidates promise is resolved.

PR #58744, opened 2023-11-21
fix waitUntil handling in pipeToNodeResponse

ztanner:
### What?
Calls to `revalidateTag` when using the data cache would sometimes be ignored. This would result in stale data being displayed even after manually triggering a revalidation. 

### Why?
Revalidation calls are pushed onto a `pendingRevalidates` array and, in the case of a route handler, are awaited [here](https://github.com/vercel/next.js/blob/8b11264ea95c9ed2f7dfd1e8d66a36b57ebc0da1/packages/next/src/server/send-response.ts#L53-L55).  

Previously this took place as part of the former `pipeReadable` implementation ([ref](https://github.com/vercel/next.js/blob/57bb52d37dd43fd669c5c07296daba19b6a12917/packages/next/src/server/send-response.ts#L49)) which awaited the promise before calling res.end ([ref](https://github.com/vercel/next.js/blob/57bb52d37dd43fd669c5c07296daba19b6a12917/packages/next/src/server/pipe-readable.ts#L99-L105)). There was a subtle change in behavior here in a recent refactor that seems to cause a race between awaiting the promise and ending the response, which results in this unpredictable behavior.

### How?
This applies similar logic that used to be in the `pipeReadable` function to the new `pipeToNodeResponse` function. Namely, we plumb `waitUntil` into the writeable stream close handler and await it before calling `res.end()`.

This is a very difficult edge case to test, and as such resorted to rigorous manual testing when deployed to an environment that uses the data cache. 

Fixes #52962
Fixes #57632

PR #93189, opened 2026-04-24
Fix route handler SWR blocking on Node runtime

### LINKED ISSUE #93146: use-cache: `await ignoredStream.cancel()` blocks SWR on Node runtime (waits for source close)
### Summary

On the Node runtime, `"use cache"` stale-while-revalidate never serves stale — it always blocks the user response on the full background regeneration. Root cause: `await ignoredStream.cancel()` in `use-cache-wrapper.ts`, combined with Node's Web Streams tee semantics where cancelling one branch waits until the **source stream closes**.

Offending line (canary): [`packages/next/src/server/use-cache/use-cache-wrapper.ts#L2668`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/use-cache/use-cache-wrapper.ts#L2668)

### To Reproduce

This is a platform-level behavior of Node's Web Streams, not something that needs a full Next.js reproduction to verify. Pure-Node repro (under 30 lines): https://gist.github.com/haydenshively/e5b7fb69088bac8e4d477ec967c7c891

```
$ node tee-cancel-repro.js
b read at 2003 ms, done=false
a.cancel() resolved at 6004 ms   <-- waits for source CLOSE, not cancel
```

In the cache wrapper's SWR branch, `ignoredStream` is one half of a `tee()`d stream whose source is `renderToReadableStream(resultPromise, ...)`. The *other* branch (`savedStream`) is actively being read by `collectResult` to populate `pendingCacheEntry`. Because of that active reader, the source stream does not close until the regen `fn` has fully resolved and all RSC chunks have been emitted. And because of Node's tee-cancel semantics, `await ignoredStream.cancel()` therefore waits that full duration before resuming the wrapper.

That suspends the wrapper body, which delays the `return createFromReadableStream(stream /* stale */)` at the bottom of `cache()`, which is exactly what the user's request is `await`ing. Net effect: SWR becomes "block until fresh."

To observe in a real app:

1. App Router route handler that returns JSON via an async `"use cache"` function whose body takes multiple seconds (e.g. several RPC calls).
2. `cacheLife({ revalidate: 60, expire: 86400 })` on the cached function. Use the default Node runtime.
3. Populate the cache by hitting the endpoint once. Wait > 60s. Hit again.

Expected: the second hit returns the stale value in < 100ms and triggers a background regen.
Observed: the second hit blocks for the full regen duration and then returns the freshly-computed value.

**Current:** `await ignoredStream.cancel()` at the end of the SWR branch synchronously waits for the regen source to close, which is equivalent to waiting for the regen `fn` to fully resolve. The user's response is therefore blocked on the regen for its entire duration.

**Expected:** In the SWR branch the user should receive the stale stream immediately and the regen should complete out-of-band (that's what the `"use cache"` contract and its documentation imply, and what the Edge runtime wrapper at `edge-route-module-wrapper.ts` does by handing `pendingWaitUntil` directly to `evt.waitUntil` without awaiting it).

Proposed fix (locally verified against 16.1.7):

```diff
-                    await ignoredStream.cancel()
+                    void ignoredStream.cancel().catch(() => {})
```

Nothing downstream reads from `ignoredStream`, so there is no observer waiting on the cancel's resolution — the `await` on it serves no purpose and only couples the caller to the tee-cancel's blocking semantic.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0
Binaries:
  Node: 24.14.1
  npm: 11.11.0
  pnpm: 10.32.1
Relevant Packages:
  next: 16.1.7 (also reproduces on canary — offending line unchanged)
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: standalone
```

Canary verified by inspection only (offending line still present at canary `use-cache-wrapper.ts:2668`). Happy to test against a canary build if a maintainer prefers.

Use Cache, Route Handlers, Runtime

next dev (local), next start (local), Vercel (Deployed), Other (Deployed)

### Additional context

Affects both `next dev` and deployed Node-runtime route handlers. The Edge runtime path is unaffected because `edge-route-module-wrapper.ts` hands `pendingWaitUntil` directly to `evt.waitUntil(...)` instead of awaiting it before `res.end()` (via `pipe-readable.ts`).

Also worth noting that because `pendingRevalidateWrites` is awaited in `pipe-readable.ts`'s writer close handler before `res.end()` on the Node runtime, custom `CacheHandler.set` implementations that correctly return an already-resolved promise (so the lambda/function isn't gated on the persistence write) still see the response blocked — because the block is inside the wrapper itself, before `cacheHandler.set`'s return value is even relevant. This one-line change is the only thing that unblocks Node-runtime SWR.

unstubbable:
In a Node-runtime app route handler, `'use cache'` and fetch-cache stale-while-revalidate block the client response on the full background regeneration. A second request after the cache has gone stale does not return the stale value promptly. It waits the entire regen duration and then returns the freshly regenerated value, as described in issue
 #93146. The symptom is visible both on self-hosted Node and on Vercel Node. Edge route handlers were unaffected because they route through `edge-route-module-wrapper.ts`, which hands `pendingWaitUntil` directly to `evt.waitUntil` and never awaits it before the response completes.

The root cause is independent of the `await ignoredStream.cancel()` change proposed in that issue, and also independent of #92636, which removed a separate blocking await inside `use-cache-wrapper.ts`. Both of those sit inside the `'use cache'` wrapper and affect how the stale entry is returned to the caller. The blocking that remains is in the transport layer and applies equally to any revalidation pushed into `pendingRevalidates` or `pendingRevalidateWrites`, not just `'use cache'`.

#74164 migrated pending revalidate handling from the "keep the response stream open until the promise resolves" pattern introduced in #55978 and reinforced in #58744 to a conditional hand-off. If a platform `waitUntil` is available, revalidations run out of band via `ctx.waitUntil`. Otherwise `pipe-readable.ts` keeps `res.end` deferred until they settle, so minimal-mode deployments stay alive long enough for writes to persist.

That migration landed correctly for app pages. `app-render.tsx` only assigns `options.waitUntil` in the `else` branch, so once `renderOpts.waitUntil` is present the `pipe-readable.ts` path receives nothing and does not block. The matching code for route handlers was added to `base-server.ts` in the same PR. It declared `let pendingWaitUntil = context.renderOpts.pendingWaitUntil`, cleared it when handing off to `ctx.waitUntil`, and then passed `context.renderOpts.pendingWaitUntil` into `sendResponse`. That is the unmutated property, not the cleared local. The local variable was never read. The hand-off therefore never displaced the pipe-readable await. The same promise was awaited twice. `ctx.waitUntil` registered it redundantly and `pipe-readable.ts` still held `res.end` open for the full revalidation.

#80189 copied the block verbatim into `packages/next/src/build/templates/app-route.ts`, carrying the bug forward when response handling moved into the route template. Fixing it is a one-line change. Pass the local `pendingWaitUntil` (which is `undefined` once handed off) into `sendResponse`, so Node route handlers match both app pages and edge route handlers and the original intent of #74164 is restored.

The existing `stale-cache-serving/route-handler` test in `app-static.test.ts` did not catch this. It only measured time-to-first-byte against the route start time, and first-byte was already fast pre-fix. Chunks are written before `pipe-readable.ts`'s close handler awaits `waitUntilForEnd`. Only the terminating chunk was delayed. The test is updated to also assert total response time against the route start. Pre-fix that new assertion fails at ~3000ms for the Node route handler variant while the page and edge variants continue to pass, which is exactly the scope of the regression. A new e2e test in `use-cache-swr` exercises the symmetric `'use cache'` path through a route handler fixture.

fixes #93146
closes #93177 (incorrect fix)
closes #93188 (incorrect fix)

### DIFF
--- a/packages/next/src/build/templates/app-route.ts
+++ b/packages/next/src/build/templates/app-route.ts
@@ -408,7 +408,7 @@ export async function handler(
               nodeNextReq,
               nodeNextRes,
               response,
-              context.renderOpts.pendingWaitUntil
+              pendingWaitUntil
             )
             return null
           }
--- a/test/e2e/app-dir/app-static/app-static.test.ts
+++ b/test/e2e/app-dir/app-static/app-static.test.ts
@@ -3342,9 +3342,6 @@ describe('app-dir static/dynamic handling', () => {
     // Prime the cache.
     let res = await next.fetch(path)
     expect(res.status).toBe(200)
-
-    // Consume the cache, the revalidations are completed on the end of the
-    // stream so we need to wait for that to complete.
     await res.text()
 
     for (let i = 0; i < 6; i++) {
@@ -3374,6 +3371,7 @@ describe('app-dir static/dynamic handling', () => {
           )
         }
       }
+      const finishedAt = Date.now()
 
       const startedResponding = +data.start
       if (Number.isNaN(startedResponding)) {
@@ -3387,12 +3385,17 @@ describe('app-dir static/dynamic handling', () => {
         )
       }
 
-      // We just want to ensure the response isn't blocked on revalidating the fetch.
-      // So we use the start time when route started processing not when we
-      // send off the response because that includes cold boots of the infra.
+      // The response must not be blocked on the 3s background revalidation:
+      // neither the first byte (TTFB) nor the terminating chunk (res.end).
+      // Using the route-start time excludes cold-boot/infra latency.
       if (startedStreaming - startedResponding >= 3000) {
         throw new Error(
-          `Response #${i} took too long to complete: ${startedStreaming - startedResponding}ms`
+          `Response #${i} first byte took too long: ${startedStreaming - startedResponding}ms`
+        )
+      }
+      if (finishedAt - startedResponding >= 3000) {
+        throw new Error(
+          `Response #${i} took too long to complete: ${finishedAt - startedResponding}ms`
         )
       }
     }
--- a/test/e2e/app-dir/use-cache-swr/app/delayed-route/route.ts
+++ b/test/e2e/app-dir/use-cache-swr/app/delayed-route/route.ts
@@ -1,0 +1,19 @@
+import { cacheLife } from 'next/cache'
+import { setTimeout } from 'timers/promises'
+
+async function getCachedData() {
+  'use cache'
+
+  cacheLife('seconds')
+
+  await setTimeout(1000)
+
+  return new Date().toISOString()
+}
+
+export async function GET() {
+  const cached = await getCachedData()
+  const dynamic = new Date().toISOString()
+
+  return Response.json({ cached, dynamic })
+}
--- a/test/e2e/app-dir/use-cache-swr/use-cache-swr.test.ts
+++ b/test/e2e/app-dir/use-cache-swr/use-cache-swr.test.ts
@@ -103,6 +103,49 @@ describe('use-cache-swr', () => {
     expect(duration3).toBeLessThan(1000)
   })
 
+  it('should serve stale data without blocking on the background regeneration (route handler)', async () => {
+    // Fetch 1: cold cache. The cached function blocks for ~1s.
+    const res1 = await next.fetch('/delayed-route')
+    const { cached: cached1, dynamic: dynamic1 } = await res1.json()
+    expect(cached1).toBeDateString()
+    expect(dynamic1).toBeDateString()
+
+    // Wait past the 1s revalidate window (cacheLife('seconds')).
+    await new Promise((resolve) => setTimeout(resolve, 1200))
+
+    outputIndex = next.cliOutput.length
+
+    // Fetch 2: stale hit. Should return the stale entry immediately and kick
+    // off the regeneration in the background, rather than blocking on the 1s
+    // delay in the cached function.
+    const start2 = Date.now()
+    const res2 = await next.fetch('/delayed-route')
+    const { cached: cached2, dynamic: dynamic2 } = await res2.json()
+    const duration2 = Date.now() - start2
+
+    expect(cached2).toBe(cached1)
+    expect(dynamic2).not.toBe(dynamic1)
+    expect(duration2).toBeLessThan(1000)
+
+    // Wait for the background regen to finish writing the fresh entry.
+    await retry(() => {
+      expect(next.cliOutput.slice(outputIndex)).toMatch(
+        /PersistentCacheHandler::set/
+      )
+    })
+
+    // Fetch 3: should serve the pre-warmed fresh entry from the background
+    // regen, not a new stale value.
+    const start3 = Date.now()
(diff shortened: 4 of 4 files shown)


haydenshively:
Thanks for the analysis. One observation that may matter for scope:

With this PR alone applied to a Node-runtime route handler whose custom `cacheHandlers.remote` returns `Promise.resolve()` from `set` (so `pendingRevalidateWrites` is already resolved), the response was still blocked on the full regen. The block is at `await ignoredStream.cancel()` in `use-cache-wrapper.ts` — Node's tee-branch `cancel()` waits for the **source to close**, not for the cancel itself:

```js
const src = new ReadableStream({
  pull(controller) {
    setTimeout(() => controller.enqueue(new Uint8Array([1])), 2000);  // first chunk
    setTimeout(() => controller.close(), 6000);                       // source close
  }
});
const [a, b] = src.tee();
b.getReader().read();              // active reader on the other branch
a.cancel().then(() => /* fires at ~6000ms, not 0 or 2000 */);
```

In the SWR branch, `savedStream` is being read by `collectResult`, so the source keeps producing until `fn` fully resolves. `await ignoredStream.cancel()` therefore blocks the wrapper before it returns the stale stream — independent of the `pendingWaitUntil` fix in this PR. Replacing that line with `void ignoredStream.cancel().catch(() => {})` on top of this PR restores the fast-stale behaviour.

Happy to open a follow-up if it's useful!

unstubbable:
> Thanks for the analysis. One observation that may matter for scope:
> 
> With this PR alone applied to a Node-runtime route handler whose custom `cacheHandlers.remote` returns `Promise.resolve()` from `set` (so `pendingRevalidateWrites` is already resolved), the response was still blocked on the full regen. The block is at `await ignoredStream.cancel()` in `use-cache-wrapper.ts` — Node's tee-branch `cancel()` waits for the **source to close**, not for the cancel itself:
> 
> ```js
> const src = new ReadableStream({
>   pull(controller) {
>     setTimeout(() => controller.enqueue(new Uint8Array([1])), 2000);  // first chunk
>     setTimeout(() => controller.close(), 6000);                       // source close
>   }
> });
> const [a, b] = src.tee();
> b.getReader().read();              // active reader on the other branch
> a.cancel().then(() => /* fires at ~6000ms, not 0 or 2000 */);
> ```
> 
> In the SWR branch, `savedStream` is being read by `collectResult`, so the source keeps producing until `fn` fully resolves. `await ignoredStream.cancel()` therefore blocks the wrapper before it returns the stale stream — independent of the `pendingWaitUntil` fix in this PR. Replacing that line with `void ignoredStream.cancel().catch(() => {})` on top of this PR restores the fast-stale behaviour.
> 
> Happy to open a follow-up if it's useful!

This is a flawed analysis and not useful.

ztanner (approved):

PR #93191, opened 2026-04-24
Turbopack: move get_inner_asset into ecmascript module

mischnic:
`resolve_origin.get_inner_asset` showed up in Luke's statistics at a very high place.
But in reality, there are only 1-3 modules per page that have any inner assets (the templates, and the edge wrapper modules).
`resolve_origin.get_inner_asset` is in the hottest hot path (resolving references), so let's simply move this logic into the EcmascriptModuleAsset itself.

Behavioral differences to before:
- only works with static ESM `import`s, dynamic `import()`, `require`, `require.resolve`. But these are the only places that currently use it anyway.

```
0d9738112f mischnic/remove-get-inner-asset
pnpm next build --experimental-build-mode=compile  353.40s user 52.96s system 748% cpu 54.265 total
pnpm next build --experimental-build-mode=compile  349.42s user 53.84s system 744% cpu 54.198 total
pnpm next build --experimental-build-mode=compile  352.42s user 54.46s system 741% cpu 54.874 total

3cb77c844b canary
pnpm next build --experimental-build-mode=compile  367.91s user 58.06s system 726% cpu 58.672 total
pnpm next build --experimental-build-mode=compile  362.97s user 57.68s system 746% cpu 56.379 total
pnpm next build --experimental-build-mode=compile  364.59s user 57.01s system 724% cpu 58.167 total
```

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/references/mod.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/mod.rs
@@ -1799,6 +1799,7 @@ async fn handle_dynamic_import<G: Fn(Vec<Effect>) + Send + Sync>(
         handler,
         origin,
         source,
+        &state.inner_assets,
         ignore_dynamic_requests,
         analysis,
         error_mode,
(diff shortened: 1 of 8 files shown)


turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -2203,13 +2224,24 @@ where
                         return Ok(());
                     }
                 }
+
+                let resolve_override = if let Some(inner_assets) = &state.inner_assets
```
sokra, line 2202: Do we also want to add it to `require.resolve` and `import()` or is that unsupported?

sokra (approved):

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
mischnic: will add it now. import() was required. and I guess I'll add `require.resolve` for consistency at least.

PR #93200, opened 2026-04-24
[turbo-tasks-backend] Single-item optimization for lost follower operations

sokra:
## What?

Adds performance optimizations for the common case of single-item lost follower operations in the turbo-tasks-backend aggregation update system, along with improved tracing capabilities for debugging.

## Why?

When profiling turbopack operations, we observed that many aggregation update jobs involve only a single upper losing a single follower. The previous implementation always used vectors and iteration even for these single-item cases, adding unnecessary allocation overhead.

Additionally, the existing `trace_aggregation_update` feature was incomplete (compilation errors) and lacked visibility into task data and operation statistics.

### Single-item optimization
- Added a dedicated `InnerOfUpperLostFollower` job variant and `inner_of_upper_lost_follower` function to handle the single upper + single follower case without vector allocation
- Optimized the plural variants (`InnerOfUppersLostFollowers`, `InnerOfUppersLostFollower`, `InnerOfUpperLostFollowers`) to delegate to the singular function when they contain only one item

### Tracing improvements  
- Added `trace_aggregation_update_stats` feature flag that enables counters for all aggregation update job types (new_followers, lost_followers, balance_edge, optimize_task, etc.) recorded in trace spans
- Fixed `trace_aggregation_update` feature compilation errors by using `task.get_task_description()` instead of `ctx.get_task_description(task_id)`
- Gated the `TaskDataCategory` used in aggregation update `ctx.task(...)` calls behind a const that resolves to `Meta` by default and `All` when tracing is enabled, so traces can see full task data without affecting default performance


turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2350,12 +2350,19 @@ impl<B: BackingStorage> TurboTasksBackendInner<B> {
         span.record("immutable", is_immutable || is_now_immutable);
 
         if !queue.is_empty() || !old_edges.is_empty() {
-            #[cfg(feature = "trace_task_completion")]
-            let _span = tracing::trace_span!("remove old edges and prepare new children").entered();
+            #[cfg(any(
+                feature = "trace_task_completion",
+                feature = "trace_aggregation_update_stats"
```
lukesandberg, line 2355: does this need to be an independent feature, it doesn't seem too expensive to always compute this when the tracing features are enabled?

lukesandberg (approved):

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
sokra: Not sure. It's in the hot path, so if it shouldn't be a feature we should probably measure it first.

PR #93203, opened 2026-04-24
Turbopack: unify ModuleGraph constructors

mischnic:
We have options and variable length lists. Why do we need a separate function for every combination?

### DIFF
--- a/turbopack/crates/turbopack-core/src/module_graph/mod.rs
+++ b/turbopack/crates/turbopack-core/src/module_graph/mod.rs
@@ -698,56 +698,21 @@ pub struct ModuleGraph {
 
 #[turbo_tasks::value_impl]
 impl ModuleGraph {
+    /// Analyze the module graph and potentially remove unused references (by determining the used
+    /// exports and removing unused imports).
     #[turbo_tasks::function(operation)]
-    pub async fn from_single_graph(graph: OperationVc<SingleModuleGraph>) -> Result<Vc<Self>> {
-        let graph = Self::create(vec![graph], None)
-            .read_strongly_consistent()
-            .await?;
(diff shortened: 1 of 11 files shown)


turbopack/crates/turbopack-core/src/module_graph/mod.rs
```diff
@@ -698,56 +698,21 @@ pub struct ModuleGraph {
...
-
-    /// Analyze the module graph and remove unused references (by determining the used exports and
-    /// removing unused imports).
-    ///
-    /// In particular, this removes ModuleReference-s that list only unused exports in the
-    /// `import_usage()`
-    #[turbo_tasks::function(operation)]
-    pub async fn from_graphs_without_unused_references(
+    pub async fn from_graphs(
         graphs: Vec<OperationVc<SingleModuleGraph>>,
```
lukesandberg, line 705: nit: `SmallVec<[OperationVc<SingleModuleGraph>>;2]>`

lukesandberg (approved):

turbopack/crates/turbopack-core/src/module_graph/mod.rs
mischnic:
```
the trait bound `SmallVec<[OperationVc<SingleModuleGraph>; 2]>: TaskInput` is not satisfied
the trait `turbo_tasks::TaskInput` is not implemented for `smallvec::SmallVec<[turbo_tasks::OperationVc<module_graph::SingleModuleGraph>; 2]>`
```

lukesandberg: it should be possible to derive TaskInput for smallvec, but not a big deal

PR #93207, opened 2026-04-24
[ci] Upload preview builds directly to Vercel Blob

eps1lon:
This allows us to drop granting Actions read permissions to vercel-packages. We just upload directly from GitHub to Vercel blob.

We're doing this from a separate workflow that runs after build_and_deploy. [`workflow_run` runs on the default branch](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_run) which prevents cache poisoning. You can still upload bad builds to Vercel blob but you can't also control the commit these get uploaded to.

The `PREVIEW_BUILDS_BLOB_READ_WRITE_TOKEN` secret is only available in a protected GitHub deploy environment that is only allowed on `canary` branches.

### DIFF
--- a/.github/workflows/upload_preview_tarballs.yml
+++ b/.github/workflows/upload_preview_tarballs.yml
@@ -1,0 +1,71 @@
+# This workflow uploads preview tarballs to Vercel Blob after build-and-deploy
+# completes. It uses workflow_run so it always executes the DEFAULT BRANCH
+# version of this file -- an attacker who modifies this file on a feature branch
+# cannot change the code that touches the blob write token.
+name: upload-preview-tarballs
+
+on:
+  workflow_run:
+    workflows: [ "build-and-deploy" ]
+    types: [ completed ]
+
+env:
+  NODE_LTS_VERSION: 20
+
+permissions:
+  actions: read
+
+jobs:
+  upload:
+    name: Upload preview tarballs to Blob
+    runs-on: ubuntu-latest
+    if: >-
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.event != 'workflow_dispatch'
+    environment: preview-builds
+    steps:
+      - name: Setup node
+        uses: actions/setup-node@v6
+        with:
+          node-version: ${{ env.NODE_LTS_VERSION }}
+          check-latest: true
+
+      # Checkout from the default branch (canary) -- workflow_run always uses
+      # the default branch's version of the workflow file and this checkout
+      # matches that, ensuring the upload script is trusted.
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          fetch-depth: 1
+
+      - name: Enable corepack
+        run: corepack enable
+
+      - name: Setup pnpm
+        run: corepack prepare
+
+      - name: Cache dependencies
+        uses: actions/cache@v3
+        with:
+          path: ~/.pnpm-store
+          key: ${{ runner.os }}-${{ runner.arch }}-pnpm-v2-${{
+            hashFiles('**/pnpm-lock.yaml') }}
+          # Do not use restore-keys since it leads to indefinite growth of the cache.
+
+      - name: Install node_modules
+        run: pnpm install --frozen-lockfile
+
+      - name: Download preview-tarballs artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: preview-tarballs
+          path: ${{ runner.temp }}/preview-tarballs
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+
+      - name: Upload tarballs to Vercel Blob
+        run: node scripts/upload-preview-tarballs.js "${{
+          github.event.workflow_run.head_sha }}" "${{ runner.temp
+          }}/preview-tarballs"
+        env:
+          BLOB_READ_WRITE_TOKEN: ${{ secrets.PREVIEW_BUILDS_BLOB_READ_WRITE_TOKEN }}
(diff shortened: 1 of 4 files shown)


scripts/upload-preview-tarballs.js
```diff
@@ -0,0 +1,47 @@
...
+    const stat = await fs.stat(dir)
+    if (!stat.isDirectory()) continue
+
+    const files = await fs.readdir(dir)
+    const tgzFile = files.find((f) => f.endsWith('.tgz'))
+    if (!tgzFile) continue
+
+    const blobPathname = `next/commits/${githubHeadSha}/${packageName}.tgz`
+
+    // Check-before-write: prevent overwriting existing blobs (cache poisoning mitigation)
```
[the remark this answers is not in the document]
eps1lon: Leftover comment. We do want to allow overwrite from retry. Cache poisoning is prevented due to workflow_run not being forgeable.

.github/workflows/upload_preview_tarballs.yml
```diff
@@ -0,0 +1,51 @@
...
+  NODE_LTS_VERSION: 20
+
+jobs:
+  upload:
+    name: Upload preview tarballs to Blob
+    runs-on: ubuntu-latest
+    if: >-
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.event != 'workflow_dispatch'
+    environment: preview-tarballs
```
ztanner: was this meant to be `preview-builds`? (That's what I see as a valid environment in settings)

eps1lon: yes
```suggestion
    environment: preview-builds
```

.github/workflows/upload_preview_tarballs.yml
```diff
@@ -0,0 +1,51 @@
...
+      # Checkout from the default branch (canary) -- workflow_run always uses
+      # the default branch's version of the workflow file and this checkout
+      # matches that, ensuring the upload script is trusted.
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          fetch-depth: 1
+
+      - name: Install node_modules
+        run: pnpm install --frozen-lockfile
```
ztanner, line 55: I'm not sure that `setup-node` preinstalls pnpm. Other workflows have been doing `corepack enable` before install in case this doesn't work

.github/workflows/upload_preview_tarballs.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Download preview-tarballs artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: preview-tarballs
+          path: ${{ runner.temp }}/preview-tarballs
+          run-id: ${{ github.event.workflow_run.id }}
```
ztanner, line 63: I think `run-id` is only used when a GitHub token is supplied because downloading from another workflow run is an authenticated lookup. I suspect maybe just passing `github-token` and using the default read token would be sufficient

ztanner:
(Looks like we'll need to rebase on `canary`, `build_and_deploy` is still referencing the self-hosted CI workflow)

.github/workflows/upload_preview_tarballs.yml
eps1lon: Passing the github-token makes sense. That's even in the docs: https://github.com/actions/download-artifact/tree/3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c/#inputs

I'm not following the point on the `run-id`. I see no indication in the docs that `run-id` would not be used. If we omit `run-id`, `github.run_id` which I would assume is not the run of the triggering workflow but the current workflow run which doesn't have the artifacts.

ztanner: I meant that `run-id` only takes effect for the cross-run lookup once `github-token` is supplied.

the `run-id` input says "If github-token is specified, this is the run that artifacts will be downloaded from." So without `github-token`, the action still searches the current workflow run.

.github/workflows/upload_preview_tarballs.yml
eps1lon: Going through the same steps as `release-next-rspack.yml` now

.github/workflows/upload_preview_tarballs.yml
eps1lon: Makes sense. So this should work now?

.github/workflows/upload_preview_tarballs.yml
```diff
@@ -0,0 +1,69 @@
...
+on:
+  workflow_run:
+    workflows: [ "build-and-deploy" ]
+    types: [ completed ]
+
+env:
+  NODE_LTS_VERSION: 20
+
+permissions:
+  actions: read
```
ztanner, line 16: I think we need this for the `actions/checkout` step

```suggestion
permissions:
  actions: read
  contents: read
```

.github/workflows/upload_preview_tarballs.yml
```diff
@@ -0,0 +1,69 @@
+# This workflow uploads preview tarballs to Vercel Blob after build-and-deploy
+# completes. It uses workflow_run so it always executes the DEFAULT BRANCH
+# version of this file -- an attacker who modifies this file on a feature branch
+# cannot change the code that touches the blob write token.
+name: upload-preview-tarballs
+
+on:
+  workflow_run:
+    workflows: [ "build-and-deploy" ]
```
ztanner:
```suggestion
    workflows: ['build-and-deploy']
```

.github/workflows/upload_preview_tarballs.yml
```diff
@@ -0,0 +1,69 @@
...
+    types: [ completed ]
```
ztanner:
```suggestion
    types: [completed]
```

ztanner (approved):

PR #93209, opened 2026-04-24
[next/image] Only fire `onError` once

eps1lon:
...and get rid of excessive retries during Client render due to setting `img.src` again when the ref was invalidated.

We probably want to avoid the retry entirely in favor of dispatching a synthetic event but it's not clear yet if that's possible. 

We now use a Layout Effect to handle loading and error retry instead of the ref phase.


lubieowoce (approved):

packages/next/src/client/image-component.tsx
```diff
@@ -207,18 +216,24 @@ const ImageElement = forwardRef<HTMLImageElement | null, ImageElementProps>(
     },
     forwardedRef
   ) => {
-    const ownRef = useCallback(
-      (img: ImgElementWithDataProp | null) => {
-        if (!img) {
-          return
-        }
+    const [img, setImg] = useState<HTMLImageElement | null>(null)
```
eps1lon: From offline discussion: This is put into state to future proof this code in case  the `` gets refactored into another Component that breaks any Effect -> Ref coupling. Under normal circumstances we can put `img` into a ref object that we read from in the Layout Effect (without dependencies). That would save us a cascading update due to `setState` during the ref phase.

I've just been through to many bugs caused by refactors breaking existing Effect -> Ref coupling. But if the current impl causes noticeable performance impact, @lubieowoce was right to complain about my factoring 😄

eps1lon: The cascading update is enough reason to couple the Effect to the ref: https://github.com/vercel/next.js/pull/93209/commits/a92d26e239e899547ec40f7aa5b8679d5979e675

many321 (approved):

PR #76207, opened 2025-02-19
Propagate expire time to `cache-control` header and prerender manifest

unstubbable:
For the `stale-while-revalidate` value in the `cache-control` response header, Next.js currently uses the configured
[`expireTime`](https://nextjs.org/docs/app/api-reference/config/next-config-js/expireTime), which applies to all routes, and defaults to one year.

With the introduction of `"use cache"` and granular [cache lifetimes](https://nextjs.org/docs/app/api-reference/functions/cacheLife), users can now set expire times per route — based on the minimum expire time of all cached functions used by that route.

This PR updates the `stale-while-revalidate` value to reflect the route's cache lifetime, using the difference between the minimum expire time and the minimum revalidate time. If no explicit expire time is defined in cache profiles, the globally configured `expireTime` is used.

Additionally, the collected expire time is added to the prerender manifest as `initialExpireSeconds` (and `fallbackExpire`), analogous to `initialRevalidateSeconds` (or `fallbackRevalidate`).

closes NAR-100

PR #93211, opened 2026-04-24
Honor the route-level `expire` value with blocking revalidation

### LINKED ISSUE #78269: Unexpected `cacheLife` `expire` behaviour

https://codesandbox.io/p/devbox/next-cache-life-expire-issue-c6wdvt

### To Reproduce

1. Execute commands:
```
npm run build
npm start
```
2. Browse to root page. Take note of displayed time.
3. Wait 5 mins.
4. Refresh page once. Note that old time is displayed.

This project is using the new canary `dynamicIO`, `use cache` and `cacheLife` functionality.

The [documentation](https://nextjs.org/docs/app/api-reference/functions/cacheLife) for `cacheLife` `expire` says:

> Maximum duration for which a value can remain stale before switching to dynamic fetching

The following is specified in the sample project:

```ts
cacheLife({ stale: 60, revalidate: 3 * 60, expire: 5 * 60 });
```

Therefore, when refreshing the page after waiting 5 minutes, I expect it to switch to dynamic fetching and return the current time. This is the functionality I want and expect. Instead, the old stale/expired value is being returned.

```bash
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 4242
  Available CPU cores: 2
Binaries:
  Node: 20.9.0
  npm: 9.8.1
  Yarn: 1.22.19
  pnpm: 8.10.2
Relevant Packages:
  next: 15.3.1-canary.11 // Latest available version is detected (15.3.1-canary.11).
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.3.3
Next.js Config:
  output: N/A
```

dynamicIO, Dynamic Routes, Use Cache

next start (local)

### Additional context

_No response_

unstubbable:
A prerendered route's `expire` — set via `cacheLife({ expire })` inside `'use cache'` or via the `expireTime` config fallback — lands in the prerender manifest as `initialExpireSeconds` / `fallbackExpire` (#76207), but the runtime never read it: `IncrementalCache.get` only considered `revalidate`. So past expire, Next.js served stale with a background refresh instead of the blocking regeneration the `cacheLife` `expire` docs describe.

The fix is three coordinated changes. The render-time `responseGenerator` in `app-page.ts`, `app-route.ts`, and `pages-handler.ts` now applies the `expireTime` fallback as soon as it has the render's `cacheControl`, so every downstream consumer (the cache stored via `IncrementalCache.set`, the response `Cache-Control` header, the entry returned to `handleResponse`) sees a finalized `cacheControl` with a populated `expire` — mirroring the build-time fallback. `IncrementalCache.get` then returns `isStale = -1` when `lastModified + expire * 1000 < now`, and `response-cache.handleGet` skips its early `resolve(previousEntry)` for `isStale === -1` so the blocking revalidation inside `responseGenerator` (which already picks `BLOCKING_STATIC_RENDER` on that signal) can return its fresh output to the user. Previously the early resolve committed the stale value to the response first, so even though `responseGenerator` still ran a fresh render its output only warmed the cache for the next request. As a side effect this also closes the same early-resolve hole on the existing tag-expired `isStale = -1` path.

On Vercel, ISR cache decisions live at the Proxy and the Proxy currently ignores `staleExpiration` (using a hard-coded one-year value instead). It is also expected, once it starts honoring `staleExpiration`, to pick up updated values from the `stale-while-revalidate` response header. Until that lands this change is only observable on `next start` — deploy-mode behavior is tracked independently of Next.js.

Two test suites cover the new behavior. `test/production/app-dir/use-cache-expire` uses `cacheComponents` + `cacheLife({ expire: 300 })` with a custom cache handler that shifts `lastModified` via an `x-test-cache-age-offset-ms` header, exercising the fully-static shell, the partially-static route shell for a known param, and the partially-static fallback shell for unknown params. `test/e2e/app-dir/expire-time` covers classic ISR (`revalidate = 1`, `expireTime: 2`) with a real three-second wait and is `it.failing` on deploy, so it will flip the moment the Proxy honors the expire value.

fixes #78269

### DIFF
--- a/packages/next/src/build/templates/app-page.ts
+++ b/packages/next/src/build/templates/app-page.ts
@@ -948,8 +948,15 @@ export async function handler(
       // `incrementalCache.set`, the response Cache-Control header, the outgoing
       // entry returned to `handleResponse`) sees a finalized `cacheControl`
       // with a populated `expire`. This mirrors the build-time fallback in
-      // `next build` (see `build/index.ts`).
-      if (cacheControl && cacheControl.expire === undefined) {
+      // `build/index.ts` so we don't apply an expire to routes that opt out of
+      // revalidation entirely (`revalidate: false`) or that are dynamic
+      // (`revalidate: 0`).
+      if (
+        cacheControl &&
+        cacheControl.revalidate !== false &&
+        cacheControl.revalidate > 0 &&
+        cacheControl.expire === undefined
+      ) {
         cacheControl.expire = nextConfig.expireTime
       }
 
--- a/packages/next/src/build/templates/app-page.ts
+++ b/packages/next/src/build/templates/app-page.ts
@@ -943,6 +943,16 @@ export async function handler(
         fetchMetrics,
       } = metadata
 
+      // Apply the `expireTime` fallback as soon as we have the render's
+      // `cacheControl`, so every downstream consumer (the cache stored via
+      // `incrementalCache.set`, the response Cache-Control header, the outgoing
+      // entry returned to `handleResponse`) sees a finalized `cacheControl`
+      // with a populated `expire`. This mirrors the build-time fallback in
+      // `next build` (see `build/index.ts`).
+      if (cacheControl && cacheControl.expire === undefined) {
+        cacheControl.expire = nextConfig.expireTime
+      }
+
       if (cacheTags) {
         headers[NEXT_CACHE_TAGS_HEADER] = cacheTags
       }
@@ -1605,7 +1615,7 @@ export async function handler(
 
             cacheControl = {
               revalidate: cacheEntry.cacheControl.revalidate,
-              expire: cacheEntry.cacheControl?.expire ?? nextConfig.expireTime,
+              expire: cacheEntry.cacheControl.expire,
             }
           }
           // Otherwise if the revalidate value is false, then we should use the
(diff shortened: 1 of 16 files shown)


packages/next/src/build/templates/app-page.ts
```diff
@@ -943,6 +943,16 @@ export async function handler(
         fetchMetrics,
       } = metadata
 
+      // Apply the `expireTime` fallback as soon as we have the render's
+      // `cacheControl`, so every downstream consumer (the cache stored via
+      // `incrementalCache.set`, the response Cache-Control header, the outgoing
+      // entry returned to `handleResponse`) sees a finalized `cacheControl`
+      // with a populated `expire`. This mirrors the build-time fallback in
+      // `next build` (see `build/index.ts`).
+      if (cacheControl && cacheControl.expire === undefined) {
```
ztanner: To match the build time behavior, shouldn't this look more like this?
https://github.com/vercel/next.js/blob/521ae9653e7c45ee1690323c86252b82d5740ffa/packages/next/src/build/index.ts#L3068-L3071

unstubbable: Good catch, fixed.

ztanner (approved):

PR #93213, opened 2026-04-24
Turbopack: simplify asset ident constructors

lukesandberg:
### What?

Removes the per-method turbo-task constructors on `AssetIdent` (`from_path`, `with_query`, `with_fragment`, `with_modifier`, `with_part`, `with_path`, `with_layer`, `with_content_type`, `with_asset`, `rename_as`, and `path`). Each of those was its own cached task that returned a small projection or a one-field-changed copy. They are now plain Rust builder methods on the owned value, with a single `into_vc()` at the end of the chain that goes through the existing cached `new_inner` constructor.

Call sites that previously chained `Vc` methods now look like:

```rust
module
    .ident()
    .owned()
    .await?
    .with_modifier(rcstr!("async loader"))
    .into_vc()
```

### Why?

These constructors were tiny "projection" turbo-tasks that paid the cost of a task lookup, cell allocation, and dependency tracking but whose cache layer didn't meaningfully prevent recomputation. The trade-off is invalidation semantics:

- **Before:** a caller doing `module.ident().path()` depended on the cached `path()` projection. If the source `AssetIdent` changed but its `.path` field was unchanged (e.g. a new modifier was added), `path()` re-ran, returned the same `FileSystemPath` cell, and the caller did not re-run.
- **After:** the same caller does `module.ident().await?.path` and depends directly on the `AssetIdent` cell. Any change to the ident (modifier, query, layer, …) invalidates the caller, even if the path is unchanged.

In practice this is rarely a real loss: when an ident changes, the `Module` typically changes too, and the dependent task was going to re-run anyway. `new_inner` already deduplicates structurally-equal idents, so the wrappers were paying overhead per call without buying meaningful invalidation isolation.

Measured on a `vercel-site` build via `NEXT_TURBOPACK_TASK_STATISTICS` and `turbopack/scripts/analyze_cache_effectiveness.py`:

| Task                              | canary (hits / misses) | this branch |
| --------------------------------- | ---------------------- | ----------- |
| `AssetIdent::path`                | 778,273 / 98,018       | removed     |
| `AssetIdent::with_modifier`       | 27,895 / 22,801        | removed     |
| `AssetIdent::from_path`           | 2,954 / 29,650         | removed     |
| `AssetIdent::with_part`           | 2 / 5,440              | removed     |
| `AssetIdent::with_layer`          | 7 / 4,356              | removed     |
| `AssetIdent::rename_as`           | 4,969 / 2,269          | removed     |
| `AssetIdent::with_query`          | 0 / 521                | removed     |
| `AssetIdent::with_content_type`   | 0 / 79                 | removed     |
| `AssetIdent::new_inner`           | 628 / 129,777          | 29,213 / 120,650 |

Aggregate over the whole build:

- Total cached tasks: 1,300 → 1,292
- Total task invocations: 39,361,186 → 38,208,036 (~1.15M fewer lookups)
- Total cache misses: 6,812,198 → 6,639,937 (~172k fewer)
- Overall hit rate: 82.7% → 82.6% (essentially unchanged)

`new_inner` absorbs the construction work that used to be split across the wrappers. Four upstream tasks gained +519 cache hits each (`EsmAssetReference::resolve_reference`, `ReferencedAsset::from_resolve_result`, `NextServerUtilityModule::ident`, `NodeJsChunkingContext::chunk_item_id_strategy`); no task gained any new misses.

### How?

- `AssetIdent::from_path` and the `with_*` methods are now plain `&mut self`/`self`-by-value builder methods on the struct itself, not `#[turbo_tasks::function]`s.
- A new `AssetIdent::into_vc(self)` finalizes the builder by going through the still-cached `new_inner`.
- `AssetIdent::path()` is removed; callers use `.path` on an owned `AssetIdent`.
- All call sites across `turbopack-*` and `next-*` crates are updated. Most go from `ident.with_modifier(m)` (returning `Vc`) to `ident.owned().await?.with_modifier(m).into_vc()`.
- A follow-up commit removes a few `.clone()`s introduced in the conversion that aren't needed once lifetimes are bound to a local.

### Follow-ups (out of scope)

While migrating call sites, two pre-existing entry builders surfaced as candidates for cleanup. Not addressed here, but worth noting:

- `get_app_page_entry` (`crates/next-core/src/next_app/app_page_entry.rs`) replaces the *content* of the source returned by `load_next_js_template` (prefixing imports onto `result.build()`) but reuses the template's `ident` with a `?page=...` query suffix as a disambiguator. The new `VirtualSource` ends up with content from one place and an ident chain pointing at another. A cleaner shape would be to mint a fresh ident from the page path, since the caller already knows what it's building.
- `create_page_ssr_entry_module` (`crates/next-pages/page_entry.rs`) has the same shape on the instrumentation-conflict branch: it appends `export const register = hoist(...)` to the template content and constructs a `VirtualSource` with the original `source.ident()` unchanged. Lower-frequency than the app-page case (fires at most once per build), but the ident still misrepresents the constructed content.

### DIFF
--- a/turbopack/crates/turbopack-core/src/ident.rs
+++ b/turbopack/crates/turbopack-core/src/ident.rs
@@ -103,6 +103,7 @@ impl AssetIdent {
 
     /// Finalizes the builder by turning the owned [`AssetIdent`] into a cached [`Vc<AssetIdent>`].
     pub fn into_vc(self) -> Vc<Self> {
+        // This optimizes cache misses in cold builds by only storing one copy of the AssetIdent.
         AssetIdent::new_inner(ReadRef::new_owned(self))
     }
 
--- a/turbopack/crates/turbopack-core/src/ident.rs
+++ b/turbopack/crates/turbopack-core/src/ident.rs
@@ -6,8 +6,8 @@ use once_cell::sync::Lazy;
 use regex::Regex;
 use turbo_rcstr::RcStr;
 use turbo_tasks::{
-    NonLocalValue, ResolvedVc, TaskInput, ValueToString, ValueToStringRef, Vc, trace::TraceRawVcs,
-    turbofmt,
+    NonLocalValue, ReadRef, ResolvedVc, TaskInput, ValueToString, ValueToStringRef, Vc,
+    trace::TraceRawVcs, turbofmt,
 };
 use turbo_tasks_fs::FileSystemPath;
 use turbo_tasks_hash::{DeterministicHash, Xxh3Hash64Hasher, encode_base38, hash_xxh3_hash64};
--- a/turbopack/crates/turbopack-core/src/ident.rs
+++ b/turbopack/crates/turbopack-core/src/ident.rs
@@ -84,25 +84,77 @@ pub struct AssetIdent {
 }
 
 impl AssetIdent {
-    pub fn new(ident: AssetIdent) -> Vc<Self> {
-        AssetIdent::new_inner(ReadRef::new_owned(ident))
+    /// Creates an [AssetIdent] from a [FileSystemPath].
+    ///
+    /// Returns an owned value; call [`AssetIdent::into_vc`] at the end of the builder chain to
+    /// turn it into a [`Vc<AssetIdent>`].
+    pub fn from_path(path: FileSystemPath) -> Self {
+        AssetIdent {
+            path,
+            query: RcStr::default(),
+            fragment: RcStr::default(),
+            assets: Vec::new(),
+            modifiers: Vec::new(),
+            parts: Vec::new(),
+            layer: None,
+            content_type: None,
+        }
     }
 
-    pub fn add_modifier(&mut self, modifier: RcStr) {
+    /// Finalizes the builder by turning the owned [`AssetIdent`] into a cached [`Vc<AssetIdent>`].
+    ///
+    /// Internally delegates to `new_inner` which is the only cached turbo-task constructor for
+    /// `AssetIdent`, avoiding redundant task executions for every builder step.
+    pub fn into_vc(self) -> Vc<Self> {
+        AssetIdent::new_inner(ReadRef::new_owned(self))
+    }
+
+    pub fn with_query(mut self, query: RcStr) -> Self {
+        self.query = query;
+        self
+    }
+
+    pub fn with_fragment(mut self, fragment: RcStr) -> Self {
+        self.fragment = fragment;
+        self
+    }
+
+    pub fn with_modifier(mut self, modifier: RcStr) -> Self {
         debug_assert!(!modifier.is_empty(), "modifiers cannot be empty.");
         self.modifiers.push(modifier);
+        self
     }
 
-    pub fn add_asset(&mut self, key: RcStr, asset: ResolvedVc<AssetIdent>) {
+    pub fn with_part(mut self, part: ModulePart) -> Self {
+        self.parts.push(part);
+        self
+    }
+
+    pub fn with_path(mut self, path: FileSystemPath) -> Self {
+        self.path = path;
+        self
+    }
+
+    pub fn with_layer(mut self, layer: Layer) -> Self {
+        self.layer = Some(layer);
+        self
+    }
+
+    pub fn with_content_type(mut self, content_type: RcStr) -> Self {
+        self.content_type = Some(content_type);
+        self
+    }
+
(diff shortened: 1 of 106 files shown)


lukesandberg:
* **#93297** <a href="https://app.graphite.com/github/pr/vercel/next.js/93297?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93213** <a href="https://app.graphite.com/github/pr/vercel/next.js/93213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93213?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbopack-core/src/ident.rs
```diff
@@ -133,75 +98,81 @@ impl AssetIdent {
...
-    pub fn with_query(&self, query: RcStr) -> Vc<Self> {
-        let mut this = self.clone();
-        this.query = query;
-        Self::new(this)
+    /// Finalizes the builder by turning the owned [`AssetIdent`] into a cached [`Vc<AssetIdent>`].
+    ///
+    /// Internally delegates to `new_inner` which is the only cached turbo-task constructor for
+    /// `AssetIdent`, avoiding redundant task executions for every builder step.
+    pub fn into_vc(self) -> Vc<Self> {
+        AssetIdent::new_inner(ReadRef::new_owned(self))
```
mischnic, line 107: What is the `ReadRef::new_owned` doing here? Is there any benefit compared to just passing the owned value as a parameter?

turbopack/crates/turbopack-core/src/issue/mod.rs
```diff
@@ -1042,7 +1042,7 @@ impl PlainSource {
 
         Ok(PlainSource {
             ident: asset.ident().to_string().await?,
-            file_path: asset.ident().path().to_string().await?,
+            file_path: ReadRef::new_owned(asset.ident().await?.path.to_string_ref().await?),
```
mischnic: Should we just store `RcStr` instead of these type gymnastics?

turbopack/crates/turbopack-core/src/ident.rs
lukesandberg: I think the benefit is 1 copy of the AssetIdent instead of 2

The ReadRef is packed into the TaskType for the task so it gets stored both in TaskType and in a cell, but this is cheap because it is an `Arc::clone`, if it was by value we would have two copies of the AssetIdent struct.

mischnic: Or if I rephrase the question: this is a way of passing an owned value to a turbo task. Why don't we do `ReadRef::new_owned` everywhere?

lukesandberg: do we have many 'pure interning' turbotasks? i think the pattern is reasonable for large owned data....

of course persistence does defeat this pattern 😢 .  We avoid duplicate copies in memory, but when we persist we do duplicate the readref target and so end up with 2 on the next execution (or post eviction 🙀 )

lukesandberg: so yeah i guess this is pure waste

in the cache hit case we needlessly box and then drop the AssetIdent (making a cache _hit_ more expensive) and in the cache miss case we do avoid a copy but only until the next persist cycle after which we cannot dedupe the two copies.  

So taking this by value avoids a temporary and a permanent heap allocation at the expensive of slightly more memory in a cold build

lukesandberg: ok after the discuission in office hours i restored this optimization.  on the plus side, for asset ident it is probably not a big deal in either direction

mischnic (approved):

lukesandberg:
### Merge activity

* **May 6, 8:35 AM UTC**: [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93213) couldn't merge this PR because it failed for an unknown reason (unsigned commits detected).
* **May 6, 5:37 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93213).
* **May 6, 5:38 PM UTC**: @lukesandberg merged this pull request with [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93213).

PR #93218, opened 2026-04-24
docs: add guide and references for experimental.useOffline

icyJoseph:
Closes: https://linear.app/vercel/issue/DOC-6390/useoffline-guide-and-api-ref

 - [x] release flag status: default / opt-out
 - [ ] publish repo and demo
 - [x] ensure learning path is correct (basic -> w/ client hook -> server actions)


docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+  description: The hook and config flag used in this guide.
+  links:
+    - app/api-reference/functions/use-offline
+    - app/api-reference/config/next-config-js/useOffline
+    - app/guides/progressive-web-apps
+---
+
+Network failures during a soft navigation, a data fetch, or a mutation throw errors in the client. Without explicit handling, the UI either breaks or you have to build a fallback UI that asks the user to retry.
+
+With [`experimental.useOffline`](/docs/app/api-reference/config/next-config-js/useOffline) enabled, a failed navigation, RSC data fetch, prefetch, or Server Action no longer throws when the network is down. Next.js keeps it pending and retries it once the connection returns.
```
aurorascharff, line 17: Andrew mentioned this is not going to be a flag but rather default enabled

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+```tsx filename="app/dashboard/page.tsx" highlight={3,9} switcher
+import { Suspense } from 'react'
+import { getLiveMetrics } from '../lib/data'
+import { ConnectivityFallback } from './connectivity-fallback'
+
+export default function Dashboard() {
+  return (
+    <section>
+      <h1>Live metrics</h1>
+      <Suspense fallback={<ConnectivityFallback />}>
```
aurorascharff, line 199: Interesting pattern. If we lose connection randomly on a page, would the fallback reappear though? If no (which I assume), that means we would always want to pair this with a banner (as below)?

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+
+  function handleSubmit() {
+    startTransition(async () => {
+      const pong = await ping()
+      setPongs((prev) => [pong, ...prev])
+    })
+  }
+
+  const label = pending
+    ? isOffline
```
aurorascharff, line 356: So cool!

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+Requests you issue directly with `fetch()` inside a Client Component, or through a client-side data library like React Query or SWR, stay under that library's own retry policy. See [How retry works](/docs/app/api-reference/config/next-config-js/useOffline#how-retry-works) for the framework's detection and polling behavior.
+
+## Example
+
+We will build a live-metrics page that fetches fresh data on every request, plus a ping form that calls a Server Action.
+
+{/* PLACEHOLDER: replace with the actual URLs before publishing */}
+
+- Source: [github.com/vercel-labs/use-offline](https://github.com/vercel-labs/use-offline)
+- Live demo: [DEMO_URL]
```
aurorascharff: Add the live URL

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+
+The companion demo has two versions of this dashboard: `/without-feedback` uses a generic loading fallback, `/with-feedback` uses a connectivity-aware one. The next sections of this guide walk through building each.
+
+## Enable offline detection and see the default behavior
+
+Turn on `experimental.useOffline`. This guide also enables [Cache Components](/docs/app/api-reference/config/next-config-js/cacheComponents), which lets you place the Suspense boundary as close as possible to the uncached data, with the [static shell](/docs/app/glossary#static-shell) rendered around it.
+
+Without Cache Components, a route-level [`loading.tsx`](/docs/app/api-reference/file-conventions/loading) gives you the same offline behavior at the segment level. See [Without Cache Components](#without-cache-components) below.
```
aurorascharff, line 36: We'd also want to enable partial prefetching here because otherwise you'd have blocking navigations on Offline because you wouldn't have a shareable app shell. 

So I think `useOffline` should really only be used with both of those features enabled, right?

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+      <h1>Live metrics</h1>
+      <Suspense fallback={<ConnectivityFallback />}>
+        <MetricsTable />
+      </Suspense>
+    </section>
+  )
+}
+```
+
+Navigating to the dashboard while offline, the fallback now reads "Waiting for connection to load this section..." Restore connectivity and the metrics stream in; the fallback disappears.
```
aurorascharff: could remove this semicolon

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+      <body>
+        <OfflineBanner />
+        {children}
+      </body>
+    </html>
+  )
+}
+```
+
+Pick whichever fits your use case. One banner in a root layout covers every route. A scoped fallback is more targeted but only shows on routes where you add it. Both together is also a reasonable default.
```
aurorascharff: We should maybe mention the process here that the default is fine. Adding a banner is nice. If you want to specialize, add a fallback UI.

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+
+The banner shows on every route while offline, and hides when connectivity returns.
+
+The same pattern extends to parameterized routes. In this example we used a static route (`/dashboard`), but a route like `/chats/[id]` with a static shell renders that shell when navigating to `/chats/42` offline, and the dynamic messages behind its `<Suspense>` boundary load when the connection returns.
```
aurorascharff: Do we want to mention runtime prefetching? If you've runtime prefetched your data, it would still be accessibl offline.

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,434 @@
...
+
+## Testing
+
+Test this feature with `next build && next start`. Dev mode is not a reliable reference for offline behavior.
+
+In Chrome, use [DevTools > Network > **Offline**](https://developer.chrome.com/docs/devtools/network/reference#offline); in Firefox, use the [Network Monitor's throttling menu](https://firefox-source-docs.mozilla.org/devtools-user/network_monitor/throttling/index.html). For a real-world test, toggle airplane mode on your laptop or phone, disconnect WiFi, or unplug the network cable.
+
+## Without Cache Components
+
+A route-level [`loading.tsx`](/docs/app/api-reference/file-conventions/loading) does the same job. Next.js wraps the segment's page content in a `<Suspense>` boundary with the `loading.tsx` export as its fallback; the layout above stays rendered. While offline, the cached layout and the `loading.tsx` fallback render from the prefetch, and the page resumes once the network returns. The `useOffline` hook, the banner, and the Server Action retry all work the same way.
```
aurorascharff: s fallback; the layout above stays rendered.

This sentence could be improved and remove the semicolon

docs/01-app/02-guides/prefetching.mdx
```diff
@@ -203,7 +203,7 @@ Next.js maintains a small task queue, which prefetches in the following order:
 3. Newer links replace older ones
 4. Links scrolled off-screen are discarded
 
-The scheduler prioritizes likely navigations while minimizing unused downloads.
+The scheduler prioritizes likely navigations while minimizing unused downloads. Pending prefetches resume through the same queue when the app recovers from an offline period; see [Handling connectivity drops](/docs/app/guides/offline-support).
```
aurorascharff:
```suggestion
The scheduler prioritizes likely navigations while minimizing unused downloads. Pending prefetches resume through the same queue when the app recovers from an offline period. See [Handling connectivity drops](/docs/app/guides/offline-support).
```

docs/01-app/01-getting-started/04-linking-and-navigating.mdx
```diff
@@ -258,6 +258,10 @@ export default function LoadingIndicator() {
 
 You can "debounce" the hint by adding an initial animation delay (e.g. 100ms) and starting as invisible (e.g. `opacity: 0`). This means the loading indicator will only be shown if the navigation takes longer than the specified delay. See the [`useLinkStatus` reference](/docs/app/api-reference/functions/use-link-status#gracefully-handling-fast-navigation) for a CSS example.
 
+Next.js handles connectivity drops gracefully: users can keep navigating between prefetched routes, and any route that needs server data completes once the network returns. Use the [`useOffline`](/docs/app/api-reference/functions/use-offline) hook to surface that state in your UI. See the [offline support guide](/docs/app/guides/offline-support).
```
aurorascharff: Next.js handles connectivity drops gracefully: 

hm well while its still a flag, not really true. it CAN handle though

docs/01-app/02-guides/progressive-web-apps.mdx
```diff
@@ -671,6 +671,6 @@ Learn more about defining [Content Security Policies](/docs/app/guides/content-s
...
-3. **Offline Support**: To provide offline functionality, one option is [Serwist](https://github.com/serwist/serwist) with Next.js. You can find an example of how to integrate Serwist with Next.js in their [documentation](https://github.com/serwist/serwist/tree/main/examples/next-basic). **Note:** this plugin currently requires webpack configuration.
+3. **Offline Support**: Next.js ships an experimental [`useOffline`](/docs/app/api-reference/functions/use-offline) hook and matching [`experimental.useOffline`](/docs/app/api-reference/config/next-config-js/useOffline) config for connectivity-aware UI and automatic retry of failed navigation and Server Action requests. For full service-worker-based offline caching, one option is [Serwist](https://github.com/serwist/serwist) with Next.js. You can find an example of how to integrate Serwist with Next.js in their [documentation](https://github.com/serwist/serwist/tree/main/examples/next-basic). **Note:** this plugin currently requires webpack configuration.
```
aurorascharff: Next.js ships an experimental 

"provides"

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,436 @@
...
+      <body>
+        <OfflineBanner />
+        {children}
+      </body>
+    </html>
+  )
+}
+```
+
+Start with the default: the pending loading state is enough for many apps. Add a banner in the root layout when you want connectivity visible on every route. Add a scoped fallback when a specific route deserves a tailored message. The three layer in that order, so you can adopt as much as you need.
```
aurorascharff: some AI here to fix

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,436 @@
...
+
+The banner shows on every route while offline, and hides when connectivity returns.
+
+The same pattern extends to parameterized routes. A route like `/chats/[id]` renders its shared [App Shell](/docs/app/glossary#app-shell) when you navigate to `/chats/42` offline, and the dynamic messages behind its `<Suspense>` boundary load when the connection returns.
+
```
aurorascharff, line 309: FYI this allow runtime will need to be removed later

docs/01-app/02-guides/offline-support.mdx
```diff
@@ -0,0 +1,436 @@
...
+## Testing
+
+Test this feature with `next build && next start`. Dev mode is not a reliable reference for offline behavior.
+
+In Chrome, use [DevTools > Network > **Offline**](https://developer.chrome.com/docs/devtools/network/reference#offline); in Firefox, use the [Network Monitor's throttling menu](https://firefox-source-docs.mozilla.org/devtools-user/network_monitor/throttling/index.html). For a real-world test, toggle airplane mode on your laptop or phone, disconnect WiFi, or unplug the network cable.
+
+## Without Cache Components
+
+A route-level [`loading.tsx`](/docs/app/api-reference/file-conventions/loading) does the same job. Next.js wraps the segment's page in a `<Suspense>` boundary using the `loading.tsx` export as its fallback, leaving the layout above rendered. While offline, the cached layout and fallback render from the prefetch, and the page resumes once the network returns. The `useOffline` hook, banner, and Server Action retry all behave the same way.
```
aurorascharff: should we link down here when we reference this further up?

aurorascharff: `Next.js wraps the segment's page in a `<Suspense>` boundary using the `loading.tsx` export as its fallback, leaving the layout above rendered.`

its not clear how this is different from normal suspense boundaries and why they dont work without cache components

aurorascharff (approved):

docs/01-app/02-guides/progressive-web-apps.mdx
adhamfayrouzamf: Hi Aurora
just wanted to mention there is serwist example with turbopack (it uses esbuild though)
[Here is the link](https://github.com/serwist/serwist/tree/main/examples/next-turbo-basic)

aurorascharff: @icyJoseph

icyJoseph: @adhamfayrouzamf PR's are always welcome :)

adhamfayrouzamf: done 👍

PR #93219, opened 2026-04-24
[ci] Populate GITHUB_TOKEN for auto-labeler

eps1lon:


.github/workflows/pull_request_auto_label.yml
```diff
@@ -59,9 +59,9 @@ jobs:
     if: github.repository_owner == 'vercel'
     runs-on: ubuntu-latest
     permissions:
-      # To checkout the action
-      contents: read
```
eps1lon, line 63: revert from prior attempt to fix that was actually fixed by https://github.com/vercel/next.js/pull/93217

ztanner (approved):

PR #93235, opened 2026-04-25
[ci] Automatically start React sync when new `react@experimental` is available

eps1lon:
We now check with a GitHub cronjob every 5 minutes if a new `react@experimental` is out. We detect changes by persisting the last seen version with GitHub Actions cache.

This is slower than the previous Vercel cronjob (every 2 minutes) but gets rid of one more token to distribute.

### DIFF
--- a/.github/workflows/update_react_poller.yml
+++ b/.github/workflows/update_react_poller.yml
@@ -1,0 +1,69 @@
+name: Poll React Sync
+
+on:
+  schedule:
+    # GitHub runs cronjobs at most every 5 minutes.
+    # Ideally we get 2 minutes or lower which is what we used to have with Vercel cronjobs.
+    - cron: '*/5 * * * *'
+
+concurrency:
+  group: poll-react-sync
+  cancel-in-progress: true
+
+jobs:
+  poll:
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/update_react_poller.yml
```diff
@@ -0,0 +1,69 @@
...
+          if [ -z "$CURRENT" ]; then
+            echo "Failed to resolve react@experimental version" >&2
+            exit 1
+          fi
+          echo "version=$CURRENT" >> "$GITHUB_OUTPUT"
+          echo "Current react@experimental: $CURRENT"
+
+      - name: Check whether this version was already seen
+        id: cache
+        uses: actions/cache@v4
```
ztanner, line 35: Could we use cache restore here and only save the `react-experimental-${version}` marker after `update_react.yml` completes successfully? I think this will save the cache when the dispatch succeeds but it's possible the sync later fails, and then the poller will skip it as though it's successful.

eps1lon: That wasn't really a problem before. It's also not trivial to determine when we should retry. If it fails on creating a PR, a retry would be a noop since the branch already exists. 

And if we only set the marker once the sync is up, we risk restarting an in-progress sync.

ztanner (approved):

PR #93236, opened 2026-04-25
[ci] Create test summary comment if job immediately passes

eps1lon:
Not sure if this was intentional or just very pessimistic about our test flakiness 😄 E.g. https://github.com/vercel/next.js/pull/93235 passed immediately and no comment was posted: https://github.com/vercel/next.js/actions/runs/24928412554/job/73002141215#step:4:7

Two reasons to always post the comment:
1. confusing if the comment is absent when it would appear when we go from failing -> passing
2. Means the comment may be pushed further down in the conversation if your initial push is passing, a review is posted, new code is pushed with a failing build. Now the comment is further down when it's usually way up.

### DIFF
--- a/scripts/pr-ci-comment.mjs
+++ b/scripts/pr-ci-comment.mjs
@@ -130,13 +130,7 @@ class GitHubClient {
     )
   }
 
-  async upsertIssueComment(
-    issueNumber,
-    marker,
-    body,
-    fallbackHeadings = [],
-    { createIfMissing = true } = {}
-  ) {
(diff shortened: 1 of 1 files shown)


ztanner (approved):
Copied this from [here](https://github.com/vercel/next-pr-webhook/blob/7929ebc3f5e08397bc028cd7d9423ad598761ce4/src/lib/post-comment.ts#L80-L85) but I do appreciate a consistent celebration when my tests pass, so lfgtm.

eps1lon:
We need a bit of positivity.

PR #93240, opened 2026-04-25
[ci] Reserve comments for stats and tests

eps1lon:
This increases the chances of finding the comments in the same place: at the tippity top of the PR conversation.

### DIFF
--- a/scripts/pr-ci-comment.mjs
+++ b/scripts/pr-ci-comment.mjs
@@ -562,24 +562,6 @@ async function handleBuildAndTestWorkflow({
     return
   }
 
-  if (workflowRun.conclusion === 'cancelled') {
-    const sha = pr.headSha || workflowRun.head_sha
-    const body = [
-      TEST_COMMENT_MARKER,
-      '## Tests cancelled',
-      '',
-      `Commit: ${sha}`,
-      `[View workflow run](${workflowRun.html_url})`,
-      '',
-    ].join('\n')
-
-    await github.upsertIssueComment(pr.number, TEST_COMMENT_MARKER, body, [
-      '## Failing test suites',
-      '## Failing CI jobs',
-    ])
-    return
-  }
-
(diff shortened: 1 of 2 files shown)


scripts/pr-ci-comment.mjs
```diff
@@ -398,7 +458,27 @@ async function readPullRequestMetadataArtifact() {
   }
 }
 
-async function handleStatsWorkflow({ github, workflowRun, pr }) {
+async function handleStatsWorkflow({ github, workflowRun, pr, phase }) {
+  if (phase === 'requested') {
+    const sha = pr.headSha || workflowRun.head_sha
+    const body = [
+      STATS_COMMENT_MARKER,
+      '## Stats in progress',
```
ztanner, line 466: Docs-only changes won't trigger the stats workflow, so we'd potentially end up in a state where there's a "Stats in progress" comment that is never fulfilled

eps1lon: Right. Also did the same for build-and-test in case people cancel

eps1lon: A docs-only Generate stats actually counts as success: https://github.com/vercel/next.js/actions/runs/24933442991/job/73014965795?pr=93243 -> https://api.github.com/repos/vercel/next.js/actions/runs/24933442991

ztanner: It has `conclusion: success`, but the aggregate/post step is skipped.

So when the reporter handles the completed run, it enters `handleStatsWorkflow`, sees it was not cancelled, scans the aggregate logs, finds no `--stats start--` block, and returns without updating the existing in progress placeholder

eps1lon: https://github.com/vercel/next.js/pull/93240/changes/772eec5a470be23f49d0e76cc94cd52c03d10908

unstubbable (approved):

PR #93245, opened 2026-04-25
[ci]: app-based release workflow

ztanner:
Moves all release workflows off of a GH PAT and uses an app with a short-lived token instead.

Test Plan:
Dry run [here](https://github.com/vercel/next.js/actions/runs/24934593874).

However, this workflow is blocked until we figure out commit signing for the bot app. Some options:
- The bot account generates a signing key and we use it in CI (not great, bypasses the app)
- The org bypasses signature verification for the bot user (also not great, requires an exemption rule)
- We need to rework the commit step so Lerna does not do the push, and instead trigger it via the app + GH API. This seems like the best option, will be added in a follow-up PR.

Note: `create-release-branch` workflow is broken in its current form, as we will not be restoring administrator privileges to adjust environment settings. This will become a manual step in the future.

### DIFF
--- a/.github/workflows/trigger_release.yml
+++ b/.github/workflows/trigger_release.yml
@@ -1,4 +1,13 @@
 on:
+  push:
+    branches:
+      - 'release-app-dry-run/**'
+      - 'release-app-dry-run-with-writes/**'
+    paths:
+      - '.github/workflows/trigger_release.yml'
+      - 'scripts/start-release.js'
+      - 'scripts/release-github-auth.js'
+
   schedule:
     # run every day at 23:15
     - cron: '15 23 * * *'
@@ -28,6 +37,16 @@ on:
         default: false
         type: boolean
 
+      dryRun:
+        description: validate release auth and local versioning without pushing commits, tags, or releases
+        default: false
+        type: boolean
+
+      verifyGitWrites:
+        description: create and delete disposable refs during a dry run
+        default: false
+        type: boolean
+
 name: Trigger Release
 
 env:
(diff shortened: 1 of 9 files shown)


.github/workflows/trigger_release.yml
```diff
@@ -61,10 +79,13 @@ jobs:
         # Ignoring failures for now to check if a failure truly implies a failed publish.
         continue-on-error: true
         env:
-          GITHUB_TOKEN: ${{ secrets.RELEASE_BOT_GITHUB_TOKEN }}
+          GH_TOKEN: ${{ steps.release-app-token.outputs.token }}
```
eps1lon, line 82: Thought for a second that this was the bug we sometimes saw but it shouldn't matter. `GH_TOKEN` is preferred though if `GITHUB_TOKEN` is also set globally: https://cli.github.com/manual/gh_help_environment

.github/workflows/trigger_release.yml
```diff
@@ -61,10 +79,13 @@ jobs:
...
 
       - name: Get commit of the latest tag
-        run: echo "LATEST_TAG_COMMIT=$(git rev-list -n 1 $(git describe --tags --abbrev=0))" >> $GITHUB_ENV
+        run: |
+          git fetch --tags --force --deepen=1000 origin
```
eps1lon, line 86: Kinda shocked we always had fewer than 25 (fetch depth) commits between releases

eps1lon (approved):
> We need to rework the commit step so Lerna does not do the push, and instead trigger it via the app + GH API. This seems like the best option, will be added in a follow-up PR.

PR #93264, opened 2026-04-26
[turbopack-trace-server] optimize loading

lukesandberg:
Land a few optimizations to the trace server

* Change `SpanEvent` so it is 32 bytes instead of 40 bytes by triggering a `niche` optimization
* Change `args` and `events` to be a `smallvec` with inline size 1
  * for `args` it is size <=1 ~31% of the time
  * for `events` it is size <=1 69% of the time
* Compute min/max timestamps in a single pass instead of 2 when inserting into the selftimetree
* Bundle dynamically computed 'total' fields behind a single OnceLock
   * saves 40 bytes per span due to `Oncelock` overheads
* Inline SpanTimeData and SpanNames into Span
   * We get little benefit from deferring the allocations and by inlining we save time and improve memory locality.
     * post load SpanTimeData is allocated for 94% of spans, but after loading `trace.nextjs.org` it is 100%
     * post load SpanNames is allocated for 0% of spans, but after loading it is 96.2% of spans
* Remove the `inner` `OnceLocks` from `SpanNames` we can just allocate these all together

Measuring with one 10gb trace file I see loading times progress from 75.7s (33G of ram)  to 60.5s (19.5G of ram).  With loading times hitting >200mb/s occasionally

### DIFF
--- a/turbopack/crates/turbopack-trace-server/src/store.rs
+++ b/turbopack/crates/turbopack-trace-server/src/store.rs
@@ -48,16 +48,11 @@ fn new_root_span() -> Span {
         args: SpanArgs::new(),
         events: Default::default(),
         is_complete: true,
-        max_depth: OnceLock::new(),
         self_allocations: 0,
         self_allocation_count: 0,
         self_deallocations: 0,
         self_deallocation_count: 0,
-        total_allocations: OnceLock::new(),
-        total_deallocations: OnceLock::new(),
-        total_persistent_allocations: OnceLock::new(),
-        total_allocation_count: OnceLock::new(),
-        total_span_count: OnceLock::new(),
+        totals: OnceLock::new(),
         time_data: OnceLock::new(),
         extra: OnceLock::new(),
         names: OnceLock::new(),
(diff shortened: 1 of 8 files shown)


lukesandberg:
* **#93264** <a href="https://app.graphite.com/github/pr/vercel/next.js/93264?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#93265](https://github.com/vercel/next.js/pull/93265) <a href="https://app.graphite.com/github/pr/vercel/next.js/93265?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#93332](https://github.com/vercel/next.js/pull/93332) <a href="https://app.graphite.com/github/pr/vercel/next.js/93332?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93264?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbopack-trace-server/src/store.rs
```diff
@@ -213,15 +211,18 @@ impl Store {
         end: Timestamp,
         outdated_spans: &mut FxHashSet<SpanIndex>,
     ) {
+        let Some(event) = SpanEvent::self_time(start, end) else {
+            return;
```
sokra: It should set `time_data.self_end` in that case and also `outdated_spans.insert(span_index);`

sokra (approved):

Issue #92676, opened 2026-04-12
Fail to load worker-entrypoint.js in Next.js 16.2+ when COEP is 'require-corp'


lucach:
### Link to the code that reproduces this issue

https://github.com/lucach/nextjs-coep-worker-entrypoint-bug

### To Reproduce

1. Start the application (`next dev`)
2. Open the browser console

### Current vs. Expected behavior

Up to Next.js 16.1.*, the worker is successfully loaded and the console shows `Message from worker: hello`.

With Next.js 16.2+ (all the way to canary), the message doesn't appear because the newly introduced `worker-entrypoint.js` file is blocked by the COEP policy.

<img width="1002" height="391" alt="Image" src="https://github.com/user-attachments/assets/0dc359ef-5f5c-4be3-915a-911affd154b0" />

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.4.0: Thu Mar 19 19:30:44 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6000
  Available memory (MB): 65536
  Available CPU cores: 10
Binaries:
  Node: 24.11.1
  npm: 9.6.2
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.2.1-canary.33 // Latest available version is detected (16.2.1-canary.33).
  eslint-config-next: N/A
  react: 19.2.5
  react-dom: 19.2.5
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Turbopack

### Which stage(s) are affected? (Select all that apply)

next dev (local)

### Additional context

The problem has been introduced with https://github.com/vercel/next.js/commit/d23af53268d062ff70b7c12322750b85ed3a76ba

I suspect that `worker-entrypoint.js` is not fetched with the appropriate headers and thus blocked by COEP.

andrewimm:
You're probably right about the blame commit.
Previously the worker was loaded with a `blob://` URL, which leads to a null origin. This doesn't cause CORS related issues, but it does mean fetch and importScript fail when trying to use the worker's origin.
16.2 fixed that by changing the bootstrapping code, but having an origin can lead to obscure CORS issues.

Thanks for the PR. I'll investigate the issue and continue the discussion there

GuinsooRocky:
[...]

PR #88602, opened 2026-01-15
Turbopack: Use a real file entrypoint for Workers (and SharedWorkers)

mmastrac:
# What

Closes https://github.com/vercel/next.js/issues/84782  
Closes https://github.com/vercel/next.js/issues/74842  
Closes https://github.com/vercel/next.js/issues/82520
Closes PACK-3723

Replace blob URLs for web workers with real entrypoint files, fixing relative URL resolution inside workers. Implements support for `SharedWorkers` since it's now a simple variation of the `Worker` case.

Note that all generated files (and shared-worker querystrings) are appropriately content-hashed and do not rely on deployment IDs and will work w/immutable resources.

# Why

Workers created via `new Worker(new URL('./worker.js', import.meta.url))` previously got blob URLs. This breaks relative URL resolution inside workers - WASM imports like `new URL('./file.wasm', import.meta.url)` fail because `import.meta.url` returns `blob:http://...` instead of a real path.

# How

We emit a real worker entrypoint file instead of blob URLs. The entrypoint reads bootstrap config from URL hash params (`#params=....`), then loads the worker's chunks via `importScripts`.

The `ChunkingContext` trait gets a new `worker_entrypoint()` method returning a singleton bootstrap asset. Node.js `bail!`s for now but could implement its own in the future. The browser implementation (`EcmascriptBrowserWorkerEntrypoint`) compiles and minifies `worker-entrypoint.ts`, which parses `#params={globals,chunks}` from the URL at runtime.

`WorkerLoaderChunkItem` now emits `__turbopack_worker_url__(entrypoint, chunks, shared)` instead of the old blob URL shim. SharedWorker passes `shared=true` which adds a querystring - the browser handles worker identity via URL, so same URL means same instance.

Generated code changes (pseudocode):

```javascript
__turbopack_export_value__(__turbopack_worker_blob_url__(["chunk1.js", "chunk2.js"]));
```

After:

```javascript
__turbopack_export_value__(__turbopack_worker_url__("worker-entrypoint.js", ["chunk1.js", "chunk2.js"], false));
```

# Notes

Webpack doesn't currently correctly support `SharedWorkers`. The e2e test behaves properly in turbopack after this change, but webpack spins up two separate workers where it should only be spinning up one.

I generated snapshot tests for workers in the first and last commits to help understand the generated code better.

# Testing

- Snapshot tests for Worker, SharedWorker, and minified output
- E2E tests for WASM loading in workers (`test/e2e/app-dir/worker/`)

Issue #74621, opened 2025-01-07
Typescript error TP1001 when creating a Web Worker in Next.js 15.1.3


farzadsoltani:
### Link to the code that reproduces this issue

https://github.com/farzadso/workers-issue

### To Reproduce

Run `npm run dev` and you can see the issue.

### Current vs. Expected behavior

There should be no errors in the console.

This is the error I see:

```
 ⚠ ./app/page.tsx:9:22
error TP1001 new Worker("/workers/timer.js") is not statically analyse-able
   7 |
   8 |   useEffect(() => {
>  9 |     ticker.current = new Worker("/workers/timer.js");
     |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  10 |   }, []);
  11 |   return <div>Test</div>;
  12 | }
```

If I remove `--turbopack` from the dev script it works fine.

I noticed this happening after my upgrade to Next.js 15 and using Turbopack for dev builds.

The Web Worker is being created dynamically, and it works fine, so I don't see any issues with the functionality.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Dec  6 19:02:12 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6031
  Available memory (MB): 36864
  Available CPU cores: 14
Binaries:
  Node: 20.15.1
  npm: 10.7.0
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 15.2.0-canary.0 // Latest available version is detected (15.2.0-canary.0).
  eslint-config-next: N/A
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.7.2
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Not sure

### Which stage(s) are affected? (Select all that apply)

next dev (local)

sam-skydio:
also experiencing this issue after attempting to turn on Turbopack, would appreciate some guidance here.

farzadsoltani:
@sam-skydio I'm just ignoring it for now as it seems to be working on the client side and not causing any issues when building for prod as the prod build doesn't use Turbopack.

bbsweb:
use `window.Worker` instead of `Worker`

farzadsoltani:
> use `window.Worker` instead of `Worker`
> 
> ```
[...]

Issue #29468, opened 2021-09-28
Asset Prefix breaks Webpack 5 Web Workers


TimoWilhelm:
### What version of Next.js are you using?

11.1.2

### What version of Node.js are you using?

v14.16.0

### What browser are you using?

Chrome, Firefox

### What operating system are you using?

Windows

### How are you deploying your application?

Other

### Describe the Bug

Browsers will not allow you to create a worker with a URL pointing to a different domain. This creates an issue when using the Webpack 5 syntax for creating web workers `new URL('/webworker/sampleworker', import.meta.url)` in combination with the [`assetPrefix`](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix) key in the config pointing at a CDN origin, since it will try to load the worker script from the CDN origin instead.

### Expected Behavior

The [Webpack 5 Web Worker Syntax](https://webpack.js.org/guides/web-workers/)  `new URL('/webworker/sampleworker', import.meta.url)` should work with [`assetPrefix`](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix).

### To Reproduce

- Set the [`assetPrefix`](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix) to any domain (e.g. a CDN) other than the NEXT.js site.
- Try to import a Web Worker using the [Webpack 5 Web Worker Syntax](https://webpack.js.org/guides/web-workers/)  `new URL('/webworker/sampleworker', import.meta.url)`.

See also: https://github.com/vercel/next.js/discussions/29415

Jassu225:
Experiencing the same issue.

Jassu225:
@TimoWilhelm Thanks for the snippet. But it didn't work for me.
I had to find one more workaround.
The following code worked for me.

```javascript
// worker-util.js
const isProd = process.env.NODE_ENV === 'production';

export const getWorker = () => {
  if (isProd) {
    return new Worker(
      new URL(
        '/_next/static/chunks/main.web-worker.js',
        window.location.href
      ),
      {
        name: 'main.web-worker'
      }
    );
  }
  return new Worker(
    new URL('../workers/main/worker.js', import.meta.url),
    {
      name: 'main.web-worker'
    }
  );
};

// next.config.js
...
webpack(config, { isServer, webpack, dev }) {
  const chunkName = isServer
        ? '[name].js'
[...]

PR #93271, opened 2026-04-27
Turbopack: add `experimental.turbopackWorkerAssetPrefix` to keep Workers same-origin when `assetPrefix` is a CDN

GuinsooRocky:
### What?

Adds `experimental.turbopackWorkerAssetPrefix` — a Turbopack counterpart to webpack's [`output.workerPublicPath`](https://webpack.js.org/configuration/output/#outputworkerpublicpath). Opt-in, off by default, fully backward compatible.

```js
// next.config.js
module.exports = {
  assetPrefix: 'https://cdn.example.com',
  experimental: {
    // Same shape as `assetPrefix`: bare prefix, no trailing slash, no `/_next`.
    // `/_next/` is appended automatically. `''` → same-origin `/_next/...`.
    turbopackWorkerAssetPrefix: '',
  },
}
```

### Why?

When `assetPrefix` points to a cross-origin CDN (common production setup), Turbopack emits Worker URLs under that CDN origin. Browsers reject cross-origin `new Worker(url)` with `SecurityError`, breaking every feature that uses `new Worker(new URL('./worker.ts', import.meta.url))` — image decoders, WebP/GIF renderers, compression, OffscreenCanvas, etc.

webpack solved this with `output.workerPublicPath`. Turbopack ignores the `webpack()` callback in `next.config.js`, so the existing same-origin workaround silently breaks when projects switch to Turbopack.

Full design doc, alternatives considered, related issues (#88602, #92676, #74621, #29468): **#93044**

Repro repo (`pnpm dev` reproduces, `pnpm dev:webpack` works): https://github.com/GuinsooRocky/next-turbopack-worker-cdn-repro

### How?

Threads a new optional `turbopack_worker_asset_prefix: Option<RcStr>` from `ExperimentalConfig` → `BrowserChunkingContext` → the JS runtime, where `createWorker` reads it via a new `WORKER_BASE_PATH` global emitted alongside the existing `CHUNK_BASE_PATH`. When unset, behavior is identical to today.

Semantics, modeled on `assetPrefix`:

- **Bare prefix.** `/_next/` is appended automatically (same as `computed_asset_prefix`); the user supplies just the origin (or empty string for same-origin).
- **`undefined` vs `''`.** `undefined` falls back to the regular chunk base path; `''` is a literal empty prefix that resolves to same-origin `/_next/...`. The runtime distinguishes the two by injecting a JS `null` literal vs a quoted string, and uses `WORKER_BASE_PATH ?? CHUNK_BASE_PATH` (not `||`).
- **Applies to entrypoint and module chunks.** The Turbopack worker bootstrap rejects cross-origin module chunks (`Refusing to load script from foreign origin`), so the override has to cover both — overriding only the entrypoint would leave the worker unable to load any chunks under a cross-origin `assetPrefix`.

Touchpoints:

- `packages/next/src/server/config-{shared,schema}.ts` — public option + zod under `experimental`
- `crates/next-{api,core}/...` — wiring through to `ClientChunkingContextOptions`; `NextConfig::turbopack_worker_asset_prefix()` resolves to `Some("<prefix>/_next/")` or `None`
- `turbopack/crates/turbopack-browser/src/chunking_context.rs` — new field + builder + getter
- `turbopack/crates/turbopack-ecmascript-runtime/src/browser_runtime.rs` — emit `WORKER_BASE_PATH` declaration as `null` or a quoted string
- `turbopack/.../runtime-base.ts` — `createWorker` resolves `workerBasePath` and uses it for both the entrypoint URL and module-chunk URLs delivered via params
- `test/e2e/turbopack-worker-asset-prefix/` — e2e fixture using real cross-origin (`localhost` page, `127.0.0.1` `assetPrefix`); intercepts `new Worker()` to assert on the URL the runtime helper resolved. Three cases: no override (cross-origin URL → `SecurityError`), explicit override origin, and `''` (relative `/_next/...`).
- `docs/01-app/03-api-reference/08-turbopack.mdx` — option reference
- `telemetry/events/version.ts` — `useTurbopackWorkerAssetPrefix: boolean` (set on any explicit value, including empty string)

### Notes

- Opening as **Draft** to invite API-shape feedback before going to review. Per the contributing guide, the design discussion is at #93044.
- All checklist items the PR template requires for a feature are covered: e2e tests, docs, telemetry, no error-link path applies.

cc @sokra

Closes #93044


turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -312,15 +318,28 @@ function createWorker(
...
+  // `assetPrefix` points to a cross-origin CDN (browsers reject cross-origin
+  // Worker construction).
+  const workerBasePath =
+    typeof WORKER_PUBLIC_PATH === 'string' && WORKER_PUBLIC_PATH.length > 0
+      ? WORKER_PUBLIC_PATH
+      : CHUNK_BASE_PATH
+
   const chunkUrls = moduleChunks
-    .map((chunk) => getChunkRelativeUrl(chunk))
+    .map((chunk) => getChunkRelativeUrl(chunk, workerBasePath))
```
sokra, line 334:
```suggestion
    .map((chunk) => getChunkRelativeUrl(chunk))
```

The "chunks" of the worker should still be in the normal location. We only want to change the path of the entry script (`url` below)

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -312,15 +318,28 @@ function createWorker(
 ): Worker {
   const isSharedWorker = WorkerConstructor.name === 'SharedWorker'
 
+  // When `workerPublicPath` is set, route both the Worker entrypoint and its
+  // module chunks through that same-origin prefix instead of CHUNK_BASE_PATH.
+  // This mirrors webpack's `output.workerPublicPath` and is needed when
+  // `assetPrefix` points to a cross-origin CDN (browsers reject cross-origin
+  // Worker construction).
```
sokra: The "chunks" should use `CHUNK_BASE_PATH`. Workers can load cross origin scripts.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -312,15 +318,26 @@ function createWorker(
...
+  // same-origin when `assetPrefix` is a cross-origin CDN. Module chunks loaded
+  // inside the worker can be cross-origin, so they always use CHUNK_BASE_PATH.
+  const entrypointBasePath =
+    typeof WORKER_BASE_PATH === 'string' && WORKER_BASE_PATH.length > 0
+      ? WORKER_BASE_PATH
+      : CHUNK_BASE_PATH
+
   const chunkUrls = moduleChunks
-    .map((chunk) => getChunkRelativeUrl(chunk))
+    .map((chunk) => getChunkRelativeUrl(chunk, CHUNK_BASE_PATH))
```
sokra:
```suggestion
    .map((chunk) => getChunkRelativeUrl(chunk))
```

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -312,15 +318,26 @@ function createWorker(
 ): Worker {
   const isSharedWorker = WorkerConstructor.name === 'SharedWorker'
 
+  // `WORKER_BASE_PATH` overrides only the entrypoint URL to keep `new Worker(...)`
+  // same-origin when `assetPrefix` is a cross-origin CDN. Module chunks loaded
+  // inside the worker can be cross-origin, so they always use CHUNK_BASE_PATH.
+  const entrypointBasePath =
+    typeof WORKER_BASE_PATH === 'string' && WORKER_BASE_PATH.length > 0
+      ? WORKER_BASE_PATH
+      : CHUNK_BASE_PATH
```
sokra:
```suggestion
  const entrypointBasePath = WORKER_BASE_PATH || CHUNK_BASE_PATH
```

That's a bit shorter

sokra: WORKER_BASE_PATH will be an empty string in the default case

sokra (approved):

PR #93280, opened 2026-04-27
Apply sourcemaps by default during prerender in `next build`

eps1lon:
This config was previously only enabled for Cache Components.

We don't expect any meaningful impact for `next build` performance. It's been on with Cache Components for a while so there's less of a chance for surprises.

Apps with default configs using Webpack will see no differenced. Those apps would need to enable `experimental.serverSourceMaps`

### DIFF
--- a/test/e2e/app-dir/root-suspense-dynamic/fixtures/default/next.config.js
+++ b/test/e2e/app-dir/root-suspense-dynamic/fixtures/default/next.config.js
@@ -2,8 +2,6 @@
  * @type {import('next').NextConfig}
  */
 const nextConfig = {
-  enablePrerenderSourceMaps: false,
-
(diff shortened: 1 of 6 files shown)


test/e2e/app-dir/root-suspense-dynamic/fixtures/default/next.config.js
```diff
@@ -2,8 +2,6 @@
  * @type {import('next').NextConfig}
  */
 const nextConfig = {
-  enablePrerenderSourceMaps: false,
```
eps1lon, line 5: Unclear what this was used for. Test works just fine with `enablePrerenderSourceMaps`

unstubbable (approved):

PR #93281, opened 2026-04-27
Stop ignore-listing original stack frames if source maps are disabled

eps1lon:
If source maps are disabled, we're most likely running in a prod server where sourcemapping happens lazily when logs are inspected. Ignore-listing generated frames would forever hide n_m frames. 

It also avoids doing a bunch of work that just won't have any effect.

As a follow-up, I'll experiment with showing all frames during next-build if all original frames are ignore-listed since that's a one-shot process. For next-dev, we'll continue showing no frames if everything is ignore-listed since you can attach a debugger to get a high-fidelity UI to see ignore-listed frames.


eps1lon:
Closing in favor of https://github.com/vercel/next.js/pull/92590. Huge PITA to get `static name` to work

PR #93285, opened 2026-04-27
[ci]: trigger signed release commit via API

ztanner:
Stacked on https://github.com/vercel/next.js/pull/93245

Commit signing is required for anything that lands on `canary`. Our previous workflow of using a PAT to push a commit no longer works, since that would have been an unsigned commit.

In #93245 we switched to an app token for release workflow steps.  This continues by:

- Telling Lerna to bump packages but not commit
- Creating a signed commit with the staged changes via GitHub's API
- Then running the publish flow

Test Plan:
Dry run [here](https://github.com/vercel/next.js/actions/runs/25001287978/job/73211826479)
Commit [here](https://github.com/vercel/next.js/commit/3a8456a62b56615e2f8c1d1e6e7b4e14ce89e078)

### DIFF
--- a/scripts/release-github-api.js
+++ b/scripts/release-github-api.js
@@ -226,10 +226,21 @@ async function alignLocalBranchWithGitHubReleaseCommit(
   tagName,
   commitSha
 ) {
-  await execa('git', ['tag', '-d', tagName], {
-    stdio: 'inherit',
-    reject: false,
-  })
+  const tagExists = await execa(
+    'git',
+    ['show-ref', '--verify', '--quiet', `refs/tags/${tagName}`],
+    {
+      stdio: 'ignore',
+      reject: false,
+    }
+  )
+
+  if (tagExists.exitCode === 0) {
+    await git(['tag', '-d', tagName])
+  } else {
+    console.log(`Local tag ${tagName} does not exist; skipping delete`)
+  }
+
   await git([
     'fetch',
     'origin',
--- a/scripts/release-github-api.js
+++ b/scripts/release-github-api.js
@@ -1,0 +1,338 @@
+// @ts-check
+
+const execa = require('execa')
+const fs = require('fs/promises')
+
+const REPO_API_PATH = '/repos/vercel/next.js'
+
+async function git(args, options = {}) {
+  const { captureOutput = false, ...execaOptions } = options
+  const { stdout } = await execa('git', args, {
+    stdio: captureOutput ? 'pipe' : 'inherit',
+    ...execaOptions,
+  })
+
+  return typeof stdout === 'string' ? stdout.trim() : stdout
+}
+
+/**
+ * Call the GitHub REST API with the release app token and include response
+ * bodies in thrown errors so workflow failures show actionable details.
+ */
+async function githubRequest(token, method, path, body) {
+  const response = await fetch(`https://api.github.com${path}`, {
+    method,
+    headers: {
+      Accept: 'application/vnd.github+json',
+      Authorization: `Bearer ${token}`,
+      'X-GitHub-Api-Version': '2022-11-28',
+      ...(body ? { 'Content-Type': 'application/json' } : {}),
+    },
+    body: body ? JSON.stringify(body) : undefined,
+  })
+
+  if (!response.ok) {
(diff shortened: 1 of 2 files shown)


scripts/release-github-api.js
```diff
@@ -0,0 +1,349 @@
...
+ * unusual file names do not affect parsing.
+ */
+async function getChangedFiles(baseSha, headSha) {
+  const stdout = await git(
+    ['diff-tree', '-r', '--name-only', '--no-renames', '-z', baseSha, headSha],
+    { captureOutput: true, encoding: 'utf8' }
+  )
+
+  return String(stdout).split('\0').filter(Boolean)
+}
```
eps1lon, line 107: I know we do this in a lot of places but want to take note of this: This breaks once diffs get big. The robust way is to stream the stdout from diff-tree and then work on the chunks. This becomes a async generator which yields each changed file e.g. https://github.com/eps1lon/eslint-focus/blob/4c587001f3ff621d1a48db6c61f379129776cc97/bin/eslint-focus.js#L15-L52

scripts/start-release.js
```diff
@@ -70,26 +71,46 @@ async function main() {
...
+    const releaseChild = execa(
+      'pnpm',
+      ['release', '--pre', '--skip-questions', '--show-url'],
+      {
+        stdio: 'pipe',
+      }
+    )
+
+    releaseChild.stdout?.pipe(process.stdout)
+    releaseChild.stderr?.pipe(process.stderr)
```
eps1lon: Does execa not support `stdio: ['inherit', 'inherit', 'ignore']` (https://nodejs.org/api/child_process.html#optionsstdio) or is that different? The optional chain just seems unexpected to me.

scripts/release-github-api.js
```diff
@@ -0,0 +1,349 @@
...
+
+  return blob.sha
+}
+
+/**
+ * Build a GitHub tree matching the local Lerna release commit, creating new
+ * blob objects for changed files in parallel while preserving deletions and
+ * submodules.
+ */
+async function createTreeFromLocalCommit(token, baseSha, localReleaseSha) {
```
eps1lon: Don't tell this Hendrik but I think an object would be safer. there are two SHAs in here and the parameter order is quite important.

scripts/release-github-api.js
```diff
@@ -0,0 +1,349 @@
...
+      }
+
+      if (treeEntry.type === 'commit') {
+        return {
+          path: filePath,
+          mode: treeEntry.mode,
+          type: treeEntry.type,
+          sha: treeEntry.sha,
+        }
+      }
```
eps1lon, line 194: Can this ever be a commit if we're passing a file?

eps1lon (approved):
I'm not super comfortable with the manual tree creation but then we're only doing this once or twice a day on average for changes to package.json. So even if that's not hyper-optimized, it won't significantly degrade git operations on the repo. And even if, we don't really have a choice if we want to sign commits.

scripts/release-github-api.js
ztanner: Yeah, agreed. This release commit should be bounded to the files Lerna changes, so I’m comfortable keeping this simple for now, but streaming stdout would be the more robust shape if this ever starts handling larger diffs

scripts/start-release.js
ztanner: Good call. I don’t need to inspect the output here, so inherited stdio would be simpler. I’ll switch that.

scripts/release-github-api.js
ztanner: 👍  I’ll switch this to named object args

scripts/release-github-api.js
ztanner: Only for submodules. `git ls-tree <commit> -- <path>` can return `type: commit` when the path is a submodule pointer. I don’t expect the release bump to touch one so this is probably unnecessary, but preserving it keeps the tree recreation faithful to Git’s object model

eps1lon: makes sense. Thanks for checking!

eps1lon (approved):

PR #92638, opened 2026-04-10
Redesign blocking route dev overlay and build errors

aurorascharff:
### What?

Redesigns the blocking-route error overlay for instant navigation errors with a distinct "Instant" overlay path, visual technique cards, and updated error wording framed around navigation impact.

### Why?

The current overlay dumps every possible cause and fix in one block of text. The new design is friendlier — amber "Instant" badge, a short headline framed around navigation, and responsive code snippet cards showing each fix pattern.

### Demo

- **Runtime template** (e.g. `cookies()`): https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/26-cookies-ssr-no-instant
- **Dynamic template** (e.g. uncached `fetch`): https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/27-fetch-ssr-no-instant

### How?

**Overlay**
- New early-return path in `errors.tsx` for `blocking-route` errors without refinement — renders `InstantRuntimeError` with CodeFrame → description → technique cards → CallStack → ErrorCause
- `InstantGuidance` component with responsive CSS grid of fix technique cards (3 per variant)
- Color-coded cards with colored borders and matching highlight text (blue, purple, red)
- "Make route params static" card (runtime only) has a dashed border indicating it's conditional

**Build & CLI messages**
- Build output messages extracted into `blocking-route-messages.ts` and deduplicated across `dynamic-rendering.ts`
- `dynamicOrRuntimeBodyMessage` added for build-time static validation where the specific cause can't be pinpointed — lists all APIs (`fetch(...)`, `cookies()`, `headers()`, `params`, `searchParams`, `connection()`)
- `isRuntimeVariant()` replaces the old `includes('cookies()')` heuristic which broke because both templates mention `cookies()`
- `logBuildDebugHint()` extracted and shared between `logDisallowedDynamicError` and instant validation — adds "run `next dev`" and "`next build --debug-prerender`" hints to instant validation build output

Results:
Google Chrome 2026-04-17 16 37 00
Google Chrome 2026-04-17 16 36 37

PR #93287, opened 2026-04-27
Extend instant error overlay to metadata, viewport, and sync IO errors

aurorascharff:
### What?

Extends the card-based instant error overlay (#92638) to metadata, viewport, and sync IO errors, and updates the matching build/CLI messages to a consistent structured format.

### Why?

After the blocking-route redesign, metadata, viewport, and sync IO errors still used the old prose format with no visual fix guidance.

### Demo

- Metadata: [runtime](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/12-cookies-in-metadata) · [dynamic](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/13-fetch-in-metadata)
- Viewport: [runtime](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/14-cookies-in-viewport) · [dynamic](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/15-fetch-in-viewport)
- Sync IO: [Math.random()](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/38-math-random-no-instant) · [Date.now()](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/39-date-now-no-instant) · [crypto.randomUUID()](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/40-crypto-random-no-instant)
- Sync IO in Client: [Math.random()](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/44-client-math-random-no-suspense) · [Date.now()](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/43-client-date-no-suspense) · [crypto.randomUUID()](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/45-client-crypto-no-suspense)
- [Fix overview (all cards)](https://error-messages-overhaul-ibsl.labs.vercel.dev/fix-overview)

### How?

**Overlay**

- New `dynamic-metadata`, `dynamic-viewport`, and `sync-io` error types in `errors.tsx`, all rendered through `InstantRuntimeError` with kind-specific fix cards
- `isSyncIOError()` detects sync IO via the docs URL pattern, mirroring `isRuntimeVariant()`

**Build & CLI messages**

- Headlines aligned to "Next.js encountered..." across all error families
- `blocking-route-messages.ts`: metadata/viewport errors restructured from prose to "Ways to fix this:" bullets
- Sync IO messages extracted into `sync-io-messages.ts`, mirroring `blocking-route-messages.ts`. Helpers renamed to `createSyncIOError` / `createSyncIORuntimeError` / `createSyncIOClientError` to match the `create*Error` pattern


packages/next/src/server/app-render/dynamic-rendering.ts
```diff
@@ -1248,7 +1248,9 @@ export function throwIfDisallowedDynamic(
     // you need to opt into that by adding a Suspense boundary above the body
     // to indicate your are ok with fully dynamic rendering.
     if (dynamicValidation.hasDynamicViewport) {
-      console.error(disallowedDynamicViewportMessage(workStore.route))
+      console.error(
```
aurorascharff, line 1251: @lubieowoce i used the same pattern as we did together  here

packages/next/src/next-devtools/dev-overlay/container/errors.tsx
```diff
@@ -611,35 +268,78 @@ function isRuntimeVariant(message: string): boolean {
...
+  'crypto.randomUUID()',
+]
+
+const SYNC_IO_DOCS_PATTERN =
+  /https:\/\/nextjs\.org\/docs\/messages\/next-prerender-(?:runtime-)?(random|current-time|crypto)(?!-)/
+
+// Matches sync IO errors from `createSyncIOError` and
+// `createSyncIORuntimeError`, plus the "needs to bail out of prerendering"
+// abort/throw reason from `dynamic-rendering.ts`.
+function isSyncIOError(message: string): boolean {
```
aurorascharff, line 306: @lubieowoce can you also verify this pattern? same issue as with the dynamic/runtime in the previous pr, not sure how to classify.

packages/next/src/server/node-environment-extensions/io-utils.tsx
```diff
@@ -55,22 +87,10 @@ export function io(expression: string, type: ApiType) {
...
-            message = `Route "${workStore.route}" used ${expression} inside a Client Component without a Suspense boundary above it. See more info here: https://nextjs.org/docs/messages/next-prerender-crypto-client`
-            break
-          default:
-            throw new InvariantError(
-              'Unknown expression type in abortOnSynchronousPlatformIOAccess.'
-            )
-        }
+        const docsUrl = SYNC_IO_CLIENT_DOCS[type]
+        const message =
+          `Route "${workStore.route}" used \`${expression}\` inside a Client Component without a Suspense boundary above it. ` +
```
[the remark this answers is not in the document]
aurorascharff: Resolved as part of the sync IO message extraction in this PR. The double-backtick was a `prerender-client` symptom of inlining the message in `io-utils.tsx` while callers passed already-quoted expressions. The new helper `createSyncIOClientError` in `packages/next/src/server/app-render/sync-io-messages.ts` builds the message with `${expression}` (no extra backticks), so the rendered output is single-backtick on both server and client paths.

packages/next/src/next-devtools/dev-overlay/components/instant/instant-guidance.tsx
```diff
@@ -110,16 +55,27 @@ function CardGrid({ cards }: { cards: FixCard[] }) {
...
+  explanation?: string
+  cause?: string
 }) {
-  const cards = variant === 'navigation' ? dynamicCards : runtimeCards
+  const cards = getCards(kind, variant, cause)
+  const docsUrl =
+    kind === 'sync-io' && cause
+      ? SYNC_IO_DOCS[cause] || DOCS_URLS[kind]
+      : DOCS_URLS[kind]
+  const defaultExplanation = explanation || EXPLANATIONS[kind]
```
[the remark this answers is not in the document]
aurorascharff: Good catch — fixed. `InstantGuidance` now conditionally omits the `Learn more` link when `docsUrl` is empty, and skips the whole explanation paragraph when neither an explanation nor a docs URL is available, so we never render `href=""`. Per-API URLs in `SYNC_IO_DOCS` cover all current sync IO causes; the empty `'sync-io'` entry in `DOCS_URLS` is now only used as a fallback for unknown causes (which now degrades gracefully instead of producing a broken link).

packages/next/src/next-devtools/dev-overlay/container/errors.tsx
```diff
@@ -611,35 +268,78 @@ function isRuntimeVariant(message: string): boolean {
...
+  return (
+    message.includes('needs to bail out of prerendering') ||
+    SYNC_IO_DOCS_PATTERN.test(message)
+  )
+}
```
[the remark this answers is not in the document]
aurorascharff: The `(?!-)` lookahead is intentional. Client sync IO errors (`prerender-client`) keep the legacy concise message until a dedicated overlay design lands — see Step 7 in the planning doc, where this is tracked as a separate workstream with open questions on whether the recommendation should be "add Suspense", "move the read", or "avoid client sync IO". Routing them through the existing `'sync-io'` cards now would imply the previous "Render on the client" suggestion was wrong, which we want to handle as a deliberate follow-up.

Added a code comment above the regex in this PR explaining the intent and noting that the lookahead should be dropped once the client-variant cards ship.

packages/next/src/server/app-render/sync-io-messages.ts
```diff
@@ -0,0 +1,72 @@
...
+      `Without a prior data access, Next.js doesn't know whether to prerender this value or compute it on each request.\n\n` +
+      `Ways to fix this:\n` +
+      `  - Render at request time by adding a dynamic data access (e.g. \`await connection()\`) before this call\n` +
+      `  - Prerender and cache the value with \`"use cache"\`\n` +
+      `  - Render the value on the client with \`"use client"\`\n` +
+      elapsedTimeBullet(type) +
+      `\n` +
+      `Learn more: ${SYNC_IO_RUNTIME_DOCS[type]}`
+  )
+}
```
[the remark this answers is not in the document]
aurorascharff: Refactored — extracted a `createSyncIOErrorImpl(route, expression, type, docsUrl)` internal helper. The two public helpers now just delegate with their respective docs URL map, so the message template lives in exactly one `new Error(...)` call. Future text changes only need to touch one place.

packages/next/errors.json
```diff
@@ -1178,5 +1178,52 @@
...
   "1179": "Invariant: %s This is a bug in Next.js.",
-  "1180": "Cookies can only be modified in a Server Action or Route Handler. Read more: https://nextjs.org/docs/app/api-reference/functions/cookies#options"
+  "1180": "Cookies can only be modified in a Server Action or Route Handler. Read more: https://nextjs.org/docs/app/api-reference/functions/cookies#options",
+  "1181": "Route \"%s\": Next.js encountered uncached or runtime data during the initial render.\\n\\n\\`fetch(...)\\`, \\`cookies()\\`, \\`headers()\\`, \\`params\\`, \\`searchParams\\`, or \\`connection()\\` accessed outside of \\`<Suspense>\\` prevents the route from being prerendered, blocking navigation and leading to a slower user experience.\\n\\nWays to fix this:\\n  - Cache the data access with \\`\"use cache\"\\`\\n  - Move the data access into a child component within a \\`<Suspense>\\` boundary\\n  - Use \\`generateStaticParams\\` to make route params static\\n  - Set \\`export const instant = false\\` to allow a blocking route\\n\\nLearn more: https://nextjs.org/docs/messages/blocking-route",
```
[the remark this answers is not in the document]
aurorascharff: `errors.json` is append-only by design (see `crates/next-error-code-swc-plugin/README.md`: _"Append-only, increment-based mapping of error codes to messages"_). Each new template gets a new code; existing entries stay as historical snapshots so anonymous error-code reporting from older releases keeps mapping back to the message that was emitted at the time. Editing `1181` in place would break that contract.

The current emitted code from `createDynamicOrRuntimeBodyError` is one of the higher entries (1218–1227), all of which use the new "Provide a placeholder with `<Suspense fallback={...}>`" wording you cite. The intermediate codes are stale by construction and can be ignored.

packages/next/errors.json
unstubbable, line 1: Don't edit preexisting error messages in this file. Let them be regenerated with new codes. This allows us to better track the error feedback, i.e. ideally an improved error message (identified by its new code) receives more positive feedback than the previous message (with the old code).

aurorascharff: Is this for the "was this helpful" button?? I think its being removed. but yep, will do!

unstubbable: Yes, it is.

unstubbable (approved):

packages/next/src/server/app-render/sync-io-messages.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export function createSyncIOClientError(
+  route: string,
+  expression: string,
+  type: SyncIOApiType
+): Error {
+  return new Error(
+    `Route "${route}": Next.js encountered ${expression} in a Client Component.\n\n` +
+      `This value would be evaluated during the prerender and fixed at build time, instead of recomputed on each visit.\n\n` +
```
gnoff, line 74: It's not build time per se, it is prerendering. We can and do prerender at runtime through revalidation.

Maybe 

> This expression will produce a unique value every time it is evaluated so it cannot be prerendered in HTML.

packages/next/src/server/app-render/sync-io-messages.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+function createSyncIOErrorImpl(
+  route: string,
+  expression: string,
+  type: SyncIOApiType,
+  docsUrl: string
+): Error {
+  return new Error(
+    `Route "${route}": Next.js encountered ${expression} without an explicit rendering intent.\n\n` +
```
gnoff, line 34: don't love `an explicit rendering intent.`

WDYT about the framing below of

> Next.js encountered ${expression} while prerendering. This expression produces a unique value on every evaluation so it cannot be prerendered directly. Ways to fix...

packages/next/src/server/app-render/sync-io-messages.ts
```diff
@@ -0,0 +1,81 @@
...
+      `This value can change between renders, so it must be either prerendered or computed later.\n\n` +
+      `Ways to fix this:\n` +
+      `  - Render at request time by adding a dynamic data access (e.g. \`await connection()\`) before this call\n` +
```
gnoff, line 37: should we talk about `io()`? for sync IO it's probably the better call in cases where the sync IO isn't actually part of the rendered output (i.e. incidental new Date() for telemetry/logging) so you can't say one way or another whether the containing scope is necessarily dynamic. Whereas something like relative time shouldn't be prerenderable so connection() would be more appropriate since you don't want someone to accidentally use it in a "use cache"

gnoff (approved):
I have nits but it might make sense to address in a followup vs churning the massive PR

PR #93301, opened 2026-04-27
Support configuring a default instant validation level 

gnoff:
Adds `experimental.instantInsights.validationLevel` so an app can opt into instant validation for all routes by default.

There are two public levels
* `manual-warning`: currently the default. this mode won't enable instant validation except where you add `export const unstable_instant = true` into your Layouts, Pages, and Default files
* `warning`: This mode will enable instant validation for all Page and Default files. You can override this with `export const unstable_instant = false` in Layouts, Pages, and Default files to tune what navigations are validated.

In addition this change disables build time validation.

There is a now undocumented way to get back to running build validation with
* `experimental-error`
* `experimental-manual-error`

The build validation isn't ready for users because we need to overhaul the setup where you provide cookies and other sample values so we can run validation without a live session like we do in dev.

To further support this the two remaining object properties that are also not ready for active use, `samples` and `from` are being updated to `unstable_samples` and `unstable_from` respectively.

We will continue to iterate on build time validation but may remove these APIs or alter their semantics as we figure out a more ergonomic way to support this setup.

In anticipation of multiple levels the object form of `instant` can be provided a level value `warning` or `experimental-error`. this latter option is similarly undocumented so documented behavior will only cover `warning` which is generally a synonym of true unless you are activating error by default

### DIFF
--- a/packages/next/src/build/static-paths/app.ts
+++ b/packages/next/src/build/static-paths/app.ts
@@ -868,6 +868,10 @@ export async function buildAppStaticPaths({
       staticPageGenerationTimeout,
       supportsDynamicResponse: true,
       cacheComponents,
+      // generateStaticParams evaluation doesn't render pages, so instant
+      // validation never runs here. The level value is irrelevant.
+      // TODO: remove validationLevel and other global config out of renderOpts
+      validationLevel: 'warning',
       experimental: {
         authInterrupts,
         useCacheTimeout,
--- a/packages/next/src/server/app-render/instant-validation/instant-config.tsx
+++ b/packages/next/src/server/app-render/instant-validation/instant-config.tsx
@@ -1,6 +1,15 @@
 import { getLayoutOrPageModule } from '../../lib/app-dir-module'
 import type { LoaderTree } from '../../lib/app-dir-module'
 import { parseLoaderTree } from '../../../shared/lib/router/utils/parse-loader-tree'
+import {
+  PAGE_SEGMENT_KEY,
+  DEFAULT_SEGMENT_KEY,
+} from '../../../shared/lib/segment'
+import {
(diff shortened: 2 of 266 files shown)


packages/next/src/build/static-paths/app.ts
```diff
@@ -866,6 +866,9 @@ export async function buildAppStaticPaths({
       staticPageGenerationTimeout,
       supportsDynamicResponse: true,
       cacheComponents,
+      // generateStaticParams evaluation doesn't render pages, so instant
+      // validation never runs here. The level value is irrelevant.
+      defaultValidationLevel: 'disabled',
```
gnoff: obv codesmell. the plan for config is to make it a global. semantically config must be static for the life of the process and so we don't need to force plumbing through request contexts. Not trying to reinvent the wheel here in this PR however...

packages/next/src/server/app-render/instant-validation/instant-config.tsx
```diff
@@ -68,40 +108,98 @@ export async function isPageAllowedToBlock(tree: LoaderTree): Promise<boolean> {
...
-  // (If there's only `false`, there's no need to run validation).
-  // If any segment has `unstable_disableValidation`, we skip validation for the whole tree.
   let needsValidation = false
-  for (const { config } of segments) {
-    if (config === true) {
-      needsValidation = true
-    } else if (typeof config === 'object') {
+  let disabled = false
+
+  async function visit(tree: LoaderTree): Promise<void> {
```
gnoff, line 168: The bulk of the logic changes are here. semantically a default validation level of warning or error sets a virtual config on any Page or Default segment if an actual config isn't exported there

acdlite (approved):

PR #91999, opened 2026-03-27
Stop server HMR from invalidating past the nearest layout segment boundary

wbinnssmith:
### What?

When a page or nested component changes during `next dev`, server HMR was re-evaluating all ancestor layout modules in the loader-tree — all the way up to the root layout. This meant a small change to a leaf page was causing the root layout module to be re-initialized on every update.

This PR limits server HMR invalidation to the nearest layout segment boundary by registering each layout as an HMR boundary in the generated loader-tree module.

### Why?

Layout modules should be stable across page-level HMR updates. Re-evaluating them is wasteful, can cause unexpected side effects (e.g. resetting module-level state, re-running expensive initialization), and defeats the purpose of layout caching. A layout should only re-evaluate when it (or one of its own dependencies) actually changes.

### How?

In the generated loader-tree module for each app-router page entry, immediately after each layout chunk's `require()`, we now emit:

```js
import.meta.turbopackHot.accept("MODULE_<n>");
```

This registers the layout's loader-tree module as an HMR boundary for that layout chunk. When a page (child) changes, HMR propagation stops at the nearest ancestor layout rather than bubbling all the way to the page entry and re-evaluating all layouts in the chain.

**Implementation details:**

- Added `server_hmr: bool` field to `BaseLoaderTreeBuilder` and plumbed it from `ProjectOptions` → `get_app_page_entry` → `AppPageLoaderTreeModule::build` → `BaseLoaderTreeBuilder`.
- Added a `server_hmr()` accessor on `Project` (via turbo-tasks function) so `AppEndpoint::app_page_entry` can read it without accessing the private `Project` struct fields directly.
- The `accept()` call is only emitted when `server_hmr` is `true` (i.e. `next dev` with server fast refresh enabled). Build mode never sets `serverHmr`, which maps to `false` via `unwrap_or(false)` in the NAPI binding layer — so production builds are unaffected.

**Test:**

Added a new `layout segment boundary` test in `test/development/app-dir/server-hmr/server-hmr.test.ts` that:
1. Fetches `/layout-boundary` and captures the root layout's module-eval timestamp (a `Date.now()` captured at module initialization, rendered into a hidden element).
2. Patches `app/layout-boundary/page.tsx` to trigger a server HMR update.
3. Waits for the page content to reflect the update, then asserts the root layout's eval timestamp is unchanged — confirming the layout module was not re-evaluated.

The test fixture uses two nested layouts (root layout + `app/layout-boundary/layout.tsx`) above the page.

PR #93322, opened 2026-04-28
Stop server HMR from invalidating past the loader tree

wbinnssmith:
Rebase of #91999 — the original branch could not be updated due to GitHub's commit signing requirements for force pushes to existing branch refs.

Add `import.meta.turbopackHot.accept(dependencyPath)` calls in the loader-tree module after each layout chunk's require(). This registers each layout as an HMR boundary, preventing page changes from propagating upward and re-evaluating parent layouts.

Thread `server_hmr` from Project through AppPageLoaderTreeModule::build and BaseLoaderTreeBuilder so the accept() call is only emitted when server HMR is enabled.

Add a test that verifies the root layout module is not re-evaluated when a nested page changes.


turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/hmr-runtime.ts
```diff
@@ -165,9 +165,15 @@ function getAffectedModuleEffects(
...
-      (hotState.selfAccepted && !hotState.selfInvalidated)
+      !module ||
+      // In server HMR mode (autoAcceptRootModules=true), individual module
+      // self-accepts do not stop propagation. The real accept boundary is
+      // always the layout tree's acceptedDependencies. Stopping here would
+      // leave consuming modules (e.g. page.tsx) cached with stale namespace
+      // references to the old module instance.
+      (!autoAcceptRootModules &&
+        hotState.selfAccepted &&
+        !hotState.selfInvalidated)
```
sokra: This change is weird.

>              Stopping here would
>       // leave consuming modules (e.g. page.tsx) cached with stale namespace
>       // references to the old module instance.

Self accept means the module updated its own exports themselves.
So a self accepted module doesn't have a new namespace object that could be stale.

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/hmr-runtime.ts
```diff
@@ -182,6 +188,10 @@ function getAffectedModuleEffects(
 
     if (runtimeModules.has(moduleId)) {
       if (autoAcceptRootModules) {
+        // The runtime module was added to outdatedModules above before we knew
+        // it was a runtime boundary. Remove it so disposePhase doesn't evict
+        // the runtime module, which would break the entire module system.
+        outdatedModules.delete(moduleId)
         continue
       }
```
sokra:
```suggestion
```

I think this whole `if (autoAcceptRootModules) continue` is not correct.
It should push the runtime module to the queue so it can be accepted in line 149 above

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/hmr-runtime.ts
sokra:
```suggestion
```

This code is just not needed. It's at the end of the loop anyway, just doing nothing will `continue`

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/hmr-runtime.ts
```diff
@@ -572,6 +583,11 @@ function disposePhase(
         if (idx >= 0) {
           module.children.splice(idx, 1)
         }
+        // Dispose and evict the accepted dependency so the new factory is
+        // used when it is next required (either by the accept callback or
+        // by the next incoming request).
+        disposeModule(dep, 'replace')
+        delete devModuleCache[dep]
```
sokra:
```suggestion
```

This is handled by `outdatedModules` in line 556.

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/hmr-runtime.ts
sokra:
```suggestion
        continue
```

An early return "accepted" is not correct. It need to check other items from the queue. Only if all of them are accepted it should return accepted at the end of this method.

This early return will also cause bugs as some modules are not correctly marked as outdated.

crates/next-core/src/base_loader_tree.rs
```diff
@@ -96,17 +99,34 @@ impl BaseLoaderTreeBuilder {
...
-                r#"
-                const {} = () => require(/*turbopackChunkingType: shared*/"MODULE_{}");
-                "#,
-                identifier,
-                i
-            )
-            .into(),
+        let require_line = formatdoc!(
+            r#"
+            const {} = () => require(/*turbopackChunkingType: shared*/"MODULE_{}");
```
lukesandberg: nit: can we write `const {identifier}` and `MODULE_{i}` does formatdoc! support that?

wbinnssmith: yeah, looks like it 👍

PR #93331, opened 2026-04-28
feat(turbopack): support `false` values for `resolveAlias` config

sokra:
### What?

Adds support for `false` as a value in `experimental.turbopack.resolveAlias`, which resolves the aliased module to an empty stub — matching webpack's long-standing behavior.

```js
// next.config.js
module.exports = {
  experimental: {
    turbopack: {
      resolveAlias: {
        'some-server-only-module': false,
      },
    },
  },
}
```

With this alias in place:
- `import * as ns from 'some-server-only-module'` → `ns` is `{}`
- `import { foo } from 'some-server-only-module'` → `foo` is `undefined`
- `import def from 'some-server-only-module'` → `def` is `undefined`
- `await import('some-server-only-module')` → resolves to `{}`
- `require('some-server-only-module')` → returns `{}`

### Why?

Webpack supports `false` as an alias target to stub out modules (e.g. to exclude Node-only packages from client/edge bundles). Turbopack did not, causing migration friction for projects that rely on this pattern.

### How?

Threads the `false` sentinel through the full resolution pipeline as a new `Empty` variant:

```
false (config)
  → SubpathValue::Empty      (next-core import map)
  → ImportMapping::Empty     (turbopack-core remap)
  → ReplacedImportMapping::Empty
  → ResolveResultItem::Empty
  → ModuleResolveResultItem::Empty
  → ReferencedAsset::Empty   (ESM static imports)
  → SinglePatternMapping::Ignored  (CJS require / dynamic import)
```

For static ESM imports, `ReferencedAsset::Empty` is kept as a distinct variant so that `binding.rs` can distinguish a namespace import (yields `{}`) from a named/default import (yields `undefined`).

For CJS `require()` and dynamic `import()`, `ModuleResolveResultItem::Empty` maps to the existing `SinglePatternMapping::Ignored`, which already generates the correct `{}` / `Promise.resolve({})` stubs — so no new code-generation path was needed there (the redundant `Empty` variant introduced during development was removed in a follow-up cleanup commit).

**Files changed:**
- `packages/next/src/server/config-shared.ts` — add `false` to `resolveAlias` type
- `packages/next/src/server/config-schema.ts` — add `z.literal(false)` to schema
- `crates/next-core/src/next_import_map.rs` — map `false` → `SubpathValue::Empty`
- `turbopack/crates/turbopack-core/src/resolve/remap.rs` — handle `ImportMapping::Empty`
- `turbopack/crates/turbopack-core/src/resolve/mod.rs` — propagate `ModuleResolveResultItem::Empty`
- `turbopack/crates/turbopack-ecmascript/src/references/esm/base.rs` — `ReferencedAsset::Empty`
- `turbopack/crates/turbopack-ecmascript/src/references/esm/binding.rs` — empty-module binding code-gen
- `turbopack/crates/turbopack-ecmascript/src/references/esm/url.rs` — handle Empty in URL refs
- `turbopack/crates/turbopack-ecmascript/src/references/async_module.rs` — skip Empty in async module analysis
- `turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs` — map Empty → Ignored

**Tests:** `test/e2e/turbopack-resolve-alias-false/` covers all five import styles (namespace, named, default, dynamic, CJS).


test/e2e/app-dir/turbopack-resolve-alias-false/turbopack-resolve-alias-false.test.ts
```diff
@@ -0,0 +1,44 @@
+import { nextTestSetup } from 'e2e-utils'
+
+// `resolveAlias: false` is a Turbopack-only feature. The `turbopack.resolveAlias`
+// config is not read by webpack.
+;(process.env.IS_TURBOPACK_TEST ? describe : describe.skip)(
```
mischnic, line 5: Add a `webpack.config.resolve.alias` in next.config.js with the same rules as the Turbopack `resolveAlias` to run this test in both cases and ensure that they indeed behave the same

turbopack/crates/turbopack-core/src/resolve/remap.rs
```diff
@@ -58,6 +58,11 @@ pub enum SubpathValue {
     /// An excluded subpath, defined with `"path": null`, prevents importing
     /// this subpath.
     Excluded,
+
+    /// An empty subpath, defined with `"path": false`, resolves to an empty
+    /// module (empty object for namespace/CJS imports, `undefined` for named
+    /// bindings).
+    Empty,
```
mischnic, line 65: How is empty different from excluded though? Don't both cause the module to not be imported?

sokra: Excluded will throw at runtime, while empty return an empty object/undefined bindings

mischnic (approved):

PR #93334, opened 2026-04-28
Turbopack: add all keys to dynamic exports before sealing the object

andrewimm:
User surfaced an issue where a barrel file had `export * from` a chain of CJS files. In this barrel-file-exports + dynamic-exports case, the Turbopack runtime wraps `module.exports` in a `Proxy` whose `ownKeys` trap synthesizes keys from the re-exported namespace. The constructed proxy conflicts with `Object.seal`, which `esm()` calls to ensure the namespace is immutable (per spec).
When you try to execute the code in the browser or for build, you get a `TypeError: 'ownKeys' on proxy: trap returned extra keys but proxy target is non-extensible`

I've constructed a test case that fails today. Nothing about this code is "wrong" -- a CJS module gets imported and re-exported, and `Page.tsx` does an `import *` from that file. But `ownKeys` returns something different from the defined properties, which causes the JS engine to trigger an exception.

To ensure that the set of properties exactly matches the output of `ownKeys`, we make sure all of those properties exist when the object is `seal`ed. With this change in place, the test passes.


mischnic:
Can we instead add that test to turbopack/crates/turbopack-tests/tests/execution/turbopack/ ?

andrewimm:
PR was a bit premature -- I can't get builds running locally due to tsc issues, so I hoped to get some insight from running tests on GH

lukesandberg:
as discussed in slack we need

* `has` handler
* `getOwnPropertyDescriptor` handler

speculative:
 * hardcode isExtensible to false and override `preventExtensions` to always return true but do nothing otherwise

andrewimm:
Original scope: re-exporting dynamic exports leads to runtime errors. Legitimate code can trigger this, and we should handle it gracefully. This is done by removing the `Object.seal()` for dynamic modules. `seal` does two things:
* Prevent extension of the object
* Set each key's `[[Configurable]]` property to false
We want this for true ESModules, but for dynamic exports we need to allow flexibility. We do this by passing the dynamic flag to the codgen, and avoiding the `seal` call in those cases. We use Proxy traps to ensure exported values can't be modified or deleted externally, while still allowing the original dynamic module to change them.

While investigating, I also found that `Object.keys(importNamespace)` would miss re-exported dynamic keys. Similarly, `'exportName' in starImports` would always return false for dynamic imports. Since we're already touching this code, I also fixed those cases.

Test cases cover the scenarios that seemed relevant to real user code, as well as the invariants we care about.

Lots of comments in the code to explain how all of these things work together, for the next person who ventures this way. Nothing is more humbling for me than trying to reason about JS Proxies...

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-utils.ts
```diff
@@ -264,6 +320,22 @@ function ensureDynamicExports(
         }
```
lukesandberg, line 320: <p class='graphite__hidden'><i>[Re: line +318]</i></p>
This is a trivially quadratic algorithm.... probably not a big deal but dumping all the keys into a set and then flattening to an array at the end would be faster for modules with lots of re-exports
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/vercel/next.js/93334?utm_source=unchanged-line-comment">Graphite</a>.</p>

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-utils.ts
```diff
@@ -264,6 +320,22 @@ function ensureDynamicExports(
...
+        if (own || prop === 'default' || prop === '__esModule') return own
+        const obj = reexportOwning(prop)
+        if (obj) {
+          // Synthetic keys don't exist on the target, so they MUST be
+          // reported as configurable. However the set/delete traps above will
+          // prevent them from actually being changed
+          return {
+            enumerable: true,
+            configurable: true,
+            get: () => Reflect.get(obj, prop),
```
lukesandberg, line 334: should we call `reexportOwning` again in here?  in case the property moves around or is deleted?

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-utils.ts
```diff
@@ -240,6 +248,22 @@ function ensureDynamicExports(
...
+          if (hasOwnProperty.call(obj, prop)) return obj
+        }
+      }
+      return undefined
+    }
+    // Modules with dynamic re-exports are not sealed by `esm()`, so the
+    // target beneath the namespace stays extensible. That is what lets the
+    // `ownKeys` and `getOwnPropertyDescriptor` traps legally report keys that
+    // exist on `reexportedObjects` but not on the target itself.
     module.exports = module.namespaceObject = new Proxy(exports, {
```
lukesandberg, line 267: do we still have the problem that if a user calls `seal` or prevent extension then it breaks the object?

do we need to override preventExtensions for that reason?

lukesandberg (approved):
this looks good and the tests are great

maybe add a test for a user calling Object.freeze on the namespace to see what happens?

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-utils.ts
andrewimm: Will implement that a separate commit. Separation of concerns.

PR #93352, opened 2026-04-29
Preserve error sublcass constructor names after mangling

eps1lon:
SWC mangles constructor names by default when minification is enabled (i.e. in prod).
For example `class ApiError extends Error {}` will end up as `class y extends Error {}`.

The problem with that is that Error constructor names appear in Node.js' default error inspection i.e. `console.log(new ApiError('message'))` will log `y [Error]: message`.
We patch Node.js error inspection to avoid that but I want to stop patching error inspection if sourcemapping is disabled (for performance reasons). It also ensure nice error inspection for other runtimes.

We could stop mangling Error subclasses to preseve the constructor name. However, this negatively impacts parsed size if Error subclasses have multiple callsites. E.g. instead of `new z(); new z()`, we'd emit `new ApiError(); new ApiError()`. For gzip size it largely doesn't matter.

So instead we enforce emitting a `static name` class property repeating the constructor name. This ensures equivalent inspection display. We could've done this automatically. Instead, we're doing an explicit opt-in. This avoids compiler magic and leaking constructor names that aren't meant to be leaked (React Flight doesn't transport Error names either in prod). It's not a strong concern since I'm not even sure that you can 100% rely on error subclass constructor name minification but better be safe than sorry.

Missing `static name` on Error subclasses are caught with a lint rule. Like any other Lint rule, this can be disabled in situations where you don't want to leak the constructor name.

---

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -401,6 +401,7 @@ export default defineConfig([
     },
     rules: {
       '@next/internal/typechecked-require': 'error',
+      '@next/internal/error-subclass-static-name': 'error',
       'jsdoc/no-types': 'error',
       'jsdoc/no-undefined-types': 'error',
     },
--- a/packages/create-next-app/create-app.ts
+++ b/packages/create-next-app/create-app.ts
@@ -23,7 +23,9 @@ import { runTypegen } from './helpers/typegen'
 import type { Bundler, TemplateMode, TemplateType } from './templates'
 import { getTemplateFile, installTemplate } from './templates'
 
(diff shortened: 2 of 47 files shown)


eps1lon:
`static name` is a PITA to get right across Webpack/Turbopack. Webpack transpiles some next library with SWC and respects `useDefineForClassFields` from tsconfig which is off by default and would emit broken code. Turbopack ignores `useDefineForClassFields` so static name works by not being on par with Webpack. I don't want to mess with SWC options based on filenames for now.

PR #92277, opened 2026-04-02
Resolve `compilerOptions.paths` in tsconfigs without `compilerOptions.baseUrl`

hamidrezahanafi:
## Summary
When `compilerOptions.baseUrl` is missing (including after TS 6 compatibility rewrites that strip `baseUrl`), Next.js fell back to the app’s `tsconfig.json` directory as the base for webpack’s `JsConfigPathsPlugin`. Path entries inherited from a workspace-level base config (e.g. `tsconfig.base.json`) are relative to **that** directory, so joins were wrong and imports like `@scope/pkg` could fail to resolve.
This change uses TypeScript 5.0+’s internal `pathsBasePath` (the directory of the config that defines the effective `paths` map) as the base when `baseUrl` is absent, falling back to the previous implicit app-dir behavior when `pathsBasePath` is not set.
## Why
- Aligns webpack path alias resolution with how TypeScript resolves `paths` when `baseUrl` is optional / removed.
- Supports monorepos that define shared `paths` in a root `tsconfig.base.json` without relying on `baseUrl` (relevant as `baseUrl` is deprecated/removed in newer TypeScript).
## Test plan
- [x] `pnpm test test/integration/typescript-paths/test/index.test.ts`

PR #93356, opened 2026-04-29
fix(typescript): set pathsBasePath after deleting baseUrl for TS6 compat

### LINKED ISSUE #93336: `getTypeScriptConfiguration`: TypeScript 6 path rewriting deletes `baseUrl` but does not update `pathsBasePath`, breaking module resolution in monorepos with shared base tsconfig

https://github.com/gardsa/my-minimal-nextjs-issue-reproduction

### To Reproduce

1. Clone the reproduction repository (linked above)
2. From the repo root, install dependencies:
   ```bash
   npm install
   ```
3. Run the build from the app directory:
   ```bash
   cd apps/my-app && npx next build
   ```
4. Observe that Turbopack/webpack compilation succeeds but the TypeScript type-check step immediately after fails:
   ```
   Type error: Cannot find module '@scope/shared-lib/index' or its corresponding type declarations.
   ```

**Expected:** `next build` completes successfully. Path aliases defined in the shared `tsconfig.base.json` resolve correctly for both webpack and the TypeScript type-checker.

**Actual:** Webpack/Turbopack compilation succeeds, but the TypeScript check step that runs immediately after fails:

```
Running TypeScript ...
Failed to type check.

./src/app/page.tsx:1:26
Type error: Cannot find module '@scope/shared-lib/index' or its corresponding type declarations.
```

This affects every path alias defined in the shared base tsconfig. Aliases defined directly in the app-level tsconfig are not affected.

```bash
Operating System:
  Platform: linux
  Arch: arm64
Binaries:
  Node: 24.14.1
  npm: 11.11.0
  Yarn: 1.22.22
Relevant Packages:
  next: 16.2.4
  react: 19.2.5
  react-dom: 19.2.5
  typescript: 6.0.3
```

TypeScript, Module Resolution

next build (local)

## Root cause

`getTypeScriptConfiguration.js` adds a TypeScript 6 compatibility block that rewrites `paths` to be relative to the app dir and deletes `baseUrl`:

```js
if (semver.gte(typescript.version, '6.0.0')) {
    const parsedBaseUrl = result.options.baseUrl;
    if (typeof parsedBaseUrl === 'string' && parsedBaseUrl.length > 0) {
        const normalizedBaseUrl = getNormalizedBaseUrlForPaths(parsedBaseUrl, tsConfigPath);
        result.options.paths = rewritePathAliasesWithoutBaseUrl(normalizedBaseUrl, result.options.paths);
        delete result.options.baseUrl;
        // ← pathsBasePath is NOT updated here
    }
}
```

For an app at `apps/my-app/`, `normalizedBaseUrl` is `'../..'`, so `"libs/shared-lib/src/*"` becomes `"../../libs/shared-lib/src/*"` — correct relative to the app dir.

After `baseUrl` is deleted, TypeScript's module resolver calls `getPathsBasePath`:

```js
// typescript/lib/typescript.js
function getPathsBasePath(options, host) {
    if (!options.paths) return undefined;
    return options.baseUrl ??
        options.pathsBasePath ||
        host.getCurrentDirectory();
}
```

TypeScript sets `pathsBasePath` to the directory of the tsconfig that *defines* the `paths` key. In a monorepo this is the **repo root** (where `tsconfig.base.json` lives), not the app directory.

TypeScript therefore resolves `"../../libs/shared-lib/src/*"` relative to the repo root:

```
path.join('/repo-root', '../../libs/shared-lib/src/foo')
// → '/libs/shared-lib/src/foo'   ← does not exist
```

**Why webpack/Turbopack works but TypeScript does not:** `load-jsconfig.js` uses `path.dirname(tsConfigPath)` (the app dir) as the implicit `baseUrl` fallback when `baseUrl` is absent. `JsConfigPathsPlugin` resolves the rewritten `../../...` paths relative to the app dir, which is correct. The TypeScript checker goes through the same `getTypeScriptConfiguration` call but uses `pathsBasePath` instead, which points at the wrong base directory.

## Proposed fix

After deleting `baseUrl`, set `pathsBasePath` to `path.dirname(tsConfigPath)` so TypeScript uses the same resolution base as webpack:

```js
result.options.paths = rewritePathAliasesWithoutBaseUrl(normalizedBaseUrl, result.options.paths);
delete result.options.baseUrl;
result.options.pathsBasePath = path.dirname(tsConfigPath); // ← add this
```

This ensures the rewritten `../../libs/...` paths are resolved from the app's tsconfig directory, producing the correct absolute path regardless of where in the extends chain `paths` were originally defined.

## Workaround

None that avoids modifying `node_modules`. Pinning to Next.js 16.1.0 is not viable because that version predates the TypeScript 6 compatibility code entirely — TypeScript 6 rejects `baseUrl` as deprecated before Next.js can influence resolution. The `getTypeScriptConfiguration.js` compat code in 16.2.x is the correct approach; it is just missing the one-line `pathsBasePath` update.

sleitor:
## What?

When the TypeScript 6 compatibility code deletes `baseUrl` and rewrites path aliases to relative form (e.g. `../../libs/shared-lib/src/*`), TypeScript's module resolver falls back to `pathsBasePath` to determine the base directory for resolving those aliases.

In a monorepo, `paths` is typically defined in a shared base tsconfig at the repo root, so `pathsBasePath` points to the repo root. TypeScript then resolves the rewritten `../../...` paths **relative to the repo root** instead of the app dir, producing paths that don't exist.

## Why?

`JsConfigPathsPlugin` (webpack) always resolves rewritten paths from `path.dirname(tsConfigPath)` (the app dir) — making compilation succeed. The TypeScript checker goes through the same `getTypeScriptConfiguration` call but uses `pathsBasePath` instead, which points at the wrong base directory — causing type checking to fail with `Cannot find module`.

## How?

After deleting `baseUrl`, set `pathsBasePath` to `path.dirname(path.resolve(tsConfigPath))` so TypeScript uses the same resolution base as webpack.

```diff
 result.options.paths = rewritePathAliasesWithoutBaseUrl(...)
 delete (result.options as { baseUrl?: unknown }).baseUrl
+result.options.pathsBasePath = path.dirname(path.resolve(tsConfigPath))
```

Fixes #93336

### DIFF
--- a/packages/next/src/lib/typescript/getTypeScriptConfiguration.test.ts
+++ b/packages/next/src/lib/typescript/getTypeScriptConfiguration.test.ts
@@ -1,0 +1,84 @@
+import { mkdtemp, writeFile, mkdir } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join, dirname } from 'node:path'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import ts from 'typescript'
+import { getTypeScriptConfiguration } from './getTypeScriptConfiguration'
+
+describe('getTypeScriptConfiguration()', () => {
+  let tmpDir: string
+  let tsConfigPath: string
+
+  beforeEach(async () => {
+    tmpDir = await mkdtemp(join(tmpdir(), 'nextjs-ts-config-test-'))
+    tsConfigPath = join(tmpDir, 'tsconfig.json')
+  })
+
+  describe('TypeScript 6 compat — pathsBasePath', () => {
+    it('removes baseUrl and sets pathsBasePath to tsconfig dir for TS6 compat', async () => {
+      // The scenario not addressed by #92277 (which handled paths-without-baseUrl):
+      // when a tsconfig HAS baseUrl, the TS6 compat code rewrites paths to
+      // relative form and deletes baseUrl. TypeScript then needs pathsBasePath
+      // to determine the root for resolving those rewritten aliases. Without
+      // setting pathsBasePath, TypeScript falls back to the repo root (or CWD)
+      // which breaks monorepo setups where the app lives in a subdirectory.
+      await writeFile(
+        tsConfigPath,
+        JSON.stringify({
+          compilerOptions: {
+            baseUrl: '.',
+            paths: {
+              '@lib/*': ['./src/lib/*'],
+            },
+          },
+        }),
+        { encoding: 'utf8' }
+      )
+
+      const result = await getTypeScriptConfiguration(ts, tsConfigPath, true)
+
+      // baseUrl must be removed so TS6 deprecation checks don't fail type checking
+      expect(result.options.baseUrl).toBeUndefined()
+      // pathsBasePath must be set to the tsconfig directory so the rewritten
+      // ../../... aliases resolve relative to the app dir, not the repo root
+      expect(result.options.pathsBasePath).toBe(dirname(tsConfigPath))
+    })
+
+    it('sets pathsBasePath to the app tsconfig dir when baseUrl is inherited from an extended root tsconfig', async () => {
+      // Monorepo scenario: root tsconfig defines baseUrl/paths, app tsconfig extends it.
+      // The pathsBasePath must point to the *app* tsconfig dir so that rewritten
+      // relative aliases (../../packages/...) resolve from the app, not the root.
+      const rootTsConfigPath = join(tmpDir, 'tsconfig.base.json')
+      await writeFile(
+        rootTsConfigPath,
+        JSON.stringify({
+          compilerOptions: {
+            baseUrl: '.',
+            paths: {
+              '@shared/*': ['./packages/shared/src/*'],
+            },
+          },
+        }),
+        { encoding: 'utf8' }
+      )
+
+      const appDir = join(tmpDir, 'apps', 'web')
+      await mkdir(appDir, { recursive: true })
+      const appTsConfigPath = join(appDir, 'tsconfig.json')
+      await writeFile(
+        appTsConfigPath,
+        JSON.stringify({
+          extends: '../../tsconfig.base.json',
+          compilerOptions: {},
+        }),
+        { encoding: 'utf8' }
+      )
+
+      const result = await getTypeScriptConfiguration(ts, appTsConfigPath, true)
+
+      expect(result.options.baseUrl).toBeUndefined()
+      // Must be the app dir, not the root dir where tsconfig.base.json lives
+      expect(result.options.pathsBasePath).toBe(dirname(appTsConfigPath))
+    })
+  })
+})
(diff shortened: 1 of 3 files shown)


eps1lon (commented):
Can you write a test that covers the specific scenario that wasn't fixed with https://github.com/vercel/next.js/pull/92277?

sleitor:
@eps1lon Added tests in commit [eb85bca](https://github.com/sleitor/next.js/commit/eb85bca907) — a new `getTypeScriptConfiguration.test.ts` file with two cases covering the scenario not addressed by #92277:

1. **Direct `baseUrl` in tsconfig** — verifies that after TS6 compat removes `baseUrl`, `pathsBasePath` is set to the tsconfig directory (not left undefined/defaulting to CWD/repo root)
2. **Inherited `baseUrl` via `extends`** — the monorepo scenario where `baseUrl` comes from a root tsconfig extended by the app tsconfig; verifies `pathsBasePath` is the *app* tsconfig dir, not the root dir where `tsconfig.base.json` lives

Both tests run with the installed TypeScript 6 version so the TS6 compat branch is exercised. Happy to adjust scope or add more cases if needed!

eps1lon:
An e2e test please with an actual Next.js app. These are located in the top level `test/` directory.

sleitor:
@eps1lon Done! Added an e2e production test in commit [206f89095a](https://github.com/sleitor/next.js/commit/206f89095a) at `test/production/typescript-paths-base-path/typescript-paths-base-path.test.ts`.

The test:
1. Creates a temp "repo root" directory (outside the Next.js project) with a `tsconfig.base.json` that has `baseUrl: "."` and `paths: { "@shared/*": ["packages/shared/src/*"] }`, plus a shared module at `packages/shared/src/index.ts` — simulating the exact monorepo layout from #93336
2. Creates a Next.js App Router project whose `tsconfig.json` **extends** that base tsconfig (via absolute path) — this is the scenario where TypeScript sets `pathsBasePath` to the base tsconfig dir rather than the app dir
3. Has a page that imports from `@shared/index` via the inherited path alias
4. Uses `typescript: "^6.0.0"` so the TS6 compat code path in `getTypeScriptConfiguration` is exercised
5. Asserts the build produces no `Cannot find module` / `Type error` output and that the page renders the value from the aliased module

Without the `pathsBasePath` fix the type-check step fails; with it the build succeeds end-to-end.

eps1lon:
Thanks for the repro. I cherry-picked the test into https://github.com/vercel/next.js/pull/93377 which we'll land instead. We want to preserve parity between `tsc` and `next` with regards to typechecking so we'll just stop rewiring `paths` based on `baseUrl`. That just suppressed deprecation warnings. 

If people want to upgrade to TypeScript 6.0 but continue using `baseUrl`, they should add `"ignoreDeprecations": "6.0"` to their `tsconfig.json`. Though strongly recommend addressing the deprecation warnings and rewriting `paths` that aren't relative already to be ready for TypeScript 7.0.

sleitor:
@eps1lon Understood, thanks for the explanation! Happy that the e2e test was useful. Will watch [#93377](https://github.com/vercel/next.js/pull/93377) for the proper fix.

PR #93362, opened 2026-04-29
Report process memory footprint (RSS) in TurboMalloc and trace samples

sokra:
### What?
Adds the process memory footprint (RSS / "Memory" column on macOS) to TurboMalloc's trace samples, mirroring how OS memory pressure was wired up in #93333.
### Why?
The trace viewer already plots `memory` (TurboMalloc's accounted bytes) and, since #93333, OS-level `memory_pressure`. Neither captures the actual resident size of the process. Tracking the OS-reported footprint alongside the allocator's view exposes overhead from things the allocator doesn't account for (e.g. mimalloc reserves, mmap'd files, executable pages, native dependencies) and makes leaks/regressions in real RSS visible in traces.
### How?
- New `TurboMalloc::memory_footprint() -> Option<usize>` (bytes), implemented in `turbopack/crates/turbo-tasks-malloc/src/memory_footprint.rs` with one `mod platform` per OS:
  - **Linux**: parses the resident-pages field of `/proc/self/statm` and multiplies by `sysconf(_SC_PAGESIZE)`.
  - **macOS**: `task_info(mach_task_self(), TASK_VM_INFO, …).phys_footprint` — the same value Apple's Activity Monitor labels "Memory". Guards against older kernels that return a shorter struct by checking the returned `count` covers `phys_footprint`.
  - **Windows**: `GetProcessMemoryInfo(GetCurrentProcess(), …).WorkingSetSize`. New `windows-sys` features `Win32_System_ProcessStatus` and `Win32_System_Threading`.
  - Other targets (incl. wasm) return `None`.
- `TraceRow::MemorySample` gains a `memory_footprint: u64` field. `RawTraceLayer::maybe_report_memory_sample` captures it via `TurboMalloc::memory_footprint().map(|v| v as u64).unwrap_or(0)` at the same sampling point as `memory` and `memory_pressure`.
- Trace-server `Store::MemorySample` becomes `(Timestamp, u64, u8, u64)` carrying `(ts, memory, memory_pressure, memory_footprint)`. A new `memory_footprint_samples_for_range` mirrors `memory_samples_for_range` and `memory_pressure_samples_for_range`, sharing the existing `memory_samples_slice` helper so the three result vectors stay index-aligned and use the same `MAX_MEMORY_SAMPLES`-cap max-per-chunk downsampling.
- `ServerToClientMessage::QueryResult` exposes a parallel `memory_footprint_samples: Vec<u64>` next to `memory_samples` and `memory_pressure_samples`.
This is a wire-format break for the postcard-encoded raw trace (added a serde field to `TraceRow::MemorySample`); old trace files cannot be read by the new trace-server. This matches the precedent set by #93333.
### Testing
- `cargo test -p turbo-tasks-malloc` — 10 pass, including a new `parses_typical_statm_content` / `returns_none_on_malformed_statm` parser pair and a runtime `memory_footprint_is_reported_on_supported_platforms` check that asserts `Some(_)` and a positive, plausibly bounded byte count on Linux/macOS/Windows.
- `cargo build`, `cargo clippy --all-targets`, and `cargo fmt --check` clean for `turbo-tasks-malloc`, `turbopack-trace-utils`, and `turbopack-trace-server`.

### DIFF
--- a/turbopack/crates/turbo-tasks-malloc/src/memory_footprint.rs
+++ b/turbopack/crates/turbo-tasks-malloc/src/memory_footprint.rs
@@ -1,0 +1,142 @@
+//! Per-OS memory footprint detection.
+//!
+//! All implementations return the resident memory size of the current
+//! process in bytes. Platforms that do not expose a memory footprint signal
+//! (or for which the query fails) return `None`.
+
+/// See [`super::TurboMalloc::memory_footprint`].
+pub fn memory_footprint() -> Option<usize> {
+    platform::memory_footprint()
+}
+
+#[cfg(all(target_os = "linux", not(target_family = "wasm")))]
+mod platform {
+    /// Reads `/proc/self/statm` and returns the resident set size in bytes
+    /// (the second whole number, multiplied by the system page size).
+    pub fn memory_footprint() -> Option<usize> {
+        let content = std::fs::read_to_string("/proc/self/statm").ok()?;
+        let resident_pages = parse_resident_pages(&content)?;
+        let page_size = page_size()?;
+        resident_pages.checked_mul(page_size)
+    }
+
+    fn parse_resident_pages(content: &str) -> Option<usize> {
+        // Expected format: "<size> <resident> <shared> <text> 0 <data> 0"
+        // where each value is in pages.
+        let mut iter = content.split_ascii_whitespace();
+        let _size = iter.next()?;
+        let resident = iter.next()?;
+        resident.parse().ok()
+    }
+
+    fn page_size() -> Option<usize> {
+        // Safety: `sysconf` is thread-safe and never accesses memory through
+        // its argument.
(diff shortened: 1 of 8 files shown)


turbopack/crates/turbo-tasks-malloc/src/memory_footprint.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    fn page_size() -> Option<usize> {
+        // Safety: `sysconf` is thread-safe and never accesses memory through
+        // its argument.
+        let value = unsafe { libc::sysconf(libc::_SC_PAGESIZE) };
+        if value <= 0 {
+            return None;
+        }
+        usize::try_from(value).ok()
```
lukesandberg: wrap this in a `LazyLock`?

lukesandberg (dismissed):

PR #93364, opened 2026-04-29
Stop patching error inspection if source maps are disabled

eps1lon:
If source maps are disabled, we're most likely running in a prod server where sourcemapping happens lazily when logs are inspected. Ignore-listing generated frames would forever hide n_m frames. 

It also avoids doing a bunch of work that just won't have any effect.

As a follow-up, I'll experiment with showing all frames during next-build if all original frames are ignore-listed since that's a one-shot process. For next-dev, we'll continue showing no frames if everything is ignore-listed since you have an attached a debugger with the browser to get a high-fidelity UI to see ignore-listed frames.

### DIFF
--- a/packages/next/src/server/node-environment-extensions/error-inspect.tsx
+++ b/packages/next/src/server/node-environment-extensions/error-inspect.tsx
@@ -1,3 +1,19 @@
+import * as module from 'module'
+import * as process from 'process'
 import { patchErrorInspectNodeJS } from '../patch-error-inspect'
 
-patchErrorInspectNodeJS(globalThis.Error)
+const sourceMappingEnabled =
+  // @ts-expect-error Node.js 22+ only
+  typeof module.getSourceMapsSupport === 'function'
+    ? // @ts-expect-error Node.js 22+ only
+      module.getSourceMapsSupport().enabled
+    : process.env.NODE_OPTIONS?.includes('--enable-source-maps') ||
+      process.execArgv.includes('--enable-source-maps')
+
+if (sourceMappingEnabled) {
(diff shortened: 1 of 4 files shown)


packages/next/src/server/node-environment-extensions/error-inspect.tsx
```diff
@@ -1,3 +1,19 @@
+import * as module from 'module'
+import * as process from 'process'
 import { patchErrorInspectNodeJS } from '../patch-error-inspect'
 
-patchErrorInspectNodeJS(globalThis.Error)
+const sourceMappingEnabled =
+  // @ts-expect-error Node.js 22+ only
+  typeof module.getSourceMapsSupport === 'function'
+    ? // @ts-expect-error Node.js 22+ only
+      module.getSourceMapsSupport().enabled
```
eps1lon, line 9: https://nodejs.org/api/module.html#modulegetsourcemapssupport

test/e2e/app-dir/actions/app-action-size-limit-invalid.test.ts
```diff
@@ -149,7 +149,7 @@ describe('app-dir action size limit invalid config', () => {
       if (!isNextDeploy) {
         await retry(() => {
           expect(logs).toContainEqual(
-            expect.stringContaining('Error: Body exceeded 2mb limit')
+            expect.stringContaining('[Error]: Body exceeded 2mb limit')
```
eps1lon: These are subclasses of `Error`. We minify classnames. Node.js includes the constructor name when inspecting errors if the constructor name does not end with "Error" so `class y extends Error {}` ends up being logged as `y [Error]: `

PR #93368, opened 2026-04-29
Fix client router cache collision for repeated search param   keys  

### LINKED ISSUE #92152: Next 16.2 router cache issue, if search param name is used multiple times

https://github.com/JonasSeegers/next-16-error-example/tree/main/src/app

### To Reproduce

1. Start the application
2. Select every color
3. Deselect "green"

Current behavior: I don't see the Suspense fallback, the list doesn't show the state correctly
Expected behavior: After deselecting "green", I want to see "loading..." immediately  and the correct list after 1500ms

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:51:28 PST 2026; root:xnu-12377.91.3~2/RELEASE_ARM64_T6041
  Available memory (MB): 65536
  Available CPU cores: 16
Binaries:
  Node: 24.11.0
  npm: 11.8.0
  Yarn: 1.22.22
  pnpm: 10.33.0
Relevant Packages:
  next: 16.2.1 // Latest available version is detected (16.2.1).
  eslint-config-next: N/A
  react: 19.2.3
  react-dom: 19.2.3
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Linking and Navigating

### Additional context

In our application we're following [this guide](https://nextjs.org/learn/dashboard-app/adding-search-and-pagination) to provide a page with filtering options. This used to work very well, but since we updated `next` to `16.2.x` we run into the same issue that you can reproduce from the given repository

genie-ru:
When a search param key was used multiple times in a URL (e.g. `?color=red&color=green&color=blue`), three client-side call sites computed the page cache key by passing `Object.fromEntries(new URLSearchParams(...))` into `addSearchParamsIfPageSegment`. That collapsed repeated entries down to the last value, so the URL above and the shorter `?color=red&color=blue` both reduced to `{"color":"blue"}`. The server passed a `Record<string, string | string[]>` instead and preserved arrays, so the two formats disagreed. When the user unchecked a value other than the last one, the new URL collided with the previous cache entry, the client served the stale render, and the Suspense boundary never re-suspended to reflect the change.

This change introduces a `searchStringToRecord` helper in `packages/next/src/shared/lib/router/utils/` that builds the same `Record<string, string | string[]>` shape as the server, preserving arrays for keys that appear more than once. The three client call sites now use it: `createSegmentFromRouteTree` in `router-reducer/ppr-navigations.ts` (the regression source, added by #87256 in v16.2.0), `doesCurrentSegmentMatchCachedSegment` in `segment-cache/scheduler.ts`, and `getCacheKeyForDynamicParam` in `route-params.ts`. The latter two have used the same buggy pattern since v16.0 and are updated together so cache-key generation stays consistent across the routing layer.

A new e2e suite at `test/e2e/app-dir/repeated-search-params-cache` reproduces the scenario from the issue: the test navigates from `?color=red&color=green&color=blue` to `?color=red&color=blue` and asserts the result list updates from `[red, green, blue]` to `[red, blue]`. Without the fix the second navigation hits the stale cache entry and the list keeps `green`; with the fix it updates correctly. A unit test for `searchStringToRecord` covers single values, repeated keys, empty values, and mixed cases. Existing `prefetch-searchparam`, `searchparams-reuse-loading`, `search-params-react-key`, and `conflicting-search-and-route-params` suites continue to pass.

fixes #92152


transfer1992:
Because we are severely impacted by https://github.com/vercel/next.js/issues/92152, we decided to build a local `patch-package` setup from this. We cannot wait until this is merged.
This is very critical for many projects.

PR #93374, opened 2026-04-30
Update 05-server-and-client-components.mdx

zana-abdi2002:
Added an important exception (for when client components wrap server components) under an original line that seems to claim a universal law.

### DIFF
--- a/docs/01-app/01-getting-started/05-server-and-client-components.mdx
+++ b/docs/01-app/01-getting-started/05-server-and-client-components.mdx
@@ -173,9 +173,11 @@ export default function Counter() {
 
 `"use client"` is used to declare a **boundary** between the Server and Client module graphs (trees).
 
-Once a file is marked with `"use client"`, **all its imports and child components are considered part of the client bundle**. This means you don't need to add the directive to every component that is intended for the client.
+Once a file is marked with `"use client"`, **all of its imports and the components it directly renders are included in the client bundle**. This means you don’t need to add the directive to every component that is intended for the client.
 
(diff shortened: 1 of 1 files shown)


docs/01-app/01-getting-started/05-server-and-client-components.mdx
```diff
@@ -175,6 +175,8 @@ export default function Counter() {
 
 Once a file is marked with `"use client"`, **all its imports and child components are considered part of the client bundle**. This means you don't need to add the directive to every component that is intended for the client.
 
+Note: Server Components passed as **props** to a Client Component, (like in [this example](https://nextjs.org/docs/app/getting-started/server-and-client-components#interleaving-server-and-client-components)) remain Server Components. The Client Component receives only the RSC Payload of those Server Components, not their code, preserving their server-side execution.
```
icyJoseph:
```suggestion
Once a file is marked with `"use client"`, **all of its imports and the components it directly renders are included in the client bundle**. This means you don’t need to add the directive to every component that is intended for the client.

This behavior applies to components that are part of the Client Component’s [module graph](/docs/app/glossary#module-graph), which includes the modules it imports and the components it renders directly. It does not apply to Server Components passed as children or other props. Those components are not imported into the Client Component’s module graph. They are rendered on the server and passed to the Client Component as rendered output.

See [Interleaving Server and Client Components](/docs/app/getting-started/server-and-client-components#interleaving-server-and-client-components) for how Server and Client Components can be combined.
```

aurorascharff (approved):

PR #93378, opened 2026-04-30
Only log partytown setup error once

eps1lon:
Which allows removing `FatalError`. There are more instances of errors that are "fatal" that aren't using this special Error subclass. We don't even document what's so special about "FatalError".


packages/next/src/lib/verify-partytown-setup.ts
```diff
@@ -65,35 +64,25 @@ export async function verifyPartytownSetup(
...
-          )} is installed as a dependency.`
-        )
-      }
-    }
-  } catch (err) {
-    // Don't show a stack trace when there is an error due to missing dependencies
-    if (err instanceof FatalError) {
-      console.error(err.message)
-      // Throw to allow finally blocks to run (e.g., telemetry flush)
-      throw err
```
eps1lon, line 95: This was duplicating the error message. It's the only reference to `FatalError` that's not instantiating it.

unstubbable (approved):

PR #93381, opened 2026-04-30
docs: clarify rendering flow in Server/Client interleaving example

icyJoseph:
follow up to: https://github.com/vercel/next.js/pull/93374

### DIFF
--- a/docs/01-app/01-getting-started/05-server-and-client-components.mdx
+++ b/docs/01-app/01-getting-started/05-server-and-client-components.mdx
@@ -340,8 +340,6 @@ export default function Page() {
 
 In this pattern, Server Components are rendered on the server ahead of time, even when passed as props to Client Components. The React Server Component Payload contains the rendered result of those Server Components, plus placeholders for where Client Components should be rendered and references to their JavaScript files.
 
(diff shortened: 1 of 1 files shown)


aurorascharff (approved):

docs/01-app/01-getting-started/05-server-and-client-components.mdx
```diff
@@ -338,7 +338,9 @@ export default function Page() {
 }
 ```
 
-In this pattern, all Server Components will be rendered on the server ahead of time, including those as props. The resulting RSC payload will contain references of where Client Components should be rendered within the component tree.
+In this pattern, Server Components are rendered on the server ahead of time, even when passed as props to Client Components. The React Server Component Payload contains the rendered result of those Server Components, plus placeholders for where Client Components should be rendered and references to their JavaScript files.
+
+Here, the RSC Payload contains the rendered `<Cart>`, plus a placeholder for `<Modal>` with a reference to its JavaScript file. On the client, React shows the prerendered HTML, reconciles the server and client trees from the RSC Payload, and uses `<Modal>`'s JavaScript during hydration and for ongoing interactivity. See [How do Server and Client Components work in Next.js?](#how-do-server-and-client-components-work-in-nextjs) for the full rendering flow.
```
icyJoseph: Consider tossing this

```suggestion
````

PR #93382, opened 2026-04-30
docs: single file name, without pipe divider

icyJoseph:
In certain markdown processing engines, the pipe `|` in the fence produces _strange_ results, though it is allowed by spec.


icyJoseph:
Not needed anymore. The tool that was parsing our docs fixed this upstream.

PR #93395, opened 2026-04-30
fix(swc): apply swcEnvOptions to server-side env

### LINKED ISSUE #93372: experimental.swcEnvOptions is not applied to server-side (Node) SWC compilation

https://github.com/izemo/izemo-next-app

### To Reproduce

1. Clone the reproduction repo and `pnpm install`.
2. The repo is configured with:
   - next: 16.3.0-canary.5
   - engines.node: "22.22.2"
   - .browserslistrc: `chrome 110`
   - next.config.ts: `experimental.swcEnvOptions: { mode: "usage", coreJs: "3.49" }`
3. `app/server/page.tsx` is a Server Component (with `export const dynamic = "force-dynamic"`) that calls `RegExp.escape("a.b.c")`. `RegExp.escape` is TC39 Stage 4 (ES2025) and is missing in both Chrome 110 and Node 22.22.2; it landed in Chrome 136 / Node 24.
4. `app/client/page.tsx` is the client-side mirror (call inside `useEffect`).
5. Run `pnpm build --webpack` and `pnpm start`.
6. `curl -i http://localhost:3000/server`.

**Expected:** With `experimental.swcEnvOptions: { mode: "usage", coreJs: "3.49" }`,
SWC's preset-env should auto-inject the `core-js` polyfill for `RegExp.escape`
into BOTH the client bundle and the Node-side server bundle, since neither
browserslist target (chrome 110) nor `engines.node` (22.22.2) ships it natively.

**Actual:** The polyfill is injected only into the client bundle; the
server bundle emits the call as-is.

Evidence after `pnpm build --webpack`:

  Client bundle (.next/static/chunks/app/client/page-*.js):
    - `escape:function` polyfill body  → 1 occurrence (core-js inlined)  ✅

  Server bundle (.next/server/app/server/page.js):
    - `escape:function` polyfill body  → 0 occurrences                   ❌
    - Verbatim user code in bundle    : RegExp.escape("a.b.c")

Hitting the route on Node 22.22.2 fails:

  $ curl -i http://localhost:3000/server
  HTTP/1.1 500 Internal Server Error
  ...
  ⨯ TypeError: RegExp.escape is not a function
      at f (.next/server/app/server/page.js:2:13549)
      at stringify (<anonymous>) {
    digest: '891653235'
  }

**Root cause** (verified by reading the installed source):
In `packages/next/src/build/swc/options.ts` (webpack path), the `isServer`
branch hardcodes `env: { targets: { node: process.versions.node } }` and
never spreads `swcEnvOptions`, while the client branch correctly does
`env: { targets: supportedBrowsers, ...swcEnvOptions }`.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
Binaries:
  Node: 22.22.2
  npm: 10.9.7
  pnpm: 10.33.2
Relevant Packages:
  next: 16.3.0-canary.5
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Turbopack

### Additional context

_No response_

ShaharAviram1:
## What?

Apply `experimental.swcEnvOptions` to the server-side SWC configuration in `getLoaderSWCOptions`.

## Why?

Currently, `swcEnvOptions` are only applied to client-side compilation, leading to inconsistent behavior between client and server builds.

This PR ensures that server-side SWC compilation respects the same configuration, while preserving the existing Node target behavior.

## How?

- Extract `targets` from `swcEnvOptions`
- Merge remaining fields into the server `env` configuration
- Safely merge `targets` to preserve the default Node target

## Notes

- Scope is intentionally limited to the webpack SWC loader path
- Jest and other SWC configuration paths are unchanged
- No tests added since this is a small, isolated configuration fix and existing SWC behavior is covered indirectly through integration tests

Fixes #93372

### DIFF
--- a/packages/next/src/build/swc/options.ts
+++ b/packages/next/src/build/swc/options.ts
@@ -519,6 +519,8 @@ export function getLoaderSWCOptions({
       }
     : {}
 
+  const { targets: userTargets, ...restEnvOptions } = swcEnvOptions ?? {}
+
   let options: any
   if (isServer) {
     options = {
@@ -533,9 +535,11 @@ export function getLoaderSWCOptions({
       preferEsm: !!esm,
       isPageFile,
       env: {
+        ...restEnvOptions,
         targets: {
           // Targets the current version of Node.js
           node: process.versions.node,
+          ...(userTargets ?? {}),
         },
       },
     }


izemo:
Thanks for your PR to fix my issue. ❤️

Verified locally on next@16.3.0-canary.5 with my [repo](https://github.com/izemo/izemo-next-app): applying PR #93395 makes .next/server/app/server/page.js include the inlined core-js RegExp.escape polyfill, and curl /server on Node 22.22.2 now returns HTTP 200 instead of TypeError: RegExp.escape is not a function. **The patch fixes my issue for next build --webpack.** 

Note it does not touch the **Turbopack** code path — users on the default Turbopack build will still see the bug

ShaharAviram1:
Thanks for checking this out and verifying it 🙏 really appreciate the detailed feedback.

Yeah, good point about Turbopack. I kept this PR focused on the webpack path to keep it small and targeted, but it definitely looks like the same issue exists there as well.

I can take a look and open a follow-up PR for Turbopack if that makes sense.

PR #93106, opened 2026-04-21
perf(ecmascript): shrink JsValue 64→32 bytes

mmastrac:
Three changes that together shrink `JsValue` from **64 → 32 bytes** (−50%). JsValue is the analyzer's value type and is instantiated millions of times during large-app analysis, so this directly reduces peak memory and improves cache locality on the hot `link`/`replace_builtin` paths.

### 1. Box `RequireContextValue` inside `WellKnownFunctionKind`

`RequireContextValue` carries an `FxIndexMap` and is only used by three variants (`RequireContextRequire{,Keys,Resolve}`). Boxing it pulls those variants' payload down from 48B to 8B, dropping `WellKnownFunctionKind` from 64 → 16 bytes and `JsValue` from 64 → 48 bytes.

### 2. Unify `MemberCall`/`New`/`Call` payload into a single `Vec`

`MemberCall` was `(u32, Box<JsValue>, Box<JsValue>, Vec<JsValue>)` (48B payload). Collapsed into `(u32, Vec<JsValue>)` (24B payload) with storage layout `[args..., prop, obj]`.

The reversed order is the key trick: on the common fallthrough in `builtin::replace_builtin` (convert `obj.prop(args)` into a `Call(Member(obj, prop), args)`), we pop `obj` then `prop` off the tail of the Vec and reuse the remainder **as** the args Vec — zero extra allocations on the hot path.

A similar optimization was applied to `New` and `Call`

A tricky think about this layout is avoiding reallocations of 'arg' vecs when constructing, To make this easier I added parallel factory methods `JsValue::call_from_parts` and `JsValue::call_from_iter`  to eliminate reallocations when constructing

Drops JsValue from 48 → 40 bytes.

### 3. Replace `Unknown::reason` with `RcStr`

The `reason` field of Unknown was a `Cow<&'static, str>` but nearly all uses were `&'static str` values so they were migrated to `rcstr!` which drops the reason field from 24 bytes to 8 bytes.

Drops JsValue from 40 -> 32 bytes

## Test plan

- [x] `cargo test -p turbopack-ecmascript --lib` — all 352 tests pass (snapshot fixtures byte-identical)

## Benchmark results

Ran `cargo bench -p turbopack-ecmascript --bench analyzer` on this branch vs `canary` (59 fixtures × 2 benches each, criterion baselines).

| Bench | Sum (canary → branch) | Δ | Geomean ratio |
|---|---|---|---|
[...]

PR #93111, opened 2026-04-22
perf(ecmascript): SlimVec experiment — 32 B JsValue

mmastrac:
Stacked on #93106. **Experiment** — see Codspeed verdict before landing.

Introduces `SlimVec<T>`, a `Vec`-alike with `u32` len + capacity (16 B total vs 24 B for `Vec`). Same size as `Box<[T]>` but conversion from `Vec` is **zero-cost** — keeps whatever capacity the `Vec` had rather than calling `shrink_to_fit`. Swaps in as the args storage for `Call`/`New`, and combines with the `Unknown.reason` `Cow<'static, str>` → `&'static str` shrink to bring `JsValue` to 32 B.

## Why this PR exists

Earlier revision used `Box<[JsValue]>` for args. Codspeed flagged a +4.73 % regression on `app-page-turbo.runtime.prod.js[full]`, pointing at `Vec::into_boxed_slice`'s `shrink_to_fit` call as the likely instruction-count culprit (it reallocates whenever `capacity != length`).

`SlimVec` keeps the capacity-free 16 B footprint but drops the realloc.

## Local wall-clock

On Apple Silicon the Full version (SlimVec + Unknown shrink, `JsValue` = 32 B) still shows +2 to +5 % vs parent — the local regression tracks `JsValue` hitting 32 B independent of the realloc path. Codspeed's measurement methodology (instruction counting) is the one that matters here since it flagged the original; pushing to get its verdict.

## Variant sizes

| | Before | After |
|---|---|---|
| `Unknown` | 40 B | 32 B |
| `Call` / `New` | 40 B | 32 B |
| **`JsValue` enum** | **40 B** | **32 B** |

`SlimVec<JsValue>`: 8 B ptr + 4 B len + 4 B cap = 16 B. u32 capacity supports up to 4 billion args — JsValue call sites are bounded at ~10.

## Test plan
- [x] 351 lib tests pass, no snapshot changes
- [ ] Codspeed verdict (this push)

PR #93402, opened 2026-04-30
perf(ecmascript): wrap recursive JsValue children in Arc

lukesandberg:
WIP: benchmark results are not compelling.

## What this branch does

Wrap recursive `JsValue` children in `triomphe::Arc` and remove a number of follow-on clones. The intent was to eliminate the per-argument `Box<JsValue>` allocations the analyzer was doing for `Call` / `MemberCall` argument lists and reduce make-mut clones in the visitor APIs.

Changes layered on top of the original commit:

- `triomphe::Arc<JsValue>` for all recursive children, replacing `Box<JsValue>` and `std::sync::Arc<JsValue>` mixed usage.
- `Vec<Arc<JsValue>>` for argument lists, with `[args..., callee]` and `[args..., prop, obj]` tail layouts so `into_parts` / `from_parts` can round-trip without re-allocating.
- Constructor surface forced to take `Arc<JsValue>` directly so heap allocations are visible at the call site (no JsValue-wrapping convenience overloads). Renamed the surviving `_arc` variants to drop the suffix.
- Pushed `Arc<JsValue>` down through `well_known_function_call`, the prop-dispatch helpers (`*_module_member`), and `require_context_visitor` so we don't unwrap-and-rewrap at the boundary.
- `for_each_children_mut` (and the `_early` / `_late` siblings) now take `&mut Arc<JsValue>` instead of `&mut JsValue`, letting the visitor decide whether to replace the slot, mutate the inner via `Arc::make_mut`, or just read.
- Pushed `Arc::make_mut` audit through `builtin.rs` — match on `&*obj` for variant dispatch, then `Arc::unwrap_or_clone` only inside the matched branch.
- `take_arc_slot` helper with a shared `PLACEHOLDER` static, used in linker.rs and builtin.rs to swap out a child without paying for `Arc::default()`'s per-call allocation.

## Benchmark results

Compared against canary using the in-tree `analyzer` criterion benches (`create_graph` and `link` on every fixture in `tests/analyzer/graph/`).

### Headline (full run, 87 benchmarks)

|        | n  | median Δ | mean Δ | worst |
|---|---:|---:|---:|---|
| create_graph | 57 | **+8.8%** | +9.7% | process-and-os: +34.2% |
| link         | 30 | **+4.0%** | +4.2% | member-call: +15.8% |
| **All**      | 87 | **+6.7%** | +7.8% | |

62 regressed (criterion p < 0.05), 8 improved, 17 noise/no-change.

### Worst regressions (full run)

```
create_graph/process-and-os       +34.2%
create_graph/logical              +29.9%
create_graph/member-call          +29.7%
create_graph/md5-reduced          +27.2%
create_graph/cycle-cache          +25.1%
create_graph/md5_2                +24.9%
create_graph/md5                  +23.0%
create_graph/assign               +22.0%
create_graph/esbuild              +18.5%
create_graph/peg                  +16.4%
link/esbuild                      +17.4%
link/member-call                  +15.8%
link/cycle-cache                  +15.7%
link/peg                          +13.5%
link/md5-reduced                  +13.6%
link/pack-2521                    +11.8%
```

Top improvements were small (1–6%) and inconsistent: `arrow`, `iife`, `nested`, `declarations`, `class_super`, `webpack-target-node`, `unreachable`, `import-meta-prop`.

### Re-measurement

Reran a targeted subset (large Δ, non-trivial absolute time) against a freshly-saved canary baseline to check for system interference. Canary's own absolute times drifted 1–4% between baseline runs (system noise floor); the regression sizes were stable within ~3pp.

| Benchmark | 1st Δ | 2nd Δ |
|---|---:|---:|
| create_graph/process-and-os | +34.2% | +33.9% |
| create_graph/md5-reduced    | +27.2% | +28.8% |
| create_graph/cycle-cache    | +25.1% | +26.6% |
| create_graph/assign         | +22.0% | +25.3% |
| create_graph/md5_2          | +24.9% | +23.0% |
| create_graph/esbuild        | +18.5% | +20.9% |
| link/cycle-cache            | +15.7% | +21.1% |
| link/esbuild                | +17.4% | +18.8% |
| create_graph/peg            | +16.4% | +13.6% |
| link/peg                    | +13.5% | +10.6% |
| link/md5-reduced            | +13.6% | +16.0% |
| link/process-and-os         |  +8.2% | +10.4% |

The regressions are reproducible.

### Note about the canary baseline

Between the two measurement runs, canary picked up `ed1c6ccbdb perf(ecmascript): shrink JsValue 64→32 bytes (#93106)` and a couple of other turbopack commits. The branch was rebased onto that newer canary before the second measurement, so both rounds compare like-for-like — but the gap is being measured against a faster canary than the first round. The fact that regression sizes still match across rounds means those canary changes didn't shift the comparison.

## Why probably

`create_graph` regresses harder than `link` (median 8.8% vs 4.0%), but `create_graph` doesn't go through the linker at all, so the `&mut Arc<JsValue>` visitor API and the `take_arc_slot` work aren't the cause. `create_graph` is dominated by JsValue construction — `call_from_iter`, `member`, `tenary`, `equal`, etc. — which is exactly the surface where this branch forced explicit `Arc::new(...)` at every call site instead of wrapping inside the constructor.

Suspected sources of extra allocations:

1. `linked_args` and similar in `references/mod.rs` now do `args.iter().cloned().map(Arc::new)` — Arc-allocs per arg per call.
2. `graph.rs` builds alternatives/concat lists with `.into_iter().map(Arc::new).collect()` — Arc-allocs per element.
3. Constructors that previously consumed `JsValue` (already-flat, no inner Arc) now demand `Arc<JsValue>`, so callers Arc-wrap eagerly even when the value would have been immediately destructured.
4. linker.rs leave-phase: every `*child = Arc::new(val)` is a fresh alloc — `done` would need to be `Vec<Arc<JsValue>>` for that allocation to be avoidable.

## Things still worth exploring

- **Bisect the branch.** Several commits' worth of changes are layered here. Running the bench at intermediate commits would tell us whether the storage change (Arc child layout) is itself fine and the regression is concentrated in the constructor-surface change, the visitor-API change, or something else. Cheapest next step.
- **Push `Arc<JsValue>` further into the linker.** `Step::Enter`/`Leave`/`Visit` and `done` currently hold `JsValue`; flipping them to `Arc<JsValue>` lets the original Arc allocation flow back into a child slot when the visitor doesn't modify it. That's the only path I see to recovering the leave-phase `Arc::new(val)` cost.
- **Profile the worst case.** `create_graph/process-and-os` (+34%) on a flame graph would tell us where the extra time is going. If it's `Arc::new` / `alloc::alloc`, that confirms the alloc-count theory; if it's `Arc::clone` / `Drop`, we're paying refcount overhead instead.
- **Selectively un-revert convenience constructors.** The "callers wrap Arcs explicitly" rule catches us at every leaf JsValue construction. We could keep that rule for the list-shaped constructors (where the `_arc` form is the natural one) and restore the JsValue-taking variants for the small fixed-arity ones (`tenary`, `equal`, etc.) where there's no real ambiguity to surface.

Raw criterion logs for the runs in this PR are at `/tmp/bench-canary.log`, `/tmp/bench-feature.log`, `/tmp/bench-canary-rerun.log`, `/tmp/bench-feature-rerun.log` on my machine.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
@@ -15,14 +15,14 @@ pub fn early_replace_builtin(value: &mut JsValue) -> bool {
         JsValue::Call(_, call) => {
             let (args, callee) = call.as_parts_mut();
             let args_have_side_effects = || args.iter().any(|arg| arg.has_side_effects());
-            match Arc::make_mut(callee) {
+            match &**callee {
                 // We don't know what the callee is, so we can early return
-                &mut JsValue::Unknown {
+                JsValue::Unknown {
                     original_value: _,
                     reason: _,
                     has_side_effects,
                 } => {
-                    let has_side_effects = has_side_effects || args_have_side_effects();
+                    let has_side_effects = *has_side_effects || args_have_side_effects();
                     value.make_unknown(has_side_effects, rcstr!("unknown callee"));
                     true
                 }
@@ -48,27 +48,27 @@ pub fn early_replace_builtin(value: &mut JsValue) -> bool {
         JsValue::MemberCall(_, call) => {
             let (args, prop, obj) = call.as_parts_mut();
             let args_have_side_effects = || args.iter().any(|arg| arg.has_side_effects());
-            match Arc::make_mut(obj) {
+            match &**obj {
                 // We don't know what the callee is, so we can early return
-                &mut JsValue::Unknown {
+                JsValue::Unknown {
                     original_value: _,
                     reason: _,
                     has_side_effects,
                 } => {
                     let side_effects =
-                        has_side_effects || prop.has_side_effects() || args_have_side_effects();
+                        *has_side_effects || prop.has_side_effects() || args_have_side_effects();
                     value.make_unknown(side_effects, rcstr!("unknown callee object"));
                     true
                 }
                 // otherwise we need to look at the property
-                _ => match Arc::make_mut(prop) {
+                _ => match &**prop {
                     // We don't know what the property is, so we can early return
-                    &mut JsValue::Unknown {
+                    JsValue::Unknown {
                         original_value: _,
                         reason: _,
                         has_side_effects,
                     } => {
-                        let side_effects = has_side_effects || args_have_side_effects();
+                        let side_effects = *has_side_effects || args_have_side_effects();
                         value.make_unknown(side_effects, rcstr!("unknown callee property"));
                         true
                     }
@@ -102,7 +102,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
             // numeric addition
             let mut sum = 0f64;
             for arg in list {
-                let JsValue::Constant(ConstantValue::Num(num)) = Arc::make_mut(arg) else {
+                let JsValue::Constant(ConstantValue::Num(num)) = &**arg else {
                     return false;
                 };
                 sum += *num.0;
@@ -123,10 +123,10 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                 values,
                 logical_property: _,
             } => {
-                *value = JsValue::alternatives_arc(
+                *value = JsValue::alternatives(
                     take(values)
                         .into_iter()
-                        .map(|alt| Arc::new(JsValue::member_arc(alt, prop.clone())))
+                        .map(|alt| Arc::new(JsValue::member(alt, prop.clone())))
                         .collect(),
                 );
                 true
@@ -139,19 +139,19 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
             } => {
                 fn items_to_alternatives(
                     items: &mut Vec<Arc<JsValue>>,
-                    prop: &mut Arc<JsValue>,
+                    prop: &Arc<JsValue>,
                 ) -> JsValue {
                     items.push(Arc::new(JsValue::unknown(
-                        JsValue::member_arc(
+                        JsValue::member(
                             Arc::new(JsValue::array(Vec::new())),
-                            Arc::new(take(Arc::make_mut(prop))),
+                            prop.clone(),
                         ),
                         false,
                         rcstr!("unknown array prototype methods or values"),
                     )));
-                    JsValue::alternatives_arc(take(items))
+                    JsValue::alternatives(take(items))
                 }
-                match Arc::make_mut(prop) {
+                match &**prop {
                     // accessing a numeric property on an array like `[1,2,3][1]`
                     // We can replace this with the value at the index
                     JsValue::Constant(ConstantValue::Num(num @ ConstantNumber(_))) => {
@@ -164,7 +164,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                                 true
                             } else {
                                 *value = JsValue::unknown(
-                                    JsValue::member_arc(take(obj), take(prop)),
+                                    JsValue::member(obj.clone(), prop.clone()),
                                     false,
                                     rcstr!("invalid index"),
                                 );
@@ -189,10 +189,12 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                         logical_property: _,
                     } => {
                         let obj_arc: Arc<JsValue> = obj.clone();
-                        *value = JsValue::alternatives_arc(
-                            take(values)
-                                .into_iter()
-                                .map(|alt| Arc::new(JsValue::member_arc(obj_arc.clone(), alt)))
+                        *value = JsValue::alternatives(
+                            values
+                                .iter()
+                                .map(|alt| {
+                                    Arc::new(JsValue::member(obj_arc.clone(), alt.clone()))
+                                })
                                 .collect(),
                         );
                         true
@@ -224,7 +226,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                             }
                             ObjectPart::Spread(_) => {
                                 values.push(Arc::new(JsValue::unknown(
-                                    JsValue::member_arc(
+                                    JsValue::member(
                                         Arc::new(JsValue::object(vec![take(part)])),
                                         prop.clone(),
                                     ),
@@ -236,7 +238,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                     }
                     if include_unknown {
                         values.push(Arc::new(JsValue::unknown(
-                            JsValue::member_arc(
+                            JsValue::member(
                                 Arc::new(JsValue::object(Vec::new())),
                                 take(prop),
                             ),
@@ -244,7 +246,7 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                             rcstr!("unknown object prototype methods or values"),
                         )));
                     }
-                    JsValue::alternatives_arc(values)
+                    JsValue::alternatives(values)
                 }
 
                 /// Convert a list of potential values into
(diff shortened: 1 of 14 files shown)


lukesandberg:
* **#93402** <a href="https://app.graphite.com/github/pr/vercel/next.js/93402?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93402?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#93106** <a href="https://app.graphite.com/github/pr/vercel/next.js/93106?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#93111](https://github.com/vercel/next.js/pull/93111) <a href="https://app.graphite.com/github/pr/vercel/next.js/93111?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `canary`


<h2></h2>

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -329,10 +335,11 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                         values,
                         logical_property: _,
                     } => {
-                        *value = JsValue::alternatives(
+                        let obj_arc: Arc<JsValue> = obj.clone();
```
lukesandberg: why does this clone need to be outside?

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -401,8 +409,11 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                                             }
                                         }
                                     }
-                                    obj.update_total_nodes();
-                                    *value = take(obj);
+                                    // Drop our mut borrow of items (held via the outer
+                                    // `Arc::make_mut(obj)` arm) before reading `obj` again.
+                                    let _ = items;
+                                    Arc::make_mut(obj).update_total_nodes();
+                                    *value = Arc::unwrap_or_clone(take(obj));
```
lukesandberg: we should just unwrap_or_clone, mutate and then assign

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -414,11 +425,11 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
                                             .enumerate()
                                             .map(|(i, item)| {
                                                 let mut b = CallBuilder::with_arg_count(2);
-                                                b.push_arg(item);
+                                                b.push_arg_arc(item);
```
lukesandberg: we need to audit all these `_arc` apis to see if the un `Arc` ones are needed

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -506,18 +517,46 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
...
+                            Ok(JsValue::Object {
+                                parts: inner_parts,
+                                mutable: inner_mutable,
+                                ..
+                            }) => {
+                                parts.extend(inner_parts);
+                                *mutable |= inner_mutable;
+                            }
+                            Ok(other) => {
+                                parts.push(ObjectPart::Spread(Arc::new(other)));
```
lukesandberg, line 555: can we preserve the original spread arc here?

turbopack/crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -3186,11 +3187,11 @@ impl Analyzer<'_> {
                         if let Some(box value) = value {
                             let value = self.eval_context.eval(value);
                             JsValue::alternatives(vec![
-                                JsValue::member(Box::new(pat_value.clone()), Box::new(key_value)),
+                                JsValue::member(pat_value.clone(), key_value),
```
lukesandberg: apis like this should just call `member_arc` and inline the Arc::new calls just like how they used to inline the Box::new calsl

turbopack/crates/turbopack-ecmascript/src/analyzer/linker.rs
```diff
@@ -185,19 +187,26 @@ where
                     for arg in args.iter() {
                         total_nodes -= arg.total_nodes();
                     }
-                    entry.insert(args);
+                    // Materialize back to `Vec<JsValue>` for the legacy `fun_args_values`
```
lukesandberg, line 191: we should change this signature if possible

turbopack/crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -423,7 +423,7 @@ pub enum JsValue {
     /// An array of nested values
```
lukesandberg, line 426: <p class='graphite__hidden'><i>[Re: line +381]</i></p>
lol done
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/vercel/next.js/93402?utm_source=unchanged-line-comment">Graphite</a>.</p>

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -17,7 +18,7 @@ pub fn early_replace_builtin(value: &mut JsValue) -> bool {
         JsValue::Call(_, call) => {
             let (args, callee) = call.as_parts_mut();
             let args_have_side_effects = || args.iter().any(|arg| arg.has_side_effects());
-            match callee {
+            match Arc::make_mut(callee) {
```
sokra: This doesn't need a mutable reference and should not use make_mut

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -50,7 +51,7 @@ pub fn early_replace_builtin(value: &mut JsValue) -> bool {
         JsValue::MemberCall(_, call) => {
             let (args, prop, obj) = call.as_parts_mut();
             let args_have_side_effects = || args.iter().any(|arg| arg.has_side_effects());
-            match obj {
```
sokra, line 50: Same

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -63,7 +64,7 @@ pub fn early_replace_builtin(value: &mut JsValue) -> bool {
                     true
                 }
                 // otherwise we need to look at the property
-                _ => match prop {
+                _ => match Arc::make_mut(prop) {
```
sokra: Same

turbopack/crates/turbopack-ecmascript/src/analyzer/graph.rs
```diff
@@ -373,7 +373,7 @@ pub struct EvalContext {
     /// Should be the same [`Mark`] used by [`swc_core::ecma::transforms::base::resolver`].
     pub(crate) top_level_mark: Mark,
     pub(crate) imports: ImportMap,
-    pub(crate) force_free_values: Arc<FxHashSet<Id>>,
+    pub(crate) force_free_values: std::sync::Arc<FxHashSet<Id>>,
```
sokra: Why are we mixing triomphe and std Arcs here?

sokra (commented):
There are many cases where make_mut is used, but it's not necessary. In general we should avoid it until necessary to avoid cloning the data.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph.rs
lukesandberg: std::Arc was preexisting here and i was playing with using triomphe to fix some of the perf issues

lukesandberg: we can probably just switch this one

PR #93408, opened 2026-05-01
fix(swc): apply swcEnvOptions to server-side SWC compilation

### LINKED ISSUE #93372: experimental.swcEnvOptions is not applied to server-side (Node) SWC compilation

sleitor:
## Summary

Fixes #93372

In `packages/next/src/build/swc/options.ts`, the `isServer` branch in `getLoaderSWCOptions` builds its `env` block as:

```ts
env: {
  targets: {
    node: process.versions.node,
  },
},
```

while the client branch correctly does:

```ts
env: {
  targets: supportedBrowsers,
  ...swcEnvOptions,
},
```

This means `experimental.swcEnvOptions` (e.g. `{ mode: 'usage', coreJs: '3.49' }`) is applied only to the client bundle; the server bundle emits polyfillable calls as-is, causing runtime errors on Node versions that don't support those APIs (e.g. `RegExp.escape` on Node 22).

## Fix

Spread `swcEnvOptions` in the server branch's `env` object, parallel to what the client branch already does.

## Testing

- Verified `experimental.swcEnvOptions: { mode: 'usage', coreJs: '3.49' }` now injects core-js polyfills into the server bundle for a Server Component calling `RegExp.escape`.


sleitor:

PR #93411, opened 2026-05-01
CI speed improvements

timneutkens:
## What?

- Restore pnpm store for test timing fetch to speed up `pnpm install` in that step.
- Skip playwright install when it's not needed. Multiple steps don't need Playwright in order to run. Saves 41 seconds per step that doesn't need it.
- Removed extra playwright install on the `test-firefox-safari` as Playwright already gets installed in the reusable workflow.

### Before (canary)

CleanShot 2026-05-01 at 16 17 37@2x

### After (This PR)

CleanShot 2026-05-01 at 16 17 25@2x

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -136,6 +136,7 @@ jobs:
     name: build-next
     uses: ./.github/workflows/build_reusable.yml
     with:
+      needsPlaywright: 'no'
       skipNativeBuild: 'yes'
       stepName: 'build-next'
     secrets: inherit
@@ -210,6 +211,7 @@ jobs:
     needs: ['build-next']
     uses: ./.github/workflows/build_reusable.yml
     with:
+      needsPlaywright: 'no'
       skipNativeBuild: 'yes'
       skipNativeInstall: 'yes'
       afterBuild: |
@@ -241,6 +243,7 @@ jobs:
 
     uses: ./.github/workflows/build_reusable.yml
     with:
+      needsPlaywright: 'no'
       afterBuild: pnpm types-and-precompiled
       stepName: 'types-and-precompiled'
     secrets: inherit
@@ -710,6 +713,7 @@ jobs:
 
     uses: ./.github/workflows/build_reusable.yml
     with:
+      needsPlaywright: 'no'
       nodeVersion: ${{ matrix.node }}
       afterBuild: node run-tests.js --type unit
       stepName: 'test-unit-${{ matrix.node }}'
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -162,6 +163,23 @@ jobs:
...
+      # Match the reusable workflow's pnpm-store cache path so this setup-only
+      # job does not bottleneck the larger test matrix.
+      - name: Get pnpm store directory
+        if: ${{ runner.environment == 'github-hosted' }}
+        id: get-store-path
+        run: echo STORE_PATH=$(pnpm store path) >> $GITHUB_OUTPUT
+
+      - name: Cache pnpm store
+        if: ${{ runner.environment == 'github-hosted' }}
+        uses: actions/cache@v4
```
bgw:
```suggestion
        uses: actions/cache@v5
```

.github/workflows/build_and_test.yml
```diff
@@ -162,6 +163,23 @@ jobs:
         with:
           fetch-depth: 25
 
+      # Match the reusable workflow's pnpm-store cache path so this setup-only
+      # job does not bottleneck the larger test matrix.
+      - name: Get pnpm store directory
+        if: ${{ runner.environment == 'github-hosted' }}
+        id: get-store-path
+        run: echo STORE_PATH=$(pnpm store path) >> $GITHUB_OUTPUT
```
bgw:
```suggestion
        run: echo STORE_PATH="$(pnpm store path)" >> "$GITHUB_OUTPUT"
```

.github/workflows/build_and_test.yml
```diff
@@ -162,6 +163,23 @@ jobs:
         with:
           fetch-depth: 25
 
+      # Match the reusable workflow's pnpm-store cache path so this setup-only
+      # job does not bottleneck the larger test matrix.
```
bgw, line 167: - This is a little risky to keep in sync. It's a lot more boilerplate heavy, but should we use a composite action for this? https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-an-action-metadata-file

- `.github/workflows/trigger_release.yml` and `.github/workflows/build_and_deploy.yml` also has this logic? `build_and_deploy.yml` is already using `actions/cache@v5`.

- [It does look like they implement good security boundaries](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache), but IMO, we should consider reworking the `build_and_deploy` workflow to avoid reading *anything* from cache (unless it's doing a preview build), as there's too much risk the cache can be poisoned (maliciously or just due to a bug in our logic), and the release workflows are too critical.

bgw (approved):

.github/workflows/build_and_test.yml
timneutkens: Yeah that'd be useful, we can follow up here in a follow-up.

PR #93337, opened 2026-04-29
Add MCP `compile_route` tool

lukesandberg:
### What?

Adds a `compile_route` MCP tool that triggers on-demand compilation of a specific route (app or pages) without issuing an HTTP request, and returns any compilation issues.

### Why?

Coding agents and benchmarking workflows need a way to warm the module graph or measure compile time for a route without standing up a live backend to satisfy the request. The existing path — hitting the URL — requires the route's runtime dependencies to be available and couples compile timing to request handling.

### How?

- New tool `mcp/compile_route` registered in `get-or-create-mcp-server.ts`, backed by a `compileRoute({ page, clientOnly })` callback plumbed from the Turbopack hot reloader
- Reuses the dev server's existing on-demand entry path (`ensurePage` / `handleRouteType`), so the call path matches a first navigation.
- Adds a `subscribeToChanges` opt-out on `ensurePage` and threads it through `handleRouteType` / `handlePagesErrorRoute`. One-shot MCP compilations skip HMR subscription wiring — without this, each call would leak a subscription that fires on every subsequent file change for the life of the dev server.
- Telemetry: registers `mcp/compile_route` in the `McpToolName` union.
- e2e test in `test/development/mcp-server/mcp-server-compile-route.test.ts`.

PR #92951, opened 2026-04-17
Add MCP `compile_route` tool

lukesandberg:
### What?

Adds a `compile_route` MCP tool that triggers on-demand compilation of a specific route (app or pages) without issuing an HTTP request, and returns any compilation issues.

### Why?

Coding agents and benchmarking workflows need a way to warm the module graph or measure compile time for a route without standing up a live backend to satisfy the request. The existing path — hitting the URL — requires the route's runtime dependencies to be available and couples compile timing to request handling.

### How?

- New tool `mcp/compile_route` registered in `get-or-create-mcp-server.ts`, backed by a `compileRoute({ page, clientOnly })` callback plumbed from both the Turbopack and webpack hot reloaders.
- Reuses the dev server's existing on-demand entry path (`ensurePage` / `handleRouteType`), so the call path matches a first navigation.
- Adds a `subscribeToChanges` opt-out on `ensurePage` and threads it through `handleRouteType` / `handlePagesErrorRoute`. One-shot MCP compilations skip HMR subscription wiring — without this, each call would leak a subscription that fires on every subsequent file change for the life of the dev server.
- Telemetry: registers `mcp/compile_route` in the `McpToolName` union.
- Error 1172 reserved for "compilation failed but no issues were recorded".
- e2e test in `test/development/mcp-server/mcp-server-compile-route.test.ts`.

PR #91790, opened 2026-03-22
Proof of concept: task eviction after snapshot for turbo-tasks-backend

lukesandberg:
> **Note:** This is a **proof of concept** implementation. It is not yet ready for production use.

## Summary

Implements memory eviction for the turbo-tasks engine. After a persistence snapshot completes, tasks that are safe to remove are evicted from in-memory storage and transparently restored from disk on next access.

### Eviction levels

- **Full eviction**: Entire task removed from the in-memory map (restored from disk on access). Only possible when the task has no meaningful transient state (and other state is already on disk)
- **DataAndMeta eviction**: Both data and meta categories cleared, but the task stays in the map to preserve transient state (e.g. `current_session_clean`, aggregated session-clean counts).
- **DataOnly eviction**: Only data-category fields cleared; meta (graph structure, output, dirty state) stays in memory.
- **MetaOnly eviction**: Only meta-category fields cleared; data stays in memory.

Data and meta evictability are computed independently — if one category is modified but the other is clean, the clean category can still be dropped.

Eviction is gated behind `BackendOptions::evict_after_snapshot` (off by default), and can be enabled in Next.js via the `TURBO_ENGINE_EVICT_AFTER_SNAPSHOT=1` env var for testing.

## Key changes

- **Orthogonal eviction decision tree** (`storage_schema.rs`): Data and meta evictability are computed independently. Full eviction additionally requires no meaningful transient state (session-clean flags, aggregated session-clean counts). Replaces the previous sequential bail-out approach which was too aggressive on full eviction (losing transient session state on leaf tasks) and not aggressive enough on partial eviction (blocking all eviction when only one category was modified).

- **`drop_partial()` codegen** (`task_storage_macro.rs`): New generated methods to drop data

- **`restore_from_*()` codegen changes** (`task_storage_macro.rs`): New semantics for merging persistent data from the backend with transient data stored in memory. 

- **`task_cache` moved into `Storage`** (`storage.rs`): The `CachedTaskType → TaskId` deduplication map was previously a separate field on `TurboTasksBackendInner`. It is now owned by `Storage` so eviction can remove entries when a task is fully evicted. Because `task_cache` is a pure performance cache (entries are re-populated by `task_by_type()` on miss once the task type is persisted to backing storage), evicting entries is safe. After bulk eviction the map is shrunk when it is less than half full.

[...]

PR #93416, opened 2026-05-01
turbo-tasks-backend: fix snapshot coordination races + extract SnapshotCoordinator

lukesandberg:
Hold a lock while persisting so two snapshots cannot execute concurrently

  * Currently if `stop` is called while an idle snapshot is running, then snapshotting can race with itself, this can corrupt the use of the `in_progress_operations` parameter since two threads will `fetch_or` with it and wait for the bit to be cleared

Abort the process if `panic!` occurs during task spawning
  * Currently if `try_start_task_execution` panics it ends up hanging a task which can deadlock the process, in this case we have no better option than to just log and abort.
  * I considered strategies that would 'poison' the task or possibly all of turbo-tasks and this is attractive but i believe fundamentally unsafe, the most likely cause of these panics is something wrong with state tracking in the backend, so exiting is all we can do

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -1,6 +1,7 @@
 mod cell_data;
 mod counter_map;
 mod operation;
+mod snapshot_coordinator;
 mod storage;
 pub mod storage_schema;
 
@@ -13,7 +14,7 @@ use std::{
     pin::Pin,
     sync::{
         Arc, LazyLock,
-        atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
+        atomic::{AtomicBool, AtomicU64, Ordering},
     },
     time::SystemTime,
 };
@@ -21,7 +22,7 @@ use std::{
 use anyhow::{Context, Result, bail};
 use auto_hash_map::{AutoMap, AutoSet};
 use indexmap::IndexSet;
-use parking_lot::{Condvar, Mutex};
+use parking_lot::Mutex;
 use rustc_hash::{FxHashMap, FxHashSet, FxHasher};
 use smallvec::{SmallVec, smallvec};
 use tokio::time::{Duration, Instant};
@@ -63,6 +64,7 @@ use crate::{
             connect_children, get_aggregation_number, get_uppers, is_root_node,
             make_task_dirty_internal, prepare_new_children,
         },
+        snapshot_coordinator::{OperationGuard, SnapshotCoordinator},
         storage::Storage,
         storage_schema::{TaskStorage, TaskStorageAccessors},
     },
@@ -75,7 +77,6 @@ use crate::{
     utils::{
         dash_map_drop_contents::drop_contents,
         dash_map_raw_entry::{RawEntry, get_shard, raw_entry_in_shard, raw_get_in_shard},
-        ptr_eq_arc::PtrEqArc,
         shard_amount::compute_shard_amount,
     },
 };
@@ -83,9 +84,7 @@ use crate::{
 /// Threshold for parallelizing making dependent tasks dirty.
 /// If the number of dependent tasks exceeds this threshold,
 /// the operation will be parallelized.
-const DEPENDENT_TASKS_DIRTY_PARALLIZATION_THRESHOLD: usize = 10000;
-
-const SNAPSHOT_REQUESTED_BIT: usize = 1 << (usize::BITS - 1);
+const DEPENDENT_TASKS_DIRTY_PARALLELIZATION_THRESHOLD: usize = 10000;
 
 /// Configurable idle timeout for snapshot persistence.
 /// Defaults to 2 seconds if not set or if the value is invalid.
@@ -96,20 +95,6 @@ static IDLE_TIMEOUT: LazyLock<Duration> = LazyLock::new(|| {
         .map(Duration::from_millis)
         .unwrap_or(Duration::from_secs(2))
 });
-
-struct SnapshotRequest {
-    snapshot_requested: bool,
-    suspended_operations: FxHashSet<PtrEqArc<AnyOperation>>,
-}
-
-impl SnapshotRequest {
-    fn new() -> Self {
-        Self {
-            snapshot_requested: false,
-            suspended_operations: FxHashSet::default(),
-        }
-    }
-}
 
 pub enum StorageMode {
     /// Queries the storage for cache entries that don't exist locally.
@@ -178,25 +163,14 @@ struct TurboTasksBackendInner<B: BackingStorage> {
 
     storage: Storage,
 
-    /// Number of executing operations + Highest bit is set when snapshot is
-    /// requested. When that bit is set, operations should pause until the
-    /// snapshot is completed. When the bit is set and in progress counter
-    /// reaches zero, `operations_completed_when_snapshot_requested` is
-    /// triggered.
-    in_progress_operations: AtomicUsize,
-
-    /// Serializes calls to `snapshot_and_persist`. The internal protocol
-    /// (snapshot_mode flag, snapshot_request bit, suspended_operations) assumes
-    /// only one snapshot runs at a time; this mutex enforces that contract.
+    /// Coordinates the operation/snapshot interleaving protocol. See
+    /// [`SnapshotCoordinator`] for details.
+    snapshot_coord: SnapshotCoordinator<AnyOperation>,
+    /// Serializes calls to `snapshot_and_persist`. The coordinator's
+    /// `begin_snapshot` asserts that snapshots don't overlap; this mutex
+    /// enforces that contract for our two callers (background loop and
+    /// `stop_and_wait`).
     snapshot_in_progress: Mutex<()>,
-    snapshot_request: Mutex<SnapshotRequest>,
-    /// Condition Variable that is triggered when `in_progress_operations`
-    /// reaches zero while snapshot is requested. All operations are either
-    /// completed or suspended.
-    operations_suspended: Condvar,
-    /// Condition Variable that is triggered when a snapshot is completed and
-    /// operations can continue.
-    snapshot_completed: Condvar,
     /// The timestamp of the last started snapshot since [`Self::start_time`].
     last_snapshot: AtomicU64,
 
@@ -251,11 +225,8 @@ impl<B: BackingStorage> TurboTasksBackendInner<B> {
             ),
             task_cache: FxDashMap::default(),
             storage: Storage::new(shard_amount, small_preallocation),
-            in_progress_operations: AtomicUsize::new(0),
+            snapshot_coord: SnapshotCoordinator::new(),
             snapshot_in_progress: Mutex::new(()),
-            snapshot_request: Mutex::new(SnapshotRequest::new()),
-            operations_suspended: Condvar::new(),
-            snapshot_completed: Condvar::new(),
             last_snapshot: AtomicU64::new(0),
             stopping: AtomicBool::new(false),
             stopping_event: Event::new(|| || "TurboTasksBackend::stopping_event".to_string()),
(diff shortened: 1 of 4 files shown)


lukesandberg:
* **#91790** <a href="https://app.graphite.com/github/pr/vercel/next.js/91790?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#92951](https://github.com/vercel/next.js/pull/92951) <a href="https://app.graphite.com/github/pr/vercel/next.js/92951?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#93337](https://github.com/vercel/next.js/pull/93337) <a href="https://app.graphite.com/github/pr/vercel/next.js/93337?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#93416** <a href="https://app.graphite.com/github/pr/vercel/next.js/93416?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93416?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -1186,6 +1188,26 @@ impl<B: Backend + 'static> TurboTasks<B> {
...
+async fn abort_on_panic<F: Future>(f: F) -> F::Output {
+    use futures::FutureExt;
+    match AssertUnwindSafe(f).catch_unwind().await {
+        Ok(r) => r,
+        Err(_) => {
+            eprintln!(
+                "[turbo-tasks] task execution loop panicked outside the user-task panic boundary. \
+                 Backend state is now inconsistent; aborting to avoid a silent hang. See the \
+                 panic message above."
+            );
```
lukesandberg: This message should be scarier

```
[turbo-tasks] Fatal Internal Error aborting

A fatal error occurred and turbopack cannot continue.

File: turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs [hunk 2]
```diff
@@ -0,0 +1,575 @@
...
+    /// uncompleted-operations log.
+    suspended_operations: FxHashSet<PtrEqArc<O>>,
+}
+
+/// Coordinates operation/snapshot interleaving.
+///
+/// Generic over the operation type the caller wants to suspend. The
+/// coordinator only requires `O: Send + Sync + 'static`; it never inspects
+/// the value, just stores it via [`PtrEqArc`].
+pub struct SnapshotCoordinator<O> {
```
bgw: I understand this needs to be generic for testing, but giving it a default might make it easier to understand how this is actually used.

```suggestion
pub struct SnapshotCoordinator<O = AnyOperation> {
```

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+        #[cold]
+        fn suspend_point_cold<O>(this: &SnapshotCoordinator<O>, suspend: impl FnOnce() -> O) {
+            let op = Arc::new(suspend());
+            let mut state = this.state.lock();
+            if !state.snapshot_requested {
+                // Race: snapshot finished between the `snapshot_pending` check
+                // and acquiring the mutex. Nothing to do.
+                return;
+            }
+            state.suspended_operations.insert(op.clone().into());
```
bgw:
```suggestion
            state.suspended_operations.insert(PtrEqArc::from(op.clone()));
```

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+            );
+            if prev - 1 == SNAPSHOT_REQUESTED_BIT {
+                this.operations_drained.notify_all();
+            }
+            // Wait for the snapshot to finish.
+            this.snapshot_completed
+                .wait_while(&mut state, |s| s.snapshot_requested);
+            // Resume: re-increment and remove ourselves from the suspended set.
+            this.in_progress_operations.fetch_add(1, Ordering::AcqRel);
+            state.suspended_operations.remove(&op.into());
```
bgw:
```suggestion
            state.suspended_operations.remove(&PtrEqArc::from(op));
```

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+
+/// State protected by the mutex. Kept tiny so critical sections stay short.
+struct State<O> {
+    /// `true` between `begin_snapshot` and `SnapshotPhase::drop`.
+    snapshot_requested: bool,
+    /// Operations that called [`SnapshotCoordinator::suspend_point`] and have
+    /// not yet resumed. Returned to the snapshotter via
+    /// [`SnapshotPhase::suspended_operations`] so it can persist them in the
+    /// uncompleted-operations log.
+    suspended_operations: FxHashSet<PtrEqArc<O>>,
```
bgw, line 36: I agree that `FxHashSet<PtrEqArc<O>>` probably makes sense here for simplicity, but another option is a simple `Vec<Option<O>>` plus a freelist (`Vec<usize>`) that allows reusing indexes. Basically a dumber version of [slab](https://docs.rs/slab/latest/slab/index.html).

That might also let us get rid of `PtrEqArc` which is only used here?

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+/// Generic over the operation type the caller wants to suspend. The
+/// coordinator only requires `O: Send + Sync + 'static`; it never inspects
+/// the value, just stores it via [`PtrEqArc`].
+pub struct SnapshotCoordinator<O> {
+    /// Combined count + bit. See [`SNAPSHOT_REQUESTED_BIT`].
+    in_progress_operations: AtomicUsize,
+    state: Mutex<State<O>>,
+    /// Notified by the last operation to drain (count drops to `BIT` while
+    /// `BIT` is set). Awaited by [`begin_snapshot`].
+    operations_drained: Condvar,
```
bgw, line 50: Comment should say `SNAPSHOT_REQUESTED_BIT` not `BIT`.

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+        let coord2 = coord.clone();
+        let snap_thread = thread::spawn({
+            let started_snapshot = started_snapshot.clone();
+            move || {
+                let _phase = coord2.begin_snapshot();
+                started_snapshot.store(1, Ordering::Release);
+            }
+        });
+
+        // Give the snapshotter time to set the bit and start waiting.
```
bgw: It should be okay to inspect private members of the struct in the unit test to verify this is true.

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+        assert_eq!(snapshotter_done.load(Ordering::Acquire), 1);
+
+        snap_thread.join().unwrap();
+        drop(g);
+    }
+
+    /// Spawn a watchdog thread that aborts the process if the test doesn't
+    /// signal completion within the timeout. Aborting (vs. panicking) is the
+    /// only way to fail a test cleanly when its main thread is parked on a
+    /// missed-wakeup — a panic in another thread doesn't unblock a join.
```
bgw: Move this comment about aborting vs panicking right above the abort call. It's an implementation detail and doesn't need to be a doc comment.

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+                    if done_watch.load(Ordering::Acquire) {
+                        return;
+                    }
+                    thread::sleep(Duration::from_millis(50));
+                }
+                eprintln!(
+                    "[watchdog] {label}: timed out after {timeout:?}, missed-wakeup race likely; \
+                     aborting"
+                );
+                std::process::abort();
```
bgw: Just import `AtomicBool`, `Instant`, and `abort`.

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+
+        snap_thread.join().unwrap();
+        drop(g);
+    }
+
+    /// Spawn a watchdog thread that aborts the process if the test doesn't
+    /// signal completion within the timeout. Aborting (vs. panicking) is the
+    /// only way to fail a test cleanly when its main thread is parked on a
+    /// missed-wakeup — a panic in another thread doesn't unblock a join.
+    fn spawn_watchdog(
```
bgw: Another idea:
Really this is just used to implement a test timeout for `stress_no_missed_wakeups`. A watchdog is a much more general concept that doesn't seem needed here.

Make the main thread handle the timeout and run the body of the test (pass it in via a `FnOnce` closure). If the child thread panics, then propagate the panic to the main thread.

It may even be possible to implement that using `CondVar::wait_timeout`/`CondVar::wait_timeout_while` avoiding the hacky 50ms loop.

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -1186,6 +1188,26 @@ impl<B: Backend + 'static> TurboTasks<B> {
 
 struct TurboTasksExecutor;
 
+/// Run a future and abort the process if a panic is reported
+///
+/// Turbtasks catches panics from user code and propagates throught the task tree, but if it happens
+/// as part of state management we have to abort
+async fn abort_on_panic<F: Future>(f: F) -> F::Output {
+    use futures::FutureExt;
```
bgw: Move this import to the top of the file

bgw (approved):

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -992,44 +906,31 @@ impl<B: BackingStorage> TurboTasksBackendInner<B> {
...
-                    });
-            }
-            suspended_operations = snapshot_request
-                .suspended_operations
-                .iter()
-                .map(|op| op.arc().clone())
-                .collect::<Vec<_>>();
-        }
+            self.snapshot_coord.begin_snapshot()
+            // `phase` drops here, releasing the snapshot bit and waking any
```
sokra: Which `phase`?

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -0,0 +1,575 @@
...
+use parking_lot::{Condvar, Mutex};
+use rustc_hash::FxHashSet;
+
+use crate::utils::ptr_eq_arc::PtrEqArc;
+
+/// High bit: set while a snapshot is requested or in flight.
+/// Low bits: count of operations currently executing (not suspended).
+const SNAPSHOT_REQUESTED_BIT: usize = 1 << (usize::BITS - 1);
+
+/// State protected by the mutex. Kept tiny so critical sections stay short.
```
sokra:
```suggestion
/// State protected by the mutex.
```

How is the size of this struct related the length of the critical section

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: stale comment

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
lukesandberg: this isn't about catching panics, it is about missing signals

but yeah we can do this in another way that is a little simpler

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
lukesandberg: deferring this idea.

it is reasonable given that we are only using 'set' semantics to support remove, not dedup.  So an indexed based approach could work, but it is probably overkill for this usecase since the set of outstanding operations is strictly bounded by tokio threads.

PR #93431, opened 2026-05-02
turbo-tasks: support `&T` parameters in `#[turbo_tasks::function]` bodies

lukesandberg:
### What?

Lets `#[turbo_tasks::function]` bodies declare arguments as `&T`, and migrates the high-impact owned-but-borrowed parameters across the codebase to do so.

### Why?

For every cache miss on a turbo-task, the runtime has to clone each non-`Copy` argument out of the cached `Box<dyn DynTaskInputs>` so the executing future (which is `'static`) can own its captures. The `task_fn_impl!` blanket impls call `get_args::<T>(arg).cloned()` to do this — a deep clone of every argument tuple element, on every miss. For a function whose body only ever takes `&path` or `items.iter()`, that clone is pure overhead.

The pain shows up most clearly in things like `DiskFileSystem::read(&self, fs_path: FileSystemPath)`: callers pass an owned `FileSystemPath` (the cache key needs it), the runtime then clones it again before handing it to the body, and the body proceeds to take a borrow.

This PR adds the language-level support to skip that runtime clone and migrates the bodies that benefit.

#### Implementation (the framework change)

The exposed (caller-facing) signature is unchanged in every case — callers always pass owned values, because that's still what the cache key is built from. Only the *body* signature gets to declare `&T`.

- **Macro** (`turbo-tasks-macros/src/func.rs`): `inline_signature_and_block` now wraps every non-receiver typed parameter as `&T'` in the inline closure. For arguments the user wrote as owned `T`, it inserts a `let arg = arg.clone();` shadow up front, so existing bodies keep compiling. `mut` is preserved on the rebind. Operations (`#[turbo_tasks::function(operation)]`) get the same `&T` wrapping but skip the `FromTaskInput` rewrite. `expand_exposed_input_type` strips a leading `&` from the user's type when emitting the *exposed* signature so callers always see the owned form.
- **Runtime** (`turbo-tasks/src/native_function.rs`, `task/function.rs`, `turbo-tasks-backend/src/backend/mod.rs`): `NativeFunction::execute` and `TaskFn::functor` now take `Arc<CachedTaskType>` instead of `&dyn DynTaskInputs`. The Arc is captured by the executing future, which lets references downcast out of `&*task.arg` stay valid for the duration of the future without an intermediate clone.
- **Blanket impls** (`task_fn_impl!`): six blanket impls per arity, by-reference, covering all four `Mode` × receiver-shape combinations (`FunctionMode`, `AsyncFunctionMode`, `MethodMode`, `AsyncMethodMode`, plus the Vc-receiver variants). Arity 0 keeps a separate by-value module (no user args means nothing to take by reference). The previous by-value blanket impls for arity ≥ 1 are gone — every macro-generated inline now flows through the by-reference path.
- **Smoke test**: `turbopack/crates/turbo-tasks-backend/tests/by_ref.rs` exercises `&Vec<u32>` arguments on a non-method, a method with mixed by-value + by-ref args, and a `&Recv` method receiver.
- **Documentation**: a new "When to take `&T`" section in `turbopack/crates/turbo-tasks/function.md` covers the rule, the carve-outs (`&'a T`, `&mut T`, `&dyn Trait`, `&[T]`, `&str` are out; `Copy` types like `Vc<T>` and primitives stay by-value), and the operation-function caveat.
- **Lint**: `.config/ast-grep/rules/turbo-task-pass-by-ref-hint.yml` is a `hint`-severity advisory rule that fires on `#[turbo_tasks::function]` parameters typed as `Vec<…>`, `HashMap<…>`, `String`, `PathBuf`, `OsString`, `FileSystemPath`, etc. (Vc-shaped types and primitives are skipped.) It cannot tell whether a migration would force a new explicit `.clone()` to compensate, so it never offers a `fix:` and the message points at `function.md`.

#### Migrations

Each migration follows a "net-clone test": migrate **only** when the body has a non-owning use that survives without adding a new `.clone()` somewhere in user code. If the body's only use is to forward the parameter by value to another callee, migrating just trades the macro shadow's implicit clone for an explicit one — wash, skip. If the body moves the parameter into a struct, mutates it, or consumes it via `.into_*()`, also skip. The conservative rule means a syntactic candidate list of 458 functions yielded ~125 actually-applied migrations; ~250 of the rest are wash cases or moves-into-struct, and the remainder are out-of-scope (`mut` parameters, etc.).

Migrations land grouped by type and crate for reviewability:

- **`Vec<…>`**: 4 functions (`output_of_endpoints`, `module_graphs_of_endpoints`, `from_graphs_inner`, `EcmascriptChunkType::chunk`).
- **`FileSystemPath`**: 75 functions across `turbo-tasks-fs` (39 incl. `DiskFileSystem::read`, `read_link`, `write`, `write_link`, `metadata`, `raw_read_dir`), the turbopack ecosystem (18 across `turbopack-core`'s resolution hot path — `find_context_file`, `find_package`, `resolve_raw`, `resolve_internal`, `read_matches`, plus `ServerFileSystem` impls and friends), and the Next.js bindings (36 across `next-core`/`next-api` — `next_config::config_file_path`, `next_server::context::get_server_resolve_options_context`, `paths::all_asset_paths`, etc.).
- **`RcStr`**: 45 functions across `turbo-tasks-fs`/`turbo-tasks-env`/`turbo-tasks-fetch` framework crates (HTTP fetch hot path, env lookup), the turbopack ecosystem (resolution helpers, runtime embedders, dev-server route tree), and Next.js bindings (HMR pipeline, napi error stack, page entry / loader, font handling).

The exposed signature is unchanged in every migration, so no caller had to be updated.

### Binary size impact

Release build of `next-napi-bindings` (`cdylib`) on `darwin-aarch64`:

| Build                   | canary             | this branch        | Δ                       |
| ----------------------- | ------------------ | ------------------ | ----------------------- |
| Release `.dylib`        | 124,604,704 (118.83 MiB) | 124,791,792 (119.01 MiB) | **+187,088 B  (+0.15 %)** |
| Stripped                | 84,230,008 (80.33 MiB)   | 84,396,280 (80.49 MiB)   | **+166,272 B  (+0.20 %)** |
| Stripped + `gzip -9`    | 29,227,202 (27.87 MiB)   | 29,241,655 (27.89 MiB)   | **+14,453 B   (+0.05 %)** |

The framework change itself doubles the number of `task_fn_impl!` blanket impls instantiated (now 6 by-ref variants per arity for arities 1..12), and each macro-generated inline gains a per-non-receiver-arg shadow. Net binary-size impact is essentially a wash — within noise on the gzipped output that actually ships.

### Performance

```
# This branch
$ hyperfine --warmup 1 --runs 5 --prepare 'rm -rf .next'  'pnpm next build   --experimental-build-mode=compile'
Benchmark 1: pnpm next build   --experimental-build-mode=compile
  Time (mean ± σ):     66.799 s ±  0.979 s    [User: 497.617 s, System: 79.704 s]
  Range (min … max):   65.893 s … 67.952 s    5 runs
```

```
# canary
$ hyperfine --warmup 1 --runs 5 --prepare 'rm -rf .next'  'pnpm next build   --experimental-build-mode=compile'
Benchmark 1: pnpm next build   --experimental-build-mode=compile
  Time (mean ± σ):     64.329 s ±  2.484 s    [User: 488.328 s, System: 81.013 s]
  Range (min … max):   62.112 s … 68.050 s    5 runs
```

### DIFF
--- a/crates/next-core/src/util.rs
+++ b/crates/next-core/src/util.rs
@@ -216,11 +216,10 @@ pub enum PathType {
 #[turbo_tasks::function]
 pub async fn pathname_for_path(
     server_root: &FileSystemPath,
-    server_path: FileSystemPath,
+    server_path: &FileSystemPath,
     path_ty: PathType,
 ) -> Result<Vc<RcStr>> {
-    let server_path_value = server_path.clone();
-    let path = if let Some(path) = server_root.get_path_to(&server_path_value) {
+    let path = if let Some(path) = server_root.get_path_to(server_path) {
         path
     } else {
         turbobail!("server_path ({server_path}) is not in server_root ({server_root})");
--- a/crates/next-core/src/util.rs
+++ b/crates/next-core/src/util.rs
@@ -215,7 +215,7 @@ pub enum PathType {
 /// Converts a filename within the server root into a next pathname.
 #[turbo_tasks::function]
 pub async fn pathname_for_path(
-    server_root: FileSystemPath,
+    server_root: &FileSystemPath,
     server_path: FileSystemPath,
     path_ty: PathType,
 ) -> Result<Vc<RcStr>> {
--- a/crates/next-napi-bindings/src/next_api/project.rs
+++ b/crates/next-napi-bindings/src/next_api/project.rs
@@ -2307,11 +2307,12 @@ pub fn get_source_map_rope_operation(
 #[turbo_tasks::function(operation)]
 pub async fn project_trace_source_operation(
     container: ResolvedVc<ProjectContainer>,
-    frame: StackFrame,
+    frame: &StackFrame,
     current_directory_file_url: &RcStr,
 ) -> Result<Vc<OptionStackFrame>> {
     let Some(map) =
-        &*SourceMap::new_from_rope_cached(get_source_map_rope(*container, frame.file)).await?
+        &*SourceMap::new_from_rope_cached(get_source_map_rope(*container, frame.file.clone()))
+            .await?
     else {
         return Ok(Vc::cell(None));
     };
--- a/crates/next-napi-bindings/src/next_api/project.rs
+++ b/crates/next-napi-bindings/src/next_api/project.rs
@@ -2234,9 +2234,9 @@ pub struct OptionStackFrame(Option<StackFrame>);
 #[turbo_tasks::function]
 pub async fn get_source_map_rope(
     container: Vc<ProjectContainer>,
-    source_url: RcStr,
+    source_url: &RcStr,
 ) -> Result<Vc<FileContent>> {
-    let (file_path_sys, module) = match Url::parse(&source_url) {
+    let (file_path_sys, module) = match Url::parse(source_url) {
         Ok(url) => match url.scheme() {
             "file" => {
                 let path = match url.to_file_path() {
@@ -2308,7 +2308,7 @@ pub fn get_source_map_rope_operation(
 pub async fn project_trace_source_operation(
     container: ResolvedVc<ProjectContainer>,
     frame: StackFrame,
-    current_directory_file_url: RcStr,
+    current_directory_file_url: &RcStr,
 ) -> Result<Vc<OptionStackFrame>> {
     let Some(map) =
         &*SourceMap::new_from_rope_cached(get_source_map_rope(*container, frame.file)).await?
@@ -2352,7 +2352,7 @@ pub async fn project_trace_source_operation(
             // Client code uses file://
             (
                 RcStr::from(
-                    get_relative_path_to(&current_directory_file_url, &original_file)
+                    get_relative_path_to(current_directory_file_url, &original_file)
                         // TODO(sokra) remove this to include a ./ here to make it a relative path
                         .trim_start_matches("./"),
                 ),
@@ -2364,7 +2364,7 @@ pub async fn project_trace_source_operation(
             (
                 RcStr::from(
                     get_relative_path_to(
-                        &current_directory_file_url,
+                        current_directory_file_url,
                         &format!("{project_root_uri}{source_file}"),
                     )
                     // TODO(sokra) remove this to include a ./ here to make it a relative path
@@ -2434,10 +2434,10 @@ pub async fn project_get_source_for_asset(
             #[turbo_tasks::function(operation)]
             async fn source_content_operation(
                 container: ResolvedVc<ProjectContainer>,
-                file_path: RcStr,
+                file_path: &RcStr,
             ) -> Result<Vc<FileContent>> {
                 let project_path = container.project().project_path().await?;
-                Ok(project_path.fs().root().await?.join(&file_path)?.read())
+                Ok(project_path.fs().root().await?.join(file_path)?.read())
             }
 
             let source_content = &*source_content_operation(container, file_path.clone())
--- a/turbopack/crates/turbo-tasks-fetch/src/client.rs
+++ b/turbopack/crates/turbo-tasks-fetch/src/client.rs
@@ -83,10 +83,10 @@ impl FetchClientConfig {
     #[turbo_tasks::function(network)]
     pub async fn fetch(
         self: Vc<FetchClientConfig>,
-        url: RcStr,
+        url: &RcStr,
         user_agent: Option<RcStr>,
     ) -> Result<Vc<FetchResult>> {
-        let url_ref = &*url;
+        let url_ref: &str = url;
         let this = self.await?;
         let response_result: reqwest::Result<HttpResponse> = async move {
             let reqwest_client = this.try_get_cached_reqwest_client()?;
@@ -123,7 +123,7 @@ impl FetchClientConfig {
                 // the client failed to construct or the HTTP request failed
                 mark_session_dependent();
                 Ok(Vc::cell(Err(
-                    FetchError::from_reqwest_error(&err, &url).resolved_cell()
+                    FetchError::from_reqwest_error(&err, url).resolved_cell()
                 )))
             }
         }
--- a/turbopack/crates/turbo-tasks-fetch/src/client.rs
+++ b/turbopack/crates/turbo-tasks-fetch/src/client.rs
@@ -84,20 +84,19 @@ impl FetchClientConfig {
     pub async fn fetch(
         self: Vc<FetchClientConfig>,
         url: &RcStr,
-        user_agent: Option<RcStr>,
+        user_agent: &Option<RcStr>,
     ) -> Result<Vc<FetchResult>> {
-        let url_ref: &str = url;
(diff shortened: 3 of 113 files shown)


lukesandberg:
* **#93431** <a href="https://app.graphite.com/github/pr/vercel/next.js/93431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93431?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

crates/next-core/src/util.rs
```diff
@@ -215,7 +215,7 @@ pub enum PathType {
 /// Converts a filename within the server root into a next pathname.
 #[turbo_tasks::function]
 pub async fn pathname_for_path(
-    server_root: FileSystemPath,
+    server_root: &FileSystemPath,
     server_path: FileSystemPath,
```
lukesandberg: can be a ref, the clone below is unneeded

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -2309,7 +2309,7 @@ pub fn get_source_map_rope_operation(
 pub async fn project_trace_source_operation(
     container: ResolvedVc<ProjectContainer>,
     frame: StackFrame,
```
lukesandberg: can be a ref

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -83,10 +83,10 @@ impl FetchClientConfig {
     #[turbo_tasks::function(network)]
     pub async fn fetch(
         self: Vc<FetchClientConfig>,
-        url: RcStr,
+        url: &RcStr,
         user_agent: Option<RcStr>,
     ) -> Result<Vc<FetchResult>> {
-        let url_ref = &*url;
+        let url_ref: &str = url;
```
lukesandberg: is this binding needed?

turbopack/crates/turbo-tasks-fetch/tests/fetch.rs
```diff
@@ -316,7 +316,7 @@ async fn client_cache() {
 
     // a simple fetch that should always succeed
     #[turbo_tasks::function(operation)]
-    async fn simple_fetch_operation(server_url: RcStr, path: RcStr) -> anyhow::Result<()> {
+    async fn simple_fetch_operation(server_url: &RcStr, path: RcStr) -> anyhow::Result<()> {
```
lukesandberg: path should be a ref

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -1755,7 +1755,7 @@ impl std::fmt::Display for FileSystemPath {
 
 #[turbo_tasks::function]
 pub async fn rebase(
-    fs_path: FileSystemPath,
+    fs_path: &FileSystemPath,
     old_base: FileSystemPath,
     new_base: FileSystemPath,
```
lukesandberg: these can be references as well

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -3280,7 +3280,7 @@ mod tests {
         #[turbo_tasks::function(operation)]
         async fn write_symlink_stress_batch(
             fs: ResolvedVc<DiskFileSystem>,
-            symlinks_dir: FileSystemPath,
+            symlinks_dir: &FileSystemPath,
             updates: Vec<(usize, usize)>,
```
lukesandberg: this can be a ref and we can just iterate below, for a copy type like this that is superior

turbopack/crates/turbo-tasks-fs/src/read_glob.rs
```diff
@@ -26,11 +26,11 @@ pub async fn read_glob(directory: FileSystemPath, glob: Vc<Glob>) -> Result<Vc<R
 
 #[turbo_tasks::function(fs)]
 async fn read_glob_inner(
-    prefix: RcStr,
+    prefix: &RcStr,
     directory: FileSystemPath,
```
lukesandberg: can be a ref, we just need to propogate the ref into the _internal function

turbopack/crates/turbo-tasks-fs/src/read_glob.rs
```diff
@@ -144,12 +144,12 @@ pub async fn track_glob(
 
 #[turbo_tasks::function(fs)]
 async fn track_glob_inner(
-    prefix: RcStr,
+    prefix: &RcStr,
     directory: FileSystemPath,
```
lukesandberg: ditto

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -603,7 +603,7 @@ impl ChunkingContext for BrowserChunkingContext {
     }
 
     #[turbo_tasks::function]
-    async fn asset_url(&self, ident: FileSystemPath, tag: Option<RcStr>) -> Result<Vc<RcStr>> {
+    async fn asset_url(&self, ident: &FileSystemPath, tag: Option<RcStr>) -> Result<Vc<RcStr>> {
```
lukesandberg: tag can be a ref

turbopack/crates/turbo-tasks-testing/src/lib.rs
```diff
@@ -52,7 +52,15 @@ impl VcStorage {
     ) -> RawVc {
         let this = self.this.upgrade().unwrap();
         let handle = tokio::runtime::Handle::current();
-        let future = func.execute(this_arg, &*arg);
+        // The executor now takes ownership of the cached task type so the future can
+        // capture it and borrow `arg` for the entire execution. The testing harness doesn't
+        // wire CachedTaskType through its own task store, so we build a one-shot Arc here.
+        let task = std::sync::Arc::new(turbo_tasks::backend::CachedTaskType {
```
lukesandberg: import both Arc and CachedTaskType

turbopack/crates/turbo-tasks/src/task/from_task_input.rs
```diff
@@ -31,6 +50,23 @@ where
...
+where
+    T: Send + Sync + ?Sized,
+{
+    fn ref_from_task_input(from: &Vc<T>) -> &ResolvedVc<T> {
+        debug_assert!(
+            from.is_resolved(),
+            "Outer `Vc`s are always resolved before this is called"
+        );
+        // SAFETY: see the safety comment on the `unsafe impl`.
+        unsafe { &*(from as *const Vc<T> as *const ResolvedVc<T>) }
```
lukesandberg, line 66: should this be a transmute?

PR #93453, opened 2026-05-04
Include deployment id in `cacheHandlers` keys

mischnic:
How was this ever working for dev? In that case, `buildId = "development"` and so it always uses the same cache key


mischnic:
* **#93453** <a href="https://app.graphite.com/github/pr/vercel/next.js/93453?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93453?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

test/production/app-dir/use-cache-cross-deployment/app/prerender/page.tsx
unstubbable, line 1: This page seems to be unused.

test/e2e/app-dir/use-cache-cross-deployment/app/page.tsx
```diff
@@ -0,0 +1,30 @@
+import { Suspense } from 'react'
+import { cacheLife, cacheTag } from 'next/cache'
+import { connection } from 'next/server'
+
+async function getData() {
+  'use cache'
+
+  cacheLife({ revalidate: 6 })
+  cacheTag('modern')
```
unstubbable: nit: not needed for the test

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
```diff
@@ -0,0 +1,52 @@
...
+    // In the future, this assertion can be relaxed to only prevent sharing if the implementation
+    // changed.
+    it('should not have the same cache key across deployments', async () => {
+      async function execute(id: string) {
+        await next.stop()
+        if (envKey !== 'default') {
+          next.env[envKey] = id
+        }
+        try {
+          await next.start()
```
unstubbable: Wouldn't the test also succeed without implementing the deployment ID handling because we restart the server here and the cache handler only stores in memory?

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
```diff
@@ -0,0 +1,52 @@
...
+          let logs = next.getCliOutputFromHere()
+
+          const browser = await next.browser(`/`)
+          const initialData = await browser.elementById('data').text()
+          expect(initialData).toMatch(isoDateRegExp)
+
+          let match = logs().match(
+            /ModernCustomCacheHandler::get \["([A-Za-z0-9_-]+)","([0-9a-f]{2})+",\[\]\] \[ '_N_T_\/layout', '_N_T_\/page', '_N_T_\/', '_N_T_\/index' \]/
```
unstubbable: Why are you even using the custom cache handler if you're not asserting on the deployment or build ID being used for the cache key here? By the way, we could also use `NEXT_PRIVATE_DEBUG_CACHE=1` in `env` in `nextTestSetup` to generate logs with cache keys.

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
```diff
@@ -0,0 +1,52 @@
...
+          if (envKey !== 'default') {
+            delete next.env[envKey]
+          }
+        }
+      }
+
+      let key1 = await execute('value-1')
+      let key2 = await execute('value-2')
+      // Second run should not use the same key
+      expect(key1).not.toBe(key2)
```
unstubbable: This assertion doesn't work in dev.

unstubbable (commented):
> How was this ever working for dev? In that case, `buildId = "development"` and so it always uses the same cache key

In dev we also have the `hmrRefreshHash` as part of the cache key.

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
mischnic: I'm not asserting the cache handler's content. I'm asserting the cache keys that the cache handler prints out

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
mischnic: I was trying to assert behavior, as opposed to asserting implementation details.

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
mischnic: It indeed fails because it reuses the cache. But is that intended behavior? What happens if you use a remote cache in dev?

unstubbable: `'use cache: remote'` doesn't have that problem because it also only uses the in-memory one with `next dev` (or `next start`). But a custom remote cache handler would indeed have the problem with Turbopack, because the `hmrRefreshHash` is just an incrementing number there and not a real hash. I guess another reason we need the implementation hash.

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
unstubbable: Oh right, I missed that. Makes sense now.

test/e2e/app-dir/use-cache-cross-deployment/use-cache-cross-deployment.test.ts
mischnic: So I'll just move this to `test/production` for now?

unstubbable: yeah, sounds good to me

test/production/app-dir/use-cache-cross-deployment/app/prerender/page.tsx
mischnic: I wanted to assert this as well. Added now

unstubbable (approved):

PR #93454, opened 2026-05-04
Turbopack: lazy aggregation optimize via persistent pending flag

sokra:
### What?

Make the aggregation `optimize_queue` in the Turbopack persistent backend bounded and lazy. Cap the in-memory queue size, persist an `optimization_pending` flag per task, drain the queue per `process()` call with a per-queue lifetime budget, and recover dropped optimizations opportunistically via the flag instead of via an unbounded scheduler-side queue.

### Why?

The previous implementation pushed an `OptimizeJob` for every `push_optimize_task` call into a single in-memory queue with no upper bound. On large workloads (or pathological aggregation churn), this queue could grow very large and cause the thread that scheduled the optimization to do unbounded work, regressing latency for the operations that triggered the schedule.

The goals of this change:

- Bound the worst-case work any single `process()` call does for optimizations (per-queue budget).
- Bound the in-memory queue size so memory use is predictable.
- Avoid losing optimizations: anything we drop must be eventually recovered.
- Keep the common fast path cheap — no extra `Meta`-category guard acquisitions when the optimization flows through normally.

### How?

Persist a new `optimization_pending` flag on `TaskStorage` (`storage_schema.rs`) and use it to drive lazy recovery in `aggregation_update.rs`:

- `push_optimize_task` only enqueues an in-memory `OptimizeJob` if the queue is under `MAX_OPTIMIZE_QUEUE_SIZE` (10000) and the per-queue lifetime budget `MAX_OPTIMIZATIONS_PER_QUEUE` (1000) hasn't been exhausted. If we can't enqueue, we set `optimization_pending = true` on the task so a future operation that visits this task will re-discover and re-enqueue the optimization.
- Every `AggregationUpdateJob` handler calls `check_optimization_pending` on the primary task(s) it touches, which re-enqueues the optimize job if the flag is set (and the queue/budget allow).
- `process()` drains the `optimize_queue` one job at a time (preserving the original "root first" ordering), counting against the per-queue budget. Once the budget is exhausted, further `OptimizeJob`s in the queue are dropped and the flag is left set on those tasks (so they recover later).
- `optimize_task` clears `optimization_pending` at entry so the recovery loop eventually settles.
- The flag is **only** written on the drop path — the common case (enqueue → process normally) does not touch `optimization_pending`, so no `Meta`-category guard contention is added on the hot path.
- `OptimizeJob` carries a best-effort `flag_already_set` snapshot so that when a job is dropped at process time and the snapshot says the flag was already set, we skip the redundant write entirely. Most jobs originating from `check_optimization_pending` (the recovery path) and `optimize_task`'s self-re-enqueue carry this hint.
- `try_enqueue_optimize_job` is `#[must_use]` so the contract \"if this returns false, set the flag\" is enforced at the type level. `lock_and_mark_optimization_pending` is shared between `push_optimize_task_by_id` and the budget-exhausted drop branch.
- `optimizations_executed` is intentionally persisted with the queue so that suspending and resuming the queue cannot reset the per-queue budget.


turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
```diff
@@ -736,15 +752,36 @@ impl Eq for BalanceJob {}
...
+    ///   record a different snapshot — the original entry's value is kept).
+    ///
+    /// A stale snapshot is still correct, just possibly suboptimal:
+    /// - If the snapshot is `false` but the flag is now `true`: at drop we acquire a guard and
+    ///   re-set the flag (a no-op write — slightly wasted work, but correct).
+    /// - If the snapshot is `true` but the flag is now `false`: at drop we skip the guard. The
+    ///   only way this can happen is that some other path ran `optimize_task` for this task and
+    ///   cleared the flag — i.e., the optimization we were about to drop has already been
+    ///   performed. Skipping the flag set is exactly what we want.
+    flag_already_set: bool,
```
lukesandberg: nit: optimization_pending_flag_already_set?  wordy, but as is it is a bit ambiguous

lukesandberg (approved):
is there a way to leverage this to achieve more parallelism in large queues

e.g. if a queue starts spilling we can spawn another thread to start processing them?

sokra:
> is there a way to leverage this to achieve more parallelism in large queues
> 
> e.g. if a queue starts spilling we can spawn another thread to start processing them?

I think it will already do that in the current implementation.
Thread 1 starts doing some work, found a lot of optimizations, but can't handle all of them. So it defers them by setting the flag and stops. After that some other threads find these flags and do some work. So they are naturally parallelizing by deferring within a single compilation.

PR #93455, opened 2026-05-04
[ci] Move Turbopack, Rspack and font data workflows off of `RELEASE_BOT_GITHUB_TOKEN`

eps1lon:
`RELEASE_BOT_GITHUB_TOKEN` is no longer working. Instead we use a dedicated GH app to push, sign and create commits that trigger workflows.

Based off of https://github.com/vercel/next.js/pull/93245

I extracted https://github.com/vercel/next.js/pull/93285 into separate utils since we're no using those in 4 places (5 once I migrated the React sync which requires more work).

### DIFF
--- a/scripts/automated-update-workflow.js
+++ b/scripts/automated-update-workflow.js
@@ -1,19 +1,37 @@
+// @ts-check
 const { promisify } = require('util')
 const { Octokit } = require('octokit')
 const { exec: execOriginal } = require('child_process')
+const {
+  createSignedCommit,
+  upsertBranchRef,
+} = require('./github-utils/signed-commit')
 
 const exec = promisify(execOriginal)
 
 const {
-  GITHUB_TOKEN = '',
+  RELEASE_GITHUB_TOKEN = '',
+  PR_GITHUB_TOKEN = '',
+  RELEASE_GITHUB_APP_SLUG = '',
+  RELEASE_GITHUB_APP_USER_ID = '',
   SCRIPT = '',
   BRANCH_NAME = 'unknown',
   PR_TITLE = 'Automated update',
   PR_BODY = '',
 } = process.env
 
-if (!GITHUB_TOKEN) {
-  console.log('missing GITHUB_TOKEN env')
+if (!RELEASE_GITHUB_TOKEN) {
+  console.log('missing RELEASE_GITHUB_TOKEN env')
+  process.exit(1)
+}
+if (!PR_GITHUB_TOKEN) {
+  console.log('missing PR_GITHUB_TOKEN env')
+  process.exit(1)
+}
+if (!RELEASE_GITHUB_APP_SLUG || !RELEASE_GITHUB_APP_USER_ID) {
+  console.log(
+    'missing RELEASE_GITHUB_APP_SLUG or RELEASE_GITHUB_APP_USER_ID env'
+  )
   process.exit(1)
 }
 if (!SCRIPT) {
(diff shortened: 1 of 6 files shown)


scripts/automated-update-workflow.js
```diff
@@ -1,34 +1,57 @@
...
+if (!PR_GITHUB_TOKEN) {
+  console.log('missing PR_GITHUB_TOKEN env')
+  process.exit(1)
+}
+if (!RELEASE_GITHUB_APP_SLUG || !RELEASE_GITHUB_APP_USER_ID) {
+  console.log(
+    'missing RELEASE_GITHUB_APP_SLUG or RELEASE_GITHUB_APP_USER_ID env'
+  )
   process.exit(1)
 }
```
bgw:
```suggestion
if (!RELEASE_GITHUB_APP_SLUG) {
  console.log('missing RELEASE_GITHUB_APP_SLUG env')
  process.exit(1)
}
if (!RELEASE_GITHUB_APP_USER_ID) }
  console.log('missing RELEASE_GITHUB_APP_USER_ID env')
  process.exit(1)
}
```

scripts/github-utils/signed-commit.js
```diff
@@ -0,0 +1,302 @@
...
+  })
+
+  return blob.sha
+}
+
+/**
+ * Build a GitHub tree matching a local commit, creating new blob objects for
+ * changed files in parallel while preserving deletions and submodules.
+ */
+async function createTreeFromLocalCommit({
```
bgw, line 110: GitHub offers a GraphQL API that lets you do this in a single request, which they seem to recommend over the REST blob+tree+commit APIs: https://github.blog/changelog/2021-09-13-a-simpler-api-for-authoring-commits/

That's what I used for front: https://github.com/vercel/front/blob/main/tooling/github-actions/utils/ghcommit.ts

I also used octokit there. Raw `fetch` calls are fine too, but it looks like `scripts/automated-update-workflow.js` is already pulling in octokit.

scripts/github-utils/signed-commit.js
```diff
@@ -0,0 +1,302 @@
...
+ * Call the GitHub REST API and include response bodies in thrown errors so
+ * workflow failures show actionable details.
+ */
+async function githubRequest(token, method, path, body) {
+  const response = await fetch(`https://api.github.com${path}`, {
+    method,
+    headers: {
+      Accept: 'application/vnd.github+json',
+      Authorization: `Bearer ${token}`,
+      'X-GitHub-Api-Version': '2022-11-28',
```
bgw, line 15: LLMs love using an ancient API version

```suggestion
      'X-GitHub-Api-Version': '2026-03-10',
```

bgw (approved):

bgw:
FYI, we'll probably need to move again from app tokens to OIDC via https://vercel.slack.com/archives/C0AM84PRSGL/p1777319001082569 at some point.

scripts/github-utils/signed-commit.js
eps1lon: I'll do that together with the GraphQL refactor in a follow-up. This is just moved code. Priority is getting everything back up.

scripts/github-utils/signed-commit.js
eps1lon: Will do in a follow-up

scripts/github-utils/signed-commit.js
eps1lon: Decided to not do that. That's the default version so it can't be that bad. Changing it does put as through breaking changes so there should be a benefit.

PR #93460, opened 2026-05-04
Trace-server: Fix bottom up, and reduce allocations in turbopack-trace-server bottom-up grouping

sokra:
### What?

Speeds up the bottom-up grouping pass in `turbopack-trace-server` by removing per-span `RcStr` allocations and fixes the hash map using the correct hasher for the grouping `HashMap`.

### Why?

When loading large traces, building the bottom-up graph spent measurable time (1) allocating fresh `RcStr` values from `&str` keys for every span just to use them as `HashMap` keys, and (2) hashing those keys with the default randomized hasher. Both are avoidable: the underlying spans already own `RcStr`s, and the `StringTupleRef` equivalence-based lookup needs `FxHasher` anyway because `RcStr`'s `Hash` impl only matches `&str`'s `Hash` under `FxHasher`.

### How?

- Change `nice_name`/`group_name`/`args` accessors on `SpanRef` and friends to return `&RcStr` instead of `&str`. The bottom-up grouping code can now clone the existing `RcStr` (a cheap ref-count bump) instead of allocating a new one from a `&str`.
- Switch the `(RcStr, RcStr) -> SpanBottomUpBuilder` map in `bottom_up.rs` to use `FxBuildHasher`. This is required for the `StringTupleRef` equivalence lookup to produce matching hashes for the owned `(RcStr, RcStr)` keys, and it also removes the overhead of the default randomized hasher.
- Update the `string_tuple_ref` test to use `FxBuildHasher` accordingly.

No behavior changes; this is a pure refactor for performance inside the trace viewer tool.

### DIFF
--- a/turbopack/crates/turbopack-trace-server/src/bottom_up.rs
+++ b/turbopack/crates/turbopack-trace-server/src/bottom_up.rs
@@ -1,6 +1,7 @@
 use std::{env, sync::Arc};
 
 use hashbrown::HashMap;
+use rustc_hash::FxBuildHasher;
 use turbo_rcstr::RcStr;
 
 use crate::{
@@ -12,7 +13,7 @@ use crate::{
 pub struct SpanBottomUpBuilder {
     // These values won't change after creation:
     pub self_spans: Vec<SpanIndex>,
-    pub children: HashMap<(RcStr, RcStr), SpanBottomUpBuilder>,
+    pub children: HashMap<(RcStr, RcStr), SpanBottomUpBuilder, FxBuildHasher>,
     pub example_span: SpanIndex,
 }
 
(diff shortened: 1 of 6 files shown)


turbopack/crates/turbopack-trace-server/src/bottom_up.rs
```diff
@@ -65,7 +66,7 @@ pub fn build_bottom_up_graph<'a>(
                 .from_key(&StringTupleRef(category, name))
```
lukesandberg, line 66: do we need StringTupleRef, i think we can actually just use borrowed keys in the roots map since roots doesn't survive this function

lukesandberg (approved):

turbopack/crates/turbopack-trace-server/src/bottom_up.rs
sokra: It doesn't work without it, due to the tuple `(&RcStr, &RcStr) != &(RcStr, RcStr)`

PR #93461, opened 2026-05-04
[turbopack] Resolve inherited tsconfig living outside Next.js app directory

eps1lon:
Turbopack used to crash during `next build` if the `tsconfig.js` inherited (`extends`) from a `tsconfig` outside the directory:

> Build error occurred
Error: Turbopack build failed with 1 errors:
./tsconfig.json
An issue occurred while parsing a tsconfig.json file.
extends: "../tsconfig.base.json" doesn't resolve correctly
>
>
>    at ignore-listed frames

Hit this case when working on https://github.com/vercel/next.js/pull/93377


eps1lon:
Can't be supported. File must at least be inside the monorepo

PR #93475, opened 2026-05-04
fix: `Infinity` cacheLife in next config should be respected

ztanner:
Raw `Infinity` is not JSON-safe and can stringify to `null` when config is serialized, causing cache life profiles to behave incorrectly. 

This PR normalizes `Infinity` cache life values to `INFINITE_CACHE` for both `next.config` profiles and inline `cacheLife({ ... })` profiles and uses a shared normalization/validation util in both places. This keeps `Infinity` as an accepted user-facing value while ensuring internals use the existing sentinel.

closes NAR-95


gaearon:
should be rolled into https://github.com/vercel/next.js/pull/95493

PR #92300, opened 2026-04-03
Refactor effects system: dedup/conflict detection, simplified Effect trait

sokra:
### What?

Refactors the turbo-tasks effects system to support duplicate detection, conflict detection, and state tracking across effect applications. Also adds a ptr_eq fast path to `ReadRef::PartialEq`.

### Why?

The existing effects system had several problems:

- **No duplicate detection**: The same effect (e.g., writing the same content to the same path) would be re-applied every time, wasting I/O
- **No conflict detection**: Two effects writing different content to the same path could cause an invalidation loop where the two writes invalidate each other, keeping a CPU busy-looping. On production builds this would crash with `Dependency tracking is disabled so invalidation is not allowed`
- **Unnecessarily generic**: The old `DynEffect`→`DynEffectApplyFuture` chain wrapped errors in `Arc<dyn EffectError>`, plus a `DynEffectApplyFuture` chain, all to work around the fact that `anyhow::Error` doesn't implement `std::error::Error`
- **Complex dedup state machine**: `EffectInstance` used a `Mutex<EffectState>` with 4 states (`NotStarted`/`Started`/`Finished`/`Invalid`), `Event`/`EventListener`, and `spawn()` to ensure idempotent application — all of which is now unnecessary since dedup is handled at the `Effects::apply` level
- **Task-local context**: `ApplyEffectsContext` was a task-local typed map used only for a directory-creation cache, requiring `spawn()` + scope propagation
- **Slow `ReadRef` equality**: `ReadRef::PartialEq` always performed deep content comparison even when both refs pointed to the same allocation

#### Changed files

| File | Change |
|------|--------|
| `turbo-tasks/src/effect.rs` | Rewrote `Effect` trait, `DynEffect`, `EffectStateStorage`, `Effects::apply()` |
| `turbo-tasks/src/lib.rs` | Updated re-exports (`EffectError`, `EffectStateStorage`) |
| `turbo-tasks/src/read_ref.rs` | Added `ptr_eq` fast path to `ReadRef::PartialEq` |
| `turbo-tasks-fs/src/lib.rs` | Updated `WriteEffect`/`WriteLinkEffect` impls, restored `AnyhowWrapper`, removed `register_write_invalidator` |
| `turbo-tasks-fs/src/invalidator_map.rs` | Simplified — removed `WriteContent` enum, invalidators now use `()` values |
| `turbo-tasks-fs/src/invalidation.rs` | Deleted — `Write`/`WriteKind` invalidation reason types no longer needed |

[...]

PR #93476, opened 2026-05-04
Turbopack: Cache effect errors, replace the write_lock Mutex with a lazily created Notify instance

bgw:
This is a mostly-by-hand refactor of some of the caching logic in https://github.com/vercel/next.js/pull/92300

- **Cache errors.** We often end up applying effects multiple times in frequent succession, and many filesystem effects already have their own backoff/retry loops, which can make them very slow when they fail, so it seems generally better to cache the failures.
- **Add `EffectLastApplied` enum:** It's a matter of opinion, but I think the `EffectLastApplied` enum makes it easier to reason about the different states in the system. `Option` was difficult to reason about, and it would've been even more confiusing if it was an `Option<Result<(), Arc<dyn EffectError>>`.
- `tokio::sync::Mutex<()>` is overkill here, `tokio::sync::Notify` is a simpler (and smaller) primitive that does the core of what we needed the `Mutex` for. We could also use `EventListener` here, but I didn't. (`EventListener` is really just smol's version of `Notify`)
- We can lazily allocate the `Notify` instance inside of `last_applied`. We have to grab this lock anyways to check for the cached fast-path. If we're not in the fast-path, we can re-use the lock guard to create and store the `Notify` instance.
- **Perf:** The old code's slow path had to always lock `last_applied` three times. Now, if there's no concurrent writer (common case), it only locks twice: Once before application, and once after.
- **Perf:** Use a smallvec when computing unique indices to avoid a bunch of heap allocations. Unless there's a bug, we don't expect any of the keys to have more than one value (returns an error).
- **Perf:** Store keys as `Box<[u8]>` instead of `Vec<u8>`, saving 8 bytes per key.

Currently, if `effect.dyn_apply()` panics, we just do our best to mark the effect as unapplied to avoid a deadlock. This could alternatively poison the object.

## Memory Usage

Memory usage can increase if we're storing errors (heap allocated, stored in `Arc`) we previously wouldn't have, but if we assume that errors are uncommon, this refactor decreases the size of the data structures involved.

Before this change, eagerly allocating a `tokio::sync::Mutex` was costing 40 bytes per effect, making the entire `EffectEntryState` struct 64 bytes:

After the changes, the entire `EffectLastApplied` enum is 40 bytes

After wrapping the `EffectLastApplied` enum in a `parking_lot::Mutex`, it comes in at 48 bytes (after alignment), 16 bytes smaller than the old struct.

### DIFF
--- a/turbopack/crates/turbo-tasks/src/effect.rs
+++ b/turbopack/crates/turbo-tasks/src/effect.rs
@@ -6,10 +6,12 @@ use std::{
     sync::{Arc, OnceLock},
 };
 
-use anyhow::{Result, bail};
+use anyhow::Result;
 use futures::{StreamExt, TryStreamExt};
-use parking_lot::Mutex;
+use parking_lot::{Mutex, MutexGuard};
 use rustc_hash::{FxHashMap, FxHashSet};
+use smallvec::SmallVec;
+use tokio::sync::Notify;
 use tracing::Instrument;
 use turbo_dyn_eq_hash::DynPartialEq;
 
@@ -71,31 +73,25 @@ pub trait Effect: TraceRawVcs + NonLocalValue + Send + Sync + 'static {
 pub trait EffectError: StdError + TraceRawVcs + NonLocalValue + Send + Sync + 'static {}
 impl<T> EffectError for T where T: StdError + TraceRawVcs + NonLocalValue + Send + Sync + 'static {}
 
-/// Per-key entry in the effect state storage.
-///
-/// - `last_applied`: the value that was last successfully written (sync-readable for fast-path
-///   dedup)
-/// - `write_lock`: async mutex held during the actual write; ensures only one concurrent write per
-///   key
-struct EffectStateEntry {
-    last_applied: Mutex<Option<Box<dyn Any + Send + Sync>>>,
-    write_lock: tokio::sync::Mutex<()>,
+enum EffectLastApplied {
+    Unapplied,
+    InProgress {
+        write_notify: Arc<Notify>,
+    },
+    Applied {
+        value: Box<dyn Any + Send + Sync>,
+        result: Result<(), Arc<dyn EffectError>>,
+    },
 }
 
-impl Default for EffectStateEntry {
-    fn default() -> Self {
-        Self {
-            last_applied: Mutex::new(None),
-            write_lock: tokio::sync::Mutex::new(()),
-        }
-    }
-}
+/// Per-key entry in the effect state storage.
+type EffectStateEntry = Arc<Mutex<EffectLastApplied>>;
 
 /// Shared state storage for tracking applied effects. Stored on the filesystem implementation
 /// (e.g. DiskFileSystemInner).
 #[derive(Default)]
 pub struct EffectStateStorage {
-    effect_state: Mutex<FxHashMap<Vec<u8>, Arc<EffectStateEntry>>>,
+    effect_state: Mutex<FxHashMap<Vec<u8>, EffectStateEntry>>,
 }
 
 // Private wrapper trait to allow dynamic dispatch of an `Effect`. This is similar to the pattern
@@ -130,11 +126,16 @@ where
     }
 
     fn dyn_apply<'a>(&'a self) -> DynEffectApplyFuture<'a> {
-        Box::pin(async move { Effect::apply(self).await.map_err(anyhow::Error::from) })
+        Box::pin(async move {
+            Effect::apply(self)
+                .await
+                .map_err(|err| Arc::new(err) as Arc<_>)
+        })
     }
 }
 
-type DynEffectApplyFuture<'a> = Pin<Box<dyn Future<Output = Result<()>> + Send + 'a>>;
+type DynEffectApplyFuture<'a> =
+    Pin<Box<dyn Future<Output = Result<(), Arc<dyn EffectError>>> + Send + 'a>>;
 
 /// A trait to emit a task effect as collectible. This trait only has one implementation,
 /// `EffectInstance` and no other implementation is allowed. The trait is private to this module so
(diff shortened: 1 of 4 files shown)


bgw:
* **#93552** <a href="https://app.graphite.com/github/pr/vercel/next.js/93552?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93476** <a href="https://app.graphite.com/github/pr/vercel/next.js/93476?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93476?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -71,37 +73,30 @@ pub trait Effect: TraceRawVcs + NonLocalValue + Send + Sync + 'static {
...
-        }
-    }
+enum EffectLastApplied {
+    Unapplied,
+    InProgress {
+        write_notify: Arc<Notify>,
+    },
+    Applied {
+        value: Box<dyn Any + Send + Sync>,
+        result: Result<(), Arc<dyn EffectError>>,
```
sokra, line 84:
```suggestion
        result: Result<(), SharedError>,
```

We have an helper for this: SharedError

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -310,90 +316,127 @@ impl Effects {
...
-                                ));
+                    let mut indices = Vec::with_capacity(by_key.len());
+                    for (key, group) in by_key {
+                        if group.len() > 1 {
+                            let first_value = self.effects[group[0]].inner.value_dyn();
+                            for &idx in &group[1..] {
+                                if !self.effects[idx].inner.eq_value_dyn(&*first_value) {
+                                    return Err(Arc::new(ConflictingEffectError {
+                                        key_len: key.len(),
+                                    }));
```
sokra, line 336: Might also want to use `SharedError` here

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -310,90 +316,127 @@ impl Effects {
...
+                                        return result.clone();
+                                    } else {
+                                        break create_notify_guard(last_applied_guard);
+                                    }
+                                }
+                                EffectLastApplied::InProgress { write_notify } => {
+                                    in_progress_notified =
+                                        Box::pin(write_notify.clone().notified_owned());
+                                    // enable the notify before we drop the last_applied_guard
+                                    in_progress_notified.as_mut().enable();
```
sokra: That's an "interesting" API...

I feel like it would be better to use `Event` and `EventListener` here:
* We already use that everywhere else
* We have a hanging_detection feature that depends on it.
* It doesn't need to be an `Arc<Notify>` since `Event` is already a lazy initialized arc internally.

turbopack/crates/turbo-tasks/src/effect.rs
bgw: I don't like `SharedError`, and in this specific case `SharedError` doesn't work well because:
- `SharedError` doesn't implement `TraceRawVcs` or `NonLocalValue`, and it can't correctly implement those because it wraps `anyhow::Error`, which could contain anything. `EffectError` includes these bounds, forcing new `EffectError`s to implement them correctly.
- `SharedError` is a wrapper around `anyhow::Error`, and I'm trying to encourage effects to use structured errors. You can stuff structured errors inside of `anyhow::Error` and get them out later, but that's not what the signature is encouraging. The API design doesn't encourage a pit of success.
  - We want effects to use structured errors, because we want most of those errors to be convertible into issues.
- We don't need `Encode`/`Decode` here, but `SharedError`'s `Encode`/`Decode` is a footgun. If you accidentally start using it, you'll lose the structured representation of the error. It's forced to convert to/from a string, losing any structured representation, because it wraps `anyhow::Error` which can't properly implement `Encode`/`Decode`.

turbopack/crates/turbo-tasks/src/effect.rs
bgw: Sure, I can switch it to `Event`+`EventListener`. I'm pretty neutral on this and the APIs are very similar.

bgw: > It doesn't need to be an `Arc<Notify>` since `Event` is already a lazy initialized arc internally.

It looks like this isn't true. Internally it's an `AtomicPtr` and it does not implement `Clone`.

However, it turns out that we can be a bit more clever about how we store this (not specific to `Event` or `Notify`). We end up taking the lock in all the places we need access to the `Event`/`Notify`, so we can just grab references there and avoid the `Arc`+`clone`. It does require adding a few `unreachable` assertions though that make the code a little harder to follow.

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -310,90 +317,134 @@ impl Effects {
...
-                    by_key.entry(key).or_default().push(i);
-                }
+            let unique_indices = self
+                .unique_indices
+                .get_or_init(|| {
+                    let mut by_key: FxHashMap<Box<[u8]>, SmallVec<[usize; 1]>> =
+                        FxHashMap::default();
+                    for (i, effect) in self.effects.iter().enumerate() {
+                        let key = effect.inner.key();
+                        by_key.entry(key).or_default().push(i);
```
bgw, line 325: Move the ConflictingEffectError up here, and then `by_key` doesn't need to be stored in a vec

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -130,11 +126,16 @@ where
     }
 
     fn dyn_apply<'a>(&'a self) -> DynEffectApplyFuture<'a> {
-        Box::pin(async move { Effect::apply(self).await.map_err(anyhow::Error::from) })
+        Box::pin(async move {
+            Effect::apply(self)
+                .await
+                .map_err(|err| Arc::new(err) as Arc<_>)
```
bgw:
```suggestion
                .map_err(|err| Arc::new(err) as _)
```

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -310,90 +317,134 @@ impl Effects {
...
-                    *entry.last_applied.lock() = None;
-
-                    // Apply the effect
-                    effect.dyn_apply().await?;
-
-                    // Store the new value (sync)
-                    *entry.last_applied.lock() = Some(effect.value_dyn());
-
-                    Ok(())
+                    let create_event_guard = |mut last_applied_guard: MutexGuard<'_, _>| {
```
bgw: Maybe name this differently

sokra (approved):

turbopack/crates/turbo-tasks/src/effect.rs
lukesandberg: begin_effect?  start_effect?  enter_effect?

lukesandberg (approved):

turbopack/crates/turbo-tasks/src/effect.rs
bgw: I'm moving this into https://github.com/vercel/next.js/pull/93552, because I've realized that `DynEffect::value_dyn` should just be `DynEffect::value_hash`, and return a u128 instead of a `Box<dyn Any>`, which lets us simplify a lot of other things.

PR #93478, opened 2026-05-05
[turbopack] Simplify local task tracking

lukesandberg:
### What?

Replace the `FuturesUnordered<Either<JoinHandle, ...>>` that tracks a global task's in-flight local tasks with a small in-place counter+`Event` bundled into a new `LocalTaskTracker` that lives directly on `CurrentTaskState`.

### Why?

`wait_for_local_tasks` only needs a barrier — "have all in-flight local tasks completed?" — not a stream of outputs. `FuturesUnordered` was overkill for that, and its supporting machinery cost an allocation per parent task plus an allocation per local task, plus repeated lock acquisitions per registration and per completion.

Per parent task execution, this PR removes:

- **One `FuturesUnordered` allocation.** The lazy `Option<FuturesUnordered<…>>` is gone; the counter+event is inline on `CurrentTaskState`.
- **One `tokio::sync::oneshot::channel()` allocation per local task.** `priority_runner::schedule_with_join_handle` (and `JoinHandle`) is deleted; local tasks now use plain `schedule`.
- **One intrusive `FuturesUnordered::Task` node allocation per local task.**
- **One `RwLock<CurrentTaskState>` write-lock acquisition per local task spawn.** The pre-existing `create_local_task` write lock now also bumps the in-flight counter, so we no longer take a separate lock to push into the tracker.
- **One `RwLock<CurrentTaskState>` write-lock acquisition per local task completion.** The pre-existing `Scheduled → Done` write lock now also decrements the counter, notifies per-task waiters, and (if the count hit zero) notifies the collective wait event — all under one lock.

Net effect on hot paths:

- **Spawn**: was 2 allocations + 2 locks; now 0 allocations + 1 lock (the existing `create_local_task` lock).
- **Wait when nothing was spawned**: a single read-lock that finds `in_flight == 0` and returns. No allocation, no listener, no await.

### How?

- New `LocalTaskTracker` ([`turbopack/crates/turbo-tasks/src/local_task_tracker.rs`](turbopack/crates/turbo-tasks/src/local_task_tracker.rs)) bundles:
  - `tasks: Vec<LocalTask>` (was a separate field on `CurrentTaskState`),
  - `in_flight: u32` (plain integer; the surrounding `RwLock` provides synchronization),
  - `done: Event` (notified on transitions to zero).
- `CurrentTaskState::local_tasks` is now `LocalTaskTracker` instead of `Vec<LocalTask>`. The wait-group is part of the same struct, so a parent task that doesn't spawn any local tasks pays nothing beyond the inline fields.
- `LocalTaskTracker::create` pushes a new `Scheduled` entry and increments the counter.
- `LocalTaskTracker::complete` does the `Scheduled → Done` swap, fires the per-task `done_event` (waking `try_read_local_output` waiters), decrements the counter, and notifies the collective `done` event if the count hit zero — all while holding one write lock.
- `wait_for_local_tasks` is the standard double-check pattern (snapshot the counter, register a listener, re-check, await) against the tracker. No loop after the await — the codebase contract is that no new local tasks (or detached test futures) can be registered after the parent's user body returns, except from inside an already-in-flight scope, so a single notify is sufficient.
- `spawn_detached_for_testing` uses a tiny `register_detached` / `dec_in_flight` pair (test path; no RAII guard, panics abort upstream).
- `priority_runner::schedule_with_join_handle`, `JoinHandle`, the `Option<Sender<()>>` field on `HeapItem`, and the matching `tx.send(())` in `WorkerFuture::poll` are all deleted; nothing else used them.

Unit tests for the tracker cover balanced create/complete, balanced detached register/dec, and that `complete` wakes per-task listeners. The existing detached integration tests (including the nested `spawn_detached_for_testing` case in `detached.rs`) still pass.

### DIFF
--- a/turbopack/crates/turbo-tasks/src/local_task_tracker.rs
+++ b/turbopack/crates/turbo-tasks/src/local_task_tracker.rs
@@ -1,5 +1,7 @@
 //! Tracker for the local tasks of a single global task execution.
 
+use std::fmt::Display;
+
 use crate::{
     OutputContent,
     event::{Event, EventListener},
@@ -36,11 +38,14 @@ impl LocalTaskTracker {
     /// Tasks should call [`complete`] when the task transitions to `Done`.
     ///
     /// [`complete`]: LocalTaskTracker::complete
-    pub(crate) fn create(&mut self, task: LocalTask) -> LocalTaskId {
-        debug_assert!(
-            matches!(task, LocalTask::Scheduled { .. }),
-            "newly created local tasks must start in the Scheduled state"
-        );
+    pub(crate) fn create(
+        &mut self,
+        task_type: impl Display + Send + Sync + 'static,
+    ) -> LocalTaskId {
+        let task = LocalTask::Scheduled {
+            done_event: Event::new(move || move || format!("LocalTask({task_type})::done_event")),
+        };
+
         self.tasks.push(task);
         self.in_flight += 1;
         // generate a one-indexed id from len() -- we just pushed so len() is >= 1
@@ -62,7 +67,7 @@ impl LocalTaskTracker {
         // Do this before decrementing in-flight so we can ensure that tasks are always completed
         // before the task completes.
         done_event.notify(usize::MAX);
-        self.dec_in_flight();
+        self.decrement_in_flight();
     }
 
     /// Test-only: register an in-flight detached future (`spawn_detached_for_testing`). No
@@ -79,7 +84,7 @@ impl LocalTaskTracker {
     /// [`complete`] which decrements as part of the slot transition.
     ///
     /// [`complete`]: LocalTaskTracker::complete
-    pub(crate) fn dec_in_flight(&mut self) {
+    pub(crate) fn decrement_in_flight(&mut self) {
         debug_assert!(
             self.in_flight > 0,
             "LocalTaskTracker::dec_in_flight without matching increment"
@@ -92,6 +97,7 @@ impl LocalTaskTracker {
 
     /// Current in-flight count. Cheap snapshot for the early-return path in
     /// `wait_for_local_tasks`.
+    #[cfg(test)]
     pub(crate) fn in_flight(&self) -> u32 {
         self.in_flight
     }
@@ -99,8 +105,12 @@ impl LocalTaskTracker {
     /// Listen for the next "in-flight reached zero" notification. Used by
     /// `wait_for_local_tasks` together with `in_flight()` for the standard double-check
     /// pattern that avoids lost wakeups.
-    pub(crate) fn listen(&self) -> EventListener {
-        self.done.listen()
+    pub(crate) fn listen_for_in_flight(&self) -> Option<EventListener> {
+        if self.in_flight > 0 {
+            Some(self.done.listen())
+        } else {
+            None
+        }
     }
 }
 
@@ -121,9 +131,7 @@ mod tests {
         let mut tracker = LocalTaskTracker::new();
         assert_eq!(tracker.in_flight(), 0);
 
-        let id = tracker.create(LocalTask::Scheduled {
-            done_event: Event::new(|| || "test".to_string()),
-        });
+        let id = tracker.create("test");
         assert_eq!(tracker.in_flight(), 1);
 
         tracker.complete(id, dummy_output());
@@ -136,9 +144,9 @@ mod tests {
         tracker.register_detached();
         tracker.register_detached();
         assert_eq!(tracker.in_flight(), 2);
-        tracker.dec_in_flight();
+        tracker.decrement_in_flight();
         assert_eq!(tracker.in_flight(), 1);
-        tracker.dec_in_flight();
+        tracker.decrement_in_flight();
         assert_eq!(tracker.in_flight(), 0);
     }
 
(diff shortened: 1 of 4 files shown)


lukesandberg:
* **#93478** <a href="https://app.graphite.com/github/pr/vercel/next.js/93478?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#93474** <a href="https://app.graphite.com/github/pr/vercel/next.js/93474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/local_task_tracker.rs
```diff
@@ -0,0 +1,161 @@
...
+    /// Create a new `Scheduled` local task and returns the new task id.
+    ///
+    /// Tasks should call [`complete`] when the task transitions to `Done`.
+    ///
+    /// [`complete`]: LocalTaskTracker::complete
+    pub(crate) fn create(&mut self, task: LocalTask) -> LocalTaskId {
+        debug_assert!(
+            matches!(task, LocalTask::Scheduled { .. }),
+            "newly created local tasks must start in the Scheduled state"
+        );
```
sokra: You should probably just pass the LocalTask::Scheduled fields to this method and construct the LocalTask enum here. That would bake this debug_assert into the type system.

turbopack/crates/turbo-tasks/src/local_task_tracker.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub(crate) fn register_detached(&mut self) {
+        self.in_flight += 1;
+    }
+
+    /// Decrement the in-flight counter and notify the collective `done` event if it reached
+    /// zero. Used by the test-only detached path; the production path goes through
+    /// [`complete`] which decrements as part of the slot transition.
+    ///
+    /// [`complete`]: LocalTaskTracker::complete
+    pub(crate) fn dec_in_flight(&mut self) {
```
sokra:
```suggestion
    pub(crate) fn decrement_in_flight(&mut self) {
```

turbopack/crates/turbo-tasks/src/local_task_tracker.rs
```diff
@@ -0,0 +1,161 @@
...
+    pub(crate) fn in_flight(&self) -> u32 {
+        self.in_flight
+    }
+
+    /// Listen for the next "in-flight reached zero" notification. Used by
+    /// `wait_for_local_tasks` together with `in_flight()` for the standard double-check
+    /// pattern that avoids lost wakeups.
+    pub(crate) fn listen(&self) -> EventListener {
+        self.done.listen()
+    }
```
sokra, line 114: Should this be a combined method that returns `Option<EventListener>`?

turbopack/crates/turbo-tasks/src/local_task_tracker.rs
sokra: Actually you only need `task_type` as argument and you can construct the done_event in this method too

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -1730,13 +1691,31 @@ impl<B: Backend + 'static> TurboTasksBackendApi<B> for TurboTasks<B> {
...
+        }
+    });
+    let Some(listener) = listener else {
+        return;
+    };
+    // Re-check after registering the listener.
+    let still_in_flight = CURRENT_TASK_STATE.with(|ts| ts.read().unwrap().local_tasks.in_flight());
+    if still_in_flight == 0 {
+        return;
     }
```
sokra: That is not needed. The in_flight counter is within a lock, so you don't need that. It would only be needed when in_flight would be an atomic counter

sokra (approved):

turbopack/crates/turbo-tasks/src/local_task_tracker.rs
lukesandberg: to make the tests simpler it now accepts `impl Display` but yeah, this is better

turbopack/crates/turbo-tasks/src/local_task_tracker.rs
lukesandberg: Done

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -1629,17 +1582,21 @@ impl<B: Backend + 'static> TurboTasksApi for TurboTasks<B> {
...
+        global_task_state
+            .write()
+            .unwrap()
+            .local_tasks
+            .register_detached();
+        let wrapped = async move {
+            fut.await;
+            // Pair for `register_detached`. Tasks panic-aborting upstream means we don't need
+            // RAII guard semantics here; if `fut` panics, the process aborts before this dec.
+            CURRENT_TASK_STATE.with(|ts| ts.write().unwrap().local_tasks.decrement_in_flight());
```
bgw:
```suggestion
            struct DropGuard;
            impl Drop for DropGuard {
                fn drop(&mut self) {
                    CURRENT_TASK_STATE.with(|ts| ts.write().unwrap().local_tasks.decrement_in_flight());
                }
            }
            let _guard = DropGuard;
            fut.await;
```

bgw (approved):

PR #93483, opened 2026-05-05
feat(turbopack): `next internal prewarm-dev` command to seed the dev cache

sokra:
### What?

Adds a new internal CLI command, `next internal prewarm-dev [directory]`, that compiles every entrypoint in a project against the Turbopack dev bundler and persists the result to the on-disk cache. After running it, a subsequent `next dev` cold start finds its persistent cache already populated and skips most of the upfront compilation work.

```bash
next internal prewarm-dev          # current directory
next internal prewarm-dev ./apps/web
```

The command is Turbopack-only and requires `experimental.turbopackFileSystemCacheForDev: true` — both are validated up front, so misconfigured projects fail fast with an actionable message instead of silently wasting time.

### Why?

The Turbopack persistent dev cache (`turbopackFileSystemCacheForDev`) makes the second cold start of `next dev` fast, but the first one still pays the full compile cost. For environments where dev is started repeatedly from a clean state (CI runners, ephemeral cloud dev environments, freshly cloned worktrees, demo/preview boxes), it's useful to do that compilation as a separate step ahead of time so the user-facing `next dev` always starts warm.

#### CLI / process model

* `packages/next/src/bin/next.ts` registers the command under the existing `next internal` group.
* `packages/next/src/cli/internal/prewarm-dev.ts` is the parent entry point. It validates the project directory, then forks a child worker that reuses the same `start-server.ts` bootstrap path that `next dev` uses, with `__NEXT_PRIVATE_PREWARM_DEV=1` and `TURBOPACK=1` set in the env. SIGINT/SIGTERM are forwarded to the child; the child's stdio is inherited so all output appears directly on the user's terminal.
* `packages/next/src/server/lib/start-server.ts` detects `__NEXT_PRIVATE_PREWARM_DEV` in its existing IPC `nextWorkerOptions` handler and dispatches to `prewarmDevServer(...)` instead of starting an HTTP server. Errors are caught, logged, and converted into a non-zero exit so the parent surfaces them.

#### Worker (`packages/next/src/server/lib/prewarm-dev-server.ts`)

* Validates that the bundler is Turbopack and the persistent cache flag is on (otherwise throws).
* Boots the dev bundler via the new shared `bootstrapDevBundler` helper (see refactor below).
* Calls `hotReloader.getEntrypoints()`, a new helper on the Turbopack hot reloader that awaits the in-flight entrypoints subscription event and returns the resulting snapshot.
* Builds a list of "prewarm units":
  * App-router and pages-router pages (compiled via `hotReloader.ensurePage`).
  * Pages-router globals `_app`/`_document`/`_error` (compiled via `endpoint.writeToDisk()`), but **only** when the project has user-defined pages routes — app-only projects skip them to avoid compiling unused defaults.
  * Middleware and instrumentation are eagerly compiled by `handleEntrypoints` during `getEntrypoints()`, so they aren't re-enqueued; the worker just counts them so the final summary stays accurate.
* Splits the unit list into growing batches (10, 20, 40, …) up front and processes each batch with the new `runWithConcurrency` helper. Concurrency starts at 1 and grows by 1 per completed unit up to `os.cpus().length`. After each batch finishes (no compilations in flight), `flushPersistentCache(hotReloader)` is called — flushes never overlap with active compilation.
* The final `closeProject(hotReloader)` is wrapped in a try/finally-like flow so it runs even when the batch loop aborts on a worker error, mirroring `turbopack-build/impl.ts`.
* SIGINT/SIGTERM in the worker exit immediately with `128 + signal` (the standard Unix convention) — Ctrl+C should feel like Ctrl+C. The cache is left in whatever state the most recent batch flush reached.

#### `runWithConcurrency` (`packages/next/src/lib/run-with-concurrency.ts`)

A small generic helper extracted into `lib/` since it's project-independent. Drains a list through a worker function with an adaptive concurrency cap (re-read on every scheduling pass). Aborts the whole drain — waiting for in-flight units to settle, then rethrowing — when the worker rejects or an external `AbortSignal` fires.

#### `bootstrapDevBundler` (`packages/next/src/server/lib/router-utils/setup-dev-bundler.ts`)

The dev-bundler bootstrap (telemetry sink, pages-dir lookup, traced `setupDevBundler` invocation) was duplicated between `router-server.ts` and the new prewarm worker. It's now a single helper consumed by both. As part of this:

* `SetupOpts` gains an optional `cliCommand` field. Telemetry sessions and the dev lockfile owner string both pick it up. `next dev` continues to default to `'dev'`; the prewarm worker passes `'prewarm-dev'`. As a result, running `next dev` while a prewarm is in progress (and vice versa) gets a clear "Another `next prewarm-dev` server is already running" message via the existing lockfile path.

#### Hot-reloader API change

`hot-reloader-turbopack.ts` now exports `TurbopackHotReloader` (a superset of `NextJsHotReloaderInterface`) which adds `getEntrypoints(): Promise<Entrypoints>`. Returning the snapshot of the FIRST entrypoints subscription event is good enough for prewarm — anything discovered later would need to be recompiled on `next dev` anyway.

### What this does **not** do

* It is not exposed as a top-level command (`next prewarm-dev`); it lives under `next internal` because it's intended for tooling/orchestration rather than a typical user workflow. It is not covered by semver.
* It does not compile a guaranteed-fresh snapshot — `getEntrypoints()` only resolves on the first subscription event. Any file added during the prewarm itself will get recompiled by `next dev` on first request, which is the same behaviour you'd get without the command.
* No documentation page is added — the command is internal and discoverable via `next internal --help`.

### Telemetry

The prewarm worker records a regular `eventCliSession` with `cliCommand: 'prewarm-dev'`, threaded through the new `cliCommand` option on `SetupOpts`. No new event types are added.

### Test plan

* `node packages/next/dist/bin/next internal --help` lists the new command.
* `node packages/next/dist/bin/next internal prewarm-dev --help` prints the expected usage.
* `pnpm --filter=next types` and `pnpm --filter=next build` are clean.


packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -1858,5 +1872,19 @@ export async function createHotReloaderTurbopack(
...
+      async getEntrypoints(): Promise<Entrypoints> {
+        // Wait for the in-flight subscription event to finish processing so
+        // the snapshot is consistent.  This resolves on the FIRST event
+        // (after the initial scan); entrypoints discovered later will not
+        // be reflected.
+        await currentEntriesHandling
+        return currentEntrypoints
+      },
+    }
+  )
```
mischnic: Why do you need object.assign here? Just put it next to the other methods a few lines above where `hotReload` is declared

packages/next/src/lib/run-with-concurrency.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  // Outer loop: keeps going until both the queue is empty *and* nothing is
+  // in flight.  Each iteration either tops up the active set (inner loop
+  // below) or waits for at least one in-flight unit to settle.
+  while (queue.length > 0 || active.size > 0) {
+    // Stop scheduling on abort or worker error.  Drain the in-flight set
+    // so the cache (and any other side effects) end up in a consistent
+    // state, then rethrow.
+    if (signal?.aborted || workerError !== undefined) {
+      while (active.size > 0) await Promise.race(active)
```
lukesandberg: should this be Promise.allSettled to avoid the loop?

sokra: fixed

packages/next/src/server/dev/hot-reloader-turbopack.ts
sokra: fixed

packages/next/src/server/lib/prewarm-dev-server.ts
```diff
@@ -0,0 +1,388 @@
...
+      await compileUnit(hotReloader, unit)
+    } catch {
+      // Per-unit errors are intentionally not logged: they would clutter
+      // output and the count is reported in the final summary.
+      failed++
+    }
+    completed++
+    // Grow the concurrency cap by 1 after each completed unit so the first
+    // few compiles run sequentially (cheap when the cache is warm) but we
+    // ramp up quickly when there's real work to do.
```
lukesandberg: i don't understand this justification

why not always use maxiumum concurrency?

packages/next/src/server/lib/prewarm-dev-server.ts
```diff
@@ -0,0 +1,388 @@
...
+  // captures the mutable `concurrency` variable.
+  const getConcurrency = () => concurrency
+
+  // Process units in growing batches.  After each batch finishes (no
+  // compilations in flight), flush the persistent cache to disk.  The very
+  // last flush is performed by the `closeProject` call below, so we skip
+  // the intermediate flush after the final batch.
+  let runError: unknown
+  try {
+    for (const batch of batchUnits(units)) {
```
lukesandberg: i wonder if instead of size heuristics we should use time heuristics where we flush after every n-seconds of work

it is common i think for some routes to be 100x as expensive as others, so a time based heuristic would be more consistent

lukesandberg: not sure though

packages/next/src/server/lib/prewarm-dev-server.ts
```diff
@@ -0,0 +1,388 @@
...
+  const fsChecker = await setupFsCheck({
+    dev: true,
+    dir,
+    config,
+    minimalMode: false,
+  })
+
+  // Force Turbopack for prewarm.  `bootstrapDevBundler` reads `TURBOPACK`
+  // from the env, which the CLI already sets, but we're defensive here.
+  process.env.TURBOPACK = '1'
```
lukesandberg: just assert that this is the case?

packages/next/src/server/lib/prewarm-dev-server.ts
```diff
@@ -0,0 +1,388 @@
...
+  }
+}
+
+/**
+ * Persist the Turbopack cache to disk while keeping the project alive.
+ *
+ * Internally this triggers Turbopack's `project.onExit()` API, which runs
+ * the registered exit handlers — those write the in-memory persistent cache
+ * to disk.  Despite the underlying name, the project remains usable and
+ * this can be called multiple times during a single run.
```
lukesandberg: can we rename it? this just looks wrong

packages/next/src/server/lib/prewarm-dev-server.ts
sokra: Each page already has concurrency due to turbo tasks, so even when running sequentially it would use multiple cpus. Especially the first page will compile a lot, so I didn't want to overwhelm it which too much work and too much memory usage. Once the first page is compiled we would end up with more cached tasks, so tasks can complete quickly instead of all concurrently waiting on some shared tasks to finish. More a gut feeling...

packages/next/src/server/lib/prewarm-dev-server.ts
sokra: Yeah I think that probably make sense.

The idea was also to run eviction after every batch, so maybe we could also go for a memory pressure heuristic here...

That might be something for a follow-up PR.

packages/next/src/server/lib/prewarm-dev-server.ts
sokra: yep we should

packages/next/src/server/lib/prewarm-dev-server.ts
lukesandberg: yeah this is tough.... from benchmarking vercel-site routes most are tiny but there are a few monsters so slow ramp doesn't help really

but maybe this also means that javascript level concurrency doesn't help too much unless we think that turbopack leave concurrency on the table at certain parts of the build (e.g. long tail tasks)

so maybe a simple policy like 4 routes at a time, snapshot after every 10 routes is simple and works?

its not really a big deal and i am probably overly oriented on pathological vercel-site patterns

packages/next/src/server/lib/prewarm-dev-server.ts
lukesandberg: yeah evict after every batch makes sense

lukesandberg (changes requested):
can you add an integration test that runs create-next-app and this?

asserting on what is cached is difficult but we should at least excersize the command

maybe this could be done on the filesystem-cache.test.ts project

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -332,6 +349,22 @@ impl<B: BackingStorage> TurboTasksBackendInner<B> {
...
+    /// that may not be persisting should check `should_persist()` first.
+    pub fn snapshot_and_persist_now(
+        &self,
+        turbo_tasks: &dyn TurboTasksBackendApi<TurboTasksBackend<B>>,
+    ) -> Result<(), anyhow::Error> {
+        assert!(
+            self.should_persist(),
+            "snapshot_and_persist_now requires persistence"
+        );
+        self.snapshot_and_persist(None, "explicit", turbo_tasks)
```
lukesandberg: i think you need to call snapshot_and_evict to ensure eviction runs

PR #88790, opened 2026-01-20
TURBOPACK: support chunk_loading_global method

fireairforce:
Support https://webpack.js.org/configuration/output/#outputchunkloadingglobal for turbopack Runtime code.

In utoo case, we need to support the iframe apps or micro frontend apps. If the master apps and slave apps both use turbopack to build, it will cause the variable `globalThis.TURBOPACK` conflict.

More details refer to: https://github.com/utooland/utoo/issues/2526

PR #93488, opened 2026-05-05
feat(turbopack): add chunkLoadingGlobal config option

tim123abc:
### What?

Adds a new `chunkLoadingGlobal` option to the `turbopack` config in `next.config.js` that overrides the global variable name used for Turbopack chunk loading (defaults to `TURBOPACK`).

```js
// next.config.js
module.exports = {
  turbopack: {
    chunkLoadingGlobal: 'myApp',
  },
}
```

### Why?

When multiple Turbopack-built apps run on the same page (e.g. horizontal micro-frontends), they each try to use `globalThis.TURBOPACK` which causes conflicts. This option allows each app to use its own isolated global, mirroring webpack's `output.chunkLoadingGlobal`.

The `chunk_loading_global` method was already wired into Turbopack in #88790 — this PR exposes it via Next.js config.

A reason is [vercel/remote-components#239](https://github.com/vercel/remote-components/pull/239), which uses this option to eliminate four brittle `globalThis.TURBOPACK` regex rename transforms in the remote-components runtime. Those transforms target Turbopack's internal variable naming and have already required updates as Turbopack's output format evolved. With `chunkLoadingGlobal` set at build time, Turbopack handles the renaming natively and those transforms become unnecessary.

### How?

- `config-shared.ts` / `config-schema.ts` — adds `chunkLoadingGlobal?: string` to `TurbopackOptions` with Zod validation
- `next_config.rs` — adds the field to `TurbopackConfig` and a `turbopack_chunk_loading_global()` accessor
- `next_client/context.rs` — threads it through `ClientChunkingContextOptions` and calls `builder.chunk_loading_global()` when set
- `project.rs` — wires the config into the client chunking context options

Only applies to the browser/client chunking context — server and edge runtimes don't have this concept.

### Test plan

- `pnpm test-start-turbo test/production/app-dir/turbopack-chunk-loading-global/turbopack-chunk-loading-global.test.ts`


test/production/app-dir/turbopack-chunk-loading-global/turbopack-chunk-loading-global.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      // Custom global must be registered as a chunk-loading global
+      expect(allContent).toContain('globalThis.myApp')
+      // And the derived chunk list global must use the same name
+      expect(allContent).toContain('globalThis.myApp_CHUNK_LISTS')
+
+      // Default TURBOPACK chunk-loading globals must not appear — if they
+      // do, chunkLoadingGlobal was ignored by the Rust layer.
+      expect(allContent).not.toContain('globalThis.TURBOPACK')
+      expect(allContent).not.toContain('globalThis.TURBOPACK_CHUNK_LISTS')
```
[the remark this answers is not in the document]
tim123abc: :robot: Fixed in the current code — the test now asserts `globalThis.TURBOPACK_myApp` (dot notation, matching what Turbopack emits). The config layer also auto-prefixes user-supplied values with `TURBOPACK_`, so `chunkLoadingGlobal: 'myApp'` becomes `TURBOPACK_myApp` before reaching the Rust layer.

packages/next/src/server/config-shared.ts
```diff
@@ -240,6 +240,16 @@ export interface TurbopackOptions {
     title?: string | RegExp
     description?: string | RegExp
   }>
+
+  /**
+   * (`next --turbopack` only) Override the global variable name used for
```
mischnic: It's inside of the Turbopack settings already
```suggestion
   * Override the global variable name used for
```

mischnic (approved):

packages/next/src/server/config.ts
```diff
@@ -1012,6 +1012,13 @@ function assignDefaultsAndValidate(
     }
   }
 
+  if (result?.turbopack?.chunkLoadingGlobal) {
+    const g = result.turbopack.chunkLoadingGlobal
+    if (!g.startsWith('TURBOPACK_')) {
+      result.turbopack.chunkLoadingGlobal = `TURBOPACK_${g}`
+    }
```
mischnic, line 1019: I am debating if we really want this or if this should just be the user's responsibility.
Just like Webpack just uses the value you pass to it

PR #93499, opened 2026-05-05
Refactor: Decouple request store creation from `req` / `res`

unstubbable:
This PR refactors `createRequestStoreImpl` into a new exported `createRequestStore` function that takes a serializable `RequestStoreInputs` shape (`headers`, optional `onUpdateCookies`, `url`, etc.) instead of a concrete `req` / `res` pair. The existing `createRequestStoreForRender` and `createRequestStoreForAPI` wrappers continue to take `(req, res, ...)` and translate to the new shape internally — there is no behavioral change for any existing code path.

Two transitive refactors are included in this change, motivated by the same goal of removing the dependency on `IncomingMessage` / `BaseNextRequest` / `NextRequest` from the inner construction:

- `checkIsOnDemandRevalidate(req, previewProps)` becomes `checkIsOnDemandRevalidate(rawHeaders, previewProps)`. Every call site now passes `req.headers` (or equivalent) rather than the whole request. The function only ever read `req.headers`, so this is purely a signature cleanup.
- `DraftModeProvider`'s constructor takes `headers` instead of `req`. Its draft-mode lookup also only needed the headers to call `checkIsOnDemandRevalidate`, so the dependency on the request types is no longer load-bearing.

`mergeMiddlewareCookies` similarly takes `headers` instead of `req`, but keeps its original `'x-middleware-set-cookie' in headers` + bracket-access pattern — preserving the pre-refactor behavior exactly (silent no-op for Web `Headers`, full merge for Node `IncomingHttpHeaders`). A `TODO` comment flags the open question of whether `Headers` callers need this merge or already handle it elsewhere. That's an investigation for a separate change.

This change is groundwork for an upcoming `'use cache'` deadlock-detection probe (see the `RequestStoreInputs` doc comment). The probe runs the same cache fill in an isolated worker process when the main fill stalls, and needs to construct a real `RequestStore` from a serialized snapshot of the outer request's headers / cookies / etc. — without access to the original `req` / `res` objects, which can't cross the worker boundary. With this refactor, the probe worker can call `createRequestStore` directly with the forwarded inputs and get a fully-functional store (real `headers`, `cookies`, `draftMode`, `mutableCookies`, `userspaceMutableCookies`), rather than building a stub that throws on access.

### DIFF
--- a/packages/next/src/build/templates/edge-ssr-app.ts
+++ b/packages/next/src/build/templates/edge-ssr-app.ts
@@ -96,7 +96,7 @@ async function requestHandler(
   const isPossibleServerAction = getIsPossibleServerAction(req)
   const botType = getBotType(req.headers.get('User-Agent') || '')
   const { isOnDemandRevalidate } = checkIsOnDemandRevalidate(
-    req,
+    req.headers,
     prerenderManifest.preview
   )
 
--- a/packages/next/src/server/api-utils/index.ts
+++ b/packages/next/src/server/api-utils/index.ts
@@ -1,5 +1,4 @@
 import type { IncomingHttpHeaders, IncomingMessage } from 'http'
-import type { BaseNextRequest } from '../base-http'
 import type { CookieSerializeOptions } from 'next/dist/compiled/cookie'
 import type { NextApiResponse } from '../../shared/lib/utils'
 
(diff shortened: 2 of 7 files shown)


unstubbable:
* **#93500** <a href="https://app.graphite.com/github/pr/vercel/next.js/93500?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93499** <a href="https://app.graphite.com/github/pr/vercel/next.js/93499?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93499?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

gnoff (approved):

unstubbable:
### Merge activity

* **May 6, 6:31 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93499).
* **May 6, 6:32 AM UTC**: @unstubbable merged this pull request with [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93499).

PR #93500, opened 2026-05-05
Detect `'use cache'` module-scope deadlocks early in dev

unstubbable:
When a `'use cache'` fill stalls in dev today, the user has to wait the full `useCacheTimeout` — 54 seconds by default — before any error surfaces, and the resulting `UseCacheTimeoutError` is too generic to point at the cause. A common cause is module-scoped state that ends up joining a promise from the outer render scope — for example a top-level `Map<string, Promise>` used to dedupe fetches, where the cache body and the outer scope both await the same promise. That promise hangs because Next.js intentionally converts uncached fetches into hanging promises during prerendering (and in dev while filling caches in the static or runtime stage), so the cache function ends up waiting forever on an outer-scope fetch that will never resolve. Most users reload long before the timeout appears, so they never see any signal that something specific is wrong.

This change adds a dev-only probe that surfaces this class of deadlock earlier. Once a cache fill has been idle for ten seconds, the dev server re-runs the same cache function in a worker thread with a fresh module scope. If it completes there, the hang in the main process is attributable to outer-scope state, and we abort the fill with `UseCacheDeadlockError` whose message points the user at the dedupe pattern and how to fix it. If the probe also hangs or fails for any other reason — decode failure, missing module, the body throwing — we fall back to the regular cache-fill timeout, so the probe is a positive signal only and never false-positives a deadlock.

The probe is gated on `__NEXT_DEV_SERVER` and tree-shakes out of the production runtime entirely. The worker pool is lazy, with no process forked until a probe actually fires, reused across probes for the same dev session, and torn down on HMR or worker crash. A snapshot of the outer request store is forwarded to the worker so that cache functions — including private caches that read `cookies()`, `headers()`, or `draftMode()` — see the same values they would in a real invocation.

### DIFF
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -59,7 +59,7 @@ import { DYNAMIC_EXPIRE, RUNTIME_PREFETCH_DYNAMIC_STALE } from './constants'
 import { NEXT_CACHE_ROOT_PARAM_TAG_ID } from '../../lib/constants'
 import type { CacheHandler } from '../lib/cache-handlers/types'
 import { getCacheHandler } from './handlers'
-import { UseCacheTimeoutError } from './use-cache-errors'
+import { UseCacheDeadlockError, UseCacheTimeoutError } from './use-cache-errors'
 import {
   createHangingInputAbortSignal,
   postponeWithTracking,
@@ -87,6 +87,10 @@ interface PrivateCacheContext {
     | PrerenderStoreModernRuntime
   readonly skipPropagation: boolean
   readonly outerOwnerStack: string | undefined
+  /** The `'use cache'` function's server reference id (second arg of `cache()`). */
+  readonly functionId: string
+  /** The cache handler kind (first arg of `cache()`, e.g. 'default'). */
+  readonly handlerKind: string
 }
 
 interface PublicCacheContext {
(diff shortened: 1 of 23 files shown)


unstubbable:
* **#93500** <a href="https://app.graphite.com/github/pr/vercel/next.js/93500?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93500?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#93499** <a href="https://app.graphite.com/github/pr/vercel/next.js/93499?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -1353,35 +1414,56 @@ export async function cache(
...
   // Private caches are currently only stored in the Resume Data Cache (RDC),
   // and not in cache handlers.
-  const cacheHandler = isPrivate ? undefined : getCacheHandler(kind)
+  const cacheHandler =
+    isPrivate || isProbeMode ? undefined : getCacheHandler(kind)
 
-  if (!isPrivate && !cacheHandler) {
+  if (!isPrivate && !isProbeMode && !cacheHandler) {
     throw new Error('Unknown cache handler: ' + kind)
   }
```
gnoff: think this would be cleaner with an if else block

gnoff (approved):

PR #92891, opened 2026-04-16
Turbopack: move import_usage computation into ImportMap

mischnic:
I'm moving all of the export-related logic into `EvalContext` (and thus `ImportMap`) so that `VarGraph` is only needed for computing the references, and not the exports.

PR #93504, opened 2026-05-05
Turbopack: fix import_usage for functions behind member expressions

mischnic:
Manifested as `ReferenceError: foo is not defined` where `foo` is an import

A regression from https://github.com/vercel/next.js/pull/92891


turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -815,6 +816,20 @@ mod analyzer_state {
...
+            self.state.is_in_static_member_expr_obj = true;
+        }
+
+        pub(super) fn take_is_in_static_member_expr_obj(&mut self) -> bool {
+            // Doesn't use visitor, because this propery is only valid for the current node, not the
+            // whole subtree.
+            let v = self.state.is_in_static_member_expr_obj;
+            self.state.is_in_static_member_expr_obj = false;
+            v
+        }
```
lukesandberg: can we use the closure pattern to manage the state transitions more safely

`self.enter_static_member_expr_obj(|this| {...})`

mischnic: the problem is that this isn't a transitive closure situation
but only 1 level deep, as I said in my comment

lukesandberg: i see, then maybe it would be clearer to just do the import handling inline since we know the receiver is an ident... we could just call `visit_ident` or extract a method from `visit_ident` to call here

mischnic: > call visit_ident

good idea. done

lukesandberg (approved):

PR #93516, opened 2026-05-05
Turbopack: make rcstr! macro expansion const

mischnic:
Brought to you by
```
destructor of `std::option::Option<turbo_rcstr::RcStr>` cannot be evaluated at compile-time
```
and
```
cannot perform conditionally-const deref coercion on `std::mem::ManuallyDrop<std::option::Option<RcStr>>` in constants
```


lukesandberg (approved):
woohoo!

sokra (approved):

PR #93520, opened 2026-05-06
bundle-analyzer: record historical snapshots on each analyze run

wbinnssmith:
After each `next experimental-analyze` run, snapshot the output data into a rolling `history/` directory so later builds can use it as a comparison baseline.

Each snapshot gets a `metadata.json` with git branch/sha/dirty status, Next.js version, and route count. A `history/history.json` index (newest-first, capped at 20 entries) lets the UI list available baselines without re-reading every snapshot directory. Old snapshots beyond the cap are pruned automatically.

The client-side `SnapshotMetadata` type mirror lives in `apps/bundle-analyzer/lib/snapshot.ts` so the compare UI can reference it without a build-time dependency on the Next.js package.

### DIFF
--- a/apps/bundle-analyzer/lib/snapshot.ts
+++ b/apps/bundle-analyzer/lib/snapshot.ts
@@ -1,0 +1,59 @@
+/**
+ * Mirrors {@link SnapshotMetadata} written by the build (see
+ * `packages/next/src/build/analyze/snapshot.ts`). Field semantics must stay in
+ * sync.
+ */
+export interface SnapshotMetadata {
+  id: string
+  createdAt: string
+  nextVersion?: string
+  gitBranch?: string
+  gitSha?: string
+  gitShortSha?: string
+  gitDirty?: boolean
+  appDirOnly?: boolean
+  noMangling?: boolean
+  routeCount: number
+}
+
+/** Mirrors {@link HistoryIndex}. */
+export interface HistoryIndex {
+  snapshots: SnapshotMetadata[]
+}
+
+/**
+ * Returns a short, human-friendly label for a snapshot. Used by the picker
+ * and the diff header bar. Format prefers branch + short sha, falling back to
+ * timestamp when neither is available.
+ */
+export function formatSnapshotLabel(metadata: SnapshotMetadata): string {
+  const sha = metadata.gitShortSha ? metadata.gitShortSha : null
+  const branch = metadata.gitBranch ?? null
+  if (branch && sha) return `${branch}@${sha}${metadata.gitDirty ? '*' : ''}`
+  if (sha) return `${sha}${metadata.gitDirty ? '*' : ''}`
+  if (branch) return branch
+  return formatRelativeTime(metadata.createdAt)
+}
+
+/**
+ * Returns a relative time string for an ISO timestamp ("3m ago", "2h ago",
+ * "yesterday", "Jan 4"). Designed for compact list rows.
+ */
+export function formatRelativeTime(iso: string): string {
+  const then = new Date(iso).getTime()
+  if (Number.isNaN(then)) return iso
+  const diffMs = Date.now() - then
+  const seconds = Math.round(diffMs / 1000)
+  if (seconds < 45) return 'just now'
+  const minutes = Math.round(seconds / 60)
+  if (minutes < 60) return `${minutes}m ago`
+  const hours = Math.round(minutes / 60)
+  if (hours < 24) return `${hours}h ago`
+  const days = Math.round(hours / 24)
+  if (days === 1) return 'yesterday'
+  if (days < 14) return `${days}d ago`
+  return new Date(iso).toLocaleDateString(undefined, {
+    month: 'short',
+    day: 'numeric',
+  })
+}
--- a/apps/bundle-analyzer/lib/snapshot.ts
+++ b/apps/bundle-analyzer/lib/snapshot.ts
@@ -11,8 +11,12 @@ export interface SnapshotMetadata {
   gitSha?: string
   gitShortSha?: string
   gitDirty?: boolean
+  /** First line of the HEAD commit message when available. */
+  gitMessage?: string
   appDirOnly?: boolean
   noMangling?: boolean
+  /** Optional label overriding branch/sha in display. See `--snapshot-label`. */
+  snapshotLabel?: string
   routeCount: number
 }
 
@@ -27,6 +31,7 @@ export interface HistoryIndex {
  * timestamp when neither is available.
  */
 export function formatSnapshotLabel(metadata: SnapshotMetadata): string {
+  if (metadata.snapshotLabel) return metadata.snapshotLabel
   const sha = metadata.gitShortSha ? metadata.gitShortSha : null
   const branch = metadata.gitBranch ?? null
   if (branch && sha) return `${branch}@${sha}${metadata.gitDirty ? '*' : ''}`
@@ -44,7 +49,7 @@ export function formatRelativeTime(iso: string): string {
   if (Number.isNaN(then)) return iso
   const diffMs = Date.now() - then
   const seconds = Math.round(diffMs / 1000)
-  if (seconds < 45) return 'just now'
+  if (seconds < 60) return 'just now'
   const minutes = Math.round(seconds / 60)
   if (minutes < 60) return `${minutes}m ago`
   const hours = Math.round(minutes / 60)
--- a/apps/bundle-analyzer/lib/snapshot.ts
+++ b/apps/bundle-analyzer/lib/snapshot.ts
@@ -15,8 +15,8 @@ export interface SnapshotMetadata {
   gitMessage?: string
   appDirOnly?: boolean
   noMangling?: boolean
-  /** Optional label overriding branch/sha in display. See `--snapshot-label`. */
-  snapshotLabel?: string
+  /** User-supplied baseline name, overriding branch/sha in display. See `--baseline-name`. */
+  baselineName?: string
   routeCount: number
 }
 
@@ -31,7 +31,7 @@ export interface HistoryIndex {
  * timestamp when neither is available.
  */
 export function formatSnapshotLabel(metadata: SnapshotMetadata): string {
-  if (metadata.snapshotLabel) return metadata.snapshotLabel
+  if (metadata.baselineName) return metadata.baselineName
   const sha = metadata.gitShortSha ? metadata.gitShortSha : null
   const branch = metadata.gitBranch ?? null
   if (branch && sha) return `${branch}@${sha}${metadata.gitDirty ? '*' : ''}`
--- a/apps/bundle-analyzer/lib/snapshot.ts
+++ b/apps/bundle-analyzer/lib/snapshot.ts
@@ -34,8 +34,9 @@ export function formatSnapshotLabel(metadata: SnapshotMetadata): string {
   if (metadata.baselineName) return metadata.baselineName
   const sha = metadata.gitShortSha ? metadata.gitShortSha : null
   const branch = metadata.gitBranch ?? null
-  if (branch && sha) return `${branch}@${sha}${metadata.gitDirty ? '*' : ''}`
-  if (sha) return `${sha}${metadata.gitDirty ? '*' : ''}`
+  if (branch && sha)
+    return `${branch}@${sha}${metadata.gitDirty ? ' (dirty)' : ''}`
+  if (sha) return `${sha}${metadata.gitDirty ? ' (dirty)' : ''}`
   if (branch) return branch
   return formatRelativeTime(metadata.createdAt)
 }
--- a/packages/next/src/build/analyze/snapshot.ts
+++ b/packages/next/src/build/analyze/snapshot.ts
@@ -1,0 +1,207 @@
+import * as path from 'node:path'
+import { cp, mkdir, readFile, readdir, rm, writeFile } from 'node:fs/promises'
+
+import { getGitBranch, getGitCommit, getGitDirty } from '../../lib/helpers/git'
+
+/**
+ * Maximum number of historical snapshots to keep on disk by default. When the
+ * history exceeds this number, the oldest snapshots are pruned.
+ */
+const MAX_HISTORY = 20
+
+/**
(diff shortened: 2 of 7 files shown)


wbinnssmith:
* **#93523** <a href="https://app.graphite.com/github/pr/vercel/next.js/93523?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93522** <a href="https://app.graphite.com/github/pr/vercel/next.js/93522?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93521** <a href="https://app.graphite.com/github/pr/vercel/next.js/93521?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93520** <a href="https://app.graphite.com/github/pr/vercel/next.js/93520?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93520?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

apps/bundle-analyzer/lib/snapshot.ts
```diff
@@ -0,0 +1,59 @@
...
+ * "yesterday", "Jan 4"). Designed for compact list rows.
+ */
+export function formatRelativeTime(iso: string): string {
+  const then = new Date(iso).getTime()
+  if (Number.isNaN(then)) return iso
+  const diffMs = Date.now() - then
+  const seconds = Math.round(diffMs / 1000)
+  if (seconds < 45) return 'just now'
+  const minutes = Math.round(seconds / 60)
+  if (minutes < 60) return `${minutes}m ago`
```
lukesandberg, line 55: if it is 55s ago, this will say 0m ago

i think the 'just now' cutoff needs to be at `seconds<60`

apps/bundle-analyzer/lib/snapshot.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Mirrors {@link SnapshotMetadata} written by the build (see
+ * `packages/next/src/build/analyze/snapshot.ts`). Field semantics must stay in
+ * sync.
+ */
+export interface SnapshotMetadata {
+  id: string
+  createdAt: string
+  nextVersion?: string
+  gitBranch?: string
+  gitSha?: string
```
lukesandberg, line 11: should we capture the first line of the commit message as well?

packages/next/src/build/analyze/snapshot.ts
```diff
@@ -0,0 +1,207 @@
...
+}
+
+const DATA_DIRNAME = 'data'
+const HISTORY_DIRNAME = 'history'
+const METADATA_FILENAME = 'metadata.json'
+const HISTORY_INDEX_FILENAME = 'history.json'
+
+interface BuildSnapshotInputs {
+  /** Project root, used to resolve git metadata. */
+  projectDir: string
```
lukesandberg, line 65: nit: a bit ambiguous, would this be front/apps/vercel-site or front/?

packages/next/src/build/analyze/snapshot.ts
```diff
@@ -0,0 +1,207 @@
...
+  /** Absolute path of the analyzer output directory (`.next/diagnostics/analyze`). */
```
lukesandberg: doesn't next nuke this directory on startup?

https://github.com/vercel/next.js/blob/f172ba9c41afb268804c84a88d45482608698f21/packages/next/src/build/index.ts#L1121

how does this data survive?

lukesandberg: maybe `.next-profiles` is a better location?

packages/next/src/build/analyze/snapshot.ts
```diff
@@ -0,0 +1,207 @@
...
+
+  // Newest first.
+  snapshots.sort((a, b) => (a.createdAt < b.createdAt ? 1 : -1))
+
+  // Prune: anything past the cap is removed from disk.
+  const kept = snapshots.slice(0, maxHistory)
+  const pruned = snapshots.slice(maxHistory)
+  await Promise.all(
+    pruned.map((snapshot) =>
+      rm(path.join(historyDir, snapshot.id), {
```
lukesandberg, line 206: use recursiveDeleteSyncWithAsyncRetries?  solves some windows problems i think

lukesandberg (dismissed):
it would be useful to support an optional `--profile-name` parameter that overrides the git branch information

i imagine a usecase where i fork a new branch collect a baseline  `--profile-name=baseline` and then iterate on some changes... but maybe i don't commit so timestampes are all the would distinguish things

bgw:
Criterion calls @lukesandberg's idea "baselines": https://bheisler.github.io/criterion.rs/book/user_guide/command_line_options.html#baselines

I think I like that as it's less ambiguous than "profiles".

packages/next/src/build/analyze/snapshot.ts
bgw: I really don't like adding more top-level `.next*` directories because we force users to add more things to their `.gitignore` every time we do this, and some percentage of users aren't going to notice and will accidentally commit a bunch of stuff.

I view this as a breaking change, but maybe I'm alone in this.

apps/bundle-analyzer/lib/snapshot.ts
wbinnssmith: yeah, I like that.

packages/next/src/build/analyze/snapshot.ts
wbinnssmith: We put the data itself in .next/diagnostics/analyzer (or similar), I think if this is clearly distinguished we can put them there as well

apps/bundle-analyzer/lib/snapshot.ts
wbinnssmith: :robot: Fixed — changed the cutoff from `seconds < 45` to `seconds < 60`, so 55 s correctly returns 'just now' rather than '0m ago'.

apps/bundle-analyzer/lib/snapshot.ts
wbinnssmith: :robot: Done — added `gitMessage?: string` to `SnapshotMetadata` (both packages/next and apps/bundle-analyzer sides) and a `getGitMessage` helper in `lib/helpers/git.ts` that returns `git log -1 --pretty=%s` (or `VERCEL_GIT_COMMIT_MESSAGE` on Vercel).

packages/next/src/build/analyze/snapshot.ts
wbinnssmith: :robot: Clarified the JSDoc: now says 'Absolute path to the Next.js project root (the directory containing `next.config.*`)' to remove ambiguity about monorepo nesting.

packages/next/src/build/analyze/snapshot.ts
wbinnssmith: :robot: Agreed that history must survive builds. Fixed by adding `diagnostics` to the exclusion regex in the `cleanDistDir` step in `build/index.ts` (`/^(cache|dev|diagnostics|lock|trace)/`). That way `.next/diagnostics/analyze/history/` is preserved across builds without adding a new top-level directory.

packages/next/src/build/analyze/snapshot.ts
wbinnssmith: :robot: Switched both the snapshot-replace path and the pruning loop to use `recursiveDeleteSyncWithAsyncRetries` + `fs.rmdirSync` so the retry-on-EBUSY/EPERM logic handles Windows-locked files.

apps/bundle-analyzer/lib/snapshot.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Mirrors {@link SnapshotMetadata} written by the build (see
+ * `packages/next/src/build/analyze/snapshot.ts`). Field semantics must stay in
+ * sync.
+ */
```
bgw, line 5: It'd be cool if we could just share these types between the two typescript files, but maybe it's difficult because of the dependency hierarchy.

apps/bundle-analyzer/lib/snapshot.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Returns a short, human-friendly label for a snapshot. Used by the picker
+ * and the diff header bar. Format prefers branch + short sha, falling back to
+ * timestamp when neither is available.
+ */
+export function formatSnapshotLabel(metadata: SnapshotMetadata): string {
+  if (metadata.baselineName) return metadata.baselineName
+  const sha = metadata.gitShortSha ? metadata.gitShortSha : null
+  const branch = metadata.gitBranch ?? null
+  if (branch && sha) return `${branch}@${sha}${metadata.gitDirty ? '*' : ''}`
+  if (sha) return `${sha}${metadata.gitDirty ? '*' : ''}`
```
bgw: Nobody is going to figure out what the `*` character means. Just put `"(dirty)"`.

packages/next/src/build/analyze/snapshot.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  // 2. Snapshot the entire data dir into history/<id>/.
+  const snapshotDir = path.join(historyDir, id)
+  await mkdir(historyDir, { recursive: true })
+  // If the same id already exists (same second + same sha) replace it so the
+  // latest run wins. Use Windows-safe deletion with retry logic.
+  if (fs.existsSync(snapshotDir)) {
+    await recursiveDeleteSyncWithAsyncRetries(snapshotDir)
+    fs.rmdirSync(snapshotDir)
+  }
```
bgw: Instead of implementing the retry logic yourself, use `fs.rmSync` with `maxRetries` and `recursive`: https://nodejs.org/api/fs.html#fsrmsyncpath-options

packages/next/src/build/analyze/snapshot.ts
```diff
@@ -0,0 +1,230 @@
...
+  const snapshots: SnapshotMetadata[] = []
+  for (const entry of entries) {
+    if (entry === HISTORY_INDEX_FILENAME) continue
+    const metadataPath = path.join(historyDir, entry, METADATA_FILENAME)
+    try {
+      const text = await readFile(metadataPath, 'utf8')
+      const parsed = JSON.parse(text) as SnapshotMetadata
+      // Trust the on-disk metadata's id if present, otherwise fall back to
+      // the directory name (backwards compat / hand-edited cases).
+      snapshots.push({ ...parsed, id: parsed.id ?? entry })
```
bgw: Do we actually care about this backwards compat?

packages/next/src/build/analyze/snapshot.ts
```diff
@@ -0,0 +1,230 @@
...
+  const pruned = snapshots.slice(maxHistory)
+  await Promise.all(
+    pruned.map(async (snapshot) => {
+      const snapshotDir = path.join(historyDir, snapshot.id)
+      await recursiveDeleteSyncWithAsyncRetries(snapshotDir)
+      try {
+        fs.rmdirSync(snapshotDir)
+      } catch {
+        // Already gone or non-empty due to a race — not fatal.
+      }
```
bgw: use `fs.rm` with recursive and maxRetries

packages/next/src/build/analyze/snapshot.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  const snapshots: SnapshotMetadata[] = []
+  for (const entry of entries) {
+    if (entry === HISTORY_INDEX_FILENAME) continue
+    const metadataPath = path.join(historyDir, entry, METADATA_FILENAME)
+    try {
+      const text = await readFile(metadataPath, 'utf8')
+      const parsed = JSON.parse(text) as SnapshotMetadata
+      // Trust the on-disk metadata's id if present, otherwise fall back to
```
wbinnssmith: there's no "backwards" versions in backwards compat here, only in the work iteration. no need. just push the `parsed` into the snapshots index

packages/next/src/build/index.ts
```diff
@@ -4279,27 +4279,34 @@ export default async function build(
       await shutdownPromise
 
       if (NextBuildContext.analyze) {
-        await cp(
-          path.join(__dirname, '../bundle-analyzer'),
-          path.join(dir, '.next/diagnostics/analyze'),
-          { recursive: true }
-        )
+        const analyzeDir = path.join(dir, '.next/diagnostics/analyze')
```
bgw: This should use `distDir` instead of hardcoding `.next`

packages/next/src/build/index.ts
```diff
@@ -4279,27 +4279,34 @@ export default async function build(
...
-            null,
-            2
-          )
+          path.join(analyzeDir, 'data/routes.json'),
+          JSON.stringify(routes, null, 2)
         )
+
+        // Capture this build alongside any prior builds so the analyzer UI
+        // can offer it as a comparison baseline in the future.
+        const { writeAnalyzeSnapshot } = await import('./analyze/snapshot')
```
bgw: uhhhhhhhhhhhhh

packages/next/src/lib/helpers/git.ts
bgw: We should really not be using a shell for this. Should use https://nodejs.org/api/child_process.html#child_processspawnsynccommand-args-options

Otherwise it's an extra process spawn for the shell and introducing a bunch of possibility for string escaping bugs.

packages/next/src/lib/helpers/git.ts
```diff
@@ -44,3 +44,32 @@ export function getGitCommit(cwd: string): string | undefined {
...
 }
+
+/**
+ * Returns true if the working tree has uncommitted changes. Returns undefined
+ * when the dirty status cannot be determined (not a git repo, git not
+ * installed, etc.).
+ */
+export function getGitDirty(cwd: string): boolean | undefined {
+  try {
+    const output = gitExec('status --porcelain', cwd)
```
wbinnssmith:
```suggestion
    return gitExec('status --porcelain', cwd).length > 0
```

bgw (approved):

PR #74482, opened 2025-01-03
perf(rcstr): Support inline string

kdy1:
### What?

Support inline string for `RcStr`. Most of the code is copied from https://github.com/dudykr/ddbase/tree/67c7cec871073f5c8c93d030d40fb8226170797f/crates/hstr/src (I own the code)

### Why?

To see at the perf difference.

#### Command: `ddt profile instruments run -t Memory -- $(which node) $(which pnpm) build-turbopack`

 - Old:
 
image

 - New:

image

### How?

Closes PACK-3692

PR #93524, opened 2026-05-06
[turbopack] correct a fencepost error in our inline string descision

lukesandberg:
## What?

Fix an off-by-one in the `RcStr` inline-length check. Strings of length exactly `MAX_INLINE_LEN` should be stored inline, not on the heap.

## Why?

`<` should have been `<=`. The buffer holds `MAX_INLINE_LEN` bytes and the 4-bit tag field can encode that length, so the strict comparison was wasting one byte of inline capacity for no reason — 7-byte strings were allocating on the default 64-bit build. Bug dates back to #74482, the original inline-string change.

This covers fun things like: "default" (×11), "project" (×8), "process" (×8), "postcss" (×7), "require" (×6), "browser" (×4)

### DIFF
--- a/turbopack/crates/turbo-rcstr/src/dynamic.rs
+++ b/turbopack/crates/turbo-rcstr/src/dynamic.rs
@@ -58,7 +58,7 @@ pub unsafe fn restore_arc(v: TaggedValue) -> Arc<PrehashedString> {
 pub(crate) fn new_atom<T: AsRef<str> + Into<String>>(text: T) -> RcStr {
     let len = text.as_ref().len();
 
-    if len < MAX_INLINE_LEN {
+    if len <= MAX_INLINE_LEN {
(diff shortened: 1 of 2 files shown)


lukesandberg:
* **#93524** <a href="https://app.graphite.com/github/pr/vercel/next.js/93524?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93524?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

bgw (approved):
@bgw approved this PR from Slack with [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93524)

sokra (approved):

PR #93526, opened 2026-05-06
Narrow per-segment `allowQuery` to reachable params

acdlite:
The build adapter previously gave every segment of a route the same page-level `allowQuery`. For routes with high-cardinality params, that forced segments whose responses don't depend on a param to still use it as part of their ISR cache key — producing many distinct cache entries of byte-identical output.

Each segment now gets an `allowQuery` narrowed to just the params it can actually reach. For the route tree prefetch, that set is always empty. This significantly reduces the number of ISR writes for affected routes.

### DIFF
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -37,6 +37,7 @@ import {
   CACHE_ONE_YEAR_SECONDS,
   HTML_CONTENT_TYPE_HEADER,
   JSON_CONTENT_TYPE_HEADER,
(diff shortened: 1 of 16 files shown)


acdlite:
I'm converting this back to a draft because I'm a bit skeptical about how fallback routes are supposed to work; it seems like those have an `allowQuery` that omits the fallback params... but then that suggests we should be able to have different `allowQuery`s for the same route across the lifetime of a deployment: one before a route is upgraded from fallback -> full, and then another one after that. In which case... we should be able to more generally change the allow query to track the exact set of params accessed during render, no?

PR #93529, opened 2026-05-06
Fix debug build paths Pages Router support entries

### LINKED ISSUE #93515: Turbopack debug-build-paths omits Pages Router _document in monorepo externalDir app

https://github.com/Stanzilla/next-debug-build-paths-pages-repro

### To Reproduce

```bash
git clone https://github.com/Stanzilla/next-debug-build-paths-pages-repro.git
cd next-debug-build-paths-pages-repro
npm install
npm run build
```

The `build` script runs:

```bash
cd apps/web && next build --debug-build-paths=pages/[gameId]/rank/matchmaking/index.jsx
```

This also reproduces with canary:

```bash
npm run build:canary
```

Current behavior:

The selective build compiles successfully, then fails while collecting page data:

```text
unhandledRejection Error [PageNotFoundError]: Cannot find module for page: /_document
    at ignore-listed frames {
  type: 'PageNotFoundError',
  code: 'ENOENT'
}
```

Expected behavior:

`next build --debug-build-paths=pages/[gameId]/rank/matchmaking/index.jsx` should build the selected Pages Router route and include the Pages Router support entries required to render/export it, such as `_app`, `_document`, `_error`, `404`, and `500`.

Reproduced with:

```text
Next.js 16.2.4
Next.js 16.3.0-canary.10
React 18.3.1
macOS
Turbopack build
```

The repro uses:

- a monorepo-shaped app in `apps/web`
- a Pages Router dynamic route selected with `--debug-build-paths`
- a custom `pages/_document.jsx`
- a custom `pages/_app.jsx`
- an App Router route in `app/`
- `i18n` config
- `experimental.externalDir: true`
- `outputFileTracingRoot` pointing at the repo root
- a Pages route importing code from `packages/rank-ui` outside the app directory

Turbopack, Pages Router, CLI (`next build`)

`next build`

### Additional context

A similar single-folder app with the same Pages route, `_document`, App Router route, and i18n config does not reproduce this. The failure appears when the app is monorepo-shaped and the selected Pages Router route imports from outside the app directory with `experimental.externalDir` enabled.

In a larger app we traced this to selective build path handling: route discovery can identify the selected Pages Router route, but the Turbopack build output can omit required Pages Router entries from `server/pages-manifest.json`, causing `PageNotFoundError` for `/_document` during page data collection.

Stanzilla:
## Summary
- Preserve Pages Router support entries during Turbopack `--debug-build-paths` selective builds
- Include custom Pages Router `404` and `500` entries alongside `_app`, `_document`, and `_error` when a Pages route is selected
- Add regression coverage for selected Pages routes retaining support entries in `pages-manifest.json`

## Testing
- `RUSTC="$(rustup which --toolchain nightly-2026-06-24 rustc)" cargo check -p next-api`
- `RUSTC="$(rustup which --toolchain nightly-2026-06-24 rustc)" pnpm --dir packages/next-swc build-native`
- `HEADLESS=true pnpm test-start-turbo test/production/debug-build-path/debug-build-paths.test.ts -t "single page with pages/ prefix"`
- `HEADLESS=true pnpm test-start-turbo test/production/debug-build-path/debug-build-paths.test.ts`

Fixes #93515


Stanzilla:
@sokra any chance this could get a look?

mischnic:
https://github.com/vercel/next.js/pull/94341 already fixed the index normalization
Can you rebase your branch?

Stanzilla:
Thanks, yeah that makes sense. I rebased on top of #94341 and dropped the overlapping nested `index` route coverage from this PR.

What’s left here is the separate support-entry bit: when `--debug-build-paths` selects a Pages Router route, we still need to keep the Pages support entries around (`/_app`, `/_document`, `/_error`, plus custom `/404` and `/500`). Otherwise Turbopack can compile the selected route, but later page-data/static generation can still miss those entries, which is the `Cannot find module for page: /_document` shape from the repro.

So the test now uses a plain `pages/foo.tsx` selected build and checks that `pages-manifest.json` still includes those support entries. That should keep this PR focused on the remaining gap after #94341.

packages/next/src/build/route-discovery.ts
```diff
@@ -439,7 +446,23 @@ export async function discoverRoutes(
   ): string[] => {
     if (debugPaths.length > 0) {
       const debugPathsSet = new Set(debugPaths)
-      return paths.filter((p) => debugPathsSet.has(p))
+      const filteredPaths = paths.filter((p) => debugPathsSet.has(p))
+      const hasPagesRoute = filteredPaths.some(
+        (p) => !isReservedPage(getPageFromPath(p, pageExtensions))
```
mischnic: What is this `isReservedPage` check for?
That checks for `/^\/(_app|_error|_document|api(\/|$))/`.

Stanzilla: :robot: Good point — isReservedPage() was broader than the intent here. The check only needs to tell whether the selected paths contain a renderable Pages route: for an API-only selection, or one containing only framework entries, we should not pull custom _app, _document, _error, 404, and 500 sources into the filtered build. I replaced it with an explicit isRenderablePagesRoute() predicate, aligned the Rust-side check for /api and /api/*, and added an API-only regression test that verifies the custom support pages are not built.

Stanzilla:
@mischnic Can I get a re-review please? :)

mischnic (approved):

PR #93535, opened 2026-05-06
fix(typescript): handle export default with arrow function

### LINKED ISSUE #55332: Next TS plugin warning 'TS71007: Props must be serializable for components in the "use client" entry file' does not work with export default
### Link to the code that reproduces this issue or a replay of the bug

https://codesandbox.io/p/sandbox/brave-northcutt-3h899t?file=%2Fapp%2Fpage.tsx%3A22%2C1

### To Reproduce

```typescript
type myProps = {
  myFunc: () => void;
};
const myComponent = ({ myFunc }: myProps) => {
  myFunc();
  return <></>;
};

export default myComponent;

export const myComponent2 = ({ myFunc }: myProps) => {
  myFunc();
  return <></>;
};

export function myComponent3({ myFunc }: myProps) {
  myFunc();
  return <></>;
}```

only `myComponent2` and `myComponent3` show the warning. 

Warning should also show for `myComponent`

### Verify canary release

- [X] I verified that the issue exists in the latest Next.js canary release

```bash
Operating System:
      Platform: darwin
      Arch: arm64
      Version: Darwin Kernel Version 23.0.0: Thu Aug 17 21:23:05 PDT 2023; root:xnu-10002.1.11~3/RELEASE_ARM64_T6000
    Binaries:
      Node: 20.3.1
      npm: 10.1.0
      Yarn: 1.22.19
      pnpm: 8.7.4
    Relevant Packages:
      next: 13.4.19
      eslint-config-next: 13.4.19
      react: 18.2.0
      react-dom: 18.2.0
      typescript: 5.2.2
    Next.js Config:
      output: standalone
```

App Router

### Additional context

Code sandbox is a little redundant but see the `components.tsx` file inside `app` folder for the same code as above. The plugin code for this warning is here: https://github.com/vercel/next.js/blob/canary/packages/next/src/server/typescript/rules/client-boundary.ts

Liuwei1125:
## Bug Description

The Next.js TypeScript plugin's `forEachChild` loop in `getSemanticDiagnostics` was missing a handler for **'export default someVar'** where `someVar` is an arrow function.

This pattern (`export default const MyComponent = ...`) produces an `ExportAssignment` AST node, which had no handler in the existing if-else chain.

## Impact

As a result, the **TS71007 warning** ("Props must be serializable for components in the 'use client' entry file") was **NOT shown** for default-exported arrow functions — even though it was correctly shown for:
- `export default function Foo() {}`
- `export function Foo() {}`
- `export const Foo = () => {}`

## Fix

Add an `ExportAssignment` branch that:
1. Uses the TypeChecker to resolve the identifier (`node.expression`) to its symbol
2. Gets the declaration of that symbol
3. Checks if it's an arrow function
4. Delegates to `clientBoundary.getSemanticDiagnosticsForFunctionExport` (same as other function export patterns)

## Testing

```tsx
// In a 'use client' file:
type myProps = { myFunc: () => void };
const myComponent = ({ myFunc }: myProps) => {  // TS71007 warning was MISSING here
  myFunc();
  return <></>;
};
export default myComponent;
```

Before fix: no warning.  
After fix: TS71007 warning appears as expected.

## Related

Fixes #55332

---

**Root cause analysis**: @Zelys-DFKH identified the missing `ExportAssignment` AST node handler in the issue comments.

### DIFF
--- a/packages/next/src/server/typescript/index.ts
+++ b/packages/next/src/server/typescript/index.ts
@@ -349,7 +349,6 @@ export const createTSPlugin: tsModule.server.PluginModuleFactory = ({
               }
             }
           }
-        }
         } else if (ts.isExportDeclaration(node)) {
           // export { ... }
           if (isAppEntry) {
--- a/packages/next/src/server/typescript/index.ts
+++ b/packages/next/src/server/typescript/index.ts
@@ -15,6 +15,7 @@ import {
   isDefaultFunctionExport,
   isPositionInsideNode,
   getSource,
+  getTypeChecker,
   isInsideApp,
 } from './utils'
 import { NEXT_TS_ERRORS } from './constant'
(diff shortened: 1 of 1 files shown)


Zelys-DFKH:
@Liuwei1125 — the `ExportAssignment` path is the right approach, but `symbol.getDeclarations()` returns a `VariableDeclaration`, not an `ArrowFunction`. For `const Foo = () => {}`, the AST looks like this:

```
VariableDeclaration        ← what getDeclarations()[0] returns
  Identifier: Foo
  ArrowFunction            ← the initializer you want
```

So `ts.isArrowFunction(declarations[0])` is always `false`. The arrow function sits at `.initializer`:

```ts
if (
  declarations &&
  declarations[0] &&
  ts.isVariableDeclaration(declarations[0]) &&
  declarations[0].initializer &&
  ts.isArrowFunction(declarations[0].initializer)
) {
  const arrowFn = declarations[0].initializer as tsModule.ArrowFunction
  // rest unchanged
}
```

Same thing the Vercel bot flagged at line 328, just with the context of why. Happy to open a follow-up PR.

PR #93549, opened 2026-05-06
Track `searchParams` access through a Proxy to catch missing-key reads

unstubbable:
Under `experimental.varyParams`, when a page reads `searchParams` from a URL without a query string, `createVaryingSearchParams` previously returned a plain object whose only own properties were getters for keys that already existed in the URL. With no keys present, that's an empty `{}` — so `Object.entries`, spread, `in` checks, and missing-key reads (e.g. `searchParams.foo` returning `undefined`) all silently no-opped without registering an access. The segment ended up keyed at the `Fallback` search slot in the segment cache, where it shadowed subsequent prefetches with non-empty search params via `Fallback` resolution, silently serving the stale empty-query cache entry on navigation.

This switches the implementation to a `Proxy` with `get`, `has`, and `ownKeys` traps that record the `'?'` sentinel on every string access. Search params have no fixed schema, so any access — including missing-key reads, `in` checks, and enumeration — must register as varying. The `get` trap has one carve-out: `.then` reads are skipped when no `then` key exists on the target, since `Promise.resolve` and React Flight probe `.then` to test thenability and those probes shouldn't count as user dependencies. As a trade-off, a page that reads `searchParams.then` to compute its output and is prefetched without a `then` key won't register the access — accepted because `then` is reserved by the Promise protocol and essentially never used as a query parameter name.

The added regression test prefetches the no-query URL first and asserts that a subsequent `?foo=1` prefetch still triggers a fresh request. Without the fix the lookup resolves to the empty-query cache entry through `Fallback`, no request is initiated, and the test times out waiting for one.

### DIFF
--- a/packages/next/src/server/app-render/vary-params.ts
+++ b/packages/next/src/server/app-render/vary-params.ts
@@ -301,20 +301,30 @@ export function createVaryingSearchParams(
   accumulator: VaryParamsAccumulator,
   originalSearchParamsObject: SearchParams
 ): SearchParams {
-  const underlyingSearchParamsWithVarying: SearchParams = {}
-  for (const searchParamName in originalSearchParamsObject) {
-    Object.defineProperty(underlyingSearchParamsWithVarying, searchParamName, {
-      get() {
-        // TODO: Unlike path params, we don't vary track each search param
-        // individually. The entire search string is treated as a single param.
-        // This may change in the future.
+  // Search params have no fixed schema, so any access — missing-key reads, `in`
+  // checks, or enumeration — must register as varying. A Proxy is required
+  // (rather than per-property getters) so that enumeration of an empty
+  // searchParams object still triggers a vary. All accesses bucket into the
+  // single sentinel '?'; the segment is keyed by the whole query string.
+  // TODO: Split into per-param tracking if the cache key evolves.
+  return new Proxy(originalSearchParamsObject, {
+    get(target, prop, receiver) {
+      if (typeof prop === 'string') {
         accumulateVaryParam(accumulator, '?')
-        return originalSearchParamsObject[searchParamName]
-      },
-      enumerable: true,
-    })
-  }
-  return underlyingSearchParamsWithVarying
+      }
+      return Reflect.get(target, prop, receiver)
+    },
+    has(target, prop) {
+      if (typeof prop === 'string') {
+        accumulateVaryParam(accumulator, '?')
+      }
+      return Reflect.has(target, prop)
+    },
+    ownKeys(target) {
+      accumulateVaryParam(accumulator, '?')
+      return Reflect.ownKeys(target)
+    },
+  })
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/next/src/server/app-render/vary-params.ts
```diff
@@ -301,20 +301,50 @@ export function createVaryingSearchParams(
...
+  return new Proxy(originalSearchParamsObject, {
+    get(target, prop, receiver) {
+      if (typeof prop === 'string') {
+        // `.then` reads are ambiguous: `Promise.resolve` (when constructing the
+        // wrapping promise) and React Flight serialization (when checking if a
+        // resolved value is itself a thenable) both probe `.then` as part of
+        // the thenable protocol, and we don't want those probes to register as
+        // user dependencies. But user code can also read a `then` query
+        // parameter as a normal key. We can't tell the two apart, so we use key
+        // presence as a proxy: if the URL has `?then=...`, treat the access as
```
acdlite: Is this part implemented? I don't see the URL check

Also I think this would be relevant to the Proxy we use to track optional param access, too, right?

unstubbable: The `hasOwnProperty` check covers that.

acdlite: I fear this special case around "then"-named search params is worse than just deopting.

The deopt wouldn't be so bad if we tracked each search param individually (and converted to a bloom filter).

So maybe we should just accept the Flight serialization-triggered deopt for now and address later with per-param tracking.

unstubbable: Agreed after thinking it through. The filter created a (rare) correctness hole. A page that reads `searchParams.then`, prefetched at a URL without `?then=...`, would have had the access mis-suppressed and could have served a stale cache entry on a later navigation that does include `?then=...`. `then` being reserved by the Promise protocol made the combination unlikely, but a silent staleness bug is worse than a scoped perf deopt that we know how to address with per-param tracking.

Quick correction to my earlier comment. I'd called the late `.then` reads "React Flight serialization", but a stack trace shows they're just Promise `[[Resolve]]` probing the Proxy for thenable assimilation. The `promiseInitialized` flag in `makeUntrackedSearchParamsWithDevWarningsImpl` works there because the Proxy is only involved in one `[[Resolve]]` call, the initial `Promise.resolve`. In the runtime prerender chain, every derived Promise that resolves with the Proxy as its value triggers another probe, and the microtask ordering of a flag flip versus the next `[[Resolve]]` running on the Proxy isn't something we can reliably control. Per-param tracking still resolves it cleanly because the probe reads the same prop name with the same value across query strings that don't include `?then=...`.

I dropped the filter and skipped `shares cached segment across search params when not accessed (runtime prefetch)` with a TODO referencing the follow-up. This is the only test in the suite that exercises a runtime-prefetched page that doesn't read `searchParams` across multiple URLs differing only by query string, which is exactly the case that loses cache sharing under the deopt.

acdlite (approved):

PR #92621, opened 2026-04-10
[devtools] Preserve full code frames in browser overlay

aurorascharff:
Summary:
- stop devtools code frames from inheriting the server terminal width
- allow the overlay code frame to scroll horizontally instead of clipping content

Problem:
The browser error overlay requests originalCodeFrame from the dev server. That code frame was being generated with process.stdout.columns as the default max width, so if the dev server terminal was narrow the browser overlay received an already-truncated frame with collapsed ellipsis segments.

Fix:
- pass a large explicit maxWidth when generating devtools code frames on the server
- switch the overlay code frame container to horizontal scrolling instead of hidden overflow

Result:
The overlay no longer changes based on the terminal size running next dev, and long source lines remain inspectable in the browser.

Before:
Google Chrome 2026-04-10 17 51 17

After:
CleanShot 2026-04-10 at 18 16 48@2x

PR #93550, opened 2026-05-06
[devtools] Force ANSI colors on overlay code frames regardless of TTY

aurorascharff:
### What?

The dev overlay's code frame loses syntax highlighting whenever `next dev` runs under a non-TTY parent (VS Code / Cursor's integrated terminal, Docker without `-t`, `concurrently`, the JavaScript Debug Terminal, CI logs, etc.). Tokens render as plain monochrome text instead of the usual colored output.

### Why?

The "syntax highlighting" in the overlay isn't real highlighting. The server embeds ANSI escape codes into the code frame string, and the browser overlay parses them via Anser to render colored `<span>`s. Whether the dev server's stdout is a TTY is irrelevant on this path — the consumer is the browser.

Since #71860, `getOriginalCodeFrame` defaults `colors` to `process.stdout.isTTY`, and the overlay middlewares relied on that default. The moment stdout was a pipe, the overlay had nothing to color.

### How?

`createOriginalStackFrame` and `getOriginalStackFrames` are shared by five consumers (overlay HTTP endpoint, webpack `Module not found` plugin, browser-log forwarding, MCP error formatter, hot-reloader resolver). Hardcoding `colors: true` would leak escape sequences into the four non-overlay paths.

Thread a `codeFrameOptions: { colors?, maxWidth? }` parameter through both functions. The overlay HTTP middlewares opt in to `colors: true` and `maxWidth: DEVTOOLS_CODE_FRAME_MAX_WIDTH`; everyone else keeps the existing TTY-gated colors and terminal-width wrapping. (The 1000-column override from #92621 was suffering from the same leak — also scoped to the overlay now.)

Tested from Cursor's integrated terminal with/without fix:

Before:
CleanShot 2026-05-06 at 22 34 45@2x

After:
CleanShot 2026-05-06 at 22 34 30@2x

### DIFF
--- a/packages/next/src/server/dev/browser-logs/receive-logs.ts
+++ b/packages/next/src/server/dev/browser-logs/receive-logs.ts
@@ -1,5 +1,4 @@
 import { configure } from 'next/dist/compiled/safe-stable-stringify'
-import stripAnsi from 'next/dist/compiled/strip-ansi'
 import { cyan, dim, red, yellow } from '../../../lib/picocolors'
 import type { Project } from '../../../build/swc/types'
 import util from 'util'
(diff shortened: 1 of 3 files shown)


eps1lon (approved):

eps1lon (approved):
🌈

PR #93553, opened 2026-05-06
docs: server actions guide

icyJoseph:
Let's centralize our Server Actions education into a guide

### DIFF
--- a/docs/01-app/02-guides/server-actions.mdx
+++ b/docs/01-app/02-guides/server-actions.mdx
@@ -1,0 +1,214 @@
+---
+title: Server Actions and Mutations
+nav_title: Server Actions
+description: How Server Actions work in Next.js, including the single-roundtrip response model, sequential dispatch, security, and caching integration.
+related:
+  title: Next steps
+  description: Learn more about creating, securing, and configuring Server Actions.
+  links:
+    - app/getting-started/mutating-data
+    - app/guides/forms
+    - app/guides/data-security
+    - app/api-reference/directives/use-server
+    - app/api-reference/functions/revalidatePath
+    - app/api-reference/functions/revalidateTag
+    - app/api-reference/functions/redirect
+    - app/api-reference/functions/refresh
+    - app/api-reference/config/next-config-js/serverActions
+---
+
+A **Server Action** is a [React Server Function](https://react.dev/reference/rsc/server-functions) dispatched inside a [transition](https://react.dev/reference/react/startTransition). React wraps the call automatically when the function is passed to a `<form action>` or a `<button formAction>`. In Next.js, this typically maps to a data mutation: writing to a database, updating cookies, or invalidating cached data. You create one by adding the [`'use server'`](/docs/app/api-reference/directives/use-server) directive, then invoke it from a form, or from an event handler or `useEffect` wrapped in `startTransition`. For the basics of creating and invoking Server Functions, see [Mutating data](/docs/app/getting-started/mutating-data) and the [Forms guide](/docs/app/guides/forms).
+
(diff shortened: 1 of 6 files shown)


docs/01-app/02-guides/server-actions.mdx
```diff
@@ -0,0 +1,214 @@
...
+
+```ts filename="app/posts/actions.ts" highlight={6,7,8}
+'use server'
+
+import { auth } from '@/lib/auth'
+
+export async function deletePost(postId: string) {
+  const session = await auth()
+  if (!session?.user) throw new Error('Unauthorized')
+  if (!(await canDelete(session.user, postId))) throw new Error('Forbidden')
```
aurorascharff, line 105: Worth calling out authInterrupts?

aurorascharff (approved):

aurorascharff (approved):

docs/01-app/02-guides/server-actions.mdx
```diff
@@ -0,0 +1,219 @@
...
+- [`revalidateTag`](/docs/app/api-reference/functions/revalidateTag): stale-while-revalidate refresh of a tag with a cache-life profile. Subsequent reads get the stale value while a fresh fetch happens in the background, so the action's own re-render does **not** wait for the new data.
+- [`revalidatePath`](/docs/app/api-reference/functions/revalidatePath): invalidate by URL path. Use when one route is affected and tagging is overkill.
+- [`refresh`](/docs/app/api-reference/functions/refresh): refetch the current route's RSC Payload without invalidating cached data. Use when the view depends on state outside the cache that the action just changed.
+
+When `updateTag`, `revalidatePath`, or `refresh` runs, Next.js re-renders the current route server-side and includes a newly rendered [RSC Payload](/docs/app/glossary#rsc-payload) in the action's response, so the page reflects the change in the same roundtrip. `revalidateTag` with a stale-while-revalidate profile intentionally skips that immediate re-render.
```
aurorascharff, line 150: It is a bit wierd that revalidatePath behaves diffferently than revalidateTag in terms of SWR behavior

docs/01-app/02-guides/server-actions.mdx
```diff
@@ -0,0 +1,219 @@
...
+}
+```
+
+## Choosing a cache update
+
+After mutating data, on-demand revalidation updates the server cache, the client router, or both. Choose based on what needs to change:
+
+- [`updateTag`](/docs/app/api-reference/functions/updateTag): immediate expiration of a tag. The next read (including the route re-render that ships with the action's response) waits for fresh data. Use when the action needs **read-your-own-writes** so the user immediately sees their change. Server Actions only.
+- [`revalidateTag`](/docs/app/api-reference/functions/revalidateTag): stale-while-revalidate refresh of a tag with a cache-life profile. Subsequent reads get the stale value while a fresh fetch happens in the background, so the action's own re-render does **not** wait for the new data.
+- [`revalidatePath`](/docs/app/api-reference/functions/revalidatePath): invalidate by URL path. Use when one route is affected and tagging is overkill.
```
aurorascharff, line 147: It's unclear that this behaves differently than the tagin terms of swr behavior. It's stated below, but maybe it should be clear here right away

docs/01-app/02-guides/server-actions.mdx
```diff
@@ -0,0 +1,219 @@
...
+Unlike [`redirect`](/docs/app/api-reference/functions/redirect), none of these throw, so an action can call them and still return a value to the caller. See [How revalidation works](/docs/app/guides/how-revalidation-works) for the underlying model.
+
+## Configuration
+
+The [`serverActions`](/docs/app/api-reference/config/next-config-js/serverActions) option in `next.config.js` controls framework-level behavior:
+
+```js filename="next.config.js"
+/** @type {import('next').NextConfig} */
+module.exports = {
+  experimental: {
```
aurorascharff, line 161: Why is this still experimental btw?

PR #93578, opened 2026-05-07
Fix: Improved rewrite detection during optimistic routing

acdlite:
When optimistic routing is enabled, the client uses responses it receives from the server (during a navigation or a prefetch) to resolve the routes of structurally similar URLs.

However, if the server rewrites a URL to a different path than the one requested by the client, then we disable optimistic routing for that pattern.

We detect a rewrite by comparing the route tree returned from the server to the requested URL. If the route tree does not match, according to the Next.js routing algorithm, then the route must have been rewritten; we do not write this pattern into the client's route table.

The basic mechanism for this was already implemented, but there were some cases we weren't handling correctly:

- A static or non-optional dynamic segment is reached with no URL part to consume.
- A page leaf is reached with URL parts still pending.
- A dynamic segment whose URL part matches one of its known static siblings (which would have taken precedence under normal routing).

These cases have been fixed. Regression tests live in `optimistic-routing-rewrite-detection-regression/` and `optimistic-routing-parallel-slot-catchall-regression/`.

### DIFF
--- a/packages/next/src/client/components/segment-cache/optimistic-routes.ts
+++ b/packages/next/src/client/components/segment-cache/optimistic-routes.ts
@@ -209,8 +209,6 @@ export function discoverKnownRoute(
   const tree = routeTree
 
   const pathnameParts = pathname.split('/').filter((p) => p !== '')
-  const firstPart = pathnameParts.length > 0 ? pathnameParts[0] : null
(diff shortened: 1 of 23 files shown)


acdlite:
Re-opened as #93619 from a vercel/next.js branch to satisfy the CI requirement that PRs come from origin refs rather than forks.

PR #93582, opened 2026-05-07
fix(turbopack): use actual push index for module_evaluation chunk to prevent index-out-of-bounds panic

### LINKED ISSUE #93424: Turbopack panic: index out of bounds in tree_shake/graph.rs:743 when experimental.turbopackTreeShaking is enabled (next build, 16.2.4)

https://github.com/AbdQaddora/next-turbopack-tree-shaking-true-issue

### To Reproduce

``

1. Clone the linked reproduction repository.
2. Open `next.config.ts` and confirm `experimental.turbopackTreeShaking: true` is set (already enabled in the repro). This is the trigger — without it, the build completes successfully.
3. Run `pnpm install`.
4. Run `pnpm run build`.
5. The build aborts with a Rust panic in `turbopack-ecmascript`:

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Home
  Available memory (MB): 32648
  Available CPU cores: 12
Binaries:
  Node: 25.4.0
  npm: 11.7.0
  Yarn: N/A
  pnpm: 10.28.2
Relevant Packages:
  next: 16.1.6 // There is a newer version (16.2.4) available, upgrade recommended! 
  eslint-config-next: N/A
  react: 19.2.3
  react-dom: 19.2.3
  typescript: 5.9.3
```

Turbopack

next build (local)

### Additional context

- This is a build-time crash in Turbopack itself, not a runtime/deployment issue. It happens locally (Windows 11 x64) before any deploy target is reached.

sleitor:
## Problem

Fixes #93424

When `experimental.turbopackTreeShaking` is enabled, Turbopack panics with:
```
index out of bounds: the len is X but the index is Y
```
at `crates/turbopack-ecmascript/src/tree_shake/graph.rs` in `split_module`.

## Root Cause

In `split_module`, the second `for (ix, group) in groups.graph_ix.iter().enumerate()` loop processes all groups and builds chunks. When a group containing `is_module_evaluation = true` is found, the code immediately stores the enumerate index `ix` as the module evaluation index:

```rust
module_evaluation_ix = Some(ix as u32);
outputs.insert(Key::ModuleEvaluation, ix as u32);
```

However, later in the same loop iteration, there is a guard:
```rust
if chunk.body.is_empty() {
    continue;  // chunk is NOT pushed to modules
}
modules.push(chunk);
```

Empty chunks are skipped and not pushed to the `modules` vector. If any empty chunks appear **before** the module evaluation chunk, the enumerate index `ix` (e.g. `2`) no longer matches the actual position in the `modules` vector (e.g. `0`). Later, when the code accesses:

```rust
modules[module_evaluation_ix.unwrap() as usize].body.push(...)
```

it panics with index out of bounds.

## Fix

Instead of recording the module evaluation index at detection time (using the enumerate index), defer the assignment until the chunk is actually pushed to `modules`. A boolean flag `is_module_eval_chunk` marks the current iteration's chunk as the module evaluation chunk, and the actual `modules.len()` index is recorded just before pushing:

```rust
if is_module_eval_chunk {
    let actual_ix = modules.len() as u32;
    module_evaluation_ix = Some(actual_ix);
    outputs.insert(Key::ModuleEvaluation, actual_ix);
}
modules.push(chunk);
```

This ensures `module_evaluation_ix` always reflects the true position in the `modules` vector, regardless of how many empty chunks were skipped before it.


sokra:
The "advanced" tree shaking implementation is currently unmaintained (experimental anyway). We are planning to revisit that in future. But I feel good about this change and we could merge it. Best would be if you also add a test case for the bug.

sleitor:

jimmyhmiller:
@sleitor I took a look at your fix and while it does indeed fix the immediate problem, the code assumes that index and id always line up so skipping it causes problems later. If you check my change, I fixed this with adding a dummy entry. Not the most elegant, but keeps the code with its invariants in place and doesn't cause problems further down the line.

PR #93601, opened 2026-05-07
Encode non-ASCII characters in cache tags at construction

### LINKED ISSUE #93142: App Router: non-ASCII dynamic route params crash with ERR_INVALID_CHAR on x-next-cache-tags header

https://github.com/ornakash/nextjs-hebrew-slug-repro

### To Reproduce

```bash
git clone https://github.com/ornakash/nextjs-hebrew-slug-repro
cd nextjs-hebrew-slug-repro
pnpm install
pnpm build
pnpm start
```

Then in another shell:

```bash
curl -I http://localhost:3000/hello   # → 200 OK
curl -I http://localhost:3000/מידע    # → 500 Internal Server Error
```

The Hebrew slug is statically pre-rendered by `generateStaticParams`. The same 500 occurs when deployed to Vercel.

**Current:** any dynamic App Router route whose matched `params` value contains a non-ASCII character (Hebrew, Arabic, Chinese, emoji, ...) responds with 500 on every request. Server log:

```
TypeError: Invalid character in header content ["x-next-cache-tags"]
    at p (.next/server/chunks/ssr/..._dist_esm_build_templates_app-page_....js)
  code: 'ERR_INVALID_CHAR'
```

Root cause: Next.js writes the matched route path into the internal `x-next-cache-tags` HTTP header (used by the edge cache to index ISR responses by tag). HTTP headers are ASCII-only per RFC 7230, so any byte outside `0x00..0x7F` in the path triggers Node's `validateHeaderValue`. The tag value is not percent-encoded or otherwise sanitized before being written to the header.

**Expected:** Next.js should `encodeURI` (or similarly ASCII-escape) the tag value before writing `x-next-cache-tags`, so non-ASCII paths don't crash ISR. Internal tag comparisons already round-trip through a normalized form, so encoding should be compatible with invalidation.

#### Workarounds that do NOT work

- URL-encoding slugs in `generateStaticParams` — Next.js decodes and re-tags with the raw non-ASCII form at request time.
- Middleware `NextResponse.rewrite` to an ASCII path (including double-encoded and base64url-encoded variants) — the cache tag is built from the original request URL, not the rewritten internal path.
- Monkey-patching `Headers.prototype.set`/`.append`, `http.OutgoingMessage.prototype.setHeader`, `http.ServerResponse.prototype.setHeader`, `.writeHead`, `.setHeaders` from `instrumentation.ts` — all patches install but the throw originates from a bundled path in `.next/server/chunks/ssr/..._dist_esm_build_templates_app-page_...` that bypasses every JS-reachable setter.
- Patching `http.validateHeaderValue` — export is a non-configurable getter in the Vercel Node runtime.
- External caches (CDN / Redis / custom `cacheHandler`) — the origin response is already a 500; there is nothing valid to cache.

The **only** workaround found is `export const dynamic = 'force-dynamic'` on the affected route, which disables ISR entirely.

```
Operating System:
  Platform: win32
  Arch: x64
Binaries:
  Node: 22.22.0
  npm: 10.9.4
  pnpm: 10.33.0
Relevant Packages:
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.7.3
Next.js Config:
  output: N/A
```

Also reproduces on Vercel deployment; also reproduces with `next@16.2.2`.

Dynamic Routes, Headers, Internationalization (i18n), Runtime

`next start (local)`, Vercel (Deployed)

### Additional context

- `next@16.2.4` confirmed broken
- `next@16.2.2` confirmed broken
- Live Vercel deployment of the reproduction: https://nextjs-hebrew-slug-repro.vercel.app (`/hello` → 200, `/מידע` → 500)
- Reproduces on both local `next start` and Vercel production deployment, confirming it's a Next.js bug rather than a Vercel-runtime specific issue.
- Haven't bisected canary yet; happy to bisect if requested.
- Re-file of #93140 — the previous submission had an extra line under the reproduction-link section which tripped the auto-closing bot. Body and reproduction are otherwise identical.

unstubbable:
When a cache tag contains a non-ASCII character (Hebrew, CJK, emoji, …) it gets written into the internal `x-next-cache-tags` HTTP header on ISR responses. Node's `validateHeaderValue` rejects any byte outside `\t\x20-\x7e`, so the response crashes with `ERR_INVALID_CHAR`. On Vercel deploys stale-if-error masks the 500 from clients, but revalidation itself keeps failing and the cache stops refreshing for affected routes.

This change introduces a single `encodeCacheTag` helper and applies it at every public boundary — `validateTags` (which `cacheTag()`, `unstable_cache()`, and `fetch` tags all funnel through), `getImplicitTags` for path-derived tags, and `revalidatePath` / `revalidateTag` / `updateTag` for invalidation inputs. The encoder matches runs of out-of-class code units so surrogate pairs reach `encodeURIComponent` intact, and it is idempotent on already-encoded `%xx` sequences, so callers can pass either the raw or the encoded form interchangeably.

PR #93139 already encodes path-derived tags at construction, but it misses every user-supplied tag entry point and uses a `decodeURIComponent` round-trip that silently mangles literal `%xx` characters in tag values. PR #93167 encodes only at the `setHeader` sites, which leaves storage and invalidation diverging and requires every new write site to remember the encoding step. The canonical-form-at-the-boundary approach taken here covers all entry points and keeps storage, comparison, and the wire in sync.

fixes #93142
closes #93139
closes #93167

---

Co-authored-by: @swarnava 
Co-authored-by: @ornakash

### DIFF
--- a/packages/next/src/server/lib/encode-cache-tag.ts
+++ b/packages/next/src/server/lib/encode-cache-tag.ts
@@ -1,0 +1,37 @@
+/**
+ * Percent-encode every character outside printable ASCII so a tag value can be
+ * safely serialized as part of the `x-next-cache-tags` HTTP header.
+ *
+ * Node's `validateHeaderValue` rejects any code unit outside `\t\x20-\x7e`, so
+ * a matched route path or user-supplied tag containing a non-ASCII character
+ * (Hebrew, Arabic, Chinese, emoji, …) would otherwise throw `ERR_INVALID_CHAR`
+ * and crash ISR on every affected request.
+ *
+ * This is applied at the public boundaries — tag construction
+ * (`getImplicitTags`, `validateTags`) and invalidation input (`revalidatePath`,
+ * `revalidateTag`, `updateTag`) — so storage, comparison, and the wire all see
+ * the same canonical ASCII-safe form.
+ *
+ * The character class `[\t\x20-\x7e]` mirrors Node's `validHdrChars` table —
+ * `\t` plus printable ASCII through `~`. Anything outside that is rejected
+ * by `validateHeaderValue`, so we encode runs of those characters and leave
+ * everything else (`,`, `/`, `%`, `[`, `]`, `_`, `-`, `\t`, …) byte-for-byte
+ * unchanged. This preserves the comma-separated header format and the
+ * dynamic-segment markers in derived tags (`_N_T_/[slug]/page`).
+ *
+ * Properties:
+ * - Fast-path: input that already fits the validation class is returned
+ *   unchanged. This makes the encoder idempotent on already-encoded `%xx`
+ *   sequences.
+ * - Matches *runs* of out-of-class code units so surrogate pairs (e.g. an
+ *   emoji) are handed to `encodeURIComponent` as a complete code point — a
+ *   per-code-unit regex would split the pair and throw `URIError`.
+ */
+const OUT_OF_CLASS_CHAR = /[^\t\x20-\x7e]/
+const OUT_OF_CLASS_RUN = /[^\t\x20-\x7e]+/g
+
+export function encodeCacheTag(tag: string): string {
+  return OUT_OF_CLASS_CHAR.test(tag)
+    ? tag.replace(OUT_OF_CLASS_RUN, (run) => encodeURIComponent(run))
+    : tag
+}
--- a/packages/next/src/server/lib/implicit-tags.test.ts
+++ b/packages/next/src/server/lib/implicit-tags.test.ts
@@ -73,6 +73,34 @@ describe('getImplicitTags()', () => {
         '_N_T_/foo/bar/baz',
       ],
     },
+    {
+      // Non-ASCII pathname must be percent-encoded so it can be safely
+      // serialized into the `x-next-cache-tags` HTTP header. Surrogate-pair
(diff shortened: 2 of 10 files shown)


unstubbable:
* **#93601** <a href="https://app.graphite.com/github/pr/vercel/next.js/93601?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93601?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#93617** <a href="https://app.graphite.com/github/pr/vercel/next.js/93617?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

gnoff (approved):
I think this is landable but we should consider perturbing encoded keys so they cannot be matched synonymous key inputs

unstubbable:
### Merge activity

* **May 7, 10:42 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93601).
* **May 7, 10:43 PM UTC**: [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93601) couldn't merge this PR because it failed for an unknown reason (unsigned commits detected).

PR #93603, opened 2026-05-07
Bugfix: Instant navs devtools was not showing the final Page Load state

samselikoff:
The instant navs devtools wasn't showing the final state for Page Load because it wasn't setting `null` as the third value of the cookie.

Added a test that was failing to cover this as well.

### DIFF
--- a/packages/next/src/next-devtools/dev-overlay/components/instant-navs/instant-navs-panel.tsx
+++ b/packages/next/src/next-devtools/dev-overlay/components/instant-navs/instant-navs-panel.tsx
@@ -44,7 +44,7 @@ export function InstantNavsPanel() {
     if (typeof cookieStore !== 'undefined') {
       cookieStore.set({
         name: COOKIE_NAME,
-        value: JSON.stringify([0, `p${Math.random()}`]),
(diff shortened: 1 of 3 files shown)


acdlite (approved):

packages/next/src/next-devtools/dev-overlay/components/instant-navs/instant-navs-panel.tsx
```diff
@@ -44,7 +44,7 @@ export function InstantNavsPanel() {
     if (typeof cookieStore !== 'undefined') {
       cookieStore.set({
         name: COOKIE_NAME,
-        value: JSON.stringify([0, `p${Math.random()}`]),
+        value: JSON.stringify([0, `p${Math.random()}`, null]),
```
acdlite, line 47: perhaps let's import the type into here to prevent future drift?

```ts
const cookie: InstantCookie = [...]
```

samselikoff: Oh great call 👍

samselikoff:
Superseded by https://github.com/vercel/next.js/pull/94027

PR #93606, opened 2026-05-07
[Turbopack] Add graph-based CSS chunking algorithm behind experimental.cssChunking: "graph"

sokra:
### What?

Adds an alternative CSS chunking algorithm to Turbopack, opted into via:

```js
// next.config.js
module.exports = {
  experimental: {
    cssChunking: 'graph',
    // or, with explicit cost overrides:
    // cssChunking: { type: 'graph', requestCost: 20_000, moduleFactorCost: 1 },
  },
}
```

The new algorithm is **off by default** — Turbopack still uses the existing
"loose"/dependencies algorithm unless this flag is set, so this PR is a pure
addition for users that don't opt in.

While we were here, the `experimental.cssChunking` shape was also generalized
so every existing string accepts an object form too:

| Value | Bundler | Notes |
|---|---|---|
| `true` / `'loose'` / `{ type: 'loose' }` | both | default heuristic-based chunking |
| `'strict'` / `{ type: 'strict' }` | webpack | unchanged |
| `false` | webpack | unchanged (one chunk per CSS module) |
| `'graph'` / `{ type: 'graph', requestCost?, moduleFactorCost? }` | Turbopack | new |

Cross-bundler combinations are rejected at config-validation time:
- `'graph'` with webpack throws.
- `'strict'` and `false` with Turbopack throw.

### Why?

The existing Turbopack CSS chunker (loose / dependencies) is good at
preserving CSS ordering but doesn't share chunks across pages well — every
page tends to load its own chunk per CSS module, which scales poorly for apps
with many pages and shared component libraries.

The new "graph" algorithm models the per-chunk-group CSS ordering as a
weighted DAG over modules, then greedily merges adjacent runs in the global
topological order whenever the merge reduces total cost. The cost model
charges every CSS request and overshipped byte, with two tunable knobs
(`requestCost` and `moduleFactorCost`).

**Trade-off vs. the loose default.** With the default cost parameters
(`requestCost: 20_000`, `moduleFactorCost: 1`) the graph algorithm typically
ships **less CSS per chunk group at the cost of more requests** than the
loose algorithm. The cost model is tuned to avoid overshipping unrelated CSS
into pages that don't need it; on apps where the loose algorithm was already
collapsing a lot into one big chunk that some pages didn't actually use, the
graph algorithm will split it. Apps that prefer fewer requests can raise
`requestCost`; apps that prefer less overshipping can raise
`moduleFactorCost`.

This is opt-in and Turbopack-only because:
- The cost model is sensitive to per-app properties (number of pages, size
  distribution of CSS modules, …) — keeping it experimental gives us room to
  tune defaults from real usage.
- Webpack already has its own `CssChunkingPlugin` and `'strict'` mode that
  cover the equivalent design space; we don't want to fork that.

### Performance

Measured on `vercel.com` (the full graph algorithm spans
`create_graph → make_acyclic → linearize → split_into_chunks → assemble`):

- **~3s** end-to-end for the synchronous chunking pipeline on a realistic
  production input.

Implementation choices that matter for that throughput:

- Tarjan SCC uses `Vec<u32>` / `Vec<bool>` scratch arrays indexed by
  `NodeIndex` — no hashing on `indices` / `lowlinks` / `on_stack`.
- `make_acyclic` batches multiple cuts per SCC pass by seeding successive
  short-cycle searches at the previous cut's target, only re-running Tarjan
  when no further cycle is reachable from the seed.
- `find_short_cycle` is a bidirectional Dijkstra over a `BinaryHeap` with
  predecessor pointers (no path cloning) and skips its refinement loop for
  trivial 2-cycles.
- `split_into_chunks` picks the next merge from a `BinaryHeap` keyed on the
  cost delta instead of an O(N) linear scan per merge.
- `chunk_cost` reads a once-built `module_to_groups` inverse index instead of
  scanning every chunk group on every call; the GlobalStyle leakage check
  uses binary search on the inverse index rather than scanning each group's
  module list.

#### Module layout (`turbopack/crates/turbopack-core/src/module_graph/`)

The two algorithms are deliberately split so neither imports from the other:

- `style_groups/` — algorithm-neutral output types (`StyleGroups`,
  `StyleItemInfo`, `make_style_groups`). Both algorithms produce these.
- `style_groups_loose/` — the existing ("loose") algorithm plus the shared
  config types (`StyleGroupsAlgorithm`, `StyleGroupsConfig`,
  `F32TaskInput`).
- `style_groups_graph/` — the new algorithm. Pure Rust, no `Vc`, with
  `petgraph::DiGraph` plus a thin `SubgraphView` wrapper and a small
  `ReadonlyGraph` trait that lets the same pipeline run against either a
  `&DiGraph` or a filtered view of one SCC.

#### Algorithm

```text
create_graph → make_acyclic → linearize → split_into_chunks → assemble batches
```

1. **`create_graph`** — for each chunk group, every `(later, earlier)` pair
   inside the group's CSS-module list becomes an edge `later → earlier`
   (weight 1, accumulated). Heavy edges = strong co-occurrence.
2. **`make_acyclic`** — co-occurrence almost always introduces cycles; each
   multi-node SCC has its lowest-weight cycle edge cut until the graph is a
   DAG.
3. **`linearize`** — Kahn-style topological sort with a tie-break on edge
   weight, so strongly co-occurring modules end up adjacent in the global
   order.
4. **`split_into_chunks`** — greedy bottom-up merger over the global order.
   At every active split point we score the merge as
   `cost(merged) - cost(left) - cost(right)`, take the most-negative score
   from a min-heap, and repeat until no merge would reduce cost.
   `max_chunk_size` and "global CSS must not leak into unrelated chunk
   groups" are enforced as `+infinity` cost.

The cost model is:

```text
cost_per_group(chunk, group)
  = chunk_size
  + (chunk_size / group_total_size) * module_factor_cost
  + request_cost
```

summed over the chunk groups that load the chunk.

#### Wiring

- `StyleGroups::shared_chunk_items` is a
  `FxIndexMap<ChunkItemWithAsyncModuleInfo, StyleItemInfo>` where
  `StyleItemInfo { order: Option<u32>, batch: Option<…> }`. The graph
  algorithm fills `order` so `style_production.rs` can stable-sort chunks
  globally; the legacy algorithm leaves `order = None`, which makes the sort
  a no-op for it. `flatten_and_sort` returns the `StyleItemInfo` references
  alongside each chunk item so the per-item loop doesn't re-query the map.
- A new `StyleGroupsAlgorithm` enum on `ChunkingConfig` selects the
  algorithm at chunking time; `ModuleGraph::style_groups` dispatches
  to either `compute_style_groups` (existing) or `compute_style_groups_graph`
  (new).
- `next-core` exposes `NextConfig::css_chunking() -> Vc<CssChunkingAlgorithm>`
  resolving the JS `experimental.cssChunking` to the Rust enum, with cost
  defaults applied (`requestCost: 20_000`, `moduleFactorCost: 1`). All three
  chunking-context constructors (`next_client`, `next_edge`,
  `next_server`) thread it through.

#### Configuration

- `experimental.cssChunking` zod schema accepts the new shapes; cost params
  are `z.number().nonnegative().finite().optional()`.
- `config-shared.ts` exports a `CssChunkingConfig` type alias and a
  `resolveCssChunkingMode(value)` helper that normalizes any input to one
  of `'off' | 'loose' | 'strict' | 'graph'`. Both `webpack-config.ts` (plugin
  wiring) and `config.ts` (bundler-compat validation) use the helper.
- New `errors.json` entries for the three bundler-compatibility validation
  errors (E1193 graph-on-webpack, E1194 strict-on-Turbopack, E1195
  false-on-Turbopack).

#### Tests

- 53 Rust unit tests in `style_groups_graph/tests.rs` cover `create_graph`,
  Tarjan SCC, `find_short_cycle` (bidirectional Dijkstra), `make_acyclic`,
  `linearize`, `split_into_chunks`, and end-to-end pipeline scenarios.
- `test/e2e/app-dir/css-order/css-order.test.ts` is parametrised over
  `[label, value]` pairs. The Turbopack matrix now includes `'graph'` and an
  object-form `{ type: 'graph', requestCost: 1, moduleFactorCost: 1 }` in
  addition to the existing default. Per-page expectations grew a `requests`
  object encoding distinct request counts for `loose` and `graph` where they
  differ.
- A new `sandwich` e2e fixture (`/sandwich/a`, `/sandwich/b`) exercises the
  case where two pages share a leading and trailing chunk around a unique
  middle stylesheet — including a global stylesheet that the algorithm must
  not leak into unrelated chunk groups. The graph algorithm hits the optimal
  3 chunks per page on this fixture; loose mode falls short.

#### Documentation

- `ExperimentalConfig.cssChunking` JSDoc describes every accepted shape and
  what each cost knob does.
- The `style_groups_graph` module-level docs describe the pipeline, cost
  model and constraints with diagrams.

Closes NEXT-


crates/next-core/src/next_config.rs
```diff
@@ -1040,6 +1041,119 @@ pub struct TurbopackIgnoreIssueRule {
...
+    pub request_cost: Option<f32>,
+    pub module_factor_cost: Option<f32>,
+}
+
+impl CssChunkingConfig {
+    /// Normalize all input shapes (booleans, strings, object form) to the canonical object form.
+    pub fn normalize(&self) -> CssChunkingObject {
+        match self {
+            // `false` is treated like `true`/loose: the "off" path is selected via the existing
+            // `ChunkingContext` defaults, not via this option.
```
mischnic: I don't understand this comment. 
Also, the PR description says that `false` should throw with Turbopack?

crates/next-core/src/next_config.rs
```diff
@@ -1040,6 +1041,119 @@ pub struct TurbopackIgnoreIssueRule {
...
+            opts.request_cost.unwrap_or(DEFAULT_REQUEST_COST),
+            opts.module_factor_cost
+                .unwrap_or(DEFAULT_MODULE_FACTOR_COST),
+        ),
+    }
+}
+
+/// `Vc`-wrapped variant of [`StyleGroupsAlgorithm`] returned from [`NextConfig::css_chunking`].
+#[turbo_tasks::value(transparent)]
+pub struct CssChunkingAlgorithm(pub StyleGroupsAlgorithm);
```
mischnic: Why not make StyleGroupsAlgorithm cellable instead?

crates/next-core/src/next_config.rs
sokra: :robot: Good catch — addressed in 93becc6dc27. `normalize` now maps `false` to a new `CssChunkingObject::None` (rather than collapsing it to `Loose`), and `resolve_css_chunking_algorithm` returns `Result<StyleGroupsAlgorithm>` and `bail!`s on both `None` and `Strict` — the cases that are bundler-incompatible with Turbopack and should never reach this code path after JS-side validation.

crates/next-core/src/next_config.rs
sokra: :robot: Done in 93becc6dc27. `StyleGroupsAlgorithm` is now `#[turbo_tasks::value(shared, operation)]` directly, so `Vc<StyleGroupsAlgorithm>` is the cell type and the `CssChunkingAlgorithm(pub StyleGroupsAlgorithm)` wrapper is gone. `css_chunking()` returns `Result<Vc<StyleGroupsAlgorithm>>` (also picking up the error path from the other comment).

turbopack/crates/turbopack-core/src/module_graph/style_groups_graph/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+
+    // Defensive fallback: emit any module that the algorithm didn't reach (e.g. a node dropped
+    // by `linearize` because a residual cycle survived `make_acyclic`) as a singleton at the
+    // end. In practice this loop is a no-op — `linearize` operates on a DAG and processes every
+    // node — but it keeps the result complete if a future change ever breaks that invariant.
+    for data in module_data {
+        if !shared_chunk_items.contains_key(&data.chunk_item) {
+            push(&mut shared_chunk_items, data.chunk_item, None);
+        }
+    }
```
mischnic: This should just be a (debug) assertion that all of them were included?

turbopack/crates/turbopack-core/src/module_graph/style_groups_graph/mod.rs
```diff
@@ -0,0 +1,365 @@
...
+///
+/// The outer traversal — `chunk_group_info`, `batches_graph`, the parallel-edge filter inside
+/// `traverse_edges_from_entries_dfs`, the `ModuleOrBatch` match — is structurally identical to
+/// [`super::style_groups_loose::collect_style_modules_per_chunk_group`]. The two diverge in the
+/// per-module bookkeeping they need (loose builds [`super::style_groups_loose::ModuleInfo`]
+/// with per-chunk-group positions, an `index_sum`, an `ident`, etc.; the graph algorithm only
+/// needs a densely-packed CSS module id per group plus a parallel sidecast vector), so an
+/// earlier extraction attempt got reverted (commits `37d983f9` → `c05c4587`). Keep the two
+/// walks separate until something concrete changes; a generic "collect modules per chunk
+/// group" callback isn't worth the API surface.
```
mischnic: this is irrelevant?

sokra: Yeah should be removed

turbopack/crates/turbopack-core/src/module_graph/style_groups_graph/mod.rs
sokra: Yep

packages/next/errors.json
```diff
@@ -1238,7 +1238,8 @@
...
-  "1245": "Route \"%s\": Next.js encountered %s in a Client Component.\\n\\nThis value would be evaluated during the prerender and fixed at build time, instead of recomputed on each visit.\\n\\nWays to fix this:\\n  - Wrap the Client Component in \\`<Suspense fallback={...}>\\`\\n  - Move the read into a \\`useEffect\\` or event handler\\n\\nLearn more: %s",
+  "1243": "\\`experimental.cssChunking: \"graph\"\\` is only supported with Turbopack. Please remove the option or run Next.js with Turbopack in %s.",
+  "1244": "\\`experimental.cssChunking: \"strict\"\\` is only supported with webpack. Please remove the option or run Next.js with webpack in %s.",
+  "1245": "\\`experimental.cssChunking: false\\` is only supported with webpack. Please remove the option or run Next.js with webpack in %s.",
```
mischnic: Now it just overwrites existing error codes?
Did some AI update errors.json instead of using the script?

sokra: Stupid AI rebase...

sokra: Actually it was using the script:

```
$ cd /vercel/sandbox/repo && git rebase origin/canary 2>&1 | tail -10
Rebasing (1/29)
merge-errors-json: added 1 new message to errors.json
Rebasing (2/29)
Rebasing (3/29)
Rebasing (4/29)
Rebasing (5/29)
Rebasing (6/29)
merge-errors-json: added 0 new messages to errors.json
Rebasing (7/29)
Rebasing (8/29)
Rebasing (9/29)
Rebasing (10/29)
Rebasing (11/29)
Rebasing (12/29)
Rebasing (13/29)
Rebasing (14/29)
Rebasing (15/29)
Rebasing (16/29)
Rebasing (17/29)
Rebasing (18/29)
Rebasing (19/29)
Rebasing (20/29)
Rebasing (21/29)
Rebasing (22/29)
Rebasing (23/29)
Rebasing (24/29)
Rebasing (25/29)
Rebasing (26/29)
Rebasing (27/29)
Rebasing (28/29)
Rebasing (29/29)
Successfully rebased and updated refs/heads/sokra/css-chunking.
```

There was not conflict rebasing...

mischnic (approved):

raquelpn1979-tech:
¡Entendido! Si necesitas analizar algo del archivo, por favor, copia y pega aquí su contenido o describeme qué clase de información contiene. Si tienes problemas accediendo al archivo o necesitas ayuda con Dropbox o la integración del contenido en un contexto como GitHub o APIs, házmelo saber. ¡Estoy aquí para ayudar!

PR #93612, opened 2026-05-07
Fix "type: module" in project dir when using standalone or adapters

timneutkens:
### What?

Add `.next/package.json` (the distDir commonjs boundary marker) to the `required-server-files.json` manifest.

### Why?

`next build` writes `.next/package.json` with `{"type": "commonjs"}` so that everything in `.next/**` is loaded as CJS even when the user's project `package.json` has `"type": "module"`. Without that file, Node walks up to the project `package.json` when loading `.next/server/**/*.js` and tries to evaluate the compiled server bundles as ESM, which fails at runtime.

This file is necessary on disk in every server deployment artifact (the standalone bundle, every adapter Node function), but it was not declared in `required-server-files.json`. As a result, adapters consuming `requiredServerFiles.files` did not include the boundary file.

`.next/package.json` is conceptually a required server file: any deployment that wants to run `.next/server/**/*.js` correctly under a `"type": "module"` project must ship it. Declaring it in the manifest is the right place to express that contract once for every consumer.

This error case would only happen if the project package.json gets included into the Node.js functions. For example when file tracing analysis (node-file-trace) decides it needs to include it (i.e. glob patterns or such).

### How?

Add `'package.json'` to the `requiredServerFilesManifest.files` array in `packages/next/src/build/index.ts`. Every entry in that array is joined with `config.distDir`, so this resolves to `.next/package.json`.
Both existing consumers pick it up automatically with no further changes:
- `packages/next/src/build/adapter/build-complete.ts` iterates `requiredServerFiles` into `sharedNodeAssets`, which is merged into every Node function's `output.assets` regardless of bundler. The Vercel adapter (and any other adapter) gets the file for free.
- `copyTracedFiles` in `packages/next/src/build/utils.ts` iterates `requiredServerFiles.files` and copies each entry into `.next/standalone/<relative>`, so the standalone output now contains `.next/standalone/.next/package.json`.

### Tests
- `test/production/required-server-files-package-boundary/` — a focused, non-standalone test that builds a `"type": "module"` app and asserts:
  - `.next/package.json` is `{"type": "commonjs"}` on disk
  - `.next/required-server-files.json` lists `.next/package.json`
- `test/production/standalone-mode/type-module/` — extended with a `pages/dynamic.js` that uses `getServerSideProps`. The test now:
  - Asserts `.next/standalone/.next/package.json` exists with `{"type": "commonjs"}`
  - Boots the standalone `server.js` and fetches `/dynamic`, which forces Node to actually evaluate `.next/server/pages/dynamic.js` at runtime — the exact code path the boundary file makes work

### DIFF
--- a/packages/next/src/build/index.ts
+++ b/packages/next/src/build/index.ts
@@ -1852,6 +1852,11 @@ export default async function build(
             appDir: dir,
             relativeAppDir: path.relative(outputFileTracingRoot, dir),
             files: [
+              // Marks distDir as commonjs so server output is loaded correctly
+              // when the user's project package.json is "type": "module".
+              // See `writeFileUtf8(path.join(distDir, 'package.json'), '{"type": "commonjs"}')`
+              // earlier in this file.
+              'package.json',
               ROUTES_MANIFEST,
               path.relative(distDir, pagesManifestPath),
               BUILD_MANIFEST,
--- a/packages/next/src/build/index.ts
+++ b/packages/next/src/build/index.ts
@@ -1852,10 +1852,18 @@ export default async function build(
             appDir: dir,
             relativeAppDir: path.relative(outputFileTracingRoot, dir),
             files: [
-              // Marks distDir as commonjs so server output is loaded correctly
+              // Marks distDir as commonjs so server output is loaded as CJS
               // when the user's project package.json is "type": "module".
-              // See `writeFileUtf8(path.join(distDir, 'package.json'), '{"type": "commonjs"}')`
-              // earlier in this file.
+              // See `writeFileUtf8(path.join(distDir, 'package.json'),
+              // '{"type": "commonjs"}')` earlier in this file.
+              //
+              // Standalone already picks this up via `next-server.js.nft.json`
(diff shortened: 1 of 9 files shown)


mischnic:
Turbopack lists it in the NFT files, not sure about Webpack though

packages/next/src/build/index.ts
```diff
@@ -1852,6 +1852,19 @@ export default async function build(
...
+              // See `writeFileUtf8(path.join(distDir, 'package.json'),
+              // '{"type": "commonjs"}')` earlier in this file.
+              //
+              // Standalone already picks this up via `next-server.js.nft.json`
+              // (nft traces it as `./package.json`), but per-page nft files do
+              // not include it. The adapter codepath (`handleBuildComplete`)
+              // builds each page's `assets` from `requiredServerFiles` plus
+              // the per-page trace, so without listing it here the boundary
+              // is missing from adapter outputs and Node walks up to the
+              // user's `"type": "module"` package.json at runtime.
```
mischnic: that comment is outdated now?

mischnic (approved):

PR #93633, opened 2026-05-08
bfcacheId: Opt out of state preservation

acdlite:
When `cacheComponents` is enabled, the App Router preserves state across navigations by rendering inactive routes inside React `<Activity>` boundaries.

As a default behavior, this is a huge convenience because it lets you navigate between routes without resetting or losing ephemeral UI state (scroll position, expand/collapse, in-progress form edits). Previously the only way to do this was to explicitly track each state with an external state manager, or hoist it to a parent component.

It's still the often the case that you should be explicitly tracking all important UI state, anyway, so that it survives a hard refresh of the app, or the browser window being accidentally closed. For example, forms draft states should be persisted to a server-side database or local storage engine. If you're already doing that, then it doesn't matter so much whether  client state is preserved via `<Activity>` boundaries or not.

For the long tail of ephemeral state that is not tracked, Next.js's philosophy is that it's a better UX default to preserve as much ephemeral state as possible. It's easier to model the cases where you _do_ want state to be reset on navigation as exceptions, compared to the other way around.

However, this is a significant change compared to the pre-Cache Components previous behavior of Next.js, and compared to other web frameworks, and indeed the browser's own native bfcache (which implements state restoration for history traversal navigations only, not push/replace).

There's are also lots of existing codebases that may rely on the current behavior, and may break subtly under these new semantics. Even if this new default unlocks better UX patterns, we don't want to force everyone to migrate all their code all at once.

So, this PR introduces a drop-in mechanism for opting out of state preservation when navigating to a previously visited route.

The API is exposed as `useRouter().bfcacheId`. It's intended to be passed to a React `key`:

  <form key={useRouter().bfcacheId}>

The id is contextual: read from a layout, you get the layout's id; read from a page, you get the page's id. It's stable across back/forward navigations, `router.refresh()`, server actions that call `refresh()`, and search-param- or hash-only navigations — i.e., any time the surrounding segment is preserved. It changes when the segment is freshly created by a push or replace into a different route.

An important detail is that the previous id is restored during a back/ foward navigation. So state preservation will still work if you navigate via the browser's back button.

Why add this to `useRouter()` instead of giving it its own hook? The intent is communicate that `bfcacheId` is not considered an idiomatic pattern — the recommended fix for "I want this state to reset on navigation" is almost always something else: an explicit reset in a submit handler, or a key derived from the underlying data (e.g., a draft id from the server). `useRouter` is the hook where we expose low-level APIs that are supported but are only recommended for advanced or exceptional cases. (For example, instead of `router.push()`, you should almost always use a `<Link>` component instead.)

### DIFF
--- a/test/e2e/app-dir/use-router-bfcache-id/use-router-bfcache-id.test.ts
+++ b/test/e2e/app-dir/use-router-bfcache-id/use-router-bfcache-id.test.ts
@@ -1,0 +1,172 @@
+import { nextTestSetup } from 'e2e-utils'
+import type * as Playwright from 'playwright'
+import { createRouterAct } from 'router-act'
+
+describe('use-router-bfcache-id', () => {
+  const { next } = nextTestSetup({
+    files: __dirname,
+  })
+
+  async function setup(initialPath: string) {
+    let page: Playwright.Page
+    const browser = await next.browser(initialPath, {
+      beforePageLoad(p: Playwright.Page) {
+        page = p
+      },
+    })
+    const act = createRouterAct(page!)
+    return { browser, act }
+  }
+
+  it('preserves leaf form state on browser back navigation', async () => {
+    const { browser, act } = await setup('/x/1')
+    await browser.elementByCss('[data-testid="leaf-input"]').type('hello')
+
+    await act(async () => {
+      await browser.elementByCss('input[data-link-accordion="/x/2"]').click()
+      await browser.elementByCss('a[href="/x/2"]').click()
+    })
+
+    await browser.back()
+    expect(await browser.elementByCss('[data-testid="pathname"]').text()).toBe(
+      '/x/1'
+    )
+    expect(
+      await browser.elementByCss('[data-testid="leaf-input"]').getValue()
+    ).toBe('hello')
+  })
+
+  it('resets leaf form state when re-entering a route via fresh push', async () => {
+    const { browser, act } = await setup('/x/1')
+    await browser.elementByCss('[data-testid="leaf-input"]').type('hello')
+
+    await act(async () => {
+      await browser.elementByCss('input[data-link-accordion="/x/2"]').click()
+      await browser.elementByCss('a[href="/x/2"]').click()
+    })
+
+    // Navigate back to /x/1 via a fresh push (not the browser back button).
+    await act(async () => {
+      await browser.elementByCss('input[data-link-accordion="/x/1"]').click()
+      await browser.elementByCss('a[href="/x/1"]').click()
+    })
+
+    expect(
+      await browser.elementByCss('[data-testid="leaf-input"]').getValue()
+    ).toBe('')
+  })
+
+  it('preserves shared layout state across sibling leaf navigations', async () => {
+    const { browser, act } = await setup('/x/1')
+    await browser.elementByCss('[data-testid="layout-input"]').type('layout')
+
+    await act(async () => {
+      await browser.elementByCss('input[data-link-accordion="/x/2"]').click()
+      await browser.elementByCss('a[href="/x/2"]').click()
+    })
+
+    // The [group] layout is shared across /x/1 and /x/2, so its form state
+    // survives the leaf navigation.
+    expect(
+      await browser.elementByCss('[data-testid="layout-input"]').getValue()
+    ).toBe('layout')
+  })
+
+  it('resets shared layout state when navigating across groups', async () => {
+    const { browser, act } = await setup('/x/1')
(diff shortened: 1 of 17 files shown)


test/e2e/app-dir/use-router-bfcache-id/use-router-bfcache-id.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      await browser
+        .elementByCss('[data-testid="server-action-refresh"]')
+        .click()
+    })
+
+    expect(
+      await browser.elementByCss('[data-testid="leaf-input"]').getValue()
+    ).toBe('hello')
+  })
+})
```
samselikoff, line 172: Was playing with the test app and think I found two buggy paths, if I'm understanding all the behavior changes correctly.

The first was when navigating from "x/1" to "x/2" to "x/3", then back, then back. The first entry showed the third entry's input text.

Here's a failing test:

```tsx
it('preserves leaf form state after multiple entries to the same url', async () => {
  // Entry 1
  const { browser, act } = await setup('/x/1')
  await browser.elementByCss('[data-testid="leaf-input"]').type('hello 1')

  // Forward nav (Entry 2 at "x/2")
  await act(async () => {
    await browser.elementByCss('input[data-link-accordion="/x/2"]').click()
    await browser.elementByCss('a[href="/x/2"]').click()
  })
  await browser.elementByCss('[data-testid="leaf-input"]').type('hello 2')

  // Forward nav (Entry 3 at "x/1")
  await act(async () => {
    await browser.elementByCss('input[data-link-accordion="/x/1"]').click()
    await browser.elementByCss('a[href="/x/1"]').click()
  })
  await browser.elementByCss('[data-testid="leaf-input"]').type('hello 3')

  // Back to Entry 2, assert
  await browser.back()
  expect(await browser.elementByCss('[data-testid="pathname"]').text()).toBe(
    '/x/2'
  )
  expect(
    await browser.elementByCss('[data-testid="leaf-input"]').getValue()
  ).toBe('hello 2')

  // Back to Entry 1, assert
  await browser.back()
  expect(await browser.elementByCss('[data-testid="pathname"]').text()).toBe(
    '/x/1'
  )
  expect(
    await browser.elementByCss('[data-testid="leaf-input"]').getValue()
  ).toBe('hello 1')
})
```

And the second one was when navigating from "/x/1" to "/x/1?q=2" to "/x/2", then back. The state didn't seem to be restored.

Failing test:

```ts
it('preserves leaf form state after a search param update, a forward nav, then a browser back', async () => {
  const { browser, act } = await setup('/x/1')
  await browser.elementByCss('[data-testid="leaf-input"]').type('hello')

  // Forward to /x/1?q=2
  await act(async () => {
    await browser
      .elementByCss('input[data-link-accordion="/x/1?q=2"]')
      .click()
    await browser.elementByCss('a[href="/x/1?q=2"]').click()
  })

  expect(
    await browser.elementByCss('[data-testid="search"][data-value="q=2"]')
  ).toBeDefined()
  expect(
    await browser.elementByCss('[data-testid="leaf-input"]').getValue()
  ).toBe('hello')

  // Forward nav to /x/2
  await act(async () => {
    await browser.elementByCss('input[data-link-accordion="/x/2"]').click()
    await browser.elementByCss('a[href="/x/2"]').click()
  })

  // Back to Entry 1 /x/1?q=2, assert
  await browser.back()
  expect(await browser.elementByCss('[data-testid="pathname"]').text()).toBe(
    '/x/1'
  )
  expect(
    await browser.elementByCss('[data-testid="search"][data-value="q=2"]')
  ).toBeDefined()
  expect(
    await browser.elementByCss('[data-testid="leaf-input"]').getValue()
  ).toBe('hello')
})
```

samselikoff: My first scenario was a misunderstanding: we don't automatically add the bfcacheId to the cache key for segments, so re-visits to the same url (e.g. "/x/1") via browser back/forward should in fact show the same uncontrolled client state.

The second scenario is now working.

samselikoff (dismissed):

samselikoff (approved):

Chibuike-web:
I don't know, but this feels like a bolted-on solution.

Can we just configure preservation explicitly in cacheComponents instead of manually keying trees with router.bfcacheId?

Something like:

cacheComponents: {
  preserveRoutes: [
    "/dashboard",
    "/patients",
    "/settings/profile"
  ]
}
This feels more declarative and easier to reason about.

PR #93648, opened 2026-05-08
(7/7) Add route replay miss coverage for incomplete router data

feedthejim:
## Stack position

This is PR 7/7 in the first `offlineNavigations` client-router persistence stack.

It sits after #93647, which teaches the client to replay a dynamic route from the persisted known-route table when the exact target URL entry is absent. This PR is intentionally test-only: it makes the new replay path prove that incomplete persisted router-cache data turns into a visible offline miss instead of a partial or stale render.

The stack is still behind `experimental.offlineNavigations`, which requires Cache Components and implies the existing offline surface. The service worker continues to own document survival only; the cache-components client router owns persisted route/segment/head data, route matching, and replay eligibility.

## Why this exists

Dynamic route replay is the riskiest part of the first client-side stack because the route table can know that `/dynamic-prefetch/[value]` is replayable even when a specific target URL has never had an exact URL record. That is useful, but only if replay refuses to continue when one of the router-cache records needed to reconstruct the target tree is missing.

This test covers that boundary directly. It creates the successful PR6 setup, deletes the required target page segment record from IndexedDB, then verifies the offline hard load misses with the expected diagnostics and user-visible fallback.

## What changed

- Refactored the dynamic-pattern prefetch setup into a shared helper so the success and failure cases assert the same persisted state before replay.
- Added a test helper that deletes selected persisted `segment-data` records from the offline navigation IndexedDB database.
- Added an e2e for the missing-segment case:
  - the source dynamic route record is persisted
  - the target exact URL record is intentionally absent
  - the target route record is intentionally absent
  - the target head and segment records are prefetched
  - the target page segment record is deleted
  - the server is stopped and the browser is offline
  - the hard load reports `router-cache-reconstruction-miss` with `missing-segment`
  - the final document shows the normal visible offline cache miss UI

## Not included

This PR does not add new runtime behavior. It also does not cover the later production-hardening matrix for middleware, rewrites, redirects, interception, optional/catch-all routes, app/session invalidation, or storage quotas. Those belong in the next stack because they need their own fixtures and policy decisions.

## Reviewer focus

- The test should fail if dynamic replay accidentally renders with incomplete router-cache data.
- The destructive IndexedDB helper should delete only the intended page segment record, leaving the route table and other target records in place.
- The assertions should prove both internal diagnostics and the user-visible miss path.

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses dynamic route pattern replay when a required segment record is missing"`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `pnpm --filter=next build`
- `pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses dynamic route pattern replay when a required segment record is missing"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses dynamic route pattern replay when a required segment record is missing"`
- `git diff --check`

### DIFF
--- a/packages/next/errors.json
+++ b/packages/next/errors.json
@@ -1190,5 +1190,6 @@
   "1189": "Route \"%s\" accessed header \"%s\" which is not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. Add it to the sample's \\`headers\\` array, or \\`[\"%s\", null]\\` if it should be absent.",
   "1190": "Route \"%s\" accessed param \"%s\" which is not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. Add it to the sample's \\`params\\` object.",
   "1191": "Route \"%s\" called %s but param%s %s %s not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. %s requires all route params to be provided.",
-  "1192": "Route \"%s\" accessed root param \"%s\" which is not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. Add it to the sample's \\`params\\` object."
+  "1192": "Route \"%s\" accessed root param \"%s\" which is not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. Add it to the sample's \\`params\\` object.",
+  "1193": "\\`experimental.offlineNavigations\\` requires \\`cacheComponents\\` to be enabled. Please update your %s accordingly."
 }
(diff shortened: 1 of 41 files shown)


feedthejim:

PR #93657, opened 2026-05-08
(1/1) Cover offline navigation service worker pass-through stress

feedthejim:
## Stack position

This is PR 1/1 in a standalone `offlineNavigations` service-worker stress coverage slice.

It sits after #93656, which added the app-owned cache reset API. This PR does not change runtime behavior. It tightens the production e2e contract around an important architectural invariant: the generated service worker owns document survival only, while RSC, route handlers, server actions, and static asset fetches continue to pass through to the network.

## Why this exists

The offline navigation architecture intentionally keeps route semantics out of the service worker. The worker may serve the generated fallback document for same-origin document navigation network failures, but it must not become a generic offline cache or router. That boundary is easy to regress later when adding output-export artifacts or more Cache Storage behavior.

This test makes that boundary visible in browser coverage instead of relying only on the service worker source predicate.

## What changed

- Expanded the existing exact-URL/pass-through production stress e2e.
- The test now registers a fallback-served message collector immediately before non-document offline fetches.
- While offline, it verifies these requests reject instead of receiving the fallback document:
  - RSC fetch
  - API route `GET`
  - API route `POST`
  - static offline navigation manifest fetch
- It also verifies none of those non-document requests emit the `fallback-served` service-worker message.

## Not included

- No service worker runtime changes.
- No output-export bridge behavior.
- No route/segment replay changes.
- No lifecycle upgrade/multi-tab coverage, which remains a separate production-hardening slice.

## Reviewer focus

- The important behavior is negative coverage: non-document fetches should fail normally under browser offline mode, not be answered by the offline navigation fallback.
- The positive document-fallback path remains covered earlier in the same stress test before these pass-through assertions run.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "covers exact URL shape and pass-through stress cases"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "covers exact URL shape and pass-through stress cases"`

### DIFF
--- a/packages/next/errors.json
+++ b/packages/next/errors.json
@@ -1190,5 +1190,6 @@
   "1189": "Route \"%s\" accessed header \"%s\" which is not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. Add it to the sample's \\`headers\\` array, or \\`[\"%s\", null]\\` if it should be absent.",
   "1190": "Route \"%s\" accessed param \"%s\" which is not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. Add it to the sample's \\`params\\` object.",
   "1191": "Route \"%s\" called %s but param%s %s %s not defined in the \\`unstable_samples\\` of \\`unstable_instant\\`. %s requires all route params to be provided.",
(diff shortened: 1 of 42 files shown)


feedthejim:

PR #93658, opened 2026-05-08
(1/1) Cover offline navigation IDB unavailable fallback

feedthejim:
## Stack position

This is PR 1/1 in a standalone `offlineNavigations` storage-failure stress coverage slice.

It sits after #93657, which tightened the service-worker pass-through boundary. This PR is test-only and covers the next production-hardening invariant: if durable browser storage is unavailable during fallback boot, the generated fallback document must settle to visible cache-miss UI instead of hanging or crashing.

## Why this exists

The offline navigation design treats IndexedDB as the client-owned durable router cache. Real browsers can still make IDB unavailable because of private-mode-like restrictions, storage failures, user settings, or transient platform failures. The feature should degrade to a deterministic miss, because guessing or blanking during an offline document reload is worse than showing the app/user that the route is unavailable offline.

## What changed

- Added a production e2e that disables `window.indexedDB` with an init script before app code runs.
- Verifies the service worker still installs and can serve the generated fallback document.
- Stops the app server, switches the browser offline, and hard-loads a route with no available durable storage.
- Asserts the fallback document renders visible cache-miss UI with `missing-entry` diagnostics.

## Not included

- No runtime changes.
- No quota-exceeded simulation.
- No IDB blocked/versionchange stress.
- No migration failure coverage.

Those remain separate storage hardening slices.

## Reviewer focus

- The key assertion is that an unavailable IDB path produces a named visible miss, not a hidden fallback shell or an uncaught boot error.
- The test intentionally disables IDB before both the online setup page and the offline fallback document so it exercises the same browser condition throughout.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "shows cache miss when IndexedDB is unavailable during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "shows cache miss when IndexedDB is unavailable during fallback boot"`

Note: the focused Turbopack run reported a stale package-build warning, but this PR changes only the e2e test file. The packed webpack run also passed through an isolated package install.

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -2954,6 +2954,94 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('shows cache miss when IndexedDB is unavailable during fallback boot', async () => {
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        async beforePageLoad(p: Playwright.Page) {
+          page = p
+          await p.addInitScript(() => {
+            Object.defineProperty(window, 'indexedDB', {
+              configurable: true,
+              value: undefined,
+            })
+          })
+        },
+      })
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93650, opened 2026-05-08
offline navigations: cover persisted router replay (10/11)

feedthejim:
## Context

This is PR 10/11 in the offline navigations stack. The stack builds the initial experimental feature in two chapters: build and service worker artifacts first, then client router persistence and fallback bootstrap. The service worker owns bootstrap delivery only; route data stays in the Cached Navigations router cache and is persisted by the client when the flag is enabled. PR 11 adds the docs and examples layer.

Review guide: https://gist.github.com/feedthejim/007cb3a3c0dd39b3806213c4ce95d31a

## What Changed

Adds the hardening rows for the router-cache based offline navigation design. These assertions make incomplete replay state explicit: when the router cache does not contain the records needed for the URL, the app shows an offline miss.

## Reviewer Focus

Review the request-sensitive and URL-shape assertions. These should visibly miss offline rather than replaying guessed or partial UI.

## What Does Not Work Yet

Runtime replay behavior for the initial experimental scope is complete after this PR. PR 11 adds docs and examples. Static export integration, dev simulation, custom miss UI, and broader production hardening remain follow-up work.

## Proof In This PR

Adds production e2e coverage for request-sensitive hard-load misses and encoded/search/hash URL shape stress misses.

## Deferred Coverage

Docs land in PR 11. Broader production hardening can be follow-up work once this stack lands.

## Docs Status

Docs ship in #93738 (11/11).

## Stack Map

1. #93736 `offline navigations: add gated build primitives (1/11)` (sibling)
2. #93737 `offline navigations: emit fallback manifest data (2/11)` (sibling)
3. #93625 `offline navigations: register pass-through worker (3/11)` (sibling)
4. #93626 `offline navigations: cache fallback artifacts (4/11)` (sibling)
5. #93627 `offline navigations: serve fallback document offline (5/11)` (sibling)
6. #93630 `offline navigations: add router-cache persistence primitives (6/11)` (sibling)
7. #93640 `offline navigations: persist cached router records (7/11)` (sibling)
8. #93644 `offline navigations: bootstrap fallback from router records (8/11)` (sibling)
9. #93647 `offline navigations: support dynamic route patterns (9/11)` (sibling)
10. **#93650 `offline navigations: cover persisted router replay (10/11)` (current)**
11. #93738 `offline navigations: add docs and examples (11/11)` (sibling)

PR #93659, opened 2026-05-08
(1/1) Cover offline navigation build namespace misses

feedthejim:
## Stack position

This is PR 1/1 in a standalone `offlineNavigations` build-namespace stress coverage slice.

It sits after #93658, which covered the IDB-unavailable fallback boot path. This PR is test-only and covers another production-hardening invariant: persisted data from a previous build/deployment namespace must not be replayed into the current fallback boot.

## Why this exists

Offline navigation data is durable browser storage. A user can have records from an older build sitting in IndexedDB when a newer build is loaded. The fallback document must scope lookups to the current build/deployment ID and miss visibly if only stale-build data exists for the requested URL.

That matters for correctness, but it also keeps the architecture aligned with the cache-components router cache: durable replay is a mirror of the current router cache namespace, not a cross-build best-effort cache.

## What changed

- Added a production e2e that first waits for the current build to persist a valid exact-URL initial-load entry.
- Copies that valid payload to a different URL under a fake stale build ID.
- Stops the app server, switches the browser offline, and hard-loads the stale-build URL.
- Asserts fallback boot uses the current build namespace, ignores the stale record, and renders visible `missing-entry` cache-miss UI with structured diagnostics.

## Not included

- No runtime changes.
- No physical old-build cleanup changes. #93650 covers build-scoped cleanup at the storage layer.
- No deployment-ID skew/multi-tab upgrade test yet.
- No route/segment stale-build record matrix yet.

## Reviewer focus

- The copied payload is intentionally valid. The miss should happen because of the build namespace, not because the payload is malformed.
- The diagnostic should report the current build/deployment ID, proving the fallback boot is scoped to the active build.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "ignores persisted offline navigation data from another build"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "ignores persisted offline navigation data from another build"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -3042,6 +3042,164 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('ignores persisted offline navigation data from another build', async () => {
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93660, opened 2026-05-08
(1/1) Cover expired exact URL offline navigation misses

feedthejim:
## Stack position

This is PR 1/1 in a standalone `offlineNavigations` exact-URL freshness stress coverage slice.

It sits after #93659, which covered current-build namespace isolation. This PR is test-only and covers the next fallback-boot invariant: even when an exact-URL payload is otherwise valid and belongs to the current build, it must not replay after its expiration time.

## Why this exists

Offline navigation data is durable browser storage, but it still needs to follow the same freshness model as the cache-components router cache. Expired data should miss visibly and be cleaned up, rather than becoming a stale offline-only route replay.

This is especially important before adding output-export and broader production hardening, because the fallback document should treat persistent storage as a scoped, freshness-checked router cache mirror, not as an eternal archive.

## What changed

- Added a production e2e that waits for a real exact-URL initial-load entry to be persisted.
- Copies that valid current-build RSC payload to a new URL with an expired `staleAt`/`expiresAt` timestamp.
- Stops the app server, switches the browser offline, and hard-loads the expired URL through the fallback document.
- Asserts fallback boot renders visible `missing-entry` cache-miss UI with structured diagnostics.
- Asserts the expired exact-URL entry is deleted from IndexedDB after the miss.

## Not included

- No runtime changes.
- No route/segment expired-record matrix yet.
- No stale-but-not-expired policy change.
- No quota or LRU behavior.

## Reviewer focus

- The copied payload is intentionally valid and uses the current build ID. The miss should happen because the entry is expired.
- The final IndexedDB assertion is the cleanup check: expired exact-URL entries should be removed during fallback boot.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted exact-URL data during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted exact-URL data during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -3200,6 +3200,178 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses and deletes expired persisted exact-URL data during fallback boot', async () => {
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93661, opened 2026-05-08
(1/1) Cover expired route record offline navigation misses

feedthejim:
## Why

This is a small test-only stress slice in the `offlineNavigations` stack. It sits directly after the expired exact-URL stress PR and covers the next P0 freshness case from the advanced e2e matrix: expired persisted route records must not be replayed during fallback boot.

The production contract here is that the service worker only gets the app to the fallback document. Once the app boots, the cache-components client router owns the IndexedDB route records and must reject stale or expired route data before rendering. A visible miss is the correct result when the route record is expired, even if the matching segment and head records are still present.

## What changed

- Adds a test helper that expires selected persisted `route-data` records in IndexedDB.
- Adds an e2e that prefetches `/docs/prefetched`, deletes the exact-URL entry, expires the route record, stops the server, and hard-loads the URL offline.
- Asserts the fallback boot emits `router-cache-reconstruction-miss` with `missing-route`, reaches visible cache-miss UI, and deletes the expired route record from IndexedDB.

## Stack position

Base: `feedthejim/offline-navigations-expired-exact-url-stress`

This PR is intentionally narrow and test-only. It does not add new route persistence behavior. It verifies the route-record expiration behavior that already exists in the client router cache read path, using a full browser fallback boot so we know the app cannot accidentally replay expired durable route data.

- Expired segment and head records
- Stale-but-not-expired route policy boundaries
- Route epoch, schema, build, deployment, basePath, assetPrefix, and app/session scope mismatch matrix

Those are separate stress slices so reviewers can inspect one failure mode at a time.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted route records during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted route records during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -319,6 +319,63 @@ describe('offlineNavigations build artifacts', () => {
     }, urlSubstring)
   }
 
+  async function expirePersistedOfflineNavigationRouteRecords(
+    browser: Awaited<ReturnType<typeof next.browser>>,
+    pathnameSubstring: string
+  ) {
+    return browser.eval(async (substring) => {
+      const database = await new Promise<IDBDatabase>((resolve, reject) => {
+        const request = indexedDB.open('next-offline-navigation-cache', 3)
+        request.onsuccess = () => resolve(request.result)
+        request.onerror = () => reject(request.error)
+      })
+
+      try {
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93662, opened 2026-05-08
(1/1) Cover expired segment record offline navigation misses

feedthejim:
## Why

This is the next narrow stress slice in the `offlineNavigations` stack, directly on top of the expired route-record PR. It covers the page-segment half of `GRAPH-04b`: a route must not reconstruct offline from an expired persisted segment record, even when the route record and head record are still present.

The important contract is still the same: the service worker only serves the fallback document. The cache-components client router owns IndexedDB route and segment freshness. If a required segment is expired, fallback boot must miss visibly instead of rendering partial or stale UI.

## What changed

- Adds a test helper that expires selected persisted `segment-data` records in IndexedDB.
- Adds an e2e that prefetches `/docs/prefetched`, deletes the exact-URL entry, expires the prefetched page segment, stops the server, and hard-loads the route offline.
- Asserts fallback boot emits `router-cache-reconstruction-miss` with `missing-segment`, reaches visible cache-miss UI, and deletes the expired page segment from IndexedDB.

## Stack position

Base: `feedthejim/offline-navigations-expired-route-record-stress`

This PR is intentionally test-only. The prior PR covers expired route records with `missing-route`; this PR covers expired page segment records with `missing-segment`. Keeping those separate should make the freshness behavior easier to review.

- Expired head records, which should exercise the `missing-head` path
- Expired parent layout, default slot, and parallel route segment records
- Stale-but-not-expired freshness policy boundaries
- Segment schema, build, deployment, route epoch, and app/session scope mismatch matrices

Those stay as separate stress slices so the route graph completeness matrix remains reviewable.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted segment records during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted segment records during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -376,6 +376,70 @@ describe('offlineNavigations build artifacts', () => {
     }, pathnameSubstring)
   }
 
+  async function expirePersistedOfflineNavigationSegmentRecords(
+    browser: Awaited<ReturnType<typeof next.browser>>,
+    options: {
+      keySubstring: string
+      requestKeySuffix: string
+    }
+  ) {
+    return browser.eval(async ({ keySubstring, requestKeySuffix }) => {
+      const database = await new Promise<IDBDatabase>((resolve, reject) => {
+        const request = indexedDB.open('next-offline-navigation-cache', 3)
+        request.onsuccess = () => resolve(request.result)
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93663, opened 2026-05-08
(1/1) Cover expired head record offline navigation misses

feedthejim:
## Why

This is a focused follow-up to the expired route-record and page-segment stress PRs. It covers the separate head reconstruction branch: a route can have valid route and page segment records but still must miss if the persisted head record is expired.

That distinction matters because fallback boot reconstructs the cache-components route in two phases: route/segment seed data first, then head data. We want e2e coverage that the `missing-head` reason is observable and does not collapse into an unrelated cache miss.

## What changed

- Reuses the segment-record expiration helper from the previous PR.
- Adds an e2e that prefetches `/docs/prefetched`, deletes the exact-URL entry, expires persisted `/_head` records, stops the server, and hard-loads the route offline.
- Asserts fallback boot emits `router-cache-reconstruction-miss` with `missing-head`, reaches visible cache-miss UI, and deletes the expired head record from IndexedDB.

## Stack position

Base: `feedthejim/offline-navigations-expired-segment-record-stress`

This PR is test-only. The prior slices cover expired exact URL, expired route, and expired page segment records. This one completes the current route/head freshness trio by covering the distinct `missing-head` reconstruction path.

- Expired parent layout, default slot, and parallel route segment records
- Stale-but-not-expired freshness policy boundaries
- Segment schema, build, deployment, route epoch, and app/session scope mismatch matrices

Those stay as separate stress slices so each PR keeps a single reviewable claim.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted head records during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes expired persisted head records during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -1682,6 +1682,172 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses and deletes expired persisted head records during fallback boot', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93664, opened 2026-05-08
(1/1) Cover missing exact URL body offline navigation misses

feedthejim:
## Why

This is a narrow `EXACT-05a` payload-damage stress slice in the `offlineNavigations` stack. Earlier coverage already checks a broad malformed payload and unsupported request kind inside the service-worker stress test, but this PR isolates one concrete exact-URL payload-capability failure: the persisted RSC response shape is otherwise fresh and current-build, but the response body is missing.

The client fallback bootstrap should reject that payload as `invalid-payload`, show visible cache-miss UI, and avoid trying to render stale or partial route data.

## What changed

- Adds an e2e that starts from a valid initial-load exact-URL entry, copies it to a new URL, and changes only `payload.body` to `null`.
- Stops the server, hard-loads the damaged URL offline, and asserts the fallback document boots to visible cache-miss UI.
- Asserts the public diagnostic reason is `invalid-payload` for the requested URL and current build/deployment namespace.

## Stack position

Base: `feedthejim/offline-navigations-expired-head-record-stress`

This PR is test-only. It follows the route/segment/head freshness stress slices and starts the exact-URL payload damage queue with one reviewable failure mode.

- Truncated but non-null RSC framing
- Exact-URL schema mismatch
- Deployment ID mismatch
- Unsupported exact-URL response shapes beyond the existing broad malformed/unsupported coverage
- Output export artifact corruption

Those stay as separate stress slices so each PR owns one concrete claim.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL data with a missing RSC response body during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL data with a missing RSC response body during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -3815,6 +3815,173 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses exact-URL data with a missing RSC response body during fallback boot', async () => {
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93665, opened 2026-05-08
(1/1) Cover incompatible exact URL payload version offline misses

feedthejim:
## Stack position

This is a one-PR stress-test slice stacked on top of #93664, `(1/1) Cover missing exact URL body offline navigation misses`.

The surrounding offline navigation stack is proving the cache-components-only architecture where the service worker only keeps document navigations alive, while the client runtime owns IndexedDB replay decisions. This PR does not add a new feature path. It adds another destructive exact-URL payload compatibility case for the fallback boot contract introduced earlier in the stack.

## What this covers

When the fallback document boots offline, it may find an exact-URL IndexedDB entry for the requested URL. That entry is only replayable if the stored RSC response payload still matches the payload schema the current runtime knows how to restore.

This PR copies a valid initial-load exact-URL RSC payload, changes only the RSC response payload `version`, stops the server, goes offline, and hard-loads the damaged URL. The expected behavior is a visible cache miss with an `invalid-payload` diagnostic, not an attempted replay from an incompatible schema and not a silent blank page.

## What works after this PR

- Exact-URL fallback boot has e2e coverage for an incompatible RSC payload version.
- The test proves the service worker still serves the generated fallback document while the app server is unavailable.
- The client rejects the damaged payload with the same visible cache-miss path used for other invalid exact-URL payloads.
- Diagnostics include the current build ID, requested URL, and `invalid-payload` reason.

- This does not cover truncated RSC framing, unsupported response shapes, or deployment-ID mismatch. Those remain separate payload-damage stress cases.
- This does not add route/segment schema mismatch coverage. Route and segment records have separate stores and should stay in their own stress slices.
- This does not change any runtime behavior, storage schema, or service-worker logic.

## Reviewer focus

Please focus on whether this test protects the right production contract: fallback boot must not replay exact-URL data whose payload schema is incompatible with the current client runtime.

The PR is intentionally test-only so it can be reviewed independently from implementation changes.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL data with an incompatible RSC response payload version during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL data with an incompatible RSC response payload version during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -3982,6 +3982,174 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses exact-URL data with an incompatible RSC response payload version during fallback boot', async () => {
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
+      await waitForOfflineNavigationServiceWorker(browser, page!)
+
+      await retry(async () => {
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93666, opened 2026-05-08
(1/1) Cover incompatible exact URL entry version offline misses

feedthejim:
## Stack position

This is a one-PR stress-test slice stacked on top of #93665, `(1/1) Cover incompatible exact URL payload version offline misses`.

The previous PR covers the inner RSC response payload schema. This PR covers the outer exact-URL IndexedDB entry schema. Keeping them separate makes the review question narrow: does fallback boot delete an incompatible durable entry before it can be treated as replayable data?

## What this covers

The exact-URL store has an entry-level schema version in addition to the stored RSC response payload shape. If that outer entry version does not match the current runtime, the client should treat the entry as incompatible, delete it, and render the normal offline cache-miss UI.

This test copies a valid initial-load exact-URL entry, changes only the outer entry `version`, stops the app server, goes offline, and hard-loads the damaged URL. The expected result is a visible `missing-entry` miss with current-build diagnostics, followed by deletion of the incompatible IndexedDB record.

## What works after this PR

- Fallback boot has e2e coverage for incompatible exact-URL entry versions.
- The incompatible record is deleted during the read path.
- The app shows deterministic cache-miss UI instead of replaying stale or schema-incompatible data.
- The diagnostic records the current build ID, requested URL, and `missing-entry` reason.

- This does not cover route or segment record schema mismatch. Those stores have separate replay paths and should be tested in route/segment stress slices.
- This does not cover truncated RSC body framing or deployment-ID mismatch.
- This does not change production behavior, storage schema, or service-worker behavior.

## Reviewer focus

Please focus on the entry-level compatibility contract: an exact-URL record whose outer schema is incompatible must not survive fallback boot or produce a replay hit.

The PR is intentionally test-only and stacked directly after the payload-version test so the two schema layers are easy to review separately.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes exact-URL data with an incompatible entry version during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses and deletes exact-URL data with an incompatible entry version during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -4150,6 +4150,183 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses and deletes exact-URL data with an incompatible entry version during fallback boot', async () => {
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
+      await waitForOfflineNavigationServiceWorker(browser, page!)
+
+      await retry(async () => {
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93667, opened 2026-05-08
(1/1) Handle malformed exact URL RSC payload offline misses

feedthejim:
## Stack position

This is a one-PR payload-damage slice stacked on top of #93666, `(1/1) Cover incompatible exact URL entry version offline misses`.

The previous two slices cover schema compatibility before React tries to consume the cached response. This PR covers the next failure boundary: a valid-shaped exact-URL entry whose RSC response bytes are malformed.

## What this changes

Fallback boot already validates that an exact-URL payload has the expected RSC response shape. A corrupt body can still pass that shape check because the body is an `ArrayBuffer`, then fail when React parses the Flight response.

This PR catches that parse failure in the fallback bootstrap path, switches the document to visible `invalid-payload` cache-miss UI, and returns the same never-resolving placeholder payload used by other cache-miss states. This avoids treating a malformed cached response as a successful replay and avoids leaving the generated fallback document blank or stuck without a reason.

## What this covers

The e2e copies a valid initial-load exact-URL RSC payload, replaces the body with a malformed complete Flight row, stops the server, goes offline, and hard-loads the damaged URL. The expected result is a visible `invalid-payload` miss with the current build ID and requested URL in diagnostics.

## What works after this PR

- Valid-shaped but malformed exact-URL RSC response bodies miss visibly during fallback boot.
- The miss reason is `invalid-payload`, consistent with other payload compatibility failures.
- The service worker still only serves the fallback document; the client bootstrap owns the payload validation and parse failure behavior.

- Incomplete partial streams that React treats as allowed partial data are not reclassified here. This PR is specifically about corrupt complete RSC rows that reject during parsing.
- Deployment-ID mismatch and route/segment malformed payloads remain separate stress slices.
- This does not change the IndexedDB schema or service-worker behavior.

## Reviewer focus

Please focus on the small bootstrap change in `app-index.tsx`: the catch is scoped to offline fallback exact-URL RSC parsing, and it falls back to the existing cache-miss mechanism rather than inventing a new render path.

## Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write packages/next/src/client/app-index.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `npx eslint --config eslint.config.mjs --fix packages/next/src/client/app-index.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `pnpm --filter=next build`
- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL data with a malformed RSC response body during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL data with a malformed RSC response body during fallback boot"`

### DIFF
--- a/packages/next/src/client/app-index.tsx
+++ b/packages/next/src/client/app-index.tsx
@@ -543,8 +543,9 @@ if (instantTestStaticFetch) {
       }
 
       const fallbackResponse = bootstrap.response
-      const fallbackInitialRSCPayload =
-        await createFromFetch<InitialRSCPayload>(
+      let fallbackInitialRSCPayload: InitialRSCPayload
+      try {
+        fallbackInitialRSCPayload = await createFromFetch<InitialRSCPayload>(
           Promise.resolve(fallbackResponse.response),
           {
             callServer,
@@ -553,6 +554,10 @@ if (instantTestStaticFetch) {
             unstable_allowPartialStream: true,
           }
         )
+      } catch {
+        showOfflineNavigationCacheMiss('invalid-payload', bootstrap.buildId)
+        return await neverResolveInitialRSCPayload()
+      }
 
       if (fallbackResponse.requestKind === 'initial-load') {
         return fallbackInitialRSCPayload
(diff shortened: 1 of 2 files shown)


feedthejim:

PR #93668, opened 2026-05-08
(1/1) Cover exact URL query identity offline misses

feedthejim:
## Stack position

This is a one-PR URL-identity stress slice stacked on top of #93667, `(1/1) Handle malformed exact URL RSC payload offline misses`.

The previous payload-damage slices prove fallback boot rejects incompatible or malformed exact-URL data. This PR moves to the next exact-URL correctness contract: nearby query strings must not collide and replay the wrong cached response.

## What this covers

Exact-URL offline replay is intentionally keyed by the normalized request URL, with the hash excluded but the query string preserved. This matters for duplicate params and encoding differences where two URLs may look similar but produce different app data.

The e2e first loads and persists:

`/docs/url-stress/space%20value/?token=a%2Bb&tag=one&tag=two`

Then it stops the server, goes offline, and hard-loads two nearby variants:

- `token=a+b`, which is not the same URL as `token=a%2Bb`
- reversed duplicate query param order, `tag=two&tag=one`

Both variants must miss visibly with `missing-entry` and must not render the cached route data from the original URL.

## What works after this PR

- Exact-URL replay cannot accidentally collapse `+` and `%2B` query encodings.
- Exact-URL replay cannot accidentally collapse duplicate query params with different order.
- Miss diagnostics include the current build ID, requested URL, and `missing-entry` reason.
- The test asserts the app route content is absent on miss, so fallback UI cannot be mistaken for a replay hit.

- This does not cover every URL identity case in the plan. Base path, asset prefix, hash behavior, and basic reordered query coverage already exist elsewhere; encoded path variants, canonical redirects, trailing slash matrix, and deployment URL changes still need their own slices if they are not already covered by the owning implementation PRs.
- This does not change runtime behavior or storage schema.

## Reviewer focus

Please focus on whether these two collision variants protect the exact-URL cache key contract without expanding the implementation surface. This PR is test-only.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL replay for query identity collision variants"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL replay for query identity collision variants"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -3569,6 +3569,146 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses exact-URL replay for query identity collision variants', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93669, opened 2026-05-08
(1/1) Cover stale exact URL offline replay

feedthejim:
## Stack position

This is a one-PR freshness-policy stress slice stacked on top of #93668, `(1/1) Cover exact URL query identity offline misses`.

The previous slices cover exact-URL payload damage and URL identity misses. This PR covers the positive side of freshness: stale-but-not-expired exact-URL data should still replay offline in the current v1 policy, while hard-expired entries are already covered by #93660.

## What this covers

The test starts from a valid initial-load exact-URL RSC payload, copies it to a never-loaded URL, mutates `staleAt` into the past while keeping `expiresAt` in the future, then navigates away online. After stopping the server and going offline, it hard-loads the never-loaded URL.

The expected behavior is a real fallback-document boot that renders from the stale-but-not-expired exact-URL data, reports `useOffline()` as offline, and emits a `cache-hit` diagnostic for the requested URL.

## What works after this PR

- Exact-URL replay has e2e coverage for stale-but-not-expired records.
- The test proves the URL was never loaded as an HTML document, so the result cannot be browser HTTP cache reuse.
- The paired hard-expired exact-URL miss remains covered in #93660.

- This does not define every future stale policy. It documents the current v1 behavior: replay while offline until hard expiry.
- Route, segment, and head stale-but-not-expired behavior still need their own cache-components route-cache slices.
- Clock skew and controlled-clock boundary tests remain separate `CACHE-01`/`EXACT-09` follow-ups.

## Reviewer focus

Please focus on whether this test captures the intended v1 offline freshness contract without changing runtime behavior: stale-but-not-expired exact-URL data is allowed to replay in browser-private storage, hard-expired data is not.

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "replays stale-but-not-expired exact-URL data during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "replays stale-but-not-expired exact-URL data during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -4638,6 +4638,168 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('replays stale-but-not-expired exact-URL data during fallback boot', async () => {
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93670, opened 2026-05-08
(1/1) Cover missing route record offline navigation misses

feedthejim:
## Stack Position

This is a focused stress-test slice stacked on top of `feedthejim/offline-navigations-stale-exact-url-stress`.

It belongs to the route-graph replay hardening work in the offlineNavigations stack. Earlier PRs added the cache-components-only offline navigation foundation, exact-URL fallback boot, persisted route/segment replay, invalidation coverage, and exact-URL payload/freshness stress tests. This PR covers one destructive `GRAPH-02b` route-graph completeness case: the requested route has segment and head data, but the required persisted route record is missing.

## Why This Exists

Route/segment replay must not accidentally render from incomplete router state. It also must not pass through the simpler exact-URL RSC replay path when we are trying to prove route-graph behavior.

This test makes that contract explicit:

- prefetch `/docs/prefetched` so route, segment, and head records are present
- delete the persisted route record for `/prefetched`
- keep the segment and head records present
- assert a visible cache miss and structured `missing-route` diagnostics

## Reviewer Focus

Please focus on whether the test proves the intended missing-route-record contract without relying on exact-URL replay or the live server.

The helper only deletes route-data records whose persisted route pathname matches the target substring. It intentionally leaves segment-data records alone so this negative case is specifically about the missing route record.

This does not finish the whole `GRAPH-02b` matrix. Remaining route-graph missing-record rows still need their own focused slices:

- missing root layout record
- missing page segment record outside the existing dynamic-pattern case
- missing head/metadata record in the non-expiry path
- missing default slot record

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses persisted router records when a required route record is missing during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses persisted router records when a required route record is missing during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -319,6 +319,56 @@ describe('offlineNavigations build artifacts', () => {
     }, urlSubstring)
   }
 
+  async function deletePersistedOfflineNavigationRouteRecords(
+    browser: Awaited<ReturnType<typeof next.browser>>,
+    pathnameSubstring: string
+  ) {
+    return browser.eval(async (substring) => {
+      const database = await new Promise<IDBDatabase>((resolve, reject) => {
+        const request = indexedDB.open('next-offline-navigation-cache', 3)
+        request.onsuccess = () => resolve(request.result)
+        request.onerror = () => reject(request.error)
+      })
+
+      try {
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93671, opened 2026-05-08
(1/1) Cover missing head record offline navigation misses

feedthejim:
## Stack Position

This is a focused stress-test slice stacked on top of `feedthejim/offline-navigations-missing-route-record-stress`.

It continues the `GRAPH-02b` route-graph completeness matrix for offlineNavigations. The previous slice covers a deleted persisted route record. This slice covers a deleted persisted head record while route and page segment records are still present.

## Why This Exists

Route/segment replay needs complete route graph data. A page segment by itself is not enough if the matching head record is missing, and this test makes that behavior explicit without relying on exact-URL replay.

The test flow is intentionally destructive:

- prefetch `/docs/prefetched` so route, page segment, and head records exist
- delete the persisted `/_head` record
- verify the route and page segment records are still present
- assert a visible cache miss and structured `missing-head` diagnostics

## Reviewer Focus

Please focus on whether this proves the missing-head-record contract distinctly from the existing expired-head-record coverage.

The existing expired-head test proves cleanup of stale head data. This PR proves the absent-record path: the client router must reject an otherwise present route/segment replay set if the required head record was never available.

This does not finish the whole `GRAPH-02b` matrix. Remaining route-graph missing-record rows still need focused slices:

- missing root layout record
- missing page segment record outside the existing dynamic-pattern case
- missing default slot record

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses persisted router records when a required head record is missing during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses persisted router records when a required head record is missing during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -1722,6 +1722,164 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses persisted router records when a required head record is missing during fallback boot', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93672, opened 2026-05-08
(1/1) Cover missing page segment offline navigation misses

feedthejim:
## Stack Position

This is a focused stress-test slice stacked on top of `feedthejim/offline-navigations-missing-head-record-stress`.

It continues the `GRAPH-02b` route-graph completeness matrix for offlineNavigations. The previous two slices cover deleted persisted route and head records. This slice covers the regular prefetched-route page segment record.

## Why This Exists

The stack already has a dynamic-pattern missing segment test, but regular route/segment replay should also have its own direct destructive case. A persisted route record and head record are not enough if the page segment body is missing.

The test flow is intentionally destructive:

- prefetch `/docs/prefetched` so route, page segment, and head records exist
- delete only the persisted page segment record for `/prefetched`
- verify the route and head records are still present
- assert a visible cache miss and structured `missing-segment` diagnostics

## Reviewer Focus

Please focus on whether this proves the missing page-segment contract on the regular route/segment replay path, without falling back to exact-URL replay or live server data.

This is deliberately separate from the expired-segment test. Expiry coverage proves stale data cleanup. This PR proves absent data cannot be treated as replayable.

This does not finish the whole `GRAPH-02b` matrix. Remaining route-graph missing-record rows still need focused slices:

- missing root layout record
- missing default slot record

## Verification

- `git diff --check`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses persisted router records when a required page segment record is missing during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses persisted router records when a required page segment record is missing during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -1880,6 +1880,164 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses persisted router records when a required page segment record is missing during fallback boot', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93673, opened 2026-05-08
(1/1) Cover workspace shell offline navigation replay

feedthejim:
## Stack Position

This is a focused stress-test slice stacked on top of `feedthejim/offline-navigations-missing-page-segment-stress`.

It continues the route-graph hardening work for offlineNavigations. The previous slices cover missing route, head, and regular page segment records. This PR adds a denser workspace-shell fixture and proves the positive replay path before we start deleting nested layout/default/parallel-slot records in follow-up PRs.

## Why This Exists

The existing fixture only had a simple root layout and page routes. That was enough for exact-URL, route-record, head, and page-segment coverage, but not enough to honestly test complex app-shell reconstruction.

This PR adds a small cache-components-compatible route shape:

- `/workspace/[workspace]/layout.tsx`
- a default `@sidebar` slot
- an `@activity` parallel slot for the thread route
- `/workspace/[workspace]/channel/[channel]/thread/[thread]/page.tsx`
- static params for the concrete `/workspace/acme/channel/general/thread/123` test route

The e2e then proves:

- the route can be prefetched online
- exact-URL data for the document URL is deleted before reload
- route, segment, and head records remain available
- the server is stopped and the browser is offline
- fallback boot renders from `router-cache`
- the workspace layout, page segment, default sidebar slot, and activity parallel slot all render together

## Reviewer Focus

Please focus on whether this is a good minimal fixture for later destructive route-graph tests.

The point of this PR is not to finish every `SHELL-01` negative case. It establishes the positive replay contract for a realistic nested app shell so follow-up PRs can delete one required layout or slot record at a time and prove the miss reason.

## What This Does Not Cover Yet

Remaining follow-ups still need focused destructive coverage for:

- missing root shell/layout record
- missing nested workspace layout record
- missing default sidebar slot record
- missing activity parallel slot record
- metadata/head/style ordering for the workspace shell
- app/session scope rotation on a workspace-like route

## Verification

[...]

PR #93674, opened 2026-05-08
(1/1) Cover missing default slot offline navigation misses

feedthejim:
## Stack position

This is a test-only stress slice on top of #93673, which introduced the Slack-like workspace shell fixture and the positive route/segment replay coverage for a nested app shell.

Stack context:

1. #93670 covers a missing persisted route record.
2. #93671 covers a missing persisted head record.
3. #93672 covers a missing persisted page segment record.
4. #93673 proves the complex workspace shell can replay when all required route, segment, slot, and head records are present.
5. This PR covers the paired negative case for the default sidebar slot inside that shell.

## What changed

Adds one focused production e2e for `offlineNavigations` with Cache Components:

- online-loads `/docs` and waits for the offline navigation service worker
- prefetches `/docs/workspace/acme/channel/general/thread/123`
- verifies the durable route record, page segment, activity parallel slot, default sidebar slot, and head record were persisted
- deletes the exact URL entry so the hard reload must use route/segment reconstruction
- deletes only the default sidebar slot segment record
- stops the server, puts the browser offline, and hard-loads the workspace route
- asserts the fallback boot emits `router-cache-reconstruction-miss` with `missing-segment`, falls back to visible `missing-entry` UI, and does not render the workspace page

## Why this is separate

The positive workspace-shell PR proves the intended app-shape can replay. This PR keeps the destructive completeness case isolated so reviewers can focus on one question: if a required default slot record is missing, do we safely miss instead of rendering a partial shell around stale or absent slot content?

This also advances the advanced e2e stress matrix in `.private-investigation/offline-navigations-production-plan.md`: route, head, page segment, and default slot deletions now each have their own focused browser-backed negative test.

## Reviewer focus

Please focus on whether the test proves the intended failure mode without passing through an unrelated exact-URL miss. The important guards are that the route, page segment, activity slot, and head records remain present after deletion, while the default sidebar slot is the only missing required segment.

## Intentionally not covered here

Still pending in later stress slices:

- missing root layout record
- missing nested workspace layout record
- missing activity parallel slot record
- metadata/head style ordering integrity on successful replay
- app/session or workspace-scope partitioning

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses a workspace shell route when a required default slot record is missing during fallback boot"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses a workspace shell route when a required default slot record is missing during fallback boot"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -1530,6 +1530,192 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses a workspace shell route when a required default slot record is missing during fallback boot', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    await next.start({ skipBuild: true })
+
+    const workspaceRoute = '/workspace/acme/channel/general/thread/123'
+    const workspaceUrlPath = `/docs${workspaceRoute}`
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
+        },
+      })
+      await waitForOfflineNavigationServiceWorker(browser, page!)
+
+      await browser.elementById('prefetch-workspace-shell').click()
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93675, opened 2026-05-08
(1/1) Cover missing parallel slot offline navigation misses

feedthejim:
## Stack position

This is a test-only stress slice on top of #93674, which covers the destructive missing default-slot case for the Slack-like workspace shell fixture introduced in #93673.

Stack context:

1. #93670 covers a missing persisted route record.
2. #93671 covers a missing persisted head record.
3. #93672 covers a missing persisted page segment record.
4. #93673 proves the complex workspace shell can replay when all required route, segment, slot, and head records are present.
5. #93674 covers a missing required default sidebar slot record.
6. This PR covers the paired missing required activity parallel-slot record.

## What changed

Adds one focused production e2e for `offlineNavigations` with Cache Components:

- online-loads `/docs` and waits for the offline navigation service worker
- prefetches `/docs/workspace/acme/channel/general/thread/123`
- verifies the durable route record, page segment, default sidebar slot, activity parallel slot, and head record were persisted
- deletes the exact URL entry so the hard reload must use route/segment reconstruction
- deletes only the `@activity` parallel-slot page segment record
- stops the server, puts the browser offline, and hard-loads the workspace route
- asserts the fallback boot emits `router-cache-reconstruction-miss` with `missing-segment`, falls back to visible `missing-entry` UI, and does not render the workspace page

The test-only IndexedDB deletion helper now accepts an optional `requestKeySubstring` so this PR can target the activity parallel slot without also deleting the main page segment.

## Why this is separate

#93673 proves the full shell can replay, and #93674 proves a missing default slot misses safely. This PR keeps the parallel-route failure mode isolated so reviewers can answer one question: if a required parallel slot record is absent, do we miss instead of rendering a partial shell with the primary page or another slot still present?

This also advances the advanced e2e stress matrix in `.private-investigation/offline-navigations-production-plan.md`: route, head, page segment, default slot, and parallel slot deletions now each have focused browser-backed negative coverage.

## Reviewer focus

Please focus on the false-positive guards:

- the exact URL entry is deleted first, so the test cannot pass through direct exact-URL replay
- the route, page segment, default sidebar slot, and head records are asserted present after the activity slot is deleted
- the expected result is a structured `missing-segment` reconstruction miss plus visible `missing-entry` UI, not a partially rendered workspace page

## Intentionally not covered here

Still pending in later stress slices:

- missing root layout record
- missing nested workspace layout record
- metadata/head style ordering integrity on successful replay
- app/session or workspace-scope partitioning

## Verification

[...]

PR #93677, opened 2026-05-08
(1/1) Cover missing root layout offline navigation misses

feedthejim:
## Stack position

This is a test-only stress slice on top of #93676, which covers the destructive missing nested workspace-layout case for the Slack-like workspace shell fixture introduced in #93673.

Stack context:

1. #93670 covers a missing persisted route record.
2. #93671 covers a missing persisted head record.
3. #93672 covers a missing persisted page segment record.
4. #93673 proves the complex workspace shell can replay when all required route, segment, slot, and head records are present.
5. #93674 covers a missing required default sidebar slot record.
6. #93675 covers a missing required activity parallel-slot record.
7. #93676 covers a missing required nested workspace layout record.
8. This PR covers a missing required root layout record.

## What changed

Adds one focused production e2e for `offlineNavigations` with Cache Components:

- online-loads `/docs` and waits for the offline navigation service worker
- prefetches `/docs/workspace/acme/channel/general/thread/123`
- verifies the durable route record, root layout segment, nested workspace layout segment, page segment, default sidebar slot, activity parallel slot, and head record were persisted
- deletes the exact URL entry so the hard reload must use route/segment reconstruction
- deletes only the root layout segment record, whose persisted request key is `''`
- stops the server, puts the browser offline, and hard-loads the workspace route
- asserts the fallback boot emits `router-cache-reconstruction-miss` with `missing-segment`, falls back to visible `missing-entry` UI, and does not render the workspace page

The test-only IndexedDB deletion helper now accepts an exact `requestKey` so the root segment can be targeted without relying on `endsWith('')`, which would match every segment.

## Why this is separate

The previous workspace-shell stress slices cover route, head, page, default slot, parallel slot, and nested layout deletion. This PR isolates the root layout case so reviewers can verify that replay cannot boot a route graph when the root app shell segment is missing, even if the nested workspace shell and page records still exist.

This completes the core missing-record matrix in `.private-investigation/offline-navigations-production-plan.md` for route, head, page segment, default slot, parallel slot, nested layout, and root layout records.

## Reviewer focus

Please focus on the targeting and false-positive guards:

- the test asserts the root segment request key `''` exists before deleting it
- the exact URL entry is deleted first, so this cannot pass through direct exact-URL replay
- the route, nested workspace layout, page segment, default sidebar slot, activity parallel slot, and head records remain present after the root layout is deleted
- the expected result is a structured `missing-segment` reconstruction miss plus visible `missing-entry` UI, not a partially rendered workspace page

## Intentionally not covered here

Still pending in later stress slices:

- metadata/head style ordering integrity on successful replay
- app/session or workspace-scope partitioning
- broader freshness, race, and storage-pressure cases

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses a workspace shell route when the required root layout record is missing during fallback boot"`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "required (root|nested) layout record"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "required (root|nested) layout record"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -229,12 +229,18 @@ describe('offlineNavigations build artifacts', () => {
     browser: Awaited<ReturnType<typeof next.browser>>,
     options: {
       keySubstring: string
+      requestKey?: string
       requestKeySubstring?: string
-      requestKeySuffix: string
+      requestKeySuffix?: string
     }
   ) {
     return browser.eval(
-      async ({ keySubstring, requestKeySubstring, requestKeySuffix }) => {
+      async ({
+        keySubstring,
+        requestKey,
+        requestKeySubstring,
+        requestKeySuffix,
+      }) => {
         const database = await new Promise<IDBDatabase>((resolve, reject) => {
           const request = indexedDB.open('next-offline-navigation-cache', 3)
           request.onsuccess = () => resolve(request.result)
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93678, opened 2026-05-08
(1/1) Cover session reset offline navigation misses

feedthejim:
## Stack Position

This is a test-only stress PR stacked on #93677, `(1/1) Cover missing root layout offline navigation misses`. It is the `stress-session-reset` follow-up from the offline navigations production plan and covers scenario `SESSION-02b`: same-browser request-sensitive exact URL replay after an explicit app-owned session reset.

This PR does not change production code. It only adds fixture UI and one focused production e2e so reviewers can evaluate this policy boundary separately from the broader route graph and service worker work.

## Why This Exists

The browser-private replay policy intentionally does not treat cookie or auth usage as a blanket reason to skip offline replay. If this browser already received a fresh payload, an offline navigation may replay it for the same build and same app/session scope.

The important safety boundary is what happens after the app knows the session scope changed. This PR proves the current reset primitive can make previously replayable exact URL data unreachable without relying on deletion of the raw record.

## What Changed

- Adds `ResetOfflineNavigationSessionButton` to the offline navigation fixture.
- The button clears the fixture `offline-session` cookie and calls `clearOfflineNavigationCache()`.
- Adds the button to the fixture root page so the e2e can trigger an app-owned reset through browser UI.
- Adds `misses request-sensitive exact URLs after an app-owned session reset`.

## Behavior Covered

The e2e walks the full contract in one browser profile:

1. Load `/docs/request-sensitive/` online with `offline-session=alpha`.
2. Assert an exact URL initial-load record was persisted with the current build/deployment ID.
3. Go offline and hard-load the same URL, proving same-session personalized data can replay and `useOffline()` reports offline.
4. Reconnect, click the reset button, and assert the session cookie is gone.
5. Assert exact URL, route, and segment cache epochs advanced.
6. Assert the old raw exact URL entry can still exist with its old epoch.
7. Go offline and hard-load the same URL again.
8. Assert fallback boot now renders visible cache-miss UI with `missing-entry` and does not render the personalized page.

This protects the intended production posture: same-browser personalized replay is allowed while scope is stable, and explicit app-owned scope reset makes old data unreachable.

## Intentionally Not Covered Yet

- Account switch.
- Workspace switch.
- Entitlement or session-version rotation.
- Multi-profile isolation.
- Races where reset overlaps a write or fallback boot.

Those remain separate stress rows so this PR stays reviewable.

## Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write test/production/app-dir/offline-navigations/app/prefetch-button.tsx test/production/app-dir/offline-navigations/app/page.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `npx eslint --config eslint.config.mjs --fix test/production/app-dir/offline-navigations/app/prefetch-button.tsx test/production/app-dir/offline-navigations/app/page.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `git diff --check -- test/production/app-dir/offline-navigations/app/prefetch-button.tsx test/production/app-dir/offline-navigations/app/page.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses request-sensitive exact URLs after an app-owned session reset"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses request-sensitive exact URLs after an app-owned session reset"`

### DIFF
--- a/test/production/app-dir/offline-navigations/app/page.tsx
+++ b/test/production/app-dir/offline-navigations/app/page.tsx
@@ -14,6 +14,7 @@ import {
   DynamicPatternTargetPrefetchButton,
   PrefetchButton,
   RefreshButton,
+  ResetOfflineNavigationSessionButton,
   WorkspaceShellPrefetchButton,
 } from './prefetch-button'
 
@@ -95,6 +96,7 @@ export default function Page() {
       <WorkspaceShellPrefetchButton />
       <RefreshButton />
       <ClearOfflineNavigationCacheButton />
+      <ResetOfflineNavigationSessionButton />
     </>
   )
 }
--- a/test/production/app-dir/offline-navigations/app/prefetch-button.tsx
+++ b/test/production/app-dir/offline-navigations/app/prefetch-button.tsx
@@ -81,3 +81,23 @@ export function ClearOfflineNavigationCacheButton() {
     </>
   )
 }
+
+export function ResetOfflineNavigationSessionButton() {
+  const [result, setResult] = useState('idle')
+  return (
+    <>
+      <button
+        id="reset-offline-navigation-session"
+        onClick={async () => {
+          document.cookie = 'offline-session=; Max-Age=0; path=/; SameSite=Lax'
+          setResult(
+            (await clearOfflineNavigationCache()) ? 'cleared' : 'not-cleared'
+          )
+        }}
+      >
+        Reset offline navigation session
+      </button>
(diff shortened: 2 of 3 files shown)


feedthejim:

PR #93679, opened 2026-05-08
(1/1) Cover workspace metadata offline replay

feedthejim:
## Stack Position

This is a focused stress PR stacked on #93678, `(1/1) Cover session reset offline navigation misses`. It covers the positive metadata/head portion of the `META-01` stress row and the complex app-shell follow-up for `SHELL-01`.

The earlier workspace-shell PRs proved that a fully prefetched app shell can replay from persisted route and segment records and that missing layout/default/parallel/page records miss visibly. This PR adds the next narrow claim: the same route/segment replay path also restores route-owned metadata instead of leaving stale fallback document head state behind.

## What Changed

- Adds `generateMetadata()` to the workspace thread fixture route.
- The fixture now emits a dynamic title and meta description derived from `workspace`, `channel`, and `thread` params.
- Adds `restores workspace shell metadata from persisted router records`.

## Behavior Covered

The new e2e:

1. Prefetches `/workspace/acme/channel/general/thread/123`.
2. Asserts the exact URL record, route record, workspace segment record, and `/_head` record exist.
3. Deletes the exact URL record so fallback boot cannot pass through the simpler exact URL replay path.
4. Stops the server and hard-loads the route offline.
5. Asserts a `router-cache` cache-hit diagnostic for the requested URL.
6. Asserts the workspace route UI renders with `useOffline()` reporting offline.
7. Asserts `document.title` and `meta[name="description"]` match the route metadata.

That gives reviewers a direct browser-level proof that route/segment replay includes the head record, not only the visible page body.

## Intentionally Not Covered Yet

- Canonical link restoration.
- Head deduplication and ordering.
- Stale head record behavior.
- CSS, font preload, and runtime chunk coherence.
- Blank-root or missing-runtime red-team cases.

Those stay in later `META-*`, `ASSET-*`, and runtime integrity stress slices so this PR remains one claim.

## Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write test/production/app-dir/offline-navigations/app/workspace/[workspace]/channel/[channel]/thread/[thread]/page.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `npx eslint --config eslint.config.mjs --fix test/production/app-dir/offline-navigations/app/workspace/[workspace]/channel/[channel]/thread/[thread]/page.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
[...]

PR #93680, opened 2026-05-08
(1/1) Cover encoded path offline navigation misses

feedthejim:
## Stack Position

This is a focused URL-identity stress PR stacked on #93679, `(1/1) Cover workspace metadata offline replay`. It covers the next small `EXACT-04` row after the existing query-collision coverage: encoded path identity.

This PR is intentionally test-only. It does not change the offline navigation keying implementation; it adds an e2e that would fail if the exact URL cache normalizer accidentally treated an encoded plus path and a literal plus path as the same offline document.

## What Changed

- Adds `misses exact-URL replay for encoded path identity collision variants`.
- Uses the existing `url-stress/[value]` fixture route.

## Behavior Covered

The e2e:

1. Loads `/docs/url-stress/a%2Bb/?token=encoded-path` online.
2. Asserts the page rendered the encoded path value and the exact URL initial-load record was persisted for the encoded URL.
3. Stops the server and switches the browser offline.
4. Hard-loads nearby `/docs/url-stress/a+b/?token=encoded-path`.
5. Asserts the fallback boot renders visible `missing-entry` UI.
6. Asserts the cache diagnostic is for the literal-plus requested URL.
7. Asserts no `url-stress-page` content rendered, so the test cannot pass by replaying the cached encoded-path route.

This protects the exact URL cache contract: nearby URL spellings do not collide just because they decode to visually similar strings.

## Intentionally Not Covered Yet

- Encoded slash variants.
- Path case-sensitivity matrix.
- Canonical redirect identity.
- Deployment URL changes.
- Remaining basePath, assetPrefix, and trailingSlash combinations beyond the current fixture settings.

Those stay as separate URL-identity stress rows so this PR remains one reviewable claim.

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL replay for encoded path identity collision variants"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "misses exact-URL replay for encoded path identity collision variants"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -5466,6 +5466,132 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('misses exact-URL replay for encoded path identity collision variants', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    const navigationBuildId = next.deploymentId ?? buildId
+    await next.start({ skipBuild: true })
+
+    let page: Playwright.Page | undefined
+    try {
+      const browser = await next.browser('/docs', {
+        beforePageLoad(p: Playwright.Page) {
+          page = p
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93688, opened 2026-05-08
(1/1) Cover offline route handler mutation boundaries

feedthejim:
## Stack Position

This is a focused request-shape stress PR stacked on #93687, `(1/1) Cover offline redirecting server action boundaries`.

It covers the fourth small `MUTATION-01` row from the offline navigations stress plan: route-handler mutation side effects while offline. This PR only changes the offline navigation test fixture and e2e coverage. It does not change service worker routing, fallback boot, or production mutation behavior.

## What Changed

- Adds a fixture route handler at `/docs/api/offline-mutation/` that sets a test cookie on POST.
- Adds `does not reach route handler mutation side effects while offline` to the offline navigations production e2e suite.
- Uses a real browser form POST targeted at a hidden iframe so the app page remains inspectable after the offline submission.

## Behavior Covered

The e2e:

1. Builds and starts the offline navigations fixture.
2. Loads `/docs` online and waits for the generated service worker.
3. Registers a listener for `next-offline-navigation-fallback-served` messages.
4. Creates a plain POST form targeting the mutation route handler.
5. Switches the browser context offline.
6. Submits the form and asserts the POST was issued but failed with an offline network error.
7. Asserts no fallback-served message was emitted to the app page.
8. Asserts the iframe did not reach the fallback target document.
9. Asserts the mutation cookie was not set while offline.
10. Reconnects, submits the same form online, and asserts the route handler sets the cookie.

This protects the service worker boundary while proving the fixture still exercises a real mutation path when online.

## Intentionally Not Covered Yet

- Automatic reconnect retry UI for failed offline mutations.
- Successful online retry invalidating durable offline records before a later offline replay.

## Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write test/production/app-dir/offline-navigations/offline-navigations.test.ts test/production/app-dir/offline-navigations/app/api/offline-mutation/route.ts`
- `npx eslint --config eslint.config.mjs --fix test/production/app-dir/offline-navigations/offline-navigations.test.ts test/production/app-dir/offline-navigations/app/api/offline-mutation/route.ts`
- `git diff --check -- test/production/app-dir/offline-navigations/offline-navigations.test.ts test/production/app-dir/offline-navigations/app/api/offline-mutation/route.ts`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "does not reach route handler mutation side effects while offline"`
[...]

PR #93690, opened 2026-05-08
(1/1) Cover offline request shape boundaries

feedthejim:
## Why

This is a small test-only stress slice on top of the offline mutation-boundary PRs. It covers the next P0 request-boundary case from `.private-investigation/offline-navigations-production-plan.md`: the generated service worker should only serve fallback HTML for failed same-origin document navigations.

The earlier pass-through stress coverage already checks broad non-document requests like RSC fetches, API fetches, POST fetches, and static manifest fetches. This PR adds a narrower matrix for request shapes that are easy to accidentally misclassify when the request looks document-ish or unusual.

## Stack Position

Base PR: #93688, `(1/1) Cover offline route handler mutation boundaries`

This PR is intentionally independent from the route replay and output-export work. It does not change implementation behavior. It only adds browser-backed coverage around the service-worker boundary contract introduced earlier in the offlineNavigations stack.

## What Changed

Adds one production e2e:

- `does not serve fallback HTML to offline non-document request shapes`

The test boots the fixture app, waits for the offline navigation service worker, sets the browser offline, then issues these exact same-origin request shapes from the page:

- `HEAD` route handler request
- `OPTIONS` route handler request
- `DELETE` route handler request
- custom-header `POST` route handler request
- `GET` fetch with `Accept: text/html`
- static manifest fetch with a `Range` header

For each shape, the test asserts two things:

- Playwright observed the exact method, URL, and relevant headers.
- No `next-offline-navigation-fallback-served` service-worker message was posted.

That avoids the false-positive pattern where a test passes because some unrelated request failed first after going offline.

## Intentionally Not Covered Here

This PR does not cover download navigations, navigation preload, aborted document requests, service-worker cache corruption, or server-stopped non-document request paths. Those stay as separate stress slices in the plan so this PR remains reviewable.

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "does not serve fallback HTML to offline non-document request shapes"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "does not serve fallback HTML to offline non-document request shapes"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -5326,6 +5326,213 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('does not serve fallback HTML to offline non-document request shapes', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const buildResult = await next.build()
+    expect(buildResult.exitCode).toBe(0)
+
+    const { buildId } = await getOfflineNavigationArtifactPaths()
+    await next.start({ skipBuild: true })
+
+    const requestShapes: Array<{
+      expected: {
+        accept?: string
+        key: string
+        method: string
+        range?: string
+        url: string
+      }
+      init: RequestInit
+      input: string
+      key: string
+    }> = [
+      {
+        expected: {
+          key: 'head',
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93693, opened 2026-05-08
(1/1) Cover offline manifest cache damage

feedthejim:
## Stack position

This is a focused service-worker cache-damage stress slice on top of #93692, `feedthejim/offline-navigations-corrupted-fallback-cache-stress`.

The parent PR validates that the worker does not serve a corrupted fallback document. This PR covers the adjacent post-install manifest cases: the cached offline-navigation manifest is missing or corrupted, while the generated fallback document itself is still intact.

## Why this slice exists

The manifest is required for install/update caching, but document fallback at request time should not turn into a router or depend on reparsing a cached manifest. Once the worker is installed, document fallback should be driven by the generated worker metadata and the cached fallback document.

So this PR proves a different contract from #93691 and #93692:

- Missing fallback document: fail normally, no fake fallback boot.
- Corrupted fallback document: fail normally, no fake fallback boot.
- Missing or corrupted cached manifest after install: still serve the valid fallback document, then let the client-owned offline restore path produce the visible cache miss.

## What changed

- Adds an e2e that runs both post-install manifest damage variants:
  - delete the cached `_offline-navigation-manifest.json`, and
  - replace it with corrupted JSON.
- In both cases, the test keeps the cached fallback HTML intact.
- It then takes the browser offline and hard-navigates to a route without durable route data.
- It asserts the browser receives the generated fallback HTML, the worker emits `fallback-served`, and the client reaches visible `missing-entry` cache-miss UI with diagnostics.

- Deleted fallback document, covered by #93691.
- Corrupted fallback document, covered by #93692.
- Install-time manifest fetch or JSON failure.
- Service worker update failures.
- Runtime chunk, CSS, font, or build-manifest damage after a valid fallback document has been served.

Those remain separate stress slices because they exercise different ownership boundaries.

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "serves fallback HTML when the cached manifest is missing or corrupted"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "serves fallback HTML when the cached manifest is missing or corrupted"`
- `pnpm --filter=next build`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -5728,6 +5728,189 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('serves fallback HTML when the cached manifest is missing or corrupted', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    for (const manifestDamage of [
+      {
+        kind: 'missing',
+        targetPath: '/docs/prefetched?missing-manifest-cache=1',
+      },
+      {
+        kind: 'corrupted',
+        targetPath: '/docs/prefetched?corrupted-manifest-cache=1',
+      },
+    ]) {
+      const buildResult = await next.build()
+      expect(buildResult.exitCode).toBe(0)
+
+      const { buildId } = await getOfflineNavigationArtifactPaths()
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93694, opened 2026-05-08
(1/1) Cover offline manifest install failures

feedthejim:
## Stack position

This PR sits directly on top of #93693, `(1/1) Cover offline manifest cache damage`. That prior slice covers post-install Cache Storage damage where the fallback document is still valid and the cached manifest is missing or corrupted.

This PR covers the earlier install-time boundary: the generated manifest is missing or malformed before the service worker can finish installing. It is a focused service-worker lifecycle stress slice for `SW-04d-manifest-install-failure` in the private production plan.

## Why

The generated worker caches the offline navigation manifest during `install`, reads it, then uses it to cache the fallback document. If that manifest cannot be fetched or parsed, the worker should fail closed. It should not activate a broken controller, fake a fallback boot, or leave the app in a blank offline state.

The important review question is narrow: when the install-time manifest dependency is damaged, do we keep normal online navigation working and avoid claiming offline fallback ownership?

## What changed

- Adds a production e2e for missing and corrupted generated manifest files before service-worker install.
- Verifies the online app page still loads after the damaged build artifact.
- Verifies the `/docs/` service worker does not become an active controller.
- Verifies an offline document navigation fails as a normal browser network failure.
- Verifies no `next-offline-navigation-fallback-served` message is emitted.

- Service-worker update failures after a previous worker is already active.
- Runtime chunk, CSS, build manifest, or asset damage after the fallback document is served.
- Client-side exact URL or route/segment replay behavior.

Those are separate stress slices in the plan so this PR stays reviewable.

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "does not activate the service worker when install-time manifest caching fails"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "does not activate the service worker when install-time manifest caching fails"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -1,4 +1,4 @@
-import { existsSync } from 'fs'
+import { existsSync, unlinkSync, writeFileSync } from 'fs'
 import { join } from 'path'
 import { nextTestSetup } from 'e2e-utils'
 import { retry } from 'next-test-utils'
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93695, opened 2026-05-08
(1/1) Clean up stale offline navigation workers

feedthejim:
## Stack position

This PR sits directly on top of #93694, `(1/1) Cover offline manifest install failures`. The previous slice proves a worker whose manifest dependency is damaged before install fails closed and never activates.

This slice handles the next service-worker lifecycle boundary: a user had `experimental.offlineNavigations` enabled, installed the generated worker, then deploys a build where `offlineNavigations` is disabled and no offline navigation artifacts are emitted.

## Why

Disabling the flag should not leave an old generated worker serving stale fallback HTML. At the same time, disabled builds should not need a new client-side cleanup path that runs for apps that never opted into the feature.

This PR keeps the cleanup responsibility inside the generated worker from the enabled build. On successful online document navigations, the worker refreshes its required offline-navigation resources in `waitUntil`. If the current deployment no longer has those resources, the old worker fails closed by deleting its own Cache Storage namespace and unregistering itself. Offline fallback responses do not trigger this cleanup path.

## What changed

- Adds generated-worker cleanup helpers for stale offline navigation registrations.
- Keeps document navigation network-first and continues to serve fallback only on document network failure.
- Refreshes required worker resources after successful online document responses.
- Deletes only the current worker cache namespace and unregisters when that refresh fails.
- Adds an e2e that installs the worker with the flag enabled, rebuilds the same app with `offlineNavigations` disabled, proves the disabled build emits no artifacts, verifies the old cache namespace is removed, and verifies a later offline document navigation fails normally without `fallback-served`.

- Multi-tab old/new controller update races.
- Runtime chunk, CSS, build manifest, or asset damage after fallback document boot.
- Client router exact URL or route/segment replay behavior.

Those remain separate stress rows so this PR is reviewable around one lifecycle contract.

## Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write packages/next/src/build/offline-navigation-service-worker.ts test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `npx eslint --config eslint.config.mjs --fix packages/next/src/build/offline-navigation-service-worker.ts test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `git diff --check -- packages/next/src/build/offline-navigation-service-worker.ts test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "emits request-invariant offline navigation artifacts when enabled|cleans up stale service worker artifacts when offlineNavigations is disabled"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "emits request-invariant offline navigation artifacts when enabled|cleans up stale service worker artifacts when offlineNavigations is disabled"`
- `pnpm --filter=next build`

### DIFF
--- a/packages/next/src/build/offline-navigation-service-worker.ts
+++ b/packages/next/src/build/offline-navigation-service-worker.ts
@@ -58,6 +58,18 @@ async function cacheOfflineNavigationResources(){
   const fallbackResponse=await fetchRequiredResource(manifest.fallbackDocument.href);
   await cache.put(manifest.fallbackDocument.href,fallbackResponse);
 }
+async function unregisterOfflineNavigationServiceWorker(){
+  const metadata=self.__NEXT_OFFLINE_NAVIGATION_SW;
+  await caches.delete(metadata.cacheNamespace);
+  await self.registration.unregister();
+}
+async function refreshOfflineNavigationResources(){
+  try{
+    await cacheOfflineNavigationResources();
+  }catch(err){
+    await unregisterOfflineNavigationServiceWorker();
+  }
+}
 function isDocumentNavigationRequest(request){
   if(request.method!=='GET'||request.mode!=='navigate'||request.destination!=='document'){
     return false;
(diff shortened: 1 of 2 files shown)


feedthejim:

PR #93696, opened 2026-05-08
(1/1) Cover stale offline manifest refreshes

feedthejim:
## Stack position

This PR sits directly on top of #93695, `(1/1) Clean up stale offline navigation workers`. The previous PR adds the cleanup primitive in the generated worker: after successful online document responses, the worker refreshes its required offline resources and fails closed if they are unavailable or invalid.

This PR is a test-only stress slice for that primitive. It covers the old-controller case where the worker was already active, then the generated manifest becomes corrupted before a later online navigation.

## Why

Install-time manifest failure is covered by #93694. Flag-disabled cleanup is covered by #93695. This slice proves the same cleanup behavior protects a stale active worker during a later manifest refresh, so it cannot keep serving old fallback HTML after its required manifest is no longer valid.

The review question is narrow: if an already-active generated worker sees a corrupted manifest during a later online refresh, does it lose the ability to serve stale offline fallback?

## What changed

- Adds a production e2e that first installs and controls the page with a valid offline navigation worker.
- Rebuilds the app on the same origin, corrupts the generated manifest file, and starts the app again.
- Navigates online so the old worker attempts to refresh its required resources.
- Verifies the offline-navigation Cache Storage namespace is deleted.
- Verifies a later offline document navigation fails normally and emits no `next-offline-navigation-fallback-served` message.

- Multi-tab old/new controller races.
- Runtime JS, CSS, font, or build-manifest damage after fallback document boot.
- Client router exact URL or route/segment replay behavior.

Those remain separate production-hardening slices.

## Verification

- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "does not serve stale fallback HTML when a later manifest refresh is corrupted"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "does not serve stale fallback HTML when a later manifest refresh is corrupted"`

### DIFF
--- a/test/production/app-dir/offline-navigations/offline-navigations.test.ts
+++ b/test/production/app-dir/offline-navigations/offline-navigations.test.ts
@@ -6231,6 +6231,151 @@ describe('offlineNavigations build artifacts', () => {
     }
   })
 
+  it('does not serve stale fallback HTML when a later manifest refresh is corrupted', async () => {
+    if (shouldSkipReplayWithCachedNavigations) {
+      return
+    }
+
+    const initialBuildResult = await next.build()
+    expect(initialBuildResult.exitCode).toBe(0)
+
+    const initialArtifacts = await getOfflineNavigationArtifactPaths()
+    await next.start({ skipBuild: true })
+
+    const previousForcedPort = next.forcedPort
+    let isStarted = true
+    let page: Playwright.Page | undefined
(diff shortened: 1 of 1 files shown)


feedthejim:

PR #93697, opened 2026-05-08
(1/1) Cover fallback runtime bootstrap failures

feedthejim:
## Stack position

This PR sits on top of #93696, after the service-worker manifest refresh failure stress slice. It is a focused production-hardening slice for the `ASSET-01` / `RUNTIME-01` runtime bootstrap gap in the offlineNavigations stress plan.

The service worker still only owns document survival. This PR does not teach the service worker to route, inspect RSC, or resolve assets. It covers the narrower case where the generated fallback HTML is served correctly, but the app runtime scripts referenced by that fallback cannot execute.

## Why

Before this slice, a fallback document could be served from the service worker and still leave the user on a blank fallback root if the client runtime never loaded. That is a dangerous false positive for offline navigations: the service worker did its job, but the app never reached either a cache hit or a visible cache miss.

For production apps, a missing or stale runtime chunk should be observable as a bounded fallback boot failure. It should not look like a pending cache hit, and it should not require route-data logic in the service worker.

## What changed

- Adds a small inline watchdog to the generated offline navigation fallback document.
- The watchdog reveals the existing generated cache-miss UI and records a `runtime-error` diagnostic if the app runtime never starts or a root `/_next/static/*.js` script fails to load.
- The real app runtime cancels that watchdog as soon as `app-index` starts executing, so normal exact-URL and route/segment fallback boot still own their own hit/miss diagnostics.
- Adds a production e2e that corrupts the cached fallback document so its runtime script URLs point at a missing chunk, then hard-loads offline through the service worker and proves the page reaches a visible `runtime-error` miss instead of a blank root.

## Review focus

The important boundary is that the fallback document only handles the "runtime never started" case. Once `app-index` executes, the existing client-owned offline navigation bootstrap remains responsible for exact URL replay, route/segment replay, IndexedDB reads, and normal cache hit/miss diagnostics.

## Intentionally not covered yet

This is not the full runtime/bootstrap matrix. Follow-up stress slices still need to cover stale runtime chunks after build updates, CSS/font preload failures, build-manifest skew, cross-origin `assetPrefix` runtime failures, and old/new controller races.

## Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write packages/next/src/build/offline-navigation-fallback.ts packages/next/src/client/app-index.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `npx eslint --config eslint.config.mjs --fix packages/next/src/build/offline-navigation-fallback.ts packages/next/src/client/app-index.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `git diff --check -- packages/next/src/build/offline-navigation-fallback.ts packages/next/src/client/app-index.tsx test/production/app-dir/offline-navigations/offline-navigations.test.ts`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "emits request-invariant offline navigation artifacts when enabled|shows cache miss when fallback runtime scripts fail to load"`
- `pnpm test-start-webpack test/production/app-dir/offline-navigations/offline-navigations.test.ts -t "emits request-invariant offline navigation artifacts when enabled|shows cache miss when fallback runtime scripts fail to load"`
- `pnpm --filter=next build`

`pnpm --filter=next build` passed with existing Rspack/devtools warnings. The background `pnpm --filter=next dev` watcher later exited with the known `RspackResolver(NotFound("../../runtime-reacts.external"))` failure after the explicit build completed; the focused tests and explicit build above passed.

### DIFF
--- a/packages/next/src/build/offline-navigation-fallback.ts
+++ b/packages/next/src/build/offline-navigation-fallback.ts
@@ -33,6 +33,10 @@ function getScriptAttributes(
   return ` crossOrigin="${htmlEscapeAttributeString(crossOrigin)}"`
 }
 
+function createRuntimeWatchdogScript(): string {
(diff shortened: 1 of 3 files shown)


feedthejim:

PR #92053, opened 2026-03-28
fix: prevent infinite request loop when forwarding server actions with middleware rewrites

claygeo:
## Problem

Fixes #84504

When a server action fires right before navigation, Next.js forwards the action to the correct worker via `createForwardedActionResponse`. If middleware applies a rewrite to the forwarded request, the rewritten URL arrives at a worker that also wants to forward, creating an infinite request loop that never resolves.

## Root Cause

In `action-handler.ts`, the forwarding guard at line 713 checks `if (actionId)` but does not check whether the action was already forwarded:

```typescript
const actionWasForwarded = Boolean(req.headers['x-action-forwarded'])  // line 711: READ but not used as guard

if (actionId) {  // line 713: missing !actionWasForwarded check
  const forwardedWorker = selectWorkerForForwarding(actionId, page)
  if (forwardedWorker) {
    return createForwardedActionResponse(...)  // forwards again → infinite loop
  }
}
```

The `x-action-forwarded` header is already set on outgoing forwarded requests (`createForwardedActionResponse`, line 221) and read on incoming requests (line 711). The loop detection mechanism exists, it's just not wired into the forwarding guard.

## Fix

Add `!actionWasForwarded` to the forwarding condition:

```typescript
if (actionId && !actionWasForwarded) {
```

If an action has already been forwarded once (header present), skip the forwarding logic entirely. The action will be handled by the current worker instead of bouncing infinitely between workers.

5 lines changed in 1 file.

### Note on PR #84525
PR #84525 addresses the same issue but has been abandoned for 6 months with zero reviews. This PR takes the same conceptual approach (check `x-action-forwarded` header) with a cleaner implementation at the caller level rather than inside `createForwardedActionResponse`.

PR #84525, opened 2025-10-05
Fix/action forward loop guard

LiamBoz:
## What?

Fixes #84504

Add a minimal loop-detection guard to `createForwardedActionResponse` so a Server Action request that has **already** been forwarded (marked by `x-action-forwarded=1`) is **not** forwarded again. Instead, we return early (no second forward).

- File: `packages/next/src/server/app-render/action-handler.ts`
- Behavior change: if `x-action-forwarded=1` is present on the **incoming** request, return from `createForwardedActionResponse` without forwarding again.

## Why?

In apps that use Middleware rewrites, a forwarded Server Action request can be rewritten to a path that triggers forwarding **again**, causing an infinite loop (e.g. `/rewrite/other → /rewrite/... → /rewrite/rewrite/... → …`). This leads to repeated logs and a server action that never resolves.

This guard breaks the cycle with very low risk by only affecting requests that were already forwarded once.

- Repro repo: https://github.com/blairmcalpine/forwarded-server-action-request-loop-repro
- Steps:
  1) `npm i && npm run dev`
  2) Visit `http://localhost:3000`
  3) Click **Run action**, then quickly (<500ms) click **Navigate to other page**
  4) **Before this change:** infinite loop/log spam; action never resolves.
  5) **After this change:** no loop; request is not re-forwarded again.

## How?

At the start of `createForwardedActionResponse`, inspect the incoming headers:

- If `x-action-forwarded=1` is present → **return** (do not forward again).
- Otherwise, proceed as before and set `x-action-forwarded=1` on the outgoing forwarded request.

Works in both Node and Edge:
- If `req.headers` is already a `Headers` (Edge), use it.
- Otherwise adapt Node headers via `HeadersAdapter`.

## Test Plan

**Manual (using the public repro)**

- With this patch built and packed into the repro app:
  - Trigger the fast **Run action → Navigate** sequence.
  - Confirm:
    - No repeated `/rewrite/...` requests or log spam.
    - The page no longer gets stuck due to an infinite forwarding loop.

**Notes:** This PR intentionally focuses on **loop prevention**. A broader follow-up could make forwarded requests mirror the *pre-rewrite* target to ensure resolution semantics remain identical under rewrites. That would be a separate change.

## Risk

Low:
- Only affects requests that already bear `x-action-forwarded=1`.
- Does not change the initial legitimate forward.
- Keeps all other server action behavior intact.

---

## For Contributors Checklist

- [x] Bug fix
- [x] Related issue linked (`Fixes #84504`)
[...]

PR #93710, opened 2026-05-08
fix: skip worker forwarding when action was already forwarded

### LINKED ISSUE #84504: Forwarded server actions with middleware rewrites cause an infinite request loop

https://github.com/blairmcalpine/forwarded-server-action-request-loop-repro

### To Reproduce

1. npm i
2. npm run dev
3. Visit localhost:3000
4. Click "run action" and then quickly (<500ms in between) click "Navigate to other page".
5. Notice that (1) the server action fired by the "run action" button never resolves and (2) the server logs print out a ton of messages. 

Current: When a server action gets forwarded (which happens when it fires after navigation to another page that doesn't know about that server action) and gets rewritten by middleware, it causes an infinite request loop.

Expected: Server action properly resolves and no request loop happens.

For full context, [here](https://github.com/vercel/next.js/pull/64227/files) is where this server action forwarding functionality was added.

In this case, the forwarding makes a new request to /rewrite/ from /rewrite/other, because /rewrite/ is the page that defines the server action. However when the server action makes it to middleware, it gets rewritten to /rewrite/rewrite and the process continues forever.

My suggestion for a solution would be (1): At a minimum, add a loop detection check using the x-action-forwarded header. If an incoming action has been forwarded already, don't forward it again no matter what.

A real solution though would likely look like making the forwarded server action request in the same format that the incoming one had. In this case, because the incoming request was made to /other and rewritten to /rewrite/other, the forwarded server request should not include the rewrite.

One can argue that you just shouldn't do a rewrite on these forwarded actions (which is what we've done at Faire to work around this issue), however this forwarded action behaviour is completely internal and undocumented, so it's not clear that anywhere that this would even cause that problem. Regardless of a solution here, I believe this behaviour should be documented as well. It can cause a lot of weirdness and problems in self-hosted setups and custom server wrappers especially where __NEXT_PRIVATE_ORIGIN is undefined, in which the server ends up making a request all the way back to the public domain.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 22.15.1
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 15.5.4 // Latest available version is detected (15.5.4).
  eslint-config-next: 15.5.4
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Server Actions, Middleware

next dev (local), next start (local), Other (Deployed)

### Additional context

_No response_

<sub>[NEXT-4736](https://linear.app/vercel/issue/NEXT-4736/forwarded-server-actions-with-middleware-rewrites-cause-an-infinite)</sub>

elishagreenwald:
### What?

Stops the server action handler from forwarding to another worker when the incoming request was already forwarded (`x-action-forwarded: 1`).

### Why?

When middleware rewrites a Server Action POST to a page that doesn't bundle the action, the receiving worker forwards the request to a worker that does. The forwarded request hits middleware again and gets rewritten the same way, so the new receiving worker forwards again — looping until undici's headers timeout (~300s) or memory pressure brings the server down (#84504).

### How?

- **Runtime:** add `!actionWasForwarded` to the existing forwarding guard in `action-handler.ts`. The `x-action-forwarded` header was already set on forwarded requests but wasn't being used to gate the forwarding decision.
- **Tests:** Browser-driven e2e at `test/e2e/app-dir/action-forward-loop`. `proxy.ts` rewrites POSTs to `/with-action` (GETs still render normally, so the user can see the page); the page renders a `<form>` with an inline server action wrapped in a React error boundary. The test clicks the button and waits for the boundary's error UI. Without the fix, the action POST loops on the server and Playwright's 10s `waitForSelector` times out.

Follow-up TODO left in `createForwardedActionResponse`: when the forwarded worker returns an action-not-found 404, the forwarder currently swallows it and returns a generic JSON body, so the client throws a generic "unexpected response" error instead of `UnrecognizedActionError`. Fixing that would let the boundary branch on the specific error class (the boundary has a matching TODO).

Fixes #84504
Closes #84525
Closes #92053

Co-Authored-By: @LiamBoz
Co-Authored-By: @claygeo

- [x] Bug fix — issue linked with `Fixes #`
- [x] Tests added (e2e)
- [x] Signed / verified commits on the branch


elishagreenwald:
@unstubbable it looks like you may have written much of the surrounding action-handler code so I'm hoping this would be something you can review. This is causing a memory leak in prod for us that we had to patch nextjs to avoid. Issue #84504 has been open since October with two stalled prior PRs (#84525  & #92053 ) so I'm really hoping this one is shippable. Thank you kindly!

PR #93759, opened 2026-05-11
Honor Suspense-above-body opt-in for dynamic `generateViewport`

unstubbable:
When a route's only dynamic source was `generateViewport()`, the documented mitigation of wrapping `<body>` in `<Suspense>` did not work: the dev redbox still appeared and the build still failed. The opt-in is detected by matching a Suspense-above-body pattern against a dynamic hole's component stack, but the viewport's component subtree is a sibling of `RootLayoutBoundary` in `AppRouter` (rendered inside the framework `<Head>` component, not inside the user's root layout), so its stack never passes through the user's root layout where the Suspense lives. The outlet boundary, by contrast, renders alongside the page content and its stack does pass through the user's layout chain. We previously short-circuited the outlet branch without inspecting it.

This change introduces a `trackOutletSuspenseAboveBody` helper that runs the existing Suspense-above-body regex against the outlet's component stack and records the opt-in on `dynamicValidation.hasSuspenseAboveBody`. The helper is invoked from `trackAllowedDynamicAccess`, `trackDynamicHoleInRuntimeShell`, and `trackDynamicHoleInStaticShell`. It deliberately does not set `hasAllowedDynamic`, because that flag tracks whether the body itself is partially dynamic (which is the prerequisite for accepting dynamic metadata), while the outlet only tells us about the structural opt-in for an empty shell.

`hasSuspenseAboveBody` and `allowEmptyStaticShell` (which covers both `unstable_instant = false` and the build-phase fallback-shell case) express the same intent — "this shell is allowed to be empty/blocking" — so they're treated as synonyms for the purpose of bypassing shell-failure errors. The dynamic-`generateMetadata` error is a separate mistake-detection signal that fires when the rest of the shell is otherwise fully static apart from metadata. It's lifted above the opt-in bypass in both `getStaticShellDisallowedDynamicReasons` and `throwIfDisallowedDynamic` so it fires regardless of how the user opted in. The `if (!allowEmptyStaticShell)` wrapper at the `throwIfDisallowedDynamic` call site is dropped — the function now decides what to bypass internally.

Adds four e2e cases under `cache-components-errors`: `dynamic-viewport-static-with-suspense` and `dynamic-viewport-static-with-instant-false` (build succeeds, no dev redbox), and `dynamic-metadata-static-with-suspense-above-body` and `dynamic-metadata-static-with-instant-false` (metadata error still surfaces in both dev and build).

### DIFF
--- a/packages/next/src/server/app-render/dynamic-rendering.ts
+++ b/packages/next/src/server/app-render/dynamic-rendering.ts
@@ -790,6 +790,42 @@ function resolveInstantStack(
   return slotStacks[0] ?? null
 }
 
+/**
+ * Inspects the component stack of an outlet boundary to discover whether the
+ * user placed a Suspense boundary above the document body, and records the
+ * opt-in on `dynamicValidation.hasSuspenseAboveBody` if so.
+ *
+ * The outlet itself isn't a meaningful source of dynamic — it only resolves
+ * when metadata/viewport are dynamic, which we track via their own boundaries.
+ * However, the outlet renders alongside the page content, so its stack passes
+ * through the user's layout chain (typically reaching into `<body>` via the
+ * root layout). That makes the outlet stack our best opportunity to spot a
+ * Suspense boundary above the body, even when no real body content is dynamic.
+ * Without this, a route whose only dynamic source is `generateViewport()` would
+ * miss the Suspense-above-body opt-in, because the viewport's stack lives in
+ * the head and never sees the user's root layout.
+ *
+ * We deliberately only set `hasSuspenseAboveBody`, not `hasAllowedDynamic`. The
+ * latter tracks whether the body has dynamic content that's been wrapped in
+ * Suspense (i.e., the page is partially dynamic). The outlet rendering tells us
+ * about the structural opt-in for an empty shell, not about the body being
+ * partially dynamic. The distinction matters because dynamic metadata is only
+ * acceptable when the page is partially dynamic (via real body holes), and we
+ * don't want this outlet-based detection to mask that case.
+ */
+function trackOutletSuspenseAboveBody(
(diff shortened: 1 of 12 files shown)


packages/next/src/server/app-render/dynamic-rendering.ts
```diff
@@ -1282,10 +1315,19 @@ export function getStaticShellDisallowedDynamicReasons(
   dynamicValidation: DynamicValidationState,
   configAllowsBlocking: boolean
 ): Array<Error> {
-  if (configAllowsBlocking || dynamicValidation.hasSuspenseAboveBody) {
-    // This route has opted into allowing fully dynamic rendering
-    // by including a Suspense boundary above the body. In this case
-    // a lack of a shell is not considered disallowed so we simply return
+  if (
+    configAllowsBlocking ||
+    (dynamicValidation.hasSuspenseAboveBody && prelude !== PreludeState.Full)
```
gnoff: I don't quite get why this is right? config and suspenseAboveBody should be synonyms. If we would error still even with SuspenseAboveBody then we shoudl also with config allows blocking.

It's almost like the dynamic generateMetadata error isn't the same as all the other disallowed dynamic. We believe you that you want to allow dynamic but even still it might be a mistake that your metadata was dynamic when teh page wasn't so we separately error for this seeming incongruity

unstubbable: Addressed in 2232b94. The two opt-ins are now symmetric, and the dynamic-metadata mistake-detection check is lifted above the bypass in both `getStaticShellDisallowedDynamicReasons` and `throwIfDisallowedDynamic`.

gnoff (approved):

PR #93766, opened 2026-05-11
Craigandrews/ensure isr lru is written if requests collapsed

gnoff:
Adopted from: #93699

## Summary

When multiple ISR revalidation invocations (different `x-invocation-id`) arrive concurrently for the same path, the Batcher correctly deduplicates them into a single render. The LRU is only written inside `handleRevalidate`, and only the winning invocation executes this code. Batched invocations return the shared promise directly and never write to the LRU. Their follow-up requests miss the LRU and trigger a new render, cascading under load.

Moved the minimal mode LRU write from `handleRevalidate` to `get()`, after the Batcher resolves. Every caller now writes to the LRU under its own invocation key, so follow-up requests always find a cache entry.

## Test Plan

- Unit test: concurrent invocations with different `invocationID` values verify only one render executes and follow-up requests hit the LRU

### DIFF
--- a/packages/next/src/server/response-cache/index.test.ts
+++ b/packages/next/src/server/response-cache/index.test.ts
@@ -1,0 +1,116 @@
+import ResponseCache from './index'
+import { CachedRouteKind, type ResponseCacheEntry } from './types'
+import { RouteKind } from '../route-kind'
+import RenderResult from '../render-result'
+import { HTML_CONTENT_TYPE_HEADER } from '../../lib/constants'
+
+function mockIncrementalCache() {
+  return {
+    get: jest.fn().mockResolvedValue(null),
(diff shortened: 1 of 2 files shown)


feedthejim (approved):

gnoff (commented):
I can't approve since this is now my PR :)  but while this change is good in that it fixes a subtle but quite bad bug related to atomicity of revalidations in minimal mode it does highlight that our LRU cache isn't optimally modelling cross crequest deduping.

The bytes in the cache for different request IDs can be entirely the same but it doesn't trrack this and so if you do end up with concurrent revalidations that are meant to deduplicate across requests the LRU counts them multiple times (once per invocation) and could in theory evict one (without saving any real space).

The LRU should be updated to track novel bytes

PR #93767, opened 2026-05-11
Convert tests using createNext -> nextTestSetup

timneutkens:
## What?

Converts existing `createNext()` usage into `nextTestSetup()`. 

`createNext()` was the setup step we had before `nextTestSetup()` was added.

This PR focused on the simple conversion cases. There will be a follow-up to complete the last few.

### DIFF
--- a/test/development/pages-dir/client-navigation/index.test.ts
+++ b/test/development/pages-dir/client-navigation/index.test.ts
@@ -7,7 +7,7 @@ import {
   getRedboxTotalErrorCount,
 } from 'next-test-utils'
 import path from 'path'
-import { nextTestSetup } from 'e2e-utils'
+import { isReact18, nextTestSetup } from 'e2e-utils'
 
 describe('Client Navigation', () => {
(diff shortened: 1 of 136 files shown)


test/development/pages-dir/client-navigation/index.test.ts
```diff
@@ -64,19 +64,20 @@ describe('Client Navigation', () => {
...
         .elementByCss('#self-reload-link')
         .click()
-        .elementByCss('p')
-        .text()
 
-      // counts (page change + two clicks)
-      expect(countAfterClicked).toBe('COUNT: 3')
+      // counts (page change + two clicks). Use `retry()` because the third
+      // click's state update may not have flushed when we read the text.
+      await retry(async () => {
```
timneutkens, line 78: There was a flake here. The flake detection caught as the test was changed it ran.

unstubbable (approved):
We might wanna follow up with adding the merge commit to `.git-blame-ignore-revs`.

PR #93770, opened 2026-05-11
Instant Insights: When unable to complete validation provide a filename for the unvalidated boundary

gnoff:
The error message previously did not say which parallel route slot couldn't be validated. Now it poitns to the nearest segment below the validation boundary. The idea is you go look abive this and figure out if it is intentional that this segment isn't rendered in the current context. If so you can ignored the warning (it should ideally not even be presented as a normal error) but if it is unintentional you investigate why you aren't SSRing this slot or you mark the slot as intentionally blocking so it doesn't get validated.


packages/next/src/server/app-render/dynamic-rendering.ts
```diff
@@ -1449,10 +1449,33 @@ export function getNavigationDisallowedDynamicReasons(
...
-      error.name = 'Error'
-      error.message = message
+      const missingFiles: string[] = []
+      for (const [id, filePath] of boundaryState.requiredIds) {
+        if (!boundaryState.renderedIds.has(id) && filePath !== null) {
+          let normalized = filePath
+            .replace(/^\[project\][\\/]?/, '')
+            .replace(process.cwd() + '/', '')
+            .replace(process.cwd() + '\\', '')
+          missingFiles.push(normalized)
```
gnoff: @eps1lon seems like we don't have a standard util for this sort of thing but maybe I missed it or there is a different way to get a project local path for filenames?

acdlite (approved):

PR #93778, opened 2026-05-12
Wait for async page chunks before route-loader timeout in Turbopack

### LINKED ISSUE #93758: Pages Router Turbopack route loader times out before page registration despite successful chunks

https://github.com/Stanzilla/next-turbopack-pages-router-route-loader-repro

### To Reproduce

2. Run `pnpm install`.
3. Run `pnpm repro`.
4. Observe that the command exits non-zero after `pageLoader.loadPage()` rejects for several Pages Router routes.

`pnpm repro` runs a production Turbopack build, finds the `window.__NEXT_P.push` page-registration chunks, serves static assets from a separate `assetPrefix` origin, delays only those registration chunks, then runs a Playwright probe against `window.next.router.pageLoader.loadPage()`.

Example failing probe summary:

```json
{
  "pageRoute": "/track",
  "message": "Route did not complete loading: /track",
  "chunkLoaderCallsForRoute": 1,
  "nextPCallsForRoute": 0,
  "entrypointCallsForRoute": 0,
  "staticChunkResponses": 4,
  "nonOkResponses": 0,
  "failures": 0
}
```

A healthy control without delayed registration chunks is documented in the repository README and reports `chunkLoaderCallsForRoute: 1`, `nextPCallsForRoute: 1`, and `entrypointCallsForRoute: 1`.

Expected: `pageLoader.loadPage('/track')` should not fail once the route-list script has requested the route chunks and those static chunk requests succeed. The page should register through `window.__NEXT_P.push` / `routeLoader.onEntrypoint`, or the loader should continue/recover when the registration chunk arrives late. If a timeout is intended, the failure should distinguish a late registration chunk from a missing/failed chunk.

Actual: `pageLoader.loadPage('/track')` rejects with `Route did not complete loading: /track` even though the route chunk loader was invoked and all observed static chunk requests returned `200`. In the failure case, the route never registers through `window.__NEXT_P.push` and `routeLoader.onEntrypoint` is never called. In a real production app with large Turbopack Pages Router route graphs, this causes `next/link` client navigation to fall back to full document navigations.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.4.0: Thu Mar 19 19:31:56 PDT 2026; root:xnu-12377.101.15~1/RELEASE_ARM64_T6030
  Available memory (MB): 36864
  Available CPU cores: 12
Binaries:
  Node: 24.15.0
  npm: 11.14.1
  Yarn: 1.22.22
  pnpm: 10.33.4
Relevant Packages:
  next: 16.3.0-canary.19 // Latest available version is detected (16.3.0-canary.19).
  eslint-config-next: N/A
  react: 18.3.1
  react-dom: 18.3.1
  typescript: N/A
Next.js Config:
  output: N/A
```

Pages Router, Turbopack, Linking and Navigating, Runtime, Internationalization (i18n)

`next build (local)`, `next start (local)`, deployed production builds

### Additional context

The linked reproduction is intentionally deterministic: it delays only the page-registration chunks to reproduce the same route-loader failure signature observed in production. The production failure had the same shape without artificial delay: the route-list script called `__turbopack_load_page_chunks__`, all target static chunks returned `200`, but `window.__NEXT_P.push` and `routeLoader.onEntrypoint` never fired before `loadPage()` rejected.

The reproduction includes Pages Router, production Turbopack, i18n locale-prefixed URLs, `assetPrefix`, `deploymentId`, and a separate static asset origin. It is pinned to `next@16.3.0-canary.19`, where the issue still reproduces.

sokra:
### What?

In Turbopack production builds, `pageLoader.loadPage()` for the Pages Router can reject with

```
Route did not complete loading: <route>
```

even though the page's chunks were requested successfully and ultimately responded with `200`. This causes `next/link` client navigations to silently fall back to full document navigations.

### Why?

In Turbopack production, the per-route `/_next/static/.../pages/<route>.js` file is **not** the page itself — it is a tiny stub:

```js
__turbopack_load_page_chunks__("/track", [{ /* chunkData */ }, ...])
```

`__turbopack_load_page_chunks__` (installed in `packages/next/src/client/next-turbopack.ts`) calls `__turbopack_load__(...)` for each chunk and returns immediately. The chunks are then fetched asynchronously, and the page entry's `window.__NEXT_P.push([...])` registration lives **inside** those chunks.

`createRouteLoader` in `packages/next/src/client/route-loader.ts`:

1. Loads the page-loader stub via `<script>` injection,
2. Treats `script.onload` as "done loading scripts",
3. Starts a 3.8s timeout waiting for `whenEntrypoint(route)` to resolve via `__NEXT_P.push` → `routeLoader.onEntrypoint`.

If the actual page chunks take longer than 3.8s to download (slow network, large graph, cold edge cache, etc.) the timeout fires even though everything is succeeding — the chunks just haven't arrived yet, so the entrypoint isn't registered yet.

This contrasts with webpack production, where `next-client-pages-loader` inlines the entrypoint registration directly into the file the route loader is fetching, so `script.onload` already implies registration. Dev mode has its own gate (`devBuildPromise`), so this only affects Turbopack `next start`.

Original bug report: https://github.com/vercel/next.js/issues/93758 — the reproduction combines i18n, `assetPrefix`, and `deploymentId`, but those are incidental. Any chunk download exceeding 3.8s reproduces the failure.

### How?

The fix gives the route loader a way to know "the page's chunks are still being downloaded — don't start counting yet":

- `packages/next/src/client/next-turbopack.ts`
  - Eagerly create `window.__TURBOPACK_PAGE_CHUNK_PROMISES__ = new Map()`.
  - Have `__turbopack_load_page_chunks__` set `map[page] = Promise.all(chunkPromises)` after invoking `__turbopack_load__` for each chunk.
  - Eager initialization (rather than lazy `??=`) lets the route loader use the map's *presence* as a "this is a Turbopack build" detector.

- `packages/next/src/client/route-loader.ts`
  - In `loadRoute`, when `window.__TURBOPACK_PAGE_CHUNK_PROMISES__` is defined (i.e. Turbopack build), construct a `chunksLoaded` promise that resolves once both the loader script has finished and the corresponding `__TURBOPACK_PAGE_CHUNK_PROMISES__.get(route)` promise has settled.
  - Pass `chunksLoaded` as the delay argument to `resolvePromiseWithTimeout`; the 3.8s timer only starts ticking once chunks have finished downloading.
  - Delete the map entry after consumption so the map stays bounded.

- `packages/next/src/client/lib/promise.ts`
  - Rename the third parameter of `resolvePromiseWithTimeout` from `devPromise` to `delayPromise` and unify the dev/prod branches; the function already gracefully handled the `undefined` case, so production callers that pass `undefined` are unaffected.

Webpack production behavior is unchanged: `__TURBOPACK_PAGE_CHUNK_PROMISES__` is `undefined`, no `chunksLoaded` promise is constructed, and the route loader passes `undefined` for `delayPromise` exactly as before. Dev mode continues to gate on `devBuildPromise`.

### Test

`test/e2e/pages-router-route-loader-chunk-delay/` is a new e2e test that:

1. Loads the index page in a real browser via Playwright.
2. Installs a Playwright `page.route('**/_next/static/**/*.js*', ...)` handler that artificially delays every `/_next/static` JS response by 5s — i.e. longer than the route-loader's 3.8s deadline.
3. Calls `window.next.router.pageLoader.loadPage('/track')` directly (avoiding `router.push`, which masks the failure by triggering a hard reload on asset errors).
4. Asserts the call resolves rather than rejecting with `"Route did not complete loading: /track"`.

The test is scoped to turbopack `start` mode via `describe.skip` because the bug only applies there; webpack `start` and dev mode would either be unaffected or use a different gate.

Verified locally:
- Turbopack `start`: ✓ passes (~11s, reflecting the 5s chunk delay).
- Webpack `start`: properly skipped (no false pass).
- Existing `test/development/basic/asset-prefix` still passes in both bundlers.
- Reverting only the `route-loader.ts` and `next-turbopack.ts` changes reproduces the original failure with the documented error message.

Fixes #93758


Stanzilla:
@sokra 👀

Stanzilla:
@sokra I'd still love to see this fixed, want me to work on it if you don't have time?

Stanzilla:
@sokra poke :)

PR #1001, opened 2017-02-05
[POC] Pretty url routing

BDav24:
Hi, here is (yet another) POC to handle route customisation.
It can also handle reverse routing (feature I needed).

To see more details: https://github.com/BDav24/next-url-prettifier

PR #93794, opened 2026-05-12
ci: run Linux Playwright jobs in prebuilt Microsoft container

timneutkens:
## Summary

Runs Linux `build_reusable.yml` jobs that need Playwright inside Microsoft's prebuilt [`mcr.microsoft.com/playwright:v1.58.2-noble`](https://mcr.microsoft.com/en-us/product/playwright/about) image. The image already contains Chromium, Firefox, WebKit, and their system dependencies under `/ms-playwright`, so Linux test shards no longer spend each job downloading browsers and installing apt deps.

- Adds a `playwrightDockerImage` input, defaulting to the Ubuntu 24.04 (`-noble`) Playwright image so glibc matches the `ubuntu-latest` runners that build `next-swc.linux-x64-gnu.node`.
- Adds a `usePlaywrightContainer` input so jobs can opt out. `test-next-swc-wasm` opts out because its `afterBuild` step needs `rustup`, which is not in the Playwright image.
- Skips `pnpm playwright install --with-deps` for Linux jobs running in the image; Windows jobs and opted-out jobs still install Playwright at runtime.
- Enables the pnpm store cache inside container jobs because the container filesystem is ephemeral even when the host runner is persistent.

### Container condition

The Playwright image is used when all of the following are true:

- `inputs.needsPlaywright != 'no'`
- `inputs.skipInstallBuild != 'yes'`
- `inputs.usePlaywrightContainer != 'no'`
- The runner is not Windows

Otherwise `container.image` evaluates to `''` and the job runs directly on the runner.

### Compatibility fixes

The Microsoft image is browser-focused, so this PR adds a few CI setup fixes before the existing test workflow runs:

| Concern | Fix |
| --- | --- |
| Missing `jq` / `unzip` / native build toolchain | Install `jq`, `unzip`, and `build-essential` inside Linux container jobs. |
| Firefox refuses `$HOME` not owned by the current user | Re-own GitHub Actions' mounted `/github/home` directory for the root user used by the job. |
| Git sees the mounted workspace as dubious ownership | Mark `$GITHUB_WORKSPACE` as a safe directory for Git. |
| Standalone tests bind to Docker's generated `HOSTNAME` | Set container `HOSTNAME=0.0.0.0`. |
| `getStorageDirectory()` treats Docker storage as ephemeral | Add `NEXT_IGNORE_IS_DOCKER=1` for CI container jobs and honor it in `cache-dir.ts` so deterministic-build tests keep using `dist/cache`. |
| `localhost` IPv4/IPv6 mismatch in analyze tests | Let the analyze server listen on both IPv4 and IPv6 localhost addresses. |
| Nx fixture can race in CI before Next.js starts | Disable Nx isolated plugins for the `nx-handling` fixture scripts. |

## Timing

Earlier comparison runs showed the Playwright install/setup portion across 82 matched Linux jobs dropping from about **3 h 03 m total** on canary to about **39 m 24 s total** with the containerized setup, roughly **1 m 45 s saved per job** on average. Test execution itself is excluded from that comparison because shard balancing makes it noisy.

## Test plan

- [x] Latest PR CI run [`26121297957`](https://github.com/vercel/next.js/actions/runs/26121297957) completed successfully with 0 failed jobs.
- [x] Linux Playwright matrices are green, including `test dev`, `test prod`, `test turbopack *`, `test cache components *`, `test node streams *`, and changed-test lanes.
- [x] Linux non-Playwright jobs remain green on the bare runner, including `lint`, `types and precompiled`, `rust check`, `test cargo unit`, and `test unit`.
- [x] `test-next-swc-wasm` is green while opted out of the container via `usePlaywrightContainer: 'no'`.
- [x] Windows jobs are green and continue to use runtime Playwright installation.


scripts/check-playwright-image-version.mjs
```diff
@@ -0,0 +1,105 @@
...
+ * used by `.github/workflows/build_reusable.yml` for Linux CI jobs.
+ *
+ * Bumping `playwright` in `package.json` without updating the image tag would
+ * cause Playwright at runtime to look for a browser revision that does not
+ * exist in the container, with a confusing "Executable doesn't exist" error
+ * deep in the test run. Failing the lint job here surfaces the mismatch as a
+ * single, obvious CI error instead.
+ *
+ * Run via `node scripts/check-playwright-image-version.mjs`.
+ */
```
bgw: This seems over-engineered? The error described would only block PRs that try to update the version of playwright, which we don't do very often?

I don't think this check script is needed at all.

.github/workflows/build_reusable.yml
```diff
@@ -328,15 +480,37 @@ jobs:
...
+      # `container:` block above), which ships Chromium/Firefox/WebKit plus
+      # all system deps at `/ms-playwright`. We only need to verify the layout
+      # matches expectations so a mismatched image fails loudly here instead
+      # of mid-test.
+      - name: Verify preinstalled Playwright browsers
+        if: ${{ inputs.skipInstallBuild != 'yes' && inputs.needsPlaywright != 'no' && inputs.usePlaywrightContainer != 'no' && runner.os != 'Windows' }}
+        run: |
+          echo "PLAYWRIGHT_BROWSERS_PATH=${PLAYWRIGHT_BROWSERS_PATH:-<unset>}"
+          if [ ! -d /ms-playwright ]; then
+            echo "::error::Expected /ms-playwright to exist inside the Playwright container (image: ${{ inputs.playwrightDockerImage }}). Did the image tag drift from package.json?"
```
bgw: > Did the image tag drift from package.json?

That wouldn't explain the error here though? This would happen if the docker image was broken or Microsoft changed the filesystem structure of the container image.

I also don't really know that we need this verification step at all. It's *maybe* useful in debugging, but that seems about it.

We certainly don't need this `if` block and fancy error message: we can just let `ls /ms-playwright` exit with a non-zero code, which should fail this step. The `run` block uses `bash -e {0}` by default, which should exit if any command gives a non-zero exit code.

.github/workflows/build_reusable.yml
```diff
@@ -177,6 +246,70 @@ jobs:
...
+      # for the string "docker"; on Ubuntu 24.04 (cgroups v2) the cgroup
+      # file does not contain that string, so deleting `/.dockerenv` is
+      # enough to make the detection return false. This is scoped to
+      # root so it's a no-op on host runners.
+      - name: Disable in-container Docker detection (Linux container)
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          if [ "$(id -u)" = "0" ] && [ -e /.dockerenv ]; then
+            rm -f /.dockerenv
+          fi
```
bgw: It seems better to modify the detection code to support an environment variable override?

.github/workflows/build_reusable.yml
```diff
@@ -177,6 +246,70 @@ jobs:
...
+          if [ "$(id -u)" = "0" ]; then
+            MISSING=()
+            command -v jq >/dev/null 2>&1 || MISSING+=(jq)
+            command -v unzip >/dev/null 2>&1 || MISSING+=(unzip)
+            command -v make >/dev/null 2>&1 || MISSING+=(build-essential)
+            if [ ${#MISSING[@]} -gt 0 ]; then
+              apt-get -o DPkg::Lock::Timeout=60 update
+              apt-get -o DPkg::Lock::Timeout=60 install -y --no-install-recommends "${MISSING[@]}"
+            fi
+          fi
```
bgw: - If we know we're in the container always need these, we should just call `apt-get`. The `command` checks are over-engineered.
- Instead of using the upstream docker image, we can make a dockerfile that extends it, and as part of that installs these packages: https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support
- The `-o DPkg::Lock::Timeout=60` thing is only needed if you're installing packages outside of the docker container. Inside the docker container, we can be confident that there's no contention for the dpkg lockfile. (And probably it's not an issue with GH hosted runners either, it was more of a self-hosted runner issue)
- Why the check for root at the beginning of this?

.github/workflows/build_reusable.yml
```diff
@@ -150,6 +174,51 @@ jobs:
...
+        # whatever order `/etc/hosts` lists them. The Docker-managed
+        # `/etc/hosts` inside the Playwright container can list `::1`
+        # before `127.0.0.1` for `localhost`, so a `server.listen(0,
+        # 'localhost', …)` call ends up bound on `::1` while the test's
+        # subsequent `fetch('http://localhost:<port>')` resolves to
+        # `127.0.0.1` and gets ECONNREFUSED (e.g.
+        # `test/e2e/next-analyze/next-analyze.test.ts`). On bare
+        # runners (where `127.0.0.1` is listed first) the historical
+        # `ipv4first` ordering matches the resolved address, so this
+        # also keeps container behaviour aligned with the host runners.
```
bgw: We should always listen on both IPv4 and IPv6? Is this a bug in Next? Maybe just in the `experimental-analyze` subcommand?

.github/workflows/build_reusable.yml
```diff
@@ -177,6 +246,70 @@ jobs:
...
+      # launch with "Firefox is unable to launch if the $HOME folder
+      # isn't owned by the current user". Re-own the directory tree to
+      # the user we're actually executing as. Scoped to root so it's a
+      # no-op on host runners (which run as the `runner` user).
+      - name: Take ownership of /github/home (Linux container)
+        if: ${{ runner.os == 'Linux' }}
+        run: |
+          if [ "$(id -u)" = "0" ] && [ -d /github/home ]; then
+            chown -R root:root /github/home
+          fi
```
bgw, line 257: Playwright has documented a much cleaner solution that involves passing `--user` to docker: https://playwright.dev/docs/ci#via-containers

bgw (changes requested):

.github/workflows/build_reusable.yml
timneutkens: Turns out analyze had a bug, good catch!

.github/workflows/build_reusable.yml
timneutkens: :robot: Addressed by documenting why this reusable workflow intentionally keeps the Playwright container running as root: the job apt-installs jq/unzip/build-essential from the upstream image before bootstrapping. I also scoped the /github/home ownership fix to the Linux Playwright-container path so it only runs where that root/container behavior applies.

bgw: IMO, there are two better ways to solve this:
1. Define a custom dockerfile and run the apt steps there: https://docs.github.com/en/actions/reference/workflows-and-actions/dockerfile-support -- That may have additional caching benefits for those apt install steps.
2. OR: Keep running as root, but drop permissions before we run the e2e tests with `sudo -user="#1001"`

.github/workflows/build_reusable.yml
```diff
@@ -233,9 +312,28 @@ jobs:
...
-      # Cache pnpm store on GitHub-hosted runners (self-hosted runners have their own persistent storage)
+      # Defense in depth: inside the Playwright container we run as root
+      # and $GITHUB_WORKSPACE is mounted from the host (owned by a
+      # different UID), so Git emits "fatal: detected dubious ownership"
+      # unless `safe.directory` is configured. `actions/checkout` adds it
+      # already, and `container.env: HOME=/root` (set above) makes its
+      # JS-action HOME match the HOME of subsequent `bash -l` steps so
+      # they share the same `/root/.gitconfig`. This step re-adds the
+      # safe.directory from a login shell as a guard against future env
+      # changes that could desync those two HOMEs.
```
bgw: "Defense in depth" implies this is a security thing. It is not. I think we should get rid of this (see my comment above), but if we do not, just get rid of this giant comment. It's clear what the bash command is doing.

.github/workflows/build_reusable.yml
```diff
@@ -328,15 +426,21 @@ jobs:
       - run: ANALYZE=1 pnpm build
         if: ${{ inputs.skipInstallBuild != 'yes' }}
 
+      # Fallback path used when:
+      #   - The runner is Windows (no Linux container), or
+      #   - The job opted out via `usePlaywrightContainer: 'no'` because it
+      #     needs tooling not shipped in the container image (e.g. rustup).
+      # The stamp file at `$HOME/.cache/playwright-deps-*` lets persistent
+      # self-hosted runners skip the slow `--with-deps` step on subsequent
+      # jobs.
```
bgw: We don't have self-hosted runners anymore. Get rid of this comment and delete the code for managing the stamp file.

packages/next/src/build/analyze/index.ts
```diff
@@ -203,7 +203,13 @@ function startServer(dir: string, port: number): Promise<void> {
 
     server.on('error', onError)
 
-    server.listen(port, 'localhost', () => {
+    // Omit the host argument so Node binds on `::` (dual-stack on Linux,
+    // macOS, and Windows). This matches `next dev`/`next start`, which
+    // also leave the host unset (see `packages/next/src/cli/next-dev.ts`).
+    // Passing `'localhost'` here would force DNS resolution and bind to a
+    // single address — whichever `/etc/hosts` lists first — so a client
+    // that resolves `localhost` to the other family gets ECONNREFUSED.
```
bgw:
```suggestion
    // Listen on localhost (both IPv4 and IPv6)
```

test/e2e/app-dir/nx-handling/nx-handling.test.ts
```diff
@@ -13,9 +13,10 @@ describe('nx-handling', () => {
       private: true,
       packageManager: 'npm@10.9.2',
       scripts: {
-        build: 'rm -rf dist; nx run next-nx-test:build',
-        dev: 'nx run next-nx-test:dev',
-        start: 'nx run next-nx-test:serve:production',
+        build:
+          'rm -rf dist; NX_ISOLATE_PLUGINS=false nx run next-nx-test:build',
+        dev: 'NX_ISOLATE_PLUGINS=false nx run next-nx-test:dev',
+        start: 'NX_ISOLATE_PLUGINS=false nx run next-nx-test:serve:production',
```
bgw, line 21: Why do we need to set this environment variable?

.github/workflows/build_reusable.yml
```diff
@@ -101,6 +101,29 @@ on:
...
+        description: >
+          Docker image to use on Linux runners that need Playwright. Pinned to
+          the same version as the `playwright` dependency in `package.json`;
+          the image ships Chromium/Firefox/WebKit and all system deps
+          preinstalled, so the per-job install step is skipped on Linux.
+          Use the `-noble` (Ubuntu 24.04) variant so its glibc (>=2.39)
+          matches the `ubuntu-latest` host runners that build the
+          `next-swc.linux-x64-gnu.node` native binary; an older base (e.g.
+          `-jammy`) fails to dlopen the native binary at runtime with
+          `GLIBC_2.39 not found`.
```
bgw, line 114: Outside the scope of this PR: We need to stop using `ubuntu-latest` and specify an exact version of ubuntu, otherwise this is going to randomly break every 2 years when there's a new LTS.

I think that'll need help from IT though to define new runner types at the GitHub org level.

bgw (changes requested):

.github/workflows/build_reusable.yml
timneutkens: :robot: Removed the oversized safe-directory comment while keeping the explicit safe.directory step.

.github/workflows/build_reusable.yml
timneutkens: :robot: Removed the self-hosted runner stamp-file path and now install Playwright browsers with --with-deps directly in this fallback step.

packages/next/src/build/analyze/index.ts
timneutkens: :robot: Applied the shorter comment wording.

test/e2e/app-dir/nx-handling/nx-handling.test.ts
timneutkens: :robot: Added a short comment explaining this env var. It avoids the Nx isolated plugin worker race we saw in CI so this fixture keeps testing Next.js integration rather than Nx plugin isolation.

.github/workflows/build_reusable.yml
timneutkens: :robot: Agreed this is outside the scope of this PR, so I left the runner label as-is here.

bgw (approved):

PR #93795, opened 2026-05-12
Add act() testing diagnostics

timneutkens:
### What?

Adds diagnostic tooling for investigating flaky tests that use `createRouterAct`:

- An **opt-in watchdog** inside `createRouterAct` that periodically writes a labeled diagnostic to stderr identifying the stuck phase, in-flight RSC fetch URLs, and the test call site.
- `scripts/repro-flake.mjs`: a local helper that runs a single test in a loop with the watchdog auto-enabled and captures per-iteration logs for analysis.
- A new "Diagnosing Hangs (Watchdog)" section in the router-act skill documenting the diagnostic format, phase reference, and common patterns to interpret.

No production code or test fixture changes — this PR is purely test-tooling.

### Why?

When an `act` scope hangs in CI the failure surfaces as Jest's opaque `Exceeded timeout of 60000 ms for a test.`, with no indication of:

- which `act` call was stuck (in nested or repeated scopes),
- which phase inside `act` was blocked (`scope`, `wait-fetch-result`, `wait-browser-finished`, …),
- which RSC fetch URLs were in flight, or
- which test source line called the failing `act`.

That made root-causing a recent `app-prefetch/prefetching.test.ts` flake into hours of artifact spelunking. The watchdog converts those 60s opaque timeouts into actionable stderr output well before Jest gives up.

Example output (captured live during the original investigation that motivated this PR):

```
[router-act #2] scope blocked for 30999ms (threshold 2000ms)
  phase: wait-browser-finished (stuck 30999ms)
  phase meta: {"url":"http://localhost:.../prefetch-auto/foobar?_rsc=..."}
  in-flight RSC fetches: 0
  call site:
    at Object.act (.../prefetching.test.ts:333:31)
```

That diagnostic pinpoints the exact `act` call site and the stuck phase, which previously required reading CI artifacts, mirroring env vars, and instrumenting locally just to figure out where to look.

### How?

**Watchdog** — `test/lib/router-act.ts`

Tracks the current `act` phase (`scope`, `wait-first-request`, `wait-idle-callback-after-scope`, `wait-pending-checks-after-scope`, `wait-fetch-result`, `fulfill-response`, `wait-browser-finished`, `wait-redirect`, `wait-idle-callback-loop`, `wait-pending-checks-loop`, `cleanup`) and the URLs / elapsed time of any in-flight RSC fetches. A periodic timer writes a diagnostic to stderr when a phase stays stuck longer than the configured threshold.

**Off by default** — opt in with `ROUTER_ACT_WATCHDOG_MS=<ms>`. Slow CI jobs can have legitimately long phases (e.g. cold compile), so always-on warnings would be noise. Diagnostics are written via `process.stderr.write` (not `console.warn`) so they always reach the CI job log regardless of how the test runner / framework patches `console`.

**Repro helper** — `scripts/repro-flake.mjs`

Runs a single test N times via `pnpm test-<mode>-<bundler>`, writes each iteration's stdout/stderr to `scripts/repro-flake/<timestamp>/NN-{pass,fail}.log`, and writes a `summary.json` with pass rate / timing. The watchdog is enabled automatically with a tunable threshold (`--watchdog <ms>`, default 5000). Output directory is gitignored.

Usage:

```bash
node scripts/repro-flake.mjs \
  --test test/e2e/app-dir/app-prefetch/prefetching.test.ts \
  --grep "should immediately render the loading state" \
  --runs 25 --mode start --bundler webpack
```

**Docs**

- New "Diagnosing Hangs (Watchdog)" section in `.agents/skills/router-act/SKILL.md` with the phase reference table, sample output, and common patterns to look for.
- Pointer in `AGENTS.md` so future agents discover the tooling.

**Tests**

- `test/unit/router-act-watchdog.test.ts`: 4 unit tests covering watchdog enable/disable behavior, threshold gating, and the "disabled by default" case.


timneutkens:
Let's add this if we need to investigate in the future.

PR #93801, opened 2026-05-12
[App Shells] App shells in runtime prefetches

lubieowoce:
Reimplements app shells in runtime prefetches in terms of a staged render.

when `appShells` is on:
 - a `next-router-prefetch: 2` request (i.e. "including link data") will have a shell stage that only includes session data. we communicate the byte offset to the client so it can rewind the response to a session fallback
 - a `next-router-prefetch: 3` request (i.e. "session shell") stops rendering after the shell stages, i.e. link data is never resolved. this replaces the previous `forceOmitParams` implementation.
 
Note that In order to produce a session shell, we have to resolve static `params` and rootParams in the Runtime stage (previously, they'd resolve in the static stage). This is because params are link data and cannot be part of the session shell. We don't really need to rewind a runtime prefetch to a static stage, so this divergence in behavior shouldn't be meaningfully observable.

Notable call-out: root params getters can also be called inside caches. if a cache accesses root params, then the cache depends on link data and must be excluded from the shell. We rely on existing mechanisms for tracking root param access and delay the cache appropriately in the final render.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -255,8 +255,11 @@ import { ImageConfigContext } from '../../shared/lib/image-config-context.shared
 import { imageConfigDefault } from '../../shared/lib/image-config'
 import {
   getNextStage,
+  isAdvanceableRenderStage,
+  RENDER_STAGE_ADVANCE_ORDER,
   RenderStage,
   StagedRenderingController,
+  type AdvanceableRenderStage,
 } from './staged-rendering'
 import {
   anySegmentHasRuntimePrefetchEnabled,
@@ -283,6 +286,7 @@ import type {
 } from '../../build/segment-config/app/app-segment-config'
 import { ResponseCookies } from '../web/spec-extension/cookies'
 import { isInstantValidationError } from './instant-validation/instant-validation-error'
+import { createPromiseWithResolvers } from '../../shared/lib/promise-with-resolvers'
 
 export type GetDynamicParamFromSegment = (
   // The LoaderTree to extract the dynamic param from
@@ -397,7 +401,7 @@ function parseRequestHeaders(
   const isAppShellPrefetchRequest = headers[NEXT_ROUTER_PREFETCH_HEADER] === '3'
 
   // App Shell prefetches are a subtype of runtime prefetch — same code path,
-  // with `forceOmitParams` set on the prerender store.
+  // but with less resolved content (omitting link data)
   const isRuntimePrefetchRequest =
     headers[NEXT_ROUTER_PREFETCH_HEADER] === '2' || isAppShellPrefetchRequest
 
@@ -597,6 +601,7 @@ async function generateDynamicRSCPayload(
     skipPageRendering?: boolean
     staleTimeIterable?: AsyncIterable<number>
     staticStageByteLengthPromise?: Promise<number>
+    shellByteLengthPromise?: Promise<number | null>
     runtimePrefetchStream?: ReadableStream<Uint8Array>
   }
 ): Promise<RSCPayload> {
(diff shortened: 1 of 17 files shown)


lubieowoce:
* **#94442** <a href="https://app.graphite.com/github/pr/vercel/next.js/94442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94441** <a href="https://app.graphite.com/github/pr/vercel/next.js/94441?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93801** <a href="https://app.graphite.com/github/pr/vercel/next.js/93801?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93801?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94430** <a href="https://app.graphite.com/github/pr/vercel/next.js/94430?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/staged-rendering.ts
```diff
@@ -304,6 +416,10 @@ export class StagedRenderingController {
     }
   }
 
+  getFirstLateStage(): AdvanceableRenderStage {
+    return this.hasShells ? FIRST_SHELL_LATE_STAGE : FIRST_LATE_STAGE
+  }
```
gnoff: don't love this. feels very hard to reason about

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -1664,21 +1701,35 @@ async function generateRuntimePrefetchResult(
...
 
+  const mode: RuntimePrerenderMode = isAppShellsEnabled
+    ? isShellPrefetch
+      ? { type: 'session-shell-only' }
+      : {
+          type: 'rewindable-session-shell',
+          shellByteLengthDeferred: createPromiseWithResolvers(),
+        }
+    : { type: 'runtime-only' }
+
```
gnoff, line 1731: generally don't love these object args for internal functions. wdyt about passing in null | promise | sentinel object.

In the long rung isAppShellsEnabled is just always enabled so it would then become promise === rewindable and no promise === session-shell-only

Alternatively just pass isAppShellsEnabled in as well and do the logic to select runtime-only based on the flag

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -1943,38 +2019,85 @@ async function finalRuntimeServerPrerender(
...
+          // Resolve link data.
+          () => {
+            finalStageController.advanceStage(RenderStage.EarlyRuntime)
+          },
+          () => {
+            // Advance to Runtime stage: resolve cookies/headers for
+            // non-prefetchable segments. Sync IO is allowed here.
+            finalStageController.advanceStage(RenderStage.Runtime)
+          },
+        ]),
```
gnoff: I'd rather just make these last couple tasks noops when session-shell-only than do this iterable expansion thing.

Alternatively push tasks into an array and refactor the scheduler to accept an array of arbitrary length.

packages/next/src/server/app-render/app-render.tsx
lubieowoce: i don't like passing `isAppShellsEnabled` as a separate arg because then you get an invalid state where `isAppShellsEnabled = false` but `isShellPrefetch = true`. i know objects aren't your style but at least this way it's readable and we can do e.g.
```ts
finalStage:
      mode.type === 'session-shell-only'
        ? RenderStage.ShellRuntime
        : RenderStage.Runtime,
```
which is much more self explanatory than what we'd get with null/sentinel checks. i'd rather keep it this way if u don't mind

gnoff (approved):

PR #90959, opened 2026-03-06
Re-implement router-act with browser-based fetch instrumentation

acdlite:
## Summary

Re-implements the `router-act` test utility to use a browser-side fetch monkey-patch instead of Playwright's `page.route()` API. The old approach was flaky due to issues with streaming responses, redirect handling, and execution context destruction.

The new implementation (`@next/router-act`) is a private workspace package with three files:
- `setup.ts` — browser-side runtime that intercepts fetch calls and runs the drain loop
- `index.ts` — thin Playwright orchestrator for config parsing, error formatting, and assertions
- `component.tsx` — `<RouterAct />` client component rendered in test fixture layouts

The `<RouterAct />` component installs the fetch patch via `useEffect` so it doesn't interfere with hydration. This works because router-act is only used in test fixtures we control.

All 30 test files have been migrated. The old implementation is preserved at `test/lib/deprecated-router-act.ts` for a handful of tests that trigger hard navigations destroying the browser context mid-act — those will be reworked in a follow-up PR.

## Test plan

- All migrated test files pass locally with `NEXT_SKIP_ISOLATE=1 NEXT_TEST_MODE=start`
- Tests covering: basic prefetch/navigation, nested act scopes, blocking, response ordering, forbidden responses, server actions, interception routes, memory pressure (200+ requests), cache staleness, revalidation, deployment skew, and more

PR #93802, opened 2026-05-12
Patch `playwright-core` to resolve `_finishedPromise` on `requestFailed`

unstubbable:
The client-side `BrowserContext._onRequestFailed` handler in `playwright-core` does not resolve the corresponding `Response._finishedPromise`, while `_onRequestFinished` does. When Chromium fires `Network.loadingFailed` for a request that already received a response — e.g. an RSC response that was delivered to the response buffer but is no longer needed because a soft navigation has already committed — the server-side `_finishedPromise` is resolved correctly (in `chromium/crNetworkManager._onLoadingFailed`), but the client side is never told. Any caller of `response.finished()` on such a response hangs indefinitely, surfacing in tests as a 60s Jest timeout with no clear stack trace.

This is the root cause of intermittent timeouts in router-act-using tests.

The patch mirrors the resolve step from `_onRequestFinished`: look up the response via the existing async `request.response()` RPC and fire-and-forget `resolve(null)` on its `_finishedPromise`. `catch` is a no-op so a tearing-down context doesn't produce an unhandled rejection.

Verified empirically against workflow run https://github.com/vercel/next.js/actions/runs/25760869258, which flake-detected all 40 router-act consumers (each 3 times) with diagnostic logging enabled. One `Network.loadingFailed canceled=true` event fired during the run (the canonical bug scenario: an outer `act` releasing a child `block: true` response that Chrome canceled mid-flight; see https://github.com/vercel/next.js/actions/runs/25760869258/job/75669131568) — pre-patch this would have hung until Jest's 60s test timeout fired, post-patch `response.finished()` resolved in 1ms. Alongside ~4000 normal resolutions, with zero `wait-browser-finished` hangs and zero router-act-related test failures.

This is a stop-gap until the router-act rewrite in #90959 lands, which replaces Playwright's `route.fulfill` + `response.finished()` flow with a browser-side `fetch` interception and removes the dependency on this Playwright code path entirely. The patch can be dropped at that point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -345,7 +345,8 @@
       "@rspack/core@1.6.7": "patches/@rspack__core@1.6.7.patch",
       "@modelcontextprotocol/sdk": "patches/@modelcontextprotocol__sdk.patch",
       "@vercel/blob": "patches/@vercel__blob.patch",
-      "postcss-scss": "patches/postcss-scss.patch"
+      "postcss-scss": "patches/postcss-scss.patch",
+      "playwright-core@1.58.2": "patches/playwright-core@1.58.2.patch"
     }
   }
 }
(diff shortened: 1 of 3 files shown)


lubieowoce (approved):

lubieowoce:
damn, nice find

PR #93805, opened 2026-05-12
make rcstrs on the heap/static slightly smaller

lukesandberg:
### What?

Splits the previously unified `PrehashedString` (which held a `Payload` enum of `String | &'static str`) into two separate types: `StaticPrehashedString { value: &'static str, hash: u64 }` for atoms produced by `rcstr!` / `make_const_prehashed_string`, and `DynamicPrehashedString { value: Box<str>, hash: u64 }` for atoms held in an `Arc`. The static and dynamic paths were already distinguished by the `STATIC_TAG` / `DYNAMIC_TAG` bits in `RcStr`, so the runtime branch on enum discriminant was redundant.

### Why?

- **16 bytes saved per heap-allocated `RcStr` value .** Dynamic atoms drop the `String::capacity` field (which was always equal to `len` since the contents are immutable) and because they are stored in a `triomphe::Arc` this drops the Arc payload to 32 bytes instead of 40 which matches a mimalloc bucket (previously we were rounded up to a 48 byte bucket) so we save 16 bytes.
- **8 bytes saved per static `RcStr` value.** The linker will optimally align our 24 byte struct.
- Removes a layer of dispatch on the hot path (`as_str`, `==`, `Hash`) — typed deref to the correct variant instead of matching on `Payload`. (i.o.w. one 'descreminent' traversal instead of 2)

### How?

- `dynamic.rs`: `Payload` enum removed; two structs replace `PrehashedString`. `deref_from` split into `deref_static` and `deref_dynamic`. `restore_arc` returns `Arc<DynamicPrehashedString>`.
- `lib.rs`: `as_str` and `into_owned` dispatch on `tag()` (STATIC vs DYNAMIC vs INLINE) rather than `location()`. New `heap_hash_and_str` helper for `PartialEq` and `Hash` to share the static/dynamic branch. `into_owned`'s `try_unwrap` arm uses `String::from(Box<str>)` which reuses the box allocation (still O(1)).
- `turbo-rcstr-macros`: emit `::turbo_rcstr::StaticPrehashedString` instead of `::turbo_rcstr::PrehashedString`.
- Added a comment on `DynamicPrehashedString` noting the future move to `triomphe::ThinArc` to fold the two heap allocations (Arc header + boxed bytes) into one. Deferred because that change would make `RcStr::from(String)` copy the bytes, invalidating the documented cheap `String -> RcStr -> String` round-trip — wants a separate evaluation.

### DIFF
--- a/turbopack/crates/turbo-rcstr/src/dynamic.rs
+++ b/turbopack/crates/turbo-rcstr/src/dynamic.rs
@@ -63,6 +63,8 @@ pub(crate) fn new_atom<T: AsRef<str> + Into<String>>(text: T) -> RcStr {
     let hash = hash_bytes(text.as_bytes());
 
     let prehashed = DynamicPrehashedString {
+        // NOTE: This will capture as a Box<str> which will essentially
+        // `shrink_to_fit` the bytes.
         value: text.into(),
         hash,
     };
--- a/turbopack/crates/turbo-rcstr/src/dynamic.rs
+++ b/turbopack/crates/turbo-rcstr/src/dynamic.rs
@@ -7,49 +7,41 @@ use crate::{
     is_atom_inlineable, tagged_value::TaggedValue,
 };
 
-pub enum Payload {
-    String(String),
-    Ref(&'static str),
-}
-
-impl Payload {
-    pub(crate) fn as_str(&self) -> &str {
-        match self {
-            Payload::String(s) => s,
-            Payload::Ref(s) => s,
-        }
-    }
-    pub(crate) fn into_string(self) -> String {
-        match self {
-            Payload::String(s) => s,
-            Payload::Ref(r) => r.to_string(),
-        }
-    }
-}
-impl PartialEq for Payload {
-    fn eq(&self, other: &Self) -> bool {
-        self.as_str() == other.as_str()
-    }
-}
-
-pub struct PrehashedString {
-    pub value: Payload,
+/// Read-only header for atoms allocated in static storage by the `rcstr!`
+/// macro. The value lives for `'static`, so we store a `&'static str` rather
+/// than owning the bytes.
+pub struct StaticPrehashedString {
+    pub value: &'static str,
     /// This is not the actual `fxhash`, but rather it's a value that passed to
     /// `write_u64` of [rustc_hash::FxHasher].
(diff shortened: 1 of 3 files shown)


lukesandberg:
* **#93805** <a href="https://app.graphite.com/github/pr/vercel/next.js/93805?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93805?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-rcstr/src/dynamic.rs
```diff
@@ -70,20 +62,20 @@ pub(crate) fn new_atom<T: AsRef<str> + Into<String>>(text: T) -> RcStr {
 
     let hash = hash_bytes(text.as_bytes());
 
-    let prehashed = PrehashedString {
-        value: Payload::String(text.into()),
+    let prehashed = DynamicPrehashedString {
+        value: text.into(),
```
sokra, line 68: This might clone the String here if it has more capacity than len. So it basically runs `shrink_to_fit()` here.
That's probably fine, but worth a comment.

Note that small strings probably doesn't reallocate due to mimalloc size classes anyway.

bgw (approved):

sokra (approved):

turbopack/crates/turbo-rcstr/src/dynamic.rs
```diff
@@ -70,20 +62,20 @@ pub(crate) fn new_atom<T: AsRef<str> + Into<String>>(text: T) -> RcStr {
...
         hash,
     };
```
lukesandberg, line 70:
```suggestion
    let prehashed = DynamicPrehashedString {
        // NOTE: This will capture as a Box<str> which will essentially
        // `shrink_to_fit` the bytes.
        value: text.into(),
        hash,
    };
```

turbopack/crates/turbo-rcstr/src/dynamic.rs
lukesandberg: dropped a comment

PR #93807, opened 2026-05-13
Stop pinning compiled chunk source on `EcmascriptBuildNodeChunkVersion`.


lukesandberg:
## What

Stop pinning compiled chunk source on `EcmascriptBuildNodeChunkVersion`.

The struct previously held a `Vec<ReadRef<CodeAndIds>>` for every module in the chunk, even though the HMR update path only needs the bytes for the *changed* subset and the bytes are already on disk. That field was also forcing the upstream `CodeAndIds` / `BatchGroupCodeAndIds` tasks to stay `serialization = "skip"`, so warm restarts had to re-walk every module and re-hash its source to rebuild `entries_hashes`.

This change mirrors the browser-side pattern: a new `EcmascriptBuildNodeChunkContentEntries` task lives on the chunk content and holds `ResolvedVc<Code>` + `ResolvedVc<u64>` per module. The version struct shrinks to `{ chunk_path, minify_type, entries_hashes }`, drops `serialization = "skip"`, and switches `chunk_path` from `String` to `RcStr`.
`update_ecmascript_node_chunk_content` now resolves entries lazily, only when an added or modified module actually needs its code shipped.

## Why

Two wins, both for dev sessions starting against a warm filesystem cache:

- **Memory.** The version no longer transitively pins every module's compiled   `Rope` in heap — those bytes can stay on disk until HMR actually needs them.
- **Warm-restart CPU.** `entries_hashes` is sourced from the per-module `Code::source_code_hash()` task (already cached) and the version itself now round-trips through the persistent cache, so we don't re-hash anything on warm start.

The HMR payload shape is unchanged.

## Perf
This should speed up warm builds a bit but the major benefit is not recomputing node outputs and keeping them in ram

measuring v0 after loading the main route
Branch: Cold: 12.3G Warm: 7.34G
Canary: Cold 12.3G Warm: 8.5G

The trace file confirms the recomputations are gone and the heap measurements confirm we trimming ~1.1g of ram

Using the devlow benchmarks i was able to confirm a possible small progression

```
# canary
chat dev startup build=warm: root page = 23.96 s (from root page/start)
chat dev startup build=warm: root page = 21.21 s (from root page/start)
chat dev startup build=warm: root page = 22.70 s (from root page/start)

# branch
chat dev startup build=warm: root page = 20.94 s (from root page/start)
chat dev startup build=warm: root page = 19.43 s (from root page/start)
chat dev startup build=warm: root page = 23.49 s (from root page/start)

```

## Tests

I added a new integration test to ensure we don't accidentally regress here.  Which confirms that 'clean warm builds' run nothing and 'clean warm dev sessions' just set up HMR session infra


lukesandberg:
* **#93807** <a href="https://app.graphite.com/github/pr/vercel/next.js/93807?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93807?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/task_statistics.rs
```diff
@@ -71,9 +71,19 @@ impl Serialize for TaskStatistics {
     where
         S: Serializer,
     {
-        let mut map = serializer.serialize_map(Some(self.inner.len()))?;
-        for entry in &self.inner {
-            map.serialize_entry(entry.key().ty.global_name, entry.value())?;
+        // Sort by `global_name` so the emitted JSON is deterministic — the
+        // underlying `FxDashMap` has unspecified iteration order. The map is
```
wbinnssmith, line 75: don't we have an abstraction for sorted maps?

turbopack/crates/turbopack-ecmascript/src/chunk/content_entry.rs
```diff
@@ -18,36 +19,35 @@ use turbopack_ecmascript::chunk::{
 /// computing updates.
 #[turbo_tasks::value]
 #[derive(Debug)]
-pub struct EcmascriptDevChunkContentEntry {
+pub struct EcmascriptChunkContentEntry {
```
wbinnssmith, line 22: 🙏

wbinnssmith (approved):

turbopack/crates/turbo-tasks/src/task_statistics.rs
lukesandberg: BtreeMap or FrozenMap

We could use BTreemap but the expensive part is actually updating this map which needs the high concucrrency random access of DashMap.  So doing a monolithic sort during serialization is better than using a bunch of O(log-N) updates.

e.g. each entry might be incremented 10K times, and we would pay the log-N on each of those so it is really N-logN*K where K is the average number of task executions.  which is around 7000 for vercel-site.

sorting at the end will be faster despite the extra allocation

PR #93821, opened 2026-05-13
fix: GNU xargs warns about incompatible --replace/--max-args

lubieowoce:
we were invoking

```
xargs --null -n1 -I'{}' ...
```
to tell xargs that we want to:
- use 1 argument per invocation (`-n1`)
- `{}` is the placeholder (`-I'{}'`)

but it seems like `-I'{}'` already implies `-n1`. BSD xargs doesn't care, but GNU xargs warns about it, so we get this in CI (e.g. [here](https://github.com/vercel/next.js/actions/runs/25789455488/job/75751231855?pr=93801#step:36:54)):
```
xargs: warning: options --max-args and --replace/-I/-i are mutually exclusive, ignoring previous --max-args value
```

seems like both GNU and BSD xargs are happy if we just drop the `-n1` and only use `-I{}`.


bgw (approved):
@bgw approved this PR from Slack with [Graphite](https://app.graphite.com/github/pr/vercel/next.js/93821)

PR #93826, opened 2026-05-13
Fix warning in new apps caused by incorrect image dimensions

samselikoff:
Fixes the default create-next-app templates so the Vercel SVG logomark uses dimensions that match its intrinsic aspect ratio, avoiding the next/image development warning on freshly generated apps. Adds a regression test that creates the default app, runs it in dev mode, loads it in a browser, and asserts there are no browser warnings or errors.


timneutkens (approved):
Missing PR description. Changes look good though.

timneutkens (changes requested):
We run the application so it's possible to add a test so that it doesn't regress in the future.

timneutkens (approved):

PR #93827, opened 2026-05-13
[CC] fix: cachedNavigations missing asyncApiPromises in resumes

lubieowoce:
When rendering dynamically for `cachedNavigations`, we do a staged render so that we can recover a static prefetch from the stream. However, we weren't delaying params properly because we were missing `asyncApiPromises` and `fallbackParams` from the store, so dynamic params would incorrectly resolve in the static stage.

This was masked in the tests, which accidentally had a bunch of `await setTimeout(300)` delays left in. so components that were meant to test when params resolve were always resolving in the dynamic stage, because timeouts are tasky.

I'm not confident in the fix for missing `fallbackParams`, it's robot-assisted and i'm not very familiar with those bits, but it seems to work

### DIFF
--- a/packages/next/src/build/templates/app-page.ts
+++ b/packages/next/src/build/templates/app-page.ts
@@ -1922,14 +1922,29 @@ export async function handler(
       const transformer = new TransformStream<Uint8Array, Uint8Array>()
       body.push(transformer.readable)
 
+      // Plumb fallback params via request meta so the RequestStore created
+      // downstream in app-render.tsx knows which params to defer during the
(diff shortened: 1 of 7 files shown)


lubieowoce:
* **#93827** <a href="https://app.graphite.com/github/pr/vercel/next.js/93827?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93827?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

unstubbable (approved):

Issue #77449, opened 2025-03-24
Playwright tests with the `next` fixture always fail (using experimental testProxy)


GRA0007:
### Link to the code that reproduces this issue

https://github.com/GRA0007/next-test-proxy-repro

### To Reproduce

1. Set up a test with Playwright that uses the `NextFixture` to mock a request (following instructions from [here](https://github.com/vercel/next.js/tree/canary/packages/next/src/experimental/testmode/playwright#use-the-nextexperimentaltestmodeplaywright-to-create-tests))
2. Run the tests, notice that if the `NextFixture` is used, the test fails with the error `TypeError: Cannot destructure property 'complete' of 'undefined' as it is undefined.`

### Current vs. Expected behavior

I expect mocking a request to successfully mock the request without causing an error. All the tests I've set up in my reproduction should pass.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Mon Aug 12 20:51:54 PDT 2024; root:xnu-11215.1.10~2/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
Binaries:
  Node: 20.11.1
  npm: 10.2.4
  Yarn: 1.22.21
  pnpm: 10.6.5
Relevant Packages:
  next: 15.2.3 // Latest available version is detected (15.2.3).
  eslint-config-next: N/A
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.8.2
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Testing

### Which stage(s) are affected? (Select all that apply)

next dev (local), next start (local)

### Additional context

I initially started making this reproduction because I notice that the `testProxy` didn't work with `GET` route handlers, only with pages, but it seems something else is going on because I can't get any of the mocked tests to work.

GRA0007:
Fixed by #77659

PR #93831, opened 2026-05-13
fix(testmode): preserve request body on passthrough

gmeligio:
Related to: #82913, #77449

When `experimental.testProxy` is enabled, any server-side `POST`, `PUT`, or `PATCH` request that is not mocked by a test handler reaches the real server with an empty body. This bug has been present since `testProxy` was added in `14.2.0-canary.36`.

## Why it happens

1. Next.js reads the full request, including its body, to send it to the test proxy.
2. The proxy decides whether to mock the request or pass it through.
3. If the proxy says "pass it through" (the `'continue'` case), Next.js forwards the original request to the real server.

The problem is in step 3: a request body in JavaScript can only be read once. Step 1 already drained it, so the request that reaches the real server has an empty body.

## Fix

1. Clone the request **before** reading the body in step 1.
2. On `'continue'`, forward the clone instead of the original.

## Tests

- [x] New test in `test/e2e/testmode/testmode.test.ts` that sends a POST with a JSON body through a route that triggers the `'continue'` case. It checks that a local echo server receives the full body.
- [x] Confirmed the new test fails on `canary` without the fix and passes with it.


jitendra-bhandure:
Hi @gmeligio, please merge and release this changes, also in [next.config.js docs](https://nextjs.org/docs/app/api-reference/config/next-config-js) not mention or not added testProxy related documents, So we miss or intentionally removed, are we not going to continue this ?

jitendra-bhandure:
Hi @gmeligio,  sometimes getting error like: `Failed to load resource: A TLS error caused the secure connection to fail.`

jitendra-bhandure:
Hi @gmeligio, i used playwright test, in that my test passed but in terminal showing some error like `Error: Proxy request aborted` there should be way/flag to stop such things.

Also in --ui mode some extra step are showing like `next.onFetch` for every api this should be some flag base so as per need use set it.

**Thanks in Advance**

gmeligio:
Hi @jitendra-bhandure . Thank you so much for reviewing. I'll answer your comments below. Thank you for looking into this.

> please merge and release this changes

I want to merge it. I need a review and merge from the maintainers because I don't have permissions to do it myself.

> in [next.config.js docs](https://nextjs.org/docs/app/api-reference/config/next-config-js) not mention or not added testProxy related documents, So we miss or intentionally removed

`testProxy` is marked as experimental so it's only [documented in the repo](https://github.com/vercel/next.js/blob/canary/packages/next/src/experimental/testmode/playwright/README.md) and not on the website.

---

Regarding the testing errors, thank you for testing! Could you please also run the same tests from the [canary branch](https://github.com/vercel/next.js/tree/canary) to confirm it's already present in canary 🙏 ? 
1. TLS errors: `sometimes getting error like: Failed to load resource: A TLS error caused the secure connection to fail.`
2. Request aborted: `i used playwright test, in that my test passed but in terminal showing some error like Error: Proxy request aborted there should be way/flag to stop such things.`
3. Verbose `next.onFetch`: `Also in --ui mode some extra step are showing like next.onFetch for every api this should be some flag base so as per need use set it.`

The fixes for these errors can be made in separate PRs/issues if you want to. I want to keep this PR's scope small to just fix empty body on request passthrough.

gmeligio:
Would it be possible to review this small fix? Maybe @dvoytenko or @agadzik since you've contributed in the past to testProxy.

jitendra-bhandure:
Hi @dvoytenko or @agadzik, Please review it

jitendra-bhandure:
Hi @gmeligio, can you add them and other people reviewers so somebody will review it.

gmeligio:
Hi @jitendra-bhandure . I don't have permissions to add them. I'm not sure who else can review it.

image

gmeligio:
@jitendra-bhandure, I created [this post in the NextJS discord](https://discord.com/channels/752553802359505017/1522935296671613070) to get help with reviewing. Feel free to discuss there too.

PR #93833, opened 2026-05-13
devlow: add run sampling snapshots and comparison reports

wbinnssmith:
A lot of this was written with Claude. We lean a lot on published libraries for the statistics, so I've got confidence in them.

This adds top-level commands to devlow, `run`, and `compare`, though we're still backwards compatible with `run` being the default if no command is specified.

- `--n=<N>` runs each variant N times, `--warmup=<N>` discards the first N runs (don't use on cold-start metrics). Default behavior is unchanged at n=1.
- Every run now writes a snapshot CSV to `.devlow-bench/snapshots/<ts>.csv` (long format, one row per sample, so mixed-unit metrics share a file). Override with `--snapshot=<path>`.
- `--compare` prints a side-by-side table at the end of a run against the newest snapshot. `--baseline=<path>` points at a specific file or directory.
- `devlow-bench compare <baseline.csv> <current.csv>` is a separate subcommand for diffing two existing snapshots without rerunning anything.

For every scenario/variant/metric combination, it shows baseline vs current `mean / p50 / p90 (n)`, the absolute and percent delta, and two p-values.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2004,6 +2004,9 @@ importers:
       inquirer:
         specifier: ^9.2.7
         version: 9.2.23
+      jstat:
+        specifier: ^1.9.6
+        version: 1.9.6
       minimist:
         specifier: ^1.2.8
         version: 1.2.8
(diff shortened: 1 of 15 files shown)


sokra (approved):
I do not known enough math to review that, but I trust Claude, so lets go

wbinnssmith:
> but I trust Claude

Thankfully this leans on `jstat` for the actual stats, so no vibecoded statistics..

PR #93835, opened 2026-05-13
turbo-tasks: replace TaskStorage TinyVec lazy area with bitmap + byte-packed tail in one allocation

lukesandberg:
### What?

Replaces TaskStorage's `TinyVec<LazyField>` lazy area with a presence bitmap plus a byte-packed payload tail that lives in the same heap allocation as the inline head. The result is one allocation per task instead of two, no per-entry discriminant tax, and the same 8-byte machine-word footprint for an owning `TaskStorage` (still a thin pointer).

### Why?

`TinyVec<LazyField>` was paying:
- Per-entry 8-byte discriminant + alignment padding on every stored variant (~12B wasted per slot at p99=7 entries × 6.5M tasks ≈ 200MB upper bound).
- A second heap allocation per task for the `TinyVec` buffer, with its own allocator metadata and fragmentation cost.
- Pointer chasing from the head into the separately-allocated lazy buffer (extra cache miss on every lazy access).

We can't afford to make `Box<TaskStorage>` a fat pointer (the DashMap stores 6.5M of them; fat pointers would add ~70MB of map overhead and erode the savings), so the new layout keeps the head sized and lays the tail bytes out *immediately after* the head in the same allocation, accessed via raw pointer arithmetic from `(self.ptr).add(size_of::<TaskStorageInner>())`. `TaskStorage` is a 1-word `NonNull<TaskStorageInner>` and `Option<TaskStorage>` still niches.

### How?

The PR is intentionally many small commits — each a green build — so the review can follow the migration:

1. **Encapsulate**: move all `lazy.push` / `lazy.iter` call sites behind typed accessors so the lazy storage representation is opaque to the rest of the crate.
2. **Macro emits tag tables**: `LAZY_TAG_<NAME>: Tag`, `LAZY_SIZE`, `LAZY_ALIGN`, `LAZY_PADDED_SIZE`, `LAZY_*_MASK`, plus `lazy_drop_dispatch` / `lazy_is_empty_dispatch` / `lazy_debug_dispatch` per-tag match arms.
3. **Swap the storage**: drop `LazyField`/`TinyVec`, introduce `LazyTail { present: u32, len: u16, cap: u16 }` (8B head metadata) addressing a tag-ordered byte buffer.
4. **Unify the allocation**: the tail bytes move into `TaskStorage`'s own heap block. `TaskStorageBox` (now `TaskStorage`) owns alloc / dealloc / grow / shrink with mimalloc-bin awareness (`mi_good_size`).
5. **Polish**: `Tag(NonZeroU8)` newtype with `bit/mask_below` helpers, schema-emitted `LAZY_TYPE_IDS` table so `Tag::debug_assert_type::<T>()` catches wrong-type casts in debug, custom `Debug` impl that pretty-prints each present lazy variant by name. Reorder lazy variants `Meta → Data → Transient` so transient churn shifts the fewest bytes. Switch `encode_meta/data` to a single bitmap walk via `LazyTail::try_for_each_present`.

Miri-clean under default Stacked Borrows mode for the 25-test `backend::lazy_tail` + `backend::storage_schema` suites. Clippy-clean.

### Benchmarks

5 runs of a large Next.js production build (`pnpm next build --experimental-build-mode=compile`, `TURBOPACK_PERSISTENT_CACHE=0`, `nice -n -20`, Spotlight + low-power mode disabled).

| metric | upstack baseline (`05-09-...-16-b-lazyvec`) | this PR | delta | ~t |
|---|---|---|---|---|
| wall (s) | 39.84 ± 0.90 | 40.03 ± 2.27 | +0.19s (+0.5%) | +0.17 |
| user CPU (s) | 278.80 ± 1.32 | **275.51 ± 1.15** | **−3.29s (−1.18%)** | **−4.19** |
| sys CPU (s) | 67.92 ± 3.67 | 69.26 ± 3.02 | +1.34s (+2.0%) | +0.63 |
| MaxRSS | 11.920 GB ± 57 MB | 11.902 GB ± 77 MB | −19 MB (−0.15%) | −0.44 |

The flattened-storage wall-time mean is dragged up by a single 44.07s outlier in run 3; the other four runs average 39.02s ± 0.27s.

**Read of the numbers**:
- **Small CPU win** (~3s, statistically clean at this sample size) — consistent with the more cache-friendly layout: one allocation hot in cache instead of head + lazy buffer, no per-entry discriminant chasing, presence bitmap iteration touches one cache line.
- **Small memory win** at peak RSS from better packing — within noise here because MaxRSS at this build's peak is dominated by parsing/bundling phases that don't touch task storage. The per-task savings would show up in a "size of the task graph at snapshot time" measurement, not at peak.
- We pay for the better layout with a bit more allocator work (slight sys CPU bump, also within noise) — the byte tail grows/shrinks via `realloc` instead of `TinyVec`'s amortized `Vec` push, but that cost is unavoidable for the layout to work.

More samples would tighten the wall/RSS confidence intervals, but the CPU signal is already clean.

### Follow-ups not in this PR

The real upside from this layout flexibility is letting `AutoSet` / `AutoMap` keep more of their content inline now that the lazy variant size isn't forced into a fixed `LazyField` enum slot. This PR doesn't change those types' inline sizes — it just removes the constraint that was preventing it.


lukesandberg:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/vercel/next.js/93835?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#93835** <a href="https://app.graphite.com/github/pr/vercel/next.js/93835?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93835?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#93720** <a href="https://app.graphite.com/github/pr/vercel/next.js/93720?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks-backend/src/backend/lazy_tail.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// Install a payload for `tag`. Caller must have already ensured the
+    /// buffer has enough capacity (`self.cap - self.len >= padded_size`).
+    ///
+    /// # Safety
+    /// - `T` must match the payload type for `tag`.
+    /// - The variant must NOT already be present (caller should `take` it first to replace).
+    /// - `tail_base` must point to a buffer of at least `self.cap` bytes.
+    pub(crate) unsafe fn install_unchecked<T>(&mut self, tag: u8, value: T, tail_base: *mut u8) {
+        debug_assert!(!self.has(tag), "lazy variant tag {tag} already present");
+        let bit = 1u32 << (tag - 1);
```
lukesandberg: we should have a newtype for tag

to improve readability and then we can just add methods to compute masks

turbopack/crates/turbo-tasks-backend/src/backend/lazy_tail.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    /// Sum of `LAZY_PADDED_SIZE[k+1]` over every set bit `k` in `mask`.
+    #[inline]
+    pub(crate) fn sum_padded_sizes(mut mask: u32) -> usize {
+        let mut offset = 0usize;
+        while mask != 0 {
+            let bit_idx = mask.trailing_zeros();
+            // `bit_idx` is in 0..32; tag is `bit_idx + 1`, in 1..=LAZY_N.
+            offset += LAZY_PADDED_SIZE[bit_idx as usize + 1] as usize;
+            mask &= mask - 1;
```
lukesandberg, line 200: this should align with https://lemire.me/blog/2018/02/21/iterating-over-set-bits-quickly/ and reference it

lukesandberg: in fact i wonder if there is a stdlib fn for this

turbopack/crates/turbo-tasks-backend/src/backend/lazy_tail.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /// Install a payload for `tag`. Caller must have already ensured the
+    /// buffer has enough capacity (`self.cap - self.len >= padded_size`).
+    ///
+    /// # Safety
+    /// - `T` must match the payload type for `tag`.
+    /// - The variant must NOT already be present (caller should `take` it first to replace).
+    /// - `tail_base` must point to a buffer of at least `self.cap` bytes.
+    pub(crate) unsafe fn install_unchecked<T>(&mut self, tag: u8, value: T, tail_base: *mut u8) {
```
lukesandberg: insert_unchecked would be a better name

turbopack/crates/turbo-tasks-backend/src/backend/lazy_tail.rs
```diff
@@ -0,0 +1,230 @@
...
+        self.present &= !bit;
+        Some(value)
+    }
+
+    /// Install a payload for `tag`. Caller must have already ensured the
+    /// buffer has enough capacity (`self.cap - self.len >= padded_size`).
+    ///
+    /// # Safety
+    /// - `T` must match the payload type for `tag`.
+    /// - The variant must NOT already be present (caller should `take` it first to replace).
```
lukesandberg: this should reference `replace_in_place` which is more efficient

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -3787,7 +3791,7 @@ fn encode_task_data(
     data.encode(category, &mut encoder)?;
 
     if cfg!(feature = "verify_serialization") {
-        TaskStorage::new()
+        crate::backend::task_storage::TaskStorage::new()
```
lukesandberg: import this

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -190,12 +190,12 @@ impl<'e, B: BackingStorage> ExecuteContextImpl<'e, B> {
         &self,
         task_id: TaskId,
         category: SpecificTaskDataCategory,
-    ) -> Result<TaskStorage> {
+    ) -> Result<crate::backend::task_storage::TaskStorage> {
```
lukesandberg: import

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -149,14 +153,14 @@ pub struct Storage {
     ///   Contains a copy of the pre-snapshot state that needs to be persisted.
     /// - `None`: Task was first modified during snapshot mode (not part of current snapshot). Will
     ///   be marked as modified at the beginning of the next snapshot cycle.
-    snapshots: FxDashMap<TaskId, Option<Box<TaskStorage>>>,
+    snapshots: FxDashMap<TaskId, Option<TaskStorage>>,
```
lukesandberg, line 168: we should have a test that TaskStorage maintains a niche optimization for option

`const _: () = const { assert_eq!(sizeof:<Option<TaskStorage>>(), sizeof:<Option<TaskStorage>>()); }`

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -437,7 +441,7 @@ impl Storage {
     /// Evict tasks from in-memory storage after a successful snapshot.
     ///
     /// Iterates all tasks and applies the eviction level returned by
-    /// `TaskStorage::evictability()`:
+    /// `TaskStorageInner::evictability()`:
```
lukesandberg: i think this method is on TaskStorage?

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -1,7 +1,8 @@
 //! Task Storage Schema Definition
 //!
-//! This module defines the complete schema for task storage using the TaskStorage derive macro.
-//! The schema covers all CachedDataItem variants with appropriate storage types and categories.
+//! This module defines the complete schema for task storage using the TaskStorageInner derive
+//! macro. The schema covers all CachedDataItem variants with appropriate storage types and
```
lukesandberg: stale comment referencing the long dead CachedDataItem

turbopack/crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -122,6 +122,30 @@ impl TurboMalloc {
...
+    /// growing past `n` up to the returned size is "free"). Without
+    /// `custom_allocator` we fall back to rounding `n` up to the next power
+    /// of two, which approximates the system allocator's slab behavior on
+    /// most platforms.
+    ///
+    /// Useful for "should we realloc?" decisions: a realloc that doesn't
+    /// cross a size-class boundary is wasted work, so callers can compare
+    /// `good_size(current_len)` against `good_size(current_cap)` and skip
+    /// the shrink when they're equal.
+    pub fn good_size(n: usize) -> usize {
```
lukesandberg: good_size isn't a great name (despite mimalloc), this is really get_bucket_size

Issue #34365, opened 2022-02-15
getServerSideProps not called if book-ended by shallow routes. 


dcoulter45:
### What example does this report relate to?

basic

### What version of Next.js are you using?

12.0.10

### What version of Node.js are you using?

16.7.0

### What browser are you using?

Chrome

### What operating system are you using?

macOS

### How are you deploying your application?

next dev

### Describe the Bug

Currently, I have an issue where if calling `router.push` is book-ended by calls to `router.replace` with `shallow: true`, it will stop `getServerSideProps` being called, causing the URL to update in the browser but the content of the page doesn't change.

### Expected Behavior

If I navigate to a page using push and shallow:false, I would expect `getServerSideProps` to always be called when going back. 

### To Reproduce

Here is the example I'm using currently to re-produce this issue:

If you're using the basic next template
```
npx create-next-app nextjs-blog --use-npm --example "https://github.com/vercel/next-learn/tree/master/basics/learn-starter"
```

Just add this to pages/post/[...slug].js

```
import router from 'next/router';
import Link from 'next/link';
import React, { useEffect, useRef, useState } from 'react';

export const getServerSideProps = (ctx) => {
  console.log('getServerSideProps', ctx.query);
  return {
    props: {
      page: ctx.query.slug[0],
    },
  };
};

const Post = (props) => {
  const [index, setIndex] = useState(0);

  const handleClick = () => {
    setIndex(index + 1);

    router.replace(`/post/${props.page}/clicks/${index + 1}`, undefined, {
      shallow: true,
    });
  };

  useEffect(() => {
    setIndex(0);
  }, [props.page]);

  return (
    <div>
      <h1>This is post {props.page}</h1>

      <h3>Clicks {index}</h3>

      <button onClick={handleClick}>bump</button>

      <Link href="/post/1" passHref>
        <a>
          Post 1
        </a>
      </Link>

      <Link href="/post/2" passHref>
        <a>
          Post 2
        </a>
      </Link>

      <Link href="/post/3" passHref>
        <a>
          Post 3
        </a>
      </Link>
    </div>
  )
};

export default Post;
```

To re-produce the normal behavior, click bump a couple of times, click another page link and click back. This will update the heading tag and call getServerSideProps each time `router.push` or history back is used. 

![Artboard](https://user-images.githubusercontent.com/2168465/154044614-758aadaf-1367-4cd7-affc-ccc6a0ec2c4a.png)

[...]

Issue #16028, opened 2020-08-09
Navigating Between Dynamic Routes with Browser Back Button Breaks After Shallow Routing


onderonur:
# Bug report

## Describe the bug

I have a `/movie/[movieId]` page. It has `getServerSideProps` function for SEO purposes and fetches base movie info at server-side. In this page, there is an image list and when you click an image it opens a modal by adding a query string like `?view=/xXBnM6uSTk6qqCf0SRZKXcga9Ba.jpg`. To prevent unnecessary SSR, I use shallow routing like this:
```ts
const query = { view: "/xXBnM6uSTk6qqCf0SRZKXcga9Ba.jpg" };
router.push(
   pathname: "/movie/[movieId]", query },
   pathname: `/movie/${movie.id}`, query },
   { shallow: true },
);
```
and when user closes the modal, it routes to the movie page like this:
```ts
const query = {  };
router.push(
   pathname: "/movie/[movieId]", query },
   pathname: `/movie/${movie.id}`, query },
   { shallow: true },
);
```
After that, user clicks another movie and navigates to its page (without shallow routing). Still no problem. But when we go back with browser's back button, `getServerSideProps` is not called, so we have the movie info from the previous movie (before clicking the back button).

At client-side, `router` has the right params etc, so I can check if the id of data from SSR and `router`'s current `movieId` are same or not and then fetch proper info at client-side. But it adds a little bit of overhead for more complex scenarios. Another solution is, when I disable shallow routing for closing the modal, there is no problem too but it ends up with unnecessary SSR.

I've seen similar issues but many of them were about `getInitialProps` and I'm not sure if they were a match for this issue.

Thanks for your hard work and thanks for this great framework!

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Go to https://tmdb-explorer.vercel.app/movie/516486
2. Scroll a bit down, get to "Images" section.
3. Click one of the images. It will open a modal for viewing the selected image (with shallow routing).
4. Click close button at the top right of the screen. This will trigger a shallow routing too.
5. There is "Recommendations" section at the right (or at the bottom if you are on a small device). Click one of the movies. Wait for new page to load. (This is not shallow routing.)
6. Click back button of your browser.
7. Now we have the previous movie's info at the top of screen. Title, genres, overview, poster etc. Videos and Recommendations are proper. Because they are fetched at client-side with current `movieId` param. But "Images" list is wrong too. Because it depends on the fetched movie info.

[...]

PR #16477, opened 2020-08-22
Correct shallow routing behavior through history

ijjk:
When navigating from a non-shallow page to a page that was navigated to shallowly in history it causes incorrect page data to be shown since an incorrect shallow navigation is done. We can only maintain shallow routing when the current page was navigated to shallowly and the page being navigated to in history was also navigated to shallowly. Additional tests to ensure this behavior is working has also been added. 

Closes: https://github.com/vercel/next.js/issues/7395
Closes: https://github.com/vercel/next.js/issues/14928
Fixes: https://github.com/vercel/next.js/issues/16028

PR #93870, opened 2026-05-15
fix(router): do not inherit shallow flag from history state on browser back/forward (popstate)

### LINKED ISSUE #93844: router.replace with shallow: true persists shallow flag in history state, causing getServerSideProps to be skipped on browser back navigation

https://github.com/JanKaifer/next-repro-34365.

### To Reproduce

When `router.replace` is called with `shallow: true`, Next.js writes the shallow flag into the `window.history` state of that entry. This is correct for the immediate navigation, it suppresses the getServerSideProps call as intended. However, the flag is never cleared, so when the user later navigates back to that entry via the browser, Next.js reads the flag from the history state and incorrectly skips the data fetch again. The page URL updates but the page is rendered with stale data.

| Step | Action | URL | History stack (oldest → newest) | Data fetch? |
|---|---|---|---|---|
| 1 | `router.push('/page1')` | `/page1` | • **`/page1 {shallow:false}`** | Yes |
| 2 | `router.replace('/page1?count=1', {shallow:true})` | `/page1?count=1` | • **`/page1?count=1 {shallow:true}`** | No (shallow) |
| 3 | `router.push('/page2')` | `/page2` | • `/page1?count=1 {shallow:true}` <br> • **`/page2 {shallow:false}`** | Yes |
| 4 | `router.replace('/page2?count=1', {shallow:true})` | `/page2?count=1` | • `/page1?count=1 {shallow:true}` <br> • **`/page2?count=1 {shallow:true}`** | No (shallow) |
| 5 | Browser back | `/page1?count=1` | • **`/page1?count=1 {shallow:true}`** ⚠️ <br> • `/page2?count=1 {shallow:true}` | No ❌ Bug |

#### Expected behaviour
Navigating back to /page1?count=1 via the browser should trigger a getServerSideProps call and render fresh data.

#### Actual behaviour
getServerSideProps is not called. The URL updates to /page1?count=1 but the page retains stale data from the previous render.

```bash
n/a
```

#### Related issues
This appears to be a variant of #34365 and #16028.

sleitor:
## What

Fixes #93844

When `router.replace()` is called with `shallow: true`, Next.js stores the shallow flag inside `window.history.state` for that history entry. This is correct for the immediate navigation — it suppresses the `getServerSideProps` call as intended.

However, the flag is **never cleared**, so when the user later navigates back (or forward) via the browser, the `onPopState` handler reads `options.shallow` from the stored history state and re-applies it, incorrectly skipping the data fetch again.

## How

In the `onPopState` handler, force `shallow: false` regardless of what is stored in the history state:

```ts
// Before
shallow: options.shallow && this._shallow,

// After — browser back/forward must never be shallow
shallow: false,
```

Browser back/forward navigations (popstate events) should always trigger a fresh `getServerSideProps` call. The `_shallow` flag in the history state was only meant to suppress the fetch at the time of the original `replace`/`push` call.

## Testing

Added a regression test to the existing `client-shallow-routing` E2E suite that:
1. Visits a page
2. Calls `router.replace()` with `shallow: true` (no data fetch expected)
3. Navigates to another page
4. Presses browser back — verifies that `getServerSideProps` **is** called (new `random` value)

## Related

- #34365 — original report of similar behavior
- #16028 — related history state / shallow routing issue


owennicoljlp:
We are experiencing the same bug, please prioritise as it's breaking critical functionality on johnlewis.com

sleitor:
@owennicoljlp Thanks for the confirmation — really appreciate you sharing the real-world impact on johnlewis.com. The fix is straightforward: strip the `_shallow` flag from history state in the popstate handler so browser back/forward always triggers a fresh `getServerSideProps` call rather than inheriting the shallow flag from a prior `router.replace`. The PR is ready and awaiting CI approval + maintainer review. Hopefully this gets prioritized given the impact.

owennicoljlp:
@sleitor Thanks, I reached out to Vercel and @timneutkens suggested this:

>Having looked at the PR his current view is that the existing behaviour is intentional when you explicitly use shallow: true, Next.js treats that as a developer contract to manage your own data freshness, and that persists in the history state by design (see this initial PR which introduced the behaviour https://github.com/vercel/next.js/pull/16477).

>When you add shallow: true you're responsible for making sure the data stays up to date. You're explicitly skipping the fetch, and that persists in memory as otherwise it'll break other cases where you rely on Next.js not triggering getServerSideProps / getStaticProps / getInitialProps

>What Tim encouraged is submission by John Lewis of a minimal reproduction in a separate GitHub issue. I advise making this as specific to the production issue as possible (for example if the navigation moves between the same route pattern, keep that in the reproduction).

petewarman:
This issue is specifically about using `shallow: true` with `router.replace` together. The correct fix would be for `router.replace` to restore the shallow flag in the history state to whatever value it held before the replace was called, rather than leaving it set to true. This way the history state accurately reflects the entry's original provenance rather than the operation used to update it.

sleitor:
@owennicoljlp @petewarman Thanks for the continued context — these are really helpful perspectives!

I want to clarify the scope of this PR:

**This fix is specifically for the `popstate` case (browser Back/Forward) — not for explicit developer `shallow: true` calls.**

I agree completely with @timneutkens' point: when a developer explicitly calls `router.push(url, { shallow: true })`, that's a deliberate contract and should absolutely persist as intended. The developer is saying "I'll manage my own data freshness for this navigation."

The bug this PR addresses is different: when a user presses the **browser Back button**, Next.js fires a `popstate` handler that reads the `shallow` flag stored in the *previous* history entry (which had `shallow: true`). The user hasn't made any such contract — they just pressed Back. The natural expectation is that navigating Back triggers a fresh data fetch, not that it inherits a developer's earlier "skip fetching" instruction.

@petewarman's suggestion of restoring the shallow flag in `router.replace` is a valid alternative direction — it would address the root cause by not *storing* shallow=true in history state in the first place (or clearing it after use). Both approaches work; this PR takes the "don't *read* shallow from history state on `popstate`" approach.

Given the real-world production impact at johnlewis.com and the +4 👍 on the linked issue, I'd love to find the right path forward. Happy to pivot to the `router.replace` approach or adjust the fix if @timneutkens or the team has a preference on where to address this.

sleitor:

PR #93871, opened 2026-05-15
Trace middleware/proxy source files in webpack NFT pipeline

unstubbable:
Webpack's source-level NFT pass in `next-trace-entrypoints-plugin.ts` previously ran only for `pages/` and `app/` entries, so middleware fell back to the post-bundle chunk trace in `collect-build-traces.ts`. That trace cannot see through webpack's numeric module-ID indirection (`f = c(3024)` where module 3024 re-exports `require("node:fs")`), so runtime file references inside Node middleware — e.g. `fs.readFile(path.join(process.cwd(), 'package.json'))` — never made it into `middleware.js.nft.json`. Vercel deploys then 500'd with `ENOENT '/var/task/package.json'` because the file was not shipped into the function bundle. Turbopack avoids this through its own pre-bundle asset graph trace, which is why only the webpack deploy path was affected; the test that surfaced it (the `/test-node-fs` case in `test/e2e/middleware-general`) had been skipped on deploy until #93614 re-enabled it.

The fix extends the Pass 1 entry filter to also process middleware entries, keyed off `isMiddlewareFilename` and using `buildInfo.route.absolutePagePath` as the entry source — middleware lives at the project root rather than under `pagesDir`/`appDir`, so `getPageFilePath` does not apply. The existing `createTraceAssets` emission and the `collect-build-traces.ts` merge already handle arbitrary entry names, so no further changes are needed downstream.

`test/e2e/middleware-general/app/middleware-node.js` is also cleaned up to use static `import * as fs from 'node:fs'` / `node:path`, drop the redundant `process.env.NEXT_RUNTIME === 'nodejs'` guard, and wrap the callback `fs.readFile` in a Promise. `@vercel/nft`'s symbol table only recognizes the callback and sync `fs` APIs, not `fs.promises.*`, so the call site has to remain a direct `fs.readFile(...)` for the path argument to be backtracked into the trace; a TODO references the upstream gap.

### DIFF
--- a/packages/next/src/build/webpack/plugins/next-trace-entrypoints-plugin.ts
+++ b/packages/next/src/build/webpack/plugins/next-trace-entrypoints-plugin.ts
@@ -19,6 +19,7 @@ import { getModuleBuildInfo } from '../loaders/get-module-build-info'
 import { getPageFilePath } from '../../entries'
 import { resolveExternal } from '../../handle-externals'
 import { isMetadataRouteFile } from '../../../lib/metadata/is-metadata-route'
+import { isMiddlewareFilename } from '../../utils'
 import { getCompilationSpan } from '../utils'
 
 const PLUGIN_NAME = 'TraceEntryPointsPlugin'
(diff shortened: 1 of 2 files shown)


timneutkens (approved):

timneutkens:
Looks good to me but @mischnic is the expert on this.

unstubbable:
Got an out-of-band approval from @mischnic: "Looks good if it works" 😄

PR #93873, opened 2026-05-15
Prerender static metadata under dynamic segments to canonical pathname

timneutkens:
### What?

Fixes an invariant failure (`failed to find source route /dynamic/[id]/apple-icon.png for prerender /dynamic/[id]/apple-icon.png`) raised by the build adapter when an app contains static metadata files (`apple-icon.png`, `icon.png`, `opengraph-image.png`, `twitter-image.png`, `sitemap.xml`) under a dynamic segment.

### Why?

Static metadata files under a dynamic segment were treated as full dynamic prerenders: they ended up in `prerenderManifest.dynamicRoutes` with `fallbackRouteParams`, but no `.body` file was exported and they never received an `appOutputMap` entry. The adapter's `dynamicRoutes` loop then asked `getParentOutput` for a source route that didn't exist and threw.

These files are content-stable across params — the route handler returns the same bytes for every URL that matches the dynamic pattern — so they should prerender once under a canonical pathname rather than be treated like per-param dynamic routes.

### How?

- New `getStaticMetadataPrerenderPathname` helper that maps `/dynamic/[id]/apple-icon.png` to `/dynamic/-/apple-icon.png`. This is the canonical pathname already used by `fillStaticMetadataSegment` for HTML link emission.
- In `build/utils.ts`, route `APP_ROUTE` entries that are static metadata files through a new `buildStaticMetadataStaticPaths` (one prerender path, no fallback params) instead of the regular `buildAppStaticPaths`.
- In `build/index.ts`:
  - Rewrite each prerendered static metadata route to its `-`-placeholder pathname before splitting into known/unknown buckets, so it lands in `prerenderManifest.routes` with a `.body` file.
  - Skip static metadata entries in the `dynamicPrerenderedRoutes` loop so they don't get a stray `PRERENDER` manifest entry.
  - Defensive sweep before write: drop any leftover bracketed metadata keys from `prerenderManifest.dynamicRoutes`.
- In the adapter, look up `prerenderManifest.routes` under the canonical placeholder pathname (with locale prefix when applicable) and skip static metadata in the `dynamicRoutes` loop.
- Unit tests in `get-metadata-route.test.ts` lock in the placeholder behavior for the tricky cases: catchall, optional catchall, multi-level dynamic, literals between dynamics, dynamic + catchall mix.

Reproduction: `test/e2e/app-dir/metadata-static-file/metadata-static-file-dynamic-route.test.ts` configured with `adapterPath` previously crashed at `handleBuildComplete`; all three cases in that file now pass.

Runtime stays the same: arbitrary param URLs continue to be served by the dynamic route handler shipped with the app route; only the canonical `-` URL is statically served.

### DIFF
--- a/packages/next/errors.json
+++ b/packages/next/errors.json
@@ -1277,5 +1277,6 @@
   "1276": "Compilation failed but no issues were recorded",
   "1277": "no route matched for path \"%s\"",
   "1278": "compileRoute: either routeSpecifier or path is required",
-  "1279": "\\`experimental.appShells\\` requires the following to also be enabled: %s. Please update your %s accordingly."
+  "1279": "\\`experimental.appShells\\` requires the following to also be enabled: %s. Please update your %s accordingly.",
+  "1280": "Invariant: expected static metadata route to have a prerender pathname (%s)"
 }
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -43,6 +43,7 @@ import {
 } from '../../lib/constants'
 
 import { normalizeLocalePath } from '../../shared/lib/i18n/normalize-locale-path'
+import { getStaticMetadataPrerenderPathname } from '../../lib/metadata/get-metadata-route'
 import { isStaticMetadataFile } from '../../lib/metadata/is-metadata-route'
 import { addPathPrefix } from '../../shared/lib/router/utils/add-path-prefix'
 import { getRedirectStatus, modifyRouteRegex } from '../../lib/redirect-status'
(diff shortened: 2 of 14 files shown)


gnoff (approved):
This is a nice improvement.

To make sure we're aligned on where we can go from here it seems to me the ideal impl is

1. static metadata files are persisted in /public or wherever we store other static assets
2. this public path is used in place of `-` param substitution when rendering these files as metadata
3. a redirect route matcher is used to take ad-hoc .../[slug]/... pathnames and map them to the single static instance of this image/icon/etc...

The critical component is there should not need to be a route handler to serve this static image nor should there be any ISR.

If we had to sacrifice 3 to simply I think that'd also be acceptable

timneutkens:
Thanks, that matches the direction I had in mind.

This branch implements the critical subset for now: static metadata under dynamic segments is canonicalized to one known pathname like `/dynamic/-/icon.png`, added to `prerenderManifest.routes` instead of `dynamicRoutes`, and the adapter output emits the generated `.body` as a `STATIC_FILE` while skipping the app route/dynamic prerender output. So the canonical static metadata path should not require ISR or an app route handler.

It does not yet switch the rendered URL away from the `-` placeholder to a separate public asset path, and it does not add the ad-hoc dynamic URL redirect matcher. I think those can be a follow-up, with this PR focused on removing the dynamic/ISR behavior and making adapter output work for the single static instance.

PR #93874, opened 2026-05-15
[otel] honor traceparent for verbose mode spans

kakadiadarpan:
## Problem
Fixes the trace propagation issue reported in https://github.com/vercel/otel/issues/107.

`NEXT_OTEL_VERBOSE=1` opened the `NextServer.getRequestHandler` span before request header propagation, so incoming `traceparent` was ignored and framework spans started a new trace.

## Solution
Extract propagated request context before wrapper request-handler spans. Add focused e2e coverage in the existing OpenTelemetry suite for verbose spans inheriting the incoming `traceparent`, including the parent span id.

Validated with `pnpm --filter=next build` and `pnpm test-dev-turbo test/e2e/opentelemetry/instrumentation/opentelemetry.test.ts`.


packages/next/src/server/next.ts
```diff
@@ -149,10 +149,13 @@ export class NextServer implements NextWrapperServer {
       res: ServerResponse,
       parsedUrl?: NextUrlWithParsedQuery
     ) => {
-      return getTracer().trace(NextServerSpan.getRequestHandler, async () => {
-        const requestHandler = await this.getServerRequestHandler()
-        return requestHandler(req, res, parsedUrl)
-      })
+      const tracer = getTracer()
+      return tracer.withPropagatedContext(req.headers, () =>
```
timneutkens, line 153: For perf reasons you'll have to change the withPropagatedContext to early return when otel is not enabled similar to `trace()`. At this point: https://github.com/vercel/next.js/blob/bd070d06f8e738286b633b69e29e3784434faad7/packages/next/src/server/lib/trace/tracer.ts#L251

Similar early return in trace(): https://github.com/vercel/next.js/blob/bd070d06f8e738286b633b69e29e3784434faad7/packages/next/src/server/lib/trace/tracer.ts#L298

kakadiadarpan: fixed here: https://github.com/vercel/next.js/pull/93874/commits/e130594e708565b110fc9d7e090bec9fee6811a2

timneutkens (approved):

PR #93875, opened 2026-05-15
turbopack: emit assets before duplicate check in emit_assets

sokra:
### What?

Reworks `emit_assets` in `crates/next-core/src/emit.rs` so that:

- The actual `emit()` / `emit_rebase()` write runs **before** `check_duplicates()`.
- Per-duplicate diff checks inside `check_duplicates` run in parallel.
- At most one `EmitConflictIssue` is emitted per asset path, even when multiple duplicates conflict.
- The conflict-issue emission is wrapped in a `turbo_tasks::function` and reported via `as_side_effect()`.

### Why?

Under eventual consistency, all children might be removed from `emit_assets`. This causes all `content()` and `emit()` tasks being inactive and keeping the stale value (which is potentially an error).

On recomputation of `emit_assets()` this only makes very few `content()` tasks active, as tasks are still in a stale error state and make `emit_assets()` exit early. So it takes N recomputations of `emit_assets()` to be up-to-date, where N is the number of duplicate assets. This can slow down builds drastically and did cause build time up to 30min.

Emitting the asset first ensures the file is written to disk even if the subsequent duplicate diagnostic crashes under eventual consistency. Doing the diff checks in parallel and folding conflicts into a single issue keeps the diagnostic path cheap and avoids noisy duplicate output for the same path. Tracking the issue emission as a turbo-tasks side effect means it is properly attributed to a task instead of being emitted ad-hoc from `emit_assets`.

### How?

- Swap the order in both branches of `emit_assets` (server-side `emit` and client-side `emit_rebase`) so the write happens first, then `check_duplicates` runs as a diagnostic afterwards. A comment is left explaining the ordering requirement.
- `check_duplicates` now returns `Result<()>` instead of returning the picked asset. The caller picks `assets.first()` directly, since `check_duplicates` only validated equivalence — it didn't choose a different asset.
- The inner loop in `check_duplicates` uses `iter.map(async |next| { ... }).try_flat_join().await?` to run the diff comparisons in parallel and collect conflict details. `ext` is computed once before the loop and cloned into each task.
- Only the first conflict from the collected list is reported, instead of one issue per conflicting asset.
- The `EmitConflictIssue` emission is moved into a local `#[turbo_tasks::function] fn emit_conflict_issue(...)` invoked via `.as_side_effect().await?`, so the issue is tracked as a task side effect.

### DIFF
--- a/crates/next-core/src/emit.rs
+++ b/crates/next-core/src/emit.rs
@@ -104,22 +104,22 @@ pub async fn emit_assets(
         let mut iter = assets.into_iter();
         let first = iter.next().unwrap();
         let ext: RcStr = path.extension().unwrap_or_default().into();
-        iter.map(async |next| {
-            if let Some(detail) = assets_diff(*next, *first, ext.clone(), node_root.clone())
-                .owned()
-                .await?
-            {
-                EmitConflictIssue {
-                    asset_path: path.clone(),
-                    detail,
-                }
-                .resolved_cell()
-                .emit();
+        let conflicts = iter
+            .map(async |next| {
+                assets_diff(*next, *first, ext.clone(), node_root.clone())
+                    .owned()
+                    .await
+            })
+            .try_flat_join()
+            .await?;
+        if let Some(detail) = conflicts.into_iter().next() {
+            EmitConflictIssue {
+                asset_path: path.clone(),
+                detail,
             }
-            Ok(())
-        })
-        .try_join()
-        .await?;
+            .resolved_cell()
+            .emit();
+        }
         Ok(())
     }
 
--- a/crates/next-core/src/emit.rs
+++ b/crates/next-core/src/emit.rs
@@ -103,9 +103,9 @@ pub async fn emit_assets(
     ) -> Result<()> {
         let mut iter = assets.into_iter();
         let first = iter.next().unwrap();
-        for next in iter {
-            let ext: RcStr = path.extension().unwrap_or_default().into();
-            if let Some(detail) = assets_diff(*next, *first, ext, node_root.clone())
+        let ext: RcStr = path.extension().unwrap_or_default().into();
+        iter.map(async |next| {
+            if let Some(detail) = assets_diff(*next, *first, ext.clone(), node_root.clone())
                 .owned()
                 .await?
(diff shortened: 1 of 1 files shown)


crates/next-core/src/emit.rs
```diff
@@ -116,8 +116,11 @@ pub async fn emit_assets(
                 .resolved_cell()
```
lukesandberg: this cell is now non-deterministic

at least add a comment, but we could fix it by moving construction into a new task and using `as_side_effect` (i think)

lukesandberg (approved):

crates/next-core/src/emit.rs
sokra: fixed

raquelpn1979-tech:
[]() ¡Me parece una idea fantástica! Viajar por la historia paso a paso es la mejor forma de entender cómo el pasado ha moldeado nuestro presente. Yo te guiaré planteándote preguntas en cada etapa para que vayamos descubriendo juntos los acontecimientos. 🧭
Para empezar esta aventura, ¿qué periodo histórico te llama más la atención explorar? Aquí tienes tres opciones muy interesantes para elegir:
 1. **La Antigua Roma 🏛️**: Desde sus humildes comienzos como una pequeña aldea junto al río Tíber hasta convertirse en un imperio colosal que dominó el Mediterráneo.
 2. **La Revolución Francesa 🗼**: Un torbellino de ideas, revueltas y cambios drásticos que comenzó en 1789 y transformó la política mundial para siempre.
 3. **La Era de los Descubrimientos 🗺️**: La época a finales del siglo XV donde exploradores audaces cruzaron océanos desconocidos y conectaron continentes por primera vez.
¿Con cuál de estos viajes en el tiempo prefieres que empecemos?

raquelpn1979-tech:
¡Entendido! Si necesitas analizar algo del archivo, por favor, copia y pega aquí su contenido o describeme qué clase de información contiene. Si tienes problemas accediendo al archivo o necesitas ayuda con Dropbox o la integración del contenido en un contexto como GitHub o APIs, házmelo saber. ¡Estoy aquí para ayudar!

PR #93876, opened 2026-05-15
turbo-tasks: prioritize recomputed tasks with Recomputation priority

sokra:
### What?

Introduces a new `Recomputation` priority in `TaskPriority` that schedules recomputed tasks (tasks re-run because their output or cell wasn't available when read) ahead of all invalidation phases. Also documents every `TaskExecutionReason` variant and removes the unused `Local` variant.

### Why?

When a task is recomputed because a downstream consumer needed its output/cell, that recomputation should run before lower-priority invalidations so the consumer can make progress. Previously these recomputations were scheduled with `Initial` priority, which placed them alongside fresh work and behind ongoing invalidation phases, hurting latency for the task actually being awaited.

Care is needed so the new priority doesn't spread virally: any invalidation triggered during a `Recomputation` task must not itself be tagged `Recomputation`, otherwise the priority would leak across the graph and break ordering. The change therefore resets the priority to `Initial` for invalidations triggered from a recomputation context.

### How?

- `turbo-tasks/src/manager.rs`: Add `TaskPriority::Recomputation` variant, ordered before `Invalidation`. Update `Display` and `update` impls.
- `turbo-tasks-backend/src/backend/mod.rs`: Schedule output-not-available and cell-not-available recomputations with `TaskPriority::Recomputation` instead of `Initial`.
- `turbo-tasks-backend/src/backend/operation/invalidate.rs`: When `make_task_dirty_internal` runs under a parent task with `Recomputation` priority, downgrade the propagated priority to `Initial` so the recomputation marker does not spread. Also tighten the "already dirty" branch so a stored `Initial` priority is always replaced by a more specific parent priority.
- `turbo-tasks/src/task_execution_reason.rs`: Document each variant and drop the unused `Local` variant (and its `as_str` arm).

Closes NEXT-

### DIFF
--- a/turbopack/crates/turbo-tasks/src/task_execution_reason.rs
+++ b/turbopack/crates/turbo-tasks/src/task_execution_reason.rs
@@ -1,13 +1,28 @@
 #[derive(Debug)]
 pub enum TaskExecutionReason {
+    /// A root task was initially scheduled and is executed
     Initial,
-    Local,
+    /// A task output was read, but the output is not available, so the task was scheduled and
+    /// executed to produce the output
     OutputNotAvailable,
+    /// A task cell was read, but the cell content is not available, so the task was scheduled and
+    /// executed to produce the cell content
     CellNotAvailable,
+    /// A task was marked as dirty and is active on it's own (maybe root or currently awaited), so
+    /// it was scheduled and executed to update the task output
     Invalidated,
+    /// A dirty task has been activated, so it was scheduled and executed to update the task output
     ActivateDirty,
+    /// A task has been activated for the first time (no output yet), so it was scheduled and
+    /// executed to produce the task output
     ActivateInitial,
+    /// A task was connected as child in `active_tracking == false` mode for the first time (no
+    /// output yet), so it was scheduled and executed to produce the task output.
+    /// Or a task was called inside of a turbo_tasks::run closure for the first time (no output
+    /// yet), so it was scheduled and executed to produce the task output.
     Connect,
+    /// An in-progress task was marked as stale, so it was scheduled again after execution and is
+    /// executing again to update the task output.
     Stale,
 }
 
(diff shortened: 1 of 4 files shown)


turbopack/crates/turbo-tasks/src/task_execution_reason.rs
```diff
@@ -1,21 +1,35 @@
 #[derive(Debug)]
 pub enum TaskExecutionReason {
+    /// A root task was initially scheduled and is executed
```
lukesandberg, line 3: it isn't necessarily a 'root' task, right?

lukesandberg: or is this the difference between Initial and ActivateInitial?

maybe this should be RootInitial?

turbopack/crates/turbo-tasks/src/task_execution_reason.rs
```diff
@@ -1,21 +1,35 @@
...
+    /// A dirty task has been activated, so it was scheduled and executed to update the task output
     ActivateDirty,
+    /// A task has been activated for the first time (no output yet), so it was scheduled and
+    /// executed to produce the task output
     ActivateInitial,
+    /// A task was connected as child in `active_tracking == false` mode for the first time (no
+    /// output yet), so it was scheduled and executed to produce the task output.
+    /// Or a task was called inside of a turbo_tasks::run closure for the first time (no output
+    /// yet), so it was scheduled and executed to produce the task output.
     Connect,
```
lukesandberg, line 23: this one is kind of confusing to me... this could be a new root task connecting a child?

but how is that different from `Initial`.  i see that this is set if a child is connected with no parent, so isn't that limited to root tasks

lukesandberg (approved):

turbopack/crates/turbo-tasks/src/task_execution_reason.rs
sokra: The difference is due to different operation modes: track_activeness = true vs false.
One scheduled on activation. The other on connecting.

turbopack/crates/turbo-tasks/src/task_execution_reason.rs
sokra: Yeah I can rename that in a follow-up PR.

PR #93877, opened 2026-05-15
Turbopack: don't crash the dev server when node_modules/next is briefly unresolvable

lukesandberg:
### What?

When `node_modules/next` is briefly unresolvable mid-session (for example because pnpm is mid-install of a package with a `next` peer dependency, reshuffling its symlinks) Turbopack would:

- Emit a `MissingNextFolderIssue` with severity `Fatal`
- Propagate `anyhow!("Next.js package not found")` through ~10 levels of the build graph
- Surface the failure at the napi boundary as a `TurbopackInternalError`, which the Next.js JS-side treats as fatal and shuts the dev server down

The reported user impact was a dev session that became wedged and could not recover even after restarting `next dev`, because the persistent cache had stored the failed operation state. Users reported it as "stuck", "catastrophic", and required wiping `.next/` to recover.

Unfortunately while i was able to reproduce a number of bad outcomes, i couldn't actually reproduce the issue of `next dev` getting stuck in a bad state.  Still this PR will improve a number of apis and error messages

See internal discussion: https://vercel.slack.com/archives/C046HAU4H7F/p1778792876550309

### Why?

A transient filesystem race during an install is the canonical recoverable error — once the install finishes, `next` is resolvable again. There is no reason this should bring down the dev server or poison the persistent cache.

The bug had two contributing causes:

1. The error message was poor. It assumed the only cause was a mis-set `turbopack.root`, with the entire message in `title()`. It didn't acknowledge the install-race case or any of the other realistic causes (broken symlink, monorepo hoisting, removed `node_modules/next`, global install). Users had no way to know that the right response was to wait or refresh.

2. The Rust→napi boundary did not treat this as a recoverable error. Two specific code paths propagated `Err` instead of catching it:
   - `issue_filter_from_endpoint` in `endpoint.rs` calls `endpoint_op.connect().await?` — if the upstream endpoint resolution errors (as it does when `directory_tree_to_loader_tree` fails), this `?` propagates **before** the surrounding `strongly_consistent_catch_collectables` gets a chance to convert the failure into Issues.
   - `Project::hmr_version_state` is called bare from `project_hmr_events` at the napi boundary. There is no `_with_issues_operation` wrapper, so any error from the underlying `hmr_content`/`versioned_content_map.compute_entry` chain propagates straight to JS.

### How?

Three changes:

**1. Rewrite `MissingNextFolderIssue`** ([next_import_map.rs](crates/next-core/src/next_import_map.rs))
- Severity downgraded from `Fatal` to `Error`. `Fatal` triggers an explicit JS-side server shutdown in `print-build-errors.ts`; `Error` lets the dev server recover.
- Title is a single line: `"Could not find the Next.js package (next/package.json)"`.
- Description lists all realistic causes (concurrent install, missing/broken symlink, mis-set workspace root, monorepo hoisting, global install).
- Doc link moved to `documentation_link()` so it doesn't pollute the inline message.

**2. Make `issue_filter_from_endpoint` Err-tolerant** ([endpoint.rs](crates/next-napi-bindings/src/next_api/endpoint.rs))
- Wrap the `endpoint_op.connect().await` in a `match` and fall back to `IssueFilter::warnings_and_foreign_errors()` on `Err`.
- This unblocks the existing `strongly_consistent_catch_collectables` recovery path, which then absorbs the upstream failure as Issues that surface through the napi boundary normally.

**3. Add `hmr_version_state_with_issues_operation`** ([project.rs](crates/next-napi-bindings/src/next_api/project.rs))
- Mirrors the existing `hmr_update_with_issues_operation` pattern.
- Drives `hmr_version_operation` as an `OperationVc`, catches any `Err` and falls back to `NotFoundVersion`, and collects issues from the operation via `peek_issues`.
- The napi `project_hmr_events` handler now uses this wrapper instead of calling `Project::hmr_version_state` bare. Issues from both `hmr_version_state` and `hmr_update` are merged before being sent to JS.
- Required lifting the inner `hmr_version_operation` out of `Project::hmr_version_state`'s closure to make it callable from the napi layer.

A new regression test at [test/development/app-dir/concurrent-install/](test/development/app-dir/concurrent-install/) reproduces the failure by moving `node_modules/next` aside mid-HMR-session and asserts that the dev server does not emit `FATAL: An unexpected Turbopack error occurred` or `TurbopackInternalError`, and that the friendly Issue text surfaces. The test is skipped under `NEXT_SKIP_ISOLATE=1` since that mode doesn't produce a manipulable `node_modules`.

Out of scope: after recovery, the dev server still returns 500 on subsequent requests until the user manually refreshes (the JS-side per-route error state isn't invalidated by Turbopack's successful recompile). That's a separate dev-server caching issue.

### DIFF
--- a/crates/next-api/src/project.rs
+++ b/crates/next-api/src/project.rs
@@ -2411,6 +2411,19 @@ impl Project {
         // sessions.
         let _ = session;
 
+        #[turbo_tasks::function(operation, root)]
+        async fn hmr_version_operation(
+            this: ResolvedVc<Project>,
+            chunk_name: RcStr,
+            target: HmrTarget,
+        ) -> Result<Vc<Box<dyn Version>>> {
+            let content = this.hmr_content(chunk_name, target).await?;
+            if let Some(content) = &*content {
+                Ok(content.version())
+            } else {
+                Ok(Vc::upcast(NotFoundVersion::new()))
+            }
+        }
         let version_op = hmr_version_operation(self, chunk_name, target);
 
         // INVALIDATION: This is intentionally untracked to avoid invalidating this
@@ -2722,18 +2735,4 @@ fn all_assets_from_entries_operation(
 ) -> Result<Vc<ExpandedOutputAssets>> {
     let assets = operation.connect();
     Ok(all_assets_from_entries(assets))
-}
-
-#[turbo_tasks::function(operation, root)]
-pub async fn hmr_version_operation(
-    project: ResolvedVc<Project>,
-    chunk_name: RcStr,
-    target: HmrTarget,
-) -> Result<Vc<Box<dyn Version>>> {
-    let content = project.hmr_content(chunk_name, target).await?;
-    if let Some(content) = &*content {
-        Ok(content.version())
-    } else {
-        Ok(Vc::upcast(NotFoundVersion::new()))
-    }
 }
--- a/crates/next-api/src/project.rs
+++ b/crates/next-api/src/project.rs
@@ -2411,19 +2411,6 @@ impl Project {
         // sessions.
         let _ = session;
 
-        #[turbo_tasks::function(operation, root)]
-        async fn hmr_version_operation(
-            this: ResolvedVc<Project>,
-            chunk_name: RcStr,
-            target: HmrTarget,
-        ) -> Result<Vc<Box<dyn Version>>> {
-            let content = this.hmr_content(chunk_name, target).await?;
-            if let Some(content) = &*content {
-                Ok(content.version())
-            } else {
-                Ok(Vc::upcast(NotFoundVersion::new()))
-            }
-        }
         let version_op = hmr_version_operation(self, chunk_name, target);
 
         // INVALIDATION: This is intentionally untracked to avoid invalidating this
@@ -2735,4 +2722,18 @@ fn all_assets_from_entries_operation(
 ) -> Result<Vc<ExpandedOutputAssets>> {
     let assets = operation.connect();
     Ok(all_assets_from_entries(assets))
+}
+
+#[turbo_tasks::function(operation, root)]
+pub async fn hmr_version_operation(
+    project: ResolvedVc<Project>,
+    chunk_name: RcStr,
+    target: HmrTarget,
+) -> Result<Vc<Box<dyn Version>>> {
+    let content = project.hmr_content(chunk_name, target).await?;
+    if let Some(content) = &*content {
+        Ok(content.version())
+    } else {
+        Ok(Vc::upcast(NotFoundVersion::new()))
+    }
 }
--- a/crates/next-core/src/next_import_map.rs
+++ b/crates/next-core/src/next_import_map.rs
@@ -1195,7 +1195,7 @@ impl Issue for MissingNextFolderIssue {
     }
 
     fn severity(&self) -> IssueSeverity {
-        IssueSeverity::Fatal
+        IssueSeverity::Error
     }
 
     fn stage(&self) -> IssueStage {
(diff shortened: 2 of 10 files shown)


lukesandberg:
* **#93877** <a href="https://app.graphite.com/github/pr/vercel/next.js/93877?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93877?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

crates/next-api/src/project.rs
```diff
@@ -2726,10 +2713,24 @@ async fn any_output_changed(
...
+    chunk_name: RcStr,
+    target: HmrTarget,
+) -> Result<Vc<Box<dyn Version>>> {
+    let content = project.hmr_content(chunk_name, target).await?;
+    if let Some(content) = &*content {
+        Ok(content.version())
+    } else {
+        Ok(Vc::upcast(NotFoundVersion::new()))
+    }
+}
```
lukesandberg: why was this moved

crates/next-core/src/next_import_map.rs
```diff
@@ -1205,56 +1205,75 @@ impl Issue for MissingNextFolderIssue {
     }
 
     fn severity(&self) -> IssueSeverity {
-        IssueSeverity::Fatal
+        IssueSeverity::Error
```
lukesandberg, line 1201: we should justify why this isn't fatal

crates/next-core/src/next_import_map.rs
```diff
@@ -1205,56 +1205,75 @@ impl Issue for MissingNextFolderIssue {
...
+
+    async fn description(&self) -> Result<Option<StyledString>> {
+        let context_path: RcStr = match to_sys_path(self.path.clone()).await? {
+            Some(path) => path.to_str().unwrap_or("{unknown}").into(),
+            _ => rcstr!("{unknown}"),
         };
-        let root_path = match to_sys_path(self.root.clone()).await? {
-            Some(path) => path.to_str().unwrap_or("{unknown}").to_string(),
-            _ => "{unknown}".to_string(),
+        let root_path: RcStr = match to_sys_path(self.root.clone()).await? {
```
lukesandberg, line 1219: do we need a type annotation here

crates/next-napi-bindings/src/next_api/endpoint.rs
```diff
@@ -101,14 +101,25 @@ impl Deref for ExternalEndpoint {
...
+///
+/// If the upstream endpoint operation fails to resolve (e.g. because the build
+/// graph cannot be evaluated transiently — for example during a mid-session
+/// `node_modules` reshuffle), this falls back to a default filter rather than
+/// propagating the error. The caller then drives the underlying operation via
+/// `strongly_consistent_catch_collectables`, which converts the same upstream
+/// failure into Issues that this default filter can surface.
 async fn issue_filter_from_endpoint(
     endpoint_op: OperationVc<OptionEndpoint>,
 ) -> Result<Vc<IssueFilter>> {
```
lukesandberg: reconsider the return type, and see if this should be inlined to clarify error behavior.

test/development/app-dir/concurrent-install/concurrent-install.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  return <p>hello world (while-missing)</p>
+}
+`
+      )
+
+      // The Issue rendered to stdout should mention the package name and
+      // the recovery hint about a concurrent install.
+      await retry(
+        async () => {
+          expect(getOutput()).toContain('Could not find the Next.js package')
```
lukesandberg: we should use a toMatchInlineSnapshot to see the full error

crates/next-api/src/project.rs
mischnic: why was this moved?

lukesandberg: answer is, it is getting called from a second location

lukesandberg: see hmr_version_state_with_issues_operation

crates/next-core/src/next_import_map.rs
```diff
@@ -1205,56 +1205,78 @@ impl Issue for MissingNextFolderIssue {
...
+                 workspace root and is not reachable from there."
+            ))]),
+            StyledString::Line(vec![StyledString::Text(rcstr!(
+                "  - Next.js is installed globally rather than as a project dependency. This is \
+                 rare and not recommended; install it locally."
+            ))]),
+            StyledString::Line(vec![StyledString::Text(rcstr!(""))]),
+            StyledString::Line(vec![StyledString::Text(rcstr!(
+                "Note: For security and performance reasons, files outside of the workspace root \
+                 are not compiled."
```
bgw: Not really security or performance reasons IMO, just cache portability.

```suggestion
                "Note: To ensure a hermetic build and a portable cache, files outside of the workspace root are not compiled."
```

(that needs to be wrapped to 100 columns)

crates/next-core/src/next_import_map.rs
```diff
@@ -1205,56 +1205,78 @@ impl Issue for MissingNextFolderIssue {
...
-            ]),
-        ]))
+            StyledString::Line(vec![StyledString::Text(rcstr!(
+                "  - In a monorepo, the Next.js package is hoisted to a directory above the \
+                 workspace root and is not reachable from there."
+            ))]),
+            StyledString::Line(vec![StyledString::Text(rcstr!(
+                "  - Next.js is installed globally rather than as a project dependency. This is \
+                 rare and not recommended; install it locally."
+            ))]),
```
bgw, line 1258:
```suggestion
            StyledString::Line(vec![StyledString::Text(rcstr!(
                "  - In a monorepo, the Next.js package may only exist in a directory above the \
                 closest directory containing a package manager lockfile. The workspace root is \
                 detected by locating the nearest package manager lockfile."
            ))]),
            StyledString::Line(vec![StyledString::Text(rcstr!(
                "  - Next.js is installed globally rather than as a project dependency. This is \
                 not supported; install it locally."
            ))]),
```

test/development/app-dir/concurrent-install/concurrent-install.test.ts
```diff
@@ -0,0 +1,188 @@
...
+          - node_modules is being reorganized by a concurrent install (e.g. pnpm adding a package with a \`next\` peer dependency). This is transient and should clear once the install completes.
+          - node_modules/next was removed, renamed, or has a broken symlink.
+          - The workspace root is incorrect — see turbopack.root in the Next.js config docs for how to configure it.
+          - In a monorepo, the Next.js package is hoisted to a directory above the workspace root and is not reachable from there.
+          - Next.js is installed globally rather than as a project dependency. This is rare and not recommended; install it locally.
+
+        Note: For security and performance reasons, files outside of the workspace root are not compiled.
+
+        https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack#root-directory"
+      `)
```
bgw: This is excessive and will break every time we change any of the wording, just match for some part of this string

test/development/app-dir/concurrent-install/concurrent-install.test.ts
```diff
@@ -0,0 +1,188 @@
...
+// This test manipulates the test-dir's `node_modules/next`, which only exists
+// when the harness performs the real isolated install. With NEXT_SKIP_ISOLATE
+// (the local dev-loop optimization), no install is performed and the dev server
+// resolves `next` from the repo's dist/ instead — there's nothing to move.
+//
+// It also only exercises the Turbopack resolve path: the
+// `MissingNextFolderIssue` and "Could not find the Next.js package" output are
+// emitted by `crates/next-core/src/next_import_map.rs`, which webpack does not
+// use. Under webpack the dev server keeps serving from the already-compiled
+// graph and the failure mode this guards against doesn't manifest.
```
bgw: Why do we have both a multiline comment and a bunch of single line comments?

bgw (approved):

test/development/app-dir/concurrent-install/concurrent-install.test.ts
lukesandberg: toMatchInlineSnapshot does make maintenance trivial since jest can update it

but i can drop it

PR #93879, opened 2026-05-15
Redesign the unrendered-segment instant validation overlay

aurorascharff:
### What?

A focused dev-overlay redesign and CLI/build message refresh for the "expected segment was not rendered" instant validation warning (PR #93770). Now integrated with the Errors/Insights tab split from #94073.

### Why?

The wrapper carries no useful source location to point at — the code-frame chrome and Call Stack the overlay rendered before were misleading. Users mostly need to know which segment didn't render and where to set `instant = false` to silence the warning.

### Demo

- [Demo 1](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/81-instant-wrapper-unrendered-segment)
- [Demo 2](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/82-nested-layouts-unrendered-segment)
- [Demo 3](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/83-conditional-render-unrendered-segment)
- [Demo 4](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/84-parallel-slot-unrendered-segment)

### How

- Restructured framework message in `dynamic-rendering.ts`: leads with `Route "<path>":` prefix like every other validation factory, drops the verbose paragraphs in favour of a one-sentence explanation + `Ways to fix this:` list, and uses `instant` not `unstable_instant` in the fix bullet.
- New `unrendered-segment` `GuidanceKind` and a dedicated case in `errors.tsx` that drops the code-frame and Call Stack.
- New `UnrenderedSegmentInfo` component renders the route + each unrendered file with the same chrome as `CodeFrame`, plus open-in-editor.
- Two fix cards: `Render the missing segment` (render, gray) and `Allow no validation` (silence, red).
- **Errors/Insights tab integration**: `isInstantNavigationError` now returns `true` for unrendered-segment errors, so they appear under the **Insights** tab with the amber pill (not the red Issues tab). They clear on navigation like other instant errors. Added `renderTabBar` to the unrendered-segment overlay layout.


packages/next/errors.json
```diff
@@ -1277,14 +1277,15 @@
...
+  "1280": "\\`experimental.appShells\\` requires the following to also be enabled: %s. Please update your %s accordingly.",
+  "1281": "Invariant: expected static metadata route to have a prerender pathname (%s)",
+  "1282": "Cache Components HTTP fallback recovery unexpectedly produced a DynamicServerError",
+  "1283": "Cache Components error recovery expected an original Flight prerender result",
+  "1284": "Cache Components error recovery expected to know whether the original Flight prerender result was dynamic",
+  "1285": "Cache Components error recovery expected an original prerender resume data cache",
+  "1286": "Cache Components error recovery expected an original prerender store",
+  "1287": "Cache Components error recovery unexpectedly produced a dynamic HTML hole",
+  "1288": "Expected stream state to be initialized before reading",
+  "1289": "Cache Components error recovery expected an original resume data cache"
```
gnoff: should we regenerate the error list from the base commit so this isn't reordered. we shouldn't change error numbers

gnoff: unfortunately means changing snapshots too

gnoff (approved):

packages/next/errors.json
aurorascharff: Hah! You seem to forget Claude is really good at that! No problem at all!

PR #93914, opened 2026-05-18
TURBOPACK: fix subpath imports resolving to dependencies

yuzheng14:
## Summary

- Fix Turbopack's `package.json#imports` handling so subpath imports can resolve to bare package specifiers.
- Preserve the existing relative-path behavior for `exports` and relative subpath import targets.
- Add a snapshot fixture for a subpath import that resolves to a dependency from `node_modules`.

Previously, the existing `#dep` case could appear to work because Turbopack always prepended `./` to the resolved target. That turned `"dep"` into `"./dep"`, and the fixture happened to contain a matching local `./dep` directory. This did not cover the valid subpath imports case where the target should be resolved as an external package dependency.

## Test plan

- Commit hook ran `lint-staged`, including `rustfmt`, Prettier, and ESLint for the staged files.
- Updated the `subpath-imports` snapshot fixture and output.

Made with [Cursor](https://cursor.com)


turbopack/crates/turbopack-core/src/resolve/mod.rs
mischnic, line 3286: If the non-`./`-appended pattern above matched, then this would be wrong.

daltino (commented):
## Good work on fixing subpath imports! 👍

This PR addresses an important gap in Turbopack's package.json imports handling. The changes look solid overall, but I have a few suggestions for improvement.

### Correctness & Implementation

✅ **Good**: Making `ExportImport` enum public to pass context between resolve functions is the right approach
✅ **Good**: Adding test coverage with a concrete `dep-in-nm` dependency scenario

### Issues & Suggestions

#### 1. **Incomplete patch context** 🔍
The main logic change in `resolve_into_package` is cut off in the diff. Could you share the complete implementation? It's hard to review the core fix without seeing:
- How `ExportImport` parameter is being used
- The logic that differentiates between exports and imports behavior
- How bare specifiers are now handled differently

#### 2. **Test coverage could be more comprehensive** 📝
```javascript
// Consider adding test cases for:
// - Mixed relative and bare specifier imports
// - Invalid/missing dependencies
// - Nested subpath patterns with dependencies
// - Edge case: circular dependency resolution
```

#### 3. **Missing package.json imports field** 📋
The test adds `#dep-in-nm` import but I don't see the corresponding `package.json` with the `imports` field definition. Should there be an update to the main package.json showing:
```json
{
  "imports": {
    "#dep-in-nm": "dep-in-nm"
  }
}
```

#### 4. **Documentation opportunity** 📚
Since this fixes a significant behavior difference between exports and imports, consider adding inline comments explaining:
- Why imports can resolve to bare specifiers while exports use relative paths
- The Node.js specification compliance aspect

### Minor Nits
- The test file could benefit from a comment explaining what `#dep-in-nm` is testing specifically
- Consider adding an assertion/test that verifies the resolved module actually comes from `node_modules`

### Performance
No concerns - the changes appear to add minimal overhead to the existing resolution logic.

Overall this looks like a solid fix! Just need to see the complete implementation and consider expanding test coverage slightly.

turbopack/crates/turbopack-core/src/resolve/mod.rs
yuzheng14: > If the non-`./`-appended pattern above matched, then this would be wrong.

This branch won't be reached when resolving Imports fields, because we're using `Pattern::Constant`. In that case, `req.as_constant_string()` will alwyas return `Some`, so the branch condition never matches.

PR #93915, opened 2026-05-18
[test] split `og-api` test suite into default and standalone variants

unstubbable:
The fixture's `next.config.js` had `output: 'standalone'` hardcoded, which is meaningless on Vercel deployments (the standalone directory is ignored) but became a build-time crash after #93684 made the Turbopack NFT generator skip `next-server.js.nft.json` whenever an adapter is configured. With `output: 'standalone'` still active, `writeStandaloneDirectory` then fails with `ENOENT` when it tries to read that file ([x-ref](https://github.com/vercel/next.js/actions/runs/26006483168/job/76441905221)).

This PR splits the `og-api` E2E suite into a default variant and a `standalone.test.ts` wrapper that enables `output: 'standalone'` via the `TEST_OUTPUT_STANDALONE` env var, and excludes the standalone wrapper from the deploy test manifest.

The default variant still runs in `deploy` mode and preserves coverage of `next/og` from Pages Router API routes, App Router route handlers (edge and node), and middleware — paths not covered by the metadata-* fixtures. The standalone variant continues to exercise the trace-copy assertion in `next start` mode.

### DIFF
--- a/test/deploy-tests-manifest.json
+++ b/test/deploy-tests-manifest.json
@@ -118,6 +118,7 @@
       "test/e2e/socket-io/index.test.ts",
       "test/e2e/middleware-matcher/index.test.ts",
       "test/e2e/next-script/index.test.ts",
+      "test/e2e/og-api/standalone.test.ts",
       "test/production/standalone-mode/**/*"
     ]
   }
--- a/test/deploy-tests-manifest.json
+++ b/test/deploy-tests-manifest.json
@@ -118,7 +118,6 @@
       "test/e2e/socket-io/index.test.ts",
       "test/e2e/middleware-matcher/index.test.ts",
       "test/e2e/next-script/index.test.ts",
-      "test/e2e/og-api/standalone.test.ts",
       "test/production/standalone-mode/**/*"
     ]
   }
--- a/test/e2e/og-api/index.test.ts
+++ b/test/e2e/og-api/index.test.ts
@@ -4,9 +4,11 @@ import fs from 'fs-extra'
 import { join } from 'path'
 
 describe('og-api', () => {
-  const { next } = nextTestSetup({
+  const { next, skipped } = nextTestSetup({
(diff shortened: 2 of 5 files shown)


test/e2e/og-api/index.test.ts
timneutkens, line 8:
```suggestion
    files: new FileRef(join(__dirname, 'app')),
    skipDeployment:  process.env.TEST_OUTPUT_STANDALONE === 'true',
```

timneutkens:
```suggestion
  const { next, skipped } = nextTestSetup({
```

timneutkens:
```suggestion
if (skipped) return

```

unstubbable: It's already excluded via the manifest.

test/deploy-tests-manifest.json
```diff
@@ -118,6 +118,7 @@
       "test/e2e/socket-io/index.test.ts",
       "test/e2e/middleware-matcher/index.test.ts",
       "test/e2e/next-script/index.test.ts",
+      "test/e2e/og-api/standalone.test.ts",
```
unstubbable:
```suggestion
```

test/e2e/og-api/index.test.ts
```diff
@@ -4,9 +4,11 @@ import fs from 'fs-extra'
 import { join } from 'path'
 
 describe('og-api', () => {
-  const { next } = nextTestSetup({
-    files: new FileRef(join(__dirname, 'app')),
+  const { next, skipped } = nextTestSetup({
+    files: new FileRef(join(__dirname, 'app')),
+    skipDeployment:  process.env.TEST_OUTPUT_STANDALONE === 'true',
   })
+if (skipped) return
```
unstubbable:
```suggestion

  if (skipped) return
```

timneutkens (approved):

PR #93917, opened 2026-05-18
[backport] Fix catch-all `router.query` corruption with `basePath` + `rewrites`

unstubbable:
Backports:

- #93294

### DIFF
--- a/packages/next/src/shared/lib/router/router.ts
+++ b/packages/next/src/shared/lib/router/router.ts
@@ -227,9 +227,14 @@ function getMiddlewareData<T extends FetchDataOutput>(
           const parsedSource = getNextPathnameInfo(
             parseRelativeUrl(source).pathname,
             {
-              nextConfig: process.env.__NEXT_HAS_REWRITES
-                ? undefined
-                : nextConfig,
+              // Pass basePath (and trailingSlash) so the basePath prefix is
+              // stripped before the `_next/data/` check, but omit `i18n` so the
+              // locale prefix is preserved here — the rewrite resolver below
+              // handles the locale-prefixed `as`.
+              nextConfig: {
+                basePath: nextConfig.basePath,
+                trailingSlash: nextConfig.trailingSlash,
+              },
               parseData: true,
             }
           )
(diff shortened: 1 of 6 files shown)


daltino (commented):
## Overall Assessment

This looks like a solid backport that addresses a specific edge case with router query corruption when using basePath + rewrites. The fix and test coverage are well-structured.

## Code Review

### ✅ Strengths

- **Clear intent**: The comment explains exactly why the config is being modified - stripping basePath but preserving locale prefix
- **Targeted fix**: The change is minimal and focused on the specific issue
- **Good test coverage**: Added comprehensive e2e test that reproduces the exact scenario
- **Realistic test setup**: Uses a proper middleware + basePath + rewrites configuration that matches real-world usage

### 🤔 Minor Concerns

**1. Incomplete patch display**
```typescript
// The router.ts patch appears truncated
+              // locale prefix is preserved here — the rewrite resolver below
+              // handles the locale-prefixed `as`
```
The actual code changes after the comment aren't visible. Could you confirm the complete implementation?

**2. Test assertion incomplete**
```typescript
// In the test file, this line is cut off:
expect(await browser.elementById('query-
```
What's the complete assertion checking?

**3. Test robustness**
Consider adding:
- Verification of the initial state before navigation
- Multiple query parameters to ensure comprehensive coverage
- Error cases (e.g., invalid paths)

### 📝 Suggestions

**Test enhancement example:**
```typescript
it('preserves router.query on client-side navigation to a catch-all page', async () => {
  const browser = await next.browser('/docs')
  
  // Verify initial state
  expect(await browser.url()).toContain('/docs')
  
  await browser.elementById('catchall-link').click()
  
  await retry(async () => {
    expect(await browser.elementById('page-title').text()).toBe('CatchAll')
  })
  
  // More specific query validation
  const queryPath = await browser.elementById('query-path').text()
  expect(JSON.parse(queryPath)).toEqual(expectedPathArray)
})
```

### 🎯 Questions

1. Does this fix handle nested catch-all routes (`[...path].tsx`) correctly?
2. Are there any performance implications of the config change during route parsing?
3. Should we add a test case for the scenario without middleware to ensure no regression?

## Verdict

The core fix appears sound and the test coverage is appropriate for a backport. Just need clarification on the truncated patches and consider the minor enhancements suggested above.

timneutkens (approved):

PR #93617, opened 2026-05-07
Hold `cacheSignal` across `unstable_cache` foreground revalidation

unstubbable:
When `unstable_cache` finds a stale entry during the prospective phase of a prerender, it dispatches a foreground revalidation and returns the in-flight promise. The wrapper's `cacheSignal.beginRead` / `endRead` pair was scoped only to the synchronous lookup.`endRead` in the outer `finally` fires the moment the function returns the promise, before the recompute and `cacheNewResult` have completed. The prospective prerender's `cacheSignal` then resolves `cacheReady` while the recompute is still in flight, ending the prospective phase early. Any `'use cache'` invocation rendered after the `unstable_cache` await in the same component never gets reached, so its RDC entry is never populated, and the final prerender phase throws "Unexpected cache miss after cache warming phase during prerendering."

Returning `await pendingRevalidates[invocationKey]` instead of returning the promise directly keeps the wrapper function suspended until the recompute and `cacheNewResult` actually complete. The outer `finally` then runs `endRead` at the right time, the prospective phase keeps waiting, and downstream `'use cache'` invocations are reached and populate the RDC as expected.

Adds a regression test under the `use-cache` suite that exercises an `unstable_cache` await followed by a downstream `'use cache'` component. Without the fix, the test fails consistently under `__NEXT_CACHE_COMPONENTS=true` with the cache-miss-after-warming error.

PR #93918, opened 2026-05-18
[backport] Encode non-ASCII characters in cache tags at construction

unstubbable:
Backports:

- #93617
- #93601

### DIFF
--- a/packages/next/src/server/lib/encode-cache-tag.ts
+++ b/packages/next/src/server/lib/encode-cache-tag.ts
@@ -1,0 +1,37 @@
+/**
+ * Percent-encode every character outside printable ASCII so a tag value can be
+ * safely serialized as part of the `x-next-cache-tags` HTTP header.
+ *
+ * Node's `validateHeaderValue` rejects any code unit outside `\t\x20-\x7e`, so
+ * a matched route path or user-supplied tag containing a non-ASCII character
+ * (Hebrew, Arabic, Chinese, emoji, …) would otherwise throw `ERR_INVALID_CHAR`
+ * and crash ISR on every affected request.
+ *
+ * This is applied at the public boundaries — tag construction
+ * (`getImplicitTags`, `validateTags`) and invalidation input (`revalidatePath`,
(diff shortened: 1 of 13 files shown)


daltino (commented):
## Overall Assessment

This is a solid backport that addresses an important issue with non-ASCII characters in cache tags causing ISR crashes. The implementation looks correct and well-tested.

## Positive Notes

✅ **Clear problem identification**: The comments clearly explain the issue with Node's `validateHeaderValue` rejecting non-ASCII characters

✅ **Good test coverage**: The test includes a surrogate-pair emoji (🎉) which exercises the more complex Unicode encoding scenarios

✅ **Consistent application**: The encoding is applied at all the right boundaries - tag construction, validation, and revalidation

✅ **Safe approach**: Percent-encoding is the correct solution for HTTP header safety

## Minor Suggestions

### 1. Missing Code in Patches
Several patches appear truncated in the diff:
- `encode-cache-tag.ts` cuts off mid-sentence in the comment
- `implicit-tags.test.ts` is incomplete 
- `patch-fetch.ts` comment is cut off

Could you verify the complete implementation is included?

### 2. Error Handling Consideration
While percent-encoding should handle most cases, consider if there are any edge cases where `encodeCacheTag()` might throw (e.g., malformed Unicode). The current usage doesn't seem to have error handling.

### 3. Performance Note
The encoding happens on every tag creation. For high-traffic sites with many ASCII-only tags, this adds overhead. Consider if a fast-path check for ASCII-only strings would be beneficial:

```typescript
// Quick ASCII check before encoding
if (/^[\x20-\x7E]*$/.test(tag)) {
  return tag;
}
// Proceed with encoding
```

## Questions

1. Are there any benchmarks showing the performance impact of encoding all tags?
2. Is the `encodeCacheTag` function handling all Unicode normalization forms correctly?

## Verdict

This addresses a real crash scenario and the approach is sound. The minor points above are suggestions for completeness rather than blockers.

timneutkens (approved):

PR #93921, opened 2026-05-18
Remove vercel approvers

gnoff:
Outdated. we don't really use these

### DIFF
--- a/packages/next/src/client/.vercel.approvers
+++ b/packages/next/src/client/.vercel.approvers
@@ -1,3 +1,0 @@
-use-intersection.tsx @timneutkens:notify
-use-intersection.tsx @ijjk:notify
-use-intersection.tsx @shuding:notify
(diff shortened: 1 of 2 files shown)


daltino (approved):
## ✅ LGTM - Clean removal of outdated approvers

This is a straightforward cleanup PR that removes outdated `.vercel.approvers` files. The changes look good:

### What's being removed:
- Client-side approvers for `use-intersection.tsx` 
- Server-side approvers for `config.ts` and `serve-static.ts`
- All entries were notification-only (`@username:notify`)

### Positives:
- Clean deletion with no leftover references
- Aligns with the stated goal of removing unused tooling
- No risk to functionality since these were just notification configs

### Minor consideration:
If there are any other `.vercel.approvers` files in the codebase that are also outdated, you might want to audit and remove those in a follow-up PR for consistency.

Great housekeeping! 🧹

timneutkens (approved):

PR #93930, opened 2026-05-19
fix: support pageExtensions in no-html-link-for-pages

zeinitsu03:


zeinitsu03:
Hi maintainers, thanks for taking a look.

This PR fixes #53473 by adding `pageExtensions` support to `@next/next/no-html-link-for-pages` while keeping the existing `string | string[]` `pagesDir` option behavior.

I also added a regression test for a Pages Router file using a compound extension like `about.page.tsx`.

Local checks run:
- `tsc --project packages/eslint-plugin-next/tsconfig.json --noEmit --skipLibCheck`
- focused regression check for the custom `pageExtensions` case

PR #93936, opened 2026-05-19
docs: use 403 for admin role check in API route examples

fecommunity:
Align Pages API route examples with Route Handlers in the authentication guide: return 403 when the user is authenticated but lacks admin privileges, not 401.

Fixes #80051


Dharya-dev (commented):
Small but great detail! A lot of people mix up 401 and 403, but since we already know who the user is (they just lack the admin role), 403 is 100% the correct HTTP status code here. Good stuff making the docs more accurate for everyone learning Next.js. LGTM 🔥

PR #93938, opened 2026-05-19
Remove `experimental.useNodeStreams` flag as it's enabled

timneutkens:
### What?

Remove `experimental.useNodeStreams` from the public config surface and make Node streams always-on for Node.js App Router rendering.

The smaller default-on change and dedicated CI cleanup landed in #94311. The standalone compatibility update landed separately in #94347.

### Why?

Node streams are now the default Node.js runtime rendering path, so the opt-in config and runtime selection plumbing are stale. Edge bundles continue to use the web-stream path because Node streams are unavailable there.

### How?

- Define `__NEXT_USE_NODE_STREAMS` as true for Node.js app runtime bundles and false for edge user bundles.
- Remove config, schema, runtime, export, and server plumbing for `experimental.useNodeStreams`.
- Delete the obsolete env-precedence fixture.
- Clean benchmark docs and scripts that still set or compare `experimental.useNodeStreams`.

### Verification

- `pnpm --filter=next types`
- `pnpm build-all`


bench/render-pipeline/benchmark.ts
```diff
@@ -16,7 +16,7 @@ const MINIMAL_SERVER = resolve(
 )
 
 type Scenario = 'minimal-server' | 'e2e'
-type StreamMode = 'web' | 'node' | 'both'
+type StreamMode = 'node'
```
unstubbable, line 19: Why don't you remove the mode completely?

unstubbable (approved):

PR #93940, opened 2026-05-19
Fix cross-param leak with `prefetch={true}` and `cachedNavigations`

unstubbable:
When both `cachedNavigations` and `varyParams` are enabled and a `<Link prefetch={true}>` points to a page with fallback params, the resulting Full prefetch could leak param-specific content into the cache entry of a different param value. As of today, `varyParams` tracking only works within the static stage portion of a response. A Full prefetch response covers all stages, so the server-reported set is incomplete and can't be trusted — it may arrive empty. The client then re-keys the entry under a vary path where every path param is replaced with `<Fallback>`, so a subsequent lookup for a different param value collides on that entry and reads the previously prefetched page's content.

This change makes `fulfillEntrySpawnedByRuntimePrefetch` skip the re-keying when `fetchStrategy === FetchStrategy.Full` and keep the entry pinned to its concrete vary path. That aligns Full prefetches with how they're already keyed when `cachedNavigations` is disabled, where the server sends no `varyParams` and the client never re-keys. The trade-off is that we miss out on cross-param sharing for Full prefetches even when a segment genuinely doesn't depend on a given param. Ideally we'd track `varyParams` through the dynamic stage too, but that's not possible today: the `varyParams` thenable would need to resolve before the render completes, while React waits for the thenable to resolve before completing the render — a deadlock in the Flight response.

This limitation should naturally go away as `prefetch={true}` evolves under Cache Components, where it will likely become a runtime prefetch — a render aborted at a known stage — at which point we can track `varyParams` properly across stages for these requests, likely switching from a thenable to an async iterable.


acdlite:
Could we fix the server response instead? ie emit `null` as the varyParams instead of an empty set

acdlite (approved):

unstubbable:
> Could we fix the server response instead? ie emit `null` as the varyParams instead of an empty set

Discussed out of band: On the server, we currently can not differentiate a full prefetch request from a dynamic RSC request. And for the latter we do need to track the `varyParams` for the static stage portion of the response.

PR #93944, opened 2026-05-19
Remove dead cacheComponents web streams path

timneutkens:
### What?

Remove the cacheComponents + web streams rendering branches from `app-render`.

### Why?

`cacheComponents` does not support the edge runtime, and Node streams are now always-on for the Node.js runtime in the parent PR. That makes the cacheComponents web-stream branches dead code.

### How?

- Collapse cacheComponents staged RSC rendering to the Node-stream implementation.
- Remove the web-stream cacheComponents helper functions and static-stage byte counter.
- Keep the ordinary non-cacheComponents web-stream paths for edge/runtime compatibility.

### Verification

- `pnpm --filter=next types`
- `pnpm --filter=next build`
- `pnpm test-start-webpack test/e2e/app-dir/app/standalone.test.ts`

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -902,178 +902,6 @@ async function generateDynamicFlightRenderResult(
 }
 
 /**
- * Production-only staged dynamic flight render for cache components. Uses
- * staged rendering to separate static (RDC-backed) from runtime/dynamic
- * content.
- */
-async function generateStagedDynamicFlightRenderResultWeb(
-  req: BaseNextRequest,
-  ctx: AppRenderContext,
-  requestStore: RequestStore
-): Promise<RenderResult> {
-  const { componentMod, workStore, renderOpts } = ctx
-  const { renderToReadableStream, routeModule } = componentMod
-  const { loaderTree } = routeModule.userland
-  const { onInstrumentationRequestError, experimental } = renderOpts
-  const { appShells } = experimental
-
-  function onFlightDataRenderError(err: DigestedError, silenceLog: boolean) {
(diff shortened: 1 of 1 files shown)


lubieowoce:
> **Verification**
>    pnpm --filter=next types
>    pnpm --filter=next build
>    pnpm test-start-webpack test/e2e/app-dir/app/standalone.test.ts

you can probably tell your agent to stop including this bit, CI does all of this and more anyway?

timneutkens:
> you can probably tell your agent to stop including this bit, CI does all of this and more anyway?

verification is part of the template. I didn't add it, but it seems relevant to add tbh. Though not as useful in cases like this one.

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -3593,537 +3331,269 @@ async function renderToStream(
...
-            staleTimeIterable,
-            selectStaleTime
+          setReactDebugChannel(
+            { readable: debugChannelClientStream.createReplayStream() },
+            htmlRequestId,
+            requestId
           )
+        }
+      } else if (cacheComponents && cachedNavigations) {
+        // MARK: nodeStreams cacheComponents RSC
```
lubieowoce, line 3455: are these `// MARK:` comments used for anything anymore? can we remove this? if they are, we should probably drop "nodeStreams" from here bc there's no non-nodeStreams now

lubieowoce (approved):

packages/next/src/server/app-render/app-render.tsx
timneutkens: It's still useful imo because there's a split between cacheComponents vs not. And the not case is also split on nodestreams vs webstreams.

PR #93949, opened 2026-05-19
[ci] Switch to precompiled action for upload-preview-tarballs

bgw:
Switching this job to a precompiled script means that we don't need to perform the messy checkout+pnpm install stuff, and we don't have to read from any caches.

We do this for a lot of other actions.

Aside from the fact that the compiled `dist` directory is near-impossible to audit, I think this is a good risk-reduction.

This is the same pattern that we use for https://github.com/vercel/next.js/blob/338d59cc46e1b56b56ab88c6bbadab867dcbef49/.github/workflows/pull_request_auto_label.yml

## Testing

It's not really practical to test this in CI prior to merging. In the worst case, it could break preview tarballs, which isn't the worst thing in the world. I'll keep an eye on the workflow after merging.


bgw:
* **#93949** <a href="https://app.graphite.com/github/pr/vercel/next.js/93949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93949?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

lukesandberg (changes requested):
can we add a test to ensure this stays up to date?

probably isn't the biggest deal if we update TSC and don't update this, but if we  bump vercel/blob it shouldn't be possible to not update this.

of course this just means we are shifting the cost of bundling from one CI job to another.... but something something turborepo?

bgw:
It wouldn't be a normal unit or e2e test because it requires `pnpm i` inside `.github` plus running `pnpm build`, but I can have claude create a custom GH action on `pull_request` that rebuilds and verifies all of these things in `.github`.

That's pretty non-trivial though and would need to be a second PR.

bgw:
Feedback from @eps1lon: https://vercel.slack.com/archives/C0B0LKK5QD7/p1779802692448059?thread_ts=1779221084.271219&cid=C0B0LKK5QD7

- The existing workflow does not have a security vulnerability -- I agree with this assessment. My thought was that this is more about risk reduction, especially with potential future changes, than there being a vulnerability today.
- It appears that @eps1lon believes this does not reduce complexity or risk.

PR #93954, opened 2026-05-20
use the action cache for passing tests

lukesandberg:
## Summary

Memoize passing tests across attempts of the same workflow run, so a re-attempt after a timeout or cancellation can skip what already passed.

## What changed

- `run-tests.js` appends each passing test's filename to `$NEXT_TEST_PASSED_FILE` as it finishes (sync `write` + `fsync`, fd kept open across calls). On startup it reads the same file and skips any entries that match.
- `build_reusable.yml` wraps the test step with `actions/cache/restore` (before) and `actions/cache/save` (after, with `if: always()`). The save step runs on success, failure, and cancel — which is the whole point: a cancelled attempt still gets to preserve its progress for retry.
- Cache key is scoped to `${input_step_key}-${run_id}-attempt${run_attempt}`, with `restore-keys` falling back to any earlier attempt of the same run.

The previous implementation wrote a single Turbo-remote-cache blob *after* `Promise.allSettled` resolved — so a cancellation (like [run 26127371192](https://github.com/vercel/next.js/actions/runs/26127371192/job/76845288867)) lost every passed test from that attempt. 

This also removes one dependency on the turbo-remote cache (there are still two more uses of `scripts/turbo-cache.mjs`)

### DIFF
--- a/.github/workflows/build_reusable.yml
+++ b/.github/workflows/build_reusable.yml
@@ -128,10 +128,7 @@ env:
   KV_REST_API_URL: ${{ secrets.KV_REST_API_URL }}
   KV_REST_API_TOKEN: ${{ secrets.KV_REST_API_TOKEN }}
   NEXT_TEST_JOB: 1
-  # Per-job memoization of passed test files. run-tests.js appends to
-  # this file as each test passes; the cache steps below restore it at
-  # job start and save it at job end (including on cancel) so that
-  # re-attempts skip what already passed.
+  # Per-job memoization of passed test files.
   NEXT_TEST_PASSED_FILE: .next-test-passed.txt
   VERCEL_TEST_TOKEN: ${{ secrets.VERCEL_TEST_TOKEN }}
   VERCEL_TEST_TEAM: vtest314-next-e2e-tests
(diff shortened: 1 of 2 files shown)


lukesandberg:
* **#93954** <a href="https://app.graphite.com/github/pr/vercel/next.js/93954?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93954?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

.github/workflows/build_reusable.yml
```diff
@@ -371,6 +373,17 @@ jobs:
...
+      # Restore the list of tests that already passed in an earlier
+      # attempt of this same workflow run.
+      - name: Restore passed-tests cache
+        if: ${{ inputs.afterBuild && runner.environment == 'github-hosted' }}
+        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: ${{ env.NEXT_TEST_PASSED_FILE }}
+          key: next-test-passed-${{ steps.var.outputs.input_step_key }}-${{ github.run_id }}-attempt${{ github.run_attempt }}
+          restore-keys: |
+            next-test-passed-${{ steps.var.outputs.input_step_key }}-${{ github.run_id }}-
```
bgw, line 385: Trailing hyphen?

run-tests.js
```diff
@@ -132,44 +131,34 @@ class TestProfile {
...
-      const parsed = JSON.parse(data.toString())
-      return new Set(parsed.passed || [])
-    } catch {
+      const data = fs.readFileSync(file, 'utf8')
+      const passed = new Set()
+      // Tolerate a partial trailing line from a hard kill mid-append by
+      // requiring an explicit '\n' terminator. Lines without it are
+      // dropped.
+      const lines = data.split('\n')
+      const last = data.endsWith('\n') ? lines.length : lines.length - 1
```
bgw: Use null terminated strings

bgw (approved):

.github/workflows/build_reusable.yml
lukesandberg: this will return the most recently written value that matches this prefix.  So the actual key contains a `run_attempt` after this hyphen, the hyphen just makes it slightly more specific

Issue #90433, opened 2026-02-24
Memory leak causing OOM still occurs in Next.js 16.0.10 (also tested with 16.0.1) when using output: standalone with fetch requests.


haoyanwang:
### Link to the code that reproduces this issue

https://github.com/austin-raey/next-memory-issue

### To Reproduce

1. Create a Next.js app with `output: 'standalone'`
2. Make fetch requests to an external API
3. With just 2 requests per second, OOM occurs after a few minutes

### Current vs. Expected behavior

- **Current**: Memory continuously grows until OOM after ~2 requests/second. GC cannot release the memory.
- **Expected**: Memory should stabilize, GC should work properly, no OOM.

### Provide environment information

```bash
- Platform: Linux
- Node: v22.16.0
- Next.js: 16.0.10
- output: standalone
- Router: App Router
```

### Which area(s) are affected? (Select all that apply)

Performance

### Which stage(s) are affected? (Select all that apply)

Other (Deployed)

### Additional context

Related to PR #88577 (fix for issue #85914). The fix should properly cancel both branches of a tee'd ReadableStream, but the issue still persists in 16.0.10.

@icyJoseph   I've tested with next@16.2.0-canary.51 (which should include the fix from #88577), but the memory leak still occurs.

Heap Timeline Analysis
The heap timeline shows that string and object types are being retained and continuously accumulated. These objects are referenced by cacheController

[nextjs-memory-leak.heaptimeline.zip](https://github.com/user-attachments/files/25521230/nextjs-memory-leak.heaptimeline.zip)

haoyanwang:
@timneutkens  I'm experiencing the same issue in Next.js 16.0.10 with `output: standalone`.

Heap timeline analysis shows `string` and `object` types are retained by `cacheController`. Reproduction is very easy - only ~2 requests/second is enough to trigger OOM.

Is the fix in #88577 complete? Happy to provide more details if needed.

timneutkens:
Please provide a reproduction of your own making. Looking at the heaptimeline it seems to be leaking something in your application code, but I can't tell if that's the case without the code.

icyJoseph:
@haoyanwang Could you make sure and verify that you did run your diagnostic with `next@canary`? You've been trying to open the issue for the past 10-15 minutes, perhaps you made a mistake while testing, worth trying 🙏

C-Jeril:
[...]

Issue #85914, opened 2025-11-07
Memory Leak with node 22 (also 20, 24) + fetch + Next 16.0.1 + `output: standalone`


austin-raey:
### Link to the code that reproduces this issue

https://github.com/austin-raey/next-memory-issue

### To Reproduce

1. Run `pnpm install`
2. In `src/app/page.tsx`, replace `"http://put-large-api-json-request-here.test"` with some JSON endpoint. I didn't want to put someone's URL here as to not DDOS them haha. Preferably one where you are not rate limited. You can probably separately set up a seperate folder with an [`express`](https://expressjs.com) (or similar) API endpoint that returns a large JSON payload as to not DOS anyone. As to not pollute the test, it may be best for you to not make this JSON endpoint in this example Next repo itself (this may not make a difference though tbf)
3. With Docker Desktop, Rancher Desktop, or etc running, do `make run`.
4. Open an Inspector session in Chrome / Edge / etc. (`about:inspect` in Chromium browsers)
  _4a._ Optionally, go to the `http://localhost:3000` page in your browser once, reload the page a few (2 or 3) times, then click "Collect garbage" a few times to ensure a clean heap, then "Take snapshot" of the heap to collect a baseline.

<details><summary>Image of the "Collect Garbage" button in Chromium browsers</summary>

![Image](https://github.com/user-attachments/assets/176bb8a5-8457-43e4-bf9c-2e5209ee9138)

</details>

5. In a seperate terminal tab in the example repository, run `pnpm exec autocannon -R 30 -f http://localhost:3000`. Let this run for a few minutes or so. `-R 30` may be adjusted for more requests, but 30 is probably fine.
6. After waiting, hit CTRL + C to stop autocannon.
7. Go to the Inspector. Before taking the heap snapshot, click the "Collect garbage" broom icon to ensure the heap is clean. Note the heap is now larger than baseline and collecting garbage will not reduce this size.
8. Inspecting the heap, it seems a large portion is now taken up by fetch. There is also `Array` and `InternalReadableByteStream` which seems suspicious, but I am not sure if they are relevant:

<details><summary>Image of the suspicious items in heap</summary>

![Image](https://github.com/user-attachments/assets/16b5d4fc-ddb9-4732-9d5d-03ce3546adb8)

![Image](https://github.com/user-attachments/assets/023dffce-403c-426a-9033-a896a0a412a7)

</details>

### Current vs. Expected behavior

[...]

PR #93965, opened 2026-05-20
fix(load-manifest): avoid per-page V8 context allocation in evalManifest

### LINKED ISSUE #93964: Native V8 context leak in load-manifest.external.ts: per-page client-reference-manifest evaluation accumulates contexts until OOM
### Summary

In production Next.js 16.2.6 standalone, RSS climbs linearly with the number of distinct app-router pages served, reaching OOM on apps with hundreds of pages. The leak is attributable to `vm.runInNewContext` in `packages/next/src/server/load-manifest.external.ts`. Each page's `*_client-reference-manifest.js` is evaluated once via `runInNewContext`, which allocates a fresh V8 native context (~3 to 4 MB residual RSS each) retained for the process lifetime by the manifest cache.

### Reproduction

* Next.js 16.2.6, `output: 'standalone'`, Node 24.15.0
* App router with ~360 prerendered pages
* Sustained crawler traffic that walks distinct pages
* 1 GB Fly machine

Result: RSS grows ~10 MB/min, OOM after roughly 30 to 60 minutes uptime. Restart, repeat.

### Diagnostic data

Captured via `kill -USR1 <pid>` + CDP `Runtime.evaluate` on the live process:

```
T1 (same PID): uptime  5 min, RSS 424 MB, nativeContextCount 28
T2 (same PID): uptime 14 min, RSS 511 MB, nativeContextCount 51
```

Same process, +9 min, +23 native contexts, +87 MB RSS. ~3.8 MB RSS per context.

Forced major GC via `HeapProfiler.collectGarbage` reclaims ~80 MB of heap but does not reduce `nativeContextCount`. Contexts are pinned by the load-manifest `sharedCache` `Map`, which retains the `contextObject` returned from each `evalManifest` call. The `contextObject` is the manifest's `globalThis`, so V8 keeps the entire native context alive.

Enumerating contexts via CDP `Runtime.executionContextCreated` shows they are all named `"VM Context N"` with barebones globals (V8 intrinsics only, no Edge runtime APIs). This rules out the Edge sandbox path and points at `vm.runInNewContext` in `load-manifest.external.ts` as the only producer.

### Code location

`packages/next/src/server/load-manifest.external.ts` in `evalManifest`:

```ts
let contextObject = {
  process: { env: { NEXT_DEPLOYMENT_ID: process.env.NEXT_DEPLOYMENT_ID } },
}
runInNewContext(content, contextObject)
```

Manifest content (`page_client-reference-manifest.js`) only does:

```js
globalThis.__RSC_MANIFEST = globalThis.__RSC_MANIFEST || {};
globalThis.__RSC_MANIFEST["/_not-found/page"] = { ... };
```

No prototype mutation, no `this` access, no eval, no implicit globals. The full V8 context isolation provided by `runInNewContext` is unused by the actual manifest content, but the cost is paid per page.

### Projection

~360 pages × ~3.8 MB ≈ 1.4 GB ceiling, before the fetch cache leak under investigation in #85914 / #90433 adds anything.

### Distinction from other reports

The reported memory leaks in #85914, #90433, #88603 all point at the fetch response cache. This is a different, additive source. Heap analysis on a process with both leaks active should show:

* fetch cache leak: string and object accumulation referenced from `cacheController`
* this leak: rising `nativeContextCount` and per-context `globalThis.__RSC_MANIFEST` retention

The two compound; apps with many pages will hit this one first.

### Proposed fix

Replace `runInNewContext` with a `Function` wrapper that shadows `globalThis` / `self` / `process` via parameters. Same sandbox semantics for current manifest content, no V8 context allocation. PR follows.

### Environment

* Next.js: 16.2.6
* Node: 24.15.0
* Output: standalone
* Container: 1 GB / 1 vCPU

ilanoh:
Closes #93964.

### What

Replace `vm.runInNewContext` with a `new Function` wrapper inside `evalManifest`.

### Why

Each `runInNewContext(content, contextObject)` allocates a fresh V8 native context that the `sharedCache` `Map` retains for the process lifetime via the returned `contextObject` (which IS the new context's `globalThis`). On `output: 'standalone'` deploys, every distinct page's `*_client-reference-manifest.js` triggers one such allocation through `tryLoadClientReferenceManifest`. Empirically each context costs ~3 to 4 MB residual RSS.

On app-router apps with ~360 pages this accumulates to ~1.4 GB before the fetch cache leak tracked in #85914 / #90433 adds anything. 1 GB containers OOM after 30 to 60 min uptime, restart, repeat. Diagnostic data and CDP samples are in #93964.

### How

The Function wrapper exposes `contextObject` as `globalThis`, `self`, and `process` inside the body, so the manifest's `globalThis.__RSC_MANIFEST[entry] = ...` writes still land on `contextObject` and the existing cache + `deepFreeze` path continues to work unchanged.

```ts
new Function('globalThis', 'self', 'process', content as string)(
  contextObject,
  contextObject,
  contextObject.process
)
```

### Known semantic divergence

Generated manifest content today only writes `globalThis.X = ...` and `self.Y = ...`. The Function wrapper differs from `runInNewContext` in:

* `this` in a non-strict Function body is the host global, not the sandbox. Current manifests don't use `this`.
* Top-level `var x` becomes a local to the wrapper rather than a property on `contextObject`. Current manifests don't declare top-level `var`.
* `Object.prototype` is shared with the host instead of isolated. Current manifests don't touch prototypes.

If any of those constraints are important to preserve for future manifest generators, alternatives are:

1. Reuse a single shared `vm.Context` across all manifests by namespacing reads/writes on a host-owned object.
2. Tear down the context after eval by extracting just the needed properties from `contextObject` and discarding the rest.
3. Compile-once-invoke-many: cache the compiled `Function` per manifest path.

Happy to follow guidance on which approach the team prefers; this PR takes the smallest-diff route that fixes the observed leak.

### Verified

Locally on a Next.js 16.2.6 standalone deploy (~360 pages), patched node_modules: `nativeContextCount` no longer grows per page first-access, RSS plateau at ~430 MB instead of climbing to OOM at ~1 GB. Will re-verify on canary once this lands.

### Notes for reviewers

* Removes the now-unused `runInNewContext` import.
* Comment block in source explains the trade-off so future readers don't restore `runInNewContext` without understanding the memory cost.
* Commit is unsigned; happy to re-push signed once a reviewer chimes in on the approach.

### DIFF
--- a/packages/next/src/server/load-manifest.external.ts
+++ b/packages/next/src/server/load-manifest.external.ts
@@ -2,7 +2,6 @@ import type { DeepReadonly } from '../shared/lib/deep-readonly'
 
 import { join } from 'path'
 import { readFileSync } from 'fs'
-import { runInNewContext } from 'vm'
 import { deepFreeze } from '../shared/lib/deep-freeze'
 
 const sharedCache = new Map<string, unknown>()
@@ -123,7 +122,23 @@ export function evalManifest<T extends object>(
   let contextObject = {
     process: { env: { NEXT_DEPLOYMENT_ID: process.env.NEXT_DEPLOYMENT_ID } },
   }
-  runInNewContext(content, contextObject)
+  // Evaluate the manifest in the existing V8 context instead of a fresh one.
+  // The previous `runInNewContext(content, contextObject)` allocated a new
+  // V8 native context per call; on app-router builds with many pages the
+  // per-page client-reference-manifest evaluations accumulate retained
+  // contexts (~3-4 MB RSS each) for the process lifetime via the sharedCache
+  // map, leading to OOM on memory-constrained standalone deployments. See
+  // the linked issue for diagnostic data.
+  //
+  // The Function wrapper exposes contextObject as `globalThis`, `self`, and
+  // `process` inside the body, preserving the manifest's existing
+  // `globalThis.__RSC_MANIFEST[entry] = ...` write pattern without spawning
+  // a new V8 context.
+  new Function('globalThis', 'self', 'process', content as string)(
+    contextObject,
+    contextObject,
+    contextObject.process
+  )
 
   // Freeze the context object so it cannot be modified if we're caching it.
   if (shouldCache) {


bykarantelicom:
Production data point for this PR (Next 16.3.0, App Router, Turbopack, `next start` in Docker, ~800 URLs / 543 `*_client-reference-manifest.js` files):

- V8 `number_of_native_contexts` grew 124 → 214 in 3.1 h of normal traffic (heap snapshot diff: +1.15M closures, +182k contexts, +40k Sets, all realm intrinsics), i.e. one retained context per first-hit manifest, ~2.1 MB each, ~58 MB/h until every manifest has been evaluated (ceiling here ≈1.1 GB on a 2.3 GB heap). Combined with anything else that leaks, this was the base of a ~40 h OOM cycle.
- Root cause confirmed in-container: `runInNewContext(content, contextObject)` contextifies `contextObject`, and `sharedCache` keeps that object, so the context can never be collected. 150 manifests evaluated the upstream way → 151 native contexts after a forced GC; the same 150 evaluated in a throwaway context with the result copied out as plain JSON → 2 contexts.
- The manifests are pure JSON (no functions; `JSON.parse(JSON.stringify(ctx))` deep-equals the original), so any variant that keeps the evaluated data out of the sandbox object fixes it. Our interim workaround, applied from `instrumentation.ts` without patching Next, wraps `vm.runInNewContext` only for calls whose code contains `__RSC_MANIFEST` and whose sandbox is a plain object: run in a throwaway sandbox, copy globals back as JSON, never contextify the cached object. After deploy: 214 distinct routes served → 9 native contexts (previously 124 at the same point), heap flat.

Happy to share the snapshot diff tooling or a minimal repro if useful. 16.3.3 and canary still have the `runInNewContext` + cache pattern.

PR #93970, opened 2026-05-20
Turbopack: respect the module-sync export condition

mischnic:
I set it to `::Unknown` because the runtime may or may not support module-sync. So let's just trace both for now

There is one problem though:

https://nodejs.org/docs/latest-v26.x/api/packages.html#conditional-exports
> "module-sync" - matches no matter the package is loaded via import, import() or require(). The format is expected to be ES modules that does not contain top-level await in its module graph - if it does, ERR_REQUIRE_ASYNC_MODULE will be thrown when the module is require()-ed.

So we have multiple options:
1. We definitely want to respect module-sync for tracing. We don't really have to throw ERR_REQUIRE_ASYNC_MODULE here, Node.js will do it at runtime anyway.
2. ~~What about bundling? Either~~
   - ~~Respect module-sync always, but don't throw ERR_REQUIRE_ASYNC_MODULE.~~
   - ~~Respect module-sync always, and do throw a build error the module requires TLA. This would require a separate validation pass though.~~
   - I'm not enabling it for bundling for now


turbopack/crates/turbopack-tracing/tests/unit.rs
```diff
@@ -123,10 +124,12 @@ static ALLOC: turbo_tasks_malloc::TurboMalloc = turbo_tasks_malloc::TurboMalloc;
 #[case::module_create_require_no_mixed("module-create-require-no-mixed")]
 // #[case::module_register("module-register")]
 // #[case::module_require("module-require")]
-// #[case::module_sync_condition_cjs("module-sync-condition-cjs")]
+#[case::module_sync_condition_cjs("module-sync-condition-cjs")]
+// Turbopack always includes the module-sync version, regardless of the current Node version
```
lukesandberg, line 128: technically next.js has a version compatibility going back to 20.9.0 and module-sync was backported in 20.19.0

but both of those versions are recently out of support so we shouldn't be too concerned...

is that your thinking? "this is fine, next needs to update minimum node to v22" or is it "this is a legit bug but this PR improves on the status quo" so lfg?

lukesandberg (approved):

turbopack/crates/turbopack-tracing/tests/unit.rs
mischnic: No. For tracing, it's fine to just include more file than are needed at runtime.
So we just include both: module-sync for Node versions that use it, and then also the following condition for the older versions that don't

PR #93972, opened 2026-05-20
Stabilize deploy cache tests

timneutkens:
### What?

- Retry Vercel build log inspection while deploy metadata is still propagating.
- Use the client-style `_rsc` URL for both `resume-data-cache` prefetch checks.

### Why?

Deploy CI was failing/flaking when `vercel inspect --logs` returned a ready deployment before metadata log lines appeared, and when a deploy-mode RSC prefetch reused `/` without the cache-busting query real clients send.

### How?

- Wrap `vercel inspect --logs` plus metadata parsing in a 60s retry for Vercel-backed deploy tests.
- Factor `getPrefetchRscUrl()` in `resume-data-cache.test.ts` and use it for the prefetch check after revalidation too.

### Verification

- `pnpm --filter=next build`
- `pnpm test-start-webpack test/e2e/app-dir/resume-data-cache/resume-data-cache.test.ts`
- `npx eslint --config eslint.config.mjs test/lib/next-modes/next-deploy.ts test/e2e/app-dir/resume-data-cache/resume-data-cache.test.ts`
- `pnpm tsc -p test/lib/tsconfig.json --noEmit`
- `git diff --check`
- Not run: `NEXT_TEST_MODE=deploy ...` (`requires Vercel CI deploy credentials/environment`)

### DIFF
--- a/test/e2e/app-dir/resume-data-cache/resume-data-cache.test.ts
+++ b/test/e2e/app-dir/resume-data-cache/resume-data-cache.test.ts
@@ -18,14 +18,7 @@ describe('resume-data-cache', () => {
   ])(
     'should have consistent data between static and dynamic renders with $name',
     async ({ id }) => {
-      // First render the page statically, getting the random number from the
-      // HTML.
-      let $ = await next.render$('/')
-      const first = $(`p#${id}`).text()
-
-      // Then get the Prefetch RSC and validate that it also contains the same
-      // random number.
-      await retry(async () => {
+      const getPrefetchRscUrl = async () => {
         const url = new URL('/', 'http://localhost')
 
         url.searchParams.set(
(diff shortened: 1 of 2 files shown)


test/e2e/app-dir/resume-data-cache/resume-data-cache.test.ts
```diff
@@ -38,8 +31,19 @@ describe('resume-data-cache', () => {
           )
         )
```
mischnic: Why do you need a getter function for the URL? Isn't the URL always the same and could just be a variable?

timneutkens: :robot: Good call, simplified this to compute the prefetch RSC URL once before the assertions and reuse the string.

PR #93974, opened 2026-05-20
request insights: record local framework spans (1/5)

feedthejim:
## What

Adds a dev-only local recorder for framework OTEL spans. This lets Next mirror span data into an in-memory store when local span capture is enabled, without requiring a user to run an OTEL provider.

Stack:

1. request insights: record local framework spans (1/5) ← current
2. request insights: derive request history and fetch data (2/5)
4. request insights: add agent diagnosis access (4/5)

## Review Focus

- Local recorder behavior when no OTEL provider is installed.
- Preserving existing OTEL span API behavior when a provider is installed.
- Request identity propagation through existing app render async storage.

## Proof In This PR

- `pnpm jest packages/next/src/server/lib/trace/tracer.test.ts packages/next/src/server/lib/trace/span-store.test.ts --runInBand`
- `pnpm --filter=next types`

## Deferred Coverage

- Request-level aggregation, fetch enrichment, dev transport, CLI/MCP access, and the DevTools panel are later PRs in this stack.

### DIFF
--- a/packages/next/src/server/lib/trace/tracer.ts
+++ b/packages/next/src/server/lib/trace/tracer.ts
@@ -224,7 +224,7 @@ class NextTracerImpl implements NextTracer {
     return trace.getTracer('next.js', '0.0.1')
   }
 
-  private isTracingEnabled(): boolean {
+  private isOpenTelemetryEnabled(): boolean {
     const activeSpan = trace.getSpan(context.active())
     if (activeSpan?.isRecording()) {
       return true
@@ -269,7 +269,7 @@ class NextTracerImpl implements NextTracer {
    * edge middleware which runs in a detached sandbox.
    */
   public runWithDetachedContext<T>(fn: () => T): T {
-    if (!NEXT_OTEL_PERFORMANCE_PREFIX && !this.isTracingEnabled()) {
+    if (!NEXT_OTEL_PERFORMANCE_PREFIX && !this.isOpenTelemetryEnabled()) {
       return fn()
     }
     return context.with(ROOT_CONTEXT, fn)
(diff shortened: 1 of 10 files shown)


packages/next/src/server/lib/trace/tracer.ts
```diff
@@ -360,11 +367,30 @@ class NextTracerImpl implements NextTracer {
       ...options.attributes,
     }
 
+    if (!tracingEnabled) {
+      return traceLocalSpanOnly(spanName, options, fn)
```
timneutkens: Should be development-only

packages/next/src/server/lib/trace/tracer.ts
```diff
@@ -360,11 +367,30 @@ class NextTracerImpl implements NextTracer {
...
+    }
+
     return context.with(spanContext.setValue(rootSpanIdKey, spanId), () =>
       this.getTracerInstance().startActiveSpan(
         spanName,
         options,
         (span: Span) => {
+          const spanStoreRecorder = localSpanStoreEnabled
```
timneutkens: Should be development-only

packages/next/src/server/lib/trace/tracer.ts
```diff
@@ -210,7 +225,8 @@ class NextTracerImpl implements NextTracer {
   }
 
   private isTracingEnabled(): boolean {
-    if (this.getActiveScopeSpan()?.isRecording()) {
+    const activeSpan = trace.getSpan(context.active())
+    if (activeSpan?.isRecording()) {
       return true
     }
```
mischnic: This is really "user tracing enabled" then.

timneutkens: :robot: Renamed the helper to isUserTracingEnabled and updated all call sites, clarifying that it detects user-configured OpenTelemetry tracing while framework performance tracing and local span recording are handled separately.

timneutkens: :robot: Follow-up: refined the helper name to isOpenTelemetryEnabled, which identifies the mechanism directly and avoids implying that application/user tracing itself is unsafe.

timneutkens (approved):

PR #93975, opened 2026-05-20
request insights: derive request history and fetch data (2/5)

feedthejim:
## What

Builds the bounded request history store on top of local spans and fetch metrics. Request Insights now keeps the last 100 requests, dedupes fetch records, and preserves raw span details for later verbose consumers.

Stack:

1. request insights: record local framework spans (1/5)
2. request insights: derive request history and fetch data (2/5) ← current
4. request insights: add agent diagnosis access (4/5)

## Review Focus

- The request insight schema shared by server, tools, and DevTools.
- Fetch cache/status mapping from existing `AppRender.fetch` spans.
- Dedupe between completed fetch spans and direct fetch metrics.

## Proof In This PR

- `pnpm jest packages/next/src/server/lib/trace/request-insights.test.ts packages/next/src/server/lib/patch-fetch.test.ts --runInBand`
- `pnpm --filter=next types`

## Deferred Coverage

- Browser/HMR transport, CLI/MCP access, and the visual panel are later PRs.

### DIFF
--- a/packages/next/src/server/lib/trace/request-insights.ts
+++ b/packages/next/src/server/lib/trace/request-insights.ts
@@ -1,0 +1,337 @@
+import type { AttributeValue } from 'next/dist/compiled/@opentelemetry/api'
+import type {
+  RequestInsight,
+  RequestInsightFetch,
+  RequestInsightsSnapshot,
+} from '../../../next-devtools/shared/request-insights'
+import type { SpanStoreRecord } from './span-store'
+export { isRequestInsightsEnabled } from './span-store'
+
+const MAX_REQUEST_INSIGHTS = 100
+const REQUEST_INSIGHTS_STORE_KEY = Symbol.for('@next/request-insights-store')
+
+type RequestInsightsListener = (insight: RequestInsight) => void
+type RequestInsightIdentity = {
+  requestId?: string
+  htmlRequestId?: string
+  route?: string
+  url?: string
+}
+
+const REDACTED_VALUE = 'redacted'
+const SAFE_SPAN_ATTRIBUTE_KEYS = new Set([
+  'http.method',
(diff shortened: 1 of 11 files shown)


packages/next/src/server/lib/trace/request-insights.ts
```diff
@@ -0,0 +1,337 @@
...
+  'next.fetch.cache_status',
+  'next.fetch.idx',
+  'next.route',
+  'next.rsc',
+  'next.segment',
+  'next.span_name',
+  'next.span_type',
+])
+const SENSITIVE_PARAM_NAME_RE =
+  /(?:^|[_-])(?:access[_-]?token|api[_-]?key|auth|authorization|code|cookie|credential|id[_-]?token|jwt|key|password|secret|session|signature|sig|token)(?:$|[_-])/i
```
mischnic, line 39: This looks overly specific? it would block `access-token` but not `accesstoken`?
I'd just do "contains any of these"?

packages/next/src/server/lib/trace/request-insights.ts
```diff
@@ -0,0 +1,337 @@
...
+
+class InMemoryRequestInsightsStore {
```
mischnic, line 41: Why do we need both `InMemorySpanStore` and `InMemoryRequestInsightsStore`?
Isn't everything just a span in the end, regardless of where it came from?

packages/next/src/server/lib/trace/request-insights.ts
timneutkens: Let's keep it as-is and verify if it's even relevant first by dogfooding the tab. I suspect it's not needed to even have this filter.

timneutkens (approved):

PR #93977, opened 2026-05-20
request insights: add agent diagnosis access (4/5)

feedthejim:
## What

Adds `next experimental-diagnose` and an MCP `get_request_insights` tool so agents and local users can inspect Request Insights without opening the DevTools overlay.

Stack:

1. request insights: record local framework spans (1/5)
2. request insights: derive request history and fetch data (2/5)
4. request insights: add agent diagnosis access (4/5) ← current

## Review Focus

- CLI error handling and raw JSON output.
- MCP gating when Request Insights is unavailable.
- Telemetry registration for the new MCP tool name.

## Proof In This PR

- `pnpm jest packages/next/src/telemetry/events/mcp-telemetry.test.ts --runInBand`
- `pnpm --filter=next types`
- Stack-tip smoke verifies `next experimental-diagnose --url http://localhost:3003 --json` against a real dev server.

## Deferred Coverage

- A more curated agent summary format is intentionally deferred until the raw private snapshot contract settles.


packages/next/src/cli/next-diagnose.ts
```diff
@@ -0,0 +1,86 @@
+import { printAndExit } from '../server/lib/utils'
+
+export type NextDiagnoseOptions = {
+  url?: string
+  json?: boolean
+}
+
+export async function nextDiagnose(options: NextDiagnoseOptions) {
+  const devServerUrl = normalizeDevServerUrl(
+    options.url ?? 'http://localhost:3000'
```
timneutkens: This hardcoded url is not ideal imo. Is there another way to handle it. For example making the dev server url written to a file in distDir (.next) or such.

packages/next/src/cli/next-diagnose.ts
```diff
@@ -0,0 +1,86 @@
...
+  )
+  const endpoint = new URL('/__nextjs_request_insights', devServerUrl)
```
timneutkens: This url should be under `/_next/development` imo.

packages/next/src/cli/next-diagnose.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (!response.ok) {
+    printAndExit(data?.error ?? `Request failed with ${response.status}`, 1)
+  }
+
+  if (options.json) {
+    console.log(JSON.stringify(data, null, 2))
+    return
+  }
+
+  const requests = Array.isArray(data.requests) ? data.requests : []
```
timneutkens: If requests is not an array it should printAndExit instead of hiding the failure

packages/next/src/cli/next-diagnose.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (requests.length === 0) {
+    console.log('No request insights captured yet.')
+    return
+  }
+
+  for (const request of requests.slice(-20).reverse()) {
```
timneutkens: requests is sliced by 20 but this should be configurable so that the agent can do pattern matching if needed.

timneutkens: The amount of requests logged should be logged as well so that the agent knows it's "the last 20 requests" instead of an assumption around that it's all requests.

packages/next/src/cli/next-diagnose.ts
```diff
@@ -0,0 +1,86 @@
...
+    const route = request.route ?? request.url ?? request.requestId
+    const duration = formatDuration(request.durationMs)
+    const fetches = Array.isArray(request.fetches) ? request.fetches : []
+    console.log(`${route} ${duration} ${request.status ?? 'pending'}`)
+    console.log(
+      `  request ${shortId(request.requestId)} page ${shortId(request.htmlRequestId)}`
+    )
+
+    for (const fetch of fetches.slice(0, 5)) {
```
timneutkens: Only 5 fetches are being shown. This should be logged so that the agent knows it's a smaller subset instead of all requests.

packages/next/src/cli/next-diagnose.ts
timneutkens: :robot: Done -- the command now discovers the running server from the project's dev lockfile, including custom distDir support and a short startup retry. --url remains available as an explicit override, and the lockfile now stores the actual __NEXT_PRIVATE_ORIGIN instead of hardcoded localhost.

packages/next/src/cli/next-diagnose.ts
timneutkens: :robot: Done -- formatted output now states exactly how many retained requests are shown and that they are ordered newest first.

packages/next/src/cli/next-diagnose.ts
timneutkens: :robot: Done -- the CLI now validates that requests is an array and that each request has a fetches array before either formatted or --json output, and exits with an explicit invalid-response error otherwise.

packages/next/src/cli/next-diagnose.ts
timneutkens: :robot: Done -- when fetch output is truncated, the CLI now reports that it is showing the first 5 of the request's total fetches.

packages/next/src/cli/next-diagnose.ts
timneutkens: :robot: Done -- added --limit <count> for formatted output, with the existing 20-request behavior as the default.

packages/next/src/cli/next-diagnose.ts
timneutkens: :robot: Done -- moved the HTTP endpoint to /_next/development/request-insights and shared the path constant between the CLI and router.

packages/next/src/bin/next.ts
```diff
@@ -600,6 +601,35 @@ program
   )
   .usage('[directory] [options]')
 
+program
+  .command('experimental-diagnose')
+  .description(
+    'Inspect experimental Request Insights from a running Next.js dev server.'
+  )
```
mischnic: The name is a bit misleading.`diagnose` sounds like it does a lot but in reality it just returns the Request Insight spans

timneutkens (approved):

PR #93978, opened 2026-05-20
request insights: add DevTools request panel (5/5)

feedthejim:
## What?

Adds the experimental Request Insights panel to the Next.js DevTools overlay.

The panel provides:

- A retained request list with stable selection as new requests arrive.
- A “Page load” marker for the exact initial document request.
- An end-to-end trace timeline that follows the recorded span hierarchy.
- A focused default view of high-level spans and a verbose view for deeper investigation.
- Fetch, cache, request, status, duration, and span-count details.
- Readable sub-millisecond durations and consistent right-aligned request metrics.

This PR is intentionally limited to the Request Insights UI. Additional OpenTelemetry coverage, span display names, and Server Action tracing are split into the dependent PRs below.

## Why?

The Request Insights data was already available to tools, but inspecting it required consuming the development endpoint or CLI output. The DevTools panel makes the same request, fetch, cache, and timing information directly inspectable while developing an application.

The focused/verbose split keeps the default trace useful for understanding the request lifecycle without hiding the lower-level spans needed for investigation.

## How?

- Adds a Request Insights panel and request list to the DevTools overlay.
- Builds timeline rows from recorded parent/child relationships instead of assigning synthetic sections.
- Uses the request’s full duration as the timeline range and clips outlier spans to that range.
- Merges recorded fetch metrics with their corresponding fetch spans.
- Preserves the selected request ID until that request is no longer retained.
- Compares `request.requestId` with `self.__next_r` to identify only the initial document request.

## Stack

The first four Request Insights foundation PRs have already merged. The remaining review stack is:

1. **#93978 — Request Insights UI (this PR)**
2. #95765 — Missing OpenTelemetry spans
3. #95766 — Human-readable OpenTelemetry span names
4. #95767 — Server Action tracing and resolution

## Verification

- `pnpm --filter=next build`
- `pnpm exec jest packages/next/src/next-devtools/dev-overlay/components/request-insights/trace-viewer.test.ts --runInBand`


packages/next/src/server/app-render/action-handler.ts
```diff
@@ -1410,30 +1473,39 @@ function areAllActionIdsValid(
...
 
       if (isInvalidStringActionDescriptor(actionField, serverModuleMap)) {
-        return false
+        return null
+      }
+      if (!seenActions.has(key)) {
+        seenActions.add(key)
+        actionId = actionField.slice(
+          ACTION_DESCRIPTOR_ID_PREFIX.length,
+          ACTION_DESCRIPTOR_ID_PREFIX.length + ACTION_ID_EXPECTED_LENGTH
```
timneutkens: This seems to not check if the `ACTION_DESCRIPTOR_ID_PREFIX` is part of the `actionField`.

packages/next/src/server/app-render/action-handler.ts
```diff
@@ -1394,11 +1456,12 @@ const ACTION_ID_EXPECTED_LENGTH = 42
  * It pre-parses the FormData to ensure that any action IDs referred to are actual action IDs for
  * this Next.js application.
  */
-function areAllActionIdsValid(
+function getValidatedMPAActionId(
```
timneutkens: This function should still be `areAllActionIdsValid` for production. Only return the ID in development. Since this is a hot path for server actions it should be optimized and `areAllActionIdsValid` is more optimized than the newly changed function.

packages/next/src/server/app-render/action-handler.ts
```diff
@@ -1321,8 +1359,32 @@ async function executeActionAndPrepareForRender<
...
-    const actionResult = await workUnitAsyncStorage.run(requestStore, () =>
-      action.apply(null, args)
+    const actionResult = await getTracer().trace(
+      AppRenderSpan.executeServerAction,
+      {
+        spanName: `run Server Action ${serverAction.name}`,
+        attributes: {
+          'next.span.category': 'application',
+          'next.server_action.name': serverAction.name,
+          'next.server_action.file': serverAction.file || undefined,
```
timneutkens: Is this data available in production?

packages/next/src/server/app-render/action-handler.ts
timneutkens: Looking at the rest of the implementation it is potentially valid that it reads the function name. However, it's unclear to me if this is available in production or not.

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -3326,6 +3369,27 @@ async function renderToStream(
...
+      AppRenderSpan.initializeRender,
+      {
+        startTime: initializationStart,
+        attributes: {
+          'next.span_name': 'initialize app render',
+          'next.span_type': AppRenderSpan.initializeRender,
+        },
+      }
+    )
+    initializationSpan.updateName('initialize app render')
```
timneutkens: It seems `initializationSpan.updateName` is incorrect now that `next.span_name` is used. All usage should be updated.

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -3355,6 +3419,21 @@ async function renderToStream(
     renderSpan.end()
   }
 
+  const trackHTMLRenderCompletion = (allReady: Promise<void>) => {
+    const completion = getTracer().trace(
+      AppRenderSpan.waitForHTMLCompletion,
+      { spanName: 'wait for HTML completion' },
```
timneutkens: Should be `next.span_name` instead of `spanName`

packages/next/src/next-devtools/dev-overlay/components/request-insights/request-insights-panel.tsx
```diff
@@ -0,0 +1,646 @@
...
+    kind: isRsc ? 'RSC request' : 'HTML request',
+    fetchSummary: request.fetches.length
+      ? `${request.fetches.length} fetch${request.fetches.length === 1 ? '' : 'es'}`
+      : 'No fetches',
+    cacheSummary:
+      request.fetches.length === 0
+        ? 'No cache data'
+        : knownCacheCount === 0
+          ? 'Cache status unknown'
+          : `Cache ${cacheCounts.hit} hit, ${cacheCounts.skip} skip${
```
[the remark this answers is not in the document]
timneutkens: :robot: Fixed in e289ccec27. The cache summary now includes the tracked miss count alongside hits, skips, and unknown statuses.

packages/next/src/server/app-render/app-render.tsx
timneutkens: :robot: Fixed in e289ccec27. All spanName options added by this PR were replaced with next.span_name attributes, including wait for HTML completion.

packages/next/src/server/app-render/action-handler.ts
timneutkens: :robot: Fixed in e289ccec27. Descriptor ID extraction now goes through getActionIdFromStringDescriptor, which validates the full descriptor (including ACTION_DESCRIPTOR_ID_PREFIX) before slicing out the ID.

packages/next/src/server/app-render/action-handler.ts
timneutkens: :robot: Fixed in e289ccec27. Production uses the original areAllActionIdsValid path; only the dev-server branch extracts an action ID for display metadata. The function name/file lookup is also gated behind __NEXT_DEV_SERVER.

packages/next/src/server/app-render/app-render.tsx
timneutkens: :robot: Fixed in e289ccec27. I removed every updateName call added by this PR; manually created spans now carry the human-readable label only in next.span_name while retaining the stable internal span name.

packages/next/src/server/app-render/action-handler.ts
timneutkens: :robot: Fixed in e289ccec27. The new Server Action metadata lookup and tracing are now dev-server-only. Production keeps the direct workUnitAsyncStorage.run/action.apply path without reading or attaching the display metadata.

timneutkens (approved):

PR #93982, opened 2026-05-20
fix : cap react debug channel data sent over hmr in dev

### LINKED ISSUE #93967: Excessive data sent over hmr socket to client, causing renderer to crash

https://github.com/irnc/next-hmr-excessive-data

### To Reproduce

1. Start app using `npm run dev`
2. Open Chrome DevTools > Network > `hmr` socket
3. Notice that string returned from `readFile` are sent to client in full.

In the reproduction code I create a JSON file with array of 10m numbers, which results in 76 MB file. Content of this file is read using `fs.readFile`.

https://github.com/irnc/next-hmr-excessive-data/blob/68501fa4d47ef0a39106e5098f3241acf604d5cd/app/page.tsx#L1-L15

Running Next.js in dev mode I see that content of that file is streamed to browser using `hmr` socket. This data is used to show _suspended by_ in React DevTools.

It may be helpful for debugging. It works on small data. But on my local project I read JSON files of 300 MB in server components (data is consumed on server, not passed to client), which is sent in full to a browser over hmr socket, which causes render process crash.

Thus app in dev mode crashes browser render process.

```bash
Operating System:
  Platform: linux
  Arch: arm64
  Version: #1 SMP Fri May  1 14:35:41 UTC 2026
  Available memory (MB): 15972
  Available CPU cores: 14
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: 1.22.22
  pnpm: 10.31.0
Relevant Packages:
  next: 16.3.0-canary.24 // Latest available version is detected (16.3.0-canary.24).
  eslint-config-next: N/A
  react: 19.2.6
  react-dom: 19.2.6
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

React

next dev (local)

### Additional context

_No response_

nehaprasad-dev:
#### what?

- Cap dev HMR debug channel traffic at 1 MB.

#### Why?

- Large server-only reads were forwarded whole to the browser and could crash the renderer.

#### How?

- Limit bytes in connectReactDebugChannel, truncate the last chunk, cancel the reader, send chunk: null.

Fixes : #93967

### DIFF
--- a/packages/next/src/server/dev/debug-channel.test.ts
+++ b/packages/next/src/server/dev/debug-channel.test.ts
@@ -1,0 +1,39 @@
+import { takeReactDebugChunkForHmr } from './debug-channel'
+
+describe('takeReactDebugChunkForHmr', () => {
+  it('passes through chunks under the limit', () => {
+    const chunk = new Uint8Array([1, 2, 3])
+    const result = takeReactDebugChunkForHmr(chunk, 0, 10)
+
+    expect(result.chunk).toBe(chunk)
+    expect(result.bytesSent).toBe(3)
+    expect(result.done).toBe(false)
+  })
+
+  it('truncates a chunk that would exceed the limit', () => {
+    const chunk = new Uint8Array([1, 2, 3, 4, 5])
+    const result = takeReactDebugChunkForHmr(chunk, 3, 10)
+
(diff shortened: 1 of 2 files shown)


unstubbable:
This is not the right fix, see https://github.com/vercel/next.js/issues/93967#issuecomment-4620359682.

PR #93955, opened 2026-05-20
Replace VcStorage with real TurboTasks in tests/benches


lukesandberg:
`VcStorage` is a partial in-memory implementor of `TurboTasksApi`/`TurboTasksCallApi` in `turbo-tasks-testing`. It gave a handful of tests/benchmarks a way to allocate cells and call functions without spinning up a real backend, at the cost of stubbing roughly half of the trait surface with `unreachable!()`/`unimplemented!()`.

This PR migrates the four call sites to construct a real `TurboTasks` over a `noop_backing_storage()`, then deletes `VcStorage` and the test-only `with_turbo_tasks_for_testing`/`current_task_for_testing` helpers it was the sole consumer of. After this PR, `TurboTasks<B>` is the only concrete implementor of `dyn TurboTasksApi`.

### Migrated call sites

- **`turbo-tasks-fs` `with_extension` and `file_stem` tests** — direct `tt.run_once(async move { … })` replacement, matching the existing `test_try_from_sys_path` test in the same file.
- **`turbopack-ecmascript` `analyzer::tests::fixture`** — the body is now in a `#[turbo_tasks::function(operation, root)] async fn fixture_op(input: RcStr)` invoked via `fixture_op(input).read_strongly_consistent().await`. The synchronous SWC parser+resolver setup is wrapped in `GLOBALS.set(&globals, || …)` returning owned `(eval_context, var_graph)`, after which the async resolve/link work proceeds without GLOBALS — mirroring the production pattern in `analyze_ecmascript_module_internal`. `run_test` is no longer needed.
- **`turbopack-ecmascript` `bench_link`** — restructured to mirror `turbo-tasks-backend/benches/overhead.rs`: `iter_custom` with `TurboTasks` constructed per measurement batch (`storage_mode: None`, `dependency_tracking: false`), `Instant::now()` placed after construction so it isn't timed. `BenchInput.var_graph` is now `Arc<VarGraph>` so the future can be `Send + 'static`. Quick run shows no change in performance for `link/typeof`.

### Deleted

- `VcStorage` and its trait impls (`turbo-tasks-testing/src/lib.rs` shrinks from 339 lines to 9).
- `with_turbo_tasks_for_testing`, `current_task_for_testing`, and the `test_helpers` re-export module — all unused after `VcStorage` is gone.

### Verification

- `cargo check --workspace --tests --benches` — clean.
- `cargo clippy` on changed crates — clean.
- `cargo test -p turbo-tasks-fs --lib` — 110/110 pass.
[...]

PR #93910, opened 2026-05-18
[turbo-tasks]: Manually devirtualize our internal calling conventions

lukesandberg:
Manually devirtualize our main TurboTasksApi trait using extern "Rust" functions inspired by bun

This does work, dissassembly reveals that functions like `dynamic_call` do get devirtualized and often inlined.  However, it does not improve performance :-(

Open issues:
* this requires a lot of manual work to ensure implementations get linked, not great

PR #93983, opened 2026-05-20
Simplify turbo-tasks-backend: collapse single-impl traits and remove redundant Arc layers

lukesandberg:
### What?

Simplifies `turbo-tasks-backend` by removing abstraction layers that each had a
single implementation, plus several redundant `Arc` indirections that were no
longer load-bearing.

**Traits removed / collapsed (each had exactly one implementor):**

- **`KeyValueDatabase`** — folded into inherent methods on its only impl,
  `TurboKeyValueDatabase`.
- **`BackingStorage` + `BackingStorageSealed`** — folded into inherent methods on
  the renamed concrete type `TurboBackingStorage` (was the
  `KeyValueDatabaseBackingStorage<T>` generic + a type alias).
- **`ConcurrentWriteBatch`** — folded into inherent methods on `TurboWriteBatch`
  (it existed to support an old backend implementation that's gone).
- The **`B: BackingStorage` generic parameter** is dropped from
  `TurboTasksBackend`, `TurboTasksBackendInner`, `ExecuteContextImpl`, and
  `ChildExecuteContextImpl`. Every instantiation in the workspace already
  resolved to one concrete type.

The only remaining trait in the crate's public role is `Backend` (defined in
`turbo-tasks`), which stays — removing it would invert the
`turbo-tasks` → backend dependency.

**`Arc` indirections removed:**

- `TurboKeyValueDatabase.db: Arc<TurboPersistence>` → `TurboPersistence`
  (the write batch only borrowed it; the clone in `new` was gratuitous).
- The `self: &Arc<Self>` receivers and `self.clone()` calls in `run_backend_job`,
  `idle_start`, and `try_read_task_output` — the deferred/background work reaches
  the backend through the pinned `turbo_tasks` handle instead, whose `Arc` already
  keeps the backend alive.
- With no remaining sharer, the backend's own `Arc<TurboTasksBackendInner>` and
  the `TurboTasksBackend` newtype were collapsed: the inner fields now live
  directly in `TurboTasksBackend`, stored inline in `TurboTasks` (which already
  shares it via its own `Arc<TurboTasks>`).

**Other cleanups surfaced by the above:**

- `noop_backing_storage()` is now an empty, read-only instance of the real
  `TurboPersistence` instead of a separate `NoopKvDb` impl — same concrete type as
  `turbo_backing_storage()`, so the `Either<…>` wrappers in napi and turbopack-cli
  go away.
- `TurboTasksBackend::backing_storage()` accessor replaced with a focused
  `invalidate_storage(reason_code)` method, so the backend no longer leaks its
  storage object.
- Dead code removed (`CellDependency::key`, `ExecuteContext::schedule` /
  `suspending_requested`, the non-stats `execute_with_stats` variant) and
  feature/test-only items gated behind their corresponding `cfg`.

### Why?

The no-op storage work removed the last thing forcing these abstractions to be
generic/dynamic, so the single-impl traits and the extra `Arc` layers were pure
indirection. This is a self-contained simplification: behavior is unchanged.

### Tradeoff

- One `Either`/match in every backing-storage method becomes one `is_empty`
  atomic load (`Relaxed` `AtomicBool`) in `should_restore` — essentially
  equivalent.
- The backend is now stored inline in `TurboTasks` rather than behind a pointer:
  one fewer indirection on every backend access, at the cost of a larger
  `TurboTasks` struct.


lukesandberg:
* **#93910** <a href="https://app.graphite.com/github/pr/vercel/next.js/93910?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93983** <a href="https://app.graphite.com/github/pr/vercel/next.js/93983?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/93983?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94007** <a href="https://app.graphite.com/github/pr/vercel/next.js/94007?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#93955** <a href="https://app.graphite.com/github/pr/vercel/next.js/93955?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

crates/next-build-test/src/lib.rs
```diff
@@ -15,11 +15,11 @@ use turbo_rcstr::{RcStr, rcstr};
 use turbo_tasks::{
     Effects, ReadConsistency, ReadRef, ResolvedVc, TransientInstance, TurboTasks, Vc, take_effects,
 };
-use turbo_tasks_backend::{NoopBackingStorage, TurboTasksBackend};
+use turbo_tasks_backend::{TurboBackingStorage, TurboTasksBackend};
 use turbo_tasks_malloc::TurboMalloc;
 
 pub async fn main_inner(
-    tt: &TurboTasks<TurboTasksBackend<NoopBackingStorage>>,
+    tt: &TurboTasks<TurboTasksBackend<TurboBackingStorage>>,
```
bgw: If we only have one implementation of BackingStorage now, we can likely eliminate the BackingStorage trait

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -280,15 +280,43 @@ impl<S: ParallelScheduler + Default, const FAMILIES: usize> TurboPersistence<S,
...
+            )
+        };
+        let (value_estimated_items, value_weight_capacity) = if empty && read_only {
+            (0, 0)
+        } else {
+            (
+                VALUE_BLOCK_CACHE_SIZE as usize / VALUE_BLOCK_AVG_SIZE,
+                VALUE_BLOCK_CACHE_SIZE,
+            )
+        };
```
bgw: Nit

```suggestion
        let (
            key_estimated_items,
            key_weight_capacity,
            value_estimated_items,
            value_weight_capacity,
        ) = if empty && read_only {
            (0, 0, 0, 0)
        } else {
            (
                KEY_BLOCK_CACHE_SIZE as usize / KEY_BLOCK_AVG_SIZE,
                KEY_BLOCK_CACHE_SIZE,
                VALUE_BLOCK_CACHE_SIZE as usize / VALUE_BLOCK_AVG_SIZE,
                VALUE_BLOCK_CACHE_SIZE,
            )
        };
```

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -280,15 +280,43 @@ impl<S: ParallelScheduler + Default, const FAMILIES: usize> TurboPersistence<S,
...
         Self::open_read_only_with_parallel_scheduler(path, config, Default::default())
     }
+
+    /// Construct an empty, read-only `TurboPersistence` that owns no on-disk state and never
+    /// touches the filesystem. Reads return None; writes bail via the existing `read_only` guard.
+    /// Used to provide a "noop" backing storage with the same concrete type as the real one.
+    pub fn empty_in_memory_with_config(config: DbConfig<FAMILIES>) -> Self {
+        // `path` is `PathBuf::new()` but never read because `meta_files` is empty and
+        // `read_only` is true (so no write/compaction path is reachable).
+        Self::new(PathBuf::new(), true, true, Default::default(), config)
```
bgw:
```suggestion
        Self::new(PathBuf::new(), /*read_only*/ true, /*empty*/ true, Default::default(), config)
```

Alternative options:
- Define a struct for these
- Define an enum each value rather than using bool

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -336,7 +364,7 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
         config: DbConfig<FAMILIES>,
         parallel_scheduler: S,
     ) -> Result<Self> {
-        let mut db = Self::new(path, false, parallel_scheduler, config);
+        let mut db = Self::new(path, false, false, parallel_scheduler, config);
```
bgw:
```suggestion
        let mut db = Self::new(path, /*read_only*/ false, /*empty*/ false, parallel_scheduler, config);
```

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -348,7 +376,7 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
         config: DbConfig<FAMILIES>,
         parallel_scheduler: S,
     ) -> Result<Self> {
-        let mut db = Self::new(path, true, parallel_scheduler, config);
+        let mut db = Self::new(path, true, false, parallel_scheduler, config);
```
bgw:
```suggestion
        let mut db = Self::new(path, /*read_only*/ true, /*empty*/ false, parallel_scheduler, config);
```

bgw (commented):
I'll review after you restack, the diff is a bit messed up right now.

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: went with a struct

crates/next-build-test/src/lib.rs
lukesandberg: this is done

BackingStorage

quite the thread to pull on, this revealed that various other methods were dead

we can do the same thing to TaskGuard and ExecutionContext but i think we should prefer that work

lukesandberg: s/prefer/defer

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -280,15 +288,50 @@ impl<S: ParallelScheduler + Default, const FAMILIES: usize> TurboPersistence<S,
...
-        read_only: bool,
-        parallel_scheduler: S,
-        config: DbConfig<FAMILIES>,
+        OpenOpts {
+            path,
+            read_only,
+            empty,
+            parallel_scheduler,
+            config,
+        }: OpenOpts<S, FAMILIES>,
```
bgw, line 316: Getting fancy here with the destructured arguments

bgw (approved):
Thanks for cleaning this all up!

PR #93994, opened 2026-05-20
fix(instrumentation): await ensureInstrumentationRegistered in RouteModule.prepare

### LINKED ISSUE #93993: RouteModule.prepare doesn't await ensureInstrumentationRegistered

https://github.com/mgm1313/next-instrumentation-race-repro

### To Reproduce

Deploy to Vercel, then hit `/api/check` on a cold lambda with a few concurrent requests:

```sh
URL=https://<your-deployment>.vercel.app/api/check
seq 5 | xargs -P 5 -n 1 -I {} curl -s "$URL"
```

`instrumentation.ts#register()` sets `globalThis.__registerStarted`, awaits 200ms (to stand in for the top-level `await import(...)` chain that real OTel / app-bootstrap hooks do), then sets `globalThis.__registerFinished`. The route reports both timestamps. On a cold lambda, responses come back with `"registerFinished": null` and `"registerFinishedBeforeRequest": false` — the handler ran while `register()` was still awaiting. Won't reproduce with `next start` locally because that path goes through `BaseServer.handleRequest`, which awaits `prepare()` correctly.

**Current:** On a cold Vercel lambda, an App Router route handler can begin executing before `instrumentation.ts#register()` resolves. Any module-scoped state that `register()` is supposed to populate (OTel patches, config globals, DB pools, etc.) reads as `null`/`undefined` on the first request. A subsequent request to the now-warm lambda sees the populated state.

In the reproduction, `register()` sets `globalThis.__registerStarted`, awaits 200ms, then sets `globalThis.__registerFinished`. On a cold start, `/api/check` returns:

```json
{
  "requestEntered": 1716330000123,
  "registerStarted": 1716330000050,
  "registerFinished": null,
  "registerFinishedBeforeRequest": false
}
```

**Expected:** Per the [instrumentation docs](https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation) ("code in the register function will be executed before any other code"), the route handler should not run until `register()` has resolved. Every response should have `"registerFinished"` populated and `"registerFinishedBeforeRequest": true`. This is what happens via `next start` locally (`BaseServer.handleRequest` awaits `prepare`) and via the edge adapter — only the Vercel app-route path is racy.

```bash
Operating System: Linux (Vercel serverless, Node 22)
Binaries:
  Node: 22.x
  pnpm: 10.24.0
Relevant Packages:
  next: 16.2.6 (also reproduced against 16.3.0-canary.24 source)
  react: 19.x
Next.js Config:
  default Vercel build, app router
```

Instrumentation, Runtime, Route Handlers

Vercel (Deployed)

### Additional context

We started seeing intermittent `TypeError`s in production where module-scoped state our `register()` is supposed to populate was still `null` when an App Router POST handler ran. Six errors over an hour, all clustered with bursts of concurrent webhook deliveries to a fresh lambda. Retries always succeeded.

Tracing it down: in `packages/next/src/server/route-modules/route-module.ts`, `RouteModule.prepare` calls `ensureInstrumentationRegistered` without awaiting it (canary line 674, 16.2.6 line 654):

```ts
const { ensureInstrumentationRegistered } = await import(
  '../lib/router-utils/instrumentation-globals.external.js'
)
// ensure instrumentation is registered and pass
// onRequestError below
ensureInstrumentationRegistered(absoluteProjectDir, this.distDir)
```

The two other call sites for the same function do await it:

| Path | File | Awaited? |
|---|---|---|
| Classic Node server | `packages/next/src/server/next-server.ts:373` (via `BaseServer.handleRequest` → `prepareImpl`) | ✅ |
| Edge adapter | `packages/next/src/server/web/adapter.ts:114` | ✅ |
| App Router per-request `RouteModule.prepare` | `packages/next/src/server/route-modules/route-module.ts:674` | ❌ |

The compiled app-route template (`build/templates/app-route.ts`) awaits `routeModule.prepare(...)`, but `prepare` returns before the instrumentation hook finishes. On a fresh Vercel lambda, the userland POST body runs while `register()` is still working through its top-level `await import(...)` chain. `ensureInstrumentationRegistered` is memoized, so it does eventually finish — just not before the first request gets dispatched.

The docs say "code in the register function will be executed before any other code", which matches the two awaited paths but not this one.

## Suggested fix

Just add the missing `await`:

```diff
- ensureInstrumentationRegistered(absoluteProjectDir, this.distDir)
+ await ensureInstrumentationRegistered(absoluteProjectDir, this.distDir)
```

The function memoizes a single promise per process, so the await is a microtask no-op on every request after the first. The cold-start request would block until `register()` is actually done, which is the behavior the docs describe.

## What we're doing in the meantime

We exposed a "contexts ready" promise from our `register()` chain and `await` it at the top of each affected route handler. Works, but it's opt-in per route and easy to forget.

mgm1313:
The per-request call to ensureInstrumentationRegistered() inside RouteModule.prepare was missing an await, so when an adapter dispatches the compiled app-route template's handler without going through BaseServer.handleRequest (where prepareImpl awaits it via runInstrumentationHookIfAvailable), a cold-start request could begin executing the userland route handler before instrumentation.ts#register() resolved. The other two call sites (next-server.ts and web/adapter.ts) already await; this aligns the third.

The function memoizes a single promise per process, so the added await is a microtask no-op on every request after the first cold start.

Adds a with-async-node-app-route mirror of the existing with-async-node-page case to lock in the invariant for app-route handlers.

### What?

`RouteModule.prepare` calls `ensureInstrumentationRegistered(...)` from `lib/router-utils/instrumentation-globals.external` without `await`ing it. This PR adds the missing `await` so that adapters which dispatch the compiled `app-route.ts` template's handler directly (bypassing `BaseServer.handleRequest`) still observe `instrumentation.ts#register()` resolving before the userland route handler runs.

### Why?

The `ensureInstrumentationRegistered` helper is awaited from the other two call sites:

| Path | File | Awaited |
|---|---|---|
| Classic Node server | `packages/next/src/server/next-server.ts:373` (via `BaseServer.handleRequest` → `prepareImpl`) | ✅ |
| Edge adapter | `packages/next/src/server/web/adapter.ts:114` | ✅ |
| App Router per-request | `packages/next/src/server/route-modules/route-module.ts:674` | ❌ (before this PR) |

When an adapter invokes the compiled app-route template's `handler` directly (e.g. on a freshly cold-started Vercel serverless function under concurrent burst load) the `RouteModule.prepare` path is the only chance to register. The unawaited call lets the userland POST handler begin executing before `register()` finishes its top-level `await import(...)` chain. Module-scoped state that `register()` is supposed to populate (OTel patches, app-bootstrap globals, DB pools) reads as `null`/`undefined` on the first request; retries on the now-warm lambda succeed.

The function memoizes a single promise per process, so the added `await` is a microtask no-op on every request after the first cold start.

### How?

One-line `await` added at the call site. Adds `with-async-node-app-route`, mirroring the existing `with-async-node-page` regression case: `register()` sleeps 1000ms and sets `globalThis.instrumentationFinished`; the app-route handler reads it and asserts `{ finished: true }`.

**Note on test coverage**: under `next start` the request still enters via `BaseServer.handleRequest`, which already awaits `prepare()`, so the new test passes both before and after this change. It documents the intended invariant for the App Router route-handler path and locks it in against future regressions. A test that fails-without-fix would require exercising an adapter-level dispatch path (e.g. Vercel's `@vercel/next`) that bypasses `BaseServer`, which isn't covered by the existing test infrastructure.

Repro: https://github.com/mgm1313/next-instrumentation-race-repro

Fixes #93993


test/e2e/instrumentation-hook/with-async-node-app-route/instrumentation.node.js
```diff
@@ -0,0 +1,8 @@
+/*global globalThis*/
+
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
+
+export async function register() {
+  await sleep(1000)
```
timneutkens, line 6: I checked and the fix looks correct to me. It should be awaited there.

For the test I'm a bit concerned that this can introduce flakes or worst case is not actually failing because of instrumentation running earlier (e.g. during server bootup).

I'm pretty sure it will correctly fail when deployed though, so it's still good to have coverage.

Can't think of a better way to test this though. Potentially it shouldn't be 1 second but 5 seconds so that it's less likely that it's a false positive/negative.

mgm1313:
Thanks, @timneutkens. I looked into this a bit more and decided not to bump the sleep to 5s for now.

The existing `next start` request-based test can still pass without this fix, because that path already goes through `BaseServer.prepareImpl()`, which awaits instrumentation earlier. So increasing the timeout would make the test slower, but it still would not prove this specific `RouteModule.prepare()` await is needed I believe.

Instead I added a `next start`-only regression test that imports the built app-route module from `.next/server/app/api/check/route.js` and invokes its exported `handler` directly. That bypasses the `BaseServer` preparation path and exercises the route-module path this change fixes. Locally, this new test fails without the `await` (`finished: false`) and passes with it.

I used the direct compiled-handler pattern from [`cache-components.test.ts`](https://github.com/vercel/next.js/blob/canary/test/e2e/app-dir/cache-components/cache-components.test.ts#L356-L430) as precedent.

test/e2e/instrumentation-hook/instrumentation-hook.test.ts
```diff
@@ -74,6 +75,99 @@ describe('Instrumentation Hook', () => {
...
+  describeCase('with-async-node-app-route', ({ next, isNextStart }) => {
+    it('with-async-node-app-route should run the instrumentation hook before the app-route handler', async () => {
+      const res = await next.fetch('/api/check')
+      const body = await res.json()
+      expect(body).toEqual({ finished: true })
+    })
+
+    if (isNextStart) {
+      it('with-async-node-app-route should wait for instrumentation when invoking the app-route handler directly', async () => {
+        const routeModulePath = path.join(
```
timneutkens: The tests should use isolation and the setup you had before is actually more in line with what I'd want to merge, increasing the timeout would be fine for that case.

We don't want to introduce this type of direct requiring of the built path, that also makes it impossible to run the test when deployed through adapters.

timneutkens: To clarify: The test you added will actually reproduce the failure during the deployment tests. So it's important to make it run during deployment tests too.

timneutkens (commented):
I'll take over getting this PR merged 🙏

mgm1313:
Thanks for taking this over, @timneutkens 🙏

That makes sense. I went down the direct-handler route mostly because I was struggling to make the local `next start` test fail without the fix: that path seems to go through `BaseServer.prepareImpl()`, which already awaits instrumentation earlier, so it kept passing before and after the change. The direct import was just my attempt to isolate the `RouteModule.prepare()` path and prove the missing `await` mattered locally.

Appreciate you taking it from here.

test/e2e/instrumentation-hook/with-async-node-app-route/app/api/check/route.ts
```diff
@@ -0,0 +1,7 @@
+export const dynamic = 'force-dynamic'
```
mischnic: This breaks with the cache components test variant

test/e2e/instrumentation-hook/instrumentation-hook.test.ts
timneutkens: :robot: Done -- removed the direct built-handler test and enabled the request-based app-route assertion for deployment mode, so adapters exercise the regression without importing generated output directly.

test/e2e/instrumentation-hook/with-async-node-app-route/app/api/check/route.ts
timneutkens: :robot: Done -- replaced the incompatible `dynamic = 'force-dynamic'` segment config with `await cookies()`, which keeps the route dynamic under cache components. I verified the focused app-route test in cache-components start mode and node-streams cache-components dev mode.

test/e2e/instrumentation-hook/with-async-node-app-route/instrumentation.node.js
timneutkens: :robot: Handled by running the app-route request coverage in deployment mode, where the adapter path reaches the affected route-module preparation. I kept the delay at 1s to avoid adding several seconds across each mode variant now that the relevant path is covered.

mischnic:
Unfortunately, our tests currently need some secrets to run deploy tests, and those are not available on external PRs, so CI will always fail here.
I've recreated this PR with the commits here: https://github.com/vercel/next.js/pull/94306

PR #93998, opened 2026-05-20
Respond to App Shell prefetches on the server

acdlite:
Introduces a new prefetch request type — the App Shell — and the server-side handler that renders it. With this PR, a client (or a test harness) can issue a request with `NEXT_ROUTER_PREFETCH_HEADER: '3'` and receive the param-independent shell of a route: param- and searchParam-bearing segments suspend, leaving everything that doesn't depend on them intact.

The App Shell render is a subtype of the existing runtime prerender, distinguished by a single bit on the prerender store. When that bit is set, `await params` and `await searchParams` both hang forever, so any segment that depends on either suspends to its nearest fallback. The "shell" of the route is whatever renders above those suspense boundaries — layouts, cookie-derived UI, and other request-context-derived content.

Static shell prefetches are not covered by these changes; those will be addressed by a future PR and use a slightly different strategy. (Instead of creating a new response type, the existing static prefetch response will be updated to support extraction of the shell via "rewinding", similar to what we do for cached navigations.)

No client-side changes yet, only the server parts.

### DIFF
--- a/packages/next/src/server/request/params.ts
+++ b/packages/next/src/server/request/params.ts
@@ -177,6 +177,7 @@ export function createServerParamsForRoute(
         return createRuntimePrerenderParams(
           underlyingParams,
           null,
+          workStore,
           workUnitStore,
           varyParamsAccumulator,
           isRuntimePrefetchable
@@ -249,6 +250,7 @@ export function createServerParamsForServerSegment(
         return createRuntimePrerenderParams(
           underlyingParams,
           optionalCatchAllParamName,
+          workStore,
           workUnitStore,
           varyParamsAccumulator,
           isRuntimePrefetchable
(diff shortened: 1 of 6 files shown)


packages/next/src/server/request/params.ts
```diff
@@ -427,6 +438,17 @@ function createRuntimePrerenderParams(
   varyParamsAccumulator: VaryParamsAccumulator | null,
   isRuntimePrefetchable: boolean
 ): Promise<Params> {
+  if (workUnitStore.forceOmitParams) {
+    // App Shell prefetch: any `await params` suspends. Segments that depend
+    // on params render as holes, leaving the param-independent shell.
+    const workStore = workAsyncStorage.getStore()
+    return makeHangingPromise<Params>(
+      workUnitStore.renderSignal,
+      workStore?.route ?? '',
```
unstubbable: The `workStore` is well defined on both call sites of this function. I suggest we pass it in as an argument to avoid this dead and nonsensical fallback.

packages/next/src/server/request/search-params.ts
```diff
@@ -245,6 +245,19 @@ function createRuntimePrerenderSearchParams(
...
 ): Promise<SearchParams> {
+  if (workUnitStore.forceOmitParams) {
+    // App Shell prefetch: any `await searchParams` suspends. Segments that
+    // depend on search params render as holes, leaving the
+    // search-param-independent shell. Matches the behavior in
+    // createRuntimePrerenderParams for path params.
+    const workStore = workAsyncStorage.getStore()
+    return makeHangingPromise<SearchParams>(
+      workUnitStore.renderSignal,
+      workStore?.route ?? '',
```
unstubbable: Same here (only one call site).

unstubbable (approved):

lubieowoce (approved):

unstubbable (approved):

PR #93999, opened 2026-05-20
Prefetch App Shells on the client

acdlite:
Gated behind `experimental.appShells`. Clicking a link to a route the user has never specifically prefetched now renders the route's App Shell instantly; the per-link concrete request continues in the background and streams in the param-specific content.

### Motivation

Today, prefetching a parameterized route like `/chat/[id]` requires a concrete `id`. The framework caches a separate prefetch entry per link, so the cost of prefetching scales with the number of visible links — not with the number of routes. On a page with high link cardinality (a feed, a search-results page, a chat list) it is impractical to prefetch every concrete URL up front. If a per-link prefetch for `/chat/123` is still in flight when the user clicks the link, the navigation blocks until that prefetch completes; there's no cached generic shell to fall back to. The cost of a prefetch miss is a full blocking navigation.

The property we want: once a user has visited a Next.js app, every subsequent navigation should transition to _something_ instantly — at minimum an App Shell — regardless of whether per-link prefetches for the concrete destination have completed. This matters most under adverse conditions (slow networks, offline, high-cardinality routes), but the guarantee is unconditional.

App Shells make that property hold. A shell is a per-_route_ resource, not a per-link one — the number of shells in flight at any time scales with filesystem routes, which is bounded and small. Aggressive App Shell prefetching is affordable in a way that aggressive per-link runtime prefetching is not.

### Mechanism

A new `Shell` phase sits in the prefetch scheduler between the existing `RouteTree` and `Speculative` (formerly `Segments`) phases. Shell-phase tasks issue an App Shell request and write the response under a param-independent vary path. Concurrent shell tasks for sibling links to the same route dedupe at this keypath, so the cache holds at most one shell entry per route.

The headline property: if N links on a page resolve to the same route under different params, they share _one_ App Shell request collectively. Once it lands, every one of those navigations can render an instant shell, regardless of whether the param-specific concrete prefetch has completed.

The Speculative phase is mostly unchanged — it still issues the per-link concrete prefetches that fill in param-specific content over time. Routes that are fully static (no runtime data anywhere in their tree) skip the Shell phase entirely, since their existing static prefetches are already shell-like in shape.

### Navigation-time cache lookup

A small change in how the cache is read at navigation time is what makes the instant-shell guarantee actually hold. The cache normally returns the _most-specific_ matching entry — the right semantics for prefetch dedup, but the wrong semantics for navigation. If a fulfilled shell entry coexists with an in-flight Pending entry for a more-specific keypath that the navigation also matches, the most-specific entry is the empty Pending one, and the navigation would block on it instead of rendering the shell.

Navigation now does a two-pass lookup: first prefer Fulfilled entries anywhere along the vary path (so a less-specific shell beats a more-specific Pending), then fall back to the regular behavior if nothing fulfilled is found. Prefetch reads keep the original semantics, since they need to see in-flight entries to dedupe.

### Scope

The shape of the prefetch request and the response interpretation differ between static (`PPR`) and runtime (`PPRRuntime`) prefetches. Only the runtime path is covered here. The static path uses a different strategy — rewinding a single response into a shell prefix and a concrete suffix, rather than issuing a separate shell-only request — and will be added in a future PR alongside the server-side byte-offset machinery it depends on.

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -78,9 +78,11 @@ import {
   setSizeInCacheMap,
   deleteFromCacheMap,
   isValueExpired,
+  EntryStatus,
   type CacheMap,
   type UnknownMapEntry,
 } from './cache-map'
+export { EntryStatus } from './cache-map'
 import {
   appendSegmentRequestKeyPart,
   convertSegmentPathToStaticExportFilename,
@@ -181,18 +183,6 @@ type RouteCacheEntryShared = {
   size: number
   staleAt: number
   version: number
-}
-
-/**
- * Tracks the status of a cache entry as it progresses from no data (Empty),
- * waiting for server data (Pending), and finished (either Fulfilled or
- * Rejected depending on the response from the server.
- */
-export const enum EntryStatus {
-  Empty = 0,
-  Pending = 1,
-  Fulfilled = 2,
-  Rejected = 3,
 }
 
 export type PendingRouteCacheEntry = RouteCacheEntryShared & {
@@ -458,7 +448,8 @@ export function readRouteCacheEntry(
     getCurrentRouteCacheVersion(),
     routeCacheMap,
     varyPath,
-    isRevalidation
+    isRevalidation,
+    false
   )
   if (existingEntry !== null) {
     return existingEntry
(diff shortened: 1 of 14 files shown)


packages/next/src/client/components/segment-cache/cache.ts
```diff
@@ -2221,11 +2259,15 @@ export async function fetchSegmentPrefetchesUsingDynamicRequest(
...
+    // marks the response as '~' (Partial). RuntimeShell additionally omits
+    // every dynamic suspense boundary below the App Shell, so its segments
+    // are always partial regardless of what the server marker says.
+    // Full/LoadingBoundary prefetches are always complete.
     const isResponsePartial =
-      fetchStrategy === FetchStrategy.PPRRuntime &&
-      (cacheData?.isResponsePartial ?? false)
+      fetchStrategy === FetchStrategy.RuntimeShell ||
+      (fetchStrategy === FetchStrategy.PPRRuntime &&
+        (cacheData?.isResponsePartial ?? false))
```
unstubbable, line 2270: > RuntimeShell additionally omits every dynamic suspense boundary below the App Shell, so its segments are always partial regardless of what the server marker says.

Hm, why wouldn't this be reflected in the marker byte?

packages/next/src/client/components/segment-cache/vary-path.ts
```diff
@@ -362,3 +369,34 @@ export function getFulfilledSegmentVaryPath(
...
+  // Re-keys a segment's vary path to identify the "App Shell" entry for this
+  // segment position — a reusable, param-free loading state that can be served
+  // for any concrete navigation to this segment. Every param node (path
+  // params, search params) is replaced with Fallback; only structural nodes
+  // (request keys, etc.) keep their concrete value.
+  //
+  // NOTE: In the future we may allow root params to be accessed in the shell,
+  // since they change less frequently, and if they do change during a
+  // navigation, we perform an MPA navigation. So they don't affect SPA
+  // navigations. However, because in the future we plan to use the App Shell
```
unstubbable: This comment is stale, as we now realized we don't perform an MPA nav when root params change.

test/e2e/app-dir/segment-cache/prefetch-app-shell/app/posts/[id]/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <ParamsDependent params={params} />
+      </Suspense>
+    </main>
+  )
+}
+
+async function CookieDependent() {
+  const cookieStore = await cookies()
+  const value = cookieStore.get('testCookie')?.value ?? 'none'
+  return <p id="cookie-value">{`Cookie: ${value}`}</p>
```
Comment (unstubbable) [inline test/e2e/app-dir/segment-cache/prefetch-app-shell/app/posts/[id]/page.tsx:30]: Did you intend to assert that this content is included in the App Shell?

unstubbable (approved):

test/e2e/app-dir/segment-cache/prefetch-app-shell/app/posts/[id]/page.tsx
Comment (acdlite) [reply test/e2e/app-dir/segment-cache/prefetch-app-shell/app/posts/[id]/page.tsx:30]: yeah we might as well

packages/next/src/client/components/segment-cache/cache.ts
acdlite: Once Janka's PR lands, the marker byte represents the whole response, not any particular stage. But the server also only emits a stage offset if it's actually distinct from the subsequent stage. So this is just explaining that by the time we decide to extract a stage, we've already determined that the stage is partial.

PR #94005, opened 2026-05-21
[ci] Use Turbopack in non-webpack-specific e2e test jobs

bgw:
Turbopack is the default bundler, so we should prioritize testing that in these variants of the CI runs (cache components, node streams, windows).

There's also some limited evidence that indicates that it might be faster for e2e test runs, though that's not conclusive (because of different shard sizes): https://vercel.slack.com/archives/C0B0LKK5QD7/p1779320615501259

### DIFF
--- a/test/e2e/app-dir/non-root-project-monorepo/non-root-project-monorepo.test.ts
+++ b/test/e2e/app-dir/non-root-project-monorepo/non-root-project-monorepo.test.ts
@@ -48,17 +48,20 @@ describe('non-root-project-monorepo', () => {
     })
   })
 
+  // TODO: These file URIs are wrong on turbopack+windows:
+  // https://en.wikipedia.org/wiki/File_URI_scheme
+  // The URIs should always use forward slashes, but use backslashes on windows.
   describe('import.meta.url', () => {
     it('should work during RSC', async () => {
       const $ = await next.render$('/import-meta-url-rsc')
-      expect($('p').text()).toMatch(
+      expect($('p').text().replaceAll('\\', '/')).toMatch(
         /^file:\/\/.*\/next-install-[^/]+\/apps\/web\/app\/import-meta-url-rsc\/page.tsx$/
       )
     })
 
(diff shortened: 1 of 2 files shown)


bgw:
* **#94005** <a href="https://app.graphite.com/github/pr/vercel/next.js/94005?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94005?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

test/e2e/app-dir/non-root-project-monorepo/non-root-project-monorepo.test.ts
```diff
@@ -48,17 +48,20 @@ describe('non-root-project-monorepo', () => {
     })
   })
 
+  // TODO: These file URIs are wrong on turbopack+windows:
```
lukesandberg, line 51: 😱  

im guessing this is the cause

https://github.com/vercel/next.js/blob/6dbba05eeb747aa14c263d5d998b3cdf7fa17420/turbopack/crates/turbopack-core/src/source_map/utils.rs#L207

we pull filenames out of source maps and then compose uris with them without normalizing dir-separators

sokra (approved):

test/e2e/app-dir/non-root-project-monorepo/non-root-project-monorepo.test.ts
sokra: I let an agent fix it as follow-up

bgw: Discussed IRL: There are deeper problems where we also have code that strips `file://` prefixes and expects a system path, so this isn't an easy fix: https://github.com/vercel/next.js/pull/84255#discussion_r2388504301

PR #94016, opened 2026-05-21
Make `--debug-build-paths` work with `src/` directory projects

unstubbable:
The `--debug-build-paths` option silently dropped routes in projects that keep their `app` or `pages` directory under `src/`. When a user passed `src/app/foo/page.tsx`, the glob matched the file but `categorizeAndAddPath` only recognized paths starting with `app/` or `pages/`, so nothing was added to the resolved app paths and the build produced no user routes. Conversely, when a user passed the bare `app/foo/page.tsx` form against a `src/app` project, glob found nothing and emitted `did not match any files`.

`resolveBuildPaths` now detects whether `src/app` or `src/pages` exists and prepends `src/` to bare patterns before globbing, and `categorizeAndAddPath` strips a leading `src/` from matched files so route keys are normalized either way. Added a `src-dir` fixture and four tests covering explicit `src/` prefixes, bare patterns with auto-prefixing, and glob patterns through route groups.

### DIFF
--- a/packages/next/src/lib/resolve-build-paths.ts
+++ b/packages/next/src/lib/resolve-build-paths.ts
@@ -40,6 +40,12 @@ export async function resolveBuildPaths(
   const appPaths: Set<string> = new Set()
   const pagePaths: Set<string> = new Set()
 
+  // Detect whether the project keeps its routes under `src/` so we can accept
+  // patterns written with or without that prefix (e.g. both `app/foo/page.tsx`
+  // and `src/app/foo/page.tsx`).
+  const useSrcApp = fs.existsSync(path.join(projectDir, 'src', 'app'))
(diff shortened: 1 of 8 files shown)


packages/next/src/lib/resolve-build-paths.ts
```diff
@@ -40,6 +40,12 @@ export async function resolveBuildPaths(
   const appPaths: Set<string> = new Set()
   const pagePaths: Set<string> = new Set()
 
+  // Detect whether the project keeps its routes under `src/` so we can accept
+  // patterns written with or without that prefix (e.g. both `app/foo/page.tsx`
+  // and `src/app/foo/page.tsx`).
```
mischnic, line 45: I think we should just say that `--debug-build-paths` accepts file paths to the page file. If you forget `src`, then that's on the user

mischnic (approved):

PR #94020, opened 2026-05-21
docs: add Interactive Apps guide

aurorascharff:
### What?

Adds the Interactive Apps guide (`docs/01-app/02-guides/interactive-apps.mdx`) with a companion [Taskboard app](https://github.com/vercel-labs/async-react-demo).

### Why?

There is no end-to-end guide that shows how to coordinate pending work across Server Functions, transitions, optimistic UI, and caching in a single app.

### How?

The guide walks through building a task management board in 8 steps, from a plain Server Component app to one with optimistic UI, streaming, caching, and prefetching:

1. Stream slow data with Suspense — promise-passing to split reads into independent boundaries
2. Respond instantly to a toggle — `useOptimistic` + `startTransition` for a priority cycle
3. Filter with pending feedback — `useTransition` + `data-pending` for label chips, action props for `ChipGroup`
4. Add comments with instant feedback — `useOptimistic([])` for a pending-only comment list
5. Move cards between columns — optimistic reducer for drag-and-drop status updates
6. Manage form lifecycle with useActionState — `isPending`, key-based reset, double-transition for dialog close
7. Signal pending deletion to a parent — `useOptimistic(false)` + `data-pending` CSS attribute
8. Make repeat navigation instant — `use cache`, `cacheTag`, `updateTag`, `unstable_prefetch`

The companion app lives at [vercel-labs/async-react-demo](https://github.com/vercel-labs/async-react-demo). The `plain` branch is the starting point, `main` has all 8 steps applied.


docs/01-app/02-guides/interactive-apps.mdx
```diff
@@ -0,0 +1,794 @@
...
+  description: Learn more about the APIs used in this guide.
+  links:
+    - app/api-reference/functions/refresh
+    - app/api-reference/directives/use-cache
+    - app/api-reference/functions/cacheTag
+    - app/api-reference/functions/updateTag
+    - app/guides/streaming
+---
+
+Server Components render the current server data. Client Components start interactions before the next server render is ready. Without additional state, the UI keeps showing the old server props while a Server Function, form action, or navigation is still running.
```
icyJoseph: Wonder if this intro fits a bit better.

```suggestion
When a user interaction requires server-side work, the result is not available immediately. Network requests take an unknown amount of time to complete and may succeed or fail. While that work is in progress, the client should provide feedback and update the UI accordingly. Without it, users are left looking at stale data and may wonder whether their action had any effect.
```

icyJoseph: Or maybe this 🤔 - up to you hehe

```suggestion
When a user interaction requires server-side work, the result is not available immediately. Network requests take an unknown amount of time to complete and may succeed or fail. While waiting, the client should provide feedback and respond to those states. Otherwise, users are left looking at stale data and may wonder whether anything is happening.
```

docs/01-app/02-guides/interactive-apps.mdx
```diff
@@ -0,0 +1,794 @@
...
+    <button onClick={() => cyclePriority(id)} className={priorityDot[priority]}>
+      {priority}
+    </button>
+  )
+}
+```
+
+If you click the dot, the Server Function runs, but the button keeps rendering the old `priority` prop until the next server render arrives.
+
+Replace the direct call with `useOptimistic` and `startTransition`. `useOptimistic` renders a temporary value while the Server Function is pending. `startTransition` scopes the async work so React knows when to keep or discard the optimistic value:
```
icyJoseph: A little nit, trying to replace `scopes the async work` 

```suggestion
Replace the direct call with `useOptimistic` and `startTransition`. `useOptimistic` provides a value to render in place of the stale prop. `startTransition` runs Server Function as part of a transition. The optimistic value applies for as long as that transition is pending.
```

Something like that, could be a little easier to read, without having to learn what startTransition does

docs/01-app/02-guides/interactive-apps.mdx
```diff
@@ -0,0 +1,794 @@
...
+    >
+      {optimisticPriority}
+    </button>
+  )
+}
+```
+
+The `setOptimisticPriority` call updates the UI on the current frame with the next value in the cycle. Reading from `optimisticPriority` (not the prop) means rapid double-clicks cycle correctly instead of reading a stale closure value.
+
+`startTransition` wraps the async work so React tracks it as a pending transition. If the Server Function throws, the optimistic value reverts. The standalone `startTransition` import does not bubble errors to error boundaries. If you need errors to reach an error boundary, use `useTransition` instead (shown in Step 3).
```
icyJoseph: Paired with the comment above, because the first sentence here felt repetitive-ish

```suggestion
If the Server Function inside `startTransition` throws, the transition ends and the optimistic value reverts. The standalone `startTransition` import does not bubble errors to error boundaries. If you need errors to reach an error boundary, use `useTransition` instead (shown in Step 3).
```

docs/01-app/02-guides/interactive-apps.mdx
```diff
@@ -0,0 +1,794 @@
...
+    </div>
+  )
+}
+```
+
+The `data-pending` attribute on the wrapper lets parent components react via CSS. The board on the home page uses `group-has-data-pending:opacity-50` to dim while the transition runs, without replacing the board with a skeleton.
+
+If you click a chip now, the board dims while the filtered data loads. The `data-pending` attribute stays on the wrapper until the navigation completes, then the board updates with the filtered tasks.
+
+But the chip itself still doesn't highlight until after the transition. The `isPending` boolean tells us the navigation is in flight, but it doesn't help the chip show a selected state before the new data arrives.
```
icyJoseph:
```suggestion
But the chip itself still doesn't highlight until the navigation completes. The `isPending` boolean tells us the navigation is in flight, but it doesn't help the chip show a selected state before the new data arrives.
````

File: docs/01-app/02-guides/interactive-apps.mdx [hunk 5]
```diff
@@ -0,0 +1,798 @@
...
+| ------------------------------------- | ---------------------------- | -------------------------------------------------------- |
+| `<Suspense>`                          | None                         | Any uncached read that can stream                        |
+| `'use cache'` + `cacheTag`            | Server memory                | Reads that return the same value for the next viewer     |
+| `unstable_prefetch = 'force-runtime'` | Server work per visible link | Destinations where the read is keyed on the real request |
+
+## Next steps
+
+Each pattern adds state for a different kind of pending work:
+
+| Pattern            | Use it when                                                   |
```
icyJoseph: I think we could switch to a situation -> API look up table?

```suggestion
| Situation | Use |
| --- | --- |
| One value should update before the server responds | `useOptimistic` |
| A URL change triggers a new server render | `useTransition` |
| Slow data should stream in without blocking the page shell | `<Suspense>` |
| A reusable component owns part of the pending UI | Action prop |
| A new item should appear before it exists on the server | `useOptimistic` (pending-only list) |
| A collection needs a temporary structural change | `useOptimistic` (reducer form) |
| A form needs pending state, reset behavior, and result state | `useActionState` |
| A parent should style itself while a child action is pending | `data-pending` attribute |
| A reusable read should survive across renders and navigations | `'use cache'` + `cacheTag` / `updateTag` |
| A cached destination should be filled before the user clicks | Runtime prefetch (`unstable_prefetch = 'force-runtime'`) |
```

docs/01-app/02-guides/interactive-apps.mdx
```diff
@@ -0,0 +1,759 @@
...
+- [`useActionState`](https://react.dev/reference/react/useActionState) wraps a form action with pending state, reset behavior, and result handling.
+- [`'use cache'`](/docs/app/api-reference/directives/use-cache) with [`cacheTag`](/docs/app/api-reference/functions/cacheTag) and [`updateTag`](/docs/app/api-reference/functions/updateTag) keeps reusable reads warm across navigations, and [runtime prefetching](/docs/app/guides/prefetching) fills the next destination before the click.
+
+Most patterns mix two or more of these. Reach for whichever primitive fits the constraint you're solving.
+
+Related guides:
+
+- [Streaming](/docs/app/guides/streaming) for loading boundaries and `<Suspense>` patterns
+- [Instant Navigation](/docs/app/guides/instant-navigation) for validating that navigations stay instant
+- [View Transitions](/docs/app/guides/view-transitions) for animating state changes
```
icyJoseph, line 779: Should we add a callout/x-ref about INP? the contents of this guide are the kind of thing that can also help people with their INP metrics, although often this can happen from heavy client modules, or blocking roundtrips - could be good to do a quick research and add a note about it

aurorascharff: Updated, check last two commits!

icyJoseph:
Looks good to me, there's a few things I want to try out myself after reading the guide, often a good sign!

I am ready to approve the guide, but in terms of discoverability, no other document points to this guide, https://nextjs.org/docs/app/getting-started/mutating-data should point as Next Steps, or in prose somewhere relevant

icyJoseph (approved):

PR #93988, opened 2026-05-20
Prerender HTTP access fallbacks with Cache Components semantics

unstubbable:
The body of `if (cacheComponents)` in `prerenderToStream` is extracted into a local helper, `prerenderWithCacheComponents(getPayload)`. The catch-block error paths now route through that helper whenever Cache Components is enabled, which means the legacy `prerender-legacy` store is no longer reachable from a Cache Components prerender.

As a result, `notFound()`, `forbidden()`, and `unauthorized()` recovery renders the matching fallback boundary under `prerender` and `prerender-client` semantics. Dynamic API access in those boundaries — for example `useSearchParams()` without a surrounding `<Suspense>` — now surfaces as a blocking-route error instead of the legacy `BailoutToCSRError`.

The dev side will be handled in a follow-up. The prerender handling diverged a bit from the dev rendering in #92231, and we'll likely need to re-align the dev rendering first.

> [!TIP]
> Best reviewed with hidden whitespace changes.

PR #92231, opened 2026-04-01
fix: preserve HTTP access fallbacks during prerender recovery

ztanner:
When a `notFound()`, `forbidden()`, or `unauthorized()` error escapes into the outer prerender recovery path, we were falling back to the generic error shell flow.

In the `cacheComponents` case, that could leave us with:
- error HTML rendered from `ErrorApp`
- Flight data reused from the aborted prerender prelude
- references to Flight chunks that were never emitted

That is what caused the client-side `Connection closed` failure in #86251.

Instead of rerendering the full Flight tree or always using the generic error RSC payload, this change:

- finds the deepest matching HTTP fallback boundary
- rerenders the normal app router payload with that segment-scoped fallback
- tees the replacement Flight stream so Fizz can render from one copy while prerender buffering consumes the other
- only takes this path for recoverable HTTP access fallbacks that have a real boundary (ie a defined not-found/unauthorized/etc)

If no matching boundary exists, we keep the existing generic error handling.

Fixes #86251
Fixes #90837
Closes #87041
Closes #86251

Closes NAR-711
Closes NEXT-4876

PR #94037, opened 2026-05-22
Improve notFound recovery in error pathway

gnoff:
Effectively

Relands #93988
Reverts #92231

Prerendering the notFound loaderTree turns out to not be the correct behavior in many cases. The original motivation of 92231 as to solve the connection closed problem caused by statically prerendering an incomplete HTTP event like notFound() when cacheComponents is enabled because it would leave holes in the inlined flight data that never resolved and led to reported errors on hydration.

This change now continues to fork the CC behavior of the error handler but it treats the error page as a resumable page so fresh hydration data is generated if it wasn't able to be entirely static.

This is good for correctness but it also means the page is not going to statically prerender the notFound UI. This will be handled in a followup

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -7339,107 +7339,228 @@ async function prerenderToStream(
 
   let prerenderStore: PrerenderStore | null = null
 
-  try {
-    if (cacheComponents) {
-      /**
-       * cacheComponents with PPR
-       *
-       * The general approach is to render the RSC stream first allowing any cache reads to resolve.
-       * Once we have settled all cache reads we restart the render and abort after a single Task.
-       *
-       * Unlike with the non PPR case we can't synchronously abort the render when a dynamic API is used
-       * during the initial render because we need to ensure all caches can be filled as part of the initial Task
-       * and a synchronous abort might prevent us from filling all caches.
-       *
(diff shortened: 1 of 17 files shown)


packages/next/src/server/app-render/app-render.tsx
```diff
@@ -8527,7 +8508,352 @@ async function prerenderToStream(
...
+            : INFINITE_CACHE,
+        stale:
+          typeof prerenderStore?.stale !== 'undefined'
+            ? prerenderStore.stale
+            : INFINITE_CACHE,
+        tags: [...(prerenderStore?.tags || implicitTags.tags)],
+        prerenderResumeDataCache: errorPrerenderResumeDataCache,
+        renderResumeDataCache: renderOpts.renderResumeDataCache ?? null,
+        hmrRefreshHash: undefined,
+        varyParamsAccumulator: null,
```
unstubbable, line 8581: Is it correct not to accumulate vary params for this prerender?

unstubbable (approved):
In the follow-up we might want to move the not-failing-anymore test cases out of `test/e2e/app-dir/cache-components-errors`, and also add behavioral assertions on top of `expectStaticRouteArtifacts`.

packages/next/src/server/app-render/app-render.tsx
gnoff: yeah. we need to use the main render's params since it covers the RSC segment data which we can still extract.

PR #94044, opened 2026-05-22
[CC] Streaming prerender

lubieowoce:
We want to be able to recover a shell from static and runtime prerenders. This means that prerenders need to become staged renders, so that we can collect the outputs as we progress through the stages. This is a first step before #93801, where we'll actually implement the staging

We do this with a regular `renderToReadableStream` so we can get the results streamingly. We run the render, abort it, and omit any chunks emitted after the abort (because in a render, pending chunks become errored instead of halting). This logic is encapsulated in `collectPrerenderChunksWeb`, which also tracks if the stream is still pending.

In order to get debug info for hanging promises in `--debug-prerender`, we also need to still collect the debug chunks emitted after the abort and give them to the client just before aborting (so that the debug info can be read, but nothing new can render). We already do this during the dev staged render, so we can just re-use the existing `createNodeStreamWithLateRelease` to achieve this.

I've also removed the unecessary microtasks from stale time and vary params tracking. we need to wait an immediate for them to actually get written into the stream, and the awaits weren't doing anything useful.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -189,8 +189,8 @@ import {
   type OpaqueFallbackRouteParams,
 } from '../request/fallback-params'
 import {
+  ReactServerPrerenderResult,
   createReactServerPrerenderResult,
-  type ReactServerPrerenderResult,
   ReactServerResult,
   ReplayableNodeStream,
   createReactServerPrerenderResultFromRender,
@@ -219,7 +219,6 @@ import {
   StaleTimeIterable,
   createSelectStaleTime,
   trackStaleTime,
-  finishStaleTimeTracking,
 } from './stale-time'
 
 import { HTML_CONTENT_TYPE_HEADER, INFINITE_CACHE } from '../../lib/constants'
@@ -1029,11 +1028,11 @@ async function generateStagedDynamicFlightRenderResultWeb(
     },
     () => {
       // This is a separate task that doesn't advance a stage. It forces
-      // draining the microtask queue so that the stale time iterable and vary
-      // params accumulators are closed before we advance to the dynamic stage.
-      void finishStaleTimeTracking(staleTimeIterable)
+      // draining the immediate queue so that the stale time iterable and vary
+      // params accumulators are flushed before we advance to the dynamic stage.
+      staleTimeIterable.close()
       if (requestStore.varyParamsAccumulator) {
-        void finishAccumulatingVaryParams(requestStore.varyParamsAccumulator)
+        finishAccumulatingVaryParams(requestStore.varyParamsAccumulator)
       }
     },
     () => {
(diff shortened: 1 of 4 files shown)


packages/next/src/server/app-render/app-render.tsx
```diff
@@ -1934,16 +1934,17 @@ async function finalRuntimeServerPrerender(
...
     async () => {
       // EarlyStatic stage: render begins.
       // Runtime-prefetchable segments render immediately.
       // Non-prefetchable segments are gated until the Static stage.
       finalStageController.advanceStage(RenderStage.EarlyStatic)
-      const prerenderResult = await workUnitAsyncStorage.run(
+      const stream = workUnitAsyncStorage.run(
         finalServerPrerenderStore,
-        getServerPrerender(ComponentMod),
+        ComponentMod.renderToReadableStream,
```
lubieowoce, line 1947: i'm avoiding forking this for node streams for now. this is not a regression, because at the time of writing, `getServerPrerender` is not actually forked and always returns a web-streams function anyway:
https://github.com/vercel/next.js/blob/d6d392a492c233b1eaafa3caa08b8f57c57e2fe1/packages/next/src/server/app-render/stream-ops.node.ts#L1093

https://github.com/vercel/next.js/blob/d6d392a492c233b1eaafa3caa08b8f57c57e2fe1/packages/next/src/server/app-render/entry-base.ts#L11

lubieowoce:
* **#94192** <a href="https://app.graphite.com/github/pr/vercel/next.js/94192?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94044** <a href="https://app.graphite.com/github/pr/vercel/next.js/94044?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94044?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -8970,6 +8998,76 @@ async function prerenderToStream(
...
+    set isPending(value) {
+      _isPending = value
+    },
+  }
+}
+
+function createPrerenderChunksAccumulator(): PrerenderChunksAccumulator {
+  return {
+    prerenderChunks: [],
+    // In dev, we  alsocollect chunks that the render emits after aborting,
```
unstubbable:
```suggestion
    // In dev, we also collect chunks that the render emits after aborting,
```

packages/next/src/server/app-render/stale-time.ts
```diff
@@ -93,18 +93,9 @@ export function trackStaleTime(
...
- * after. We can't use a macrotask (that would allow dynamic IO to sneak into
- * the response), so we use microtasks instead. The exact number of awaits
- * isn't important as long as we wait enough ticks for Flight to finish writing.
+ * NOTE: This function does not wait for the resulting flight chunks to flush.
+ * The appropriate wait time depends on whether we're using a prerender or a render.
  */
-export async function finishStaleTimeTracking(
-  iterable: StaleTimeIterable
-): Promise<void> {
+export function finishStaleTimeTracking(iterable: StaleTimeIterable): void {
```
unstubbable: The jsdoc (before the `NOTE`) is kinda stale now. Should we just inline the `iterable.close()` call at the call sites and delete this function?

unstubbable (approved):
Clean change! 👍

packages/next/src/server/app-render/stale-time.ts
lubieowoce: hmm yeah i was thinking about it too. i'll do that

PR #94051, opened 2026-05-22
Add Chromium-only Playwright image publisher

timneutkens:
### What?

Add the Next.js-owned Chromium-only Playwright image definition and a workflow that builds and pushes it to GHCR.

### Why?

This needs to merge before the reusable CI workflow starts using the image, so the GHCR tag exists when job containers are initialized.

### How?

- Add a Noble-based Chromium image with the CI build dependencies, Playwright Chromium, and the repository's Node.js LTS version preinstalled.
- Add a manual/canary-only publisher workflow that derives Playwright from `package.json`, Node.js from `build_reusable.yml`, and tags images with a Next.js-owned revision suffix.
- Keep the image revision in the publisher workflow so Dockerfile-only changes can publish a distinct immutable tag.

### Verification

- `git diff --cached --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/playwright_chromium_image.yml")'`
- Not run: `docker buildx build --check --build-arg PLAYWRIGHT_VERSION=1.58.2 .github/images/playwright-chromium` (Docker daemon is not running locally)
- Not run: `pnpm prettier --with-node-modules --ignore-path .prettierignore --write .github/workflows/playwright_chromium_image.yml .github/images/playwright-chromium/Dockerfile` (`prettier` command is not available in this checkout)

### DIFF
--- a/.github/images/playwright-chromium/Dockerfile
+++ b/.github/images/playwright-chromium/Dockerfile
@@ -1,0 +1,43 @@
+# syntax=docker.io/docker/dockerfile:1
+FROM ubuntu:noble
+
+ARG NODE_VERSION=20.9.0
+ARG PLAYWRIGHT_VERSION
+ARG IMAGE_NAME_TEMPLATE=ghcr.io/vercel/next.js-playwright-chromium:v%version%-noble-r1
+
+ENV DEBIAN_FRONTEND=noninteractive
+ENV FNM_DIR=/usr/local/share/fnm
+ENV LANG=C.UTF-8
+ENV LC_ALL=C.UTF-8
+ENV PATH=/usr/local/share/fnm:$PATH
+ENV PLAYWRIGHT_BROWSERS_PATH=/ms-playwright
+ENV TZ=Etc/UTC
+
+RUN apt-get update && \
+    test -n "${NODE_VERSION}" && \
+    test -n "${PLAYWRIGHT_VERSION}" && \
+    apt-get install -y --no-install-recommends \
+      bash \
+      build-essential \
+      ca-certificates \
+      curl \
+      git \
+      jq \
+      openssh-client \
+      unzip \
+      wget && \
+    curl -fsSL https://fnm.vercel.app/install | bash && \
+    fnm install "${NODE_VERSION}" && \
+    fnm default "${NODE_VERSION}" && \
+    eval "$(fnm env)" && \
+    npm install -g yarn && \
+    mkdir -p /ms-playwright /ms-playwright-agent && \
+    cd /ms-playwright-agent && \
+    npm init -y && \
+    npm i "playwright-core@${PLAYWRIGHT_VERSION}" && \
+    npm exec --no -- playwright-core mark-docker-image "${IMAGE_NAME_TEMPLATE}" && \
+    npm exec --no -- playwright-core install --with-deps chromium && \
+    rm -rf /var/lib/apt/lists/* /ms-playwright-agent ~/.npm && \
+    chmod -R 777 /ms-playwright && \
+    useradd --uid 1001 --create-home --shell /bin/bash pwuser && \
+    chown -R pwuser:pwuser "${FNM_DIR}" /home/pwuser
--- a/.github/workflows/playwright_chromium_image.yml
+++ b/.github/workflows/playwright_chromium_image.yml
@@ -1,0 +1,72 @@
+name: build-playwright-chromium-image
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - canary
+    paths:
+      - '.github/images/playwright-chromium/**'
+      - '.github/workflows/playwright_chromium_image.yml'
+      - '.github/workflows/build_reusable.yml'
+      - 'package.json'
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  IMAGE_NAME: ghcr.io/vercel/next.js-playwright-chromium
+  IMAGE_REVISION: 1
+
+jobs:
+  build:
+    if: github.repository == 'vercel/next.js'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v6.0.0
+
+      - name: Resolve Playwright version
+        id: version
+        run: |
+          version="$(yq -r '.devDependencies.playwright' package.json)"
+          node_version="$(yq -r '.env.NODE_LTS_VERSION' .github/workflows/build_reusable.yml)"
+          if [ -z "$version" ] || [ "$version" = "null" ]; then
+            echo "Unable to resolve Playwright version from package.json"
+            exit 1
+          fi
+          if [ -z "$node_version" ] || [ "$node_version" = "null" ]; then
+            echo "Unable to resolve Node.js version from build_reusable.yml"
+            exit 1
+          fi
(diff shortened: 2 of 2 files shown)


.github/workflows/build_reusable.yml
```diff
@@ -236,27 +237,12 @@ jobs:
...
-      # Playwright's GitHub Actions example runs the container as a
-      # non-root user, but this reusable workflow needs root so it can
-      # install jq/unzip/build-essential from the upstream image before
-      # bootstrapping. Firefox still requires $HOME to be owned by the
-      # current user, so re-own GitHub's mounted home directory in the
-      # Playwright container before tests run.
+      # Keep GitHub's mounted home directory owned by the current container
+      # user before bootstrapping tools that write under $HOME.
       - name: Take ownership of /github/home (Linux container)
-        if: ${{ runner.os == 'Linux' && inputs.needsPlaywright != 'no' && inputs.skipInstallBuild != 'yes' && inputs.usePlaywrightContainer != 'no' && !contains(inputs.runs_on_labels, 'windows') }}
```
bgw, line 259: The argument for not being able to run docker with `--user` and needing to do this chown hack as well as the "Mark workspace as safe for Git (Linux container)" hack later in this file was that we needed root to run apt.

Since we're no longer running apt inside the container, let's just pass `container: options: --user pwuser` and finally kill this whole chown thing entirely.

.github/workflows/playwright_chromium_image.yml
```diff
@@ -0,0 +1,60 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v6.0.0
+
+      - name: Resolve Playwright version
+        id: version
+        run: |
+          version="$(node -p "require('./package.json').devDependencies.playwright")"
+          node_version="$(node -e "const fs = require('fs'); const workflow = fs.readFileSync('.github/workflows/build_reusable.yml', 'utf8'); const match = workflow.match(/^\\s*NODE_LTS_VERSION:\\s*([^\\s#]+)/m); if (!match) process.exit(1); process.stdout.write(match[1]);")"
```
bgw: GitHub Actions runner images include `yq`, so let's use that instead of `node` and this regex hack.

- https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md
- https://github.com/mikefarah/yq

.github/workflows/playwright_chromium_image.yml
```diff
@@ -0,0 +1,60 @@
...
+        run: |
+          docker buildx create --use --name nextjs-playwright-builder
+          # Keep this GHCR package public: job containers are pulled before any
+          # workflow step can authenticate to a private registry.
+          docker buildx build \
+            --file .github/images/playwright-chromium/Dockerfile \
+            --build-arg NODE_VERSION="${{ steps.version.outputs.node_version }}" \
+            --build-arg PLAYWRIGHT_VERSION="${{ steps.version.outputs.version }}" \
+            --build-arg IMAGE_NAME_TEMPLATE="ghcr.io/vercel/next.js-playwright-chromium:v%version%-noble" \
+            --tag "ghcr.io/vercel/next.js-playwright-chromium:${{ steps.version.outputs.tag }}" \
```
bgw: Don't interpolate GitHub Actions environment variables this way as it can cause escaping issues. Set these values as intermediate environment variables and then use them inside bash as environment variables.

https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

.github/workflows/playwright_chromium_image.yml
```diff
@@ -0,0 +1,60 @@
...
+            echo "Unable to resolve Node.js version from build_reusable.yml"
+            exit 1
+          fi
+          echo "version=${version}" >> "$GITHUB_OUTPUT"
+          echo "node_version=${node_version}" >> "$GITHUB_OUTPUT"
+          echo "tag=v${version}-noble" >> "$GITHUB_OUTPUT"
+
+      - name: Login to GHCR
+        run: |
+          echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u "${{ github.actor }}" --password-stdin
```
bgw: Same interpolation issue as below.

.github/workflows/build_reusable.yml
```diff
@@ -103,24 +103,27 @@ on:
...
         default: 'mcr.microsoft.com/playwright:v1.58.2-noble'
+      playwrightChromiumDockerImage:
+        description: >
+          Chromium-only Docker image to use for the default Chromium Linux
+          test jobs. This image is maintained in this repository and also
+          includes jq, unzip, and build-essential so container jobs do not need
+          to apt-install them on every run.
+        required: false
+        type: string
+        default: 'ghcr.io/vercel/next.js-playwright-chromium:v1.58.2-noble'
```
bgw: I'm wondering about this version numbering. Don't we need to include a suffix with our own version number? What if `.github/images/playwright-chromium/Dockerfile` is updated but playwright is not? Don't we need to be able to publish a new copy of the docker image with a distinct version number? And how do we remember to update the version here?

.github/images/playwright-chromium/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+      bash \
+      build-essential \
+      ca-certificates \
+      curl \
+      git \
+      jq \
+      openssh-client \
+      unzip \
+      wget && \
+    curl -fsSL https://fnm.vercel.app/install | bash && \
```
bgw, line 29: Out of scope for this PR: We still need to fix how we install fnm so that we pin it and validate a checksum.

bgw (changes requested):

.github/workflows/build_reusable.yml
timneutkens: :robot: Updated this to run Chromium container jobs with `--user pwuser`, removed the `/github/home` chown step, and removed the root-only Git safe.directory workaround.

.github/workflows/playwright_chromium_image.yml
timneutkens: :robot: Switched the version lookup to `yq` for both `package.json` and `NODE_LTS_VERSION` in `build_reusable.yml`.

.github/workflows/playwright_chromium_image.yml
timneutkens: :robot: Moved the resolved image tag, Node version, and Playwright version into step `env` values and the build script now only reads shell variables.

.github/workflows/playwright_chromium_image.yml
timneutkens: :robot: Moved the token and actor into intermediate environment variables before calling `docker login`.

.github/workflows/build_reusable.yml
timneutkens: :robot: Added our own image revision suffix: the default image is now `v1.58.2-noble-r1`, the publisher workflow has `IMAGE_REVISION: 1`, and the input description calls out bumping that revision when the Dockerfile changes without a Playwright version bump.

.github/images/playwright-chromium/Dockerfile
timneutkens: :robot: Agreed. I kept fnm pinning/checksum validation out of this PR so the Chromium-image change stays focused.

bgw (approved):

PR #94061, opened 2026-05-22
turbopack tracing: add chunk_name to `get HMR events` span

wbinnssmith:
This adds the chunk's file path to the span as metadata.

### DIFF
--- a/crates/next-api/src/project.rs
+++ b/crates/next-api/src/project.rs
@@ -2412,6 +2412,12 @@ impl Project {
         // sessions.
         let _ = session;
 
+        #[tracing::instrument(
+            level = "info",
+            name = "get HMR version",
+            skip_all,
+            fields(chunk_name = %chunk_name, target = %target),
+        )]
         #[turbo_tasks::function(operation, root)]
         async fn hmr_version_operation(
             this: ResolvedVc<Project>,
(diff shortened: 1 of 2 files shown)


crates/next-api/src/project.rs
```diff
@@ -2418,6 +2418,7 @@ impl Project {
             chunk_name: RcStr,
             target: HmrTarget,
         ) -> Result<Vc<Box<dyn Version>>> {
+            tracing::info!(chunk_name = %chunk_name, target = %target, "hmr subscription");
```
sokra: This just emits an event. You want to make it a field on a span instead.

e.g. `let _span = tracing::trace_span!("hmr subscription", chunk_name).entered();`
but you need the async variant here (instrument)

crates/next-api/src/project.rs
```diff
@@ -2412,6 +2412,12 @@ impl Project {
         // sessions.
         let _ = session;
 
+        #[tracing::instrument(
+            level = "info",
+            name = "hmr subscription",
```
sokra: hmr_version_operation and hmr_update_with_issues_operation shouldn't have the same name. Make it unique

lukesandberg (approved):

PR #94062, opened 2026-05-23
Turbopack: reduce hmr chunk list subscriptions

wbinnssmith:
Previously, we created chunk list register chunks for every reachable chunk in the chunk graph on a page. Now, we only create one that subscribes to all recursively reachable assets.

This has to pass around an explicit list of client references chunks as those cannot be discovered via the chunk graph alone.

This results in a significant performance improvement when loading pages with the dev server, improving performance of a 60s cold dev startup in a large app by about 10s.


packages/next-routing/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "resolveJsonModule": true,
     "esModuleInterop": true,
     "skipLibCheck": false,
-    "lib": ["es2022", "dom"],
+    "lib": ["es2022", "dom", "dom.iterable"],
```
wbinnssmith, line 10: Typescript was failing without this..

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -879,6 +861,28 @@ impl ChunkingContext for BrowserChunkingContext {
             );
 
             if this.enable_hot_module_replacement {
+                // inner=true follows manifest chunk references to get actual dynamic component
+                // chunks, so the single HMR chunk list covers all lazily-loaded modules.
+                let all_dynamic_chunks = expand_output_assets(
+                    references
+                        .iter()
+                        .copied()
+                        .map(ExpandOutputAssetsInput::Reference),
```
sokra: You can `chain` the extra_chunks here, so it visits them all and dedupes them.
This will make it expand the extra_chunks too, which is relevant for dynamic imports.

crates/next-core/src/next_app/app_client_references_chunks.rs
```diff
@@ -337,3 +337,29 @@ pub async fn get_app_client_references_chunks(
     .instrument(tracing::info_span!("process client references"))
     .await
 }
+
+/// Flattens all client-side output assets from `client_references_chunks` for
+/// use as `extra_chunks` in the entry's evaluated chunk group, so the single
+/// HMR chunk list covers client references that have no chunk list of their own.
+#[turbo_tasks::function]
+pub async fn get_recursive_client_references_chunks_for_hmr(
```
sokra: The `recursive` in the name is confusing. It doesn't recurse?

crates/next-core/src/next_app/app_client_shared_chunks.rs
```diff
@@ -15,13 +16,14 @@ pub async fn get_app_client_shared_chunk_group(
     app_client_runtime_entries: Vc<EvaluatableAssets>,
     module_graph: Vc<ModuleGraph>,
     client_chunking_context: Vc<Box<dyn ChunkingContext>>,
+    extra_chunks: Vc<OutputAssets>,
```
sokra: When you pass page-specific extra chunks here, the `app_client_shared_chunk_group` is no longer shared between pages.
I think it shouldn't pass `extra_chunks` here and use `empty` instead.
The extra chunks are already passed to the client components evaluated chunk group instead, which is page specific

turbopack/crates/turbopack-browser/src/chunking_context.rs
wbinnssmith: :robot: Done — chained \`assets.iter().copied().map(ExpandOutputAssetsInput::Asset)\` into the iterator passed to \`expand_output_assets\`, so both the direct chunk assets and references are expanded together (with deduplication via the adjacency map).

packages/next-routing/tsconfig.json
lukesandberg: turborepo was hiding this error 🤦

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -1314,7 +1314,6 @@ export async function createHotReloaderTurbopack(
           const parsedData = JSON.parse(
             typeof data !== 'string' ? data.toString() : data
           )
-
```
lukesandberg, line 1317: nit: revert file?

test/development/app-dir/hmr-dynamic-component/chunk-list-subscriptions.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    //      reference chunks (built outside the shared module graph).
+    // Previously each `chunk_group(IsolatedMerged)` call for a client
+    // component group produced its own Dynamic chunk list, resulting in N+1
+    // subscriptions per page.
+    await retry(async () => {
+      const chunkListPaths = new Set(
+        sentSubscribes.filter((m) => m.path.endsWith('.js')).map((m) => m.path)
+      )
+      expect(chunkListPaths.size).toBeGreaterThan(0)
+      expect(chunkListPaths.size).toBeLessThanOrEqual(2)
```
lukesandberg: we can't be specific about 1 or 2 for the page in question?

test/development/app-dir/hmr-dynamic-component/hmr-dynamic-component.test.ts
```diff
@@ -0,0 +1,41 @@
+import { join } from 'path'
+import { nextTestSetup } from 'e2e-utils'
+import { retry } from 'next-test-utils'
+
+describe('hmr-dynamic-component', () => {
```
lukesandberg, line 5: don't we have a ton of coverage for basic hmr flows.... what is the unique coverage here?

test/development/app-dir/hmr-dynamic-component/next.config.js
```diff
@@ -0,0 +1,6 @@
+/**
```
lukesandberg, line 1: i think we can just omit these trivial config files

test/development/app-dir/hmr-dynamic-component/chunk-list-subscriptions.test.ts
lukesandberg: or ideally have several tests to trigger both the 1 and 2 cases (pages with and without client components i guess?)

wbinnssmith: yeah this needs to be exact.

test/development/app-dir/hmr-dynamic-component/hmr-dynamic-component.test.ts
wbinnssmith: this one specifically covers the change introduced here: that changes further down the chunk graph from the page entry (i.e. module changes in chunks created because of next/dynamic) are reflected. It's been really helpful making sure we don't regress here, and I don't believe we already had something that did exactly this.

crates/next-api/src/app.rs
```diff
@@ -1320,6 +1299,33 @@ impl AppEndpoint {
...
+                        client_runtime_entries
+                            .iter()
+                            .map(|v| ResolvedVc::upcast(*v))
+                            .collect(),
+                    ),
+                    *module_graphs.full,
+                    AvailabilityInfo::root(),
+                )
+                .await?
+                .availability_info
```
sokra: This is very weird. If you use the avilablility info you probably also need to use the assets for something. They should be made available by adding the asset, so that availablilty_info is actual valid.

sokra: I feel like this is releated to `client_shared_chunks`, but in this case it should be combined and assign `client_shared_chunks` and `availablilty_info` at the same time.

crates/next-api/src/app.rs
```diff
@@ -1351,6 +1374,35 @@ impl AppEndpoint {
...
+        // chunk_group(IsolatedMerged) calls and aren't reachable from the shared client
+        // chunk group's module graph, so they need their own HMR subscription.
+        //
+        // The register chunk is page-specific and must NOT go into client_shared_chunks
+        // (root_main_files), which is shared across all pages. Instead it goes into
+        // root_main_files_per_page so it is serialized under rootMainFilesTree[page] in
+        // the build manifest. The server renderer reads rootMainFilesTree[pagePath] first
+        // (required-scripts.tsx), so only the correct page's register chunk is loaded.
+        let mut page_hmr_chunks: Vec<ResolvedVc<Box<dyn OutputAsset>>> = vec![];
+        if is_app_page && !is_production {
```
sokra: `!is_production` -> `is_hot_module_replacement_enabled`

I think we have a separate flag for that. You may run a development build without hmr enabled

sokra:
```suggestion
        let mut page_hmr_chunks = if is_app_page && !is_production {
```

crates/next-api/src/app.rs
```diff
@@ -1351,6 +1374,35 @@ impl AppEndpoint {
...
+                AssetIdent::from_path(project.project_path().owned().await?)
+                    .with_modifier(rcstr!("client-components"))
+                    .with_modifier(app_entry.original_name.clone())
+                    .into_vc();
+            let client_reference_chunks =
+                get_client_references_chunks_for_hmr(*client_references_chunks);
+            for &asset in client_chunking_context
+                .hmr_chunk_list(client_components_chunks_ident, client_reference_chunks)
+                .await?
+                .iter()
```
sokra:
```suggestion
            asset in client_chunking_context
                .hmr_chunk_list(client_components_chunks_ident, client_reference_chunks)
                .await?
                .iter().copied().collect::<Vec<_>>()
```

crates/next-api/src/app.rs
```diff
@@ -1351,6 +1374,35 @@ impl AppEndpoint {
...
+            {
+                client_assets.insert(asset);
+                page_hmr_chunks.push(asset);
+            }
+        }
+
```
sokra, line 1387: `client_assets.extend(page_hmr_chunks.iter().copied());`

test/development/app-dir/hmr-dynamic-component/hmr-dynamic-component.test.ts
```diff
@@ -0,0 +1,41 @@
...
+        const div = await browser.elementByCss('#dynamic-component')
+        expect(await div.text()).toContain('Dynamic Component UPDATED')
+      }, 10000)
+
+      // Ensure the page was updated via HMR and not a full reload
+      expect(await browser.eval('performance.timeOrigin')).toEqual(timeOrigin)
+    } finally {
+      await next.patchFile(componentPath, originalContent)
+    }
+  })
```
sokra, line 40: Add two more test cases here.

* A page where the dynamic component is only loaded on button click.
    * Click button first, then change code
    * Change code first, then click button.

sokra: Might be easy to add styles to the dynamic component and change these too, to test style HMR

test/development/app-dir/hmr-dynamic-component/hmr-dynamic-component.test.ts
sokra: Isn't test/development/basic/hmr/run-hot-module-reload-hmr-test.util.ts the same?

wbinnssmith: that one only ever tests pages router. might be worth centralizing there though.

crates/next-core/src/next_app/app_client_references_chunks.rs
```diff
@@ -337,3 +337,33 @@ pub async fn get_app_client_references_chunks(
...
+    client_references_chunks: Vc<ClientReferencesChunks>,
+) -> Result<Vc<OutputAssets>> {
+    let client_references_chunks_ref = client_references_chunks.await?;
+    let mut extras: FxIndexSet<ResolvedVc<Box<dyn OutputAsset>>> = FxIndexSet::default();
+    for &assets in client_references_chunks_ref
+        .layout_segment_client_chunks
+        .values()
+    {
+        extras.extend(assets.primary_assets().await?.iter().copied());
+    }
```
lukesandberg: nit: this could be

let mut extras: FxIndexSet<ResolvedVc<Box<dyn OutputAsset>>> = client_references_chunks_ref.layout_segment_client_chunks.values().map(|assets|  async move { Ok(assets.primary_assets().await?.iter().copied())}).try_flat_join.await?.into_iter.collect();

lukesandberg: and then chained onto the below loop

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -903,6 +909,28 @@ impl ChunkingContext for BrowserChunkingContext {
...
     }
 
+    #[turbo_tasks::function]
+    async fn hmr_chunk_list(
+        self: Vc<Self>,
+        ident: Vc<AssetIdent>,
+        chunks: Vc<OutputAssets>,
+    ) -> Result<Vc<OutputAssets>> {
+        let this = self.await?;
+        if !this.enable_hot_module_replacement || chunks.await?.is_empty() {
```
lukesandberg: calling this with enable_hot_module_replacement == false would seem to be a logical error, tag as unreachable?

lukesandberg (approved):

PR #94095, opened 2026-05-25
Extract App Shell from static prefetches

acdlite:
Adds support for extracting an App Shell from a more concrete prerender response. The server sends down a byte offset that represents the subset of the stream that corresponds to the reusable App Shell.

This does _not_ yet implement shell extraction for per-segment prefetch responses. Implementing this adds an additional layer of complexity, because those responses are generated during a separate phase of the build process. We do intend to implement this, but it's a non-essential optimization that can come later.

This also does not yet implement shell extraction from a navigation response (the Cached Navigations feature), though both features are based on essentially the same mechanism. I'm deferring this to a subsequent PR because some of the existing implementation needs to be rethought in light of the new App Shells based model; for example, the "static stage" boundary might not make sense to track separately from the App Shell.

The main practical upshot of the PR is that if you have a fully statically prerendered page with no dynamic holes, that page's App Shell can now be fetched by the client without incurring any runtime server execution cost: the server will return the full static page, and the client will extract the App Shell from that concrete response.

### DIFF
--- a/packages/next/src/client/components/segment-cache/scheduler.ts
+++ b/packages/next/src/client/components/segment-cache/scheduler.ts
@@ -790,8 +790,24 @@ function pingRootRouteTree(
             return PrefetchTaskExitStatus.InProgress
           }
 
-          if (tree.prefetchHints & PrefetchHint.SubtreeHasRuntimePrefetch) {
-            // The route has segments that require a runtime prefetch.
+          // We may need to do a runtime prefetch for one or more segments.
+          // Before checking, we can do some fast checks to bail out of this
+          // branch early.
+          if (
+            // Do any segments have runtime prefetching configured? This is
+            // sent by the server as part of the prefetch hints.
+            tree.prefetchHints & PrefetchHint.SubtreeHasRuntimePrefetch ||
+            // Are we in the Shell prefetching phase? The Shell phase is allowed
+            // to perform runtime prefetches even without an explicit opt-in
+            // because the Shell for a given route is reusable across all given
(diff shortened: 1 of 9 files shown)


packages/next/src/client/components/segment-cache/scheduler.ts
```diff
@@ -1073,7 +1089,27 @@ function pingNewPartOfCacheComponentsTree(
...
+  // (`'3'`) request shape — it has no runtime stage to abort at, so it
+  // responds with the full URL-specific prerender. The follow-up commit
+  // (server-side rewindable prerender) makes the server emit a byte
+  // offset so the client can extract just the shell prefix. Until that
+  // commit lands, this branch sends the request anyway and the client
+  // gracefully falls back to caching whatever the server returned (which
+  // is technically the wrong content at the Fallback vary path, but is
+  // only reachable behind `experimental.appShells`). Note that even
+  // before this widening, the same shape exists today: a `RuntimeShell`
+  // request for a fully-static route already gets the entire page back.
```
lubieowoce: > The follow-up commit (server-side rewindable prerender) makes the server emit a byte offset so the client can extract just the shell prefix. Until that commit lands, [...]

this has already landed, does the logic here need updating?

acdlite: Yeah the comment is outdated. Logic is correct for now. I updated the TODO to describe our plan to send down a hint that indicates whether or not to attempt a static shell prefetch first.

lubieowoce (approved):

PR #94105, opened 2026-05-25
[Skills] next-dev-loop: persist login state across loops

gaojude:
## Summary

The `next-dev-loop` skill currently asks the user to log in at the start of every preflight when the target page is gated. With `agent-browser` 0.27.0's `--state <path>` flag and `--session-name <slug>` auto-save, the saved login can carry from one loop to the next.

This patch updates `skills/next-dev-loop/SKILL.md`:

- **Preflight** — open with `--session-name <slug>` (auto-save on close) and `--state ~/.agent-browser/sessions/<slug>-default.json` when that file already exists (omit on first run; a missing path fails the open).
- **Teardown** — note that `--session-name` writes the state on close, so the next loop's `--state` restores login.

No code changes; doc only.

## Test plan

Verified end-to-end with a headless `claude -p` invocation in a real logged-in Next.js project:

- The fresh agent reads the updated skill, opens `agent-browser` with `--session-name <slug> --state <path>`, and finds itself behind the login — no user input needed.
- After close, the state file's `mtime` advances (auto-save fired).
- Daemon `env` at PID confirms `AGENT_BROWSER_SESSION_NAME` and `AGENT_BROWSER_STATE` are set correctly.

### DIFF
--- a/skills/next-dev-loop/SKILL.md
+++ b/skills/next-dev-loop/SKILL.md
@@ -62,6 +62,7 @@ Once per session, confirm both views are live.
    you can't add `--enable react-devtools` after the session is
    open, and `cookies set` on a not-yet-opened session creates a
    sessionless cookie that silently fails to apply.
+
 2. POST `tools/list` to `/_next/mcp`. Send
    `Accept: application/json, text/event-stream`; responses are
    SSE-framed, strip the `data: ` prefix before parsing JSON.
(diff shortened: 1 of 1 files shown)


aurorascharff (approved):

CertifiedWebMaster:
Wow seriously good work we applaud you for being a pioneer in the industry just last year I was wondering what your systems and software was and started to learn it and now it's so vitaly system and operation in the workplace today!!

PR #94119, opened 2026-05-26
fix(next-mdx): forward plugin load errors to webpack callback

### LINKED ISSUE #94118: @next/mdx: unhandled promise rejection when plugin path fails to resolve in mdx-js-loader

https://github.com/Jinoko01/mdx-catch-bug

### To Reproduce

1. Clone the reproduction repo
2. Run `npm install`
3. Run `npm run build`  (next build --webpack)
4. Observe the build hangs indefinitely or throws UnhandledPromiseRejection

Current:
  The build process hangs without any error message, or Node.js throws
  an UnhandledPromiseRejection without webpack properly reporting the error.

  node:internal/process/promises:288
    UnhandledPromiseRejection: Cannot find module 'remark-plugin-does-not-exist'

Expected:
  Webpack should receive the error via callback and report it clearly:
  
  Error: Cannot find module 'remark-plugin-does-not-exist'
  Require stack: ...

```bash
Operating System:
  Platform: win32
  Arch: x64
  Available memory (MB): 32488
  Available CPU cores: 16
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: N/A
  pnpm: 11.1.1
Relevant Packages:
  next: 16.3.0-canary.29 // Latest available version is detected (16.3.0-canary.29).
  eslint-config-next: N/A
  react: 19.2.6
  react-dom: 19.2.6
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Markdown (MDX), Webpack

next build (local)

### Additional context

Root cause: In `packages/next-mdx/mdx-js-loader.js`, the Promise returned by `getOptions()` is missing a `.catch()` handler. When a plugin string path
fails to resolve (e.g., module not found, ESM import failure), the rejection
is never forwarded to webpack's callback, causing the build to stall.

// Current code (line 61)
getOptions(options, this.context).then((userProvidedMdxOptions) => {
  mdxLoader.call(proxy, ...args)
})
// Missing: .catch((error) => callback(error))

Jinoko01:
### What?
Add `.catch()` handler to the Promise in `mdx-js-loader.js` so that
plugin resolution failures are properly forwarded to webpack's error callback.

### Why?
When a plugin string path fails to resolve (e.g. typo in plugin name,
missing package), `getOptions()` returns a rejected Promise. Without
`.catch()`, webpack's callback is never called, causing the build to
hang indefinitely or throw an unhandled promise rejection instead of
a clear error message.

### How?
Chain `.catch((error) => callback(error))` after the existing `.then()`
in `nextMdxLoader`.

Fixes #94118


timneutkens (changes requested):
Can you add a test for the changes. You can use `pnpm new-test` (or ask an AI agent)

Jinoko01:
> Can you add a test for the changes. You can use `pnpm new-test` (or ask an AI agent)

Thanks for the review! I've added a test in test/e2e/app-dir/mdx/mdx.test.ts that verifies the error is properly forwarded to webpack's callback when a plugin path fails to resolve.

The test patches next.config.ts with a non-existent remark plugin, attempts a production build, and asserts two things:

1. The plugin name appears in the CLI output — confirming the error was passed to webpack's callback
2. UnhandledPromiseRejection does not appear — confirming the rejection is handled, not silently swallowed

test/e2e/app-dir/mdx/mdx.test.ts
```diff
@@ -1,5 +1,54 @@
...
+      '@mdx-js/loader': '^2.2.1',
+      '@mdx-js/react': '^2.2.1',
+    },
+    env: {
+      WITH_MDX_RS: 'false',
+    },
+    skipStart: true,
+  })
+
+  if (!isNextStart) {
```
timneutkens: This will still create the isolated dir before skipping for dev. It's preferable to skip based on importing `isNextStart` from e2e-utils in this case.

```
import { nextTestSetup, isNextStart } from 'e2e-utils'
```

Then use it to skip the describe:

```
;(isNextStart ? describe : describe.skip)('mdx without-mdx-rs - invalid plugin error handling', () => {
```

Jinoko01: Good catch! Updated to import isNextStart directly from e2e-utils and wrap the describe with (isNextStart ? describe : describe.skip)(...). This way nextTestSetup is never called in dev mode, so the isolated directory won't be created unnecessarily.

PR #94128, opened 2026-05-26
Fix Firefox refresh loop on initial load of streaming pages in dev mode

unstubbable:
Follow-up to #92892 and #93486. Those PRs introduced `wasServedFromCache()` to detect HTTP cache restores so the dev-mode debug channel could be replayed from `sessionStorage` (or, as a fallback, the document could be force-reloaded). The check used `PerformanceNavigationTiming.transferSize === 0` as the sole signal.

In Firefox, `transferSize` stays at `0` for the entire window while a streaming response body is still arriving. On the initial load of a page that uses `Suspense` + dynamic content, the bootstrap ran before the body had finished streaming, `wasServedFromCache()` returned `true`, no `sessionStorage` entry existed for the request, and the `location.reload()` safety net fired — only to land in the same condition again, producing an infinite refresh loop. Chrome and Safari report a non-zero `transferSize` during streaming, so they were unaffected.

`wasServedFromCache()` now combines two signals that fire on a real cache restore but not on a streaming load:

- `type === 'back_forward'` for browser-history navigation (in all three browsers) and for tab duplication in Chrome and Firefox.
- `responseStart === 0 && responseEnd > 0` for Safari tab duplication, where `type` stays `'navigate'` and `responseStart` is `0` when no first-body-byte arrived over the network.

A regression test under `test/e2e/app-dir/bfcache-regression` covers a streaming page and asserts that no extra `load` event fires after the initial navigation. The test is also wired into the Firefox and Safari dev jobs in CI so future regressions are caught there. Tab duplication cannot be simulated in Playwright and was verified manually across all three browsers, same as in the original PR.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -916,10 +916,12 @@ jobs:
           test/production/chunk-load-failure/chunk-load-failure.test.ts
 
         TURBOPACK_DEV=1 NEXT_TEST_MODE=dev BROWSER_NAME=firefox node run-tests.js \
-          test/e2e/app-dir/scss/hmr-module/hmr-module.test.ts
+          test/e2e/app-dir/scss/hmr-module/hmr-module.test.ts \
+          test/e2e/app-dir/bfcache-regression/bfcache-regression.test.ts
 
         TURBOPACK_DEV=1 NEXT_TEST_MODE=dev BROWSER_NAME=safari node run-tests.js \
-          test/e2e/app-dir/scss/hmr-module/hmr-module.test.ts
+          test/e2e/app-dir/scss/hmr-module/hmr-module.test.ts \
+          test/e2e/app-dir/bfcache-regression/bfcache-regression.test.ts
 
         TURBOPACK_BUILD=1 NEXT_TEST_MODE=start BROWSER_NAME=safari node run-tests.js \
           test/production/pages-dir/production/test/index.test.ts \
--- a/packages/next/src/client/dev/debug-channel.ts
+++ b/packages/next/src/client/dev/debug-channel.ts
@@ -51,7 +51,36 @@ function wasServedFromCache(): boolean {
     // There is exactly one PerformanceNavigationTiming entry per page load.
     const entry = performance.getEntriesByType('navigation')[0]
 
-    return entry?.transferSize === 0
+    if (!entry) {
+      return false
+    }
+
+    // HTTP cache restore detection isn't uniform across browsers, so we combine
+    // two signals:
+    //
+    //   1. type === 'back_forward' — set on browser-history navigations
+    //      (back/forward) in all three browsers, and on tab duplication in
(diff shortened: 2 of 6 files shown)


lubieowoce (approved):
not gonna lie, the wall-of-text explanation makes me feel like this solution is a bit brittle... but i guess browser differences like are just a fact of life, and we have tests so it should be fine 🤷‍♀️

NanningR:
Could this be backported to the 16.2.x branch?

16.2.7 backported #92892 as #93492, which introduced the `transferSize === 0` cache
detection and the `location.reload()` fallback in `client/dev/debug-channel.js`. The
Firefox refresh loop that this PR fixes therefore exists in every 16.2.x release from
16.2.7 onward, including 16.2.12, which is currently `latest`.

This PR landed on 2026-05-26, before 16.2.12 shipped on 2026-07-25, but does not appear
in the 16.2 backport branch.

It is not a rare edge case: any dev page that streams reproduces it in Firefox. Users on
the stable line also cannot avoid it by staying current, since 16.2.11 carries the
security fixes.

### Reproduction on 16.2.12

`app/page.tsx`, no other dependencies or config:

```tsx
import { Suspense } from "react";

async function Slow() {
	await new Promise((r) => setTimeout(r, 400));
	const rows = Array.from({ length: 4000 }, (_, i) => <li key={i}>row {i} lorem ipsum dolor sit amet</li>);
	return <ul>{rows}</ul>;
}

	return (
		<main>
			<Suspense fallback={<p>loading…</p>}>
				<Slow />
			</Suspense>
		</main>
	);
}
```

`next dev`, then load the page.

| | Navigations | Reloads | `transferSize` at `readyState === "interactive"` |
| --- | --- | --- | --- |
| Chromium 151 | 1 | 0 | 104305 |
| Firefox 153 | 36 | 35 | 0 |

Numbers above are Playwright's bundled builds so they are scriptable, but this was first
hit in regular desktop Firefox; Chrome and Edge are fine. A page small enough not to
stream does not reproduce, which is why a bare `create-next-app` looks healthy.

Sampling from an init script in every document, Firefox logs
`{"type":"reload","transferSize":0}` repeatedly: the response is still streaming when
`wasServedFromCache()` runs during hydration, so `transferSize` is still 0,
`restoreDebugChannelFromSessionStorage` finds no entry for the freshly minted
`requestId`, and `location.reload()` mints another one. It never converges.

On `16.3.0-canary.104` the same page is 1 navigation and 0 reloads in Firefox, even
though `transferSize` is still 0 at that point, which is exactly the distinction this
PR draws between a streaming response and a real cache restore.

### Workaround for anyone hitting this on 16.2.x

```ts
// next.config.ts
experimental: {
	reactDebugChannel: false
}
```

Environment for the reproduction above: Next 16.2.12, React 19.2.0, Node 24.15.0,
Linux x64, Turbopack.

unstubbable:
@NanningR You can now upgrade to 16.3.0 to get this fix.

PR #94139, opened 2026-05-27
fix: dispatch ACTION_RESTORE synchronously after browser back/forward navigation

### LINKED ISSUE #93905: Bug: Scripts disabled due to cache error when navigating back via browser history (Regression) Next JS 16.2.6

https://github.com/barisdemirr/nextjs-16.2.6-bug-report

### To Reproduce

1. Start the development server (`npm run dev`).
2. Open the application in any browser (Chrome, Edge, etc.).
3. Navigate from the initial page to any other route using client-side navigation.
4. Click the browser's "Back" button to return to the previous page.
5. Try interacting with any element that requires JavaScript (or check if initial scripts run). 
6. Observe that scripts are not executing and no errors are logged in the console.

When navigating back via the browser history, the page should restore correctly, and all client-side scripts should execute and remain functional without silent failures.

```bash
- OS: Windows 10 Pro
- Browser: Tested on multiple browsers (Chrome, Edge, etc.)
- Next.js version: 16.2.6
- Node.js version: v24.11.1
- React version: 19.2.4
```

cacheComponents, Script (next/script), Route Handlers

### Additional context

This issue was highly frustrating to debug because of the absolute lack of error logs, warnings, or stack traces in the browser console. The page visually restores itself from the cache perfectly, but all JavaScript execution completely freezes up and client-side interactions become totally unresponsive only after using the browser's "Back" button.

I tested this behavior thoroughly across multiple browsers (including Chrome and Edge) on Windows 10 Pro, and the result is entirely consistent. It acts as a major blocker for daily local development since history navigation is a frequent part of the workflow. 

The issue only triggered immediately after upgrading to Next.js v16.2.6. Rolling back to the previous working version instantly resolved the issue, which strongly indicates a regression within the router's client-side caching mechanism or bfcache handling in this specific release.

arab971:
### What?

Fixes a critical issue where browser back/forward navigation leaves the page visually restored but functionally "frozen" (no JS execution, no interactivity). Fixes #93905.

### Why? (The BFCache vs Hydration Distinction)

This is NOT a hydration issue (which was fixed in #92892 and #93486). Those addressed **HTTP Cache Restore**, where the page re-bootstraps from scratch.

This is a **BFCache (Back/Forward Cache) Restore** issue. In BFCache, the JS heap is preserved. The page "wakes up," but the critical state sync (\`ACTION_RESTORE\`) is deferred by three layers:
1. \`startTransition\` in \`app-router.tsx\`
2. \`useTransition\` in \`use-action-queue.ts\`
3. \`async runAction\` in \`app-router-instance.ts\`

These layers defer the React commit indefinitely, leaving the page as a frozen snapshot.

### How to Reproduce (The "Freeze")

1. Use a page with a \`useEffect\` that logs to the console.
2. Navigate to a second page.
3. Press the Browser Back button.
4. **Observation**: The page restores visually, but the \`useEffect\` does NOT fire, and the page is unresponsive to clicks.

### How?

- **Synchronous Dispatch**: Removed \`startTransition\` from \`onPopState\` and \`applyUrlFromHistoryPushReplace\`.
- **Bypass Deferral**: Modified \`use-action-queue.ts\` to bypass \`appDevRenderingIndicator\` for \`ACTION_RESTORE\` actions.
- **Eliminate Microtask**: Removed \`async\` from \`runAction\`. Since the reducers are synchronous, the \`async\` wrapper was adding a microtask delay that prevented immediate commit.

### Verification

- \`pnpm test-dev-turbo test/e2e/app-dir/popstate-useeffect-timing/popstate-useeffect-timing.test.ts\` passed.
- Verified that \`useEffect\` now fires immediately upon back navigation.

### DIFF
--- a/packages/next/src/client/components/app-router-instance.ts
+++ b/packages/next/src/client/components/app-router-instance.ts
@@ -92,7 +92,7 @@ function runRemainingActions(
   }
 }
 
-async function runAction({
+function runAction({
   actionQueue,
   action,
   setState,
@@ -225,7 +225,7 @@ export function createMutableActionQueue(
     state: initialState,
     dispatch: (payload: ReducerActions, setState: DispatchStatePromise) =>
       dispatchAction(actionQueue, payload, setState),
-    action: async (state: AppRouterState, action: ReducerActions) => {
+    action: (state: AppRouterState, action: ReducerActions) => {
       const result = reducer(state, action)
       return result
     },
--- a/packages/next/src/client/components/app-router.tsx
+++ b/packages/next/src/client/components/app-router.tsx
@@ -1,7 +1,6 @@
 import React, {
   useEffect,
   useMemo,
-  startTransition,
   useInsertionEffect,
   useDeferredValue,
 } from 'react'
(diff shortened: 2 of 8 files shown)


unstubbable:
@arab971 How is this related to #93905 and #94254? Those two issues should already be fixed in canary (via #92892 and #93486).

PR #94146, opened 2026-05-27
fix(dev): enable polling for file watching in Docker containers

### LINKED ISSUE #71622: Hot reload doen't work inside Docker container

https://github.com/ArtoszBart/next-issue

### To Reproduce

1. `docker-compose up`
2. make change in page.tsx
3. save changes

Hot reload should reload the page to show changes, but changes is not visible in the browser

```bash
Physical PC:
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 10 Education
  Available memory (MB): 16319
  Available CPU cores: 16
Binaries:
  Node: 20.18.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 15.0.0 // Latest available version is detected (15.0.0).
  eslint-config-next: 15.0.0
  react: 19.0.0-rc-65a56d0e-20241020
  react-dom: 19.0.0-rc-65a56d0e-20241020
  typescript: 5.6.3
Next.js Config:
  output: N/A

Container:
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP Fri Mar 29 23:14:13 UTC 2024
  Available memory (MB): 7911
  Available CPU cores: 16
Binaries:
  Node: 20.18.0
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 15.0.0 // Latest available version is detected (15.0.0).
  eslint-config-next: 15.0.0
  react: 19.0.0-rc-65a56d0e-20241020
  react-dom: 19.0.0-rc-65a56d0e-20241020
  typescript: 5.6.3
Next.js Config:
  output: N/A
```

Developer Experience

next dev (local)

### Additional context

After saving changes in page.tsx on the PC, the .next/server/chunks/ssr on PC is updated same as page.tsx in container. However .next/server/chunks/ssr in container remains not updated.

SpencerJung:
## Description
This PR adds automatic Docker environment detection and enables polling mode for file watching when running inside Docker containers.

## Problem
Next.js 15 changed the file watching mechanism, which broke hot reload inside Docker containers. Users had to manually set WATCHPACK_POLLING=true or configure webpackDevMiddleware as workarounds.

## Solution
- Added isDocker() utility function that detects Docker environment by checking for /.dockerenv file or docker in /proc/self/cgroup
- Updated Webpack aseWatchOptions to automatically enable polling (1000ms interval) when running in Docker
- Updated Turbopack pollIntervalMs to automatically enable polling when running in Docker
- Added unit tests for Docker detection

## Verification
- [x] LSP diagnostics clean on changed files
- [ ] Type check passed (requires pnpm install)
- [ ] Tests passed (requires pnpm install)

Closes #71622

### DIFF
--- a/packages/next/src/build/webpack-config.ts
+++ b/packages/next/src/build/webpack-config.ts
@@ -7,6 +7,7 @@ import path from 'path'
 import fs from 'fs'
 
 import { getDefineEnv } from './define-env'
+import { isDocker } from '../lib/is-docker'
 import { escapeStringRegexp } from '../shared/lib/escape-regexp'
 import { WEBPACK_LAYERS, WEBPACK_RESOURCE_QUERIES } from '../lib/constants'
 import type { WebpackLayerName } from '../lib/constants'
@@ -147,6 +148,7 @@ const nodePathList = (process.env.NODE_PATH || '')
 
 const baseWatchOptions: webpack.Configuration['watchOptions'] = Object.freeze({
   aggregateTimeout: 5,
+  poll: isDocker() ? 1000 : undefined,
   ignored:
     // Matches **/node_modules/**, **/.git/** and **/.next/**
     /^((?:[^/]*(?:\/|$))*)(\.(git|next)|node_modules)(\/((?:[^/]*(?:\/|$))*)(?:$|\/))?/,
--- a/packages/next/src/lib/is-docker.test.ts
+++ b/packages/next/src/lib/is-docker.test.ts
@@ -1,0 +1,33 @@
+import { isDocker } from './is-docker'
+import fs from 'fs'
+
+jest.mock('fs', () => ({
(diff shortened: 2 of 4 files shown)


bgw:
`WATCHPACK_POLLING=true` is never a good solution. We do not want to encourage it. It has abysmal performance characteristics.

It's also not always needed, depending on how Docker is being used. On a Linux host, it should never be needed, as Docker uses a bind mount and inotify is handled as-normal by the kernel. This change would be a significant regression for those users.

Rather, this is a long-standing feature gap in VirtioFS, which Docker Desktop uses to run Docker inside of a VM on macOS and Windows.

If we're not doing a good job detecting and notifying users if inotify isn't working, we could consider changes there, but we should not automatically turn this on.

PR #94159, opened 2026-05-27
Fix short circuit evaluation of AND / OR in the Turbopack analyzer for string & nullish-related methods

sampoder:
Here is the original description of the bug:

```
1. shortcircuit_if_known predicates are swapped for Logical(And) and
Logical(Or) (lines 2541–2548, 2573–2582, 2634–2643)

is_nullish, is_string, and is_empty_string all do this for AND/OR:

LogicalOperator::And => shortcircuit_if_known(list, JsValue::is_truthy, …),
LogicalOperator::Or => shortcircuit_if_known(list, JsValue::is_falsy, …),

shortcircuit_if_known returns item_value(item) for the first item where
use_item == Some(true). But:
- a && b returns a when a is falsy, else b → use_item should be is_falsy.
- a || b returns a when a is truthy, else b → use_item should be is_truthy.

Concretely:
- is_nullish(1 && null): code does is_truthy(1) = Some(true) → return
is_nullish(1) = Some(false). JS answer is true. Wrong.
- is_nullish(null || 2): code does is_falsy(null) = Some(true) → return
is_nullish(null) = Some(true). JS answer is false. Wrong.
```

As mentioned, in the issue `is_string`, `is_empty_string`, and `is_nullish`'s short circuit evaluation of `&&` and `||` was reversed. This PR reverses them so that for AND expressions, the first false-y value is evaluated (otherwise the last value is). For OR expressions, the first truth-y value is evaluated (otherwise the last value is). 

I’ve added unit tests to this PR. To verify these unit tests with JavaScript’s behaviour I used Claude Code to generate this node script:

```javascript
const PREDICATES = {
  is_string: (v) => typeof v === 'string',
  is_empty_string: (v) => v === '',
  is_nullish: (v) => v == null,
  is_not_nullish: (v) => v != null,
}

const cases = [
  {
    test: 'is_string_constant',
    rust: `assert_eq!(value.is_string(), Some(true));`,
    expr: `'hello'`,
    predicate: 'is_string',
    expected: true,
  },

  {
    test: 'is_string_and_short_circuit',
    rust: `assert_eq!(eval("'hello' && 2").is_string(), Some(false));`,
    expr: `'hello' && 2`,
    predicate: 'is_string',
    expected: false,
  },
  {
    test: 'is_string_and_short_circuit',
    rust: `assert_eq!(eval("2 && 1 && 'hello'").is_string(), Some(true));`,
    expr: `2 && 1 && 'hello'`,
    predicate: 'is_string',
[...]

PR #94163, opened 2026-05-27
Update `shortcircuit_if_known` to `eval_shortcircuit()`

sampoder:
Originally this PR was a comment fix that was flagged by @bgw's Claude:

```

  ▎ "Otherwise returns None."

  The function actually returns item_value(last_item) when no element matches.
  Comment doesn't describe behavior.

  ---
  The two most consequential bugs are #1 (swapped AND/OR predicates) and #2 
  (regex). #1 silently produces wrong nullishness/string-ness results for any
  code path that asks about &&/|| chains and feeds into linker decisions; #2
  silently corrupts every default-filter require.context resolution.
``` 

Related to #94159, updates the comment to say that it returns the last item. 

Which appears to be what is happening:

```rust
if it.peek().is_none() {
    return item_value(item);
} 
```

Afterwards, it now includes a reworking of `shortcircuit_if_known` into `eval_shortcircuit()`: https://github.com/vercel/next.js/pull/94163#discussion_r3313537931


turbopack/crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2789,7 +2789,7 @@ fn any_if_known<T: Copy>(
 }
 
 /// Selects the first element of the list where `use_item` is compile-time true.
-/// For this element returns the result of `item_value`. Otherwise returns None.
+/// For this element returns the result of `item_value`. Otherwise returns the last item.
 fn shortcircuit_if_known<T: Copy>(
     list: impl IntoIterator<Item = T>,
     use_item: impl Fn(T) -> Option<bool>,
```
lukesandberg: <p class='graphite__hidden'><i>[Re: line +2796]</i></p>
i think the item_value function is kind of confusing

perhaps this should be renamed to `return_matching_or_last(list: impl IntoIterator<Item = T>, matches: impl Fn(T) -> Option<bool>) -> Option<T>`

and then callers can just `.map(item_value)` as needed?

im not sure if this would be clearer but it makes some sense.

the empty case is also interesting, should we just reject it and require the list be non-empty
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/vercel/next.js/94163?utm_source=unchanged-line-comment">Graphite</a>.</p>

lukesandberg: lets try `eval_shortcircuit` for a name

turbopack/crates/turbopack-ecmascript/src/analyzer/mod.rs
```diff
@@ -2525,12 +2527,12 @@ impl JsValue {
...
             JsValue::Logical(_, op, list) => match op {
-                LogicalOperator::And => {
-                    shortcircuit_if_known(list, JsValue::is_truthy, JsValue::is_nullish)
-                }
-                LogicalOperator::Or => {
-                    shortcircuit_if_known(list, JsValue::is_falsy, JsValue::is_nullish)
-                }
+                LogicalOperator::And => eval_shortcircuit(list, JsValue::is_truthy)
+                    .map(JsValue::is_nullish)
+                    .flatten(),
```
bgw:
```suggestion
                LogicalOperator::And => eval_shortcircuit(list, JsValue::is_truthy)?.is_nullish(),
```

bgw (changes requested):

sokra (approved):

bgw (approved):

PR #94173, opened 2026-05-27
[turbopack] Enable Effects to be evicted

lukesandberg:
### What?
Enable Effects to be evicted.

Currently `EffectInstance` transitively holds a `ReadRef<PersistentFileContent>` which means eviction cannot not free the memory for the outputs.  Because the `ReadRefs` are captured by the task producing the `Effect` we cannot just drop them (otherwise we could need to somehow re-execute the write to recover them).  So instead we need a new 'effect processing phase'

Here is our new state machine:
 * Phase 1: `emit` effects.
      - At this point we capture the content as a `ResolvedVc` and the hash as a value.
      - In principle this means the hash and content can get out of sync, but only within an eventual consistency race which callers should be handling.
  * Phase 2: `take_effects`.
      - This will call `capture()` on all the effects which will conditionally read the content cell if the hash doesn't match the `EffectStateStorage`
  * Phase 3: `Effects::apply`.
      - This will apply the effects for all effects whose hashes don't match storage.
      - At this point we may discover that a hash in storage changed between `capture` and `apply` which will trigger a special `Retry` error and `invalidate` `Phase 2`.

To facilitate the `Retry` logic a `read_strongly_consistent_and_apply_effects` helper function handles it.  Which allows for a simple circuit breaker and some verbose warnings/errors when we retry too many times.

Specifically:
- Split the `Effect` trait into `Effect` (emit-side, returns a `Captured` payload) and `CapturedEffect` (apply-side). `
- Switch `Effects` to `cell = "new"`. Every producer re-execution allocates a fresh cell value; the prior cell (and its captured payload) drops on overwrite, cascade-releasing upstream `ReadRef<PersistedFileContent>` strong counts. The captured payload also drops naturally when the cell is evicted.
- Push the per-key state machine (`Unapplied` / `InProgress` / `Applied { value_hash, result }`, `EventGuard` panic recovery, `Event` listener coordination) down into `EffectStateStorage::run_apply`. `Effects::apply` becomes a thin aggregator over `dyn_apply()` calls.
- New `ApplyOutcome<E> { Failed(E), Retry }`. `CapturedEffect::apply` returns this; `Retry` signals "capture elided content but storage diverged before apply" — `Effects::apply` collects across the batch, fires the producer's invalidator once, and surfaces `EffectsError::Retry`.
- New `EffectStateStorage::matches_applied(key, hash)`. `WriteEffect::capture` / `WriteLinkEffect::capture` consult this and skip the `ReadRef<PersistedFileContent>` / `ReadRef<LinkContent>` materialization when storage already holds `Applied { matching, Ok(()) }` — avoiding the disk-read + decompression hit on producer re-runs that don't actually change output.
- 10 integration tests in `turbopack/crates/turbo-tasks-backend/tests/effects.rs` cover the new lifecycle: duplicate apply, sibling stomp re-apply, repeated apply with unchanged state dedupe, `cell = "new"` producing distinct cells per producer run, capture-skip on storage match, and the capture-skip-then-stomp race that exercises the `Retry` path.

### Why?

A memory audit showed `EffectInstance` cells accumulating the majority of the ram in an eviction session

#### Capture-time storage check (the perf story)

`Effect::capture` may consult `EffectStateStorage::matches_applied` and elide the content materialization when storage already records the target hash. `CapturedWriteEffect.content` becomes `Option<ReadRef<PersistedFileContent>>` — `None` means "capture observed `Applied { matching }`; no `body` to pass to `run_apply`". The apply-side state machine then either dedup-hits (storage still matches → cached `Ok`) or enters the no-body branch and returns `ApplyOutcome::Retry`.

Reading shared mutable `EffectStateStorage` from inside a turbo-tasks task is normally suspect (the state isn't a tracked input). It is sound here because the apply-time re-check fires the producer's invalidator on mismatch, turning the otherwise-untracked read into an explicitly-tracked one via the `Retry` pathway. The producer reruns, `capture` sees the new storage state, materializes content, and the next `apply` succeeds.

#### `ApplyOutcome::Retry` and the invalidator

`Effects::apply` collects `Retry` signals across the parallel batch via `AtomicBool`. `Failed(e)` errors fail-fast. After the batch, if any `Retry` fired (and no `Failed`), the producer's invalidator runs once and `EffectsError::Retry` propagates. Callers re-read the operation; the producer reruns; fresh `capture` either materializes content (storage diverged) or dedup-hits cleanly.

#### `turbo-tasks-fs`

`WriteEffect::capture` / `WriteLinkEffect::capture` call `matches_applied` first and conditionally `.await?` the content `ResolvedVc`. `CapturedWriteEffect::apply` / `CapturedWriteLinkEffect::apply` dispatch through `run_apply(key, hash, body)`, where `body` is `Some(closure)` iff content is `Some`. `apply_inner` takes the unwrapped `&ReadRef<…>` as a parameter so the option-unwrap is at the dispatch boundary (type-enforced).

### Risk: infinite retry loop

**Failure mode.** Two sibling producers A and B contend on the same key with different hashes (e.g. two routes that both want to write `chunks/foo.js` with different content). Each `Retry` from A's apply invalidates A's producer; each `Retry` from B's apply invalidates B's producer. In the worst case, A and B perpetually re-capture, race to apply, stomp each other's storage state, and re-trigger Retry — invalidating themselves on each cycle.

**Why we believe it terminates in practice.** A `Retry` only fires when (a) capture observed `Applied { matching }` and elided content, and (b) by apply time storage was stomped to a different hash. Each `Retry` invalidates only its own producer (no cross-producer invalidation), so without external input churn the system reaches a fixed point — whichever producer applies last wins, and subsequent re-runs of the loser's producer materialize content (because storage no longer matches) and apply cleanly. The pathological case requires a real, ongoing race fundamentally triggered by external invalidations (file edits).

### DIFF
--- a/crates/next-napi-bindings/src/next_api/project.rs
+++ b/crates/next-napi-bindings/src/next_api/project.rs
@@ -51,7 +51,7 @@ use turbo_tasks::{
     TransientInstance, TryJoinIterExt, TurboTasksApi, TurboTasksCallApi, UpdateInfo, Vc,
     mark_top_level_task,
     message_queue::{CompilationEvent, Severity},
-    take_effects,
+    read_strongly_consistent_and_apply_effects, take_effects,
     trace::TraceRawVcs,
     unmark_top_level_task_may_leak_eventually_consistent_state,
 };
@@ -1406,17 +1406,16 @@ pub async fn project_write_all_entrypoints_to_disk(
                 first_phase,
             );
 
-            // Read and compile the files
+            let read =
+                read_strongly_consistent_and_apply_effects(entrypoints_with_issues_op, |v| {
+                    &v.effects
+                })
+                .await?;
             let AllWrittenEntrypointsWithIssues {
                 entrypoints,
                 issues,
-                effects,
-            } = &*entrypoints_with_issues_op
-                .read_strongly_consistent()
-                .await?;
-
-            // Apply phase side effects. Asset emission is performed once at the end.
-            effects.apply().await?;
+                ..
+            } = &*read;
 
             Ok((
                 entrypoints.clone(),
@@ -1471,16 +1470,16 @@ pub async fn project_write_all_entrypoints_to_disk(
                     EntrypointsWritePhase::Deferred,
                 );
 
+                let read =
+                    read_strongly_consistent_and_apply_effects(entrypoints_with_issues_op, |v| {
+                        &v.effects
+                    })
+                    .await?;
                 let AllWrittenEntrypointsWithIssues {
                     entrypoints,
                     issues,
-                    effects,
-                } = &*entrypoints_with_issues_op
-                    .read_strongly_consistent()
-                    .await?;
-
-                // Apply phase side effects. Asset emission is performed once at the end.
-                effects.apply().await?;
+                    ..
+                } = &*read;
 
                 Ok((
                     entrypoints.clone(),
@@ -1503,17 +1502,16 @@ pub async fn project_write_all_entrypoints_to_disk(
                 app_dir_only,
                 has_deferred_entrypoints,
             );
-            let OperationResult { issues, effects } =
-                &*emit_result_op.read_strongly_consistent().await?;
+            let read =
+                read_strongly_consistent_and_apply_effects(emit_result_op, |v| &v.effects).await?;
+            let OperationResult { issues, .. } = &*read;
 
-            effects.apply().await?;
-
-            Ok(issues.iter().cloned().collect::<Vec<_>>())
+            Ok(issues.clone())
         })
         .or_else(|e| ctx.throw_turbopack_internal_result(&e.into()))
         .await?;
 
-    issues.extend(emit_issues);
+    issues.extend(emit_issues.iter().cloned());
 
     Ok(TurbopackResult {
         result: if let Some(entrypoints) = entrypoints {
@@ -1745,15 +1743,16 @@ pub fn project_entrypoints_subscribe(
         move || {
             async move {
                 let entrypoints_with_issues_op = get_entrypoints_with_issues_operation(container);
+                let read =
+                    read_strongly_consistent_and_apply_effects(entrypoints_with_issues_op, |v| {
+                        &v.effects
+                    })
+                    .await?;
                 let EntrypointsWithIssues {
                     entrypoints,
                     issues,
-                    effects,
-                } = &*entrypoints_with_issues_op
-                    .read_strongly_consistent()
-                    .await?;
-
-                effects.apply().await?;
+                    ..
+                } = &*read;
                 Ok((entrypoints.clone(), issues.clone()))
             }
             .instrument(tracing::info_span!("entrypoints subscription"))
@@ -1865,17 +1864,14 @@ pub fn project_hmr_events(
                         state,
                         hmr_target,
                     );
-                    let update = update_op.read_strongly_consistent().await?;
-                    let HmrUpdateWithIssues {
-                        update,
-                        issues,
-                        effects,
-                    } = &*update;
                     // HACK(bgw): Remove this mark call
                     mark_top_level_task();
-                    effects.apply().await?;
+                    let read =
+                        read_strongly_consistent_and_apply_effects(update_op, |v| &v.effects)
+                            .await?;
                     // HACK(bgw): Remove this unmark call
                     unmark_top_level_task_may_leak_eventually_consistent_state();
+                    let HmrUpdateWithIssues { update, issues, .. } = &*read;
                     match &**update {
                         Update::Missing | Update::None => {}
                         Update::Total(TotalUpdate { to }) => {
@@ -1987,14 +1983,16 @@ pub fn project_hmr_chunk_names_subscribe(
         move || async move {
             let hmr_chunk_names_with_issues_op =
                 get_hmr_chunk_names_with_issues_operation(container, hmr_target);
+            let read =
+                read_strongly_consistent_and_apply_effects(hmr_chunk_names_with_issues_op, |v| {
+                    &v.effects
+                })
+                .await?;
             let HmrChunkNamesWithIssues {
                 chunk_names,
                 issues,
-                effects,
-            } = &*hmr_chunk_names_with_issues_op
-                .read_strongly_consistent()
-                .await?;
-            effects.apply().await?;
+                ..
+            } = &*read;
 
             Ok((chunk_names.clone(), issues.clone()))
         },
@@ -2453,11 +2451,10 @@ pub async fn project_write_analyze_data(
         .turbo_tasks()
         .run_once(async move {
             let analyze_data_op = write_analyze_data_with_issues_operation(container, app_dir_only);
-            let WriteAnalyzeResult { issues, effects } =
-                &*analyze_data_op.read_strongly_consistent().await?;
-
             // Write the files to disk
-            effects.apply().await?;
+            let read =
+                read_strongly_consistent_and_apply_effects(analyze_data_op, |v| &v.effects).await?;
+            let WriteAnalyzeResult { issues, .. } = &*read;
             Ok(issues.clone())
         })
         .await
--- a/turbopack/crates/turbo-tasks-backend/tests/effects.rs
+++ b/turbopack/crates/turbo-tasks-backend/tests/effects.rs
@@ -1,0 +1,736 @@
+#![feature(arbitrary_self_types)]
+#![feature(arbitrary_self_types_pointers)]
+#![allow(clippy::needless_return)] // tokio macro-generated code doesn't respect this
+
+use std::sync::{
+    Arc,
+    atomic::{AtomicU64, Ordering},
+};
+
+use anyhow::Result;
+use async_trait::async_trait;
+use parking_lot::Mutex;
+use rustc_hash::FxHashMap;
+use turbo_rcstr::{RcStr, rcstr};
+use turbo_tasks::{
+    ApplyError, CapturedEffect, Effect, EffectExt, EffectStateStorage, Effects, EffectsError,
+    NonLocalValue, OperationValue, ReadRef, ResolvedVc, State, TurboTasks, Vc, take_effects,
+    trace::TraceRawVcs,
+};
+use turbo_tasks_backend::{BackendOptions, TurboTasksBackend, noop_backing_storage};
+
+/// Per-test counters + `EffectStateStorage`. Tests assert against
+/// `applies_by_key` to check how many times each effect ran.
+///
+/// A single `SharedState` can be associated with multiple `TestInput` cells (each
+/// representing a separate producer task); they then share the same
+/// `EffectStateStorage`, mirroring how on disk two producers writing the
+/// same path contend on the same per-key state entry.
+#[derive(TraceRawVcs, NonLocalValue)]
+struct SharedState {
+    #[turbo_tasks(trace_ignore)]
+    applies_by_key: Mutex<FxHashMap<RcStr, u64>>,
+    #[turbo_tasks(trace_ignore)]
+    total_applies: AtomicU64,
+    /// Counts captures that materialized content (i.e. those where
+    /// `EffectStateStorage::matches_applied` returned false). Lets tests
+    /// assert that capture skipped content materialization on re-runs.
+    #[turbo_tasks(trace_ignore)]
+    captures_with_content: AtomicU64,
+    /// Shared `EffectStateStorage` used by both `matches_applied` (in `capture`)
+    /// and `run_apply` (in `apply`).
+    #[turbo_tasks(trace_ignore)]
+    state_storage: EffectStateStorage,
+}
+
+impl SharedState {
+    fn new() -> Arc<Self> {
+        Arc::new(Self {
+            applies_by_key: Mutex::new(Default::default()),
+            total_applies: AtomicU64::new(0),
+            captures_with_content: AtomicU64::new(0),
+            state_storage: EffectStateStorage::default(),
+        })
+    }
+
+    fn applies_for(&self, key: &RcStr) -> u64 {
+        self.applies_by_key.lock().get(key).copied().unwrap_or(0)
+    }
+
+    fn total(&self) -> u64 {
+        self.total_applies.load(Ordering::Relaxed)
+    }
+
+    fn captures_with_content(&self) -> u64 {
+        self.captures_with_content.load(Ordering::Relaxed)
+    }
+}
+// Emit-side effect. A `#[turbo_tasks::value]` cell implementing the `Effect` value_trait. Marked
+// `serialization = "skip"` because it holds `Arc<SharedState>` (counters + storage, not
+// serializable) — these tests use `noop_backing_storage`, so persistence isn't exercised.
+#[turbo_tasks::value(serialization = "skip", cell = "new", eq = "manual")]
+struct TestEffect {
+    key: RcStr,
+    value_hash: u128,
+    #[turbo_tasks(trace_ignore, debug_ignore)]
+    shared: Arc<SharedState>,
+}
+
+#[async_trait]
+#[turbo_tasks::value_impl]
+impl Effect for TestEffect {
+    async fn capture(&self) -> Result<Box<dyn CapturedEffect>> {
+        // Consult storage. If the per-key state already records `Applied { value_hash }`
+        // matching our hash, elide content materialization (`content = false`). Otherwise
+        // bump the captures-with-content counter (the test's stand-in for a `ReadRef` /
+        // disk-read at this point).
+        let content = if self
+            .shared
+            .state_storage
+            .matches_applied(self.key.as_bytes(), self.value_hash)
+        {
+            false
+        } else {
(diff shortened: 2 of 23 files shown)


lukesandberg:
* **#94173** <a href="https://app.graphite.com/github/pr/vercel/next.js/94173?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94173?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -926,27 +927,62 @@ impl FileSystem for DiskFileSystem {
...
+                // pathway and the producer will rerun with a fresh capture.
+                let key_bytes: Box<[u8]> = self.full_path.as_os_str().as_encoded_bytes().into();
+                let content = if self
+                    .inner
+                    .effect_state_storage
+                    .matches_applied(&key_bytes, self.content_hash)
+                {
+                    None
+                } else {
+                    Some(self.content.await?)
```
lukesandberg: this read should be untracked

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -926,27 +927,62 @@ impl FileSystem for DiskFileSystem {
...
-        let content = content.persist().await?;
-
-        let inner = self.inner.clone();
+        let content = content.persist().to_resolved().await?;
+        let content_hash = u128::from_le_bytes(hash_xxh3_hash128(&*content.await?));
 
         #[derive(TraceRawVcs, NonLocalValue, Clone)]
         struct WriteEffect {
             full_path: Arc<PathBuf>,
             inner: Arc<DiskFileSystemInner>,
```
lukesandberg: pass this to capture instead

or get a Vec<Self> and capture can extract the disk filesystem inner to recover

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -926,27 +927,62 @@ impl FileSystem for DiskFileSystem {
...
         // require recomputing the content on cache restore — avoiding unnecessary downstream
         // recomputation.
-        let content = content.persist().await?;
-
-        let inner = self.inner.clone();
+        let content = content.persist().to_resolved().await?;
+        let content_hash = u128::from_le_bytes(hash_xxh3_hash128(&*content.await?));
 
         #[derive(TraceRawVcs, NonLocalValue, Clone)]
         struct WriteEffect {
```
lukesandberg, line 945: derive Encode/Decode

turbopack/crates/turbo-tasks-fs/src/lib.rs
lukesandberg: :robot: Done — this read is now untracked. `capture` does `(*self.content).untracked().await?` (and `(*self.target).untracked().await?` for links) so it no longer takes a tracked dependency on the content cell that would pin it.

turbopack/crates/turbo-tasks-fs/src/lib.rs
lukesandberg: :robot: Done as part of the serializability redesign. `WriteEffect`/`WriteLinkEffect` are now `#[turbo_tasks::value]` cells holding `fs: ResolvedVc<DiskFileSystem>` instead of `Arc<DiskFileSystemInner>`. `capture` recovers the inner via an untracked read (`(*self.fs).untracked().await?.inner().clone()`) and stores the `Arc` into the (session-only, plain) captured struct. Added a `pub(crate) DiskFileSystem::inner()` accessor for this.

turbopack/crates/turbo-tasks-fs/src/lib.rs
lukesandberg: :robot: Done. The emit-side effects now derive Encode/Decode (via default-serialization `#[turbo_tasks::value]`). To make this actually meaningful I made `Effect` a `#[turbo_tasks::value_trait]` and dropped `EffectInstance`/`EffectCollectible`/`DynEffect` — effects are emitted/collected as `Box<dyn Effect>` collectibles directly. Since each concrete effect is now a serializable cell, the collectible edge persists with its `CellId.type_id` discriminant and restores on a warm cache start without re-running the producer. (`CapturedEffect` stays a plain session-only trait; the captured `ReadRef` content doesn't need to be serializable.)

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -1424,17 +1424,18 @@ pub async fn project_write_all_entrypoints_to_disk(
                 first_phase,
             );
 
-            // Read and compile the files
+            // Read and compile the files, applying phase side effects. Asset emission is
+            // performed once at the end.
```
bgw: This comment basically reads like

```
// apply the effects
apply_effects()
```


```suggestion
```

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -1489,16 +1490,17 @@ pub async fn project_write_all_entrypoints_to_disk(
                     EntrypointsWritePhase::Deferred,
                 );
 
+                // Apply phase side effects. Asset emission is performed once at the end.
```
bgw:
```suggestion
```

turbopack/crates/turbo-tasks-backend/tests/effects.rs
```diff
@@ -0,0 +1,742 @@
+//! Integration tests for `take_effects` and `Effects::apply` semantics.
+//!
+//! These tests use a custom `Effect` impl (`TestEffectEmit`) whose `apply`
+//! increments a shared counter, so we can directly assert how many times each
+//! effect's side-effect actually ran across various producer/apply scenarios:
+
```
bgw:
```suggestion
```

This comment is already out of date. The struct is called `TestEffect`

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -3,54 +3,72 @@ use std::{
...
+    async fn apply(&self) -> Result<(), ApplyOutcome>;
+}
 
-    /// Perform the side effect (write file, create symlink, etc.).
-    fn apply(&self) -> impl Future<Output = Result<(), Self::Error>> + Send;
+/// Outcome of [`CapturedEffect::apply`]. Distinguishes a side-effect failure (terminal) from a
+/// soft failure where the captured form had no content and storage state diverged between
+/// capture and apply (recoverable via [`Effects::apply`]'s invalidator path).
+#[derive(Debug)]
+pub enum ApplyOutcome {
```
bgw: Call this `ApplyError`? Maybe this doesn't need to be `pub`?

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -155,10 +250,16 @@ impl EffectCollectible for EffectInstance {}
...
-    ));
+///
+/// The concrete `effect` is celled and emitted as a `Box<dyn Effect>` collectible. Because each
+/// concrete `Effect` is a serializable [`macro@turbo_tasks::value`] cell, the collectible persists
+/// in the backing store and is restored on a warm cache start without re-running the producer.
+pub fn emit_effect<T>(effect: T)
+where
+    T: Effect + VcValueType<Read = VcDefaultRead<T>> + UpcastStrict<Box<dyn Effect>>,
+{
+    let cell = ResolvedVc::<T>::cell_private(effect);
```
bgw: Follow the pattern that `Issue` does to avoid needing to call the private constructor here.

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -214,22 +315,38 @@ pub fn emit_effect(effect: impl Effect) {
...
+        .take_collectibles::<Box<dyn Effect>>()
         .into_iter()
-        .map(|effect| {
-            if let Some(effect) = ResolvedVc::try_downcast_type::<EffectInstance>(effect) {
-                effect
-            } else {
-                unreachable!("EffectCollectible must only be implemented by EffectInstance");
-            }
-        })
+        .collect();
```
bgw: Why are we recollecting this into a `Vec`? We can just use the `AutoSet` that `take_collectibles` returns directly.

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -238,202 +355,357 @@ struct ConflictingEffectError {
...
+    Conflict(usize),
+
+    /// The captured effects were dropped after a previous successful apply, and the shared
+    /// per-key state for at least one effect no longer carries a matching `Applied { value_hash
+    /// }`. The producing operation has been invalidated; the caller should re-read the operation
+    /// and call `apply()` again on the fresh [`Effects`] value.
+    ///
+    /// Carries the producing task's name and the conflicting key name(s) for diagnostics.
+    /// [`read_strongly_consistent_and_apply_effects`] consumes these to log a warning when a
+    /// retry repeats.
```
bgw: This is a useless comment

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -238,202 +355,357 @@ struct ConflictingEffectError {
...
-type UniqueEffectIndices = Result<Vec<(usize, EffectStateEntry)>, Arc<ConflictingEffectError>>;
+/// Error returned by [`Effects::apply`]. Callers should retry on `Retry`; everything else is
+/// terminal.
+#[derive(thiserror::Error, Debug, Clone)]
+pub enum EffectsError {
+    /// A side effect failed during apply. Holds the first error encountered.
+    #[error(transparent)]
+    Apply(Arc<dyn EffectError>),
+
+    /// Two effects emitted the same key with different value_hashes; no apply happened.
```
bgw: Redundant comment

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -238,202 +355,357 @@ struct ConflictingEffectError {
...
     ///
     /// See [`take_effects`] for example usage.
-    pub async fn apply(&self) -> Result<(), Arc<dyn EffectError>> {
+    ///
+    /// **Do not call this directly.** External callers must go through
+    /// [`read_strongly_consistent_and_apply_effects`] or
+    /// [`read_strongly_consistent_and_apply_effects_with`], which own the read+apply+retry loop
+    /// required to recover from [`EffectsError::Retry`]. Exposed publicly only as
+    /// [`Effects::apply_for_testing`] (`#[doc(hidden)]`) so integration tests can drive the apply
+    /// state machine directly.
```
bgw, line 463: Something about this should go way higher in the doc comment. We should mention something on the first line.

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -238,202 +355,357 @@ struct ConflictingEffectError {
...
+    /// }`. The producing operation has been invalidated; the caller should re-read the operation
+    /// and call `apply()` again on the fresh [`Effects`] value.
+    ///
+    /// Carries the producing task's name and the conflicting key name(s) for diagnostics.
+    /// [`read_strongly_consistent_and_apply_effects`] consumes these to log a warning when a
+    /// retry repeats.
+    #[error(
+        "effect state was reset for {}; producing task {task_name} invalidated, retry required",
+        keys.join(", ")
+    )]
```
bgw, line 346: `keys` could be a super long vec. Maybe let's truncate after a certain point?

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -238,202 +355,357 @@ struct ConflictingEffectError {
...
+
+/// Like [`read_strongly_consistent_and_apply_effects`], but the [`Effects`] directly accessed by
+/// calling [`take_effects`] on the supplied operation.
+///
+/// Unlike [`read_strongly_consistent_and_apply_effects`], this may be called from *inside* a
+/// turbo-tasks task (it owns the `mark`/`unmark` around `apply`). The consequence is that the
+/// effects may be re-applied if that enclosing task is invalidated — acceptable for lazily-created
+/// resources.
+///
+/// AVOID CALLING THIS UNLESS DEEPLY REQUIRED
```
bgw: Move this into the first line of the doc comment

turbopack/crates/turbo-tasks/src/invalidation.rs
```diff
@@ -33,6 +33,13 @@ pub struct Invalidator {
 }
 
 impl Invalidator {
+    /// The [`TaskId`] of the task this invalidator targets. Used for diagnostics
+    /// (e.g. resolving the producing task's name via
+    /// [`TurboTasksApi::get_task_name`][crate::TurboTasksApi::get_task_name]).
+    pub fn task_id(&self) -> TaskId {
+        self.task
+    }
```
bgw: Where is this called? It's typically not actually useful to output this in diagnostics.

turbopack/crates/turbopack-cli/src/build/mod.rs
```diff
@@ -157,10 +158,9 @@ impl TurbopackBuildBuilder {
                     self.scope_hoist,
                 ));
 
-                // Await the result to propagate any errors and capture effects.
-                let effects = wrapper_op.read_strongly_consistent().await?;
-
-                effects.apply().await?;
+                // Await the result to propagate any errors and apply effects (retrying on
+                // `EffectsError::Retry`).
```
bgw: Delete this comment
```suggestion
```

turbopack/crates/turbopack-node/src/evaluate.rs
```diff
@@ -191,23 +191,24 @@ async fn create_evaluate_pool_assets_operation(
...
+    // (rather than a `serialization = "skip"` wrapper carrying the effects) keeps this task's
+    // output restorable from the persistent cache, so a warm restart does not re-run the
+    // effect-producing tasks.
     //
-    // That's not possible here because we lazily create the pool, so instead, use
-    // `mark_top_level_task` to avoid the debug assertion. The consequence is that these effects
-    // might get evaluated more than once if this function is invalidated.
-    mark_top_level_task();
-    effects.apply().await?;
+    // HACK (retained from before the effect refactor): applying effects from inside a task means
```
bgw: > (retained from before the effect refactor)

bgw (approved):

turbopack/crates/turbo-tasks/src/invalidation.rs
lukesandberg: dead

turbopack/crates/turbo-tasks/src/effect.rs
lukesandberg: needs to be pub but ApplyError is a better name

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -955,19 +1003,29 @@ impl FileSystem for DiskFileSystem {
...
+                // spawned task so that multiple pending write effects can execute in parallel
+                // rather than serially on the caller's future (see #94140).
+                let this = self.clone();
+                spawn(async move {
+                    let body = this.content.as_ref().map(|content| {
+                        || async { this.apply_inner(content).await.map_err(AnyhowWrapper::from) }
+                    });
+                    this.inner
+                        .effect_state_storage
+                        .run_apply::<AnyhowWrapper, _, _>(this.key(), this.content_hash, body)
```
lukesandberg: this all seems a bit too complex, perhaps our caller should handle the `spawns`

turbopack/crates/turbo-tasks/src/effect.rs
```diff
@@ -89,78 +120,140 @@ pub struct EffectStateStorage {
...
+    /// Coordinate an apply for `(key, value_hash)` against the per-key state machine.
+    ///
+    /// On a dedup hit (`Applied { value_hash == target, result }`), short-circuits and returns the
+    /// cached result without calling `body`. Otherwise, transitions the entry to `InProgress`,
+    /// awaits any in-flight apply for the same key, calls `body` to perform the side effect, then
+    /// writes back the result as `Applied { value_hash, result }`.
+    ///
+    /// If `body` is `None` the captured effect has no materialized content (capture observed
+    /// matching storage and elided materialization). In that case, a non-matching state forces a
+    /// `Retry` — we have nothing to apply. The state entry is left in `Unapplied` for waiters.
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks/src/invalidation.rs
```diff
@@ -33,6 +33,13 @@ pub struct Invalidator {
 }
 
 impl Invalidator {
+    /// The [`TaskId`] of the task this invalidator targets. Used for diagnostics
+    /// (e.g. resolving the producing task's name via
+    /// [`TurboTasksApi::get_task_name`][crate::TurboTasksApi::get_task_name]).
+    pub fn task_id(&self) -> TaskId {
```
lukesandberg: come up with a better mechanism for debug messages, task name is probably not important

PR #94174, opened 2026-05-27
[ci] Update all actions/* dependencies

bgw:
Previously we pinned all of these to their current versions at the time. That meant that some of these got pinned to really old versions. This updates all of the "first-party" `actions/*` dependencies.

Prompted claude with:

```
For all of these uses of actions/*, make sure we're pulling in the latest stable version (check the gh cli). When upgrading across major versions, fetch the release notes from the GH releases using the gh CLI and make sure that there are no problems that the upgrade will cause.
```

Claude read all the release notes and seems to think these upgrades are safe.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -31,7 +31,7 @@ jobs:
     permissions:
       contents: read
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 25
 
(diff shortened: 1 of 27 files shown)


bgw:
* **#94174** <a href="https://app.graphite.com/github/pr/vercel/next.js/94174?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94174?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

.github/workflows/build_and_test.yml
```diff
@@ -175,7 +175,7 @@ jobs:
 
       - name: Cache pnpm store
         if: ${{ runner.environment == 'github-hosted' }}
-        uses: actions/cache@v5
```
bgw, line 178: This one wasn't actually pinned yet 🙃

eps1lon (approved):

PR #89077, opened 2026-01-27
[Instant] Instant validation in Dev

lubieowoce:
This PR implements the initial version of dev-time validation for `export const instant = ...`

```tsx
export const instant = { prefetch: 'static' }
```
```tsx
export const instant = { prefetch: 'runtime', ... }
```
```tsx
export const instant = false
```
(the config is currently expected to be named `unstable_instant`, but i'll be shortening it to `instant` below.)

When a segment specifies an instant config with `prefetch: 'static'` or `prefetch: 'runtime'`, we'll validate that all (prefetched) navigations into that segment will render instant UI, i.e. that the navigation won't block. On the other hand, if a segment is allowed to block, it must be marked with `export const instant = false`.

Note that our existing static shell validation is a special case of this, but for now, we're keeping both. A future PR will reimplement static shell validation as a special case of instant validation.

Like static shell validation, these validations currently run in dev whenever we render the full page (i.e. no segments are omitted). This means it'll happen on an initial load (or refresh) and HMR.

#### Validation approach

The goal is to simulate what the browser would display for a prefetched client navigation. We do this by re-assembling the segments extracted from the original stream for the page.
- The outer (shared) segments should already fully resolved (i.e. in the Dynamic stage), to represent the fact that the browser loaded them before
- The new segments are either in Static or Runtime stage, depending on how each segment would have been prefetched. 

Similar to static shell validation, we then perform a Fizz prerender on the combined payload, abort it, and track the locations of `onError` calls. The difference here is that unlike static shell validation, we require a Suspense boundary *inside* the new subtree. This is detected by adding a `InstantValidationBoundary` around the new subtree when we construct the payload, and checking if there's a suspense below that in the component stack.

Currently, for each `instant` config we find, we take each of its parent layouts as a possible navigation parent (unless it's marked as blocking). So if we have a page with parent layouts like this:
```
/layout.tsx
/foo/layout.tsx
/foo/bar/layout.tsx
/foo/bar/page.tsx <- `export const instant = { prefetch: 'static' }`
```
Then we'll check for navigations where each of `/layout.tsx`, `/foo/layout.tsx` and `/foo/bar/layout.tsx` is a fully resolved shared parent.

If a validation fails, we do a "discriminated error message" flow analogous to static shell validation. Static segments are replaced with Runtime segments to see if the error goes away, which lets us determine if the hole is caused by runtime or dynamic data.

#### Building the combined payload

[...]

PR #88556, opened 2026-01-14
fix: preserve cache behavior for PPR fallback shells with root params

ztanner:
When generating PPR fallback shells, a cache miss during warmup was being treated as a dynamic hole, so the warmup render never filled new cache entries. This caused root‑param fallback shells (e.g. `/en/blog/[slug]`) to suspend even though root params were already known.

This change always creates a fresh prerender RDC for warmup and copies any seed hits into it. The final render then uses the warmed cache, so fallback shells can reuse populated entries without short‑circuiting. On the export side, we now choose a seed RDC by matching known params (e.g. locale) so /en/... fallback shells don’t get seeded with /fr/... data. (Previously, the RDC seed was "last-write-wins". 

Fixes NAR-716

PR #94194, opened 2026-05-28
Consolidate Resume Data Caches and partially revert #88556

unstubbable:
This consolidates the two Resume Data Caches into a single discriminated union and reverts three non-load-bearing changes from #88556. The result sets up a follow-up that fixes a bug where pages exporting `unstable_instant = false` cause their `'use cache'` reads to be incorrectly treated as dynamic.

The cache wrapper short-circuits a `'use cache'` miss to a dynamic hole when prerendering a fallback shell that was seeded with a read-only RDC from an earlier phase. The reasoning is that the seed was filled by a phase-1 prerender of a more-specific sibling route, so a miss in phase-2 means the cache key depends on a fallback param. In PR #79743, we set up an XOR on the work unit store that directly encoded this distinction: a prerender carried either a mutable `PrerenderResumeDataCache` (a phase-1 prerender filling caches as it runs) or an immutable `RenderResumeDataCache` (a phase-2 fallback shell reading from a phase-1-seeded cache), never both. The bail-out didn't use that signal directly, though; it was gated on `workUnitStore.allowEmptyStaticShell`, which at the time happened to coincide with "we have a read-only seed" because the fallback-shell flow was the only thing setting it. #89077 later widened that flag to also cover `unstable_instant = false`, which has no seed, so the bail-out started firing for routes that should be allowed to fill their caches normally. #88556 broke the XOR by carrying both caches at the same time. In a follow-up, the flag check is replaced with a precise `workUnitStore.resumeDataCache?.mutable === false`, which only lands cleanly once the XOR is back.

The three reverts: (1) at the start of the Cache Components branch of `prerenderToStream`, phase-2 no longer creates a fresh mutable RDC alongside the prefilled one — it just uses the seed; (2) the swap between the prospective and final prerender passes that replaced the prefilled RDC with a freeze of the (now-gone) mutable one is removed, so the final pass reads from the same seed as the prospective pass; (3) the line in the cache wrapper that mirrored prefilled-RDC hits into the mutable RDC (only useful to feed the swap) is removed. The actual fix from #88556 — selecting the seed RDC by matching known params, `buildRDCCacheByPage` in `export/index.ts` — is kept, and the `ppr-root-param-fallback` test suite that was added in that PR still passes.

On top of the revert, the two RDC types now carry a `mutable: true | false` discriminator, the store fields collapse into one `resumeDataCache: ResumeDataCache | null`, and the specialized `getPrerenderResumeDataCache` / `getRenderResumeDataCache` helpers are replaced by a single `getResumeDataCache` returning the union. Call sites narrow via `resumeDataCache?.mutable` directly, and the typing-lie in the old `getRenderResumeDataCache` (coercing a mutable cache to read-only via spread) is gone. Restoring the XOR as two separate fields would have worked, but the discriminated union makes the invariant compiler-enforced and removes a typing-lie that the original design already carried — and the revert touches the same call sites anyway, so bundling the refactor in one pass was cleaner than splitting it.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -229,6 +229,7 @@ import {
   createRenderResumeDataCache,
   type PrerenderResumeDataCache,
   type RenderResumeDataCache,
+  type ResumeDataCache,
 } from '../resume-data-cache/resume-data-cache'
 import type { MetadataErrorType } from '../../lib/metadata/resolve-metadata'
 import isError from '../../lib/is-error'
@@ -969,7 +970,7 @@ async function generateStagedDynamicFlightRenderResultWeb(
   if (hasRuntimePrefetch) {
     // Create a mutable cache that gets filled during the dynamic render.
     const prerenderResumeDataCache = createPrerenderResumeDataCache()
-    requestStore.prerenderResumeDataCache = prerenderResumeDataCache
+    requestStore.resumeDataCache = prerenderResumeDataCache
 
     const cacheSignal = new CacheSignal()
     trackPendingModules(cacheSignal)
@@ -1128,7 +1129,7 @@ async function generateStagedDynamicFlightRenderResultNode(
   if (hasRuntimePrefetch) {
     // Create a mutable cache that gets filled during the dynamic render.
     const prerenderResumeDataCache = createPrerenderResumeDataCache()
-    requestStore.prerenderResumeDataCache = prerenderResumeDataCache
+    requestStore.resumeDataCache = prerenderResumeDataCache
 
     const cacheSignal = new CacheSignal()
     trackPendingModules(cacheSignal)
@@ -1229,7 +1230,6 @@ async function spawnRuntimePrefetchWithFilledCaches(
       ctx,
       generateDynamicRSCPayload.bind(null, ctx, { staleTimeIterable }),
       prerenderResumeDataCache,
-      null, // renderResumeDataCache
       rootParams,
       requestStore.headers,
       requestStore.cookies,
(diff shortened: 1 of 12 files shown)


packages/next/src/server/app-render/app-render.tsx
```diff
@@ -7399,14 +7379,15 @@ async function prerenderToStream(
       // to cut the render off.
       const cacheSignal = new CacheSignal()
 
-      // Always start with a fresh prerender RDC so warmup can fill misses,
-      // even when we have a prefilled render RDC to seed from.
-      const prerenderResumeDataCache = createPrerenderResumeDataCache()
```
unstubbable, line 7404: This is revert 1.

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -7681,13 +7659,6 @@ async function prerenderToStream(
         initialClientReactController.abort()
       }
 
-      if (renderOpts.renderResumeDataCache) {
-        // Swap to the warmed cache so the final render uses entries produced during warmup.
-        renderResumeDataCache = createRenderResumeDataCache(
-          prerenderResumeDataCache
-        )
-      }
```
unstubbable, line 7689: This is revert 2.

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -2209,10 +2206,6 @@ export async function cache(
       if (rdcResult !== undefined) {
         debug?.('Resume Data Cache entry found', serializedCacheKey)
 
-        if (prerenderResumeDataCache) {
-          prerenderResumeDataCache.cache.set(serializedCacheKey, rdcEntry)
-        }
```
unstubbable, line 2214: This is revert 3.

gnoff (approved):

PR #94201, opened 2026-05-28
[turbopack] Shrink ReadRawVcFuture from 80 to 64 bytes

lukesandberg:
Smaller is better

Probably very minor but these are hot futures so shrinking must be useful.  This optimization caries over to the ResolveVcFuture and ReadVcFuture as well.

PR #94202, opened 2026-05-28
[turbopack] Optimize TaskInput implementations

lukesandberg:
# What

Replace the `derive` macros for TaskInput with a proc macro `[turbo_tasks::task_input]` (and analogous support inside of the `turbo_tasks::value` macro.

Introduce a hand written `CloneResolved` future type for TaskInput implementations that are always resolved.

# Why

This allows us to take advantage of the `NonLocalValue` marker trait.  If a type is `NonLocalValue` then `TaskInput` can have trivial `is_resolved` and `resolve_input` implementations.  This applies to the vast majority of TaskInputs but the current `derive` implementation always produces a recursive `async` `resolve_input` implementation but it is rarely needed and rustc/llvm do a bad job of optimizing it.

## Code-size impact

`next-swc.darwin-arm64.node` — release profile, `pnpm swc-build-native --release`, macOS aarch64.

| Metric | canary | this branch | Δ bytes | Δ % |
| Raw `.node` | 124,494,592 | 124,217,280 | −277,312 | −0.22% |
| Stripped (`strip -x`) | 84,294,000 | 84,013,136 | −280,864 | −0.33% |
| Stripped + gzip -9 | 29,123,538 | 28,898,920 | −224,618 | −0.77% |

## Performance

A similar measurement of performance was neutral

| Metric | canary | temporary | Δ | % |
|---|---|---|---|---|
| wall time (s) | 42.659 ± 1.851 | 42.501 ± 2.028 | −0.158 | −0.4% |
| user time (s) | 294.322 ± 1.834 | 293.207 ± 2.070 | −1.115 | −0.4% |
| sys time (s) | 74.158 ± 4.197 | 73.543 ± 5.725 | −0.615 | −0.8% |
| MaxRSS (MB) | 13650.1 ± 66.0 | 13610.1 ± 57.5 | −40.0 | −0.3% |

All deltas are smaller than one standard deviation — statistically indistinguishable from noise on this benchmark, as expected for a small optimization.


lukesandberg:
* **#94215** <a href="https://app.graphite.com/github/pr/vercel/next.js/94215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94213** <a href="https://app.graphite.com/github/pr/vercel/next.js/94213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94202** <a href="https://app.graphite.com/github/pr/vercel/next.js/94202?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94202?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94201** <a href="https://app.graphite.com/github/pr/vercel/next.js/94201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

scripts/bench-build.sh
lukesandberg, line 1: drop this file

scripts/bench-quiet.sh
lukesandberg, line 1: drop this

turbopack/crates/turbo-tasks/src/lib.rs
```diff
@@ -391,10 +416,6 @@ pub use turbo_tasks_macros::value_impl;
 #[rustfmt::skip]
 pub use turbo_tasks_macros::task_storage;
```
lukesandberg, line 422: why is this rexported here?

turbopack/crates/turbopack-core/src/chunk/availability_info.rs
```diff
@@ -2,20 +2,20 @@ use anyhow::Result;
...
 use turbo_rcstr::RcStr;
-use turbo_tasks::{NonLocalValue, ResolvedVc, TaskInput, Vc, trace::TraceRawVcs};
+use turbo_tasks::{ResolvedVc, Vc, trace::TraceRawVcs};
 
 use crate::chunk::available_modules::{AvailableModules, AvailableModulesSet};
 
 bitfield! {
-    #[derive(Clone, Copy, Default, TaskInput, TraceRawVcs, NonLocalValue, PartialEq, Eq, Hash, Encode, Decode)]
+    #[turbo_tasks::task_input]
+#[derive(Clone, Copy, Default, TraceRawVcs, PartialEq, Eq, Hash, Encode, Decode)]
```
lukesandberg: formatting

turbopack/crates/turbo-tasks/src/lib.rs
```diff
@@ -269,6 +269,31 @@ pub use turbo_tasks_macros::function;
 #[rustfmt::skip]
 pub use turbo_tasks_macros::value;
 
+/// Attribute macro for declaring a [`TaskInput`] type. Emits:
+///
+/// - `unsafe impl NonLocalValue for X {}` (unless `contains_unresolved_vcs` is set).
```
lukesandberg, line 279: we should emit a derive for this

turbopack/crates/turbo-tasks-macros/src/expand.rs
```diff
@@ -1,10 +1,88 @@
...
+            generics: generics.clone(),
+            data: Data::Enum(DataEnum {
+                enum_token: Default::default(),
+                brace_token: Default::default(),
+                variants: variants.clone(),
+            }),
+        }),
+        _ => None,
+    }
+}
```
bgw: I think we could modify `match_expansion` to accept a `Data` or some other simpler data structure rather than contorting an `Item` into a `DeriveInput`.

The only things that `match_expansion` uses are `data` and the `ident`, and it's not even clear that there's much value to it reading the `ident`.

turbopack/crates/turbo-tasks-macros/src/task_input_attr_macro.rs
```diff
@@ -0,0 +1,292 @@
...
+        );
+        quote! {
+            #[automatically_derived]
+            unsafe impl #impl_generics turbo_tasks::NonLocalValue for #ident #ty_generics
+            where
+                #(#existing_predicates,)*
+                #(#generic_type_params: turbo_tasks::NonLocalValue,)*
+            {}
+            #assertions
+        }
```
bgw: Is there some easy way we can share code with `derive_non_local_value`?

turbopack/crates/turbo-tasks-macros/src/task_input_attr_macro.rs
```diff
@@ -0,0 +1,292 @@
...
+    let non_local_value_impl = if contains_unresolved_vcs.is_some() {
+        quote! {}
+    } else {
+        let mut assertion_generics = generics.clone();
+        let where_clause = assertion_generics.make_where_clause();
+        for param in &generic_type_params {
+            where_clause
+                .predicates
+                .push(parse_quote!(#param: turbo_tasks::NonLocalValue));
+        }
```
bgw: Is there some reason this can't use `split_for_impl` as `derive_non_local_value` does?

turbopack/crates/turbo-tasks/src/raw_vc.rs
```diff
@@ -408,37 +408,60 @@ impl Unpin for ResolveRawVcFuture {}
...
+    /// because it is configured by the builder before phase 1 starts.
     read_cell_options: ReadCellOptions,
-    /// Phase 2: the resolved task and cell identity, set when phase 1 completes.
-    resolved: Option<(TaskId, CellId)>,
-    /// Phase 2: listener for the cell read wait.
-    listener: Option<EventListener>,
+    /// Whether phase 1 was a strongly-consistent read. Needed in phase 2 to re-apply
+    /// [`suppress_top_level_task_check`]. Stored here so we can drop the [`ResolveRawVcFuture`]
+    /// once phase 1 completes.
+    strongly_consistent: bool,
```
bgw: - These two fields could be stored inside `ReadRawVcState::Reading`, and that would save space assuming `Reading` is currently smaller than `Resolving`.
- Both `ResolveRawVcFuture` and `ReadOutputOptions` contain consistency information, there might be some further options for cleaning this up if you're interested.

turbopack/crates/turbo-tasks/src/task/task_input.rs
```diff
@@ -28,6 +30,29 @@ use crate::{
...
+/// An 8-byte hand-rolled [`Future`] that immediately resolves to `Ok(self.clone())` of the
+/// referenced value.
+///
+/// Used by the [`TaskInput::resolve_input`] default implementation
+struct CloneReady<'a, T> {
+    pub inner: Option<&'a T>,
+}
+
+impl<'a, T: Clone> Future for CloneReady<'a, T> {
+    type Output = Result<T>;
```
bgw, line 42: This would be a good usecase for `Infallible`: https://doc.rust-lang.org/std/convert/enum.Infallible.html

bgw (approved):

turbopack/crates/turbo-tasks-macros/src/expand.rs
lukesandberg: :robot: Done. Dropped the synthetic-`DeriveInput` contortion: `match_expansion` now takes `(ident: &Ident, data: &Data)` directly, and `item_to_derive_input` is replaced by a small `item_data(&Item) -> Option<Data>` helper that just extracts the field shape. All callers (trace/hash/value_debug derives, the value macro, and the task_input attr macro) updated to pass `ident` + `data`.

turbopack/crates/turbo-tasks-macros/src/task_input_attr_macro.rs
lukesandberg: :robot: Done. Extracted a shared `non_local_value_impl(ident, generics, data, add_bounds)` in `non_local_value_macro.rs`. `#[derive(NonLocalValue)]` now calls it with `add_bounds = false` (relies on the type's existing where-clause), and the `task_input` attr macro calls it with `add_bounds = true` (appends a `T: NonLocalValue` predicate per generic type param so the `unsafe impl` is sound and the field assertions see the bound). Single code path now.

turbopack/crates/turbo-tasks-macros/src/task_input_attr_macro.rs
lukesandberg: :robot: Done — it does now. The shared `non_local_value_impl` helper clones the generics, appends the `T: NonLocalValue` predicates into the where-clause, then uses `split_for_impl` (same as `derive_non_local_value`). The previous hand-rolled `#(#existing_predicates,)* #(#generic_type_params: ...,)*` interpolation is gone.

turbopack/crates/turbo-tasks/src/raw_vc.rs
lukesandberg: :robot: Done — moved both into the state enum, `ReadRawVcFuture` shrinks 64 → 56 bytes. `read_cell_options` now lives in `Resolving { resolve, read_cell_options }` (builder writes it there) and is moved into `Reading` at the phase-1→2 transition. `strongly_consistent` was redundant with the state already on `ResolveRawVcFuture` (your second bullet), so I dropped the outer copy and read it off the resolve future at the transition, storing it only in `Reading`. The outer struct is now just `{ state }`.

Measured: `Reading` 24B < `Resolving` 32B before; after the move `Reading` is 48B and `Resolving` 56B (it carries the 24B `read_cell_options`), so the enum is 56B and the future is 56B vs 64B. Couldn't get to the full 48B without reconstructing `ReadCellOptions` from `ReadOutputOptions` at the transition, which loses `track_with_key`'s key — left that alone.

turbopack/crates/turbo-tasks/src/task/task_input.rs
lukesandberg: :robot: I looked at this but don't think `Infallible` fits cleanly here, and wanted to check with you before changing it. The trait fixes `fn resolve_input(&self) -> impl Future<Output = anyhow::Result<Self>>`, and the `contains_unresolved_vcs` impl genuinely fails (it does `resolve_input(field).await?` per field). Since the `Output` is shared across all impls, the default `CloneReady` future has to also yield `anyhow::Result<Self>` — I can't give it `Output = Result<T, Infallible>` or `Output = T` without an adapter future that wraps it back into the `anyhow::Result`, which would defeat the 8-byte point (the whole reason it's hand-rolled rather than `std::future::ready`, which would hold a full `T`).

So I think `Infallible` would only apply if we split the trait into infallible-default vs fallible-resolve methods, which seems heavier than it's worth. Happy to do that or apply `Infallible` some way I'm missing — let me know. Leaving this thread open for now.

turbopack/crates/turbo-tasks/src/raw_vc.rs
lukesandberg: :robot: Correction to my earlier reply — went with the simpler layout you suggested. `read_cell_options` stays on the wrapper struct (builder methods write it directly, no per-variant matching), and only `strongly_consistent` moves into `Reading`. Same 64 → 56 bytes.

The win is entirely the bool, not alignment and not the cell options: dropping the bool from the wrapper leaves `read_cell_options`(24) + `state`(32) = 56 with no padding waste. `Reading` with the bool is 24B, still ≤ `Resolving`'s 32B, so the enum stays 32. Pushing `read_cell_options` into the enum (what I did first) also lands at 56 but bloats `Resolving` to carry 24B it never reads — so the wrapper layout is the right one. `strongly_consistent` is read off the resolve future at the phase transition (it was already stored there).

lukesandberg: because ReadCellOptions needs to be mutated during Resolve it should stay on the future directly but the boolean can move

lukesandberg: saves 8 bytes

PR #94208, opened 2026-05-28
[turbopack] Separate `JsValue` into it's own folder in the analyzer

sampoder:
This is the first in a stack of PRs to break up the analyzer and, eventually, move it to its own crate. There shouldn't be any code changes - everything was just copy and pasted into separate files (let me know if this split makes sense).


turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/constants.rs
```diff
@@ -0,0 +1,334 @@
+#![allow(clippy::redundant_closure_call)]
+
+use super::*;
```
bgw: - Don't use wildcard imports.
- Don't import from `super` unless you've got an inline module (e.g. `mod tests { ... }`).
- Use explicit `use crate::foo::Bar` imports

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/constants.rs
```diff
@@ -0,0 +1,334 @@
+#![allow(clippy::redundant_closure_call)]
```
bgw: Can we remove this and re-add it on the specific members that need it?

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/explain.rs
```diff
@@ -0,0 +1,662 @@
...
+                        } else {
+                            ""
+                        }
+                    );
+                    format!("???*{i}*")
+                }
+            }
+            JsValue::WellKnownObject(obj) => {
+                let (name, explainer) = match obj {
+                    WellKnownObjectKind::Generator => (
```
bgw, line 399: Maybe (in a separate PR) `WellKnownObjectKind` and `WellKnownFunctionKind` should have their own `explain` methods that this calls?

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/mod.rs
```diff
@@ -0,0 +1,1226 @@
...
+use turbopack_core::compile_time_info::{
+    CompileTimeDefineValue, DefinableNameSegmentRef, DefinableNameSegmentRefs, FreeVarReference,
+    TotalOrderF64,
+};
+
+use super::{
+    WellKnownFunctionKind, WellKnownObjectKind,
+    graph::{EvalContext, VarGraph},
+    imports::ImportAnnotations,
+};
```
bgw: Let's write these as crate imports.

The problem is that the `super` imports make grepping much harder.

bgw (changes requested):
Other than the minor comments about the imports and the clippy lint, LGTM

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/constants.rs
sampoder: Sounds good, fixed!

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/explain.rs
sampoder: Added on to this stack: #94217

bgw (approved):

PR #94213, opened 2026-05-28
make `TaskInput::is_resolved` inlinable everywhere

lukesandberg:
## What

Removes the `ArgMeta::is_resolved` function-pointer field and the 592 monomorphized `closure#3` trampolines it generated. The `TaskInput::is_resolved` check that runs on the synchronous fast path of every turbo-tasks function call now happens at the macro-generated callsite on the concrete tuple type, before erasure into `&dyn StackDynTaskInputs`. For trait calls with `filter_owned`, the check is fused into the filter functor so the filter and the resolved-check monomorphize together.

## Why

The `is_resolved` method is almost always trivial after optimization but putting it behind an opaque function pointer means we always pay a small code size and perf cost.  Inlining actually saves both.

Bonus rename: `OwnedStackDynTaskInputs` → `HeapDynTaskInputsStorage`. The \"Stack\" was the trait, not the storage — the inner value is a `Box<dyn DynTaskInputs>` on the heap.

## Hot-path microcheck

The synchronous wrapper for a representative 6-input turbo function (`directory_tree_to_loader_tree`) now performs the entire `is_resolved` check as ~15 register-only ARM instructions inlined into the caller — three `cmp/csel/tbnz` chains testing the discriminant byte of each `Vc<...>` input. The four trivially-resolved inputs (`FileSystemPath`, `RcStr`, `AppPage`, `AppPath`) contribute zero instructions because their `is_resolved` constant-folded to `true`. No indirect call, no downcast trampoline.

## Code-size impact

Release `next-swc.darwin-arm64.node`, macOS aarch64, vs the `temporary` branch this PR is stacked on:

| Metric | prior (temporary) | this PR | Δ bytes | Δ % |
| Raw `.node` | 124,217,280 | 123,845,712 | −371,568 | −0.30% |
| Stripped (`strip -x`) | 84,013,136 | 83,880,384 | −132,752 | −0.16% |
| Stripped + gzip -9 | 28,898,920 | 28,887,624 | −11,296 | −0.04% |


lukesandberg:
* **#94215** <a href="https://app.graphite.com/github/pr/vercel/next.js/94215?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94213** <a href="https://app.graphite.com/github/pr/vercel/next.js/94213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94213?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94202** <a href="https://app.graphite.com/github/pr/vercel/next.js/94202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94201** <a href="https://app.graphite.com/github/pr/vercel/next.js/94201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/dyn_task_inputs.rs
```diff
@@ -101,31 +101,35 @@ impl<T: DynTaskInputs> StackDynTaskInputs for StackDynTaskInputsSlot<T> {
 }
 
 /// Adapter for an already-boxed value (e.g., from async resolution tasks).
-pub struct OwnedStackDynTaskInputs {
+///
+/// Implements the [`StackDynTaskInputs`] trait over a heap-allocated `Box<dyn DynTaskInputs>`.
+/// The "Stack" in the trait name refers to the trait itself, not where the inputs live; this
+/// type's payload is on the heap.
```
bgw: Is there a better name we can give the `StackDynTaskInputs` trait then?

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -810,10 +819,25 @@ impl<B: Backend + 'static> TurboTasks<B> {
             match registry::get_value_type(type_id).get_trait_method(trait_method) {
                 Some(native_fn) => {
                     if let Some(filter) = native_fn.arg_meta.filter_owned {
-                        let mut arg = (filter)(arg);
-                        return self.dynamic_call(native_fn, Some(this), &mut arg, persistence);
+                        // TODO: filter should recompute task persistence as well? maybe minor
+                        // issue.  It is also unfortunate that our caller computed resolved and then
+                        // we recompute it when filtering, but the cost is quite cheap.
```
bgw: I don't think argument filtering should change persistence.

The task persistence behavior is confusing enough as-is, but right now the only determination of persistence happens at the callsite.

If we recomputed persistence after filtering, persistence would be decided by a combination of the callsite and the function implementation, which seems even more confusing? Changing the implementation of a function could affect if certain callers get a persistent version or not.

Generally speaking, argument filtering shouldn't cause any (or at least cause only minimal) eternally-visible changes to how the function works, unless you're performing `Vc` equality, internal mutability, or comparing task IDs, which is always a little sketchy (and implies "I know what I'm doing").

turbopack/crates/turbo-tasks/src/native_function.rs
```diff
@@ -21,11 +21,13 @@ use crate::{
...
-type IsResolvedFunctor = fn(&dyn DynTaskInputs) -> bool;
-
+/// Filters out arguments that aren't used by the function body, returning the post-filter args
+/// together with the precomputed `TaskInput::is_resolved` answer for those filtered args. Fusing
+/// the two operations into one functor lets LLVM monomorphize and constant-fold them together,
+/// avoiding a second indirect call.
 #[doc(hidden)]
 pub type FilterOwnedArgsFunctor =
-    for<'a> fn(&'a mut dyn StackDynTaskInputs) -> OwnedStackDynTaskInputs;
+    for<'a> fn(&'a mut dyn StackDynTaskInputs) -> (bool, BoxedDynTaskInputs);
```
bgw: This would be a good time for a custom enum instead of a bool.

bgw (approved):

turbopack/crates/turbo-tasks/src/dyn_task_inputs.rs
lukesandberg: done,  but i got confused and did it up-stack 🤦

PR #94214, opened 2026-05-29
[ci] Set `persist-credentials: false` for all GH actions

bgw:
The actual security implications of this are likely *very* minor, but in the interest of reducing how many credentials we have, there's little reason for us not to set this to `false`.

In cases where we create a new commit via GH actions, we already do that via the GitHub API, so those don't need the on-disk git credentials to push.

---

From: https://github.com/actions/checkout :

> The auth token is persisted in the local git config. This enables your scripts to run authenticated git commands. The token is removed during post-job cleanup. Set persist-credentials: false to opt-out.

```
    # Whether to configure the token or SSH key with the local git config
    # Default: true
    persist-credentials: ''
```

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -50,6 +50,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 1
+          persist-credentials: false
       - run: echo "${{ github.event.after }}"
       - name: Setup node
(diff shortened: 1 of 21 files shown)


bgw:
* **#94214** <a href="https://app.graphite.com/github/pr/vercel/next.js/94214?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94214?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

lukesandberg (approved):

PR #94216, opened 2026-05-29
fix: Avoid Set in missing slots RSC payload

### LINKED ISSUE #73323: `Only plain objects can be passed from Client Components to Server Components` on route handler and while using the pinata-sdk

https://github.com/tjwelde/next15-pinata-issue

### To Reproduce

1. `npm install`
2. `npm run dev`
3. open page (e.g. http://localhost:3000/)
4. everything still normal
5. reload page
6. see error in server logs

Expected:
a. No error to be thrown
b. If there is an actual issue with the library used: since the error seems to stem from a route handler, I would've expected the error to not mention client and server components and/or being more precise

Current:
Following error is logged:
```
Error: [ Server ] Only plain objects can be passed to Client Components from Server Components. Set objects are not supported.
  {P: </>, b: ..., p: "", c: ["", ""], i: ..., f: ..., m: Set, G: ..., s: ..., S: ...}
                                                          ^^^
    at createUnhandledError (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/client/components/react-dev-overlay/internal/helpers/console-error.js:27:49)
    at handleClientError (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/client/components/react-dev-overlay/internal/helpers/use-error-handler.js:44:56)
    at console.error (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/client/components/globals/intercept-console-error.js:48:56)
    at react-stack-bottom-frame (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.browser.development.js:2446:58)
    at resolveConsoleEntry (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.browser.development.js:1961:9)
    at processFullStringRow (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.browser.development.js:2095:11)
    at processFullBinaryRow (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.browser.development.js:2059:7)
    at progress (webpack-internal:///(app-pages-browser)/./node_modules/next/dist/compiled/react-server-dom-webpack/cjs/react-server-dom-webpack-client.browser.development.js:2262:17)
    ```

```bash
Node.js v20.15.0

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.1.0: Thu Oct 10 21:03:11 PDT 2024; root:xnu-11215.41.3~2/RELEASE_ARM64_T6020
  Available memory (MB): 16384
  Available CPU cores: 12
Binaries:
  Node: 20.15.0
  npm: 10.7.0
  Yarn: 4.2.2
  pnpm: N/A
Relevant Packages:
  next: 15.0.4-canary.30 // Latest available version is detected (15.0.4-canary.30).
  eslint-config-next: N/A
  react: 19.0.0-rc-b01722d5-20241114
  react-dom: 19.0.0-rc-b01722d5-20241114
  typescript: 5.3.3
Next.js Config:
  output: N/A
```

Not sure

next dev (local)

### Additional context

I tested my reproduction against multiple versions up to 14.0.0 

14.1.0 still shows the error with a bit more details:
```
Warning: Only plain objects can be passed to Client Components from Server Components. Set objects are not supported.
  <... buildId=... assetPrefix="" urlParts=... initialTree=... initialSeedData=... couldBeIntercepted=... initialHead=... globalErrorComponent=... missingSlots={Set}>
                                                                                                                                                                ^^^^^
```

14.0.0 does not show the error

as background:
- I am trying to upgrade an existing project from next@13.5.7 to next@15.x
- the `@pinata/sdk` package is deprecated and we will switch to the new `pinata-web3` package, which does not seem to have this issue in my early tests. 
- Still, I think other libraries might also trigger an error like this, so I think it is worth investigating, what is actually going on
- My main issue here is, that it took a long time to debug, because the error said it would have something to do with the client/server component boundary, so I was looking at the wrong things for quite some time.
  - Having a more precise error message would already help immensely
- Still not sure, if there is an actual error happening somewhere. Wrapping the affected call in a try..catch did not catch anything though

hukuda222:
### What?

Avoid sending the dev-only `missingSlots` value as a `Set` in the initial RSC payload. The payload now carries missing slots as a string array only when there are actually missing slots, and the client reconstructs the `Set` before providing `MissingSlotContext`.

### Why?

In development, pages without missing parallel route slots can still include an empty `Set` at `initialRSCPayload.m`, which triggers this warning:

```txt
Only plain objects can be passed to Client Components from Server Components. Set objects are not supported.
```

This warning points users at application Server/Client Component props even though the `Set` comes from Next.js internals.

### How?

- Serialize `missingSlots` as `string[] | undefined` in `InitialRSCPayload`.
- Omit `m` when the set is empty.
- Recreate `Set<string>` on the client only when missing slots are present.
- Add a regression assertion to the existing app dir RSC basics test.

Fixes #73323

### Testing

- `pnpm install`
- `pnpm build-all`
- `pnpm --filter=next types`
- `pnpm exec prettier --check packages/next/src/client/app-index.tsx packages/next/src/server/app-render/app-render.tsx packages/next/src/shared/lib/app-router-types.ts test/e2e/app-dir/rsc-basic/rsc-basic.test.ts`
- `pnpm lint-eslint packages/next/src/client/app-index.tsx packages/next/src/server/app-render/app-render.tsx packages/next/src/shared/lib/app-router-types.ts test/e2e/app-dir/rsc-basic/rsc-basic.test.ts`
- `pnpm test-dev-turbo test/e2e/app-dir/rsc-basic/rsc-basic.test.ts`


hukuda222:
Hi @acdlite, @unstubbable — could one of you take a quick look at this or point me to the right reviewer?

This is a small dev-only serialization fix for `missingSlots` in the initial RSC payload. It fixes #73323 by avoiding sending a `Set` through the RSC payload, omitting `m` when there are no missing slots, and reconstructing the `Set` on the client only when needed.

I've rebased onto `canary` today and verified:

- `pnpm install`
- `pnpm build-all`
- `pnpm --filter=next types`
- `pnpm exec prettier --check ...`
- `pnpm lint-eslint ...`
- `pnpm test-dev-turbo test/e2e/app-dir/rsc-basic/rsc-basic.test.ts`

Happy to adjust the shape if you'd prefer a different payload representation or test coverage.

unstubbable:
This is not the correct fix. Sets are supported in the RSC payload. The error is a false positive, and possibly a bug in React. I'll take a closer look.

PR #94197, opened 2026-05-28
Trace cacheHandler and cacheHandlers when using adapters

timneutkens:
### What?

Add focused production adapter coverage for configured `cacheHandler` and `cacheHandlers` module graphs in Node `output.assets`, including transitive helper dependencies and Turbopack `assetsHashes` assertions.

### Why?

Adapter deployments dynamically load configured cache handlers at runtime. This coverage demonstrates that Node adapter outputs currently omit those handler files and their transitive dependencies.

### How?

Add an App Router Cache Components fixture using `adapterPath`, both cache-handler APIs, and a `use cache: remote` route. Its adapter writes `onBuildComplete` context to disk and the test inspects Node outputs for each handler and uniquely named helper.

### Verification

- `NEXT_TEST_PREFER_OFFLINE=1 pnpm --filter=next build`
- Expected failure: `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-turbo test/production/app-dir/adapter-cache-handlers/adapter-cache-handlers.test.ts` (no Node output contains all four handler assets)
- Expected failure: `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-webpack test/production/app-dir/adapter-cache-handlers/adapter-cache-handlers.test.ts` (same adapter handoff gap)
- `git diff canary --check`
- Not run continuously: `pnpm --filter=next dev` (`EMFILE: too many open files` occurred after startup in the local watcher)

PR #93363, opened 2026-04-29
Turbopack: fix bundle-analyzer for NFT change

mischnic:
We have no testing infrastructure, unfortunately

Bildschirmfoto 2026-05-28 um 20 29 58

PR #94224, opened 2026-05-29
Turbopack: refactor NFT to add Endpoint.traced_files

mischnic:
Refactor the module-graph bit of NFT so that we can provide `Endpoint.traced_files()` which is needed for the bundle analyzer

There should be no logic change whatsoever


mischnic:
* **#94224** <a href="https://app.graphite.com/github/pr/vercel/next.js/94224?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 4  dependent PRs ([#93363](https://github.com/vercel/next.js/pull/93363) <a href="https://app.graphite.com/github/pr/vercel/next.js/93363?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#94197](https://github.com/vercel/next.js/pull/94197) <a href="https://app.graphite.com/github/pr/vercel/next.js/94197?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#94207](https://github.com/vercel/next.js/pull/94207) <a href="https://app.graphite.com/github/pr/vercel/next.js/94207?utm_source=stack-comment-icon" target="_blank">Graphite</a> and 1 other) 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94224?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,672 @@
...
+                )
+                .collect(),
+        ))
+    }
+}
+
+#[turbo_tasks::function]
+pub async fn trace_endpoint(
+    project: ResolvedVc<Project>,
+    page_name: Option<RcStr>,
```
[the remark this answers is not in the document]
mischnic: tracing_exclude_glob does this now

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+                        use turbo_tasks_fs::FileSystemEntryType;
+
+                        if current_path.is_root() {
+                            break;
+                        }
+
+                        if matches!(
+                            &*current_path.get_type().await?,
+                            FileSystemEntryType::Symlink
+                        ) {
```
sokra: Could this be `referenced_chunk_path != referenced_chunk_path.realpath()`?

mischnic: No, a `referenced_chunk_path` itself could be a symlink (e.g. `node_modules/react -> node_modules/.pnpm/react@18`), which is valid.

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+
+            // Process includes
+            if let Some(includes_config) = output_file_tracing_includes
+                && let Some(includes_obj) = includes_config.as_object()
+            {
+                for (glob_pattern, include_patterns) in includes_obj {
+                    // Check if the route matches the glob pattern
+                    let glob =
+                        Glob::new(glob_pattern.as_str().into(), GlobOptions { contains: true })
+                            .await?;
```
sokra: This could be moved to NextConfig::output_file_tracing_includes() so we only create the glob once and not per route. That would save the N x M turbo-tasks function lookups.

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+                    if glob.matches(route)
+                        && let Some(patterns) = include_patterns.as_array()
+                    {
+                        for pattern in patterns {
+                            if let Some(pattern_str) = pattern.as_str() {
+                                let (glob, root) =
+                                    relativize_glob(pattern_str, project_path.clone())?;
```
sokra: Could also be precomputed in NextConfig::output_file_tracing_includes()

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+    }
+}
+impl PartialOrd for SortableFileSystemPath {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// Apply outputFileTracingIncludes patterns to find additional files
+async fn apply_includes(
```
sokra: The method name sounds like it's mutating something, but it actually returns a list.

apply_includes -> glob_includes?

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+
+            result.insert(SortableFileSystemPath(file_path.clone()));
+        }
+
+        for nested_result in glob_result.inner.values() {
+            let nested_result_ref = nested_result.await?;
+            stack.push_back(nested_result_ref);
+        }
+    }
+    Ok(result)
```
sokra, line 203: I think it's simpler and more efficient if you get rid of the BTreeSet and use a Vec instead. Just call `sort_by_key` here. That should be faster than maintaining a set.

A glob shouldn't result in duplicates anyway.

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+                    let glob = Glob::new(
+                        format!("{{{}}}", globs.join(",")).into(),
+                        GlobOptions { contains: true },
+                    );
+                    apply_includes(root, glob)
+                })
+                .try_join()
+                .await?;
+
+            includes.into_iter().flatten().map(|path| path.0).collect()
```
sokra: This potentially results in duplicates. When do we deduplicate the includes?
Do we want to sort and deduplicate here?

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+
+#[turbo_tasks::function]
+pub async fn tracing_exclude_glob(
+    page_name: Option<RcStr>,
+    project_path: FileSystemPath,
+    next_config: ResolvedVc<NextConfig>,
+) -> Result<Vc<OptionGlob>> {
+    Ok(if let Some(page_name) = &page_name {
+        let route = format!("/{page_name}");
+        let output_file_tracing_excludes = next_config.output_file_tracing_excludes().await?;
```
sokra: Same. output_file_tracing_excludes could precompute the globs and relativize_glob

crates/next-api/src/nft.rs
```diff
@@ -0,0 +1,673 @@
...
+        Some(exclude_glob.await?)
+    } else {
+        None
+    };
+    let module_idents = if exclude_glob.is_some() {
+        let data = traced_module_data_for_graph(module_graph, entries_are_traced).await?;
+        Some(data.idents.await?)
+    } else {
+        None
+    };
```
sokra: You could combine it into a `Option<(exclude_glob, module_idents)>` to avoid the unwrap() later

crates/next-api/src/nft.rs
mischnic: - EndpointTraceResult might contain duplicates
- That then gets read in NftJsonAsset which will dedupe everything at the end:

https://github.com/vercel/next.js/blob/c215524086408114e3212f755b146e7b181226c9/crates/next-api/src/nft_json.rs#L265-L268

crates/next-api/src/nft.rs
sokra: `referenced_chunk_path.parent() != referenced_chunk_path.parent().realpath()`?

sokra (approved):

Issue #92287, opened 2026-04-03
Next.js 16.2.2 standalone + Cache Components: cached internal streamed fetches cause unbounded arrayBuffers growth and OOM


mdotk:
### Link to the code that reproduces this issue

https://github.com/mdotk/next-standalone-memory-repro

### To Reproduce

1. Clone the repro repo and install dependencies.

```bash
npm install
```

2. Build the app.

```bash
npm run build
```

3. Start the standalone server.

```bash
PORT=3025 npm run start:standalone
```

4. In another shell, run the reproducing load.

```bash
BASE_URL=http://127.0.0.1:3025 \
DURATION_MS=180000 \
CONCURRENCY=64 \
PAGE_WEIGHT_KB=2048 \
API_WEIGHT_KB=2048 \
MIX=page,api,page,page \
MAX_LOGGED_FAILURES=20 \
npm run load
```

5. Sample memory during the run.

```bash
curl 'http://127.0.0.1:3025/api/health?sample=1'
```

### Current vs. Expected behavior

Current:

On `next@16.2.2` with `output: "standalone"` and `cacheComponents: true`, the standalone server shows rapid unbounded memory growth under sustained high-cardinality traffic once the app is doing cached internal server-side `fetch()` calls against a streamed JSON route.

In one local run:

- baseline was about `95 MB rss`
- after `~28s`, memory reached about `1.58 GB rss / 589 MB arrayBuffers`
- after `~38s`, memory reached about `2.36 GB rss / 1.04 GB arrayBuffers`
- after `~62s`, memory reached about `2.42 GB rss / 1.95 GB arrayBuffers`
- after `~180s`, memory reached about `3.43 GB rss / 4.31 GB arrayBuffers`

The standalone server then died with:

```text
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

While the server is destabilizing, the load generator starts receiving:

- `ECONNRESET`
- `UND_ERR_SOCKET`
- `ECONNREFUSED`

Expected:

Memory should stabilize or at least be reclaimed after responses complete. Large temporary spikes under load are one thing, but the standalone process should not continue retaining `arrayBuffers`/`external` memory until it exits with OOM.

### Provide environment information

```bash
Operating System:
  macOS 15.x arm64
Node.js version:
  25.1.0
Next.js version:
  16.2.2
Output mode:
  standalone
Other config:
  cacheComponents: true
  compress: false
```

### Additional context

[...]

PR #90897, opened 2026-03-04
fix: backport memory leak fixes (FinalizationRegistry + LRU zero-size) to stable

Tib-Gridello:
## Summary

Backports three critical memory leak fixes from canary to the 16.x stable branch. These fixes prevent ArrayBuffer/WriteWrap retention causing OOM kills in production `output: standalone` deployments.

### Changes

1. **`clone-response.ts`**: Register **both** tee'd response clones with `FinalizationRegistry` (from #88577)
   - Previously only `cloned1` was registered. `cloned2` (stored in caches) was never auto-cancelled, leaving its tee buffer (backed by ArrayBuffers) alive indefinitely.

2. **`lru-cache.ts`**: Enforce minimum size of 1 and throw on zero-size entries (from #89040)
   - `calculateSize` returning 0 meant entries never incremented `totalSize`, so the eviction loop (`totalSize > maxSize`) never fired. The `Map` grew without bound.

3. **Caller fixes** (`filesystem.ts`, `next-dev-server.ts`): Ensure all `calculateSize` callbacks return >= 1

### Evidence

Heap dump from production (ARM64 ECS Fargate, Next.js 16.1.6, Node 22.22.0) at 1 GB ArrayBuffers:
- 977 MB in 183 `JSArrayBufferData` objects (mostly 32 MB chunks)
- Retainer chain: `WriteWrap._chunks` → GC roots (20/31 large ArrayBuffers)
- Root cause: un-cancelled tee'd ReadableStream branches holding native HTTP write buffers

| Version | After 60s load | After settling |
|---------|---------------|----------------|
| **16.1.6 stable** | 663 MB (growing) | 663 MB+ (never reclaimed, OOM) |
| **canary.79** | 1040 MB (peak) | **32 MB (back to baseline)** |

Related: #85914, #89033, #90433, #90895

## Test plan

- [x] Existing `lru-cache.test.ts` updated for new `set()` return type and zero-size rejection
- [x] Verified on production ECS Fargate (ARM64): canary.79 with these fixes resolves the leak completely

Made with [Cursor](https://cursor.com)

PR #89040, opened 2026-01-26
fix: ensure LRU cache items have minimum size of 1 to prevent unbounded growth

timneutkens:
### What?

Prevent LRU cache unbounded growth by requiring `calculateSize` to return values > 0.

### Why?

When `calculateSize` returned 0 (e.g., for `null` values in the filesystem route cache), items were never evicted because they didn't contribute to `totalSize`. This caused unbounded memory growth when handling dynamic routes like `/api/logs/[id]` where each unique ID created a new cache entry with size 0.

### How?

1. **LRUCache now throws an error if size ≤ 0** - This catches bugs at development time rather than silently masking them:
   ```typescript
   if (size <= 0) {
     throw new Error(
       \`LRUCache: calculateSize returned ${size}, but size must be > 0. \` +
       \`Items with size 0 would never be evicted, causing unbounded cache growth.\`
     )
   }
   ```

2. **Fixed filesystem.ts** - Null entries (negative cache) now return size 1 instead of 0:
   ```typescript
   if (!value) {
     // Null entries (negative cache) still need a non-zero size for LRU eviction
     return 1
   }
   ```

3. **Fixed next-dev-server.ts** - Static paths cache now uses `|| 1` instead of `?? 0`:
   ```typescript
   return JSON.stringify(value.staticPaths)?.length || 1
   ```

Fixes #89033

PR #88577, opened 2026-01-15
Add 2nd cloned body to FinalizationRegistry. Only used RuntimeStyles component for webpack

timneutkens:
### What?

Reduces memory retention in two areas: response stream cleanup in `cloneResponse` and unnecessary runtime style initialization in Turbopack mode. Also adds a test fixture for standalone mode fetch memory leak detection.

### Why?

Based on the reproduction provided in #85914, investigating potential sources of memory retention. When a `ReadableStream` is tee'd, both branches must be cancelled to release connection resources — previously only one branch was registered for GC cancellation. Additionally, `RuntimeStyles` and `_N_E_STYLE_LOAD` were always initialized even in Turbopack mode where they are unused.

### How?

**`clone-response.ts`**: Register both `cloned1` and `cloned2` bodies with the `FinalizationRegistry` so that when either response is garbage collected, its stream branch is properly cancelled.

**`app-router.tsx`**: Guard `_N_E_STYLE_LOAD` and style tracking `Set` initialization behind `!process.env.TURBOPACK && typeof window !== 'undefined'`. Rename the component to `RuntimeStylesForWebpack` and skip rendering it entirely when using Turbopack.

Fixes #85914

PR #94225, opened 2026-05-29
fix(cache-components): deterministically release retained fetch-dedupe clones on render abort

chanpark-xyz:
## What

Under Cache Components, cached server-side `fetch()`es can leak off-heap memory (`arrayBuffers` / `external`) until the process OOMs (`output: "standalone"`, #92287).

The leak is a **GC-timing race**. To cache a response *and* hand it to the render, `cloneResponse()` tees the body; the React **fetch dedupe cache** retains the second clone (`entry[2]`) per URL, and draining the sibling fills that retained clone with the whole body. It's released only when the `FinalizationRegistry` runs — i.e. on GC — but the bytes are off-heap, and V8 schedules GC mainly off JS-heap pressure, so under sustained load the retained clones' bytes pile up faster than GC reclaims them.

This PR makes that cleanup **deterministic** for cached fetches issued during a render: the retained clone is cancelled the moment the render is torn down, instead of waiting for GC.

## Root cause (confirmed)

Registering both clones with the `FinalizationRegistry` (#88577 / #90897) is necessary but not sufficient: while the retained clone stays reachable, the finalizer can't run, and once it's unreachable the off-heap bytes still wait on a GC that lags under load. Confirmed on a production `standalone` heap snapshot — the dedupe-retained undici response streams were pinned via `Node / CleanupQueue` / global handles.

## Change

- **`clone-response.ts`** — `cloneResponse` stays a pure "split a `Response` in two" utility. A new export, `cancelUnconsumedBodyOnAbort(signal, body)`, cancels a *retained, un-read* clone deterministically when `signal` aborts (a `locked`/in-flight body is left untouched; one shared `abort` listener per signal avoids `MaxListenersExceededWarning`). The `FinalizationRegistry` stays as a backstop.
- **`dedupe-fetch.ts`** — registers the retained clone (`entry[2]`) against the render's abort signal.
- **`work-unit-async-storage.external.ts`** — `getRenderAbortSignal(workUnitStore)`.

## Scope & follow-up (honest)

- **Covers** cached fetches issued at the **render level** (prerender stores), where the render's `renderSignal` is reachable — the cleanup becomes deterministic there.
- **Does not yet cover** fetches issued *inside* `"use cache"`: those run in a `'cache'` work-unit store, where the render signal isn't currently reachable, so `getRenderAbortSignal` returns `null` and the cancellation is a no-op. Confirmed by instrumenting the running build (`storeType=cache, hasRenderSignal=false`), and a heavy-load matched run of the minimal repro (which fetches inside `"use cache"`) shows patched ≈ unpatched. **So this PR does not by itself resolve that repro.** Threading an abort/teardown signal through the cache-store lifecycle (`use-cache-wrapper.ts`) is a follow-up I'm investigating; this change is the groundwork (pure utility + explicit, reusable cancel helper).

## Verification

- **Mechanism (isolation A/B, deterministic):** a retained clone pins ~800 MB off-heap until `cancelUnconsumedBodyOnAbort` cancels it, after which GC reclaims to ~0. Always reproduces.
- **Root cause (heap snapshot):** production `standalone` snapshot shows the dedupe-retained streams pinned via `Node / CleanupQueue`.
- **Unit tests:** `clone-response.test.ts` + `dedupe-fetch.test.ts` (41 pass) — cover the helper and the render-store threading.
- **End-to-end:** the minimal repro uses `"use cache"`, which this PR doesn't yet reach (see *Scope*), so it isn't the validation here; render-level coverage is validated by the isolation A/B above.

Related: #88577, #89040, #90897.


chanpark-xyz:
Closing this draft. The original end-to-end verification was invalid for the "use cache" repro path: instrumenting the running build showed the leaking fetch runs inside a `'cache'` work-unit store, where the render abort signal isn't reachable (`getRenderAbortSignal` returns null), so this patch is a no-op for that pattern — a matched heavy-load run of the official repro shows patched ≈ unpatched.

The dedupe-clone retention is a real mechanism in isolation, but I have not confirmed it is the dominant off-heap retainer for the repro (which fully recovers at idle). A fresh PR will follow once a peak heap snapshot pins the actual retainer (dedupe clone vs use-cache result-stream tee vs transient) and the fix is verified against the heavy-load repro (patched ≪ unpatched). Tracking #92287.

PR #94234, opened 2026-05-29
Turbopack: compute code hash per `use cache` function

mischnic:
This computes the implementation hash of a given `use cache`:

- it includes the module containing the `use cache` function, plus all transitive imports of that module (including externals via NFT)
- what does it hash
    - for bundled code, it hashes the generated code (so including any AST transforms, inlined env vars, etc)
    - for NFTd files, it hashes the file content on disk
- So this is basically equivalent to hashing the generated JS chunks, except that it's scoped to only what the given actions module uses
- (so notably, if you have multiple use cache functions in a given file, then they will all have the same hash)

<br/>

- [x] Gate behind experimental flag
- [x] Only compute hash for `use cache` functions, not all server actions
- [ ] `final_read_hint` is a problem. Some chunk items are now codgened twice, so it recomputes the AST
- [x] disable for dev for now
- [ ] Measure perf impact

```jsonc
// .next/server/server-reference-manifest.json
{
  "node": {
    "806f4954cfbb75404a19d6d405065ed9059cc0cab2": {
      "workers": {
        "app/rsc/page": {
          "moduleId": "[project]/bench/app-router-server/.next-internal/server/app/rsc/page/actions.js { ACTIONS_MODULE0 => \"[project]/bench/app-router-server/app/rsc/logic.js [app-rsc] (ecmascript)\" } [app-rsc] (server actions loader, ecmascript)",
          "async": false,
          "exportedName": "$$RSC_SERVER_CACHE_0",
          "filename": "bench/app-router-server/app/rsc/logic.js",
          "codeHash": "413e394d9597b35df12828a6a7fd6363" // <------------
        }
      },
      "filename": "bench/app-router-server/app/rsc/logic.js",
      "exportedName": "$$RSC_SERVER_CACHE_0"
    }
  },
  "edge": {},
  "encryptionKey": "jOF2TbpNLjp2oOhl3VPBwGE7luodnei9clJPq/gaYQo="
}
```


mischnic:
* **#94234** <a href="https://app.graphite.com/github/pr/vercel/next.js/94234?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94234?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#95232** <a href="https://app.graphite.com/github/pr/vercel/next.js/95232?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

crates/next-api/src/server_actions.rs
```diff
@@ -225,6 +246,7 @@ impl Asset for ServerActionManifestAsset {
         struct ActionMetadata<'a> {
             exported_name: &'a str,
             filename: Cow<'a, str>,
+            code_hash: Option<ReadRef<RcStr>>,
```
bgw, line 253: `ReadRef` of an `RcStr` is a little silly.

```suggestion
            code_hash: Option<RcStr>,
```

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
     Ok(module)
 }
 
+#[turbo_tasks::function]
+async fn compute_subtree_content_hash(
+    module_graph: ResolvedVc<ModuleGraph>,
+    entry: ResolvedVc<Box<dyn Module>>,
+    chunking_context: Vc<Box<dyn ChunkingContext>>,
+) -> Result<Vc<RcStr>> {
```
bgw, line 364: Maybe this should return a u128 and we should defer conversion of the hash to a string to right before serialization?

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
...
+                    // modules themselves.
+                    modules.insert(target);
+                    Ok(GraphTraversalAction::Exclude)
+                } else {
+                    modules.insert(target);
+                    Ok(GraphTraversalAction::Continue)
+                }
+            },
+            |_, _, _| Ok(()),
+            true,
```
bgw:
```suggestion
            /* include_traced */ true,
```

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
...
+                    // Include the client reference proxy module, but not the referenced client
+                    // modules themselves.
+                    modules.insert(target);
+                    Ok(GraphTraversalAction::Exclude)
+                } else {
+                    modules.insert(target);
+                    Ok(GraphTraversalAction::Continue)
+                }
+            },
+            |_, _, _| Ok(()),
```
bgw:
```suggestion
            /* visit_postorder */ |_, _, _| Ok(()),
```

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
...
+    );
+    match async {
+        let module_graph_value = module_graph.await?;
+        let async_module_info = module_graph.async_module_info();
+
+        let mut modules = FxIndexSet::default();
+        module_graph_value.traverse_edges_dfs(
+            std::iter::once(entry),
+            &mut (),
+            |_, target, _| {
```
bgw, line 378:
```suggestion
            /* visit_preorder */ |_, target, _| {
```

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
...
+        entry = display(entry.ident_string().await?)
+    );
+    match async {
+        let module_graph_value = module_graph.await?;
+        let async_module_info = module_graph.async_module_info();
+
+        let mut modules = FxIndexSet::default();
+        module_graph_value.traverse_edges_dfs(
+            std::iter::once(entry),
+            &mut (),
```
bgw:
```suggestion
            /* state */ &mut (),
```

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
...
+    }
+    .instrument(span)
+    .await
+    {
+        Ok(hash) => Ok(hash),
+        // ast-grep-ignore: no-context-turbofmt
+        Err(e) => Err(e.context(
+            turbofmt!(
+                "Failed to compute use-cache code hash {}",
+                entry.ident_string().await?
```
bgw: Isn't the whole point of the `turbofmt` macro that you don't need to await the `Vc`?

```suggestion
                entry.ident_string(),
```

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
...
+async fn module_hash(
+    module_graph: ResolvedVc<ModuleGraph>,
+    chunking_context: ResolvedVc<Box<dyn ChunkingContext>>,
+    async_module_info: ResolvedVc<AsyncModulesInfo>,
+    m: ResolvedVc<Box<dyn Module>>,
+) -> Result<Vc<RcStr>> {
+    let ident = m.ident();
+    let ident_value = ident.await?;
+    let ident_str = ident.to_string().await?;
+    Ok(Vc::cell(
```
bgw: Move the wrapping `Ok(Vc::cell(` down into the last statement of each block so that this giant if/else condition doesn't get an extra indent.

crates/next-api/src/server_actions.rs
```diff
@@ -316,6 +354,123 @@ pub async fn to_rsc_context(
...
+                ResolvedVc::try_downcast::<Box<dyn EcmascriptChunkItem>>(chunk_item).unwrap();
+            let async_info = if async_module_info.is_async(m).await? {
+                Some(module_graph.referenced_async_modules(*m))
+            } else {
+                None
+            };
+            let code = chunk_item.code(async_info).await?;
+            RcStr::from(deterministic_hash(
+                "",
+                (ident_str, code.source_code()),
```
bgw: What about `.source_code_hash()`? I'm guessing the issue is that's not a 128 bit hash? Maybe it should?

turbopack/crates/turbo-tasks-hash/src/deterministic_hash.rs
```diff
@@ -112,6 +112,15 @@ impl DeterministicHash for [u8] {
     }
 }
 
+impl<T: DeterministicHash, const N: usize> DeterministicHash for [T; N] {
+    fn deterministic_hash<H: DeterministicHasher>(&self, state: &mut H) {
+        state.write_usize(self.len());
+        for v in self {
+            v.deterministic_hash(state);
+        }
+    }
```
bgw: It would be good to have a specialized version of this for `[T; u8]` that uses `write_bytes` like `String` does?

This would probably be a reasonable use of the nightly min specialization feature? If we later had to turn off the feature to compile on a stable compiler, it wouldn't break anything.

packages/next/src/server/config-shared.ts
```diff
@@ -1133,6 +1133,11 @@ export interface ExperimentalConfig {
    */
   useCache?: boolean
 
+  /**
+   * Enables durable `"use cache"` remote cache entries across deployments.
+   */
+  durableUseCacheEntries?: boolean
+
```
bgw, line 1224: Document that this is turbopack-only? Maybe put that in the config name?

bgw (approved):

crates/next-api/src/server_actions.rs
mischnic: The current version does cache the `encode_hex()` operation though

crates/next-api/src/server_actions.rs
mischnic: Is it though? If I make it a RcStr, then I'd have to add a `clone()` somewhere.

turbopack/crates/turbo-tasks-hash/src/deterministic_hash.rs
mischnic: I just changed the callsites to use u128 directly, so this impl isn't needed anymore at all.

packages/next/src/server/config-shared.ts
mischnic: Well, in that sense, `supportsImmutableAssets` is Turbopack only as well.

PR #94238, opened 2026-05-29
fix: release retained fetch dedupe clones at cache-scope completion

### LINKED ISSUE #92287: Next.js 16.2.2 standalone + Cache Components: cached internal streamed fetches cause unbounded arrayBuffers growth and OOM

chanpark-xyz:
Fixes #92287

## What

Under Cache Components, especially in `output: "standalone"` deployments, cached server-side `fetch()` calls can grow `arrayBuffers` / `external` until the process OOMs (#92287). This releases those off-heap bytes **deterministically when the cache scope completes**, instead of waiting for GC to run soon enough.

## Root cause

To both cache a response and hand it to the caller, `cloneResponse()` `tee()`s the body. The React fetch-dedupe cache retains the **second** branch (`entry[2]`) per URL so a later duplicate fetch in the same scope can be served. Draining the first branch force-fills the retained branch with the whole body (`tee()` buffering), so the retained clone holds the entire response **off-heap**.

That clone lives in a `React.cache` map keyed by the cache fill's `RequestInstance`, and on the normal-completion path only becomes unreachable once React drops the instance **and** GC runs. But the bytes are off-heap, so they create little JS-heap pressure and a major GC isn't scheduled promptly; under sustained, high-cardinality load, new cache-fill scopes create retained clones faster than completed scopes become unreachable and collected, and the retained clones pile up until OOM.

This is reachability/scope-lifetime, **not** collectable garbage:

- A forced full `global.gc()` at peak reclaims **< 10%** of `arrayBuffers` (the rest is genuinely reachable).
- Memory **fully recovers at idle** once the load stops and the fills' `RequestInstance`s become unreachable.

The existing `FinalizationRegistry` in `cloneResponse` is a backstop, but it can only run on the GC that lags here.

## Fix

Register the retained clone's body for cancellation on **`React.cacheSignal()`** — the signal React aborts when the cache scope settles (success, error, or explicit abort): *"All cacheSignals are now aborted to allow clean up of any unused resources."* When it fires we `cancel()` the un-read `tee()` branch, releasing the off-heap bytes immediately rather than waiting for GC.

- **`clone-response.ts`** — `cancelUnconsumedBodyOnAbort(signal, body)`: cancels a retained, un-read body when `signal` aborts. A `locked`/in-flight body is left untouched, and a single `abort` listener is shared per signal (fanning out to each retained clone) so a fill that performs many cached fetches doesn't add one listener per fetch and trigger a `MaxListenersExceededWarning`. The `FinalizationRegistry` stays as the backstop.
- **`dedupe-fetch.ts`** — registers the retained clone (`entry[2]`) against `React.cacheSignal()` at both clone sites.

Why this is safe:

- Cancellation runs when the cache scope's signal aborts — at **scope completion**, or on the next microtask if the scope already settled before the fetch resolved — always after any in-scope duplicate fetch has re-cloned the spare, so it cannot race a live consumer.
- It touches only the **retained spare** (`cloned2`), never the clone returned to the caller (`cloned1`).
- `React.cacheSignal` is accessed defensively (`typeof … === 'function'`). Outside a React cache/render scope, or on older React builds without `cacheSignal`, this is a no-op and behavior is unchanged — the `FinalizationRegistry` remains the backstop.

This covers fetches issued **inside `"use cache"`** (whose work-unit store carries no render abort signal — the gap that made earlier render-signal-based attempts a no-op for the reported repro) as well as render-level fetches, uniformly.

## Verification

Environment: the official minimal repro (`output: "standalone"`, `cacheComponents: true`, a `"use cache"` data fetcher hit from a route handler over an internal `fetch()` streaming 512 KB bodies) — Next `16.3.0-canary.33`, Node `24.12.0`, React `19.1.0`.

Load A/B on the **same build with the cancel toggled on/off** (no build-pipeline confound), 96-concurrency API load, 150 s:

| | peak `arrayBuffers` | 30 s-tail average | end of load | requests | failed |
|---|---:|---:|---:|---:|---:|
| off | 876 MB (still rising) | 807 MB | 876 MB | 18,378 | 0 |
| on | 272 MB | 222 MB | 214 MB | 18,546 | 0 |

`arrayBuffers` trajectory (~10 s apart) — the trend, not a single noisy peak:

```
off: 208 568 659 626 643 635 642 637 637 711 740 738 750 837 819   (climbs toward OOM)
on:  205 192 188 214 189 226 191 238 246 254 222 228 237 207 214   (flat — bounded working set)
```

(The table's peak is the max across all 2 s samples; the trajectory lists points ~10 s apart, so the 876 MB peak falls between the shown 837/819 samples.)

≈ **−72%** on the noise-robust 30 s-tail average; the unpatched arm **accumulates** (would OOM given time), the patched arm stays **bounded** at the in-flight working set. Instrumentation confirmed each completed fill's retained spare was cancelled at scope completion (**18,334 cancels, 0 errors**). The small gap to 18,546 requests is fills still in flight when the load was stopped, not failures — **0 responses failed** in either arm, so the deterministic cancel never truncates a served body.

- **Mechanism (forced-GC probe):** a major `global.gc()` at peak frees < 10%, so the retention is genuine reachability and a deterministic cancel — not GC tuning — is what's required.
- **Unit tests:** `clone-response.test.ts` + `dedupe-fetch.test.ts` (42 pass) — cover the helper and the cache-scope-signal threading on the cache-miss path, cache-hit path, and the no-scope (`null`) case.
- **Types:** `tsc --noEmit` on `packages/next` is clean.

Related: #88577, #89040, #90897.

### DIFF
--- a/packages/next/src/server/lib/clone-response.test.ts
+++ b/packages/next/src/server/lib/clone-response.test.ts
@@ -1,0 +1,175 @@
+/**
+ * @jest-environment node
+ */
+import { cloneResponse, cancelUnconsumedBodyOnAbort } from './clone-response'
+
+// Drain microtasks/macrotasks so a Web Streams `cancel()` settles. This waits on
+// stream-internal tasks, not application state — there is no condition to poll,
+// so a fixed drain (rather than a `retry()`-style wait) is the right tool here.
+async function flush() {
+  for (let i = 0; i < 6; i++) {
+    await new Promise((resolve) => setTimeout(resolve, 0))
+  }
+}
+
+function streamedResponse(text: string) {
+  const bytes = new TextEncoder().encode(text)
+  return new Response(
+    new ReadableStream<Uint8Array>({
+      start(controller) {
+        controller.enqueue(bytes)
+        controller.close()
+      },
+    }),
+    { status: 200, statusText: 'OK', headers: { 'x-test': '1' } }
+  )
+}
+
+// A cancelled stream yields `{ done: true }` with no data on the first read.
+async function firstReadDone(body: ReadableStream | null) {
+  if (!body) return true
+  const reader = body.getReader()
+  const { done } = await reader.read()
+  reader.releaseLock()
+  return done
+}
+
+describe('cloneResponse', () => {
+  it('returns the same response twice when there is no body', () => {
+    const original = new Response(null, { status: 204 })
+    const [cloned1, cloned2] = cloneResponse(original)
+    expect(cloned1).toBe(original)
+    expect(cloned2).toBe(original)
+  })
+
+  it('produces two independently readable clones', async () => {
+    const [cloned1, cloned2] = cloneResponse(new Response('hello'))
+    expect(await cloned1.text()).toBe('hello')
+    expect(await cloned2.text()).toBe('hello')
+  })
+
+  it('preserves status, statusText and headers on both clones', () => {
+    const [cloned1, cloned2] = cloneResponse(
+      new Response('x', {
+        status: 201,
+        statusText: 'Created',
+        headers: { 'x-test': 'yes' },
+      })
+    )
+    for (const clone of [cloned1, cloned2]) {
+      expect(clone.status).toBe(201)
+      expect(clone.statusText).toBe('Created')
+      expect(clone.headers.get('x-test')).toBe('yes')
+    }
+  })
+})
+
+describe('cancelUnconsumedBodyOnAbort', () => {
+  // Regression test for https://github.com/vercel/next.js/issues/92287: a
+  // retained, un-read clone must be released when the render aborts, instead of
+  // being held off-heap until the FinalizationRegistry happens to run.
+  it('cancels a retained, un-read body when the signal aborts', async () => {
+    const controller = new AbortController()
+    const [, retained] = cloneResponse(streamedResponse('leak'))
+    cancelUnconsumedBodyOnAbort(controller.signal, retained.body)
+
+    controller.abort()
+    await flush()
+
+    expect(await firstReadDone(retained.body)).toBe(true)
+  })
+
+  // Models the real leak ordering: the cache side drains the sibling clone
+  // (which fills the retained clone's tee buffer), then the render aborts.
+  it('releases a retained body whose sibling was already drained', async () => {
+    const controller = new AbortController()
+    const [consumed, retained] = cloneResponse(streamedResponse('leak-body'))
+    cancelUnconsumedBodyOnAbort(controller.signal, retained.body)
+
+    expect(await consumed.text()).toBe('leak-body')
+    expect(retained.body!.locked).toBe(false)
+
+    controller.abort()
+    await flush()
+
+    expect(await firstReadDone(retained.body)).toBe(true)
+  })
+
+  it('does nothing without a signal', async () => {
+    const [, retained] = cloneResponse(streamedResponse('keep'))
+    cancelUnconsumedBodyOnAbort(null, retained.body)
+    cancelUnconsumedBodyOnAbort(undefined, retained.body)
+    await flush()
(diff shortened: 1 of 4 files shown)


0-don:
looking forward to this

0-don:
Confirmed in production on 16.3.0 (standalone, Turbopack, cacheComponents). Note for testers: on Turbopack standalone builds the `dist/server/lib` backport is dead code; the executing copy is minified inside `dist/compiled/next-server/app-route-turbo.runtime.prod.js`. Our first A/B was a placebo for exactly this reason.

After applying this PR's semantics to the compiled bundle (verified executing in the live process): never-drained tee branches went from ~100/min unbounded growth to a bounded steady state, arrayBuffers slope down 3-4x. Remaining growth is the separate segmentData/prerenderChunks family (details on the issue), so: necessary but not sufficient for cacheComponents workloads. Hope this lands soon.

unav4ila8le:
this is needed

andreasdc:
> this is needed

If that fixes the big issue, yes.

PR #94241, opened 2026-05-29
[CLI] Exit non-zero with a clear message when next typegen fails

gaojude:
The `typegen` command chained `nextTypegen(...).then(() => process.exit(0))` with no rejection handler. When `next.config` throws during `loadConfig`, the rejection is orphaned by the synchronous `program.parse` and then absorbed by Next's `unhandledRejection` filter, so the process exits `0` having written no route types. A later `tsc` then reports `Cannot find name 'PageProps'`/`'LayoutProps'` (and missing `*.png` modules) with nothing tying it back to typegen. Add a `.catch` that prints an actionable message about the missing route types and `process.exit(1)`, so `next typegen && tsc` halts on failure. Includes an e2e test whose fixture config throws and asserts a non-zero exit with the diagnostic.

### DIFF
--- a/packages/next/src/bin/next.ts
+++ b/packages/next/src/bin/next.ts
@@ -518,7 +518,26 @@ program
     // ensure process exits after typegen completes so open handles/connections
     // don't cause process to hang
     import('../cli/next-typegen.js').then((mod) =>
-      mod.nextTypegen(options, directory).then(() => process.exit(0))
+      mod
+        .nextTypegen(options, directory)
+        .then(() => process.exit(0))
+        .catch((err: unknown) => {
+          // Surface failures loudly with an actionable message and a non-zero
+          // exit code. Without this, a failed typegen (e.g. `next.config`
+          // throwing) is swallowed as an unhandled rejection that exits 0 and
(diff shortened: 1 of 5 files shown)


packages/next/src/bin/next.ts
```diff
@@ -518,7 +518,26 @@ program
...
+          // throwing) is swallowed as an unhandled rejection that exits 0 and
+          // writes no types, leaving the user to chase cryptic
+          // `Cannot find name 'PageProps'/'LayoutProps'` errors from a later
+          // `tsc` run.
+          console.error(
+            '\n> Failed to generate route types. The global `PageProps` and ' +
+              '`LayoutProps` helpers and other generated route types were not ' +
+              "written, which can surface as `Cannot find name 'PageProps'` " +
+              'errors during type checking. This is often caused by ' +
+              '`next.config` failing to load. Original error:\n'
```
lukesandberg: this seems a little wordy,  do we really need to say that the types were not written? (actually they may have been written if the failure was late enough)

A simple: `Unexpected error while generating route types, Original error:\n` is probably enough.

lukesandberg (approved):

PR #94243, opened 2026-05-29
Persist debug channel via `IndexedDB` without blocking hydration

unstubbable:
In development the debug channel that streams React's Server Component debug information to the browser is buffered for the initial document and persisted, so that it can be replayed when the browser later serves the page from its HTTP cache — for example on back/forward navigation or tab duplication — instead of forcing a full reload. This changes where that buffer is persisted, moving it from `sessionStorage` to `IndexedDB`.

The largest win comes from no longer having to serialize the data. `sessionStorage` can only hold strings, so the binary debug chunks had to be encoded into a string before being written and parsed back out again on restore, which becomes expensive once the payload grows to several megabytes. `IndexedDB` stores the chunks directly as the `Uint8Array`s they already are, so there is no encode and parse round trip on either side. Its asynchronous API helps a little on top of that, since the write itself no longer has to happen synchronously.

The other improvement is that persistence no longer competes with hydration. The initial document's debug stream closes while hydration is still running, and the previous synchronous `sessionStorage` write happened at exactly that moment, taking main thread time away from hydration. The write is now deferred with `requestIdleCallback`, so it only runs once the main thread is genuinely idle, which is after hydration has drained, and it is skipped entirely if the page navigates away before that happens, in which case a later restore simply falls back to a reload. This is visible in the profiles below: with `sessionStorage` there is still hydration work running after the persistence task, whereas with the idle-scheduled `IndexedDB` write the persistence only runs once hydration is done.

In a profile of a test page that deliberately transfers a large amount of debug information, the persistence work on the main thread dropped from more than 700 ms to roughly 25 to 45 ms. A real application with far less debug data will see a smaller difference, so the test page amplifies the effect, but the direction is the same.

The number of persisted entries stays bounded to 10, with the oldest pruned on each write, and an end-to-end test covers the case where an entry is pushed out by newer page loads so that navigating back to it recovers through a page reload.

**Before with `sessionStorage`**:

sessionStorage

**After with idle-scheduled `IndexedDB`:**

indexedDB

### DIFF
--- a/packages/next/src/client/dev/debug-channel.ts
+++ b/packages/next/src/client/dev/debug-channel.ts
@@ -1,6 +1,22 @@
 import { NEXT_REQUEST_ID_HEADER } from '../components/app-router-headers'
 import { InvariantError } from '../../shared/lib/invariant-error'
 
+declare global {
+  interface Window {
+    /**
+     * Test-only flag, set once this document's debug channel entry is durably
+     * committed. Persistence is deferred to an idle callback and its write is
+     * async, so exposing this flag is a deliberate compromise: it lets e2e
+     * tests await persistence deterministically — rather than relying on
+     * timing/idle hacks — while coupling only to "an entry was persisted" and
+     * not to how or where it is stored. It resets naturally on each navigation
+     * since every document gets a fresh window. Only set when
+     * `process.env.__NEXT_TEST_MODE` is enabled.
+     */
+    __NEXT_DEBUG_CHANNEL_PERSISTED?: boolean
+  }
+}
+
 export interface DebugChannelReadableWriterPair {
   readonly readable: ReadableStream<Uint8Array>
   readonly writer: WritableStreamDefaultWriter<Uint8Array>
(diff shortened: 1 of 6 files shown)


packages/next/src/client/dev/debug-channel.ts
```diff
@@ -1,51 +1,173 @@
...
+            const getReq: IDBRequest<DebugChannelEntry | undefined> =
+              store.get(requestId)
+            getReq.onsuccess = () => resolve(getReq.result)
+            getReq.onerror = () => reject(getReq.error)
+          })
+        } finally {
+          db.close()
+        }
+      } catch {
+        // Treat any IDB failure as "no entry" and fall through to reload.
```
timneutkens, line 145: I know `setItem` previously also had a `catch {}` but in the past there have been very hard to debug issue reports caused by not logging errors if they happen. Can we log these failures, potentially as verbose level `console.debug()`.

packages/next/src/client/dev/debug-channel.ts
```diff
@@ -143,12 +233,29 @@ export function getOrCreateDebugChannelReadableWriterPair(
...
+        // The initial document's debug stream closes while hydration is still
+        // running, so persisting here would steal main-thread time from it.
+        // Wait for genuine idle (no timeout): persistence is best-effort, so if
+        // the page never idles before navigation we skip it and a later restore
+        // falls back to a reload, rather than forcing a blocking write.
+        await whenIdle()
+        await persistDebugChannelToIndexedDB(requestId, chunks)
+      })
+      .catch(() => {
+        // writer.closed rejected (e.g., stream aborted) — nothing to persist.
```
timneutkens, line 345: Smae as https://github.com/vercel/next.js/pull/94243/changes#r3332902952

timneutkens (approved):
Looks good, only the two small nits on the catch.

packages/next/src/client/dev/debug-channel.ts
```diff
@@ -1,51 +1,179 @@
 import { NEXT_REQUEST_ID_HEADER } from '../components/app-router-headers'
 import { InvariantError } from '../../shared/lib/invariant-error'
 
+declare global {
+  interface Window {
+    /**
+     * Test-only flag, set once this document's debug channel entry is durably
```
eps1lon: If it's test-only, we shouldn't ship it in declaration files since that augmentation will be visible to user code.

Adding `@internal` might strip it from the emitted declaration files. Either way, please manually verify this isn't shipped.

packages/next/src/client/dev/debug-channel.ts
```diff
@@ -1,51 +1,179 @@
...
-      let binary = ''
-      for (let i = 0; i < chunk.byteLength; i++) {
-        binary += String.fromCharCode(chunk[i])
-      }
-      return btoa(binary)
-    })
-  )
+function openDebugChannelDB(): Promise<IDBDatabase> {
+  return new Promise((resolve, reject) => {
+    const req = indexedDB.open(DB_NAME, 1)
```
eps1lon: "req" stands for "Request"? Can we use something meaningful to not be grepped on "req" which is commonly used for http requests.

test/e2e/app-dir/bfcache-regression/bfcache-regression.test.ts
```diff
@@ -82,4 +82,113 @@ describe('bfcache-regression', () => {
...
+    isNextDev &&
+    // Excludes Safari: the WebKit build Playwright uses re-fetches the document
+    // from the server on back-navigation instead of restoring it from the HTTP
+    // cache like Chrome and Firefox, so each back-navigation gets a fresh
+    // request ID that was never persisted, the request-id-keyed restore always
+    // misses, and every page reloads — the pruned-vs-cached distinction this
+    // test relies on never appears. Real Safari serves back-navigations from
+    // cache without a server request, so this is a limitation of the test's
+    // WebKit rather than of the feature.
+    global.browserName !== 'safari' &&
```
eps1lon: Can we encode the current behavior of safari in test assertions instead of outright skipping? That way we get a notification when this Playwright behavior changes.

test/e2e/app-dir/bfcache-regression/bfcache-regression.test.ts
```diff
@@ -82,4 +82,113 @@ describe('bfcache-regression', () => {
 
     await assertNoConsoleErrors(browser)
   })
+
+  if (
+    // Persistence only exists in dev.
+    isNextDev &&
+    // Excludes Safari: the WebKit build Playwright uses re-fetches the document
```
eps1lon: Why does it do that? Is that a Playwright bug or a config bug on our side?

packages/next/src/client/dev/debug-channel.ts
unstubbable: Removed in favor of a type cast at the usage side.

packages/next/src/client/dev/debug-channel.ts
unstubbable: done

test/e2e/app-dir/bfcache-regression/bfcache-regression.test.ts
unstubbable: done. this also uncovered an edge case bug which I've addressed in the downstack PR:
- #94317

test/e2e/app-dir/bfcache-regression/bfcache-regression.test.ts
unstubbable: It's not something we can flip from `launch()`. `--disable-back-forward-cache` is a Chromium-only flag (I tried passing it via `ignoreDefaultArgs` to `webkit.launch()` with no effect on the test), and Playwright's WebKit defaults are just `--inspector-pipe`, `--headless`, `--user-data-dir`, and proxy plumbing. WebKit's cache behavior is governed by `WKPreferences` and network-session settings, which Playwright doesn't expose at launch, so this looks like a Playwright-WebKit build choice rather than a config bug on our side.

eps1lon (approved):

PR #94252, opened 2026-05-29
fix(types): correct AppType generic parameter usage

EduardF1:
## Summary

Fixes the generic type parameter on AppType so that app-level initial props (from getInitialProps) are correctly merged at the top level of the component's props, rather than being applied to pageProps.

## Problem

AppType<P> passes P as the second argument to AppPropsType<any, P>, making pageProps typed as P. This means:
- props.myProp causes a TypeScript error (doesn't exist on type)
- props.pageProps.myProp incorrectly passes type-checking (false negative)

This is inconsistent with runtime behaviour, where getInitialProps result from _app is spread onto the component's top-level props.

## Fix

Change the third generic argument from \AppPropsType<any, P>\ to \AppPropsType<any, any> & P\, so the app-level initial props are correctly intersected at the top level.

\\\diff
 export type AppType<P = {}> = NextComponentType<
   AppContextType,
   P,
-  AppPropsType<any, P>
+  AppPropsType<any, any> & P
 >
\\\

Fixes #42846


EduardF1:
Friendly ping! CI is green and this type fix is ready for review.

EduardF1:
Gentle ping for a types maintainer review — this is the `AppType<P>` fix for #42846.

One thing I'd specifically like guidance on: this change makes `props.pageProps` resolve to `any` (the `P` moves from the `pageProps` slot to a top-level intersection). That matches the runtime spreading of `_app.getInitialProps()`, but it does drop the `pageProps` typing that `AppPropsType<any, P>` used to provide. If the team prefers to preserve `pageProps` typing (e.g. `Omit<AppPropsType<any, any>, 'pageProps'> & P & { pageProps: ... }`), I'm happy to adjust.

I know #42846 has a lot of competing PRs — happy to align with whatever shape the maintainers prefer, or close this if a preferred fix already exists.

PR #94257, opened 2026-05-30
fix: #14244 - ISG with GIP in _app overrides cache-control

### LINKED ISSUE #14244: ISG with GIP in _app overrides cache-control of ISG page
# Bug report

When using the `unstable_revalidate` param with ISG, a header set by _app should never win.

## Additional context

From @lfades:

> Update: this is currently not the case, SSG finishes the request before it gets to _app.
>
> the custom _app in the Next.js site was setting a header, but that causes an error because the request was sent, this happens in production and with next start, but in dev you don't get any error.

duongynhi000005-oss:
## Summary

Fixes #14244

When a custom `_app` has `getInitialProps` that sets a `Cache-Control` header, it was preventing the ISG/ISR page's cache-control from being applied. The `sendRenderResult` function and the pages-handler both checked `!res.getHeader('Cache-Control')` before applying the ISG revalidation cache-control header, meaning any earlier header would win.

## Changes

- **`packages/next/src/server/send-payload.ts`**: Removed the `!res.getHeader('Cache-Control')` guard so ISG/ISR cache-control always takes precedence.
- **`packages/next/src/server/route-modules/pages/pages-handler.ts`**: Same fix for the pages route handler path.

## What was tested

- Added a regression test (`test/production/prerender-revalidate-with-app-gip`) that verifies an ISG page's `s-maxage` cache-control is not overridden by a `Cache-Control` header set in `_app`'s `getInitialProps`.

### DIFF
--- a/packages/next/src/server/route-modules/pages/pages-handler.ts
+++ b/packages/next/src/server/route-modules/pages/pages-handler.ts
@@ -638,9 +638,9 @@ export const getHandler = ({
           }
         }
 
-        // If cache control is already set on the response we don't
-        // override it to allow users to customize it via next.config
-        if (cacheControl && !res.getHeader('Cache-Control')) {
+        // ISG/ISR cache-control must always take precedence over any
+        // Cache-Control header set earlier (e.g. by _app's getInitialProps)
+        if (cacheControl) {
           res.setHeader('Cache-Control', getCacheControlHeader(cacheControl))
         }
 
--- a/packages/next/src/server/send-payload.ts
+++ b/packages/next/src/server/send-payload.ts
@@ -55,9 +55,12 @@ export async function sendRenderResult({
     res.setHeader('X-Powered-By', 'Next.js')
   }
 
-  // If cache control is already set on the response we don't
-  // override it to allow users to customize it via next.config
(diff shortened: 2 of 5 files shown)


iknoul:
The test correctly covers the `_app` GIP case, but as the Vercel bot pointed out, removing the `!res.getHeader('Cache-Control')` guard also causes a regression for `Cache-Control` headers set via `next.config` — and there's no test for that scenario here.

Could you add a case like this to the test suite?

```js
// next.config.js
module.exports = {
  async headers() {
    return [
      {
        source: '/isg',
        headers: [{ key: 'Cache-Control', value: 'public, max-age=9999' }],
      },
    ]
  },
}
```

Then a second test assertion:

```ts
it('should respect Cache-Control set via next.config headers', async () => {
  const res = await next.fetch('/isg')
  const cacheControl = res.headers.get('Cache-Control')
  expect(cacheControl).toContain('max-age=9999')
})
```

With the current fix, this test would fail — the `next.config` header gets overridden too, which is the regression. The fix needs to distinguish between a header set by `_app`'s `getInitialProps` (should be overridden) vs one set by user config (should be respected).

Issue #93556, opened 2026-05-07
Turbopack fails to resolve next/package.json with pnpm v11 enableGlobalVirtualStore


yamaaaaaa31:
### Link to the code that reproduces this issue

https://github.com/yamaaaaaa31/turbopack-pnpm-v11-repro

### To Reproduce

1. Clone the reproduction repository.
2. Use pnpm v11.
3. Make sure pnpm global virtual store is enabled:

```yaml
# pnpm-workspace.yaml
enableGlobalVirtualStore: true
```

4. Install dependencies:

```bash
pnpm install
```

5. Start the development server with Turbopack:

```bash
pnpm dev
```

6. Open the app route in the browser.

### Current vs. Expected behavior

When `enableGlobalVirtualStore: true` is enabled in pnpm v11, `node_modules/next` is a symlink whose real path points outside the project root, for example:

```text
node_modules/next -> ~/Library/pnpm/store/v11/links/.../node_modules/next
```

With Turbopack, `next dev` fails when serving the app because it cannot resolve `next/package.json`:

```text
Couldn't find the Next.js package (next/package.json)
from project directory: <project>/src/app
```

Expected behavior: Turbopack should handle pnpm v11 global virtual store symlinks for normal runtime dependencies, including Next.js itself, or provide a clear error/workaround.

This does not happen when using webpack dev mode.

Additional verification:

- This also reproduces with `next@canary`.
- This is not specific to `output: "standalone"`. The linked reproduction does not set `output: "standalone"` and still reproduces.
- Setting `experimental.externalDir: true` does not address this. The failure appears to happen at Turbopack's filesystem root / symlink resolution layer, before regular external source imports.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.4.0
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 22.22.2
  npm: 10.9.7
  Yarn: N/A
  pnpm: 11.0.6
Relevant Packages:
  next: 16.2.4
  eslint-config-next: N/A
  react: 19.2.5
  react-dom: 19.2.5
  typescript: N/A
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Turbopack

### Which stage(s) are affected? (Select all that apply)

next dev (local)

### Additional context

This seems related to pnpm v11's `enableGlobalVirtualStore: true`.

[...]

PR #94275, opened 2026-05-30
Improve Turbopack error for external next package symlinks

puneetdixit200:
### What?

Improve the recoverable Turbopack issue shown when `next/package.json` cannot be resolved and `node_modules/next` is a symlink to a package outside the configured Turbopack root.

### Why?

With package-manager layouts such as pnpm's global virtual store, the generic "Could not find the Next.js package" message does not explain that the symlink target is outside the root that Turbopack is allowed to compile. This adds a more specific cause and points users at `turbopack.root` or disabling pnpm's global virtual store for the project.

Refs #93556

### How?

- Detect a `node_modules/next` symlink during the missing-package issue path.
- If its canonical target is outside the canonical Turbopack root, include that target in the issue description.
- Add a development Turbopack regression test that simulates the external symlink without moving the package-manager store entry needed by `next dev` itself.

### Tests

- `cargo check -p next-core`
- `pnpm swc-build-native`
- `pnpm test-dev-turbo test/development/app-dir/concurrent-install/concurrent-install.test.ts`
- `git diff --check`

### DIFF
--- a/crates/next-core/src/next_import_map.rs
+++ b/crates/next-core/src/next_import_map.rs
@@ -1185,6 +1185,7 @@ pub async fn get_next_package(context_directory: FileSystemPath) -> Result<FileS
 struct MissingNextFolderIssue {
     path: FileSystemPath,
     root: FileSystemPath,
+    external_next_symlink: Option<RcStr>,
 }
 
 #[async_trait]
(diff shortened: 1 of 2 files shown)


stovmascript:
But I don't want to configure `turbopack.root` outside of my project.

PR #94283, opened 2026-05-31
fix(router-server): return 400 instead of 500 for malformed URLs when…

### LINKED ISSUE #94278: Malformed Request URL causes 500 response

https://github.com/Oudwins/nextjs-bugs/tree/bug/malformed-url-casuses-500

### To Reproduce

npm run build
npm run start
curl --path-as-is -i 'http://localhost:3000/50%%2050%%'
curl --path-as-is -i 'http://localhost:3000/50%25%2050%25'

Requirements: you must have a catch all route

The malformed request should return 400 Bad Request, not 500 Internal Server Error.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 36864
  Available CPU cores: 12
Binaries:
  Node: 25.8.1
  npm: 11.11.0
  Yarn: N/A
  pnpm: 10.32.1
Relevant Packages:
  next: 16.3.0-canary.35 // Latest available version is detected (16.3.0-canary.35).
  eslint-config-next: N/A
  react: 19.2.6
  react-dom: 19.2.6
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Error Handling

Other (Deployed), next start (local)

### Additional context

I have a patch that fixes the issue. But I have not read in depth into nextjs internals or made any contributions to next before. If someone guides me a little I'm happy to make the PR.

Here is the patch

```patch
diff --git a/dist/esm/server/lib/router-server.js b/dist/esm/server/lib/router-server.js
index 95fb3da7fc5bef160b7e49d3d573d745eb6422a6..f0bb1d030e80fbe1d1e70684cd50b11a92010356 100644
--- a/dist/esm/server/lib/router-server.js
+++ b/dist/esm/server/lib/router-server.js
@@ -461,16 +461,22 @@ async function initialize(opts) {
             try {
                 let invokePath = '/500';
                 let invokeStatus = '500';
                 if (err instanceof DecodeError) {
                     invokePath = '/400';
                     invokeStatus = '400';
                 } else {
                     console.error(err);
                 }
                 res.statusCode = Number(invokeStatus);
                 return await invokeRender(parseUrlUtil(invokePath), invokePath, 0, {
                     invokeStatus: res.statusCode
                 });
             } catch (err2) {
+                if (err instanceof DecodeError) {
+                    res.statusCode = 400;
+                    res.setHeader('Content-Type', 'text/plain; charset=utf-8');
+                    res.end('Bad Request');
+                    return;
+                }
                 console.error(err2);
             }
diff --git a/dist/server/lib/router-server.js b/dist/server/lib/router-server.js
index f2b00575c0f5f888f1b2d20ed4f8cb8fd46533c5..2890d68ec4edc3443e490c35ce17f1213f4c0699 100644
--- a/dist/server/lib/router-server.js
+++ b/dist/server/lib/router-server.js
@@ -516,16 +516,22 @@ async function initialize(opts) {
             try {
                 let invokePath = '/500';
                 let invokeStatus = '500';
                 if (err instanceof _utils.DecodeError) {
                     invokePath = '/400';
                     invokeStatus = '400';
                 } else {
                     console.error(err);
                 }
                 res.statusCode = Number(invokeStatus);
                 return await invokeRender((0, _parseurl.parseUrl)(invokePath), invokePath, 0, {
                     invokeStatus: res.statusCode
                 });
             } catch (err2) {
+                if (err instanceof _utils.DecodeError) {
+                    res.statusCode = 400;
+                    res.setHeader('Content-Type', 'text/plain; charset=utf-8');
+                    res.end('Bad Request');
+                    return;
+                }
                 console.error(err2);
             }
```

---

I accidentally created this same issue but with the repo being private. Fixed now... See: https://github.com/vercel/next.js/issues/94272#issuecomment-4584078567

vishalranaut:
## What?

Fixes #94278.

When a malformed URL triggers a `DecodeError`, Next.js attempts to render the `/400` page. In applications containing catch-all routes, rendering `/400` can itself fail, causing the request to fall through and return a 500 Internal Server Error.

This change adds a fallback path for `DecodeError` handling. If rendering the `/400` page fails, the server now responds with a plain-text `400 Bad Request` response instead of incorrectly returning a 500.

## Why?

Malformed URLs are client errors and should consistently produce a 400 response. Returning a 500 suggests a server-side failure and makes debugging more difficult for both users and operators.

## How?

* Keep the existing behavior of rendering `/400` for `DecodeError`.
* Add a fallback in the render-error path.
* If the original error is a `DecodeError`, respond directly with:

  * Status: `400`
  * Content-Type: `text/plain; charset=utf-8`
  * Body: `Bad Request`

## Testing

Added a regression test that:

1. Creates an application with a catch-all route.
2. Sends a request to a malformed encoded URL (`/50%%2050%%`).
3. Verifies the response status is `400`.
4. Ensures valid encoded URLs continue to behave normally.

## Risk

Low.

The change only affects the error fallback path after rendering the `/400` page has already failed. Normal request handling and existing custom 400 page rendering remain unchanged.


Oudwins:
Thank you! Wasn't sure best practice around creating a PR if I was meant to first create the issue, could submit a PR directly or what was the process

PR #94293, opened 2026-06-01
[turbopack] Treat assignments to `module.exports` as side-effect free

sampoder:
I noticed this section of the documentation in `turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs`: 

https://github.com/vercel/next.js/blob/89562628d27b672352d212ecd8c6941873e83222/turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs#L36

This PR implements this change; as part of a stack of PRs that attempts to work towards tracking local variable mutations.


mischnic:
That is unfortunately not always true:
```javascript
module.exports = {
  set foo(v) {
    console.log('side effect')
  },
}
module.exports.foo = 123
```

sampoder:
> That is unfortunately not always true:
> 
> ```js
> module.exports = {
>   set foo(v) {
>     console.log('side effect')
>   },
> }
> module.exports.foo = 123
> ```

:( - I believe I have patched this case though and added a test for it!

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -735,10 +866,26 @@ impl<'a> Visit for SideEffectVisitor<'a> {
...
+                // `module.exports`, `module.exports.x`) is the CJS equivalent of an
+                // ESM `export` declaration.
+                //
+                // An assigned value that carries an accessor will be marked as
+                // having side effects: it would attach a getter/setter to the 
+                // exports object, so a later member write (e.g. `module.exports.foo = 1`)
+                // could invoke the accessor. Running the accessor could have side effects.
+                if assign.op == AssignOp::Assign
+                    && self.is_cjs_export_target(&assign.left)
+                    && !contains_accessor(&assign.right)
```
lukesandberg: contains_setter might be a better name

lukesandberg: contains_getters_or_setters for explicitness

property_access_might_have_side_effecs for policy

sampoder: went with `contains_getters_or_setters`!

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -267,13 +267,164 @@ static KNOWN_PURE_REGEXP_PROTOTYPE_METHODS: phf::Set<&'static str> = phf_set! {
...
+    any_top_level_assign(program, |assign| {
+        if assign.op != AssignOp::Assign {
+            return false;
+        }
+        let AssignTarget::Simple(SimpleAssignTarget::Member(member)) = &assign.left else {
+            return false;
+        };
+        is_cjs_export_member(member, unresolved_mark) && contains_getters_or_setters(&assign.right)
+    })
+}
```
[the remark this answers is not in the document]
sampoder: This was resolved, however, it meant that I switched back to a visitor.

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -267,13 +267,160 @@ static KNOWN_PURE_REGEXP_PROTOTYPE_METHODS: phf::Set<&'static str> = phf_set! {
...
+    }
+    impl<F: FnMut(&AssignExpr)> Visit for Collector<F> {
+        noop_visit_type!();
+        fn visit_assign_expr(&mut self, n: &AssignExpr) {
+            (self.f)(n);
+            n.visit_children_with(self);
+        }
+        // Function/method bodies do not execute during module evaluation.
+        fn visit_function(&mut self, _: &Function) {}
+        fn visit_arrow_expr(&mut self, _: &ArrowExpr) {}
```
lukesandberg, line 365: there can be 'top level assignments' in strange places

just make sure we have coverage for
```
class {
  static {
      foo = bar ;// top level
  }
  constructor() {
     baz = quuz; // not top level
 }
}
```

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -345,7 +505,7 @@ impl<'a> SideEffectVisitor<'a> {
             Callee::Expr(expr) => {
                 let expr = unparen(expr);
                 if let Expr::Ident(ident) = expr {
-                    ident.ctxt.outer() == self.unresolved_mark && ident.sym.as_ref() == "require"
+                    is_unresolved(ident, "require", self.unresolved_mark)
```
lukesandberg: `is_global` might be a better name?  `unresolved` is accurate but a little misleading

also maybe make it a method `self.is_global(ident, require)`?

lukesandberg (approved):

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
sampoder: added tests to cover these and confirmed they pass

PR #94294, opened 2026-06-01
[turbopack] Treat local `const` assignments as side-effect free

sampoder:
This is a follow-up to this comment: https://github.com/vercel/next.js/blob/89562628d27b672352d212ecd8c6941873e83222/turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs#L22

I wouldn't consider it a comprehensive tracker of local variables. However, it's a start. Only `const`s are tracked at the moment. This is because with `let`s you can do things like:

```javascript
let x = { a: "b" };
x = globalThis;
x.a = "b" // side effect!
```
I do think this is a helpful improvement, though as it handles the common pattern that was mentioned in the comment.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
@@ -19,11 +19,10 @@
 //! has side effects. This is safe for tree-shaking purposes as it prevents
 //! incorrectly removing code that might be needed, and can simply be improved over time.
 //!
-//! ## Future Enhancement: Local Variable Mutation Tracking
+//! ## Local Variable Mutation Tracking
 //!
-//! Currently, all assignments, updates, and property mutations are treated as side effects.
-//! However, mutations to locally-scoped variables that never escape the module evaluation scope
-//! could be considered side-effect free. This would handle common patterns like:
+//! Currently, assignments to local unaliased constants and `module.exports` are considered
+//! side-effect free. This handles the common pattern:
 //!
 //! ```javascript
 //! // Currently marked as having side effects, but could be pure:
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -285,6 +287,54 @@ fn is_object_or_array_literal(expr: &Expr) -> bool {
...
+/// c.fetch = sideEffects();
+/// ```
+///
+/// Has side-effects.
+fn collect_safe_assignment_constant_ids(program: &Program) -> HashSet<Id> {
+    struct Collector {
+        ids: HashSet<Id>,
+    }
+    impl Visit for Collector {
+        noop_visit_type!();
```
lukesandberg, line 318: we are only interested in variables defined in the root scope.... does this traverse all functions and local scopes

e.g. we shouldn't track function local `const`

lukesandberg: the normal visitor does avoid traversing into, so you could just compute this as you go in the theory that we must have visited the `const` declaration before visiting an assignment to the same id.

This isn't strictly guaranteed due to things like
```
function f() {  a.b =1;}
const a = {}
f()
```

but i think we would mark such code effectful anyway due to the function call so it should be fine

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -471,6 +544,8 @@ impl<'a> SideEffectVisitor<'a> {
             return false;
         }
         self.is_cjs_export_member(member)
+            || matches!(root_identifier(&member.obj), Some(root)
+                if self.safe_assignment_constant_ids.contains(&root.to_id()))
```
lukesandberg: can you break this out into its own if statement

putting the `if` inside the `matches!` is a little too cute for me

lukesandberg (approved):

PR #94296, opened 2026-06-01
[turbopack] Add `BumpVec` (vector implementation that takes a `Bump`)

sampoder:
This PR implements `BumpVec` this is a custom vector implementation that takes in a `Bump` as a type (from bumpalo) so that we get `Send`/`Sync` on our `JSValue` vectors when we make that switch in: https://github.com/vercel/next.js/pull/94297.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
@@ -2,14 +2,15 @@
 //!
 //! `JsValue` analysis allocates all of its nodes into a per-thread [`Bump`](bumpalo::Bump) that is
 //! freed in one shot when analysis finishes. For the list children that grow or are rebuilt after
-//! construction, this module provides [`BumpVec`]: a `Send`/`Sync` growable vector that stores
-//! nothing but an arena-allocated buffer and a length, with no `unsafe impl`.
+//! construction, this module provides [`BumpVec`]: a `Send`/`Sync` growable vector that stores a
+//! raw pointer into the arena, a capacity, and a length.
 
 use std::{
     alloc::Layout,
     fmt,
     hash::{Hash, Hasher},
-    mem::{ManuallyDrop, MaybeUninit},
+    marker::PhantomData,
+    mem::ManuallyDrop,
     ops::{Deref, DerefMut},
     ptr::{self, NonNull},
 };
@@ -21,14 +22,23 @@ use bumpalo::Bump;
 /// construction (e.g. `Array.items`, `Object.parts`, `Alternatives.values`, `Add` operands, and the
 /// `Call`/`MemberCall` lists).
 ///
-/// It holds nothing but an arena-allocated buffer and a length, so it is `Send`/`Sync` (a
-/// `&'a mut [T]` is, when `T` is) with no `unsafe impl`. The growth methods take the `&'a Bump` to
-/// allocate from. The only `unsafe` here is the localized `MaybeUninit` bookkeeping every growable
-/// buffer needs.
+/// It owns its `T` elements like `Vec<T>` but stores them through a raw pointer into the arena,
+/// alongside a capacity and a length. The raw pointer makes it neither `Send` nor `Sync` on its
+/// own, so those are declared via `unsafe impl` (sound because it owns the `T`s exactly like a
+/// `Vec<T>`). The growth methods take the `&'a Bump` to allocate from.
 pub struct BumpVec<'a, T> {
-    buf: &'a mut [MaybeUninit<T>],
+    /// Points at `cap` allocated (possibly uninitialized) `T` slots; dangling when `cap == 0`.
+    ptr: NonNull<T>,
+    cap: usize,
     len: usize,
+    /// Binds the arena lifetime and signals ownership/variance over `T` to the compiler.
+    _marker: PhantomData<&'a mut [T]>,
 }
+
+// SAFETY: `BumpVec` owns its `T` elements just like `Vec<T>`; the raw pointer is merely an owning
+// handle into the arena, so it is `Send`/`Sync` exactly when `T` is.
+unsafe impl<T: Send> Send for BumpVec<'_, T> {}
+unsafe impl<T: Sync> Sync for BumpVec<'_, T> {}
 
 impl<T> Default for BumpVec<'_, T> {
     fn default() -> Self {
@@ -39,15 +49,31 @@ impl<T> Default for BumpVec<'_, T> {
 impl<'a, T> BumpVec<'a, T> {
     pub fn new() -> Self {
         Self {
-            buf: &mut [],
+            ptr: NonNull::dangling(),
+            cap: 0,
             len: 0,
+            _marker: PhantomData,
         }
+    }
+
+    /// Allocate room for `cap` uninitialized `T` slots from the arena, or a dangling pointer when
+    /// `cap == 0` (no allocation needed).
+    fn alloc_uninitialized(bump: &'a Bump, cap: usize) -> NonNull<T> {
+        if cap == 0 {
+            return NonNull::dangling();
+        }
+        let layout = Layout::array::<T>(cap).expect("capacity overflow");
+        bump.allocate(layout)
+            .expect("bump allocation failed")
+            .cast::<T>()
     }
 
     pub fn with_capacity_in(bump: &'a Bump, capacity: usize) -> Self {
         Self {
-            buf: bump.alloc_slice_fill_with(capacity, |_| MaybeUninit::uninit()),
+            ptr: Self::alloc_uninitialized(bump, capacity),
+            cap: capacity,
             len: 0,
+            _marker: PhantomData,
         }
     }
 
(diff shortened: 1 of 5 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/arena.rs
```diff
@@ -0,0 +1,135 @@
...
+
+impl<T: fmt::Debug> fmt::Debug for ArenaVec<'_, T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&*self.0, f)
+    }
+}
+
+/// A bump-allocation arena scoped to a single module analysis.
+pub struct Arena(Bump);
+unsafe impl Sync for Arena {}
```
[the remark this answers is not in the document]
sampoder: If we can't do this the unfortunately I don't think we can use `bumpalo` as this is a limitation it has.

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,377 @@
...
+    pub fn from_iter_in(bump: &'a Bump, iter: impl IntoIterator<Item = T>) -> Self {
+        let iter = iter.into_iter();
+        let mut vec = Self::with_capacity_in(bump, iter.size_hint().0);
+        vec.extend(bump, iter);
+        vec
+    }
+
+    /// Reallocate the buffer to `new_cap` elements (`new_cap >= len`), moving the live prefix into
+    /// the fresh arena allocation. The old buffer is abandoned (the arena frees it in bulk).
+    fn realloc_to(&mut self, bump: &'a Bump, new_cap: usize) {
```
bgw: make this unsafe, because we don't verify that `new_cap >= self.len` in release builds

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,377 @@
...
+        debug_assert!(new_cap >= self.len);
+        let new_buf = bump.alloc_slice_fill_with(new_cap, |_| MaybeUninit::uninit());
+        for (dst, src) in new_buf.iter_mut().zip(self.buf.iter()).take(self.len) {
```
bgw: This should use `Allocator::grow` from the `allocator_api2` crate or from the stdlib with the nightly feature enabled (either are fine).

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+        self.buf[self.len].write(value);
+        self.len += 1;
+    }
+
+    pub fn extend(&mut self, bump: &'a Bump, iter: impl IntoIterator<Item = T>) {
+        for value in iter {
+            self.push(bump, value);
+        }
+    }
```
bgw, line 146: If we can, we should implement `extend_from_slice` instead, and use memcopy to implement that.

This is also not ideal as-implemented because it could lead to multiple reallocs, when there should only be one.

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// Split the vec in two at `at`: `self` retains the prefix `[0, at)` and the returned vec owns
+    /// the suffix `[at, len)`, moved into a fresh arena allocation.
+    pub fn split_off(&mut self, bump: &'a Bump, at: usize) -> Self {
+        assert!(at <= self.len, "split_off index out of bounds");
+        let tail_len = self.len - at;
+        let mut tail = Self::with_capacity_in(bump, tail_len);
+        for (dst, src) in tail.buf.iter_mut().zip(self.buf[at..self.len].iter()) {
+            // SAFETY: indices `[at, len)` are initialized; move each element into `tail`.
+            dst.write(unsafe { ptr::read(src.as_ptr()) });
+        }
```
bgw: This should use a non-overlapping memcopy too

bgw (changes requested):

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,397 @@
...
+    pub fn new() -> Self {
+        Self {
+            buf: &mut [],
+            len: 0,
+        }
+    }
+
+    pub fn with_capacity_in(bump: &'a Bump, capacity: usize) -> Self {
+        Self {
+            buf: bump.alloc_slice_fill_with(capacity, |_| MaybeUninit::uninit()),
```
bgw: This is potentially doing some work to copy data to the slice, we can use the `unsafe` `Allocator::allocate` API

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,397 @@
...
+/// A minimal growable vector for the list children of a `JsValue` that grow or are rebuilt after
+/// construction (e.g. `Array.items`, `Object.parts`, `Alternatives.values`, `Add` operands, and the
+/// `Call`/`MemberCall` lists).
+///
+/// It holds nothing but an arena-allocated buffer and a length, so it is `Send`/`Sync` (a
+/// `&'a mut [T]` is, when `T` is) with no `unsafe impl`. The growth methods take the `&'a Bump` to
+/// allocate from. The only `unsafe` here is the localized `MaybeUninit` bookkeeping every growable
+/// buffer needs.
+pub struct BumpVec<'a, T> {
+    buf: &'a mut [MaybeUninit<T>],
```
bgw: If we're using the unsafe allocator APIs for everything, this should probably just be a pointer, likely a `NonNull<T>`

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// is the arena's most recent allocation, [`Allocator::grow`] extends it in place.
+    unsafe fn realloc_to(&mut self, bump: &'a Bump, new_cap: usize) {
+        debug_assert!(new_cap >= self.len);
+        let new_layout = Layout::array::<MaybeUninit<T>>(new_cap).expect("capacity overflow");
+        let new_data = if self.buf.is_empty() {
+            // Nothing allocated yet, so there is no prior block to grow from.
+            bump.allocate(new_layout)
+        } else {
+            // The only caller (`push`) reallocates when full, so `buf.len() == self.len`: every
+            // slot in the old buffer is initialized and `grow` moves them all into the new block.
```
bgw: This comment won't be valid if you fix `extend`/`extend_from_slice` to reallocate before pushing multiple items.

But also moving uninitialized memory is fine?

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,397 @@
...
+            let old_layout =
+                Layout::array::<MaybeUninit<T>>(self.buf.len()).expect("capacity overflow");
+            let old_data = NonNull::from(&mut *self.buf).cast::<u8>();
+            // SAFETY: `old_data`/`old_layout` describe the current arena buffer and `new_layout` is
+            // strictly larger, so growing it (and moving the bytes) is sound.
```
bgw: https://doc.rust-lang.org/std/alloc/trait.Allocator.html#method.grow

> ptr must denote a block of memory [currently allocated](https://doc.rust-lang.org/std/alloc/trait.Allocator.html#currently-allocated-memory) via this allocator.

Double check that it's okay for bumpalo to grow data from a different arena. We can't actually guarantee the arena is the same here, but I imagine that `Bump` is simple enough that it doesn't actually care about this.

Include some test coverage for this.

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,397 @@
...
+    fn into_iter(self) -> IntoIter<'a, T> {
+        // Disable `Drop` (it would also drop the elements) and move the buffer out.
+        let me = ManuallyDrop::new(self);
+        // SAFETY: we move the `&mut` buffer out of the (forgotten) `BumpVec`; it is not aliased.
+        IntoIter {
+            buf: unsafe { ptr::read(&me.buf) },
+            len: me.len,
+            idx: 0,
+        }
+    }
```
bgw, line 223: If there's some easy way for us to convert ourselves into a `Box<[T]>`, maybe we could use https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-IntoIterator-for-Box%3C%5BI%5D,+A%3E

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,377 @@
...
+        unsafe { std::slice::from_raw_parts_mut(self.buf.as_mut_ptr() as *mut T, self.len) }
+    }
+}
+
+impl<T> Drop for BumpVec<'_, T> {
+    fn drop(&mut self) {
+        // SAFETY: `0..len` is initialized; drop each element (the arena only frees memory).
+        unsafe { ptr::drop_in_place(&mut **self) }
+    }
+}
```
bgw, line 209: Potential follow up: See if we can eliminate reference-counted types in `JsValue` and develop a unit test that checks that `std::mem::needs_drop::<JsValue>() == false`.

bgw (changes requested):

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
sampoder: I had a look into this, it requires passing an allocator so this would look something like:

```rust
let boxed: Box<[T], NoOpAllocator> = unsafe { Box::from_raw_in(slice, NoOpAllocator) };
boxed.into_iter()
```
And `NoOpAllocator` does nothing on deallocations and errors if you attempt to allocate:

```rust
unsafe impl std::alloc::Allocator for NoOpAllocator {
    fn allocate(&self, _layout: Layout) -> Result<NonNull<[u8]>, std::alloc::AllocError> {
        Err(std::alloc::AllocError)
    }
    unsafe fn deallocate(&self, _ptr: NonNull<u8>, _layout: Layout) {}
}
```

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,517 @@
...
+/// alongside a capacity and a length. The raw pointer makes it neither `Send` nor `Sync` on its
+/// own, so those are declared via `unsafe impl` (sound because it owns the `T`s exactly like a
+/// `Vec<T>`). The growth methods take the `&'a Bump` to allocate from.
+pub struct BumpVec<'a, T> {
+    /// Points at `cap` allocated (possibly uninitialized) `T` slots; dangling when `cap == 0`.
+    ptr: NonNull<T>,
+    cap: usize,
+    len: usize,
+    /// Binds the arena lifetime and signals ownership/variance over `T` to the compiler.
+    _marker: PhantomData<&'a mut [T]>,
```
bgw:
```suggestion
    /// Binds the arena lifetime and signals ownership/variance over `T` to the compiler.
    _marker: PhantomData<&'a ()>,
```

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -0,0 +1,517 @@
...
+            cap: capacity,
+            len: 0,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Collect `iter` into a growable [`BumpVec`].
+    pub fn from_iter_in(bump: &'a Bump, iter: impl IntoIterator<Item = T>) -> Self {
+        let iter = iter.into_iter();
+        let mut vec = Self::with_capacity_in(bump, iter.size_hint().0);
```
bgw: Maybe prefer the upper bound if it exists

bgw (approved):

PR #94297, opened 2026-06-01
[turbopack] Use an arena for `JSValue`s

sampoder:
This PR takes the arena introduced in https://github.com/vercel/next.js/pull/94296 and uses it! It's large but shouldn't change functionality beyond the use of the arena. Some syntax related to pattern-matching had to change because `bumpalo`'s `Box` doesn't support the same pattern matching that the standard library's `Box` does but the code should be equivalent.

`bumpalo`'s `Bump` type implements `Send` but not `Sync`, to allow us to pass a `Bump` into async-methods we wrap it in `ThreadLocal` (see https://docs.rs/thread_local/latest/thread_local/). `ThreadLocal` implements `Sync` by creating a new bumpalo allocator for each thread. We've tried to minimise the usage of `ThreadLocal` to just where it was needed. The downside of `ThreadLocal` is that we may find ourselves allocating memory for multiple arenas instead of just one. This is more time allocating if there are multiple threads compared to one shared allocation, however, it still appears to be a performance improvement over individually allocating memory for each `JSValue`

### DIFF
--- a/crates/next-core/src/segment_config.rs
+++ b/crates/next-core/src/segment_config.rs
@@ -30,7 +30,9 @@ use turbopack_core::{
 };
 use turbopack_ecmascript::{
     EcmascriptInputTransforms, EcmascriptModuleAssetType,
-    analyzer::{ConstantNumber, ConstantValue, JsValue, ObjectPart, graph::EvalContext},
+    analyzer::{
+        Bump, ConstantNumber, ConstantValue, JsValue, ObjectPart, ThreadLocal, graph::EvalContext,
+    },
     parse::{ParseResult, parse},
 };
 
@@ -413,12 +415,25 @@ pub async fn parse_segment_config_from_source(
         return Ok(Default::default());
     };
 
+    // Arena for the `JsValue`s produced while evaluating config expressions;
+    // freed when this function returns.
+    let arena = ThreadLocal::new();
     let config = WrapFuture::new(
         async {
             let mut config = NextSegmentConfig::default();
 
             let mut parse = async |ident, init, span| {
-                parse_config_value(source, mode, &mut config, eval_context, ident, init, span).await
+                parse_config_value(
+                    source,
+                    mode,
+                    &mut config,
+                    eval_context,
+                    &arena,
+                    ident,
+                    init,
+                    span,
+                )
+                .await
             };
 
             for item in &module_ast.body {
@@ -575,7 +590,7 @@ async fn invalid_config(
     key: &str,
     span: Span,
     error: RcStr,
-    value: Option<&JsValue>,
+    value: Option<&JsValue<'_>>,
     severity: IssueSeverity,
 ) -> Result<()> {
     let detail = if let Some(value) = value {
@@ -604,6 +619,7 @@ async fn parse_config_value(
     mode: ParseSegmentMode,
     config: &mut NextSegmentConfig,
     eval_context: &EvalContext,
+    arena: &ThreadLocal<Bump>,
     key: Cow<'_, str>,
     init: Option<Cow<'_, Expr>>,
     span: Span,
@@ -619,17 +635,18 @@ async fn parse_config_value(
             | Some(Expr::TsSatisfies(TsSatisfiesExpr { expr, .. })) => Some(&**expr),
             _ => init,
         };
-        init.map(|init| eval_context.eval(init)).map(|v| {
-            // Special case, as we don't call `link` here: assume that `undefined` is a free
-            // variable.
-            if let JsValue::FreeVar(name) = &v
-                && name == "undefined"
-            {
-                JsValue::Constant(ConstantValue::Undefined)
-            } else {
-                v
-            }
-        })
+        init.map(|init| eval_context.eval(arena.get_or_default(), init))
+            .map(|v| {
+                // Special case, as we don't call `link` here: assume that `undefined` is a free
+                // variable.
+                if let JsValue::FreeVar(name) = &v
+                    && name == "undefined"
+                {
+                    JsValue::Constant(ConstantValue::Undefined)
+                } else {
+                    v
+                }
+            })
     };
 
     match &*key {
@@ -1001,7 +1018,7 @@ async fn parse_static_string_or_array_from_js_value(
     span: Span,
     key: &str,
     sub_key: &str,
-    value: &JsValue,
+    value: &JsValue<'_>,
 ) -> Result<Option<Vec<RcStr>>> {
     Ok(match value {
         // Single value is turned into a single-element Vec.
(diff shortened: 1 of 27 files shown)


crates/next-core/src/segment_config.rs
```diff
@@ -600,11 +615,12 @@ async fn invalid_config(
     Ok(())
 }
 
-async fn parse_config_value(
+async fn parse_config_value<'a>(
     source: ResolvedVc<Box<dyn Source>>,
     mode: ParseSegmentMode,
     config: &mut NextSegmentConfig,
     eval_context: &EvalContext,
+    arena: &'a ThreadLocal<Bump>,
```
bgw:
```suggestion
async fn parse_config_value(
    source: ResolvedVc<Box<dyn Source>>,
    mode: ParseSegmentMode,
    config: &mut NextSegmentConfig,
    eval_context: &EvalContext,
    arena: &ThreadLocal<Bump>,
```

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/effects.rs
```diff
@@ -110,40 +113,40 @@ impl EffectArg {
 }
 
 #[derive(Debug)]
-pub enum Effect {
+pub enum Effect<'a> {
```
bgw, line 117: Follow up: Maybe also make `Effect` use the arena for everything? Right now it has some global-heap-allocated types.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -28,18 +28,18 @@ use crate::{
     utils::{AstPathRange, unparen},
 };
 
-enum EarlyReturn {
+enum EarlyReturn<'a> {
```
bgw, line 32: Possible future PR: Also use an arena here

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1048,7 +1073,7 @@ impl Analyzer<'_> {
     }
 }
 
-impl VisitAstPath for Analyzer<'_> {
+impl<'a, 'e> VisitAstPath for Analyzer<'a, 'e> {
```
bgw:
```suggestion
impl VisitAstPath for Analyzer<'_, '_> {
```

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -75,7 +77,7 @@ pub(super) struct Analyzer<'a> {
     /// slightly less correct circular import errors) for EsmModuleItem
     pub(super) supports_block_scoping: bool,
 
-    pub(super) eval_context: &'a EvalContext,
+    pub(super) eval_context: &'e EvalContext,
```
bgw: Rename `'a` here to `'arena` and `'e` here to `'eval` for clarity

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -2036,7 +2064,8 @@ impl VisitAstPath for Analyzer<'_> {
         ast_path: &mut swc_core::ecma::visit::AstNodePath<'r>,
     ) {
         if n.op == UnaryOp::TypeOf && self.analyze_mode.is_code_gen() {
-            let arg_value = Box::new(self.eval_context.eval(&n.arg));
+            let arg_value =
+                bumpalo::boxed::Box::new_in(self.eval_context.eval(self.arena, &n.arg), self.arena);
```
bgw: Import these as `BumpBox`, there's enough of them this seems more annoying than it's worth

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -2141,23 +2173,24 @@ impl VisitAstPath for Analyzer<'_> {
     }
 }
 
-impl Analyzer<'_> {
+impl<'a, 'e> Analyzer<'a, 'e> {
```
bgw:
```suggestion
impl<'a> Analyzer<'a, '_> {
```

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/mod.rs
```diff
@@ -1089,7 +1174,14 @@ impl JsValue {
...
 
 // Unknown management
-impl JsValue {
+impl<'a> JsValue<'a> {
+    /// Move the value out of an arena-allocated box, leaving a default ([`JsValue::unknown_empty`])
+    /// in the box's slot. Needed because `bumpalo::boxed::Box`, unlike `std::Box`, does not support
+    /// moving its contents out via `*box`.
+    pub fn unbox(mut boxed: bumpalo::boxed::Box<'a, JsValue<'a>>) -> JsValue<'a> {
+        take(&mut *boxed)
+    }
```
bgw: `JsValue::unbox(boxed)` is more characters than `take(&mut *boxed)`. Let's just inline the usage everywhere that calls this.

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/normalize.rs
```diff
@@ -73,18 +75,20 @@ impl JsValue {
                     if set.len() == 1 {
                         *self = set.into_iter().next().unwrap().0;
                     } else {
-                        *values = set.into_iter().map(|v| v.0).collect();
+                        values.extend(arena, set.into_iter().map(|v| v.0));
```
bgw: `values` points to an empty `BumpVec` at this point. It's probably cleaner to just do `*values = BumpVec::from_iter_in(...)` here.

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/normalize.rs
```diff
@@ -73,18 +75,20 @@ impl JsValue {
...
                         self.update_total_nodes();
                     }
                 }
             }
             JsValue::Concat(_, v) => {
-                // Remove empty strings
-                v.retain(|v| v.as_str() != Some(""));
-
                 // TODO(kdy1): Remove duplicate
                 let mut new: Vec<JsValue> = vec![];
```
bgw: Allocate this vec on the arena

OR... if we can't do a good size of estimating a reasonable maximum size/capacity and we're concerned about all the bumpalo resizes that might happen, this could be a good candidate for a SmallVec (https://docs.rs/smallvec/latest/smallvec/) with a fairly generous on-stack size and that would at least eliminate the heap allocation in the common case.

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/normalize.rs
```diff
@@ -104,21 +108,22 @@ impl JsValue {
                 if new.len() == 1 {
                     *self = new.into_iter().next().unwrap();
                 } else {
-                    *v = new;
+                    v.extend(arena, new);
```
bgw: This should go back to `*v = new` after switching `new` to a `BumpVec`.

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/normalize.rs
```diff
@@ -104,21 +108,22 @@ impl JsValue {
...
             }
             JsValue::Add(_, v) => {
                 let mut added: Vec<JsValue> = Vec::new();
-                let mut iter = take(v).into_iter();
+                let taken = take(v);
+                let mut iter = taken.into_iter();
                 while let Some(item) = iter.next() {
                     if item.is_string() == Some(true) {
-                        let mut concat = match added.len() {
+                        let mut concat: Vec<JsValue> = match added.len() {
```
bgw: Similar thing here. We should use a `BumpVec` or a `SmallVec` to avoid a heap allocation in the common case.

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/predicates.rs
```diff
@@ -101,9 +101,9 @@ impl JsValue {
                     shortcircuit_if_known(list, JsValue::is_not_nullish, JsValue::is_truthy)
                 }
             },
-            JsValue::Binary(_, box a, op, box b) => {
+            JsValue::Binary(_, a, op, b) => {
```
bgw, line 104: RIP box patterns

turbopack/crates/turbopack-ecmascript/src/analyzer/well_known/mod.rs
```diff
@@ -43,34 +45,45 @@ pub async fn replace_well_known(
...
+                Some("map") => (
+                    JsValue::WellKnownFunction(WellKnownFunctionKind::ArrayMap),
+                    true,
+                ),
+                _ => (
+                    JsValue::member(arena.get_or_default(), take(&mut *obj), take(&mut *prop)),
+                    false,
+                ),
+            }
         }
```
bgw, line 85: We want to avoid reconstructing the member at the end with `JsMember` because that has to create new `BumpBox`es and copy data, and there's no good reason to so because we already have the `JsMember` as `value`.

We just have to be careful not to move `obj` or `prop` out of `value` in the `match` expression so that we can use `value` later.

```suggestion
        JsValue::Member(_, ref obj, ref prop) if matches!(**obj, JsValue::Array { .. }) => {
            match prop.as_str() {
                Some("filter") => (
                    JsValue::WellKnownFunction(WellKnownFunctionKind::ArrayFilter),
                    true,
                ),
                Some("forEach") => (
                    JsValue::WellKnownFunction(WellKnownFunctionKind::ArrayForEach),
                    true,
                ),
                Some("map") => (
                    JsValue::WellKnownFunction(WellKnownFunctionKind::ArrayMap),
                    true,
                ),
                _ => (value, false),
            }
        }
```

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -77,18 +77,17 @@ pub fn early_replace_builtin(value: &mut JsValue) -> bool {
...
+            if let JsValue::Unknown {
+                has_side_effects, ..
+            } = &**obj
+            {
+                let side_effects = *has_side_effects || prop.has_side_effects();
+                value.make_unknown(side_effects, rcstr!("unknown object"));
+                true
+            } else {
+                false
+            }
```
bgw, line 91: We can just put the `if` on the match arm as a guard.

```suggestion
        JsValue::Member(_, obj, prop)
            if let JsValue::Unknown {
                has_side_effects, ..
            } = &**obj =>
        {
            let side_effects = *has_side_effects || prop.has_side_effects();
            value.make_unknown(side_effects, rcstr!("unknown object"));
            true
```

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -151,14 +154,21 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
...
                             if index < items.len() {
-                                *value = items.swap_remove(index);
+                                // Equivalent to `swap_remove(index)`: move the last item into
+                                // the slot and return the displaced one.
+                                let last = items.pop().unwrap();
+                                *value = if index < items.len() {
+                                    std::mem::replace(&mut items[index], last)
+                                } else {
+                                    last
+                                };
```
bgw: Let's just add the `swap_remove` method onto `BumpVec` using this implementation.

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -230,28 +242,30 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
...
+                                arena,
+                                JsValue::object(BumpVec::new()),
+                                take(prop),
                             ),
                             true,
                             rcstr!("unknown object prototype methods or values"),
                         ));
                     }
-                    JsValue::alternatives(values)
+                    JsValue::alternatives(BumpVec::from_iter_in(arena, values))
```
bgw: This is constructing an intermediate vec and then throwing it away and converting it to a `BumpVec`. Let's just make a `BumpVec`.

```
                    let mut values = BumpVec::with_capacity_in(
                        arena, parts.len() + if include_unknown { 1 } else { 0 }
                    );
                    for part in parts {
                        match part {
                            ObjectPart::KeyValue(_, value) => {
                                values.push(arena, take(value));
                            }
                            ObjectPart::Spread(_) => {
                                values.push(arena, JsValue::unknown(
                                    JsValue::member(
                                        arena,
                                        JsValue::object(BumpVec::from_iter_in(arena, [take(part)])),
                                        prop.clone_in(arena),
                                    ),
                                    true,
                                    rcstr!("spread object"),
                                ));
                            }
                        }
                    }
                    if include_unknown {
                        values.push(arena, JsValue::unknown(
                            JsValue::member(
                                arena,
                                JsValue::object(BumpVec::new()),
                                take(prop),
                            ),
                            true,
                            rcstr!("unknown object prototype methods or values"),
                        ));
                    }
                    JsValue::alternatives(values)

```

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
```diff
@@ -230,28 +242,30 @@ pub fn replace_builtin(value: &mut JsValue) -> bool {
...
-                    parts: &mut Vec<ObjectPart>,
-                    prop: &mut Box<JsValue>,
+                    parts: &mut BumpVec<'a, ObjectPart<'a>>,
+                    prop: &mut JsValue<'a>,
                     include_unknown: bool,
-                ) -> JsValue {
+                ) -> JsValue<'a> {
                     // Note: potential_values are already in reverse order
-                    let mut potential_values = take(parts)
+                    let mut potential_values: Vec<ObjectPart<'a>> = take(parts)
```
bgw: Two options:
- More ideal: Switch `parts_to_alternatives` to accept an `impl IntoIterator<Item = ObjectPart<'a>>`, so that we don't need to collect here at all. It doesn't look like `parts_to_alternatives` actually benefits from having a slice instead of an iterator, aside from a stronger guarantee of a size bound.
- Less ideal: Switch this from `Vec` to `SmallVec` with a reasonable on-stack size so that most of the time this at least doesn't do a heap allocation.

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
bgw: This is a temporary data structure. Let's make this a `SmallVec`.

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/normalize.rs
sampoder: 👍 to avoid the resizes, I did:

```rust
 BumpVec::with_capacity_in(arena, taken.len())
```

(`taken`'s length should be an upper bound)

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
sampoder: Oh good catch, yep!

turbopack/crates/turbopack-ecmascript/src/analyzer/builtin.rs
sampoder: Went with the `impl IntoIterator<Item = ObjectPart<'a>>` option

lukesandberg (dismissed):
it would be useful to describe the `thread_local!` design in the PR description, how this is solving a problem with async and the associated risks

turbopack/crates/turbopack-ecmascript/src/analyzer/well_known/mod.rs
```diff
@@ -241,18 +269,22 @@ fn path_join(args: Vec<JsValue>) -> JsValue {
...
+                arena,
+                [rcstr!("/").into(), rcstr!("").into()],
+            )));
         }
         results.push(part);
         last_is_str = is_str;
     }
-    JsValue::concat(results)
+    JsValue::concat(BumpVec::from_iter_in(arena, results))
 }
```
bgw, line 282: Let's avoid some heap allocation:

```suggestion
fn path_join<'a>(arena: &'a Bump, args: BumpVec<'a, JsValue<'a>>) -> JsValue<'a> {
    if args.is_empty() {
        return rcstr!(".").into();
    }
    let mut results_final: SmallVec<JsValue<'a>, [_; 16]> = SmallVec::new();
    let mut results: SmallVec<JsValue<'a>, [_; 16]> = SmallVec::new();
    for arg in args {
        let arg_parts = if let Some(str) = arg.as_str() {
            let split = str.split('/');
            Either::Left(split.map(|s| s.into()))
        } else {
            Either::Right(iter::once(arg))
        };
        for item in arg_parts {
            if let Some(str) = item.as_str() {
                match str {
                    "" | "." => {
                        if results_final.is_empty() && results.is_empty() {
                            results_final.push(item);
                        }
                    }
                    ".." => {
                        if results.pop().is_none() {
                            results_final.push(item);
                        }
                    }
                    _ => results.push(item),
                }
            } else {
                results_final.append(&mut results);
                results_final.push(item);
            }
        }
    }
    results_final.append(&mut results);
    let mut iter = results_final.into_iter();
    let first = iter.next().unwrap();
    let mut last_is_str = first.as_str().is_some();
    results.push(first);
    for part in iter {
        let is_str = part.as_str().is_some();
        if last_is_str && is_str {
            results.push(rcstr!("/").into());
        } else {
            results.push(JsValue::alternatives(BumpVec::from_iter_in(
                arena,
                [rcstr!("/").into(), rcstr!("").into()],
            )));
        }
        results.push(part);
        last_is_str = is_str;
    }
    JsValue::concat(BumpVec::from_iter_in(arena, results))
}
```

Follow-up: It would be good to add some unit tests for this, because it's very confusing, and `results_final` is a terrible name.

turbopack/crates/turbopack-ecmascript/src/analyzer/well_known/mod.rs
```diff
@@ -241,18 +269,22 @@ fn path_join(args: Vec<JsValue>) -> JsValue {
...
 
-fn path_resolve(cwd: JsValue, mut args: Vec<JsValue>) -> JsValue {
+fn path_resolve<'a>(
```
bgw, line 284: Let's to the same optimizations to this.

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -280,8 +293,15 @@ impl<T> Iterator for IntoIter<'_, T> {
...
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let remaining = self.len - self.idx;
+        (remaining, Some(remaining))
+    }
 }
 
+impl<T> ExactSizeIterator for IntoIter<'_, T> {}
+
```
bgw, line 304: Nice, this should help let `.collect()` allocate the correct-sized `Vec`/`BumpVec` in more cases.

bgw (approved):
All the remaining comments can be handled as follow-ups. :shipit:  🚢

PR #94319, opened 2026-06-01
[ci] Reduce usage of turborepo cache in `build_and_deploy.yml`

bgw:
We're consolidating on `actions/cache` now that we got rid of self-hosted runners. This doesn't remove all of the uses of turborepo's cache (notably sccache), but this does eliminate the remote turborepo cache for the native build cache and the docker image.

These build steps use docker so that we're building against an old pinned glibc (2.31), and so that we can have access to both `aarch64` and `x86_64` musl sysroots.

## Testing

**Local Build:** `node scripts/docker-native-build.js` still works (this script just exists for being able to debug these builds locally)

**CI Build:**
- Run an automated-preview CI job, but cancel it before the native build succeeds to test writing to the docker cache but not the native cache: https://github.com/vercel/next.js/actions/runs/26986955366/job/79638633136?pr=94319
- Re-run the job with only the docker image cache, see that we get a cache hit for the docker image: https://github.com/vercel/next.js/actions/runs/26986955366/job/79639365126?pr=94319
- Re-run the job again and see that we hit the native build cache hit and don't touch docker at all: https://github.com/vercel/next.js/actions/runs/26986955366/job/79647324952?pr=94319
- Manually trigger the job to do a full preview build: https://github.com/vercel/next.js/actions/runs/26989861549


bgw:
* **#94319** <a href="https://app.graphite.com/github/pr/vercel/next.js/94319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

.github/workflows/build_and_deploy.yml
```diff
@@ -270,13 +270,11 @@ jobs:
       - name: normalize versions
         run: node scripts/normalize-version-bump.js
 
-      - name: Cache on ${{ github.ref_name }}
+      - name: Cache Rust targets on ${{ github.ref_name }}
         if: ${{ !matrix.docker && !env.NEXT_SKIP_BUILD_CACHE }}
```
eps1lon: Do we even use this step nowadays? `NEXT_SKIP_BUILD_CACHE` is true for production, staging, and force-preview. So we only enable the build cache for the automated preview. However, the automated preview only builds Linux which runs in Docker. Seems like this always evaluates to false.

bgw: You're right. `gh cache list --key 'build-' --limit 1000` confirms.

.github/workflows/build_and_deploy.yml
```diff
@@ -270,55 +270,118 @@ jobs:
...
-
       # Try to restore the native binary BEFORE loading the Docker image.
       # If the binary cache hits, we skip the expensive Docker image restore entirely.
       - name: Rust fingerprint
         if: ${{ matrix.docker && !env.NEXT_SKIP_BUILD_CACHE }}
-        run: pnpm dlx turbo@${TURBO_VERSION} run rust-fingerprint ${TURBO_ARGS}
+        id: rust-fingerprint
+        run: |
+          pnpm dlx turbo@${TURBO_VERSION} run rust-fingerprint ${TURBO_ARGS}
+          echo "hash=$(cat target/.rust-fingerprint)" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
bgw: Practically that shouldn't happen, and if it did, it wouldn't really poison the cache for any reasonable definition of "poison".

bgw:
@eps1lon I increased the scope of this PR to migrate more of the native build logic to GHA cache.

There's still some remaining dependency on the turborepo remote cache where we use it for sccache. I intend to remove sccache entirely, just caching the rust `targets` directory instead.

.github/workflows/build_and_deploy.yml
```diff
@@ -270,55 +270,121 @@ jobs:
...
+            const names = [
+              'ACTIONS_RUNTIME_TOKEN',
+              'ACTIONS_RESULTS_URL',
+              ...Object.keys(process.env).filter((name) =>
+                name.startsWith('ACTIONS_CACHE_')
+              ),
+            ]
+            for (const name of names) {
+              core.exportVariable(name, process.env[name] || '')
+            }
```
eps1lon, line 307: Why can't this be listed in `env`? This seems pretty sketchy to dynamically expose variables. We don't know what future variables this may cover.

bgw: I originally just had a fixed list, but then it turns out that it missed `ACTIONS_CACHE_SERVICE_V2`, and so I figured this might be more robust to future changes on GitHub's side.

I don't think there's any real security implication here, as this action is running under the same level of permissions as other steps here, it's just that GitHub seems to assume that you'd only want these environment variables within JS-based actions.

The third-party action that the docker documentation recommends using (https://github.com/crazy-max/ghaction-github-runtime) actually exposes all `ACTIONS_*` environment variables, but I felt like that was excessive.

I can switch it back to a fixed list though.

lukesandberg (approved):

PR #92623, opened 2026-04-10
Remove a ton of turbo task functions from Issue trait items

lukesandberg:
## What?

Remove `#[turbo_tasks::function]` from all methods on the `Issue` value_trait, following the same pattern as #92593 which did the same for `ValueDebug`.

The `Issue` trait previously had 8 methods decorated with `#[turbo_tasks::function]`, which caused the turbo-tasks machinery to generate `NativeFunction` statics, task registration, vtable entries, and scheduling/caching infrastructure for each method on each of the ~48 `impl Issue for` types in the codebase — most of which are trivially synchronous (returning a constant string, enum variant, or stored struct field).

## How?

- Change all 8 `Issue` trait methods to plain methods. Methods that are always synchronous become direct return-type methods (`severity`, `stage`, `source`, `documentation_link`). Methods that are genuinely async on at least one impl use `Pin<Box<dyn Future<Output = Result<T>> + Send + '_>>` for object safety.
- Callers that need to invoke trait methods on a `Vc<Box<dyn Issue>>` use `issue.into_trait_ref().await?` to get a `TraitRef<Box<dyn Issue>>` and call methods directly on that.
- `PlainIssue::from_issue` (the sole hot caller that materializes all fields) resolves one `trait_ref` and calls all methods on it, similar to the `ValueDebug` pattern.
- `ResolvingIssueWithLocation` (which wraps an inner `ResolvedVc<Box<dyn Issue>>`) eagerly copies `severity`, `stage`, `documentation_link`, and `source` at construction time rather than delegating async calls to the inner issue.
- Delete `OptionIssueSource` and `OptionStyledString` wrapper types, which existed solely as `Vc<T>` return type wrappers for the old turbo-tasks methods.

## Tradeoffs

**Lost per-call caching:** Previously each `Issue` method was a turbo-tasks task, so results were cached per `(Vc<Issue>, method)`. In practice this caching was useless — issues are emitted once and their fields read once by `PlainIssue::from_issue`, so there was never a second call to hit the cache. The only real cost is that `IssueFilter::matches` previously cached per `(filter, issue)` pair; that caching is preserved because `matches` itself remains a `#[turbo_tasks::function]` and now calls `into_trait_ref().await?` once at the top.

**`ResolvingIssueWithLocation` eagerly reads inner fields:** Instead of lazily delegating to the inner issue, `stage` and `documentation_link` are copied at construction. This is a minor semantic change (the values are snapshotted rather than live), but in practice these fields are always constant values so it makes no difference.

[...]

PR #94324, opened 2026-06-01
[turbopack] Remove turbotask functions from `trait ResolveOrigin`

lukesandberg:
### What?

Removes the `#[turbo_tasks::function]` annotations from the `ResolveOrigin` trait methods (`origin_path`, `asset_context`, `resolve_options`), turning them into plain synchronous trait methods. Callers obtain the value once via `Vc::into_trait_ref().await?` and then call the methods directly on the resulting `TraitRef`.

This mirrors the earlier de-functionification of the `Issue` trait (#92623).

### Why?

The `ResolveOrigin` implementations are all trivial — they return a stored field or delegate a single call.

| Task (misses, before → after) | Misses removed |
| --- | --- |
| `ResolveOrigin::resolve_options` (dyn) | 92,608 → 0 |
| `EcmascriptModuleAsset::origin_path` | 64,779 → 0 |
| `ResolveOriginWithTransition::asset_context` | 52,292 → 0 |
| `ResolveOriginWithTransition::origin_path` | 52,292 → 0 |
| `EcmascriptModuleAsset::asset_context` | 41,164 → 0 |
| `PlainResolveOrigin::{origin_path, asset_context}` | 105 + 105 → 0 |
| `EcmascriptCssModule::{origin_path, asset_context}` | 32 + 32 → 0 |
| `CssModule::{origin_path, asset_context}` | 19 + 19 → 0 |

Net effect in that run: **~303k fewer cache misses** (total misses 6,197,068 → 5,893,545) — The ~1.35M eliminated cached-task *hits* additionally remove the per-call task lookup/scheduling overhead.

The caching that genuinely matters is preserved one layer down — `AssetContext::resolve_options` (which `ResolveOrigin::resolve_options` delegated to) is still a cached task; its miss count is unchanged (39,668) since those distinct results were always computed, and it simply absorbs the call volume directly now (hit rate 57% → 92%).

## Build Performance: `ResolveOrigin` optimization

Benchmark across 10 runs per branch.

| Metric | Base | This branch | Δ |
|---|---|---|---|
| **Wall time** | 43.302 s | 42.692 s | **−0.61 s (−1.4%)** |
| **User time** | 314.249 s | 310.919 s | **−3.33 s (−1.1%)** |
| **Sys time** | 84.225 s | 84.618 s | +0.39 s (+0.5%) |
| **MaxRSS** | 14106.1 MB | 13883.4 MB | **−222.8 MB (−1.6%)** |

### Notes on significance

- **User time** is the cleanest signal: −1.1% with very tight variance (base σ 0.4%, branch σ 0.3%). The drop (~3.3 s) is well outside the noise — a real reduction in CPU work.
- **MaxRSS** is down a solid 1.6% (~223 MB) with tight variance on both sides (σ ≈ 0.5–0.6%) — a clear, reproducible memory win.
- **Wall time** improves 1.4%, but variance is large (base σ 6.6%, branch σ 4.5%, with one 50 s outlier in the base set), so treat this as directional. It tracks the user-time improvement.
- **Sys time** is essentially flat (+0.5%, within its ~10% run-to-run noise) — no meaningful change, as expected for a CPU/allocation optimization.

### Invalidation tradeoff

The main behavioral risk is invalidation granularity. Removing the per-method tasks removes their invalidation boundary: a consumer that previously depended only on the `origin_path()` (or `asset_context()`) task output now reads the value off the origin directly, so it is tied to the origin cell rather than to that one method's cached output.

In practice this is narrow, because every `ResolveOrigin` implementor is an immutable, content-addressed `#[turbo_tasks::value]` (only `ResolvedVc`/owned fields) — a different field value means a different cell, so "reading the whole value vs. one field" does not by itself widen invalidation.

### DIFF
--- a/.config/ast-grep/rules/resolved-vc-in-return-type.yml
+++ b/.config/ast-grep/rules/resolved-vc-in-return-type.yml
@@ -1,34 +1,24 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/main/schemas/rule.json
 
 id: resolved-vc-in-return-type
-message: Returning a `ResolvedVc` in a turbo task is not recommended. Consider replacing `$TYPE` in `$FN_NAME` with `$RESOLVED_VC` instead.
+message: Returning a `ResolvedVc` in a turbo task is not recommended. Consider replacing `ResolvedVc<$A>` in `$FN_NAME` with `Vc<$A>` instead.
 severity: warning # error, warning, info, hint
 language: Rust
 rule:
-  # turbo tasks function
-  - all:
-      - pattern:
-          context: 'let x: ResolvedVc<$A> = 1;'
-          selector: generic_type
-      - pattern: $TYPE
-    inside:
-      kind: function_item
-      field: return_type
+  # The `ResolvedVc<$A>` generic type itself, captured as $TYPE.
+  pattern:
+    context: 'let x: ResolvedVc<$A> = 1;'
+    selector: generic_type
+  # ...appearing in the return type of a `#[turbo_tasks::function]`.
+  inside:
+    kind: function_item
+    field: return_type
+    stopBy: end
+    follows:
+      pattern: '#[turbo_tasks::function]'
       stopBy: end
-      follows:
-        pattern: '#[turbo_tasks::function]'
-      has:
-        kind: identifier
-        pattern: $FN_NAME
+    has:
+      kind: identifier
+      pattern: $FN_NAME
 
-rewriters:
-  - id: rewrite-vc
-    rule:
-      pattern: $TYPE
-    fix: Vc<$A>
-transform:
-  RESOLVED_VC:
-    rewrite:
-      rewriters: [rewrite-vc]
-      source: $TYPE
-fix: $RESOLVED_VC
+fix: Vc<$A>
--- a/.config/ast-grep/rules/resolved-vc-in-return-type.yml
+++ b/.config/ast-grep/rules/resolved-vc-in-return-type.yml
@@ -5,40 +5,22 @@ message: Returning a `ResolvedVc` in a turbo task is not recommended. Consider r
 severity: warning # error, warning, info, hint
 language: Rust
 rule:
-  any:
-    # turbo tasks function
-    - all:
-        - pattern:
-            context: 'let x: ResolvedVc<$A> = 1;'
-            selector: generic_type
-        - pattern: $TYPE
-      inside:
-        kind: function_item
-        field: return_type
-        stopBy: end
-        follows:
-          pattern: '#[turbo_tasks::function]'
-        has:
-          kind: identifier
-          pattern: $FN_NAME
-      # turbo tasks value trait
-    - all:
-        - pattern:
-            context: 'let x: ResolvedVc<$A> = 1;'
-            selector: generic_type
-        - pattern: $TYPE
-      inside:
-        kind: function_signature_item
-        field: return_type
-        stopBy: end
-        inside:
-          inside:
-            kind: trait_item
-            follows:
-              pattern: '#[turbo_tasks::value_trait]'
-        has:
-          kind: identifier
-          pattern: $FN_NAME
+  # turbo tasks function
+  - all:
+      - pattern:
+          context: 'let x: ResolvedVc<$A> = 1;'
+          selector: generic_type
+      - pattern: $TYPE
+    inside:
+      kind: function_item
+      field: return_type
+      stopBy: end
+      follows:
+        pattern: '#[turbo_tasks::function]'
+      has:
+        kind: identifier
+        pattern: $FN_NAME
+
 rewriters:
   - id: rewrite-vc
     rule:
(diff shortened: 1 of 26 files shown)


lukesandberg:
* **#94465** <a href="https://app.graphite.com/github/pr/vercel/next.js/94465?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#94466](https://github.com/vercel/next.js/pull/94466) <a href="https://app.graphite.com/github/pr/vercel/next.js/94466?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#94467](https://github.com/vercel/next.js/pull/94467) <a href="https://app.graphite.com/github/pr/vercel/next.js/94467?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#94324** <a href="https://app.graphite.com/github/pr/vercel/next.js/94324?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94324?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbopack-cli/src/build/mod.rs
```diff
@@ -293,7 +294,7 @@ async fn build_internal(
                 let ty = ReferenceType::Entry(EntryReferenceSubType::Undefined);
                 let request = request_vc.await?;
                 origin
```
lukesandberg: resolve_asset immediately derefs the trait to a traitref, a little bit of inlining would remove that work

turbopack/crates/turbopack-cli/src/dev/web_entry_source.rs
```diff
@@ -134,12 +134,14 @@ pub async fn create_web_entry_source(
...
     let origin = PlainResolveOrigin::new(asset_context, root_path.join("_")?);
+    // `resolve_options` is loop-invariant; read it once instead of per entry.
+    let resolve_options = origin.await?.resolve_options()?;
     let entries = entry_requests
         .into_iter()
         .map(|request| async move {
             let ty = ReferenceType::Entry(EntryReferenceSubType::Web);
             origin
-                .resolve_asset(request, origin.resolve_options(), ty)
+                .resolve_asset(request, resolve_options, ty)
```
lukesandberg: ditto

turbopack/crates/turbopack-ecmascript/src/references/esm/base.rs
```diff
@@ -386,12 +386,14 @@ pub struct EsmAssetReference {
...
 
 impl EsmAssetReference {
-    fn get_origin(&self) -> Vc<Box<dyn ResolveOrigin>> {
-        if let Some(transition) = self.annotations.as_ref().and_then(|a| a.transition()) {
-            self.origin.with_transition(transition.into())
-        } else {
-            *self.origin
-        }
+    async fn get_origin(&self) -> Result<Vc<Box<dyn ResolveOrigin>>> {
+        Ok(
```
lukesandberg: just like with EsmAsyncAssetReference it would be better to precompute this

it is true we need other data from the ImportAnnotations but we should be able to just capture that eagerly and drop the btreemap

turbopack/crates/turbopack-ecmascript/src/references/esm/dynamic.rs
```diff
@@ -37,9 +37,9 @@ use crate::{
 #[derive(Hash, Debug, ValueToString)]
 #[value_to_string("dynamic import {request}")]
 pub struct EsmAsyncAssetReference {
+    /// The resolve origin, with any annotation-driven transition already applied at construction.
```
lukesandberg: drop comment

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -435,6 +435,9 @@ struct AnalysisState<'a> {
     module: ResolvedVc<EcmascriptModuleAsset>,
     source: ResolvedVc<Box<dyn Source>>,
     origin: ResolvedVc<Box<dyn ResolveOrigin>>,
+    /// The resolved origin path, precomputed once so that `Sync`-constrained visitor futures don't
```
lukesandberg: i think we could just pass down a TraitRef for the origin?  instead of both this and the origin iteelf

bgw (approved):

.config/ast-grep/rules/resolved-vc-in-return-type.yml
```diff
@@ -1,52 +1,24 @@
...
-        has:
-          kind: identifier
-          pattern: $FN_NAME
-rewriters:
-  - id: rewrite-vc
-    rule:
-      pattern: $TYPE
-    fix: Vc<$A>
-transform:
-  RESOLVED_VC:
```
mischnic, line 48: What was this change for?

lukesandberg: I was attempting to remove the rule targeting all members of value_trait impl blocks

Since this rules should only apply to functions with turbo_task::function annotations.

mischnic: see the description in https://github.com/vercel/next.js/pull/94547

PR #93953, opened 2026-05-19
[ci] Explicitly disable package-manager-cache for uses of actions/setup-node

bgw:
If this action detects that you're likely using `npm` it (by default) enables a cache for `npm`. We use `pnpm`, so it doesn't do anything (and we manually set up a pnpm cache in some places), but the implicit behavior here is a little dangerous: e.g. https://adnanthekhan.com/posts/angular-compromise-through-dev-infra/#finding-the-pivot

So let's explicitly disable this everywhere. It wasn't doing anything for us anyways.

PR #93609, opened 2026-05-07
[ci] Also pin first-party GH actions

bgw:
We were already pinning nearly all of our third-party GitHub actions, but this also pins our first-party ones (things starting with `actions/*`.

This gets us closer to being able to enforcing pinning at the repository level:

The one remaining blocker is this self-reference in a `pull_request_target` action: https://github.com/vercel/next.js/blob/c06d94ba22d0156e8bff28c81f7874b73d80ed03/.github/workflows/pull_request_auto_label.yml#L65

I'm still figuring out the best approach to do there, and I'll submit that in a separate PR.

Discussion here about enforcing this org-wide: https://vercel.slack.com/archives/C0AM84PRSGL/p1778110550384279

PR #92757, opened 2026-04-13
Pin upload-artifact to v7.0.1 and download-artifact to v8.0.1

mmastrac:
## What
- Pin all `actions/upload-artifact` references to `v7.0.1` (`043fb46d1a93c77aae656e7c1c64a875d1fc6a0a`)
- Pin all `actions/download-artifact` references to `v8.0.1` (`3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c`)

PR #94344, opened 2026-06-02
[15.5.x] Align Actions dependencies with Canary

eps1lon:
The main motivation is getting the latest actions/checkout version which ensures we use the expected token in trigger_release i.e. this should fix the trigger release workflow for 16.2.

Cherry-picks Action version changes from:
1. https://github.com/vercel/next.js/pull/91477
1. https://github.com/vercel/next.js/pull/92757
1. https://github.com/vercel/next.js/pull/92016
1. https://github.com/vercel/next.js/pull/93609
1. https://github.com/vercel/next.js/pull/93953
1. https://github.com/vercel/next.js/pull/94174

## Test plan

- Does not fix the trigger release. Still want to land this in case GitHub magically fixes itself after this lands on a protected branch. And we want to pin Actions dependencies anyway.

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -9,7 +9,7 @@ runs:
   using: 'composite'
   steps:
     - name: 'Setup Rust toolchain'
-      uses: actions-rust-lang/setup-rust-toolchain@v1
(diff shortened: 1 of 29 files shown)


bgw (approved):
Thank you for backporting this!

PR #94347, opened 2026-06-02
Delay Node Fizz piping until after render task

timneutkens:
### What?

Delay dynamic Node Fizz piping by one React render task after `onShellReady()`, and keep the dedicated Node-stream CI jobs scoped to `test/use-node-streams-tests-manifest.json`.

### Why?

The Node adapter started `pipeable.pipe(pt)` inside `onShellReady()`, before the existing continuation delay could take effect. In App Router renders backed by RSC, that caused short-lived Suspense boundaries to flush fallback and reveal markup that the web-stream path avoids by waiting before its first pull.

The previous test assertion relaxations hid this runtime mismatch. The original assertions pass once piping starts after the settling window, so they have been restored.

The workflow correction remains necessary because merging the cache-components manifest into the ordinary Node-stream jobs filtered out relevant e2e coverage.

### How?

- Resolve the Node Fizz shell promise from `onShellReady()` without immediately piping.
- For dynamic renders, wait one React render task and then start piping into the `PassThrough`.
- Preserve the existing `onAllReady()` behavior for static generation.
- Use the Node-stream manifest alone for ordinary Node-stream dev and prod CI jobs.

### Verification

- `pnpm --filter=next build`
- `__NEXT_USE_NODE_STREAMS=true __NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-turbo test/e2e/app-dir/ppr-root-param-fallback/ppr-root-param-fallback.test.ts`
- `__NEXT_USE_NODE_STREAMS=true __NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-turbo test/e2e/app-dir/use-server-inserted-html/use-server-inserted-html.test.ts`

### DIFF
--- a/test/e2e/app-dir/ppr-root-param-fallback/ppr-root-param-fallback.test.ts
+++ b/test/e2e/app-dir/ppr-root-param-fallback/ppr-root-param-fallback.test.ts
@@ -14,29 +14,17 @@ describe('ppr-root-param-fallback', () => {
     // filled in for all pregenerated locales.
 
     for (const locale of ['en', 'fr']) {
+      // next.render$ doesn't stream, so we get just the shell content
       const $ = await next.render$(`/${locale}/blog/new-post`)
-      const html = $.html()
 
       // The shell should have the locale-header with cached content,
-      // even if Node streams flush the Suspense fallback before the resolved
-      // content that replaces it.
+      // NOT the locale-loading Suspense fallback
       expect($('#locale-header').length).toBe(1)
       expect($('#locale-header').text()).toContain(`Locale: ${locale}`)
       expect($('#translations').text()).toContain(`Home (${locale})`)
 
-      const localeLoading = $('#locale-loading')
-      if (localeLoading.length > 0) {
-        const fallbackTemplateId = localeLoading.prev('template').attr('id')
-        const resolvedContentId = $('#locale-header')
-          .parent('div[hidden]')
-          .attr('id')
-
-        expect(fallbackTemplateId).toBeTruthy()
-        expect(resolvedContentId).toBeTruthy()
-        expect(html).toContain(
-          `$RC("${fallbackTemplateId}","${resolvedContentId}")`
-        )
-      }
+      // The Suspense fallback should NOT be in the shell
+      expect($('#locale-loading').length).toBe(0)
     }
   })
 })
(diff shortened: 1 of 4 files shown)


test/e2e/app-dir/ppr-root-param-fallback/ppr-root-param-fallback.test.ts
```diff
@@ -14,17 +14,29 @@ describe('ppr-root-param-fallback', () => {
...
+      // even if Node streams flush the Suspense fallback before the resolved
+      // content that replaces it.
       expect($('#locale-header').length).toBe(1)
       expect($('#locale-header').text()).toContain(`Locale: ${locale}`)
       expect($('#translations').text()).toContain(`Home (${locale})`)
 
-      // The Suspense fallback should NOT be in the shell
-      expect($('#locale-loading').length).toBe(0)
+      const localeLoading = $('#locale-loading')
+      if (localeLoading.length > 0) {
```
unstubbable: Something's not right here, I think. This element should not be in the document at all. If it is, then it's maybe because `.next/server/app/[locale].html` was used for resuming and not `.next/server/app/en/blog/[slug].html`, which would be a bug.

unstubbable: Or it's suspending on the client and showing the fallback based on the inlined RSC data. But why?

unstubbable: Oh wait, did the test only fail in dev? Can you link to a failed CI run?

timneutkens: :robot: Found and fixed the underlying Node-stream adapter bug. The Node path started pipeable.pipe(pt) directly inside onShellReady(), before the existing render-task settling window could take effect. The fix resolves shellReady first, waits one React render task for dynamic renders, and then starts piping. I restored the original strict assertion that #locale-loading is absent and verified it passes with __NEXT_USE_NODE_STREAMS=true.

unstubbable (approved):
This makes sense to me. It's basically the same reasoning we applied in #86311. In fact, I think we should follow up by removing `waitAtLeastOneReactRenderTask` from the Node streams version of `continueFizzStream`, as it has no effect on the shell on the Node path.

PR #94349, opened 2026-06-02
refactor: make stage advancing logic more generic

lubieowoce:
In #93801, we'll be adding 4 new stages. `StagedRenderingController` already contains a lot of duplication around advancing stages, and i want to avoid adding to that, so this PR does some cleanup.

Previously, for a stage, we had
- `resolve<name>Stage` - triggers listener callbacks and resolves the promise for the stage
  - `<name>StageListeners` - added via `onStage`, triggered when advancing to the stage but before resolving the promise
  - `<name>StagePromise` - resolved when we enter the stage (or rejected on abort/abandon)
- sometimes, `<name>EndTime`

these behaviors are all now packaged up into `StageTrigger`, which can be "fired" (when advancing to a stage) or "cancelled" (when aborting/abandoning) and tracks the time.

In addition, we now have `RENDER_STAGE_ADVANCE_ORDER`, an array that defines the order in which advanceable stages (i.e. excluding Before/Abandoned) should happen. We can loop through this to advance stages without relying on a huge switch statement with fallthroughs. Re-working the logic into this shape will be handy in the future, when we want each route param to have its own stage, and the set of stages is no longer statically known.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -252,7 +252,11 @@ import {
 import type { Params } from '../request/params'
 import { ImageConfigContext } from '../../shared/lib/image-config-context.shared-runtime'
 import { imageConfigDefault } from '../../shared/lib/image-config'
-import { RenderStage, StagedRenderingController } from './staged-rendering'
+import {
+  getNextStage,
+  RenderStage,
+  StagedRenderingController,
+} from './staged-rendering'
 import {
   anySegmentHasRuntimePrefetchEnabled,
   isPageAllowedToBlock,
@@ -5321,7 +5325,8 @@ async function countStaticStageBytes(
   let byteLength = 0
   const reader = stream.getReader()
 
-  stageController.onStage(RenderStage.EarlyRuntime, () => {
+  const endStage = getNextStage(RenderStage.Static)
+  stageController.onStage(endStage, () => {
     reader.cancel()
   })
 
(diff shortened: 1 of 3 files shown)


lubieowoce:
* **#94350** <a href="https://app.graphite.com/github/pr/vercel/next.js/94350?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94358** <a href="https://app.graphite.com/github/pr/vercel/next.js/94358?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94349** <a href="https://app.graphite.com/github/pr/vercel/next.js/94349?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94349?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -5321,7 +5325,8 @@ async function countStaticStageBytes(
   let byteLength = 0
   const reader = stream.getReader()
 
-  stageController.onStage(RenderStage.EarlyRuntime, () => {
+  const endStage = getNextStage(RenderStage.Static)
+  stageController.onStage(endStage, () => {
     reader.cancel()
   })
```
gnoff, line 5331: Is this pattern common? Should we model this as afterStage and let you pass in Static and it fires after you transition past this stage as opposed to on entering the next stage?

gnoff (approved):

packages/next/src/server/app-render/app-render.tsx
lubieowoce: i haven't seen it elsewhere (other than the node/web forks). but also importantly there's an implicit assumption here that we'll reach a stage after the target stage at all. here we know we will, because it's working off of a full staged render, but in other codepaths, if we're doing a prerender, we might not reach the target stage at all and would need to deal with that somehow. we'd probably need to to call something at the end of the task where we do this stage. anyway i wanted to avoid making API design decisions on that, so i didn't make this generic

PR #94350, opened 2026-06-02
refactor: remove delayUntilRuntimeStage

lubieowoce:
In #93801, we'll be removing `getRuntimeStage` because we want to differentiate session data and link data, which resolve in different stages. `delayUntilRuntimeStage` isn't all that useful anyway (we can do `stagedRendering.delayUntilStage` instead) and gets in the way of that refactor. This should have no behavior changes otherwise.

### DIFF
--- a/packages/next/src/server/request/cookies.ts
+++ b/packages/next/src/server/request/cookies.ts
@@ -22,9 +22,9 @@ import {
 } from '../app-render/dynamic-rendering'
 import { StaticGenBailoutError } from '../../client/components/static-generation-bailout'
 import {
-  delayUntilRuntimeStage,
   makeDevtoolsIOAwarePromise,
   makeHangingPromise,
+  getRuntimeStage,
 } from '../dynamic-rendering-utils'
 import { createDedupedByCallsiteServerErrorLoggerDev } from '../create-deduped-by-callsite-server-error-logger'
 import { isRequestAPICallableInsideAfter } from './utils'
@@ -104,11 +104,18 @@ export function cookies(): Promise<ReadonlyRequestCookies> {
             workStore,
             workUnitStore
           )
-        case 'prerender-runtime':
-          return delayUntilRuntimeStage(
-            workUnitStore,
-            makeUntrackedCookies(workUnitStore.cookies)
-          )
+        case 'prerender-runtime': {
+          const { stagedRendering } = workUnitStore
+          if (stagedRendering) {
+            return stagedRendering.delayUntilStage(
+              getRuntimeStage(stagedRendering),
+              'cookies',
+              workUnitStore.cookies
+            )
+          } else {
+            return makeUntrackedCookies(workUnitStore.cookies)
+          }
+        }
         case 'private-cache':
           // Private caches are delayed until the runtime stage in use-cache-wrapper,
           // so we don't need an additional delay here.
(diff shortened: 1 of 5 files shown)


lubieowoce:
* **#94350** <a href="https://app.graphite.com/github/pr/vercel/next.js/94350?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94350?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94358** <a href="https://app.graphite.com/github/pr/vercel/next.js/94358?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94349** <a href="https://app.graphite.com/github/pr/vercel/next.js/94349?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/request/cookies.ts
```diff
@@ -104,11 +104,18 @@ export function cookies(): Promise<ReadonlyRequestCookies> {
...
-            workUnitStore,
-            makeUntrackedCookies(workUnitStore.cookies)
-          )
+        case 'prerender-runtime': {
+          const { stagedRendering } = workUnitStore
+          if (stagedRendering) {
+            return stagedRendering.delayUntilStage(
+              getRuntimeStage(stagedRendering),
+              'cookies',
+              workUnitStore.cookies
```
gnoff, line 113: the old code looks wrong. why is it delaying the untracked cookies? the new code seems right but flagging it as odd that it was ever set up this way

packages/next/src/server/request/cookies.ts
```diff
@@ -104,11 +104,18 @@ export function cookies(): Promise<ReadonlyRequestCookies> {
             workStore,
             workUnitStore
           )
-        case 'prerender-runtime':
-          return delayUntilRuntimeStage(
-            workUnitStore,
-            makeUntrackedCookies(workUnitStore.cookies)
-          )
+        case 'prerender-runtime': {
+          const { stagedRendering } = workUnitStore
```
gnoff, line 108: I wonder if having a store for the warmup render would simplify the types of the other stores. Having this be optional is interesting because it's only optional when cache filling but maybe the behavior of cookies on cache filling is consistent regardless of static or runtime prospective renders.

gnoff (approved):

packages/next/src/server/request/cookies.ts
lubieowoce: it's not really wrong per se, it's gonna resolve to a (already resolved) promise with a delay, which is the same as resolving to its value because the promise will get unwrapped. `p.then(() => x)` is basically equivalent to `p.then(() => Promise.resolve(x))`

seems like i set it up that way when i originally added it https://github.com/vercel/next.js/pull/82475/changes#diff-20c354509ae1e93e143d91b67b75e3df592c38b7d1ec6ccf7c4a2f72b32ab17dR845, not sure what the thinking was at the time, i'm guessing that this was just the easiest way to drop an additional delay into the existing promise code

packages/next/src/server/request/cookies.ts
lubieowoce: not sure if we want a whole separate store type, but it would be useful to discriminate which things are actually set together instead of having them all be independently nullable. so, like this
```ts
type RuntimePrerenderStore = { type: "prerender-runtime" } & (
  | { cacheSignal: CacheSignal; stagedRendering: null }
  | { cacheSignal: null; stagedRendering: StagedRenderingController }
);

PR #94356, opened 2026-06-02
Expose Turbopack NFT functionality as separate entrypoint

mischnic:
Replicate what `require("@vercel/nft").nodeFileTrace` does


turbopack/crates/turbopack-nft-napi/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+                max_depth.map(|d| d as usize),
+            )
+            .await
+        })
+        .await
+        .map_err(|e| napi::Error::from_reason(format!("{e:#}")))?;
+
+    // Intentionally leak the TurboTasks instance to avoid expensive cleanup,
+    // same pattern as turbopack-nft/src/main.rs
+    std::mem::forget(tt);
```
mischnic:
```suggestion
```

PR #94357, opened 2026-06-02
Keep Instant DevTools locked when opening error overlay

devjiwonchoi:
### Why?

The Instant DevTools closes when opening the error overlay and continues hydration. This is annoying when I wanted to check the error, but didn't mean to stop the capture.

Before:

[github.com/user-attachments/assets/f1de72a9-5a2b-4e4e-bd7e-e7f36c975ea7](https://github.com/user-attachments/assets/f1de72a9-5a2b-4e4e-bd7e-e7f36c975ea7)

After:

https://github.com/user-attachments/assets/f060e11e-9d60-43f6-a1fd-ca735fad1aa1

### DIFF
--- a/packages/next/src/next-devtools/dev-overlay/components/devtools-indicator/next-logo.tsx
+++ b/packages/next/src/next-devtools/dev-overlay/components/devtools-indicator/next-logo.tsx
@@ -512,7 +512,12 @@ export function NextLogo({
                         return
                       }
                       dispatch({ type: ACTION_ERROR_OVERLAY_OPEN })
-                      setPanel(null)
+                      // Keep the instant navigation panel mounted so its capture
+                      // survives and it stays behind the error overlay backdrop.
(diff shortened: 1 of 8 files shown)


samselikoff (changes requested):
Ok this one looks good!

The test is currently being skipped on CI because of flakiness there. I ran it locally and everything works. Let's keep the `.skip` for now around the whole thing, because I don't want to introduce new flakiness that could affect others, and I worry the helpers that are in the new tests in this PR are still flaky.

Maybe add the skip back around the root `describe` with this comment:

```
/*
  FIXME: This entire module is flaky on CI. When making changes, remove the
  .skip and run locally, but add it back before pushing.
  
  We can re-enable the module on CI when the root cause of the flakiness is
  determined (#94196).
*/
```

samselikoff (approved):

PR #89979, opened 2026-02-13
Only error for sync IO after runtime in segments that would be runtime prefetched

gnoff:
Stacked on #89984

When we initially implemented runtime prefetching we decided to make sync IO that happens after runtime but before async IO an error. This is similar to but more aggressive than the error for sync IO before runtime data but is important because if you have sync IO before uncached IO and we are runtime prefetching the page then we aren't going to get anything useful from that prefetch but it might be very expensive.

The problem with the current implementation is that if you turn on runtime prefetching for a Page segment then in practice the router will only ever runtime prefetch that page segment but the sync IO validator will error for sync IO before uncached IO anywhere in the route. This makes adding runtime prefetching to leaf Segments much harder b/c you have to address sync IO in the entire Segment stack.

This change updates the sync IO heuristic to only error in segments that would be runtime prefetched and below. This way when you add runtime prefetching to a segment you only have to true up the sync IO for the part of the tree that has the changed behavior.

This PR only partially implements the intended semantics for runtime prefetching validation of sync IO. One thing that will need to be done in a follow up is handling sync IO in generateMetadata. In this case while the functions are defined in layouts and pages they are semantically always "owned" by the page itself. This means that while the layout might not be runtime prefetchable and thus allow sync IO after runtime data access the generateMetadata in that same layout might not allow sync IO after runtime if there is a runtime prefetch config at the page level or some layout higher. This will be done in a follow up

PR #94358, opened 2026-06-02
refactor: make early/late checks stricter

lubieowoce:
While working on #93801 , i added some stricter checks about early/late stages. In particular, if we're in the `Before`stage (generally, while running `createComponentTree`), we can't know if we're supposed to use early or late stages. Making this stricter revealed that `createServerPathnameForMetadata` was accidentally always using the late stage because it runs in `Before`. Instead, it should check `isRuntimePrefetchable` like we do for `params` and `searchParams`, so i'm refactoring it to do that here.

### DIFF
--- a/packages/next/src/server/app-render/work-unit-async-storage.external.ts
+++ b/packages/next/src/server/app-render/work-unit-async-storage.external.ts
@@ -21,7 +21,7 @@ import type { WorkStore } from './work-async-storage.external'
 import { NEXT_HMR_REFRESH_HASH_COOKIE } from '../../client/components/app-router-headers'
 import { InvariantError } from '../../shared/lib/invariant-error'
 import type { StagedRenderingController } from './staged-rendering'
-import { RenderStage } from './staged-rendering'
+import { isEarlyRenderStage, RenderStage } from './staged-rendering'
 import type { ValidationBoundaryTracking } from './instant-validation/boundary-tracking'
 import type { InstantValidationSampleTracking } from './instant-validation/instant-samples'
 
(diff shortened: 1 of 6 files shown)


lubieowoce:
* **#94350** <a href="https://app.graphite.com/github/pr/vercel/next.js/94350?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94358** <a href="https://app.graphite.com/github/pr/vercel/next.js/94358?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94358?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94349** <a href="https://app.graphite.com/github/pr/vercel/next.js/94349?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/work-unit-async-storage.external.ts
```diff
@@ -131,10 +131,13 @@ export type AsyncApiPromises = {
 export function isInEarlyRenderStage(requestStore: RequestStore): boolean {
```
gnoff, line 131: the change is good, but it's odd this function even exists in this file? why not just move this whole thing into staged-rendering and import it from there everywher it is needed? gets rid of the duplicate name

gnoff (approved):

packages/next/src/server/app-render/work-unit-async-storage.external.ts
lubieowoce: you tell me why it's in this file, [you added it](https://github.com/vercel/next.js/pull/89979/changes#diff-71df16755b3b1fd33b004a6354977d6ce77930f1efed49910db2896f6e332eb8R108) :p

agree we should get rid of, but i don't have the brainpower to untangle this right now, and i don't want to inline this `Before` check into every callsite. will follow up to clean this up later

PR #94360, opened 2026-06-02
docs: add GitHub SSH authentication guidance

timneutkens:
### What?

Add GitHub SSH authentication guidance to `AGENTS.md` for user-configured SSH agents and key providers.

### Why?

Remote operations can fail when an SSH agent or key provider is unavailable or locked. The guidance prevents unrequested authentication workarounds and directs agents to involve the user when signing fails.

### How?

Document recognized SSH signing failure symptoms, the required stop-and-ask behavior, and a lightweight preflight recommendation before force-pushes or stack rebases that may hydrate partial-clone objects.

### Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write AGENTS.md`
- `git diff --cached --check`
- `pnpm build-all`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -321,6 +321,27 @@ Always treat environment variable values as sensitive unless they are known test
 - Never commit local secret files; if documenting env setup, use placeholder-only examples.
 - When sharing command output, summarize and redact sensitive-looking values.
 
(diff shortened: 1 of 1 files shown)


aurorascharff (approved):
great addition

PR #94366, opened 2026-06-02
Revert "ci: run Linux Playwright jobs in prebuilt Microsoft container (#93794)"

bgw:
We're going to go with custom images instead, though we don't have an implementation for this yet.

This does save a meaningful amount of minutes, but it also significantly increases complexity.

This is a clean revert, minus the fix that the PR included for listening on localhost in `packages/next/src/build/analyze/index.ts`, which is still good.

This also pins our exact version of Node 24 in some e2e tests because there is a bug with older versions of `playwright install` on newer versions of Node: https://github.com/microsoft/playwright/issues/40724

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -588,7 +588,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [22, 24]
+        # Playwright hangs on Node 24.16+, we can remove this exact node version
+        # pin once we update to playwright 1.60.0 or newer.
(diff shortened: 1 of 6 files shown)


bgw:
* **#94366** <a href="https://app.graphite.com/github/pr/vercel/next.js/94366?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94366?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

eps1lon (approved):

PR #94370, opened 2026-06-03
fix(eslint-plugin): respect pageExtensions in no-html-link-for-pages rule

InukaWijerathna:
Fixes #53473

## What

The `no-html-link-for-pages` ESLint rule hardcoded `.js/.jsx/.ts/.tsx` as the only recognized page extensions (there was even a TODO comment flagging this). When a project configures custom `pageExtensions` in `next.config.js`, the rule would not recognize those files as pages and either miss violations or produce false positives.

## How

**`packages/eslint-plugin-next/src/utils/url.ts`**
- `parseUrlForPages` and `parseUrlForAppDir` now accept a `pageExtensions: string[]` parameter (defaults to `['js','jsx','ts','tsx']`)
- Extension regexes are built dynamically from the provided extensions instead of being hardcoded
- `getUrlFromPagesDirectories` and `getUrlFromAppDirectory` forward `pageExtensions` through

**`packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts`**
- Adds `readPageExtensionsFromConfig` which tries to `require()` `next.config.js` / `next.config.cjs` from each root dir and extract `pageExtensions` (safe try/catch for ESM/errors)
- Adds an optional second rule option `{ pageExtensions: string[] }` for projects that cannot expose CJS config
- Resolution order: **rule option > next.config.js > default**

## Testing

Existing tests should pass unchanged (defaults are preserved). A project with `pageExtensions: ['page.tsx', 'page.jsx']` in `next.config.js` will now have those extensions respected by the rule.

### DIFF
--- a/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
+++ b/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
@@ -10,6 +10,8 @@ import {
   getUrlFromAppDirectory,
 } from '../utils/url'
 
+const DEFAULT_PAGE_EXTENSIONS = ['js', 'jsx', 'ts', 'tsx']
+
 const pagesDirWarning = execOnce((pagesDirs) => {
   console.warn(
     `Pages directory cannot be found at ${pagesDirs.join(' or ')}. ` +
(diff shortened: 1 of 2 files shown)


InukaWijerathna:
Hi! Just checking in to see if there's any update on this PR — happy to make any changes needed to get it merged. Thanks for your time!

PR #94372, opened 2026-06-03
[turbopack] Remove worker helpers from the default runtime

sampoder:
This PR is looking to reduce the size of the runtime we're shipping. The goal is to not include worker helpers in the default runtime, if workers aren't being used / imported. 

The `createWorker()` function has been removed from the runtime. It now exists in `turbopack-ecmascript/js/src/worker/browser/createWorker.ts` and `turbopack-ecmascript/js/src/worker/node/createWorker.ts`. The reason behind having them in `turbopack-ecmascript` is that in `turbopack-ecmascript/src/worker_chunk/module.rs` we can then create a worker module (see `create_worker_module()`). This is then returned by `references()` as a `SingleChunkableModuleReference`.

Instead of using `createWorker()` from the runtime we can instead now import it from that module. 

There are a couple of things in this PR that I'd like to flag for discussion:

1) The location of the JavaScript. Shipping it in `turbopack-ecmascript` doesn't quite feel natural; however, `turbopack-ecmascript-runtime` imports `turbopack-ecmascript` so I didn't want to create a circular dependency by importing `turbopack-ecmascript-runtime` in `turbopack-ecmascript`.

2) There a couple of things in the runtime that we need inside of `createWorker()`, they are:

```javascript
TURBOPACK_ASSET_SUFFIX, TURBOPACK_CHUNK_BASE_PATH
```

which are values from the runtime we needed to references

and:

```javascript
function getChunkRelativeUrl(
  chunkPath: ChunkPath | ChunkListPath,
  basePath: string = CHUNK_BASE_PATH
): ChunkUrl
```

which is `TURBOPACK_CHUNK_RELATIVE_URL`.

These currently exist in the runtime. To make them available, I've placed them in `browserContextPrototype` and `contextPrototype` in the node version.

We also have a code gen that inserts the values of `_TURBOPACK_WORKER_FORWARDED_GLOBALS_` and `_TURBOPACK_WORKER_BASE_PATH_` during the analyzer's pass through free variables (takes them out of the chunking context).

@lukesandberg curious to hear your thoughts about these!


turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -393,6 +337,18 @@ function getChunkRelativeUrl(
     .join('/')}${ASSET_SUFFIX}` as ChunkUrl
 }
 
+// Workers-specific runtime config for the bundled `createWorker` helper.
+browserContextPrototype.b = {
+  CHUNK_BASE_PATH,
+  WORKER_BASE_PATH,
+  ASSET_SUFFIX,
+  WORKER_FORWARDED_GLOBALS,
```
lukesandberg: use a 'tuple type' aka an array

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-types.d.ts
```diff
@@ -146,12 +151,14 @@ interface TurbopackBaseContext<M> {
   M: ModuleFactories
   l: LoadChunk
   L: LoadChunkByUrl
+  // Shared runtime primitive: builds a chunk URL (browser); used by bundled helpers.
+  h: Function
```
lukesandberg: better type please

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-types.d.ts
```diff
@@ -128,6 +123,16 @@ interface ModuleWithDirection extends Module {
   parents: ModuleId[]
 }
 
+// Exposed so that bundled `createWorker` can read the runtime internals.
+// All fields are set by the browser, `WORKER_FORWARDED_GLOBALS` is the
+// only one of these set by node.
+type WorkerRuntimeBindings = {
+  WORKER_FORWARDED_GLOBALS: string[]
+  CHUNK_BASE_PATH?: string
+  WORKER_BASE_PATH?: string | null
```
lukesandberg: WORKER_FORWARDED_GLOBALS and WORKER_BASE_PATH should be inlined into the helper module


you can use a strategy like `expand_next_js_template_no_imports` but more manual and transform the source file to include these 'manually'

turbopack/crates/turbopack-ecmascript/js/src/worker/browser/createWorker.ts
```diff
@@ -0,0 +1,78 @@
...
+ * `[TURBOPACK_NEXT_CHUNK_URLS, ASSET_SUFFIX, ...WORKER_FORWARDED_GLOBALS values]`
+ *
+ * @param bindings worker runtime config from `__turbopack_context__.b`
+ * @param WorkerConstructor The Worker or SharedWorker constructor
+ * @param entrypoint path to the worker entrypoint chunk
+ * @param moduleChunks list of module chunk paths to load
+ * @param workerOptions options to pass to the Worker constructor (optional)
+ */
+export default function createWorker(
+  bindings: WorkerBindings,
```
lukesandberg: access these by calling a new shortcut __turbopack_chunk_base_and_asset_suffix_

turbopack/crates/turbopack-ecmascript/src/runtime_functions.rs
```diff
@@ -131,4 +131,8 @@ pub const TURBOPACK_RUNTIME_FUNCTION_SHORTCUTS: [(&str, &TurbopackRuntimeFunctio
     ),
     ("__turbopack_wasm__", TURBOPACK_WASM),
     ("__turbopack_wasm_module__", TURBOPACK_WASM_MODULE),
+    (
```
lukesandberg, line 135: add a new shortcut for your new thing

turbopack/crates/turbopack-ecmascript/src/worker_chunk/module.rs
```diff
@@ -130,6 +133,31 @@ impl WorkerLoaderModule {
         ))
     }
 
+    /// `createWorker` is stored in a module; for each worker we need to
+    /// load, we require this module and then use it.
+    #[turbo_tasks::function]
+    async fn create_worker_module(self: Vc<Self>) -> Result<Vc<Box<dyn Module>>> {
```
lukesandberg, line 140: you can just accept `&self`

lukesandberg: or even better, make this a free function and just accept WorkerType, you may need to derive TaskInput for WorkerType

turbopack/crates/turbopack-ecmascript/src/worker_chunk/module.rs
```diff
@@ -267,11 +313,12 @@ impl EcmascriptChunkPlaceable for WorkerLoaderModule {
                 formatdoc! {
```
lukesandberg, line 309: this is
module.exports = function(ctor, opts) 
 {
    return require(id).default(stuff thats going away, {entrypoint}, ctor opts)
}

but instead it could be
module.exports = require(id).default(entrypoint)

and the create worker module would return the factory function

turbopack/crates/turbopack-ecmascript/src/worker_chunk/module.rs
```diff
@@ -130,6 +133,31 @@ impl WorkerLoaderModule {
...
+        Ok(this
+            .asset_context
+            .process(
+                Vc::upcast(FileSource::new(
+                    embed_directory!("turbopack-ecmascript", "$CARGO_MANIFEST_DIR/js/src")
+                        .root()
+                        .await?
+                        .join(&helper)?,
+                )),
+                ReferenceType::EcmaScriptModules(EcmaScriptModulesReferenceSubType::Import),
```
lukesandberg: use EcmaScriptModulesReferenceSubType::ImportPart(ModulePart:;Export(rcstr!('default'))

turbopack/crates/turbopack-ecmascript/src/worker_chunk/module.rs
sampoder: Talked IRL, needed to use `this.asset_context` which made this difficult

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-types.d.ts
sampoder: I ended up passing these as arguments because I don't have access to `chunking_context` in `references()`, which calls `create_worker_module()`.

turbopack/crates/turbopack-ecmascript/src/worker_chunk/module.rs
sampoder: talked irl, `self` is needed later in the function as well

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: ended up moving away from this as the object was replaced with simple constants

turbopack/crates/turbopack-ecmascript/src/worker_chunk/module.rs
```diff
@@ -264,14 +311,20 @@ impl EcmascriptChunkPlaceable for WorkerLoaderModule {
                     .map(|chunk_data| EcmascriptChunkData::new(chunk_data))
                     .collect();
 
+                // The worker base-path override (`WORKER_BASE_PATH`) and the list
+                // of globals to forward (`WORKER_FORWARDED_GLOBALS`).
+                let worker_base_path = chunking_context.worker_asset_prefix().await?;
+                let forwarded_globals = chunking_context.worker_forwarded_globals().await?;
```
lukesandberg: nit: combine these into a single accessor on chunking_context

turbopack/crates/turbopack-tests/tests/snapshot/imports/ignore-comments/output/[root-of-the-server]__08ofy-h._.js
```diff
@@ -0,0 +1,112 @@
...
+//
+// The chunk-URL builder, the chunk base path and the asset suffix are read from
+// the shared `__turbopack_chunk_relative_url__` / `__turbopack_chunk_base_path__`
+// / `__turbopack_chunk_asset_suffix__` runtime primitives. The worker base-path
+// override and forwarded-global names are baked into the loader at build time and
+// the forwarded-global values are read from `globalThis`.
+__turbopack_context__.s([
+    "default",
+    ()=>generateCreateWorker
+]);
```
Comment (lukesandberg) [inline turbopack/crates/turbopack-tests/tests/snapshot/imports/ignore-comments/output/[root-of-the-server]__08ofy-h._.js:25]: im still confused why are generating this structur

it should be 

```
__turbopack_context__.s([

    "default", 0 generateCreateWorker
]);
```

but at the bottom of the file

so something must be wrong with how we are either creating the refference or instantiating the module

Comment (sampoder) [reply turbopack/crates/turbopack-tests/tests/snapshot/imports/ignore-comments/output/[root-of-the-server]__08ofy-h._.js:25]: So I think something is triggering `export_usage_info.is_circuit_breaker` :/

Comment (sampoder) [reply turbopack/crates/turbopack-tests/tests/snapshot/imports/ignore-comments/output/[root-of-the-server]__08ofy-h._.js:25]: talked IRL - related to the testing enviroment

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-types.d.ts
sampoder: added a code gen to do this!

turbopack/crates/turbopack-ecmascript/js/src/worker/browser/createWorker.ts
```diff
@@ -0,0 +1,91 @@
...
+  // other when `CHUNK_BASE_PATH` (= `assetPrefix`) is a cross-origin CDN.
+  // `null` falls back; an empty string is treated as a literal empty prefix.
+  const workerBasePath =
+    _TURBOPACK_WORKER_BASE_PATH_ ?? __turbopack_chunk_base_path__
+
+  const chunkUrls = moduleChunks
+    .map((chunk) => __turbopack_chunk_relative_url__(chunk, workerBasePath))
+    .reverse()
+  const params: unknown[] = [chunkUrls, __turbopack_chunk_asset_suffix__]
+  for (const globalName of _TURBOPACK_WORKER_FORWARDED_GLOBALS_) {
```
lukesandberg: nit:
```
const globals = _TURBOPACK_WORKER_FORWARDED_GLOBALS_
for (let i =0; i < globals.length; i++ ) {
   params.push((globalThis as ...)[globals[i]])
}
```

indexed iteration remains faster than `for...of`

turbopack/crates/turbopack-ecmascript/js/src/worker/node/createWorker.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param workerOptions options to pass to the Worker constructor (optional)
+ */
+function createWorker(
+  WorkerConstructor: { new (path: string, options?: object): unknown },
+  workerPath: string,
+  workerOptions?: { workerData?: unknown; [key: string]: unknown }
+): unknown {
+  // Build the forwarded globals object
+  const forwardedGlobals: Record<string, unknown> = {}
+  for (const name of _TURBOPACK_WORKER_FORWARDED_GLOBALS_) {
```
lukesandberg: ditto

turbopack/crates/turbopack-ecmascript/src/references/worker.rs
```diff
@@ -400,3 +402,55 @@ impl WorkerAssetReferenceCodeGen {
...
+    ) -> Result<CodeGeneration> {
+        let options = chunking_context.worker_configuration_options().await?;
+        let value: RcStr = match self.placeholder {
+            WorkerGlobalPlaceholder::ForwardedGlobals => {
+                StringifyJs(&options.forwarded_globals).to_string()
+            }
+            WorkerGlobalPlaceholder::BasePath => StringifyJs(&options.asset_prefix).to_string(),
+        }
+        .into();
+        let replacement = parse_single_expr_lit(&value);
```
lukesandberg: im guessing there is a more direct/faster way to construct an expr than parsing an expr

e.g. for basepath it should be `Expr::Lit(Str(options.asset_prefix.into())` and for the array case there should be a similar pattern

turbopack/crates/turbopack-ecmascript/src/worker_chunk/module.rs
```diff
@@ -23,10 +26,15 @@ use crate::{
...
-    runtime_functions::{TURBOPACK_CREATE_WORKER, TURBOPACK_EXPORT_VALUE},
-    utils::StringifyJs,
+    runtime_functions::{TURBOPACK_EXPORT_VALUE, TURBOPACK_REQUIRE},
+    utils::{StringifyJs, StringifyModuleId},
 };
 
+#[turbo_tasks::function]
+fn embed_fs() -> Vc<Box<dyn FileSystem>> {
+    embed_directory!("turbopack-ecmascript", "$CARGO_MANIFEST_DIR/js/src")
+}
```
lukesandberg: this should be somewhere more central since it is for the whole crate, just move to `lib.rs`?

lukesandberg (approved):
nice!

lukesandberg:
oh be sure to update the PR description it has become a little bit stale

PR #94387, opened 2026-06-03
docs: document Cache Components behavior for navigation hooks

aurorascharff:
### What?

Adds a `## Behavior` → `### Cache Components` section to the API reference pages for `usePathname`, `useSelectedLayoutSegment`, and `useSelectedLayoutSegments`.

### Why?

When `cacheComponents` is enabled, these hooks suspend during prerendering on routes that have dynamic params not covered by `generateStaticParams`. This was previously only hinted at with a one-line note on `usePathname` and undocumented on the layout segment hooks, even though they share the same behavior.

### How?

- `usePathname`: replaced the terse one-line note with a `## Behavior` → `### Cache Components` section explaining when it resolves on the server vs. suspends. Also added a link to the [Preventing flash before hydration](/docs/app/guides/preventing-flash-before-hydration) guide in the existing rewrite-mismatch example, and cleaned up a stray em dash.
- `useSelectedLayoutSegment`: added the same `### Cache Components` section.
- `useSelectedLayoutSegments`: added the same `### Cache Components` section.

The rewrite/Proxy hydration-mismatch gotcha is intentionally not duplicated to the layout segment pages — those hooks derive their value from the routing tree, not the URL pathname, so rewrites don't produce the same mismatch.

Documents existing behavior only — no API surface changes.

### DIFF
--- a/docs/01-app/03-api-reference/04-functions/use-pathname.mdx
+++ b/docs/01-app/03-api-reference/04-functions/use-pathname.mdx
@@ -5,8 +5,6 @@ description: API Reference for the usePathname hook.
 
 `usePathname` is a **Client Component** hook that lets you read the current URL's **pathname**.
 
-> **Good to know**: When [`cacheComponents`](/docs/app/api-reference/config/next-config-js/cacheComponents) is enabled `usePathname` may require a `Suspense` boundary around it if your route has a dynamic param. If you use `generateStaticParams` the `Suspense` boundary is optional
-
 ```tsx filename="app/example-client-component.tsx" switcher
 'use client'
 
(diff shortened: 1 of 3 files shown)


docs/01-app/03-api-reference/04-functions/use-pathname.mdx
```diff
@@ -67,6 +65,17 @@ const pathname = usePathname()
...
+
+- **Static routes and routes with [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params)**: the pathname is known during prerendering, so `usePathname` resolves on the server and no `Suspense` boundary is required.
+- **Routes with dynamic params not covered by `generateStaticParams`**: the param is a [fallback param](/docs/app/api-reference/functions/generate-static-params#all-paths-at-build-time) that isn't known until request time. `usePathname` suspends during prerendering, so the closest `Suspense` boundary renders its fallback. Wrap the component (or a parent) in `Suspense`, otherwise the build fails.
+
+This applies even when the component that calls `usePathname` is itself static. For example, a sidebar with active links rendered in a layout suspends on any page below it that has an unknown dynamic param. To keep the rest of the layout prerendered, wrap the component that calls `usePathname` (or a parent) in a `Suspense` boundary with a fallback.
```
icyJoseph:
```suggestion
When [`cacheComponents`](/docs/app/api-reference/config/next-config-js/cacheComponents) is enabled, `usePathname` may require a [`Suspense`](https://react.dev/reference/react/Suspense) boundary. This depends on whether the pathname can be resolved during prerendering.

- **Static routes and routes with [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params)**: every route segment, including dynamic params, is known at build time. The pathname can be resolved during prerendering, so `usePathname` resolves on the server and no `Suspense` boundary is required.
- **Routes with dynamic params not covered by `generateStaticParams`**: the param is a [fallback param](/docs/app/api-reference/functions/generate-static-params#all-paths-at-build-time) that is not known until request time. The pathname cannot be resolved during prerendering, so `usePathname` suspends. Wrap the component (or a parent) in a `Suspense` boundary so its fallback can be rendered during prerendering; otherwise, the build fails.

This applies even when the component that calls `usePathname` is itself static. For example, a sidebar with active links rendered in a layout suspends on any page below it that has an unknown dynamic param. To keep the rest of the layout prerendered, wrap the component that calls `usePathname` (or a parent) in a `Suspense` boundary with a fallback.
```

icyJoseph (approved):

PR #94422, opened 2026-06-03
test(instant-validation): document usePathname() stack-trace attribution today

aurorascharff:
Three bug-witness fixtures + snapshots that pin down today's `usePathname()` stack-frame attribution. The snapshots are intentionally wrong — they bake the current bug so a fix shows up clearly in diff review.

- **Build, no wrapper** — function frame is `<unknown>`.
- **Build, three wrappers deep** — function frame is an SWC-minified letter; non-deterministic across module-graph orderings, snapshot normalizes to `<minified>`.
- **Dev, fallback param** — overlay attributes the error to the parent's JSX line, reports `E1265` "uncached data" with Cache/Stream/Block cards, and drops the hook caller from the stack.

Dev fixture lives under `app/default/` (no root `<Suspense>` above body) so the validation hole isn't swallowed before the Insight fires.

Related: #94235, #94387.

### DIFF
--- a/test/e2e/app-dir/instant-validation-build/instant-validation-build.test.ts
+++ b/test/e2e/app-dir/instant-validation-build/instant-validation-build.test.ts
@@ -821,11 +821,7 @@ describe('instant-validation-build', () => {
       const result = await prerender(
         '/(default)/pathname/invalid-use-pathname-missing-params-no-wrapper/[one]/[two]'
       )
-      // Snapshot the full validation error so any change to the stack frame
-      // attribution (good or bad) shows up clearly. Today's behavior: the file,
-      // line, and column are correct, but the function name is reported as
-      // `<unknown>` instead of `PathnameReader`. When the framework starts
-      // recovering the real function name from sourcemaps, update this snapshot.
+      // Today the function name is `<unknown>` instead of `PathnameReader`.
       expect(extractBuildValidationError(result.cliOutput))
         .toMatchInlineSnapshot(`
        "Error: Route "/pathname/invalid-use-pathname-missing-params-no-wrapper/[one]/[two]" called usePathname() but param "two" is not defined in the \`unstable_samples\` of \`unstable_instant\`. usePathname() requires all route params to be provided.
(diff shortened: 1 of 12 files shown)


test/e2e/app-dir/instant-validation/instant-validation.test.ts
```diff
@@ -1893,6 +1893,50 @@ describe('instant validation', () => {
...
+           "label": "Blocking Route",
+           "source": "app/default/invalid-use-pathname-no-samples/[slug]/page.tsx (11:7) @ Page
+         > 11 |       <PathnameLabel />
+              |       ^",
+           "stack": [
+             "Page app/default/invalid-use-pathname-no-samples/[slug]/page.tsx (11:7)",
+           ],
+         }
+        `)
+      })
```
eps1lon: This should include a dev assertion to ensure there is a proper stack.

aurorascharff: Like this? https://github.com/vercel/next.js/pull/94422/changes/d7285c650edd2a21283951055c36362641093b22

test/e2e/app-dir/instant-validation/instant-validation.test.ts
```diff
@@ -1893,6 +1893,37 @@ describe('instant validation', () => {
           expectNoBuildValidationErrors(result)
         }
       })
+
+      // Today the dev overlay misattributes usePathname() errors to the
+      // parent's JSX line instead of the actual hook call site. When the
```
eps1lon: FYI this is a React decision. Error stacks have less fidelity in production for performance reasons.

aurorascharff: Thats what i learned, but @gnoff has a PR to change that i think?

aurorascharff: https://github.com/facebook/react/pull/36586

test/e2e/app-dir/instant-validation-build/instant-validation-build.test.ts
```diff
@@ -816,6 +816,69 @@ describe('instant-validation-build', () => {
...
+
+    it('error - usePathname() at the top of a Client Component body (no ensureThrows wrapper)', async () => {
+      const result = await prerender(
+        '/(default)/pathname/invalid-use-pathname-missing-params-no-wrapper/[one]/[two]'
+      )
+      // Today the function name is `<unknown>` instead of `PathnameReader`.
+      expect(extractBuildValidationError(result.cliOutput))
+        .toMatchInlineSnapshot(`
+       "Error: Route "/pathname/invalid-use-pathname-missing-params-no-wrapper/[one]/[two]" called usePathname() but param "two" is not defined in the \`unstable_samples\` of \`unstable_instant\`. usePathname() requires all route params to be provided.
+           at <unknown> (app/(default)/pathname/invalid-use-pathname-missing-params-no-wrapper/[one]/[two]/pathname-reader.tsx:6:20)
```
eps1lon, line 828: That doesn't really tell us stack attribution is working because it's just the sync error from misusing samples. The codeframe is the one we finally want. But it's broken at the moment.

PR #94433, opened 2026-06-03
Keep the dev React debug channel on Node streams end to end

unstubbable:
With Node streams enabled (now the default), the dev-only React debug channel was still being round-tripped through web streams repeatedly. The Node variant of `teeStream` wrapped the readable with `Readable.toWeb`, teed it, then unwrapped with `Readable.fromWeb`, applied up to twice as the channel was split for validation and then for SSR and the browser, and `connectReactDebugChannel` additionally converted the readable to a web `ReadableStream` and ran every chunk through a web `TransformStream` batcher before forwarding it over HMR. Each conversion allocates web-stream wrappers and adds per-chunk microtask scheduling that contends for CPU with the render, which is wasteful because the debug payload is non-trivial in dev (async debug info, console replay, large debug strings) and the whole point of the Node streams flag is to avoid that overhead.

This change keeps the channel Node-native whenever Node streams are in use. The Node variant of `teeStream` now fans out through `ReplayableNodeStream` instead of the web round-trip, which fixes every Node-mode caller at once. The Cache Components Node render path, which needs three independent consumers (validation, SSR, and the browser), fans out through a single `ReplayableNodeStream` directly rather than nesting binary tees. `connectReactDebugChannel` now branches on the stream type and, for Node input, batches with a Node `Transform` (`createNodeBufferedTransformStream`, now exported with a byte-length cap to match the web batcher) consumed via `data`/`end`/`error` events, with no `toWeb` conversion. A `NodeDebugChannelPair` type narrows the client-side readable to a Node `Readable` so Node call sites can consume it without casting `AnyStream` down. The web path is left untouched and remains conversion-free.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -3451,7 +3451,7 @@ async function renderToStream(
       ) {
         if (process.env.__NEXT_USE_NODE_STREAMS) {
           // MARK: nodeStreams dev CacheComponents RSC
-          let debugChannel: DebugChannelPair | undefined
+          let debugChannelClientStream: ReplayableNodeStream | undefined
 
           // eslint-disable-next-line @typescript-eslint/no-shadow
           const getPayload = async (requestStore: RequestStore) => {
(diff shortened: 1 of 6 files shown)


timneutkens (approved):

eps1lon:
Somehow this causes the runtime to use installed instead of bundled React. Reverting in https://github.com/vercel/next.js/pull/94459

PR #94430, opened 2026-06-03
refactor: indicate which stage is the last stage in staged prerenders

lubieowoce:
Split out of #93801.

Adds a new config option to `StagedRenderingController` - `finalStage`. Staged prerenders must indicate which stage they plan to abort after.

This lets us generalize the "wait for the runtime stage" bit  in `createHangingInputAbortSignal` to now mean "wait until the last stage of the prerender". This will be used more in #93801, where 
- we need to be able to abort a runtime prerender earlier than the Runtime stage (only rendering a session shell, without link data), meaning that we can't hardcode the stage there
- we add additional shell stages to static prerenders, meaning that we must wait until the last stage (Static) until hanging inputs can be aborted

PR #94441, opened 2026-06-03
[App Shells] staged shell rendering in cached navs

lubieowoce:
Continuing the app shell implementation work. This PR implements recovering *static* shells from staged dynamic requests, so that we can recover a shell from a navigation. This is done using a byte offset, analogous to existing recovery of the static stage.

in #93801 we already implemented the session shell codepath, in `spawnRuntimePrefetchWithFilledCaches`, so we're not concerned with session shells here. This means that for a navigation, we can resolve static params in the static stage -- we don't care about recovering a  session shell, so we don't have to delay link data until the runtime stage.

One notable difference from the previous PR is that we don't set the byte length to `null` if the whole response is the same as the shell stage. We don't really expect this to happen -- if a page is fully static, it shouldn't be doing a navigation request anyway, so this doesn't seem worth optimizing.

Highlight: changing the timing of when cookies/headers resolve affected some Sync IO tests -- params/searchParams were now unresolved at the abort point, and triggered a bunch of unhandled rejections. This led to the discovery that we our unhandled rejection filter is not applied to build validation, which is why we were seeing mysterious duplicated error logs in those tests. This is also fixed here since now it became more of a problem.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -287,6 +287,7 @@ import type {
 import { ResponseCookies } from '../web/spec-extension/cookies'
 import { isInstantValidationError } from './instant-validation/instant-validation-error'
 import { createPromiseWithResolvers } from '../../shared/lib/promise-with-resolvers'
+import { RENDER_STAGES_BY_DATA_KIND } from '../dynamic-rendering-utils'
 
 export type GetDynamicParamFromSegment = (
   // The LoaderTree to extract the dynamic param from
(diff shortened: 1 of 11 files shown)


lubieowoce:
* **#94442** <a href="https://app.graphite.com/github/pr/vercel/next.js/94442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94441** <a href="https://app.graphite.com/github/pr/vercel/next.js/94441?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94441?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94430** <a href="https://app.graphite.com/github/pr/vercel/next.js/94430?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

gnoff (approved):

gnoff (approved):

PR #91426, opened 2026-03-15
Expose instant devtools for production builds under `exposeInstantDevToolsInProductionBuild`

devjiwonchoi:
This PR exposes the initial version of Instant DevTools for builds for testing purposes. It is not recommended to ship on the user-facing production builds.

https://github.com/user-attachments/assets/ef34ad1e-4500-4f22-ac5c-51fac5bd0d17

PR #94443, opened 2026-06-03
Expose Instant Devtools for production builds

devjiwonchoi:
### What?

Exposes the **Instant Navigation insights** devtools widget in **production builds**, behind a new experimental flag `experimental.exposeInstantDevToolsInProductionBuild`. When enabled, a small widget renders in production pages, letting you capture and inspect the instant (prefetched / static) UI shown during navigations: freeze the next client navigation or page reload, view the prefetched / pre-rendered UI, then continue rendering to stream dynamic data. This is the same insight the dev overlay's Instant Navigation panel provides, now available against a real production build.

Re-implements #91426 on top of current `canary` (the original went stale while the instant-navigation feature was reworked: cookie format, cookie-logic split, dev-panel rewrite, `self.__next_instant_test` static-shell flow, separate webpack dev/prod client entries).

https://github.com/user-attachments/assets/960d7f0a-857d-481b-b79b-f17aeb0394f2

### Why?

The instant-navigation capture mechanism is already production-capable (gated by `__NEXT_EXPOSE_TESTING_API`), but was only surfaced through the dev overlay (dev only). This flag exposes a self-contained widget so the same navigation insights can be gathered against a production build for profiling / testing.

> [!WARNING]
> Intended for profiling and testing purposes only. Not recommended for user-facing production deployments.

### Verification

- root `tsc --noEmit` (test files) and the `next` build (types) — pass
- eslint + prettier on changed files — pass
- `pnpm test-start-turbo test/production/app-dir/instant-devtools-prod` — 5/5 pass
- `pnpm test-start-webpack test/production/app-dir/instant-devtools-prod` — 5/5 pass
- **DCE**: built flag-on vs flag-off apps in both webpack and turbopack; the widget appears in a client chunk only when the flag is set (eliminated otherwise)
- Dev overlay instant-navs panel suite still passes (16/16) with the cookie move


devjiwonchoi:
will re-impl when we're sure of use case

PR #94457, opened 2026-06-04
Stream Cache Components dev render instead of restarting on cache miss

unstubbable:
When Cache Components is enabled, `next dev` previously simulated a production loading experience on every cold request. The render did a prospective pass to detect cache misses, and on any miss it waited for every cache to fill via `cacheSignal.cacheReady()` and then restarted the render with warm caches before streaming anything, so the browser saw nothing until the slowest cache had filled. Every cold load blocked on cache population.

This change replaces the restart-on-cache-miss flow with a single non-abandoning staged render that streams immediately and fills caches as a side effect. On a cold load the Suspense fallbacks stream right away and the cached content resolves as its cache fills; on a warm reload the staged progression matches the previous no-cache-miss path. The render is split into clearly owned pieces: `setUpStagedDevRender` builds the staged controller, cache signal, and resume cache; `streamStagedRenderInDev{Node,Web}` runs the streaming render and reports a result once the stream has fully finished; and `stagedRenderWithCachesInDev{Node,Web}` returns the stream and leaves the validation follow-up detached so it never blocks the response.

The render advances its stages in sequential tasks, and the stream is not handed back the instant it exists: it is held until the render has advanced through the stage whose content belongs in the shell. That is the static stage for initial loads, HMR refreshes, and plain navigations, or the runtime stage for client navigations to a route with a runtime prefetch config, whose runtime-prefetchable content the navigation's prefetch would have settled. It never waits for the dynamic stage. Buffering the shell before the first flush keeps the streaming renderer from emitting a premature Suspense fallback for content that belongs in the shell, and it mirrors production, where the static shell (plus runtime-prefetchable content where configured) is served and the remaining holes stream in as fallbacks.

Two internal reads that would otherwise register as synchronous IO and wrongly force the render to the dynamic stage, the cache handler's tag-expiry clock check and the hot reloader's module-scope dev client id, are now read untracked: via `performance.timeOrigin + performance.now()` like the `'use cache'` handler, and only in the browser where the HMR connection reads it, respectively.

Cache Components rules validation now runs in that background follow-up, once the streamed render has fully settled. `planDevValidation` inspects the finished render and picks one of three paths: forward an invalid dynamic usage error the streamed render already recorded and stop (for example a request API used inside `'use cache'`); validate the streamed render's own chunks when it neither missed caches nor hit sync IO; or, when it did either, validate a dedicated warm-cache render instead. Because that warm render reads the filled caches back rather than filling them, it can surface an invalid dynamic usage error the cold streamed render cannot, such as a nested dynamic `use cache` cache life that propagated to a parent with no explicit `cacheLife`; that error is forwarded and validation is skipped, just as one recorded by the streamed render is.

Since cold loads no longer block on cache fills, the transient cache-status indicator that reflected that wait is no longer emitted; a follow-up will instead add an indicator that tells the user whether a render streamed with cache misses, and so wasn't representative of production.


packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4767,314 +4688,628 @@ async function renderToStream(
...
+  stream: AnyStream
+  resultPromise: Promise<StagedDevRenderResult>
+}> {
+  const { ComponentMod } = ctx.renderOpts
   const { clientModules } = getClientReferenceManifest()
 
-  // Note: The stage controller starts out in the `Before` stage,
-  // where sync IO does not cause aborts, so it's okay if it happens before render.
-  const initialRscPayload = await getPayload(requestStore)
+  const streamReady = new DetachedPromise<{
```
lubieowoce: nit: i feel like we should standardize on `createPromiseWithResolvers`, they're functionally the same but cPWR is based on a standard

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4767,314 +4688,628 @@ async function renderToStream(
...
-      advanceStageIfNoCacheMiss(RenderStage.Static)
+      checkForCacheMiss()
+      stageController.advanceStage(RenderStage.Static)
     },
     () => {
-      advanceStageIfNoCacheMiss(RenderStage.ShellEarlyRuntime)
+      checkForCacheMiss()
+      // The static stage's chunks flushed during the macrotask gap since the
+      // previous task advanced to `Static`, so the static shell is buffered
+      // now. Hand the stream to the caller before advancing into the runtime
```
lubieowoce: is "macrotask gap" an actual phrase? i haven't heard that before. either way, this phrasing is a bit confusing to me. it'd be enough to say:

> The static stage's chunks flushed in the previous task, so the static shell is buffered now.

lubieowoce: (same goes for the runtime one below)

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4767,316 +4687,628 @@ async function renderToStream(
...
+    runtimeStageEndTime,
+    accumulatedChunks,
+  } = result
+
+  // A cache-miss render records its `invalidDynamicUsageError` while filling,
+  // so its verdict isn't final until the fills settle. Wait for that here (a
+  // no-op when the render didn't miss) before reading the work store below.
+  if (hadCacheMiss) {
+    await cacheSignal.cacheReady()
+  }
```
lubieowoce: nit: seems a bit weird to have this here in a function called "plan" (which seems like it should already get all the inputs it needs w/o extra waiting). can't we wait for everything necessary outside, before calling this?

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4767,316 +4687,628 @@ async function renderToStream(
...
+ * once the render's stream has fully finished.
+ */
+interface StagedDevRenderResult {
+  readonly hadCacheMiss: boolean
+  readonly syncInterruptReason: Error | null
+  readonly startTime: number
+  readonly staticStageEndTime: number
+  readonly runtimeStageEndTime: number
+  readonly accumulatedChunks: AccumulatedStreamChunks
+}
```
lubieowoce: 5 of these fields overlap, can we do a `Pick<...>` or something here? don't love copy pasting these around. or at least give them a consistent order

Image

gnoff (approved):
Codex claims this comment is stale: https://github.com/vercel/next.js/blob/50f2ebdeb66557c062108ebb2d90aebdccb5036d/packages/next/src/server/app-render/app-render.tsx#L4528-L4536

And that the Runtime branch from https://github.com/vercel/next.js/blob/50f2ebdeb66557c062108ebb2d90aebdccb5036d/packages/next/src/server/app-render/app-render.tsx#L1328-L1350 is also untested

PR #94460, opened 2026-06-04
Allow `unstable_retry` from TS plugin check in Next.js Error APIs

devjiwonchoi:
### What?

Stop the Next.js TypeScript plugin from flagging the framework-injected `unstable_retry` prop in `error.tsx` and `global-error.tsx` with error `71007` (`INVALID_CLIENT_ENTRY_PROP`). The plugin already exempts `reset`; this extends that exemption to `unstable_retry`.

### Why?

`unstable_retry` is injected by the framework into error boundaries: it is part of the `ErrorInfo` type (`error`, `reset`, `unstable_retry`) in `packages/next/src/client/components/error-boundary.tsx` and is passed to the user's error component, not supplied by the user. Like `reset`, it therefore does not need to be serializable.

The file-convention docs (`docs/01-app/03-api-reference/03-file-conventions/error.mdx`) document `unstable_retry` as an error-component prop (added in `v16.2.0`) and recommend it over `reset`. The exact documented pattern, `export default function Error({ error, unstable_retry })`, was being false-flagged by the plugin, so a user copying the docs would hit a spurious diagnostic.

image

### How?

Extend the existing error-file `reset` exemption in `packages/next/src/server/typescript/rules/client-boundary.ts` to also allow `unstable_retry`. The exemption stays scoped to those two prop names in `error` / `global-error` files; every other non-serializable function prop is still flagged.

Add regression tests for both `error.tsx` and `global-error.tsx`, each with a `_notExempt` negative control that confirms the exemption stays scoped (an ordinary function prop is still flagged). Update the manual VSCode fixtures and the test README to match.

### Verification

- `pnpm test-dev-turbo test/development/typescript-plugin/client-boundary/client-boundary.test.ts` (5/5 passed; the harness loads the compiled plugin from `dist`, so green also confirms the build reflects the change)
- `pnpm --filter=next types` (passed)
- `pnpm --filter=next build` (passed)
- `pnpm typescript` root `tsc` for the changed test files (passed)
- Not run: full CI suite (will run on the PR)

### DIFF
--- a/packages/next/src/server/typescript/rules/client-boundary.ts
+++ b/packages/next/src/server/typescript/rules/client-boundary.ts
@@ -75,13 +75,17 @@ const clientBoundary = {
                   const maybeServerAction =
                     propName === 'action' || /.+Action$/.test(propName)
 
-                  // There's a special case for the error file that the `reset` prop
-                  // is allowed to be a function:
+                  // There's a special case for error files where the
+                  // framework-injected `reset` and `unstable_retry` props are
+                  // allowed to be functions. Next.js provides these props, the
+                  // user doesn't pass them, so they don't need to be
+                  // serializable.
                   // https://github.com/vercel/next.js/issues/46573
-                  const isErrorReset =
-                    (isErrorFile || isGlobalErrorFile) && propName === 'reset'
+                  const isErrorBoundaryFunctionProp =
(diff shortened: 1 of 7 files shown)


icyJoseph (approved):
Verified on a project too. 

Image

PR #94462, opened 2026-06-04
Reland the debug channel on native Node streams, fixed for React 18

unstubbable:
This relands #94433, which was reverted in #94459 after it broke app rendering in dev on React 18. The first commit reapplies the original change verbatim, and the second commit fixes the regression that prompted the revert.

The regression came from `connectReactDebugChannel` reaching into `app-render/stream-ops.node` for its Node byte batcher transform. That render-layer module carries unconditional `react-dom/server` and `react-dom/static` value imports, and the dev server's hot-reloaders run in the precompiled server runtime where `react-dom/*` is not aliased to the vendored React. The `require` therefore resolved `react-dom/static` against the app's installed React, which on React 18 has no `./static` subpath export and threw `ERR_PACKAGE_PATH_NOT_EXPORTED`.

The fix extracts the batcher into its own React-free module, `stream-utils/node-buffered-transform-stream`, which depends only on `node:stream`. Both `stream-ops.node` and the debug channel import it from there, so the dev server can batch Node stream chunks without dragging the render layer's React imports into an unaliased runtime. The debug channel's runtime branch on the stream type is unchanged and the web path already used a React-free batcher, so no `.node`/`.web` split is needed.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -3593,7 +3593,7 @@ async function renderToStream(
       ) {
         if (process.env.__NEXT_USE_NODE_STREAMS) {
           // MARK: nodeStreams dev CacheComponents RSC
-          let debugChannel: DebugChannelPair | undefined
+          let debugChannelClientStream: ReplayableNodeStream | undefined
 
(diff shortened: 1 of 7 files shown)


eps1lon (approved):
Makes review very smooth if this is split into revert and fix, thank you!

PR #94464, opened 2026-06-04
Trace viewer: store span IDs in a Vec instead of a HashMap

andrewimm:
The tracing layer allocates span IDs sequentially, so when we index a trace file we expect nearly every ID to be encountered. This was backed up with analyzing some actual traces: I measured seen ID density to be somewhere between 0.997 and 0.999 (where 1.0 == every ID was encountered). This allows us to just store the span id -> index mapping in a `Vec<Option<T>>`, with linear lookup and a 3-4x reduction in memory.

The dense map itself is just a convenience data structure. Most of the methods are inlined for speed.

On medium-to-large traces I'm seeing an 8% reduction in file processing time.

### DIFF
--- a/turbopack/crates/turbopack-trace-server/src/reader/turbopack.rs
+++ b/turbopack/crates/turbopack-trace-server/src/reader/turbopack.rs
@@ -67,9 +67,53 @@ enum InternalRowType {
     },
 }
 
+/// Dense map from external span id to `SpanIndex`.
+///
+/// The turbopack tracing layer assigns span ids sequentially starting at 1, so
+/// the key space is effectively contiguous (measured density ≈ 1.0 on real
+/// traces). Indexing a `Vec` beats hashing an `FxHashMap` on the ingest hot path
+/// and uses less memory; `Option<SpanIndex>` is a single `usize` thanks to the
+/// `NonZeroUsize` niche. Ids that have not been seen yet read back as `None`,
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-trace-server/src/reader/turbopack.rs
```diff
@@ -67,9 +67,53 @@ enum InternalRowType {
...
+/// Dense map from external span id to `SpanIndex`.
+///
+/// The turbopack tracing layer assigns span ids sequentially starting at 1, so
+/// the key space is effectively contiguous (measured density ≈ 1.0 on real
+/// traces). Indexing a `Vec` beats hashing an `FxHashMap` on the ingest hot path
+/// and uses less memory; `Option<SpanIndex>` is a single `usize` thanks to the
+/// `NonZeroUsize` niche. Ids that have not been seen yet read back as `None`,
+/// matching the old `HashMap::get` semantics relied on for out-of-order parents.
+#[derive(Default)]
+struct DenseIdMap {
```
lukesandberg, line 79: LinearIdMap might be a more relatable name?

should we just use spanids as spanindex directly and accept that while span ids do not arrive in order they do eventually 'fill out'?

lukesandberg (approved):
this id good to go, curious about your thoughts on just collapsing the spand-id spanindex mapping altogether

PR #94490, opened 2026-06-05
Warn on prefetch={true} navigation without Partial Prefetching (dev)

acdlite:
Under Cache Components, a `<Link prefetch={true}>` pointing at a route that hasn't enabled Partial Prefetching quietly falls back to a legacy full prefetch: it pulls down the route's dynamic data instead of just the static shell, defeating the static/dynamic split. In development we now surface this with a console error that explains the fallback and points at how to opt in — app-wide with `partialPrefetching`, or per-route with `unstable_prefetch = 'partial'`.

The check runs when the navigation happens, not when the link is prefetched. Dev doesn't prefetch, so navigation is the only point where both the originating link's fetch strategy and the resolved route tree are in hand; it also means an app that has just turned on Cache Components isn't flooded with warnings for every `prefetch={true}` link on the page, since you only hear about a route when you actually navigate to it. The decision reads a single hint at the root of the resolved route tree that records whether any segment in the route opted into Partial Prefetching.

Scope: the warning fires on every qualifying navigation, including ones to fully static routes where a full prefetch costs nothing. Suppressing it for fully-static targets is deferred — the route-level "has dynamic data" signal isn't cheaply available on cache-served repeat navigations.

### DIFF
--- a/packages/next/src/build/segment-config/app/app-segment-config.ts
+++ b/packages/next/src/build/segment-config/app/app-segment-config.ts
@@ -39,13 +39,19 @@ const InstantConfigSchema = z.union([
 const PrefetchSchema = z.enum([
   'auto',
   'partial',
+  'unstable_eager',
   'force-disabled',
   'force-runtime',
 ])
 
 export type Instant = InstantConfig | true | false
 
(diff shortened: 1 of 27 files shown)


acdlite:
Closing — superseded/not needed; the App Shells work continues in #94510.

PR #94494, opened 2026-06-05
Specialize client hook prerender abort reasons

gnoff:
Specialize Cache Components prerender abort errors for `useParams`, `usePathname`, `useSearchParams`, `useSelectedLayoutSegment`, and `useSelectedLayoutSegments` instead of reporting a generic abort reason.

React now allows errors observed after abort begins but before the final abort task runs to replace the generic abort reason. Client hook promises can use this window to report which hook blocked prerendering while preserving higher-priority synchronous I/O errors.


packages/next/src/server/dynamic-rendering-utils.ts
```diff
@@ -48,15 +90,35 @@ export function makeHangingPromise<T>(
...
+export function makeClientHookHangingPromise<T>(
+  signal: AbortSignal,
+  route: string,
+  expression: string
+): Promise<T> {
+  return makeHangingPromiseWithError(
+    signal,
+    expression === 'useSearchParams()'
+      ? new ClientHookDynamicError(route, expression)
+      : new ParamClientHookDynamicError(route, expression)
```
lubieowoce: nit: i don't love this branching here, feels like `expression` should only be used for error text. can't we have the caller pick if the promise is params-dependent or not? `useDynamicParams()` and `useDynamicSearchParams()` know which variant they are

lubieowoce (approved):
looks sensible, i'd prefer we got rid of the `expression === 'useSearchParams()'` bit, but i'm not gonna die on that hill

packages/next/src/server/dynamic-rendering-utils.ts
gnoff: agreed

PR #94498, opened 2026-06-06
[tubopack] migrate rcstr! to use scattered collect

lukesandberg:
### What?
Adopt `[scattered_collect`](https://docs.rs/scattered-collect/latest/scattered_collect/) for collecting static rcstr values. 

### Why?

Built by our good friend @mmastrac to solve some of our problems.  Replaces inventory with a simpler linker driver approach.

Basically inventory works using constructor functions to build a linked list, then iterating follows the list.  `scattered_collect` on the otherhand uses a linker section for the data, so it all gets moved to a dedicated part of the binary and bingo, a slice!.   This means we don't need `init` time logic.

This is preparation for migrating other uses of `inventory` just to prove it out.  Which was a good idea because we found and fixed a bug! See https://github.com/mmastrac/linktime/releases/tag/link-section-0.18.2

## Wasm

scatter_collect supports wasm, but requires us to register a 'section reader' function with the runtime, see https://docs.rs/link-section/latest/link_section/#wasm.  The instructions are staightforward but a little tricky since we rely on `wasm-bindgen` to constrct the `Wasm.Module` instance and it is encapsulated.  So to 'inject' that function we would need to patch some of the bootstrapping.  Given that this registry is only an optimization for deserialization which never happens in a wasm build i just flagged it all off.


lukesandberg:
* **#94503** <a href="https://app.graphite.com/github/pr/vercel/next.js/94503?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94498** <a href="https://app.graphite.com/github/pr/vercel/next.js/94498?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94498?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-rcstr/src/lib.rs
```diff
@@ -520,45 +521,57 @@ pub const fn make_const_prehashed_string(text: &'static str) -> StaticPrehashedS
...
-/// Forwarder around [`inventory::submit!`] that lets the `rcstr!` proc macro
-/// emit a single path it can rely on, without depending on whether
-/// `turbo_rcstr::inventory` is reachable as a macro path in the call site
-/// crate. Macros emitted from a proc macro lose access to the proc macro
-/// crate's deps, so the submission has to bounce through this declarative
-/// macro defined where `inventory::submit!` is in scope.
+// Link-time collection of every `rcstr!` static.
+#[doc(hidden)]
+#[scattered_collect::gather]
+pub static STATIC_RCSTRS: ScatteredSlice<StaticRcStr>;
```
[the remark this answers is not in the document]
mmastrac: Note that WASM should be supported, but requires a supporting JS function: https://docs.rs/link-section/latest/link_section/#wasm

```js
/**
 * Support function for `link-section` crate.
 */
export function readCustomSection(
  wasmModule: WebAssembly.Module,
  wasmInstance: WebAssembly.Instance,
  namePtr: number,
  nameLength: number,
  targetPtr: number,
  targetLength: number,
): number {
    const memory = wasmInstance.exports.memory as WebAssembly.Memory;
    const nameBytes = new Uint8Array(memory.buffer, namePtr, nameLength);
    const sectionName = new TextDecoder().decode(nameBytes);

    const sections = WebAssembly.Module.customSections(wasmModule, sectionName);
    if (sections.length === 0) {
        return 0;
    }

    const section = sections[0];
    const need = section.byteLength;
    if (targetLength < need) {
        return need;
    }

    new Uint8Array(memory.buffer, targetPtr, need).set(new Uint8Array(section));
    return need;
}
```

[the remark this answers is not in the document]
lukesandberg: Thanks im going to document this and just flag off this under wasm

our use of wasm-bindgen makes it tricky to install this env function (and close over the module/instance) and this feature isn't really needed under wasm

mmastrac (commented):
LGTM!

It might be worth using a ScatteredSortedReferenceSlice in the future which would give you two benefits: the hashes would be distributed effectively randomly so you could intern arbitrary strings in `log log N` via interpolation search. 

In addition, each `rcstr` could make use of the cheap `offset_of` method to get an ID that reduces to a trivial subtraction (https://docs.rs/scattered-collect/latest/scattered_collect/sorted_referenced_slice/struct.ScatteredSortedReferencedSlice.html).

There's also `ScatteredMap` that builds the swiss-table-like map in a ctor and should give you allocation-free O(1) lookup after an O(N) ctor runs at boot time.

turbopack/crates/turbo-rcstr/src/lib.rs
```diff
@@ -520,45 +522,56 @@ pub const fn make_const_prehashed_string(text: &'static str) -> StaticPrehashedS
...
 
-/// Wrapper for collecting `rcstr!` static constants via `inventory`.
+/// Wrapper for collecting `rcstr!` static constants at link time.
 #[doc(hidden)]
 pub struct StaticRcStr(pub &'static StaticPrehashedString);
 
-inventory::collect!(StaticRcStr);
+// Link-time collection of every `rcstr!` static.
+//
+// Disabled under wasm because scattered-collect relies on a environment provided function described in https://docs.rs/link-section/latest/link_section/#wasm and installing it is tricky using wasm-bindgen.  Also this is only hear to support desrialization of rcstrs which shouldn't happen under wasm anyway.
```
bgw:
```suggestion
// Disabled under wasm because scattered-collect relies on a environment provided function 
// described in <https://docs.rs/link-section/latest/link_section/#wasm> and installing it is tricky
// using wasm-bindgen. Also this is only here to support deserialization of rcstrs which shouldn't
// happen under wasm anyway.
```

This is possibly problematic for utoo, which does support wasm, though it does seem unlikely they'd care about persistent caching in the browser?

utoo does have some custom environment stuff already though, so maybe they could still support `scattered_collect` without too much trouble? We should just ask them.

bgw (approved):

turbopack/crates/turbo-rcstr/src/lib.rs
lukesandberg: yeah my take is that this is very feasible, just not important.  if utoo wants it it shoudn't be too bad to add

lukesandberg: wait! that means the next pr will break utoo...

PR #94503, opened 2026-06-06
migrate turbo-tasks to scattered collect

lukesandberg:
# What
This uses `scatter_collect` from @mmastrac to register all values, traits and value impls of traits.  now all data can be consistently accessed when constructing the VALUEs registry, eliminating a number of racy access opportunities.

# Why
Scatter collect uses link sections to gather registries at link time, this resolve some inherent brittleness in constructor functions since the relative execution order is **not guaranteed** and it is also not even possible to  _detect_ execution that is too early (early access to an `inventory` collection will miss some entries silently)

We have a single report that appears to be due to a race like this

```
web:dev: thread 'tokio-rt-worker' (41574444) panicked at turbopack/crates/turbo-tasks/src/macro_helpers.rs:274:13:
web:dev: no trait impl registered for value type turbopack_core::module_graph::ModuleGraphImportTracer
web:dev: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I do not entirely understand how this is possible since constructors should complete prior to the first napi entry point (other than `napi::module_init` which is itself a constructor).

Our initialization logic is fundamentally brittle since it relies on

* an `inventory` of all value types
* an `inventory `of all trait types
* a set of ctor functions to register trait implementations (basically (value, trait) pairs)

All of these need to be initialized/invoked before the `VALUES` registry is accessed, but the above panic implies that the ctor function for `ModuleGraphImportTracer` to register its implementation of `ImportTracer` hasn't occurred yet.

By moving the registies to link time we completely eliminate all startup races since the linker does all the 'gathering', now the only possible concern is that we attempt to downcast a trait prior the the `VALUES` registry completing which does not appear to be possible.

# Binary size impact

Measured on `next-swc.darwin-arm64.node` built with `pnpm --filter @next/swc build-native-release` (release profile), this branch vs `canary`. Stripped = `strip -x`; gzip = `gzip -9` of the stripped binary.

| Metric | branch | canary | Δ | Δ% |
|---|---:|---:|---:|---:|
| **raw** | 122,879,888 B | 124,184,976 B | **−1,305,088 B** | **−1.051%** |
| **stripped** | 83,505,264 B | 84,127,232 B | **−621,968 B** | **−0.739%** |
| **gzip+stripped** | 28,800,547 B | 28,925,679 B | **−125,132 B** | **−0.433%** |

The addon shrinks on every metric. Replacing the generated per-impl registration functions (and their `ctor`/`inventory` machinery) with link-time scattered collection removes real code/data — ~622 KB even after stripping — and laying the registries out contiguously at link time compresses better, so the gzip win holds up too.

### DIFF
--- a/turbopack/crates/turbo-tasks/src/macro_helpers.rs
+++ b/turbopack/crates/turbo-tasks/src/macro_helpers.rs
@@ -174,17 +174,6 @@ pub const fn strip_trailing_segments(s: &str, count: usize) -> &str {
 }
 
 /// A registry of all the impl vtables for a given VcValue trait.
-///
-/// `const`-constructed as a plain `static`. The map (`inner`) is built once during
-/// [`crate::value_type::register_all_trait_methods`], which runs inside the `VALUES` `LazyLock`
-/// initializer after `ValueType` ids are assigned. Each `impl Trait for Concrete` contributes via
-/// a [`TraitImplRecord`] gathered at link time (see [`TRAIT_IMPLS_SLICE`]); its
-/// `install_vtable` thunk calls [`Self::insert`].
-///
-/// The cast path is a single hashmap `.get()` keyed by `u16` — no `LazyLock` check, no
-/// `get_value_type(id)` indirection. The vtable pointer for each `impl Trait for Concrete` is
-/// materialized at compile time (a `const DynMetadata`), so there's no runtime `transmute` or
-/// indirect fn-pointer call on the cast path either.
 pub struct VTableRegistry<T>
 where
     T: Pointee<Metadata = DynMetadata<T>> + ?Sized,
--- a/turbopack/crates/turbo-tasks/src/macro_helpers.rs
+++ b/turbopack/crates/turbo-tasks/src/macro_helpers.rs
@@ -7,9 +7,9 @@ use std::{
 
 pub use async_trait::async_trait;
 pub use bincode;
-pub use ctor;
-pub use inventory;
 use rustc_hash::FxHashMap;
+pub use scattered_collect;
+use scattered_collect::slice::ScatteredSlice;
 pub use shrink_to_fit;
 pub use tracing;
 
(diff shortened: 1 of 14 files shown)


lukesandberg:
* **#94503** <a href="https://app.graphite.com/github/pr/vercel/next.js/94503?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94503?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94498** <a href="https://app.graphite.com/github/pr/vercel/next.js/94498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/macro_helpers.rs
```diff
@@ -289,34 +243,28 @@ where
     }
 }
 
-pub struct CollectableTraitMethods {
+/// One `impl Trait for ConcreteType` registration, gathered at link time into
+/// [`TRAIT_IMPLS_SLICE`]. Replaces the old `inventory`-collected `CollectableTraitMethods` plus the
+/// per-impl `#[ctor]` that seeded the vtable: both the method table and the Rust vtable now come
+/// from the same link-time record, so a partial view is impossible regardless of constructor order.
```
bgw:
```suggestion
/// [`TRAIT_IMPLS_SLICE`].
```

turbopack/crates/turbo-tasks/src/registry/mod.rs
```diff
@@ -14,12 +15,13 @@ pub use registry_type::RegistryType;
...
 /// Declare a type as a compile-time-collected registry item.
 ///
-/// Generates pointer-based `Eq`, `PartialEq`, `Hash`, `Ord`, `PartialOrd` impls
-/// and an `inventory::collect!` call for `&'static $ty`.
+/// Generates pointer-based `Eq`, `PartialEq`, `Hash`, `Ord`, `PartialOrd` impls.
+///
+/// The actual link-time gather slice for each registry is declared separately (see
+/// [`FUNCTIONS_SLICE`], [`VALUES_SLICE`], [`TRAITS_SLICE`]) because each needs a distinct
+/// module-level static and a wrapping [`LazyLock`].
 macro_rules! turbo_registry {
```
bgw: This macro name doesn't make a ton of sense anymore if it's not actually doing the collection.

bgw (approved):

turbopack/crates/turbo-tasks/src/registry/mod.rs
lukesandberg: concur

lukesandberg:
@mmastrac see the updated size metrics 👍

PR #94509, opened 2026-06-06
fix: preserve original sourcesContent in React Compiler sourcemaps

### LINKED ISSUE #94450: React Compiler + Turbopack: browser source map `sourcesContent` is the compiler output (wrong line numbers) for client→client imports

https://github.com/atkaiser/rc-sourcemap-repro

### To Reproduce

1. Clone the reproduction and install deps:
   ```bash
   git clone https://github.com/atkaiser/rc-sourcemap-repro
   cd rc-sourcemap-repro
   npm install
   ```
   The repo is two files: `app/page.tsx` (`'use client'`) imports `app/widget.tsx`
   (`'use client'`, a single `useState`). `next.config.mjs` sets
   `reactCompiler: true` and `productionBrowserSourceMaps: true`.

2. Build with the React Compiler enabled (the default):
   ```bash
   npm run build
   ```

3. Inspect the generated browser source map:
   ```bash
   npm run inspect
   ```
   This finds the `.next/static/chunks/*.js.map` whose `sources` includes
   `app/widget.tsx` and prints its `sourcesContent`. Observe **27 lines**
   containing `import { c as _c } from "react/compiler-runtime"` →
   `VERDICT: BUG`. (Manual equivalent: open the chunk `.js.map`, find the entry
   whose `sources` contains `app/widget.tsx`, read `sourcesContent`.)

4. Rebuild with the compiler disabled:
   ```bash
   npm run build:no-rc   # = reactCompiler:false via REACT_COMPILER=false next build
   ```

5. Inspect again:
   ```bash
   npm run inspect
   ```
   Now `sourcesContent` is **9 lines — identical to the real file** →
   `VERDICT: OK`.

**Current** (`reactCompiler: true`): the browser source map for
`app/widget.tsx` has its `sourcesContent` set to the React Compiler **output**
(27 lines, beginning `import { c as _c } from "react/compiler-runtime"` plus a
generated cache and hoisted helpers) instead of the 9-line source file. Resolved
stack-frame line numbers therefore point **past the end of the real file**,
breaking devtools "original source" view and the line-accurate "open in
repository" links that error trackers (Datadog / Sentry) generate from the map.

**Expected**: `sourcesContent` should be the original `app/widget.tsx`, with line
numbers mapping back to the real source — exactly what happens with
`reactCompiler: false`.

| `reactCompiler` | `app/widget.tsx` `sourcesContent` | line numbers |
| --- | --- | --- |
| `true`  | 27 lines (compiler output, `c as _c`) | wrong |
| `false` | 9 lines (= the real file) | correct |

This is **not** "expected React Compiler behavior" — it is **inconsistent**.
With `reactCompiler: true`, a standalone `'use client'` page, and a `'use client'`
component imported by a **server** component, both map correctly back to source.
Only a client component imported by **another client component** breaks. So the
pipeline is able to chain the compiler's map back to the original source; it just
fails on the client→client path.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:56 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6041
  Available memory (MB): 24576
  Available CPU cores: 14
Binaries:
  Node: 22.14.0
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 16.3.0-canary.40 // Latest available version is detected (16.3.0-canary.40).
  eslint-config-next: N/A
  react: 19.2.6
  react-dom: 19.2.6
  typescript: 5.7.3
Next.js Config:
  output: N/A

Tested and reproduces on **`next@16.3.0-canary.40`** (per the "test on canary" guidance) as well as `next@16.2.6` stable.
```

Turbopack, Output

next build (local)

### Additional context

_No response_

scarab-systems:
### What?

Fixes #94450.

Preserves original source content in production browser source maps when React Compiler transforms client components under Turbopack.

### Why?

React Compiler transforms can carry source maps whose `sourcesContent` still points back to the original client source, but final browser source-map composition can lose that source authority and expose compiler output instead. That makes generated browser source maps point at transformed compiler code rather than the original component source.

### How?

- Preserve the original loader input source in the Babel loader source map.
- Populate a missing source-map `file` from the resolved origin path so Turbopack composition can match the transform map back to the generated intermediate source.
- Add regression coverage for React Compiler production browser source maps.

### Tests

- `scripts/run-jest.sh --mode=start --bundler=turbo --headless -- test/production/production-browser-sourcemaps -t "React Compiler"`
- `cargo test -p turbopack-core --lib source_map::utils::tests::test_resolve_source_map_sources`
- `rustfmt --check turbopack/crates/turbopack-core/src/source_map/utils.rs`
- `git diff --check`

### DIFF
--- a/packages/next/src/build/babel/loader/transform.ts
+++ b/packages/next/src/build/babel/loader/transform.ts
@@ -72,6 +72,45 @@ function transformAst(file: any, babelConfig: any, parentSpan: Span) {
   }
 }
 
+function normalizeSourceMapPath(filename: string): string {
+  return filename.replace(/\\/g, '/')
+}
+
+function findLoaderInputSourceIndex(
+  sources: string[],
+  filename: string
+): number {
+  const normalizedFilename = normalizeSourceMapPath(filename)
+  const sourceIndex = sources.findIndex(
+    (sourceName) => normalizeSourceMapPath(sourceName) === normalizedFilename
+  )
+
+  if (sourceIndex !== -1) {
+    return sourceIndex
+  }
+
+  return sources.length === 1 ? 0 : -1
+}
+
+function preserveLoaderInputSourceContent(
+  map: GeneratorResult['map'],
+  source: string,
+  filename: string
+) {
+  const sources = map?.sources
+  if (!sources) {
+    return
+  }
+
+  const sourceIndex = findLoaderInputSourceIndex(sources, filename)
+  if (sourceIndex === -1) {
+    return
+  }
+
+  map.sourcesContent ??= []
+  map.sourcesContent[sourceIndex] = source
+}
+
 export default async function transform(
   ctx: NextJsLoaderContext,
   source: string,
@@ -108,6 +147,8 @@ export default async function transform(
   const generateSpan = parentSpan.traceChild('babel-turbo-generate')
   const { code, map } = generate(file.ast, file.opts.generatorOpts, file.code)
   generateSpan.stop()
+
+  preserveLoaderInputSourceContent(map, source, filename)
 
   return { code, map }
 }
--- a/test/production/production-browser-sourcemaps/app/client.js
+++ b/test/production/production-browser-sourcemaps/app/client.js
@@ -1,0 +1,14 @@
+'use client'
+
+import { useState } from 'react'
+
+export function ReactCompilerSourceMapClient() {
+  const [count, setCount] = useState(0)
+  const originalSourceMapMarker = 'original-source-map-marker'
+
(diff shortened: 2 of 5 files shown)


scarab-systems:
Gentle follow-up on this PR. The visible checks are not showing failures on my side.

Is there anything you’d like changed here, or should I close it if this is not a direction the project wants to take?

PR #94518, opened 2026-06-06
docs: fix onRequestError error type example

### LINKED ISSUE #94499: Instrumentation.onRequestError error parameter type does not match the docs

https://github.com/sculli-net/instrumentation-types-reproduction

### To Reproduce

1. Clone repo
2. Install packages `pnpm i`
3. Open [instrumentation.ts](/instrumentation.ts)

The code within `instrumentation.ts` is an exact copy of the sample within the [next.js docs](https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation#onrequesterror-optional). Despite the docs saying the following:
> Types: 
> ```typescript
> export function onRequestError(
>   error: { digest: string } & Error,
>   request: {
>     path: string // resource path, e.g. /blog?name=foo
>     method: string // request method. e.g. GET, POST, etc
>     headers: { [key: string]: string | string[] }
>   },
>   context: {
>     routerKind: 'Pages Router' | 'App Router' // the router type
>     routePath: string // the route file path, e.g. /app/blog/[dynamic]
>     routeType: 'render' | 'route' | 'action' | 'proxy' // the context in which the error occurred
>     renderSource:
>       | 'react-server-components'
>       | 'react-server-components-payload'
>       | 'server-rendering'
>     revalidateReason: 'on-demand' | 'stale' | undefined // undefined is a normal request without revalidation
>     renderType: 'dynamic' | 'dynamic-resume' // 'dynamic-resume' for PPR
>   }
> ): void | Promise<void>
> ```
> - **`error`: The caught error itself (type is always Error), and a digest property which is the unique ID of the error.**
> - `request`: Read-only request information associated with the error.
> - `context`: The context in which the error occurred. This can be the type of router (App or Pages Router), and/or (Server Components (`'render'`), Route Handlers (`'route'`), Server Actions (`'action'`), or Proxy (`'proxy'`)).

Type of `err` shows up as unknown, despite the docs describing it as `error: { digest: string } & Error`. This requires either a cast or a type check to enable this code to pass typescript build checks.
Image

The docs do specify the following:
> The error instance might not be the original error instance thrown, as it may be processed by React if encountered during Server Components rendering. If this happens, you can use digest property on an error to identify the actual error type.

Though if a type check is expected then the sample should show something like `message: err instanceof Error ? err.message : String(err)`, otherwise we should modify the Type with the [source code](https://github.com/vercel/next.js/blob/f4828c13e778dbe04c4a1e308ff29941536f091f/packages/next/src/server/instrumentation/types.ts#L14) to match the expected `{ digest: string } & Error` type.

```bash
Operating System:
  Platform: darwin
  Arch: x64
  Version: Darwin Kernel Version 23.6.0: Mon Jan 19 22:02:22 PST 2026; root:xnu-10063.141.1.710.3~1/RELEASE_X86_64
  Available memory (MB): 32768
  Available CPU cores: 16
Binaries:
  Node: 22.18.0
  npm: 10.9.3
  Yarn: 1.22.22
  pnpm: 11.5.0
Relevant Packages:
  next: 16.3.0-canary.42 // Latest available version is detected (16.3.0-canary.42).
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Instrumentation

next build (local)

### Additional context

I am using the project/workspace version of typescript.

danyalahmed1995:
### Summary

The `Instrumentation.onRequestError` type currently exposes the error parameter as `unknown`, but the docs described it as `Error & { digest: string }` and the example accessed `err.message` directly.

This updates the docs/example to match the current type and narrows the caught value before reading `message`.

### Tests:
- `git diff --check`

### DIFF
--- a/docs/01-app/03-api-reference/03-file-conventions/instrumentation.mdx
+++ b/docs/01-app/03-api-reference/03-file-conventions/instrumentation.mdx
@@ -48,10 +48,12 @@ export const onRequestError: Instrumentation.onRequestError = async (
   request,
   context
 ) => {
+  const message = err instanceof Error ? err.message : String(err)
+
   await fetch('https://.../report-error', {
     method: 'POST',
     body: JSON.stringify({
-      message: err.message,
+      message,
       request,
       context,
     }),
@@ -64,10 +66,12 @@ export const onRequestError: Instrumentation.onRequestError = async (
 
 ```js filename="instrumentation.js" switcher
 export async function onRequestError(err, request, context) {
+  const message = err instanceof Error ? err.message : String(err)
+
   await fetch('https://.../report-error', {
     method: 'POST',
     body: JSON.stringify({
-      message: err.message,
+      message,
       request,
       context,
     }),
@@ -84,7 +88,7 @@ The function accepts three parameters: `error`, `request`, and `context`.
 
 ```ts filename="Types"
 export function onRequestError(
-  error: { digest: string } & Error,
+  error: unknown,
   request: {
     path: string // resource path, e.g. /blog?name=foo
     method: string // request method. e.g. GET, POST, etc
@@ -104,7 +108,7 @@ export function onRequestError(
 ): void | Promise<void>
 ```
 
-- `error`: The caught error itself (type is always `Error`), and a `digest` property which is the unique ID of the error.
+- `error`: The caught value is typed as `unknown`. Narrow it before reading properties like `message` or `digest`.
 - `request`: Read-only request information associated with the error.
 - `context`: The context in which the error occurred. This can be the type of router (App or Pages Router), and/or (Server Components (`'render'`), Route Handlers (`'route'`), Server Actions (`'action'`), or Proxy (`'proxy'`)).
 
--- a/docs/01-app/03-api-reference/03-file-conventions/instrumentation.mdx
+++ b/docs/01-app/03-api-reference/03-file-conventions/instrumentation.mdx
@@ -49,11 +49,16 @@ export const onRequestError: Instrumentation.onRequestError = async (
   context
 ) => {
   const message = err instanceof Error ? err.message : String(err)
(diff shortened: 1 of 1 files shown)


docs/01-app/03-api-reference/03-file-conventions/instrumentation.mdx
```diff
@@ -64,10 +66,12 @@ export const onRequestError: Instrumentation.onRequestError = async (
 
 ```js filename="instrumentation.js" switcher
 export async function onRequestError(err, request, context) {
+  const message = err instanceof Error ? err.message : String(err)
```
icyJoseph, line 74: All in all this change is good. 

Thanks for taking the time.

One thing that's missing from this PR is that, we have now lost knowledge of the digest, which is not an standard property of Error.

```js
const message = err instanceof Error ? err.message : String(err)
const digest =
  typeof err === 'object' && err !== null && 'digest' in err
    ? String(err.digest)
    : undefined
```

Perhaps we need to show the whole thing? or add a small snippet after line 113 on how to extract this property. We talk about digest further up too.

danyalahmed1995: Oh yes right , I added `digest` narrowing to both examples so the docs still show how to preserve it while keeping the parameter typed as `unknown`.

icyJoseph (approved):

icyJoseph:
Nice, once again thanks. We require signed commits. Do you mind rewriting the commits on this branch so that they are signed?

danyalahmed1995:
Oh sure

danyalahmed1995:
Done, I rewrote the branch with signed commits.

icyJoseph (approved):

PR #94519, opened 2026-06-06
fix(instrumentation): update onRequestError error type to match docs

EduardF1:
## Summary
Fixes the type signature for `onRequestError` within `InstrumentationModule` to reflect the documentation and reality of the Next.js runtime, solving #94499.

Currently, the error parameter is typed as `unknown`, forcing developers to implement unnatural casting when using the framework's own boilerplate code. This updates it to `Error & { digest?: string }` to exactly mirror the documentation at https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation#onrequesterror-optional

## Testing
Verified against TS compiler.

### DIFF
--- a/packages/next/src/server/instrumentation/types.ts
+++ b/packages/next/src/server/instrumentation/types.ts
@@ -11,7 +11,7 @@ export type RequestErrorContext = {
 }
 
 export type InstrumentationOnRequestError = (
-  error: unknown,
+  error: Error & { digest?: string },
   errorRequest: Readonly<{
     path: string
     method: string


sculli-net:
I think @danyalahmed1995 has the correct approach here: https://github.com/vercel/next.js/pull/94518
The error parameter can be an `unknown` type after being forwarded through the `pages-handler.ts` - so updating the docs is preferred to updating the type itself.

danyalahmed1995:
@sculli-net Thanks and idk what the author of this PR is trying to do here , it felt like they read the whole issue and went completely opposite direction 🤷‍♂️

Also thank you for providing the detailed triage it really helped.

icyJoseph:
We went with https://github.com/vercel/next.js/pull/94518 -

PR #94520, opened 2026-06-07
docs: clarify custom barrel file optimization in optimizePackageImports

### LINKED ISSUE #52462: Docs: How to modularize imports from a barrel file that exports from multiple directories in Next.js
### What is the improvement or update you wish to see?

I would like to use the [`modularizeImports`](https://nextjs.org/docs/architecture/nextjs-compiler#modularize-imports) option of the Next.js compiler on a library that its barrel file exports from multiple directories within the library package.

What I mean is the `index.ts` file for the library looks something like this:

```
//index.ts
export {default as SomeComponent} from './components/SomeComponent';
export {default as someFun} from './utils/someFunction';
export {default as someHook} from './hooks/someHook';
```

ideally I would like to have something like this

```
//next.config.js
  modularizeImports: {
    '@some/library': {
      transform: '@some/library/dist/(components|hooks|utils)/{{member}}',
    },
  },
```

but 

1. I am not sure if this functionality is even possible
1. if it is I am not sure how to to write it.

I also tried passing it an array of transforms but it only accepts a single string.

Under the hood this option is implemented with https://docs.rs/handlebars/latest/handlebars/ afaik

-

_No response_

<sub>[DX-1758](https://linear.app/vercel/issue/DX-1758/docs-how-to-modularize-imports-from-a-barrel-file-that-exports-from)</sub>

EduardF1:
## Description
Fixes #52462

Added a targeted section to the `optimizePackageImports` documentation providing an explicit example of how developers can use the feature to modularize their own internal or multi-directory barrel files, rather than only relying on the built-in ecosystem packages.

## Testing
Docs only change.


EduardF1:
Follow-up: I pushed a small fix (23fad0c) — the custom barrel-file example was using single backticks instead of a triple-backtick fenced code block, so it would have rendered as broken inline code rather than a JS block. The fix:

- Use a proper ` ```js filename="next.config.js" ` fence (matching the existing example earlier in the page).
- Wrap `optimizePackageImports` and the `import { Button } from '@my-org/ui-library'` snippet in inline code.
- Drop a stray trailing space.

Formatting matches the repo Prettier config (`singleQuote`, `semi: false`). Ready for review — happy to adjust wording or squash the two commits if preferred.

EduardF1:
Friendly ping — this docs change is ready for review (the fenced-code-block fix is in, and the checks are green). Happy to adjust wording or squash the two commits if preferred.

PR #94522, opened 2026-06-07
fix(eslint-plugin-next): allow \
o-html-link-for-pages\ to respect \pageExtensions\ configurations

EduardF1:
## Description
Fixes #53473

Currently, the `@next/next/no-html-link-for-pages` rule is hardcoded with a Regex that only looks for `.js`, `.ts`, `.jsx`, and `.tsx` extensions. If a developer maps `pageExtensions` (e.g. to .page.tsx) in Next.js, the linter completely misses the directory mappings and fails to enforce HTML anchor tracking.

This patch dynamically extracts `pageExtensions` either from the eslint context options (via `context.settings?.next?.pageExtensions` or rule params) and passes it down through the cached URL parse pipeline, constructing a dynamic Extension RegExp rather than the hardcoded one.

## Testing
Verified logic bounds.


EduardF1:
Addressed the review feedback: `parseUrlForAppDir` was not updated alongside `parseUrlForPages` when `pageExtensions` was threaded through the URL helpers.

**What was broken**
- `getUrlFromAppDirectory` called `parseUrlForAppDir(urlPrefix, directory, pageExtensions)` (3 args) while the function still had a 2-arg signature → TS `Expected 2 arguments, but got 3`.
- At runtime, `parseUrlForAppDir` recursed into nested subdirectories via `parseUrlForPages(...)` without forwarding `pageExtensions`, so `pageExtensions.join('|')` threw `TypeError: Cannot read properties of undefined (reading 'join')` for any app dir with a nested subdirectory.

**Fix** (`packages/eslint-plugin-next/src/utils/url.ts`)
- `parseUrlForAppDir` now accepts `pageExtensions: string[]`, builds its `page`/`layout`/extension patterns from it (replacing the hard-coded `(j|t)sx?` regexes), and forwards `pageExtensions` in the recursive call.
- Also dropped the stray `dirent.isDirectory(dirPath)` argument to match `parseUrlForPages`.

**Tests** (`test/unit/eslint-plugin-next/no-html-link-for-pages.test.ts` + new `with-app-dir-custom-extensions` fixture)
- Nested app-dir route is reported without crashing (reproduces the previous `TypeError`).
- A `.mdx` app route is ignored when `pageExtensions` is not configured, and flagged when `settings.next.pageExtensions` includes `mdx`.

Verified locally by running the rule through ESLint's `Linter` against the fixtures: the unpatched code throws the exact `Cannot read properties of undefined (reading 'join')`, and the patched code passes all three cases. `prettier` + `eslint` are clean on the changed files.

EduardF1:
Follow-up: while type-checking the package (`tsc --noEmit -p packages/eslint-plugin-next/tsconfig.json`) I found a second, separate build blocker introduced by this change — `context.settings.next` is typed as `unknown`, so reading `context.settings.next.pageExtensions` directly failed with `TS2339: Property 'pageExtensions' does not exist on type 'unknown'`. Fixed by assigning `settings.next` to a typed local first, matching the existing pattern in `get-root-dirs.ts`. `tsc`, `eslint`, and `prettier` are now all clean on the changed files, and the rule's behavior is unchanged (verified via the unit tests).

EduardF1:
Follow-up cleanup: dropped the redundant second positional schema option (`pageExtensions` now comes solely from `settings.next.pageExtensions` or the default) and escaped each extension before joining it into the URL `RegExp`, so exotic `pageExtensions` containing regex metacharacters are matched literally.

Re-ran the Docker verification (node:20): `tsc` exit 0 and all 3 Linter scenarios pass — default extensions ignore `.mdx`; `pageExtensions` including `mdx` flags `/blog/`; nested subdir recursion does not throw. Prettier clean.

PR #94524, opened 2026-06-07
fix(dev): prevent static indicator from rendering on dynamic searchParams access

### LINKED ISSUE #72133: Static indicator shouldn't be shown when searchParams is being used

https://codesandbox.io/p/devbox/awesome-shockley-46z7c5

### To Reproduce

1. Start the dev server
2. Create a new page and use `searchParams`
```tsx
export default async function Page({
  searchParams,
}: {
  searchParams: Promise<{a: string}>;
}) {
  const sp = await searchParams;

  return (
    <div>
      <p>sp: {sp.a}</p>
    </div>
  );
}
```
3. Visit the page on the browser and the static indicator should not be shown here since `searchParams` is considered a Dynamic API

Build time:
In the build time process, though, the page is correctly rendered as a dynamic page (the `/sp` route):
image

The static indicator should not be shown here since `searchParams` is considered a Dynamic API.

```bash
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 4102
  Available CPU cores: 2
Binaries:
  Node: 20.9.0
  npm: 9.8.1
  Yarn: 1.22.19
  pnpm: 8.10.2
Relevant Packages:
  next: 15.0.3-canary.2 // Latest available version is detected (15.0.3-canary.2).
  eslint-config-next: N/A
  react: 19.0.0-rc-603e6108-20241029
  react-dom: 19.0.0-rc-603e6108-20241029
  typescript: 5.3.3
Next.js Config:
  output: N/A
```

Developer Experience

next dev (local)

### Additional context

_No response_

EduardF1:
## Description
Fixes #72133

In `next dev`, a Server Component that reads `searchParams` still shows the `Static` indicator in the dev overlay, even though the page is dynamically rendered based on query parameters (`next build` correctly evaluates it as dynamic).

The root cause is in the dev-time `searchParams` proxy: unlike `cookies()` and `headers()` — which call `trackDynamicDataInDynamicRender` and set `usedDynamic` — accessing `searchParams` never marks the render as dynamic. So the static-flag calculation (`!requestStore.usedDynamic && !workStore.forceDynamic`) incorrectly stays `true` in dev.

## Fix
In `search-params.ts`, call `trackDynamicDataInDynamicRender(requestStore)` from the dev-warning `searchParams` proxy when a value is actually accessed (`get` / `has` / `ownKeys`, gated on `promiseInitialized`). This marks the route dynamic only when `searchParams` is really read, so pages that never read it stay Static.

## Testing
Added cases to the existing `test/development/dev-indicator` suite using `nextTestSetup` + `waitForStaticIndicator`, covering searchParams access → Dynamic (with and without a query string), and (under Cache Components) that the indicator stays hidden.

## Validation (local)
- `pnpm --filter=next types` → clean
- `pnpm --filter=next build` → success
- `dev-indicator` suite → 15/15 pass


EduardF1:
Addressed the review finding in fdfd1158.

**Root cause**

`requestStore.searchParams` does not exist on `RequestStore`, so the `app-render.tsx` change was a TypeScript build failure and, at runtime, a no-op — the indicator was never actually corrected. Keying off the URL query string would also be wrong: a page that never reads `searchParams` would be marked Dynamic just because `?a=…` was present.

The real gap is that, in dev, accessing `searchParams` never set `requestStore.usedDynamic` — unlike `cookies()` / `headers()`, which call `trackDynamicDataInDynamicRender`. So `isStatic = !usedDynamic && !forceDynamic` stayed `true`.

**Fix**

- Revert `app-render.tsx` back to the original `isStatic` calculation.
- In `search-params.ts`, call `trackDynamicDataInDynamicRender(requestStore)` from the dev-warning `searchParams` proxy when a value is actually accessed (`get` / `has` / `ownKeys`, gated on `promiseInitialized`). This marks the route dynamic only when `searchParams` is really read, so pages that never read it stay Static.

**Tests**

The original e2e test used `createNextDescribe`, which no longer exists in this version (and `check` is deprecated). Replaced it with cases in the existing `test/development/dev-indicator` suite using `nextTestSetup` + `waitForStaticIndicator`, covering: searchParams access → Dynamic (with and without a query string) and, under Cache Components, the indicator stays hidden.

**Validation (local)**

- `pnpm --filter=next types` → clean, 0 errors (the previous `requestStore.searchParams` type error is gone)
- `pnpm --filter=next build` → success
- `dev-indicator` suite (dev / Turbopack): **15/15 pass**, including "static page stays Static" (confirms no over-triggering) and the new searchParams → Dynamic cases
- Cache Components mode: searchParams case passes (indicator hidden)

PR #94526, opened 2026-06-07
server component HMR: refresh affected segments (3/3)

feedthejim:
### Summary

Refresh the smallest affected App Router subtree after a Server Component edit instead of refetching the entire active route.

Today, editing a page-level Server Component can rerender the root layout, nested layouts, sibling parallel routes, and their data dependencies.

This PR makes the request scale with the part of the route that changed. Unaffected router cache nodes are reused.

The behavior is enabled by default, with an opt-out:

```js
experimental: {
  serverComponentsHmrSegmentRefresh: false,
}
```

### Design

The previous PR in this stack adds route filtering:

> Does this edit affect the route currently open in the browser?

This PR adds segment targeting:

> Within that route, what is the smallest subtree containing every affected Server Component owner?

The flow is:

```text
changed module
  -> owning page/layout/default boundaries
  -> targets in the active router tree
  -> minimal set of refetch markers
  -> targeted RSC response
```

Webpack and Turbopack trace a changed Server Component to the `page`, `layout`, or `default` files that own its output. The server sends those targets to the browser.

The browser resolves the targets against the active router tree, including dynamic segments, route groups, and parallel routes. Overlapping targets are collapsed when an ancestor already covers a descendant.

The existing App Router pipeline then places `refetch` markers only at the selected subtrees. Unaffected `CacheNode`s remain in place.

### Examples

| Edit | Refresh behavior |
| --- | --- |
| Page or component owned only by that page | Refresh the page subtree; reuse parent layouts and sibling slots |
| Nested layout | Refresh the layout and its descendants; reuse ancestors and sibling branches |
| `@sidebar/default.tsx` | Refresh only the sidebar parallel route |
| Shared module owned by a layout and its page | Refresh the layout once because it covers both owners |
| Root layout | Refresh the route root, equivalent to current behavior |

For example, given:

```text
app/layout.tsx
├── app/@sidebar/default.tsx
└── app/[team]/dashboard/layout.tsx
    └── app/[team]/dashboard/(content)/page.tsx
```

editing the dashboard page previously marked the route root for refetch. With this PR, only the dashboard page subtree is marked. The root layout, dashboard layout, and sidebar remain outside the request.

This matters most when unaffected layouts or sibling slots perform expensive uncached I/O.

### Fallbacks

Targeting is an optimization, not a correctness requirement. The active route is refreshed from the root when:

- ownership cannot be traced completely
- an owner or file convention cannot be represented as a target
- a target cannot be matched to the active router tree
- the dev server requires full invalidation

Disabling the flag also preserves the existing full-route behavior and avoids the additional ownership tracking.

### Results

Compiler and bundler time is excluded. Timing begins when the browser receives the HMR event.

#### v0 Projects route

Editing a leaf Server Component on the authenticated Projects route:

| Metric | Full route | Targeted | Change |
| --- | ---: | ---: | ---: |
| Router refetch | Route root | Projects page | Smaller subtree |
| Encoded response bytes | 96,474 | 16,598 | 82.8% lower |
| RSC request duration | 1,189 ms | 104 ms | 91.3% lower |

Both cases issued one RSC request. The improvement came from rendering and encoding less work. The sibling sidebar was not refetched.

#### vercel-site controlled route

Median of two edits on a Server Component route using `Suspense`, `connection()`, and filesystem I/O:

| Metric | Full route | Targeted | Change |
| --- | ---: | ---: | ---: |
| HMR event to visible update | 200.4 ms | 157.9 ms | 21.2% faster |
| RSC request wall time | 189.6 ms | 147.9 ms | 22.0% faster |
| RSC response bytes | 136,106 | 136,112 | No material change |

This route had little unaffected sibling payload, so response size stayed flat while latency improved. The real `/marketplace/aws` route was also smoke tested.

### Tradeoffs

- Development performs additional module-owner tracking.
- Shared modules can initially produce several targets before overlaps are collapsed.
- Targeted responses cannot validate the complete route shell; full-tree refreshes continue to perform that validation.
- Conservative fallback means ambiguous module graphs retain current behavior.

This does not change production rendering, initial loads, normal navigations, Client Component Fast Refresh, or cache semantics.

### Stack and verification

1. #94521: cancel superseded refreshes
2. #94525: skip inactive route refreshes
3. **This PR:** refresh affected route segments

Verified with `pnpm build-all`, native SWC build, focused Turbopack and Webpack development tests, Turbopack unit tests, v0, vercel-site, and the full CI matrix.


packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -753,14 +787,75 @@ export async function createHotReloaderTurbopack(
...
+      nextConfig.pageExtensions
+    )
+  }
+
+  function enqueueServerHmrRefresh() {
+    const targets = completedServerComponentTargets
+    const routes =
+      targets === undefined
+        ? undefined
+        : [...new Set(targets.map((target) => normalizeAppPath(target.path)))]
```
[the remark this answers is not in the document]
feedthejim: :robot: Fixed by deriving the route from the target owner path before normalization, so layout and default targets map to their owning URL. The targeted HMR suite covers both conventions.

packages/next/src/client/components/router-reducer/hmr-refresh.ts
```diff
@@ -0,0 +1,117 @@
...
+  FlightRouterState,
+  HmrRefreshTarget,
+} from '../../../shared/lib/app-router-types'
+import {
+  DEFAULT_SEGMENT_KEY,
+  PAGE_SEGMENT_KEY,
+} from '../../../shared/lib/segment'
+import { segmentToSourcePagePathname } from './compute-changed-path'
+
+function getTargetPath(
```
feedthejim, line 11: code smell

packages/next/src/server/dev/hmr-refresh-targets.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+export function getTurbopackChangedFiles(
+  moduleIds: ReadonlySet<string>,
+  projectRoot: string
+): Set<string> {
+  const changedFiles = new Set<string>()
+
+  for (const moduleId of moduleIds) {
+    const match = /^\[project\]\/(.+?) \[app-rsc\] \(ecmascript\)$/.exec(
```
feedthejim: code smell

turbopack/crates/turbopack-ecmascript-runtime/js/src/nodejs/dev/dev-nodejs.ts
```diff
@@ -33,6 +33,88 @@ function __turbopack_server_hmr_apply__(update: NodeJsHmrPayload): boolean {
...
 
+type HmrRefreshOwners = {
+  found: boolean
+  complete: boolean
+  owners: string[]
+}
+
+const appRscModuleIdRegex = /^\[project\]\/.+ \[app-rsc\] \(ecmascript\)$/
+const hmrRefreshOwnerModuleIdRegex =
+  /^\[project\]\/.+\/(?:page|layout|default)\.[^/ ]+ \[app-rsc\] \(ecmascript\)$/
```
feedthejim: surely there's a better way of doing this

packages/next/src/client/components/router-reducer/hmr-refresh.ts
feedthejim: Addressed. Refresh targets are now canonical owner-path strings, and the client derives the convention from the path so the representation cannot contain contradictory path/kind fields.

packages/next/src/server/dev/hmr-refresh-targets.ts
feedthejim: Addressed. Target classification now returns an explicit all-or-targets scope, and Turbopack returns project-relative owner paths directly instead of reparsing module IDs in the dev server.

turbopack/crates/turbopack-ecmascript-runtime/js/src/nodejs/dev/dev-nodejs.ts
feedthejim: Addressed. The runtime traversal now returns a discriminated unrelated/all/owners result, relies on cycle detection instead of an arbitrary visit cap, and emits canonical owner paths. The dev-server lifecycle is isolated in a tested coordinator.

PR #94528, opened 2026-06-07
fix(dynamic): do not force Suspense for ssr: false if loading is explicitly undefined

### LINKED ISSUE #73830: dynamic() ignores <Suspense> hierarchy resulting in layout flicker

https://codesandbox.io/p/devbox/determined-firefly-9dq24w

### To Reproduce

Open codesandbox on main page and observe the flicker

My app suffers from a rare (when on prod-optimized build) race condition, which results in layout flicker when navigating between pages.
It got more intense as I moved more pages out of SSR to improve performance, using `dynamic`.
I recalled an earlier issue in which the Router could make a full browser page refresh under some conditions resulting in a white-page flicker, but looks like that issue is long gone (with some NextJS update) and I wasn't able to race it that way.

With Chrome frame-by-frame recording, I finally raced it and seen that the flickered frame contains just the <Footer> element without the webpage content.

I wrapped my whole layout code with <Suspend, but the issue remained

So the layout code was
```
const Layout: FC<AppProps> = ({ Component, pageProps }) => {
  return (
    <Suspense>
      <div>
        <Header />
        <Component {...pageProps} />
        <Footer />
      </div>
    </Suspense>
  );
};
export default Layout;
```

Why?
NextJS uses <Suspend in dynamic(), but it will always force fallback to be either a provided element, or it's own element which returns empty content when Suspended.
That means React can't see the parent <Suspend which only works when the child has fallback=null, which would avoid rendering incomplete component with just <Footer> inside.

I don't see rendering "empty" as a default fallback making sense, IMO it would make sense to use no fallback= at all and refer the error handling, which is part of current "fallback" function, to the ErrorBoundary, and otherwise just properly Suspense.

Alternatively, I tried forcing `loading` to `undefined` which should override it properly, but there's another part of code that expects it to be valid and results in crash:

https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/loadable.shared-runtime.tsx#L142

```bash
"next": "15.1.1-canary.1",
    "react": "^19.0.0",
    "react-dom": "^19.0.0"
```

Not sure

### Additional context

_No response_

EduardF1:
## Description
Fixes #73830

Currently, in App Router, using `next/dynamic` with `ssr: false` unconditionally forces an inner `<Suspense>` boundary (with `fallback={null}` if no loading component is provided) during both SSR and hydration. This prevents users from gracefully propagating the client-side rendering bailout out to their layout-level `<Suspense>` boundaries.

This surgical fix modifies `loadable.tsx` to detect if `loading` was explicitly passed in the options object by the user (`'loading' in options`). When `loading: undefined` is passed, the wrapper will no longer force an inner Suspense boundary, allowing the `BailoutToCSRError` to successfully bubble up to the parent `<Suspense>` boundary.

Additionally, it handles `loading: undefined` in Pages router's `noSSR` implementation to avoid server crashes.

## Testing
Tested manually, and added an integration test for `next/dynamic` in `test/e2e/app-dir/dynamic-suspense-bailout` ensuring parent Suspense boundaries trigger correctly when `ssr: false` and `loading: undefined` are combined.


EduardF1:
Follow-up pushed for the Vercel review note: the invalid/self-referential test line now uses the standard test helper form const $ = await next.render/. I did not mark this draft ready because the targeted dynamic-suspense-bailout e2e should still be run in the full Next.js test environment.

PR #94538, opened 2026-06-07
switch our test benchmark runs to release-with-assertions

lukesandberg:
In ci when testing our benchmarks use our test profile `release-with-assertions`

This avoids an expensive `lto` step and ensures that our tests run with debug asserts.

I noticed that our workflow for 'test cargo benches' was very slow and a lot of that time was the build presumably due to lto overheads across so many benchmark binaries.

Compare:

* cargo benches for this pr: https://github.com/vercel/next.js/runs/79974366863?pr=94538
   - build=4m42s
   - test=1m46s
* cargo benches before this pr: https://github.com/vercel/next.js/actions/runs/27097792020/job/79973082715#logs
   - build=8m23s
   - test=6m56s

Of course doing this revealed a few things

* a birthday paradox panic in one of the persistence tests
* some 'top level read' errors to address in the benchmarks and in the turbopack-cli server used for some benchmarks.

And finally, that the persistence tests are just extremely slow in test mode due to their initialization overhead (populating >1GB dbs), so that harness is rewritten to remove some variants and also to reduce max sizes in test mode

### DIFF
--- a/turbopack/crates/turbo-persistence/benches/mod.rs
+++ b/turbopack/crates/turbo-persistence/benches/mod.rs
@@ -24,10 +24,30 @@ static ALLOC: TurboMalloc = TurboMalloc;
 // =============================================================================
 
 const MB: u64 = 1024 * 1024;
-/// Data amount for batch read benchmarks (1 GiB)
-const BATCH_READ_DATA_AMOUNT: usize = 1024 * MB as usize;
+/// Data amount for batch read benchmarks (128 MiB).
+const BATCH_READ_DATA_AMOUNT: usize = 128 * MB as usize;
 /// Maximum memory to use for storing keys during prefill (4 GiB)
 const MAX_KEY_MEMORY: usize = 4 * 1024 * MB as usize;
+
+/// When running under `cargo test --benches` / `cargo nextest run --benches`, criterion
+/// executes each benchmark for a single iteration to verify it doesn't panic.
+const TEST_MODE_MAX_ENTRIES: usize = 100 * 1024;
+
+/// True when the benchmark harness is running in test mode, i.e. via `cargo test --benches` or
+/// `cargo nextest run --benches`, rather than a real `cargo bench`.
+static IS_TEST_MODE: LazyLock<bool> =
+    LazyLock::new(|| !std::env::args().any(|arg| arg == "--bench"));
+
+/// Cap an entry count to [`TEST_MODE_MAX_ENTRIES`] when running in test mode; otherwise return
+/// it unchanged. Used only to bound prefill/setup work — benchmark `id` strings keep using the
+/// configured (unscaled) sizes so real-run reports are unaffected.
+fn scaled(full: usize) -> usize {
+    if *IS_TEST_MODE {
+        full.min(TEST_MODE_MAX_ENTRIES)
+    } else {
+        full
+    }
+}
 
 // =============================================================================
 // Helper Types and Functions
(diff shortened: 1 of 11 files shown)


lukesandberg:
* **#94538** <a href="https://app.graphite.com/github/pr/vercel/next.js/94538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94538?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-persistence/benches/mod.rs
```diff
@@ -24,11 +24,31 @@ static ALLOC: TurboMalloc = TurboMalloc;
...
 const MAX_KEY_MEMORY: usize = 4 * 1024 * MB as usize;
 
+/// When running under `cargo test --benches` / `cargo nextest run --benches`, criterion
+/// executes each benchmark for a single iteration to verify it doesn't panic.
+const TEST_MODE_MAX_ENTRIES: usize = 100 * 1024;
+
+/// True when the benchmark harness is running in test mode, i.e. via `cargo test --benches` or
+/// `cargo nextest run --benches`, rather than a real `cargo bench`.
+static IS_TEST_MODE: LazyLock<bool> =
+    LazyLock::new(|| !std::env::args().any(|arg| arg == "--bench"));
```
bgw: Does this actually work?

The `--benches` (named differently) arg is passed to `cargo` or `nextest`, not to the actual child test process.

lukesandberg: :robot: Yes, this works. The check is for \`--bench\` (singular), not the cargo-level \`--benches\` flag.

When you run \`cargo bench\`, cargo invokes the benchmark binary with \`--bench\` in its argv. When you run \`cargo test --benches\`, cargo invokes the *same* binary but with \`--test\` instead — it does not pass \`--bench\` through. So checking the child process's own \`std::env::args()\` for \`--bench\` reliably distinguishes the two.

This is exactly the signal criterion itself uses. From `codspeed-criterion-compat-walltime` (`Criterion::configure_from_args`):

```rust
let bench = matches.get_flag("bench");
let test = matches.get_flag("test");
let test_mode = match (bench, test) {
    (true, true) => true,   // cargo bench -- --test should run tests
    (true, false) => false, // cargo bench should run benchmarks
    (false, _) => true,     // cargo test --benches should run tests
};
```

Verified empirically by building the turbo-persistence bench binary with `--profile=release-with-assertions --no-run` and running it directly:
- `mod-<hash> --test` (what `cargo test --benches` passes) → enters test mode, full suite runs in ~20s with scaled-down entry counts.
- `mod-<hash> --bench ...` → benchmark mode, no scaling.

So `IS_TEST_MODE` is `false` only under a real `cargo bench`, which is the intended behavior.

turbopack/crates/turbo-persistence/benches/mod.rs
```diff
@@ -24,11 +24,31 @@ static ALLOC: TurboMalloc = TurboMalloc;
...
+
+/// Cap an entry count to [`TEST_MODE_MAX_ENTRIES`] when running in test mode; otherwise return
+/// it unchanged. Used only to bound prefill/setup work — benchmark `id` strings keep using the
+/// configured (unscaled) sizes so real-run reports are unaffected.
+fn scaled(full: usize) -> usize {
```
lukesandberg, line 44: instead add a LARGE_DB` env var to unlock the larger tests

turbopack/crates/turbopack-trace-server/src/reader/mod.rs
```diff
@@ -23,6 +23,10 @@ use crate::{
 
 const MIN_INITIAL_REPORT_SIZE: u64 = 100 * 1024 * 1024;
 
+/// Width to pad the in-place progress line to, so shorter updates fully
+/// overwrite the leftover characters of a previous, longer update.
+const LINE_CLEAR_WIDTH: usize = 100;
+
```
bgw: Should be a different PR?

bgw (approved):

turbopack/crates/turbo-persistence/benches/mod.rs
lukesandberg: :robot: Done — replaced the implicit `--bench` argv detection with an explicit `LARGE_DB` env var. The benches now scale down by default for both `cargo bench` and `cargo test --benches`; setting `LARGE_DB` unlocks the full (large) sizes. (`std::env::var_os("LARGE_DB").is_some()`, so any value enables it.) Committed locally, pushing shortly.

PR #94547, opened 2026-06-08
Revert internal ast-grep lint rule change

mischnic:
@lukesandberg changed this in https://github.com/vercel/next.js/pull/94324

There were many false lint warnings after that, e.g. on this non-turbotask-function:
```rust
async fn create_app_paths_manifest(
    node_root: FileSystemPath,
    original_name: &str,
    filename: RcStr,
) -> Result<ResolvedVc<Box<dyn OutputAsset>>> {  // <----
```

or

```rust
async fn merge_modules(
    mut contents: Vec<(ResolvedVc<Box<dyn EcmascriptChunkPlaceable>>, CodeGenResult)>,
    entry_points: &Vec<(ResolvedVc<Box<dyn EcmascriptChunkPlaceable>>, usize)>,
    globals_merged: &'_ Globals,
) -> Result<(
    Program,
    Vec<CodeGenResultComments>,
    Vec<CodeGenResultSourceMap>,
    SmallVec<[ResolvedVc<Box<dyn GenerateSourceMap>>; 1]>, // <----
    Arc<Mutex<Vec<ModulePosition>>>,
)> {
```

### DIFF
--- a/.config/ast-grep/rules/resolved-vc-in-return-type.yml
+++ b/.config/ast-grep/rules/resolved-vc-in-return-type.yml
@@ -1,24 +1,52 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/ast-grep/ast-grep/main/schemas/rule.json
 
 id: resolved-vc-in-return-type
-message: Returning a `ResolvedVc` in a turbo task is not recommended. Consider replacing `ResolvedVc<$A>` in `$FN_NAME` with `Vc<$A>` instead.
+message: Returning a `ResolvedVc` in a turbo task is not recommended. Consider replacing `$TYPE` in `$FN_NAME` with `$RESOLVED_VC` instead.
 severity: warning # error, warning, info, hint
 language: Rust
 rule:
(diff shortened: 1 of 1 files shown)


.config/ast-grep/rules/resolved-vc-in-return-type.yml
```diff
@@ -1,24 +1,52 @@
...
+        - pattern:
+            context: 'let x: ResolvedVc<$A> = 1;'
+            selector: generic_type
+        - pattern: $TYPE
+      inside:
+        kind: function_item
+        field: return_type
+        stopBy: end
+        follows:
+          pattern: '#[turbo_tasks::function]'
```
lukesandberg, line 20: i think this condition needs to get copied down to the `function_signature_item` rule, since it should only apply to value_trait items with a function annotation

lukesandberg: i would suspect without that it will complain about the ResolveOrigin trait members

PR #94556, opened 2026-06-08
fix(eslint-plugin-next): Support pageExtensions in no-html-link-for-pages and resolve app directory static route bug

Myself-Praveen:
### What?
Updates the `@next/next/no-html-link-for-pages` ESLint rule to dynamically support custom `pageExtensions`. Additionally, this fixes two URL resolution bugs in the App Router linting logic.

### Why?
The rule previously hardcoded regex checks for standard extensions and ignored custom `pageExtensions` (e.g., `.page.tsx`), resulting in false negatives for invalid internal HTML links. 

Furthermore, while investigating, I found two bugs in the `app` directory linting logic:
1. `parseUrlForAppDir` was erroneously calling `parseUrlForPages` recursively on its subdirectories.
2. `normalizeAppPath` omitted trailing slashes for static app routes (e.g. `/about`). Because `hrefPath` normalizes with a trailing slash (e.g. `/about/`), the regex mismatch meant that an invalid `<a href="/about">` inside the `app` directory would silently bypass linting checks without warning the user.

### How?
- Extracts `pageExtensions` from `context.settings.next.pageExtensions` within the ESLint config (defaulting to standard extensions if not provided).
- Refactors `parseUrlForPages` and `parseUrlForAppDir` to dynamically strip these configured extensions instead of relying on a hardcoded `/(\.(j|t)sx?)$/` regex.
- Fixes the recursive bug in `parseUrlForAppDir` to properly call itself instead of `parseUrlForPages`.
- Wraps `normalizeAppPath` with `normalizeURL` in `getUrlFromAppDirectory` to ensure static app directory routes consistently match the trailing slash format of normalized `href` paths.

Fixes #53473

### DIFF
--- a/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
+++ b/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
@@ -83,7 +83,9 @@ export default defineRule({
     if (typeof pageExtensions === 'string') {
       pageExtensions = [pageExtensions]
     }
-    pageExtensions = pageExtensions.map((ext) => ext.trim().replace(/^\.+/, ''))
+    pageExtensions = pageExtensions.map((ext) =>
+      ext.trim().replace(/^\.+/, '').toLowerCase()
+    )
 
     const rootDirs = getRootDirs(context)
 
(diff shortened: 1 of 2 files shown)


Myself-Praveen:
@copilot all review comments have been fully addressed and resolved in the latest commits!

PR #94559, opened 2026-06-08
Revert "Remove dead cacheComponents web streams path"

timneutkens:
Reverts vercel/next.js#93944 so that we can merge [94457](https://github.com/vercel/next.js/pull/94457) first.


unstubbable:
The merge of `canary` into #94457 did work out, so this is not needed after all.

PR #94561, opened 2026-06-08
[turbopack] Drop data while serializing in `next build`

lukesandberg:
When turbopack persists it needs to allocate scratch buffers and gather and sort large amounts of data before it can start writing to disk.  Fundamentally this is because in memory the data is organized by FxHash and on disk it is by a different hash function and sorted. We get more efficient output files if we buffer more entries in ram before flushing thus persistence tends to consume a lot of additional ram.

To mitigate that cost we need to proactively drop data as we go treating persistence as a terminal operation in `next build`.

Currently this doesn't do anything for `next dev` however a future PR will attempt to integrate 'eviction' logic into these same locations to mitigate the same concern.

With this PR

Without this PR

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
@@ -9,6 +9,8 @@ use std::{
     },
 };
 
+use dashmap::SharedValue;
+use hashbrown::raw::RawIntoIter;
 use thread_local::ThreadLocal;
 use tracing::span::Id;
 use turbo_bincode::TurboBincodeBuffer;
@@ -271,8 +273,10 @@ impl Storage {
     /// entries) are filtered out, but shards may still yield no items if all entries produce
     /// empty `SnapshotItem`s (this is rare and only happens under error conditions).
     ///
-    /// When `drain_entries` is true, each task entry is removed from the map and dropped as soon as
-    /// it has been serialized, draining the core datastructures.
+    /// When `drain_entries` is true (shutdown only), the scan drains the map: unmodified entries
+    /// are erased and freed immediately, and the modified entries are moved out into the
+    /// returned shard iterators, which free each task's memory as it is serialized rather than
+    /// after the whole batch is written.
     pub fn take_snapshot<
         'l,
         P: for<'a> Fn(TaskId, &'a TaskStorage, &mut TurboBincodeBuffer) -> SnapshotItem + Sync,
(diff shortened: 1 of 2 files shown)


lukesandberg:
* **#94842** <a href="https://app.graphite.com/github/pr/vercel/next.js/94842?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94841** <a href="https://app.graphite.com/github/pr/vercel/next.js/94841?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94752** <a href="https://app.graphite.com/github/pr/vercel/next.js/94752?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94717** <a href="https://app.graphite.com/github/pr/vercel/next.js/94717?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94561** <a href="https://app.graphite.com/github/pr/vercel/next.js/94561?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94561?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94570** <a href="https://app.graphite.com/github/pr/vercel/next.js/94570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -289,47 +293,59 @@ impl Storage {
...
+                        // them. new_task always comes with modified flags (set_persistent_task_type
+                        // calls track_modification), so any_modified() is sufficient.
+                        (*key, shared_value.get().flags.any_modified())
+                    };
+                    if modified_task {
+                        debug_assert!(
+                            !key.is_transient(),
+                            "found a modified transient task: {key:?}"
+                        );
+                        modified.push(key);
```
sokra: In drain_entries mode we technically don't need to keep track of the modified tasks. Since we erase all non-modified tasks we could just iterate over the remaining entries in the map. That would avoid the Vec and a lot of map lookups.

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -270,13 +270,17 @@ impl Storage {
...
+    /// it has been serialized, draining the core datastructures.
     pub fn take_snapshot<
         'l,
         P: for<'a> Fn(TaskId, &'a TaskStorage, &mut TurboBincodeBuffer) -> SnapshotItem + Sync,
     >(
         &'l self,
         guard: SnapshotGuard<'l>,
         process: &'l P,
+        drain_entries: bool,
     ) -> Vec<SnapshotShard<'l, P>> {
```
sokra, line 288: I think it might make sense to split the logic into two separate methods `take_snapshot` and `into_snapshot`. This would allow to move the shards into the returned type and iterator.

lukesandberg: i moved the shards into the iterator in 'drain' mode, but kept a single function and single iterator type

the tiny bit of 'match' overhead is no big deal but splitting would add duplication, plus in a followup i want to add an `evict` path here as well which is kind of like `drain`

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
lukesandberg: Done.  drain now `takes` and drains the shard directly.

sokra (approved):

PR #94573, opened 2026-06-08
Turbopack: add experimental React compiler support

wbinnssmith:
- Adds experimental.rustReactCompiler config option to the next.js config to opt in. It should be used alongside `config.reactCompiler` and errors if not set.
- Only runs the compiler on client code (and SSR), not RSC
- When enabled, runs the compiler directly on Turbopack's swc AST with no gen+reparse. This required passing through the original source text into source transforms.


packages/next/src/server/config-shared.ts
```diff
@@ -1164,6 +1164,12 @@ export interface ExperimentalConfig {
    */
   globalNotFound?: boolean
 
+  /**
+   * @experimental Use the Rust port of the React compiler (Turbopack only).
+   * Requires `reactCompiler` to be enabled.
+   */
+  rustReactCompiler?: boolean
```
bgw: Maybe this should be `turbopackRustReactCompiler`?

crates/next-core/src/next_config.rs
```diff
@@ -2128,6 +2133,29 @@ impl NextConfig {
...
     }
 
+    /// Returns compilation mode when both `reactCompiler` and `rustReactCompiler` are set; `None`
+    /// otherwise.
+    #[turbo_tasks::function]
+    pub fn rust_react_compiler(&self) -> Vc<OptionRustReactCompilerCompilationMode> {
+        let use_rust = self.experimental.rust_react_compiler.unwrap_or(false);
+        let mode = match (use_rust, &self.react_compiler) {
+            (true, Some(ReactCompilerOptionsOrBoolean::Boolean(true))) => {
+                Some(RustReactCompilerCompilationMode::default())
```
bgw:
```suggestion
                Some(RustReactCompilerCompilationMode::default())  // infer
```

Maybe just remove the `Default` implementation for `RustReactCompilerCompilationMode` and centralize all the logic here?

crates/next-core/src/next_config.rs
```diff
@@ -2128,6 +2133,29 @@ impl NextConfig {
...
+            }
+            (true, Some(ReactCompilerOptionsOrBoolean::Option(opts))) => {
+                Some(match opts.compilation_mode {
+                    ReactCompilerCompilationMode::Infer => RustReactCompilerCompilationMode::Infer,
+                    ReactCompilerCompilationMode::Annotation => {
+                        RustReactCompilerCompilationMode::Annotation
+                    }
+                    ReactCompilerCompilationMode::All => RustReactCompilerCompilationMode::All,
```
bgw: Instead of doing this conversion, we could just rename `RustReactCompilerCompilationMode` to `ReactCompilerCompilationMode` and make it implement the required serde traits? Unless we think that the rust implementation is going to expose different options than the JS version in the future?

turbopack/crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -352,6 +387,97 @@ impl EcmascriptInputTransform {
...
+        "enableReanimated": false,
+        "isDev": ctx.node_env != "production",
+        "compilationMode": compilation_mode.as_str(),
+        "panicThreshold": "none",
+        "flowSuppressions": false,
+        "noEmit": false,
+        "ignoreUseNoForget": false,
+        "filename": ctx.file_name_str,
+        "environment": {}
+    }))
```
bgw: Is this really the best way to construct this object? Via serde?

crates/next-core/src/next_config.rs
wbinnssmith: Yeah I like this.

lukesandberg (approved):

PR #94581, opened 2026-06-09
Call `register_assignment_scope()` in  `visit_export_default_expr()` and `visit_export_default_decl()`

sampoder:
Talked IRL, we found that calling this in `visit_export_default_expr()` cut down the `module-cost` benchmark on ESM execution by about 10ms (approx. 35%).

Doing this unlocks a codegen optimization to avoid arrow functions, this special case was needed to support `export default ...` which is a kind of declaration all on its own.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
@@ -1076,6 +1076,7 @@ impl Visit for Analyzer<'_> {
             }
         };
 
+        self.register_assignment_scope(id.clone());
         self.data.exports.insert(
             rcstr!("default"),
             Export::LocalBinding(RcStr::from(id.0.as_str()), false),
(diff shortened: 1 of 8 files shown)


mischnic:
> Doing this unlocks a codegen optimization to avoid arrow functions, this special case was needed to support export default ... which is a kind of declaration all on its own.

Can we add a test case for that? Or does this just unblock some future change?

sampoder:
> Can we add a test case for that? 

Added a snapshot test that demonstrates it!

turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -1077,6 +1076,7 @@ impl Visit for Analyzer<'_> {
             }
         };
 
+        self.register_assignment_scope(id.clone());
```
lukesandberg, line 1089: we need to add some defense in depth

i think this would go in the get_export_ident_liveness function

in the else block for when there is no key in the map there are two theories for why we don't have a map entry, we should just `debug_assert` that that is the case (the var is free or is an import)

lukesandberg (approved):
this looks good and is a nice win

but we should add some defense in depth to prevent regressions

PR #94595, opened 2026-06-09
[PP] Instant validation - error for unguarded static params

lubieowoce:
Follow-up to #95151, implementing support for validating awaits of static params (i.e. when using `generateStaticParams`)

An app shell cannot contain any link data. This poses a challenge when we're trying to re-use the dev render for instant validation, because we have to make a choice:
1. either we resolve static params (from `generateStaticParams`) in the `Static` stage, and get an accurate *static HTML shell* (used for the initial navigation), but an incorrect app shell
2. or we resolve them in the `ShellRuntime` stage and we get an accurate App Shell, but cannot validate a static HTML shell (which would contain static params)

We'll use order 2 in the main render whenever client-navigating to a page with `partialPrefetching` enabled. We still need a render with order 1 for Static Shell Validation, so we perform a second, partial render that aborts before the dynamic stage (because only instant navigation needs the dynamic stage). As an optimization, we can skip the secondary render if the page doesn't have static params, because in that case the two orders are equivalent.

If we're performing an initial load, we'll use order 1 (to reflect the HTML shell). in this case, we'll do a full secondary render with order 2. Same as above, we can skip the secondary render if there's no static params.

---

Where params resolve (order 1 vs order 2) is controlled by `requestStore.needsSessionShell`. We then end up with two sets of "validation inputs" (mainly rendered chunks) that we can feed into static and instant validation respectively.

Also note that this PR doesn't touch `environmentName`, because that required changing too many tests. This will be addressed in a follow-up.


lubieowoce:
* **#95090** <a href="https://app.graphite.com/github/pr/vercel/next.js/95090?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95276** <a href="https://app.graphite.com/github/pr/vercel/next.js/95276?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94595** <a href="https://app.graphite.com/github/pr/vercel/next.js/94595?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94595?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#95106** <a href="https://app.graphite.com/github/pr/vercel/next.js/95106?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95151** <a href="https://app.graphite.com/github/pr/vercel/next.js/95151?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95279** <a href="https://app.graphite.com/github/pr/vercel/next.js/95279?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/lib/params-utils.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+  return false
+}
+
+function isStaticParam(
+  paramName: string,
+  fallbackParams: OpaqueFallbackRouteParams | null | undefined
+) {
+  // NOTE: Assume that undefined fallback params mean that none of the params are static
```
unstubbable:
```suggestion
  // NOTE: Assume that undefined fallback params mean that all of the params are static.
```

Right?

packages/next/src/server/request/params.ts
```diff
@@ -590,9 +595,13 @@ function createStagedRenderParamsImpl(
     !allParamsAreRootParams(underlyingParams, workUnitStore.rootParams)
   ) {
     // For a dynamic request we generally want to recover a static shell,
-    // so static params can resolve in the static stage.
-    // Session shells are handled with a separate render.
-    const staticParamsStages = RENDER_STAGES_BY_DATA_KIND.staticLinkData
+    // so static params can resolve in the static stage, because session
+    // shells are handled with a separate render.
+    // However, in dev where we might need to recover a session shell for instant svalidation.
```
unstubbable:
```suggestion
    // However, in dev we might need to recover a session shell for instant validation.
```

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4487,68 +4414,245 @@ function runDevValidationInBackground(
...
+    hadCacheMiss,
+    syncInterruptReason,
+    startTime,
+    staticStageEndTime,
+    runtimeStageEndTime,
+    accumulatedChunks,
+  } = result
+
+  if (!hadCacheMiss && syncInterruptReason === null) {
+    const inputsFromNavigation = {
```
unstubbable:
```suggestion
    const inputsFromNavigation: DevValidationInputs = {
```

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4645,6 +4775,15 @@ type StreamRevealStage =
   | RenderStage.ShellRuntime
   | RenderStage.Runtime
 
+function navigationHasAppShell(navigationKind: DevNavigationKind): boolean {
+  // TODO(app-shells): when we implement `<Link prefetch={true}>/`prefetch = "unstable_eager"` in dev,
+  // this might need to be adjusted, becase we'll use `Runtime` for the stage
```
unstubbable:
```suggestion
  // this might need to be adjusted, because we'll use `Runtime` for the stage
```

test/e2e/app-dir/instant-validation/app/suspense-in-root/non-app-shell/valid-unguarded-static-params/[slug]/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+  params,
+}: {
+  params: Promise<{ slug: string }>
+}) {
+  return (
+    <main>
+      <p>
+        This page is missing a suspense around a link data access, but this page
+        is intended to run without partialPrefetching is not enabled, so this is
+        still allowed.
```
Comment (unstubbable) [inline test/e2e/app-dir/instant-validation/app/suspense-in-root/non-app-shell/valid-unguarded-static-params/[slug]/page.tsx]:
```suggestion
        is intended to run without partialPrefetching, so this is still allowed.
```

unstubbable (approved):
Just a couple of nits, looks good otherwise.

packages/next/src/server/lib/params-utils.ts
lubieowoce: yeah i meant to type "none are fallback" and got confused lol

PR #94596, opened 2026-06-09
[cd] Ensure release can be triggered on old branches

eps1lon:
Creating a ref requires `contents: write` or `contents: write` and `workflows: write`. The API didn't make it clear when we need `workflows: write`. Turns out `workflows: write` is needed when running on commits that are far removed from the default branch according to GitHub support. Adding `workflows: write` fixed permission issues when creating refs on old, diverged commits so this seems to have been the issue indeed.

IT already updated the application to have `workflows: write`. Now we need to update the requested token.

Needs to be backported to be effective.

## test plan

- [x] ref creation works on old (`next-15-5`) branches: https://github.com/vercel/next.js/actions/runs/27212237311

### DIFF
--- a/.github/workflows/trigger_release.yml
+++ b/.github/workflows/trigger_release.yml
@@ -62,6 +62,10 @@ jobs:
           owner: ${{ github.repository_owner }}
           repositories: next.js
           permission-contents: write
+          # Even though this permission may seem optional, it's required for
+          # creating refs on branches that have diverged a lot from the default
+          # branch. This is not documented but was confirmed by GitHub support.
+          permission-workflows: write
 
       - name: Get GitHub App user ID
         id: release-app-user


.github/workflows/trigger_release.yml
```diff
@@ -62,6 +62,10 @@ jobs:
           owner: ${{ github.repository_owner }}
           repositories: next.js
           permission-contents: write
+          # Even though this permission may seem optional, it's required for
+          # creating refs on branches that have diverged a lot from the default
+          # branch. This is not documented but was confirmed by GitHub support.
```
bgw, line 67: This is technically documented, but the documentation is rather confusing in how it's explained. https://docs.github.com/en/rest/git/refs?apiVersion=2026-03-10#create-a-reference

> The fine-grained token must have at least one of the following permission sets:
>
> - "Contents" repository permissions (write)
> - "Contents" repository permissions (write) and "Workflows" repository permissions (write)

What this doesn't tell you though is *when* the workflows write permission is needed, as it's not needed if you don't have modified workflows?

bgw (approved):

PR #94600, opened 2026-06-09
examples/with-ably: update to App Router + Ably v2

owenpearson:
## Summary

Modernizes the `examples/with-ably` example off the deprecated Pages Router and `@ably-labs/react-hooks` package, onto the App Router and the React hooks that ship with ably-js v2.

The previous example created the Realtime client at module scope inside `pages/_app.tsx`, which caused connections to be created during SSR. The rewrite creates it inside a `useEffect` in a client component (`app/ably-client-provider.tsx`), gated by an `AblyReadyContext` so consumer components don't try to call `ably/react` hooks before the provider is in place.

## Notable changes

- Replace Pages Router with App Router.
- Upgrade `ably` to v2; drop `@ably-labs/react-hooks` (hooks now ship as `ably/react`).
- Bump React, react-dom and their `@types` to v19 to match recently-modernized examples like `with-supabase`.
- `app/api/createTokenRequest/route.ts` now returns 400 when `clientId` is missing rather than coalescing to a shared `"NO_CLIENT_ID"` value.
- Various improvements to README.md

## How I tested these changes

Ran the example with Chrome and Firefox. Verified in two browser windows:

- Pub/sub: messages published from one window appear in both.
- Server publish: `POST /api/send-message` round-trips and broadcasts.
- Presence: events propagate between windows.
- No errors or warnings in browser console or dev console.


owenpearson:
Hey @icyJoseph, I've just amended this with a commit signature. Anything additional we need to do to get eyes on this? 🙂

examples/with-ably/app/ably-client-provider.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export default function AblyClientProvider({
+  children,
+}: {
+  children: ReactNode;
+}) {
+  const [clientId] = useState(() => randomClientId());
+  const [client, setClient] = useState<Ably.Realtime | null>(null);
+
+  useEffect(() => {
```
icyJoseph, line 22: Looking at the Ably docs and some other React bits, I think one could refactor here

```tsx
const able = useMemo(() => new Ably.Realtime(), [clientId])
```

Or if one is concerned with the useMemo is an optimization, not a constant lock, then `useState` (like the clientId) - one could fold both into the same state initializer then

Then you only need the effect to close the connection.

That being said, `useSyncExternalStore` could be used to both close and read the connection state?

```tsx
const isAblyConnected = useSyncExternalStore((callback) => {
  ably.connection.on('connected', callback)
  return () =>  realtime.connection.off('connected', callback);
}, () => ably.connection.state , () => false )
```

Something like that, so you can see the connection state from outside the Ably Provider here.

All of that to try and drop the `if(!client)` branch that changes the React tree, leading to a full client remount when the client is set.

<details>
<summary>Why changing a tree like that is not a good idea</summary>

Consider a page like this:

```jsx
"use client";
import { useEffect } from "react";

export default function Home() {
  useEffect(() => {
    console.log("sync once");
  }, []);
  return <p>Home page</p>;
}
```

If there's a provider in the tree, for example in the root layout, structured like this:

```jsx
"use client";

import type { ReactNode } from "react";

import { useEffect, useState } from "react";

export function Provider({ children }: { children: ReactNode }) {
  const [isLoading, setIsLoading] = useState(true);

  useEffect(() => {
    const timeout = setTimeout(() => {
     /* simulate some synchronization work */
      setIsLoading(false);
    }, 1_000);

    return () => clearTimeout(timeout);
  });

  if (isLoading)
    return (
      <>
        <p>Loading...</p>

        {children}
      </>
    );

  return <>{children}</>;
}
```

This causes `sync once` to print twice, in production, and can lead to bad UX (measurable via web vitals), and so on, because everything is discarded and recreated.

Whereas, restructuring to something like:

```tsx
  return (
    <>
      <>{isLoading && <p>Loading...</p>}</>
      {children}
    </>
```

Won't do the double `sync once`.
</details>

owenpearson: Hey, thanks for the detailed review 🙂

I take your point about `if (!client)`. I've reworked this a bit now so the AblyProvider is nested deeper in the tree around the components that need it and children only mount once when the client is ready. Let me know if you still have any issues with the patterns used here.

As for the suggestions to use `useMemo` or `useState` I don't think that would work here, the Ably client creates a stateful websocket connection and typically doesn't make sense to be instantiated during SSR so we want to ensure that it only gets created on the client side.

I like your suggestion for `useSyncExternalStore` 👍 we aren't exposing connection state in this demo at the moment so I'd err on the side of keeping it simple, but I will raise it upstream because our connection state hooks should probably be using that pattern.

icyJoseph: Ah right, I forgot about the SSR pass, haha, it happens. 

Nice. Thanks for the updates. I'll take one more look.

examples/with-ably/app/page.tsx
```diff
@@ -0,0 +1,5 @@
+import HomeClient from "./home-client";
+
+export default function Home() {
+  return <HomeClient />;
```
icyJoseph, line 4: Nice, now I am just being picky haha, but most of HomeClient could be at this Server Component level, and instead one could compose both the ChatArea and Provider here. 

A nit some might say, but it showcases the composition of client components. Approving though, if you wanna do that change still, I'll merge this and other community PRs, in about 10-hours ish (Europe morning)

icyJoseph (dismissed):

icyJoseph:
@owenpearson there's two lint issues:

```console
  19:10  error  'ChatArea' was used before it was defined                      @typescript-eslint/no-use-before-define
  32:12  error  Definition for rule '@next/next/no-img-element' was not found  @next/next/no-img-element
```

owenpearson:
@icyJoseph Apologies for this! I've amended these and ran the checks locally so should be green this time

icyJoseph (approved):

PR #94602, opened 2026-06-09
Remove render server

timneutkens:
### What?

Remove the obsolete `render-server.ts` lifecycle adapter and instantiate `NextServer` directly from `router-server.ts`.

### Why?

The adapter no longer provides a meaningful process boundary. It duplicates server initialization and keeps directory-keyed lifecycle state that can instead be owned directly by each router-server initialization.

### How?

Move `ServerInitResult` into shared server types, capture the prepared `NextServer` and request handler in `router-server.ts`, pass explicit dev lifecycle callbacks into bundler setup, route requests through the prepared handler, import sandbox cache-clearing helpers directly from their context module, and remove the obsolete render-worker compatibility environment variable.

### Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --write <changed files>`
- `npx eslint --config eslint.config.mjs --fix <changed files>`
- `pnpm --filter=next types`
- `pnpm --filter=next build`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-turbo test/e2e/custom-server/custom-server.test.ts`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-webpack test/e2e/custom-server/custom-server.test.ts`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-webpack test/e2e/custom-server/custom-server.test.ts`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-turbo test/development/app-hmr/hmr-env.test.ts`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-webpack test/development/app-hmr/hmr-env.test.ts`


packages/next/src/server/config.ts
```diff
@@ -1782,15 +1782,13 @@ export default async function loadConfig(
   }
 
   // Original implementation continues below...
-  if (!process.env.__NEXT_PRIVATE_RENDER_WORKER) {
```
timneutkens, line 1928: This env var was leftover from when render-server was a separate process. No longer relevant.

PR #94603, opened 2026-06-09
Properly set response-based OTEL attributes with adapters 

mischnic:
**Ignore whitespace changes when viewing the diff**

This fixes some of the failures (particularly the ones where a 500 returned from an App API route showed up as status:200 in OTEL).
But the newly added tests have more failures that I skipped for now

1. Missing `setStatus` call
2. Attribute handling was too early, before `sendResponse` was called:

https://github.com/vercel/next.js/blob/6f4d94ac889db3d23a77847ec4f3a9dadf9dd775/packages/next/src/build/templates/app-route.ts#L289-L296
```js
  return routeModule.handle(nextReq, context).finally(() => { // <-- handle() returns the response object from the user code
        if (!span) return
        span.setAttributes({
          'http.status_code': res.statusCode,          // <---------- this is using some default response object, not what handle() returned
          'next.rsc': false,
        })
```

### DIFF
--- a/packages/next/src/build/templates/app-route.ts
+++ b/packages/next/src/build/templates/app-route.ts
@@ -408,6 +408,7 @@ export async function handler(
           }
           throw err
         } finally {
+          // An IIFE to make early returns easier.
           ;(() => {
             if (!currentSpan) {
               return
--- a/packages/next/src/build/templates/app-route.ts
+++ b/packages/next/src/build/templates/app-route.ts
@@ -286,51 +286,6 @@ export async function handler(
 
   try {
     let parentSpan: Span | undefined
-    const invokeRouteModule = async (span?: Span) => {
-      return routeModule.handle(nextReq, context).finally(() => {
-        if (!span) return
-
-        span.setAttributes({
-          'http.status_code': res.statusCode,
-          'next.rsc': false,
-        })
-
-        const rootSpanAttributes = tracer.getRootSpanAttributes()
-        // We were unable to get attributes, probably OTEL is not enabled
-        if (!rootSpanAttributes) {
-          return
-        }
-
-        if (
-          rootSpanAttributes.get('next.span_type') !==
-          BaseServerSpan.handleRequest
-        ) {
-          console.warn(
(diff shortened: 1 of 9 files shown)


mischnic:
* **#94817** <a href="https://app.graphite.com/github/pr/vercel/next.js/94817?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94603** <a href="https://app.graphite.com/github/pr/vercel/next.js/94603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94603?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/build/templates/app-route.ts
```diff
@@ -447,6 +407,62 @@ export async function handler(
             )
           }
           throw err
+        } finally {
+          ;(() => {
```
timneutkens, line 412: Maybe worth adding a comment that the IIFE is used for early returns logic.

test/e2e/opentelemetry/instrumentation/custom-entrypoint-server.ts
```diff
@@ -81,20 +91,31 @@ async function main() {
...
-    })
+    if (withoutParentSpan) {
+      ;(async () => {
+        try {
+          await handle()
+        } catch (err) {
+          res.statusCode = 500
+          res.end('Internal Server Error')
+        }
+      })()
```
timneutkens, line 105: Dangling promise created here. Probably needs a .catch at least, even though it does try{}catch{} already if it changes in the future it'll be hard to debug. I.e. if res.end() were to throw.

mischnic: This mirrors the existing `tracer.startActiveSpan(...)` version below which is also a dangling promise.
Isn't this the usual pattern?

timneutkens (approved):

PR #94614, opened 2026-06-09
[turbopack] Allocate `Effect`s in an arena

sampoder:
This is a follow-up PR to https://github.com/vercel/next.js/pull/94297#discussion_r3351215197; also adds an `append()` method to `BumpVec`.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
@@ -159,12 +159,6 @@ impl<'a, T> BumpVec<'a, T> {
         }
     }
 
-    /// Move every element out of `other` into `self`, leaving `other` empty. Mirrors
-    /// [`Vec::append`], but takes the `&'a Bump` since growth may reallocate into the arena.
-    pub fn append(&mut self, bump: &'a Bump, other: &mut Self) {
-        self.extend(bump, std::mem::take(other));
-    }
-
     pub fn pop(&mut self) -> Option<T> {
         if self.len == 0 {
             return None;
@@ -517,16 +511,6 @@ mod tests {
 
         let collected: Vec<i32> = v.into_iter().collect();
         assert_eq!(collected, vec![2, 4, 6]);
-    }
-
-    #[test]
-    fn append_moves_and_empties_other() {
-        let bump = Bump::new();
-        let mut a = BumpVec::from_iter_in(&bump, [1, 2]);
-        let mut b = BumpVec::from_iter_in(&bump, [3, 4, 5]);
-        a.append(&bump, &mut b);
-        assert_eq!(&*a, &[1, 2, 3, 4, 5][..]);
-        assert!(b.is_empty());
     }
 
     #[test]
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
@@ -16,7 +16,7 @@ use std::{
 };
 
 use allocator_api2::alloc::Allocator;
-use bumpalo::Bump;
+use bumpalo::{Bump, boxed::Box as BumpBox};
 
 /// A minimal growable vector for the list children of a `JsValue` that grow or are rebuilt after
 /// construction (e.g. `Array.items`, `Object.parts`, `Alternatives.values`, `Add` operands, and the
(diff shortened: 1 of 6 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/graph/effects.rs
```diff
@@ -4,13 +4,13 @@ use turbo_rcstr::RcStr;
...
 use crate::{
-    analyzer::{Bump, JsValue},
+    analyzer::{Bump, BumpVec, JsValue},
     utils::AstPathRange,
 };
 
 #[derive(Debug)]
 pub struct EffectsBlock<'a> {
-    pub effects: Vec<Effect<'a>>,
+    pub effects: BumpVec<'a, Effect<'a>>,
```
bgw: This can likely be a `BumpBox<'a, [...]>` assuming that nothing mutates it in-place.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/mod.rs
```diff
@@ -30,7 +30,7 @@ pub struct VarGraph<'a> {
     /// non-trivial values.
     pub free_var_ids: FxHashMap<Atom, Id>,
 
-    pub effects: Vec<Effect<'a>>,
+    pub effects: BumpVec<'a, Effect<'a>>,
```
bgw: We only have one var graph per analyzed file? I don't think this is worth switching to the arena allocated version.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/effects.rs
```diff
@@ -23,33 +23,33 @@ impl EffectsBlock<'_> {
...
     /// The blocks of an `if ... else` statement.
-    Else { r#else: Box<EffectsBlock<'a>> },
+    Else { r#else: EffectsBlock<'a> },
     /// The blocks of an `if { ... return ... } else { ... } ...` or `if { ... }
     /// else { ... return ... } ...` statement.
     IfElseMultiple {
-        then: Vec<Box<EffectsBlock<'a>>>,
-        r#else: Vec<Box<EffectsBlock<'a>>>,
+        then: Box<[EffectsBlock<'a>]>,
+        r#else: Box<[EffectsBlock<'a>]>,
```
bgw: These weren't getting mutated in-place, so I pushed a commit on top of your PR to convert this to a boxed slice, but this should be a `BumpBox`.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/effects.rs
```diff
@@ -120,15 +120,15 @@ pub enum Effect<'a> {
     /// to determine which effects are executed and remove the others.
     Conditional {
         condition: BumpBox<'a, JsValue<'a>>,
-        kind: Box<ConditionalKind<'a>>,
+        kind: BumpBox<'a, ConditionalKind<'a>>,
         /// The ast path to the condition.
         ast_path: Vec<AstParentKind>,
```
bgw: Boxed slice (`BumpBox<'a, [T]>`) (if it's not mutated in-place) or `BumpVec`?

Same for other `ast_path`s here.

turbopack/crates/turbopack-ecmascript/src/analyzer/bump_vec.rs
```diff
@@ -159,6 +159,12 @@ impl<'a, T> BumpVec<'a, T> {
         }
     }
 
+    /// Move every element out of `other` into `self`, leaving `other` empty. Mirrors
+    /// [`Vec::append`], but takes the `&'a Bump` since growth may reallocate into the arena.
+    pub fn append(&mut self, bump: &'a Bump, other: &mut Self) {
+        self.extend(bump, std::mem::take(other));
+    }
+
```
bgw: We can probably get rid of this? It's not adding enough value, and I think it's mostly getting used for when we're reading `effects` out of `VarGraph`, which probably should just be a `Vec` and not a `BumpVec`.

The real benefit of `Vec::append` (and why it's different than `Vec::extend`) is that it has a slightly more specialized implementation, but this isn't a specialized implementation, nor is it worth it for us to add a specialized implementation.

bgw (approved):
Approving as this is an improvement as-is. Follow-ups can be addressed here if they're small or in a separate PR if they're more complicated.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/effects.rs
sampoder: Going to do this as a follow up

sampoder: nvm https://github.com/vercel/next.js/pull/94614/commits/b8a45da45ede728d318841ff1708c73b6484e213

PR #94615, opened 2026-06-09
[turbopack] Allocate `EarlyReturn` on an arena

sampoder:
Follow-up PR to: https://github.com/vercel/next.js/pull/94297#discussion_r3351233367.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
@@ -165,7 +165,7 @@ mod analyzer_state {
         /// return statement handler.
         cur_fn_return_values: Option<Vec<JsValue<'a>>>,
         /// Stack of early returns for control flow analysis.
-        early_return_stack: Vec<EarlyReturn<'a>>,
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -325,8 +325,9 @@ mod analyzer_state {
         }
 
         /// Helper to access the early_return_stack mutably (for push operations)
-        pub(super) fn early_return_stack_mut(&mut self) -> &mut Vec<EarlyReturn<'a>> {
-            &mut self.state.early_return_stack
+        pub(super) fn push_early_return(&mut self, early_return: EarlyReturn<'a>) {
+            let arena = self.arena;
+            self.state.early_return_stack.push(arena, early_return);
```
bgw, line 330:
```suggestion
            self.state.early_return_stack.push(self.arena, early_return);
```

bgw (changes requested):
I think this is probably not worth it because `AnalyzerState` is only created once per analyzed file.

`EarlyReturn` itself should use arenas internally.

sharaddh (commented):
let arena = self.aren;
self.state.early_return_stack.push(self.arena, early_return);

PR #94622, opened 2026-06-09
[cd] More robust publish pipeline

eps1lon:
Includes multiple changes each separate as commits to make cherry-picking easier. Review by commit is advised.

## Use SemVer prerelease type for NPM dist-tag

...or, if it's not a prerelase, use `latest` if it's the highest version or `backport` otherwise

Removes one more place where the tags are hardcoded. We always mirrored SemVer prerelease types and NPM dist tags (except how stable releases are handled). We can just use the SemVer prerelease as the source of truth to avoid future mistakes.

## Allow tagging semver-lower releases as `@latest` if `@latest` points at a prerelease

In case we accidentally tag a prerelase as `@latest`, we need to be able to move `@latest` to a stable release. Only way to do this with Trusted Publishing is publishing a new version that's tagged with `@latest` (even if that's SemVer lower)

## Undraft any prerelease not just canary

Felt like an oversight when we added `beta` or `rc`.

## topological publishing

Previously, we published all native bindings in parallel and ignored failure (which resulted in missing that `@next/swc-wasm-web` wasn't published in a while). JS code was  published in parallel without bailing if individual publishes failed. Though the overall publish was marked as failed if at least one JS package failed to publish.

This could lead to an inconsistent publish where `@next/env` failed, but `next` succeeded. That version of `next` could not have been installed because its dependency `@next/env` was missing.

Now we publish in topological order. `publish-native` runs first aborting the workflow if any of its publish step fails. Then we run `publish-release` which now uses `pnpm publish`. `pnpm publish` uses topological order, ignores private packages, and skips already published packages by default. We retry the whole `pnpm publish` now to keep the existing retry behavior.

We basically hardcode that at least one package in `publish-release` depends on any package in `publish-native`.

## test plan

Starting and publishing of a release now have a dry mode (`--dry-run`). This makes it considerably easier to replay and test what CI may do.

<details>
<summary>Preparation regardless of release/semver type</summary>

```console
pnpm install && \
pnpm build && \
pnpm dlx turbo@2.9.4 run build-native-release -- --target aarch64-apple-darwin && \
strip -x packages/next-swc/native/next-swc.*.node && \
ls -lh packages/next-swc/native/ && \
rm -rf crates/wasm/pkg crates/wasm/pkg-nodejs && \
pnpm --filter=@next/swc build-wasm --target nodejs && \
mv crates/wasm/pkg crates/wasm/pkg-nodejs && \
rm -rf crates/wasm/pkg crates/wasm/pkg-web && \
pnpm --filter=@next/swc build-wasm --target web && \
mv crates/wasm/pkg crates/wasm/pkg-web && \
ls crates/wasm/pkg-nodejs crates/wasm/pkg-web
```
</details>

```console
$ NEXT_SKIP_NATIVE_POSTINSTALL=1 node scripts/start-release.js --release-type canary --dry-run
$ git log --oneline -1
2877fc31d9 (HEAD -> sebbie/less-tag-hardcoding, tag: v16.3.0-canary.54) v16.3.0-canary.54
$ node scripts/publish-release.js --dry-run
Dry run: not publishing to npm
Publishing 16.3.0-canary.54
Publishing as "canary" dist tag...
Binary next-swc.darwin-x64.node not found, but ignoring due to dry run
[...]
npm notice 📦  @next/swc-darwin-arm64@16.3.0-canary.54
[...]
+ @next/third-parties@16.3.0-canary.54
Dry run: skipping GitHub release un-draft
```

Also tested the same flow with `--release-type stable` and `--release-type beta`.


scripts/publish-native.js
```diff
@@ -2,87 +2,153 @@
...
 const path = require('path')
 const execa = require('execa')
+const semver = require('semver')
 const { Sema } = require('async-sema')
 const { readFile, readdir, writeFile, cp } = require('fs/promises')
 
 const cwd = process.cwd()
+const dryRun = process.argv.includes('--dry-run')
+const maxPublishAttempts = 4
+const publishRetryDelaySeconds = 15
```
bgw: It seems like we have some duplicate logic between `publish-native` and `publish-release` that might be good to move into shared functions and/or constants.

Or maybe this should just be combined into one larger script? When would we run one script but not the other? If pnpm handles the topological sort problem, it doesn't seem like there's benefit from them being separate scripts?

scripts/publish-native.js
```diff
@@ -2,87 +2,153 @@
...
+    }
+
+    console.log(`Publishing as "${tag}" dist tag...`)
+
+    const publish = async (label, pkgPath, attempt = 1) => {
+      let output = ''
+      try {
+        await publishSema.acquire()
+        const child = execa(
+          `npm`,
```
bgw: We should use `pnpm` for publishing in both scripts just for consistency?

scripts/publish-release.js
```diff
@@ -4,56 +4,38 @@
...
-    console.log(err)
-
-    if (err.message && err.message.includes('no tag exactly matches')) {
-      console.log('Nothing to publish, exiting...')
-      return
-    }
-    throw err
+  const version = JSON.parse(
+    await fs.promises.readFile(path.join(cwd, 'lerna.json'), 'utf-8')
+  ).version
```
bgw: This uses `lerna.json`, but the publish-release` script uses `let version = require('@next/swc/package.json').version`?

bgw (approved):

scripts/publish-native.js
eps1lon: Yeah I was conservative with the diffs but at this point it's just annoying for testing to have them separate. And unifying resolves the other review points

PR #94625, opened 2026-06-09
[turbopack] Rename variables in `path_join` and add tests

sampoder:
`results` became `segments` and `results_final` became `locked_prefix`, the `segments` name I'm not too attached to, I'm just trying to communicate that that vector stores the "other" parts of the path that aren't fixed and can change. Meanwhile, the `locked_prefix` has to be at the start - it contains variables we don't know the value of (and anything that came before it) or also the starting "." or "..". I think this is what "final" was originally trying to communicate. 

I also had Claude generate tests for this function based on the pre-existing tests in the `node` repo. There was some divergence related to absolute paths if anyone is curious:

| Input | turbopack | node |
|---|---|---|
| `["", "foo"]` | `/foo` | `foo` |
| `["", "", "foo"]` | `/foo` | `foo` |
| `["", "..", "..", "/foo"]` | `/../../foo` | `../../foo` |
| `["/"]` | `` (empty) | `/` |
| `["/", "."]` | `` (empty) | `/` |
| `["/foo", "../../../bar"]` | `/../../bar` | `/bar` |
| `["/", ".."]` | `/..` | `/` |
| `["/", "..", ".."]` | `/../..` | `/` |
| `["", "."]` | `` (empty) | `.` |
| `[""]` | `` (empty) | `.` |
| `["", ""]` | `` (empty) | `.` |

I can follow-up and fix these as well if that'd be worthwhile, though I think it was mentioned that we don't support absolute paths

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/well_known/mod.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/well_known/mod.rs
@@ -230,8 +230,8 @@ fn path_join<'a>(arena: &'a Bump, args: BumpVec<'a, JsValue<'a>>) -> JsValue<'a>
     if args.is_empty() {
         return rcstr!(".").into();
     }
-    let mut results_final: SmallVec<[JsValue<'a>; 16]> = SmallVec::new();
-    let mut results: SmallVec<[JsValue<'a>; 16]> = SmallVec::new();
+    let mut locked_prefix: SmallVec<[JsValue<'a>; 16]> = SmallVec::new();
+    let mut segments: SmallVec<[JsValue<'a>; 16]> = SmallVec::new();
     for arg in args {
         let arg_parts = if let Some(str) = arg.as_str() {
             let split = str.split('/');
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/well_known/mod.rs
```diff
@@ -1096,3 +1099,145 @@ fn protobuf_loader<'a>(arena: &'a Bump, prop: JsValue<'a>) -> JsValue<'a> {
...
+        render(&path_join(arena, args))
+    }
+
+    /// Cases where `path_join`'s static-analysis result matches the runtime
+    /// behaviour of Node's `path.posix.join`.
+    ///
+    /// Mirrors the `joinTests` table in Node's `test/parallel/test-path-join.js`:
+    /// <https://github.com/nodejs/node/blob/main/test/parallel/test-path-join.js>
+    #[test]
+    fn matches_node_path_posix_join() {
```
lukesandberg, line 1149: can you add the cases where we diverge from node here also, either commented out or with a comment saying that node computes something different

lukesandberg (approved):

PR #94626, opened 2026-06-09
[cd] Allow tagging semver-lower releases as `@latest` if `@latest` points at a prerelease

eps1lon:
`@latest` pointing at a prerelease is always a mistake so subsequent stable release should automatically get `@latest`.


gnoff (approved):

eps1lon:
Targeted wrong branch. Need to be merged into `next-16-2` so that `16.2.9` is tagged at latest. Can't rebase because `next-16-2` contains unsigned commits.

PR #94630, opened 2026-06-09
Fix PPR resume metadata tree mismatch for bot user agents

### LINKED ISSUE #93370: PPR resume mismatch on RSC prefetch with htmlLimitedBots renders __next_metadata_boundary__

I do not have a small public repo ready yet. This was reproduced in a production Next.js 16.2.3 App Router application, and the minimal request/config shape is below.

Reference workaround guard script: https://gist.github.com/mdotk/3c05d464194466800a8ec56df817dd7d

### To Reproduce

Use an App Router application with Cache Components / PPR enabled and metadata streaming disabled globally with `htmlLimitedBots: /.*/`.

```ts
// next.config.ts
import type { NextConfig } from "next";

  cacheComponents: true,
  htmlLimitedBots: /.*/,
  experimental: {
    ppr: true,
  },
};

export default nextConfig;
```

Create any public PPR route that renders metadata and contains a client-side path that triggers App Router prefetch, for example `router.prefetch("/some-public-route")` from a client component after hover/focus.

Then request the same public route through the RSC prefetch request shape:

```bash
curl -i \
  -H 'RSC: 1' \
  -H 'Next-Router-Prefetch: 1' \
  -H 'Accept: text/x-component' \
  https://example.com/some-public-route
```

Normal document requests to the same URL return clean HTML and complete metadata in `<head>`. Bot-like document requests also return clean HTML when `htmlLimitedBots: /.*/` is set.

Current behavior:

The server logs a React resume mismatch on the RSC prefetch path:

```txt
Error: Expected the resume to render <div> in this slot but instead it rendered <__next_metadata_boundary__>. The tree doesn't match so React will fallback to client rendering.
digest: '3219170805'
```

Expected behavior:

The RSC prefetch path should not render a different metadata subtree than the route's prerender/resume expectation. With `htmlLimitedBots: /.*/`, metadata rendering should be invariant for this route, or the prefetch response should avoid producing a resume tree that collides with `__next_metadata_boundary__`.

```txt
Next.js: 16.2.3
React: version bundled by Next 16.2.3
Node.js: 20.20.2
Runtime: self-hosted next start
App Router: yes
Cache Components: enabled
PPR: enabled
htmlLimitedBots: /.*/
```

App Router, Metadata, Partial Prerendering, Runtime

next start (local), Other self-hosted production runtime

### Additional context

This was initially suspected to be a route-level Cache Components boundary problem, but the error reproduced only on the App Router RSC prefetch request shape. Normal browser HTML, Googlebot HTML, and Facebook external hit HTML were clean.

The production workaround was two-layered:

1. Stop using `router.prefetch()` on public content links and use document prefetch instead:

```ts
const link = document.createElement("link");
link.rel = "prefetch";
link.as = "document";
link.href = href;
link.dataset.gtaboomDocumentPrefetch = "true";
document.head.appendChild(link);
```

2. Add a proxy/middleware safety valve for public RSC prefetch requests:

```ts
function isAppRouterRscPrefetchRequest(request: NextRequest) {
  return (
    request.headers.get("rsc") === "1" &&
    request.headers.get("next-router-prefetch") === "1"
  );
}

if (!isAdminPath && isAppRouterRscPrefetchRequest(request)) {
  return new Response(null, {
    status: 204,
    headers: {
      "Cache-Control": "no-store",
      "CDN-Cache-Control": "no-store",
      "X-GTA-Boom-Prefetch": "rsc-disabled",
    },
  });
}
```

We also added a guard script in our app to prevent public `router.prefetch()` from being reintroduced while this class of issue exists. The guard scans public source for `router.prefetch(` and fails pre-deploy if found: https://gist.github.com/mdotk/3c05d464194466800a8ec56df817dd7d

### LINKED ISSUE #93401: PPR metadata resume mismatch despite htmlLimitedBots fully disabling streaming metadata

https://github.com/mdotk/next-ppr-metadata-repro

### To Reproduce

Clone the repo and run:

```bash
npm install
npm run build
npm run start
```

In another terminal:

```bash
npm run repro:check
```

Or manually request the PPR route with different user agents:

```bash
curl -sS \
  -A 'Mozilla/5.0 (Macintosh; Intel Mac OS X 14_7) AppleWebKit/537.36 Chrome/124 Safari/537.36' \
  -o /tmp/browser.html \
  http://127.0.0.1:3099/posts/alpha

curl -sS \
  -A 'Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Googlebot/2.1; +http://www.google.com/bot.html) Safari/537.36' \
  -o /tmp/googlebot.html \
  http://127.0.0.1:3099/posts/alpha
```

The app uses:

```ts
const nextConfig = {
  cacheComponents: true,
  htmlLimitedBots: /.*/,
}
```

The route is a PPR route in the production build:

```txt
└ ◐ /posts/[slug]
  ├ /posts/[slug]
  └ /posts/alpha
```

Current behavior:

- `htmlLimitedBots: /.*/` should fully disable streaming metadata per the docs.
- Browser-like UA returns 200, but the initial head does not contain the route metadata and the response contains the hidden streaming metadata wrapper.
- Googlebot/Bingbot return 200 with route metadata in the initial head.
- Empty/no-UA requests also stream metadata outside the initial head.
- The server logs a React resume mismatch:

```txt
Error: Expected the resume to render <div> in this slot but instead it rendered <__next_metadata_boundary__>. The tree doesn't match so React will fallback to client rendering.
digest: '2491006542'
```

Expected behavior:

With `htmlLimitedBots: /.*/`, the prerender/no-UA path and runtime paths should use a consistent metadata component tree shape, or otherwise avoid a PPR resume mismatch. A documented "fully disable streaming metadata" config should not still allow the PPR shell to be generated with a streaming metadata wrapper.

### What appears to be happening

In Next 16.2.3, `shouldServeStreamingMetadata(userAgent, ".*")` returns `false` for a non-empty browser UA, Googlebot, and Bingbot. But the app-page template still has a no-UA branch that forces streaming metadata:

```ts
const serveStreamingMetadata =
  botType && isRoutePPREnabled
    ? false
    : !userAgent
      ? true
      : shouldServeStreamingMetadata(userAgent, nextConfig.htmlLimitedBots)
```

That means the PPR prerender/export shell can be built with the streaming metadata tree shape, while runtime requests with non-empty UAs can render the blocking metadata tree shape. The metadata component tree changes from a hidden `<div>` wrapper to direct `<__next_metadata_boundary__>`, which matches the resume mismatch.

This is in the same problem area as:

https://github.com/vercel/next.js/issues/92087

And this closed PR appears directly relevant:

https://github.com/vercel/next.js/pull/90259

```txt
Next.js: 16.2.3
React: 19.1.2
React DOM: 19.1.2
Node.js tested locally: 23.3.0
Runtime: next build + next start
App Router: yes
Cache Components: enabled
PPR: enabled through cacheComponents
htmlLimitedBots: /.*/
```

App Router, Metadata, Partial Prerendering, Runtime

next build, next start

### Additional context

The minimal reproduction digest is `2491006542`. The important signal is the metadata tree-shape mismatch: expected `<div>`, got `<__next_metadata_boundary__>`.

pnodet:
### Fixing a bug

- Related issues: fixes #93401, fixes #93370
- Tests added: `test/production/standalone-mode/ppr-resume-bot-metadata`

## What

Requests from bot user agents to PPR-enabled routes fail to resume the postponed state and fall back to client rendering:

```
Error: Expected the resume to render <div> in this slot but instead it rendered <__next_metadata_boundary__>. The tree doesn't match so React will fallback to client rendering.
```

This happens in production (minimal mode) with the **default config** — no `htmlLimitedBots` override needed — for any recognized bot (Googlebot as a DOM bot, Twitterbot/Bingbot/etc. as HTML-limited bots). The result is that the dynamic content is deferred to client rendering for exactly the requests that need server-rendered HTML the most (crawlers), while regular browser requests are unaffected.

## Why

A resume must produce the exact tree shape that the prerender postponed. Prerenders always render with streaming metadata enabled (`serveStreamingMetadata` is set to `true` during build/export), and the metadata component tree has a different shape depending on that flag — the streaming variant wraps the metadata boundary in a `<div hidden>` (`createMetadataComponents`).

The route handler however forces the flag to `false` for any bot user agent on a PPR-enabled route:

```js
// packages/next/src/build/templates/app-page.ts
const serveStreamingMetadata =
  botType && isRoutePPREnabled
    ? false
    : ...
```

The intended behavior for bots on PPR routes is a single-pass blocking render (`shouldWaitOnAllReady`), which works when the server controls the whole response. In minimal mode however the platform serves the prerendered shell from its cache and only invokes the server to resume the postponed state — so the bot opt-out is applied to one half of a render whose other half was produced with streaming metadata enabled. When the page is dynamic at the segment level, the metadata slot is part of the postponed hole, and the resume renders `<MetadataBoundary>` where the postponed state expects the `<div>`.

The same mismatch is reproducible with a custom `htmlLimitedBots` config that matches browser user agents (#93401, #93370) — that path is also covered by this fix for resumes.

## How

When `renderOpts.postponed` is present (i.e. the render is a resume), always render with streaming metadata enabled, matching the prerender that produced the postponed state. The user-agent based opt-out still applies to single-pass renders.

The added test builds a `cacheComponents` app with a dynamic page and dynamic `generateMetadata`, runs the standalone server in minimal mode, and replays the build-time postponed state with `Next-Resume`. Before the fix, the Googlebot and Twitterbot cases fail with the tree mismatch above; the browser case passes:

```
✓ should resume without a metadata tree mismatch for a browser
✕ should resume without a metadata tree mismatch for Googlebot
✕ should resume without a metadata tree mismatch for Twitterbot
```

After the fix all three pass.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -330,6 +330,31 @@ import { RENDER_STAGES_BY_DATA_KIND } from '../dynamic-rendering-utils'
 import type { ValidationPrefetchKind } from './instant-validation/instant-validation'
 import { hasNonRootStaticParams } from '../lib/params-utils'
 
+/**
+ * Determines whether streaming metadata should be served for this render.
+ *
+ * The metadata component tree has a different shape depending on whether
+ * streaming metadata is enabled (see `createMetadataComponents`). Prerenders
+ * always render with streaming metadata enabled (`serveStreamingMetadata` is
+ * set to `true` during build/export), so a render that resumes a postponed
+ * prerender must also render with it enabled — regardless of the requesting
+ * user agent — or React will detect a tree mismatch ("Expected the resume to
+ * render <div> in this slot but instead it rendered
+ * <__next_metadata_boundary__>") and fall back to client rendering.
+ *
+ * The user-agent based opt-out (HTML-limited bots receive blocking metadata)
+ * only applies to renders that produce the whole document in a single pass.
+ */
+function getServeStreamingMetadata(
+  renderOpts: Pick<RenderOpts, 'serveStreamingMetadata' | 'postponed'>
+): boolean {
+  if (typeof renderOpts.postponed === 'string') {
+    return true
+  }
+
+  return !!renderOpts.serveStreamingMetadata
+}
+
 export type GetDynamicParamFromSegment = (
   // The LoaderTree to extract the dynamic param from
   loaderTree: LoaderTree
@@ -685,7 +710,7 @@ async function generateDynamicRSCPayload(
     url,
   } = ctx
 
-  const serveStreamingMetadata = !!ctx.renderOpts.serveStreamingMetadata
+  const serveStreamingMetadata = getServeStreamingMetadata(ctx.renderOpts)
 
   if (!options?.skipPageRendering) {
     const preloadCallbacks: PreloadCallbacks = []
@@ -2096,7 +2121,7 @@ async function getRSCPayload(
     getDynamicParamFromSegment,
     query
   )
-  const serveStreamingMetadata = !!ctx.renderOpts.serveStreamingMetadata
+  const serveStreamingMetadata = getServeStreamingMetadata(ctx.renderOpts)
   const hasGlobalNotFound = !!tree[2]['global-not-found']
 
   const { Viewport, Metadata, MetadataOutlet } = createMetadataComponents({
@@ -2249,7 +2274,7 @@ async function getErrorRSCPayload(
   let Viewport: ComponentType | null = null
   let Metadata: ComponentType | null = null
   if (shouldRenderMetadataAndViewport) {
-    const serveStreamingMetadata = !!ctx.renderOpts.serveStreamingMetadata
+    const serveStreamingMetadata = getServeStreamingMetadata(ctx.renderOpts)
     const metadataComponents = createMetadataComponents({
       tree,
       parsedQuery: query,
--- a/test/production/standalone-mode/ppr-resume-bot-metadata/app/dynamic/loading.js
+++ b/test/production/standalone-mode/ppr-resume-bot-metadata/app/dynamic/loading.js
@@ -1,0 +1,3 @@
+export default function Loading() {
+  return <p>loading...</p>
+}
--- a/test/production/standalone-mode/ppr-resume-bot-metadata/app/dynamic/page.js
+++ b/test/production/standalone-mode/ppr-resume-bot-metadata/app/dynamic/page.js
@@ -1,0 +1,24 @@
+import { cookies } from 'next/headers'
+import { connection } from 'next/server'
+
+export async function generateMetadata() {
+  // Dynamic metadata: resolved during the resume, not the prerender.
+  await connection()
+
+  return {
+    title: 'dynamic-metadata-title',
+  }
+}
+
+export default async function Page() {
+  // Dynamic at the top level of the page (no Suspense boundary): the whole
+  // segment, including the metadata slot rendered adjacent to it, is part of
+  // the postponed hole instead of the static shell.
+  const store = await cookies()
+
+  return (
+    <main>
+      <p id="content">dynamic content {store.size}</p>
+    </main>
+  )
+}
--- a/test/production/standalone-mode/ppr-resume-bot-metadata/app/layout.js
+++ b/test/production/standalone-mode/ppr-resume-bot-metadata/app/layout.js
@@ -1,0 +1,7 @@
+export default function Layout({ children }) {
+  return (
+    <html lang="en">
+      <body>{children}</body>
+    </html>
+  )
+}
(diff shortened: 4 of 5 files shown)


filiptrivan:
Confirming in production on 16.2.4 — continuous `__next_metadata_boundary__` resume mismatches (HTTP 200 + client fallback) on dynamic-metadata PPR routes in minimal mode, ~hundreds/day, matching the bot-UA path described here. Thanks for the fix.

ogabasseyy:
Added a minimal public repro + live deployment that this PR should address:

- Repo: https://github.com/ogabasseyy/nextjs-ppr-metadata-resume-repro
- Live: https://nextjs-ppr-repro.vercel.app

Reproduces the `Expected the resume to render <div> ... <__next_metadata_boundary__>` mismatch deterministically on **next@16.2.9** with `cacheComponents` + `htmlLimitedBots: /.*/` (browser UA → HTTP 500 on the Vercel edge, 200 for Googlebot; locally → 200 with a client-render fallback). Happy to retest against this branch.

sebastianhenneberg:
I desperately waited for this PR because the errors cause a lot noise in my logs (thousands every hour). I patched my app with this PR, and the errors immediately dropped to zero. Can't wait for this getting merged into one of the upcoming next releases. 🙌

PR #94631, opened 2026-06-09
[turbopack] Create a `chunk_group_bootstrap_params()` function

sampoder:
This method generates the params for this code (random example for reference):

```javascript
(globalThis["TURBOPACK"] || (globalThis["TURBOPACK"] = [])).push([
    "output/1ece_tests_snapshot_basic-tree-shake_export-namespace_input_index_0tbr8qy.js",
    {"otherChunks":["output/0_9x_turbopack-tests_tests_snapshot_basic-tree-shake_export-namespace_input_1mqgz2-._.js"],"runtimeModuleIds":["[project]/turbopack/crates/turbopack-tests/tests/snapshot/basic-tree-shake/export-namespace/input/index.js [test] (ecmascript)"]}
]);
```
This is a pretty small change on its own (no functionality change), it makes it a separate method so that it can be called in future PRs. I will flag that the shape for inline-d bootstrap code will be slightly different (further up the stack). Because it isn't a chunk itself it does not have the first item in the array (`output/1ece_tests_snapshot_basic-tree-shake_export-namespace_input_index_0tbr8qy.js`):

```javascript
(globalThis["TURBOPACK"] || (globalThis["TURBOPACK"] = [])).push({"otherChunks":["static/chunks/node_modules_react-dom_82bf80._.js","static/chunks/_4f9bce._.js","static/chunks/pages_index_tsx_5a9e._.js"],"runtimeModuleIds":[96168]});
```

This item only has `otherChunks` and `runtimeModuleIds`. New code is added in https://github.com/vercel/next.js/pull/94664 to handle this different shape.

### DIFF
--- a/turbopack/crates/turbopack-browser/src/ecmascript/evaluate/chunk.rs
+++ b/turbopack/crates/turbopack-browser/src/ecmascript/evaluate/chunk.rs
@@ -4,7 +4,7 @@ use anyhow::Result;
 use either::Either;
 use indoc::writedoc;
 use serde::Serialize;
-use turbo_rcstr::rcstr;
+use turbo_rcstr::{RcStr, rcstr};
 use turbo_tasks::{ResolvedVc, TryJoinIterExt, ValueToString, Vc, turbobail};
 use turbo_tasks_fs::{File, FileContent, FileSystemPath};
 use turbopack_core::{
@@ -32,9 +32,9 @@ use crate::{
     chunking_context::{CURRENT_CHUNK_METHOD_DOCUMENT_CURRENT_SCRIPT_EXPR, CurrentChunkMethod},
 };
 
-/// An Ecmascript chunk that:
-/// * Contains the Turbopack browser runtime code; and
-/// * Evaluates a list of runtime entries.
+/// An Ecmascript chunk that registers an entrypoint's chunks and runtime module
+/// IDs onto the `globalThis["TURBOPACK"]` queue, which the shared
+/// [`crate::ecmascript::evaluate::runtime::EcmascriptBrowserRuntimeChunk`] drains.
 #[turbo_tasks::value(shared)]
 #[derive(ValueToString)]
 #[value_to_string("Ecmascript Browser Evaluate Chunk")]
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-browser/src/ecmascript/evaluate/chunk.rs
```diff
@@ -134,28 +117,56 @@ impl EcmascriptBrowserEvaluateChunk {
...
-        )?;
+            script_or_path = script_or_path,
+            params = &**params,
+        };
+
+        let mut code = CodeBuilder::new(
+            source_maps,
+            *this.chunking_context.debug_ids_enabled().await?,
+        );
+        write!(code, "{}", statement)?;
```
bgw: Not sure why this got changed from `writedoc!` to `formatdoc!`+`write!`?

```suggestion
        let mut code = CodeBuilder::new(
            source_maps,
            *this.chunking_context.debug_ids_enabled().await?,
        );
        writedoc! {
            code,
            r#"
                (globalThis[{chunk_loading_global}] || (globalThis[{chunk_loading_global}] = [])).push([
                    {script_or_path},
                    {params}
                ]);
            "#,
            chunk_loading_global = StringifyJs(&chunk_loading_global),
            script_or_path = script_or_path,
            params = &**params,
        }?;
```

bgw (approved):

turbopack/crates/turbopack-browser/src/ecmascript/evaluate/chunk.rs
```diff
@@ -134,28 +117,58 @@ impl EcmascriptBrowserEvaluateChunk {
             runtime_module_ids,
         };
 
-        let mut code = CodeBuilder::new(
-            source_maps,
-            *this.chunking_context.debug_ids_enabled().await?,
-        );
+        Ok(Vc::cell(format!("{}", StringifyJs(&params)).into()))
```
bgw: The build manifest is a JSON file. It seems like we're storing a JSON-stringified value inside of a JSON file.

- Don't use StringifyJs here, it's overkill, directly call `serde_json::to_string` instead
- Enable the `serde_json` `raw_value` feature for this crate and `next-core`.
- use `serde_json::RawValue::from_string` to convert the string to a `RawValue`.
- Make sure `SerializedBuildManifest` in `crates/next-core/src/next_manifests/mod.rs` (later in this stack) uses the `RawValue`

`RawValue` skips the re-encoding.

PR #92507, opened 2026-04-08
docs: add Building guide

aurorascharff:
## Summary

Adds a new **Building** guide (`docs/01-app/02-guides/building.mdx`) that covers what happens when you run `next build` and how to work with the output.

### Sections

1. **What `next build` does** — high-level overview of the build phases (setup, route discovery, compilation, static analysis, prerendering, output)
2. **Read the build output** — route table symbols (`○` Static, `◐` Partial Prerender) and what they mean with Cache Components
3. **Pre-render dynamic routes** — `generateStaticParams`, how listed/unlisted params behave, and `partialFallbacks` for high-cardinality sites
4. **Fix prerender errors** — blocking-route error, `<Suspense>`/`loading.js` fix, `--debug-prerender` for stack traces, `--debug-build-paths` for building specific routes

Related: https://github.com/vercel/next.js/pull/92382

Closes: https://linear.app/vercel/issue/DOC-4946/getting-started-building

PR #92382, opened 2026-04-05
docs: add generateStaticParams to Cache Components migration guide

aurorascharff:
### What?

Adds a `generateStaticParams` section to [Migrating to Cache Components](https://nextjs.org/docs/app/guides/migrating-to-cache-components).

### Why?

Developers migrating dynamic routes (for example blogs with `[slug]`) had no guidance in that guide for how `generateStaticParams` behaves with Cache Components, why `next build` can show ◐ instead of ○, or that `return []` becomes a build error.

### How?

- Link to `generateStaticParams` with Cache Components, dynamic routes with Cache Components, and the `empty-generate-static-params` message doc instead of duplicating full behavior.
- Before/after example: empty array vs at least one sample param.
- Short note on ◐ (Partial Prerender) with links to the public pages PPR section and glossary.

Related: https://github.com/vercel/next.js/pull/92507

PR #94649, opened 2026-06-10
docs: expand the Cache Components migration guide

icyJoseph:
Continuation of - https://github.com/vercel/next.js/pull/92382

### What?

Adds a `generateStaticParams` section to [Migrating to Cache Components](https://nextjs.org/docs/app/guides/migrating-to-cache-components).

### Why?

Developers migrating dynamic routes (for example blogs with `[slug]`) had no guidance in that guide for how `generateStaticParams` behaves with Cache Components, why `next build` can show ◐ instead of ○, or that `return []` becomes a build error.

### How?

- Link to `generateStaticParams` with Cache Components, dynamic routes with Cache Components, and the `empty-generate-static-params` message doc instead of duplicating full behavior.
- Before/after example: empty array vs at least one sample param.
- Short note on ◐ (Partial Prerender) with links to the public pages PPR section and glossary.

Related: https://github.com/vercel/next.js/pull/92507

### DIFF
--- a/docs/01-app/02-guides/migrating-to-cache-components.mdx
+++ b/docs/01-app/02-guides/migrating-to-cache-components.mdx
@@ -623,7 +623,7 @@ async function Dashboard() {
 }
 ```
 
-Your page receives `params` and `searchParams` as props, and both are promises. Apply the same pattern: pass the promise straight through to the `<Suspense>`-wrapped component as a prop and `await` it there, rather than at the top of the page. You can also unwrap the promise inline with `.then()` and pass a plain value down; see [Streaming](/docs/app/guides/streaming) for the full pattern.
+Your page receives `params` and `searchParams` as props, and both are promises. Apply the same pattern: pass the promise straight through to the `<Suspense>`-wrapped component as a prop and `await` it there, rather than at the top of the page. You can also unwrap the promise inline with `.then()` and pass a plain value down; see [Streaming](/docs/app/guides/streaming#push-dynamic-access-down) for a similar pattern.
 
 ```tsx filename="app/page.tsx" switcher
 import { Suspense } from 'react'
@@ -717,7 +717,7 @@ async function getProducts() {
 }
 ```
 
-> **Good to know:** Reading uncached or runtime data in a `GET` handler bails out of prerendering by **throwing**. A `try/catch` you already have around other operations will catch that bail-out. This is harmless (prerendering has already stopped and the value isn't used), but a `catch` that logs the error produces noisy build-time logs. To quiet them, set `experimental.hideLogsAfterAbort: true` to hide logs emitted after a bail-out.
+> **Good to know:** Reading uncached or runtime data in a `GET` handler bails out of prerendering by **throwing**. A `try/catch` you already have around other operations will catch that bail-out. If the `catch` block logs the error, it adds noise to the build output. Set `experimental.hideLogsAfterAbort: true` to hide logs emitted after a bail-out.
 
 ## `generateMetadata` and `generateViewport`
 
--- a/docs/01-app/02-guides/migrating-to-cache-components.mdx
+++ b/docs/01-app/02-guides/migrating-to-cache-components.mdx
@@ -561,7 +561,7 @@ async function Post({ params }) {
 }
 ```
 
-Client hooks that read the route need the same treatment. In the App Shell, the params, and by extension the pathname, are unknown, so reading these hooks suspends, wherever the component renders. A nav or breadcrumb in a shared layout, for instance, suspends while Next.js generates the App Shell for any route below it that has dynamic params. Wrap the component that reads the hook in `<Suspense>` (push the read down to the smallest leaf so the rest stays prerendered), or the build fails:
+The same applies to client hooks that read the route. When the route's pathname is fully known, they resolve during prerendering and need no boundary. When it depends on dynamic params not yet known, they suspend, wherever the component sits. A nav or breadcrumb in a shared layout, for instance, suspends while Next.js generates the App Shell for any route below it that has dynamic params. Wrap the component that reads the hook in `<Suspense>` (push the read down to the smallest leaf so the rest stays prerendered), or the build fails:
 
 - [`usePathname`](/docs/app/api-reference/functions/use-pathname)
 - [`useParams`](/docs/app/api-reference/functions/use-params)
--- a/docs/01-app/02-guides/migrating-to-cache-components.mdx
+++ b/docs/01-app/02-guides/migrating-to-cache-components.mdx
@@ -20,7 +20,7 @@ The migration is driven by **instant navigation validation**. With Cache Compone
 
 Start by removing the route segment configs (`dynamic`, `revalidate`, `fetchCache`), then follow the validation insights and errors. Each one names the code to fix, most often uncached data to cache with [`use cache`](/docs/app/api-reference/directives/use-cache) or runtime data to wrap in [`<Suspense>`](https://react.dev/reference/react/Suspense). If a route needs more work than you want to take on right now, you can [opt it out of validation](#opting-out-of-validation) and come back to it later.
 
-Your existing `fetch` caching keeps working as a separate layer, so let the insights and errors guide what to change.
+Your existing `fetch` and `unstable_cache` caching keeps working as a separate layer, so let the insights and errors guide what to change.
 
 Some surfaces have their own steps:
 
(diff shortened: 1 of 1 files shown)


docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
 }
 ```
 
+## Opting out of validation
+
+A validation insight or error tells you a route won't render instantly. Resolve it by following the recommendation: cache the data with [`use cache`](/docs/app/api-reference/directives/use-cache) or wrap it in [`<Suspense>`](https://react.dev/reference/react/Suspense). If you don't want to deal with it right now, set the [`instant`](/docs/app/api-reference/file-conventions/route-segment-config/instant) config to `false` on the segment that raised it (a layout, page, or parallel slot), then come back to it later.
```
aurorascharff: instead of "right now" maybe "at that point in time" or "immediately"? sounds a bit non-docsy

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
...
+  'use cache'
+  cacheLife('hours')
+  cacheTag('users')
+  return db.query.users.findFirst({ where: eq(users.id, id) })
+}
+```
+
+Similarly to the `fetch` cache, `unstable_cache` persists cached values across deployments and across serverless instances.
+
+`use cache` defaults to in-memory storage, so its entries are discarded when the serverless instance is destroyed and are scoped to a single deployment. Use [`use cache: remote`](/docs/app/api-reference/directives/use-cache-remote) or a [cache handler](/docs/app/api-reference/config/next-config-js/cacheHandlers) for storage that survives instance teardown. Even with durable storage, expect cached values to recompute after a new deployment.
```
aurorascharff: this is the same text as above? i guess thats okay to repeat?

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
...
+
+> **Good to know:** Like `unstable_cache`, a `use cache` function cannot read runtime data such as `cookies()` or `headers()`. Read it outside the cached function and pass what you need in as an argument.
```
aurorascharff: Hmmm, well kind of :D maybe link to private? would make their migration easier

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
...
+}
+
+async function getProducts() {
+  'use cache'
+  cacheLife('hours')
+  return db.query('SELECT * FROM products')
+}
+```
+
+> **Good to know:** Reading uncached or runtime data in a `GET` handler bails out of prerendering by **throwing**. A `try/catch` you already have around other operations will catch that bail-out. This is harmless (prerendering has already stopped and the value isn't used), but a `catch` that logs the error produces noisy build-time logs. To quiet them, either set `experimental.hideLogsAfterAbort: true` to hide logs emitted after a bail-out, or call [`unstable_rethrow`](/docs/app/api-reference/functions/unstable_rethrow) at the top of your `catch` so the framework's control-flow error propagates instead of being logged.
```
icyJoseph: Let's discuss this bit... I rather only say hideLogsAfterAbort - but yeah, people have ran into this

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
...
+
+async function Post({ params }) {
+  const { slug } = await params
+  // ...
+}
+```
+
+## `cookies`, `headers`, and `searchParams`
+
+**Wrap runtime data access in `<Suspense>`.** Without Cache Components, reading [`cookies()`](/docs/app/api-reference/functions/cookies), [`headers()`](/docs/app/api-reference/functions/headers), or [`searchParams`](/docs/app/api-reference/file-conventions/page#searchparams-optional) opts the whole route into dynamic rendering. With Cache Components, accessing them outside a [`<Suspense>`](https://react.dev/reference/react/Suspense) boundary raises [`Uncached data was accessed outside of <Suspense>`](/docs/messages/blocking-route). Move the access into a component wrapped in `<Suspense>` so the rest of the page prerenders as a static shell and the dynamic part streams in at request time.
```
aurorascharff: these links are the old errors, maybe make them the new or link both? it depends on version i guess

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
...
+  )
+}
+
+async function Dashboard() {
+  const theme = (await cookies()).get('theme')?.value
+  // ...
+}
+```
+
+`params` and `searchParams` arrive as props on your page, and both are promises. Apply the same pattern: pass the promise straight through to the `<Suspense>`-wrapped component as a prop and `await` it there, rather than at the top of the page.
```
aurorascharff: Dont love how the sentence starts with a code block

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
...
+
+export default function Page({ searchParams }: PageProps<'/'>) {
+  return (
+    <Suspense fallback={<p>Loading...</p>}>
+      <Results searchParams={searchParams} />
+    </Suspense>
+  )
+}
+
+async function Results({ searchParams }: Pick<PageProps<'/'>, 'searchParams'>) {
```
aurorascharff, line 648: might wanna link to streaming guide and promote then()

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -176,18 +207,593 @@ export default async function Page() {
...
+}
+```
+
+Paths you don't return are still served. Next.js serves the App Shell instantly, then upgrades it in the background once the params are known. See [ISR with Cache Components](/docs/app/guides/incremental-static-regeneration-cache-components) for the full prerender-a-subset workflow.
+
+### `dynamicParams` no longer blocks the first visit
+
+**`dynamicParams: true` (the default) now serves an App Shell instead of blocking.** Previously, visiting a param not returned by `generateStaticParams` blocked the response while the page rendered. With Cache Components, Next.js serves the App Shell instantly, then upgrades it in the background with the now-known params. `dynamicParams: false` is unchanged: unspecified paths still 404.
+
+To produce the App Shell, pass the `params` promise into a [`<Suspense>`](/docs/app/api-reference/file-conventions/loading) boundary instead of awaiting it at the top of the component, so unknown params can still prerender.
```
aurorascharff, line 506: TODO add usepathname and useselectedlayoutsegment(s), could link directly to the API refs and say to use the same pattern maybe

aurorascharff (changes requested):

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -20,7 +20,7 @@ The migration is driven by **instant navigation validation**. With Cache Compone
 
 Start by removing the route segment configs (`dynamic`, `revalidate`, `fetchCache`), then follow the validation insights and errors. Each one names the code to fix, most often uncached data to cache with [`use cache`](/docs/app/api-reference/directives/use-cache) or runtime data to wrap in [`<Suspense>`](https://react.dev/reference/react/Suspense). If a route needs more work than you want to take on right now, you can [opt it out of validation](#opting-out-of-validation) and come back to it later.
 
-Your existing `fetch` caching keeps working as a separate layer, so let the insights and errors guide what to change.
+Your existing `fetch` and `unstable_cache` caching keeps working as a separate layer, so let the insights and errors guide what to change.
```
aurorascharff, line 23: "Insights" should be capitalized maybe

aurorascharff (approved):

docs/01-app/02-guides/migrating-to-cache-components.mdx
icyJoseph: mmm then it should be like that Everywhere, not sure we are doing that just now, I can follow up

aurorascharff: yeah np can be done later

PR #94650, opened 2026-06-10
docs: add prefetchInlining config reference

### LINKED ISSUE #94458: Docs: `prefetchInlining` not described anywhere

Missing docs

https://github.com/vercel/next.js/pull/92863

_No response_

ifer47:
## What

- Adds a `prefetchInlining` `next.config.js` API reference page.
- Documents the default behavior and the `false` / object forms.
- Covers the `maxSize` and `maxBundleSize` thresholds and the request-count vs deduplication trade-off.
- Links related navigation and prefetching docs.

## Why

`experimental.prefetchInlining` is currently only discoverable through code, PRs, and release notes. This gives the option a canonical docs page and addresses the missing docs report.

Closes #94458

## Duplicate check

I checked open PRs for `prefetchInlining` and `prefetchInlining docs`; the current open matches are runtime/test work, not a config reference page.

## Checks

- `pnpm exec prettier --check docs/01-app/03-api-reference/05-config/01-next-config-js/prefetchInlining.mdx`
- pre-commit `lint-staged` (Prettier + ESLint)

Docs-only change; no runtime tests were run.

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/prefetchInlining.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/prefetchInlining.mdx
@@ -1,26 +1,24 @@
 ---
 title: prefetchInlining
-description: Configure App Router segment prefetch inlining.
+description: Override how the App Router bundles small prefetch responses together.
 version: experimental
 related:
   title: Related
   description: View related API references and guides.
   links:
-    - app/api-reference/config/next-config-js/cacheComponents
-    - app/api-reference/config/next-config-js/partialPrefetching
     - app/api-reference/components/link
-    - app/guides/runtime-prefetching
+    - app/guides/prefetching
 ---
 
-`prefetchInlining` controls whether small App Router segment prefetch responses are bundled into fewer requests. It is enabled by default.
+When the App Router prefetches a route, it can bundle small segment responses into a single response instead of requesting each one separately. This reduces the number of prefetch requests at the cost of duplicating some shared segment data across routes. This behavior is on by default, and most apps should leave it that way.
 
-The App Router can prefetch segment data independently so shared layouts can be reused across routes. Inlining reduces the number of prefetch requests by including small segment responses in another segment's response, trading some deduplication for lower request overhead.
+`experimental.prefetchInlining` is the configuration surface for that behavior. It exists so you can override the defaults or turn inlining off while debugging navigation or measuring request volume. **You generally shouldn't set this option.**
 
-Most apps do not need to configure this option. It is primarily useful when you need to tune or temporarily disable prefetch inlining while testing navigation behavior or request volume.
+> **Good to know**: The inlining behavior is a permanent part of the App Router. Only the `experimental.prefetchInlining` configuration is experimental, so its options may still change.
 
 ## Usage
 
-To disable prefetch inlining, set `experimental.prefetchInlining` to `false`:
+To turn off prefetch inlining, set `experimental.prefetchInlining` to `false`:
 
 ```ts filename="next.config.ts" switcher
 import type { NextConfig } from 'next'
@@ -45,7 +43,7 @@ const nextConfig = {
 module.exports = nextConfig
 ```
 
-To tune the inlining thresholds, pass an object:
+To override the thresholds instead of disabling inlining, pass an object. Any value you omit keeps its default:
 
 ```ts filename="next.config.ts" switcher
 import type { NextConfig } from 'next'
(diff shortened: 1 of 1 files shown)


icyJoseph:
Hi, you need to have verified commits.

docs/01-app/03-api-reference/05-config/01-next-config-js/prefetchInlining.mdx
```diff
@@ -0,0 +1,105 @@
+---
+title: prefetchInlining
```
icyJoseph, line 2: Whether or not to document this one...

You see the thing is that, you'll almost almost never want to set this back to false.

It'll be the default value moving forward. 

That being said, we can still document some of the behavior, while still registering that the behavior is correct and well stablished, but the experimental flag exists as an experimental API to control the stablished behavior.

docs/01-app/03-api-reference/05-config/01-next-config-js/prefetchInlining.mdx
```diff
@@ -0,0 +1,105 @@
...
+Inlining can reduce request count for prefetched routes, which is useful on high-latency networks or platforms where request volume affects cost. The trade-off is that shared layout data may be duplicated across multiple prefetched route responses instead of being fetched once and reused from the client cache.
+
+Lower thresholds preserve more per-segment deduplication. Higher thresholds inline more segment data and can reduce the number of prefetch requests further.
+
+## Version History
+
+| Version | Change                                 |
+| ------- | -------------------------------------- |
+| 16.3.0  | `prefetchInlining` enabled by default. |
+| 16.2.0  | `experimental.prefetchInlining` added. |
```
icyJoseph: Let's reframe: this option is an override people generally shouldn't reach for, not a tuning knob. 

I'd lead with the behavior (which is permanent and default), clarify that only the config is experimental, drop the cacheComponents link (inlining isn't tied to it), and fix the two dead related links.

`````suggestion
---
title: prefetchInlining
description: Override how the App Router bundles small prefetch responses together.
version: experimental
related:
  title: Related
  description: View related API references and guides.
  links:
    - app/api-reference/components/link
    - app/guides/prefetching
---

When the App Router prefetches a route, it can bundle small segment responses into a single response instead of requesting each one separately. This reduces the number of prefetch requests at the cost of duplicating some shared segment data across routes. This behavior is on by default, and most apps should leave it that way.

`experimental.prefetchInlining` is the configuration surface for that behavior. It exists so you can override the defaults or turn inlining off while debugging navigation or measuring request volume. **You generally shouldn't set this option.**

> **Good to know**: The inlining behavior is a permanent part of the App Router. Only the `experimental.prefetchInlining` configuration is experimental, so its options may still change.

## Usage

To turn off prefetch inlining, set `experimental.prefetchInlining` to `false`:

```ts filename="next.config.ts" switcher
import type { NextConfig } from 'next'

  experimental: {
    prefetchInlining: false,
  },
}

export default nextConfig
```

```js filename="next.config.js" switcher
/** @type {import('next').NextConfig} */
const nextConfig = {
  experimental: {
    prefetchInlining: false,
  },
}

module.exports = nextConfig
```

To override the thresholds instead of disabling inlining, pass an object. Any value you omit keeps its default:

```ts filename="next.config.ts" switcher
import type { NextConfig } from 'next'

  experimental: {
    prefetchInlining: {
      maxSize: 2048,
      maxBundleSize: 10240,
    },
  },
}

export default nextConfig
```

```js filename="next.config.js" switcher
/** @type {import('next').NextConfig} */
const nextConfig = {
  experimental: {
    prefetchInlining: {
      maxSize: 2048,
      maxBundleSize: 10240,
    },
  },
}

module.exports = nextConfig
```

## Reference

| Value    | Description                                                                  |
| -------- | ---------------------------------------------------------------------------- |
| `true`   | Inlines prefetch responses with the default thresholds. This is the default. |
| `false`  | Disables prefetch inlining. Each segment is prefetched as its own request.   |
| `object` | Inlines prefetch responses using the `maxSize` or `maxBundleSize` you set.   |

When you pass an object, the following options control the thresholds. Both are measured in bytes of the gzip-compressed segment response:

| Option          | Type     | Default | Description                                                                             |
| --------------- | -------- | ------- | --------------------------------------------------------------------------------------- |
| `maxSize`       | `number` | `2048`  | Largest a single segment response can be to still be eligible for inlining.             |
| `maxBundleSize` | `number` | `10240` | Largest total size that can be inlined into one bundled prefetch response along a path. |

Lower thresholds keep more per-segment deduplication; higher thresholds inline more data and cut request count further.

## Version History

| Version | Change                                             |
| ------- | -------------------------------------------------- |
| 16.3.0  | `experimental.prefetchInlining` enabled by default. |
| 16.2.0  | `experimental.prefetchInlining` added.             |
`````

**Comment by ifer47** [comment]
Thanks, I updated the branch with a verified SSH-signed commit.

**Review by icyJoseph — CHANGES_REQUESTED** [review]
We need to re-evaluate how we present this flag. I've left a comment with motivation and a comment with a suggestion for file contents

**Comment by icyJoseph** [comment]
@ifer47 are you still onboard? wdyt about the suggestion

**Comment by ifer47** [comment]
Thanks for the nudge, yes, I'm still onboard.

I agree with your framing: this reads better as documenting the default App Router prefetch behavior plus the experimental override surface, rather than presenting the flag as a tuning knob most apps should consider.

I applied your suggested direction in c39dd278f7a282390002aefaef01b26bf2c4df7e:

- lead with the permanent/default inlining behavior
- clarify that only `experimental.prefetchInlining` is experimental
- make the "most apps should leave this alone" guidance explicit
- remove the `cacheComponents`/`partialPrefetching`/`runtime-prefetching` related links
- keep the examples limited to disabling inlining or overriding thresholds for debugging/measurement

The new commit is SSH-verified, and the PR checks are green. Happy to adjust further if you want the wording tightened more.

File: docs/01-app/03-api-reference/05-config/01-next-config-js/prefetchInlining.mdx [hunk 3]
```diff
@@ -0,0 +1,99 @@
...
+  title: Related
+  description: View related API references and guides.
+  links:
+    - app/api-reference/components/link
+    - app/guides/prefetching
+---
+
+When the App Router prefetches a route, it can bundle small segment responses into a single response instead of requesting each one separately. This reduces the number of prefetch requests at the cost of duplicating some shared segment data across routes. This behavior is on by default, and most apps should leave it that way.
+
+`experimental.prefetchInlining` is the configuration surface for that behavior. It exists so you can override the defaults or turn inlining off while debugging navigation or measuring request volume. **You generally shouldn't set this option.**
```
icyJoseph:
```suggestion
The `experimental.prefetchInlining` option lets you override this behavior or disable inlining while debugging navigation issues or measuring request volume. For most applications, there is no need to change the default behavior.
```

icyJoseph (approved):

PR #94661, opened 2026-06-10
[turbopack] Add `inline_chunk_group_bootstrap` to `BrowserChunkingContext` and `chunk_group_bootstrap_params` to `ChunkGroupResult`

sampoder:
We used to generate the runtime file for each `ChunkGroupResult` in `evaluated_chunk_group()`, I've added `chunk_group_bootstrap_params` which are the parameters that were unique to each chunk group. These are inlined into each page's HTML in https://github.com/vercel/next.js/pull/94666. They look like:

```javascript
{"otherChunks":["output/0_9x_turbopack-tests_tests_snapshot_basic-tree-shake_export-namespace_input_1mqgz2-._.js"],
"runtimeModuleIds":["[project]/turbopack/crates/turbopack-tests/tests/snapshot/basic-tree-shake/export-namespace/input/index.js [test] (ecmascript)"]}
```

Whether or not the inline-able code is generated is based on the `inline_chunk_group_bootstrap` which is still false in all cases at the moment but will be turned to true in https://github.com/vercel/next.js/pull/94666.

### DIFF
--- a/turbopack/crates/turbopack-browser/src/chunking_context.rs
+++ b/turbopack/crates/turbopack-browser/src/chunking_context.rs
@@ -880,6 +880,7 @@ impl ChunkingContext for BrowserChunkingContext {
                 referenced_assets: OutputAssets::empty_resolved(),
                 references: ResolvedVc::cell(references),
                 availability_info,
+                chunk_group_bootstrap_params: None,
             }
             .cell())
         }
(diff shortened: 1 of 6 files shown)


turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -932,11 +943,25 @@ impl ChunkingContext for BrowserChunkingContext {
...
+                && matches!(chunk_group, ChunkGroup::Entry(_))
+            {
+                Some(
+                    evaluate_chunk
+                        .chunk_group_bootstrap_params()
+                        .owned()
+                        .await?,
+                )
+            } else {
+                assets.push(ResolvedVc::upcast(evaluate_chunk.to_resolved().await?));
```
bgw:
```suggestion
            let evaluate_chunk = self.generate_evaluate_chunk(ident, other_assets, entries).to_resolved().await?;
            let chunk_group_bootstrap_params = if this.inline_chunk_group_bootstrap
                && matches!(chunk_group, ChunkGroup::Entry(_))
            {
                Some(
                    evaluate_chunk
                        .chunk_group_bootstrap_params()
                        .owned()
                        .await?,
                )
            } else {
                assets.push(ResolvedVc::upcast(evaluate_chunk));
```

bgw (approved):

PR #94664, opened 2026-06-10
[turbopack] Add `registerEntry()` to handle inline bootstrapping

sampoder:
In https://github.com/vercel/next.js/pull/94631, we started pushing items to `globalThis["TURBOPACK"]` that were of a slightly different shape. See the comment on https://github.com/vercel/next.js/pull/94631#issue-4626372740. 

This PR adds a new method `registerEntry()` that is able to take that new shape (the object without the file name as an argument) and still instantiate the module + load the other chunks.


turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -68,6 +68,25 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
       }
```
bgw, line 70: Replace this with a call to `registerEntry`? Or some helper function shared between them?

turbopack/crates/turbopack-ecmascript-runtime/src/browser_runtime.rs
```diff
@@ -232,14 +232,23 @@ pub async fn get_browser_runtime_code(
...
     writedoc!(
         code,
         r#"
+            function registerChunkOrEntry(registration) {{
+                if (Array.isArray(registration)) {{
+                    registerChunk(registration);
+                }} else {{
+                    BACKEND.registerEntry(registration);
+                }}
+            }}
```
bgw: Let's keep as little code inside `writedoc!` as possible

bgw (approved):

PR #94666, opened 2026-06-10
[turbopack] Inline the chunk group bootstrap in Next.js to drop the per-route runtime

sampoder:
This PR builds on top of the other PRs in this stack, it is the PR that switches to inlining the bootstrap code and dropping the runtime file that each route had.

### DIFF
--- a/packages/next/src/client/route-loader.ts
+++ b/packages/next/src/client/route-loader.ts
@@ -12,6 +12,8 @@ declare global {
   interface Window {
     __BUILD_MANIFEST?: Record<string, string[]>
     __BUILD_MANIFEST_CB?: Function
+    __TURBOPACK_PAGE_BOOTSTRAP?: Record<string, unknown>
+    __TURBOPACK_CHUNK_LOADING_GLOBAL?: string
     __SERVER_FILES_MANIFEST?: RequiredServerFilesManifest
     __MIDDLEWARE_MATCHERS?: ProxyMatcher[]
     __MIDDLEWARE_MATCHERS_CB?: Function
@@ -246,6 +248,22 @@ export function createRouteLoader(assetPrefix: string): RouteLoader {
   const styleSheets: Map<string, Promise<RouteStyleSheet>> = new Map()
   const routes: Map<string, Future<RouteLoaderEntry> | RouteLoaderEntry> =
     new Map()
+  const bootstrappedRoutes: Set<string> = new Set()
+
+  // Bootstrap a client-loaded route (navigation/prefetch) so its entry registers via
+  // `window.__NEXT_P`. The initial page is bootstrapped in the document.
+  function bootstrapRoute(route: string): void {
+    // Gated for DCE
+    if (!process.env.__NEXT_TURBOPACK_SHARED_RUNTIME) return
+    if (process.env.NODE_ENV === 'development') return
+    if (bootstrappedRoutes.has(route)) return
+    const params = self.__TURBOPACK_PAGE_BOOTSTRAP?.[route]
+    const global = self.__TURBOPACK_CHUNK_LOADING_GLOBAL
+    if (params == null) return
+    // `global` is always defined alongside the params map (see manifest-loader).
+    bootstrappedRoutes.add(route)
+    ;(self as any)[global!].push(params)
+  }
 
   function maybeExecuteScript(
     src: TrustedScriptURL | string
@@ -345,7 +363,10 @@ export function createRouteLoader(assetPrefix: string): RouteLoader {
               return Promise.all([
                 entrypoints.has(route)
                   ? []
-                  : Promise.all(scripts.map(maybeExecuteScript)),
+                  : Promise.all(scripts.map(maybeExecuteScript)).then((r) => {
+                      bootstrapRoute(route)
+                      return r
+                    }),
                 Promise.all(css.map(fetchStyleSheet)),
               ] as const)
             })
--- a/packages/next/src/pages/_document.tsx
+++ b/packages/next/src/pages/_document.tsx
@@ -12,7 +12,10 @@ import type {
 import type { ScriptProps } from '../client/script'
 import type { NextFontManifest } from '../build/webpack/plugins/next-font-manifest-plugin'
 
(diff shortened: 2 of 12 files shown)


packages/next/src/client/route-loader.ts
```diff
@@ -246,6 +248,19 @@ export function createRouteLoader(assetPrefix: string): RouteLoader {
   const styleSheets: Map<string, Promise<RouteStyleSheet>> = new Map()
   const routes: Map<string, Future<RouteLoaderEntry> | RouteLoaderEntry> =
     new Map()
+  const bootstrappedRoutes: Set<string> = new Set()
+
+  // Run a navigated route's inlined bootstrap so its entry module executes and
+  // registers via `window.__NEXT_P`. The initial page is bootstrapped in the document.
```
bgw: Is this used for navigation? Or initial page load? Or both?

packages/next/src/server/get-page-files.ts
```diff
@@ -20,6 +20,31 @@ export type BuildManifest = {
...
+ * pages, seeding the runtime queue before the shared runtime chunk drains it. Returns
+ * undefined when there's nothing to inline (webpack, or dev).
+ */
+export function getTurbopackChunkGroupBootstrap(
+  buildManifest: BuildManifest,
+  pages: readonly string[]
+): string | undefined {
+  const paramsByPage = buildManifest.pagesChunkGroupBootstrapParams
+  const chunkLoadingGlobal = buildManifest.chunkLoadingGlobal
+  if (!paramsByPage || !chunkLoadingGlobal) return undefined
```
bgw: These should never be null?

packages/next/src/server/get-page-files.ts
```diff
@@ -20,6 +20,31 @@ export type BuildManifest = {
...
+  const g = JSON.stringify(chunkLoadingGlobal)
+  const statements = pages
+    .map((page) => paramsByPage[page])
+    .filter(Boolean)
+    .map(
+      (params) =>
+        `(globalThis[${g}] || (globalThis[${g}] = [])).push(${params});`
+    )
+
+  return statements.length > 0 ? statements.join('\n') : undefined
```
bgw, line 46: Can statements.length be 0? Should this be able to return `undefined`?

packages/next/src/pages/_document.tsx
```diff
@@ -128,6 +131,27 @@ function getDynamicChunks(
...
 
+// Builds one inline <script> seeding the runtime queue with the shared (`/_app`) and
+// current route's bootstrap params, before the shared runtime chunk drains it.
+function getInlineBootstrapScript(context: HtmlProps, props: OriginProps) {
+  const { buildManifest, __NEXT_DATA__, crossOrigin } = context
+  const bootstrap = getTurbopackChunkGroupBootstrap(buildManifest, [
+    '/_app',
+    __NEXT_DATA__.page,
+  ])
+  if (!bootstrap) return null
```
bgw, line 150: Should it actually be possible for `getTurbopackChunkGroupBootstrap` to return `null` or `undefined`?

packages/next/src/client/route-loader.ts
```diff
@@ -246,6 +248,19 @@ export function createRouteLoader(assetPrefix: string): RouteLoader {
...
+  function bootstrapRoute(route: string): void {
+    if (process.env.NODE_ENV === 'development') return
+    if (bootstrappedRoutes.has(route)) return
+    const params = self.__TURBOPACK_PAGE_BOOTSTRAP?.[route]
+    const global = self.__TURBOPACK_CHUNK_LOADING_GLOBAL
+    if (params == null || !global) return
```
bgw: When would this happen?

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4092,9 +4094,15 @@ async function renderToStream(
...
         nonce,
-        '/_not-found/page'
+        UNDERSCORE_NOT_FOUND_ROUTE_ENTRY
       )
 
+      const errorBootstrapScriptContent = process.env.__NEXT_DEV_SERVER
+        ? bootstrapScriptContent
+        : getTurbopackChunkGroupBootstrap(buildManifest, [
+            UNDERSCORE_NOT_FOUND_ROUTE_ENTRY,
+          ])
```
bgw: Is this meaningfully different than `bootstrapScriptContent` above? Is `page !== UNDERSCORE_NOT_FOUND_ROUTE_ENTRY`

packages/next/src/shared/lib/turbopack/manifest-loader.ts
```diff
@@ -585,11 +585,29 @@ export class TurbopackManifestLoader {
...
-    )};self.__BUILD_MANIFEST_CB && self.__BUILD_MANIFEST_CB()`
+    // Expose each route's bootstrap params and the chunk-loading global to the client
+    // so `route-loader` can instantiate a navigated page's entry module. The server
+    // stores params as a JSON string per route; deserialize into the client manifest.
+    const pageBootstrapParams: Record<string, unknown> = {}
+    let chunkLoadingGlobal: string | undefined
+    for (const m of this.buildManifests.values()) {
+      if (m.chunkLoadingGlobal) chunkLoadingGlobal = m.chunkLoadingGlobal
+      const paramsByRoute = m.pagesChunkGroupBootstrapParams
+      if (!paramsByRoute) continue
```
bgw: This parameter in `SerializedBuildManifest` is not nullable? https://github.com/vercel/next.js/blob/158bce5aea5a068aacdf95e94a2ecfa2816443b7/crates/next-core/src/next_manifests/mod.rs#L110

bgw (approved):

packages/next/src/server/get-page-files.ts
sampoder: yeah, because this is feature flagged now this is possible

packages/next/src/server/app-render/app-render.tsx
sampoder: yes - bootstrapScriptContent would be for the original page you were requesting before the error happened

PR #94669, opened 2026-06-10
test: add PPR resume segment id collision regression test

pnodet:
Related issue: #94670

### What?

Adds an e2e regression test asserting that a Partial Prerender response never contains duplicate Fizz segment ids (`<div hidden id="S:x">`), that every `$RC("B:x","S:x")` completion script references an unambiguous segment, and that boundary content is not swapped between Suspense holes.

### Why?

On Next.js 16.2.9 (current `latest`) the fixture in this PR reproduces a visible page corruption out of the box:

- At build time, the stored HTML for a PPR route is composed of two Fizz passes: the prerender prelude plus the resume-and-abort pass that materializes postponed-but-not-dynamic boundaries (outlined as hidden segments, ids allocated past the prelude's counter).
- The postponed state persisted in the route's `.meta` still carries the **pre-resume** `nextSegmentId` (e.g. `nextSegmentId: 2` while the stored document already contains `S:2`–`S:4`).
- At request time, the resume render for the dynamic boundary allocates new segment ids from that stale counter, producing duplicate `S:x` ids in a single document. Since `$RC` resolves elements via `getElementById`, the wrong fragment is moved into the wrong Suspense hole — in the fixture, the dynamic footer literally receives a cached widget's content.

Version matrix established with this fixture:

| Version | web streams path | node streams path |
|---|---|---|
| 16.2.9 (latest) | broken | flag not available |
| 16.3.0-canary.0 – .37 | broken | ok (opt-in) |
| 16.3.0-canary.38+ | broken (`experimental.useNodeStreams: false`) | ok (default since #94311) |

So current canary passes this test by default because #94311 switched the default serving path to Node streams — the web-streams resume path still has the stale-counter issue (reproducible on canary with `experimental.useNodeStreams: false`), and the entire stable 16.2.x line is affected with no opt-out.

This test pins the now-correct default behavior so the collision cannot regress, and documents the failure mode. Happy to follow up with a targeted fix for the web-streams path (persisting the advanced segment counter after the build-time resume pass) if that path is expected to remain supported — or this can simply guard the default path if web streams are on the way out.

### How?

Fixture: `cacheComponents: true`, three Suspense-wrapped `'use cache'` widgets large enough to be outlined into hidden segments in the stored document (`expire` ≥ 5 minutes so they are included in the static shell), plus one large `await connection()` boundary so the route is a Partial Prerender with a stored postponed state and a runtime resume that allocates new segment ids.

### DIFF
--- a/test/e2e/app-dir/cache-components-resume-segment-ids/app/dynamic-footer.tsx
+++ b/test/e2e/app-dir/cache-components-resume-segment-ids/app/dynamic-footer.tsx
@@ -1,0 +1,23 @@
+import { connection } from 'next/server'
+
+// Mirrors a request-time footer: a genuinely dynamic boundary that
+// postpones, making the route Partial Prerender and storing a postponed
+// state alongside the prerendered HTML. Its content is large enough that
+// the runtime resume outlines parts of it into NEW hidden segments
+// (allocated from the stored nextSegmentId counter).
+export async function DynamicFooter() {
+  await connection()
+
+  const rows = Array.from({ length: 600 }, (_, i) => `footer-row-${i}`)
+
+  return (
+    <footer id="dynamic-footer">
+      <p>rendered at request time</p>
+      <ul>
+        {rows.map((row) => (
+          <li key={row}>{row}</li>
+        ))}
+      </ul>
+    </footer>
+  )
+}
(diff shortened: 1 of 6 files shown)


eps1lon:
This was a React bug which got fixed in https://github.com/react/react/pull/36779 and synced in https://github.com/vercel/next.js/pull/94951

PR #94683, opened 2026-06-11
Add Owner Stack to "`prefetch={true}` navigation without Partial Prefetching" warning

eps1lon:
by capturing an Owner Stack eagerly during render and reusing that when we issue the warning.

Either we capture the Owner Stack eagerly or we check the route eagerly during render. We don't capture the Owner Stack eagerly all the time so render performance impact is reduced. Memoizing should be. If the parents change, the Component would be remounted so we shouldn't ever print a stale stack.

Also moved the test suite from `test/e2e` to `test/dev`. It already skipped all its test outside of dev so this makes sure it doesn't impose any test scheduling overhead which relies on static information such as test location.


packages/next/src/client/app-dir/link.tsx
```diff
@@ -595,6 +595,23 @@ export default function LinkComponent(
...
+    // gracefully.
+    if (
+      process.env.NODE_ENV !== 'production' &&
+      process.env.__NEXT_CACHE_COMPONENTS &&
+      fetchStrategy === FetchStrategy.Full
+    ) {
+      return React.captureOwnerStack()
+    }
+    return undefined
+  }, [fetchStrategy])
```
lubieowoce: hmm is it worth skipping the memo+closure in prod? we can do that at the top level
```ts
const useOwnerStackInDev =
  process.env.NODE_ENV !== "production" && process.env.__NEXT_CACHE_COMPONENTS
    ? (fetchStrategy: FetchStrategy) => {
        return useMemo(() => {
          if (fetchStrategy === FetchStrategy.Full) {
            return React.captureOwnerStack();
          }
          return undefined;
        }, [fetchStrategy]);
      }
    : (_fetchStrategy: FetchStrategy) => {
        return undefined;
      };
```

packages/next/src/client/components/segment-cache/navigation.ts
```diff
@@ -263,13 +263,29 @@ export function navigateToKnownRoute(
...
         `A <Link prefetch={true}> navigated to "${url.pathname}", but Partial ` +
           `Prefetching is not enabled for that route, so its dynamic data was ` +
           `included in the prefetch. Enable Partial Prefetching app-wide by ` +
           `setting \`partialPrefetching: true\` in next.config, or per-route by ` +
           `exporting \`const prefetch = 'partial'\` from the page or layout.`
       )
+      const ownerStack = 'ownerStack' in link ? link.ownerStack : undefined
+      if (ownerStack === undefined) {
+        console.error(
+          '' +
```
lubieowoce, line 276: `'' +` is a noop, looks like a leftover from something?
```suggestion
```

eps1lon: It's for formatting so that the lines are aligned.

packages/next/src/client/app-dir/link.tsx
eps1lon: Definitely. Can't wait to have React Compiler run over this code before publishing to eliminate that whole problem space: [`db998c3` (this PR)](https://github.com/vercel/next.js/pull/94683/changes/db998c3bb3380886f0dbd36676f35765281cb943)

packages/next/src/client/components/segment-cache/navigation.ts
lubieowoce: oh cute! alright then

lubieowoce (approved):

PR #94689, opened 2026-06-11
codemod: write pnpm overrides to pnpm-workspace.yaml on pnpm v11+

eps1lon:
## Why

pnpm v11 silently stopped reading `pnpm.overrides` (and `pnpm.patchedDependencies`) from `package.json`. The canonical location moved to `pnpm-workspace.yaml#overrides`. There is no deprecation warning — see [pnpm/pnpm#11536](https://github.com/pnpm/pnpm/issues/11536) and [pnpm settings docs](https://pnpm.io/settings).

`next-codemod upgrade` uses overrides to pin `@types/react` / `@types/react-dom` during React major upgrades. On pnpm v11 those pins were a no-op.

## What

- Detect pnpm major version via `npm_config_user_agent` (with a `pnpm --version` fallback), mirroring the helper in `create-next-app`.
- On pnpm v11+ (or when the version cannot be determined), `writeOverridesField` routes through a new `writePnpmWorkspaceOverrides` helper that reads/merges/writes `pnpm-workspace.yaml#overrides`. Existing keys in the file are preserved.
- pnpm v10 and below keep writing `package.json#pnpm.overrides`.
- `js-yaml` is `require()`d lazily inside the v11+ branch, so non-pnpm and pre-v11 upgrades don't load it.
- Monorepo behavior is unchanged (still writes relative to `cwd`); pnpm overrides have always been workspace-root-only, so this matches the prior shortcoming rather than introducing a new one.

## Test plan

- [x] `pnpm --filter=@next/codemod build` passes.
- [ ] Manual: `pnpm test:upgrade-fixture pnpm-v11-overrides` with pnpm v11 on PATH — expect overrides merged into the existing `pnpm-workspace.yaml` (see fixture README for the expected diff).
- [ ] Existing fixtures continue to pass on pnpm v10.

### DIFF
--- a/packages/next-codemod/bin/__testfixtures__/pnpm-v11-overrides/README.md
+++ b/packages/next-codemod/bin/__testfixtures__/pnpm-v11-overrides/README.md
@@ -1,0 +1,40 @@
+Verifies that on pnpm v11+, `@types/react` / `@types/react-dom` overrides are
+written to `pnpm-workspace.yaml#overrides` (merged into existing keys) rather
+than `package.json#pnpm.overrides`. pnpm v11 silently ignores
+`pnpm.overrides` in `package.json` — see
+https://github.com/pnpm/pnpm/issues/11536 and https://pnpm.io/settings.
+
+Run this fixture with pnpm v11+ on PATH.
+
+```diff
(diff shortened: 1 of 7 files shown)


eps1lon:
Reopening from upstream branch.

PR #94685, opened 2026-06-11
fix(codemod): write pnpm overrides to pnpm-workspace.yaml

KaushalMikaelson:
Description

Starting from pnpm v11, the package manager no longer reads configuration from the `pnpm` field in `package.json`, including `pnpm.overrides`.

The upgrade codemod currently writes React type overrides into `packageJson.pnpm.overrides`, which causes pnpm v11+ projects to emit warnings and ignore the generated overrides.

This change updates the pnpm path in `writeOverridesField()` to write and merge overrides into `pnpm-workspace.yaml` instead.

Changes

* Add support for updating `pnpm-workspace.yaml` with generated overrides.
* Merge generated overrides with existing overrides already present in the workspace file.
* Preserve existing npm, yarn, and bun behavior.
* Add tests covering workspace file creation and override merging.
* Update fixture documentation to reflect the new behavior.

Related Issue

Fixes #94653

Testing

* Added unit tests for override merging behavior.
* Verified linting with:

  ```bash
  pnpm exec eslint packages/next-codemod/bin/upgrade.ts
  ```
* Verified formatting with:

  ```bash
  git diff --check
  ```

PR #94690, opened 2026-06-11
[codemod] Write pnpm overrides to pnpm-workspace.yaml on pnpm v11+

### LINKED ISSUE #94653: codemod: pnpm ignores package.json overrides as of v11

https://github.com/Brycero/pnpm-override

### To Reproduce

1. Run `pnpx @next/codemod upgrade latest` against the sample project using pnpm v11+.
2. pnpm will return a warning about the pnpm overrides field no longer being read in package.json: `[WARN] The "pnpm" field in package.json is no longer read by pnpm. The following keys were ignored: "pnpm.overrides". See https://pnpm.io/settings for the new home of each setting`

Currently, Codemod adds pnpm overrides to package.json, [which is ignored by pnpm as of v11](https://pnpm.io/package_json). Codemod should create/edit overrides in pnpm-workspace.yaml instead.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:26 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T8132
  Available memory (MB): 24576
  Available CPU cores: 10
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 11.5.2
Relevant Packages:
  next: 16.3.0-canary.47 // Latest available version is detected (16.3.0-canary.47).
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Not sure

next dev (local)

### Additional context

_No response_

eps1lon:
pnpm v1 stopped reading `pnpm.overrides` (and `pnpm.patchedDependencies`) from `package.json`. The canonical location moved to `pnpm-workspace.yaml#overrides`.

`next-codemod upgrade` uses overrides to pin `@types/react` / `@types/react-dom` during React major upgrades. On pnpm v11 those pins were a no-op.

On pnpm v11+ (`npm_config_user_agent` with a `pnpm --version` fallback, mirroring the helper in `create-next-app`) we now add `overrides` to the yaml config.

Fixes https://github.com/vercel/next.js/issues/94653
Closes https://github.com/vercel/next.js/pull/94685

## Test plan

- [x] Manual: `pnpm test:upgrade-fixture bin/__testfixtures__/pnpm-v11-overrides latest` with pnpm v11 on PATH — expect overrides merged into the existing `pnpm-workspace.yaml` (see fixture README for the expected diff).

### DIFF
--- a/packages/next-codemod/lib/handle-package.ts
+++ b/packages/next-codemod/lib/handle-package.ts
@@ -1,8 +1,56 @@
 import findUp from 'find-up'
 import execa from 'execa'
+import { execSync } from 'node:child_process'
 import { basename } from 'node:path'
 
 export type PackageManager = 'npm' | 'pnpm' | 'yarn' | 'bun'
+
+/**
+ * Get the full version string for the given package manager.
+ *
+ * First tries to parse from `npm_config_user_agent` (e.g., "pnpm/9.13.2 npm/? ..."),
+ * then falls back to spawning `<packageManager> --version`.
+ *
+ * Returns null if unable to determine the version.
+ *
+ * Mirrors `packages/create-next-app/helpers/get-pkg-manager.ts`.
+ */
+export function getPackageManagerVersion(
+  packageManager: PackageManager
+): string | null {
+  const userAgent = process.env.npm_config_user_agent || ''
+  const userAgentMatch = userAgent.match(
+    new RegExp(`${packageManager}/([\\d.]+[\\w.-]*)`)
+  )
+  if (userAgentMatch) {
+    return userAgentMatch[1]
+  }
+
+  try {
+    const version = execSync(`${packageManager} --version`, {
+      encoding: 'utf8',
+      stdio: ['pipe', 'pipe', 'ignore'],
+    }).trim()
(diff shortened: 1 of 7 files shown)


devjiwonchoi (approved):

packages/next-codemod/lib/handle-package.ts
```diff
@@ -1,9 +1,57 @@
...
+}
+
+/**
+ * Get the major version of pnpm being used.
+ * Returns null if unable to determine the version.
+ */
+export function getPnpmMajorVersion(): number | null {
+  const version = getPackageManagerVersion('pnpm')
+  if (!version) return null
+  const major = parseInt(version.split('.')[0], 10)
```
icyJoseph, line 51: could swap out for `semver` (already installed)

eps1lon: We're using the same code in create-next-app already. I'd rather keep this in sync since it has little to no test coverage.

PR #94694, opened 2026-06-11
Persist `'use cache: private'` entries in dev

unstubbable:
Private caches were never stored in a cache handler, so every reload in `next dev` re-ran them from scratch and they registered as a cache miss on each load. This change persists `'use cache: private'` entries in development in a dedicated built-in in-memory handler so that warm reloads are fast. The handler is gated on `process.env.__NEXT_DEV_SERVER` and is kept out of the kind-keyed handlers map so it can never be replaced by a user-configured `default` handler: private cache entries can hold data specific to the incoming request (for example, derived from its cookies or headers) and must never reach a remote or otherwise persistent handler. Production keeps private caches non-persisted as before.

The coarse cache-handler key for a dev private cache is scoped by the request's cookies and headers so entries for requests with different request data don't collide. It excludes Next-internal cookies that aren't application data (the HMR refresh hash, already part of the cache key, and the instant-navigation cookie, which toggles while a navigation lock is held) and the transport and content-negotiation headers that vary between otherwise-equivalent requests (a browser reload adds `cache-control`, and `accept` and `sec-fetch-*` differ between an HTML navigation and an RSC request). Keying by only the cookies and headers a cache actually reads is left as a follow-up; read root params are already tracked that way, the same as for public caches. The cache life is forced to `revalidate: 0` with a 5-minute `expire`, so each read serves the stale entry immediately and warms a fresh one in the background through the existing stale-while-revalidate path.

Cross-request deduplication now applies to private caches in development too, so concurrent requests with identical request data share a single fill; it remains skipped in production where request-specific data must not be shared across requests. To make this deterministic, `saveToCacheHandler` resolves the metadata a cross-request joiner awaits only after the entry has been written to the handler, so the joiner finds it when re-reading its recomputed key. This closes a pre-existing race in that path, present for public caches too but never surfaced by their cross-request test, where the joiner's metadata could resolve before the handler write had landed.

The defensive invariant that rejected reading a private entry from a handler is removed: it only existed to narrow `cacheContext.kind` for a code path that no longer needs it, and dev now legitimately reads persisted private entries while production never registers a private handler to read from.

### DIFF
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -360,13 +360,18 @@ const COOKIES_EXCLUDED_FROM_PRIVATE_CACHE_KEY = new Set<string>([
 // spurious misses (a browser reload adds `cache-control`/`pragma` that an
 // initial navigation doesn't, and `accept`/`sec-fetch-*` differ between an HTML
 // navigation and an RSC or prefetch request for the same page), or are
-// connection- and proxy-level rather than application data. Header names are
-// lowercased by `HeadersAdapter`, so every entry here is lowercase.
+// connection- and proxy-level rather than application data. The `cookie` header
+// is excluded because cookies are keyed separately below (via the dedicated
+// cookie path, which applies `COOKIES_EXCLUDED_FROM_PRIVATE_CACHE_KEY`);
+// including the raw header would duplicate them and reintroduce the cookies
+// that path excludes. Header names are lowercased by `HeadersAdapter`, so every
+// entry here is lowercase.
 const HEADERS_EXCLUDED_FROM_PRIVATE_CACHE_KEY = new Set<string>([
   'accept',
   'accept-encoding',
   'cache-control',
   'connection',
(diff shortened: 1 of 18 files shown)


gnoff (approved):

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -339,6 +345,89 @@ function computeRootParamsCacheKeySuffix(
...
+  'sec-fetch-site',
+  'sec-fetch-user',
+  'sec-purpose',
+  'te',
+  'upgrade',
+  'upgrade-insecure-requests',
+  'x-forwarded-for',
+  'x-forwarded-host',
+  'x-forwarded-port',
+  'x-forwarded-proto',
```
lubieowoce, line 396: not sure if it matters, but these seem like they *should* actually vary the key (if they're different, they'd probably come from a different user), so we shouldn't do a cache hit
(at least `x-forwarded-for` is supposed to be the client's IP)

lubieowoce (approved):

PR #94705, opened 2026-06-11
docs: Route Handlers to be supported later on

icyJoseph:
I noticed the Restrictions section doesn't explicitly mention Route Handlers. It is only mentioned in a good to know. However, this is a temporary restriction.

### DIFF
--- a/docs/01-app/03-api-reference/04-functions/next-root-params.mdx
+++ b/docs/01-app/03-api-reference/04-functions/next-root-params.mdx
@@ -312,6 +312,8 @@ If the parameter does not exist in the current route's root layout (see [Multipl
 
 ## Restrictions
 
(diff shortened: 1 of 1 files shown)


docs/01-app/03-api-reference/04-functions/next-root-params.mdx
```diff
@@ -312,6 +312,8 @@ If the parameter does not exist in the current route's root layout (see [Multipl
 
 ## Restrictions
 
+Most of these are permanent constraints of how root parameters work. The exception is [Route Handlers](/docs/app/api-reference/file-conventions/route), which are not supported yet but are planned for a future release.
```
icyJoseph, line 315: We don't refer to future stuff, except in very niche cases. This is one of those.

aurorascharff (approved):

PR #94726, opened 2026-06-11
[turbopack] Rename the edge runtime to the self-contained runtime

sampoder:
This isn't really a functionality change but it makes later PRs up the stack make more sense. At this point though it's just a name change.


bgw (approved):

mischnic:
It's still weird that this is a subvariant of the browser runtime.
But unrelated to your change here

PR #94727, opened 2026-06-11
[turbopack] Add support for single-entry chunks

sampoder:
This PR adds support for generating single-entry chunks, based on this plan: https://gist.github.com/bgw/356344ea6fb0e897647c478d01c096d3 up until Step 6.

Also adds a basic snapshot test to capture this behaviour.


turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -704,7 +719,21 @@ impl ChunkingContext for BrowserChunkingContext {
...
+            let ecmascript_ty: ResolvedVc<Box<dyn ChunkType>> =
+                ResolvedVc::upcast(Vc::<EcmascriptChunkType>::default().to_resolved().await?);
+            let config = ChunkingConfig {
+                min_chunk_size: usize::MAX,
+                max_chunk_count_per_group: 1,
+                max_merge_chunk_size: usize::MAX,
+                ..Default::default()
+            };
+            return Ok(Vc::cell([(ecmascript_ty, config)].into_iter().collect()));
+        }
```
bgw: Look around and see if there's a better place we could put this override?

turbopack/crates/turbopack-tests/tests/snapshot.rs
```diff
@@ -96,6 +96,10 @@ struct SnapshotOptions {
     scope_hoisting: bool,
     #[serde(default)]
     production_chunking: bool,
+    /// Emit a single self-contained entry chunk (service-worker mode). Only valid with the
+    /// `Browser` runtime; drives `entry_chunk_group` instead of `evaluated_chunk_group_assets`.
```
bgw:
```suggestion
    /// Emit a single self-contained entry chunk (e.g. for service-workers). Only valid with the
    /// `Browser` runtime; drives `entry_chunk_group` instead of `evaluated_chunk_group_assets`.
```

turbopack/crates/turbopack-tests/tests/snapshot.rs
```diff
@@ -589,6 +598,29 @@ async fn run_test_operation(resource: RcStr) -> Result<Vc<FileSystemPath>> {
...
     let chunks = match options.runtime {
+        // Single self-contained entry chunk (service-worker mode): everything is inlined into one
+        // file via `entry_chunk_group`, with no separately-loaded chunks.
+        Runtime::Browser if options.single_chunk => OutputAssetsWithReferenced {
+            assets: ResolvedVc::cell(vec![
+                chunking_context
+                    .entry_chunk_group(
+                        // `expected` expects a completely flat output directory.
+                        chunk_root_path
+                            .join(entry_module.ident().await?.path.file_stem().unwrap())?
```
bgw, line 611:
```suggestion
                            .join(entry_module.ident().await?.path.file_stem().expect("entry module path must have a non-empty file name"))?
```

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -926,16 +955,89 @@ impl ChunkingContext for BrowserChunkingContext {
...
+
+            let chunks = chunks.await?;
+            if chunks.len() != 1 {
+                bail!(
+                    "single-chunk (service-worker) entry produced {} chunks, expected exactly 1. \
+                     This usually means the module graph contains non-ECMAScript chunkable items \
+                     (e.g. CSS, image or font imports), which cannot be inlined.",
+                    chunks.len()
+                );
+            }
```
bgw: We should emit an implementation of Issue: https://turbopack-rust-docs.vercel.sh/rustdoc/turbopack_core/issue/trait.Issue.html

With `IssueSeverity::Error`, and then we should continue on with "best effort" (e.g. just pick the first chunk or generate a stub chunk).

There are a few reasons for this:
- We treat anyhow errors that bubble all the way up to the top-level as internal errors, but actually this error would be the user's fault.
- If an error is the user's fault, we want to keep going after emitting the `Issue`, in case we find more issues, so we can report all of them to the user at once.
- If we do bail out on an error, it might end up invalidating a lot of other cached turbo-task functions, leading to extra recomputation later when the user fixes the issue.
- `Issue` gives us a bunch of fields (title, description, doc links) and uses `StyledString`, so it generates much nicer user-facing outputs.

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -939,16 +967,104 @@ impl ChunkingContext for BrowserChunkingContext {
...
+                SingleChunkProducedMultipleChunksIssue {
+                    path: path.clone(),
+                    chunk_count: chunks.len(),
+                }
+                .resolved_cell()
+                .emit();
+            }
+
+            // Best effort (if chunks != 1): use the first ECMAScript chunk
+            // we produced or use a stub if chunks == 0.
```
bgw:
```suggestion
            // use a stub if chunks == 0, we already emitted an issue
```

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -965,6 +1081,9 @@ impl ChunkingContext for BrowserChunkingContext {
         module_graph: Vc<ModuleGraph>,
         availability_info: AvailabilityInfo,
     ) -> Result<Vc<Box<dyn ChunkItem>>> {
+        if self.await?.single_chunk {
+            bail!("async loaders are not produced in single-chunk (service-worker) mode");
```
bgw: Minor wording:

```suggestion
            bail!("async loaders are not produced in single-chunk (e.g. service-worker) mode");
```

---

Question: Is there any sort of protection against us hitting this bail when creating a service worker?

We don't want to hit `bail`s because they're treated as internal logic errors in Turbopack (similar to a panic), and give extremely verbose output. We should also emit an issue here if it's possible to hit this codepath when defining a service worker in the later PRs.

This should also probably be covered by an e2e test that verifies that async imports in service workers generate a sane error.

Similarly if there's some way we could force a service worker file to generate multiple chunks, we should test the error for that too. (e.g. can I trigger it by adding a css dependency to a service worker?)

turbopack/crates/turbopack-browser/src/ecmascript/evaluate/single_entry_chunk.rs
```diff
@@ -0,0 +1,155 @@
...
+        ))
+    }
+}
+
+#[turbo_tasks::value_impl]
+impl GenerateSourceMap for EcmascriptBrowserSingleEntryChunk {
+    #[turbo_tasks::function]
+    fn generate_source_map(self: Vc<Self>) -> Vc<FileContent> {
+        self.code().generate_source_map()
+    }
```
bgw, line 154: Minor performance nit: This `#[turbo_tasks::function]` isn't providing any benefit from caching this because both `.code()` and `.generate_source_map()` are already turbo-task functions, and this isn't even constructing it's own cell.

```suggestion
    fn generate_source_map(self: Vc<Self>) -> Vc<FileContent> {
        self.code().generate_source_map()
    }
```

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -1044,3 +1172,48 @@ struct ChunkPathInfo {
...
+    }
+
+    async fn description(&self) -> Result<Option<StyledString>> {
+        Ok(Some(StyledString::Stack(vec![
+            StyledString::Line(vec![
+                StyledString::Text(rcstr!(
+                    "A single-chunk entry must produce exactly one ECMAScript chunk, but it \
+                     produced "
+                )),
+                StyledString::Strong(format!("{}", self.chunk_count).into()),
```
bgw: Minor nit: prefer the more explicit thing here

```suggestion
                StyledString::Strong(RcStr::from(format!("{}", self.chunk_count))),
```

bgw (approved):
I think my main concern here is just about making sure we can never hit a `bail` in actual user code, and that we always handle those cases with `Issue`.

turbopack/crates/turbopack-browser/src/ecmascript/evaluate/single_entry_chunk.rs
```diff
@@ -0,0 +1,154 @@
...
+                    .await?,
+            ))
+            .cell(),
+        ))
+    }
+}
+
+#[turbo_tasks::value_impl]
+impl GenerateSourceMap for EcmascriptBrowserSingleEntryChunk {
+    fn generate_source_map(self: Vc<Self>) -> Vc<FileContent> {
```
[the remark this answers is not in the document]
sampoder: has this been fixed?

Issue #81161, opened 2025-07-01
Turbopack dev server uses too much RAM and CPU


dokugo:
### Link to the code that reproduces this issue

https://github.com/dokugo/next-reown-appkit-memory-usage

### To Reproduce

1. Clone https://github.com/dokugo/next-reown-appkit-memory-usage
2. Install dependencies
3. Run dev server with Turbopack: `npm run dev` / `pnpm dev` / etc...
4. Open http://localhost:3000
5. Click through navbar links (routes 0 to 14)
6. Notice that each route compilation causes massive CPU spike and adds ~400 MB to the dev server process
7. See dev server process eventually taking 8000 MB RAM
8. Hard reload at any route to add ~30 MB each time, repeatable until hitting OOM

### Current vs. Expected behavior

**Current behavior:**

Turbopack dev server consumes a huge amount of RAM and CPU, even in an almost empty app with 15 blank routes. 

- Just running npm run dev and opening localhost leads to ~2000 MB RAM usage
- After visiting 15 blank routes, the dev server process takes ~8000 MB (vs ~2500 MB with Webpack)
- Each route compilation adds ~400 MB
- A high CPU spike for a few seconds while compiling each route
- Route compilation ~5x slower than with Webpack
- HMR uses ~5% of CPU instead of the usual 0.1%
- Hard page reload adds ~30 MB each time, it just keeps growing until OOM

**Expected behavior:**

Turbopack dev server consumes a reasonable amount of RAM and CPU (especially while running an almost empty app).

### Provide environment information

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 65144
  Available CPU cores: 16
Binaries:
  Node: 22.16.0
  npm: 10.9.2
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 15.3.4 // Latest available version is detected (15.3.4).
  eslint-config-next: 15.3.4
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Turbopack, Performance

### Which stage(s) are affected? (Select all that apply)

next dev (local)

### Additional context

I've been setting up a Next.js app, in which I needed to add crypto wallet functionality, so I used one of the most popular solutions, Reown Appkit (previously WalletConnect).

[...]

PR #94735, opened 2026-06-12
Filter denied paths from Turbopack watcher events

scarab-systems:
### What?

Filters `turbo-tasks-fs` watcher batches so filesystem events under denied paths are discarded before invalidation work.

### Why?

`Project::project_fs()` already denies generated output paths such as `.next/dev`, but recursive platform watchers can still receive native events from those paths. When `.next/dev/cache/turbopack` is busy, those events do not correspond to project source changes and can keep the dev server doing unnecessary watcher/invalidation work.

Related to #81161.

### How?

- Share the existing denied-path prefix logic with a helper that also handles absolute system paths.
- Drop notify events whose paths are entirely denied before queueing invalidations.
- Keep pathless watcher errors on the existing root-invalidation path.

### Testing

- `cargo +nightly-2026-05-15 fmt --package turbo-tasks-fs --check`
- `cargo +nightly-2026-05-15 test -p turbo-tasks-fs denied_path`
- `cargo +nightly-2026-05-15 test -p turbo-tasks-fs`
- `cargo +nightly-2026-05-15 clippy -p turbo-tasks-fs --tests -- -D warnings`


scarab-systems:
Gentle follow-up on this PR. I saw the automated review note, and I’m happy to adjust the patch if maintainers think this is a direction worth continuing.

If not, I’m also happy to close it rather than leave noise in the queue.

bgw:
Sorry for the slow response, @scarab-systems. I'm finally getting around to looking at the filesystem watcher again. A bit of a brain dump:

In theory, the cost of adding a path to a batch *should* be incredibly cheap (a few hashmap inserts + hashmap lookups), so there *shouldn't* be much of a benefit to filtering them out here.

In practice though, some of our code for creating batches is rather inefficient. https://github.com/vercel/next.js/pull/96114 is a step towards improving overall efficiency. There's a lot of other work we can do (e.g. using arenas to avoid cloning `PathBuf` entirely when creating parent paths).

Denied paths defined at turbopack's `DiskFileSystem` layer are defined as relative paths, but the watcher receives absolute paths from the operating system. Handling denied paths here requires converting absolute paths to relative paths (which your PR does lexically), but doing that robustly is actually surprisingly complicated and expensive (https://github.com/vercel/next.js/pull/95668). It may actually be fine to just do a lexical absolute-to-relative conversion here under the assumption that the OS will give us canonicalized paths, and the fs root is already canonicalized, but unless it's giving us a lot of benefits, I'd rather avoid that can of worms.

This approach would be a lot more valuable if we could mask the update events at the OS level, so that windows/macos doesn't need to notify us about files we don't care about, but from reading through the code of `notify-rs`, it doesn't look like the `unwatch` methods actually do that: https://docs.rs/notify/latest/notify/trait.Watcher.html#tymethod.unwatch

- https://github.com/notify-rs/notify/blob/205b50c3ede6bdb2a66476f5aa1f0ace4f928d9e/notify/src/fsevent.rs#L431-L477
- https://github.com/notify-rs/notify/blob/205b50c3ede6bdb2a66476f5aa1f0ace4f928d9e/notify/src/windows.rs#L398-L433

It looks like `unwatch` only works on a path that you previously gave to `watch`, not on arbitrary child paths.

Similarly `notify-rs` subscribes us to [`EventKind::Access`](https://docs.rs/notify/latest/notify/enum.EventKind.html#variant.Access) events that we don't care about on Linux with `inotify`, and there's no way to remove that from the mask. That might be simple enough for me to send a PR to `notify-rs` for...

PR #94738, opened 2026-06-12
[App Shells] Allow root params in app shells

lubieowoce:
Server-side part of #94773. We remove previously added code that excludes root params from the ShellStatic stage and let them resolve there. We also add some special cases for `params` objects that consist entirely of root params - those are allowed to resolve as well.


lubieowoce:
* **#94738** <a href="https://app.graphite.com/github/pr/vercel/next.js/94738?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94738?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94809** <a href="https://app.graphite.com/github/pr/vercel/next.js/94809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94857** <a href="https://app.graphite.com/github/pr/vercel/next.js/94857?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -2411,51 +2408,6 @@ export async function cache(
...
-              )
-              break
-            }
-            case 'request': {
-              // For a staged dynamic request, assume we're recovering a static shell --
-              // If a session shell is needed, we do it in a separate render
-              await stagedRendering.waitForStage(
-                getStaticLinkDataStage(stagedRendering)
-              )
-              break
```
gnoff, line 2470: removing these delays means the caches that read root params can resolve in the shell but then it's not clear if we are connecting these read params through accumulateRootVaryParam. is that set up in some way I (and codex) can't understand?

packages/next/src/server/request/params.ts
```diff
@@ -564,39 +565,52 @@ function createStagedRenderParamsImpl(
...
-      userspaceParams
-    )
-  }
-
   return makeUntrackedParams(userspaceParams)
 }
 
+function allParamsAreRootParams(underlyingParams: Params, rootParams: Params) {
+  for (const paramName in underlyingParams) {
+    if (!(paramName in rootParams)) {
```
gnoff: need to use `hasOwnProperty` rather than `in` to avoid prototype <> param collisions

test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts
```diff
@@ -289,4 +289,242 @@ describe('App Shell prefetching', () => {
...
+        // It uses a different root param value ("pl", not "en"), so we shouldn't use
+        // the App Shell we got from the link we revealed, and the navigation should block.
+        await browser
+          .elementByCss('a[href="/with-root-param/pl/posts/126"]')
+          .click()
+
+        // Make sure the click had a chance to have an effect.
+        await browser.eval(
+          () => new Promise((resolve) => setTimeout(resolve, 5))
+        )
```
gnoff: If the click isn't processed fast enough would the later assertions pass even if the intended behavior wasn't observed?

packages/next/src/server/use-cache/use-cache-wrapper.ts
lubieowoce: hm, yeah, looks like that's missing. but i don't think this was ever actually wired up correctly? 
the 'cache' branch in root params was always missing this, see how `root-params` looked before i did any of this
https://github.com/vercel/next.js/pull/93801/changes#diff-570afe674cdff7efa0457f65dad7dc94a9cf19124778105be8d2d70eb6443ec7L66
'private-cache' was doing a `break` and thus accumulating the param, but 'cache' was doing an early return. dunno why, might be a bug, but this isn't making it worse. let's investigate that separately

gnoff (approved):

packages/next/src/server/use-cache/use-cache-wrapper.ts
gnoff: maybe but let's add a test where the only use of a root param is in a cache. we can probably demonstrate that the shell will be poisoned for links to other versions of the route with a different param value

test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts
lubieowoce: yeah i guess so. negative assertions ("nothing changed") are tricky. i'm not sure if this is necessary, perhaps playwright's `click()` already accounts for this

lubieowoce: i'll add a pending indicator to the link with `useLinkStatus` and assert on that, that should do the trick. we care about whether the router actually reacted to the click, so we can wait until the status becomes pending

packages/next/src/server/request/params.ts
```diff
@@ -337,48 +331,47 @@ function createStaticPrerenderParams(
...
-          'params',
-          underlyingParamsWithVarying
-        )
+      const { stagedRendering } = prerenderStore
+      if (process.env.__NEXT_APP_SHELLS && stagedRendering) {
+        // Even if all params are static, we need to exclude them from the app shell
+        // by delaying them to the static stage. However, root params are allowed in shells,
+        // so if all the params are root params, they can be included as well.
+        if (
+          !allParamsAreRootParams(underlyingParams, prerenderStore.rootParams)
```
lubieowoce, line 363: hm. we're allowing root params here, i wonder if we need to update the tracking logic so that it uses the root params vary accumulator? i think we don't, the params get merged anyway, and unlike the root params accumulator, here we know which segment the access belongs to

PR #94754, opened 2026-06-12
fix: this in edge sandbox setTimeout/setInterval

lubieowoce:
Our attempts at setting the correct `this` in our `setTimeout`/`setInterval` polyfills were slightly incorrect, because they used the wrong global object.

(Thanks to [cyberjoker](https://hackerone.com/cyberjoker) for discovering this)

### DIFF
--- a/packages/next/src/server/web/sandbox/resource-managers.ts
+++ b/packages/next/src/server/web/sandbox/resource-managers.ts
@@ -23,9 +23,9 @@ abstract class ResourceManager<T, Args> {
 
 class IntervalsManager extends ResourceManager<
   number,
-  Parameters<typeof setInterval>
+  Parameters<typeof webSetIntervalPolyfill>
 > {
-  create(args: Parameters<typeof setInterval>) {
+  create(args: Parameters<typeof webSetIntervalPolyfill>) {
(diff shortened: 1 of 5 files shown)


packages/next/src/server/web/sandbox/resource-managers.ts
```diff
@@ -49,7 +49,10 @@ class TimeoutsManager extends ResourceManager<
   }
 }
 
+export type GlobalObject = import('edge-runtime').EdgeRuntime['context']
+
 function webSetIntervalPolyfill<TArgs extends any[]>(
+  globalObject: GlobalObject,
```
gnoff, line 55: This works, but why not store the context in the resource manager itself and save all the spreading and whatnot

gnoff (approved):

packages/next/src/server/web/sandbox/resource-managers.ts
lubieowoce: We intialize these managers at the top level of the `resource-managers` module, before a context exists. And the surrounding code is written in a way that suggests that multiple contexts can exist. I don't feel like untangling all of that right now, and binding the context at the point where we define the actual setTimeout is definitely correct, so it's kinda just the path of least resistance

PR #88647, opened 2026-01-16
[cna] Ensure created app is not considered the workspace root in pnpm

eps1lon:
Follow-up to https://github.com/vercel/next.js/pull/83168

With `packages`, pnpm will assume the package is the workspace root and demand the `-w` flag (ERR_PNPM_ADDING_TO_ROOT). Even if `packages` is empty.

Since we only need `packages` for v9 (which doesn't even support `ignoreBuiltDependencies`), we just skip generation of pnpm-workspace.yaml entirely which also means we can drop `packages`

I originally skipped the check because I wanted to avoid an expensive `pnpm` invocation. Cursor figured we can use `npm_config_user_agent` (see https://github.com/pnpm/pnpm/issues/3985) to parse the version from which avoids spawning a sub process entirely. We still fallback to `pnpm --quiet --version` if the env variable is not available.

## test plan

- [x] v9 produces no workspace.yml `p create next-app@http://vercel-packages.vercel.app/next/commits/399b272b713abcb1c63c28c93cc082d2b41b031b/create-next-app next-pnpm-built-workspace-yml`
- [x] v10 produces workspace.yml `p create next-app@http://vercel-packages.vercel.app/next/commits/399b272b713abcb1c63c28c93cc082d2b41b031b/create-next-app next-pnpm-built-workspace-yml`

PR #94762, opened 2026-06-12
[cna] Include packages field in pnpm-workspace.yaml for pnpm v10.0-10.4

srajang1805:
PR #88647 removed the packages field from pnpm-workspace.yaml for all pnpm v10+, assuming it is optional across all v10 versions. However, pnpm only made the packages field optional from v10.5.0. Versions 10.0-10.4.x still require it, causing pnpm install to fail with packages field missing or empty.

This adds a getPnpmMinorVersion() helper and uses it to distinguish between v10.0-10.4 (requires packages field) and v10.5+ (packages field can be omitted). The v11+ and v9 paths are unchanged.

### DIFF
--- a/packages/create-next-app/helpers/get-pkg-manager.ts
+++ b/packages/create-next-app/helpers/get-pkg-manager.ts
@@ -63,3 +63,14 @@ export function getPnpmMajorVersion(): number | null {
   const major = parseInt(version.split('.')[0], 10)
   return Number.isNaN(major) ? null : major
 }
+
+/**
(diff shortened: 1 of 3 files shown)


srajang1805:
Would appreciate a review. Thank you.

eps1lon (changes requested):
We're not adding workarounds for broken minor versions of package managers. This is not tenable long-term. You should be able to upgrade to latest pnpm 10.x without issues.

srajang1805:
@eps1lon Thank you.

PR #94764, opened 2026-06-12
refactor: process streaming prerenders more efficiently

lubieowoce:
Introducing a brand new concept: processing a stream by having a callback invoked whenever a chunk is emitted. No more tee()ing streams just to run two loops over the chunks from a render. Turns out node streams had it right all along

This lets us simplify the (stream) runtime prerender and static prerender codepaths, and will be used further in #94738, where we need to run a staged render with some special accumulation conditions on the chunks. This now becomes easy to add, because we can just pick what we do in our `collectChunk` callback instead of writing an entire loop

It will also let us move to switch these codepaths to node streams easier -- the iteration code is the only part that depends on the stream being web or node, the collection logic is generic

follow up work: refactor `accumulateStreamChunksInto` and `countStageBytesUntilAbortNode`  in the same way. I didn't do that here yet to keep the change focused on what upstack needs

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -1735,6 +1735,16 @@ async function finalRuntimeServerPrerender(
     mode.type === 'session-shell-only'
       ? createStageByteLengths()
       : null
+  const collectChunk = (chunk: Uint8Array) => {
+    collectPrerenderChunk(collectedChunks, finalServerController.signal, chunk)
+    if (stageByteLengths) {
+      increaseChunkByteLengths(
+        stageByteLengths,
+        finalStageController.currentStage,
+        chunk.byteLength
+      )
+    }
+  }
 
   await runInSequentialTasks(
     async () => {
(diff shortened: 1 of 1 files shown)


lubieowoce:
* **#94595** <a href="https://app.graphite.com/github/pr/vercel/next.js/94595?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94764** <a href="https://app.graphite.com/github/pr/vercel/next.js/94764?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94764?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4982,11 +4981,11 @@ async function accumulateStreamChunksInto(
     try {
       while (!cancelled) {
         const { done, value } = await reader.read()
-        if (done) {
+        if (done || cancelled) {
```
lubieowoce, line 4984: drive-by fix. the node branch never accumulates anything if `cancelled` is set. a `cancel()` call results in a `{ done: true }` chunk, so this probably doesn't matter, but this guarantees parity.

gnoff (approved):

PR #94775, opened 2026-06-12
fix(dev-overlay): correct open-in-editor path when using a `src` directory

### LINKED ISSUE #94776: Dev overlay "open in editor" 404s when using a `src` directory (path segments reversed)

https://github.com/fhfournier/next-launch-editor-src-repro

### To Reproduce

1. Clone the reproduction (an App Router app **with a `src/` directory**), `pnpm install`, `pnpm dev`.
2. With the dev server running, ask the dev "open in editor" endpoint to open an **app-relative** file (this is exactly what the DevTools route/segment explorer does):

```bash
curl -s -o /dev/null -w "%{http_code}\n" \
  "http://localhost:3000/__nextjs_launch-editor?file=layout.tsx&isAppRelativePath=1"
```

3. Observe `404`. Interactively: open the **Next.js DevTools** → route/segment view → "open in editor" on a segment such as `layout.tsx` → nothing opens, same `…isAppRelativePath=1` request 404s.

> The error overlay's **call-stack frames** are not affected — they open via the full source-mapped path (`frame.file`). The bug is specifically in the `isAppRelativePath=1` branch used by the segment explorer.

**Current:** for `isAppRelativePath` requests the endpoint returns 404 when a `src/` directory is used, so "open in editor" silently does nothing. (`REACT_EDITOR` makes no difference — the file is never found, so the editor is never launched.)

**Expected:** the endpoint resolves the file under `src/app/...` and opens it (it already works for projects without a `src/` directory) → `204`.

### Root cause

`getOverlayMiddleware` joins the `'app'` and `'src'` segments in the wrong order:

`packages/next/src/server/dev/middleware-turbopack.ts` (and the identical block in `middleware-webpack.ts`):

```ts
const appPath = path.join(
  'app',
  isSrcDir ? 'src' : '',
  relativeFilePath
)
```

With `isSrcDir === true` this yields `app/src/<file>` instead of `src/app/<file>`, so `openFileInEditor` can't find it → `found: false` → 404. When there is no `src/` directory the segment is `''` (dropped by `path.join`), which is why the non-`src` case works today.

### Fix

Reorder to `src/app/<file>`; the non-`src` case is unchanged. PR: #94775

```ts
const appPath = path.join(
  isSrcDir ? 'src' : '',
  'app',
  relativeFilePath
)
```

Dev overlay / Error overlay, Turbopack, Webpack

next dev (local)

fhfournier:
### What

The dev error overlay's "open in editor" endpoint (`/__nextjs_launch-editor`) builds an app-relative path with the `'app'` and `'src'` segments in the wrong order, so it 404s for projects that use a `src/` directory.

`packages/next/src/server/dev/middleware-turbopack.ts` (and the identical block in `middleware-webpack.ts`):

```ts
const appPath = path.join(
  'app',
  isSrcDir ? 'src' : '',
  relativeFilePath
)
```

With `isSrcDir === true` this produces `app/src/<file>` instead of `src/app/<file>`. `openFileInEditor` then can't find the file → `found: false` → 404, and the editor never opens (clicking a layout/server-component frame in the overlay does nothing).

### Why the non-`src` case is unaffected

When there is no `src/` directory the segment is `''`, and `path.join` drops empty segments. So both before and after this change the non-`src` path is `app/<file>`. The reorder only changes the result when `src/` is present:

| | before | after |
| --- | --- | --- |
| no `src` | `app/<file>` | `app/<file>` |
| with `src` | `app/src/<file>` ❌ | `src/app/<file>` ✅ |

### Fix

Reorder to `src/app/<file>` in both the Turbopack and webpack dev middleware.

Fixes #94776

### Bug

- [x] Related issues linked using `fixes #number`
- [ ] Tests added — happy to add one; pointers to the right harness for the `/__nextjs_launch-editor` middleware would be welcome.

### Verification

Patched a local Next.js 16.2.9 install in a `--src-dir` app and hit the endpoint:

```
# before
GET /__nextjs_launch-editor?file=layout.tsx&isAppRelativePath=1 → 404
# after
GET /__nextjs_launch-editor?file=layout.tsx&isAppRelativePath=1 → 204   (opens src/app/layout.tsx)
GET /__nextjs_launch-editor?file=missing.tsx&isAppRelativePath=1 → 404  (still 404, as expected)
```


eps1lon (approved):
thank you

PR #94778, opened 2026-06-12
[CC] Instrument tracers resolved through ProxyTracerProvider.getDelegateTracer

### LINKED ISSUE #94753: cacheComponents: OTel span creation bypasses the #82350 tracer patch via ProxyTracer.getDelegateTracer, failing prerenders with next-prerender-random

https://github.com/pnodet/next-prerender-random-repro

### To Reproduce

1. `pnpm install && pnpm dev`
2. `curl http://localhost:3000/early-span` → render fails with `next-prerender-random`
3. `curl http://localhost:3000/late-span` → identical `startSpan`, works

The two pages differ only in **when** `trace.getTracer()` was called. `instrumentation.ts` acquires one tracer **before** registering a plain `NodeTracerProvider` (no Sentry or any vendor SDK involved), which is the standard constructor pattern of OTel instrumentation libraries (`InstrumentationAbstract` calls `this._tracer = trace.getTracer(...)` in its constructor).

**Current:** #82350 ("Allow span creation while prerendering") patches `trace.getTracerProvider().getTracer` so span creation exits `workUnitAsyncStorage`. But a tracer acquired before the delegate is registered on the global `ProxyTracerProvider` is a `ProxyTracer`, and `ProxyTracer._getTracer()` resolves the real tracer through `ProxyTracerProvider.getDelegateTracer()` — which the patch does not wrap. Span-id generation then runs `RandomIdGenerator`'s raw `Math.random()` inside the prerender scope:

```
Error: Route "/early-span" used `Math.random()` before accessing either uncached data (e.g. `fetch()`) or Request data (...). See more info here: https://nextjs.org/docs/messages/next-prerender-random
```

**Expected:** spans created from instrumentation-acquired tracers are allowed during prerendering, per #82350's stated intent.

### Real-world impact

This is how `@sentry/nextjs` breaks `cacheComponents` apps in production: `@sentry/node`'s `_init` constructs all integrations (pg, undici, http — each acquiring a `ProxyTracer`) before `initOpenTelemetry()` registers the provider. Any instrumented operation in a prerenderable scope (e.g. a Postgres query in a PPR shell) then fails **every ISR revalidation** of the route — the static shell never refreshes, on-demand renders can 500, and Sentry reports it as a message-less `Error: null` (`StaticGenBailoutError` has no message; the real error is only in server logs). Production stack on next@16.2.9 + @sentry/nextjs@10.57.0 — note the `ProxyTracer` frame:

```
Error: Route "/product/[category]" used `Math.random()` before accessing either uncached data ...
    at RandomIdGenerator.generateId [as generateSpanId] (@opentelemetry/sdk-trace-base/src/platform/node/RandomIdGenerator.ts:31:41)
    at Tracer.startSpan (@opentelemetry/sdk-trace-base/src/Tracer.ts:82:38)
    at ProxyTracer.startSpan (@opentelemetry/api/src/trace/ProxyTracer.ts:41:30)
    at BoundPool.connect (@sentry/node/src/integrations/tracing/postgres/vendored/instrumentation.ts:549:36)
```

Likely the unresolved cause behind #72572 and the remaining reports on #72024.

### Suggested fix

In `extendTracerProviderForCacheComponents` (`packages/next/src/server/lib/router-utils/instrumentation-node-extensions.ts`), also wrap `ProxyTracerProvider.getDelegateTracer` (or wrap the delegate provider's `getTracer` when the delegate is set), so tracers resolved lazily by pre-registration `ProxyTracer`s get the same `workUnitAsyncStorage.exit` treatment.

```
Operating System:
  Platform: darwin / Linux (Vercel production)
  Arch: arm64
Binaries:
  Node: 24.14.1
  pnpm: 10.33.4
Relevant Packages:
  next: 16.2.9
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
  @opentelemetry/api: 1.9.1
  @opentelemetry/sdk-trace-node: 2.7.1
Next.js Config:
  cacheComponents: true
```

Cache Components, Instrumentation

next dev (local), next start (local), Vercel (Deployed)

pnodet:
Fixes #94753

### Problem

#82350 allows OTel span creation while prerendering by wrapping tracers returned from the global provider's `getTracer`. But tracers acquired **before a delegate provider is registered** are `ProxyTracer`s, which resolve the real tracer through `ProxyTracerProvider.getDelegateTracer()` — never hitting the patched `getTracer`. Span-id generation then runs `RandomIdGenerator`'s raw `Math.random()` inside the prerender scope and fails it with `next-prerender-random`.

This is the standard constructor pattern of OTel instrumentation libraries (`InstrumentationAbstract` calls `trace.getTracer()` in its constructor): every `@sentry/node` integration does it (constructed before `initOpenTelemetry()` registers the provider), and `@vercel/otel` with `registerOTel({ instrumentations: [...] })` as well. In production this fails **every ISR revalidation** of PPR routes that touch an instrumented resource (pg, undici, …) — the static shell never refreshes — surfacing in Sentry as message-less `Error: null` (`StaticGenBailoutError`). Standalone repro: https://github.com/pnodet/next-prerender-random-repro

### Fix

Extract the tracer instrumentation into a helper and also apply it on the `getDelegateTracer` resolution path. On current `@opentelemetry/api` versions the proxy provider's `getTracer` routes through `getDelegateTracer` internally, so the original patch is subsumed by the new one whenever the global provider is a proxy (the WeakSet makes double instrumentation a no-op); both are kept so we don't depend on that internal routing detail and still cover non-proxy global providers.

### Test

Added `t9` to `cache-components-allow-otel-spans`: a span created from a tracer acquired in `instrumentation.node.ts` before `sdk.start()`. Without the fix, runtime prerendering fails with `Next.js encountered the unstable value Math.random() while prerendering` and the page degrades to per-request dynamic rendering; with the fix, the dev suite is 4/4 green and start mode matches baseline.

Note: `should allow creating Spans during prerendering at runtime` fails identically on **unmodified canary** (verified at both 658d3786e7 and a64db6f77f: `expect(t7againValue).toEqual(t7value)` on `/novel/cache` — the runtime-prerendered page is not served from cache on the second load). Pre-existing, not introduced or affected by this change.

### DIFF
--- a/packages/next/src/server/lib/router-utils/instrumentation-node-extensions.ts
+++ b/packages/next/src/server/lib/router-utils/instrumentation-node-extensions.ts
@@ -49,66 +49,103 @@ function extendTracerProviderForCacheComponents(): void {
   // to exit the workUnitAsyncStorage context when generating spans.
   const originalGetTracer = provider.getTracer.bind(provider)
   provider.getTracer = (...args) => {
-    const tracer = originalGetTracer.apply(provider, args)
-    if (WeakTracers.has(tracer)) {
-      return tracer
+    return instrumentTracerForCacheComponents(
+      originalGetTracer.apply(provider, args)
+    )
+  }
+
+  // Tracers acquired before a delegate provider was registered are
+  // ProxyTracers. This is the standard pattern of OTel instrumentation
+  // libraries: `InstrumentationAbstract` calls `trace.getTracer()` in its
+  // constructor, and SDKs (e.g. @sentry/node, @vercel/otel with custom
+  // instrumentations) construct instrumentations before registering the
+  // provider. A ProxyTracer resolves the actual tracer lazily through
+  // `ProxyTracerProvider.getDelegateTracer()`, which does not go through the
+  // patched `getTracer` above, so we need to instrument that resolution path
+  // as well. (On current @opentelemetry/api versions the proxy provider's
+  // `getTracer` routes through this method internally, so the patch above is
+  // subsumed by this one whenever the global provider is a proxy — the
+  // WeakSet makes the double instrumentation a no-op. We keep both patches so
+  // we don't depend on that internal routing detail and still cover non-proxy
+  // global providers.)
+  const proxyProvider = provider as typeof provider & {
+    getDelegateTracer?: (
+      name: string,
+      version?: string,
+      options?: unknown
+    ) => Tracer | undefined
+  }
+  if (typeof proxyProvider.getDelegateTracer === 'function') {
+    const originalGetDelegateTracer =
+      proxyProvider.getDelegateTracer.bind(proxyProvider)
+    proxyProvider.getDelegateTracer = (...args) => {
+      const tracer = originalGetDelegateTracer(...args)
+      return tracer === undefined
+        ? undefined
+        : instrumentTracerForCacheComponents(tracer)
     }
-    const originalStartSpan = tracer.startSpan
-    tracer.startSpan = (...startSpanArgs) => {
-      return workUnitAsyncStorage.exit(() =>
-        originalStartSpan.apply(tracer, startSpanArgs)
+  }
+}
+
+function instrumentTracerForCacheComponents(tracer: Tracer): Tracer {
+  if (WeakTracers.has(tracer)) {
+    return tracer
+  }
+  const originalStartSpan = tracer.startSpan
+  tracer.startSpan = (...startSpanArgs) => {
+    return workUnitAsyncStorage.exit(() =>
+      originalStartSpan.apply(tracer, startSpanArgs)
+    )
+  }
+
+  const originalStartActiveSpan = tracer.startActiveSpan
+  // @ts-ignore TS doesn't recognize the overloads correctly
+  tracer.startActiveSpan = (...startActiveSpanArgs: any[]) => {
+    const workUnitStore = workUnitAsyncStorage.getStore()
+    if (!workUnitStore) {
+      // @ts-ignore TS doesn't recognize the overloads correctly
+      return originalStartActiveSpan.apply(tracer, startActiveSpanArgs)
+    }
+
+    let fnIdx: number = 0
+    if (
+      startActiveSpanArgs.length === 2 &&
+      typeof startActiveSpanArgs[1] === 'function'
+    ) {
+      fnIdx = 1
+    } else if (
+      startActiveSpanArgs.length === 3 &&
+      typeof startActiveSpanArgs[2] === 'function'
+    ) {
+      fnIdx = 2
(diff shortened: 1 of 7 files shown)


pnodet:
**Follow-up**: `span.end()` leaks aren't covered by this fix (and one isn't fixable here)

  This PR closes the span-creation hole — ProxyTracer-resolved tracers running `RandomIdGenerator`'s `Math.random()` at span-id generation inside the prerender scope. That's the right fix for #94753. Flagging two related gaps on
  the span-end path for completeness, since the question naturally comes up: "a span created and ended around a query during render (the pg-pool pattern) also runs time code in `end()` — is that still leaking?"

  **The common case is fine.** I checked `_getTime` in `@opentelemetry/sdk-trace-base` (identical in `1.30.1` and `2.0.1`). The canonical `startSpan()` → query → `span.end()` (no args) pattern does not leak: with no explicit `startTime`,
  `end()` resolves the end time via `otperformance.now() - _performanceStartTime` — i.e. `performance.now()`, which Next deliberately leaves unpatched (date.tsx recommends it as the safe alternative). The only `Date.now()` in the
  span lifecycle is the `SpanImpl` constructor, which runs under `startSpan` and is already escaped by this PR. That's why the original extension shipped without wrapping end() and nobody hit it.

  **Two gaps remain, neither exercised by the e2e here** (t1–t9 create spans but never call `end()`):

  1. Explicit `startTime — tracer.startSpan(name, { startTime })` (or Sentry's `startSpanManual` with a user-supplied start time). Then `end()` / `addEvent()` / `recordException()` without a timestamp take the `_startTimeProvided`
  branch → `millisToHrTime(Date.now())`, inside the render scope → fails with next-prerender-current-time. This is fixable from the tracer wrapper.
  2. `spanProcessor.onEnd` **runs synchronously** inside `end()`, still in the work-unit context. With `SimpleSpanProcessor` / Sentry's processor, export kicks off immediately and ALS propagates into the continuation, so
  `Date.now()`/`Math.random()` in exporter code (e.g. OTLP retry jitter) can still observe the prerender store. This is **not** fixable from the tracer wrapper — it's the user's exporter continuation.

**Suggested follow-up (separate PR):**
  - Have `instrumentTracerForCacheComponents` also wrap the span returned by `startSpan` and the one passed into the `startActiveSpan` callback, so `end`/`addEvent`/`recordException` run through `workUnitAsyncStorage.exit`.
  - Add a `t10` case: `startSpan(name, { startTime })` → `await` → `span.end()`, which fails without the wrap.
  - Document gap # 2 as a known limitation rather than chasing it here.

  Keeping this PR scoped to the creation-path fix; raising the above so it's tracked rather than lost.

PR #94781, opened 2026-06-12
Move `warnOnce` imports into statements guarded by `process.env.NODE_ENV === 'development'`

sampoder:
The context is this message from @icyJoseph:

```
I wonder if you've also look a little at DCE? This module, https://nextjs.org/_next/static/immutable/chunks/2uxebeysomvt2.js ships every time...
```

https://nextjs.org/_next/static/immutable/chunks/2uxebeysomvt2.js corresponds to this `warnOnce` helper: https://github.com/vercel/next.js/blob/643e34a59183f90ae9d2db9a64673781316ce786/packages/next/src/shared/lib/utils/warn-once.ts#L1

That doesn't do anything because Turbopack appears to be attempting to remove unused imports before doing dead-code removal. If the imports are moved though into the `(process.env.NODE_ENV !== 'production')` guard, this module won't get shipped into production. The patch to do this is pretty small so this PR does that, I'm going to look into whether its possible to reorder it so unused imports are looked at after dead code is removed.

### DIFF
--- a/packages/next/src/client/app-dir/link.tsx
+++ b/packages/next/src/client/app-dir/link.tsx
@@ -7,7 +7,6 @@ import { AppRouterContext } from '../../shared/lib/app-router-context.shared-run
 import { useMergedRef } from '../use-merged-ref'
 import { isAbsoluteUrl } from '../../shared/lib/utils'
 import { addBasePath } from '../add-base-path'
-import { warnOnce } from '../../shared/lib/utils/warn-once'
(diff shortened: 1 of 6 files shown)


icyJoseph:
Nice that did the trick, I see no `.next/static/chunks` files making reference to `warnOnce` in the build output. For sanity check, the browser has also one less asset to request now.

icyJoseph (approved):

lukesandberg (approved):

PR #94784, opened 2026-06-12
Make `cacheMaxMemorySize: 0` and custom cache handlers fast in dev

unstubbable:
When Cache Components is enabled, `next dev` treats a `'use cache'` value as a miss and renders as if the cache were empty whenever the read does not resolve right away. Two development configurations triggered that even for values that were already cached, so warm reloads streamed slowly instead of serving the cached value: `cacheMaxMemorySize: 0` replaced the built-in default handler with a no-op stub, so nothing was cached at all, and custom cache handlers with a slow or remote `get` did not return in time.

For the size-0 case, development now uses a real in-memory handler instead of the no-op stub, and the `'use cache'` wrapper forces a dynamic cache life (`revalidate: 0`, a 5-minute `expire`) for it, the same treatment private caches already receive, so every read serves the stale entry and re-warms a fresh one in the background. This also fixes the dev private handler, which was sized from `cacheMaxMemorySize` and so degraded to the no-op stub whenever `cacheMaxMemorySize: 0` was set.

Custom cache handlers keep their configured cache life, since their backing owns it. Instead we put a fast built-in in-memory front handler in front of the configured one through the new `TieredCacheHandler`, which serves warm reads from the front, writes through to both tiers, and reconciles the front against the backing in the background, evicting the front entry when the backing no longer has it (the handler interface has no per-key delete, so it overwrites the entry with an already-expired copy). These dev-only handlers are kept out of the registered handler set and merged in only where tag operations iterate, so `revalidateTag` still reaches them.

Everything is gated on `process.env.__NEXT_DEV_SERVER`, so production is unchanged: `cacheMaxMemorySize: 0` still caches nothing, private entries are still never persisted, and configured handlers are used directly. New development test suites cover the size-0 and custom-handler behavior.

### DIFF
--- a/packages/next/src/server/use-cache/tiered-cache-handler.ts
+++ b/packages/next/src/server/use-cache/tiered-cache-handler.ts
@@ -23,15 +23,15 @@ export type CacheReadWriteHandler = Pick<CacheHandler, 'get' | 'set'>
  * writes through.
  *
  * The handler is a per-kind singleton (the front and backing are both shared),
- * so its in-flight map can coalesce background front syncs across concurrent
- * reads of the same key.
+ * so its in-flight map can serialize background front syncs for a key across
+ * concurrent reads, running them one at a time instead of in parallel.
  */
 export function createTieredCacheHandler(
   front: CacheHandler,
   backing: CacheHandler
 ): CacheReadWriteHandler {
-  // While a reconcile or mirror is in flight for a key, concurrent reads reuse
-  // it instead of each starting another (which would hit the backing again).
+  // Holds the in-flight (or chained) background sync per key, so a sync for a
+  // key runs after any earlier one for that key rather than in parallel.
   const inFlightSyncs = new Map<string, Promise<void>>()
 
   function scheduleBackgroundSync(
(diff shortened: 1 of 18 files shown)


packages/next/src/server/use-cache/tiered-cache-handler.ts
```diff
@@ -0,0 +1,171 @@
...
+        trackBackgroundWrite(
+          workStore,
+          reconcileFrontFromBacking(
+            front,
+            backing,
+            cacheKey,
+            softTags,
+            frontEntry
+          )
+        )
```
lubieowoce: can we get two of these in parallel? should we ensure that only one is in progress per key?

packages/next/src/server/use-cache/tiered-cache-handler.ts
```diff
@@ -0,0 +1,171 @@
...
+ * it must carry at least one byte because the built-in LRU cache refuses to
+ * store size-0 entries.
+ */
+function toExpiredEntry(entry: CacheEntry): CacheEntry {
+  return {
+    ...entry,
+    expire: 0,
+    value: new ReadableStream({
+      start(controller) {
+        controller.enqueue(new Uint8Array(1))
```
lubieowoce, line 202: we have a whole byte to play with, imho we should pick a funnier number like 42

unstubbable: Are you testing whether I let my agent auto-address review feedback? 😉

lubieowoce: ~ah he figured it out, pack it up everybody~

well, i'd assume you wouldn't outsource making something *funny* to an agent, that's way too important!

packages/next/src/server/use-cache/tiered-cache-handler.ts
unstubbable: fixed

lubieowoce (approved):

PR #94786, opened 2026-06-13
fix: retry visible link prefetches on reconnect

### LINKED ISSUE #69999: Link Component does not prefetch on network recovery.

https://github.com/refirst11/reproduction-app

### To Reproduce

Sometimes unnecessary mounts are run.
And canary release e2e testing is special.

Let's say the start of the route is "/" and it goes offline in this state.
"/a", "/b", and "/c" are in the viewport at this point, so there will be a smooth transition when transitioning to them.
And it is not possible to transition to links "/d", "/e", and "/f" on pages a, b, and c at this point, but let's say the network is restored, for example in a subway or train tunnel. In that case, when transitioning to pages d, e, and f, a network request will be sent and the page will be reloaded as in the a tag.
I call this mounting on recovery in Next.js.

However, since prefetch only needs to hit when the network restarts,
I thought it would be fixed by adding the following to the dependencies in React.useEffect.

In my environment, when I tried e2e testing next.js,
pnpm test-dev test/e2e/app-dir/app-prefetch/prefetching.test.ts

It's possible that I overlooked this, but all of the test code in the describe block of prefetching.test.ts passes (even if it fails), so I decided to write this as an issue here.

When the network becomes online, link components in the viewport are prefetched.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.0.0: Sat Jul 13 00:56:26 PDT 2024; root:xnu-11215.0.165.0.4~50/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 22.8.0
  npm: 10.8.2
  Yarn: N/A
  pnpm: 9.6.0
Relevant Packages:
  next: 14.2.5 // There is a newer version (14.2.9) available, upgrade recommended! 
  eslint-config-next: 13.5.6
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.5.4
Next.js Config:
  output: N/A
```

create-next-app, Performance

next dev (local), next start (local), Vercel (Deployed), Other (Deployed)

### Additional context

Deploy is with vercel and next start.
I've been using Next.js since version 11, but even back then there was no prefetching during network restarts.

## I want to improve
```ts
// use-network.ts
import { useEffect, useState } from 'react'

export function useNetwork() {
  const isClient =
    typeof window !== 'undefined' && typeof navigator.onLine === 'boolean'

  const readOnlineStatus = isClient && window.navigator.onLine
  const [isOnline, setIsOnline] = useState(readOnlineStatus)

  useEffect(() => {
    function updateOnlineStatus() {
      setIsOnline(window.navigator.onLine)
    }

    window.addEventListener('online', updateOnlineStatus)
    window.addEventListener('offline', updateOnlineStatus)

    return () => {
      window.removeEventListener('online', updateOnlineStatus)
      window.removeEventListener('offline', updateOnlineStatus)
    }
  }, [])
  return isOnline
}
```
```ts
// link.tsx
...
    const [setIntersectionRef, isVisible, resetVisible] = useIntersection({
      rootMargin: '200px',
    })

    const isOnline = useNetwork()
    
    React.useEffect(() => {
      // in dev, we only prefetch on hover to avoid wasting resources as the prefetch will trigger compiling the page.
      if (process.env.NODE_ENV !== 'production') {
        return
      }

      if (!router) {
        return
      }

      // If we don't need to prefetch the URL, don't do prefetch.
      if (!isVisible || isOnline || !prefetchEnabled) {
        return
      }

      // Prefetch the URL.
      prefetch(
        router,
        href,
        as,
        { locale },
        {
          kind: appPrefetchKind,
        },
        isAppRouter
      )
    }, [
      as,
      href,
      isVisible,
      locale,
      prefetchEnabled,
      pagesRouter?.locale,
      router,
      isAppRouter,
      appPrefetchKind,
      isOnline,
    ])
```
## e2e test case
```ts
describe('online/offline transitions', () => {
  it('should handle transition from offline to online correctly', async () => {
    const browser = await next.browser('/static-page')
    let requests = []

    browser.on('request', (req) => {
      requests.push(new URL(req.url()).pathname)
    })

    // prefetch wait time
    await waitFor(1000)
    await browser.eval('navigator.onLine = false')

    // Link component "/" click
    // and request dashboard.
    await browser
      .elementByCss('#to-home')
      .click()
      .waitForElementByCss('#to-dashboard')

    await browser.eval('navigator.onLine = true')
    // prefetch wait time
    await waitFor(1000)
    expect(requests.filter((req) => req.includes('/dashboard')).length).toBe(0)

    await waitFor(1000)

    const before = Date.now()
    await browser
      .elementByCss('#to-dashboard')
      .click()
      .waitForElementByCss('#dashboard-layout')
    const after = Date.now()
    const timeToComplete = after - before

    // Ensure the dashboard page is prefetched
    expect(timeToComplete).toBeLessThan(20)
  })
})
```

ifer47:
### What?

Retry visible App Router link/form prefetches when the browser comes back online.

### Why?

Fixes #69999. If a visible `<Link>` attempted to prefetch while the browser was offline, the failed route or segment cache entry could remain rejected for the normal retry window. Since the link was already visible, reconnecting did not trigger a new visibility event, so the route stayed unprefetched until another user interaction or cache change happened.

### How?

- Listen for the browser `online` event in the shared link prefetch registry and reschedule every currently visible prefetchable instance.
- Treat prefetch failures that happen while `navigator.onLine === false` as immediately expired cache entries, matching the existing experimental offline behavior without requiring `experimental.useOffline`.
- Keep intentional abort/timeout errors on the existing retry path.
- Add an app-prefetch e2e regression that uses Playwright network emulation to fail a visible link prefetch offline, restore connectivity, and assert the RSC prefetch retries.

### Verification

- `pnpm --filter @next/font build`
- `pnpm --filter next build`
- `pnpm exec eslint --config eslint.cli.config.mjs packages/next/src/client/components/links.ts packages/next/src/client/components/segment-cache/cache.ts test/e2e/app-dir/app-prefetch/prefetching.test.ts`
- `NEXT_TEST_MODE=start IS_WEBPACK_TEST=1 HEADLESS=true NEXT_TEST_PREFER_OFFLINE=1 pnpm exec jest --runInBand test/e2e/app-dir/app-prefetch/prefetching.test.ts -t "should prefetch visible links after reconnecting"`
- `NEXT_TEST_MODE=start IS_WEBPACK_TEST=1 HEADLESS=true NEXT_TEST_PREFER_OFFLINE=1 NEXT_SKIP_ISOLATE=1 pnpm exec jest --runInBand test/e2e/app-dir/app-prefetch/prefetching.test.ts -t "should (not fetch again when a static page was prefetched|prefetch visible links after reconnecting)"`

### DIFF
--- a/packages/next/src/client/components/links.ts
+++ b/packages/next/src/client/components/links.ts
@@ -351,6 +351,16 @@ function rescheduleLinkPrefetch(
   }
 }
 
+function rescheduleVisibleLinksOnReconnect() {
+  for (const instance of prefetchableAndVisible) {
+    rescheduleLinkPrefetch(instance, PrefetchPriority.Default)
+  }
+}
+
+if (typeof window !== 'undefined') {
+  window.addEventListener('online', rescheduleVisibleLinksOnReconnect)
+}
+
 export function pingVisibleLinks(
   nextUrl: string | null,
   tree: FlightRouterState
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -1234,6 +1234,30 @@ function rejectSegmentCacheEntry(
   }
 }
 
+function isIntentionalAbortError(error: unknown): boolean {
+  return (
+    typeof DOMException === 'function' &&
+    error instanceof DOMException &&
+    (error.name === 'AbortError' || error.name === 'TimeoutError')
+  )
+}
+
+function shouldExpirePrefetchOnNetworkError(error: unknown): boolean {
+  if (process.env.__NEXT_USE_OFFLINE) {
+    const { checkOfflineError } =
+      require('../offline') as typeof import('../offline')
+    if (checkOfflineError(error)) {
+      return true
+    }
+  }
+
+  return (
+    typeof navigator !== 'undefined' &&
+    navigator.onLine === false &&
+    !isIntentionalAbortError(error)
+  )
+}
+
 type RouteTreeAccumulator = {
   metadataVaryPath: PageVaryPath | null
 }
@@ -1890,17 +1914,11 @@ export async function fetchRouteOnCacheMiss(
     // decoding the response. If we're offline, reject with staleAt=-1 so the
     // entry immediately expires and gets retried once the scheduler is
     // re-pinged after connectivity is restored.
-    if (process.env.__NEXT_USE_OFFLINE) {
-      const { checkOfflineError } =
-        require('../offline') as typeof import('../offline')
-      if (checkOfflineError(error)) {
-        // Unlike navigations and server actions, prefetches don't await
-        // waitForConnection — they just reject the cache entry with an
-        // immediate expiration so it gets retried once the scheduler is
-        // re-pinged after connectivity is restored.
-        rejectRouteCacheEntry(entry, -1)
-        return null
-      }
+    if (shouldExpirePrefetchOnNetworkError(error)) {
+      // Prefetches do not block on connectivity. Expire immediately so a
+      // reconnect can retry the entry.
+      rejectRouteCacheEntry(entry, -1)
+      return null
     }
     rejectRouteCacheEntry(entry, Date.now() + 10 * 1000)
     return null
@@ -2100,17 +2118,11 @@ export async function fetchSegmentsOnCacheMiss(
   } catch (error) {
     // Either the connection itself failed, or something bad happened while
     // decoding the response.
-    if (process.env.__NEXT_USE_OFFLINE) {
-      const { checkOfflineError } =
-        require('../offline') as typeof import('../offline')
-      if (checkOfflineError(error)) {
-        // Unlike navigations and server actions, prefetches don't await
-        // waitForConnection — they just reject the cache entry with an
-        // immediate expiration so it gets retried once the scheduler is
-        // re-pinged after connectivity is restored.
-        rejectRemainingSegmentsInBundle(segments, -1)
-        return null
-      }
+    if (shouldExpirePrefetchOnNetworkError(error)) {
+      // Prefetches do not block on connectivity. Expire immediately so a
+      // reconnect can retry the entry.
+      rejectRemainingSegmentsInBundle(segments, -1)
+      return null
     }
     rejectRemainingSegmentsInBundle(segments, Date.now() + 10 * 1000)
     return null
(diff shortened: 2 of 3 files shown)


refirst11:
This method is better than the one I proposed in the issue; this is a bug that causes next to remount, so I hope it will be merged immediately in the next version.

ifer47:
Thanks for confirming! Glad this approach addresses the issue. The regression test is included, and the PR is now awaiting maintainer review.

refirst11:
I filed #69999, so thank you for picking this up — it has been open for two years.

Some good news about the conflict: rebasing should shrink this PR substantially rather than cost you work. Canary has since consolidated the three duplicated `__NEXT_USE_OFFLINE` blocks you are editing into a single helper, `getPrefetchErrorStaleAt(error)` (`segment-cache/cache.ts:121`), called from the three former sites at 2077, 2132 and 2748. It returns a `staleAt` rather than a boolean, so the change becomes a few lines inside that one function instead of three near-identical edits.

The `links.ts` half is untouched by any of that, and it is still the part nothing upstream does — canary has no `online` listener and no `navigator.onLine` reference anywhere in that file. Registering it at module scope also matches how the file already keeps `prefetchableAndVisible` and the shared `IntersectionObserver`, so that reads consistent to me.

One thing worth stating explicitly before a reviewer asks: **this is a default-behaviour change, not only a bug fix.** Today the immediate-expiry path runs only under `__NEXT_USE_OFFLINE`; keying it on `navigator.onLine === false` turns it on for everyone. That may well be right, but it is the first question a maintainer will have, so it is better made explicit than left implicit in the diff.

And one correctness note. `cache.ts` uses no `AbortSignal`, `AbortController` or timeout anywhere, so this path does not produce a `TimeoutError` of its own. If one did arrive while `navigator.onLine === false`, `isIntentionalAbortError` would classify it as intentional and it would take the 10s `REJECTION_BACKOFF_MS` instead of `-1` — and then the `online` listener would reschedule the link, but the entry would not be expired, so it would not be re-fetched. That is the exact case this PR sets out to fix. `AbortError` is easier to justify, since the browser aborts in-flight requests on unload.

Putting those together, the whole `cache.ts` change after a rebase would be:

```ts
function getPrefetchErrorStaleAt(error: unknown): number {
  if (process.env.__NEXT_USE_OFFLINE) {
    const { checkOfflineError } =
      require('../offline') as typeof import('../offline')
    if (checkOfflineError(error)) {
      return -1
    }
  }

  // The browser aborts in-flight requests on unload; that is not a network
  // failure, so keep the normal backoff for it.
  const isAbort =
    typeof DOMException === 'function' &&
    error instanceof DOMException &&
    error.name === 'AbortError'

  if (
    typeof navigator !== 'undefined' &&
    navigator.onLine === false &&
    !isAbort
  ) {
    return -1
  }

  return Date.now() + REJECTION_BACKOFF_MS
}
```

That replaces the three edited blocks and both new helpers, and leaves `links.ts` and the e2e test as they are. Yours to take or leave, of course — I have not built the branch, so treat the snippet as a sketch of the shape rather than tested code.

PR #94792, opened 2026-06-13
[turbo-tasks] Shrink RawVc to 8 bytes and CellId to 4 bytes

lukesandberg:
### What?

Shrink `RawVc` (16 → **8 bytes**) and `CellId` (6 → **4 bytes**), by hand-packing them into `NonZero` integers.

### Why?

`RawVc` is the type-erased representation behind every `Vc` / `ResolvedVc` / `OperationVc`, and `CellId` keys every task cell. They live in huge numbers in the cache keys and task storage to track cells. Cutting `RawVc` in half and `CellId` by a third removes hundreds of megabytes of peak RSS on a real, large app.

### How?

| Type      | Before   | After    | Representation |
|-----------|----------|----------|----------------|
| `CellId`  | 6 bytes  | **4 bytes** | `NonZeroU32`: `ValueTypeId` in the top 10 bits, cell index in the low 22 bits |
| `RawVc`   | 16 bytes | **8 bytes** | `NonZeroU64`; bit 31 flags `LocalOutput`, and the two task variants are split by whether the high-32-bit `CellId` field is zero |

Supporting changes:

- **`TaskId` constrained to 31 bits**
- **`ValueTypeId` capped at 1023** (10 bits) and **cell index capped at ~4.19M** (22 bits), enforced at the registry and cell-allocation sites.

Together those restrictions enable us to preserve a bit to use as a discreminent in RawVc and pack `CellId` into a u32

### Perf

Building vercel-site, 5 runs each. `maxRSS` and `user CPU` are means; `wall` is the median
| Condition | Branch | maxRSS | wall (s) | user CPU (s) |
|---|---|---|---|---|
| No persistence | canary | 13.64 GiB | 41.56 | 305.64 |
| No persistence | **this PR** | **13.04 GiB** | 40.72 | 299.52 |
| | | **−4.4%** | −2.0% | −2.0% |
| Persistence | canary | 17.19 GiB | 55.20 | 466.54 |
| Persistence | **this PR** | **16.30 GiB** | 55.12 | 462.33 |
| | | **−5.2%** | −0.1% | −0.9% |

### DIFF
--- a/turbopack/crates/turbo-tasks/src/id.rs
+++ b/turbopack/crates/turbo-tasks/src/id.rs
@@ -171,16 +171,8 @@ impl Debug for TaskId {
 
 unsafe impl crate::NonLocalValue for TaskId {}
 
-/// `TaskId` values are constrained to 31 bits so the id fits alongside a
-/// [`crate::CellId`] (32 bits) inside the 64-bit packed [`crate::RawVc`], which
-/// reserves bit 31 of its word as the `LocalOutput` discriminator. Bit 30 marks
-/// transient tasks; bit 31 is always zero. The transient flag is intentionally
-/// a property of the *value* (not of any `RawVc` wrapper) because transience is
-/// computed from a bare `TaskId` in many places (storage, aggregation, child
-/// connection).
-///
-/// This splits the id space into ~1.07B persistent ids (`1..=0x3FFF_FFFF`) and
-/// ~1.07B transient ids (`0x4000_0000..=0x7FFF_FFFF`).
+/// `TaskId` values are constrained to 31 bits to preserve a niche for [`crate::RawVc`]. Bit 30
+/// marks transient tasks; bit 31 is always zero.
 pub const TRANSIENT_TASK_BIT: u32 = 0x4000_0000;
 
 /// The largest value a [`TaskId`] may hold (31 bits set).
--- a/turbopack/crates/turbo-tasks/src/id.rs
+++ b/turbopack/crates/turbo-tasks/src/id.rs
@@ -137,12 +137,17 @@ macro_rules! define_id {
 
 define_id!(
     TaskId: u32,
+    // Capped below `u32::MAX` so the id fits in 30 bits when packed into `RawVc`.
     max = TASK_ID_MAX,
     derive(Serialize, Deserialize, Encode, Decode),
     serde(transparent),
 );
+define_id!(
+    ValueTypeId: u16,
+    // Capped below `u16::MAX` so the id fits in 10 bits when packed into `CellId`.
+    max = crate::CellId::MAX_VALUE_TYPE_ID,
+);
 define_id!(FunctionId: u16);
-define_id!(ValueTypeId: u16);
 define_id!(TraitTypeId: u16);
 define_id!(
     LocalTaskId: u32,
@@ -304,5 +309,13 @@ mod tests {
         assert!(TaskId::MAX.is_transient());
         assert!(!TaskId::new(TRANSIENT_TASK_BIT - 1).unwrap().is_transient());
         assert!(TaskId::new(TRANSIENT_TASK_BIT).unwrap().is_transient());
+    }
+
+    /// `ValueTypeId::MAX` is capped to the 10-bit field it occupies in `CellId`,
+    /// not the full `u16::MAX`.
+    #[test]
+    fn value_type_id_max_is_capped() {
+        assert_eq!(*ValueTypeId::MAX, crate::CellId::MAX_VALUE_TYPE_ID);
+        assert!(*ValueTypeId::MAX < u16::MAX);
     }
 }
--- a/turbopack/crates/turbo-tasks/src/id.rs
+++ b/turbopack/crates/turbo-tasks/src/id.rs
@@ -137,7 +137,7 @@ macro_rules! define_id {
 
 define_id!(
     TaskId: u32,
-    // Capped below `u32::MAX` so the id fits in 30 bits when packed into `RawVc`.
+    // Capped below `u32::MAX` so the id fits in 31 bits when packed into `RawVc`.
     max = TASK_ID_MAX,
     derive(Serialize, Deserialize, Encode, Decode),
     serde(transparent),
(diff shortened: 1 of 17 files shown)


lukesandberg:
* **#94792** <a href="https://app.graphite.com/github/pr/vercel/next.js/94792?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94792?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/id.rs
```diff
@@ -263,3 +295,28 @@ make_registered_serializable!(
...
+    }
+
+    /// `ValueTypeId::MAX` is capped to the 10-bit field it occupies in `CellId`,
+    /// not the full `u16::MAX`.
+    #[test]
+    fn value_type_id_max_is_capped() {
+        assert_eq!(*ValueTypeId::MAX, crate::CellId::MAX_VALUE_TYPE_ID);
+        assert!(*ValueTypeId::MAX < u16::MAX);
+    }
+}
```
lukesandberg: these tests are not useful

turbopack/crates/turbo-tasks/src/registry/mod.rs
```diff
@@ -118,6 +118,13 @@ trait Registerable: 'static + Eq + std::hash::Hash {
     type Id: Copy + From<NonZeroU16> + std::ops::Deref<Target = u16> + std::fmt::Display;
     const TYPE_NAME: &'static str;
 
+    /// The largest id that may be assigned to this registry item. Ids are
+    /// assigned sequentially from 1, so this also bounds the number of items.
+    ///
+    /// `ValueType` is capped tighter than `u16::MAX` because [`ValueTypeId`] is
+    /// packed into 10 bits of [`crate::CellId`].
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks/src/vc/raw.rs
```diff
@@ -25,15 +26,100 @@ use crate::{
     turbo_tasks,
 };
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize, Encode, Decode)]
-pub struct CellId {
-    pub type_id: ValueTypeId,
-    pub index: u32,
+/// Identifies a specific cell within a task: a [`ValueTypeId`] paired with a
```
lukesandberg, line 29: simplify comment

turbopack/crates/turbo-tasks/src/vc/raw.rs
```diff
@@ -45,10 +131,41 @@ impl Display for CellId {
...
+/// bit31 = 1                  LocalOutput: 1<<31 | transient(1) | ExecutionId(16) << 1 | LocalTaskId(32) << 32
+/// bit31 = 0, bits32..63 == 0 TaskOutput:  TaskId(31)
+/// bit31 = 0, bits32..63 != 0 TaskCell:    TaskId(31) | CellId(32) << 32
+/// ```
+///
+/// The `TaskId` occupies bits `0..=30` (31 bits) unshifted, so a task variant's
+/// word equals its `TaskId` value (plus the cell field for `TaskCell`). The word
+/// is never zero: `TaskOutput`/`TaskCell` carry a `TaskId >= 1`, and
+/// `LocalOutput` always has bit 31 set.
+///
```
lukesandberg: simplify

turbopack/crates/turbo-tasks/src/registry/mod.rs
```diff
@@ -134,6 +137,7 @@ impl Registerable for NativeFunction {
 impl Registerable for ValueType {
     type Id = ValueTypeId;
     const TYPE_NAME: &'static str = "Value";
+    const MAX_ID: u16 = crate::CellId::MAX_VALUE_TYPE_ID;
```
bgw: I know that the max is defined by the space allocated in `CellId`, but I think it's a bit easier to understand in this context if you read the re-export on `ValueTypeId`.

```suggestion
    const MAX_ID: ValueTypeId = crate::ValueTypeId::MAX;
```

turbopack/crates/turbo-tasks/src/registry/mod.rs
```diff
@@ -164,6 +168,12 @@ fn init_registry<T: Registerable>(mut items: Vec<&'static T>) -> Box<[&'static T
             );
         }
         prev_name = Some(global_name);
+        assert!(
+            u16::from(id) <= T::MAX_ID,
+            "too many {ty} items registered: id {id} exceeds the cap of {max}",
+            ty = T::TYPE_NAME,
+            max = T::MAX_ID,
+        );
```
bgw, line 176: I guess the reason we need this here instead of using the check built into the id factory is that we want to avoid atomic integer operations when we haven't created any threads yet?

turbopack/crates/turbo-tasks/src/vc/raw.rs
```diff
@@ -69,69 +168,232 @@ pub enum RawVc {
     LocalOutput(ExecutionId, LocalTaskId, TaskPersistence),
 }
 
+/// Bit 31 discriminates `LocalOutput` (set) from the task variants (clear).
+/// It is free for the task variants because a `TaskId` is only 31 bits
+/// (`bits 0..=30`) and the `CellId` lives in the high 32 bits (`32..=63`).
+const RAW_VC_LOCAL_FLAG: u64 = 1 << 31;
+
+/// Mask of the `TaskId` value inside `TaskOutput` / `TaskCell` (bits `0..=30`).
+const RAW_VC_TASK_MASK: u64 = TASK_ID_MAX as u64;
```
bgw, line 180: I think you can use the `from` conversion here, which makes it clearer that there's no overflow issues.

```suggestion
const RAW_VC_TASK_MASK: u64 = u64::from(TASK_ID_MAX);
```

turbopack/crates/turbo-tasks/src/vc/raw.rs
```diff
@@ -541,3 +795,153 @@ impl Future for ReadRawVcFuture {
 }
 
 impl Unpin for ReadRawVcFuture {}
+
+#[cfg(test)]
+mod tests {
```
bgw, line 803: Maybe some tests that check that we panic if we try to create a RawVc that exceeds our bounds checks?

bgw (approved):
Very nice win!

turbopack/crates/turbo-tasks/src/vc/raw.rs
lukesandberg: runs into 'trait call in const position' problems

```
cannot call conditionally-const associated function `<u64 as From<u32>>::from` in constants
calls in constants are limited to constant functions, tuple structs and tuple variants
see issue #143874 <https://github.com/rust-lang/rust/issues/143874> for more information
```

turbopack/crates/turbo-tasks/src/registry/mod.rs
lukesandberg: the id-factory is used for `TaskIds` which are dynamically allocated.

These registries are for the 1 time static registration (function value and trait ids). so they are distinct usecases.

turbopack/crates/turbo-tasks/src/vc/raw.rs
lukesandberg: clarified in the comment

PR #94794, opened 2026-06-13
Fix sandbox timeout memory leak causing stepwise heap growth

### LINKED ISSUE #95094: Memory leak from retained sandbox timeouts causes stepwise heap growth

https://github.com/berry95/next-timeout-resource-leak-repro

### To Reproduce

2. Install dependencies with `pnpm install`.
3. Build the standalone server with `pnpm build`.
4. Start the standalone server with `pnpm start:standalone`.
5. In another terminal, run `pnpm load`.
6. Observe the reported server heap usage over repeated requests.

The reproduction sends requests to `/api/heap?timeouts=500`. The Pages Router
`middleware.ts` schedules one-shot `setTimeout(() => {}, 0)` callbacks before
the request reaches the API route. The timeout ids are not retained by user code
and `clearTimeout()` is not called, matching common fire-and-forget timeout
usage.

The `timeouts` query parameter intentionally amplifies the issue so retained
sandbox timeout resources are easier to observe locally.

Current behavior:

Naturally completed one-shot timeouts are cleared at the underlying Node timer
level, but their numeric ids remain tracked by Next.js's sandbox
`TimeoutsManager`. The ids are removed only if user code explicitly calls
`clearTimeout(id)`, or when the whole sandbox module context is cleared.

In a long-lived standalone/self-hosted server process, repeated middleware
requests can therefore make the tracked timeout resources grow monotonically.
The production symptom is step-like heap growth over time and, under sustained
traffic, possible OOM.

Expected behavior:

After a one-shot timeout callback runs, the timeout id should be released from
Next.js's sandbox resource manager even when user code does not explicitly call
`clearTimeout(id)`.

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows
Binaries:
  Node: 24.14.1
  npm: 11.11.0
  pnpm: 10.33.0
Relevant Packages:
  next: 16.3.0-canary.50
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
Next.js Config:
  output: standalone
```

Middleware, Pages Router, Runtime, Output, Performance

### Additional context

The sandbox resource managers were introduced in #57235 to avoid retaining old
Edge runtime VM contexts during dev HMR by taking ownership of timers and
clearing them when a module context is invalidated. That teardown path is still
needed.

The missing path is natural completion while the module context remains alive:
one-shot timeout callbacks can finish successfully without releasing their ids
from the sandbox manager.

The existing `clearTimeout(timeout)` inside `webSetTimeoutPolyfill` works around
the Node.js primitive timer leak described in nodejs/node#53335, but it does not
remove the id from Next.js's own resource manager.

berry95:
Fixes #95094

## What?

Fix a sandbox timer memory leak where naturally completed one-shot timeouts stay
tracked by Next.js after their callback runs.

In long-lived standalone/self-hosted processes, this can accumulate timeout ids
across middleware/proxy requests, producing step-like heap growth over time and
eventually contributing to OOM under sustained traffic.

Reproduction: https://github.com/berry95/next-timeout-resource-leak-repro

This changes the sandbox `ResourceManager` backing store from an array to a
`Set`, and makes `TimeoutsManager` remove a timeout from the tracked resource
set in a `finally` block after the timeout callback completes.

The existing `clearTimeout(timeout)` call inside `webSetTimeoutPolyfill` is kept.
That call works around the Node.js timer primitive leak described in
nodejs/node#53335. This PR fixes the separate Next.js resource-manager tracking
leak.

Background: the sandbox resource managers were introduced in #57235 to avoid
retaining old Edge runtime VM contexts during dev HMR by taking ownership of
timers and clearing them when a module context is invalidated. That covered
module-context teardown, but one-shot timeouts that complete naturally also need
to release themselves from the manager while the module context remains alive.

## Why?

Next.js installs sandbox-specific timer polyfills for middleware and edge
function execution:

1. `context.setTimeout` calls `timeoutsManager.add([context, ...args])`.
2. `ResourceManager.add()` stores the returned timeout id so it can be cleaned up
   when the sandbox module context is cleared.
3. A one-shot timeout may complete naturally without user code ever calling
   `clearTimeout(id)`.
4. Before this change, natural completion cleared the underlying Node timer but
   did not remove the timeout id from `TimeoutsManager`'s tracked resources.

In a long-lived sandbox module context, such as a self-hosted/standalone server
running middleware in a container, each naturally completed one-shot timeout
could leave its id behind until the module context was cleared or the process
restarted. Middleware code and its dependencies commonly treat `setTimeout` as a
fire-and-forget one-shot primitive, so they often do not keep the id or call
`clearTimeout()`.

Under sustained traffic this can make the tracked timeout resources grow
monotonically. The production symptom is step-like heap growth over time, and in
high-throughput long-lived processes this can contribute to an eventual OOM.

The affected path is the web sandbox used by middleware/proxy and edge function
execution. This was observed with Pages Router middleware in a standalone
self-hosted deployment, but the underlying issue is not Pages Router-specific:
any sandboxed runtime code that creates one-shot timeouts without explicitly
clearing them can hit the same tracking behavior.

## How?

- Use `Set<T>` for tracked resources because the manager tracks a live resource
  set, not an ordered list.
- Add `ResourceManager.release()` to remove a resource without destroying it.
- Wrap timeout callbacks in `TimeoutsManager.create()` so naturally completed
  one-shot timeouts release their id after callback execution.
- Preserve explicit `clearTimeout()` behavior through `TimeoutsManager.remove()`.
- Preserve timeout callback `this` binding and variadic callback arguments.

## Tests

- `pnpm prettier --check packages/next/src/server/web/sandbox/resource-managers.ts packages/next/src/server/web/sandbox/resource-managers.test.ts`
- `pnpm jest packages/next/src/server/web/sandbox/resource-managers.test.ts --runInBand`
- `pnpm --filter next typescript`
- `pnpm --filter next build`

The regression test verifies that completed one-shot timeouts are no longer
retained by checking that `removeAll()` does not clear already completed
timeouts again. This fails with the previous implementation because completed
timeout ids remain in the resource manager.

### DIFF
--- a/packages/next/src/server/web/sandbox/resource-managers.test.ts
+++ b/packages/next/src/server/web/sandbox/resource-managers.test.ts
@@ -1,0 +1,70 @@
+import { timeoutsManager } from './resource-managers'
+
+describe('sandbox resource managers', () => {
+  afterEach(() => {
+    timeoutsManager.removeAll()
+  })
+
+  it('does not retain one-shot timeouts after they run', async () => {
+    const context = {}
+    const clearTimeoutSpy = jest.spyOn(global, 'clearTimeout')
+
+    try {
+      await Promise.all(
+        [0, 1, 2].map(
+          () =>
+            new Promise<void>((resolve) => {
+              timeoutsManager.add([context, resolve, 0] as any)
+            })
+        )
+      )
+
+      expect(clearTimeoutSpy).toHaveBeenCalledTimes(3)
+
+      timeoutsManager.removeAll()
+
+      expect(clearTimeoutSpy).toHaveBeenCalledTimes(3)
+    } finally {
+      clearTimeoutSpy.mockRestore()
+    }
+  })
+
+  it('clears timeouts when they are removed explicitly', () => {
+    const context = {}
+    const clearTimeoutSpy = jest.spyOn(global, 'clearTimeout')
+    const timeout = timeoutsManager.add([context, () => {}, 1000] as any)
+
+    try {
+      timeoutsManager.remove(timeout)
+
+      expect(clearTimeoutSpy).toHaveBeenCalledTimes(1)
+      expect(clearTimeoutSpy).toHaveBeenCalledWith(timeout)
+    } finally {
+      clearTimeoutSpy.mockRestore()
+    }
+  })
+
+  it('preserves timeout callback arguments that are functions', async () => {
+    const context = {}
+    const callbackArg = () => 'value'
+    const callback = jest.fn(function (this: unknown, arg: () => string) {
+      expect(this).toBe(context)
+      expect(arg).toBe(callbackArg)
+      expect(arg()).toBe('value')
+    })
+
+    await new Promise<void>((resolve) => {
+      timeoutsManager.add([
+        context,
+        function (this: unknown, ...args: Parameters<typeof callback>) {
+          callback.apply(this, args)
+          resolve()
+        },
+        0,
+        callbackArg,
+      ] as any)
+    })
+
+    expect(callback).toHaveBeenCalledTimes(1)
+  })
+})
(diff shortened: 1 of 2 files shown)


gaearon:
Thank you. Merged via https://github.com/vercel/next.js/pull/96161 (I cherry picked your commit there).

PR #94797, opened 2026-06-13
[test] Recover from a leftover build process on test retry

unstubbable:
When a `next start` test exceeds the per-test timeout while `next build` is still running, the build process is left running and tracked by the next instance, because the build phase of `NextStartInstance.start()` clears `childProcess` only when the process exits. `jest.retryTimes` then re-runs the test body in the same process, and `start()` immediately throws `next already started`, which masks the original timeout.

`NextStartInstance` now tracks whether a build or a server currently owns the child process. When `start()` finds a leftover `next build` process from an interrupted attempt, it stops that process and continues instead of throwing, so the retry can start from a clean state. The `next already started` guard is preserved for the case where a server is genuinely already running.

### DIFF
--- a/test/lib/next-modes/next-start.ts
+++ b/test/lib/next-modes/next-start.ts
@@ -280,6 +280,7 @@ export class NextStartInstance extends NextInstance {
       exitCode: NodeJS.Signals | number | null
       cliOutput: string
     }>((resolve) => {
+      this._phase = 'building'
       const curOutput = this._cliOutput.length
       const spawnOpts = this.getSpawnOpts(options.env)
       const buildArgs = this.getBuildArgs(options.args)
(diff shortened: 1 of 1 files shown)


test/lib/next-modes/next-start.ts
```diff
@@ -88,6 +105,7 @@ export class NextStartInstance extends NextInstance {
     }
 
     if (!options.skipBuild) {
+      this._phase = 'building'
```
eps1lon, line 108: Should this also be set when running the public `next.build()`? I checked where `getBuildArgs` is called and noticed we're not setting the phase there.

unstubbable: Yeah, that makes sense, I've added it.

eps1lon (approved):

PR #94808, opened 2026-06-15
Omit unused root params from App Shell cache key

acdlite:
Addresses an oversight from #94773. Not a bugfix, just a missed optimization.

We've decided to allow root params to be accessed in the App Shell. In a previous PR, I updated the client to include root params in the cache key. However, I missed an optimization: when writing shell entries to the cache, we should not assume that the shell varies on *all* root params — only the ones that the server reports as actually being accessed.

Note: When this PR was first opened, the server-side change to allow root params in the shell hadn't landed yet. It has since landed in #94738, so this optimization is observable end-to-end (the included e2e test exercises it).

### DIFF
--- a/test/e2e/app-dir/segment-cache/app-shell-root-params/app-shell-root-params.test.ts
+++ b/test/e2e/app-dir/segment-cache/app-shell-root-params/app-shell-root-params.test.ts
@@ -1,0 +1,74 @@
+import { nextTestSetup } from 'e2e-utils'
+import type * as Playwright from 'playwright'
+import { createRouterAct } from 'router-act'
+
+// The App Shell is keyed by root params so that a locale-aware (etc.) shell can
+// still be reused across navigations. But a shell should only vary on the root
+// params it actually reads — keying it on root params it never touches stores
+// redundant copies and misses cache hits.
+//
+// This fixture's root layout reads only `lang`; `region` is a root param that
+// is read only below the shell boundary. So the shell for
+// /[lang]/[region]/posts varies on `lang` but is identical across `region`.
+//
+// The app uses Partial Prefetching, so revealing a link prefetches only the
+// shared shell — no per-link Speculative prefetch. Each reveal is therefore
+// exactly one shell prefetch (when the shell isn't already cached) or no
+// requests at all. We pass `{ includeAppShellRequests: true }` so router-act
+// asserts on App Shell requests (next-router-prefetch: '3') instead of ignoring
+// them.
+describe('App Shell varies only on the root params it reads', () => {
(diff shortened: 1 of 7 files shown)


acdlite:
Failures appear to be related to https://next-data-api-endpoint.vercel.app being down

test/e2e/app-dir/segment-cache/app-shell-root-params/app/[lang]/[region]/layout.tsx
```diff
@@ -0,0 +1,35 @@
...
+    </html>
+  )
+}
+
+// Only one combo is statically generated — just enough to give the app a
+// buildable start page (/en/us). The combos the test probes are deliberately
+// NOT listed here, so their shells are rendered at runtime (FetchStrategy.
+// RuntimeShell). A statically-prerendered shell would be cached via a different
+// code path that already narrows by the server-reported vary params, masking
+// the runtime-shell keying we're exercising here.
```
Comment (lubieowoce) [inline test/e2e/app-dir/segment-cache/app-shell-root-params/app/[lang]/[region]/layout.tsx]: i'm confused by this comment. the shells aren't static because `id` isn't a static param, so `/posts/[id]` isn't fully static. this should have no relation to what root params we render here?

test/e2e/app-dir/segment-cache/app-shell-root-params/app-shell-root-params.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  if (isNextDev) {
+    it('is skipped in dev (prefetching is disabled)', () => {})
+    return
+  }
+
+  it('reuses the shell across an unread root param, but refetches across a read one', async () => {
+    let page: Playwright.Page
+    // Start on the one statically-generated route. The combos probed below are
+    // NOT statically generated, so their shells are rendered at runtime
+    // (FetchStrategy.RuntimeShell).
```
lubieowoce: i'm confused, i thought `RuntimeShell` just means that the router will do `next-router-prefetch: 3`, and whether or not that's static depends on the server?

test/e2e/app-dir/segment-cache/app-shell-root-params/app/[lang]/[region]/layout.tsx
Comment (acdlite) [reply test/e2e/app-dir/segment-cache/app-shell-root-params/app/[lang]/[region]/layout.tsx]: I think Claude was just confused, I saw this too and forgot to delete before submitting

test/e2e/app-dir/segment-cache/app-shell-root-params/app-shell-root-params.test.ts
acdlite: same here

lubieowoce (approved):

PR #94809, opened 2026-06-15
Support vary params rewinding

acdlite:
Refactors how vary params are serialized in the Flight stream so that it's compatible with stage rewinding.

The immediate motivation is preparation for allowing root params to be accessed in the App Shell, but this new approach will work more generally for any other kind of rewinding we add in the future.

The design is similar to how stale time is already serialized: each access of a param is written to an AsyncIterable that is embedded into the Flight stream. When decoding the params, the client first buffers the entire stream, then synchronously drains the iterable using the special `status` and `value` fields used by the React protocol, until it reaches a suspended entry.

This is compatible with rewinding because the stream can be cut off at any point and the client can still decode all the param accesses up to that point.

As of this PR, this is a pure refactor: it won't change any observable behavior until the server starts allowing root params in the shell.

### DIFF
--- a/packages/next/src/server/app-render/vary-params.ts
+++ b/packages/next/src/server/app-render/vary-params.ts
@@ -1,37 +1,87 @@
 import type { Params } from '../request/params'
 import type { SearchParams } from '../request/search-params'
-import { workUnitAsyncStorage } from './work-unit-async-storage.external'
-import type {
-  VaryParamsThenable,
-  VaryParams,
-} from '../../shared/lib/segment-cache/vary-params-decoding'
+import {
+  getVaryParamsAccumulator,
+  workUnitAsyncStorage,
+} from './work-unit-async-storage.external'
 
 /**
  * Accumulates vary params for a single segment (or for metadata/rootParams).
  *
- * VaryParamsAccumulator is also a thenable that can be serialized by React
- * Flight. The accumulator starts as 'pending' and accumulates param accesses
- * during render. Call `finishTrackingVaryParams()` after rendering to resolve
- * all accumulators.
+ * A VaryParamsAccumulator is an `AsyncIterable<string>` that can be serialized
+ * by React Flight. As params are accessed during render, each newly-seen param
+ * name is `add`ed, which yields it into the Flight stream immediately. After
+ * rendering, call `close()` (via `finishAccumulatingVaryParams`) to end the
+ * iteration.
  *
- * The `status` and `value` fields follow the React Flight thenable protocol:
- * when `status === 'fulfilled'`, Flight can read `value` synchronously without
- * scheduling a microtask via `.then()`.
+ * Because each access is flushed into the stream as it happens, the set of
+ * accessed params is built up incrementally, with no step at the end of the
+ * render that has to run for the client to read anything. If the prerender is
+ * aborted by sync I/O, the params yielded before the abort are already in the
+ * stream — and they're exactly the params the partial response depends on.
+ * This mirrors how `StaleTimeIterable` works (see stale-time.ts).
+ *
+ * Each name is emitted at most once: `add` dedupes against the set of
+ * already-yielded names, so the stream never contains a duplicate.
+ *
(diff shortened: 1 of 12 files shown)


packages/next/src/server/app-render/vary-params.ts
```diff
@@ -196,16 +228,16 @@ export function getRootParamsVaryParamsAccumulator(): VaryParamsAccumulator | nu
   if (workUnitStore) {
     switch (workUnitStore.type) {
       case 'prerender':
-      case 'prerender-runtime': {
+      case 'prerender-runtime':
+      case 'request': {
```
lubieowoce: why didn't we have this for 'request' before? was that just a bug?

packages/next/src/server/app-render/vary-params.ts
```diff
@@ -1,37 +1,104 @@
...
-  ): void
+  // Resolves once Flight has pulled this iterable's terminating `done` result —
+  // i.e. it has fully consumed the iterable and serialized its rows. The
+  // prerender awaits this (see finishAccumulatingVaryParams) to know the vary
+  // params have finished flushing before it decides whether the stream is still
+  // pending on dynamic data.
+  private _flushedResolve: (() => void) | null = null
+  readonly flushed: Promise<void> = new Promise((resolve) => {
+    this._flushedResolve = resolve
+  })
```
lubieowoce: we're not doing anything like this in stale-time.ts, why do we need it here? the `await waitAtLeastOneReactRenderTask()` we do in prerenders should this anyway (or, in cached navs, a separate sequential task, but those two end up basically equivalent). unless i'm missing something?

also the comment seems misleading, we also do `_markFlushed()` when we `close()` the iterable manually, so it seems a little meaningless as a proxy for "flight pulled this". we don't actually know if flight observed our `close()`

lubieowoce: also i haven't quite worked through this but i suspect this might hang if sync IO aborts the render.

`abortIterable` doesn't seem to interact with the iterator at all unless it supports `throw()`
https://github.com/facebook/react/blob/c0cd4d5d30f3ad56380c14ac53c6f50d96e38814/packages/react-server/src/ReactFlightServer.js#L1367-L1389, so i think`flushed` would just stay unresolved?

edit: actually, we *do* close these manually in case of sync IO, which i suppose would cover this. but that was only added as a workaround because flight was hanging on them (despite the abort), so i don't think we should rely on it

packages/next/src/shared/lib/app-router-types.ts
```diff
@@ -407,8 +410,16 @@ export type InitialRSCPayload = {
...
   /**
    * headVaryParams - vary params for the head (metadata) of the response.
+   * Does not include root params (see `r`).
+   */
+  h: VaryParamsIterable | null
+  /**
+   * rootVaryParams - the root params accessed anywhere in the response, emitted
+   * once. The client unions these into the head and every segment's vary
+   * params, rather than the server folding them into each set. Only present for
+   * prerenders (root params aren't tracked for plain request renders).
```
lubieowoce: hmm, shouldn't they be (for recovering shells in cached navs)? this seems like a TODO

lubieowoce:
* **#94809** <a href="https://app.graphite.com/github/pr/vercel/next.js/94809?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94809?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94857** <a href="https://app.graphite.com/github/pr/vercel/next.js/94857?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/vary-params.ts
lubieowoce: i guess it was

lubieowoce (approved):

PR #94818, opened 2026-06-15
docs: tighten Partial Prefetching API references and adoption guide

aurorascharff:
### What?

Three friction points found while running through the Partial Prefetching migration on a Cache Components app:

- **`prefetch.mdx`** — added `'partial'`. It's a real value in the framework but the docs page didn't list it.
- **`link.mdx`** — rewrote the `prefetch={true}` bullet so it's correct on its own under Partial Prefetching, not only inside the callout below.
- **`adopting-partial-prefetching.mdx`** — restored the "Going further" section and added a before/after migration example.

### Related

[#94798](https://github.com/vercel/next.js/pull/94798) — Insight surface for the new `<Link prefetch={true}>` warning. Lands independently.

### DIFF
--- a/docs/01-app/02-guides/adopting-partial-prefetching.mdx
+++ b/docs/01-app/02-guides/adopting-partial-prefetching.mdx
@@ -90,7 +90,7 @@ export default nextConfig
 
 ## Auditing existing `<Link prefetch={true}>` calls
 
-When you flip `partialPrefetching: true`, every existing `<Link prefetch={true}>` switches semantics. Before the flag, it prefetched the destination's cached page render and its dynamic data. After the flag, it prefetches the App Shell plus the destination's cached content — never dynamic data. Walk through each existing usage and decide whether the prop still earns its keep.
+When you flip `partialPrefetching: true`, every existing `<Link prefetch={true}>` switches semantics. Before the flag, it prefetched the destination's cached page render and its dynamic data. After the flag, it prefetches the App Shell plus the destination's cached content, without the dynamic data. Walk through each existing usage and decide whether the prop is still adding value.
 
 | Destination                                                                                 | Recommendation                                                                                                                               |
 | ------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
--- a/docs/01-app/02-guides/adopting-partial-prefetching.mdx
+++ b/docs/01-app/02-guides/adopting-partial-prefetching.mdx
@@ -88,6 +88,32 @@ const nextConfig: NextConfig = {
 export default nextConfig
 ```
 
+## Auditing existing `<Link prefetch={true}>` calls
+
+When you flip `partialPrefetching: true`, every existing `<Link prefetch={true}>` switches semantics. Before the flag, it prefetched the destination's cached page render and its dynamic data. After the flag, it prefetches the App Shell plus the destination's cached content — never dynamic data. Walk through each existing usage and decide whether the prop still earns its keep.
+
+| Destination                                                                                 | Recommendation                                                                                                                               |
+| ------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
+| Fully static (no `"use cache"`, no Suspense)                                                | Drop `prefetch={true}`. The default `<Link>` already loads the full page because the App Shell is the page.                                  |
+| Has cached content worth shipping ahead of the click                                        | Keep `prefetch={true}`. The cached parts come with the App Shell.                                                                            |
+| Reads request data (`cookies()`, `headers()`, `searchParams`) the user will need on arrival | Keep `prefetch={true}` and opt the route into [runtime prefetching](/docs/app/guides/runtime-prefetching) with `prefetch = 'allow-runtime'`. |
+
+Before:
+
+```tsx filename="app/nav.tsx"
+<Link href="/about" prefetch={true}>About</Link>           {/* static page */}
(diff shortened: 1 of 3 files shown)


docs/01-app/03-api-reference/03-file-conventions/02-route-segment-config/prefetch.mdx
```diff
@@ -54,6 +54,16 @@ Allows Next.js to prefetch this segment at runtime. Today, setting this option c
 export const prefetch = 'allow-runtime'
 ```
 
+### `'partial'`
+
+Opts the segment into [Partial Prefetching](/docs/app/guides/adopting-partial-prefetching) without flipping the global [`partialPrefetching`](/docs/app/api-reference/config/next-config-js/partialPrefetching) flag. A `<Link>` pointing at a segment with `prefetch = 'partial'` loads the per-route [App Shell](/docs/app/glossary#app-shell) instead of the legacy full prefetch. Set this on the destination, not the link.
+
+Use this for incremental adoption when you can't enable `partialPrefetching` for the entire app at once. Once every route in scope has `prefetch = 'partial'`, flip the global flag and remove the per-route exports.
```
icyJoseph: Nit.
```suggestion
Opts the segment into [Partial Prefetching](/docs/app/guides/adopting-partial-prefetching) without enabling the global [`partialPrefetching`](/docs/app/api-reference/config/next-config-js/partialPrefetching) flag. A `<Link>` pointing at a segment with `prefetch = 'partial'` loads the per-route [App Shell](/docs/app/glossary#app-shell) instead of the legacy full prefetch. Set this on the destination, not the link.

Use this for incremental adoption when you can't enable `partialPrefetching` for the entire app at once. Once every route in scope has `prefetch = 'partial'`, enable the global flag and remove the per-route exports.
```

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -88,6 +88,32 @@ const nextConfig: NextConfig = {
 export default nextConfig
 ```
 
+## Auditing existing `<Link prefetch={true}>` calls
+
+When you flip `partialPrefetching: true`, every existing `<Link prefetch={true}>` switches semantics. Before the flag, it prefetched the destination's cached page render and its dynamic data. After the flag, it prefetches the App Shell plus the destination's cached content, without the dynamic data. Walk through each existing usage and decide whether the prop is still adding value.
```
icyJoseph:
```suggestion
When `partialPrefetching` is enabled, `<Link prefetch={true}>` now prefetches the App Shell plus cached content, but not dynamic data. Review each existing usage and decide whether the prop is still providing value.

Previously, `<Link prefetch={true}>` also prefetched dynamic data.
```

And then the table. 

This way we go from:
- `something changed` -> `previous` -> `new` -> `action` to
- `new behavior` -> `reminder of previous behavior` -> `action`

I think the deeper problem here is that, we are not addressing why anyone had this prop to begin with - to try and load an entire route ahead of time, to try and beat any potential blocking nav - I'd almost want to add that to the `Previously...` line but it feels like it is too much

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -88,6 +88,32 @@ const nextConfig: NextConfig = {
...
+
+| Destination                                                                                 | Recommendation                                                                                                                               |
+| ------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------- |
+| Fully static (no `"use cache"`, no Suspense)                                                | Drop `prefetch={true}`. The default `<Link>` already loads the full page because the App Shell is the page.                                  |
+| Has cached content worth shipping ahead of the click                                        | Keep `prefetch={true}`. The cached parts come with the App Shell.                                                                            |
```
icyJoseph, line 100: Still not sure we are formulating this the right way. Let's make sure this is describing the behavior correctly, it might be that in trying to shorten the description, some nuance gets dropped

docs/01-app/03-api-reference/02-components/link.mdx
```diff
@@ -300,10 +300,10 @@ Prefetching happens when a `<Link />` component enters the user's viewport (init
 The following values can be passed to the `prefetch` prop:
 
 - **`"auto"` or `null` (default)**: Prefetch behavior depends on whether the route is static or dynamic. For static routes, the full route will be prefetched (including all its data). For dynamic routes, the partial route down to the nearest segment with a [`loading.js`](/docs/app/api-reference/file-conventions/loading#instant-loading-states) boundary will be prefetched.
-- `true`: The full route will be prefetched for both static and dynamic routes.
+- `true`: Without [Partial Prefetching](/docs/app/guides/adopting-partial-prefetching), the full route is prefetched for both static and dynamic routes. Under [Partial Prefetching](/docs/app/guides/adopting-partial-prefetching), the destination's cached content is prefetched along with the [App Shell](/docs/app/glossary#app-shell). Dynamic content is not included.
```
icyJoseph:
```suggestion
`true`: The full route is prefetched for both static and dynamic routes. With [Partial Prefetching](/docs/app/guides/adopting-partial-prefetching) enabled, the prefetch is the [App Shell](/docs/app/glossary#app-shell) plus the route's cached content; dynamic data is excluded.
```

icyJoseph (approved):

icyJoseph:
There's still a handful of corrections and alignment, that'll be done in https://github.com/vercel/next.js/pull/94855

Issue #418, opened 2016-12-17
Routing questions - params patterns like:   "/profile/:id/edit"  and server db -how to pass data to react components?


Ajar-Ajar:
hi, I know this issue was discussed before but I wasn't able to find a clear answer to this pattern.
First - correct me if I'm wrong, if you don't implement a custom server (http or express) the Link component supports only query string parameters like "/profile?id=123&action=edit"  am I correct?
To be able to work with a url params pattern like "/profile/:id" you need to implement a custom server... this was the best I was able to guess going through the examples...
If so - how would you implement a pattern like "/profile/:id/edit" 
In the 'parameterized-routing' example the index.js is using links like this one:
`<Link href='/blog?ip=first' as='/blog/first'><a>My first blog post</a></Link>`

and I wasn't able to decipher what is the necessary workflow on server.js
server.js declares `const match = route('/blog/:id')`

and then it is unclear what exactly goes to the address bar, and what is the data that is being pushed to the react component? 
How do I handle routes like "/profile/:id/edit"
How do I handle many routes in a generic way (coming from react-router I was looking for the match utility equivalent...)
Also, if I have db data fetching before rendering - how can I pass the fetched data to the react component.
All of these are a bit obscure and I would seriously appreciate your help as I was trying to guess my way through with no success.

Thank you
Ajar

arunoda:
This is a set of good questions.
We'll add more info on the docs.

For now, think like this:

* Next.js only knows how to serve pages and data should be pass via params. (Like: `/blog?id=first`)
* With #310 we allow a way to customize how it looks to the outside world. (Like: `blog/:id` [See](https://github.com/zeit/next.js/blob/master/examples/custom-server-express/server.js#L11)
* And when you are doing client side routing, client side app don't know anything about custom urls.
* So, you need use Link as: `<Link href='/blog?id=first' as='/blog/first' />`
  * href: this is the actual url of the page
  * as: this is how it should looks like in the browser (typically, it should match with your custom URL pattern)

Ajar-Ajar:
Thank you @arunoda  !
Simple & clear.

arunoda:
Awesome. gonna close this issue.

PR #94819, opened 2026-06-15
Repro: PPR shell + resume emit duplicate Suspense segment ids (B:/S:)

gaojude:
## What

A minimal **failing** e2e reproducing a PPR (`cacheComponents`) bug: the served document contains **duplicate Suspense completion ids** — the same `<template id="B:n">` and `<div hidden id="S:n">` pairs appear twice.

This is a **repro only** (no fix), opened to showcase the bug. The single test is **expected to fail on `canary`**.

## Why it happens

With `cacheComponents`, a response is `[static shell] [prerender-deferred segments] [resume stream]`, stitched into one document.

1. React's `getPostponedState` snapshots `nextSegmentId` at `onAllReady` — **before** the pull-driven prelude flush.
2. The flush then **outlines** the large completed boundaries into the shell, allocating segment ids **after** that snapshot (`boundary.rootSegmentID = request.nextSegmentId++`).
3. The resume render is seeded from the stale snapshot (`request.nextSegmentId = postponedState.nextSegmentId`) and re-allocates the **same** ids for its own outlined boundaries.
4. The stitched document then has duplicate `B:`/`S:` ids. The inline `$RC` completion scripts resolve targets with `document.getElementById` (first match), so content cross-wires between boundaries — React #418/#419 recovery, duplicated/orphaned subtrees (a visible "tear" on slow networks).

It only surfaces when the shell actually **outlines** a completed boundary (large or suspensey shell); a small, fully-inlined shell leaves the snapshot equal to the final id, which is why this has been latent.

## Repro output (current `canary`, `next start`)

```
✕ serves the page without duplicate Suspense completion ids

  - Expected
  + Received
    Object {
  -   "segments": Array [],
  -   "templates": Array [],
  +   "segments": Array ["S:4", "S:5", "S:6"],
  +   "templates": Array ["B:4", "B:5", "B:6"],
    }
```

(Skips in dev — dev has no prerendered shell, so the resume path under test never runs.)

## Fix (separate)

Root cause is in React's Fizz — the `getPostponedState` snapshot timing — fixed upstream in facebook/react#36779 (finalize `nextSegmentId` after the prelude flush). A Next-side mitigation is also viable in the export path: bump the stored `nextSegmentId` past the shell's highest emitted id before writing `.meta`.

### DIFF
--- a/test/e2e/app-dir/ppr-resume-segment-collision/app/layout.tsx
+++ b/test/e2e/app-dir/ppr-resume-segment-collision/app/layout.tsx
@@ -1,0 +1,9 @@
+import type { ReactNode } from 'react'
+
+export default function RootLayout({ children }: { children: ReactNode }) {
+  return (
+    <html lang="en">
+      <body>{children}</body>
+    </html>
+  )
+}
--- a/test/e2e/app-dir/ppr-resume-segment-collision/app/page.tsx
+++ b/test/e2e/app-dir/ppr-resume-segment-collision/app/page.tsx
@@ -1,0 +1,68 @@
+import { Suspense } from 'react'
+import { cookies } from 'next/headers'
+
+// With cacheComponents (PPR) the response is:
+//   [static shell] [prerender-deferred segments] [resume stream]
+// React's getPostponedState snapshots `nextSegmentId` BEFORE the prelude flush.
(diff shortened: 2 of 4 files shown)


gaojude:
Superseded by #94820 (same branch pushed to the vercel/next.js origin instead of a fork).

PR #94830, opened 2026-06-15
[turbopack] Add `experimental.turbopackChunkingPriorities` to `next.config.js` with validation

sampoder:
These priorities will be used up this PR stack to weight whether or not to over-ship code as part of chunking.

### DIFF
--- a/packages/next/src/server/config-shared.ts
+++ b/packages/next/src/server/config-shared.ts
@@ -750,7 +750,7 @@ export interface ExperimentalConfig {
    * paths (e.g. `/products/[id]`), values are `'low' | 'medium' | 'high'`.
    * Routes not listed default to `'medium'`. Turbopack will avoid
    * over-shipping code to `high` priority routes to reduce page-load times
-   * on them.
+   * on them, at the cost of making other pages slower to load.
    */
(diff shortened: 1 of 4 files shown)


lukesandberg (approved):

packages/next/src/server/config-shared.ts
```diff
@@ -745,6 +745,15 @@ export interface ExperimentalConfig {
    */
   turbopackScopeHoisting?: boolean
 
+  /**
+   * (`next --turbopack` only) Per-route traffic priorities. Keys are route
+   * paths (e.g. `/products/[id]`), values are `'low' | 'medium' | 'high'`.
+   * Routes not listed default to `'medium'`. Turbopack will avoid
+   * over-shipping code to `high` priority routes to reduce page-load times
+   * on them.
+   */
```
bgw, line 754:
```suggestion
  /**
   * (`next --turbopack` only) Per-route traffic priorities. Keys are route
   * paths (e.g. `/products/[id]`), values are `'low' | 'medium' | 'high'`.
   * Routes not listed default to `'medium'`. Turbopack will avoid
   * over-shipping code to `high` priority routes to reduce page-load times
   * on them, at the cost of making other pages slower to load.
   */
```

bgw (approved):

PR #94831, opened 2026-06-15
[turbopack] Thread priorities for chunking through to `ChunkGroupInfo`

sampoder:
This PR takes the information from https://github.com/vercel/next.js/pull/94830 and threads it through to `ChunkGroupInfo`, which is then used in https://github.com/vercel/next.js/pull/94832. 

Probably the most important part of the PR to review is the part where it inherits the priority from its parent chunk group or the highest priority chunk group that references it. Is there a better place to put this traversal? I've tried to limit the time this takes by doing one traversal to pick up on the relationships and then just do the priority changes with those already known relationships. Maybe I can put this into an already existing traversal? Or maybe its best left separate whilst this is experimental.

### DIFF
--- a/crates/next-api/src/pages.rs
+++ b/crates/next-api/src/pages.rs
@@ -56,7 +56,7 @@ use turbopack_core::{
     module_graph::{
         GraphEntries, ModuleGraph, SingleModuleGraph, VisitedModules,
         binding_usage_info::compute_binding_usage_info,
-        chunk_group_info::{ChunkGroup, ChunkGroupEntry},
+        chunk_group_info::{ChunkGroup, ChunkGroupEntry, ChunkGroupPriority},
     },
     output::{OptionOutputAsset, OutputAsset, OutputAssets},
     reference::all_assets_from_entries,
@@ -745,9 +745,10 @@ impl PageEndpoint {
             }
 
             let graph = SingleModuleGraph::new_with_entries_visited_intern(
-                GraphEntries::from_chunk_groups(vec![ChunkGroupEntry::Entry(vec![
-                    ssr_chunk_module.ssr_module,
-                ])]),
+                GraphEntries::from_chunk_groups(vec![ChunkGroupEntry::Entry {
+                    modules: vec![ssr_chunk_module.ssr_module],
+                    priority: ChunkGroupPriority::default(),
+                }]),
                 visited_modules,
                 should_trace,
                 should_read_binding_usage,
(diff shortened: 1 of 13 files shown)


crates/next-api/src/pages.rs
```diff
@@ -1708,6 +1708,17 @@ impl Endpoint for PageEndpoint {
...
+        // The route's traffic priority from `experimental.turbopackChunkingPriorities`.
+        let priority = this
+            .pages_project
+            .project()
+            .next_config()
+            .turbopack_chunking_priorities()
+            .await?
+            .get(&this.pathname)
+            .copied()
+            .unwrap_or_default();
```
lukesandberg, line 1721: somewhere we need to validate the route names


imho a typo in a route name should be an error (in a production build) (ideally with an edit-distanced based 'did you mean' suggestion)

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -126,11 +131,49 @@ impl ChunkGroupInfo {
...
 #[derive(Debug, Clone, Hash, PartialEq, Eq, TraceRawVcs, Encode, Decode)]
 pub enum ChunkGroupEntry {
     Entry(Vec<ResolvedVc<Box<dyn Module>>>),
+    /// Same as [`ChunkGroupEntry::Entry`], but with an explicit traffic priority (e.g. from
+    /// `experimental.turbopackChunkingPriorities`). Keyed identically to `Entry` — the
+    /// priority does not affect chunk group identity.
+    EntryWithPriority {
+        modules: Vec<ResolvedVc<Box<dyn Module>>>,
+        priority: ChunkGroupPriority,
+    },
```
lukesandberg: can we just add priority to entry as is, do we need another variant (leverage the default?)

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -757,9 +820,99 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                | ChunkGroupKey::SharedMerged { parent, .. } = key
+                {
+                    let parent = parent.0 as usize;
+                    if parent != index && seen_edges.insert((parent, index)) {
+                        inherits_from[parent].push(index);
+                    }
+                }
+            }
+
+            graph.traverse_edges_unordered(|parent, target| {
```
lukesandberg, line 857: i think you can do this logic in a single pass using traverse_edges_bfs

you can use the chunk_groups_map to detect the chunk_group_boundaries and do the aggregation directly as they are discovered

Are import cycles the problem?

sampoder: Yeah, import cycles cause issues I think. I think there are some cases where you need to re-visit a node :/

crates/next-api/src/pages.rs
sampoder: Added to https://github.com/vercel/next.js/pull/94830

PR #94832, opened 2026-06-15
[turbopack] Weight the cost of overshipping code in chunks by route priority

sampoder:
This threads the priorities created down this PR stack into `make_production_chunks`. The weights are currently 1, 2, 4. So over-shipping on a high priority route is double the cost. The code comment in the code (which is most of the LOC) explains where the priority was fitted into the calculation.

### DIFF
--- a/turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
+++ b/turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
@@ -15,7 +15,7 @@ use crate::{
     },
     module_graph::{
         ModuleGraph,
-        chunk_group_info::{ModuleToChunkGroups, RoaringBitmapWrapper},
+        chunk_group_info::{ChunkGroupPriority, ModuleToChunkGroups, RoaringBitmapWrapper},
     },
 };
 
@@ -38,6 +38,8 @@ pub async fn make_production_chunks(
     async move {
         let module_chunk_groups = module_graph.chunk_group_info().module_chunk_groups();
         let merged_modules = module_graph.merged_modules().await?;
+        let chunk_group_info = module_graph.chunk_group_info().await?;
+        let chunk_group_priorities = chunk_group_info.chunk_group_priorities.as_slice();
 
         #[derive(Default)]
         struct GroupedChunkItems<'l> {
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +469,44 @@ pub async fn make_production_chunks(
...
+                                            b,
+                                            chunk_group_priorities,
+                                        ),
+                                        max_overship_weight_in_difference(
+                                            b,
+                                            a,
+                                            chunk_group_priorities,
+                                        ),
+                                    ),
+                                    _ => (OVERSHIP_WEIGHT_MEDIUM, OVERSHIP_WEIGHT_MEDIUM),
```
lukesandberg: can you comment on this case? is it basically impossible?

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +469,44 @@ pub async fn make_production_chunks(
...
+                                };
+
+                            // `pre_d` is the bracketed factor of `m * 3d` from the derivation
+                            // above: the merge benefit `d` scaled by `m
+                            // = OVERSHIP_WEIGHT_MEDIUM` and without the
+                            // common positive divisor `o_groups / (rem_g * groups)`.
+                            // Since that divisor is positive, `sign(d) == sign(pre_d)`, so we gate
+                            // on `pre_d` before computing `d`.
+                            let pre_d = OVERSHIP_WEIGHT_MEDIUM
```
lukesandberg: maybe `OVERSHIP_WEIGHT_DEFAULT` is a better name?

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: Yeah, this is impossible because the overlap would be 0: https://github.com/vercel/next.js/pull/94832/changes#diff-e03b8113c2087d0c96a1cc4edc1a8f31ec8021205e4de0f25de01a40528d538eR255

PR #94842, opened 2026-06-16
pre-cache idents

lukesandberg:
### What

`SingleModuleGraph` nodes now carry the per-module data that the "late" graph traversals previously re-derived by reading per-module turbo-tasks cells (`Module::ident`, `ident_string`, `is_self_async`, `side_effects`, `MergeableModule::is_mergeable`).

On `canary`, `SingleModuleGraphNode::Module` was a bare tuple — `Module(ResolvedVc<Box<dyn Module>>)` — and every consuming pass re-read the module's properties on demand, fanning out a turbo-task per module per pass. This PR turns it into a struct that stores the resolved data:

```rust
Module {
    module: ResolvedVc<Box<dyn Module>>,
    ident: ResolvedVc<AssetIdent>,
    ident_string: Option<ResolvedVc<RcStr>>,
    is_self_async: ResolvedVc<bool>,
    side_effects: Option<ResolvedVc<ModuleSideEffects>>,
    is_mergeable: Option<ResolvedVc<bool>>,
}
```

- `ident` / `is_self_async` are always collected (`async_module_info()` runs on every graph during chunking; idents are broadly needed).
- `ident_string` / `side_effects` / `is_mergeable` are gated by new `ModuleGraphOptions` bits (`include_ident_strings`, `include_side_effects`, `include_mergeable`), collected only when the consuming pass actually runs — `include_side_effects` ↔ `turbopackRemoveUnusedImports`, `include_mergeable` ↔ scope hoisting, `include_ident_strings` ↔ the deterministic module-id strategy. `None` on a gated field unambiguously means "the graph wasn't built with that bit," and consumers `bail!` rather than silently fall back to a per-module read.

### How

Resolution happens in a single per-module turbo-tasks function, `module_graph_node_data(module, …)`, rather than inline in the graph-construction task. The builder node holds the resolved (pointer-only, unread) `ModuleGraphNodeData` cell; its content is read **exactly once**, at the dedup point in `new_inner`, with `final_read_hint` so the backend can drop the cell afterward (no intermediary copy retained in the task cache).

This keeps the per-module dependency edges distributed across many small per-module tasks rather than concentrating them on the single construction task, while still tracking those reads in the ultimate consumer.

Consumers updated to read from the node instead of re-reading the module: `compute_async_module_info`, `compute_side_effect_free_module_info`, `compute_merged_modules` (gated on the chunking context's `is_module_merging_enabled()`), `get_global_module_id_strategy`, and `server_actions.rs`.

`ModuleGraphOptions` construction is also consolidated: `Project::module_graph_options()` builds the options for every project-derived graph in one place, with `include_ident_strings` derived from `Project::uses_deterministic_module_ids()` (which now also accounts for per-page graphs) so the per-page and whole-app graphs can't drift.

### Performance

Measured on a large internal Next.js production build (full `next build`). Wall time is within run-to-run noise; the win is in distributed/serial CPU — the per-module reads move out of the late, serial-leaning traversals into the highly-parallel graph-construction phase.

**Single-threaded (serial) CPU time, from Xcode Instruments traces** — `canary` vs. this change (two independent captures of this change):

| metric                                       | canary | this — run 1 | this — run 2 |
| -------------------------------------------- | ------ | ------------ | ------------ |
| single-threaded wall                         | ~5.3s  | ~3.1s        | ~3.5s        |
| single-threaded CPU                          | ~5.25s | ~2.98s       | ~3.44s       |
| `aggregation_update` — in serial windows     | 101ms  | 26ms         | 5ms          |
| `aggregation_update` — total CPU             | 3438ms | 3213ms       | 3236ms       |
| `module_graph_node_data` — in serial windows | n/a    | 0ms          | 0ms          |
| `module_graph_node_data` — total CPU         | n/a    | ~29ms        | ~28ms        |

Relocating the per-module child-edge aggregation into the parallel construction phase **reduces single-threaded time by ~1.8–2.2s** and cuts serial `aggregation_update` by 4–20×. The new `module_graph_node_data` task contributes **0ms** to serial windows (fully parallel) despite adding one child per module to the construction task.

**Selected late-traversal task times** (whole-build, illustrative — these reads are now satisfied from the node):

| task                        | canary | this   |
| --------------------------- | ------ | ------ |
| `get_project_feature_usage` | ~2.5s  | ~80ms  |
| collect mergeable modules   | ~391ms | ~36ms  |

`SingleModuleGraphNode` grows from 8 bytes (the bare `ResolvedVc` tuple) to 48 bytes (the added resolved fields); a `size_test` pins this. Peak RSS is up ~0.6%.

### DIFF
--- a/turbopack/crates/turbopack-core/src/module_graph/mod.rs
+++ b/turbopack/crates/turbopack-core/src/module_graph/mod.rs
@@ -1,6 +1,7 @@
 use std::{
     collections::{BinaryHeap, VecDeque},
     future::Future,
+    hash::{Hash, Hasher},
     iter::FusedIterator,
     ops::Deref,
 };
@@ -26,9 +27,10 @@ use turbo_tasks::{
 use turbo_tasks_fs::FileSystemPath;
 
 use crate::{
-    chunk::{AsyncModuleInfo, ChunkingContext, ChunkingType, TracedMode},
+    chunk::{AsyncModuleInfo, ChunkingContext, ChunkingType, MergeableModule, TracedMode},
+    ident::AssetIdent,
     issue::{ImportTracer, ImportTraces, Issue},
-    module::Module,
+    module::{Module, ModuleSideEffects},
     module_graph::{
         async_module_info::{AsyncModulesInfo, compute_async_module_info},
         binding_usage_info::BindingUsageInfo,
@@ -151,7 +153,7 @@ impl VisitedModules {
                 .await?
                 .enumerate_nodes()
                 .flat_map(|(node_idx, module)| match module {
-                    SingleModuleGraphNode::Module(module) => Some((
+                    SingleModuleGraphNode::Module { module, .. } => Some((
                         *module,
                         GraphNodeIndex {
                             graph_idx: 0,
@@ -191,7 +193,7 @@ impl VisitedModules {
                 graph
                     .enumerate_nodes()
                     .flat_map(|(node_idx, module)| match module {
-                        SingleModuleGraphNode::Module(module) => Some((
+                        SingleModuleGraphNode::Module { module, .. } => Some((
                             *module,
                             GraphNodeIndex {
                                 graph_idx: this.next_graph_idx,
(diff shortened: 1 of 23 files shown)


lukesandberg:
* **#94842** <a href="https://app.graphite.com/github/pr/vercel/next.js/94842?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94842?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94977** <a href="https://app.graphite.com/github/pr/vercel/next.js/94977?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

turbopack/crates/turbopack-core/src/module_graph/mod.rs
```diff
@@ -1076,6 +1152,28 @@ impl ModuleGraphSnapshot {
...
 
+    /// Returns a module's resolved `AssetIdent` `Vc`, preferring the eagerly-resolved ident stored
+    /// on the node (graphs built with [`ModuleGraphOptions::include_idents`]) and falling back to
+    /// `module.ident()` otherwise.
+    ///
+    /// Unlike [`Self::module_ident`] this does not bail when idents weren't collected, so it's
+    /// usable from consumers that may run on graphs built either way (e.g.
+    /// `get_global_module_id_strategy`, which runs on the whole-app production graph as well as the
+    /// CLI build graph). The returned `ResolvedVc` is resolved either way; callers read it tracked.
+    pub fn module_ident_resolved(
```
lukesandberg: this function should move down closer to the other ones

turbopack/crates/turbopack-core/src/module_graph/mod.rs
```diff
@@ -1907,14 +2185,23 @@ impl Visit<SingleModuleGraphBuilderNode, RefData> for SingleModuleGraphBuilder<'
...
             SingleModuleGraphBuilderNode::Module {
-                ident: Some(ident), ..
+                span_ident: Some(span_ident),
+                ..
+            } => {
+                // Format the span name from the in-memory ident (no extra read). `AssetIdent` has
+                // no `Display`, so use its path; span-name fidelity is
+                // non-load-bearing. `span_ident` is populated whenever `emit_spans`, so this arm is
+                // taken for every `Module` node here.
+                tracing::info_span!("module", name = display(&span_ident.path.path))
```
lukesandberg: this is a behavior change, previously we were using the ident to_string now it is just the path

turbopack/crates/turbopack-core/src/module_graph/mod.rs
lukesandberg: :robot: Moved next to `get_ids` (the other ident accessor) in 'impl ModuleGraphSnapshot'.

turbopack/crates/turbopack-core/src/module_graph/mod.rs
lukesandberg: :robot: Fixed: added a synchronous `AssetIdent::to_span_name()` that reproduces the old `ident.to_string()` span name (path, query, fragment, layer, modifiers, content type, parts) with no async read. The only difference from the old behavior is that nested `assets` are elided as `{…}` (rare, and would require an async read); span names don't need to be unique.

turbopack/crates/turbopack-core/src/module_graph/mod.rs
```diff
@@ -678,12 +769,27 @@ impl ModuleGraphImportTracer {
...
             .iter()
-            .map(|(&module, _)| async move { Ok((module.ident().await?.path.clone(), module)) })
+            .map(|(&module, &node_idx)| async move {
+                let ident = match graph
+                    .graph
+                    .node_weight(node_idx)
+                    .and_then(|n| n.ident_resolved())
+                {
+                    Some(ident) => ident,
+                    None => module.ident().to_resolved().await?,
```
lukesandberg: do we need the fallback here?

PR #94851, opened 2026-06-16
fix(eslint-plugin-next): detect async client components in named exports

### LINKED ISSUE #77243: Eslint rule no-async-client-component doesn't work

https://github.com/ThomasKientz/next-eslint

### To Reproduce

1. Run `npm run lint`

I expected ESLint to output the error 

> "no-async-client-component" 

instead, ESLint outputs 

> No ESLint warnings or errors

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T8122
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 20.14.0
  npm: 10.7.0
  Yarn: N/A
  pnpm: 10.5.0
Relevant Packages:
  next: 15.2.3 // Latest available version is detected (15.2.3).
  eslint-config-next: 15.2.3
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 5.8.2
Next.js Config:
  output: N/A
```

Linting

next dev (local)

### Additional context

_No response_

eran132:
## What

The [`no-async-client-component`](https://nextjs.org/docs/messages/no-async-client-component) ESLint rule only inspected `ExportDefaultDeclaration` nodes. As a result, async Client Components exported as **named exports** were never flagged, even though they hit the same runtime pitfall:

```jsx
"use client"

// all three were silently missed before this change:
export async function MyComponent() { return <></> }
export const MyComponent = async () => { return <></> }

async function MyComponent() { return <></> }
export { MyComponent }
```

Closes #77243.

## How

- Extracted the existing identifier-resolution logic and the inline async/capitalized check into two shared helpers (`checkExportedIdentifier`, `checkInlineDeclaration`).
- Applied them to `ExportNamedDeclaration` as well — covering inline declarations (`export async function` / `export const = async () =>`) and specifiers (`export { MyComponent }`).
- The existing capitalization guard is preserved, so lowercase helpers (e.g. `export async function useData()`) are still not flagged, and non-async components are untouched.

No public API or config changes; this only broadens detection of an existing recommended rule.

## Testing

Extended `test/unit/eslint-plugin-next/no-async-client-component.test.ts`:
- **invalid** (now reported): named inline async function, named inline async arrow, and both specifier forms.
- **valid** (still not reported): lowercase named export, and a non-async named function export.

Verified against the built rule with ESLint's `RuleTester` (all 8 valid + 9 invalid cases pass). As a regression control, the four new named-export `invalid` cases fail against the pre-change rule and pass after it. `eslint-plugin-next` `index.test.ts`, `tsc`, and `prettier` all pass.

### DIFF
--- a/packages/eslint-plugin-next/src/rules/no-async-client-component.ts
+++ b/packages/eslint-plugin-next/src/rules/no-async-client-component.ts
@@ -24,6 +24,90 @@ export default defineRule({
       Program(node) {
         let isClientComponent: boolean = false
 
+        // Reports when an inline declaration is an async, capitalized
+        // function/arrow, e.g. `export [default] async function MyComponent() {}`
+        // or `export const MyComponent = async () => {}`.
+        function checkInlineDeclaration(declaration, reportNode) {
+          if (
+            declaration?.type === 'FunctionDeclaration' &&
+            declaration.async &&
+            isCapitalized(declaration.id.name)
+          ) {
+            context.report({ node: reportNode, message })
+            return
+          }
+
+          if (declaration?.type === 'VariableDeclaration') {
+            for (const varDeclarator of declaration.declarations) {
+              if (
+                varDeclarator.id?.type === 'Identifier' &&
+                isCapitalized(varDeclarator.id.name) &&
+                varDeclarator.init?.type === 'ArrowFunctionExpression' &&
+                varDeclarator.init.async
+              ) {
+                context.report({ node: reportNode, message })
+              }
+            }
+          }
+        }
+
+        // Resolves an exported identifier (e.g. `export default MyComponent` or
+        // `export { MyComponent }`) to its top-level declaration and reports
+        // when it is an async, capitalized function/arrow.
+        function checkExportedIdentifier(targetName: string) {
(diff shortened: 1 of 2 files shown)


eran132:
Thanks for the review! Addressed in f2797c3:

- **Anonymous default async export crash** (`export default async function () {}`): `FunctionDeclaration#id` is `null` there, so `declaration.id.name` could throw. Now guarded with optional chaining — anonymous exports have no name to flag, so they are skipped (no exception either way). Added a `valid` test for it.
- **Aliased named exports** (`export { foo as MyComponent }`): the declaration is still resolved by the local name, but it is treated as a component when either the local or the exported (aliased) name is capitalized, so capitalized aliases are no longer a false negative. Added an `invalid` test for it.
- **Misleading test comment**: clarified the lowercase named-export cases to state they verify lowercase names are *not* flagged.

All 19 unit cases pass under the real `next/jest` runner; `tsc` and `prettier` are clean.

PR #94857, opened 2026-06-16
[CC] Fix accumulator flushing logic

lubieowoce:
The `waitAtLeastOnReactRenderTask()` trick we used in prerenders to wait for flushing was subtly incorrect. It only worked if the code before it resulted in React's `scheduleWork` being called synchronously before it (i.e. already having its immediate scheduled when `waitAtLeastOnReactRenderTask()` schedules its own). 

This currently works, because `finishSegmentAccumulator` calls the `then()` callbacks synchronously (which calls `pingTask` which schedules a flush): https://github.com/vercel/next.js/blob/40296fe53f40ffe6a0e2184879bd5e294d271ce4/packages/next/src/server/app-render/vary-params.ts#L369-L371

However, if `scheduleWork` was called in a microtask instead, then `waitAtLeastOnReactRenderTask()` would queue its immediate first, so it'd resolve before Flight actually flushed. This becomes the case in #94809, where the vary params thenable is replaced with an async iterable (which uses normal promises in the iterator, and thus always schedules work in a microtask)

TLDR:
- replaced `await waitAtLeastOneReactRenderTask()` with a separate task for finishing accumulators (which doesn't rely on immediate ordering, and thus is resilient to delaying the `scheduleWork` by a couple microtasks)
- as a drive-by cleanup, i moved the sync IO abort logic into a thing we check on top of every task and only do once. otherwise the code was confusing to follow, and was unnecessarily computing accumulator things for renders that are already aborted

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -1750,6 +1750,39 @@ async function finalRuntimeServerPrerender(
     }
   }
 
+  let didHandleUnexpectedAbort = false
+  /**
+   * @returns - whether or not the task should be skipped
+   * because the render was already aborted.
+   * */
+  const checkUnexpectedAbort = (): boolean => {
+    if (finalServerController.signal.aborted) {
+      // If the server controller is already aborted, then we must have encountered sync IO
+      if (!didHandleUnexpectedAbort) {
+        didHandleUnexpectedAbort = true
+        onUnexpectedAbort()
+      }
+      return true
+    }
+
+    // Not aborted.
+    return false
+  }
+
+  const onUnexpectedAbort = () => {
+    resultIsPartial = true
+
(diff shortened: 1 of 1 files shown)


lubieowoce:
* **#94809** <a href="https://app.graphite.com/github/pr/vercel/next.js/94809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94857** <a href="https://app.graphite.com/github/pr/vercel/next.js/94857?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94857?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

unstubbable (approved):

Issue #92787, opened 2026-04-14
Object.fromEntries(URLSearchParams) drops duplicate keys → stale UI on navigation


krivcikfilip:
### Link to the code that reproduces this issue

https://github.com/krivcikfilip/next-multi-searchparams-bug

### To Reproduce

1. npm install && npm run dev
2. Click "+ red" → /?color=red (2 items) ✅
3. Click "+ blue" → /?color=red&color=blue (4 items) ✅
4. Click "✕ red" → /?color=blue — still 4 items ❌ (should be 2)

### Current vs. Expected behavior

ppr-navigations.ts uses Object.fromEntries(new URLSearchParams(search)) for cache keys.
Object.fromEntries keeps only the last value per key, so:

  "?color=red&color=blue" → {color: "blue"}
  "?color=blue"           → {color: "blue"}

Same key → cache hit → no re-render.

Fix: replace Object.fromEntries() with a method preserving all values (e.g. getAll()).

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:41 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
Binaries:
  Node: 24.6.0
  npm: 11.5.1
  Yarn: 1.22.19
  pnpm: 10.18.3
Relevant Packages:
  next: 16.2.1-canary.38 // Latest available version is detected (16.2.1-canary.38).
  eslint-config-next: N/A
  react: 19.2.5
  react-dom: 19.2.5
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Linking and Navigating

### Which stage(s) are affected? (Select all that apply)

next dev (local)

marcoshernanz:
Thanks for the report. This was fixed in #94863. The fix preserves repeated search params in client page segment cache keys and includes an e2e regression test for this exact transition. I confirmed the repro fails on `16.3.0-canary.70` and passes on `16.3.0-canary.71`, `16.3.0`, and the current canary. Closing as fixed.

PR #94863, opened 2026-06-16
fix: preserve repeated search params in client page segment cache keys

### LINKED ISSUE #94821: Submitting Query Parameters via a Form does not always update the client with the newly server rendered content

https://gitlab.com/JanBayer/nextjs-stale-query-param-rendering

### To Reproduce

1. Clone repro repo
2. Run `npm install`
3. Run `npm run dev`
4. Open `http://localhost:3000` in your browser
5. Check the first box, click submit
6. Check the second box, click submit
7. Uncheck the first box, click submit

## Current Behavior:

The query param will have updated and the server will have logged the new update, but the client will not have received the newly rendered content.

## Expected Behavior:

The client should see "test2" instead of ["test1", "test2"]

```bash
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 17560
  Available CPU cores: 8
Binaries:
  Node: 24.15.0
  npm: 11.12.1
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.3.0-canary.51 // Latest available version is detected (16.3.0-canary.51).
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Form (next/form)

### Additional context

_No response_

icyJoseph:
Fixes: #94821, #92787, #93104

Using `Object.fromEntries` on URLSearchParams, drops repeated keys.

```js
const s = '?foo=bar&foo=baz'
Object.fromEntries(new URLSearchParams(s)) // { foo: 'baz' }
```

There's already a function, `urlSearchParamsToParsedUrlQuery`, which handles this case.

https://github.com/vercel/next.js/blob/40296fe53f40ffe6a0e2184879bd5e294d271ce4/packages/next/src/client/route-params.ts#L235

```js
urlSearchParamsToParsedUrlQuery(new URLSearchParams(s)) // { foo: ['bar', 'baz'] }
```

- [x] The e2e off one of the issue reports, I'll try to simplify it.


acdlite (approved):

stephenhmarsh:
it sounds like this fixes https://github.com/vercel/next.js/issues/92152 too, no?

PR #94866, opened 2026-06-16
Fix Instant Navs DevTools capture bugs and re-enable its test suite

unstubbable:
The `instant-nav-panel` devtools test suite was skipped in #94196 after failing in roughly 40% of CI runs. This pull request fixes the dev-capture defects behind that flakiness and re-enables the suite.

These would normally be separate pull requests, but they resist a clean split. Every fix is required to unblock re-enabling the suite, and each was discovered while doing so, as removing one defect exposed the next along the dev-capture path. That same suite is also the only test that exercises the path, so it is the shared verification for all of the fixes; splitting them would either ship fixes with no passing test or force the suite to be duplicated. The changes are therefore kept in one pull request, organized as **self-contained commits meant to be reviewed one at a time**. Each commit message carries the full reasoning, so the summaries below are only an index.

1. Fix instant navs capture silently stopping after "Continue Rendering". The dev overlay panel restarts capture by chaining a cookie delete and a fresh pending cookie, so it no longer relies on an effect that could miss the `renderingIndicator` transition and silently stop capturing.

2. Stop nulling a superseded segment cache entry that's still in use. `upsertSegmentEntry` stops mutating a candidate that a navigation seed is already reading, which had been resolving that read to `null` and rendering captured pages empty.

3. Create a dev debug channel for runtime prefetch responses. The server now creates the React debug channel the client always opens for these responses, so the decode can close and the stale-time read can settle.

4. Tee the dev debug channel so a response can be decoded more than once. `createDebugChannel` hands each decode its own tee branch instead of a shared, already-locked reader, fixing the crash when a prefetch response is decoded twice.

5. Retain dev debug channels past stream close for late decodes. The client keeps closed channels in an LRU-bounded map so a late decode finds the buffered stream rather than a fresh, never-fed pair that hangs.

6. Re-enable the instant-nav devtools suite under partial prefetching. Un-skips the suite, migrates it to two-phase prefetching, and adds `prefetch={true}` coverage for the runtime-data path that exercises the fixes above.

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -924,14 +924,18 @@ export function upsertSegmentEntry(
       // (TODO: can this be true if `candidateEntry.fetchStrategy >= existingEntry.fetchStrategy`?)
       (!existingEntry.isPartial && candidateEntry.isPartial)
     ) {
-      // We're going to leave revalidating entry in the cache so that it doesn't
-      // get revalidated again unnecessarily. Downgrade the Fulfilled entry to
-      // Rejected and null out the data so it can be garbage collected. We leave
-      // `staleAt` intact to prevent subsequent revalidation attempts only until
-      // the entry expires.
-      const rejectedEntry: RejectedSegmentCacheEntry = candidateEntry as any
-      rejectedEntry.status = EntryStatus.Rejected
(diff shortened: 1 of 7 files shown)


samselikoff (approved):

acdlite (approved):
very nice!

PR #94869, opened 2026-06-16
Rename experimental.useExperimentalReact to blockingSSR

gaojude:
`experimental.useExperimentalReact` names the React *build* it pulls in rather than the feature it's there for. Rename it to `blockingSSR` — the actual capability (React's blocking SSR, which emits `<link rel="expect">` to hold first paint until the streamed shell is coherent) — mirroring React's own feature flag so it's clear why the experimental channel got enabled when grepping.


lubieowoce (approved):

packages/next/src/server/config.ts
```diff
@@ -546,11 +546,11 @@ function assignDefaultsAndValidate(
...
-  // `useExperimentalReact` only opts *into* React's experimental channel; it
-  // cannot opt out of it when another feature (e.g. `taint`) structurally
-  // requires it — those APIs only exist in the experimental React build. Warn
-  // so an explicit `false` here doesn't look like it was silently ignored.
-  if (result.experimental.useExperimentalReact === false) {
+  // `blockingSSR` only opts *into* React's experimental channel; it cannot opt
+  // out of it when another feature (e.g. `taint`) structurally requires it —
+  // those APIs only exist in the experimental React build. Warn so an explicit
+  // `false` here doesn't look like it was silently ignored.
+  if (result.experimental.blockingSSR === false) {
```
lubieowoce: should remove this whole block think? it made sense when the whole flag was `useExperimentalReact` (because the other flags could be viewed as dependent on it). but if it's `blockingSSR` then it doesn't really make sense to view the rest as dependent on it

it might make sense to warn for invalid combinations of these instead, i.e. check that if one of them is `true`, then none of the other ones are an explicit `false`, bc you can't opt out like that. but that's different from what this is

lubieowoce (changes requested):

lubieowoce (approved):

eps1lon (approved):

PR #94870, opened 2026-06-16
[ci] Migrate build_and_test and pull_request_stats to use arm64 runners

bgw:
A 16 core x86-64 runner is $0.042/minute. A 16 core arm64 runner is $0.026/minute.
https://docs.github.com/en/billing/reference/actions-runner-pricing

It also looks like they're faster! After factoring the performance improvement, this appears to be a ~50% cost reduction for migrated jobs.

Datadog SQL:

```
SELECT
  REGEXP_REPLACE(ci_job_name, '\s*\([0-9]+/[0-9]+\)', '') AS ci_job_name,
  labels,
  branch,
  commit_date,
  SUM(duration) / 1000 / 1000 / 1000 / 60 as minutes,
  ci_pipeline_id
FROM dd.ci_pipelines(
    columns => ARRAY ['@git.commit.committer.date', '@ci.pipeline.id', '@duration', '@git.branch', '@ci.node.labels', '@ci.job.name'],
    event_type => 'job',
    filter => '@git.repository.id_v2:github.com/vercel/next.js @ci.pipeline.name:build-and-test @ci.provider.instance:github-actions @ci.pipeline.id:(28266192824-1 OR 28266190122-2) -@ci.job.name:*changed*',
    from_timestamp => now() - INTERVAL '30 days',
    to_timestamp => now()
  ) AS (
    commit_date VARCHAR,
    ci_pipeline_id VARCHAR,
    duration BIGINT,
    branch VARCHAR,
    labels VARCHAR,
    ci_job_name VARCHAR
  )
GROUP BY
  REGEXP_REPLACE(ci_job_name, '\s*\([0-9]+/[0-9]+\)', ''),
  labels,
  commit_date,
  branch, 
  ci_pipeline_id
ORDER BY 
  REGEXP_REPLACE(ci_job_name, '\s*\([0-9]+/[0-9]+\)', ''),
  branch,
  commit_date,
  minutes;
```

https://app.datadoghq.com/ddsql/editor?ddsql_dataset_id=34de66b9-30f4-48dd-ac98-354dc7e67fdf

Full Data: https://docs.google.com/spreadsheets/d/1SszR3RsPeJ13nVW5ggMakwy87fkQMqWMHbZDehHH8IA/edit?usp=sharing

Here's the abbreviated data for jobs longer than 5 minutes:
Screenshot 2026-06-26 at 4 38 00 PM

It seems like for the jobs switched over to arm64, they run in about ~73-82% of the time as arm.

PR #91250, opened 2026-03-12
[next-playwright] Use unique cookie values for instant navigation testing lock

eps1lon:
Chrome >= 136 doesn't fire CookieStore change events if a cookie is set with the same value (Cursor analysis). This means that nesting `instant` didn't warn in newer versions of Playwright.

We missed this due to using older Playwright versions. This PR fixes the Playwright issues by using a unique Cookie value (`performance.now()`). I'll also bump Playwright in the same change to show the breakage.

Looks like `LaunchOptions['devtools']` was removed (or never worked). Switched to using CLI args instead ([`--auto-open-devtools-for-tabs`](https://developer.chrome.com/docs/devtools/open#auto)).

Newer Chrome versions have additional policies to avoid cross-origin requests. You now get prompts when accessing "insecure content". I haven't found a way to grant that permission. Since we only access localhost, we can avoid the prompt by allowing accessing local content.

[Cursor thread for fixing unexpected CI failures](https://cursor.com/agents/bc-793179a6-5ccc-40f4-8e91-4bb50bb32a4c)

PR #94871, opened 2026-06-16
[ci] Update playwright to 1.61.0

bgw:
[Playwright v1.61.0 adds support for ubuntu-26.04](https://github.com/microsoft/playwright/releases/tag/v1.61.0). When Lindsey created our arm64 runners (see https://github.com/vercel/next.js/pull/94870), [he picked the ubuntu-26.04 image](https://vercel.slack.com/archives/C01LN7C5QR5/p1781608258610389?thread_ts=1781118174.353839&cid=C01LN7C5QR5) (which is technically still [in preview](https://github.com/actions/runner-images#available-images), but IMO that's fine).


bgw:
* **#94870** <a href="https://app.graphite.com/github/pr/vercel/next.js/94870?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94871** <a href="https://app.graphite.com/github/pr/vercel/next.js/94871?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94871?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

test/lib/browsers/playwright.ts
```diff
@@ -247,7 +247,17 @@ export class Playwright<TCurrent = undefined> {
...
+      let launchArgs: string[] = [
+        // Chrome enables Local Network Access checks by default, which block
+        // connections from a public origin to a loopback/local dev server.
+        // Some tests (i.e. test/development/basic/allowed-dev-origins.test.ts)
+        // deliberately make cross-origin requests to the local dev server so
+        // Next.js's own allowedDevOrigins logic can decide whether to block
+        // them. Disable the browser-level check so those requests reach the
+        // server.
+        // https://developer.chrome.com/blog/local-network-access
+        '--disable-features=LocalNetworkAccessChecks',
```
eps1lon: This should be per test similar to what we already did for `test/development/basic/allowed-dev-origins.test.ts` in https://github.com/vercel/next.js/pull/91250/changes#diff-6cd1a1271162bb4e44a04f9d36f212c28d3a62fdd825f4c524e54db109a42878R134

bgw: That's much better. Thanks for the pointer.

pnpm-workspace.yaml
```diff
@@ -46,3 +46,7 @@ minimumReleaseAgeExclude:
   - react-server-dom-*
   - scheduler
   - turbo
+  # these playwright exemptions can be removed once v1.61.0 is >48hrs old
+  - playwright
+  - playwright-core
+  - playwright-chromium
```
eps1lon: Using it that way is quite sketchy. We're basically vulnerable again to supply-chain attacks.

Fortunately, you can list a specific version
```suggestion
  # these playwright exemptions can be removed once v1.61.0 is >48hrs old
  - playwright@1.61.0
  - playwright-core@1.61.0
  - playwright-chromium@1.61.0
```

bgw: The good news is that it's been 48 hours, so I can just remove this.

This gets used in combination with the lockfile and doesn't apply to transitive dependencies, so I wouldn't exactly say this opens us up to supply chain attacks, but it's good to know we can limit it to a specific version.

eps1lon: > doesn't apply to transitive dependencies

Why would we have to list `playwright-core` then? That's not a direct dependency of this worktree.

eps1lon (approved):

pnpm-workspace.yaml
bgw: Precisely because `minimumReleaseAgeExclude` doesn't automatically exclude transitive dependencies.

`playwright@1.61.0` depends on `playwright-core@1.61.0`. Both of these packages were <48hrs old, so both had to be explicitly included in the list.

eps1lon: What I'm saying is that `playwright-core` could be compromised. If another package starts depending on it that we depend on, we're pwned. It makes no sense to not specify a version in `minimumReleaseAgeExclude`. You're just slowly moving back to a world without `minimumReleaseAgeExclude`

bgw: Sure, there's no reason to not specify a version. I wasn't aware of the ability to specify a version. If I was, I would've done that.

I'm just saying that the lockfile also provides some level of protection, so we're not 100% leaning on `minimumReleaseAgeExclude` to protect us against supply chain attacks. If some other package depended on a newer version of `playwright-core`, we'd have to update that other package to become compromised.

PR #94872, opened 2026-06-16
cache DelegatingImportTracer across use

wbinnssmith:
Previously, we created new `DelegatingImportTracer`s even if the collectibles were the same. This stores the collectibles in a ~Vec~FrozenSet instead, ~sorted by cell id~ so the task input is cached.

Test Plan: Performance test tbd

### DIFF
--- a/turbopack/crates/turbopack-core/src/issue/mod.rs
+++ b/turbopack/crates/turbopack-core/src/issue/mod.rs
@@ -14,11 +14,12 @@ use auto_hash_map::AutoSet;
 use bincode::{Decode, Encode};
 use serde::{Deserialize, Serialize};
 use turbo_esregex::EsRegex;
+use turbo_frozenmap::FrozenSet;
 use turbo_rcstr::{RcStr, rcstr};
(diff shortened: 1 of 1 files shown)


bgw (approved):
LGTM as long as the perf benefit is there

turbopack/crates/turbopack-core/src/issue/mod.rs
```diff
@@ -193,7 +193,15 @@ pub trait ImportTracer {
 #[turbo_tasks::value]
 #[derive(Debug)]
 pub struct DelegatingImportTracer {
-    delegates: AutoSet<ResolvedVc<Box<dyn ImportTracer>>>,
+    delegates: Vec<OrdResolvedVc<Box<dyn ImportTracer>>>,
```
bgw: Now that we do have `Ord`, this can be a `FrozenSet`, because a `FrozenSet` is just a sorted and deduped boxed slice, the advantage being that the type system enforces that it's sorted.

PR #94876, opened 2026-06-16
[ci] Remove playwright stamp file logic

bgw:
This is left over from self-hosted runners. This should never be used in gh hosted runners.

PR #94882, opened 2026-06-17
[ci] Only install chromium-headless-shell and not chromium

bgw:
Because it's not sure which one you'll want, playwright installs TWO copies of Chromium by default. A full copy and a stripped down headless version.

Most of the install time is spent in `apt-get`/`dpkg`, but this at least buys us a few seconds back.

Before:

After:

Screenshot 2026-06-16 at 6 24 50 PM

Playwright also installs ffmpeg in case you want to perform a screen recording (we don't), but it's much smaller (2.3 MiB), and there doesn't appear to be a way to disable it.

### DIFF
--- a/.github/workflows/build_reusable.yml
+++ b/.github/workflows/build_reusable.yml
@@ -327,7 +327,7 @@ jobs:
 
       - name: Install Playwright browsers
         if: ${{ inputs.skipInstallBuild != 'yes' && inputs.needsPlaywright != 'no' }}
-        run: pnpm playwright install --with-deps ${{ inputs.browser }}
+        run: pnpm playwright install --with-deps --only-shell ${{ inputs.browser }}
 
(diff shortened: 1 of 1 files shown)


bgw:
* **#94882** <a href="https://app.graphite.com/github/pr/vercel/next.js/94882?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94882?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#94876** <a href="https://app.graphite.com/github/pr/vercel/next.js/94876?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

timneutkens (approved):

eps1lon (approved):

PR #94894, opened 2026-06-17
[otel] Add more verbose-mode OTEL spans for App Router

mischnic:
This adds two new spans:
- `AppRender.renderToNodeFizzStream`
- `AppRender.waitShellReady`

Using the example from https://github.com/switz/nextjs-tracing-repro

with NEXT_OTEL_VERBOSE:

Bildschirmfoto 2026-06-17 um 18 11 28


packages/next/src/server/lib/trace/constants.ts
```diff
@@ -132,6 +134,8 @@ export const NextVanillaSpanAllowlist = new Set([
   RenderSpan.getStaticProps,
   AppRenderSpan.fetch,
   AppRenderSpan.getBodyResult,
+  AppRenderSpan.waitShellReady,
+  AppRenderSpan.renderToNodeFizzStream,
```
mischnic: Do we want these in `NextVanillaSpanAllowlist`?

timneutkens (approved):

PR #94896, opened 2026-06-17
Let LLMs create less noisy and more focused PR descriptions

eps1lon:
The default template is too noisy when used by LLMs. "what" and "how" are almost always duplicates. Verification most often includes trivial commands (e.g. static checking of .js files or running `pnpm build`). This information isn't relevant to review and can leave the important bits of a PR description unreviewed.

Oftentimes the "why" is the most important bit for maintainers. The "what" is most important for end users reading along (or for deriving docs or changelog entries from the PR).

### DIFF
--- a/.agents/skills/create-pr/SKILL.md
+++ b/.agents/skills/create-pr/SKILL.md
@@ -70,17 +70,7 @@ Use this skill when turning local work into a GitHub pull request.
 Use this PR body format:
 
 ```markdown
-### What?
-
-<what changed>
-
-### Why?
-
-<why this is needed>
-
-### How?
-
-<implementation approach>
+<what changed and why>
 
 ### Verification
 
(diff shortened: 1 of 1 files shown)


.agents/skills/create-pr/SKILL.md
```diff
@@ -70,17 +70,7 @@ Use this skill when turning local work into a GitHub pull request.
...
-<what changed>
-
-### Why?
-
-<why this is needed>
-
-### How?
-
-<implementation approach>
+<what changed and why>
```
timneutkens, line 75: Would still be helpful to have sections in the template, but maybe having the agent introduce sections by itself would be enough for this as it makes the formatting look better when sharing e.g. screenshots.

eps1lon: sure

lubieowoce (approved):

PR #94899, opened 2026-06-17
docs: GtK, caching with fetch is opt-in - extended to any method or header included

icyJoseph:
Once and for all, a note for LLMs and readers, fetch cache is opt-in. You can cache a POST request, for example a GraphQL request that uses a Bearer token to access the graph, can be cached. 

Training data, and human cognitive association to v13's caching default, seem to add noise here, let's be more authoritative.


aurorascharff (approved):

abhishekmardiya:
I think the documentation should mention that the response **status code** must be `200`.

PR #94902, opened 2026-06-17
Prefetch links nearest the top of the document first

acdlite:
### Why

Most viewport prefetches are scheduled via a shared `IntersectionObserver`. When several links enter the viewport at once (e.g. on initial load), the observer reports them in document order and we scheduled them in that order. Because the prefetch scheduler gives the highest priority to the *most recently scheduled* task, this meant the link **lowest** in the document ended up with the highest priority — the opposite of a sensible default.

### What

Iterate each observer batch in reverse, so the link nearest the **top** of the document is scheduled last and therefore prioritized. The topmost link isn't guaranteed to be the most important, but as a default heuristic it's more reasonable than favoring whichever link happens to be lowest in the document.

Also updates the two hover tests in `prefetch-scheduling`, whose expected ordering was coupled to the old bottom-first default (the set of route trees prefetched within the concurrency limit flips from pages 7,6,5,4 to 1,2,3,4). The behaviors they assert are unchanged, and a new test covers the top-of-document prioritization.

### DIFF
--- a/packages/next/src/client/components/links.ts
+++ b/packages/next/src/client/components/links.ts
@@ -247,7 +247,16 @@ export function unmountPrefetchableInstance(element: Element) {
 }
 
 function handleIntersect(entries: Array<IntersectionObserverEntry>) {
-  for (const entry of entries) {
+  // Process the entries in reverse order. The prefetch scheduler assigns the
+  // highest priority to the most recently scheduled task, so whichever link we
+  // schedule *last* wins. When multiple links enter the viewport at once (e.g.
+  // on initial load), the observer reports them in document order, so iterating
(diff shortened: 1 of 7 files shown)


samselikoff (approved):
Tried it out in a local demo and works great! Happy to see this one getting in 👍

acdlite:
Superseded by #95393. This PR was opened from a fork, so the "Test new and changed tests when deployed" jobs can never pass here — `vercel link` fails with "No existing credentials found" because `secrets.VERCEL_ADAPTER_TEST_TOKEN` isn't available to `pull_request` runs from forks. #95393 is the identical change (rebased onto latest canary) on an upstream branch.

PR #94908, opened 2026-06-17
fix(otel): mark AppRender.fetch span as ERROR for 4xx/5xx upstream responses

### LINKED ISSUE #94901: AppRender.fetch spans never set span status to ERROR for failed (4xx/5xx) responses

https://github.com/anilparlak/otel-repro/tree/main

### To Reproduce

1. git clone https://github.com/anilparlak/otel-repro && cd otel-repro
2. npm install
3. npm run dev
4. In another terminal: curl http://localhost:3000/
5. In the dev-server console, look at the `fetch GET .../api/fail` span:
   despite the upstream HTTP 503, it has status: { code: 0 } (UNSET).

Notes:
- To test 4xx, change the status in app/api/fail/route.ts to 404.
- To see the @vercel/otel path, swap instrumentation.ts with
  instrumentation.vercel.ts (the README documents both variants).

## Current behavior

Next.js's native fetch instrumentation (`patch-fetch.ts`) creates
`AppRender.fetch` spans but never calls
`span.setStatus({ code: SpanStatusCode.ERROR })` when the upstream response
status is 4xx/5xx. OTel-compatible APM backends (Azure Application Insights,
Datadog, Honeycomb, …) that consume these native spans therefore record
failed backend dependencies as successful.

Observed span status (Next 16 canary, @opentelemetry/sdk-trace-node 1.x):

| Path                      | Response | span.status.code |
|---------------------------|----------|------------------|
| native (patch-fetch.ts)   | 503      | 0 (UNSET)        |
| native (patch-fetch.ts)   | 404      | 0 (UNSET)        |
| @vercel/otel              | 503      | 2 (ERROR)        |
| @vercel/otel              | 404      | 0 (UNSET)        |

The native path marks neither 4xx nor 5xx; @vercel/otel marks only
status >= 500, so 4xx is unmarked on every path.

Native-path span (503), abbreviated console output:

````
name: 'fetch GET .../api/fail',

attributes: {

'next.span_type': 'AppRender.fetch',

'http.method': 'GET',

...

},

status: { code: 0 },   // UNSET, despite the upstream HTTP 503
````

### Concrete impact — Azure Application Insights

A server-side fetch returning HTTP 503 is recorded as a successful
dependency (resultCode 0, success True), with customDimensions showing
`next.span_type: AppRender.fetch` and no error status. In the APM UI the
Failures → Dependencies panel is empty, the Application Map shows a 0%
failure rate (looks healthy), and operators cannot diagnose backend errors
from the APM UI alone.

## Expected behavior

For 4xx/5xx responses, the `AppRender.fetch` span should be marked with
`SpanStatusCode.ERROR` so APM backends record the dependency as failed.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.1.0: Mon Oct  9 21:32:52 PDT 2023; root:xnu-10002.41.9~7/RELEASE_ARM64_T8122
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 22.21.1
  npm: 10.9.4
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.3.0-canary.53 // Latest available version is detected (16.3.0-canary.53).
  eslint-config-next: N/A
  react: 19.0.0
  react-dom: 19.0.0
  typescript: 6.0.3
Next.js Config:
  output: N/A
```

Instrumentation

next dev (local)

### Related work

Two existing PRs add a status-code attribute but neither sets the span status to ERROR:

- #57646 — adds the deprecated `http.status_code`
- #84039 — uses the current `http.response.status_code` convention

### Suggested fix

In the same place the status-code attribute is set (`packages/next/src/server/lib/patch-fetch.ts`), also set the span status for failed responses (exact variable name may differ):

```ts
import { SpanStatusCode } from '@opentelemetry/api'

if (response.status >= 400) {
  span.setStatus({
    code: SpanStatusCode.ERROR,
    message: `HTTP ${response.status}`,
  })
}
```

Note this affects `next start` (production) too, not just `next dev`.

tsushanth:
## What's the problem this PR addresses?

Closes #94901.

The native fetch instrumentation in `packages/next/src/server/lib/patch-fetch.ts` creates `AppRender.fetch` spans but never calls `span.setStatus({ code: SpanStatusCode.ERROR })` when the upstream response status is 4xx/5xx. The result is that OTel-compatible APM backends (Azure Application Insights, Datadog, Honeycomb, etc.) that consume these native spans record failed backend dependencies as successful, the failure-rate panels look healthy, and operators cannot diagnose backend errors from the APM UI alone. As the reporter observed on the latest canary against `@opentelemetry/sdk-trace-node` 1.x:

| Path                      | Response | span.status.code (before) | span.status.code (after) |
|---------------------------|----------|---------------------------|--------------------------|
| native (`patch-fetch.ts`) | 503      | 0 (UNSET)                 | 2 (ERROR)                |
| native (`patch-fetch.ts`) | 404      | 0 (UNSET)                 | 2 (ERROR)                |
| `@vercel/otel`            | 503      | 2 (ERROR)                 | 2 (ERROR) — unchanged    |
| `@vercel/otel`            | 404      | 0 (UNSET)                 | 0 (UNSET) — out of scope |

## How does this PR solve the problem?

The trace callback already determines the final `Response` through several return paths (cache hit, revalidate, original fetch). Rather than thread the span through each branch, the callback's body is wrapped in a single async IIFE so every existing return becomes a single resolved value at the outer scope. Once the response is in hand, the wrapper inspects its `status` and calls `span.setStatus({ code: SpanStatusCode.ERROR, ... })` for any 4xx/5xx code. The span's existing `http.url` / `http.method` / `net.peer.*` attributes are unchanged, so only the status field is affected; chat-completion, Responses-API, and other non-fetch code paths are untouched.

## How to verify the change?

The reporter's full repro repo at https://github.com/anilparlak/otel-repro reproduces both the native and `@vercel/otel` paths in a single Next.js project. On `canary` the native `fetch GET .../api/fail` span carries `status: { code: 0 }` despite the upstream 503; with this PR applied that span carries `status: { code: 2 }` with `message: "HTTP 503 Service Unavailable"`. The README in that repo documents both the deterministic console-exporter output and how to swap `instrumentation.ts` with `instrumentation.vercel.ts` to confirm the `@vercel/otel` path is not regressed.

I deliberately scoped this PR to the native-instrumentation gap that the issue raises. The reporter also notes that `@vercel/otel` does not mark 4xx as ERROR; that fix lives upstream in `@vercel/otel` and is intentionally out of scope here.

## Tests added?

No tests in this commit. The single-line behaviour is exercised by the reporter's repro; happy to add an e2e test against `test/e2e/opentelemetry/instrumentation/` (adding a page that does a server-side `fetch` to `/api/app/[param]/status`, which already returns 418, then asserting the matching `AppRender.fetch` span has `status: { code: 2 }`) if reviewers would prefer one before merging. Leaving the PR in draft until that decision is made.

### DIFF
--- a/packages/next/src/server/lib/patch-fetch.ts
+++ b/packages/next/src/server/lib/patch-fetch.ts
@@ -9,6 +9,7 @@ import {
   recordRequestInsightFetch,
 } from './trace/request-insights'
 import { getTracer, SpanKind } from './trace/tracer'
+import { SpanStatusCode } from '@opentelemetry/api'
 import {
   CACHE_ONE_YEAR_SECONDS,
   INFINITE_CACHE,
@@ -351,6 +352,7 @@ export function createPatchedFetcher(
         },
       },
       async (span) => {
+        const response = await (async (): Promise<Response> => {
         // If this is an internal fetch, we should not do any special treatment.
         if (isInternal) {
           return originFetch(input, init)
@@ -1309,6 +1311,18 @@ export function createPatchedFetcher(
         } else {
           return doOriginalFetch(false, cacheReasonOverride)
         }
+        })()
+
+        if (span && response.status >= 400) {
+          span.setStatus({
+            code: SpanStatusCode.ERROR,
+            message: `HTTP ${response.status}${
+              response.statusText ? ` ${response.statusText}` : ''
+            }`,
+          })
+        }
+
+        return response
       }
     )
 


tsushanth:
Rebased onto current upstream canary. Resolved conflict in `patch-fetch.ts` caused by upstream refactoring of the fetch span callback — applied the `SpanStatusCode.ERROR` status-setting logic cleanly onto the new structure.

PR #94912, opened 2026-06-17
[cd] Continue release train if `@next/swc-wasm-*` fails to publish

eps1lon:
https://github.com/vercel/next.js/pull/94622 tried to be more strict with when a release is considered healthy. However, `@next/swc-wasm-web` is not enrolled in Trusted Publishing and blocks the release train.

This PR restores the behavior for `@next/swc-wasm-*` packages prior to https://github.com/vercel/next.js/pull/94622 i.e. we continue the release train even if any `@next/swc-wasm-*` fails. Other NAPI bindings will continue to block the release train if they fail to publish.

The bail vs no-bail behavior is controlled with a repository variable so that we can mor easily test if `@next/swc-wasm-web` is enrolled in Trusted Publishing.


unstubbable (approved):

eps1lon:
Force merging because to break the CI deadlock.

PR #94914, opened 2026-06-17
Sort multi-error Redbox snapshots to deflake the hydration-error ordering test

gaojude:
The "extra node inside Suspense content" hydration test asserts the order of the three errors shown in the dev overlay, but React reports the thrown `Runtime Error` and the two `onRecoverableError` callbacks in a non-deterministic order during hydration, so the inline snapshot flakes on ordering. It's been flaky since 2024-09 (~11% failure rate) and most recently broke canary.

Add a `sortErrors` option to `toDisplayRedbox` that sorts a multi-error snapshot by serialized content before asserting, and enable it for this test (snapshot reordered to the resulting stable order). Single-error snapshots and tests that don't opt in are unaffected.

### DIFF
--- a/test/development/acceptance/hydration-error.test.ts
+++ b/test/development/acceptance/hydration-error.test.ts
@@ -676,19 +676,9 @@ describe('Error overlay for hydration errors in Pages router', () => {
     })
 
     if (isReact18) {
-      await expect(browser).toDisplayRedbox(`
+      await expect(browser).toDisplayRedbox(
+        `
        [
-         {
-           "componentStack": "<Mismatch>
-       >   <div>
-             <Suspense>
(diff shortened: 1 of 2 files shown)


eps1lon (changes requested):
I think this is the correct root cause analysis: https://github.com/vercel/next.js/pull/94555

The flakiness is caused by an actual bug. We should land the fix instead of adding APIs that allows silencing those bugs.

gaojude:
Closing in favor of #94555, which fixes the actual root cause in the dev overlay event queue (FIFO ordering) rather than making the test order-insensitive. The test was correctly catching a real ordering bug — normalizing it here would have masked the bug #94555 fixes.

PR #88107, opened 2026-01-05
Add retry logic for chunk load errors

feedthejim:
## Summary

This PR makes client chunk loading more resilient to transient network failures by retrying failed chunk fetches once at the runtime loader boundary.

That applies to the chunk-loading paths used by:
- `next/dynamic`
- pages-router route chunks
- app-router client-reference chunks loaded after a navigation

The goal is simple: if the browser briefly fails to fetch a JS or CSS chunk, Next.js should make one more attempt before surfacing `ChunkLoadError` to the user.

## Problem

A client navigation or lazy render often depends on one or more chunks that are fetched after the initial page load. If one of those requests fails because of a brief connection reset, a short CDN blip, or another transient network issue, the browser may be able to fetch the exact same asset successfully a moment later.

Today, those failures can still surface as an immediate `ChunkLoadError` even though the underlying problem is temporary.

That shows up in a few common cases:
- a `next/dynamic` component whose chunk fails on the first request
- a pages-router navigation that needs a new route chunk
- an app-router navigation where the RSC payload succeeds, but a referenced client chunk fails to download

In all of those cases, the user-visible failure mode is the same: the app gives up on the first failed chunk fetch even though retrying the same request once would often recover.

## Design

The retry belongs at the chunk-loader boundary, not in a higher-level router or React wrapper.

That gives a few important properties:
- the retry happens exactly where the network request is issued
- React continues to observe a single logical in-flight chunk load instead of layered retry loops
- the same policy can cover dynamic imports, pages-router chunks, and app-router client-reference chunks
- failures that are not chunk fetch failures are left alone

This PR keeps the policy intentionally narrow:
- retry once
- only for chunk load failures
- preserve the existing failure behavior after the retry budget is exhausted

### Shared retry policy

`packages/next/src/client/components/chunk-load-error/retry-chunk-load-error.ts` remains the shared policy entrypoint for chunk retries.

It uses the existing chunk-load error classification and delay utilities, and now tags exhausted errors so nested webpack loader wrappers do not accidentally stack multiple retry loops around the same logical chunk load.

### Webpack

Webpack chunk retry is implemented in a new runtime patch:
- `packages/next/src/client/patch-webpack-chunk-load.ts`

That patch is applied from both client webpack bootstraps:
- `packages/next/src/client/app-webpack.ts`
- `packages/next/src/client/webpack.ts`

The patch wraps both:
- `__webpack_chunk_load__`
- `__webpack_require__.e`

[...]

PR #94918, opened 2026-06-17
[turbopack] Retry chunk loading on failure

sampoder:
This is a trimmed down PR from https://github.com/vercel/next.js/pull/88107, it achieves the same thing (with much of the same code). However, it is Turbopack-only. This is because the Webpack portions of that PR required patching React - we can return to this in a different PR or I guess people could use a plugin? (https://github.com/mattlewis92/webpack-retry-chunk-load-plugin)


turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -100,6 +102,56 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
...
+    }
+    resolver.reject(error)
+  }
+
+  function getChunkLoadRetryDelayMs() {
+    const jitter = Math.floor(
+      Math.random() * (CHUNK_LOAD_RETRY_MAX_JITTER_MS + 1)
+    )
+    return CHUNK_LOAD_RETRY_BASE_DELAY_MS + jitter
+  }
```
bgw, line 126: I don't think the SWC minifier is smart enough to remove these single-callsite functions? Let's just inline them and save a few bytes in the runtime?

bgw (approved):
This LGTM, but it might also be good if we let @sokra take a quick pass on this as well because he had a lot of comments on https://github.com/vercel/next.js/pull/88107

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -100,6 +102,56 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
...
+      resolver.retryAttempts >= CHUNK_LOAD_RETRY_MAX_ATTEMPTS ||
+      chunkResolvers.get(chunkUrl) !== resolver
+    ) {
+      rejectChunkResolver(chunkUrl, resolver, error)
+      return
+    }
+
+    resolver.retryAttempts++
+    setTimeout(() => {
+      if (resolver.resolved || chunkResolvers.get(chunkUrl) !== resolver) {
```
lukesandberg, line 159: how would these cases happen?  leave a comment

(a racing load?)

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -134,7 +186,11 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
         // ignore
       } else if (isJs(chunkUrl)) {
         self.TURBOPACK_NEXT_CHUNK_URLS!.push(chunkUrl)
-        importScripts(chunkUrl)
+        try {
+          importScripts(chunkUrl)
+        } catch (error) {
+          onChunkLoadError(sourceType, chunkUrl, resolver, error as Error)
```
lukesandberg, line 208: we should probably only retry NetworkErrors?  some of the errors reported by `importScripts` (TypeError) are unrecoverable

also do you know the semantics of calling importScripts on a module like `throw new Error()`?

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
lukesandberg: if there is one callsite it should work fine, but worth confirming

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -176,9 +234,15 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
           // There is this edge where the script already failed loading, but we
           // can't detect that. The Promise will never resolve in this case.
```
lukesandberg: stale comment

lukesandberg (approved):

crates/next-core/src/next_client/context.rs
```diff
@@ -483,6 +483,14 @@ pub struct ClientChunkingContextOptions {
     pub style_groups_algorithm: StyleGroupsAlgorithm,
 }
 
+/// Next.js' chunk-load retry policy for the Turbopack browser runtime.
+/// Webpack does not currently support chunk-load retrying.
+const NEXT_CHUNK_LOAD_RETRY: ChunkLoadRetry = ChunkLoadRetry {
+    max_attempts: 1,
```
sokra: The naming is a bit confusing. `max_attempts` -> `max_retry_attempts`

sampoder: oops, yeah, good point.

test/production/chunk-load-failure/chunk-load-failure.test.ts
```diff
@@ -48,15 +50,61 @@ describe('chunk-load-failure', () => {
...
+        )
+      })
+
+      // One initial request that failed + one retry that succeeded.
+      expect(chunkRequestCount).toBe(2)
+      expect(pageError).toBeUndefined()
+    }
+  )
+  /* eslint-enable jest/no-standalone-expect */
+
```
sokra, line 107: You might want to add a test case which first fails to load the chunk and the promise rejects. Then have a retry button which retries to load the next/dynamic component and it succeeds

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -158,7 +214,9 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
           link.crossOrigin = CROSS_ORIGIN
           link.href = chunkUrl
           link.onerror = () => {
-            resolver.reject()
+            // Drop the failed tag so a retry can re-add it cleanly.
+            link.remove()
+            onChunkLoadError(sourceType, chunkUrl, resolver)
```
sokra: That will insert the link tag at a different order in the DOM. That's not ok.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
sokra, line 226: For consistency we might want to add an onerror to these link tags too.

sokra (commented):
The CSS retry is broken. It need to keep the order of link tags on retry

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
sampoder: Based on https://swc.rs/playground, I did minify + compress. It seems like they are inlined.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
sampoder: Yes - re: only retry NetworkErrors.

When it is the same-origin, you get `Error` and the original error message thrown from the web worker. If you're a different origin, you may get a NetworkError based on: https://html.spec.whatwg.org/multipage/webappapis.html#run-a-classic-script

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
sampoder: in theory we would have already retried them before if there was an error though right? like retried them on the previous page, I'm not quite sure what you mean here

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
sampoder: @sokra - I just pushed some code that instead of removing the link replaces it with a new link in the same location, does this fix this issue?

sokra (approved):

PR #94921, opened 2026-06-17
[turbopack] Create `ServiceWorkerEntryModule` and `service_worker_chunk_filename`

sampoder:
This is what will be inserted by the analyzer into a page's module graph: https://github.com/vercel/next.js/pull/94923. This allows `next-api` to discover these modules (see https://github.com/vercel/next.js/pull/94922) and compile them.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/references/service_worker.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/service_worker.rs
@@ -1,0 +1,108 @@
+use anyhow::Result;
+use turbo_rcstr::{RcStr, rcstr};
+use turbo_tasks::{ResolvedVc, Vc};
+use turbo_tasks_hash::{encode_hex, hash_xxh3_hash64};
+use turbopack_core::{
+    chunk::{AsyncModuleInfo, ChunkableModule, ChunkingContext},
+    ident::AssetIdent,
+    module::{Module, ModuleSideEffects},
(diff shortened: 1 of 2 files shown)


turbopack/crates/turbopack-ecmascript/src/references/service_worker.rs
```diff
@@ -0,0 +1,103 @@
...
+        return rcstr!("sw.js");
+    }
+    let slug: String = trimmed
+        .chars()
+        .map(|c| match c {
+            'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_' => c,
+            _ => '-',
+        })
+        .collect();
+    format!("sw-{slug}.js").into()
```
bgw: We need to do something to strongly guarantee this name is unique for the given scope. (e.g. I have two scopes: `/foo/bar` and `/foo-bar`)

I think that should be a hash at the end of the slug.

bgw: Prefer the explicit thing

```suggestion
    let slug: String = trimmed
        .chars()
        .map(|c| match c {
            'a'..='z' | 'A'..='Z' | '0'..='9' | '-' | '_' => c,
            _ => '-',
        })
        .collect();
    RcStr::from(format!("sw-{slug}.js"))
```

bgw (approved):

PR #94922, opened 2026-06-17
[turbopack] Discover `ServiceWorkerEntryModule`s in `next-api` and compile + serve those service workers

sampoder:
This is the big change in the `next-` part of the repo to support this. The code goes through the module graph to find any service workers that need compiling and then compiles them. It also has the code changes to `packages/next/src/server/lib/router-utils/filesystem.ts` that serve these files on their respective routes.

PR #94923, opened 2026-06-17
[turbopack] Discover service workers in the Turbopack analyzer

sampoder:
This PR is the PR that "turns on" the service worker feature, this is because in the analyzer we are now looking for code like:

```
await navigator.serviceWorker.register(new URL('../lib/service-workers/1.js', import.meta.url), ...)
```

And transforms it into registering the URL that Next.js will now serve based on scope.

URL scheme is:

```
scope: "/" -> `sw.js`
scope: "/offline/mode" -> `sw-offline-mode.js`
```

etc.

It also inserts the `ServiceWorkerEntryModule`s so they can be discovered by `next-api` (see https://github.com/vercel/next.js/pull/94922).

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/references/mod.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/mod.rs
@@ -41,6 +41,7 @@ use num_traits::Zero;
 use parking_lot::Mutex;
 use regex::Regex;
 use rustc_hash::{FxHashMap, FxHashSet};
+use service_worker::ServiceWorkerAssetReference;
 use swc_core::{
     atoms::{Atom, Wtf8Atom, atom},
     common::{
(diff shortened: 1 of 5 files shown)


turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -2045,6 +2046,63 @@ where
     }
 
     match func {
+        WellKnownFunctionKind::ServiceWorkerRegister => {
```
bgw: Let's move this near the end of the match block because it's a less common format of imports, could go near web workers?

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -2045,6 +2046,63 @@ where
...
+                                    ObjectPart::KeyValue(
+                                        JsValue::Constant(JsConstantValue::Str(key)),
+                                        JsValue::Constant(JsConstantValue::Str(value)),
+                                    ) if key.as_str() == "scope" => Some(value.as_str().into()),
+                                    _ => None,
+                                })
+                            }
+                            _ => None,
+                        })
+                        .unwrap_or_else(|| rcstr!("/"));
```
bgw: We shouldn't silently fall back to a `None` scope if we fail to analyze it, we should generate a diagnostic?

turbopack/crates/turbopack-ecmascript/src/references/service_worker.rs
```diff
@@ -106,3 +126,165 @@ impl EcmascriptChunkPlaceable for ServiceWorkerEntryModule {
...
+        for (request_key, item) in result_ref.primary.iter() {
+            match item {
+                ModuleResolveResultItem::Module(module) => {
+                    let marker = ServiceWorkerEntryModule {
+                        inner: *module,
+                        scope: self.scope.clone(),
+                    }
+                    .cell()
+                    .to_resolved()
+                    .await?;
```
bgw:
```suggestion
                    let marker = ServiceWorkerEntryModule {
                        inner: *module,
                        scope: self.scope.clone(),
                    }
                    .resolved_cell();
```

bgw (approved):

PR #94928, opened 2026-06-18
[CC] Remove invariant breaking output: export

gaearon:
This removes the invariant that caused `output: 'export'` to not work with `cacheComponents` at all.

The intention is to enable it for things that are actually fully static, or to err otherwise.

A similar patch got my blog running, though I haven't tried rebuilding Next from source.

### DIFF
--- a/test/e2e/app-dir/cache-components-output-export-error/cache-components-output-export-error.test.ts
+++ b/test/e2e/app-dir/cache-components-output-export-error/cache-components-output-export-error.test.ts
@@ -1,0 +1,23 @@
+import { isNextStart, nextTestSetup } from 'e2e-utils'
+
+describe('cache-components-output-export-error', () => {
+  const { next } = nextTestSetup({
+    files: __dirname,
+    skipStart: true,
+    skipDeployment: true,
+  })
+
+  it('fails the build, naming the route, when a route cannot be fully prerendered', async () => {
+    if (!isNextStart) {
+      // The enforcement is a build-time concern; nothing to assert in dev.
+      return
+    }
+
+    const { exitCode, cliOutput } = await next.build()
+
+    expect(cliOutput).toContain(
+      'Route "/dynamic-route" could not be statically exported because it uses dynamic or uncached data.'
+    )
+    expect(exitCode).toBe(1)
+  })
+})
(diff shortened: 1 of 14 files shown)


test/e2e/app-dir/cache-components-output-export/app/feed/route.ts
```diff
@@ -0,0 +1,12 @@
+export function generateStaticParams() {
+  return [{}]
+}
```
unstubbable, line 3: This route doesn't have any params so that feels like a hack. And I think it masks the fact that we need more work to allow Cache Components with `output: 'export'` mode. If you omit this, we error with `Error: export const dynamic = "force-static"/export const revalidate not configured on route "/feed" with "output: export".` which is a misleading error message because those configs are forbidden when `cacheComponents` is enabled. Using `generateStaticParams` instead is not obvious and feels like a dirty workaround.

test/e2e/app-dir/cache-components-output-export-error/cache-components-output-export-error.test.ts
```diff
@@ -0,0 +1,23 @@
+import { isNextStart, nextTestSetup } from 'e2e-utils'
+
+describe('cache-components-output-export-error', () => {
```
unstubbable, line 3: I think this test suite should also have a the case where a page uses params without `generateStaticParams`, e.g.:

```jsx
import { Suspense } from 'react'

export default function Page({
  params,
}: {
  params: Promise<{ slug: string }>
}) {
  return (
    <Suspense fallback={<p>Loading...</p>}>
      {params.then(({ slug }) => (
        <p>{slug}</p>
      ))}
    </Suspense>
  )
}
```

Without Cache Components we would error the build with `Error: Page "/with-params/[slug]" is missing "generateStaticParams()" so it cannot be used with "output: export" config.`

But with Cache Components (and the changes from this PR) we would build the partial shell without any errors, which is wrong I think.

There's also a related issue: Even if you define `generateStaticParams`, we would still generate a fallback shell for unknown params, which is kinda equivalent to using `export const dynamicParams = true`, which errored in the old world with `Error: "dynamicParams: true" cannot be used with "output: export".` Should we skip prerendering fallback shells in that mode?

---

PS: You can build individual pages in such a test suite with:

```js
await next.build({
  args: ['--debug-build-paths', 'app/page.tsx'],
})
```

gaearon:
proper approach in https://github.com/vercel/next.js/pull/95485

PR #94942, opened 2026-06-18
fix(lru-cache): include URL key length in getItemsLru size accounting

### LINKED ISSUE #94890: getItemsLru size accounting omits URL key length, causing excessive heap retention in long-lived servers

https://github.com/dmchoi77/nextjs-getitemslru-key-length-repro

### To Reproduce

1. Clone the reproduction repository and install dependencies.
2. Run `npm run build` and `npm start`.
3. In another terminal, run `NEXT_LRU_REPRO_REQUESTS=1200000 node test-memory-leak.mjs`.
4. Observe the `next start` process heap over the run.
5. Take a heap snapshot of the `next start` process. `LRUNode` instances accumulate up to roughly 1,048,576 entries; each retained node holds its URL path key string plus node overhead.

The same pattern applies to production apps on `15.5.11` or later, including current canary as of 2026-06-11. The `if (!value) return 1` fix from #89040 must be present, because this report is about the residual issue that survives that fix.

Production shape where this is easiest to observe:

1. Run a long-lived self-hosted server, such as AWS ECS, a VM, or another deployment where the process uptime is measured in days rather than minutes.
2. Generate traffic with a large number of unique dynamic-route URLs, especially long URL paths such as i18n template detail paths with long percent-encoded slugs or bot-crawled URLs with varied tails.
3. Observe `process.memoryUsage().heapUsed` or `runtime.node.mem.heap_used` over multiple days.
4. Take a heap snapshot. `LRUNode` instances accumulate up to roughly 1,048,576 entries; each retained node holds its URL path key string plus node overhead.

**Current behavior**

In current canary as of 2026-06-11 and the latest npm release I checked that day (`next@16.2.9`), `getItemsLru` in `packages/next/src/server/lib/router-utils/filesystem.ts` is constructed with `maxSize = 1024 * 1024` and a size function shaped as `length(value)`.

That function returns:

- `1` for null negative-cache entries
- `(value.fsPath || '').length + value.itemPath.length + value.type.length` for positive entries

The URL path key, which is often the dominant memory cost per entry, is not counted in the LRU size accounting.

The current `LRUCache` implementation in `packages/next/src/server/lib/lru-cache.ts` also only calls the size function with the value:

```ts
const size = this.calculateSize?.(value) ?? 1
```

This means the cache can report tiny sizes while retaining much larger key strings:

| Quantity | Reported `size` | Actual retained cost per entry |
| --- | --- | --- |
| Negative cache entry, `set(itemKey, null)` | `1` | key string + `LRUNode` overhead |
| Positive cache entry | roughly `30-100` | key string + value strings + `LRUNode` overhead |

Under traffic with many unique long URL paths, the cache can grow toward its 1,048,576-entry cap. At that ceiling, a single LRU instance can retain hundreds of MB of heap even though the configured `maxSize` is nominally `1024 * 1024`.

**Expected behavior**

`maxSize = 1024 * 1024` should include the dominant per-entry string cost, matching the v14 key-aware size approximation before the custom `LRUCache` migration. In that version, the cache size function received `(value, key)` and included key length:

```js
length(value, key) {
  if (!value) return key?.length || 0
  return (
    (key || '').length +
    (value.fsPath || '').length +
    value.itemPath.length +
    value.type.length
  )
}
```

The current custom `LRUCache` should pass `key` to `calculateSize`, and `getItemsLru` should include the key length in its size calculation.

```bash
Operating System: Linux (Docker, AWS ECS Fargate)
Binaries:
  Node: 22.x
Relevant Packages:
  next: 15.5.12 / 15.5.18 reproduce in production
  current canary as of 2026-06-11 and next@16.2.9 still have the same key-accounting behavior
  react: 19.x
Deployment:
  Self-hosted, long-lived containers with uptime measured in days
```

Performance, Dynamic Routes

Other (Deployed), next start (local)

### Additional context

This was observed in production on a long-running self-hosted Next.js 15.5.x deployment with 3 ECS tasks and no restarts over the observation window.

Short-lived or serverless deployments may not observe this before instance recycling. The primary affected shape is a long-lived self-hosted `next start` process that sees a high number of unique dynamic-route URLs.

7-day trend:

| Metric | Day 0 | Day 7 | Delta |
| --- | --- | --- | --- |
| `runtime.node.mem.heap_used` avg | 291 MB | 445 MB | +154 MB |
| `runtime.node.mem.heap_used` max | 304 MB | 470 MB | +166 MB |
| `runtime.node.mem.external` | 37 MB | 40 MB | flat |
| `runtime.node.process.uptime` | 12,327 s | 612,150 s | no restart |

Key observations:

- External memory was flat, so growth was in V8 heap rather than native memory.
- Major GC ran frequently, but retained objects survived GC.
- All 3 hosts showed similar monotonic heap growth.
- Heap growth matched accumulation of unique route-cache keys rather than total request volume.

The top retainer path in a Datadog Continuous Profiler A-to-B comparison pointed at:

```text
resolveRoutes
  -> handleRoute
    -> getItem (filesystem.ts)
      -> set (lru-cache.ts)
```

Other URL-related allocations appeared in profiles but were reclaimed by GC in local reproduction. The `lru-cache.set` path was the one producing surviving `LRUNode` instances.

#### Root cause

The key contribution to size accounting appears to have been dropped during the migration from the bundled `lru-cache` library to the self-built implementation. The old size function received `(value, key)`, while the new custom `LRUCache` stores the key on `LRUNode` but only passes `value` into `calculateSize`.

PR #89040 fixed the unbounded `size = 0` negative-cache case by making null entries return `1`, but it did not restore key-aware size accounting. As a result, the cache is now bounded by entry count for negative-cache entries rather than by the retained memory represented by those entries.

Related context:

- #71455 introduced the custom `LRUCache` migration where the key-aware `length(value, key)` behavior appears to have been dropped.
- #82633 later rewrote the custom LRU as a doubly-linked list while preserving the `calculateSize(value)` shape.
- #89033 reported the original size-0 unbounded growth.
- #89040 fixed the size-0 case by requiring positive sizes and making null filesystem cache entries return `1`, but did not restore key-aware size accounting.

#### Suggested fix

Restore key-aware size accounting by passing `key` through the `LRUCache.set` -> `calculateSize` chain:

```diff
// packages/next/src/server/lib/lru-cache.ts
- private readonly calculateSize: ((value: T) => number) | undefined
+ private readonly calculateSize: ((value: T, key: string) => number) | undefined

...

- calculateSize?: (value: T) => number,
+ calculateSize?: (value: T, key: string) => number,

...

- const size = this.calculateSize?.(value) ?? 1
+ const size = this.calculateSize?.(value, key) ?? 1
```

Then include the key length in `getItemsLru`:

```diff
// packages/next/src/server/lib/router-utils/filesystem.ts
- function length(value) {
+ function length(value, key) {
+   const keyLen = key.length
    if (!value) {
      // Null entries (negative cache) still need a non-zero size for LRU eviction
-     return 1
+     return keyLen || 1
    }
    return (
+     keyLen +
      (value.fsPath || '').length +
      value.itemPath.length +
      value.type.length
    )
  }
```

Existing one-argument `calculateSize` callbacks should continue to work with the wider callback signature.

sleitor:
## What?

The `calculateSize` callback in `LRUCache` only received the cached value, omitting the key (URL string). In the filesystem router, the URL key – which can be long in path-heavy apps – was never counted toward the cache's memory budget.

## Why?

This causes under-eviction: the LRU cache believes entries are smaller than they really are, so it evicts too infrequently, leading to excessive heap retention in long-lived servers handling many unique URLs.

## How?

1. **`lru-cache.ts`** — Extend the `calculateSize` callback signature to pass `key` as a second argument. This is fully backward-compatible: all existing callers that only use `value` continue to work unchanged (TypeScript allows functions with fewer parameters to satisfy a type requiring more).

2. **`filesystem.ts`** — Update the `getItemsLru` size function to include `key.length` for both positive cache entries and negative (null) entries, so the full URL string is counted toward the cache budget.

Fixes #94890

### DIFF
--- a/packages/next/src/server/lib/router-utils/filesystem.ts
+++ b/packages/next/src/server/lib/router-utils/filesystem.ts
@@ -113,12 +113,14 @@ export async function setupFsCheck(opts: {
   config: NextConfigRuntime
 }) {
   const getItemsLru = !opts.dev
-    ? new LRUCache<FsOutput | null>(1024 * 1024, function length(value) {
+    ? new LRUCache<FsOutput | null>(1024 * 1024, function length(value, key) {
         if (!value) {
-          // Null entries (negative cache) still need a non-zero size for LRU eviction
-          return 1
+          // Null entries (negative cache) still need a non-zero size for LRU eviction.
+          // Include the URL key length to accurately account for heap usage.
+          return 1 + key.length
         }
         return (
+          key.length +
           (value.fsPath || '').length +
           value.itemPath.length +
           value.type.length
--- a/packages/next/src/server/lib/lru-cache.ts
+++ b/packages/next/src/server/lib/lru-cache.ts
@@ -49,12 +49,14 @@ export class LRUCache<T> {
   private readonly tail: SentinelNode<T>
   private totalSize: number = 0
   private readonly maxSize: number
-  private readonly calculateSize: ((value: T) => number) | undefined
+  private readonly calculateSize:
+    | ((value: T, key: string) => number)
+    | undefined
   private readonly onEvict: ((key: string, value: T) => void) | undefined
 
   constructor(
     maxSize: number,
-    calculateSize?: (value: T) => number,
+    calculateSize?: (value: T, key: string) => number,
     onEvict?: (key: string, value: T) => void
   ) {
     this.maxSize = maxSize
@@ -124,7 +126,7 @@ export class LRUCache<T> {
    * - O(k) where k is the number of items evicted (can be O(N) for variable sizes)
    */
   public set(key: string, value: T): boolean {
-    const size = this.calculateSize?.(value) ?? 1
+    const size = this.calculateSize?.(value, key) ?? 1
     if (size <= 0) {
       throw new Error(
         `LRUCache: calculateSize returned ${size}, but size must be > 0. ` +


packages/next/src/server/lib/router-utils/filesystem.ts
```diff
@@ -113,12 +113,14 @@ export async function setupFsCheck(opts: {
...
+    ? new LRUCache<FsOutput | null>(1024 * 1024, function length(value, key) {
         if (!value) {
-          // Null entries (negative cache) still need a non-zero size for LRU eviction
-          return 1
+          // Null entries (negative cache) still need a non-zero size for LRU eviction.
+          // Include the URL key length to accurately account for heap usage.
+          return 1 + key.length
         }
         return (
+          key.length +
```
eps1lon, line 123: I'm a bit concerned here that we now cache less even though the existing heap usage was apparently fine. We should investigate how much key space we currently use and adjust the size accordingly.

sleitor: Good point — including the key length will cause the LRU to evict more aggressively, so the cache will hold fewer entries for the same size budget.

For the fix to be correct we'd want to either (a) investigate how much key space is actually consumed in production and adjust the `maxSize` upward to compensate, or (b) include the key length and update the tests to reflect the new effective capacity.

The issue reporter (#94890) described heap growth under a large unique-URL workload, which suggested the budget was being underestimated, but you're right that we should verify that with data rather than assuming.

Do you have a suggested approach for measuring key-space usage, or would a `maxSize` adjustment alongside the key-length accounting work for you?

eps1lon: Looking at actual production data. I'm not entirely sure this isn't purely a theoretical issue at this point.

sleitor: That's really valuable — if production key space shows URL lengths are small relative to the value sizes, the accounting gap may indeed be negligible. 

Happy to wait on your findings. If the data suggests this is purely theoretical, I'd rather close the PR than merge a change that makes the cache more aggressive without a measurable benefit. Alternatively, if there's a threshold or target max URL length you'd like us to guard against, I can adjust the implementation accordingly.

eps1lon: I won't have time to dig into this. As far as we can tell, this is not an issue. If you have observed otherwise in real world data, you should share that data.

sleitor:
Closing per @eps1lon's note — production data doesn't show this as a real issue. Thanks for checking!

PR #94943, opened 2026-06-18
fix(router): preserve scroll:false across async PPR retry navigations (force-dynamic)

### LINKED ISSUE #94893: `scroll: false` not preserved across async PPR retry navigations (force-dynamic pages)

https://github.com/cluehr/nextjs-scroll-issue-webapp

### To Reproduce

1. Build and start in production mode: `npm i && npm run prod`
2. Open http://localhost:3000.
3. Scroll down to the **Filters** section at the bottom.
4. Tick one colour checkbox — the page scrolls back to the top. ❌

## Current behaviour
When clicking one of the checkboxes the URL is changed. We use `router.replace` with option `scroll: false` to prevent scrolling to the top of the page and preserve the scroll position. Unfortunately, `scroll: false` gets lost and the page is scrolled to the top.

## Expected behaviour
Next.js 16 should behave the same as Next.js 15 did and respect the `scroll: false` option.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 36864
  Available CPU cores: 12
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: 4.14.1
  pnpm: 9.15.3
Relevant Packages:
  next: 16.3.0-canary.52 // No Next.js version data was found.
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Dynamic Routes, Linking and Navigating, Partial Prerendering (PPR)

Other (Deployed)

### Additional context

The issue appeared for us, when trying to upgrade our app from next.js 15.5.8 to 16.1.6. We digged a little into this and found the following.

### The buggy line

`server-patch-reducer.js` (Next.js 16.1.6) hardcodes scroll-to-top on every retry navigation:

```js
// server-patch-reducer.js (unpatched)
const shouldScroll = true;  // ← always resets scroll
```

The `shouldScroll` value from the original `router.replace(..., { scroll: false })` call is
never threaded through to the retry, so `scroll: false` is silently discarded.

### How the retry is triggered

On a `force-dynamic` page, every navigation goes through the segment cache. When there is no
exact prefetch cache hit, Next.js checks for an *optimistic* route — a cache entry for the same
pathname but with an empty search string. If one exists, the navigation takes the optimistic
path:

```
router.replace('?color=red', { scroll: false })
  → navigate()
    → requestOptimisticRouteCacheEntry()   // succeeds if '/' was prefetched
      → navigateUsingPrefetchedRouteTree()
        → spawnDynamicRequests()           // spawns a live server fetch
          → finishNavigationTask()         // awaits the response
            → writeDynamicDataIntoNavigationTask()
              // server returns different tree (force-dynamic re-rendered with ?color=red)
              → didReceiveUnknownParallelRoute = true  OR  abortRemainingPendingTasks = 1
                → dispatchRetryDueToTreeMismatch()
                  → ACTION_SERVER_PATCH
                    → server-patch-reducer (shouldScroll = true) ❌
```

The optimistic route cache entry for `/` is populated by Next.js when the user hovers over a
`<Link href="/">` (Next.js prefetches linked routes on hover/viewport entry). Without that
prefetch, every navigation falls through to `navigateDynamicallyWithNoPrefetch`, which fetches
inline and never reaches `finishNavigationTask` — so the bug is not triggered.

This is why the bug only manifests after interacting with the page (triggering a prefetch), and
why it does not reproduce in `next dev` (prefetching is disabled in development).

### Why the second checkbox tick, not the first

The first tick navigates from `/` → `?color=red`. The optimistic route is built from the cached
`/` entry. The server responds with the `?color=red` tree, which differs from the optimistic
tree (because `force-dynamic` re-renders on every navigation) → mismatch → retry with
`shouldScroll = true` → scroll reset.

The same mechanism fires on every subsequent tick.

The sample webapp already reproduces the issue on the first click.

For our real webapp we added applied a patch to fix the issue.

## Patch we applied

We patched three files in `dist/`:

```diff
diff --git a/dist/client/components/router-reducer/ppr-navigations.js b/dist/client/components/router-reducer/ppr-navigations.js
--- a/dist/client/components/router-reducer/ppr-navigations.js
+++ b/dist/client/components/router-reducer/ppr-navigations.js
@@ -821,7 +821,9 @@ function spawnDynamicRequests(task, primaryUrl, nextUrl, freshnessPolicy, accumu
 routeCacheEntry, // The original navigation's push/replace intent. Threaded through to the
 // server-patch retry logic so it can inherit the intent if the original
 // transition hasn't committed yet.
-navigateType) {
+navigateType, // The original navigation's scroll behavior. Threaded through to preserve
+// scroll: false across async retry navigations.
+scrollBehavior) {
     const dynamicRequestTree = task.dynamicRequestTree;
     if (dynamicRequestTree === null) {
         // This navigation was fully cached. There are no dynamic requests to spawn.
@@ -878,12 +880,12 @@ navigateType) {
     }
     // Further async operations are moved into this separate function to
     // discourage sequential network requests.
-    const voidPromise = finishNavigationTask(task, nextUrl, primaryRequestPromise, refreshRequestPromises, routeCacheEntry, navigateType);
+    const voidPromise = finishNavigationTask(task, nextUrl, primaryRequestPromise, refreshRequestPromises, routeCacheEntry, navigateType, scrollBehavior);
     // `finishNavigationTask` is responsible for error handling, so we can attach
     // noop callbacks to this promise.
     voidPromise.then(noop, noop);
 }
-async function finishNavigationTask(task, nextUrl, primaryRequestPromise, refreshRequestPromises, routeCacheEntry, navigateType) {
+async function finishNavigationTask(task, nextUrl, primaryRequestPromise, refreshRequestPromises, routeCacheEntry, navigateType, scrollBehavior) {
     // Wait for all the requests to finish, or for the first one to fail.
     let exitStatus = await waitForRequestsToFinish(primaryRequestPromise, refreshRequestPromises);
     // Once the all the requests have finished, check the tree for any remaining
@@ -910,7 +912,7 @@ async function finishNavigationTask(task, nextUrl, primaryRequestPromise, refres
                 // happen in a row, fall back to a hard retry.
                 const isHardRetry = false;
                 const primaryRequestResult = await primaryRequestPromise;
-                dispatchRetryDueToTreeMismatch(isHardRetry, primaryRequestResult.url, nextUrl, primaryRequestResult.seed, task.route, routeCacheEntry, navigateType);
+                dispatchRetryDueToTreeMismatch(isHardRetry, primaryRequestResult.url, nextUrl, primaryRequestResult.seed, task.route, routeCacheEntry, navigateType, scrollBehavior);
                 return;
             }
         case 2:
@@ -925,7 +927,7 @@ async function finishNavigationTask(task, nextUrl, primaryRequestPromise, refres
                 // doesn't exist yet.
                 const isHardRetry = true;
                 const primaryRequestResult = await primaryRequestPromise;
-                dispatchRetryDueToTreeMismatch(isHardRetry, primaryRequestResult.url, nextUrl, primaryRequestResult.seed, task.route, routeCacheEntry, navigateType);
+                dispatchRetryDueToTreeMismatch(isHardRetry, primaryRequestResult.url, nextUrl, primaryRequestResult.seed, task.route, routeCacheEntry, navigateType, scrollBehavior);
                 return;
             }
         default:
@@ -978,7 +980,8 @@ function dispatchRetryDueToTreeMismatch(isHardRetry, retryUrl, retryNextUrl, see
 // prediction. If the navigation results in a mismatch, we mark it as having
 // a dynamic rewrite so future predictions bail out.
 routeCacheEntry, // The original navigation's push/replace intent.
-originalNavigateType) {
+originalNavigateType, // The original navigation's scroll behavior.
+originalScrollBehavior) {
     // If the navigation used a route prediction, mark it as having a dynamic
     // rewrite since it resulted in a mismatch.
     if (routeCacheEntry !== null) {
@@ -1026,7 +1029,8 @@ originalNavigateType) {
         nextUrl: retryNextUrl,
         seed,
         mpa: isHardRetry,
-        navigateType: retryNavigateType
+        navigateType: retryNavigateType,
+        scrollBehavior: originalScrollBehavior
     };
     (0, _useactionqueue.dispatchAppRouterAction)(retryAction);
 }

diff --git a/dist/client/components/router-reducer/reducers/server-patch-reducer.js b/dist/client/components/router-reducer/reducers/server-patch-reducer.js
--- a/dist/client/components/router-reducer/reducers/server-patch-reducer.js
+++ b/dist/client/components/router-reducer/reducers/server-patch-reducer.js
@@ -41,7 +41,9 @@ function serverPatchReducer(state, action) {
     // using the tree we just received from the server.
     const retryCanonicalUrl = (0, _createhreffromurl.createHrefFromUrl)(retryUrl);
     const retryNextUrl = action.nextUrl;
-    const scrollBehavior = _routerreducertypes.ScrollBehavior.Default;
+    // Preserve the original scroll behavior from the navigation that triggered the retry.
+    // This ensures that scroll: false is respected across async retry navigations.
+    const scrollBehavior = action.scrollBehavior !== undefined ? action.scrollBehavior : _routerreducertypes.ScrollBehavior.Default;
     const now = Date.now();
     return (0, _navigation.navigateToKnownRoute)(now, state, retryUrl, retryCanonicalUrl, retrySeed, currentUrl, currentRenderedSearch, state.cache, state.tree, _pprnavigations.FreshnessPolicy.RefreshAll, retryNextUrl, scrollBehavior, navigateType, null, // Server patch (retry) navigations don't use route prediction. This is
     // typically a retry after a previous mismatch, so the route was already

diff --git a/dist/client/components/segment-cache/navigation.js b/dist/client/components/segment-cache/navigation.js
--- a/dist/client/components/segment-cache/navigation.js
+++ b/dist/client/components/segment-cache/navigation.js
@@ -144,7 +144,7 @@ routeCacheEntry) {
     const task = (0, _pprnavigations.startPPRNavigation)(now, currentUrl, currentRenderedSearch, currentCacheNode, currentFlightRouterState, navigationSeed.routeTree, navigationSeed.metadataVaryPath, freshnessPolicy, navigationSeed.data, navigationSeed.head, navigationSeed.dynamicStaleAt, isSamePageNavigation, accumulation);
     if (task !== null) {
         if (freshnessPolicy !== _pprnavigations.FreshnessPolicy.Gesture) {
-            (0, _pprnavigations.spawnDynamicRequests)(task, url, nextUrl, freshnessPolicy, accumulation, routeCacheEntry, navigateType);
+            (0, _pprnavigations.spawnDynamicRequests)(task, url, nextUrl, freshnessPolicy, accumulation, routeCacheEntry, navigateType, scrollBehavior);
         }
         return completeSoftNavigation(state, url, nextUrl, task.route, task.node, navigationSeed.renderedSearch, canonicalUrl, navigateType, scrollBehavior, accumulation.scrollRef, debugInfo);
     }
```

sleitor:
## What?

Thread `scrollBehavior` through the async PPR retry chain so that `scroll: false` passed to `router.replace()` (or `router.push()`) is preserved when a force-dynamic page triggers a tree mismatch and the navigation falls back to `ACTION_SERVER_PATCH`.

## Why?

On force-dynamic pages with PPR enabled, every navigation that hits an optimistic route cache entry ends up in a tree mismatch → `dispatchRetryDueToTreeMismatch` → `serverPatchReducer`. Before this fix, `serverPatchReducer` hardcoded `ScrollBehavior.Default` regardless of the original navigation's `scroll: false` option, causing the page to scroll back to the top on every navigation.

The bug only triggers in production (prefetching is disabled in dev) and only after the user has prefetched the root route (e.g. by hovering over a `<Link href="/">`), which is why it's subtle.

## How?

1. Add optional `scrollBehavior?: ScrollBehavior` to `ServerPatchAction` in `router-reducer-types.ts`.
2. Add `scrollBehavior` parameter to `spawnDynamicRequests()` (after `navigationLock`).
3. Thread it through: `spawnDynamicRequests` → `finishNavigationTask` → `dispatchRetryDueToTreeMismatch` → `ServerPatchAction.scrollBehavior`.
4. In `serverPatchReducer`, read `action.scrollBehavior ?? ScrollBehavior.Default` instead of always using `Default`.
5. Update `restore-reducer.ts` (the only other caller of `spawnDynamicRequests`) to pass `ScrollBehavior.Default` — history traversal already defaults to no-scroll behavior.

Fixes #94893

### DIFF
--- a/packages/next/src/client/components/router-reducer/ppr-navigations.ts
+++ b/packages/next/src/client/components/router-reducer/ppr-navigations.ts
@@ -16,6 +16,7 @@ import type { NavigationLockState } from '../segment-cache/navigation-testing-lo
 import { createHrefFromUrl } from './create-href-from-url'
 import { fetchServerResponse } from './fetch-server-response'
 import { dispatchAppRouterAction } from '../use-action-queue'
+import type { ScrollBehavior } from './router-reducer-types'
 import {
   ACTION_SERVER_PATCH,
   type ServerPatchAction,
@@ -1417,6 +1418,9 @@ export function spawnDynamicRequests(
   // transition hasn't committed yet.
   navigateType: 'push' | 'replace',
   navigationLock: NavigationLock,
+  // The original navigation's scroll behavior. Threaded through to preserve
+  // scroll: false across async PPR retry navigations.
+  scrollBehavior: ScrollBehavior,
   signal: AbortSignal | undefined
 ): void {
   const dynamicRequestTree = task.dynamicRequestTree
@@ -1512,7 +1516,8 @@ export function spawnDynamicRequests(
     primaryRequestPromise,
     refreshRequestPromises,
     routeCacheEntry,
-    navigateType
+    navigateType,
+    scrollBehavior
   )
   // `finishNavigationTask` is responsible for error handling, so we can attach
   // noop callbacks to this promise.
@@ -1527,7 +1532,8 @@ async function finishNavigationTask(
     ReturnType<typeof fetchMissingDynamicData>
   > | null,
   routeCacheEntry: FulfilledRouteCacheEntry | null,
-  navigateType: 'push' | 'replace'
+  navigateType: 'push' | 'replace',
+  scrollBehavior: ScrollBehavior
 ): Promise<void> {
   // Wait for all the requests to finish, or for the first one to fail.
   let exitStatus = await waitForRequestsToFinish(
@@ -1575,6 +1581,7 @@ async function finishNavigationTask(
         task.route,
         routeCacheEntry,
         navigateType,
+        scrollBehavior,
         FreshnessPolicy.RefreshAll
       )
       return
@@ -1594,6 +1601,7 @@ async function finishNavigationTask(
         task.route,
         routeCacheEntry,
         navigateType,
+        scrollBehavior,
         FreshnessPolicy.HistoryTraversal
       )
       return
@@ -1617,6 +1625,7 @@ async function finishNavigationTask(
         task.route,
         routeCacheEntry,
         navigateType,
+        scrollBehavior,
         FreshnessPolicy.RefreshAll
       )
       return
@@ -1688,6 +1697,9 @@ function dispatchRetryDueToTreeMismatch(
   routeCacheEntry: FulfilledRouteCacheEntry | null,
   // The original navigation's push/replace intent.
   originalNavigateType: 'push' | 'replace',
+  // The original navigation's scroll behavior. Threaded through to preserve
+  // scroll: false across async PPR retry navigations.
+  originalScrollBehavior: ScrollBehavior,
   // Freshness policy for the retry navigation. `RefreshAll` re-fetches the
   // tree's dynamic data (used for genuine tree mismatches). `HistoryTraversal`
   // reuses the data already in the tree (used when only the URL needs
@@ -1762,6 +1774,7 @@ function dispatchRetryDueToTreeMismatch(
     seed,
     mpa: isHardRetry,
     navigateType: retryNavigateType,
+    scrollBehavior: originalScrollBehavior,
     freshnessPolicy: retryFreshnessPolicy,
   }
   dispatchAppRouterAction(retryAction)
--- a/packages/next/src/client/components/router-reducer/reducers/restore-reducer.ts
+++ b/packages/next/src/client/components/router-reducer/reducers/restore-reducer.ts
@@ -3,6 +3,7 @@ import type {
   ReducerState,
   RestoreAction,
 } from '../router-reducer-types'
+import { ScrollBehavior } from '../router-reducer-types'
 import { extractPathFromFlightRouterState } from '../compute-changed-path'
 import {
   FreshnessPolicy,
@@ -95,6 +96,8 @@ export function restoreReducer(
     // History traversal always uses 'replace'.
     'replace',
     navigationLock,
+    // History traversal preserves the current scroll position by default.
+    ScrollBehavior.Default,
     // Not an HMR refresh, so there's no request generation to cancel.
     undefined
   )
--- a/packages/next/src/client/components/router-reducer/reducers/server-patch-reducer.ts
+++ b/packages/next/src/client/components/router-reducer/reducers/server-patch-reducer.ts
@@ -47,7 +47,13 @@ export function serverPatchReducer(
   // (`HistoryTraversal`), since the data we received is correct.
   const retryCanonicalUrl = createHrefFromUrl(retryUrl)
   const retryNextUrl = action.nextUrl
-  const scrollBehavior = ScrollBehavior.Default
+  // Preserve the original scroll behavior from the navigation that triggered
+  // the retry. This ensures that scroll: false is respected across async PPR
+  // retry navigations (e.g. force-dynamic pages).
+  const scrollBehavior =
+    action.scrollBehavior !== undefined
+      ? action.scrollBehavior
+      : ScrollBehavior.Default
   const navigationLock = getCurrentNavigationLock()
   const now = Date.now()
   return navigateToKnownRoute(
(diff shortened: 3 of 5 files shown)


sleitor:
Rebased onto latest canary — fix still applies. Ready for review. 🙏

PR #94948, opened 2026-06-18
Turbopack: aggregate server HMR into one subscription

wbinnssmith:
This replaces per-chunk Server HMR turbo tasks with a single firehose subscription that diffs every HMR chunk.  This significantly cuts the number of tokio task churn on projects with many server chunks and centralizes the diff/clear logic. It leads to a multi-second saving in both cold and warm builds in a large app.

This PR also renames `clear()` to `reEvaluateAllModulesExpensive()` to label directly that this is a costly operation and should only be done in exceptional circumstances.
    
A following PR will bring this to client chunks.
    
Rust:
  - New `aggregate_hmr` module: `AggregateHmrVersion` keyed by chunk path, `merged_partial_update` builder, and `is_hmr_eligible_chunk` (excludes `.map` files, which would force every diff to `Total`).
  - `Project::all_hmr_version_state` / `all_hmr_update` aggregate over the whole `hmr_root_path`. The seed transition emits an empty `Partial` so the JS consumer doesn't treat it as a restart and wipe handlers the triggering request just populated. Any chunk requiring `Total`/`Missing` escalates the batch.
  - `VersionedContentMap::hmr_chunks_in_path` lists eligible chunks with their `VersionedContent`.
    
NAPI:
  - `projectAllHmrEvents(target)` returns a single subscription.
    
JS:
  - `setupServerHmr` subscribes once via `allHmrEvents` instead of fanning out over `hmrChunkNamesSubscribe`.
  - `reEvaluateAllModulesExpensive()` evicts every chunk under `server/chunks/` from `require.cache` directly rather than tracking subscriptions. This is a bit fragile as it relies on the path prefix and scanning require.cache.


packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -165,98 +170,54 @@ function setupServerHmr(
...
       }
-    } catch (err) {
-      console.error('[Server HMR Setup] Error in chunk path subscription:', err)
     }
-  })()
-
-  return serverHmrSubscriptions
+  })().catch(async (err) => {
+    console.error('[Server HMR] Subscription error:', err)
+    await clear()
```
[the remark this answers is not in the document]
wbinnssmith: :robot: Added a re-subscription loop: when the \`allHmrEvents\` subscription throws, the catch handler calls \`clear()\` and re-invokes \`runSubscription()\` so server Fast Refresh stays active for the rest of the dev session.

crates/next-api/src/project.rs
```diff
@@ -2498,14 +2503,142 @@ impl Project {
...
+    pub async fn all_hmr_version_state(
+        self: ResolvedVc<Self>,
+        target: HmrTarget,
+        session: TransientInstance<()>,
+    ) -> Result<Vc<VersionState>> {
+        if target == HmrTarget::Client {
+            bail!("all_hmr_version_state is not yet implemented for the client target");
+        }
+
+        // The session argument keeps this from caching across sessions.
```
wbinnssmith: this is a trick reused from https://github.com/vercel/next.js/blob/98f5dd99975d46719396522e31516d1abb4462e2/crates/next-api/src/project.rs#L2449-L2451

@bgw @lukesandberg is there a more modern way of doing this with turbotasks?

lukesandberg: you can also tag it as `session_dependent` which will force a re-execution on next session

if you don't want it to be cached then you could just make VersionState as `serialization=skip`

sokra (dismissed):
I really don't like that we stay subscribed to all visited routes over the lifetime of the dev server. This leads to excessive compilation and memory usage over time.

Instead we should maintain some state in a turbo-tasks value which is the list of subscribed chunk groups. We can update that state via napi methods, when clients are subscribing/unsubscribing.

wbinnssmith:
> I really don't like that we stay subscribed to all visited routes over the lifetime of the dev server. This leads to excessive compilation and memory usage over time.

As discussed, this will be a priority going forward with the client implementation, but this is a reasonable limitation to start with for the server implementation.

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -0,0 +1,199 @@
...
+                let path = path.clone();
+                let version = TraitRef::cell(version.clone());
+                async move {
+                    let id = version.id().owned().await?;
+                    Ok::<_, anyhow::Error>((path, id))
+                }
+            })
+            .try_join()
+            .await?;
+        entries.sort_by(|a, b| a.0.cmp(&b.0));
```
lukesandberg: why do we sort?

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -203,60 +196,40 @@ function setupServerHmr(
...
+        // request will require() the chunks fresh from disk with current content.
+        const handlers = globalThis.__turbopack_server_hmr_handlers__
+        if (!handlers || handlers.size === 0) {
+          continue
+        }
+
         if (typeof __turbopack_server_hmr_apply__ === 'function') {
           const applied = __turbopack_server_hmr_apply__(update)
           if (!applied) {
             await clear()
```
lukesandberg: this implies some kind of application failure, do you know when this happens?

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+impl AggregateHmrVersion {
+    pub async fn from_map(
+        map: Vc<VersionedContentMap>,
+        root: &FileSystemPath,
+    ) -> Result<Vc<Box<dyn Version>>> {
+        let chunks = map.hmr_chunks_in_path(root).await?;
+        if chunks.is_empty() {
+            return Ok(Vc::upcast(NotFoundVersion::new()));
```
lukesandberg: is this better than an empty map, leave a comment

lukesandberg: looking at `diff_chunks_against` it would appear to work the same either way?

crates/next-api/src/aggregate_hmr.rs
wbinnssmith: this is probably overly defensive for the hasher, likely not needed with the indexmap.

packages/next/src/server/dev/hot-reloader-turbopack.ts
wbinnssmith: Yeah, this should only happen in real error cases (like if there was a failure to run the eval for the new module or something)

I think returning a boolean here is a misleading api. Going to refactor it so we rely on exceptions to signal something went wrong and only evict in that case. Along with the following PRs in this stack it should clear this up.

packages/next/src/server/app-render/entry-base.ts
```diff
@@ -102,7 +102,7 @@ declare global {
   var __next__clear_chunk_cache__: (() => void) | null | undefined
   var __turbopack_clear_chunk_cache__: () => void | null | undefined
   var __turbopack_server_hmr_apply__:
-    | ((update: NodeJsPartialHmrUpdate) => boolean)
+    | ((update: NodeJsPartialHmrUpdate) => void)
```
wbinnssmith, line 105: We now rely on exceptions to signal "something went wrong, evict everything and start over" instead of returning `false`.

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -0,0 +1,191 @@
...
+    #[turbo_tasks::function]
+    async fn id(&self) -> Result<Vc<RcStr>> {
+        let entries = self
+            .versions
+            .iter()
+            .map(|(path, version)| {
+                let path = path.clone();
+                let version = TraitRef::cell(version.clone());
+                async move {
+                    let id = version.id().owned().await?;
```
lukesandberg, line 36: it feels lame to re-cell these versions just to call `id` not sure there is a better option though (other than making `id` itself not a turbo-task.... but then we would need some new way to cache these hashes.... sigh, a OnceLock is actually kind of reasonable since we are storing a TraitRef (not a vc) so nothing is actually really changing

crates/next-api/src/versioned_content_map.rs
```diff
@@ -84,12 +89,66 @@ impl VersionedContentMap {
...
+    /// context can shift the internals of the map, making iteration order different for the same
+    /// set of paths.
+    pub async fn hmr_chunks_in_path(
+        self: Vc<Self>,
+        root: &FileSystemPath,
+    ) -> Result<Vec<HmrChunkWithContent>> {
+        let this = self.await?;
+        let paths: Vec<FileSystemPath> = {
+            let map = &this.map_path_to_op.get().0;
+            map.keys().cloned().collect()
```
lukesandberg, line 112: why clone here...  acant we just iterate over map.keys directly below?

crates/next-api/src/versioned_content_map.rs
```diff
@@ -84,12 +89,66 @@ impl VersionedContentMap {
...
+        Ok(chunks)
+    }
 }
 
 #[turbo_tasks::value_impl]
 impl VersionedContentMap {
     /// Inserts output assets into the map and returns a completion that when
     /// awaited will emit the assets that were inserted.
+    //
+    // TODO: If `VersionedContentMap` becomes transient as described above, these methods should be
```
lukesandberg: if it was transient i think they _wouldn't_ need to be session dependent since they would become transient tasks.... right?

packages/next/src/server/app-render/entry-base.ts
lukesandberg: This is the way

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -150,117 +150,95 @@ const isTestMode = !!(
...
-  function subscribeToServerHmr(chunkPath: string) {
-    if (serverHmrSubscriptions.has(chunkPath)) {
-      return
-    }
+  async function runSubscription() {
+    const subscription = project.allHmrEvents(HmrTarget.Server)
 
-    const subscription = project.hmrEvents(chunkPath, HmrTarget.Server)
-    serverHmrSubscriptions.set(chunkPath, subscription)
+    // Skip initial state
```
lukesandberg: why?

im guessing because the initial update is some kind of monolithic 'here is everything' update?

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -150,117 +150,95 @@ const isTestMode = !!(
...
+      if (update.type !== 'partial') {
+        continue
+      }
 
-        if (update.type !== 'partial') {
-          continue
-        }
+      const instruction = update.instruction
+      if (!instruction || instruction.type !== 'EcmascriptMergedUpdate') {
+        continue
```
lukesandberg, line 208: is this basically some kind of 'unknown' instruction type?

lukesandberg (approved):

packages/next/src/server/dev/hot-reloader-turbopack.ts
wbinnssmith: yeah

packages/next/src/server/dev/hot-reloader-turbopack.ts
wbinnssmith: yep. I'll add a comment describing this.

crates/next-api/src/versioned_content_map.rs
wbinnssmith: added a comment above

PR #94952, opened 2026-06-18
typegen: default to Turbopack

wbinnssmith:
While the bundler isn't super relevant for typegen, we _do_ validate the config as part of the process, so some checks (like those added with the experimental rust react compiler) can fail if we're not consistent about
the default.

Test Plan: added an e2e test to verify the config is validated as turbopack by default

### DIFF
--- a/test/e2e/app-dir/typegen-bundler-env/next.config.js
+++ b/test/e2e/app-dir/typegen-bundler-env/next.config.js
@@ -1,0 +1,9 @@
+// cssChunking: "graph" is Turbopack-only. loadConfig throws when TURBOPACK is
+// not set, letting the test assert that typegen selects the right bundler.
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  experimental: {
+    cssChunking: 'graph',
+  },
+}
+module.exports = nextConfig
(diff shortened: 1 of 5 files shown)


test/e2e/app-dir/typegen-bundler-env/next.config.js
```diff
@@ -0,0 +1,9 @@
+// cssChunking: "graph" is Turbopack-only. loadConfig throws when TURBOPACK is
+// not set, letting the test assert that typegen selects the right bundler.
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  experimental: {
```
wbinnssmith, line 5: the risk here is that this experimental config is removed and we need to find another sentinel...

lukesandberg: well you could do something boring like `if (!process.env.TURBOPACK) { throw new Error();}`

but maybe that isn't the right semantic for you?

lukesandberg (approved):

r34son:
@wbinnssmith Hi! Getting same error using turbopackRustReactCompiler and next/jest. Could you also fix it please?

PR #94954, opened 2026-06-18
fix(next): allow typegen without active bundler

### LINKED ISSUE #94950: `next typegen` fails when `experimental.turbopackRustReactCompiler` is `true`

https://github.com/nikhilsnayak/turbopack-rust-react-compiler-next-typegen-bug

### To Reproduce

1. Run `npm run type-check`

`next typegen` throws following error:

> Error: `experimental.turbopackRustReactCompiler` is only supported with Turbopack. Please remove the option or run Next.js with Turbopack in next.config.ts.

Type gen should work regardless of this config

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #22-Ubuntu SMP PREEMPT_DYNAMIC Mon May 25 15:54:34 UTC 2026
  Available memory (MB): 7068
  Available CPU cores: 8
Binaries:
  Node: 24.12.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 11.5.2
Relevant Packages:
  next: 16.3.0-canary.53 // There is a newer canary version (16.3.0-canary.57) available, please upgrade! 
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 6.0.3
Next.js Config:
  output: standalone
```

TypeScript

next dev (local)

### Additional context

`turbopackRustReactCompiler` is supported only for turbopack and the check for turbopack is done through `process.env.TURBOPACK` which wont be true during typegen

https://github.com/vercel/next.js/blob/45d26dfd6c938b6f0e62f0b3339274bccebf0ee6/packages/next/src/server/config.ts#L518

workaround: running `TURBOPACK=1 next typegen`

mturac:
### What?

Allows `next typegen` to load config without requiring an active bundler when bundler-specific experimental options are present.

### Why?

`next typegen` does not run webpack or Turbopack, but it currently loads config through the production build phase. That makes `experimental.turbopackRustReactCompiler` fail unless `TURBOPACK` is set, even though type generation itself does not need a bundler.

### How?

- Marks the `loadConfig` call from `next typegen` as typegen-only while preserving the public `PHASE_PRODUCTION_BUILD` value passed to `next.config.js`.
- Skips only the bundler-availability validation for `next typegen`; normal build/dev validation still runs.
- Adds config unit coverage and an e2e regression fixture that verifies route types are generated.

Fixes #94950

### Tests

- `pnpm jest packages/next/src/server/config.test.ts --runInBand`
- `pnpm test-dev test/e2e/app-dir/typegen-turbopack-rust-react-compiler/typegen-turbopack-rust-react-compiler.test.ts`
- `pnpm test-dev test/e2e/app-dir/typegen-error-diagnostic/typegen-error-diagnostic.test.ts`
- `pnpm --filter=next build`

### DIFF
--- a/packages/next/src/cli/next-typegen.ts
+++ b/packages/next/src/cli/next-typegen.ts
@@ -36,7 +36,9 @@ const nextTypegen = async (
     printAndExit(`> No such directory exists as the project root: ${baseDir}`)
   }
 
-  const nextConfig = await loadConfig(PHASE_PRODUCTION_BUILD, baseDir)
+  const nextConfig = await loadConfig(PHASE_PRODUCTION_BUILD, baseDir, {
+    isTypegen: true,
+  })
   await installBindings(nextConfig.experimental?.useWasmBinary)
   const distDir = join(baseDir, nextConfig.distDir)
   const { pagesDir, appDir } = findPagesDir(baseDir)
--- a/packages/next/src/server/config.test.ts
+++ b/packages/next/src/server/config.test.ts
@@ -73,6 +73,44 @@ describe('loadConfig', () => {
     })
   })
 
+  describe('bundler-specific config validation', () => {
+    it('does not require an active bundler when loading config for typegen', async () => {
+      await expect(
+        loadConfig(PHASE_PRODUCTION_BUILD, __dirname, {
+          customConfig: {
+            experimental: {
+              cssChunking: 'graph',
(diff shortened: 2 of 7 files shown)


mturac:
Closing this stale branch because it has drifted into conflicts and I am keeping the open PR queue focused on changes I can actively maintain.

Thanks for the review automation that ran here.

PR #94955, opened 2026-06-18
Remove legacy PPR codepaths

eps1lon:
These should be unreachable since we removed support for `config.experimental.ppr`.

Had to be previously reverted in https://github.com/vercel/next.js/pull/90948/.

No new findings from my side. I added new errors where in `forceDynamic && staticGeneration && cacheComponent` branches. These should not be reachable but now we throw explicitly instead of trying to reconcile that defect.

We'll check in closely with the people previously seeing regressions.


packages/next/src/server/app-render/create-component-tree.tsx
```diff
@@ -264,6 +263,15 @@ async function createComponentTreeInternal(
       : []
 
   let dynamic = layoutOrPageMod?.dynamic
+  if (typeof dynamic === 'string') {
+    if (cacheComponents) {
+      // TODO: Remove this since this should be caught during compilation.
+      // Left as a guard in case usage of Postpone was attempted
+      throw new Error(
+        `Route segment config "dynamic" is not compatible with \`nextConfig.cacheComponents\`. Please remove it.`
+      )
```
unstubbable, line 272: Nit: In other similar cases we're throwing an invariant error like this:

https://github.com/vercel/next.js/blob/f66c8a6a78e0c5ed30ec459071fc114f8f71d2e6/packages/next/src/server/request/cookies.ts#L83-L85

packages/next/src/server/dev/static-paths-worker.ts
```diff
@@ -131,9 +126,9 @@ export async function loadStaticPaths({
       )
     }
 
+    // TODO: vroom
```
unstubbable: 🚗 Did you plan to revisit this?

turbopack/crates/turbopack-ecmascript/tests/tree-shaker/analyzer/failed-2/input.js
```diff
@@ -45,21 +45,21 @@ export function trackDynamicDataAccessed(store, expression) {
         }
     }
 }
-export function Postpone({ reason, prerenderState, pathname }) {
-    postponeWithTracking(prerenderState, reason, pathname);
+export function ApiThatShallNotBeNamed({ reason, prerenderState, pathname }) {
```
unstubbable, line 48: 😆

unstubbable (approved):

packages/next/src/server/app-render/create-component-tree.tsx
eps1lon: Not sure yet if this is an invariant yet or an actual `export const dynamic` that we didn't catch at build time.

packages/next/src/server/dev/static-paths-worker.ts
eps1lon: Yeah. When inlining the flag, I wasn't sure yet if this is actually just the CC flag. But it's also based on the whether this is app dir or pages dir so I kept it. Removing the comment.

packages/next/src/server/config.ts
```diff
@@ -1570,11 +1570,6 @@ function assignDefaultsAndValidate(
     result.experimental.mcpServer = true
   }
 
-  if (result.cacheComponents) {
-    // TODO: remove once we've finished migrating internally to cacheComponents.
-    result.experimental.ppr = true
-  }
-
```
eps1lon, line 1577: Broke legacy builders that require changes like https://github.com/vercel/vercel/pull/16706.

Fixing in https://github.com/vercel/next.js/pull/95111

eps1lon: Reverting for now (https://github.com/vercel/next.js/pull/95113) and then thinking about forward fix.

PR #94957, opened 2026-06-18
Statically prerender metadata image routes under Cache Components

unstubbable:
Under Cache Components, metadata image routes such as `opengraph-image` and `icon` that return an `ImageResponse` were always rendered on demand (`ƒ`) rather than prerendered. `ImageResponse` defers rasterizing its element tree into the response body stream, and the route-handler prerender unwraps that body within a single task; the rasterization never finishes within that budget, so the route was classified as dynamic. This change renders and caches the image during the prerender so these routes become static (`○`).

During the prerender we serialize the `ImageResponse` arguments with React Flight's `prerenderToNodeStream` inside the prerender work-unit store, which runs the user's component tree once in the correct scope. This brings any user-space I/O inside that tree, such as `cookies()` or an uncached `fetch`, under the same Cache Components rules that already governed I/O elsewhere in the handler. The hanging-input abort signal bounds the serialization and decides static versus dynamic: if the tree is still waiting on dynamic input once the prerender's cache-sourced input is ready, the serialization can't complete and we return a hanging promise, so the final prerender's macrotask budget classifies the route as dynamic; a tree that resolves entirely from static data or `use cache` finishes serializing and is rendered to an image. This mirrors `encryptActionBoundArgs`, which serializes server action bound args with React Flight under the same hanging-input abort signal during a prerender.

The fully resolved element tree is then handed to satori. Because React Flight encodes an async Server Component's output as a `React.lazy` that satori can't walk, those references are resolved into plain elements first; this lets an async server component, including one that uses `use cache`, be passed as the `ImageResponse` element. Rasterization runs outside the prerender work-unit store: inside a Cache Components prerender an uncached `fetch`, such as the renderer loading a font, is turned into a hanging promise (Cache Components skips I/O that would not be cached anyway), so running satori with no store lets those framework fetches resolve normally. Crucially, because satori only walks the already-resolved tree, no user component runs in that storeless scope, so uncached user-space I/O can't be wrongly allowed there and let a route that should be dynamic render as static.

The rendered image is stored as an `ArrayBuffer` in a new in-memory `imageResponses` store on the Resume Data Cache, keyed by a base64 encoding of its serialized arguments, and the cache signal is held open until it is stored so the prospective prerender waits for it. The final prerender retrieves the array buffer from memory within microtasks. The store is in-memory only and is never serialized, so the image array buffers never enter the resume data that ships with the prerender.

The caching path lives in a separate `cache-image-response` module that is loaded only for Cache Components builds, gated behind `process.env.__NEXT_CACHE_COMPONENTS` so the `require` and its React Flight dependencies are eliminated as dead code otherwise; apps without Cache Components keep `ImageResponse`'s original streaming behavior unchanged.

### DIFF
--- a/packages/next/src/server/og/cache-image-response.ts
+++ b/packages/next/src/server/og/cache-image-response.ts
@@ -1,0 +1,169 @@
+import { PassThrough, Writable } from 'node:stream'
+
+import { workUnitAsyncStorage } from '../app-render/work-unit-async-storage.external'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { renderToPipeableStream } from 'react-server-dom-webpack/server'
+import { getClientReferenceManifest } from '../app-render/manifests-singleton'
+
+type OgModule = typeof import('next/dist/compiled/@vercel/og')
+
+function importOgModule(): Promise<OgModule> {
+  // Cache Components is Node-only (rejected for the edge runtime at compile
+  // time), so we always load the Node build. Loading it dynamically keeps the
+  // heavy `@vercel/og` renderer (satori + WASM) off the module-load path, so
+  // it's pulled in only when an image is actually rendered.
+  return import('next/dist/compiled/@vercel/og/index.node.js')
+}
+
+/**
+ * Builds the body for a Cache Components `ImageResponse`. The rendered image is
+ * cached in the Resume Data Cache during a prerender, so the prospective
+ * prerender renders it once and the final prerender retrieves it from memory
+ * within microtasks. This lets metadata image routes be statically prerendered
+ * under Cache Components instead of being treated as dynamic.
+ *
+ * The cache boundary is drawn around only the deterministic JSX -> image
+ * rasterization. The handler that constructs the `ImageResponse` still re-runs
+ * on every prerender pass, so any user-space I/O remains subject to the normal
+ * Cache Components rules (it must be wrapped in `use cache` by the user). If
+ * the inputs don't reproduce across the prospective and final passes, the cache
+ * keys diverge, the final render re-renders, and the route falls back to
+ * dynamic.
+ *
+ * Outside of a prerender (normal requests) this just renders.
+ */
+export function getCachedImageResponseBody(
+  args: ConstructorParameters<OgModule['ImageResponse']>
+): ReadableStream {
+  return new ReadableStream({
+    async start(controller) {
+      const arrayBuffer = await getCachedImageResponseArrayBuffer(args)
+      if (arrayBuffer.byteLength > 0) {
+        controller.enqueue(new Uint8Array(arrayBuffer))
+      }
+      controller.close()
+    },
+  })
+}
+
+async function getCachedImageResponseArrayBuffer(
+  args: ConstructorParameters<OgModule['ImageResponse']>
+): Promise<ArrayBuffer> {
+  const workUnitStore = workUnitAsyncStorage.getStore()
+
+  switch (workUnitStore?.type) {
+    case 'prerender':
+      // We only cache during a prerender. Metadata image routes compile to
+      // route handlers, which use the `prerender` store.
+      break
+    case undefined:
+    case 'request':
+    case 'cache':
+    case 'private-cache':
+    case 'unstable-cache':
+    case 'prerender-runtime':
+    case 'prerender-client':
+    case 'validation-client':
+    case 'prerender-ppr':
+    case 'prerender-legacy':
+    case 'generate-static-params':
+      return renderImageResponseArrayBuffer(args)
+    default:
+      return workUnitStore satisfies never
+  }
+
+  const { cacheSignal, resumeDataCache } = workUnitStore
+
+  if (!resumeDataCache) {
+    return renderImageResponseArrayBuffer(args)
+  }
+
+  // Begin a cache read synchronously, before any await, so the prospective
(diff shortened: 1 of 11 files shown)


packages/next/src/server/og/cache-image-response.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  const passThrough = new PassThrough()
+  pipe(passThrough)
+
+  // The Flight stream can contain raw binary bytes. Most notably, binary
+  // `ImageResponse` args such as a custom font (`fonts: [{ data: ArrayBuffer }]`)
+  // are serialized as verbatim typed-array bytes, which are not valid UTF-8.
+  // Decoding those bytes with a fatal UTF-8 decoder would throw, so we collect
+  // the raw bytes and build the key with a lossless binary-safe encoding
+  // (Node's `latin1`, i.e. ISO-8859-1, maps each byte to a distinct code point).
```
lubieowoce: hmm this is funky. is there an efficiency gain over using base64? i'm kinda unsure about the correctness of this so i'd rather default to the known good thing

packages/next/src/server/og/cache-image-response.ts
```diff
@@ -0,0 +1,173 @@
...
+  } finally {
+    cacheSignal?.endRead()
+  }
+}
+
+async function renderImageResponseArrayBuffer(
+  args: ConstructorParameters<OgModule['ImageResponse']>
+): Promise<ArrayBuffer> {
+  const OGImageResponse = (await importOgModule()).ImageResponse
+  const imageResponse = new OGImageResponse(...args) as Response
```
lubieowoce: why the `as Response` cast? looks weird

packages/next/src/server/og/cache-image-response.ts
```diff
@@ -0,0 +1,313 @@
...
+ * needs dynamic input the serialization can't complete, and the route falls
+ * back to dynamic. Otherwise the fully resolved tree is handed to satori,
+ * which never re-runs the user's components.
+ *
+ * Outside of a prerender (normal requests) this just renders.
+ */
+export function getCachedImageResponseBody(
+  args: ImageResponseArgs
+): ReadableStream {
+  return new ReadableStream({
```
lubieowoce:
```suggestion
): ReadableStream<Uint8Array> {
  return new ReadableStream<Uint8Array>({
```

packages/next/src/server/og/cache-image-response.ts
```diff
@@ -0,0 +1,313 @@
...
+  hangingInputAbortSignal.addEventListener(
+    'abort',
+    () => {
+      if (!prerenderCompleted) {
+        resultIsPartial = true
+        abortController.abort(hangingInputAbortSignal.reason)
+      }
+    },
+    { once: true }
+  )
```
lubieowoce: can't we just pass in `hangingInputAbortSignal` but still attach this listener before? `.abort()` runs listeners synchronously so that should be equivalent unless i'm missing something

packages/next/src/server/og/cache-image-response.ts
```diff
@@ -0,0 +1,313 @@
...
+      signal: abortController.signal,
+      filterStackFrame: undefined,
+      onError(error) {
+        // A halt (our deliberate abort) emits nothing, so this is only called
+        // for genuine serialization errors. We surface the first one.
+        if (serializationError === undefined && !resultIsPartial) {
+          serializationError = error
+        }
+      },
+    }).then((result) => {
```
lubieowoce: nit: is doing this inside `.then()` getting us anything? it looks like it could be done just fine by just putting this code below the `await`

packages/next/src/server/og/cache-image-response.ts
unstubbable: we can. fixed it.

packages/next/src/server/og/cache-image-response.ts
```diff
@@ -0,0 +1,312 @@
...
+    }
+    return resolveFlightLazies(lazy._init(lazy._payload))
+  }
+
+  if (Array.isArray(node)) {
+    return node.map(resolveFlightLazies)
+  }
+
+  const element = node as { props?: { children?: unknown } }
+  if (element.props && 'children' in element.props) {
```
lubieowoce, line 299: i don't think we can get away with only checking `children`, we can get JSX in any prop:
`<One elem={<Two />} />`
they can also be nested:
`<One stuff={{ nested: <Two /> }} />`

also this probably doesn't handle thenables as children for example. not sure if we need to be concerned about that

packages/next/src/server/og/cache-image-response.ts
unstubbable: fixed

packages/next/src/server/og/cache-image-response.ts
lubieowoce: ok considering that we don't intend to support arbitrary client components in here, it's probably fine

lubieowoce (approved):

packages/next/src/server/og/cache-image-response.ts
unstubbable: Yeah, this doesn't cover all cases, which I also don't like. Generally, this PR is a best-effort attempt at preventing the static ➞ dynamic regression for metadata image routes when upgrading to Cache Components. It does not cover all edge cases.

PR #94963, opened 2026-06-18
[Node Streams] Use node:stream/consumers

lubieowoce:
Converting a node stream to a ReadableStream just to collect it into a buffer seems a bit inefficient


eps1lon:
Why not use the built-in helpers everywhere?

PR #94974, opened 2026-06-19
fix: after(callback) called after response end

lubieowoce:
Fixes a bug where if `after(callback)`was called after the response ended, and no `after(callback)` calls occurred earlier, the callback would never run.

Implementation-wise, what happens is:
1. when the first callback is scheduled, we do `runCallbacksOnClose()`, which waits until `onClose` fires and starts the callback queue
2. but if no `after(callback)` calls occurred before the response ended, then by the time we got to `runCallbacksOnClose()`, `onClose` has already fired, so we'd wait forever

We now start listening for `onClose` when the context is initialized and track the state (`isRequestClosed`), which avoids this.

I also did another drive-by fix here: if we only had `after(promise)` calls, we wouldn't switch the workUnitStore's phase to `'after'`, because we were only doing that in the `after(callback)` codepath. This is relevant in #94964 which fixes some bugs around `after(promise)`.

### DIFF
--- a/packages/next/errors.json
+++ b/packages/next/errors.json
@@ -1365,5 +1365,6 @@
   "1364": "\\`experimental.turbopackRustReactCompiler\\` requires \\`reactCompiler\\` to be enabled. Please add \\`reactCompiler: true\\` in %s.",
   "1365": "Expected a dev tiered cache handler for kind \"%s\".",
   "1366": "Turbopack build failed with %s %s:\\n%s",
-  "1367": "createServerParamsForRoute should not be called in runtime prerenders."
+  "1367": "createServerParamsForRoute should not be called in runtime prerenders.",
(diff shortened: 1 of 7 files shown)


lubieowoce:
* **#94964** <a href="https://app.graphite.com/github/pr/vercel/next.js/94964?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94974** <a href="https://app.graphite.com/github/pr/vercel/next.js/94974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94974?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

unstubbable (approved):

PR #94977, opened 2026-06-19
Record size metrics to our compaction and persistence spans

lukesandberg:
Add data about persistence and compaction size to our spans

This will show up in trace data collected by `NEXT_TURBOPACK_TRACING` and in the default trace spans collected by next.js

### DIFF
--- a/turbopack/crates/turbo-persistence/src/db.rs
+++ b/turbopack/crates/turbo-persistence/src/db.rs
@@ -1,6 +1,7 @@
 use std::{
     borrow::Cow,
     collections::HashSet,
+    fmt::Display,
     fs::{self, File, OpenOptions, ReadDir},
     io::{BufWriter, Write},
     mem::take,
@@ -276,14 +277,27 @@ struct DeletedFile {
     size: u64,
 }
 
-/// Physical byte volume of a single commit, measured from on-disk file sizes (post-compression,
-/// including `.sst`, `.blob`, and `.meta` files).
+/// Physical byte volume of a single commit/compaction cycle, measured from on-disk file sizes
+/// (post-compression, including `.sst`, `.blob`, and `.meta` files).
 #[derive(Clone, Copy, Debug, Default)]
 pub struct CommitStats {
     /// Total bytes of new files created by this commit.
     pub bytes_written: u64,
     /// Total bytes of files removed/superseded by this commit.
     pub bytes_deleted: u64,
+}
+
+impl Display for CommitStats {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let CommitStats {
+            bytes_written,
+            bytes_deleted,
+        } = self;
+        write!(
+            f,
+            "bytes_written={bytes_written} bytes_deleted={bytes_deleted}"
+        )
+    }
 }
 
 struct OpenOpts<S: ParallelScheduler, const FAMILIES: usize> {
@@ -799,14 +813,7 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES> 
 
         new_meta_files.sort_unstable_by_key(|f| f.seq);
 
-        // Sum the on-disk size of every new file (post-compression). The sizes were captured at
-        // each file's creation site (no stat), so this is a plain in-memory sum.
-        let bytes_written = new_meta_files
-            .iter()
-            .chain(new_sst_files.iter())
-            .chain(new_blob_files.iter())
-            .map(|f| f.size)
-            .sum::<u64>();
+        let mut stats = CommitStats::default();
 
         let sync_span = tracing::trace_span!("sync new files").entered();
 
@@ -823,13 +830,16 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES> 
 
         let mut sync_items: Vec<SyncItem> =
             Vec::with_capacity(new_meta_files.len() + new_sst_files.len() + new_blob_files.len());
-        for NewFile { seq, file, .. } in new_meta_files {
+        for NewFile { seq, file, size } in new_meta_files {
+            stats.bytes_written += size;
             sync_items.push(SyncItem::Meta(seq, file));
         }
-        for NewFile { file, .. } in new_sst_files {
+        for NewFile { file, size, .. } in new_sst_files {
+            stats.bytes_written += size;
             sync_items.push(SyncItem::Sst(file));
         }
-        for NewFile { seq, file, .. } in new_blob_files {
+        for NewFile { seq, file, size } in new_blob_files {
+            stats.bytes_written += size;
             sync_items.push(SyncItem::Blob(seq, file));
         }
 
@@ -906,11 +916,9 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES> 
         let has_delete_file;
         let mut meta_seq_numbers_to_delete = Vec::new();
         let entries_to_remove;
-        // On-disk bytes of deleted SST files. The caller knows each deleted SST's size when it
-        // decides to delete it, so it's carried on `DeletedFile` — just sum it (no scan, no stat).
-        let sst_bytes_deleted = sst_files_to_delete.iter().map(|f| f.size).sum::<u64>();
-        // On-disk bytes of deleted meta files, read from each `MetaFile`'s mmap length (no stat).
-        let mut meta_bytes_deleted = 0u64;
+        // Deleted SST bytes: the caller knows each deleted SST's size when it decides to delete it,
+        // so it's carried on `DeletedFile` and summed here (no scan, no stat).
+        stats.bytes_deleted += sst_files_to_delete.iter().map(|f| f.size).sum::<u64>();
         // The rest of the commit only needs the sequence numbers of the deleted SSTs.
         let mut sst_seq_numbers_to_delete = sst_files_to_delete
             .iter()
@@ -948,7 +956,8 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES> 
             for i in (0..inner.meta_files.len()).rev() {
                 if sst_filter.apply_and_get_remove(&inner.meta_files[i]) {
                     meta_seq_numbers_to_delete.push(inner.meta_files[i].sequence_number());
-                    meta_bytes_deleted += inner.meta_files[i].byte_size();
+                    // Deleted meta bytes, read from the `MetaFile`'s mmap length (no stat).
+                    stats.bytes_deleted += inner.meta_files[i].byte_size();
                 }
             }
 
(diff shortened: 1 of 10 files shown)


lukesandberg:
* **#94842** <a href="https://app.graphite.com/github/pr/vercel/next.js/94842?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#94977** <a href="https://app.graphite.com/github/pr/vercel/next.js/94977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/94977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -770,15 +789,24 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
...
+        new_meta_files.sort_unstable_by_key(|f| f.seq);
+
+        // Sum the on-disk size of every new file (post-compression). The sizes were captured at
+        // each file's creation site (no stat), so this is a plain in-memory sum.
+        let bytes_written = new_meta_files
+            .iter()
+            .chain(new_sst_files.iter())
+            .chain(new_blob_files.iter())
+            .map(|f| f.size)
+            .sum::<u64>();
```
lukesandberg: sum as part of the loops below

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -878,6 +906,16 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
         let has_delete_file;
         let mut meta_seq_numbers_to_delete = Vec::new();
         let entries_to_remove;
+        // On-disk bytes of deleted SST files. The caller knows each deleted SST's size when it
+        // decides to delete it, so it's carried on `DeletedFile` — just sum it (no scan, no stat).
+        let sst_bytes_deleted = sst_files_to_delete.iter().map(|f| f.size).sum::<u64>();
+        // On-disk bytes of deleted meta files, read from each `MetaFile`'s mmap length (no stat).
+        let mut meta_bytes_deleted = 0u64;
```
lukesandberg: we should have one accumulator instead of several

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -910,16 +948,33 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
...
                 || !blob_seq_numbers_to_delete.is_empty()
                 || !meta_seq_numbers_to_delete.is_empty();
         }
+
+        // Bytes deleted by this commit. SST sizes came from `DeletedFile` and meta sizes from each
+        // `MetaFile`'s mmap length (both in-memory, no stat). Blob sizes aren't tracked in memory,
+        // so stat them by sequence number before Phase C unlinks them — best-effort: a file already
+        // gone reports 0 rather than failing the commit (these stats are reported, not
+        // load-bearing). Blob deletions are rare.
+        let blob_bytes_deleted = blob_seq_numbers_to_delete
```
lukesandberg: classic parallelism opportunity

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -1189,16 +1247,25 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
...
-                sst_seq_numbers_to_delete,
-                blob_seq_numbers_to_delete,
-                sequence_number: *sequence_number.get_mut(),
-                keys_written,
-            })
-            .context("Failed to commit the database compaction")?;
+            let span = tracing::info_span!(
+                "compact",
+                bytes_written = tracing::field::Empty,
+                bytes_deleted = tracing::field::Empty,
```
lukesandberg: just emit the whole struct as a `display(..)` item to the span

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -1189,16 +1247,25 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
...
-                new_meta_files,
-                new_sst_files,
-                new_blob_files: Vec::new(),
-                sst_seq_numbers_to_delete,
-                blob_seq_numbers_to_delete,
-                sequence_number: *sequence_number.get_mut(),
-                keys_written,
-            })
-            .context("Failed to commit the database compaction")?;
+            let span = tracing::info_span!(
```
lukesandberg: i believe there is already a span for compaction, we shouldn't need another one?

turbopack/crates/turbo-persistence/src/meta_file_builder.rs
```diff
@@ -43,15 +43,19 @@ impl<'a> MetaFileBuilder<'a> {
...
+    pub fn write(self, db_path: &Path, seq: u32) -> Result<(File, u64)> {
         let file = db_path.join(format!("{seq:08}.meta"));
         self.write_internal(&file)
             .with_context(|| format!("Unable to write meta file {seq:08}.meta"))
     }
 
-    fn write_internal(mut self, file: &Path) -> io::Result<File> {
-        let mut file = BufWriter::new(File::create(file)?);
+    fn write_internal(mut self, file: &Path) -> io::Result<(File, u64)> {
+        // Wrap the writer to count the bytes written, so callers can accumulate written-byte totals
```
lukesandberg, line 54: drop redundant comment

turbopack/crates/turbo-persistence/src/write_batch.rs
```diff
@@ -28,6 +28,15 @@ use crate::{
     static_sorted_file_builder::{StaticSortedFileBuilderMeta, write_static_stored_file},
 };
 
+/// A newly created database file (meta, SST, or blob), carrying its on-disk size so commit can sum
+/// written bytes without stat'ing the files afterwards.
+pub(crate) struct NewFile {
+    pub(crate) seq: u32,
+    pub(crate) file: File,
+    /// On-disk size in bytes (post-compression).
```
lukesandberg: (post-compression) is obvious

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -282,6 +282,9 @@ impl TurboBackingStorage {
                         data_items,
                         meta_items,
                         task_cache_items,
+                        // we don't know yet
+                        bytes_written: 0,
+                        bytes_deleted: 0,
```
lukesandberg: should we introduce a separate struct to accumulate?

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: :robot: Done — folded the bytes_written sum into the existing sync-item loops (no separate chain().sum() pass). It now accumulates into a single CommitStats while building sync_items.

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: :robot: Done — replaced the separate sst/meta/blob byte locals with a single CommitStats accumulator that all the loops add into, returned directly from commit().

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: :robot: Decided to keep this serial. Blob deletions are rare and few (the surrounding code already notes this), so dispatching a handful of stat() calls to the parallel scheduler would cost more in fan-out overhead than it saves. Updated the comment to make that rationale explicit.

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: :robot: Done — removed the per-field span recording. CommitStats now has a Display impl and is emitted as a single display() field on the existing 'compact database' span (see thread re: the duplicate span).

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: :robot: Good catch — removed the redundant 'compact' span. There's already an info_span!("compact database") in backend/mod.rs wrapping backing_storage.compact(). compact() now returns Result<Option<CommitStats>> (Some when a commit happened, None when there was nothing to compact), and the stats are recorded on that existing span instead of a nested one.

turbopack/crates/turbo-persistence/src/meta_file_builder.rs
lukesandberg: :robot: Done — dropped the doc comment that just restated the (File, u64) signature.

turbopack/crates/turbo-persistence/src/write_batch.rs
lukesandberg: :robot: Done — dropped the '(post-compression)' note from the NewFile.size doc.

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
lukesandberg: :robot: Addressed on the turbo-persistence side: CommitStats itself is now the single accumulator (one struct summed across all the commit loops). Here in save_snapshot the byte fields genuinely aren't known until batch.commit() returns, so they stay 0 in the per-shard SnapshotMeta and get filled from CommitStats right after the commit — clarified the comment to say so rather than introducing another struct.

turbopack/crates/turbo-tasks-backend/src/database/turbo/mod.rs
```diff
@@ -195,10 +196,19 @@ fn do_compact(
...
+            vec![
+                (
+                    "bytes_written",
+                    serde_json::Value::from(stats.bytes_written),
+                ),
+                (
+                    "bytes_deleted",
+                    serde_json::Value::from(stats.bytes_deleted),
+                ),
+            ],
```
bgw: It seems like it might be nicer if this accepted `serde_json::Value` so we could write `serde_json::json! { ... }`

The syntax in `turbopack/crates/turbo-tasks-backend/src/backend/mod.rs` would benefit from this too.

turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
```diff
@@ -59,6 +59,11 @@ pub struct SnapshotMeta {
     pub data_items: usize,
     pub meta_items: usize,
     pub task_cache_items: usize,
+    /// Physical on-disk bytes (post-compression, including `.sst`/`.blob`/`.meta` files) written
+    /// by the commit.
+    pub bytes_written: usize,
+    /// Physical on-disk bytes of files removed/superseded by the commit.
+    pub bytes_deleted: usize,
```
bgw: Consistency: `CommitStats` stores these as u64 instead of usize, and we have to do some conversion when we update `SnapshotMeta`?

`u64` seems better here because we're talking about on-disk sizes and not in-memory sizes. (e.g. `std::fs::Metadata::len` uses `u64`)

```suggestion
    /// Physical on-disk bytes (post-compression, including `.sst`/`.blob`/`.meta` files) written
    /// by the commit.
    pub bytes_written: u64,
    /// Physical on-disk bytes of files removed/superseded by the commit.
    pub bytes_deleted: u64,
```

bgw (approved):

turbopack/crates/turbo-tasks-backend/src/database/turbo/mod.rs
lukesandberg: yeah nicer, gives up a little bit of type safety but that is no big deal

PR #94992, opened 2026-06-20
fix(typegen): skip turbopackRustReactCompiler Turbopack-only check during next typegen

### LINKED ISSUE #94950: `next typegen` fails when `experimental.turbopackRustReactCompiler` is `true`

sleitor:
## What?

Skip the `experimental.turbopackRustReactCompiler` Turbopack-only validation when running `next typegen`.

## Why?

`next typegen` does not involve a bundler at all — it only loads the project config to generate TypeScript route types. However, the check at `packages/next/src/server/config.ts` verifies that the `TURBOPACK` environment variable is set when `experimental.turbopackRustReactCompiler` is enabled, and since `typegen` never sets that variable the command throws:

```
Error: `experimental.turbopackRustReactCompiler` is only supported with Turbopack.
Please remove the option or run Next.js with Turbopack in next.config.ts.
```

## How?

1. At the start of the `nextTypegen` function in `next-typegen.ts`, set `process.env.NEXT_PRIVATE_TYPEGEN = '1'` before `loadConfig` is called.
2. In `config.ts`, add `!process.env.NEXT_PRIVATE_TYPEGEN` to the condition that throws the Turbopack-only error so typegen bypasses it.

The existing workaround of running `TURBOPACK=1 next typegen` continues to work unchanged.

Fixes #94950

### DIFF
--- a/packages/next/src/cli/next-typegen.ts
+++ b/packages/next/src/cli/next-typegen.ts
@@ -30,6 +30,11 @@ export type NextTypegenOptions = {
 const nextTypegen = async (options: NextTypegenOptions, directory?: string) => {
   parseBundlerArgs(options)
 
+  // Signal to config loading that we are running typegen so that
+  // Turbopack-only config checks (e.g. turbopackRustReactCompiler) are
+  // skipped – typegen does not involve a bundler at all.
+  process.env.NEXT_PRIVATE_TYPEGEN = '1'
+
   const baseDir = getProjectDir(directory)
 
   // Check if the provided directory exists
--- a/packages/next/src/cli/next-typegen.ts
+++ b/packages/next/src/cli/next-typegen.ts
@@ -6,7 +6,7 @@ import { mkdir } from 'fs/promises'
 
 import loadConfig from '../server/config'
 import { printAndExit } from '../server/lib/utils'
-import { PHASE_PRODUCTION_BUILD } from '../shared/lib/constants'
+import { PHASE_INFO } from '../shared/lib/constants'
 import { getProjectDir } from '../lib/get-project-dir'
 import { findPagesDir } from '../lib/find-pages-dir'
 import { verifyAndRunTypeScript } from '../lib/verify-typescript-setup'
@@ -30,11 +30,6 @@ export type NextTypegenOptions = {
 const nextTypegen = async (options: NextTypegenOptions, directory?: string) => {
   parseBundlerArgs(options)
 
-  // Signal to config loading that we are running typegen so that
-  // Turbopack-only config checks (e.g. turbopackRustReactCompiler) are
-  // skipped – typegen does not involve a bundler at all.
-  process.env.NEXT_PRIVATE_TYPEGEN = '1'
-
   const baseDir = getProjectDir(directory)
 
   // Check if the provided directory exists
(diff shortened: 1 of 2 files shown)


eps1lon (commented):
> and since typegen never sets that variable the command throws:

Isn't that the bug? I think it scales worse in he future to gate every config check based on the context. Instead, whatever loads the config should set the right context.

sleitor:
Thanks @eps1lon — that's a fair point. The `NEXT_PRIVATE_TYPEGEN=1` env flag is essentially "set the right context before loading config", but you're right that putting the bypass in `config.ts` adds one more context-specific gate there.

A cleaner approach would be to propagate a structured context option through `loadConfig` (e.g. `{ phase: 'typegen' }`) and have checks like the Turbopack-only guard inspect that instead of reading env vars. That way config.ts doesn't accumulate an ever-growing list of env-var overrides.

Would you prefer I refactor toward that pattern, or is there a lighter alternative you had in mind (e.g. just skipping the turbopack guard when no bundler phase is set)?

eps1lon:
The caller of `loadConfig` should have the right context. That doesn't imply we need to pass new arguments to `loadConfig`. Just that the callsite needs to have the proper context.

sleitor:
Thanks for the clarification! That makes sense — the typegen command (as the caller) already sets the context via the env var before calling `loadConfig`, which is the right direction.

But I think the deeper concern from your review is that adding exception clauses to each Turbopack-only check in `assignDefaultsAndValidate` will scale poorly over time. A cleaner approach would be to move the `turbopackRustReactCompiler` validation out of `assignDefaultsAndValidate` entirely and into the Turbopack startup path (e.g. where Turbopack is initialized for `dev`/`build`), where it actually makes sense to enforce it.

That way `next typegen` (and other non-Turbopack commands) load config cleanly without needing any exception clause, and the constraint is only checked in contexts where Turbopack is actually running.

Would that be the direction you'd prefer? Happy to rework the PR along those lines.

eps1lon:
All I'm saying is that `loadConfig` should be called with the right context setup out of the box. How that looks, you'll have to figure out. But adding new forks based on context that's effectively just the callsite, won't scale. Maybe `next dev` and `next build` give a hint.

sleitor:
Thanks @eps1lon — you're right, an env var fork per callsite doesn't scale.

Looking at how `next dev` and `next build` handle Turbopack context gave the hint: they both set `TURBOPACK` as part of the bundler startup, and nothing else. For `next typegen`, there is no bundler, so the right signal is the *phase* — the same mechanism the rest of `loadConfig` already uses.

**New approach (just pushed):**

1. `next-typegen.ts`: switch from `PHASE_PRODUCTION_BUILD` to `PHASE_INFO`. This is already what `next info` uses for read-only config loading, and typegen is the same kind of operation — it reads config and analyses the project, no bundler starts.

2. `config.ts`: extend the existing `assignDefaultsAndValidate` guard from `phase !== PHASE_PRODUCTION_SERVER` to `phase !== PHASE_PRODUCTION_SERVER && phase !== PHASE_INFO`. This makes the bundler-specific checks (cssChunking, turbopackRustReactCompiler) skip all informational commands consistently.

3. Remove `NEXT_PRIVATE_TYPEGEN` entirely — no new env vars.

Let me know if there's a different direction you had in mind.

sleitor:
Closing this PR — issue #94950 was fixed upstream by #94952 (merged June 19). Thanks!

PR #94997, opened 2026-06-20
docs(cache-components): clarify allow-runtime, sync-IO and instant=false, CLS fallback

aurorascharff:
### What?

Docs-only follow-ups for Cache Components and Partial Prefetching, from agent migration friction logs and Joseph's review of [#95082](https://github.com/vercel/next.js/pull/95082).

- **`adopting-partial-prefetching.mdx`** — reframe the `allow-runtime` row in the audit table (it's an enhancement, not the `<Link prefetch={true}>` warning fix; the per-route opt-in is `prefetch = 'partial'`).
- **`migrating-to-cache-components.mdx`** — split the `## Enable Cache Components` section into before-flag (remove `dynamic` / `revalidate` / `fetchCache`) and after-flag (translate `revalidate`, `fetchCache`, `unstable_cache`, `fetch` cache options) steps. Flag the `revalidateTag` second-argument requirement, with a Before / After example. Add two Good-to-know callouts: off-grid `revalidate` values map to the closest `cacheLife` profile; `instant = false` is allowed-to-block, not forced-dynamic.
- **`blocking-prerender-current-time` / `-random` / `-crypto`** (and their `-client` siblings) — drop the misleading "Don't want this validation?" opt-out section. `instant = false` is an instant-navigation knob and doesn't suppress sync-IO prerender errors.
- **Remaining `blocking-prerender-*` pages** — standardize the CLS fallback callout so they all link the canonical section.

### Why?

Friction logs surfaced four recurring misreads:

1. Devs expected `prefetch = 'allow-runtime'` to silence the `<Link prefetch={true}>` warning. The docs implied it would.
2. The migration guide mixed `dynamic` removal with `unstable_cache` translation, so it wasn't clear which had to happen before the flag and which after.
3. `revalidateTag` now requires a cache profile, but the on-demand revalidation section didn't say so.
4. Sync-IO error pages still carried a "Don't want this validation?" opt-out that doesn't actually work for those errors.


docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -47,6 +47,10 @@ export default nextConfig
 
 > **Good to know:** If you were using `experimental.dynamicIO` or `experimental.useCache`, `cacheComponents` replaces them. See the [version 16 upgrade guide](/docs/app/guides/upgrading/version-16#experimentaldynamicio-and-experimentalusecache).
 
+Before enabling the flag, remove route segment configs such as `export const dynamic`, `export const revalidate`, and `export const fetchCache` from your route segments. Under `cacheComponents`, these are no longer compatible, and pages that still export them will fail to build. The sections below cover the replacements (`cacheLife` and `'use cache'`).
```
icyJoseph: This is already in line 21, this section is about enabling the flag.

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -47,6 +47,10 @@ export default nextConfig
...
+
+Your existing `unstable_cache` calls and `fetch` cache options keep working as a separate caching layer after the flag is on. Translate them as the validation insights and errors surface each one.
```
icyJoseph: This is in line 23 too

docs/01-app/02-guides/migrating-to-cache-components.mdx
aurorascharff: Sorry, i think something got messed up when i moved this section from a different PR. thanks!

errors/blocking-prerender-current-time.mdx
```diff
@@ -83,11 +83,24 @@ export async function UpdatedAt() {
 }
 ```
 
-Learn more: [Streaming patterns and boundary placement](/docs/app/guides/streaming).
+#### Alternative: `await io()`
+
+If you'd rather not tie the read to an HTTP request context, use [`io()`](/docs/app/api-reference/functions/io) from `next/cache`. It signals a request-time IO read and works the same way as `connection()` in the streaming case, plus inside `"use cache"` scopes and Client Components where `connection()` throws.
```
icyJoseph:
```suggestion
To avoid tying the read to an HTTP request context, use [`[io()](https://chatgpt.com/docs/app/api-reference/functions/io)`](/docs/app/api-reference/functions/io) from `next/cache`. It marks the read as request-time IO and behaves like `connection()` for streaming, with the added benefit that it also works inside `"use cache"` scopes and Client Components, where `connection()` throws.
```

docs/01-app/02-guides/migrating-to-cache-components.mdx
icyJoseph: No worries, it got me thinking though... cuz it is worth mentioning that using these, is a hard error once the flag is on, so you'll need to actually act on it

icyJoseph: Wydt:

```suggestion
After enabling the flag, route segments that still export `dynamic`, `revalidate`, or `fetchCache` will error. Start by replacing those configs. The sections below explain what to do for each one.
```

We can be upfront, the flag activation is not compatible with these, oyu don't even need to run a build or dev, rather, go through our guidance (thinking of the LLM case)

aurorascharff: yes, nice! fixed

docs/01-app/02-guides/migrating-to-cache-components.mdx
aurorascharff: fixed

errors/blocking-prerender-current-time.mdx
aurorascharff: fixed (without the chatgpt link)

icyJoseph (approved):

PR #94999, opened 2026-06-20
docs: add Building guide

aurorascharff:
Adds a **Building** guide (`docs/01-app/02-guides/building.mdx`): what `next build` does, how to read the route table (`○` Static, `◐` Partial Prerender, `ƒ` Dynamic), prerendering dynamic routes with `generateStaticParams`, streaming request-time work, and debugging prerender errors. Also links the guide from the deploying page and the `next build` CLI reference.


docs/01-app/02-guides/building.mdx
```diff
@@ -0,0 +1,211 @@
...
+    - app/guides/streaming
+    - app/guides/incremental-static-regeneration-cache-components
+    - app/guides/ci-build-caching
+---
+
+Run `next build` to compile your app for production: optimized bundles, prerendered pages, and a route table that shows how each URL is served (fully static, partial prerender with streaming, dynamic on demand, or a combination).
+
+## What `next build` does
+
+`next build` runs in these phases:
```
icyJoseph: I wonder if we could do an increment here:

- https://github.com/vercel/next.js/blob/canary/docs/01-app/01-getting-started/01-installation.mdx#run-the-development-server

We don't really show the commands with:

````
```bash package="pnpm"
pnpm dev
```

```bash package="npm"
npm run dev
```

```bash package="yarn"
yarn dev
```

```bash package="bun"
bun dev
```
````

Same for build? that'd render the various commands selector -

File: docs/01-app/02-guides/building.mdx [hunk 2]
```diff
@@ -0,0 +1,211 @@
...
+◐  (Partial Prerender)  prerendered as static HTML with dynamic server-streamed content
+```
+
+When a user visits an unlisted param, Next.js serves the prerendered shell immediately while the page content streams in.
+
+When there are more paths than fit in the preview, Next.js prints a short list and ends with `[+N more paths]`.
+
+> **Good to know**: The shell for unlisted params is available from Next.js 16.3. Earlier versions wait for a full server render before sending the response for unlisted params.
+
+Routes that use [`cacheLife`](/docs/app/api-reference/functions/cacheLife) also show `Revalidate` and `Expire` columns. These cached, revalidating routes are covered in [Incremental Static Regeneration with Cache Components](/docs/app/guides/incremental-static-regeneration-cache-components):
```
icyJoseph: Even routes without it because it'd be the default

icyJoseph (approved):

docs/01-app/02-guides/building.mdx
```diff
@@ -0,0 +1,376 @@
...
+
+## Reading the build output
+
+With Cache Components, Next.js prerenders as much of each route as it can at [build time](/docs/app/glossary#build-time). [Runtime data](/docs/app/glossary#request-time-apis) ([`params`](/docs/app/api-reference/file-conventions/page#params-optional), [`searchParams`](/docs/app/api-reference/file-conventions/page#searchparams-optional), [`cookies()`](/docs/app/api-reference/functions/cookies), [`headers()`](/docs/app/api-reference/functions/headers)) and uncached data (`fetch()`, database calls) are not available during that pass. Wrap runtime reads in [`<Suspense>`](https://react.dev/reference/react/Suspense), and cache data accesses with [`use cache`](/docs/app/api-reference/directives/use-cache) or wrap them in `<Suspense>` too. Otherwise, the build fails with a [`blocking-prerender-runtime`](/docs/messages/blocking-prerender-runtime) or [`blocking-prerender-dynamic`](/docs/messages/blocking-prerender-dynamic) error. [Random values and timestamps](/docs/app/getting-started/caching#random-values-and-timestamps) such as `Math.random()` or `new Date()` can also fail the build.
```
icyJoseph: I think we could've organized this as: 

Open with: 

After a successful build, Next.js prints a route table with a symbol next to each route, showing how it is served:

The table as it is today, with everything

Then, static, SSG, and dynamic are the modes applications have always rendered in Next.js, including Pages Router... and so on... 

Cache Components introduces a new type of output, Partial Prerender, and follow with what we have so far. Routes can still be fully static under this model - some explanation there on how routes move in that spectrum of static - dynamic, never being fully server rendered anymore (unless you opt out?)

Then I think the intro to this section can be moved to the Example section, right before:

> Run `next build`. It fails with a prerender-blocking error because the `params` read and the uncached `fetch` run outside a `<Suspense>` boundary

Adjusting the prose as needed.

And - I'd add a variation of this under the Example heading:

> One of the most common reasons production builds fail today is prerender validation. The remainder of this guide walks through that process.

We could frame it as an adoption, or learning exercise, but also as a safeguard that Cache Components puts in place to ensure your application is fast.

docs/01-app/02-guides/incremental-static-regeneration-cache-components.mdx
```diff
@@ -198,6 +200,10 @@ After the first visit, Next.js renders the page in the background with the known
 
 > **Good to know**: Prefetching also triggers upgrades. When a [`<Link>`](/docs/app/api-reference/components/link) enters the viewport or [`router.prefetch`](/docs/app/api-reference/functions/use-router) is called, Next.js can upgrade the App Shell in the background, so the next visitor gets the more specific version even before anyone actually navigates to the page.
 
+## High-cardinality routes
```
icyJoseph:
```suggestion
## Choosing what to prerender
```

docs/01-app/02-guides/incremental-static-regeneration-cache-components.mdx
```diff
@@ -198,6 +200,10 @@ After the first visit, Next.js renders the page in the background with the known
...
+
+Large or unbounded param sets, such as millions of product pages, aren't practical to prerender fully at build time. List the params worth prerendering in [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params) so those pages are prerendered in full. Visits to any other param get the App Shell instantly, while Next.js renders the full page in the background. Later visits to the same URL get the completed page from the cache.
```
icyJoseph: I'd say this isn't even about size. It is about whether or not it makes sense to materialize every route. 

Also associating this with cost, rather than speed, so that `generateStaticParams` isn't seen as an optimization, or a workaround for a limitation, this is a product/application consideration people have to make. For which routes should one pay the storage and CPU time to build them (at runtime if they are never hit, you don't pay for their generation either)

```suggestion
Not every route needs to be prerendered. Every page you prerender increases build work and produces output that has to be stored and deployed. Many routes may never be visited before your next deployment, making that work unnecessary.

Instead, use [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params) to prerender the routes that benefit most from being ready ahead of time, such as popular pages or predictable content. Less frequently visited routes are generated on demand and upgraded after their first visit, so you don't spend build time and storage on pages that may never be requested.
```

PR #95019, opened 2026-06-22
[turbopack] Add an experimental `chunkingHeuristics` to `next.config.js`

sampoder:
This PR adds new experimental options to `next.config.js`, they are: the request cost, “clusters” of routes where should chunk more (because the Z+Z case in `make_production_chunks` is more likely), “entry points” where we prioritize larger merged chunks for fast initial page loads (achieved by increasing P(N=1)), and a bounce rate to configure P(N=1) and P(N=2) based on site data.

### DIFF
--- a/packages/next/src/server/config-schema.ts
+++ b/packages/next/src/server/config-schema.ts
@@ -385,6 +385,14 @@ export const experimentalSchema = {
   turbopackRemoveUnusedImports: z.boolean().optional(),
   turbopackRemoveUnusedExports: z.boolean().optional(),
   turbopackScopeHoisting: z.boolean().optional(),
+  turbopackChunkingHeuristics: z
+    .object({
+      firstPageLoadPriority: z.number().min(0).max(1).optional(),
+      priorityRoutes: z.array(z.instanceof(RegExp)).optional(),
+      priorityBoost: z.number().min(1).optional(),
+      requestCost: z.number().min(0).max(1_000_000).optional(),
+    })
+    .optional(),
   turbopackWorkerAssetPrefix: z.string().optional(),
   turbopackClientSideNestedAsyncChunking: z.boolean().optional(),
   turbopackServerSideNestedAsyncChunking: z.boolean().optional(),
--- a/packages/next/src/server/config-shared.ts
+++ b/packages/next/src/server/config-shared.ts
@@ -765,6 +765,39 @@ export interface ExperimentalConfig {
   turbopackScopeHoisting?: boolean
 
   /**
+   * (`next --turbopack` only) Traffic-related hints for the production chunker. These change the
+   * assumptions Turbopack makes when making chunk merging decisions.
+   */
+  turbopackChunkingHeuristics?: {
+    /**
+     * This is a number between `0..1`, when higher, we weight the benefits of
+     * merging chunks for a signal page load higher. If you don't know a good
+     * number for this, your bounce rate is a good approximate for this value.
+     */
+    firstPageLoadPriority?: number
+    /**
+     * Regular expressions matching routes that are often the first page
+     * visited and whose client-side bundles should be merged more eagerly to reduce the single-route
+     * request cost (e.g. the homepage). This is at the cost of extra requests on other pages.
+     */
+    priorityRoutes?: RegExp[]
+    /**
+     * How much more eagerly to merge the client-side bundles of
+     * `priorityRoutes` routes, as a multiplier on their single-request probability (default
+     * `1.5`). Higher values merge more aggressively for those routes at the cost of extra requests
+     * elsewhere.
+     */
+    priorityBoost?: number
+    /**
+     * Estimated cost of an additional request, in bytes (uncompressed
+     * and unminfified bytes of code, default is 200 KB and the max is 1 MB), used by the chunker to
+     * trade off request count against preventing double-fetching. Uncompressed and unminfified code
+     * is approximately 5x the size of compressed and minified code.
+     */
+    requestCost?: number
+  }
+
+  /**
    * (`next --turbopack` only) A custom URL prefix for Web Worker URLs
    * produced by `new Worker(new URL(..., import.meta.url))` — both the
    * entrypoint URL and the module chunks loaded inside the worker —
--- a/packages/next/src/build/swc/index.ts
+++ b/packages/next/src/build/swc/index.ts
@@ -1027,6 +1027,25 @@ function bindingToApi(
       nextConfigSerializable.turbopack = turbopack
     }
 
+    // Serialize `experimental.turbopackChunkingHeuristics` route patterns: convert each RegExp to
+    // {source, flags} since RegExp objects are not JSON-serializable.
+    const chunkingHeuristics =
+      nextConfigSerializable.experimental?.turbopackChunkingHeuristics
+    if (chunkingHeuristics) {
+      const regexComponents = (regex: RegExp) => ({
+        source: regex.source,
+        flags: regex.flags,
+      })
+      nextConfigSerializable.experimental = {
+        ...nextConfigSerializable.experimental,
+        turbopackChunkingHeuristics: {
+          ...chunkingHeuristics,
+          priorityRoutes:
+            chunkingHeuristics.priorityRoutes?.map(regexComponents),
+        },
+      }
+    }
+
     return JSON.stringify(nextConfigSerializable, null, 2)
   }
 


packages/next/src/server/config-schema.ts
```diff
@@ -381,6 +380,14 @@ export const experimentalSchema = {
   turbopackRemoveUnusedImports: z.boolean().optional(),
   turbopackRemoveUnusedExports: z.boolean().optional(),
   turbopackScopeHoisting: z.boolean().optional(),
+  chunkingHeuristics: z
```
bgw: call this `turbopackChunkingHeuristics`

packages/next/src/server/config-shared.ts
```diff
@@ -748,6 +745,26 @@ export interface ExperimentalConfig {
    */
   turbopackScopeHoisting?: boolean
 
+  /**
+   * (`next --turbopack` only) Traffic-related hints for the production chunker. These change the
+   * assumptions Turbopack makes when making chunk merging decisions.
+   *
+   * - `clusters`: groups of pages commonly visited together, each defined by a list of wildcard
+   *   route patterns (`*` matches within a path segment, `**` across segments).
+   * - `bounceRate`: a number in `0..1`. Used at two-decimal (integer-percent) resolution.
```
bgw: - Why integer-percent resolution?
- This doesn't document what `bounceRate` *means*.

packages/next/src/server/config-shared.ts
```diff
@@ -748,6 +745,26 @@ export interface ExperimentalConfig {
    */
   turbopackScopeHoisting?: boolean
 
+  /**
+   * (`next --turbopack` only) Traffic-related hints for the production chunker. These change the
+   * assumptions Turbopack makes when making chunk merging decisions.
+   *
+   * - `clusters`: groups of pages commonly visited together, each defined by a list of wildcard
+   *   route patterns (`*` matches within a path segment, `**` across segments).
```
bgw: I would say `*` matches one path segment and `**` matches "one or more" path segments (or zero or more?)

packages/next/src/server/config-shared.ts
```diff
@@ -748,6 +745,26 @@ export interface ExperimentalConfig {
...
+   * - `bounceRate`: a number in `0..1`. Used at two-decimal (integer-percent) resolution.
+   * - `commonEntryPoints`: routes that are entry points and should be grouped more eagerly to
+   *   reduce the single-route request cost (e.g. the homepage).
```
bgw: This explanation doesn't clearly differentiate this from `clusters`

packages/next/src/server/config-shared.ts
```diff
@@ -748,6 +745,26 @@ export interface ExperimentalConfig {
...
+   * - `estimatedRequestCost`: estimated cost of an additional request, in bytes (uncompressed
+   *   bytes of code, default is 200 KB, max 1 MB), used by the chunker to trade off request count
+   *   against preventing double-fetching.
```
bgw: Uncompressed and unminfified. It would be neat to give the user a rough (order of magnitude) estimate of a common ratio of uncompressed+unminified bytes to actual network bytes after minification and gzip or brotli.

E.g. "we've found that this value is often about 10x the size of gzip and minified code" (or whatever it actually is)

It maybe also be useful to explain why this number is so awkwardly defined, e.g.:

"this metric represents uncompressed and unminified code because the chunking algorithm must run before the minifier."

bgw (approved):

bgw:
We should also probably also mention this option here? https://nextjs.org/docs/app/api-reference/turbopack#configuration

packages/next/src/server/config-shared.ts
```diff
@@ -755,6 +755,29 @@ export interface ExperimentalConfig {
    */
   turbopackScopeHoisting?: boolean
 
+  /**
+   * (`next --turbopack` only) Traffic-related hints for the production chunker. These change the
+   * assumptions Turbopack makes when making chunk merging decisions.
+   *
+   * - `clusters`: groups of pages commonly visited together, each defined by a list of wildcard
+   *   route patterns (`*` matches to one path segment; `**` matches to zero or more path segments).
```
bgw: We use a fork of `globset`, and I'm pretty sure we support the `[abc]` syntax, which would likely be ambiguous with next.js dynamic route segments.

@lukesandberg's suggestion was to use a RegExp object. If you do this, look through how the webpack loader logic converts RegExp to/from a JSON-serializable value.

https://docs.rs/globset/latest/globset/#syntax

bgw:
Maybe in a future version of this, we could consider moving some of the configuration into the route segment configs: https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config

There are already some undocumented options in there prefixed with `unstable_`, so we could do the same.

packages/next/src/server/config-shared.ts
sampoder: Switched to the pattern that was used by the webpack-loader logic:

```javascript
const regexComponents = (regex: RegExp) => ({
  source: regex.source,
  flags: regex.flags,
})
```

packages/next/src/server/config-shared.ts
```diff
@@ -755,6 +755,34 @@ export interface ExperimentalConfig {
...
+   *   elsewhere.
+   * - `estimatedRequestCost`: estimated cost of an additional request, in bytes (uncompressed
+   *   and unminfified bytes of code, default is 200 KB and the max is 1 MB), used by the chunker to
+   *   trade off request count against preventing double-fetching. Uncompressed and unminfified code
+   *   is approximately 5x the size of compressed and minified code.
+   */
+  turbopackChunkingHeuristics?: {
+    clusters?: RegExp[][]
+    bounceRate?: number
+    priorityEntryPoints?: RegExp[]
```
sokra:
```suggestion
    priorityRoutes?: RegExp[]
```

We usually don't use the word entrypoints publicly.

packages/next/src/server/config-shared.ts
```diff
@@ -755,6 +755,34 @@ export interface ExperimentalConfig {
...
+    priorityBoost?: number
+    estimatedRequestCost?: number
```
sokra:
```suggestion
    requestCost?: number
```

Align with css chunking `requestCost`.
I also think `heuristic` and `cost` already implies that it's a heuristic, so no `estimated` needed here

packages/next/src/server/config-shared.ts
```diff
@@ -755,6 +755,34 @@ export interface ExperimentalConfig {
...
+   *   visited and whose client-side bundles should be merged more eagerly to reduce the single-route
+   *   request cost (e.g. the homepage). This is at the cost of extra requests on other pages.
+   * - `priorityBoost`: how much more eagerly to merge the client-side bundles of
+   *   `priorityEntryPoints` routes, as a multiplier on their single-request probability (default
+   *   `1.5`). Higher values merge more aggressively for those routes at the cost of extra requests
+   *   elsewhere.
+   * - `estimatedRequestCost`: estimated cost of an additional request, in bytes (uncompressed
+   *   and unminfified bytes of code, default is 200 KB and the max is 1 MB), used by the chunker to
+   *   trade off request count against preventing double-fetching. Uncompressed and unminfified code
+   *   is approximately 5x the size of compressed and minified code.
```
sokra: Put these comments on the individual properties

packages/next/src/server/config-shared.ts
```diff
@@ -755,6 +755,34 @@ export interface ExperimentalConfig {
...
+   */
+  turbopackChunkingHeuristics?: {
+    clusters?: RegExp[][]
```
sokra: We often support glob or regex instead of only a regexp, as globs are usually easier to work with for users. But this can be a follow-up change

sampoder: Yep, will do this as a follow-up

packages/next/src/server/config-shared.ts
```diff
@@ -756,6 +756,43 @@ export interface ExperimentalConfig {
...
+    /**
+     * Groups of pages commonly visited together, each defined by a list of regular
+     * expressions matched against the route pathname.
+     */
+    clusters?: RegExp[][]
+    /**
+     * The percentage (a number between `0..1`) of visitors that "bounce" after visiting
+     * one page (they don't navigate to other pages on the site).
+     */
+    bounceRate?: number
```
bgw: Name this something more generic, like `firstPageLoadPriority`, and then in the comment suggest that if the user is unsure of what to set, bounce rate can be a good starting point.

Also: Make it clear that a priority of 0 optimizes for navigation over everything else, and a priority of 1 optimizes for cold initial page loads over navigation.

packages/next/src/server/config-shared.ts
```diff
@@ -756,6 +756,43 @@ export interface ExperimentalConfig {
...
 
+  /**
+   * (`next --turbopack` only) Traffic-related hints for the production chunker. These change the
+   * assumptions Turbopack makes when making chunk merging decisions.
+   */
+  turbopackChunkingHeuristics?: {
+    /**
+     * Groups of pages commonly visited together, each defined by a list of regular
+     * expressions matched against the route pathname.
+     */
```
bgw: This might benefit from an example code snippet in the comment, just because a 2d array is a little hard to conceptualize

PR #95021, opened 2026-06-22
[turbopack] Set `chunking_heuristics` on each `ChunkGroupInfo`

sampoder:
This PR changes `compute_chunk_group_info` so that chunks inherit heuristics from the entry chunk groups that reference them. For entry-points, if a chunk group is referenced by any entry point, it inherits a true value. For clusters, chunk groups inherit all the clusters that the groups that reference them are a part of.

### DIFF
--- a/turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
+++ b/turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
@@ -564,6 +564,9 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
             })
             .collect::<FxHashMap<_, _>>();
 
+        // `inherits_from[source]` is the set of chunk groups that inherit heuristics from `source`.
+        let mut inherits_from: FxHashMap<u32, RoaringBitmap> = FxHashMap::default();
+
         let visit_count = graph.traverse_edges_fixed_point_with_priority(
             entries
                 .iter()
@@ -658,15 +661,16 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
                         // Start of a new chunk group, don't inherit anything from parent
                         let chunk_group_ids = chunk_groups.map(|chunk_group| {
                             let len = chunk_groups_map.len();
-                            let is_merged = matches!(
-                                chunk_group,
-                                ChunkGroupKey::IsolatedMerged { .. }
-                                    | ChunkGroupKey::SharedMerged { .. }
-                            );
-                            match chunk_groups_map.entry(chunk_group) {
+                            // For merged groups, the parent group id whose heuristics they inherit.
+                            let merged_parent = match &chunk_group {
+                                ChunkGroupKey::IsolatedMerged { parent, .. }
+                                | ChunkGroupKey::SharedMerged { parent, .. } => Some(parent.0),
+                                _ => None,
+                            };
+                            let id = match chunk_groups_map.entry(chunk_group) {
                                 Entry::Occupied(mut e) => {
                                     let (id, merged_entries) = e.get_mut();
-                                    if is_merged {
+                                    if merged_parent.is_some() {
                                         merged_entries.insert(node);
                                     }
                                     **id
@@ -674,13 +678,26 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
                                 Entry::Vacant(e) => {
                                     let chunk_group_id = len as u32;
                                     let mut set = FxIndexSet::default();
-                                    if is_merged {
+                                    if merged_parent.is_some() {
                                         set.insert(node);
                                     }
                                     e.insert((ChunkGroupId(chunk_group_id), set));
                                     chunk_group_id
                                 }
+                            };
+                            // Record heuristics-inheritance edges into this chunk group: merged
+                            // groups inherit from their specific parent group; all other groups
+                            // inherit from every chunk group of the referencing module.
+                            if let Some(parent) = merged_parent {
+                                inherits_from.entry(parent).or_default().insert(id);
+                            } else if let Some((parent_module, _, _)) = parent_info
+                                && let Some(parent_groups) = module_chunk_groups.get(&parent_module)
+                            {
+                                for source in parent_groups.iter() {
+                                    inherits_from.entry(source).or_default().insert(id);
+                                }
                             }
+                            id
                         });
 
                         let chunk_groups =
@@ -803,10 +820,55 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
             }
         }
 
+        // Resolve per-chunk-group chunking heuristics. Entry chunk groups carry their route's
+        // priority-route flag; other chunk groups inherit it (OR) from their referencing chunk
+        // groups.
+        let mut priority_routes = RoaringBitmap::new();
+
+        let mut worklist: Vec<usize> = Vec::new();
+
+        for chunk_group in &entries {
+            let ChunkGroupEntry::Entry {
+                modules,
+                heuristics,
+            } = chunk_group
+            else {
+                continue;
+            };
+            if !heuristics.high_priority {
+                continue;
+            }
+            if let Some(index) =
+                chunk_groups_map.get_index_of(&ChunkGroupKey::Entry(modules.clone()))
+                && priority_routes.insert(index as u32)
+            {
+                worklist.push(index);
+            }
+        }
+
+        while let Some(source) = worklist.pop() {
+            let Some(targets) = inherits_from.get(&(source as u32)) else {
+                continue;
+            };
+            for target in targets.iter() {
+                let target = target as usize;
+                if target == source {
+                    continue;
+                }
+                if priority_routes.insert(target as u32) {
+                    worklist.push(target);
+                }
+            }
+        }
+
+        let chunk_group_priority_routes = RoaringBitmapWrapper(priority_routes);
+
         Ok(ChunkGroupInfo {
             module_chunk_groups: ResolvedVc::cell(module_chunk_groups),
             chunk_group_keys: chunk_groups_map.keys().cloned().collect(),
-            chunking_heuristics: ChunkingHeuristicsInfo::default(),
+            chunking_heuristics: ChunkingHeuristicsInfo {
+                priority_routes: chunk_group_priority_routes,
+            },
             chunk_groups: chunk_groups_map
                 .into_iter()
                 .map(|(k, (_, merged_entries))| match k {


turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -809,10 +809,141 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                if let Some(parent_groups) = module_chunk_groups.get(&parent_module) {
+                    for group in parent_groups.iter() {
+                        let group = group as usize;
+                        if group != target_index && seen_edges.insert((group, target_index)) {
+                            inherits_from[group].push(target_index);
+                        }
+                    }
+                }
+                Ok(())
+            })?;
```
sokra: This can be build up during the traversal above in the old code

sampoder: Switched

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -567,6 +567,9 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
             })
             .collect::<FxHashMap<_, _>>();
 
+        // `inherits_from[source]` is the set of chunk groups that inherit heuristics from `source`.
+        let mut inherits_from: FxHashMap<usize, FxHashSet<u32>> = FxHashMap::default();
```
sokra:
```suggestion
        let mut inherits_from: FxHashMap<usize, RoaringBitmap> = FxHashMap::default();
```

This could use a bitmap here... not sure if that's better.

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -661,34 +664,69 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
                                     }
                                     e.insert((ChunkGroupId(chunk_group_id), set));
                                     chunk_group_id
                                 }
+                            };
+                            // Merged groups inherit heuristics from their parent group.
+                            if let Some(parent) = merged_parent
+                                && parent != id
+                            {
+                                inherits_from.entry(parent as usize).or_default().insert(id);
```
sokra:
```suggestion
```

I would remove that here and integrate that into the logic below.

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -661,34 +664,69 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                                    | ChunkingType::Isolated {
+                                        merge_tag: None,
+                                        ..
+                                    }
+                                    | ChunkingType::Shared {
+                                        merge_tag: None,
+                                        ..
+                                    }
+                            )
+                            && let Some(parent_groups) = module_chunk_groups.get(&parent)
```
sokra:
```suggestion
                            && let Some(parent_groups) = merged_parent.unwrap_or_else(|| module_chunk_groups.get(&parent))
```

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -661,34 +664,69 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                                ChunkingType::Async
+                                    | ChunkingType::Isolated {
+                                        merge_tag: None,
+                                        ..
+                                    }
+                                    | ChunkingType::Shared {
+                                        merge_tag: None,
+                                        ..
+                                    }
+                            )
```
sokra:
```suggestion
```

When merged with the logic above, this is no longer needed?

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -661,34 +664,69 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                        {
+                            for source in parent_groups.iter() {
+                                for target in chunk_groups.iter() {
+                                    if source != target {
+                                        inherits_from
+                                            .entry(source as usize)
+                                            .or_default()
+                                            .insert(target);
+                                    }
+                                }
```
sokra: This could be a `|=` when using a bitmap

sokra: Direct cycles could be accepted and filtered in the walking below.

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -806,10 +844,76 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
             }
         }
 
+        // Resolve per-chunk-group chunking heuristics (`experimental.chunkingHeuristics`). Entry
+        // chunk groups carry their route's clusters / priority-entry-point flag; other chunk groups
+        // inherit the union of clusters (and OR of the flag) from their referencing chunk groups.
+        let chunk_group_count = chunk_groups_map.len();
+
+        let mut clusters: Vec<RoaringBitmap> = vec![RoaringBitmap::new(); chunk_group_count];
+        let mut priority_entry_points: Vec<bool> = vec![false; chunk_group_count];
```
sokra: This smells like with should be a RoaringBitmap instead (right from the start)

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -806,10 +844,76 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                chunk_groups_map.get_index_of(&ChunkGroupKey::Entry(modules.clone()))
+            {
+                clusters[index].extend(heuristics.clusters.iter().copied());
+                if heuristics.priority_entry_point {
+                    priority_entry_points[index] = true;
+                }
+            }
+        }
+
+        let mut worklist: Vec<usize> = (0..chunk_group_count).collect();
```
sokra: worklist could be initialized with the entries that have changes instead of all chunk groups.

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -806,10 +844,76 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+        while let Some(source) = worklist.pop() {
+            let source_clusters = clusters[source].clone();
```
sokra: This clone is very eager. You probably added it for lifetime resons, but I think we should remove it

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -806,10 +844,76 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+            let source_priority_entry_point = priority_entry_points[source];
+            let Some(targets) = inherits_from.get(&source) else {
+                continue;
+            };
+            for &target in targets.iter() {
+                let target = target as usize;
+                let mut changed = false;
+                if !source_clusters.is_empty() && !source_clusters.is_subset(&clusters[target]) {
```
sokra: You can get references to `clusters[source]` and `clusters[target]` at the same time using https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_disjoint_mut

Just add a `target != source` check somewhere

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -806,10 +844,76 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                    clusters[target] |= &source_clusters; // clusters[target] = clusters[target] | &source_clusters,
```
sokra: Instead of `is_subset` followed by `|=`, it should be cheaper to allways to `|=` and do the change detection by checking if the len changed: `changed = old_len != source_clusters.len()`

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
sampoder: oh this is neat!

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
sampoder: I ended up finding that it was easier to actually bring the logic up into this and have it all live here

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -576,6 +576,9 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
             })
             .collect::<FxHashMap<_, _>>();
 
+        // `inherits_from[source]` is the set of chunk groups that inherit heuristics from `source`.
+        let mut inherits_from: FxHashMap<usize, RoaringBitmap> = FxHashMap::default();
```
bgw:
```suggestion
        let mut inherits_from: FxHashMap<u32, RoaringBitmap> = FxHashMap::default();
```

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -815,10 +832,77 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                    continue;
+                }
+                let [source_clusters, target_clusters] =
+                    clusters.get_disjoint_mut([source, target]).unwrap();
+                let previous_target_clusters_len = target_clusters.len();
+                *target_clusters |= &*source_clusters;
+                let mut changed = previous_target_clusters_len != target_clusters.len();
+                if source_priority_route && priority_routes.insert(target as u32) {
+                    changed = true;
+                }
```
bgw:
```suggestion
                let mut changed = (
                  previous_target_clusters_len != target_clusters.len()
                  || (source_priority_route && priority_routes.insert(target as u32))
                );
```

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -815,10 +832,77 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                if changed {
+                    worklist.push(target);
+                }
+            }
+        }
+
+        let chunk_group_clusters: Vec<Vec<u16>> = clusters
```
bgw: Maybe in a PR above this stack, define a type alias for `ClusterId`? Otherwise it's a bit confusing where we're dealing with chunk group ids and where we're dealing with cluster ids.

bgw (approved):

PR #95024, opened 2026-06-22
fix(turbopack): allow #/* subpath imports

deepshekhardas:
Removes the `specifier.starts_with("#/")` check from `resolve_package_internal_with_imports_field` so that subpath imports like `#/greeting.js` are correctly resolved.

According to the Node.js spec, `#/` is a valid prefix for subpath import patterns (e.g. `"#/*": "./src/*"`). The removed check was treating these as invalid.

Closes #94461


gaearon:
https://github.com/vercel/next.js/pull/94461 landed which does the same

PR #95025, opened 2026-06-22
fix: ignore stray parent lockfiles in find-root (#94597)

deepshekhardas:
Refactors `findRootDirAndLockFiles()` to use a custom `findUpFile()` implementation instead of the `find-up` package, so that stray lockfiles in parent directories without a `package.json` (or without `workspaces` declared) are properly ignored.

Previously, any lockfile found walking up the directory tree was treated as a valid workspace root, causing incorrect root inference in monorepo setups where parent directories happened to contain a lockfile (e.g., from a different project or a container image).

Changes:
- Replace `find-up` with a lightweight `findUpFile()` that supports a candidate predicate
- Skip lockfiles in directories without `package.json`
- Skip parent directories that have `package.json` but no `workspaces` field
- Keep `pnpm-workspace.yaml` as an explicit, always-honored root marker
- Move lockfile name list to a module-level constant

Closes #94597

### DIFF
--- a/packages/next/src/lib/find-root.test.ts
+++ b/packages/next/src/lib/find-root.test.ts
@@ -1,0 +1,98 @@
+import { mkdtemp, mkdir, rm, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+import { findRootDirAndLockFiles } from './find-root'
+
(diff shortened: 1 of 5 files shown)


deepshekhardas:
Superseded by upstream #96159 (Improve root detection to handle worktrees, more workspaces & stray lockfiles) which reimplements this fix more comprehensively.

PR #95026, opened 2026-06-22
[turbopack] Use experimental chunking heuristics

sampoder:

### DIFF
--- a/turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
+++ b/turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
@@ -19,6 +19,9 @@ use crate::{
     },
 };
 
+/// Default estimated cost of an additional request, in bytes (200 KB).
+const DEFAULT_ESTIMATED_REQUEST_COST_BYTES: u64 = 200_000;
+
 pub async fn make_production_chunks(
     chunk_items: Vec<&ChunkItemOrBatchWithInfo>,
     batch_groups: Vec<ResolvedVc<ChunkItemBatchGroup>>,
@@ -37,6 +40,8 @@ pub async fn make_production_chunks(
     let span = span_outer.clone();
     async move {
         let module_chunk_groups = module_graph.chunk_group_info().module_chunk_groups();
+        let chunk_group_info = module_graph.chunk_group_info().await?;
+        let heuristics = &chunk_group_info.chunking_heuristics;
         let merged_modules = module_graph.merged_modules().await?;
 
         #[derive(Default)]
@@ -123,6 +128,9 @@ pub async fn make_production_chunks(
             min_chunk_size,
             max_chunk_count_per_group,
             max_merge_chunk_size,
+            first_page_load_priority,
+            priority_boost_percent,
+            request_cost,
             ..
         } = chunking_config;
 
@@ -217,6 +225,24 @@ pub async fn make_production_chunks(
                     unreachable!();
                 };
 
+                // Chunking-heuristics-derived constants for the maths below.
+                //
+                // The cost of a single request in transferred bytes, capped at 1MB.
+                // Defaults to 200,000 bytes (200 KB).
+                let c_req = request_cost
+                    .unwrap_or(DEFAULT_ESTIMATED_REQUEST_COST_BYTES)
+                    .min(1_000_000) as i64;
+
+                // Default `P(N = 1)`: the probability that we request exactly 1 chunk group.
+                // An integer percentage (0..=100). `firstPageLoadPriority` maps directly to
+                // it; the default is 67 (~2/3).
+                let default_p1 = first_page_load_priority.map_or(67, |percent| percent as i64);
+
+                // `priorityBoost` as an integer percentage to multiply `P(N = 1)` by for priority
+                // routes; the default is 150 (a 1.5x boost).
+                let priority_boost_percent =
+                    priority_boost_percent.map_or(150, |percent| percent as i64);
+
                 let mut iterations = 0;
                 while chunks_to_merge.len() > 1 {
                     // Find best candidate
@@ -286,15 +312,16 @@ pub async fn make_production_chunks(
                             */
 
                             /*
-                                For our calculations we assume that there is a probability of 2/3 that we request exactly 1 chunk group (`N = 1`)
-                                and a probability of 1/3 that we request 2 chunk groups (`N = 2`).
-                                This is a simplification, but it should be good enough for our purposes.
+                                By default, for our calculations we assume that there is a probability of 2/3 that
+                                we request exactly 1 chunk group (`N = 1`) and a probability of 1/3 that we request
+                                2 chunk groups (`N = 2`). This is a simplification, but it should be good enough
+                                for our purposes and it is configurable using the chunking heuristics.
 
                                 We want to compute the expected request count `e_req` and the expected total requested size `e_size` for the unmerged and merged case.
 
                                 To compute that we compute the two cases `N = 1` and `N = 2` and combine them
-                                e_size = 2/3 * e_size(N = 1) + 1/3 * e_size(N = 2)
-                                e_req = 2/3 * e_req(N = 1) + 1/3 * e_req(N = 2)
+                                e_size = P(N = 1) * e_size(N = 1) + P(N = 2) * e_size(N = 2)
+                                e_req = P(N = 1) * e_req(N = 1) + P(N = 2) * e_req(N = 2)
 
                                 We combine `e_size` with `e_req` using this formula:
                                 e_cost = e_req * c_req + e_size
@@ -311,8 +338,8 @@ pub async fn make_production_chunks(
                                 d_req = e_req_unmerged - e_req_merged
 
                                 And we can split it further for every N:
-                                d_size = 2/3 * d_size(N = 1) + 1/3 * d_size(N = 2)
-                                d_req = 2/3 * d_req(N = 1) + 1/3 * d_req(N = 2)
+                                d_size = P(N = 1) * d_size(N = 1) + P(N = 2) * d_size(N = 2)
+                                d_req = P(N = 1) * d_req(N = 1) + P(N = 2) * d_req(N = 2)
                             */
 
                             /*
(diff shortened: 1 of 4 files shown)


turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +470,57 @@ pub async fn make_production_chunks(
...
                             }
                             let rem_g = groups - 1;
-                            let c_req = 200000;
-                            // d3 = 3 * d
-                            let pre_d3 = c_req * (2 * rem_g + (o_groups - 1))
-                                - 2 * (a_rem * a_size + b_rem * b_size);
+
+                            // If a chunk group is referenced by an entry point
+                            // we increase its P(N=1) by 1.5x (and as a result
+                            // reduce P(N=2)). This is to encourage merging chunks
```
bgw:
```suggestion
                            // reduce P(N=2). This is to encourage merging chunks
```

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -217,6 +220,35 @@ pub async fn make_production_chunks(
                     unreachable!();
                 };
 
+                // `experimental.chunkingHeuristics`-derived constants for the maths below.
+                //
+                // The cost of a single request in transferred bytes, capped at 1MB.
+                // Defaults to 200,000 KB.
```
bgw:
```suggestion
                // Defaults to 200,000 Bytes.
```

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +470,57 @@ pub async fn make_production_chunks(
...
+                            // reduce P(N=2)). This is to encourage merging chunks
+                            // used on these entry points.
+                            let is_entry_point = candidate
+                                .chunk_groups
+                                .as_ref()
+                                .is_some_and(|g| !g.is_disjoint(&entry_point_groups))
+                                || other
+                                    .chunk_groups
+                                    .as_ref()
+                                    .is_some_and(|g| !g.is_disjoint(&entry_point_groups));
```
bgw: As discussed IRL, this union isn't right

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +470,57 @@ pub async fn make_production_chunks(
...
+                            } else {
+                                default_p1
+                            };
+
+                            let p2 = 100 - p1;
+
+                            // If the chunk groups sharing both chunks include two routes in
+                            // the same cluster, weight the `N = 2` Z+Z request benefit
+                            // higher so commonly co-visited pages merge more readily.
+                            let mut wz: i64 = 1;
```
bgw: Define what "w" means in this context

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -217,6 +220,31 @@ pub async fn make_production_chunks(
...
                 };
 
+                // `experimental.chunkingHeuristics`-derived constants for the maths below.
+                //
+                // The cost of a single request in transferred bytes, capped at 1MB.
+                // Defaults to 200,000 KB.
+                let c_req = heuristics
+                    .estimated_request_cost
+                    .unwrap_or(200_000)
+                    .min(1_000_000) as i64;
```
sampoder: make the default 200_000

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -217,6 +220,31 @@ pub async fn make_production_chunks(
                     unreachable!();
                 };
 
+                // `experimental.chunkingHeuristics`-derived constants for the maths below.
+                //
+                // The cost of a single request in transferred bytes, capped at 1MB.
+                // Defaults to 200,000 KB.
+                let c_req = heuristics
+                    .estimated_request_cost
+                    .unwrap_or(200_000)
```
sokra: Define some constants

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +466,66 @@ pub async fn make_production_chunks(
...
-                            let c_req = 200000;
-                            // d3 = 3 * d
-                            let pre_d3 = c_req * (2 * rem_g + (o_groups - 1))
-                                - 2 * (a_rem * a_size + b_rem * b_size);
+
+                            // If the chunk groups sharing both chunks include two that are used
+                            // by the same entry point, we increase its P(N=1) by 1.5x (and as a
+                            // result reduce P(N=2)). This is to encourage merging chunks used on
+                            // these entry points.
+                            let mut is_entry_point = false;
```
sokra: is_entry_point -> is shared by the same common entry point

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```suggestion
                            if overlap < 1 {
```

Separate PR

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +466,66 @@ pub async fn make_production_chunks(
...
+                                let mut seen = RoaringBitmap::new(); // entry points that these chunk groups are for
+                                for index in o.iter() {
+                                    let entry_points =
+                                        &heuristics.common_entry_points[index as usize].0;
+                                    if !seen.is_disjoint(entry_points) {
+                                        is_entry_point = true;
+                                        break;
+                                    }
+                                    seen |= entry_points;
+                                }
```
sokra: I think that can be simplified to be `overlap is an common entry point` so assuming you make `common_entry_points` a RoaringBitmap => `is_entry_point = !o.is_disjoint(common_entry_points)`

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +466,66 @@ pub async fn make_production_chunks(
...
+                            }
+
+                            let p1 = if is_entry_point {
+                                (default_p1 * 3 / 2).min(100)
```
sokra: Since you are defining config options for everything anyway, make the 3 / 2 here also a config options `priority_boost` or something like that

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: Made in https://github.com/vercel/next.js/pull/95102

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: Yeah - I went back over this and this was overkill for what it was doing. `common_entry_points` is now a `RoaringBitmap` and yep, that's what I'm doing. thanks for catching this!

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -374,14 +413,23 @@ pub async fn make_production_chunks(
...
+
+                                W(Z + Z) = 1
+                                W(X + Z) = 1
+                                W(Y + Z) = 1
+
+                                However, W(Z + Z) = 2, if the chunk items are in multiple groups in a
+                                "cluster" (see `experimental.turbopackChunkingHeuristics`).
+
+                                d_req_z_z = W(Z + Z) * ((o_groups/groups) * (o_groups - 1)/rem_g) * (2 - 1)
+                                          = W(Z + Z) * (o_groups * (o_groups - 1) / (groups * rem_g))
```
sokra: That sounds really weird to me... with `W(Z + Z) = 2` we have a total propability of > 1 which doesn't sound correct.

I think what we have to do here is to split `Z + Z` (and the other cases) into `Z + Z (same cluster)` and `Z + Z (different cluster)` and adjust the propabilities based on that.

But that will probably become a bit messy as technically we have to walk all chunk groups of Z and compare their clusters with all clusters of all other chunk groups of Z. That also sounds a bit expensive as it's `O(Z * Z * N)` where N is the avg number of clusters. And we have to do that for every cost computation. And also for the `X + Z` and `Y + Z` cases.

I think I would feel better if we land the initial feature without the cluster feature and only ship the `priority` initially.

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +472,66 @@ pub async fn make_production_chunks(
...
+                            // If the chunk groups sharing both chunks include two routes in
+                            // the same cluster, weight the `N = 2` Z+Z request benefit
+                            // higher so commonly co-visited pages merge more readily.
+                            // This variable is W(Z + Z) above.
+                            let mut wz: i64 = 1;
+                            if has_clusters
+                                && let (Some(a), Some(b)) =
+                                    (&candidate.chunk_groups, &other.chunk_groups)
+                            {
+                                let o = &***a & &***b; // `o_groups`
```
sokra: That looks a bit repeated from above

turbopack/crates/turbopack-core/src/chunk/chunking_context.rs
```diff
@@ -260,6 +260,19 @@ pub struct ChunkingConfig {
     /// type.
     pub style_groups_algorithm: StyleGroupsAlgorithm,
 
+    /// `experimental.turbopackChunkingHeuristics.bounceRate` as an integer percentage (`0..=100`),
```
sokra:
```suggestion
    /// `bounceRate` as an integer percentage (`0..=100`),
```

Do not mention next.js config options in Turbopack

turbopack/crates/turbopack-core/src/chunk/chunking_context.rs
```diff
@@ -260,6 +260,19 @@ pub struct ChunkingConfig {
...
+    /// or `None` to use the default. Used by the production chunker's merge heuristics.
+    pub bounce_rate_percent: Option<u32>,
+
+    /// `experimental.turbopackChunkingHeuristics.priorityBoost` as an integer percentage (e.g.
```
sokra:
```suggestion
    /// `priorityBoost` as an integer percentage (e.g.
```

turbopack/crates/turbopack-core/src/chunk/chunking_context.rs
```diff
@@ -260,6 +260,19 @@ pub struct ChunkingConfig {
...
+    /// `150` for a 1.5x boost), or `None` to use the default. Used by the production chunker's
+    /// merge heuristics.
+    pub priority_boost_percent: Option<u32>,
+
+    /// `experimental.turbopackChunkingHeuristics.estimatedRequestCost` in bytes, or `None` to use
```
sokra:
```suggestion
    /// `estimatedRequestCost` in bytes, or `None` to use
```

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: @bgw and I talked today, I think I'm going to strip out the clusters feature and put it in a separate PR for review after everything else is merged

sampoder: I've stripped the cluster feature out of this stack and into a separate PR so we should be able to ship the priority stuff!

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -19,6 +19,9 @@ use crate::{
     },
 };
 
+/// Default estimated cost of an additional request, in bytes (200 KB).
+const DEFAULT_ESTIMATED_REQUEST_COST: u64 = 200_000;
```
bgw:
```suggestion
const DEFAULT_ESTIMATED_REQUEST_COST_BYTES: u64 = 200_000;
```

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -270,6 +296,12 @@ pub async fn make_production_chunks(
                             let b_rem = b_groups - o_groups;
 
                             /*
+                                This comment describes the algorithm with the default
+                                probabilities, 2/3 for N=1 and 1/3 for N=2 as well as
+                                the default request-cost. The six combinations in the
+                                N=2 case are also weighted equally. These things can
+                                change if custom chunking heuristics are configured.
```
bgw: Delete comment

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +457,46 @@ pub async fn make_production_chunks(
...
+                            // in `o_groups` would be a chunk group that requests both chunk items.
+
+                            let mut is_priority_route = false;
+                            if let (Some(a), Some(b)) =
+                                (&candidate.chunk_groups, &other.chunk_groups)
+                            {
+                                let o = &***a & &***b; // `o_groups`
+
+                                // if there is one chunk group in `o_groups` that is a priority
+                                // route, we should prioritise merging these two chunk items.
```
bgw:
```suggestion
                                // if there is one chunk group in `o_groups` that is used by a priority
                                // route, we should prioritise merging these two chunk items.
```

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -431,16 +457,46 @@ pub async fn make_production_chunks(
...
+                                is_priority_route = !o.is_disjoint(&heuristics.priority_routes);
```
bgw, line 475: Possible future improvement? Look at what percentage of the overlap is in priority groups, and scale the boost by that?

bgw (approved):

PR #95060, opened 2026-06-22
test: support local packages in deploy tests

timneutkens:
### What?

Add a local deploy-test workflow that packages the current Next.js checkout and deploys isolated test fixtures with those tarballs. Turbopack is the default, with an explicit webpack variant.

### Why?

Deployment tests previously required a published commit tarball or an existing deployment URL, making it difficult to validate uncommitted framework changes against the Vercel deployment environment.

### How?

- Add `test-deploy-local` scripts that run `pack-next --tar` once before Jest.
- Copy the generated package tarballs into each isolated deploy fixture and patch its exact `package.json` with relative `file:` references.
- Validate conflicting deploy inputs, preserve existing package overrides, and document the local and reusable two-step workflows.

### Verification

- `pnpm exec jest --runInBand test/unit/deployable-tarballs/deployable-tarballs.test.ts`
- `VERCEL_TURBOPACK_TEST_TEAM=vtest314-next-turbo-e2e-tests NEXT_TEST_DEPLOY_TARBALLS_DIR=$PWD/tarballs NEXT_TEST_PREFER_OFFLINE=1 NEXT_E2E_TEST_TIMEOUT=240000 pnpm test-deploy-turbo test/e2e/test-utils-tests/basic/basic.test.ts`
- `bash -n scripts/run-jest.sh`
- Not passing: `pnpm --filter=next types` (existing missing compiled MCP, Vercel Blob, URLPattern, and postcss-scss declarations)


test/lib/next-modes/deployable-tarballs.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export async function prepareDeployableTarballs(
+  sourceDir: string,
+  projectDir: string
+): Promise<void> {
+  const resolvedSourceDir = path.resolve(sourceDir)
+  const targetDir = path.join(projectDir, 'tarballs')
+
+  await Promise.all(
```
timneutkens: Use `for...in` and use `fs.accessSync` or `fs.existsSync`

timneutkens: :robot: Updated the tarball preflight to use a for...in loop with existsSync before the async copy step.

PR #95062, opened 2026-06-22
Handle registries not serving JSON when checking the version

### LINKED ISSUE #95061: `getVersionInfo()` throw error in dev server when npm proxy return HTML resp instead of json

Provied small snippets to tests in addition context

### To Reproduce

`getVersionInfo()` in [packages/next/src/server/dev/hot-reloader-shared-utils.ts#L17-L23](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/dev/hot-reloader-shared-utils.ts#L17-L23) fetches `https://registry.npmjs.org/-/package/next/dist-tags` to check version staleness

In corporate environments where direct access to the public npm registry is blocked and routed through a private Artifactory proxy, this URL returns an HTML redirect/error forbidden page instead of JSON.

Current Behavior
- res.ok is true (the proxy returns HTTP 200 with an HTML body)
- res.json() throws a SyntaxError because the response is HTML, not JSON
- The dev server received error:
```
SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
    at JSON.parse (<anonymous>)
```

# Expected Behavior

`getVersionInfo()` should resolve the registry URL via the existing getRegistry() helper in [packages/next/src/lib/helpers/get-registry.ts](https://github.com/vercel/next.js/blob/canary/packages/next/src/lib/helpers/get-registry.ts), which is already used in other parts of the codebase and correctly respects the user's configured npm registry.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.50: Mon Apr 27 20:41:06 PDT 2026; root:xnu-12377.121.6-2/RELEASE_ARM64_T6030
  Available memory (MB): 18432
  Available CPU cores: 12
Binaries:
  Node: 24.16.0
  npm: 11.13.0
  Yarn: N/A
  pnpm: 9.15.9
Relevant Packages:
  next: 16.2.9 // Latest available version is detected (16.2.9)
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 6.0.3
Next.js Config:
  output: N/A
```

Not sure

next dev (local)

### Additional context

It would be easy to test by doing fetch html request and `res.json`
```
await fetch('https://httpbin.org/html').then(async res => await res.json())
```
Image

vctqs1:
Fixes #95061
- Update to get registry via `getRegistry`
- Moving step to `res.json` inside try/catch to ignore fetch errors + `.json` if response is not `JSON`


packages/next/src/server/dev/hot-reloader-shared-utils.ts
```diff
@@ -13,15 +14,25 @@ export async function getVersionInfo(): Promise<VersionInfo> {
...
+          }
+
+          return await response.json()
+        }
+      )
     } catch {
       // ignore fetch errors
     }
 
-    if (!res || !res.ok) return { installed, staleness: 'unknown' }
```
eps1lon, line 22: Wouldn't this have prevented accidentally assuming JSON for proxies protecting the access? The proxy should return the proper error code instead of claiming an ok request that actually wasn't.

I think we should be able to assume that a 2xx registry response is JSON.

vctqs1: Yeah i understand that, bu under corporate VPN if we try to access a restricted site like public npm registry then it will redirect to generic page to say that `public npm` is forbidden.
Mean it still return `res.ok` for that redirect happen. but the response now is HTML so `.json` lead to issue

vctqs1: Like this 
image

eps1lon: If it's forbidden, it should return a 403 error code. That would fix the issue automatically without requiring any new code being added to Next.js.

vctqs1anz: I understand your point, and ideally this should return a proper 403.
However, in practice this behavior is outside of our control — the current flow performs a 307 redirect to a page that displays a UI and eventually returns an HTML response with status 200.

Given this constraint, in Next.js we already handle this pattern using `try { ... } catch` with intentional error suppression around fetch. I think applying the same approach here would be a reasonable way to handle this.
And `getRegistry` fn was existing there to use together

vctqs1: hey @eps1lon Is that good or you would prefer to move res.json outsite the try/catch

eps1lon: We can't add code for every broken 3rd party service. The contract is pretty clear. We can check the content-type the registry responded with instead of a try-catch.

vctqs1: ah right, we can check content-type

vctqs1: Hey @eps1lon  I keep the things remain, then only add one more `res.headers?.get?.('content-type') !== 'application/json'` check before res.json + `getRegistry()` change above

packages/next/src/server/dev/hot-reloader-shared-utils.ts
```diff
@@ -13,13 +14,20 @@ export async function getVersionInfo(): Promise<VersionInfo> {
...
+      res = await fetch(`${registry}-/package/next/dist-tags`)
     } catch {
       // ignore fetch errors
     }
 
-    if (!res || !res.ok) return { installed, staleness: 'unknown' }
+    if (
+      !res ||
+      !res.ok ||
+      res.headers?.get?.('content-type') !== 'application/json'
```
eps1lon: if it's undefined, we should assume a correct registry i.e. one that has valid `.json()`

eps1lon (changes requested):

packages/next/src/server/dev/hot-reloader-shared-utils.ts
vctqs1: yeah nice, i updated it thanks @eps1lon

eps1lon (approved):

vctqs1:
Hi @eps1lon  coud you help have alook, anything I can do?

vctqs1:
Hi @eps1lon i saw the test still failed. anything I can do here?

packages/next/src/server/dev/hot-reloader-shared-utils.ts
```diff
@@ -13,13 +14,22 @@ export async function getVersionInfo(): Promise<VersionInfo> {
     let res
 
     try {
+      const registry = getRegistry()
```
eps1lon: This needs to be async. Otherwise it's blocking too long which is a known issue described at the callsite: https://github.com/vercel/next.js/blob/000a91a76b4c3cf479046ad65dd0aaf753df3213/packages/next/src/server/dev/hot-reloader-turbopack.ts#L1229-L1232

eps1lon (changes requested):

packages/next/src/server/dev/hot-reloader-shared-utils.ts
vctqs1: Soo I need to add a change in `getRegistry` to remove `execSync`?

eps1lon: Check is still relevant but it needs to be async.

vctqs1: sorry i dont get it. because `getRegistry` is sync function. so i would understand if you mean that is change `getRegistry` from `sync` to `async` or just call it as async here @eps1lon 
```
      const registry = await getRegistry()
```

vctqs1: image

eps1lon: And actually make its body async.

vctqs1: Updated it @eps1lon thank you

packages/next/src/lib/helpers/get-registry.ts
```diff
@@ -1,34 +1,36 @@
...
-      {
-        env: {
-          ...process.env,
-          NODE_OPTIONS: getFormattedNodeOptionsWithoutInspect(),
-        },
-      }
-    )
-      .toString()
-      .trim()
+    const { stdout } = await execFileAsync(pkgManager, args, {
```
eps1lon, line 27: Just highlighting that this also switches from "exec" to "execFile" which means we no longer spawn a new shell (faster). We always used "exec".  https://github.com/vercel/next.js/pull/44234 just copied this from existing code but I couldn't find any comments if we need really need "exec" i.e. a new shell.

vctqs1:
Hi @eps1lon do you have any suggest to fix the new fail ones ??

vctqs1:
Hi @eps1lon could you have a look into that CI failure job.

eps1lon (approved):

packages/next/src/server/dev/hot-reloader-shared-utils.ts
```diff
@@ -10,18 +11,25 @@ export async function getVersionInfo(): Promise<VersionInfo> {
...
-      // ignore fetch errors
+      return { installed, staleness: 'unknown' }
     }
+    const contentType = response.headers.get('content-type')
 
-    if (!res || !res.ok) return { installed, staleness: 'unknown' }
+    if (
+      !response.ok ||
+      // only reject if content-type is explicitly not JSON, absent content-type is treated as valid registry response
+      (contentType !== null && contentType !== 'application/json')
```
eps1lon, line 27: could be `application/json; charset=utf-8` so this is too fragile. try-catch it is then.

eps1lon (changes requested):

packages/next/src/lib/helpers/get-registry.ts
eps1lon: Actually it's not under Windows so this should use `exec`

packages/next/src/server/dev/hot-reloader-shared-utils.ts
vctqs1: I updated it to `(contentType !== null && !contentType.includes('application/json'))`

vctqs1: Pls help me review PR again @eps1lon

PR #95065, opened 2026-06-22
[turbopack] Sort modules in files by path not ID

sampoder:
Sorting-by-ID was flagged as bad for g-zipping:

> This sorts by chunk item id.
> Tobias has been saying that this is bad, because it gzips worse
> We should be sort by path, so that similar modules stay together. E.g. if you have many similar icon modules
> Sorting by chunk item results in random order

Switched to sorting by path!


turbopack/crates/turbopack-ecmascript/src/chunk/code_and_ids.rs
```diff
@@ -13,7 +14,7 @@ use crate::chunk::{
 };
 
 #[turbo_tasks::value(transparent, serialization = "skip")]
-pub struct CodeAndIds(SmallVec<[(ModuleId, ReadRef<Code>); 1]>);
+pub struct CodeAndIds(SmallVec<[(ModuleId, ReadRef<Code>, RcStr); 1]>);
```
bgw: Now this name doesn't make sense. `CodeModuleIdsAndAssetIdents`?

turbopack/crates/turbopack-ecmascript/src/chunk/code_module_ids_and_paths.rs
```diff
@@ -49,7 +50,8 @@ pub async fn item_code_and_ids(
         }) => {
             let id = chunk_item.id().await?;
             let code = chunk_item.code(async_info.map(|info| *info));
-            smallvec![(id, code.await?)]
+            let path = chunk_item.asset_ident().to_string().owned().await?;
```
bgw, line 58: This is getting path through `AssetIdent`

turbopack/crates/turbopack-browser/src/ecmascript/content.rs
```diff
@@ -119,9 +119,11 @@ impl EcmascriptBrowserChunkContent {
 
         let content = this.content.await?;
         let mut chunk_items = content.chunk_item_code_and_ids().await?;
-        chunk_items.sort_by_key(|item| item.first().map(|(id, _)| id.clone()));
+        // Sort items by their module path so that similar modules stay
+        // together so that the chunks gzips better.
+        chunk_items.sort_by_key(|item| item.first().map(|(_, _, path)| path.clone()));
```
bgw: Let's sort by path, and then the rest of the id.

If it's easy to add `Ord` to `AssetIdent` and get an asset ident here, do that, otherwise just produce tuples of `(path, id)`.

turbopack/crates/turbopack-nodejs/src/ecmascript/node/content.rs
```diff
@@ -70,7 +70,7 @@ impl EcmascriptBuildNodeChunkContent {
         let content = self.content.await?;
         let chunk_items = content.chunk_item_code_and_ids().await?;
         for item in &chunk_items {
-            for (id, item_code) in &**item {
+            for (id, item_code, _) in &**item {
```
bgw: Let's also sort this by path to get the same gzip benefits.

bgw (approved):

turbopack/crates/turbopack-ecmascript/src/chunk/code_and_ids.rs
sampoder: Fixed, changed to `CodeModuleIdsAndPaths`

turbopack/crates/turbopack-ecmascript/src/chunk/code_module_ids_and_paths.rs
sampoder: Talked IRL - deriving `Ord` on `AssetIdent` is challenging because of `FileSystemPath`

turbopack/crates/turbopack-browser/src/ecmascript/content.rs
sampoder: https://github.com/vercel/next.js/pull/95065#discussion_r3455633309

PR #95079, opened 2026-06-23
Turbopack: Avoid clone when sorting by module file path

bgw:
Minor nit: In the worst-case (module id can be an `RcStr` or a u64), this eliminates two `RcStr` clones+drops per iteration of the sort algorithm.

The code is cleaner with `sort_by_key`, but `sort_by_key` introduces an artificial lifetime issue that `sort_by` avoids.

This is a long-standing and widely documented language issue: https://github.com/rust-lang/rust/issues/34162


mischnic (approved):

sampoder:
thanks!

PR #95083, opened 2026-06-23
Fix false-positive `export const dynamic` in Cache Components detection

eps1lon:
Resolves https://github.com/vercel/next.js/pull/94955/changes#r3439357143

I missed this during rebase.

### DIFF
--- a/packages/next/src/server/app-render/create-component-tree.tsx
+++ b/packages/next/src/server/app-render/create-component-tree.tsx
@@ -285,14 +285,6 @@ async function createComponentTreeInternal(
   }
 
   if (typeof dynamic === 'string') {
-    if (cacheComponents) {
-      // TODO: Remove this since this should be caught during compilation.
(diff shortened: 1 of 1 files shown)


packages/next/src/server/app-render/create-component-tree.tsx
```diff
@@ -285,14 +285,6 @@ async function createComponentTreeInternal(
...
 
   if (typeof dynamic === 'string') {
-    if (cacheComponents) {
-      // TODO: Remove this since this should be caught during compilation.
-      // Left as a guard in case usage of Postpone was attempted
-      throw new Error(
-        `Route segment config "dynamic" is not compatible with \`nextConfig.cacheComponents\`. Please remove it.`
-      )
-    }
-
```
eps1lon, line 295: At this point `dynamic` could be a derived value (e.g. L278) not just from `export const dynamic`.

unstubbable (approved):

PR #95085, opened 2026-06-23
[cd] publish on release tags instead of branch push

eps1lon:
That way we can push multiple commits when doing a release. For canaries, nothing will change.

In a follow-up, we'll start cutting `@preview` from the `canary` branch. To support that case, we need to immediately revert the version in `lerna.json`. We shouldn't push commits indidually to avoid having an incomplete release if something else is pushed between `@preview` version bumps and `@canary` version reverts.

## test plan

After merge, the next canary release should publish from the `v…-canary.N` tag run (`deploy-target = production`), while the canary branch push for the same commit shows `skipped` — exactly one production run per release.

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -3,9 +3,13 @@ name: build-and-deploy
 
 on:
   push:
-    # Don't run when tags or graphite base branches are pushed
+    # Don't run when graphite base branches are pushed
     branches-ignore:
       - 'graphite-base/**'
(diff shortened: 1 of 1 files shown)


bgw (approved):

eps1lon:
✅  release commit being skipped: https://github.com/vercel/next.js/actions/runs/28081340901/job/83136782799#step:7:15
✅ release tag being built: https://github.com/vercel/next.js/actions/runs/28081340543/job/83136781741#step:7:15

PR #95088, opened 2026-06-23
Turbopack: replace cssChunking graph `moduleFactorCost` with `weightDistribution`

sokra:
## Summary

Reworks the cost model of the experimental graph CSS chunking algorithm (`experimental.cssChunking: 'graph'`, Turbopack only) and retunes its defaults.

- **Replaces the `moduleFactorCost` option with `weightDistribution`.** The old `moduleFactorCost` term penalized a chunk purely by `chunk_size / group_total_size`, which charged a chunk group even for CSS it fully needs and never actually measured overshipping.
- **New per-group cost** is `chunk_group_weight * (chunk_size + request_cost)`, summed over the chunk groups that load a chunk, where `chunk_group_weight = group_total_size ^ (-weightDistribution)` is precomputed once per group. `weightDistribution = 0` weights every chunk group equally; higher values give smaller chunk groups a larger weight, so the algorithm overships less to small pages at the cost of more requests. The size weighting subsumes the explicit overship penalty, so the metric stays chunk-local and the greedy merger is unchanged.
- **Retunes defaults:** `requestCost` `20000` → `100000` (bytes) and `weightDistribution` default `0.1`.

Config shape (object form): `{ type: 'graph', requestCost?, weightDistribution? }`. The change is wired end to end through the config schema/types, `StyleGroupsAlgorithm::Graph`, and the chunking algorithm.

## Verification

- `cargo test -p turbopack-core --lib style_groups_graph` (54 passed, incl. a new test asserting `weightDistribution` keeps an unneeded module out of a small group's chunk)
- `pnpm test-start-turbo test/e2e/app-dir/css-order/css-order.test.ts` (243 passed, 12 todo) — confirms chunk shapes/request counts at the new defaults across the `graph` and `graph-object` modes


lukesandberg (changes requested):
is `weightDistribution` the best name for this parameter?

it is really about preferring smaller chunk groups or not... but im not sure as a user i would really know what a chunk group is and even if i did i wouldn't really know what the size is

sokra:
> is `weightDistribution` the best name for this parameter?
> 
> it is really about preferring smaller chunk groups or not... but im not sure as a user i would really know what a chunk group is and even if i did i wouldn't really know what the size is

It's an advanced config option. I don't expect the normal user to touch that.

Yeah propose a better name. I think it's better than the previous name `moduleFactorCost`.

The config is experimental, so we can ITG that.

lukesandberg (approved):

lukesandberg:
yeah fair

i don't know what a better name would be 😂

PR #95093, opened 2026-06-23
fix(codemod): don't double-add '.' after boolean ESLint --fix flag (#84616)

### LINKED ISSUE #84616: Codemod adds extra dot path

https://github.com/boiseiben/reproduction-app

### To Reproduce

Starting from scratch (alternatively, just run the last command after cloning the repo linked above):
```bash

pnpm create next-app@canary --example reproduction-template reproduction-app
cd reproduction-app
pnpx @next/codemod@canary next-lint-to-eslint-cli .
```

The diff on package.json includes in part:
```
-    "lint": "next lint",
-    "lint:fix": "next lint --fix .",
+    "lint": "eslint .",
+    "lint:fix": "eslint --fix . .",
```

**That last line seems to have an extra ` .` at the end.**

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Mon Apr 21 17:08:54 UTC 2025
  Available memory (MB): 31770
  Available CPU cores: 12
Binaries:
  Node: 24.2.0
  npm: 11.3.0
  Yarn: N/A
  pnpm: 10.15.0
Relevant Packages:
  next: 15.6.0-canary.51 // Latest available version is detected (15.6.0-canary.51).
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: N/A
Next.js Config:
  output: N/A
```

Linting

next dev (local)

### Additional context

I am pretty sure the issue came from [this commit](https://github.com/vercel/next.js/commit/c57fe2561cda36f8a5abb41a94ea1a3f0a094226) and is on or near [this line](https://github.com/vercel/next.js/blob/0f781a6a41a7823f7236ac34aedbabddda8a7cfd/packages/next-codemod/transforms/next-lint-to-eslint-cli.ts#L673), which should have another conditional around it based on checking for what paths might already exist in the command.

Dodothereal:
Closes #84616

## Description

The `next-lint-to-eslint-cli` codemod transformed scripts like `"next lint --fix ."` into `"eslint --fix . ."`. The parser was treating the trailing path token as the value of the `--fix` flag, causing it to insert that path *and* the default `.` placeholder.

ESLint's `--fix` and `--fix-dry-run` are boolean flags (no value), so the next token should fall through to the positional directory parser instead. Adding an explicit branch in the token loop to push them through unchanged removes the double `.` artifact.

## Test plan

- Existing fixtures continue to assert `"next lint --fix && npm test"` → `"eslint --fix . && npm test"` (verified by reading the test and tracing manually).
- The existing tests are unchanged. A fixture case for `"next lint --fix ."` would be a nice follow-up but is out of scope of this single-fix PR.

## AI assistance

Used an AI coding assistant to draft the diff; verified by hand against the existing token-loop branches (`--strict` case for boolean flag handling).

### DIFF
--- a/packages/next-codemod/transforms/next-lint-to-eslint-cli.ts
+++ b/packages/next-codemod/transforms/next-lint-to-eslint-cli.ts
@@ -907,6 +907,9 @@ function updatePackageJsonScripts(packageJsonContent: string): {
 
               if (token === '--strict') {
                 eslintArgs.push('--max-warnings', '0')
+              } else if (token === '--fix' || token === '--fix-dry-run') {
+                // Boolean flags: do not consume the next token as a value.
+                eslintArgs.push(token)
               } else if (token === '--dir' && i + 1 < argTokens.length) {
                 paths.push(argTokens[++i])
               } else if (token === '--file' && i + 1 < argTokens.length) {


Dodothereal:
Hi, just a gentle ping on this in case it slipped through - no rush at all. Happy to rebase, split, or adjust anything if that helps. Thanks for maintaining this project.

PR #95100, opened 2026-06-23
Keep the resolved cache life for `cacheMaxMemorySize: 0` caches in dev

unstubbable:
To make `cacheMaxMemorySize: 0` fast in development, the `'use cache'` wrapper forced every built-in cache entry to a dynamic cache life (`revalidate: 0`, a five-minute `expire`) so that warm reads would serve the stale value and regenerate a fresh one in the background. That forcing leaked into the stored entry, and an entry with `revalidate: 0` is treated as too dynamic to live in a static shell. On a warm reload of a route that is fully cached and has no `<Suspense>` above the `'use cache'`, the dev validation prerender read the entry and threw an error claiming that a `'use cache'` with zero `revalidate` was nested inside another `'use cache'` that has no explicit `cacheLife`, even though nothing was actually nested, and reported the route as a blocking route. The cold load was unaffected because the entry did not exist yet.

With this change the size-0 case no longer forces a dynamic cache life and keeps the entry's resolved life instead: the explicit `cacheLife()` values when the user set them, otherwise the default (a 15-minute `revalidate` and an infinite `expire`). That default sits well above the dynamic thresholds, so a fully cached route can be prerendered again and the false-positive blocking error is gone. An explicitly short `cacheLife()` still makes the route legitimately dynamic without a spurious throw, because the explicit-revalidate and explicit-expire flags are then set. Private caches are unchanged and remain forced-dynamic.

So that reloads keep showing a fresh value now that the stored life is no longer always stale, the background revalidation in the cache-hit path also fires for size-0 built-in entries regardless of the stored `revalidate`, restricted to the dynamic dev request render so the dev validation prerender does not also regenerate.

### DIFF
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -1083,14 +1083,15 @@ async function collectResult(
   // DYNAMIC_EXPIRE`) for private caches, which have no real backing handler.
   // The zero revalidate makes every read serve stale-while-revalidate
   // (regenerating a fresh entry in the background), and `DYNAMIC_EXPIRE` (5
-  // minutes, the shortest expire not treated as dynamically shortened) bounds
-  // how long an entry lingers in the dedicated in-memory private handler. The
-  // size-0 case (`cacheMaxMemorySize: 0`) deliberately does NOT force this: it
-  // keeps its resolved cache life so that the cache entry can be considered
-  // prerenderable instead of being misread as a dynamic hole, and a separate
-  // dev revalidation (see the cache-hit path below) keeps its reloads showing a
-  // fresh value. Custom kinds keep their real cache life too, since their
-  // backing handler owns it.
+  // minutes) caps how long an entry lingers in the dedicated in-memory private
+  // handler. It is the shortest `expire` that isn't treated as dynamic; a
+  // smaller `expire` would exclude the entry from prerenders. The size-0 case
+  // (`cacheMaxMemorySize: 0`) deliberately does NOT force this: it keeps its
+  // resolved cache life so that the cache entry can be considered prerenderable
+  // instead of being misread as a dynamic hole, and a separate dev revalidation
+  // (see the cache-hit path below) keeps its reloads showing a fresh value.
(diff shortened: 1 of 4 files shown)


packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -3112,10 +3109,40 @@ export async function cache(
...
 
-          if (currentTime > entry.timestamp + entry.revalidate * 1000) {
-            // If this is stale, and we're not in a prerender (i.e. this is
-            // dynamic render), then we should warm up the cache with a fresh
-            // revalidated entry.
+          // Trigger a background revalidation when the entry is stale (past its
+          // `revalidate`), so the next read gets a fresh value without blocking
+          // this one. In development with the in-memory cache disabled
+          // (`cacheMaxMemorySize: 0`), built-in entries keep their resolved
+          // (potentially non-dynamic) cache life, so a warm hit from the dev
```
lubieowoce: "warm hit"?

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -1080,21 +1080,18 @@ async function collectResult(
...
-  // any built-in (non-custom) kind when the in-memory cache is disabled
-  // (`cacheMaxMemorySize: 0`). The zero revalidate makes every read serve
-  // stale-while-revalidate (re-warming a fresh entry in the background) so warm
-  // reloads stay fast, and `DYNAMIC_EXPIRE` (5 minutes, the shortest expire not
-  // treated as dynamically shortened) bounds how long an entry lingers in the
-  // dev in-memory cache. Custom kinds keep their real cache life, since their
+  // DYNAMIC_EXPIRE`) for private caches, which have no real backing handler.
+  // The zero revalidate makes every read serve stale-while-revalidate
+  // (regenerating a fresh entry in the background), and `DYNAMIC_EXPIRE` (5
+  // minutes, the shortest expire not treated as dynamically shortened) bounds
```
lubieowoce: what does "dynamically shortened" mean?

lubieowoce (approved):

PR #95101, opened 2026-06-23
test: adapter-root regression — verify repoRoot is not outputFileTracingRoot

gaojude:
### What?

Adds `test/production/adapter-root` — a regression test for the adapter `repoRoot` calculation introduced by commit `beee190d8b` (Next.js 16.3.0-canary.57).

### Why?

In canary.57, `build/adapter/build-complete.ts` was fixed so the adapter uses the true repository root (`NextConfigComplete.repoRoot`) instead of `outputFileTracingRoot` when computing relative asset paths. Before the fix, users who set `outputFileTracingRoot` to a custom value (e.g. a monorepo root) would get incorrect relative paths for route assets in the adapter output.

This regression caused a production outage in v0: the Vercel Workflows SDK's generated route files (`/.well-known/workflow/v1/flow`, etc.) were deployed at wrong relative paths, so the handlers were not found and all POSTs returned 405 — every workflow run stuck in `pending`.

### How?

The test places the Next.js project in a `sub/` subdirectory of a temp workspace (simulating a monorepo sub-package). It then verifies:

1. `repoRoot` equals the parent directory (not the project directory)
2. All adapter output asset paths are prefixed with `sub/` (relative to repo root, not project root)

Both the lockfile-detection path and the `NEXT_PRIVATE_OUTPUT_TRACE_ROOT` environment variable path are tested.

### Verification

- `pnpm test-start-turbo test/production/adapter-root/adapter-root.test.ts` — **FAILS** on pre-canary.57 dist (repoRoot set to outputFileTracingRoot); **PASSES** after the `beee190d8b` fix is compiled in

### DIFF
--- a/crates/next-core/src/next_import_map.rs
+++ b/crates/next-core/src/next_import_map.rs
@@ -1391,11 +1391,10 @@ fn insert_package_alias(import_map: &mut ImportMap, prefix: &str, package_root: 
 
 /// Handles instrumentation-client.ts bundling logic.
 ///
-/// Resolves the `private-next-instrumentation-client` alias to a virtual module
-/// that first requires each entry of `instrumentationClientInject` for side
-/// effects (in array order) and then re-exports the user's
-/// `instrumentation-client.{pageExt}` file via the
-/// `private-next-instrumentation-client-user` alias.
+/// Without injected modules, resolves `private-next-instrumentation-client`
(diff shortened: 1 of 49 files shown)


gaojude:
Closing to rebase cleanly on canary — branch was cut from wrong base

PR #95102, opened 2026-06-23
Consider merging chunks when `overlap == 1`

sampoder:
See https://github.com/vercel/next.js/pull/95026/changes#r3462912703; as discussed in office hours


turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -250,7 +250,7 @@ pub async fn make_production_chunks(
                             iterations += 1;
                             let overlap = overlap(&candidate.chunk_groups, &other.chunk_groups);
                             // It need to have at least two chunk groups in common
```
```suggestion
                            // It need to have at least one chunk group in common
```

sokra (approved):

PR #95114, opened 2026-06-24
[ci] Make automatic native binding install opt-in

eps1lon:
A `pnpm install` installs native bindings for Next.js by default. This makes sense for local development where half of the team doesn't work on native code and just needs the most recent bindings. However, CI almost exclusively wants to test with bindings that are built from the current commit. The rest doesn't need native bindings at all.

The only exception is `test_examples` which is likely not functional. Though we still support it by allowing to opt-into automatically installing native bindings in CI.

## test plan

- [x] fetch test timings still skips: https://github.com/vercel/next.js/actions/runs/28087707205/job/83157705446?pr=95114#step:7:228

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -136,9 +136,6 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NEXT_TELEMETRY_DISABLED: 1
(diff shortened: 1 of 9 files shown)


bgw (approved):
Thanks, this feels like it eliminates a possible footgun in CI

PR #95117, opened 2026-06-24
fix(eslint-plugin): replace destructured re-export with direct property access for CJS named export compatibility

### LINKED ISSUE #86504: @next/eslint-plugin-next incorrect named exports

https://github.com/thernstig/eslint-mcp-jiti

### To Reproduce

1. Install the packages
2. Execute this in a terminal: `node -e "import {configs} from '@next/eslint-plugin-next';"`

### Expected

It should return with no errors.

### Actual

```sh
> node -e "import {configs} from '@next/eslint-plugin-next';"
file:///home/tobias/code/repro-eslint-mcp-jiti/[eval1]:1
import {configs} from '@next/eslint-plugin-next';
        ^^^^^^^
SyntaxError: Named export 'configs' not found. The requested module '@next/eslint-plugin-next' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@next/eslint-plugin-next';
const {configs} = pkg;

    at #_instantiate (node:internal/modules/esm/module_job:254:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:369:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:278:26)
    at async ModuleLoader.executeModuleJob (node:internal/modules/esm/loader:275:20)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v24.11.0
```

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP Tue Nov 5 00:21:55 UTC 2024
  Available memory (MB): 31662
  Available CPU cores: 32
Binaries:
  Node: 24.11.0
  npm: 11.6.1
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.0.4 // Latest available version is detected (16.0.4).
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: N/A
Next.js Config:
  output: N/A
```

Linting

Other (Deployed)

### Additional context

I do not know what the fault is, but I compared how `@next/eslint-plugin-next` sets its named exports compared to `typescript-eslint`, which also sets named exports (both being in Common JS (CJS)). An AI summarized it as below. Again, might be wrong, but adding for context:

> @next/eslint-plugin-next defines top-level getters on exports
> `Object.defineProperty(exports, "configs", { enumerable: true, get: () => configs });`
> Node ignores top-level getters for synthetic named exports when importing CJS as ESM.
>
> typescript-eslint works because it assigns a data property at the top level:
> `exports.configs = createConfigsGetters({ ... }); // data property`
> Node sees a top-level data property → creates a synthetic named export → import { configs } works.
>
> To fix @next/eslint-plugin-next do this:
> 
> ```js
> // Instead of top-level getter:
> Object.defineProperty(exports, "configs", {
>  enumerable: true,
>  value: configs  // <- data property, not a getter
> });
> ```

deepshekhardas:
Fixes #86504

The `@next/eslint-plugin-next` package uses destructured re-exports:
```
export const { rules, configs } = plugin
```

When TypeScript compiles this to CJS, it generates `Object.defineProperty` with a **getter** for each named export. Node.js's ESM-CJS interop only creates synthetic named exports for **data properties** (assigned with `=`), not for accessor properties (getters). This causes:
```
SyntaxError: Named export 'configs' not found
```

Replaced with direct property access:
```
export const rules = plugin.rules
export const configs = plugin.configs
```

This compiles to `exports.rules = plugin.rules` / `exports.configs = plugin.configs` (data properties), which Node.js correctly recognizes as synthetic named exports.

### DIFF
--- a/packages/eslint-plugin-next/src/index.ts
+++ b/packages/eslint-plugin-next/src/index.ts
@@ -124,4 +124,5 @@ Object.assign(plugin.configs, {
 } satisfies ESLintPluginConfigs)
 
 export default plugin
-export const { rules, configs } = plugin
+export const rules = plugin.rules
+export const configs = plugin.configs


deepshekhardas:
block status test

PR #95121, opened 2026-06-24
Fix local fonts in statically prerendered `ImageResponse` metadata route

unstubbable:
Follow-up to #94957, which made cached `ImageResponse` routes statically prerenderable by serializing the constructor arguments through React Flight (to build a cache key and to run the element's async Server Components once). A custom font passed as a `Buffer` in the options was corrupted by that round-trip: Flight applies the `toJSON` method that Node's `Buffer` carries, so the font reached satori as a `{ type: 'Buffer', data: [...] }` plain object rather than binary, and satori's font parser threw `TypeError: First argument to DataView constructor must be an ArrayBuffer`. The crash happened even without a user `'use cache'` directive, because the serialization runs unconditionally during the prerender.

We now serialize only the element through Flight, which is the only part that needs async resolution and dynamic-input detection, and pair the resolved element with the original in-memory options when handing the tree to satori. The font `Buffer` is therefore never serialized and reaches satori intact. The cache key no longer comes from the serialized arguments either; it is derived from a SHA-256 hash of the serialized element combined with a content hash of the options, with binary font data hashed by its bytes and everything else by a sorted, type-tagged walk. Besides fixing the crash, this keeps font bytes out of the cache key and removes the "Uint8Array objects are not supported" warning that the dev React runtime emitted during `next build --debug-prerender`.

The custom-fonts documentation example now loads the font once at module scope. This keeps the route statically generated, since reading inside the component would be uncached I/O that Cache Components treats as dynamic, and it steers users away from wrapping the read in `'use cache'`, which would route the font `Buffer` through use-cache's own Flight serialization and reintroduce the same corruption upstream of `ImageResponse`, where this fix cannot reach it. A regression test loads a local font via a module-scope `readFile` and asserts the route prerenders statically.


test/e2e/app-dir/use-cache-metadata-route-handler/use-cache-metadata-route-handler.test.ts
```diff
@@ -17,6 +17,20 @@ describe('use-cache-metadata-route-handler', () => {
...
+    const res = await next.fetch('/custom-font/opengraph-image')
+    expect(res.status).toBe(200)
+    expect(res.headers.get('content-type')).toBe('image/png')
+
+    if (isNextStart) {
+      const [buildStatus] = next.cliOutput.match(
+        /. \/custom-font\/opengraph-image/
+      )
+
+      expect(buildStatus).toBe('○ /custom-font/opengraph-image')
```
gaearon, line 30: Does this test for static one specifically? If not, should there be a test that verifies static build when possible?

unstubbable: It does, indicated by the `○` (as opposed to `ƒ`).

packages/next/src/server/og/cache-image-response.ts
```diff
@@ -255,6 +278,82 @@ async function getCachedImageResponseArrayBuffer(
...
+ * plain objects/arrays, and binary font data. Exotic objects such as `Date` or
+ * `Map` aren't special-cased (they'd hash via their enumerable own keys), since
+ * options never contain them.
+ *
+ * The encoding is self-delimiting: every node starts with a type tag, and
+ * variable-length parts (byte runs, primitives, keys) are length-prefixed,
+ * while arrays and objects are count-prefixed. This makes it injective, so no
+ * concatenation of values can be mistaken for a differently shaped input.
+ */
+function updateHashWithOptions(hash: Hash, value: unknown): void {
```
lubieowoce, line 298: from defensive programming PoV: i don't love this, it doesn't handle Maps/Sets and whatnot (we should recurse into those like we do for objects/arrays), so we're kind of banking on Satori not having any options like that. maybe we can add some kind of check for non-plain objects and at least warn that we may be serializing them wrong? then at least there's a chance we get a bug report

lubieowoce (approved):

PR #95122, opened 2026-06-24
skill(cc-adoption): recommend next-dev-loop and add build-only path

aurorascharff:
### What?

Update the `next-cache-components-adoption` Skill so agents actively offer the [`next-dev-loop`](https://github.com/vercel/next.js/tree/canary/skills/next-dev-loop) Skill when verifying Cache Components fixes, and add a build-only verification path for environments that can't run a dev server.

- `## surfacing errors` opens with **"Prefer `next dev`, but build-only works too."** and reorders so `next dev` comes first.
- `## Verifying a fix at runtime` now leads with **"`next-dev-loop` is the recommended way to do this"**, gives the agent a draft user prompt to ask permission to install it, and labels the alternatives as "Fallback" / "Build-only verification" / "No browser and no build either" so the preference order is unambiguous.
- The build-only verification path tells the agent what it can and cannot honestly call "verified" without a browser, so agents in CI / sandboxed environments can finish the work without faking runtime verification.

### Why?

The Skill already pointed at `next-dev-loop` as "the fastest path", but the wording read as one of three equally valid options. Agents that didn't already have it installed silently picked option 2 or 3, so the user didn't see the recommendation. This update keeps the same flow but makes the agent explicitly ask whether to install `next-dev-loop` before milestone B, and softens the assumption that every agent has a dev server available.

### How?

Wording-only changes to [`skills/next-cache-components-adoption/SKILL.md`](skills/next-cache-components-adoption/SKILL.md). No code or test changes.


icyJoseph (dismissed):

docs/01-app/02-guides/migrating-to-cache-components.mdx
```diff
@@ -18,13 +18,28 @@ When [Cache Components](/docs/app/api-reference/config/next-config-js/cacheCompo
...
+## Use the migration skill (recommended)
+
+The [`next-cache-components-adoption`](https://github.com/vercel/next.js/tree/canary/skills/next-cache-components-adoption) skill lets an agent drive the migration with you. It runs in two steps:
+
+1. **Get the build green.** The agent runs a codemod that adds `export const instant = false` to every page and layout. Each opt-out gets a `// TODO: Cache Components adoption` comment as a marker. You can merge this as a standalone PR and tackle step 2 incrementally.
+2. **Remove the opt-outs one at a time.** For each route, the agent walks the subtree top-down, applies the fix the dev overlay points at (cache the data with [`use cache`](/docs/app/api-reference/directives/use-cache), wrap the dynamic part in [`<Suspense>`](https://react.dev/reference/react/Suspense), or leave the route opted out with a comment explaining why), and checks in with you before moving on.
```
icyJoseph:
```suggestion
The [`next-cache-components-adoption`](https://github.com/vercel/next.js/tree/canary/skills/next-cache-components-adoption) skill helps migrate routes to Cache Components incrementally:

1. **Get the build green.** It adds `export const instant = false` to every page and layout, with a `// TODO: Cache Components adoption` marker. You can stop here and open a checkpoint PR, or continue directly to adoption.
2. **Adopt route by route.** It removes each opt-out top-down, applying the needed fix: cache data with [`use cache`](/docs/app/api-reference/directives/use-cache), wrap dynamic UI in [`<Suspense>`](https://react.dev/reference/react/Suspense), or keep the opt-out with an explanation.

The agent checks in before moving to the next route.
```

skills/next-cache-components-adoption/SKILL.md
```diff
@@ -11,148 +11,209 @@ description: >
...
...ar) and as the final verification across the whole app. In Incremental, the build also confirms the pre-step (codemod opted every route out, no shared layout still has a sync-IO blocker) before you ship that PR. Don't reach for the build instead of the dev loop while you're working a route — even when the build surfaces a clean compile error, you still don't know what ended up in the static shell vs streamed; the build's clean error messages are the seductive part. By default the build stops at the first blocking route (it sets `experimental.prerenderEarlyExit`), so it's also poor for sizing the work. Two flags help when iterating: `--debug-build-paths` builds only the routes you name (comma-separated glob patterns of file paths relative to the project root, e.g. `--debug-build-paths="app/admin/**/page.tsx"` — not URL paths; `--debug-build-paths="app/(marketing)/about/page.tsx"` — not `/about`; `--debug-build-paths="app/admin"` matches nothing and silently builds zero routes), and `--debug-prerender` disables the early exit so the build continues past the first prerender failure, reports every blocking route, and prints a fuller stack trace that names the originating file and line.
```
icyJoseph: I think the agent Skill prob doesn't need to mention `experimental.prerenderEarlyExit`, otherwise this part is good - we shall see more usage of --debug-build-paths

icyJoseph (approved):

PR #95132, opened 2026-06-24
defer and batch leaf distance updates

lukesandberg:
### What?

Defer and batch the "leaf distance" updates in `turbo-tasks-backend` so that a task's leaf distance is computed once, when the task completes, instead of eagerly on every dependency read.

### Why?

Leaf distance is a per-task heuristic used only for **invalidation/re-execution scheduling priority** (`compute_stale_priority` and `schedule_task`'s `TaskPriority::invalidation`). On `canary` it was recomputed and propagated on *every* output-dependency read: each read built a one-shot `LeafDistanceUpdateQueue`, pushed the reader, and executed it immediately. A task reading N dependencies therefore did N separate queue executions (with N transitive propagation passes) during its execution.

This batches that work: a task reads many dependencies, but we only need its final leaf distance once it finishes executing.

### How?

- The per-read `push` calls stay exactly where they were, but instead of executing a throwaway local queue, they push into a single `LeafDistanceUpdateQueue` stored in the reader's `InProgressState` (`InProgressStateInner.leaf_distance_update_queue`).
- The queue is executed once, at task completion (`task_execution_completed_prepare`), which updates the task's own leaf distance and propagates the change to its output dependents — same logic as before, just deferred and de-duplicated. It runs before the stale check so `compute_stale_priority` still sees the freshly computed value.
- The leaf-distance computation/propagation code in `leaf_distance_update.rs` is unchanged (the queue type just gains `Debug` so it can live in the in-progress state, and its now-unused `new()` is dropped).

Behavior is otherwise identical to `canary`:

- A task that reads no dependencies never accumulates anything, so its leaf distance stays at its default (no spurious bump).
- Leaf distance is only accumulated when dependency tracking is enabled (it is meaningless when tasks never re-execute); a `debug_assert!` at completion documents and enforces this.
- Under `verify_immutable`, immutable producers are still excluded from leaf-distance accumulation (checked explicitly), so `verify_immutable` builds compute the same leaf distance as release builds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10064,6 +10064,7 @@ dependencies = [
  "codspeed-criterion-compat",
  "crossbeam-utils",
  "dashmap 6.1.0",
+ "either",
  "futures",
  "hashbrown 0.14.5",
  "indexmap 2.13.0",
--- a/turbopack/crates/turbo-tasks-backend/Cargo.toml
+++ b/turbopack/crates/turbo-tasks-backend/Cargo.toml
@@ -32,7 +32,7 @@ trace_task_completion = []
 trace_task_modification = []
 trace_task_dirty = ["trace_aggregation_update_queue", "task_dirty_cause"]
 trace_task_output_dependencies = []
-trace_task_details = []
+trace_task_details = ["dep:either"]
 trace_prepare_tasks = []
 
 [dependencies]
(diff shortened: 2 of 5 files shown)


lukesandberg:
* **#95132** <a href="https://app.graphite.com/github/pr/vercel/next.js/95132?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95132?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

PR #95135, opened 2026-06-24
[next-dev-loop] Require --session on every agent-browser command

gaojude:
The skill set `--session <name>` only on `open`/`close` and framed it as a state-persistence flag, so agents inferred it was an open/close-time concern and dropped it on later commands. A command without `--session` runs against a different session — the unnamed default, a separate browser that never navigated — so its reads come back empty and a working page looks blank, easily mistaken for a broken app.

Preflight now makes `--session <name>` an explicit every-command invariant (each invocation is a separate process; the flag selects which browser it talks to), and teardown back-references that rule. The blank-read gotcha is reduced to the skill's own cross-check principle — when the browser view and `/_next/mcp` disagree, suspect the tooling first — with the agent-browser session/tab specifics deferred to its own manual, which the skill now points at via `agent-browser skills get core --full` instead of `--help`.

### DIFF
--- a/skills/next-dev-loop/SKILL.md
+++ b/skills/next-dev-loop/SKILL.md
@@ -22,7 +22,11 @@ You verify through two views of the same running app:
   `tools/list` for the current surface.
 - **`agent-browser`** — a CLI that drives a real Chrome. Knows
   framework-agnostic browser things: DOM, console, network, React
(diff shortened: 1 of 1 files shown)


aurorascharff (approved):

gaearon:
iterated on top of this: https://github.com/vercel/next.js/pull/95153

PR #95137, opened 2026-06-24
[turbopack] Don't persist if there is little work to do

lukesandberg:
Currently we skip a persist cycle when there are no dirty tasks. This PR extends that to skip persist cycles when little compilation work has happened since the last snapshot, avoiding small writes and expensive compaction for little benefit.

Compilation time is a better proxy for "how much work a snapshot would save" than a raw dirty-task count. The background snapshot loop accumulates active (non-idle) wall-clock time and skips a cycle when less than a threshold has elapsed since the last persisted snapshot. Default is 1s, overridable via `TURBO_ENGINE_SNAPSHOT_MIN_ACTIVE_TIME_MILLIS` (set to `0` in tests). Shutdown and test snapshots bypass the gate.

The active-time bookkeeping is encapsulated in a small `Stopwatch` helper.

### DIFF
--- a/test/e2e/filesystem-cache/evict-after-snapshot.test.ts
+++ b/test/e2e/filesystem-cache/evict-after-snapshot.test.ts
@@ -8,6 +8,9 @@ import { retry, waitFor } from 'next-test-utils'
     'ENABLE_CACHING=1',
     'TURBO_ENGINE_IGNORE_DIRTY=1',
     'TURBO_ENGINE_SNAPSHOT_IDLE_TIMEOUT_MILLIS=1000',
+    // Persist even tiny snapshots so the test doesn't depend on the
+    // minimum-compilation-time threshold.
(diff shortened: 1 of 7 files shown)


lukesandberg:
* **#95213** <a href="https://app.graphite.com/github/pr/vercel/next.js/95213?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95137** <a href="https://app.graphite.com/github/pr/vercel/next.js/95137?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95137?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

lukesandberg:
discussed in office hours

the better approach would be a time based metric, only persist if we have accumulated N seconds of compilation time.   We can track this via the aggregated_update metric stored in the backend.


Edit: Done!

wbinnssmith (approved):

PR #95144, opened 2026-06-24
Improve the NFT error message and ignore comment handling

### LINKED ISSUE #95125: Turbopack: `turbopackIgnore` doesn't silence the dynamic-fs whole-project trace for `fs(path.join(...))` calls

https://github.com/nick-haiar-home/turbopack-nft-dynamic-fs-repro

### To Reproduce

1. `pnpm install` (installs `next@canary`).
2. `pnpm build`.
3. Observe the warning pointing at `lib/loader.js` (the `readdirSync(folder)` call) - see below.
4. Apply the **exact** remedy the warning recommends to the folder path: `path.join(/* turbopackIgnore: true */ process.cwd(), "data")`.
Rebuild > **the warning is unchanged.**
5. Move the comment to the `fs` call's bare variable argument - `readdirSync(/* turbopackIgnore: true */ folder)` and rebuild. **That one is silenced**, which proves the feature works; but the sibling `readFileSync(path.join(folder, f), …)` on the next line still warns and cannot be silenced by the comment in either position (before `path.join`, or inside it as the message shows).

**Current** (`next@16.3.0-canary.65`, Turbopack, `output: "standalone"`):

```
Turbopack build encountered 3 warnings:
./lib/loader.js:6:10
Warning: Dynamic filesystem access causes tracing of the whole project
> 6 |   return readdirSync(folder).map((f) =>
    |          ^^^^^^^^^^^^^^^^^^^
...
To resolve this, you can
- make sure they are statically scoped to some subfolder: path.join(process.cwd(), 'data', bar), or
- only use them in development, or
- add ignore comments: path.join(/*turbopackIgnore: true*/ process.cwd(), bar), or
- remove them.
```

The message's headline suggestion is `path.join(/*turbopackIgnore: true*/ …)`,
but the flagged operation is the **`fs` call** (`readdirSync` / `readFileSync`),
not `path.join`. Annotating `path.join` per the message has **no effect**:

| Annotation (per the recommended form) | Result |
|---|---|
| `path.join(/* turbopackIgnore: true */ process.cwd(), "data")` (folder) | warning unchanged |
| `readFileSync(/* turbopackIgnore: true */ path.join(folder, f), …)` | still warns at that line |
| `readFileSync(path.join(/* turbopackIgnore: true */ folder, f), …)` | still warns at that line |
| `readdirSync(/* turbopackIgnore: true */ folder)` (bare variable arg) | **silenced** ✅ |

So `turbopackIgnore` *is* honored, but only when the comment annotates a bare
variable passed directly to the `fs` function — **not** the
`path.join(/*turbopackIgnore: true*/ …)` form the warning explicitly tells you
to use, and not a `path.join(...)` expression nested inside the `fs` call. There
is no documented placement that silences a `readFileSync(path.join(folder, f))`.

**Expected:** the `turbopackIgnore` placement shown in the warning message
should actually suppress the trace for the flagged call — or the message should
point at the expression that can be annotated (the `fs` call's argument), so the
recommended remedy works as written.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:40:51 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T8112
  Available memory (MB): 24576
  Available CPU cores: 8
Binaries:
  Node: 24.13.1
  npm: 11.8.0
  Yarn: 1.22.22
  pnpm: 10.20.0
Relevant Packages:
  next: 16.3.0-canary.65 // Latest available version is detected (16.3.0-canary.65).
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: N/A
Next.js Config:
  output: standalone
```

Output, Turbopack

next build (local)

### Additional context

 **Prior art:** [#92639](https://github.com/vercel/next.js/issues/92639)
  reported the same family of warning (then worded "Encountered unexpected file
  in NFT list", flagging `next.config.ts`) and independently found
  `turbopackIgnore: true` had no effect and `--webpack` made it disappear. It
  was auto-closed for an invalid reproduction link, never triaged on the merits.
  Note that **canary has already improved the diagnostics** vs. that report: the
  message now names the dynamic-fs cause and points at the exact source line
  instead of vaguely flagging `next.config`. This issue is the narrower,
  still-present remainder: the *recommended escape hatch placement doesn't work*.
- **Why it matters / motivating case:** the canonical trigger is a library that
  reads files from a runtime-resolved folder at boot — e.g. `drizzle-orm`'s
  `migrate()` reading its migrations folder, which is mandatory when the DB
  lives on a runtime-mounted volume (Fly.io etc.) that doesn't exist at build
  time. There, the `fs` read is inside a `node_modules` dependency you can't
  annotate at all, so the warning has no available remedy and the whole project
  gets traced into the standalone output.
- **Affected versions:** reproduces on `next@16.2.6` (warning then worded
  "Encountered unexpected file in NFT list", vaguely flagging `next.config`)
  through `next@16.3.0-canary.65` (reworded to "Dynamic filesystem access…",
  now line-accurate). The diagnostics improved between them, but the underlying
  behavior — no working `turbopackIgnore` placement for `fs(path.join(...))` —
  did not. Long-standing, not a recent regression.
- **`--webpack` is not a viable workaround:** building with the webpack builder
  suppresses the warning (confirmed in #92639), but that isn't a long-term
  option given Turbopack is the default builder — this needs to work under
  Turbopack.
- **Not purely cosmetic:** per the warning's own text, the whole project
  (including `public/`) is included in the server output, which "can slow down
  deployments or lead to failures when size limits are exceeded" — a real risk
  on size-constrained / serverless targets, not just build noise.

lukesandberg:
Fixes #95125 in two ways

The OP admits that they misread the message in a few ways, but we can fix all of them

1. Be specific about what function might need an annotation in the error message.  A user with a nearby `path.join` might think that that is where the annotation belongs instead of on the function with the warning.

2. Be more flexible with how these `ignore` comments are interpreted `fs.readFileSync(path.join(process.cwd(), someVar))` shouldn't require 2 annotations.   Handle this by checking for and detecting `turbopackIgnore` comments in all expressions!  This will then bubble up through the linker and just cause us to ignore corresponding fs access apis.

### DIFF
--- a/crates/next-api/src/nft.rs
+++ b/crates/next-api/src/nft.rs
@@ -7,7 +7,7 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use tracing::Instrument;
 use turbo_rcstr::{RcStr, rcstr};
 use turbo_tasks::{
-    FxIndexMap, FxIndexSet, ReadRef, ResolvedVc, TryFlatJoinIterExt, TryJoinIterExt, Vc,
+    FxIndexMap, FxIndexSet, ReadRef, ResolvedVc, TraitRef, TryFlatJoinIterExt, TryJoinIterExt, Vc,
 };
 use turbo_tasks_fs::{
     DirectoryEntry, FileSystemPath,
@@ -23,6 +23,7 @@ use turbopack_core::{
     module::{Module, Modules},
     module_graph::{GraphTraversalAction, ModuleGraph},
     raw_module::RawModule,
+    reference::DynamicTraceReference,
 };
 
 use crate::project::Project;
@@ -182,7 +183,7 @@ async fn get_glob_includes(
     let glob_result = project_root_path.read_glob(glob).await?;
 
     // Walk the full glob_result using an explicit stack to avoid async recursion overheads.
-    // Use a BTreeSet to get determinstic order (return value of `read_glob` has random order).
+    // Use a BTreeSet to get deterministic order (return value of `read_glob` has random order).
     let mut result = vec![];
     let mut stack = VecDeque::new();
     stack.push_back(glob_result);
@@ -334,13 +335,14 @@ pub async fn traced_modules_for_entries(
     )?;
 
     for (parent, reference) in forbidden_issues {
-        ForbiddenTracedFileIssue::new(
-            parent.ident().await?.path.clone(),
-            reference.into_trait_ref().await?.source(),
-        )
-        .to_resolved()
-        .await?
-        .emit();
+        let reference = reference.into_trait_ref().await?;
+        let source = reference.source();
+        let origin_fn_name = TraitRef::try_downcast::<Box<dyn DynamicTraceReference>>(reference)
+            .map(|traced| traced.origin_fn_name());
+        ForbiddenTracedFileIssue::new(parent.ident().await?.path.clone(), source, origin_fn_name)
+            .to_resolved()
+            .await?
+            .emit();
     }
 
     Ok(Vc::cell(traced_modules.into_iter().collect()))
@@ -439,6 +441,9 @@ pub async fn traced_module_data_for_graph(
 struct ForbiddenTracedFileIssue {
     parent: FileSystemPath,
     issue_source: Option<IssueSource>,
+    /// The dynamic function whose access triggered the trace (e.g.
+    /// `fs.readFileSync`), used to name the offending call in the message.
+    origin_fn_name: Option<RcStr>,
 }
 
 #[turbo_tasks::value_impl]
@@ -447,10 +452,12 @@ impl ForbiddenTracedFileIssue {
     pub async fn new(
         parent: FileSystemPath,
         issue_source: Option<IssueSource>,
+        origin_fn_name: Option<RcStr>,
     ) -> Result<Vc<Self>> {
         Ok(Self {
             parent,
             issue_source,
+            origin_fn_name,
         }
         .cell())
     }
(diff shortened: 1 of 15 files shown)


lukesandberg:
* **#95144** <a href="https://app.graphite.com/github/pr/vercel/next.js/95144?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95144?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbopack-core/src/reference/mod.rs
```diff
@@ -44,6 +44,14 @@ pub trait ModuleReference: ValueToString {
...
         None
     }
+
+    /// The name of the function/call that created this reference (e.g.
+    /// `fs.readFileSync`), if any. Used to name the offending call in tracing
+    /// diagnostics so the suggested fix refers to the actual call rather than an
+    /// example.
+    fn origin_fn_name(&self) -> Option<RcStr> {
+        None
+    }
```
mischnic: I don't think that should exist for ModuleReference

crates/next-api/src/nft.rs
```diff
@@ -499,19 +527,13 @@ impl Issue for ForbiddenTracedFileIssue {
...
                 )),
                 StyledString::Code(rcstr!("path.join(process.cwd(), 'data', bar)")),
                 StyledString::Text(rcstr!(", or")),
             ]),
             StyledString::Text(rcstr!("- only use them in development, or")),
-            StyledString::Line(vec![
-                StyledString::Text(rcstr!("- add ignore comments: ")),
-                StyledString::Code(rcstr!(
-                    "path.join(/*turbopackIgnore: true*/ process.cwd(), bar)"
-                )),
```
mischnic, line 512: If the hardcoded path.join was misleading, why don't we just say `fsOperation(/*turbopackIgnore: true*/ process.cwd(), bar)` instead.

lukesandberg: yeah we could do that.  propogating the actual fn name seems not too bad though

turbopack/crates/turbopack-core/src/reference/mod.rs
lukesandberg: move it down to a sub-trait?  DynamitTraceReference?

lukesandberg: done

crates/next-api/src/nft.rs
lukesandberg: that is a better fallback

turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/mod.rs
```diff
@@ -116,6 +116,12 @@ pub enum JsValue<'a> {
         original_value: Option<Arc<JsValue<'a>>>,
         reason: RcStr,
         has_side_effects: bool,
+        /// True when this value was intentionally downgraded to unknown because
+        /// the user opted out of analysis with a `turbopackIgnore` comment (as
+        /// opposed to being unknown because it could not be statically
+        /// determined). Consumers such as the dynamic-fs/path tracing handlers
+        /// use this to skip adding references (and warnings) entirely.
+        ignored: bool,
```
lukesandberg: explicitly_ignored is a better name

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -2324,6 +2337,11 @@ where
         WellKnownFunctionKind::FsReadMethod(name) if analysis.analyze_mode.is_tracing_assets() => {
             let args = linked_args().await?;
             if !args.is_empty() {
+                // The path argument opted out of analysis via a `turbopackIgnore`
+                // comment. Skip tracing entirely (no reference, no warning).
+                if args[0].is_ignored() {
```
lukesandberg: do we still need these early returns

mischnic (approved):

PR #95106, opened 2026-06-24
refactor: split out validateStaticShell

lubieowoce:
This is a pure refactor, split from #94595 to reduce the diff
- Move static shell validation code into a separate function, `validateStaticShell`
- Rename `spawnStaticShellValidationInDev` to `runValidationInDev`. This is outdated: it no longert spawns anything, and also does more than just static shell validation

Also includes a drive-by cleanup in `renderWithWarmCachesForValidationInDev` that was not worth a separate PR : we can drop an unnecessary `accumulatedChunksPromise` wrapper, because it's immediately awaited.

PR #95105, opened 2026-06-24
test: Remove unnecessary dynamic timestamp from instant-validation root layouts

lubieowoce:
Extracted from #94595 . The root layout includes a timestamp that we use to check if the root layout changed during a navigation, but it was unnecessarily using `await connection(); Date.now()` when a `performance.now()`would be enough. Changing it to not introduce a random dynamic hole which could affect validation in some way.  
  
This change revealed that some client component tests were implicitly relying on the dynamic timestamp component as a workaround for NAR-800 . I've added a workaround dynamic component to each of the relevant tests, those can be removed when we fix the bug.

PR #95149, opened 2026-06-25
[PP] Reveal after ShellRuntime when simulating a Shell Prefetch in dev

lubieowoce:
When `partialPrefetching` is on, the shell that we usually care about displaying is the App Shell, which is represented by the ShellRuntime stage. We should only show the cold cache indicator for caches that happen up until then, and release the client-side promise appropriately. This PR extends the existing `revealAfterStage` mechanics to account for this.

> Note that this is incomplete: if we're navigating via `<Link prefetch={true}>` (or have `partialPrefetching: "unstable_eager"), we might want to use the Runtime stage instead. I'm leaving those for a follow up -- representing the common case (a shell prefetch) seems like the most useful thing.

I've re-worked the `revealAfterStage/holdStreamUntilRevealed` setup into an object (`DevNavigationKind`) that represents the navigation that we're trying to simulate -- either an initial load or a client nav. This gets rid of the invalid `holdStreamUntilRevealed = true` + `revealAfterStage = RenderStage.Runtime/ShellRuntime` combination and lets us bring the logic closer to where the stream blocking tricks actually happen.

I've also done some drive-by refactoring of `streamStagedRenderInDev` to dedupe some repetitive code, because every task was basically doing the same thing. I've also moved all `revealAfter.resolve()` calls into separate tasks -- we were inconsistent about this, and the `Static` `revealAfter.resolve()` was done together with the next stage, but the `RenderStage.Runtime` was done in a separate task.

---

Strangely, despite the changes working as expected for the Cold Cache indicator, I can't actually get link data (`await searchParams`) to trigger a fallback when navigating, so there's a failing test for that in `cache-components-dev-streaming.test.ts`. I'm not sure why what's causing this, but I'm leaving that investigation for a follow-up as well.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -1347,6 +1347,12 @@ async function generateDynamicFlightRenderResultWithStagesInDev(
       setCacheStatus('ready', htmlRequestId)
     }
 
+    const prefetchMode =
+      !!renderOpts.partialPrefetching ||
+      (await anySegmentHasPartialPrefetchingEnabled(loaderTree))
+        ? PrefetchingMode.Partial
+        : PrefetchingMode.LegacySpeculative
+
     // A client navigation into a runtime-prefetch route extends the shell
     // through the runtime-prefetchable content: it has already settled on the
     // client (via the prefetch) by the time it navigates, so it belongs in this
@@ -1354,11 +1360,23 @@ async function generateDynamicFlightRenderResultWithStagesInDev(
     // load: plain navigations, and HMR refreshes (a fresh render of the current
     // page, with no settled prefetch to draw on). Dynamic content always
     // streams in after the shell.
-    const revealAfterStage =
-      initialRequestStore.isHmrRefresh !== true &&
-      (await anySegmentHasRuntimePrefetchEnabled(loaderTree))
-        ? RenderStage.Runtime
-        : RenderStage.Static
+    let prefetchStage: StreamRevealStage
+
+    if (initialRequestStore.isHmrRefresh === true) {
+      prefetchStage = RenderStage.Static
+    } else {
+      if (prefetchMode === PrefetchingMode.Partial) {
+        // TODO(app-shells): model `partialPrefetching: "unstable_eager"`
+        // TODO(app-shells): if this navigation came from <Link prefetch={true} />,
+        // we should show the shell for a speculative prefetch
+        // (which can have more data than the app shell)
+        prefetchStage = RenderStage.ShellRuntime
+      } else {
+        prefetchStage = (await anySegmentHasRuntimePrefetchEnabled(loaderTree))
+          ? RenderStage.Runtime
+          : RenderStage.Static
+      }
+    }
 
     const result = await stagedRenderWithCachesInDev({
       ctx,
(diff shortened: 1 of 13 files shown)


lubieowoce:
* **#94595** <a href="https://app.graphite.com/github/pr/vercel/next.js/94595?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#95089](https://github.com/vercel/next.js/pull/95089) <a href="https://app.graphite.com/github/pr/vercel/next.js/95089?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#95276](https://github.com/vercel/next.js/pull/95276) <a href="https://app.graphite.com/github/pr/vercel/next.js/95276?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#95106** <a href="https://app.graphite.com/github/pr/vercel/next.js/95106?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95151** <a href="https://app.graphite.com/github/pr/vercel/next.js/95151?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95105** <a href="https://app.graphite.com/github/pr/vercel/next.js/95105?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95149** <a href="https://app.graphite.com/github/pr/vercel/next.js/95149?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95149?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4762,68 +4832,22 @@ async function streamStagedRenderInDev({
...
-    () => {
-      // Reveal on a cache miss, or at the static-shell boundary: the static
-      // stage's chunks flushed in the previous task, so the static shell is on
-      // the stream now. `checkForCacheMiss()` is the left operand so its side
-      // effect (tracking the miss / updating the dev overlay) always runs.
-      if (checkForCacheMiss() || revealAfterStage === RenderStage.Static) {
-        revealAfter.resolve()
-      }
-      stageController.advanceStage(RenderStage.ShellEarlyRuntime)
-    },
```
lubieowoce, line 4792: this used to resolve `revealAfter` in the same task as advancing to `ShellEarlyRuntime`, unlike the `Runtime` one which did it in a separate task. I don't see a reason to do this -- if we want `_resolveAfter` to flush consistently, then we should always use a separate task.

test/development/app-dir/cache-indicator-partial-prefetching/app/private/page.tsx
```diff
@@ -1,11 +1,6 @@
 import { Suspense } from 'react'
 import { setTimeout } from 'timers/promises'
 
-// Runtime-prefetchable: a client navigation reveals the runtime shell, and a
-// private cache resolves in the session-data (runtime) stage, so it's part of
-// the runtime shell for a dev client navigation.
-export const prefetch = 'allow-runtime'
-
```
lubieowoce, line 8: not necessary in `partialPrefetching`, private caches are session data so app shells can access them

unstubbable (approved):

PR #95156, opened 2026-06-25
Fix stale default parallel route slots on soft navigation

Tomoya0k:
## Summary

Reset parallel route slots to their new default state during soft navigations instead of preserving the previously active slot.

Add a regression to the parallel breadcrumbs fixture to cover navigating back to a slot's default.tsx fallback.

## Verification

- pnpm.cmd --filter=next types
- NEXT_SKIP_ISOLATE=1 ./scripts/run-jest.sh --mode=dev --bundler=turbo --headless -- test/e2e/app-dir/parallel-routes-breadcrumbs/parallel-routes-breadcrumbs.test.ts


Tomoya0k:
Closing this draft after re-checking the parallel routes semantics and the repro structure. The proposed change made soft navigations render default.js for an unmatched slot, but in this case the slot stays under the same (dashboard) layout and should preserve its active subpage during soft navigation. That means this PR was changing documented behavior rather than fixing a router bug.

PR #90847, opened 2026-03-04
Extract production chunk merge algorithm into testable merge.rs module

lukesandberg:
## Summary

- Extract the pure merge algorithm from `production.rs` into a new `merge.rs` module that operates on simple indices and `RoaringBitmap`s instead of Vc types, making it unit-testable without the turbo-tasks runtime
- `production.rs` now delegates to `merge_grouped_chunks()` and maps results back to Vc types
- Includes unit tests covering near-duplicate modules, layout nesting, deep nesting, input order stability, and the no-merge fast path

## Test plan

- [x] `cargo check -p turbopack-core` passes
- [ ] `cargo test -p turbopack-core` passes (new unit tests in `merge.rs`)
- [ ] Existing production build integration tests pass (behavioral equivalence)

PR #95157, opened 2026-06-25
[turbopack] Add `clusters` to chunking heuristics

sampoder:
The goal of this PR increase the expected benefit of merging two chunks when their overlapping chunk groups are in a "cluster" (a cluster is a set of routes that are commonly visited together). This is because in these cases the Z+Z case (one of six in the N=2 set of cases) is more likely than others. To this I have increased the probability of the Z+Z case.

At the moment, to do this I've done the following to the probabilities:

```rust
/*
MERGED CASE (N = 2):
case X + X (p = (a_rem/groups) * ((a_rem - 1)/rem_g)): size = a_size, requests = 1
case Y + Y (p = (b_rem/groups) * ((b_rem - 1)/rem_g)): size = b_size, requests = 1
case Z + Z (p = (o_groups/groups) * (o_groups - 1)/rem_g): size = (a_size + b_size), requests = 1
case X + Y (p = (a_rem/groups) * (b_rem/rem_g) + (b_rem/groups) * (a_rem/rem_g)): size = a_size + b_size, requests = 2
case X + Z (p = (a_rem/groups) * (o_groups/rem_g) + (o_groups/groups) * (a_rem/rem_g)): size = a_size + (a_size + b_size), requests = 2
case Y + Z (p = (b_rem/groups) * (o_groups/rem_g) + (o_groups/groups) * (b_rem/rem_g)): size = b_size + (a_size + b_size), requests = 2

Request count is different in this case: Z + Z (better)
Requests size is different (worse) in these cases: X + Z, Y + Z

Each cost / benefit is weighted by the probabilities above. There are cases when
we know that Z + Z is more likely due to common user behaviour. This is based on
the "cluster" chunking heuristic we provide. We increase P(Z + Z) when two or more
routes in a cluster request both chunk items together (ie. request Z). P(X + Z) and
P(Y + Z) are therefore less likely. To increase P(Z + Z) while maintaining a total
probability of 1, we do the following (when chunk items overlap in a cluster):

P'(X + Z) = (1/2) * P(X + Z)
P'(Y + Z) = (1/2) * P(Y + Z)
P'(Z + Z) = P(Z + Z) + (1/2) * P(X + Z) + (1/2) * P(Y + Z)

Otherwise:

P'(X + Z) = P(X + Z)
P'(Y + Z) = P(Y + Z)
P'(Z + Z) = P(Z + Z)
*/
```

This increases `P'(Z + Z)` while reducing `P'(X + Z)` and `P'(Y + Z)`. It is a bit of a blunt tool for doing this, however. So if anyone has other suggestions for how to change this probabilities while maintaining existing behaviour - let me know!

### DIFF
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -1146,6 +1146,9 @@ const DEFAULT_WEIGHT_DISTRIBUTION: f32 = 0.1;
 )]
 #[serde(rename_all = "camelCase")]
 pub struct ChunkingHeuristicsConfig {
+    /// Groups of pages commonly visited together, each defined by a list of regular expressions
+    /// matched against the route pathname. The cluster ID is the index in this list.
+    clusters: Option<Vec<Vec<RegexComponents>>>,
     /// A number between `0.0..=1.0`. Higher values weight the benefit of merging
     /// chunks for a single page load more heavily. A site's bounce rate is a good
     /// approximation if you don't have a better value.
@@ -1165,6 +1168,8 @@ pub struct ChunkingHeuristicsConfig {
 
 #[turbo_tasks::value]
 pub struct ChunkingHeuristics {
+    /// The route-matching regexes for each user-defined cluster.
+    clusters: Vec<Vec<EsRegex>>,
     /// First-page-load priority as an integer percentage (`0..=100`), or `None` if unset.
     pub first_page_load_priority: Option<u32>,
     /// Route-matching regexes for priority routes.
@@ -1178,20 +1183,30 @@ pub struct ChunkingHeuristics {
 
 impl ChunkingHeuristics {
     /// Compute the [`EntryHeuristics`] for a route `pathname` by matching it against the configured
-    /// priority-route regexes.
+    /// cluster and priority-route regexes.
     pub fn entry_heuristics_for(&self, pathname: &str) -> EntryHeuristics {
-        let high_priority = self
-            .priority_routes
+        // Match `pathname` against a group of regexes.
+        let matches_any = |regexes: &[EsRegex]| {
+            regexes
+                .iter()
+                .filter(|regex| regex.as_regex_str().is_none())
+                .any(|regex| regex.is_match(pathname))
+                || regex::RegexSet::new(regexes.iter().filter_map(|regex| regex.as_regex_str()))
+                    .is_ok_and(|set| set.is_match(pathname))
+        };
+
+        let clusters = self
+            .clusters
             .iter()
-            .filter(|regex| regex.as_regex_str().is_none())
-            .any(|regex| regex.is_match(pathname))
-            || regex::RegexSet::new(
-                self.priority_routes
-                    .iter()
-                    .filter_map(|regex| regex.as_regex_str()),
-            )
-            .is_ok_and(|set| set.is_match(pathname));
-        EntryHeuristics { high_priority }
+            .enumerate()
+            .filter(|(_, regexes)| matches_any(regexes))
+            .map(|(index, _)| index as u16)
+            .collect();
+        let high_priority = matches_any(&self.priority_routes);
+        EntryHeuristics {
+            clusters,
+            high_priority,
+        }
     }
 }
 
@@ -2095,12 +2110,19 @@ impl NextConfig {
     #[turbo_tasks::function]
     pub fn chunking_heuristics(&self) -> Result<Vc<ChunkingHeuristics>> {
         let config = self.experimental.turbopack_chunking_heuristics.as_ref();
+        let clusters = config
+            .and_then(|c| c.clusters.as_deref())
+            .unwrap_or_default()
+            .iter()
+            .map(|patterns| parse_route_regexes(patterns))
+            .collect::<Result<Vec<_>>>()?;
         let priority_routes = parse_route_regexes(
             config
                 .and_then(|c| c.priority_routes.as_deref())
                 .unwrap_or_default(),
         )?;
         Ok(ChunkingHeuristics {
+            clusters,
             first_page_load_priority: config
                 .and_then(|c| c.first_page_load_priority)
                 .map(|priority| (priority.clamp(0.0, 1.0) * 100.0).round() as u32),
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -7,7 +7,7 @@ use either::Either;
 use rustc_hash::FxHashSet;
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::Value as JsonValue;
-use turbo_esregex::EsRegex;
+use turbo_esregex::{EsRegex, EsRegexSet};
 use turbo_rcstr::{RcStr, rcstr};
 use turbo_tasks::{
     FxIndexMap, NonLocalValue, OperationValue, ResolvedVc, TryJoinIterExt, Vc,
@@ -1150,6 +1150,9 @@ const DEFAULT_WEIGHT_DISTRIBUTION: f32 = 0.1;
 )]
 #[serde(rename_all = "camelCase")]
 pub struct TurbopackChunkingConfig {
+    /// Groups of pages commonly visited together, each defined by a list of regular expressions
+    /// matched against the route pathname. The cluster ID is the index in this list.
+    clusters: Option<Vec<Vec<RegexComponents>>>,
     /// A number between `0.0..=1.0`. Higher values weight the benefit of merging
     /// chunks for a single page load more heavily. A site's bounce rate is a good
     /// approximation if you don't have a better value.
@@ -1183,10 +1186,12 @@ pub struct TurbopackChunkingConfig {
 
 #[turbo_tasks::value]
 pub struct TurbopackChunking {
+    /// The route-matching regexes for each user-defined cluster.
+    clusters: Vec<EsRegexSet>,
     /// First-page-load priority as an integer percentage (`0..=100`), or `None` if unset.
     pub first_page_load_priority: Option<u32>,
     /// Route-matching regexes for priority routes.
-    priority_routes: Vec<EsRegex>,
+    priority_routes: EsRegexSet,
     /// Priority-route boost as an integer percentage (e.g. `150` for a 1.5x boost), or
     /// `None` to use the default.
     pub priority_boost_percent: Option<u32>,
@@ -1206,33 +1211,35 @@ pub struct TurbopackChunking {
 
 impl TurbopackChunking {
     /// Compute the [`EntryHeuristics`] for a route `pathname` by matching it against the configured
-    /// priority-route regexes.
+    /// cluster and priority-route regexes.
     pub fn entry_heuristics_for(&self, pathname: &str) -> EntryHeuristics {
-        let high_priority = self
-            .priority_routes
+        let clusters = self
+            .clusters
             .iter()
-            .filter(|regex| regex.as_regex_str().is_none())
-            .any(|regex| regex.is_match(pathname))
-            || regex::RegexSet::new(
-                self.priority_routes
-                    .iter()
-                    .filter_map(|regex| regex.as_regex_str()),
-            )
-            .is_ok_and(|set| set.is_match(pathname));
-        EntryHeuristics { high_priority }
+            .enumerate()
+            .filter(|(_, regexes)| regexes.is_match(pathname))
+            .map(|(index, _)| index as u16)
+            .collect();
+        let high_priority = self.priority_routes.is_match(pathname);
+        EntryHeuristics {
+            clusters,
+            high_priority,
+        }
     }
 }
 
-/// Compile a list of route-matching [`RegexComponents`] into [`EsRegex`]es.
-fn parse_route_regexes(patterns: &[RegexComponents]) -> Result<Vec<EsRegex>> {
-    patterns
+/// Compile a list of route-matching [`RegexComponents`] into an [`EsRegexSet`], which builds the
+/// combined [`regex::RegexSet`] up front so that matching a route doesn't have to.
+fn parse_route_regexes(patterns: &[RegexComponents]) -> Result<EsRegexSet> {
+    let regexes = patterns
         .iter()
         .cloned()
         .map(|pattern| {
             EsRegex::try_from(pattern)
                 .context("Invalid route pattern in `experimental.turbopackChunking`")
         })
-        .collect()
+        .collect::<Result<Vec<_>>>()?;
+    Ok(EsRegexSet::new(regexes))
 }
 
 /// Resolve `experimental.cssChunking` to the [`StyleGroupsAlgorithm`] Turbopack should use.
@@ -2165,12 +2172,19 @@ impl NextConfig {
     #[turbo_tasks::function]
     pub fn turbopack_chunking(&self) -> Result<Vc<TurbopackChunking>> {
         let config = self.experimental.turbopack_chunking.as_ref();
+        let clusters = config
+            .and_then(|c| c.clusters.as_deref())
+            .unwrap_or_default()
+            .iter()
+            .map(|patterns| parse_route_regexes(patterns))
+            .collect::<Result<Vec<_>>>()?;
         let priority_routes = parse_route_regexes(
             config
                 .and_then(|c| c.priority_routes.as_deref())
                 .unwrap_or_default(),
         )?;
         Ok(TurbopackChunking {
+            clusters,
             first_page_load_priority: config
                 .and_then(|c| c.first_page_load_priority)
                 .map(|priority| (priority.clamp(0.0, 1.0) * 100.0).round() as u32),
--- a/turbopack/crates/turbo-esregex/src/lib.rs
+++ b/turbopack/crates/turbo-esregex/src/lib.rs
@@ -161,6 +161,89 @@ impl EsRegex {
     }
 }
 
+/// A group of [`EsRegex`]es matched against a haystack as a unit.
+///
+/// The members backed by the `regex` crate are compiled into a single [`regex::RegexSet`] once,
+/// when the group is built, rather than on every match. The remainder (those that fall back to
+/// `regress`, e.g. for lookahead) are matched one at a time.
+#[derive(Debug, Clone)]
+#[turbo_tasks::value(eq = "manual", shared, serialization = "custom")]
+pub struct EsRegexSet {
+    /// The members, in the order they were given. Also the source of truth for equality and
+    /// serialization, since [`regex::RegexSet`] supports neither.
+    regexes: Vec<EsRegex>,
+    /// The combined members, or `None` if the combined program couldn't be built.
+    #[turbo_tasks(trace_ignore)]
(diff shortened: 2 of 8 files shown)


turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -483,10 +530,15 @@ pub async fn make_production_chunks(
 
                             let p2 = 100 - p1;
 
+                            // `d(N = 2)` scaled by (rem_g * groups) / o_groups, see above.
+                            let d_n2 = if cluster_overlap {
+                                rem_g * c_req - (a_rem * a_size + b_rem * b_size)
```
sokra: What is the mathematically reasoning behind that?

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -473,6 +503,23 @@ pub async fn make_production_chunks(
...
+                                // the same cluster, weight the `N = 2` Z+Z request benefit higher
+                                // so commonly co-visited pages
+                                // merge more readily.
+                                if has_clusters {
+                                    // clusters that these chunk groups are in
+                                    let mut seen = RoaringBitmap::new();
+                                    for index in o.iter() {
+                                        let clusters = &heuristics.clusters[index as usize];
+                                        if clusters.iter().any(|&c| seen.contains(c as u32)) {
+                                            cluster_overlap = true;
```
```suggestion
                                            cluster_overlap += 1;
```

You want to count the number of overlapping clusters, because each cluster increases the probability.

But you also need to count the cluster in the non-overlapping parts, as each cluster there decreases the probability.

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sokra: Here are the probabilities you need to adjust to make it mathematically correct.
The `.../rem_g` parts in all these terms are probabilities of navigating from one chunk group to another.
They do change when clusters are involved.

Let's take this as example:
```
case X + X (p = (a_rem/groups) * ((a_rem - 1)/rem_g)): size = a_size, requests = 1
```

Currently we assume that the probability moving from X to X' is `(a_rem - 1)/rem_g`, but that changes with clusters.
If there is a cluster within X it increases, if there is a cluster between X and Y/Z it decreases.

So you have to determine the number of cluster between chunk groups of these three groups (X=a_rem, Y=b_rem, Z=o_groups):

| co-clustering | A rem | B rem | overlap |
|---|---|---|---|
| A rem | ? | ? | ? | 
| B rem | ? | ? | ? | 
| overlap | ? | ? | ? | 

Now we have two options:
* Being a cluster makes it X times more likely that a route is followed. So maybe there there are 100 groups in total. A cluster would change the probability from 1% to 2%. meh...
* There is a X% (e.g. 60%) probably that clusters will be followed, an the remaining (100-X)% (e.g. 40%) will be distributed on the non-cluster routes. So with 100 groups and a single cluster it would make the cluster route 60% and the remaining routes would be 0.4% instead of 1%. That sounds good.

So we change the `(a_rem - 1)/rem_g` to `(a_rem_a_rem_non_cluster - 1)/rem_g*X + a_rem_a_rem_cluster/rem_g*(1-X)`

```
a_rem_a_rem_cluster = co_clustering.a_rem.a_rem
a_rem_a_rem_non_cluster = a_rem - a_rem_a_rem_cluster - a_rem_b_rem_cluster - a_rem_overlap_cluster
```

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: in the old model, you could simplify this down algebraically. i've dropped that model for your suggestion though: https://github.com/vercel/next.js/pull/95157#discussion_r3568396460.

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: yep, +1, this has been added now.

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: thank you for this! I took a stab at implementing this version:

> There is a X% (e.g. 60%) probably that clusters will be followed, an the remaining (100-X)% (e.g. 40%) will be distributed on the non-cluster routes. So with 100 groups and a single cluster it would make the cluster route 60% and the remaining routes would be 0.4% instead of 1%. That sounds good.

Here's the code comment that runs through the updated maths (basically how we calculate the new transition probability):

X, Y and Z are three sets of chunk groups:
    X = the a_rem groups that load only chunk A
    Y = the b_rem groups that load only chunk B
    Z = the o_groups groups that load both

Two groups that sit in the same cluster form a pair. The table below
counts pairs: each cell says how many pairs have one group in one set
and the other group in another set. The diagonal cells (c_xx, c_yy,
c_zz) count pairs where both groups are in the same set.

            | X (a_rem) | Y (b_rem) | Z (overlap)
    --------+-----------+-----------+------------
    X a_rem |    c_xx   |    c_xy   |    c_xz
    Y b_rem |    c_xy   |    c_yy   |    c_yz
    Z overl |    c_xz   |    c_yz   |    c_zz

Each row sum is the total number of pairs leaving that set:
    paired_x = c_xx + c_xy + c_xz
    paired_y = c_xy + c_yy + c_yz
    paired_z = c_xz + c_yz + c_zz

`cluster_navigation_probability` (= 0.6, `cnp` below) is the chance a
navigation stays within a cluster.

For a first group in set 1, trans(1 -> 2) can be calculated using
the following probability tree:

                    Will the navigation stay
                       within a cluster?
                    /                    \
             yes (cnp)                 no (1 - cnp)
                  /                        \
       Does it go to set 2?         Does it go to set 2?
                |                            |
         c_12 / paired_1     non_paired_2 / (rem_g - paired_1)

    trans(1 -> 2) = cnp       * (c_12         / paired_1)
                  + (1 - cnp) * (non_paired_2 / (rem_g - paired_1))

When set 1 == set 2, we subtract one from count_2 and non_paired_2.
paired_1 is the number of groups that are paired with set 1 via a
cluster.

UNMERGED CASE (N = 2):
case X + X (p = (a_rem/groups) * trans(X -> X)): size = a_size, requests = 1
case Y + Y (p = (b_rem/groups) * trans(Y -> Y)): size = b_size, requests = 1
case Z + Z (p = (o_groups/groups) * trans(Z -> Z)): size = a_size + b_size, requests = 2
case X + Y (p = (a_rem/groups) * trans(X -> Y) + (b_rem/groups) * trans(Y -> X)): size = a_size + b_size, requests = 2
case X + Z (p = (a_rem/groups) * trans(X -> Z) + (o_groups/groups) * trans(Z -> X)): size = a_size + b_size, requests = 2
case Y + Z (p = (b_rem/groups) * trans(Y -> Z) + (o_groups/groups) * trans(Z -> Y)): size = a_size + b_size, requests = 2

MERGED CASE (N = 2):
case X + X (p = (a_rem/groups) * trans(X -> X)): size = a_size, requests = 1
case Y + Y (p = (b_rem/groups) * trans(Y -> Y)): size = b_size, requests = 1
case Z + Z (p = (o_groups/groups) * trans(Z -> Z)): size = (a_size + b_size), requests = 1
case X + Y (p = (a_rem/groups) * trans(X -> Y) + (b_rem/groups) * trans(Y -> X)): size = a_size + b_size, requests = 2
case X + Z (p = (a_rem/groups) * trans(X -> Z) + (o_groups/groups) * trans(Z -> X)): size = a_size + (a_size + b_size), requests = 2
case Y + Z (p = (b_rem/groups) * trans(Y -> Z) + (o_groups/groups) * trans(Z -> Y)): size = b_size + (a_size + b_size), requests = 2

Does that seem right to you? (sorry for the big chunk of text :sob:)

sokra: The math starting to get hard.

Maybe you could move it from the code comment to a separate markdown file, which explains all the math behind it with good explanations.

sampoder: Could I get a sanity check that this is what you were thinking? I wasn't quite sure what your comment meant originally

sampoder: Moved to MD: https://github.com/vercel/next.js/blob/b6d59952b8feb4dc68430922e0472482f84ba564/turbopack/crates/turbopack-core/src/chunk/chunking/chunk_merging_cost_benefit.md

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -487,14 +401,55 @@ pub async fn make_production_chunks(
...
+
+                            - pairs_to_dest: co-clustered pairs from source to dest
+                            - source_pairs: all co-clustered pairs leaving source (its row sum)
+                            - source_size: number of groups in the source set
+                            - dest_size: groups in the dest set (minus 1 if source == dest) */
+                            let transition_probability =
+                                |pairs_to_dest: f64,
+                                 source_pairs: f64,
+                                 source_size: f64,
+                                 dest_size: f64| {
```
sampoder: rename to `_groups`

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -487,14 +401,55 @@ pub async fn make_production_chunks(
...
+                                    if source_pairs == 0.0 {
+                                        // Source has no pairs: navigate uniformly.
+                                        return dest_size / rem_g;
+                                    }
+                                    let non_paired = rem_g * source_size - source_pairs;
```
sampoder: rem_g - source_pairs

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -487,14 +401,55 @@ pub async fn make_production_chunks(
...
+                                    if non_paired <= 0.0 {
+                                        // Every other group is paired: all weight on the pairs.
+                                        return pairs_to_dest / source_pairs;
+                                    }
+                                    let non_paired_to_dest =
+                                        dest_size * source_size - pairs_to_dest;
+                                    cluster_navigation_probability * (pairs_to_dest / source_pairs)
+                                        + (1.0 - cluster_navigation_probability)
+                                            * (non_paired_to_dest / non_paired)
```
sampoder: source_size *

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: @sokra, I had another read of this and I believe the document was wrong and the code was right. 

This was the original copy;

Image

`non_paired_2` really should have been `non_paired_12`. So this is the total number of navigations between 1 and 2 outside of a cluster. Therefore, the denominator should be the total number of navigations outside of the cluster. 

This is equivalent to the amount of groups in set 1 (source_size) multiplied by the total groups - 1 (rem_g) and then with the number of navigations from set 1 inside of the cluster subtracted from it (source_pairs). 

Hence, `rem_g * source_size - source_pairs`.

turbopack/crates/turbopack-core/src/chunk/chunking/chunk_merging_cost_benefit.md
```diff
@@ -0,0 +1,432 @@
...
+$$
+\begin{flalign*}
+& \begin{aligned}
+\text{non\\_paired}\_{12} &= |S_1| \cdot |S_2| - c\_{12} \\
+\text{non\\_paired}\_1 &= (\text{groups} - 1) \cdot |S_1| - \text{paired}\_1
+\end{aligned} &
+\end{flalign*}
+$$
+
+When set 1 and set 2 are the same set, $|S_2|$ becomes $|S_2| - 1$, since a group cannot navigate to itself.
```
lukesandberg: $|S_1| - 1$ ?

crates/next-core/src/next_config.rs
```diff
@@ -1192,34 +1179,34 @@ pub struct TurbopackChunking {
...
     pub fn entry_heuristics_for(&self, pathname: &str) -> EntryHeuristics {
-        let high_priority = self
-            .priority_routes
+        // Match `pathname` against a group of regexes.
+        let matches_any = |regexes: &[EsRegex]| {
+            regexes
+                .iter()
+                .filter(|regex| regex.as_regex_str().is_none())
+                .any(|regex| regex.is_match(pathname))
+                || regex::RegexSet::new(regexes.iter().filter_map(|regex| regex.as_regex_str()))
```
lukesandberg: for this to be at all useful, you should cache the regex set, ideally as part of building the EsRegex objects

lukesandberg: as part of building them would be natural, using a new type to encapsulate it

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -475,17 +326,70 @@ pub async fn make_production_chunks(
...
+                                    let y = &***b - &o; // b_rem groups: load only chunk B
+
+                                    // Map each cluster to the candidate groups it contains.
+                                    let mut cluster_groups: FxHashMap<u16, RoaringBitmap> =
+                                        FxHashMap::default();
+                                    for set in [&x, &y, &o] {
+                                        for index in set.iter() {
+                                            for &c in &heuristics.clusters[index as usize] {
+                                                cluster_groups.entry(c).or_default().insert(index);
+                                            }
```
lukesandberg, line 368: you should be able to use a union operation `cluster_groups.entry(c).or_default() |= set`?

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -475,17 +326,70 @@ pub async fn make_production_chunks(
...
+
+                                    for index in x.iter() {
+                                        let p = pairs_with(index);
+                                        c_xx += (&p & &x).len() as f64;
+                                        c_xy += (&p & &y).len() as f64;
+                                        c_xz += (&p & &o).len() as f64;
+                                    }
+                                    for index in y.iter() {
+                                        let p = pairs_with(index);
+                                        c_yy += (&p & &y).len() as f64;
```
lukesandberg: use `intersection_len`

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -475,17 +326,70 @@ pub async fn make_production_chunks(
...
+                                        c_yz += (&p & &o).len() as f64;
+                                    }
+                                    for index in o.iter() {
+                                        c_zz += (&pairs_with(index) & &o).len() as f64;
```
lukesandberg: ditto

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -497,14 +401,57 @@ pub async fn make_production_chunks(
                             let o = o_groups as f64;
                             let groups = groups as f64;
                             let rem_g = rem_g as f64;
+                            let a_rem = a_rem as f64;
+                            let b_rem = b_rem as f64;
+                            let a_size = a_size as f64;
+                            let b_size = b_size as f64;
+
+                            let cluster_navigation_probability = 0.6;
```
lukesandberg: nit: const?

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -497,14 +401,57 @@ pub async fn make_production_chunks(
...
+                                    }
+                                    let non_paired_from_source_to_dest =
+                                        dest_groups * source_groups - pairs_to_dest;
+                                    cluster_navigation_probability * (pairs_to_dest / source_pairs)
+                                        + (1.0 - cluster_navigation_probability)
+                                            * (non_paired_from_source_to_dest
+                                                / non_paired_from_source)
+                                };
+
+                            let p_zz = transition_probability(c_zz, paired_z, o, o - 1.0);
```
lukesandberg, line 455: s/trans/transition_probability/ in your markdown? maybe?

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -813,9 +831,12 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
             }
         }
 
-        // Resolve per-chunk-group chunking heuristics. Entry chunk groups carry their route's
-        // priority-route flag; other chunk groups inherit it (OR) from their referencing chunk
-        // groups.
+        // Resolve per-chunk-group chunking heuristics. Entry
+        // chunk groups carry their route's clusters / priority-route flag; other chunk groups
+        // inherit the union of clusters (and OR of the flag) from their referencing chunk groups.
+        let chunk_group_count = chunk_groups_map.len();
```
lukesandberg: nit:inline

lukesandberg:
it would really be great to have unit tests demonstrating these kinds of behaviors

i started down this path in https://github.com/vercel/next.js/pull/90847 maybe merge it?

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: we need the filter here, `heuristics.clusters` is every group in the cluster and `cluster_groups.entry(c)` contains every cluster the `c` group is in.

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: tbh, I prefer the short hand

turbopack/crates/turbo-esregex/src/lib.rs
```diff
@@ -161,6 +161,77 @@ impl EsRegex {
...
+        }
+    }
+}
+
+impl EsRegexSet {
+    /// Builds the combined matcher. Fails if the members cannot be compiled into a single
+    /// [`regex::RegexSet`]; the combined program has its own size limit, so compiling
+    /// individually is no guarantee.
+    pub fn new(regexes: Vec<EsRegex>) -> Result<Self> {
+        let set = regex::RegexSet::new(regexes.iter().filter_map(EsRegex::as_regex_str))?;
```
lukesandberg: if it fails can we just fallback to an empty set? Option<RegexSet>?

turbopack/crates/turbo-esregex/src/lib.rs
```diff
@@ -161,6 +161,77 @@ impl EsRegex {
...
+        Ok(Self { regexes, set })
+    }
+
+    /// Returns true if any member matches somewhere in the `haystack`.
+    pub fn is_match(&self, haystack: &str) -> bool {
+        self.set.is_match(haystack)
```
lukesandberg: in the common case all regexes will be in the set, so we should record which ones are in and not in the set to avoid querying every one twice if it doesn't match

lukesandberg (approved):

PR #95179, opened 2026-06-25
Turbopack: Add test for node package with incompatible module and default exports

bgw:


bgw:
It seems like this is a totally different issue than https://vercel.enterprise.slack.com/archives/C0BD40PAW6Q

This test is just showing that node doesn't implement `exports.module` and turbopack does.

PR #95133, opened 2026-06-24
Don't track modifications to transient data stored in tasks

lukesandberg:
Many fields in TaskStorage store a mix of transient and persistent data.  When we serialize we leverage the `filter_transient` tag (and the `IsTransient` trait) to avoid persisting it, however, when mutations occur we are still marking the tasks as `modified` even when we are only modifying transient data.

This leads to unnecessary persistence cycles.

The solution is simply to not call `track_modification` when the modification is about transient data items.

For most fields this is handled in the macro accessors, but for `cell_data` this is managed via some new 'handwritten' mutators implemented on `TaskGuard`.  This is desirable to decrease the number of times we do registry lookups since most callers who modify the cell_data already have the ValueType in hand

PR #95180, opened 2026-06-25
[turbopack] Support undoable modifications for many field types on TaskStorage

lukesandberg:
## What
Allow some modifications that turn out to be no-ops to be undone on TaskStorage.  This saves us some redundant check-then-mutate cycles.

This doesn't apply to every field type
* `Option<T>` - still read-check-modify
* `CounterMap<T>` - updates are unconditional tracks and removals use the undo approach
* `AutoSet`/`AutoMap` - mostly uses the new 'undo' approach 
   * `take` as a bulk operation does not, taking an empty set is rare and cheap to check for
   * `AutoMap::insert` doesn't use the undo approach or even check, we always track a modification since all callers are only inserting unique values anyway

A few other incidental cleanups were landed as well.

## Why

Avoid redundant hash lookups.

## How

Have `track_modification` return a new TrackOutcome that records what modifications were made and then in the undo case we can pass it to a new `undo_track_modification` function which can reverse it.   This is a little brittle since both operations need to occur within a single lock transition

I considered using a 'lambda' oriented approach but decided against it since it would be awkward for some of the mutators and as all the callers are generated by a macro (well _almost_ all callers) it isn't too risky.

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
@@ -1251,7 +1251,9 @@ pub trait TaskGuard: Debug + TaskStorageAccessors {
         // tracked
         self.check_access(SpecificTaskDataCategory::Data);
         if matches!(persistence, ValueTypePersistence::Persistable(..)) {
-            self.track_modification(SpecificTaskDataCategory::Data, "cell_data");
+            // Unconditional track when persistable: an insert always changes the data snapshot
+            // (a fresh cell, or a replaced value). No no-op to undo.
+            let _ = self.track_modification(SpecificTaskDataCategory::Data, "cell_data");
         }
         self.typed_mut().cell_data_mut().insert(cell, value)
     }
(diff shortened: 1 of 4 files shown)


lukesandberg:
* **#95180** <a href="https://app.graphite.com/github/pr/vercel/next.js/95180?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95180?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#95133** <a href="https://app.graphite.com/github/pr/vercel/next.js/95133?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -1466,12 +1476,21 @@ impl TaskStorageAccessors for TaskGuardImpl<'_> {
         &mut self,
         category: crate::backend::storage::SpecificTaskDataCategory,
         name: &str,
-    ) {
-        if !self.task_id.is_transient() {
-            self.task.track_modification(category, name);
+    ) -> crate::backend::storage::TrackOutcome {
```
sokra:
```suggestion
    ) -> TrackOutcome {
```

sokra (approved):
That looks good. Might be a performance win considering how often we read before write...

PR #95182, opened 2026-06-26
server actions: enforce bodySizeLimit in edge runtime for multipart and non-multipart bodies

### LINKED ISSUE #95172: Edge Server Actions do not enforce serverActions.bodySizeLimit for multipart form submissions

https://github.com/Str1ckl4nd/next-edge-server-actions-body-limit-repro

### To Reproduce

```bash
npm install
npm run make-payload
npm run dev
```

Then:

1. Open `http://localhost:3000/node`.
2. Upload the generated `payload.bin` file and submit the form.
3. Open `http://localhost:3000/edge`.
4. Upload the same `payload.bin` file and submit the form.

Expected: both runtimes reject the 3 MiB upload because the Server Actions body limit is `2mb`.

Observed: the Node runtime rejects the request with `Body exceeded 2mb limit.`, while the Edge runtime accepts the same upload and logs `File name: payload.bin size: 3145728`.

Current behavior:

The Node runtime Server Action rejects the 3 MiB multipart upload because `experimental.serverActions.bodySizeLimit` is set to `2mb`. The observed error is:

```text
Body exceeded 2mb limit.
```

The equivalent Edge runtime Server Action accepts the same 3 MiB multipart upload and executes the action. The observed server log is:

```text
File name: payload.bin size: 3145728
```

Expected behavior:

Both runtimes should enforce the configured Server Actions body limit consistently. With `experimental.serverActions.bodySizeLimit: '2mb'`, a 3 MiB multipart Server Action request should be rejected in both the Node and Edge runtimes.

Why this appears to happen:

In `v16.3.0-canary.67`, the Edge runtime multipart Server Action path calls `req.request.formData()` directly:

https://github.com/vercel/next.js/blob/v16.3.0-canary.67/packages/next/src/server/app-render/action-handler.ts#L771-L773

The Node runtime path calculates `bodySizeLimitBytes`, applies `sizeLimitTransform`, and raises the `Body exceeded ... limit` error when the stream exceeds the configured limit:

https://github.com/vercel/next.js/blob/v16.3.0-canary.67/packages/next/src/server/app-render/action-handler.ts#L901-L923

For multipart action requests, the Node path then pipes the request body through the size-limiting transform before decoding the action payload:

https://github.com/vercel/next.js/blob/v16.3.0-canary.67/packages/next/src/server/app-render/action-handler.ts#L953-L958

The result is a runtime-specific limit bypass for multipart Server Action form submissions. An application that relies on `serverActions.bodySizeLimit` to bound request memory/work gets the configured limit in Node, but the Edge action path can still parse and execute a larger multipart body.

```bash
Operating System:
  Platform: linux
Binaries:
  Node: 22.x
  npm: 10.x
Relevant Packages:
  next: 16.3.0-canary.67
  react: 19.2.7
  react-dom: 19.2.7
Next.js Config:
  experimental.serverActions.bodySizeLimit: 2mb
```

Server Actions

next dev (local)

### Additional context

_No response_

tsushanth:
Fixes #95172

## What

The edge runtime Server Action handler has a `// TODO: add body limit` comment at the top of the edge branch. Neither the multipart path nor the non-multipart (fetch-action) path enforced `experimental.serverActions.bodySizeLimit`, so a multipart upload that exceeds the configured limit was silently accepted on edge routes while the identical request was correctly rejected on Node routes.

The reporter's repro confirms:
- Node runtime: `Body exceeded 2mb limit.` (correct)
- Edge runtime: action executes with a 3 MiB payload (bug)

## How

**Size parsing.** The Node path uses `require('next/dist/compiled/bytes')` lazily so the dependency is excluded from edge bundles. Because edge bundles cannot load that module at runtime, the limit string is parsed inline with a small multiplier table covering the same kb/mb/gb/tb/pb units that `SizeLimit` accepts. Numeric values pass through directly.

**Multipart path.** `req.request.formData()` reads the body internally with no size hook. Instead, the body stream is now consumed chunk-by-chunk through `req.body.getReader()`, accumulating byte count and throwing on overflow before parsing. The validated chunks are concatenated into a `Uint8Array` and wrapped in a new `Request` that carries the original headers (including the multipart `Content-Type` boundary) so `formData()` can parse the already-buffered body correctly.

**Non-multipart fetch-action path.** The existing chunk-reading loop is extended with the same per-chunk accumulator and overflow check. The `chunks` type is widened from `Buffer[]` to `Uint8Array[]` (which `Buffer.concat` already accepts) to avoid relying on a Node global.

The error message and documentation link match the Node path exactly so client-side error handling sees the same text regardless of runtime.

## Testing

The change is in a dead-code-eliminated block (`process.env.NEXT_RUNTIME === 'edge'`), so the Node runtime path is untouched. Manual verification using the reporter's reproduction repo at https://github.com/Str1ckl4nd/next-edge-server-actions-body-limit-repro confirms that the edge `/edge` route now returns a 500 (body exceeded error) for the 3 MiB upload when `bodySizeLimit: '2mb'` is configured.

### DIFF
--- a/packages/next/src/server/app-render/action-handler.ts
+++ b/packages/next/src/server/app-render/action-handler.ts
@@ -756,7 +756,32 @@ export async function handleAction({
             throw new Error('invariant: Missing request body.')
           }
 
-          // TODO: add body limit
+          const defaultBodySizeLimitEdge = '1 MB'
+          const bodySizeLimitEdge =
+            serverActions?.bodySizeLimit ?? defaultBodySizeLimitEdge
+          const bodySizeLimitBytesEdge =
+            typeof bodySizeLimitEdge === 'number'
+              ? bodySizeLimitEdge
+              : (() => {
+                  // Parse human-readable size strings (e.g. "2mb", "512 KB") without
+                  // relying on the Node-only `bytes` module, which cannot be bundled
+                  // into the edge runtime.
+                  const match = String(bodySizeLimitEdge)
+                    .trim()
+                    .match(/^(\d+(?:\.\d+)?)\s*(k|m|g|t|p)?b?$/i)
+                  if (!match) return 1024 * 1024 // default 1 MB
+                  const num = parseFloat(match[1])
+                  const unit = (match[2] ?? '').toLowerCase()
+                  const multipliers: Record<string, number> = {
+                    '': 1,
+                    k: 1024,
+                    m: 1024 * 1024,
+                    g: 1024 * 1024 * 1024,
+                    t: 1024 * 1024 * 1024 * 1024,
+                    p: 1024 * 1024 * 1024 * 1024 * 1024,
+                  }
+                  return Math.floor(num * (multipliers[unit] ?? 1))
+                })()
 
           // Use react-server-dom-webpack/server
           const {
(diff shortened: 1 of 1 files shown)


tsushanth:
Closing — this has been independently fixed on `canary` already (same approach: size-checking loop while reading the edge-runtime body stream, enforcing `bodySizeLimit`), and the upstream version is actually better than mine: it returns a proper 413 via `ApiError` instead of a generic `Error`, and fixes a Content-Type casing bug in the multipart Blob reconstruction that I hadn't caught. Nothing left for this PR to contribute — my branch's version of this file is now byte-identical to canary's.

PR #95184, opened 2026-06-26
Return APP_PAGE instead of PAGES for production Instant Navigation tests

devjiwonchoi:
### Why?

Current behavior: https://github.com/vercel/next.js/actions/runs/28225010081/job/83615571987?pr=95184#step:31:8732

For a minimal-mode resume request carrying the Instant Navigation testing cookie, Next.js skips the postponed render so the test receives only the prefetched static shell. The resulting empty continuation was represented as `PAGES`, but the App Page handler requires `APP_PAGE`, causing error:

```
Error: Invariant app-page handler received invalid cache entry PAGES
```

### How?

Return an empty `APP_PAGE` response with the correct HTML or RSC content type and postponed metadata. If the document continuation has no `<head>` chunk, insert the Instant Navigation bootstrap while flushing the stream.

Add production regression coverage for document and RSC-prefetch resume requests using build-generated postponed state.

### DIFF
--- a/packages/next/src/build/templates/app-page.ts
+++ b/packages/next/src/build/templates/app-page.ts
@@ -61,7 +61,6 @@ import {
   CachedRouteKind,
   IncrementalCacheKind,
   type CachedAppPageValue,
-  type CachedPageValue,
   type ResponseCacheEntry,
   type ResponseGenerator,
 } from '../../server/response-cache' with { 'turbopack-transition': 'next-server-utility' }
@@ -1372,7 +1371,8 @@ export async function handler(
         }
 
         // When we're in minimal mode, if we're trying to debug the static shell,
-        // we should just return nothing instead of resuming the dynamic render.
+        // return an empty App Page response instead of resuming the dynamic
+        // render. The static shell has already been streamed by the platform.
         if (
           (isDebugStaticShell || isDebugDynamicAccesses) &&
           typeof postponed !== 'undefined'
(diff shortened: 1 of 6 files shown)


gnoff (approved):

unstubbable (commented):
Thanks for digging into this. The `PAGES → APP_PAGE` change is the right fix and I think we should land it on its own. I [deployed the branch manually](https://vtest314-e2e-tests-8zy2wutjx-hendrik-liebaus-projects.vercel.app/) and confirmed the resume no longer throws the `received invalid cache entry PAGES` invariant.

One thing I'd drop before merging: the new script append in the `flush` of `createInstantTestScriptInsertionTransformStream`. It doesn't actually do anything on a real deployment. I tested it and `self.__next_instant_test` never reaches the browser, because with the cookie set the document is served from the cached static prelude, which was prerendered without the cookie and so has no script in it. The empty-body case that branch handles only happens via the debug-shell early-return in `app-page.ts`, and that only reaches the browser through the synthetic `next-resume` POST in the new test. So it's really only asserting against the test's own setup, not real behavior.

So for this PR I'd keep the kind fix, drop the `flush` append, and simplify the test to just assert the resume returns 200 and no longer logs the invariant (for both the document and RSC cases). The `self.__next_instant_test` assertion can go since it only reflects the synthetic path.

Making the instant behavior actually work when deployed is a separate problem worth its own PR. Right now the script isn't in the cached prelude, so hydration errors on the locked load and the unlock falls back to a hard reload. The real fix might be to get the cookie-guarded `__next_instant_test` bootstrap into the prelude's `<head>` (e.g. via `bootstrapScriptContent`, gated on `exposeTestingApiInProductionBuild` and inert without the cookie) so it runs before the bootstrap script reads it. I'll follow up with that.

unstubbable (approved):

unstubbable:
> Making the instant behavior actually work when deployed is a separate problem worth its own PR. Right now the script isn't in the cached prelude, so hydration errors on the locked load and the unlock falls back to a hard reload. The real fix might be to get the cookie-guarded `__next_instant_test` bootstrap into the prelude's `<head>` (e.g. via `bootstrapScriptContent`, gated on `exposeTestingApiInProductionBuild` and inert without the cookie) so it runs before the bootstrap script reads it. I'll follow up with that.

Handled in #95222.

PR #95186, opened 2026-06-26
[dev overlay]: restructure Copy-as-prompt body into a step-by-step checklist

aurorascharff:
## What

Restructures the dev overlay's **Copy as prompt** body from three loosely-grouped instructions into a seven-step ordered checklist.

## Why

Agents tend to do one thing per step. The old step 4 bundled "reload in browser," "set up browser tooling," "check the shell," "re-check siblings," and a before/after capture into a single paragraph, so most of it got skipped.

Two other gotchas got their own steps: the empty-shell case (`<Suspense fallback={null}>` around the page body passes the build but renders nothing) and the sibling re-check on shared-code changes.

## How

Same checklist shape as the `next-cache-components-adoption` skill's per-route loop, so agents following either workflow apply the same sequence. Used for the in-overlay copy button and the prompt mirrored in `next dev` / `next build` terminal output.

### DIFF
--- a/packages/next/src/next-devtools/dev-overlay/components/instant/instant-guidance.tsx
+++ b/packages/next/src/next-devtools/dev-overlay/components/instant/instant-guidance.tsx
@@ -84,11 +84,19 @@ function CopyPromptButton({
     '',
     'Steps:',
     '',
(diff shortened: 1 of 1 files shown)


gaojude:
Let's add a before vs after screenshot to the PR description

gaojude (approved):

gaojude (approved):

Issue #86151, opened 2025-11-15
`loading.js` sometimes causes soft-navigation to get stuck and not render the new page, despite receiving it from the server


mogzol:
### Link to the code that reproduces this issue

https://github.com/mogzol/next-stuck-navigation-repro

Relevant components:
- [RefreshButton.tsx](https://github.com/mogzol/next-stuck-navigation-repro/blob/main/src/components/RefreshButton.tsx) - Calls `router.refresh()` to trigger the soft refresh that sometimes gets stuck
- [page.tsx](https://github.com/mogzol/next-stuck-navigation-repro/blob/main/src/app/page.tsx) - Renders the refresh button as well as several slow client components that increase page size and render time
- [loading.tsx](https://github.com/mogzol/next-stuck-navigation-repro/blob/main/src/app/loading.tsx) - The page's loading UI. This file existing causes the issue to occur. The actual contents of the loading UI don't seem to matter, just whether it exists or not.

### To Reproduce

1. Start the application **in production mode** (`npm run build && npm run start`)
2. Open the page in your browser and press the "Refresh" button on the page to trigger a soft refresh (calls `router.refresh()`)
3. Observe that sometimes the new page never loads, and the loading indicator stays visible indefinitely.
    - The request for the new page still completes successfully in the network monitor
    - Changing network request throttling can make this issue more or less likely ("Fast 4G" almost always triggers it, "Slow 4G" never does)
    - Changing the `GARBAGE_ITEMS` environment variable before starting the app also makes the issue more or less likely
        - `GARBAGE_ITEMS=10 npm run start` - Almost never has the issue
        - `GARBAGE_ITEMS=100 npm run start` - Almost always has the issue
    - Removing the `loading.tsx` file for the page makes the issue go away completely

Here's a screen recording showing the issue. The first couple refreshes work fine, but then it gets stuck on the third one. Then after performing a full page reload, it gets stuck again on the first refresh. Then after changing the network request throttling to "Slow 4G", the issue goes away.

https://github.com/user-attachments/assets/6ec273db-20b8-4f97-9310-04c8e13aa007

### Current vs. Expected behavior

I would expect the soft refresh to always work, and never get stuck. Currently, if a page has a `loading.js` file, it randomly gets stuck. It seems that this is more likely if your page has client components that take a while to render.

### Provide environment information

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32673
[...]

PR #95188, opened 2026-06-26
fix(app-router): don't run remaining actions when discarding a non-revalidating server action

errmakov:
### What?

Addresses a soft-navigation segment drop reported in #86151: navigating to a route with a `loading.tsx`/Suspense boundary while a Server Action is in flight commits the URL but never renders the destination segment — the resolved content stays inside the hidden boundary and is never swapped in.

### Why?

#82674 added a `runRemainingActions(actionQueue, setState)` call to the `if (action.discarded)` branch of `handleResult`. When a Server Action is discarded by a boundary-gated navigation that is still streaming, that call dispatches a `setState` mid-transition, which interrupts the destination segment's boundary completion and drops it.

The regression bisects to #82674 (re-introduced at `15.5.1-canary.8`, clean on `15.5.0`), with the vendored React identical across that boundary; grafting the pre-#82674 client files onto canary.8 also makes it clean. Re-instating the eager refresh does **not** recover the segment (a `router.refresh()` that works on a boundary-free route still drops a boundary-gated one), confirming the disturbance — not the refresh removal — is the cause.

### How?

Only run the remaining actions when the discarded action actually revalidated (the case #82674 added the refresh for). Otherwise there is no state to apply, so we return without disturbing the in-flight navigation. This keeps #82674's optimization and its revalidation refresh intact.

### Validation

- Both existing `test/e2e/app-dir/actions/app-action.test.ts` "action discarding" tests still pass (no-revalidation → no refresh; revalidation → refresh).
- Deterministic external reproduction goes 100% → 0% with this change: https://github.com/errmakov/next-soft-nav-segment-drop-repro

### Scope

Referenced (not `fixes`) #86151 deliberately: per discussion in the thread there appears to be a second, older drop (reproduces on 15.5.0, with no destination DOM at all) that this change does not address. This PR fixes the #82674 regression specifically.


deepshekhardas:
Well-documented fix with reproduction and deterministic test results. All existing \ction discarding\ e2e tests pass. This addresses a real regression from #82674 — would be great to get a review on this.

gaearon:
If this fixes a bug, it needs to come with a test that fails without the fix.

errmakov:
Added one in `test/e2e/app-dir/actions-discarded-segment-drop`. It reproduces the drop end-to-end: a provider's slow on-mount Server Actions are still in flight when a nested `staff → staff/[id]/schedule` navigation commits behind a `loading.tsx` boundary; without this change the discarded action runs the rest of the queue mid-navigation and the schedule segment is dropped (the page stays stale on `/staff`). It fails without the fix and passes with it.

The drop is a timing-sensitive concurrent-commit race, so the test makes it reliable rather than incidental: a CPU-throttled browser reloaded before each navigation reproduces it ~90% per iteration on the unfixed client and 0% with the fix. It runs in `next start` only (the drop needs the production streamed boundary) and breaks on the first reproduced drop so it fails fast.

One tradeoff worth flagging: to reproduce reliably the fixture mirrors the reported app fairly closely and pulls in `@tanstack/react-query` + `react-table` as test deps — stripped-down fixtures didn't trigger it (the provider-context-gated segment behind the boundary seems to be part of what does).

gaearon:
We'd need a more deterministic failing test and we can't have third party deps there.

errmakov:
Replaced the e2e with a deterministic unit test on the action queue — no browser, no third-party deps.

It drives `createMutableActionQueue` directly: an in-flight, non-revalidating Server Action is discarded by a navigation, and when its reducer settles while the navigation is still pending, the queue must not run the remaining queued action — that mid-navigation dispatch is what drops the destination segment. Fails without this change, passes with it. A companion case asserts the `didRevalidate` discard still runs the remaining actions, so #82674's behavior is preserved.

gaearon:
No, tests must be e2e.

errmakov:
Added a dep-free e2e in `test/e2e/app-dir/actions-discarded-segment-drop` — no third-party deps.

It reproduces the drop end-to-end: a provider's slow on-mount Server Actions are still in flight when a nested `staff → staff/[id]/schedule` navigation commits behind a `loading.tsx` boundary. Without this change the discarded action runs the rest of the queue mid-navigation and the schedule segment is dropped (the page bounces back to `/staff`); with it, the segment commits. `next start` only.

On determinism: the drop is a concurrent-commit race, so it can't be a single per-iteration assertion — I tried forcing the interleaving with server-side gates, but that removes the concurrency the bug needs (0 drops). So the test uses a CPU-throttled browser reloaded before each navigation and stops at the first drop.

gaearon:
https://github.com/vercel/next.js/pull/95391 was merged which includes a different fix to this

PR #95189, opened 2026-06-26
[test] browser-chunks test should also check `next/src/server`

mischnic:


test/production/app-dir/browser-chunks/browser-chunks.test.ts
eps1lon: Can we make that a snapshot so that it's easier to chip away?

mischnic: done
during this, I also noticed that only Turbopack is failing this test. Not Webpack

eps1lon:
Included in https://github.com/vercel/next.js/pull/95196 which already branches for Cache Components which pulls in a different set.

PR #95192, opened 2026-06-26
Split up into constants-browser.ts

mischnic:
So that we don't include all those server-only constants in the client chunk

---

> [!NOTE]
> **Medium Risk**
> Wide import churn across client navigation, rendering, and build paths could mis-route a constant if any import is wrong; the runtime throw on accidental client imports to `lib/constants` is a breaking behavior for misbundled code.
> 
> **Overview**
> Introduces **`shared/lib/constants-browser.ts`** for values that must ship in client bundles (content-type headers, nav/deployment headers, query prefixes, cache TTL, HMR websocket limits, etc.) and **strips those exports from `lib/constants.ts`**, which now throws at runtime if imported in the browser.
> 
> **Import paths** across the App Router client (segment cache, fetch, HMR websockets), shared router utilities, and assorted server/build/export code are updated to pull shared literals from `constants-browser` while server-only symbols stay on `lib/constants`.
> 
> **`is-edge-runtime`** moves from `lib/` to **`build/is-edge-runtime.ts`** with import fixes in build tooling; the server NFT allowlist is adjusted accordingly. Error **1388** documents the new client guard message.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc0213f9468504a4e4c40028e2235c8f212e845b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/next/src/build/adapter/build-complete.ts
```diff
@@ -40,7 +40,7 @@ import {
   JSON_CONTENT_TYPE_HEADER,
   NEXT_QUERY_PARAM_PREFIX,
   NEXT_RESUME_HEADER,
-} from '../../lib/constants'
+} from '../../lib/constants-browser'
```
eps1lon: this is confusing to import browser code into build. it should be moved into shared instead

PR #95194, opened 2026-06-26
feat(eslint): add pageExtensions support to no-html-link-for-pages

mmustafasenoglu:
## Description
Adds support for custom `pageExtensions` in the `@next/next/no-html-link-for-pages` ESLint rule. The rule now reads `pageExtensions` from `next.config.js` and uses those extensions when scanning for page files.

Fixes #53473

### Motivation
Currently, the rule only recognizes `.js`, `.jsx`, `.ts`, and `.tsx` files. When users configure custom `pageExtensions` in their `next.config.js`, the rule doesn't detect those pages and fails to report violations.

### Changes
- Modified `parseUrlForPages` and `parseUrlForAppDir` to accept an `extensions` parameter
- Added `getPageExtensions()` helper that reads `pageExtensions` from `next.config.js`
- Updated `getUrlFromPagesDirectories` and `getUrlFromAppDirectory` to pass extensions through
- Default extensions remain `['js', 'jsx', 'ts', 'tsx']` for backwards compatibility

### DIFF
--- a/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
+++ b/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
@@ -10,6 +10,37 @@ import {
   getUrlFromAppDirectory,
 } from '../utils/url'
 
+/**
+ * Reads pageExtensions from next.config.js
+ */
+function getPageExtensions(rootDirs: string[]): string[] {
+  const defaultExtensions = ['js', 'jsx', 'ts', 'tsx']
+  
+  for (const dir of rootDirs) {
+    const configPaths = [
+      path.join(dir, 'next.config.js'),
+      path.join(dir, 'next.config.mjs'),
+      path.join(dir, 'next.config.cjs'),
+    ]
+    
+    for (const configPath of configPaths) {
+      if (fs.existsSync(configPath)) {
+        try {
+          delete require.cache[require.resolve(configPath)]
+          const config = require(configPath)
+          if (config.pageExtensions && Array.isArray(config.pageExtensions)) {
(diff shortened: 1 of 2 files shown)


mmustafasenoglu:
Hi! Great catch from the Vercel review bot. The `extensions` parameter was indeed dead code — it was accepted by `parseUrlForPages` and `parseUrlForAppDir` but never used in the file-matching regexes.

I've updated both functions to dynamically build the regex from the `extensions` array, and also fixed `getUrlFromPagesDirectories` and `getUrlFromAppDirectory` to properly forward the `extensions` parameter to their respective parsers.

Now custom page extensions like `.mdx` or `.md` will be properly detected. Thanks!

mmustafasenoglu:
Thanks for the thorough review! 

You're right about the `extensions` parameter being dead code in the original implementation. My PR adds the forwarding of `extensions` from `getUrlFromPagesDirectories`/`getUrlFromAppDirectory` into the recursive parse calls, which was the missing piece.

Regarding the regex issue with `'\.'` collapsing to wildcard `.` - I'll update the regex patterns to use escaped dots (`\\.`) for proper literal dot matching. Will push a fix shortly.

mmustafasenoglu:
Fixed the regex escape issue! Updated all new RegExp patterns to use escaped dots for proper literal dot matching.

Changes in url.ts:
- extPattern: escaped dot before extension group
- indexPattern: escaped dot in index pattern
- pagePattern: escaped dot in page pattern
- layoutPattern: escaped dot in layout pattern

Now custom page extensions like .mdx or .md will be matched correctly.

PR #95196, opened 2026-06-26
Avoid bundling client-nodemodules used by `unstable_rethrow` in client-browser modules

eps1lon:
There are more client-node modules being bundled into `client-browser` which I'll fix in a followup (with codegen). 

Our internal `unstable-rethrow` will now have a version without suffix (unbundled or client-node) and a client-browser only version (`.browser`) that bundlers alias to for client-browser which will always be bundled.

For the specific modules being removed from client-browser, check the second commit with the actual change. The first commit fixes the `browser-chunk` test to include `next/src/server` in a snapshot.

Closes #95189


crates/next-core/src/next_import_map.rs
```diff
@@ -559,16 +563,36 @@ async fn insert_unsupported_node_internal_aliases(import_map: &mut ImportMap) ->
...
+    // regardless of which module pulls it in. Anchored at the filesystem root so it matches
+    // wherever `next` resolves from (node_modules, pnpm store, or monorepo `packages/next`).
+    let glob_mappings = vec![(
+        root.root().owned().await?,
+        Glob::new(
+            rcstr!("**/next/dist/client/components/unstable-rethrow.js"),
+            GlobOptions::default(),
+        )
+        .to_resolved()
+        .await?,
```
mischnic, line 586: Ideally we would just do `next_package.join("dist/client/components/unstable-rethrow.js")`
But that might not work well with `ResolvedMap` current glob-based API

mischnic (approved):

PR #95197, opened 2026-06-26
fix: add type check to escapePathDelimiters for non-string params

mmustafasenoglu:
## Description
Adds a type check to `escapePathDelimiters` to throw a helpful error when non-string values are passed as dynamic route params from `getStaticPaths`.

Fixes #41281

### Motivation
When `getStaticPaths` returns non-string values (e.g., numbers) for dynamic route params, the error message was unclear. Users would see a generic error instead of a message explaining that params must be strings.

### Before
```
TypeError: segment.replace is not a function
```

### After
```
TypeError: Expected a string for path segment but received number instead.
Make sure the params returned from getStaticPaths match the dynamic route.
```

### Changes
- Added type check at the beginning of `escapePathDelimiters`
- Throws `TypeError` with a descriptive message when non-string value is received


mmustafasenoglu:
Fixed! Added the missing trailing newline to comply with Prettier's `insert-final-newline` rule. Thanks for catching that!

mmustafasenoglu:
Thanks for catching the trailing newline issue! I've added the final newline to `escape-path-delimiters.ts` to satisfy Prettier's `insert-final-newline` rule.

PR #95200, opened 2026-06-26
Collect modules with browser variants statically

eps1lon:
Instead of hardcoding the list, we now automatically detect any `.browser` sibling and create a compiler alias for that module.

Together with a linting rule, we'll ensure no future regressions are handled automatically (detection of conditional require and missing .browser sibling).

This isn't the smoothest DX yet. Ideally we'd generate the list at Next.js compile time (i.e. before publishing) to avoid having to rerun a script (I'll follow-up). Generating the list during build time (next dev or next build) is just another tiny cost users would have to pay. 


crates/next-core/src/browser_variant_modules.rs
```diff
@@ -0,0 +1,14 @@
...
+//!
+//! Modules with a `.browser` sibling. The default module is aliased to its
+//! `.browser` variant for the browser bundle in
+//! `get_next_client_resolved_map` (see `next_import_map.rs`). Paths are
+//! relative to `next/dist` (extension omitted).
+
+// rustfmt::skip keeps this list one-per-line (and matching the generator
+// output) instead of collapsing short lists onto a single line.
+#[rustfmt::skip]
+pub static BROWSER_VARIANT_MODULES: &[&str] = &[
```
mischnic, line 12: we could also make this a JSON file. just as we do with serverExternalPackages
then you wouldn't need the Rust + Typescript list duplication.

mischnic (approved):

crates/next-core/src/browser_variant_modules.rs
eps1lon: Doe Rust compile this JSON away at compile time or does it stay a JSON.parse on every `next build`?

It's not actually duplication because it's auto-generated.

mischnic: It just does a runtime file read of `dist/lib/server-external-packages.jsonc`. Not inlined at compile time

eps1lon: Yeah than I'd rather keep two files since they're just auto-generated.

PR #95201, opened 2026-06-26
Split typeof-window server requires into .browser variants

eps1lon:
First batch of fixes that addresses most of the recent bundle size regressions. 

### DIFF
--- a/packages/next/src/client/components/server-async-storage.browser.ts
+++ b/packages/next/src/client/components/server-async-storage.browser.ts
@@ -1,0 +1,7 @@
+// Browser variant of `./server-async-storage`. The server-only async-storage
+// singletons don't exist in the browser, so these are `undefined`. Consumers
+// guard usage behind `typeof window === 'undefined'`, so the stubs are never
+// dereferenced.
+export const actionAsyncStorage = undefined
+export const workAsyncStorage = undefined
+export const workUnitAsyncStorage = undefined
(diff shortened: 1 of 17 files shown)


test/e2e/app-dir/cache-components-errors/cache-components-client-hook-abort-reasons.test.ts
```diff
@@ -315,7 +315,7 @@ describe('Cache Components Errors - Client Hook Abort Reasons', () => {
                      https://nextjs.org/docs/messages/blocking-prerender-client-hook#wrap-in-or-move-into-suspense
                    - [block] Set \`export const instant = false\` to silence this warning and allow a blocking route
                      https://nextjs.org/docs/messages/blocking-prerender-client-hook#allow-blocking-route
-                     at useDynamicSearchParams (webpack:///<next-src>)
+                     at Object.useDynamicSearchParams (webpack:///<next-src>)
```
mischnic, line 318: Is this change acceptable?

packages/next/src/client/components/server-async-storage.browser.ts
```diff
@@ -0,0 +1,7 @@
+// Browser variant of `./server-async-storage`. The server-only async-storage
+// singletons don't exist in the browser, so these are `undefined`. Consumers
+// guard usage behind `typeof window === 'undefined'`, so the stubs are never
+// dereferenced.
+export const actionAsyncStorage = undefined
+export const workAsyncStorage = undefined
+export const workUnitAsyncStorage = undefined
```
mischnic, line 7: Pipedream: ideally we'd have a lint that ensure that a `.browser.ts` file has the same list of exports as the non-browser sibling file. Otherwise it could happen that you forget to update one of them when changing/adding/removing an export

mischnic (approved):

test/e2e/app-dir/cache-components-errors/cache-components-client-hook-abort-reasons.test.ts
eps1lon: Yeah. That frame should be ignore-listed anyway.

PR #95207, opened 2026-06-26
Fix: Update the URL when a client navigation redirects to a rewritten route

### LINKED ISSUE #95195: Next.js 16.3 Preview: Middleware redirect not reflected in URL on client-side navigation when target is rewritten to a dynamic route

https://github.com/amannn/nextjs-16-3-bug-repro-redirect

### To Reproduce

1. `npm install && npm run build && npm start`
2. Open http://localhost:3000/two
3. Click "Go to /a"
4. Observe the address bar — it shows `/a` instead of `/`

To run the automated reproduction:
```
npx playwright install chromium
npm test
```
The test `client-side navigation to /a should redirect to /` fails.

The proxy (middleware) is:
- `/a` → redirect to `/`
- `/` → rewrite to `/a` (a dynamic, `force-dynamic` page that reads `params`)

Expected: clicking <Link href="/a"> triggers the middleware redirect (/a → /), and
the browser URL updates to /.

Actual (16.3): the URL stays at /a. The redirected content renders correctly
(slug: a is shown), but window.location.href remains /a instead of /.

Hard navigation (typing the URL, reload, curl) still redirects correctly — the
regression is specific to client-side navigation via <Link> or router.push().

All three of the following are required to trigger the bug:
1. A client-side navigation (hard navigation redirects correctly)
2. The middleware redirect target is itself rewritten to another route
3. The rewrite target is dynamically rendered and reads route params

This matches the dynamic-route shell changes in 16.3 ("Instant Navigations"):
the router appears to commit the destination URL (/a) to the address bar
optimistically before the RSC response resolves, and does not correct it after
the server follows the redirect.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 24576
  Available CPU cores: 18
Binaries:
  Node: 24.14.0
  npm: 11.14.1
  Yarn: 1.22.22
  pnpm: 10.32.1
Relevant Packages:
  next: 16.3.0-preview.5 // Latest available version is detected (16.3.0-preview.5).
  eslint-config-next: N/A
  react: 19.2.3
  react-dom: 19.2.3
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Redirects, Middleware, Linking and Navigating, Dynamic Routes

### Additional context

Regression between 16.2.9 (working) and 16.3.0-preview.5 (broken).
To verify the fix, `npm install next@16.2.9` and run `npm test` — both tests pass.

acdlite:
On a client-side navigation, when a proxy/middleware redirected the requested URL to a destination that was itself rewritten to a dynamically rendered route, the page content updated correctly but the address bar stayed on the original URL. Hard navigations (reload, typing the URL) were unaffected.

This happened because the navigation optimistically committed the predicted URL up front, and when the redirect resolved to a route with the same shape as the prediction, the router treated the navigation as already correct and never reconciled the address bar.

A redirect discovered during a navigation now invalidates the prediction and re-resolves the route, so the address bar reflects the redirect destination.

Adds an end-to-end test covering the regression.

Fixes #95195

### DIFF
--- a/test/e2e/app-dir/redirect-rewrite-dynamic/app/[slug]/page.tsx
+++ b/test/e2e/app-dir/redirect-rewrite-dynamic/app/[slug]/page.tsx
@@ -1,0 +1,53 @@
+import { connection } from 'next/server'
+import { Suspense } from 'react'
+import { LinkAccordion } from '../../components/link-accordion'
+
+async function Content({ params }: { params: Promise<{ slug: string }> }) {
+  // Opt into dynamic rendering and read the `[slug]` param.
+  await connection()
+  const { slug } = await params
+
+  // slug === "a" means we're at / (proxy rewrites / -> /a).
+  const isHome = slug === 'a'
+
+  return (
+    <main>
+      {/* Build the string in JS-land so the literal "slug: a" appears in the
+          Flight response body (see router-act skill: JSX interpolation
+          splits the string in the wire format). */}
+      <h1 id="page" data-testid="page">{`slug: ${slug}`}</h1>
+
+      {isHome ? (
+        <p>
+          This is the home page (<code>/</code>), served via a proxy rewrite
+          from <code>/a</code>.
+          <LinkAccordion href="/two" prefetch={false}>
+            Go to /two
+          </LinkAccordion>
+        </p>
+      ) : (
+        <p>
+          Click the link. The proxy redirects <code>/a</code> to <code>/</code>,
+          so the URL should change to <code>/</code>.
+          <LinkAccordion href="/a" prefetch={false}>
+            Go to /a
+          </LinkAccordion>
+        </p>
+      )}
+    </main>
+  )
+}
+
+export default function Page({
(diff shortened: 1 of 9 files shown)


test/e2e/app-dir/redirect-rewrite-dynamic/app/[slug]/page.tsx
```diff
@@ -0,0 +1,42 @@
+import { LinkAccordion } from '../../components/link-accordion'
+
+// Dynamically rendered and reads the `[slug]` param.
+export const dynamic = 'force-dynamic'
```
Comment (eps1lon) [inline test/e2e/app-dir/redirect-rewrite-dynamic/app/[slug]/page.tsx]: Claude did the same for me. Does that repro with `await connection()` as well so that we can test it with and without Cache Components enabled?

Comment (acdlite) [reply test/e2e/app-dir/redirect-rewrite-dynamic/app/[slug]/page.tsx]: ah oops, missed this. Good catch.

eps1lon (approved):

PR #95213, opened 2026-06-26
[turbopack] Don't evict when there is little memory to save

lukesandberg:
### What?

Adds a new `'auto'` mode to the experimental `turbopackMemoryEviction` config option and makes it the default. In `'auto'` mode, Turbopack only evicts in-memory cache after a snapshot once enough memory has been allocated since the last eviction to make the work worthwhile.

The option now accepts three values:

- `false`: never evict.
- `'auto'` (new default): evict after a snapshot only once a memory threshold has been crossed.
- `'full'`: evict all evictable data after every snapshot (the previous behavior).

### Why?

When the persistent (FileSystem) cache is enabled, Turbopack snapshots its in-memory state to disk and can then evict those in-memory copies to reclaim memory, reloading them from disk on demand.

Previously, with eviction enabled (`'full'`), we evicted after *every* snapshot. This is too aggressive: tasks get restored from disk and then immediately evicted again, cycle after cycle, wasting work for little memory benefit.

`'auto'` mirrors the existing persistence-threshold model (we already skip a snapshot when too little compilation time has accumulated to justify its cost). Here the proxy is memory instead of time: it isn't worth paying the restore-then-re-evict churn to reclaim a small amount of memory.

This was motivated by an example in v0.app where the client would poll the server every 5 seconds leading to a pathological behavior
* client poll ->
* next.js ensurePage ->
* turbopack writeEndpointToDisk
*  recompute and restore settings for that endpoint (3-10ms of io work)
* realize there is nothing to do
* respond to client
* 2 seconds later..... persist and evict everything saving 0.5M of ram (100ms of work! writes out 2 SSTs and then compacts them!)
* 3 seconds after that restart the loop

This is silly, #95137 will prevent the persistence loop from occurring , this PR will also just skip the 'restore and recompute' work using a similar strategy.

### How?

We can't measure exactly how much memory a sweep would reclaim, so we use `TurboMalloc::memory_usage()` (process-global net live bytes) as a proxy. In `'auto'` mode an eviction sweep runs only once the net bytes allocated since the last eviction exceed a threshold (default 128 MiB, overridable via `TURBO_ENGINE_EVICT_MIN_BYTES`). The first eviction after startup always runs. The threshold scales down under OS memory pressure (`TurboMalloc::memory_pressure()`) so we evict more eagerly when memory is tight.

**Backend (`turbo-tasks-backend`):**
- `BackendOptions.evict_after_snapshot: bool` → `eviction_mode: EvictionMode` (`Off` / `Full` / `Auto`).
- New `EvictionControl` type owns the policy: the mode plus the threshold bookkeeping. The background snapshot loop calls `should_evict(snapshot_had_new_data)` once per cycle and `record_eviction()` after a sweep, so the loop no longer branches on the mode. `'full'` and `false` behave exactly as before.

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackMemoryEviction.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackMemoryEviction.mdx
@@ -6,12 +6,13 @@ description: Learn how to control Turbopack's memory eviction strategy for the p
 
 ## Usage
 
-`turbopackMemoryEviction` controls how whether Turbopack reclaims memory while the persistent (FileSystem) cache is enabled. After Turbopack writes a snapshot of its cache to disk, it can 'evict' the in-memory copies of that data and reload them from disk on demand.
+`turbopackMemoryEviction` controls whether Turbopack reclaims memory while the persistent (FileSystem) cache is enabled. After Turbopack writes a snapshot of its cache to disk, it can 'evict' the in-memory copies of that data and reload them from disk on demand.
 
-Currently there are two options
+Currently there are three options
 
 - `false`: never evict. Cached data stays in memory for the lifetime of the process.
-- `'full'` (default): after every snapshot, evict all possible data from memory. They are reloaded from disk on demand.
+- `'auto'` (default): evict after a snapshot only once enough memory has been allocated since the last eviction to make it worthwhile. Leverages thresholds and memory pressure feedback from the operating system.
+- `'full'`: evict all possible data from memory every time we save to disk.
 
 > **Good to know:** This option only has an effect in `next dev` sessions when the [FileSystem Cache](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is enabled, since eviction relies on data already being persisted to disk. It is experimental and under active development.
 
@@ -20,8 +21,7 @@ import type { NextConfig } from 'next'
 
 const nextConfig: NextConfig = {
   experimental: {
-    // Evict in-memory cache data after each snapshot to reduce memory usage
-    turbopackMemoryEviction: 'full',
+    turbopackMemoryEviction: 'auto',
   },
 }
 
@@ -32,8 +32,7 @@ export default nextConfig
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   experimental: {
-    // Evict in-memory cache data after each snapshot to reduce memory usage
-    turbopackMemoryEviction: 'full',
+    turbopackMemoryEviction: 'auto',
   },
 }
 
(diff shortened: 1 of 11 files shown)


lukesandberg:
* **#95213** <a href="https://app.graphite.com/github/pr/vercel/next.js/95213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95213?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#95137** <a href="https://app.graphite.com/github/pr/vercel/next.js/95137?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackMemoryEviction.mdx
```diff
@@ -6,12 +6,13 @@ description: Learn how to control Turbopack's memory eviction strategy for the p
 
 ## Usage
 
-`turbopackMemoryEviction` controls how whether Turbopack reclaims memory while the persistent (FileSystem) cache is enabled. After Turbopack writes a snapshot of its cache to disk, it can 'evict' the in-memory copies of that data and reload them from disk on demand.
+`turbopackMemoryEviction` controls whether Turbopack reclaims memory while the persistent (FileSystem) cache is enabled. After Turbopack writes a snapshot of its cache to disk, it can 'evict' the in-memory copies of that data and reload them from disk on demand.
 
-Currently there are two options
+Currently there are three options
 
 - `false`: never evict. Cached data stays in memory for the lifetime of the process.
```
wbinnssmith, line 13: I know this is pre-existing, but any reason we didn't do `'off'` here to avoid mixing types here?

packages/next/src/server/config.ts
```diff
@@ -424,18 +424,22 @@ function assignDefaultsAndValidate(
...
-    const rawEnv = process.env.TURBO_ENGINE_EVICT_AFTER_SNAPSHOT
-    turbopackMemoryEvictionMode =
-      rawEnv == null || rawEnv === '1' || rawEnv === 'true' ? 'full' : 'off'
+  // 'auto' | undefined`) into the `turbopackMemoryEvictionMode` enum expected by
+  // napi (`'off' | 'full' | 'auto'`).
+  let turbopackMemoryEvictionMode: 'off' | 'full' | 'auto'
+
+  switch (result.experimental.turbopackMemoryEviction) {
+    case false:
+      turbopackMemoryEvictionMode = 'off'
```
wbinnssmith, line 434: 😭

wbinnssmith (approved):

docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackMemoryEviction.mdx
lukesandberg: https://vercel.slack.com/archives/C09R44U5HQW/p1780513578996979

the argument as i understand it is just that `turbopackMemoryEviction: false` is an intuitive way to disable it, but `'off'` would be less so

PR #95224, opened 2026-06-27
Add early OpenTelemetry tracer prerender coverage

devjiwonchoi:
### Why?

Current behavior: https://github.com/vercel/next.js/actions/runs/28284411969/job/83806126742?pr=95224#step:31:8803

```
Next.js encountered the unstable value `Math.random()` while prerendering
```

A tracer can be acquired before OpenTelemetry provider registration and then used while Cache Components prerenders a route at runtime. Because early proxy tracers resolve their registered tracer through `getDelegateTracer()`, they can bypass the existing `getTracer()` patch that exits `workUnitAsyncStorage` during span creation. Span ID generation should remain allowed and must not surface as dynamic `Math.random()` access.

This draft adds regression coverage and patches that early tracer resolution path.

### How?

- Capture a tracer before `NodeSDK.start()` and retain it on a typed `globalThis` field.
- Render a manual span on a dynamic route whose first unknown parameter is prerendered at runtime.
- Patch `getDelegateTracer()` for proxy tracer providers to apply the same tracer instrumentation used by `getTracer()`.
- Assert that the first request has no `Math.random()` prerender error and that repeated runtime prerenders create non-zero span IDs.

### DIFF
--- a/packages/next/src/server/lib/router-utils/instrumentation-node-extensions.ts
+++ b/packages/next/src/server/lib/router-utils/instrumentation-node-extensions.ts
@@ -3,7 +3,11 @@
  * This extension must only be loaded in Node environments.
  */
 
-import type { Tracer } from '@opentelemetry/api'
+import type {
+  ProxyTracerProvider,
+  Tracer,
+  TracerProvider,
+} from '@opentelemetry/api'
 import {
   type WorkUnitStore,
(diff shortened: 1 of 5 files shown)


devjiwonchoi:
Superseded by https://github.com/vercel/next.js/pull/95317

PR #95233, opened 2026-06-27
More granular cache keys for use-cache entries

mischnic:
Caveats:
- I had to skip `react[-dom][/*]` and `private-next-rsc-server-reference` and `private-next-rsc-cache-wrapper` imports in the code hash and env-var tracking. Because those all end up pulling in app-page-turbo.runtime.prod.js which reads many env vars and would cause constant deopting. But this should still be correct. The code of these imports is included via the Next.js version, and no env vars should change the semantics of any of those imports.
- Static env var reads are collected, but too dynamic accesses are silently ignored and don't lead to deopts (same goes for env var reads in native NAPI addons). This means that this cache reuse is not guaranteed to be 100% guaranteed to never lead to stale caches.
- Code hashing and env var collection happens on a per-module basis. So if you put all use-cache functions into a single file and/or together with react components, then you will see extraneous invalidations. This will be fixed by either generally enabling module splitting for all of Turbopack, or by adding a special transform that does it for use-cache functions.

Followups:
- There are some env var static analysis gaps that will be immediate followups before broader testing. These are the various TODOs added in https://github.com/vercel/next.js/pull/95310
- Client components invalidation is very coarse grained right now (statically imports any client component anywhere -> deopt completely). Followup for the future. But the current setup is correct. This would just improve effectiveness further
- Do this in dev as well. Currently there is no NFT at all in dev (for performance reasons)
- Module splitting for more granular tracking
- Include entropy when serializing server reference arguments for cache key

Todo:

- [x] Use implementation code hash (includes inlined env vars): from #94234
- [x] Include non-inlined runtime env vars: from #95310
- [x] Include client reference manifest (very coarse for now)
- [x] Include Next.js version (for wire format, etc)
- [x] This is now done for all use-cache entries now. Not just for `use cache: remote`. Is that the intended behavior? Yes
- [ ] ~~if `NEXT_DEPLOYMENT_ID` is in the env vars. just deopt and don't care about stringifing and hashing the env vars~~
- [x] Is cache key size a problem? Currently you can get this: (values are always hashed)
`CustomCacheHandler::get ["80e6f6560092f0078775e7e787c1c10ecf6dea0bc4",[],["d984fbaa996274737f3b59345a300a20","16.4.0-canary.5","__NEXT_NO_MIDDLEWARE_URL_NORMALIZE=undefined","NEXT_OTEL_PERFORMANCE_PREFIX=undefined","__NEXT_PRIVATE_ORIGIN=a04f4b9d6a8f42724740a480e9a2bc67c053dc04377be831c0e2c407a1422004","NEXT_PRIVATE_RESPONSE_CACHE_TTL=undefined","NEXT_PRIVATE_RESPONSE_CACHE_MAX_SIZE=undefined","__NEXT_CACHE_COMPONENTS=b5bea41b6c623f7c09f1bf24dcae58ebab3c0cdd90ad966bc43a45b44867e12b","__NEXT_ROUTER_BASEPATH=undefined","__NEXT_MANUAL_CLIENT_BASE_PATH=undefined","__NEXT_INSTRUMENTATION_CLIENT_ROUTER_TRANSITION_EVENTS=undefined","__NEXT_APP_NAV_FAIL_HANDLING=undefined","__NEXT_GESTURE_TRANSITION=undefined","__NEXT_USE_OFFLINE=undefined","NEXT_DEBUG_BUILD=undefined","__NEXT_VERBOSE_LOGGING=undefined...]] [["_N_T_/layout","_N_T_/page","_N_T_/","_N_T_/index"]]`

### DIFF
--- a/crates/next-api/src/server_actions.rs
+++ b/crates/next-api/src/server_actions.rs
@@ -5,7 +5,8 @@ use bincode::{Decode, Encode};
 use next_core::{
     next_client_reference::{CssClientReferenceModule, EcmascriptClientReferenceModule},
     next_manifests::{
-        ActionLayer, ActionManifestModuleId, ActionManifestWorkerEntry, ServerReferenceManifest,
+        ActionLayer, ActionManifestModuleId, ActionManifestWorkerEntry,
+        ActionManifestWorkerEntryDurability, ServerReferenceManifest,
     },
     util::NextRuntime,
 };
@@ -315,12 +316,11 @@ impl Asset for ServerActionManifestAsset {
                     is_async: async_module_info
                         .is_async(self.chunk_item.module().to_resolved().await?)
                         .await?,
-                    code_hash: data.as_ref().map(|d| d.ident_code_hash.as_str()),
-                    runtime_env_vars: data.as_ref().map(|d| d.runtime_env_vars.as_slice()),
-                    references_client_component: data
-                        .as_ref()
-                        .map(|d| d.references_client_component)
-                        .unwrap_or(false),
+                    durability: data.as_ref().map(|d| ActionManifestWorkerEntryDurability {
+                        code_hash: d.ident_code_hash.as_str(),
+                        runtime_env_vars: d.runtime_env_vars.as_slice(),
+                        references_client_component: d.references_client_component,
+                    }),
                 },
             );
 
--- a/crates/next-api/src/server_actions.rs
+++ b/crates/next-api/src/server_actions.rs
@@ -3,6 +3,7 @@ use std::{borrow::Cow, collections::BTreeMap, io::Write, sync::LazyLock};
 use anyhow::{Context, Result, bail};
 use bincode::{Decode, Encode};
 use next_core::{
+    get_next_package,
     next_client_reference::{CssClientReferenceModule, EcmascriptClientReferenceModule},
     next_manifests::{
         ActionLayer, ActionManifestModuleId, ActionManifestWorkerEntry,
@@ -10,7 +11,6 @@ use next_core::{
     },
     util::NextRuntime,
 };
-use rustc_hash::FxHashSet;
 use swc_core::{
     atoms::{Atom, atom},
     common::comments::Comments,
@@ -38,12 +38,11 @@ use turbopack_core::{
     context::AssetContext,
     file_source::FileSource,
     ident::AssetIdent,
-    module::Module,
+    module::{Module, Modules},
     module_graph::{
         GraphTraversalAction, ModuleGraph, ModuleGraphLayer, async_module_info::AsyncModulesInfo,
     },
     output::{OutputAsset, OutputAssetsReference},
-    reference::ModuleReference,
     reference_type::{EcmaScriptModulesReferenceSubType, ReferenceType},
     resolve::ModulePart,
     virtual_source::VirtualSource,
@@ -53,7 +52,6 @@ use turbopack_ecmascript::{
     chunk::{EcmascriptChunkItem, EcmascriptChunkItemExt, EcmascriptChunkPlaceable},
     module_fragments::part::module::EcmascriptModulePartAsset,
     parse::ParseResult,
-    references::esm::EsmAssetReference,
 };
 
 use crate::project::Project;
@@ -251,6 +249,42 @@ impl Asset for ServerActionManifestAsset {
             NextRuntime::NodeJs => &mut manifest.node,
         };
 
+        // These modules end up pulling in a lot of env vars and would always cause invalidations.
+        // But they don't actually read the env vars for the imports that are used. Their code is
+        // versioned anyway via the Next.js version in the cache key.
+        //
+        // `module.compiled.js -> app-page-turbo.runtime.prod.js` is imported by the following
+        // modules. But none of them have values whose runtime value depends on the runtime env
+        // vars.
+        // - react
+        // - react-dom
+        // - react-server-dom-*
+        // - next/dist/shared/lib/app-router-context.shared-runtime
+        // - next/dist/shared/lib/head-manager-context.shared-runtime
+        // - next/dist/shared/lib/hooks-client-context.shared-runtime
+        // - next/dist/shared/lib/image-config-context.shared-runtime
+        // - next/dist/shared/lib/router-context.shared-runtime
+        // - next/dist/shared/lib/server-inserted-html.shared-runtime
+        let app_project = self.project.app_project().await?.unwrap();
+        let next_dir = get_next_package(self.project.project_path().owned().await?).await?;
+        let source_to_ignore = FileSource::new(
+            next_dir.join("dist/server/route-modules/app-page/module.compiled.js")?,
+        );
+        let modules_to_ignore = Vc::cell(
+            [
+                app_project.rsc_module_context(),
+                app_project.route_module_context(),
+            ]
+            .iter()
+            .map(|c| {
+                c.process(Vc::upcast(source_to_ignore), ReferenceType::Undefined)
+                    .module()
+                    .to_resolved()
+            })
+            .try_join()
+            .await?,
+        );
+
         struct ActionMetadata<'a> {
             exported_name: &'a str,
             filename: Cow<'a, str>,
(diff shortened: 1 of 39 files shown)


mischnic:
* **#95233** <a href="https://app.graphite.com/github/pr/vercel/next.js/95233?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95233?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#97926** <a href="https://app.graphite.com/github/pr/vercel/next.js/97926?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#97921** <a href="https://app.graphite.com/github/pr/vercel/next.js/97921?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95310** <a href="https://app.graphite.com/github/pr/vercel/next.js/95310?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/app-render/manifests-singleton.ts
```diff
@@ -280,9 +291,23 @@ function createServerModuleMap(): ServerModuleMap {
...
+      } = workerEntry
+
+      return {
+        id: moduleId,
+        name: id,
+        chunks: [],
+        async,
+        codeHash,
+        runtimeEnvVars,
+        referencesClientComponent,
```
unstubbable: Hmm, should we really expose these to React as well? Maybe using `getServerActionsManifest()` directly in the cache wrapper is the better approach.

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -3616,6 +3609,85 @@ export async function cache(
...
+ * In case that granular information isn't available, fall back to
+ * buildId/deploymentId/hmrRefreshHash, which is a correct hash but over-invalidates way too often.
+ */
+async function computeCacheKeyImplementationPart(
+  workStore: WorkStore,
+  workUnitStore: WorkUnitStore,
+  id: string
+): Promise<unknown> {
+  async function hashString(input: string): Promise<string> {
+    if (process.env.NEXT_RUNTIME === 'edge') {
```
unstubbable: `'use cache'` is not supported in the Edge runtime, we can omit this branch. (I just noticed that there's also a preexisting `isEdgeRuntime` condition that we don't need.)

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -153,9 +153,13 @@ interface PublicCacheContext {
...
 
-export type CacheKeyParts =
-  | [buildId: string, id: string, args: unknown[]]
-  | [buildId: string, id: string, args: unknown[], hmrRefreshHash: string]
+const nextVersion = process.env.__NEXT_VERSION as string
+
+export type CacheKeyParts = [
+  id: string,
+  args: unknown[],
+  implementationHash: unknown,
```
unstubbable, line 163: Is there a good reason to use `unknown` here?

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -3616,6 +3609,85 @@ export async function cache(
...
+/**
+ * This returns a cache key that has to cover everything that can affect the result of the cached
+ * function (apart from the arguments). So
+ * - codeHash: the code itself that generates the return value
+ *    - Notably, this excludes the following modules.  Those are included via the Next.js version
+ *      anyway:
+ *    - react, react-dom, private-next-rsc-server-reference, private-next-rsc-cache-wrapper
+ * - runtimeEnvVars: the keys and values runtime environment variables that the code reads (and are
+ *   not inlined)
+ * - the version of Next.js (to account for RSC write format changes, or use-cache-wrapper.ts
```
```suggestion
 * - the version of Next.js (to account for RSC wire format changes, or use-cache-wrapper.ts
```

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -3616,6 +3609,85 @@ export async function cache(
...
+  }
+
+  let serverModuleMapEntry = workStore.durableUseCacheEntries
+    ? getServerModuleMap()?.[id]
+    : undefined
+  if (
+    typeof serverModuleMapEntry?.codeHash === 'string' &&
+    serverModuleMapEntry?.runtimeEnvVars &&
+    // TODO replace this with more granular tracking: a list of all client components imported
+    serverModuleMapEntry?.referencesClientComponent !== true
```
unstubbable: Looks like we can only have all three or none of them? Should we group them into one object then?

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -3616,6 +3609,85 @@ export async function cache(
...
+  ) {
+    // Hash the env var values, to not leak secrets into the cache key.
+    let runtimeEnvVarStateHash = await hashString(
```
unstubbable: Why isn't this stored as a hash in the manifest already, same as the code hash? We wouldn't want to waste the compute time on each cache function invocation.

unstubbable (commented):
Can we also run `test/development/app-dir/use-cache-dev/use-cache-dev.test.ts` with the feature flag disabled (current default) as well as enabled?

packages/next/src/server/use-cache/use-cache-wrapper.ts
mischnic: This is an opaque cache key 🤷 
We can change it if you want

packages/next/src/server/app-render/manifests-singleton.ts
mischnic: Fixed

packages/next/src/server/use-cache/use-cache-wrapper.ts
mischnic: Fixed

packages/next/src/server/use-cache/use-cache-wrapper.ts
mischnic: When we write the manifest (at build time), we don't know the runtime env var values.
Which is why this hashing happens at runtime.

unstubbable: I see.

mischnic:
> Can we also run test/development/app-dir/use-cache-dev/use-cache-dev.test.ts with the feature flag disabled (current default) as well as enabled?

I can do that if you want. But currently, none of this runs in dev at all. So it's not necessary yet. But eventually, yes.

packages/next/src/server/use-cache/use-cache-wrapper.ts
mischnic: Done

unstubbable (approved):
Didn't look at the Rust changes, but the rest looks good to me.

crates/next-api/src/server_actions.rs
```diff
@@ -372,20 +394,67 @@ async fn compute_subtree_content_hash(
...
+            .references()
+            .await?
+            .into_iter()
+            .filter_map(ResolvedVc::try_downcast_type::<EsmAssetReference>)
+            .map(async |reference| {
+                let reference_value = reference.await?;
+                Ok(
+                    // These pull in app-page-turbo.runtime.prod.js which reads basically all env
+                    // vars that exist and would always cause a deopt.
+                    if reference_value.request == "private-next-rsc-server-reference"
```
lukesandberg: it seems odd to filter based on the request instead of the resolved target 

could a configured alias confuse this logic?

resolving all the modules would be expensive i guess , but it is perhaps better?  query the module graph for identifiers?

lukesandberg (approved):

crates/next-api/src/server_actions.rs
mischnic: Yeah, that was roughly my thought process when writing this code. And I didn't want to await everything 🤷‍♂️ 

Actually, this is still broken, because the "only walk the references of the entrypoint" bit might work for `private-next-rsc-server-reference` (which will always be there), but not necessarily for react which might only be discovered further down the traversal.

mischnic: I have changed it now to synthesize the problematic module. Then we can easily skip that way: https://github.com/vercel/next.js/pull/95233/commits/b11cd41be09ceb37ad9d8a006dcd1d38e6aaf2f8

PR #95238, opened 2026-06-27
fix(cache-components): decompress postponed resume body before parsing

### LINKED ISSUE #95214: PPR resume reads the request body as UTF-8 without decompressing `Content-Encoding: gzip` → "Invariant: invalid postponed state"

https://github.com/arjenblokzijl/next-ppr-gzip-resume-repro

### To Reproduce

1. `npm install`
2. `npm run repro:unit` — deterministic, server-free proof. It calls the installed
   `parsePostponedState` exactly as `base-server.js` does on a resume
   (`body.toString("utf8")`), passing a **gzip-compressed** postponed state. Output:

   gzip header bytes: 1f 8b 08 00   (matches what we see in production)
   Failed to parse postponed state Error: Invariant: invalid postponed state <gzip garbage>
   -> parsePostponedState degraded to type:1 (logged error, no crash; HTTP 200)
3. `npm run build` — the single route reports as `◐ (Partial Prerender)`.

(The end-to-end path that *calls* the parser is driven by the platform's
internal resume routing — minimal mode + the prerender manifest — so a bare
`next start` doesn't exercise it; the unit repro isolates the failure
deterministically. See the repo README for the full code-path trace.)

On a PPR (Cache Components) resume, the postponed state is read from the POST
request body and decoded with `body.toString("utf8")` **without honoring
`Content-Encoding`**. When the resume body arrives **gzip-compressed** (as it
does behind Vercel's infrastructure), the gzip bytes decoded as UTF-8 are not a
valid postponed-state string, so `parsePostponedState` fails its `^<digits>:`
check, logs `Failed to parse postponed state` + the Invariant, and **degrades to
`type:1`** — a logged server error with an HTTP 200 fallback (the route cannot
resume its prerendered HTML). In production this fires steadily on every resumed
`◐` route (cart/checkout/category pages, etc.).

**Expected:** the resume body is decompressed according to its `Content-Encoding`
(gzip/br/deflate) before `toString("utf8")`/`parsePostponedState`.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 18432
  Available CPU cores: 12
Binaries:
  Node: 24.17.0
  npm: 11.13.0
  Yarn: N/A
  pnpm: 9.5.0
Relevant Packages:
  next: 16.3.0-preview.5 // Latest available version is detected (16.3.0-preview.5).
  eslint-config-next: N/A
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

cacheComponents

Vercel (Deployed)

### Root cause (code references)

- `server/base-server.js` — resume-body read:
  
```js
  if (this.isAppPPREnabled && this.minimalMode &&
   req.headers[NEXT_RESUME_HEADER] === '1' && req.method === 'POST') {
 const body = await readBodyWithSizeLimit(req.body, maxPostponedStateSizeBytes);
 const postponed = body.toString('utf8');   // ← no Content-Encoding decompression
 addRequestMeta(req, 'postponed', postponed);
  }
```

- server/lib/postponed-request-body.js — readBodyWithSizeLimit just concatenates
raw chunks; no decompression.
- server/app-render/postponed-state.js — parsePostponedState requires the state
to start with ^([0-9]*):, throws Invariant: invalid postponed state (E314),
and its own catch logs Failed to parse postponed state and returns a degraded
type:1.

The garbled state in the error starts with the gzip magic header 1f 8b 08.

- Not a regression — the read path (body.toString('utf8'), no decompression)
is byte-for-byte identical in 16.2.x and 16.3, so it reproduces on both. The
repro pins 16.3.0-preview.5 for convenience.
- Workaround: make affected routes fully dynamic (ƒ) so there's no postponed
state to resume (e.g. await connection() at the page root — 16.3 only). This
trades away the static shell, so it's only viable for session-specific routes,
not cacheable catalog/content pages.

lazypool:
### What?

In the PPR (Cache Components) resume flow, `base-server.js` reads the POST request body and decodes it using `body.toString('utf8')` **without respecting the `Content-Encoding` header**. 

When the resume body arrives **gzip-compressed** (which happens behind Vercel's infrastructure), the gzip binary bytes (starting with `1f 8b 08`) are misinterpreted as a UTF-8 string. This causes `parsePostponedState` to fail the `^([0-9]*):` format check, throwing an `Invariant: invalid postponed state` error (E314). The error is caught, degraded to `type:1`, and results in a **logged server error with an HTTP 200 fallback**—meaning the route cannot resume its prerendered HTML shell.

**Note:** This is **not a 16.3 regression**. The resume-body read logic in `base-server.js` is byte-for-byte identical in 16.2.x and 16.3.

---

### Why?

In `next/dist/server/base-server.js`, the resume branch reads the body without decompression:

```js
if (this.isAppPPREnabled && this.minimalMode &&
    req.headers[NEXT_RESUME_HEADER] === '1' && req.method === 'POST') {
  const body = await readBodyWithSizeLimit(req.body, maxPostponedStateSizeBytes);
  const postponed = body.toString('utf8');   // ← No Content-Encoding handling
  addRequestMeta(req, 'postponed', postponed);
}
```

`readBodyWithSizeLimit` (located in `next/dist/server/lib/postponed-request-body.js`) merely concatenates raw chunks—it does not perform any decompression. `parsePostponedState` requires the string to start with `^([0-9]*):`, otherwise it throws E314. Since the gzipped binary data decoded as UTF-8 produces garbage (starting with `1f 8b 08`), the parser fails.

---

### How?

Added a `decompressBody()` utility in `postponed-request-body.ts` that handles decompression **after reading the body** and **before calling `toString('utf8')`**:

1. **Explicit `Content-Encoding`** — When the header is present, `decompressBody` dispatches to the matching `zlib` sync method (`inflateSync` for `deflate`, `gunzipSync` for `gzip`, `brotliDecompressSync` for `br`).
2. **Magic header detection (no `Content-Encoding`)** — The PPR resume chain contract does not carry `Content-Encoding`, but Vercel's infrastructure may gzip the body without setting the header. `decompressBody` auto-detects the gzip magic number (`0x1f 0x8b`) when the header is absent. This is safe because a valid postponed state always starts with a length prefix (ASCII digits `0x30`–`0x3a`), so there is no overlap.
3. **Zip bomb protection** — Each decompression call enforces a `maxOutputLength` limit (5× the max compressed body size, or 500 MB) via `zlib`'s `maxOutputLength` option, matching the pattern already used in `resume-data-cache.ts`.
4. **Coverage** — Applied in both `base-server.ts` (router-level resume) and `app-page.ts` (template-level resume, including the server-action + postponed state path).

**`decompressBody` logic:**

```
Content-Encoding present
  ├── deflate → inflateSync(maxOutputLength)    — Next.js internal practice
  ├── gzip    → gunzipSync(maxOutputLength)     — Vercel proxy explicit
  ├── br      → brotliDecompressSync(maxOutputLength)
  └── other   → passthrough

Content-Encoding absent
  ├── gzip magic (0x1f 0x8b) → gunzipSync(maxOutputLength)
  └── no magic → passthrough
```

**E2e test** — Added `test/e2e/app-dir/resume-content-encoding/resume-content-encoding.test.ts` with 5 cases:
- Uncompressed body (baseline)
- gzip body with explicit `Content-Encoding: gzip`
- gzip body without `Content-Encoding` (magic header detection)
- deflate body with explicit `Content-Encoding: deflate`
- brotli body with explicit `Content-Encoding: br`

---

### Closes

Fixes #95214


FurryR (commented):
Looks good 👍

The pull request description didn't match the implementation but the fix worked properly.

eps1lon:
Please add an e2e test using Next.js public APIs that shows the defect. Our test suites run also deployed on Vercel so we should see the failure.

eps1lon:
Also keep in mind that Next.js is currently controlling the compression with `inflateSync`. I don't think it makes sense to support different compression algorithms unless we fix both sides.

packages/next/src/build/templates/app-page.ts
```diff
@@ -357,25 +358,30 @@ export async function handler(
         res.statusCode = 413
         res.end(
           `Request body exceeded limit. ` +
-            `To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit`
+            `To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/config/next-config-js/serverActions#bodysizelimit`
```
FurryR:
```suggestion
            `To configure the body size limit for Server Actions, see: https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#bodysizelimit`
```

Possible irrelevant change?

FurryR:
> Also keep in mind that Next.js is currently controlling the compression with `inflateSync`. I don't think it makes sense to support different compression algorithms unless we fix both sides.

Implementation already uses `inflatSync`. Possibly a mistake.

I'm with the contributor on this point. Adding support to different compression algorithms does not add a lot of overheads (just a comparison) while ensures forward compatibility (what if we retrieve Brotli, etc. data in the same situation? If we don't handle them, they'll become new defects).

If **it is 100% sure that Content-Encoding only equals to gzip**, then we can safely remove them; but in that case we will need to add an `assert` to ensure `Content-Encoding` is whether nonexistent or gzip.

Hmm, but it does break consistency. I'm pretty sure there is a better solution; @eps1lon may I hear your suggestion?

eps1lon:
We should start with an e2e test first to understand what is actually broken instead of guessing just based on source code.

lazypool:
I have addressed the review feedback: updated the description to match the actual implementation, added the e2e test under `test/e2e/app-dir/resume-content-encoding/` with 5 test cases covering explicit `Content-Encoding` (gzip, deflate, br), magic header detection, and the uncompressed baseline.

test/e2e/app-dir/resume-content-encoding/resume-content-encoding.test.ts
```diff
@@ -0,0 +1,117 @@
+import { nextTestSetup } from 'e2e-utils'
+import { brotliCompressSync, gzipSync, deflateSync } from 'node:zlib'
+
+describe('resume-content-encoding', () => {
+  const { next, isNextDev, skipped } = nextTestSetup({
+    files: __dirname,
+    skipDeployment: true,
+    env: {
+      NEXT_PRIVATE_TEST_HEADERS: '1',
```
eps1lon: This cannot be part of the test. Instead, you should be able to show breaking behavior with a custom adapter e.g. https://nextjs.org/docs/app/api-reference/adapters/implementing-ppr-in-an-adapter

test/e2e/app-dir/resume-content-encoding/resume-content-encoding.test.ts
```diff
@@ -0,0 +1,117 @@
...
+      NEXT_PRIVATE_MINIMAL_MODE: '1',
```
eps1lon: This is only valid for deploys on Vercel. You don't need to set this as e2e tests already run deployed on vercel.

eps1lon (changes requested):

lazypool:
Hey @eps1lon — I went back and reworked the e2e test based on your feedback.

The old one was using `NEXT_PRIVATE_MINIMAL_MODE` and those private headers to send a resume POST locally, which... yeah, not great. So I swapped it over to use the public API instead:

- registered a custom adapter through `adapterPath` in `next.config.js`
- removed `skipDeployment: true` so it actually runs on Vercel deployment now
- on Vercel, the CDN will hit the internal POST+gzip path for PPR resume, and that's where the decompression code gets exercised — if the fix wasn't there, the dynamic content assertion should blow up

Thanks for pointing this out, that's way cleaner. Local tests are all passed.

Let me know if anything still looks off!

arjenblokzijl:
@lazypool im seeing a lot less errors rn in the vercel runtime logs because we upgraded to 16.3

lazypool (commented):
Thanks @arjenblokzijl , I synced things up to the latest commit, so this should be in 0.16 now. I also ran the tests locally.

arjenblokzijl:
@lazypool any clues when this will be picked up?

swarnava:
Hi everyone,

Thank you for your patience while we work on this. We have pushed a fix for this and you should no longer see the Failed to parse postponed state error. 

Please redeploy your application and let us know if you are still facing any issue. This doesn't need a fix on Next.js side.

PR #95246, opened 2026-06-27
docs(opengraph-image): load assets at module scope to keep route static under Cache Components

aurorascharff:
### What?

Document loading request-independent assets at module scope, and update the OG image examples to follow it.

### Why?

Reading a local asset inside the component is async I/O during render, so the route defers to request time. A top-level `await` at module scope keeps the route prerendered without reaching for `use cache`. The pattern previously only lived as an inline snippet in the OG image docs.

### How?

- Add an "Asynchronous I/O at module scope" section to the Cache Components guide as the canonical explanation.
- Load fonts and local assets at module scope in `opengraph-image.mdx` and `image-response.mdx`, and link both to the new section.

### DIFF
--- a/docs/01-app/01-getting-started/08-caching.mdx
+++ b/docs/01-app/01-getting-started/08-caching.mdx
@@ -416,7 +416,7 @@ import { readFile } from 'node:fs/promises'
 const content = await readFile('./config.json', 'utf-8')
 const items = JSON.parse(content).items ?? []
 
-export default async function Page() {
+export default function Page() {
   return (
     <ul>
       {items.map((item) => (
--- a/docs/01-app/01-getting-started/08-caching.mdx
+++ b/docs/01-app/01-getting-started/08-caching.mdx
@@ -403,12 +403,9 @@ export default async function Page() {
 
 > **Good to know:** This includes queries to embedded databases with synchronous APIs, such as `better-sqlite3` or Node.js's built-in [`node:sqlite`](https://nodejs.org/api/sqlite.html). If you need per-request data from a synchronous source, call [`connection()`](/docs/app/api-reference/functions/connection) before the query.
 
-Asynchronous I/O is the exception: even though the read returns the same result each time, an `await readFile(...)` inside a component defers the route to request time. Request-time rendering is expected when the response should be dynamic. To keep the route prerendered, you have two options:
+Some asynchronous APIs read local resources that don't depend on the incoming request, such as fonts or configuration files. When those resources are expected to be the same for every request, read them once at module scope instead of during rendering
 
-- Read the data once at module scope with a top-level `await`, so it runs when the module loads rather than during render.
-- Wrap the read in [`use cache`](/docs/app/api-reference/directives/use-cache), so Next.js includes the cached result in the static shell.
-
-Module scope fits data that never varies per request, such as a local file, because it avoids caching entirely:
+If the data should instead be computed during rendering and reused across requests, wrap the read in [`use cache`](/docs/app/api-reference/directives/use-cache). If the data depends on the incoming request or is expected to change over time, read it during request-time rendering.
 
 ```tsx filename="page.tsx"
 import { readFile } from 'node:fs/promises'
@@ -427,6 +424,7 @@ export default function Page() {
 }
 ```
 
+In this example, the configuration file is expected to be the same for every request, so it is read once at module scope. Calling `await readFile()` inside the component would be treated as uncached data that work must be either accessed within `use cache` or behind a `<Suspense>` boundary. Since this file does not depend on the request and is not expected to change, module scope is the simplest option.
 ## Prerendering
 
 At build time, Next.js renders your route's component tree. How each component is handled depends on the APIs it uses:
--- a/docs/01-app/01-getting-started/08-caching.mdx
+++ b/docs/01-app/01-getting-started/08-caching.mdx
@@ -403,6 +403,32 @@ export default async function Page() {
 
 > **Good to know:** This includes queries to embedded databases with synchronous APIs, such as `better-sqlite3` or Node.js's built-in [`node:sqlite`](https://nodejs.org/api/sqlite.html). If you need per-request data from a synchronous source, call [`connection()`](/docs/app/api-reference/functions/connection) before the query.
 
+## Asynchronous I/O at module scope
+
+Asynchronous I/O during render, such as `await readFile(...)` inside a component, defers the route to request time. Request-time rendering is expected when the response should be dynamic. To keep the route prerendered, you have two options:
+
+- Read request-independent data once at module scope with a top-level `await`, so it runs when the module loads rather than during render.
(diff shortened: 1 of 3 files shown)


docs/01-app/03-api-reference/03-file-conventions/01-metadata/opengraph-image.mdx
```diff
@@ -108,13 +108,15 @@ export const size = {
 
 export const contentType = 'image/png'
 
+// Load assets at module scope. Reading inside the component would be treated
+// as dynamic I/O under Cache Components and opt the route out of static
+// generation.
+const interSemiBold = await readFile(
```
icyJoseph, line 111: Maybe I am not in the right channel here, but I think it can be argued that we don't need to mention CC at all here.

Top Level Await just works, and this font load, doesn't depend on request info, so it is OK, and correct, to put it at top of the module. That also implies that this work owuldn't need to be `use cached

We might want to add a section to these docs, the migration guide, or a standalone guide on Designing and Patterns with Cache Components, where we list this (move to TLA over `use cache` wrapper), the `params.then`, and other tips and tricks for CC productivity.

docs/01-app/01-getting-started/08-caching.mdx
```diff
@@ -376,9 +376,9 @@ You don't need to memorize which operations behave this way. The dev overlay sur
   />
 </FixCardGrid>
 
-## Synchronous I/O and pure computations
+## I/O and computations
 
-Unlike random or time-based APIs, synchronous I/O, module imports, and pure computations are predictable: the same inputs produce the same outputs. Components using only these operations are prerendered automatically, and their output becomes part of the static HTML at build time.
+Unlike random values or timestamps, module imports, synchronous I/O, and pure computations produce the same result every time they run. Components using only these operations are prerendered automatically, and their output becomes part of the static HTML at build time.
```
icyJoseph: Maybe I am a bit tired now, but the original sentence structure could be a bit hard to parse for some readers:

```suggestion
Unlike random values and timestamps, which can vary between renders, module imports, synchronous I/O, and pure computations produce the same result every time they run. Components using only these operations are prerendered automatically, and their output becomes part of the static HTML at build time.
```

wdyy?

docs/01-app/01-getting-started/08-caching.mdx
```diff
@@ -403,13 +403,37 @@ export default async function Page() {
...
+
+Module scope fits data that never varies per request, such as a local file, because it avoids caching entirely:
+
+```tsx filename="page.tsx"
+import { readFile } from 'node:fs/promises'
+
+const content = await readFile('./config.json', 'utf-8')
+const items = JSON.parse(content).items ?? []
+
+export default async function Page() {
```
icyJoseph: let's drop the async to stress the change

```suggestion
export default function Page() {
```

docs/01-app/01-getting-started/08-caching.mdx
```diff
@@ -376,9 +376,9 @@ You don't need to memorize which operations behave this way. The dev overlay sur
   />
 </FixCardGrid>
 
-## Synchronous I/O and pure computations
+## I/O and computations
```
icyJoseph: Predictable operations maybe, wdyt? The core of all these below is that we can be certain, to some extend that, they give the same results.

I think readFile async is nuanced but perhaps correct, assuming the file system doesn't change at all for the server runtime, as in, you are truly reading a file that won't change at all. And the section below explains that well.

icyJoseph: Or Predictable Values?

docs/01-app/01-getting-started/08-caching.mdx
```diff
@@ -403,13 +403,37 @@ export default async function Page() {
...
+  return (
+    <ul>
+      {items.map((item) => (
+        <li key={item.id}>{item.value}</li>
+      ))}
+    </ul>
+  )
+}
+```
+
```
icyJoseph:
````suggestion
Some asynchronous APIs read local resources that don't depend on the incoming request, such as fonts or configuration files. When those resources are expected to be the same for every request, read them once at module scope instead of during rendering

If the data should instead be computed during rendering and reused across requests, wrap the read in [`use cache`](/docs/app/api-reference/directives/use-cache). If the data depends on the incoming request or is expected to change over time, read it during request-time rendering.

```tsx filename="page.tsx"
import { readFile } from 'node:fs/promises'

const content = await readFile('./config.json', 'utf-8')
const items = JSON.parse(content).items ?? []

  return (
    <ul>
      {items.map((item) => (
        <li key={item.id}>{item.value}</li>
      ))}
    </ul>
  )
}
```

In this example, the configuration file is expected to be the same for every request, so it is read once at module scope. Calling `await readFile()` inside the component would be treated as uncached data that work must be either accessed within `use cache` or behind a `<Suspense>` boundary. Since this file does not depend on the request and is not expected to change, module scope is the simplest option.
````

File: docs/01-app/01-getting-started/08-caching.mdx (hunk 4 above)
Comment (aurorascharff) [reply docs/01-app/01-getting-started/08-caching.mdx]: Or Predictable Values?

maybe this one

File: docs/01-app/01-getting-started/08-caching.mdx [hunk 6]
```diff
@@ -403,13 +403,36 @@ export default async function Page() {
...
+In this example, the configuration file is expected to be the same for every request, so it is read once at module scope. Calling `await readFile()` inside the component would be treated as uncached data that work must be either accessed within `use cache` or behind a `<Suspense>` boundary. Since this file does not depend on the request and is not expected to change, module scope is the simplest option.
```
icyJoseph:
```suggestion
In this example, the configuration file is expected to be the same for every request, so it is read once at module scope. Calling `await readFile()` inside the component would be treated as uncached data that must be either accessed within `use cache` or behind a `<Suspense>` boundary. Since this file does not depend on the request and is not expected to change, module scope is the simplest option.
```

docs/01-app/03-api-reference/03-file-conventions/01-metadata/opengraph-image.mdx
```diff
@@ -164,13 +163,12 @@ export const size = {
 
 export const contentType = 'image/png'
 
+const interSemiBold = await readFile(
```
icyJoseph, line 166:
```suggestion
// Font loading, process.cwd() is Next.js project directory
const interSemiBold = await readFile(
```

docs/01-app/03-api-reference/03-file-conventions/01-metadata/opengraph-image.mdx
```diff
@@ -108,13 +108,12 @@ export const size = {
...
```
icyJoseph, line 111:
```suggestion
// Font loading, process.cwd() is Next.js project directory
const interSemiBold = await readFile(
```

icyJoseph (approved):

PR #95260, opened 2026-06-29
fix(app-router): scroll to hash target inside a Suspense boundary

### LINKED ISSUE #65960: Link does not scroll to anchor if it is within Suspense boundary

https://github.com/kachkaev/next-link-scroll-with-suspense-mwe

### To Reproduce

https://github.com/vercel/next.js/assets/608862/464a3373-1f35-4566-8da5-e0c2f6da0086

<ol dir="auto">
<li>Open <a href="http://localhost:3000/" rel="nofollow">http://localhost:3000/</a></li>
<li>Click on <code class="notranslate">/products#category-42</code></li>
<li>Observe <code class="notranslate">&lt;h2 id="category-42"&gt;</code> inside viewport</li>
<li>Go back to <code class="notranslate">/</code></li>
<li>Click on <code class="notranslate">/products-with-suspense#category-42</code> (same page, but categories load inside Suspense boundary with a 100ms delay)</li>
<li>Observe the top of the page, unlike previously</li>
</ol>

<p dir="auto">I’d expect <code class="notranslate">/products#category-42</code> and <code class="notranslate">/products-with-suspense#category-42</code> to work the same way. If <code class="notranslate">#category-42</code> shows while Suspense ‘islands’ get ‘resolved’, a user should be navigated to the hash. Perhaps, it should be smooth in this case (to explain the ‘blink’ of the top of the page), but not having it at all is a bug.</p>

<p dir="auto">This bug can emerge as a regression during refactoring. Imagine we have a product catalog that has been working for years and we’ve added <code class="notranslate">&lt;Suspense&gt;</code> around the list of products to show the skeleton of the <code class="notranslate">/products</code> page ASAP. Users will no longer be able to navigate to a specific category.</p>

<p dir="auto">Interestingly, navigating to  <code class="notranslate">/products-with-suspense#category-42</code> for the second time works correctly (i.e. it scrolls like <code class="notranslate">/products#category-42</code>). This is because of caching. When browser-side cache of <code class="notranslate">/products-with-suspense</code> expires, scrolling is skipped again.</p>

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 20.11.1
  npm: 10.5.0
  Yarn: 4.0.0
  pnpm: 8.10.5
Relevant Packages:
  next: 14.3.0-canary.70 // Latest available version is detected (14.3.0-canary.70).
  eslint-config-next: N/A
  react: 19.0.0-beta-04b058868c-20240508
  react-dom: 19.0.0-beta-04b058868c-20240508
  typescript: 5.1.3
Next.js Config:
  output: N/A
```

Navigation, Partial Prerendering (PPR)

### Additional context

_No response_

errmakov:
### Problem

Navigating to a URL with a hash fragment (`/page#section`) whose target element is rendered inside a `<Suspense>` boundary that hasn't resolved yet (e.g. streamed content) did not scroll to it — the page reset to the top and never recovered. Navigating a second time worked only because the content was cached.

Fixes #65960

### Root cause

`ScrollAndMaybeFocusHandler` (in `layout-router.tsx`) runs once when the segment commits. If the hash target isn't in the DOM at that moment, it consumed the scroll, cleared the hash, and scrolled the segment container (≈ top). Because the handler is an *ancestor* of every Suspense boundary, it does not re-render when a boundary resolves, so it never retried once the target appeared.

### Fix

When the hash target isn't in the DOM, watch for it with a `MutationObserver` and scroll to it once it appears, without consuming the scroll. The watcher is cancelled when the element is found, when a newer navigation supersedes the scroll (`scrollRef.current` is invalidated), or on unmount.

If the fragment never appears, the page no longer scrolls — matching the browser's behavior for an absent `#fragment` (previously it reset to the top). This is the one intentional behavior change; it's why no timeout/"give-up" heuristic is needed.

Applied to both the default scroll handler and the experimental `appNewScrollHandler` fork.

### Testing

Added an e2e test under `test/e2e/app-dir/navigation` whose hash target streams in behind a `<Suspense>` boundary. It fails without the fix (page stays at the top) and passes with it. The existing hash-navigation tests — including the non-existent-hash case — continue to pass.

### DIFF
--- a/packages/next/src/client/components/layout-router.tsx
+++ b/packages/next/src/client/components/layout-router.tsx
@@ -153,13 +153,17 @@ interface PendingHashFragmentScroll {
 }
 
 /**
- * Watch the DOM for a hash fragment's target element to appear (e.g. when a
- * Suspense boundary it lives in resolves), then invoke `onAppear`. The caller
- * is responsible for cancelling the returned watcher once it is no longer
- * needed via `cancelPendingHashFragmentScroll`.
+ * Watch `root`'s subtree for a hash fragment's target element to appear (e.g.
+ * when a Suspense boundary it lives in resolves), then invoke `onAppear`.
+ * Callers scope `root` to the segment's own subtree rather than the whole
+ * document, so the watcher does not react to unrelated DOM changes elsewhere on
+ * the page and does not outlive the segment. The caller is responsible for
+ * cancelling the returned watcher once it is no longer needed via
+ * `cancelPendingHashFragmentScroll`.
  */
 function watchForHashFragment(
   hashFragment: string,
+  root: Node,
   onAppear: () => void
 ): PendingHashFragmentScroll {
   const observer = new MutationObserver(() => {
@@ -167,7 +171,7 @@ function watchForHashFragment(
       onAppear()
     }
   })
-  observer.observe(document.body, { childList: true, subtree: true })
+  observer.observe(root, { childList: true, subtree: true })
   return { hashFragment, observer }
 }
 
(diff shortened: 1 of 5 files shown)


errmakov:
Pushed two follow-ups:

**1. Test for the one intentional behavior change.** The fix stops resetting scroll to the top when a hash target is absent. Added an e2e (`/hash-absent`): scroll down, navigate to a hash whose target never exists, assert the scroll position is left untouched — fails before the fix (resets to 0), passes after. (The existing non-existent-hash assertion only ran from scroll 0, so this case was previously unpinned.)

**2. Scoped the `MutationObserver` to the segment's own subtree** rather than `document.body`, so it doesn't observe the whole document and can't react to unrelated DOM elsewhere on the page (including scrolling to a late, unrelated element that happens to match a stale fragment). The experimental Fragment-ref handler has no single container node, so it still falls back to `document.body`.

Both the existing suspense test and the new one verified red→green.

One question on direction: since `appNewScrollHandler` (Fragment refs) is where this is heading — would you rather the Suspense-reveal scroll be handled there natively than via a `MutationObserver`? I've wired both so it works today, but happy to reshape if the observer isn't the long-term tool you'd want.

PR #95261, opened 2026-06-29
[turbopack] Generate component chunks for each merged group to increase cache hits

sampoder:
This is a WIP PR to attempt to achieve this:

Screenshot 2026-06-28 at 11 53 17 pm

This does involve emitting extra chunks, though in theory, they are already being emitted.

I'm reusing the `module_chunks` field on `OutputChunkRuntimeInfo`:

```rust
/// List of paths of chunks containing individual modules that are part of
/// this chunk. This is useful for selectively loading modules from a chunk
/// without loading the whole chunk.
pub module_chunks: Option<ResolvedVc<OutputAssets>>,
```

This does cause larger build sizes:

| App | Client chunk files (baseline → PR) | Difference | Client chunk bytes (baseline → PR) | Difference |
| vercel-site | 3651 → 5245 | +1594 (43.7%) | 344.5 MiB → 418.4 MiB | +73.9 MiB (21.5%) |
| vercel-marketing | 2242 → 2634 | +392 (17.5%) | 163.5 MiB → 180.7 MiB | +17.2 MiB (10.5%) |
| **Total** | **5893 → 7879** | **+1986 (33.7%)** | **508.0 MiB → 599.1 MiB** | **+91.1 MiB (17.9%)** |


turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -153,12 +154,41 @@ async function loadChunkInternal(
 
     promise = Promise.all(moduleChunksPromises)
   } else {
-    promise = loadChunkPath(sourceType, sourceData, chunkData.path)
+    // For a merged chunk (one that lists more than one module chunk), probe the HTTP cache before
+    // downloading the whole merged chunk: if enough constituent partials are already cached,
+    // load the smaller partials individually instead.
+    let cachedPartials: Map<ChunkPath, number> = new Map()
```
lukesandberg: nit: init to undefined instead of allocating an empty map

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -153,12 +154,41 @@ async function loadChunkInternal(
...
+    if (
+      !TURBOPACK_DEV &&
```
lukesandberg: is the `TURBOPACK_DEV` condition important, wouldn't the includedModuelChunksList just be empty?

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -153,12 +154,41 @@ async function loadChunkInternal(
...
+      sourceType === SourceType.Parent &&
+      includedModuleChunksList.length > 1 &&
+      typeof fetch === 'function'
```
lukesandberg: document this condition, it isn't obvious why it is important here

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -245,6 +275,75 @@ function loadChunkPath(
   return loadChunkByUrlInternal(sourceType, sourceData, url)
 }
 
+/**
+ * Approximate cost of an extra HTTP request, used to decide whether splitting a merged chunk into
+ * individually-cached partials is worthwhile. This is the runtime analog of the `c_req`, however,
+ * it is expressed in compressed transfer bytes rather than pre-minified (which the chunker's # is).
+ */
+const REQUEST_COST_BYTES = 30_000
```
lukesandberg: i wonder if we could/should use the NetworkInformationApi,  `rtt` and `saveData` might be interesting

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -245,6 +275,75 @@ function loadChunkPath(
...
+ * `TypeError` depending on the browser; both are treated as "not cached". `only-if-cached` requires
+ * `same-origin`, which chunks always are.
+ */
+async function probeCachedChunks(
+  chunkPaths: ChunkPath[]
+): Promise<Map<ChunkPath, number>> {
+  const hits: Map<ChunkPath, number> = new Map()
+  await Promise.all(
+    chunkPaths.map(async (chunkPath) => {
+      if (!isJs(chunkPath)) return
```
lukesandberg: we could do this for css chunks also, future work?

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -245,6 +275,75 @@ function loadChunkPath(
...
+      const url = getChunkRelativeUrl(chunkPath)
+      try {
+        const response = await fetch(url, {
+          cache: 'only-if-cached',
+          mode: 'same-origin',
```
lukesandberg:
```suggestion
          // required if passing only-if-cached
          mode: 'same-origin',
```

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -245,6 +275,75 @@ function loadChunkPath(
   return loadChunkByUrlInternal(sourceType, sourceData, url)
 }
 
+/**
+ * Approximate cost of an extra HTTP request, used to decide whether splitting a merged chunk into
+ * individually-cached partials is worthwhile. This is the runtime analog of the `c_req`, however,
```
lukesandberg: not quite an analog, here it is the overaly cost of a request whereas in the chunker it is `c_req + request size`

so this is the combined value

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -245,6 +275,75 @@ function loadChunkPath(
...
+async function probeCachedChunks(
+  chunkPaths: ChunkPath[]
+): Promise<Map<ChunkPath, number>> {
+  const hits: Map<ChunkPath, number> = new Map()
+  await Promise.all(
+    chunkPaths.map(async (chunkPath) => {
+      if (!isJs(chunkPath)) return
+      const url = getChunkRelativeUrl(chunkPath)
+      try {
+        const response = await fetch(url, {
```
lukesandberg: discussed IRL, it would be good to measure performance of this for both hits and misses

lukesandberg (changes requested):
it would be good also to handle the case where you load a Merged(A,B,C) chunk and then can later on tell that you don't need to load A B or C.  The http cache wouldn't work for this, but we could augment the merged chunks with a list of 'also-known-as' chunk urls and then filter by 'loaded chunk urls' as part of this (before probing the http cache)

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -22,6 +22,7 @@ declare var TURBOPACK_NEXT_CHUNK_URLS: ChunkUrl[] | undefined
 declare var CHUNK_BASE_PATH: string
 declare var ASSET_SUFFIX: string
 declare var CROSS_ORIGIN: 'anonymous' | 'use-credentials' | null
+declare var TURBOPACK_DEV: boolean
```
sokra: You don't need the TURBOPACK_ prefix, since we are in the turbopack runtime anyway

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sokra: It need to match the cross origin mode of the scripts, otherwise you are loading twice.

When using cross origin you probably can't fetch the script via fetch.

sampoder: Resolving because we removed the cache lookup for now

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: Removed ✅

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: removed this!

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: this has been removed

sampoder:
> it would be good also to handle the case where you load a Merged(A,B,C) chunk and then can later on tell that you don't need to load A B or C. 

added!

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: agreed - if we come back to caching will do

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: I can look into doing this in the future, I'm not super familiar with how they work

crates/next-core/src/next_manifests/client_reference_manifest.rs
```diff
@@ -86,11 +89,25 @@ pub struct ManifestNodeEntry {
...
+///
+/// Most chunks are a plain URL string. A *merged* chunk (one that bundles several component
+/// chunks) is instead emitted as an object carrying its URL (`u`) alongside its component chunk
+/// paths (`c`). This us to dynamically choose to load the whole chunk or individual components of
+/// it, as neeeded.
+#[derive(Serialize, Debug, Clone)]
+#[serde(untagged)]
+pub enum ClientChunk {
+    Path(RcStr),
+    Merged { u: RcStr, c: Vec<RcStr> },
```
lukesandberg: nit: for client side parsing and encoding speed just make this a tuple so it encodes as an Array

crates/next-core/src/next_manifests/client_reference_manifest.rs
```diff
@@ -461,7 +502,10 @@ async fn build_manifest(
                             inlined: inlined_css,
                             content,
                         });
-                    } else {
+                    } else if !mode.is_production()
+                        || !client_reference_chunk_paths.contains(&path)
+                    {
+                        // TODO: is this safe???
```
lukesandberg: this makes sense to me, entry_js_files are loaded eagerly on navigation but the client_reference_chunk_paths are loaded on hydration.

so technically we are deferring a load, but the load would be conditional

crates/next-core/src/next_manifests/client_reference_manifest.rs
```diff
@@ -522,6 +566,35 @@ impl From<&TurbopackModuleId> for ModuleId {
     }
 }
 
+async fn client_chunk_component_paths(
+    chunk: ResolvedVc<Box<dyn OutputAsset>>,
+    client_relative_path: &FileSystemPath,
+) -> Result<Option<Vec<RcStr>>> {
```
lukesandberg: nit: just return an empty vec instead of None (it is also allocation free), it makes the api less ambiguous imho

the caller can test for emptyiness just as easy as none

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -699,6 +738,49 @@ impl PartialEq for MergeCandidate<'_> {
...
+            component_chunks.push((part.chunk_items, part.batch_groups.into_vec()));
+        } else {
+            // we create a "remainder" chunk with smaller component chunks so that
+            // an entire chunk can be loaded by loading all of its component chunks
+            remainder_items.extend(part.chunk_items);
+            remainder_batch_groups.extend(part.batch_groups);
+        }
+    }
+    if !remainder_items.is_empty() {
+        component_chunks.push((
```
lukesandberg, line 777: should we only have 1 remainder chunk? maybe there are many the component chunks that are only slightly smaller than the min_component_chunk_size?

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -174,6 +152,45 @@ async function loadChunkInternal(
...
+): Promise<unknown> {
+  const componentChunkPromises = componentChunks
+    .map((componentChunk) => availableModuleChunks.get(componentChunk))
+    .filter((p) => p) as Array<Promise<any> | true>
+
+  if (componentChunkPromises.length > 0) {
+    // At least one component chunk is already loaded or loading. Splitting avoids
+    // re-downloading the ones we already have.
+    for (const componentChunk of componentChunks) {
+      if (!availableModuleChunks.has(componentChunk)) {
```
lukesandberg, line 235: `availableModuleChunks` seems like an odd name... should it just be availableChunks?

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -174,6 +152,45 @@ async function loadChunkInternal(
...
+  sourceData: SourceData,
+  componentChunks: ChunkPath[],
+  loadWhole: () => Promise<unknown>
+): Promise<unknown> {
+  const componentChunkPromises = componentChunks
+    .map((componentChunk) => availableModuleChunks.get(componentChunk))
+    .filter((p) => p) as Array<Promise<any> | true>
+
+  if (componentChunkPromises.length > 0) {
+    // At least one component chunk is already loaded or loading. Splitting avoids
```
lukesandberg: we are no longer applying a cost heuristic?

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -182,14 +199,92 @@ const instrumentedBackendLoadChunks = new WeakMap<
...
+  if (components !== undefined) {
+    chunkComponents.set(chunkPath, components)
+  } else {
+    // A plain URL may still be a merged chunk we already registered from its `{ u, c }`.
+    components = chunkComponents.get(chunkPath)
+  }
+
+  // If we have component chunks for this merged chunk, load only the ones we don't already have
+  // instead of the whole merged chunk.
+  if (components !== undefined && components.length > 0) {
```
lukesandberg: is there a way that `components` could be `[]` at this point... seems like an error?  don't bother checking?

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -117,51 +125,21 @@ async function loadChunkInternal(
...
+  // chunkComponents-registry lookup, which would just repeat the same split decision.
+  const promise = loadComponentChunksOrWhole(
+    sourceType,
+    sourceData,
+    componentChunks,
+    () =>
+      loadChunkByUrlWhole(
+        sourceType,
+        sourceData,
+        getChunkRelativeUrl(chunkData.path)
```
lukesandberg: nit: just pass the url instead of the closure? are we concerned about the cost of computing the url?

crates/next-core/src/next_manifests/client_reference_manifest.rs
sampoder: sgtm!

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: yeah - we can switch, this is a pre-existing name fwiw

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: oops - yes I need to add this back

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: i think that should be chill to do

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: interesting. yes, I think we could make smaller mini-component chunks but that point its starting to feel like you're re-doing chunking. maybe I'll do this as a follow-up.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
sampoder: added it back!

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
lukesandberg: yeah fair point.  leave a comment about this as a risk

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
lukesandberg: i guess this name is to distinguish between normal 'module chunks' and chunk-list-chunks, which are a dev only concept?

no big deal

crates/next-core/src/next_client/context.rs
```diff
@@ -593,6 +593,10 @@ pub async fn get_client_chunking_context(
                     first_page_load_priority: chunking_first_page_load_priority,
                     priority_boost_percent: chunking_priority_boost_percent,
                     request_cost: chunking_request_cost,
+                    // Generate component chunks alongside the merged chunk so that the browser
+                    // runtime can fetch an already-cached one instead of the whole merged chunk.
+                    generate_component_chunks: true,
```
lukesandberg: lets tie this to a new experimental next.js option so we can land this sooner and not wait for the release

lukesandberg (commented):
have you gathered any metrics about how many _more_ outputs are generated?

sampoder:
> have you gathered any metrics about how many more outputs are generated?

added this to the pr description!

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: added 👍

crates/next-core/src/next_client/context.rs
sampoder: added! lol yes

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -182,14 +250,93 @@ const instrumentedBackendLoadChunks = new WeakMap<
...
+  sourceType: SourceType,
+  sourceData: SourceData,
+  chunkEntry: ChunkUrlOrMerged
+): Promise<any> {
+  // A merged chunk arrives as a `[url, componentChunkPaths, componentChunkSizes]` array. Register
+  // the components so a by-URL load of this merged chunk — now or from a later navigation — can
+  // be split, and so `registerChunk` can mark them available when the whole chunk loads.
+  let chunkUrl: ChunkUrl
+  let components: ChunkPath[] | undefined
+  if (typeof chunkEntry === 'string') {
```
lukesandberg: to make this 'free' when disabled you could inject a new constant

`const SUPPORT_COMPONENT_CHUNKS = false|true`  here https://github.com/vercel/next.js/blob/a9f913a9f39d7f4fd3d993d19b9ca7fc8fa5e4f1/turbopack/crates/turbopack-ecmascript-runtime/src/browser_runtime.rs#L114-L115

and then add that to these conditions... i SWC should inline and fold that constant which should allow these branches to be simplified and support functions like `registerComponentChunkSizes` etc to be dropped entirely

lukesandberg (approved):

PR #95270, opened 2026-06-29
Render a code frame for build errors thrown collecting page data

unstubbable:
Errors thrown while collecting page data, such as an empty `generateStaticParams` under Cache Components, printed a source-mapped stack in the build output but no code frame, unlike prerender errors from the export worker which show the offending source lines. The static worker exposes both `isPageStatic` and `exportPages`, but only `exportPages` called `installCodeFrameSupport`, so when `isPageStatic` ran the code-frame renderer was never registered and the frame was dropped even though the stack itself was already source-mapped.

This installs the native bindings and code-frame support in the static worker entry (`build/worker.ts`) by wrapping the exposed `isPageStatic`, mirroring what `exportPages` already does. It is deliberately not done in `build/utils.ts`: that module is reachable from `next-server` through the dev server, so importing the code-frame installer there pulls the devtools renderer (`next-devtools/server/shared`) into the production server's file trace, which the `next-server-nft` test rightly rejects. The worker entry is only loaded by build workers, so the renderer stays out of the runtime server. Both installs are idempotent.

With the renderer registered, a build error thrown from `generateStaticParams` now prints the same code frame as a prerender error, pointing at the offending line. The frame is only rendered when the source map carries source content, which happens under `--debug-prerender` (and otherwise `serverSourceMaps`), so normal minified production builds are unaffected. The `empty-generate-static-params` end-to-end test is updated to assert the code frame that now appears in the build error for both the literal and the computed empty array cases.

### DIFF
--- a/packages/next/src/build/worker.ts
+++ b/packages/next/src/build/worker.ts
@@ -2,13 +2,26 @@ import '../server/require-hook'
 // Import cpu-profile to start profiling early if enabled
 import '../server/lib/cpu-profile'
 
+import { installBindings } from './swc/install-bindings'
+import { installCodeFrameSupport } from '../server/lib/install-code-frame'
+import { isPageStatic as isPageStaticImpl } from './utils'
+
 // Set the global asset suffix for Turbopack compiled code to use during prerendering
 ;(globalThis as any).NEXT_CLIENT_ASSET_SUFFIX =
   process.env.__NEXT_PRERENDER_CLIENT_ASSET_SUFFIX
 
-export {
-  getDefinedNamedExports,
-  hasCustomGetInitialProps,
-  isPageStatic,
-} from './utils'
+export { getDefinedNamedExports, hasCustomGetInitialProps } from './utils'
+
+// Install native bindings and code-frame support before collecting page data so
+// errors thrown there (e.g. an empty `generateStaticParams`) render a code
+// frame in the build output, matching prerender errors. Both are idempotent.
+// This is done in the worker entry rather than in `./utils`, which is reachable
+// from `next-server` and would otherwise pull the devtools code-frame renderer
+// into the production server's file trace.
+export const isPageStatic: typeof isPageStaticImpl = async (params) => {
+  await installBindings()
+  installCodeFrameSupport()
+  return isPageStaticImpl(params)
+}
+
 export { exportPages } from '../export/worker'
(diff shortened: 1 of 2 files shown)


packages/next/src/build/worker.ts
```diff
@@ -2,13 +2,27 @@ import '../server/require-hook'
...
+
+// Install native bindings and code-frame support before collecting page data so
+// errors thrown there (e.g. an empty `generateStaticParams`) render a code
+// frame in the build output, matching prerender errors. Both are idempotent.
+// This is done in the worker entry rather than in `./utils`, which is reachable
+// from `next-server` and would otherwise pull the devtools code-frame renderer
+// into the production server's file trace.
+export const isPageStatic: typeof worker.isPageStatic = async (params) => {
+  await installBindings()
+  installCodeFrameSupport()
```
eps1lon, line 23: This is a bit sketchy. Function sounds side-effect free but isn't anymore.

packages/next/src/build/worker.ts
```diff
@@ -2,13 +2,27 @@ import '../server/require-hook'
...
 ;(globalThis as any).NEXT_CLIENT_ASSET_SUFFIX =
   process.env.__NEXT_PRERENDER_CLIENT_ASSET_SUFFIX
 
-export {
-  getDefinedNamedExports,
-  hasCustomGetInitialProps,
-  isPageStatic,
-} from './utils'
+export const getDefinedNamedExports = worker.getDefinedNamedExports
+export const hasCustomGetInitialProps = worker.hasCustomGetInitialProps
```
eps1lon: I'm always a bit paranoid about wildcard exports not being DCEd properly. Let's just stick with static imports and exports for safety.

packages/next/src/build/worker.ts
unstubbable: This is the same pattern we already use here:

https://github.com/vercel/next.js/blob/ae6d43b8e5ebfc7531d3651d0ee4c2c7cdc315a7/packages/next/src/export/worker.ts#L349-L355

eps1lon: 😭

packages/next/src/build/worker.ts
unstubbable: fixed

eps1lon (approved):

PR #95273, opened 2026-06-29
Ignore onAllReady after a fatal shell error in the Node Fizz stream

eps1lon:
React still invokes the `onAllReady` callback after `onShellError` has fired for a fatal shell error, even though the render never completed (facebook/react#36890). Because `onShellError` is terminal, treating a subsequent `onAllReady` as a successful completion is incorrect: it resolves the `allReady` promise as if the render had succeeded, and in the `waitForAllReady` path it would pipe a stream whose shell had already errored.

This change tracks whether the shell errored in both `renderToNodeFizzStream` and `resumeToFizzStream` in `stream-ops.node.ts` and ignores the late `onAllReady` so that a fatal shell error stays terminal. The web path is unaffected because it consumes `renderToReadableStream` (whose returned `allReady` promise already rejects on shell error) rather than the callback form.


eps1lon:
Testing https://github.com/react/react/pull/36903. Seems safe

PR #95275, opened 2026-06-29
docs: Activity bfcacheId and :has caveats

icyJoseph:
- x-ref to `bfcacheId`
- add a section on `:has`

### DIFF
--- a/docs/01-app/02-guides/preserving-ui-state.mdx
+++ b/docs/01-app/02-guides/preserving-ui-state.mdx
@@ -333,7 +333,7 @@ For global state, prefer a `data-*` attribute that React owns:
 ```
 
 ```css
-html[data-modal-open="true"] {
+html[data-modal-open='true'] {
   overflow: hidden;
 }
 ```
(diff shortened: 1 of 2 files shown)


aurorascharff (approved):

eps1lon (commented):
There's a really good section on MDN explaining the problems we're trying to solve here: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:has#performance_considerations

We should reference that to make it clear that this isn't a Next.js specific problem users may encounter but a general DOM performance issue React just surfaces earlier.

icyJoseph:
Indeed, looks like I linked to https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:has#avoid_broad_anchoring - the first child of the performance section, prob a clipboard mishap - I'll try to tweak this more toward that direction though, I cut down quite a bit, and it lost some emphasis

aurorascharff (approved):

PR #95281, opened 2026-06-29
Fix workers on thread to not crash with napi-rs

jimmyhmiller:
We believed the workers were crashing due to issues related to node. We had code that guarded on which version of node you were using and warned that it was unstable. I was able to recreate the crashes regardless of node version or napi-rs version. I filed an issue with napi-rs https://github.com/napi-rs/napi-rs/issues/3357. 

This gets around the crashes we were seeing by making sure we don't drop things off isolate causing these bugs. I did explore writing some tests here, but I wasn't sure a good way to do that. I did make [jimmym/turbopack-loader-worker-changes-stress-test](https://github.com/vercel/next.js/tree/jimmym/turbopack-loader-worker-changes-stress-test) which has  a nice little reproducing stress test that now passes. I wasn't sure if we'd want something like this checked in or not. But figure it is good for people to see.

I probably do need to validate this with some existing apps out there that trigger the crash. But I did keep the default to the process workers for now, so impact should be minimal.

### DIFF
--- a/packages/next/src/build/swc/generated-native.d.ts
+++ b/packages/next/src/build/swc/generated-native.d.ts
@@ -58,7 +58,7 @@ export interface NapiTaskMessage {
 export declare function recvTaskMessageInWorker(
   workerId: number
 ): Promise<NapiTaskMessage>
-export declare function sendTaskMessage(message: NapiTaskMessage): Promise<void>
+export declare function sendTaskMessage(message: NapiTaskMessage): void
 export interface NapiLocation {
   line: number
   column?: number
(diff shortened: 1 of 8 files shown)


lukesandberg:
| "We had coder here guarding ..."

not quite parsing that


Our typical testing approach would be to 
1. try CI with this enabled, lots of tests exercise loaders so this is a reasonable way to test
2. try enabling in `front` and `v0` (once this is merged and released)
3. if that all goes well we can enable by default in `canary` releases

jimmyhmiller:
@lukesandberg 

Lol, `we had code here` not coder..

Rewrote for clarity

lukesandberg:
ok, this looks good to me, and as an experimental feature im not too worried so i think we can just proceed.

lukesandberg (approved):

PR #95289, opened 2026-06-29
fix: params/searchParams in client page crashing dev instant validation

lubieowoce:
Fixes an oversight where `params`/`searchParams` in client pages would cause dev instant validation to go down a codepath intended for build-time instant validation. We'd end up unintentionally wrapping params with an excaustive proxy which would then crash because the store is not set up as expected:
```
## Error Message
Route "/store/[slug]": Could not validate `instant` because an error prevented the target segment from rendering.

Next.js version: 16.3.0-preview.5 (Turbopack)

  [cause]: Error [InvariantError]: Invariant: Expected to have a workUnitStore that provides validationSampleTracking. This is a bug in Next.js.
      at Page (app/store/[slug]/page.tsx:6:11)
    4 |
    5 | export default function Page(props: PageProps<"/store/[slug]">) {
  > 6 |   const { slug } = use(props.params);
      |           ^
    7 |   return (
    8 |     <>
    9 |       <h1>Store {slug}</h1>
}
```
The proxy code should've been gated on `workUnitStore.validationSamples` like we do in other places that care about samples.

### DIFF
--- a/test/e2e/app-dir/instant-validation/app/suspense-in-root/static/valid-client-params/[slug]/layout.tsx
+++ b/test/e2e/app-dir/instant-validation/app/suspense-in-root/static/valid-client-params/[slug]/layout.tsx
@@ -1,0 +1,18 @@
+import { Instant } from 'next'
+import { ReactNode } from 'react'
+import { HackilyPreventFullyStaticServerPrerender } from '../../../../shared'
+
+// We need a layout to export an instant config. Client components cannot have one
+export const instant: Instant = {
+  level: 'experimental-error',
+  unstable_samples: [{ params: { slug: '123' } }],
+}
+
+export default function DummyLayout({ children }: { children: ReactNode }) {
+  return (
+    <>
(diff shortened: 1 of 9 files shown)


test/e2e/app-dir/instant-validation/app/suspense-in-root/static/valid-client-params/[slug]/layout.tsx
```diff
@@ -0,0 +1,18 @@
...
+// We need a layout to export an instant config. Client components cannot have one
+export const instant: Instant = {
+  level: 'experimental-error',
+  unstable_samples: [{ params: { slug: '123' } }],
+}
+
+export default function DummyLayout({ children }: { children: ReactNode }) {
+  return (
+    <>
+      <HackilyPreventFullyStaticServerPrerender />
```
Comment (lubieowoce) [inline test/e2e/app-dir/instant-validation/app/suspense-in-root/static/valid-client-params/[slug]/layout.tsx:14]: workaround for unrelated bug, see NAR-800

unstubbable (approved):

PR #95315, opened 2026-06-30
Fix metadata title dropped on soft navigation with Cache Components

### LINKED ISSUE #95268: App Router: document.title dropped (empty) on client-side navigation with cacheComponents (static metadata & generateMetadata)
### Summary

With `cacheComponents: true` on the App Router, **`document.title` is dropped to an empty string on client-side (soft) navigation** between routes. The `<title>` element is removed from `<head>` and never re-applied, so the browser tab falls back to showing the request URL. It only resolves correctly on a **hard reload**.

This happens for **both** `export const metadata` (static) **and** `export async function generateMetadata` routes, and it is independent of any `fetch`/`await` inside metadata. It reproduces with a plain static title.

This looks related to but distinct from #83017 (there the title *briefly* shows the URL *while async `generateMetadata` awaits a fetch*). Here the title goes **permanently empty** on soft nav, even for a fully static `metadata` export with no async work.

### Verify canary

I reproduced on `16.3.0-preview.5`. I also tested `16.3.0-canary.70`: there `generateMetadata` routes update correctly on soft nav, but **static `export const metadata` titles are still dropped on soft nav**. So the regression seems partially addressed for `generateMetadata` but not for static metadata.

**Current:** After a soft navigation (verified soft — JS context survives, `window` markers persist), `document.title === ""` and `document.querySelector('head > title') === null`. Stable for 5+ seconds; it does not eventually resolve.

**Expected:** `document.title` updates to the destination route's resolved metadata title (e.g. `"Posts | Vincent CHAN"`) on soft navigation, as it does on hard reload.

### Environment

- next: `16.3.0-preview.5` (also tested `16.3.0-canary.70`)
- react / react-dom: `19.2.6`
- node: `v24.11.1`
- platform: `win32 x64`
- Relevant config:
  ```js
  // next.config.js
  const nextConfig = {
    cacheComponents: true,
    partialPrefetching: true, // ruled out as cause (see below)
    reactCompiler: true,      // present
  };
  ```

### What I systematically ruled out

I bisected this on the real app (production-built, `next start`, real Chrome, soft-nav verified via a persistent `window` marker):

- **Static `metadata` vs `generateMetadata`** — both fail on `preview.5`. (On `canary.70`, `generateMetadata` works, static still fails.)
- **`partialPrefetching: false`** — still fails.
- **Removing React `<ViewTransition>`** (it wrapped `<body>`) — still fails.
- **`<Suspense>` placement** — moving the root `<Suspense>` so `<html>`/`<head>` sit outside it (only the dynamic body content is wrapped) — still fails.
- **Upgrading to `canary.70`** — fixes `generateMetadata` only; static `metadata` still fails.

The common factor across all failing cases is `cacheComponents: true` + a client-side navigation. (My root layout fetches dynamic data and renders behind a `<Suspense>`; I have not isolated whether a dynamic/suspended root layout is required to trigger it.)

### Minimal reproduction (best-effort recipe — not yet minimized/verified)

I observed this in a full app; I have not reduced it to the snippet below, so treat it as a starting point. I'm happy to provide a complete reproduction repository.

```js
// next.config.js
module.exports = { cacheComponents: true };
```
```tsx
// app/layout.tsx
export const metadata = { title: { default: "Home", template: "%s | Site" } };
export default function RootLayout({ children }: { children: React.ReactNode }) {
  return (<html lang="en"><body>{children}</body></html>);
}
```
```tsx
// app/page.tsx
import Link from "next/link";
export default function Home() { return <Link href="/about">About</Link>; }
```
```tsx
// app/about/page.tsx
export const metadata = { title: "About" };
export default function About() { return <p>About</p>; }
```
Build (`next build`), `next start`, open `/`, click the link. Expected `document.title === "About | Site"`; observed empty.

### Workaround

A client component that sets `document.title` from a `useEffect` on `usePathname()` change reliably restores correct titles on soft navigation (server metadata still handles hard load / SEO).

acdlite:
With `cacheComponents` + `partialPrefetching`, navigating to an already-prefetched route with a dynamic `generateMetadata` left `document.title` permanently empty until a hard reload.

The prefetch cached the head as complete using the server's `isHeadPartial` flag, which is unreliable under Cache Components. Derive the head's partiality from `isResponsePartial` instead, as segment data already does, so a dynamic head is fetched and applied on navigation while a fully static head stays complete.

Fixing the flag exposed a latent issue in the prefetch scheduler: during a speculative prefetch, the head was unconditionally runtime-prefetched, which previously went unnoticed because the head was mismarked as complete. The head is now runtime-prefetched only when a segment in the new part of the tree is a candidate for runtime prefetching — it rides along with that request rather than spawning a standalone one. If nothing in the new part of the tree is a runtime prefetch candidate, the head is fetched during the navigation instead. This makes the scheduler's metadata-only request path dead code, so it's removed.

Fixes #95268.

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -2969,18 +2969,21 @@ export function writeDynamicRenderResponseIntoCache(
 
     const head = flightDataEntry.head
     if (head !== null && metadataTree !== null) {
-      // When Cache Components is enabled, the server conservatively marks
-      // the head as partial during static generation (isPossiblyPartialHead
-      // in app-render.tsx), even for fully static pages where the head is
-      // actually complete. When the response is non-partial, we override
-      // this since the server confirmed no dynamic content exists.
+      // When Cache Components is enabled, the server's `isHeadPartial` flag
+      // (isPossiblyPartialHead in app-render.tsx) is unreliable: it's computed
+      // before the head is serialized, so it's conservatively `true` for every
+      // statically-generated PPR page — even pages whose head is actually
+      // complete — and it's `false` for runtime/dynamic responses whose head is
+      // actually partial (e.g. a route with an async `generateMetadata`). So we
+      // ignore it and derive the head's partiality from whether the response
+      // itself was partial, exactly as we do for segments (see
+      // `writeSeedDataIntoCache`). A non-partial response carries a complete
+      // head; a partial (postponed) one does not.
       //
-      // Without Cache Components, the server always sends the correct
-      // isHeadPartial value, so no override is needed.
-      const isHeadPartial =
-        !isResponsePartial && process.env.__NEXT_CACHE_COMPONENTS
-          ? false
-          : flightDataEntry.isHeadPartial
+      // Without Cache Components, the server sends the correct isHeadPartial.
+      const isHeadPartial = process.env.__NEXT_CACHE_COMPONENTS
+        ? isResponsePartial
+        : flightDataEntry.isHeadPartial
 
       fulfillEntrySpawnedByRuntimePrefetch(
         now,
--- a/packages/next/src/client/components/segment-cache/scheduler.ts
+++ b/packages/next/src/client/components/segment-cache/scheduler.ts
@@ -27,7 +27,6 @@ import {
   canNewFetchStrategyProvideMoreContent,
   attemptToFulfillDynamicSegmentFromBFCache,
   attemptToUpgradeSegmentFromBFCache,
-  MetadataOnlyRequestTree,
 } from './cache'
 import type { RouteCacheKey } from './cache-key'
 import { createCacheKey } from './cache-key'
@@ -851,18 +850,22 @@ function pingRootRouteTree(
                 ? FetchStrategy.RuntimeShell
                 : FetchStrategy.PPRRuntime
 
-            // Always ping the runtime head — it may need to be fetched
-            // independently even if all runtime segments are already
-            // cached (e.g. navigating between siblings under a shared
-            // runtime layout).
-            const spawnedEntries = new Map<
-              SegmentRequestKey,
-              PendingSegmentCacheEntry
-            >()
-            pingRuntimeHead(now, task, route, spawnedEntries, runtimeStrategy)
-
+            // spawnedRuntimePrefetches was populated during the traversal
+            // above: every segment in the new part of the tree that is a
+            // candidate for runtime prefetching. It's derived purely from
+            // server hints, not cache state — it tells us whether a runtime
+            // request would be needed even if the cache were completely empty.
+            //
+            // If it's null, nothing in the new part of the tree is a candidate
+            // for runtime prefetching, and we don't fetch the head, either —
+            // the head is runtime prefetched only if one of the segments is.
             const spawnedRuntimePrefetches = task.spawnedRuntimePrefetches
             if (spawnedRuntimePrefetches !== null) {
+              const spawnedEntries = new Map<
+                SegmentRequestKey,
+                PendingSegmentCacheEntry
+              >()
+              pingRuntimeHead(now, task, route, spawnedEntries, runtimeStrategy)
               const requestTree = pingRuntimePrefetches(
                 now,
                 task,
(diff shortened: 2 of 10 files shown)


unstubbable (commented):
The test failure looks legit and related to the change.

gaearon:
Had Claude look at the failure. Here's what it had to say:

---

Investigated the CI failure. TL;DR: it's real, deterministic, and a direct consequence of the head-partiality change — and the naive fix (special-casing shell writes) reintroduces #95268, so the two behaviors are in genuine tension and need a design tweak rather than a retry.

## What fails

All three prod jobs fail on the same single test: `prefetch-inlining.test.ts › independent head: metadata is prefetched even when no runtime segment request is needed`. They fail identically because the fixture sets `cacheComponents: true` itself. Reproduced locally with `pnpm test-start-turbo`.

## Mechanism

Verified by instrumenting every head-entry write and every scheduler read of a head entry.

`writeDynamicRenderResponseIntoCache` is also the write path for App Shell (`RuntimeShell`) responses, and for those `isResponsePartial` is always `true` (a shell by definition omits below-shell content). Shell head entries are re-keyed under the shell vary path (params → `Fallback`), so one entry is shared across sibling routes.

In the failing test, prefetching `/test-independent-head/a` now writes:

```
head WRITE strategy=RuntimeShell isPartial=true varyPath=/_head <- ?:FALLBACK <- item:FALLBACK
```

Before this PR that same write was `isPartial=false` (the server's `isHeadPartial`, which is `false` for runtime responses — the flag this PR stops trusting).

Then revealing the link to sibling `/b`:

```
head READ phase=Shell        reqStrategy=RuntimeShell  entry=Fulfilled/RuntimeShell/partial  spawned=false
head READ phase=Speculative  reqStrategy=PPRRuntime    entry=Fulfilled/RuntimeShell/partial  spawned=true   ← the failure
```

The Speculative pass finds the shared shell entry via the `item:FALLBACK` key, sees it's partial with a lower-ranked strategy (`canNewFetchStrategyProvideMoreContent(RuntimeShell, PPRRuntime) === true`), and spawns the metadata-only runtime prefetch. Its response streams `Independent Head Title: b`, which the test rejects with `block: 'reject'`: under App Shells, param-dependent heads are deferred to navigation, never fetched per-link speculatively. On canary the entry is (wrongly) marked complete, so nothing is spawned.

## Why exempting shell writes doesn't work

I tried keeping the old server-flag behavior for `RuntimeShell` writes only. That fixes prefetch-inlining but breaks this PR's own regression test — `document.title` stays `""`. Instrumentation shows `/slow`'s head is cached by the same `RuntimeShell` write path:

```
head WRITE strategy=RuntimeShell isPartial=true varyPath=/slow/__PAGE__/_head <- ?:FALLBACK
```

Marking that entry partial is exactly what makes navigation refetch the dynamic head. It's one cache entry with two consumers that want opposite things:

- **Navigation correctness** wants it partial — shell heads genuinely lack dynamic metadata. This PR gets that right; canary lies and that's #95268.
- **Speculative scheduling** treats "partial + higher-ranked strategy available" as "upgrade now", which App Shells explicitly doesn't want for param-dependent heads. Canary only got this right because the flag lied.

Also worth noting: the over-fetch isn't sibling-specific. With this PR, even a plain link prefetch to a route with dynamic metadata spawns an extra speculative `PPRRuntime` head request after the shell settles (observed on `/slow`) — per-link runtime metadata requests are the cost App Shells was designed to avoid.

## Possible direction

Keep the head marked partial (data-honest), but gate the upgrade in the scheduler — e.g. in the `EntryStatus.Fulfilled` branch reached via `pingRuntimeHead`, skip upgrading a fulfilled `RuntimeShell` head entry when `task.phase === PrefetchPhase.Speculative`, leaving navigation (`Full`) to fetch it. Or split "data is incomplete" from "eligible for speculative upgrade" on the entry. (Untested suggestion.)

acdlite:
The new failing test turned out to be a separate issue that was masked by the other one. Manifested as too many prefetch requests rather than as wrong app behavior. Updated the description:

> Fixing the flag exposed a latent issue in the prefetch scheduler: during a speculative prefetch, the head was unconditionally runtime-prefetched, which previously went unnoticed because the head was mismarked as complete. The head is now runtime-prefetched only when a segment in the new part of the tree is a candidate for runtime prefetching — it rides along with that request rather than spawning a standalone one. If nothing in the new part of the tree is a runtime prefetch candidate, the head is fetched during the navigation instead. This makes the scheduler's metadata-only request path dead code, so it's removed.

unstubbable (approved):

PR #95318, opened 2026-06-30
Turbopack: add more context to persistence file errors

lukesandberg:
### What?

Adds path- and operation-level context to the filesystem errors raised by the `turbo-persistence` commit/rollback flow and the `turbo-tasks-backend` database invalidation/versioning helpers.

### Why?

A user reported a bare `Access is denied. (OS error 5)` on Windows 11 that surfaced out of the cache `commit` flow with no indication of which file or which filesystem operation failed: https://github.com/vercel/next.js/discussions/95130#discussioncomment-17479429   The underlying issue was fixed in #95495 but this adds more information for future issues.

Raw `io::Error`s from `?` carry an OS error code but no path, so there was no way to tell whether the failure was on `CURRENT`, a `.del` file, a per-file `fsync`, the directory `fsync`, or a cleanup `remove_file` etc.

### How?

- Adopted the `fs_err` crate for all fs operations in the persistence layer
- Added additional context to some flows
- Unified `CURRENT` file writing and adopted an 'atomic write' pattern.


lukesandberg:
* **#95318** <a href="https://app.graphite.com/github/pr/vercel/next.js/95318?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95318?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -159,6 +159,23 @@ impl WriteOperationGuard<'_> {
     }
 }
 
+/// Durably writes `seq` to the `CURRENT` file in the database directory `path`, creating the file
+/// if it doesn't exist.
+fn write_current_file(path: &Path, seq: u32) -> Result<()> {
+    let current_path = path.join("CURRENT");
+    (|| {
+        let mut current_file = OpenOptions::new()
+            .create(true)
```
mischnic: This `create(true)` didn't exist before?

lukesandberg: this function is now shared from the `init` path (line 452 before) and from the 'overwrite' paths from `commit` and `cleanup_db_inner` etc

so some of those don't need `create` but afaik it isn't harmful

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -159,6 +159,23 @@ impl WriteOperationGuard<'_> {
...
+            .write(true)
+            .truncate(false)
```
bgw: Why `truncate(false)`? I guess we're always writing the same length to it, but it seems like it would be better to use `truncate(true)`?

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -159,6 +159,23 @@ impl WriteOperationGuard<'_> {
...
+            .read(false)
+            .open(&current_path)?;
+        current_file.write_u32::<BE>(seq)?;
+        current_file.sync_data()
```
bgw: Probably not entirely unreasonable, but we're assuming the 4 bytes are written atomically?

The real way that atomic file writes are supposed to happen are by writing to a different file and then calling `std::fs::rename`, overwriting the original file if it exists.

bgw (approved):

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: atomicity isn't the goal... though i guess if we do the symlinking thing it may need to be

just durability is what we are after, we do not expect or support racing readers

lukesandberg: discussed IRL: write 'tearing' is a concern so we should do the 'move' trick

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: This was preexisting.  im guessing this touches on the atomicity question? truncating the file and then writing it opens a race?

PR #95331, opened 2026-07-01
feat(create-next-app): detect nub package manager

colinhacks:
When invoked through nub, create-next-app misdetects the package manager as npm and prints npm hints. It infers the PM from `npm_config_user_agent`, matching `yarn`/`pnpm`/`bun` and defaulting to npm otherwise; [nub](https://nubjs.com) advertises `nub/<version>` in that string, so it falls through to the default.

nub is a Rust CLI with a pnpm-compatible command grammar. This adds it to the detection in `get-pkg-manager.ts` and to the update-notice map in `index.ts`. Nothing else changes: `install()` already spawns `<pm> install`, and the post-scaffold output prints `nub run dev`.


colinhacks:
Checks are green here — anything I can do to help move this along?

PR #95339, opened 2026-07-01
[ci] Add optimize-ci, changes, and fetch-test-timings jobs to the needs list for the tests-pass job

bgw:
On #95332, the `fetch-test-timings` job failed. That caused downstream jobs to get a "skip" state, instead of a failure state, which then caused `tests-pass` to succeed. This is because if a dependency job fails, GitHub's default behavior is for the downstream job to skip, and to not cascade the failure.

Depending on all of these helper jobs seems like the simplest way to prevent this from happening in the future.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1025,8 +1025,15 @@ jobs:
       stepName: 'test-cache-components-prod-${{ matrix.group }}'
     secrets: inherit
 
+  # We block merging unless this job passes. This is enforced by a "require
+  # status checks to pass" rule on the canary branch.
+  # <https://github.com/vercel/next.js/settings/rules/15507172>
   tests-pass:
     needs:
+      # Note: Some of these jobs are dependencies of other test jobs. If they
+      # fail, they'll cause the dependent test job to be skipped, which could
+      # cause this aggregation job to incorrectly pass if the dependency job is
+      # not listed here.
       - optimize-ci
       - changes
       - build-native
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1027,36 +1027,36 @@ jobs:
 
   tests-pass:
     needs:
-      [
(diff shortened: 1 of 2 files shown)


.github/workflows/build_reusable.yml
```diff
@@ -73,9 +73,12 @@ on:
...
         type: string
         default: ''
-
       testTimingsArtifact:
-        description: 'Name of an uploaded artifact containing test-timings.json. When set, download it instead of fetching via turbo.'
+        description: >
+          Name of an uploaded artifact containing test-timings.json. When set,
+          download it instead of fetching via turbo. This can be useful in
+          ensuring that when failed jobs are retried, they end up with exactly
+          the same set of timings, and therefore run the same set of tests.
```
bgw, line 81: Not directly related, but I had to do some research to figure out why we even did this fetching as a separate job, so I'm improving this description.

.github/workflows/build_and_test.yml
```diff
@@ -1027,36 +1027,36 @@ jobs:
...
-        'test-dev-windows',
-        'test-integration-windows',
-        'test-prod-windows',
-      ]
-
+      - optimize-ci
+      - changes
+      - build-native
+      - build-next
+      - fetch-test-timings
```
bgw, line 1041: This adds `optimize-ci`, `changes`, and `fetch-test-timings`. The rest of this is formatting.

lubieowoce: would it make sense to add a comment on the jobs that we depend on as a workaround for the "CI accidentally passes" problem, and not because they're actually needed? stuff like that tends to get lost and then someone will wonder why they're in here

bgw: Done!

eps1lon (approved):
> That caused downstream jobs to get a "skip" state, instead of a failure state

That's annoying. We should auto-generate the `needs` list in `tests-pass` then.

PR #95341, opened 2026-07-01
[ci] Migrate some CI jobs back to x86-64, build both arm and x86-64 native binaries, include both os and architecture in native artifact names

bgw:
This should be a more robust version of https://github.com/vercel/next.js/pull/95332 that re-introduces some x86-64 runners, but also properly segments the artifact cache.

Even if we don't end up merging this, this does some cleanup of `uploadNativeArtifact` that might be useful to pull out into a separate PR.

### DIFF
--- a/.github/workflows/build_reusable.yml
+++ b/.github/workflows/build_reusable.yml
@@ -63,11 +63,6 @@ on:
         required: false
         type: string
         default: '["ubuntu-latest-16-core-arm-oss"]'
-      buildNativeTarget:
-        description: 'Target for build-native step'
-        required: false
-        type: string
-        default: 'aarch64-unknown-linux-gnu'
       overrideProxyAddress:
         description: Override the proxy address to use for the test
         required: false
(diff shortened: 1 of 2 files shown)


.github/workflows/build_reusable.yml
```diff
@@ -166,7 +163,7 @@ jobs:
 
     defaults:
       run:
-        shell: bash -leo pipefail {0}
+        shell: bash -leou pipefail {0}
```
lubieowoce:
```suggestion
        shell: bash -leuo pipefail {0}
```
i think it's `-o pipefail` so that should go last?

lubieowoce: oh nvm, apparently both work (somehow). but it feels a bit cursed to mix switches and `-o ...` like this

bgw: Yeah, I agree. I'll fix.

eps1lon (approved):

PR #95346, opened 2026-07-01
example: Use a Redis cacheHandler w/o dependencies

icyJoseph:
Verify locally:

- Copy the example to a local directory
- Install deps
- docker compose up -d
- npm run build && npm run start 
- navigate to localhost:3000/cet (or gmt)
- open the Redis view (link in localhost:3000/cet)
- revalidate data on /cet or /gmt
- verify it on the Redis view

I had to update the app to async params, updateTag, and write the handler in such a way that it solved:

```
⨯ TypeError: p.segmentData.get is not a function
    at ignore-listed frames
⨯ Error: failed to pipe response
```

Also adding docs edits with this caveat.

An agent review, and cross checking with a community implementation, surfaced, that we had to read tags from `data.headers['x-next-cache-tags']`. This agent review added a few comments, I think they are useful, but can cut down if needed.

Last but not least, had to update the Time API endpoint (former no longer worked).

### DIFF
--- a/examples/cache-handler-redis/README.md
+++ b/examples/cache-handler-redis/README.md
@@ -54,6 +54,18 @@ The `/[timezone]` page renders a mostly static shell and, inside it, a `'use cac
 
 > **Note:** This example fetches the current time from a public API (`timeapi.io`) purely as sample data to demonstrate caching and revalidation. It is included for learning purposes only. If you reuse it, review and respect that API's terms of use and rate limits, and swap in your own data source for real applications.
 
+## Revalidation: paths, soft tags, and advisory parameters
+
+The **Revalidate** button ([`revalidate-from.tsx`](./app/revalidate-from.tsx)) calls [`updateTag('time-data')`](https://nextjs.org/docs/app/api-reference/functions/updateTag), but that is only one entry point into the remote handler's tag machinery. Two related capabilities are handled without extra code, and two handler parameters are intentionally unused:
+
+- **[`revalidatePath`](https://nextjs.org/docs/app/api-reference/functions/revalidatePath) works without extra code.** Next.js derives implicit _soft tags_ from the route path (prefixed `_N_T_`, e.g. `_N_T_/[timezone]`) and routes path revalidation through the same tags as `revalidateTag`. A `revalidatePath('/…')` call therefore reaches `updateTags(['_N_T_/…'])`, and the next read calls `getExpiration(['_N_T_/…'])`. Because both methods operate generically over any tag string, path revalidation propagates across instances through Redis exactly like an explicit tag.
+
+- **`get(cacheKey, softTags)` ignores `softTags` by design.** A handler can honor soft tags one of two ways: implement `getExpiration` to return the most recent revalidation timestamp (Next.js then performs the soft-tag staleness check itself), or return `Infinity` from `getExpiration` and compare timestamps inside `get`. This handler does the former, so it never needs to read the `softTags` argument.
+
+- **`updateTags(tags, durations)` ignores `durations` by design.** `durations` (`{ expire }`) is only populated when a revalidation call carries a [`cacheLife`](https://nextjs.org/docs/app/api-reference/functions/cacheLife) profile, such as `revalidateTag(tag, 'max')`; it defers a tag's expiration into the future instead of expiring it immediately. This example performs only immediate revalidation, so `durations` is always `undefined` — equivalent to recording the current timestamp, which is what `updateTags` already does.
+
+See the [Soft Tags](https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheHandlers#soft-tags) section of the `cacheHandlers` documentation for the full tag architecture.
+
 ## Documentation
 
 For detailed information, see the official Next.js documentation:
--- a/examples/cache-handler-redis/README.md
+++ b/examples/cache-handler-redis/README.md
@@ -1,8 +1,11 @@
 # Next.js Redis Cache Integration Example
 
-This example is tailored for self-hosted setups and demonstrates how to use Redis as a shared cache. It implements a custom Next.js [`cacheHandler`](https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheHandler) directly against the [`redis`](https://github.com/redis/node-redis) client, so no third-party adapter is required.
+This example is tailored for self-hosted setups and demonstrates how to back Next.js caching with Redis, using no third-party adapter. It wires up **both** of Next.js's cache handler APIs against the [`redis`](https://github.com/redis/node-redis) client, storing everything in a single Redis instance:
 
-The handler lives in [`cache-handler.js`](./cache-handler.js) and implements the `get`, `set`, `revalidateTag`, and `resetRequestCache` methods. It's wired up in [`next.config.js`](./next.config.js), which also sets `cacheMaxMemorySize: 0` to disable the default in-memory cache so Redis is the single shared source of truth across instances.
+- **[`cacheHandler`](https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheHandler) (singular)** — the ISR / incremental cache for pages, route handlers, and images. Implemented in [`cache-handler.js`](./cache-handler.js) with `get`, `set`, `revalidateTag`, and `resetRequestCache`.
+- **[`cacheHandlers`](https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheHandlers) (plural)** — the [`'use cache'`](https://nextjs.org/docs/app/api-reference/directives/use-cache) family. The [`remote`](https://nextjs.org/docs/app/api-reference/directives/use-cache-remote) handler in [`remote-cache-handler.js`](./remote-cache-handler.js) stores `'use cache: remote'` entries with `get`, `set`, `refreshTags`, `getExpiration`, and `updateTags`.
+
+Both are configured in [`next.config.js`](./next.config.js), which enables `cacheComponents: true` (required for `'use cache'`) and sets `cacheMaxMemorySize: 0` so Redis is the single shared source of truth across instances.
 
 Check out this [repository](https://github.com/ezeparziale/nextjs-k8s) that contains a comprehensive setup for Kubernetes.
 
@@ -28,7 +31,7 @@ Once you have installed the dependencies, you can begin running the example Redi
 docker compose up -d
 ```
 
-Then, build and start the Next.js app as usual. The custom cache handler is only used for production builds (`next start`), so run:
+Then, build and start the Next.js app as usual. The custom cache handlers are only used for production builds (`next start`), so run:
 
 ```bash
 npm run build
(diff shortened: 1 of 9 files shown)


examples/cache-handler-redis/next.config.js
timneutkens, line 7: Should use `cacheHandlers` with the `s`

examples/cache-handler-redis/README.md
Svish, line 1: Have to say I'm a bit confused about why there are both a `cacheHandler` and a `cacheHandlers`.

The `cacheHandler` point here mentions ISR, is that for the old-school pages-router ISR concept? Or are actually both needed to "fully cache" a modern Next app-router site?

The `cacheHandlers` mentioned the `"use cache"` family, does the `remote-cache-handler.js` handle both `"use cache"` and `"use cache: remote"` or only the latter? If so, what handles the regular `"use cache"`? Is that what the other `cacheHandler` is for, or would `"use cache"` just be handled by the default Next handler? Can the `remote-cache-handler` also handle `"use cache"`?

examples/cache-handler-redis/README.md
```diff
@@ -20,40 +25,54 @@ yarn create next-app --example cache-handler-redis cache-handler-redis-app
...
 
-- **Create Your Own Handlers:** Take a look at [Custom Redis Handler](https://caching-tools.github.io/next-shared-cache/usage/creating-a-custom-handler) and use it as a basis to create your own handler.
+- **Remote cache (`remote-cache-handler.js`):** stores each `'use cache: remote'` entry under a `nextjs:use-cache:` prefix (the streamed value is base64-encoded). Tag revalidation is timestamp-based: `updateTags` records `nextjs:use-cache-tag:<tag>` = now, and `getExpiration` reports the latest time so Next treats older entries as stale. Clicking **Revalidate** calls [`updateTag('time-data')`](https://nextjs.org/docs/app/api-reference/functions/updateTag), which regenerates the remote entry.
 
-- **Redis Server Setup:** Ensure your Redis server is running and properly configured before starting your Next.js application.
+- **Building without Redis:** both handlers skip connecting during `next build` (they check `NEXT_PHASE`) and degrade gracefully when Redis is unavailable, so the app still builds and runs, just without a shared cache.
```
Svish, line 51: I don't quite understand what is meant by "skip connecting during `next build`", as I thought there would be no connecting at all until the app was actually running. Is it to pre-fill the cache during build or something?

Svish (commented):
Added some questions, hopefully not dumb ones. 🙈

Would also be really helpful if the cache handlers were converted to ESM (`.mjs`) with strict type-checked JSDoc. It's so much easier to understand and use examples when everything isn't `any`. 😅

examples/cache-handler-redis/README.md
Svish: I saw the explanation in the cache handler now: "Connecting to Redis during `next build` can cause issues, so we only connect at runtime.", so I guess the cache-handler _is_ run so that one _could_ prefill the cache during build? What kind of issues are we talking about though? Connection issues when being built on a CI server perhaps?

Svish: About "degrade gracefully when Redis is unavailable", would be good if it was specified what that means exactly. Does it fall back to the built-in Next cache then, or would the app be running without any cache at all? If the latter, would it be possible to somehow defer back to the built-in cache if Redis is unavailable for some reason, or would that be awkward maybe? Maybe outside the scope of this example as well?

examples/cache-handler-redis/remote-cache-handler.js
```diff
@@ -0,0 +1,126 @@
...
+    if (!redis) return undefined;
+
+    const stored = await redis.get(ENTRY_PREFIX + cacheKey);
+    if (!stored) return undefined;
+
+    const data = JSON.parse(stored);
+
+    // Drop the entry once it's past its revalidate window.
+    if (Date.now() > data.timestamp + data.revalidate * 1000) {
+      return undefined;
```
Svish, line 50: Would it make sense if the entry was also deleted from the cache if it's past its revalidate window, or would that be handled elsewhere? Or would it just be overwritten the next time the cache was set anyways perhaps?

examples/cache-handler-redis/next.config.js
```diff
@@ -1,9 +1,21 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  // Enable Cache Components so the `'use cache: remote'` directive is available.
+  cacheComponents: true,
+  // The singular `cacheHandler` backs the ISR/incremental cache (pages, route
+  // handlers, images).
   cacheHandler:
```
Svish, line 7: The docs mentions ["Image Optimization Caching"](https://nextjs.org/docs/app/api-reference/config/next-config-js/incrementalCacheHandlerPath#image-optimization-caching) and that "This opt-in flag will become the default behavior in the next major version." -- Would be really great if dealing with images could be included in this example. 🙏

examples/cache-handler-redis/cache-handler.js
```diff
@@ -1,80 +1,127 @@
...
+    // `x-next-cache-tags` header (app page/route entries).
+    const headerTags = (data.headers?.[NEXT_CACHE_TAGS_HEADER] ?? "")
+      .split(",")
+      .filter(Boolean);
+    const tags = [...new Set([...(ctx?.tags ?? []), ...headerTags])];
+
+    await client.set(
+      CACHE_PREFIX + key,
+      serialize({ value: data, lastModified: Date.now(), tags }),
+    );
```
Svish, line 112: The remote-cache-handler also sets an option here to automatically have Redis expire the entry, `{ EX: Math.max(1, Math.ceil(entry.expire)) }`. Would be great if the cache-handler did something equivalent.

I'm poking around in the types, and seems that the `ctx` here potentially has a `cacheControl` property with this type:
```ts
// next/dist/server/lib/cache-control.d.ts
export type Revalidate = number | false;
export interface CacheControl {
    revalidate: Revalidate;
    expire: number | undefined;
}
```
Not sure how to use it correctly though. Would it be `revalidated` or `expire` that should be passed in as `EX` in that case? Or whichever is smallest?

examples/cache-handler-redis/remote-cache-handler.js
```diff
@@ -0,0 +1,126 @@
...
+      JSON.stringify({
+        value: bytes.toString("base64"),
+        tags: entry.tags,
+        stale: entry.stale,
+        timestamp: entry.timestamp,
+        expire: entry.expire,
+        revalidate: entry.revalidate,
+      }),
+      // Let Redis expire the entry automatically.
+      { EX: Math.max(1, Math.ceil(entry.expire)) },
```
Svish: Poking around in the types and it looks like `EX` is deprecated as a set option here. Looks like it's replaced by `expiration: { type: 'EX', value: number }`, so should be easy fix.

leejpsd:
Thanks for pushing this along. Three things from testing the current revision
(f916b4b7) against next 16.2.x and canary — all reproducible.

### The baseline this replaces is worse than it looks

On canary today the example still pins `@neshca/cache-handler@^1.3.2`, which
resolves to 1.9.0 with peer range `next >= 13.5.1 < 15`. With **npm** — the
path `npx create-next-app --example cache-handler-redis` takes — that silently
resolves `next: latest` down:

```bash
# canary examples/cache-handler-redis/package.json, clean dir, npm 10.9.2
npm install && node -p "require('next/package.json').version"
# → 14.2.35
```

Nothing in the install output mentions it (`npm audit` is the only place it
surfaces: *"Will install next@16.2.12, which is outside the stated dependency
range"*). pnpm and yarn do print a peer warning and keep Next 16, so it's
npm-specific. Since `self-hosting.mdx` links this example twice today — three
times once this PR lands — that's the current official self-hosting path for
npm users. Worth calling out in the PR description; it makes this a fix, not a
cleanup.

### Dropping entries past `revalidate` costs the SWR win

```js
// remote-cache-handler.js
if (Date.now() > data.timestamp + data.revalidate * 1000) {
  return undefined;
}
```

For the `'use cache'` family `revalidate` is the *stale* boundary and `expire`
is the hard one — in `use-cache-wrapper.js` the unconditional discard is
`currentTime > entry.timestamp + entry.expire * 1000`, while the `revalidate`
comparison is gated on `workStore.isStaticGeneration`. So Next expects to
receive the stale entry and decide for itself.

Measured on canary (`cacheLife({stale:1, revalidate:2, expire:300})`, 1.5 s
compute, same app, only the handler's stale policy swapped): serving the stale
entry returns in ~5 ms, dropping it returns in ~1.5 s — each request past the
boundary pays the full render instead of being served immediately. (On released
16.2.x the refresh is still awaited inline, so the gap only shows against
canary.)

In fairness this mirrors the `cacheHandlers` docs, whose own `get()` example
does the same `revalidate`-based drop — which is probably where it came from,
so I opened #96517 for the docs side rather than leaving it to be re-derived.
It also answers @Svish's question on line 48 about deleting entries past the
revalidate window: at that boundary, no; only past `expire`.

### `EX: Math.max(1, Math.ceil(entry.expire))` throws on a non-finite `expire`

Building on @Svish's open note on this line — switching to
`expiration: { type: 'EX', value }` fixes the deprecation but not the value.
`cacheLife({ expire: Infinity })` is the documented way to say
never-expire-on-the-server (Next's validator literally suggests *"Pass
`Infinity` instead of `false` if you want to cache on the server forever"*),
and that reaches `set()` as-is:

```js
await client.set("k", "v", { EX: Math.max(1, Math.ceil(Infinity)) });
// ERR value is not an integer or out of range
```

A `Number.isFinite` guard rather than a fixed ceiling keeps the built-in
profiles working — `default`'s real `expire` is `4294967294`, which Redis
accepts as-is, so clamping to e.g. a year would silently shorten it.

### Two smaller ones

- `updateTags` writes `TAG_PREFIX + tag` with no expiry, so every tag ever
  revalidated stays in Redis forever. A TTL longer than the longest entry
  lifetime bounds that without changing the semantics.
- On @Svish's "degrade gracefully" question: from the code it's the miss path —
  `getClient()` returns `null` when the client isn't ready, so `get()` returns
  `undefined` and every request re-renders. Worth saying that explicitly in the
  README, since the other plausible reading (serve stale from a local fallback)
  is a very different guarantee.

Happy to test a revision against a multi-instance setup if useful.

icyJoseph:
Hate to post a comment indicating that I'll post later, but I've ran out of time for now. I'm trying to exhaustively chase everything mentioned here. Sorry it dragged a bit, past 16.3 landing even, but I think it is nearly there, I just wanna review it all once more.

examples/cache-handler-redis/next.config.js
icyJoseph: I'll do a follow up with that. I don't think is belongs to this scope just now. Good call out though.

Svish: I tried to enable it locally now, and given that the cache handler is setup to handle Buffer instances correctly, which it needs to be regardless, it seems to just works out of the box without any other changes. Would still be nice if it was part of the sample, but definitely not needed to add now. 👍

examples/cache-handler-redis/README.md
icyJoseph: The `cacheHandler` is for ISR — incremental cache, for prerendered pages, fetch, and as you mention, optimized images, opted in by setting this handler (and also `cacheMaxMemorySize: 0` - to disable in-memory altogether):

- https://nextjs.org/docs/app/api-reference/config/next-config-js/incrementalCacheHandlerPath
- https://nextjs.org/docs/app/guides/self-hosting#caching-and-isr

This is definitely not Pages Router only, App Router does use it. The most common use case is to customize where outputs for `/article/[slug]` are saved, at runtime, normally that goes on disk into `.next`, but through the handler you can customize that. If you deploy many instances of the same app, they can all read the same source. 

The `cacheHandlers`, are for Cache Components only. Each key defines what handler will be used for a directive. You can define `default`, `remote`,... and more if you want to — only `private` is off-limits.
 
In this example we are only handling the `remote` handler, so `use cache: remote` goes through that.

Since `default` is not defined/registered by us, the app continues to use the Next.js built-in in-memory handler, but it is possible to define one _if needed_.

examples/cache-handler-redis/README.md
icyJoseph: Next.js prerenders pages at build time.

And we'd be opening a connection to try and store during build time. In this case, we skip both reading and setting back to the Redis store, and still let the build complete, to show that, even if there's no way to reach for that remote DB (legit case in some CI setups), you can still build.

It also wouldn't fall through the built in handler. And furthermore we have set `cacheMaxMemorySize: 0`, so the memory saving here is totally disabled.

examples/cache-handler-redis/next.config.js
icyJoseph: yeah — explicit over implicit, to be honest I had missed that that feature had gone in 😅

aurorascharff (approved):

PR #95358, opened 2026-07-01
fix(config): correctly validate cacheHandlers names

Partha-Shankar:
## What?

The current `cacheHandlers` validation uses `/[a-z-]/`, which only checks whether a key contains at least one lowercase letter or hyphen.

As a result, invalid keys such as `abc123`, `abc_123`, `abc.def`, and `handler!` are accepted even though they contain characters outside the documented allowed set.

## Why?

The validation error states that handler names must only contain `a-z` and `-`, but the current implementation does not enforce that constraint because the regex is not anchored.

## How?

- Replace `/[a-z-]/` with `/^[a-z-]+$/` so the entire handler name is validated.
- Add regression tests covering both valid and invalid handler names.

## Testing

- Added regression tests in `packages/next/src/server/config.test.ts`.

### DIFF
--- a/packages/next/src/server/config.test.ts
+++ b/packages/next/src/server/config.test.ts
@@ -192,4 +192,45 @@ describe('loadConfig', () => {
       expect(result.experimental.externalProxyRewritesResolve).toBe(true)
     })
   })
+
+  describe('cacheHandlers validation', () => {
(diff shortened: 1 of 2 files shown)


unstubbable (approved):
Thank you.

PR #95361, opened 2026-07-01
Rename `DYNAMIC_EXPIRE` and `DYNAMIC_STALE` to describe what they gate

unstubbable:
The `DYNAMIC_EXPIRE` and `DYNAMIC_STALE` constants read as if the value itself is a dynamic expire or stale time, but every comparison is `expire < DYNAMIC_EXPIRE` / `stale < DYNAMIC_STALE`, so the value is the exclusive upper bound of the dynamic range: it is the smallest value that is *not* dynamic, not a dynamic value. The names are backwards at the boundary, which is easy to misread.

The two thresholds also gate different things. `expire` gates whether a cache entry is included in the static prerender: below the threshold it becomes a dynamic hole, though a short-`expire` value is still served during a runtime prefetch. `stale` gates whether an entry is included in a runtime prefetch or app shell.

This renames them to name that consequence: `MIN_PRERENDERABLE_EXPIRE` and `MIN_PREFETCHABLE_STALE`. It reads correctly both at the comparison sites (`expire < MIN_PRERENDERABLE_EXPIRE` is "below the minimum prerenderable expire", i.e. dynamic) and at the one site that uses the value directly, where private caches force `expire: MIN_PRERENDERABLE_EXPIRE` precisely because it is the shortest expire that stays prerenderable (and so is retained without being reclassified as a short-expire dynamic hole). This is a pure rename with no behavior change.

---

### DIFF
--- a/packages/next/src/server/app-render/work-unit-async-storage.external.ts
+++ b/packages/next/src/server/app-render/work-unit-async-storage.external.ts
@@ -366,8 +366,8 @@ export interface PublicUseCacheStore extends CommonUseCacheStore {
   readonly readRootParamNames: Set<string>
   /**
    * The first nested public `'use cache'` invocation with a dynamic cache life
-   * (`revalidate === 0` or `expire < DYNAMIC_EXPIRE`) that propagated up to
-   * this store. Used as `cause` for the nested-dynamic cache error so the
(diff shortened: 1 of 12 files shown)


lubieowoce:
there's a couple tests that reference these constants in comments, let's update those too

lubieowoce (approved):

PR #95363, opened 2026-07-01
Skip saving `expire: 0` values in the default cache handler in prod

unstubbable:
A `'use cache'` value with an explicit `cacheLife({ expire: 0 })` is expired the moment it is produced, so in production the `'use cache'` wrapper regenerates it on every read rather than serving the stored copy. The entry written to the cache handler is never served back, so persisting it is wasted work, and for a remote handler a needless round-trip and stored payload.

Whether to skip that write is left to each cache handler. This change implements it in the built-in default handler; a custom handler can opt into the same behavior. Beyond the saved write, this makes `cacheLife({ expire: 0 })` a reliable way to keep a value out of the server-side cache handler, which covers two cases. In a conditional opt-out, a function with an otherwise normal cache life sets `expire: 0` for a result that is not worth persisting, for example when an upstream fetch errored. In client-only caching, a value is `expire: 0` by design so that it lives only in the client cache, via Cached Navigations or Runtime Prefetches when the route opts in, and never on the server.

The skip is gated on `!process.env.__NEXT_DEV_SERVER`. In production the value is regenerated on every read regardless, so not storing it changes nothing that is served. Development keeps storing it, because the default handler's minimum retention serves the previously cached value across reloads while the entry re-warms in the background.

A new `use-cache-default-handler-expire-zero` end-to-end test enables the default handler's debug logging (`NEXT_PRIVATE_DEBUG_CACHE`) and asserts on its `set()` decisions. Because the debug output is only available from the CLI, the test skips deployment.

### DIFF
--- a/packages/next/src/server/lib/cache-handlers/default.ts
+++ b/packages/next/src/server/lib/cache-handlers/default.ts
@@ -162,6 +162,18 @@ export function createDefaultCacheHandler(maxSize: number): CacheHandler {
       let size = 0
 
       try {
+        // In production an `expire: 0` entry is dynamic: the "use cache"
+        // wrapper regenerates it on every read instead of serving the stored
+        // copy, so persisting it would be a wasted write of a value that is
+        // never served back. Skip storing it so the next read is a plain miss.
+        // The dev server keeps it, because its minimum retention serves the
+        // previously cached value across reloads. The `finally` below still
+        // resolves the pending set.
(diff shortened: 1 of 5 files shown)


unstubbable (commented):
Decision: It should be up to the cache handlers to decide to skip the saving. We'll do that for the built-in default handler (and the ones used by Vercel), and custom handlers can do the same if they wish to do so.

gnoff (approved):

PR #95151, opened 2026-06-25
[PP] Validate Shell prefetches (except gSP)

lubieowoce:
Implements instant validation for `partialPrefetching`. In this mode, `<Link>` prefetches an App Shell, which cannot access link data, and we need to warn for that.

The changes in `instant-validation.tsx` are relatively simple: for an App Shell, we simply use `ShellRuntime` for all the new segments. We might also force them into `Runtime` for the purposes of discriminating dynamic holes. If a hole is present in `ShellRuntime` but disappears in `Runtime`, then we know it's caused by **link data** (as opposed to runtime or dynamic data). I've added some new error messages for this case.

Note that the implementation here is incomplete: it uses the chunks from the dev render, which resolves static params in the `Static` stage. We use `ShellRuntime` for validating the App Shell, so as a result, static params are incorrectly included in it and don't trigger link data errors. This will be implemented in a follow-up.

Note: It seems like we have some pre-existing bug in build validation where `fallbackParams` aren't populated, so params resolve statically when they shouldn't. I've marked two tests with `// TODO(app-shells): missing fallback params in build validation` so we can follow up and fix those.

---

> [!NOTE]
> **Medium Risk**
> Touches instant validation, prerender staging, and error classification in the App Router render path; behavior changes for apps using partial prefetching and how blocking/link violations are reported in dev and build.
> 
> **Overview**
> When **`partialPrefetching`** is enabled, instant navigation validation now simulates **App Shell** prefetches (via **`ValidationPrefetchKind.Shell`**) instead of only legacy speculative static/runtime staging. Segment collection and combined payloads gain a **`RenderStage.ShellRuntime`** stage, and validation picks shell vs legacy behavior from the app’s prefetch mode.
> 
> **Link data** (`params`, `searchParams`, root params) is classified separately from runtime and uncached dynamic holes: new **`DynamicHoleKind.Link`**, dedicated **`createLink*`** errors in **`blocking-route-messages`**, and error codes **1388–1390**. The dev overlay replaces **`isRuntimeVariant`** with **`getGuidanceVariant`** (`link` | `runtime` | `dynamic`) and shows tailored copy and fix cards for link-data cases in blocking routes, metadata, and viewport.
> 
[...]

PR #95365, opened 2026-07-01
[PP] Surface URL data during prefetching as an Instant insight with rule page

aurorascharff:
## What?

Adds the Instant Insight + rule page for the Partial Prefetching shell validation (Janka's #95151): with `partialPrefetching` on, a route that reads `params`/`searchParams` outside `<Suspense>` blocks the extraction of the per-route **App Shell** that every link to the route shares. Surfaced as an Instant insight (not a redbox), with two fix cards: Stream (wrap in `<Suspense>`) · Allow blocking route (`instant = false`).

The insight fires both at link-prefetch time and during a client-side navigation, so it covers a `<Link prefetch>` into the route as well as a plain client navigation to it.

Demos:

- [99-link-params-gsp](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/99-link-params-gsp) — `<Link prefetch>` to a `generateStaticParams` route that awaits `params` unguarded.
- [117-subnav-params](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/117-subnav-params) — client-side navigation into a `[slug]` route that reads `params` outside `<Suspense>`.
- [118-subnav-search-params](https://error-messages-overhaul-ibsl.labs.vercel.dev/scenario/118-subnav-search-params) — client-side navigation into a route that reads `searchParams` outside `<Suspense>`.

Follow-up: #95389 documents the route-side audit in the Adopting Partial Prefetching guide (draft, to be coordinated).

## How?

- Family-wide polish across all 16 insight pages while aligning the new one: consistent "Verifying the fix" wording, normalized `Learn more:` link texts (target title for guides, bare code name for API references), "App Shell" used consistently, CLS guidance reframed as "minimizing layout shift", prose cleanups (no semicolons, no code-initial sentences), and the insight-error-page skill updated to match.
- New `errors/instant-shell-url-data.mdx` rule page, structured like the rest of the insight family (`blocking-prerender-client-hook`, `instant-link-prefetch-partial`); added to every page's Related Insights. New **URL data** glossary entry ("varies per link, not per session"), linked from the page.
- `createLinkBodyErrorInNavigation` re-pointed to the page and classified as a navigation-time insight (`inNavigation` → Instant tab) using the shared `linkCards` fix cards.
- Common insight format for the message: headline "Next.js encountered URL data outside of Suspense", consequence "may prevent the navigation from being instant, leading to a slower user experience" ("may", since boundaries or a runtime prefetch can still keep the navigation instant), and the two fixes phrased as `[stream] Provide a placeholder with <Suspense fallback={...}> around the data access` and `[block] Set export const instant = false to allow a blocking route`. The shell mechanics are explained on the docs page rather than in the headline.
- Second fix reframed from "Disable validation" (group `ignore`) to **Allow blocking route** (group `block`), matching the `blocking-prerender-*` family: `instant = false` marks the segment as allowed to block rather than opting the route out of validation.
- Metadata/viewport variants: only the headline term is aligned ("link data" → "URL data") plus a small grammar fix; framing, fix cards, and docs URLs stay on the existing `-metadata-runtime`/`-viewport-runtime` pages unchanged.
- `errors.json` regenerated append-only (codes 1407-1409), and the docs opt-out wording updated to the new silence-free phrasing.
- e2e snapshots updated for all four configs (dev/start × default/partial-prefetching); unit tests for the matcher, cards, classification, and `isInstantNavigationError`.


errors/instant-shell-url-data.mdx
```diff
@@ -0,0 +1,214 @@
...
+  const product = await getProduct(id)
+  return <Details product={product} />
+}
+```
+
+Learn more: [Streaming](https://preview.nextjs.org/docs/app/guides/streaming).
+
+### Trade-off
+
+The shared parts of the route are prefetched, and the URL-dependent region streams in after navigation, so the user sees a fallback for that region. Design the fallback to approximate the final layout so the page doesn't visibly jump when the content arrives.
```
icyJoseph: Let's call this prevent layout shifts?

docs/01-app/04-glossary.mdx
```diff
@@ -273,6 +273,10 @@ The process of removing unused code from your JavaScript bundles during the buil
 
 # U
 
+## URL data
+
+Data that identifies a specific URL, such as the pathname and query parameters. In the App Router this means [`params`](/docs/app/api-reference/file-conventions/page#params-optional) and [`searchParams`](/docs/app/api-reference/file-conventions/page#searchparams-optional), and the client hooks that read them like [`usePathname`](/docs/app/api-reference/functions/use-pathname) and [`useSearchParams`](/docs/app/api-reference/functions/use-search-params). URL data varies per link, not per user, so it can't be part of a shared [App Shell](#app-shell).
```
icyJoseph: > not per user

Should this be instead:

> URL data varies per link, not session data, so it can't be part of the App Shell.

lubieowoce (approved):

PR #95369, opened 2026-07-01
fix: prerender complete nested generateStaticParams combinations when a sibling parent returns []

### LINKED ISSUE #95166: Child `generateStaticParams` skips pre-rendering ALL static params when returning empty array (`[]`) for SOME parent params

https://github.com/maxijonson/nextjs-prerender-issue

### To Reproduce

(Using App Router on `next@16.3.0-canary.67`)

1. Two dynamic segments where the **parent** segment generates its params from a **layout**:

   ```tsx
   // app/[lang]/layout.tsx
   export function generateStaticParams() {
     return [{ lang: "en" }, { lang: "fr" }];
   }
   export default function LangLayout({ children }: { children: React.ReactNode }) {
     return children;
   }
   ```

2. A nested `[slug]` page whose `generateStaticParams` returns a **non-empty** array for one
   parent value and **`[]`** for another:

   ```tsx
   // app/[lang]/thing/[slug]/page.tsx
   export function generateStaticParams({ params }: { params: { lang: string } }) {
     return params.lang === "en"
       ? [
           { lang: params.lang, slug: "a" },
           { lang: params.lang, slug: "b" },
         ]
       : [];
   }
   export default async function Page({ params }: { params: Promise<{ lang: string; slug: string }> }) {
     const { lang, slug } = await params;
     return (
       <div>
         {lang} {slug}
       </div>
     );
   }
   ```

3. For contrast, a sibling that returns the same **non-empty** set, but for **every** parent (the positive control):

   ```tsx
   // app/[lang]/thing-control/[slug]/page.tsx
   export function generateStaticParams({ params }: { params: { lang: string } }) {
     return [
       { lang: params.lang, slug: "a" },
       { lang: params.lang, slug: "b" },
     ];
   }
   // same default Page as above
   ```

4. Build:

   ```bash
   rm -rf .next && npx next build
   ```

**Expected:** `/en/thing/a` and `/en/thing/b` prerender (the `en` GSP returned them).
`fr` simply contributes no static children. This is the documented bottom-up/top-down
nested `generateStaticParams` behavior.

**Actual:** `thing/[slug]` prerenders **zero** pages — the valid `en` pages are dropped too.
The route is still printed as `●` (SSG) but emits no HTML and has no entries in the
prerender manifest. The only difference from the control route is the `fr` return value
(`[]` vs `[{slug:"a"},{slug:"b"}]`), yet that empty return poisons the entire route.

Build output:

```
├ ● /[lang]/thing-control/[slug]
│ ├ /en/thing-control/a
│ ├ /en/thing-control/b
│ ├ /fr/thing-control/a
│ └ /fr/thing-control/b
└ ● /[lang]/thing/[slug]        <-- no child paths, 0 pages
```

Verification:

```bash
find .next/server/app -path '*/thing/*' -name '*.html'        # bug route  => empty
find .next/server/app -path '*thing-control*' -name '*.html'  # control    => 4 files
```

## Reproduced across versions and bundlers

| Next             | Turbopack     | Webpack (`next build --webpack`) |
| ---------------- | ------------- | -------------------------------- |
| 16.2.4           | 0 pages (bug) | 0 pages (bug)                    |
| 16.2.9           | 0 pages (bug) | 0 pages (bug)                    |
| 16.3.0-canary.67 | 0 pages (bug) | —                                |

Not Turbopack-specific (webpack reproduces it), not fixed in the latest stable, and still
present on canary.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0 (macOS, arm64 / Apple Silicon)
Binaries:
  Node: 24.x
  npm: 11.x
Relevant Packages:
  next: 16.2.4 (also reproduced on 16.2.9 and 16.3.0-canary.67)
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Dynamic Routes

next build (local)

## Workaround

Returning a placeholder slug for the empty parent case works around the issue, but is not ideal. This is also the workaround documented for [Cache Components](https://nextjs.org/docs/app/api-reference/functions/generate-static-params#with-cache-components) when **all** param values are unknown. However, in this case, all params are **known**, they're just **known** to not exist under a subset of the parent segments.

```ts
export function generateStaticParams({ params }: { params: { lang: string } }) {
  return params.lang === "en"
    ? [{ lang: params.lang, slug: "a" }, { lang: params.lang, slug: "b" }]
    : [{ lang: params.lang, slug: "_" }]; // placeholder slug
}
```

```
└   /[lang]/thing/[slug]
  ├ ● /en/thing/a
  ├ ● /en/thing/b
  └ ● /fr/thing/_
```

vismayIO:
### What & Why

Fixes #95166.

When a route has two nested dynamic segments where the **parent** segment's params come from a layout's `generateStaticParams`, and the **child** page's `generateStaticParams` returns a non-empty array for some parent values but `[]` for others, **none** of the child pages were prerendered — including the valid ones.

Example: `/[lang]/thing/[slug]` where the layout generates `[{ lang: 'en' }, { lang: 'fr' }]`, and the child returns `[{ slug: 'a' }, { slug: 'b' }]` for `en` but `[]` for `fr`. Expected: `/en/thing/a` and `/en/thing/b` prerender; `fr` contributes no static children. Actual (before this change): the route emitted **0 pages**.

### Root cause

In `buildAppStaticPaths` (`packages/next/src/build/static-paths/app.ts`), an empty child result for the `fr` branch passes the *incomplete* parent params (`{ lang: 'fr' }`, missing `slug`) through to `routeParams`. That incomplete combination made `hadAllParamsGenerated` `false`, which gated off prerendering for the **entire** route — dropping the complete `en` combinations too. (The gate exists because `validateParams` throws on a missing param in non-PPR mode.)

### Fix

In non-PPR mode, filter out param combinations that are missing a pathname segment's param *before* prerendering. Complete combinations (`en`) are still statically generated; incomplete branches (`fr`) fall back to on-demand rendering — the documented behavior. The PPR/cache-components path is unchanged, since it intentionally uses every combination to generate fallback shells.

### Testing

- Added `test/production/app-dir/nested-empty-generate-static-params/` — mirrors the issue's reproduction (a mixed `thing/[slug]` route plus a positive-control `thing-control/[slug]` that returns a non-empty set for every parent). Asserts the `en` pages land in the prerender manifest and are served, while `fr` pages are not prerendered.
- Verified the existing `test/production/app-dir/empty-generate-static-params/` suite still passes (fully-empty `generateStaticParams` remains on-demand — no regression).
- `pnpm --filter=next types` passes.


vismayIO:
@acdlite, Could you review this PR?

vismayIO:
@vercel/nextjs-maintainers, Could you please review this PR? Fix + tests are included. Thanks!

vismayIO:
@aurorascharff, Could you please review this PR?

PR #95370, opened 2026-07-01
Keep the request body a plain Readable after middleware so Readable.toWeb() doesn't hang

### LINKED ISSUE #95335: NextJS copies _writableState to a Readable object, breaks Node Readable.toWeb() on POST routes

https://github.com/abir-taheer/next-js-readable-stream-bug

### To Reproduce

  1. `git clone https://github.com/abir-taheer/next-js-readable-stream-bug`
  2. `npm install && npm run dev`
  3. `bash test.sh`

Current: POST requests that pass through middleware returning NextResponse.next() hang indefinitely when the downstream handler reads the body via `Readable.toWeb()`. The request never completes and times out.

Expected: The request body should be fully readable by the downstream handler regardless of how it consumes the stream. `Readable.toWeb()` is the standard Node.js API for converting a Readable to a Web ReadableStream — it should work on the IncomingMessage after middleware runs.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 18
Binaries:
  Node: 24.17.0
  npm: 11.13.0
  Yarn: N/A
  pnpm: 10.34.4
Relevant Packages:
  next: 16.2.6
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 6.0.3
Next.js Config:
  output: N/A
```

Middleware

### Additional context

The root cause is in `replaceRequestBody()` (`body-streams.ts`), not in middleware itself. After middleware calls `NextResponse.next()` on a POST route, `runMiddleware` clones the body into a PassThrough (Duplex) and copies its properties back onto the `IncomingMessage` (Readable) via `for...in`. This copies `_writableState` and 39 Writable prototype methods onto the Readable. `Node's Readable.toWeb()` then sees `_writableState.finished === false` and hangs waiting for the writable side to close.

The timing is deterministic, `_writableState.finished` becomes true via `process.nextTick()`, but `replaceRequestBody()` runs in a microtask (from `await endPromise` in `finalize()`). Microtasks always run before nextTick, so it always copies `finished === false` regardless of body size.

`NextResponse.rewrite()` is unaffected because it creates a new internal request and doesn't go through `replaceRequestBody()`. GET requests are also fine since there's no body to clone. The issue only affects POST (or PUT/PATCH) through `NextResponse.next()`.

UditDewan:
### What's the problem?

A `POST` (or `PUT`/`PATCH`) request that passes through middleware returning `NextResponse.next()` hangs indefinitely when the downstream handler reads the body via Node's `Readable.toWeb()`. The request never completes and eventually times out.

Reproduction: https://github.com/abir-taheer/next-js-readable-stream-bug

### Root cause

When middleware runs, `runMiddleware` clones the request body and later calls `finalize()`, which grafts the buffered stream back onto the original `IncomingMessage` via `replaceRequestBody()` (`packages/next/src/server/body-streams.ts`). `replaceRequestBody` copies the buffered stream's enumerable properties onto the request.

The buffered stream (`p2`) was a `PassThrough` — a `Duplex` — so its writable-side internals (`_writableState` plus the enumerable `Writable` methods like `write`/`end`) were copied onto the `IncomingMessage`. Because that `_writableState.finished` is `false`, Node stream utilities that inspect it — including `Readable.toWeb()`, which uses `finished()`/end-of-stream detection — treat the request as a still-open writable stream and wait forever.

`NextResponse.rewrite()` is unaffected (it builds a new internal request and skips this path), and `GET`/`HEAD` requests are fine because there is no body to clone.

### The fix

`p2` is only ever fed with `.push()`, so it never needs a writable side. Making it a plain `Readable` instead of a `PassThrough` keeps the finalized request a pure `Readable`, so `Readable.toWeb()` (and any other duck-typing based on `_writableState`) behaves correctly. No behavior change for the existing consumers, which only read the stream.

### Testing

Added `test/unit/body-streams.test.ts`, which drives the real clone → `finalize()` flow and asserts the finalized request:
- is no longer writable (`_writableState` is `undefined`), and
- is fully consumable via `Readable.toWeb()` (this hangs before the fix).

Fixes #95335


timneutkens:
I've opened a new PR while keeping the attribution here: #95607

As it needs to run deployment tests which fail on forks.

Closing this one in favor of #95607.

PR #95383, opened 2026-07-01
refactor: remove unnecessary switches from StagedRenderingController

lubieowoce:
We have a couple switches in `StagedRenderingController` where the main purpose ends up being eliminating `Before/Dynamic/Abandoned` (i.e. the start and the end stages). Since we migrated everything to `RENDER_STAGE_ORDER` loops, we no longer really need to list all the intermediate stages out explicitly in these switches. It's enough to rule out the "special" stages  and let the rest of the code be generic. This saves some work when adding/removing intermediate stages.

View with "hide whitespace", otherwise it looks messy.

### DIFF
--- a/packages/next/src/server/node-environment-extensions/io-utils.tsx
+++ b/packages/next/src/server/node-environment-extensions/io-utils.tsx
@@ -9,6 +9,7 @@ import {
   createSyncIORuntimeError,
   type SyncIOApiType,
 } from '../app-render/sync-io-messages'
+import { InvariantError } from '../../shared/lib/invariant-error'
 
 export function io(expression: string, type: SyncIOApiType) {
   const workUnitStore = workUnitAsyncStorage.getStore()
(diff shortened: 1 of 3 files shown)


lubieowoce:
* **#95384** <a href="https://app.graphite.com/github/pr/vercel/next.js/95384?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95383** <a href="https://app.graphite.com/github/pr/vercel/next.js/95383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95383?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#95596** <a href="https://app.graphite.com/github/pr/vercel/next.js/95596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/node-environment-extensions/io-utils.tsx
```diff
@@ -56,21 +57,36 @@ export function io(expression: string, type: SyncIOApiType) {
       const stageController = workUnitStore.stagedRendering
       if (stageController && stageController.shouldTrackSyncInterrupt()) {
         let syncIOError: Error
-        if (
-          stageController.currentStage === RenderStage.Static ||
-          stageController.currentStage === RenderStage.EarlyStatic
-        ) {
```
lubieowoce, line 62: drive-by fix: this was missing `RenderStage.Shell[Early]Static`. It doesn't matter much because right now `createSyncIOError` and `createSyncIORuntimeError` do the exact same thing. For now I'm assuming this split was put in place for a reason so i'm preserving it and just fixing the condition

gnoff (approved):

PR #95384, opened 2026-07-02
[PPF] Sync IO is only allowed in the dynamic stage

lubieowoce:
This PR makes Sync IO behavior more restrictive when `partialPrefetching` is on (either globally or for a page).

- Previously, we allowed `await cookies(); Date.now()` in in segments that won't ever be runtime prefetched. This was achieved by separating segments into "early" and "late" stages and only erroring in the "early" ones. 
- After this PR, it will always be an error to do sync IO anywhere other than after `io()`/`connection()`/uncached IO (i.e. outside the Dynamic stage).

We're doing this mainly because with `partialPrefetching`, `cookies()` resolves in App Shells, and sync IO causes a severe deopt in any prerender. `allow-runtime` also opts a route into `partialPrefetching`, which means that even the segments above the runtime prefetch boundary will need an App Shell and can't tolerate Sync IO.

This PR removes all early/late stage separation, because we no longer need to vary the Sync IO behavior on the segment. Instead, we now pick whether a stageController should use
- `SyncIOMode.AllowedInRuntimeOrDynamic` (legacy)
- `SyncIOMode.AllowedInDynamic` (`partialPrefetching`)
- `SyncIOMode.Untracked` if it needs to opt out

I've removed the existing tests that asserted that sync IO is allowed in segments above the `allow-runtime` boundary. Instead, we check that sync io either errors (if partialPrefetching is on) or is allowed (if partialPrefetching is off)

Closes NAR-855


lubieowoce:
* **#95384** <a href="https://app.graphite.com/github/pr/vercel/next.js/95384?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95384?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#95383** <a href="https://app.graphite.com/github/pr/vercel/next.js/95383?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95596** <a href="https://app.graphite.com/github/pr/vercel/next.js/95596?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

gnoff (approved):

PR #95385, opened 2026-07-02
[DO NOT MERGE] debug: queue tracing for popstate-revalidate CI flake

gaearon:
Debugging branch for the CI-only failure of `app dir - navigation > browser back to a revalidated page > should load the page` seen on #95381 ([failing job](https://github.com/vercel/next.js/actions/runs/28552750725/job/84654270421)).

This is #95381 plus:

- Temporary `console.log` tracing of the router action queue (dispatch/run/settle/reject/advance/flush, with per-action ids). The test harness echoes browser logs into CI output, so a CI repro will include the exact action interleaving.
- A temporary CPU-throttled test that forces `back()` to discard the in-flight revalidating server action, to give CI more samples of the discard path.

In the failing runs, the server log shows no requests at all after the action `POST` — the client router appears to freeze with no console errors. Locally the failure does not reproduce (including with forced discards under 10x CPU throttling), so this collects the interleaving from CI directly.

Do not merge.

### DIFF
--- a/packages/next/src/client/components/app-router-instance.ts
+++ b/packages/next/src/client/components/app-router-instance.ts
@@ -65,27 +65,33 @@ export type ActionQueueNode = {
   discarded?: boolean
 }
 
+// Called when `action` settles, to run the next action in the queue.
 function runRemainingActions(
   actionQueue: AppRouterActionQueue,
+  action: ActionQueueNode,
   setState: DispatchStatePromise
 ) {
-  if (actionQueue.pending !== null) {
-    actionQueue.pending = actionQueue.pending.next
+  // Only the settlement of the action at the head of the queue may advance the
+  // queue. A discarded action settles while the navigation that discarded it
+  // is the head; advancing from here would start the next queued action
+  // against router state that doesn't include that navigation yet.
+  if (actionQueue.pending === action) {
+    actionQueue.pending = action.next
     if (actionQueue.pending !== null) {
       runAction({
         actionQueue,
         action: actionQueue.pending,
         setState,
       })
+      return
     }
-  } else {
-    // Check for refresh when pending is already null
-    // This handles the case where a discarded server action completes
-    // after the navigation has already finished and the queue is empty
-    if (actionQueue.needsRefresh) {
-      actionQueue.needsRefresh = false
-      actionQueue.dispatch({ type: ACTION_REFRESH }, setState)
-    }
+  }
+
+  if (actionQueue.pending === null && actionQueue.needsRefresh) {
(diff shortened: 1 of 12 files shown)


gaearon:
Root cause found — closing this debug PR. The trace it captured was decisive.

**The failure is not in the router action queue** (and not caused by #95381). In the failing attempt, the queue trace shows `navigate` → `hmr-refresh` → `server-action` all settling cleanly, the queue going idle — and then after `browser.back()` **no `restore` action is ever dispatched**, no reload, no network, no errors, for 45s.

**Mechanism (reproduced deterministically on unmodified canary, no CPU throttling needed):**

1. A navigation settles with `pushRef.pendingPush: true`. The actual `history.pushState` only happens when React commits that state (`HistoryUpdater`).
2. A refresh-family action right behind it (`hmr-refresh` from webpack's on-demand route compile in the CI runs; `router.refresh()` in the repro) returns a state built via `navigateToKnownRoute(..., navigateType: 'replace')`, i.e. a fresh `pushRef` with `pendingPush: false`.
3. When React commits only the later state (the earlier one is superseded before it ever commits — timing dependent), `HistoryUpdater` never sees `pendingPush: true` and `replaceState`s the *previous page's* history entry to the new URL. The back entry is destroyed.
4. `browser.back()` then has no in-app entry to traverse to (or lands on an entry with no state, hitting the silent `if (!event.state) return` bail in the popstate handler) — the page appears frozen: no dispatch, no fetch, no error.

Deterministic repro (fails 3/3 on canary even at 1x CPU):

```tsx
// app/popstate-refresh-race/page.tsx
'use client'
import { useRouter } from 'next/navigation'
  const router = useRouter()
  return (
    <button
      id="go"
      onClick={() => {
        router.push('/popstate-revalidate/foo')
        router.refresh()
      }}
    >
      go
    </button>
  )
}
```

```ts
const lengthBefore = await browser.eval('window.history.length')
await browser.elementById('go').click()
// ... wait for /popstate-revalidate/foo to render ...
const lengthAfter = await browser.eval('window.history.length')
expect(lengthAfter).toBe(lengthBefore + 1) // ❌ received: lengthBefore — the push was dropped
```

Why it surfaced as a burst of `navigation.test.ts` flakes on July 1 (~21:30–23:00 UTC) across unrelated branches (including a docs-only PR, 3/3 attempts): every failing run's merge base included the React upgrade `92f4fda3` → `ec0fca31` (#95368), which plausibly shifted commit timing so the dev-mode `hmr-refresh` more often supersedes the navigation commit. The underlying bug — `pendingPush` being droppable by a superseding replace-type state — is Next.js's, and is reachable deterministically as shown above.

Issue #84299, opened 2025-09-27
Transition deadlock when navigating and double calling server action


denk0403:
### Link to the code that reproduces this issue

https://codesandbox.io/p/devbox/next-js-transition-starvation-kpl45h

### To Reproduce

1. Start the application in development or production.
2. Open the network tab in the browser devtools.
3. Click the button on the home page. This will start a transition which pushes a new path and then awaits a server action.
4. Once the server action has started pending **and** has not completed yet, click the button again.

<img width="1276" height="317" alt="Image" src="https://github.com/user-attachments/assets/e207aa60-d196-4619-9ebe-be4667f90137" />

### Current vs. Expected behavior

Current Behavior:
- The navigation never completes and the second action is dropped (not visible in the network tab).
- Subsequent presses to the button also do not navigate, indicating the application is in a transition deadlock.

https://github.com/user-attachments/assets/d06388c6-14af-44f7-99ff-a195709013d3

Expected behavior:
- Both actions should fire from the home page, and once both complete, the navigation should occur.

### Provide environment information

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Sun Aug  6 20:05:33 UTC 2023
  Available memory (MB): 4102
  Available CPU cores: 2
Binaries:
  Node: 20.9.0
  npm: 9.8.1
  Yarn: 1.22.19
  pnpm: 8.10.2
Relevant Packages:
  next: 15.6.0-canary.5 // There is a newer canary version (15.6.0-canary.32) available, please upgrade! 
  eslint-config-next: N/A
  react: 19.1.1
  react-dom: 19.1.1
  typescript: 5.3.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Server Actions

### Which stage(s) are affected? (Select all that apply)

next dev (local), next start (local)

### Additional context

I tested various older versions of Next.js, and I believe this was introduced in v15.2.2. I suspect PR #75362 is related to the issue.

<sub>[NEXT-4733](https://linear.app/vercel/issue/NEXT-4733/transition-deadlock-when-navigating-and-double-calling-server-action)</sub>

Suhaib3100:
### Root Cause Analysis

I've investigated this issue and identified the root cause in the action queue logic.

**The Issue:**
[...]

PR #95391, opened 2026-07-02
Fix navigation getting reverted when a Server Action is in flight

### LINKED ISSUE #86151: `loading.js` sometimes causes soft-navigation to get stuck and not render the new page, despite receiving it from the server

gaearon:
Alternative to https://github.com/vercel/next.js/pull/95188.
Fixes https://github.com/vercel/next.js/issues/86151#issuecomment-4802297986.

This fixes a bug where a navigation would spuriously get reverted once a Server Action settles.

This also fixes another bug where a navigation would get stuck forever (https://github.com/vercel/next.js/issues/84299).

## Test Plan

New e2e tests that fail before the fix and pass after the fix. I've steered the agent to cover both the broken case now *and* the case that the regressing PR (#82674) was originally trying to fix. The test itself is reduced from https://github.com/vercel/next.js/pull/95188.

Also, verified before/after in a real tiny app based on the tests. Before the fix, it reverts to the previous route; after the fix, it stays on the last route as it should. 

The tests themselves are a bit wonky and do some timing assertions. I had no luck reducing them further.

## Summary

*Note: Everything below is written by Fable.*

Fixes a regression where a navigation could get reverted shortly after it committed if a Server Action was still in flight when the navigation started: the destination page renders, then the router snaps back to the previous URL. Regressed in 15.5.1 via #82674. Reported in #86151, with a repro and initial investigation by @errmakov in #95188.

### Why?

When a navigation starts while a Server Action is in flight, the router discards the action and the navigation takes its place at the head of the action queue. The discarded action still runs to completion. Since #82674, its completion also called `runRemainingActions`, which advances the queue — but the discarded action is no longer at the head of the queue, the navigation is. This started the next queued action while the navigation was still running, computed against router state that didn't include the navigation yet. Those actions send their requests with the previous URL, and applying their responses reverts the navigation.

The promise rejection path had the same problem even before #82674: a discarded action that rejected would also advance the queue.

### How?

Only the settlement of the action at the head of the queue may advance the queue. A discarded action's settlement (resolve or reject) now only records whether the action revalidated, and the deferred refresh is flushed once the queue goes idle. This keeps the behavior #82674 was after: no refresh for discarded actions that didn't revalidate, and a refresh for those that did — including when they settle mid-navigation, a case that could previously lose the refresh entirely.


gnoff (approved):

test/e2e/app-dir/actions-discarded-navigation-revert/app/layout.tsx
```diff
@@ -0,0 +1,26 @@
+import { Controls } from './controls'
+
+// The stamp only changes when the router refreshes: navigations preserve the
+// root layout, so a new stamp on the client means a refresh happened.
+export const dynamic = 'force-dynamic'
+
+let renders = 0
```
acdlite: This can break when running in a deploy test because the server is stateless.

The way I've been doing it is to count on the client, like this: https://github.com/vercel/next.js/blob/f14048b691ea7806ab881036b6a4fd4db71c0996/test/e2e/app-dir/use-router-bfcache-id/app/%5Bgroup%5D/%5Bpage%5D/page.tsx#L5-L11

acdlite (approved):
Test looks good! left one request to make sure the test is compatible with stateless deploy environments

PR #95401, opened 2026-07-02
[ci] Add stats comment with partial stats

eps1lon:
We used to not post a comment when a part of the bundler matrix got cancelled. Conceptually, a cancelled job is just like a failed on in the sense of not being able to produce stats.

Now we always aggregate stats and post stats from the bundlers that did produce them.

The risk here is a race between a run that was cancelled because another run superseded it and the run that superseded it. Though practically, aggregating the stats isn't slower than running the full workflow. Not impossible (especially given how slow the Webpack job is) but incredibly rare.

## test plan

Needs to land on canary since that's the context in which the PR comment runs.

### DIFF
--- a/.github/workflows/pull_request_stats.yml
+++ b/.github/workflows/pull_request_stats.yml
@@ -89,6 +89,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Keep in sync with STATS_EXPECTED_BUNDLERS
         bundler: [webpack, turbopack]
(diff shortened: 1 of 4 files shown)


.github/workflows/pull_request_stats.yml
bgw, line 93: Possibly leave a comment here too about `STATS_EXPECTED_BUNDLERS`

bgw (approved):

PR #95405, opened 2026-07-02
fix(middleware): exclude _writableState from replaceRequestBody() to fix Readable.toWeb() on POST routes

### LINKED ISSUE #95335: NextJS copies _writableState to a Readable object, breaks Node Readable.toWeb() on POST routes

sleitor:
### What?

In `replaceRequestBody()` inside `body-streams.ts`, replace the bare `for...in` property copy with a filtered copy that excludes `_writableState` and `Writable.prototype` methods when merging a PassThrough back into an IncomingMessage.

### Why?

When middleware calls `NextResponse.next()` on a POST route, `cloneBodyStream()` creates a `PassThrough` (a Duplex stream) and `replaceRequestBody()` copies its properties onto the `IncomingMessage` (a pure Readable) using `for...in`. This also copies `_writableState` and 39+ Writable prototype methods.

`Readable.toWeb()` inspects `_writableState.finished` on the object and hangs indefinitely when it is `false` — it waits for a writable side that will never close. The timing is deterministic: `_writableState.finished` only becomes `true` via `process.nextTick()`, but `replaceRequestBody()` runs in a microtask (after `await endPromise`), so microtasks always execute before the nextTick callback, meaning `finished === false` is always copied.

GET requests and `NextResponse.rewrite()` are unaffected. Only POST (or PUT/PATCH) through `NextResponse.next()` is affected.

### How?

Build a module-level `WRITABLE_ONLY_KEYS` set from `_writableState` plus `Object.getOwnPropertyNames(Writable.prototype)` and `continue` on those keys inside the `for...in` loop. This keeps all Readable-compatible properties (`_readableState`, event emitter methods, readable methods) while preventing Writable internals from leaking onto the IncomingMessage.

Fixes #95335

### DIFF
--- a/packages/next/src/server/body-streams.ts
+++ b/packages/next/src/server/body-streams.ts
@@ -1,6 +1,6 @@
 import type { IncomingMessage } from 'http'
 import type { Readable } from 'stream'
-import { PassThrough } from 'stream'
+import { PassThrough, Writable } from 'stream'
 import bytes from 'next/dist/compiled/bytes'
 
 const DEFAULT_BODY_CLONE_SIZE_LIMIT = 10 * 1024 * 1024 // 10MB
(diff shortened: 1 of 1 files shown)


sleitor:
Closing in favour of #95370 which was opened earlier and provides a more comprehensive fix.

PR #95408, opened 2026-07-02
Migrate napi-rs bindings from v2 to v3

jimmyhmiller:
## What

Migrates the native bindings from napi-rs v2 to v3:

- Bump the workspace to napi 3.10.0 / napi-derive 3.5.8 / @napi-rs/cli 3.7.2 and port all first-party bindings to the modern API: ThreadsafeFunction builder (with `Weak` instead of the deprecated `unref`), lifetime-carrying `Object`/`Function`/`Unknown` value types, `FunctionRef::borrow_back`, `External`/`ExternalRef`, and `#[napi(object)]` outputs.
- Replace the `SendableExternalRef` unsafe Send/Sync wrapper with the idiomatic v3 pattern: sync `#[napi]` fns that extract Send data from the `&External` up front and return `env.spawn_future(...)` as `PromiseRaw` (JS still sees a Promise).
- Vendor `lightningcss-napi` 0.4.6 in-repo as `turbopack-lightningcss-napi`, ported to napi v3, because upstream pins napi ^2 (a single addon can only link one napi runtime). This also fixes the JS visitor bridge, which silently no-op'd under v3 via the v2 compat path (breaking CSS `url()`/`@import` rewriting and CSS-module handling). Provenance, local modifications, and drop criteria are documented in the crate README.
- Regenerate `generated-native.d.ts` with the v3 CLI, adding `ts_arg_type`/`ts_return_type` overrides where v3 typegen is weaker than the runtime contract.

## Why

napi v2 is in maintenance mode; v3 brings handle-scope lifetimes (compiler-checked JS value scoping), the builder-based ThreadsafeFunction API, and first-class async/Promise integration, and unblocks future Node-API features.

## Notes for review

- The `generated-native.d.ts` diff is large mostly because the v3 CLI orders declarations alphabetically; the meaningful changes are the new `TransformOutputResult`/`NapiCompilationEvent` interfaces, the now-synchronous lightningCss declarations, and `lockfileUnlock`.
- JS-observable behavior is intended to be unchanged throughout, with one deliberate fix: `project_shutdown` no longer consumes the exit receiver before `stop_and_wait()`, so a panic there leaves exit handlers runnable by the fallback `onExit()` path.


jimmyhmiller:
Sorry, fable decided it wanted to create a PR...

PR #95412, opened 2026-07-02
Migrate napi-rs bindings from v2 to v3

jimmyhmiller:
Draft PR for testing


.github/workflows/build_reusable.yml
```diff
@@ -102,7 +102,12 @@ on:
         type: string
         default: 'yes'
 env:
-  NAPI_CLI_VERSION: 2.18.4
+  # Newest @napi-rs/cli that runs on Node.js 20.9.0 (NODE_LTS_VERSION below):
+  # 3.5.0+ depends on @inquirer/prompts 8, whose @inquirer/core 11 requires
+  # `util.styleText` (Node >= 20.12) and crashes at CLI startup. The CLI here
+  # only orchestrates cargo and names artifacts (types are committed), so
+  # drifting from the workspace devDependency (3.7.2) is acceptable.
```
bgw, line 109: Asking here: https://vercel.slack.com/archives/C04KC8A53T7/p1783709048113789

If we have to do this, let's make everything use version 3.4.1 of the CLI? I don't like the idea of mixing these versions between jobs if there's not a really good reason we can't just pin on 3.4.1?

crates/next-napi-bindings/src/css/mod.rs
```diff
@@ -1,17 +1,23 @@
...
-#[allow(clippy::not_unsafe_ptr_arg_deref)]
-#[js_function(1)]
-fn transform_style_attribute(ctx: CallContext) -> napi::Result<JsUnknown> {
-    lightningcss_napi::transform_style_attribute(ctx)
+#[napi(js_name = "lightningCssTransformStyleAttribute")]
+#[allow(dead_code)]
+pub fn transform_style_attribute<'env>(
+    env: &'env Env,
+    opts: Object<'_>,
+) -> napi::Result<napi::bindgen_prelude::Unknown<'env>> {
```
bgw: Nit: Just import `napi::bindgen_prelude::Unknown` from `napi::Unknown`.

packages/next/src/build/swc/generated-native.d.ts
```diff
@@ -5,7 +5,9 @@ import type { TurbopackResult } from './types'
 export type TurboTasks = { readonly __tag: unique symbol }
 export type ExternalEndpoint = { readonly __tag: unique symbol }
 export type NextTurboTasks = { readonly __tag: unique symbol }
-export type RefCell = { readonly __tag: unique symbol }
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
bgw: Huh? `RefCell` is still used by `initCustomTraceSubscriber` below?

crates/next-napi-bindings/src/next_api/endpoint.rs
```diff
@@ -145,50 +148,59 @@ async fn get_written_endpoint_with_issues_operation(
...
-                get_written_endpoint_with_issues_operation(endpoint_op);
-            let read = read_strongly_consistent_and_apply_effects(
-                written_entrypoint_with_issues_op,
-                |v| &v.effects,
-            )
-            .await?;
-            let WrittenEndpointWithIssues {
-                written, issues, ..
-            } = &*read;
+pub fn endpoint_write_to_disk<'env>(
```
bgw: This looks like the switch from `async fn` here to using a synchronous function with `spawn_future` blames to f38f75513464471e1569bedc66521751ffa251ca.

That commit claims that `External<T>` is `!Send`.

This is not true: https://docs.rs/napi/latest/napi/bindgen_prelude/struct.External.html#impl-Send-for-External%3CT%3E

What is true is that `ExternalRef<T>` is `!Send`, but it seems like in most (all?) places where we're using `ExternalRef<T>`, we could just be using `External<T>` or `&External<T>`.

I don't like switching things to use `spawn_future` without really good reason because it adds more nesting and makes this harder to read.

crates/next-napi-bindings/src/next_api/endpoint.rs
```diff
@@ -145,50 +148,59 @@ async fn get_written_endpoint_with_issues_operation(
...
 
 #[tracing::instrument(level = "info", name = "get server-side endpoint changes", skip_all)]
 #[napi(ts_return_type = "{ __napiType: \"RootTask\" }")]
 pub fn endpoint_server_changed_subscribe(
-    #[napi(ts_arg_type = "{ __napiType: \"Endpoint\" }")] endpoint: External<ExternalEndpoint>,
+    env: Env,
+    #[napi(ts_arg_type = "{ __napiType: \"Endpoint\" }")] endpoint: ExternalRef<ExternalEndpoint>,
     issues: bool,
-    func: JsFunction,
+    #[napi(ts_arg_type = "(...args: any[]) => any")] func: napi::bindgen_prelude::FunctionRef<
```
bgw: - Nit: add an import at the top of the file for `FunctionRef`
- Follow-up: Let's try to clean up these typescript `any` types wherever we reasonably can.

crates/next-napi-bindings/src/next_api/endpoint.rs
```diff
@@ -2,7 +2,10 @@ use std::{ops::Deref, sync::Arc};
 
 use anyhow::Result;
 use futures_util::TryFutureExt;
-use napi::{JsFunction, bindgen_prelude::External};
+use napi::{
+    Env, Unknown as JsUnknown,
```
bgw: Let's not rename `Unknown` to `JsUnknown`. If we think that `Unknown` is too ambiguous of a name, we can use the fully qualified `napi::Unknown` everywhere.

There are so many types in napi that have both a `Js`-prefixed and a non-`Js`-prefixed version that I thought it was a different type and this was confusing to me.

crates/next-napi-bindings/src/next_api/utils.rs
```diff
@@ -406,35 +409,42 @@ impl<T: ToNapiValue> ToNapiValue for TurbopackResult<T> {
...
-pub fn subscribe<T: 'static + Send + Sync, F: Future<Output = Result<T>> + Send, V: ToNapiValue>(
+pub fn subscribe<
+    T: 'static + Send + Sync,
+    F: Future<Output = Result<T>> + Send,
+    V: 'static + ToNapiValue,
+>(
     ctx: NextTurbopackContext,
-    func: JsFunction,
+    env: &Env,
+    func: &FunctionRef<JsUnknown<'static>, ()>,
```
bgw: Follow-up: It'd be nice if we could get rid of `Unknown` here, it looks like this could be `T`? That'll involve updating callsites too.

Something like this:

```diff
diff --git a/crates/next-napi-bindings/src/next_api/project.rs b/crates/next-napi-bindings/src/next_api/project.rs
index eb05739082..9e7a164427 100644
--- a/crates/next-napi-bindings/src/next_api/project.rs
+++ b/crates/next-napi-bindings/src/next_api/project.rs
@@ -1780,7 +1780,7 @@ pub fn project_entrypoints_subscribe(
     env: Env,
     #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
     #[napi(ts_arg_type = "(...args: any[]) => any")] func: napi::bindgen_prelude::FunctionRef<
-        JsUnknown<'static>,
+        TurbopackResult<Option<NapiEntrypoints>>,
         (),
     >,
 ) -> napi::Result<External<RootTask>> {
diff --git a/crates/next-napi-bindings/src/next_api/utils.rs b/crates/next-napi-bindings/src/next_api/utils.rs
index 4f62a09071..2445a7379e 100644
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -435,11 +435,11 @@ pub fn subscribe<
 >(
     ctx: NextTurbopackContext,
     env: &Env,
-    func: &FunctionRef<JsUnknown<'static>, ()>,
+    func: &FunctionRef<V, ()>,
     handler: impl 'static + Sync + Send + Clone + Fn() -> F,
     mapper: impl 'static + Sync + Send + FnMut(ThreadsafeCallContext<T>) -> napi::Result<V>,
 ) -> napi::Result<External<RootTask>> {
-    let js_func: Function<'_, JsUnknown<'static>, ()> = func.borrow_back(env)?;
+    let js_func: Function<'_, V, ()> = func.borrow_back(env)?;
     let func: ThreadsafeFunction<T, (), V, Status, true> = js_func
         .build_threadsafe_function::<T>()
         .callee_handled::<true>()
```

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -410,17 +410,20 @@ impl From<NapiDefineEnv> for DefineEnv {
 pub struct ProjectInstance {
     turbopack_ctx: NextTurbopackContext,
     container: ResolvedVc<ProjectContainer>,
-    exit_receiver: tokio::sync::Mutex<Option<ExitReceiver>>,
+    // `Arc` so the `'static` futures spawned by `project_on_exit`/`project_shutdown` can own a
+    // handle to it (the `External<ProjectInstance>` itself is `!Send`). Never locked across an
+    // await point.
+    exit_receiver: Arc<std::sync::Mutex<Option<ExitReceiver>>>,
```
bgw: 👍 this looked a little sketchy, but it seems to check out. This is just a poor-man's `ArcSwap`.

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -2068,95 +2126,132 @@ impl From<UpdateInfo> for NapiUpdateInfo {
...
+        .borrow_back(&env)?
+        .build_threadsafe_function::<UpdateMessage>()
+        .callee_handled::<true>()
+        .build_callback(|ctx| {
+            let message = ctx.value;
+            Ok(NapiUpdateMessage::from(message))
+        })?;
+    let tt = project.turbopack_ctx.turbo_tasks().clone();
+    // Unlike napi v2, v3 does not enter the tokio runtime context on the JS thread, so
+    // `tokio::spawn` from a sync `#[napi]` fn panics without this wrapper.
```
bgw: :+1: Makes sense, but let's delete this meta-comment about the migration.

Normally you'd use `spawn_future`, which I guess would handle this for you, but this is a callback-based API rather than a promise-returning API, so it doesn't make sense for us to construct a promise object.

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -410,17 +410,20 @@ impl From<NapiDefineEnv> for DefineEnv {
...
-pub fn project_new(
-    env: Env,
+pub fn project_new<'env>(
+    env: &'env Env,
     options: NapiProjectOptions,
     turbo_engine_options: NapiTurboEngineOptions,
     napi_callbacks: NapiNextTurbopackCallbacksJsObject,
-) -> napi::Result<JsObject> {
-    let napi_callbacks = NapiNextTurbopackCallbacks::from_js(&env, napi_callbacks)?;
+) -> napi::Result<napi::bindgen_prelude::PromiseRaw<'env, External<ProjectInstance>>> {
```
bgw: Use the thing we already imported at the top of the file.

```suggestion
) -> napi::Result<PromiseRaw<'env, External<ProjectInstance>>> {
```

crates/next-napi-bindings/src/next_api/endpoint.rs
```diff
@@ -148,12 +151,11 @@ async fn get_written_endpoint_with_issues_operation(
...
 #[napi]
 pub async fn endpoint_write_to_disk(
-    #[napi(ts_arg_type = "{ __napiType: \"Endpoint\" }")] endpoint: External<ExternalEndpoint>,
+    #[napi(ts_arg_type = "{ __napiType: \"Endpoint\" }")] endpoint: &External<ExternalEndpoint>,
 ) -> napi::Result<TurbopackResult<NapiWrittenEndpoint>> {
-    let ctx = endpoint.turbopack_ctx();
-    let endpoint_op = ***endpoint;
-    let (written, issues) = endpoint
-        .turbopack_ctx()
+    let ctx = endpoint.turbopack_ctx().clone();
```
bgw: Not needed (left over from previous refactoring commits in this PR)

```suggestion
    let ctx = endpoint.turbopack_ctx();
```

crates/next-napi-bindings/src/next_api/utils.rs
```diff
@@ -406,35 +409,42 @@ impl<T: ToNapiValue> ToNapiValue for TurbopackResult<T> {
...
+    T: 'static + Send + Sync,
+    F: Future<Output = Result<T>> + Send,
+    V: 'static + ToNapiValue,
+>(
     ctx: NextTurbopackContext,
-    func: JsFunction,
+    env: &Env,
+    func: &FunctionRef<V, ()>,
     handler: impl 'static + Sync + Send + Clone + Fn() -> F,
-    mapper: impl 'static + Sync + Send + FnMut(ThreadSafeCallContext<T>) -> napi::Result<Vec<V>>,
```
bgw, line 435: :+1: Returning `Vec<V>` here was a bit whacky before. I think this is because napi v2 tried to make the arguments generic, but by using a `Vec<_>` they made it so that all of the arguments had use the same generic which was... a choice. It seems napi v3 solved this by using tuple types for arguments.

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -727,13 +730,12 @@ async fn benchmark_file_io(turbo_tasks: &NextTurboTasks, dir: &Path) -> Result<(
...
-    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: External<ProjectInstance>,
+    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
     options: NapiPartialProjectOptions,
 ) -> napi::Result<()> {
-    let ctx = &project.turbopack_ctx;
-    let options = options.into();
+    let ctx = project.turbopack_ctx.clone();
     let container = project.container;
-
+    let options = options.into();
```
bgw: Revert this, don't need the clone:

```suggestion
    let ctx = &project.turbopack_ctx;
    let options = options.into();
    let container = project.container;

```

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -744,20 +746,19 @@ pub async fn project_update(
 /// is created with filesystem cache enabled.
 #[napi]
 pub async fn project_invalidate_file_system_cache(
-    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: External<ProjectInstance>,
+    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
 ) -> napi::Result<()> {
+    let ctx = project.turbopack_ctx.clone();
```
bgw, line 763: Okay, this clone is actually needed since `project` is now a reference and not an owned `External` 😆

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -744,20 +746,19 @@ pub async fn project_update(
...
+        // TODO: Let the JS caller specify a reason? We need to limit the reasons to ones we
+        // know how to generate a message for on the Rust side of the FFI.
+        ctx.turbo_tasks()
             .backend()
             .invalidate_storage(invalidation_reasons::USER_REQUEST)
     })
     .await
     .context("panicked while invalidating filesystem cache")??;
-    Ok(())
+    Ok::<_, napi::Error>(())
```
bgw: Revert, this is trivially inferred.

```suggestion
    Ok(())
```

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -787,13 +798,16 @@ async fn project_on_exit_internal(project: &ProjectInstance) {
...
-    project.turbopack_ctx.turbo_tasks().stop_and_wait().await;
-    project_on_exit_internal(&project).await;
+    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
+) -> napi::Result<()> {
+    let tt = project.turbopack_ctx.turbo_tasks().clone();
+    let exit_receiver = project.exit_receiver.clone();
+    tt.stop_and_wait().await;
+    run_exit_handlers(exit_receiver).await;
+    Ok::<_, napi::Error>(())
 }
```
bgw, line 818: Avoid some clones, return types can be inferred.

```suggestion
/// Runs exit handlers for the project registered using the [`ExitHandler`] API.
///
/// This is called by `project_shutdown`, so if you're calling that API, you shouldn't call this
/// one.
#[napi]
pub async fn project_on_exit(
    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
) -> napi::Result<()> {
    run_exit_handlers(&*project.exit_receiver).await;
    Ok(())
}

/// Takes the [`ExitReceiver`] out of the shared slot and runs the registered exit handlers.
///
/// The receiver is taken only once the caller is ready to run the handlers, so an earlier failure
/// (e.g. a panic in `stop_and_wait` during `project_shutdown`) leaves it in place for a fallback
/// `project_on_exit` call.
async fn run_exit_handlers(exit_receiver: &std::sync::Mutex<Option<ExitReceiver>>) {
    let exit_receiver = exit_receiver
        .lock()
        .expect("panicked while holding the exit receiver")
        .take();
    exit_receiver
        .expect("`project.onExitSync` must only be called once")
        .run_exit_handler()
        .await;
}

/// Runs `project_on_exit`, and then waits for turbo_tasks to gracefully shut down.
///
/// This is used in builds where it's important that we completely persist turbo-tasks to disk, but
/// it's skipped in the development server (`project_on_exit` is used instead with a short timeout),
/// where we prioritize fast exit and user responsiveness over all else.
#[tracing::instrument(level = "info", name = "shutdown project", skip_all)]
#[napi]
pub async fn project_shutdown(
    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
) -> napi::Result<()> {
    let tt = project.turbopack_ctx.turbo_tasks();
    tt.stop_and_wait().await;
    run_exit_handlers(&*project.exit_receiver).await;
    Ok(())
}
```

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -1301,13 +1315,22 @@ async fn app_route_filter_for_write_phase(
...
     let container = project.container;
+    project_write_all_entrypoints_to_disk_inner(ctx, container, app_dir_only).await
+}
+
+async fn project_write_all_entrypoints_to_disk_inner(
+    ctx: NextTurbopackContext,
+    container: ResolvedVc<ProjectContainer>,
+    app_dir_only: bool,
+) -> napi::Result<TurbopackResult<Option<NapiEntrypoints>>> {
+    let ctx = &ctx;
```
bgw: `project_write_all_entrypoints_to_disk_inner` is no longer providing any value, there's only one caller.

```suggestion
#[napi(ts_return_type = "Promise<TurbopackResult<Partial<NapiEntrypoints>>>")]
pub async fn project_write_all_entrypoints_to_disk(
    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
    app_dir_only: bool,
) -> napi::Result<TurbopackResult<Option<NapiEntrypoints>>> {
    let ctx = &project.turbopack_ctx;
```

crates/next-napi-bindings/src/transform.rs
```diff
@@ -84,7 +84,7 @@ fn skip_filename() -> bool {
 
 impl Task for TransformTask {
     type Output = (TransformOutput, FxHashSet<Atom>, FxHashMap<String, usize>);
-    type JsValue = Object;
+    type JsValue = crate::TransformOutputResult;
```
bgw: `TransformOutputResult` should be moved into this file, and `lib.rs` should re-export it.

bgw (changes requested):

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -2068,95 +2080,124 @@ impl From<UpdateInfo> for NapiUpdateInfo {
...
+    pub message: String,
+    pub event_json: String,
+    #[napi(ts_type = "unknown")]
+    pub event_data: External<Arc<dyn CompilationEvent>>,
+}
+
+impl From<Arc<dyn CompilationEvent>> for NapiCompilationEvent {
+    fn from(event: Arc<dyn CompilationEvent>) -> Self {
+        NapiCompilationEvent {
+            type_name: event.type_name().to_string(),
```
bgw: `napi` should be able to handle `&'static str`?

```suggestion
    pub type_name: &'static str,
    pub severity: String,
    pub message: String,
    pub event_json: String,
    #[napi(ts_type = "unknown")]
    pub event_data: External<Arc<dyn CompilationEvent>>,
}

impl From<Arc<dyn CompilationEvent>> for NapiCompilationEvent {
    fn from(event: Arc<dyn CompilationEvent>) -> Self {
        NapiCompilationEvent {
            type_name: event.type_name(),
```

crates/next-napi-bindings/src/next_api/turbopack_ctx.rs
```diff
@@ -92,15 +96,19 @@ impl NextTurbopackContext {
...
+                .call_async(Ok(TurbopackInternalErrorOpts {
                     message,
                     anonymized_location: panic_location,
                 }))
                 .await
-                .expect_err("throwTurbopackInternalError must throw an error")
+            {
+                Ok(()) => panic!("throwTurbopackInternalError must throw an error"),
+                Err(err) => err,
+            }
```
bgw: Why doesn't `expect_err` not still work here?

crates/next-napi-bindings/src/next_api/turbopack_ctx.rs
```diff
@@ -163,8 +171,18 @@ pub struct NapiNextTurbopackCallbacks {
...
+    //
+    // `Weak = true` so these ThreadsafeFunctions don't keep the Node.js event loop alive after
+    // shutdown.
+    throw_turbopack_internal_error: ThreadsafeFunction<
+        TurbopackInternalErrorOpts,
+        (),
+        TurbopackInternalErrorOpts,
+        Status,
+        true,
+        true,
```
bgw: Holy crimes against const generics, Batman!

This sort of thing would benefit a lot from `impl Trait`, but that's not valid in this position...

```suggestion
        /* CalleeHandled */ true,
        /* Weak */ true,
```

crates/next-napi-bindings/src/next_api/turbopack_ctx.rs
```diff
@@ -163,8 +171,18 @@ pub struct NapiNextTurbopackCallbacks {
...
+    >,
+    on_before_deferred_entries: Option<ThreadsafeFunction<(), Promise<()>, (), Status, true, true>>,
```
bgw:
```suggestion
    on_before_deferred_entries: Option<ThreadsafeFunction<(), Promise<()>, (), Status, /* CalleeHandled */ true, /* Weak */ true>>,
```

crates/next-napi-bindings/src/next_api/turbopack_ctx.rs
```diff
@@ -176,23 +194,35 @@ pub struct TurbopackInternalErrorOpts {
...
-        // after shutdown.
-        let _ = throw_turbopack_internal_error.unref(env);
+        let throw_turbopack_internal_error: ThreadsafeFunction<
+            TurbopackInternalErrorOpts,
+            (),
+            TurbopackInternalErrorOpts,
+            Status,
+            true,
+            true,
+        > = obj
```
bgw: Can we just let rust infer this?

```suggestion
        let throw_turbopack_internal_error = obj
```

crates/next-napi-bindings/src/next_api/turbopack_ctx.rs
```diff
@@ -176,23 +194,35 @@ pub struct TurbopackInternalErrorOpts {
...
+                // PII-containing message in anonymized telemetry.
+                Ok(ctx.value)
+            })?;
 
         let on_before_deferred_entries = obj
             .on_before_deferred_entries
             .map(|callback| {
-                let mut f = callback.create_threadsafe_function(0, |_| Ok::<Vec<()>, _>(vec![]))?;
-                let _ = f.unref(env);
+                let f: ThreadsafeFunction<(), Promise<()>, (), Status, true, true> = callback
```
bgw: Assuming this can be inferred

```suggestion
                let f = callback
```

crates/next-napi-bindings/src/lib.rs
```diff
@@ -115,6 +115,15 @@ fn init() {
...
     create_custom_tokio_runtime(rt);
+
+    // napi v2 permanently entered its tokio runtime context on the addon's main thread. Both
+    // these bindings and turbo-tasks (e.g. `PriorityRunner`) schedule tokio work from
+    // synchronous N-API calls and rely on that ambient context. napi v3 no longer provides it,
+    // so restore it: capture the runtime handle (this also forces napi to adopt the custom
+    // runtime registered above) and enter it for the lifetime of this thread.
+    let handle =
+        napi::bindgen_prelude::within_runtime_if_available(tokio::runtime::Handle::current);
+    std::mem::forget(Box::leak(Box::new(handle)).enter());
```
bgw, line 130: Looks like this prevents `tokio` from ever shutting down? That seems undesirable? I think I can understand why we might need this, but we should probably fix places that depend on this? That's probably out of scope for this PR, but in a follow-up?

packages/next/src/build/swc/types.ts
```diff
@@ -63,8 +63,10 @@ export interface Binding {
   teardownTraceSubscriber?(guardExternal: ExternalObject<RefCell>): void
   css: {
     lightning: {
-      transform(transformOptions: any): Promise<any>
-      transformStyleAttr(transformAttrOptions: any): Promise<any>
+      // Synchronous in the native bindings; callers `await` the result, which
+      // also tolerates Promise-returning custom implementations.
+      transform(transformOptions: any): any
+      transformStyleAttr(transformAttrOptions: any): any
```
bgw: Uh, if these have always been actually synchronous, we should just update the callsites to not `await` them.

turbopack/crates/turbopack-node/src/worker_pool/worker_thread.rs
```diff
@@ -18,43 +19,34 @@ use crate::worker_pool::{
     operation::{TaskMessage, WORKER_POOL_OPERATION},
 };
 
-static WORKER_CREATOR: OnceLock<ThreadsafeFunction<NapiWorkerCreation, ErrorStrategy::Fatal>> =
-    OnceLock::new();
+// napi v3 replaced the `ErrorStrategy` type parameter with a `CalleeHandled`
+// const generic. `CalleeHandled = false` is the former `ErrorStrategy::Fatal`.
+// `Weak = true` so these functions don't keep the Node.js event loop alive; they arrive from JS
+// already unref'd.
+type FatalThreadsafeFunction<T> = ThreadsafeFunction<T, Unknown<'static>, T, Status, false, true>;
```
bgw:
```suggestion
type FatalThreadsafeFunction<T> = ThreadsafeFunction<T, Unknown<'static>, T, Status, /* CalleeHandled */ false, /* Weak */ true>;
```

turbopack/crates/turbopack-node/src/worker_pool/worker_thread.rs
```diff
@@ -18,43 +19,34 @@ use crate::worker_pool::{
     operation::{TaskMessage, WORKER_POOL_OPERATION},
 };
 
-static WORKER_CREATOR: OnceLock<ThreadsafeFunction<NapiWorkerCreation, ErrorStrategy::Fatal>> =
-    OnceLock::new();
+// napi v3 replaced the `ErrorStrategy` type parameter with a `CalleeHandled`
+// const generic. `CalleeHandled = false` is the former `ErrorStrategy::Fatal`.
+// `Weak = true` so these functions don't keep the Node.js event loop alive; they arrive from JS
+// already unref'd.
```
bgw: I don't need the full history of the codebase in the comments, thanks Claude.

```suggestion
```

package.json
```diff
@@ -308,6 +308,8 @@
   "packageManager": "pnpm@10.33.0",
   "pnpm": {
     "overrides": {
+      "@emnapi/core": "1.11.1",
+      "@emnapi/runtime": "1.11.1",
```
bgw: Huh? Why is this needed?

bgw (approved):

bgw (approved):

bgw:
Maybe hold off on merging this until we figure out if 16.3.x is going to get cut from `canary` or from the `next-16-3` release branch. https://vercel.slack.com/archives/C02CDC2ALJH/p1786483656374079?thread_ts=1786372888.906399&cid=C02CDC2ALJH

PR #95439, opened 2026-07-02
Fix stale data after navigation despite revalidation

gaearon:
Claude noticed this while working on https://github.com/vercel/next.js/pull/95391.

Basically, the queue's React state always gets updated in the order of dispatching. So if a navigation displaced pending actions, it's the *navigation's* promise that was last. So when the pending action runs after it, that action's state is no longer taken into account, even if it revalidated. This is why the new data won't reflect on the screen.

With the fix, the final state is properly reflected.

## Test Plan

New test, which was red before the fix.

Also tried it on a mini-app.

Before:

https://github.com/user-attachments/assets/2235b116-82c4-4ef4-a63c-78e157c90470

After:

https://github.com/user-attachments/assets/d7b685c7-d0a5-4b91-8da7-78bb14011313

## How

We re-render at the end with the final queue state — if we know it was preempted.

### DIFF
--- a/packages/next/src/client/components/app-router-instance.ts
+++ b/packages/next/src/client/components/app-router-instance.ts
@@ -49,6 +49,7 @@ export type AppRouterActionQueue = {
 
   pending: ActionQueueNode | null
   needsRefresh?: boolean
+  wasPreempted?: boolean
   last: ActionQueueNode | null
 }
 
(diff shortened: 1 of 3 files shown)


packages/next/src/client/components/app-router-instance.ts
```diff
@@ -86,11 +87,23 @@ function runRemainingActions(
...
-    // action that revalidated data.
-    actionQueue.needsRefresh = false
-    actionQueue.dispatch({ type: ACTION_REFRESH }, setState)
+  if (actionQueue.pending === null) {
+    if (actionQueue.wasPreempted) {
+      // React renders the state promise from the latest dispatch — the
+      // navigation's, since it jumped ahead of actions that were already
+      // waiting in the queue. Those actions ran after it, so the states they
+      // produced were never rendered. Now that the queue is idle, render the
+      // state we ended up with.
```
ztanner: I’d maybe phrase this in terms of the queue invariant, to decouple the explanation from a specific navigation example

> When an action is preempted, later actions can update the queue’s state without React rendering it. Once the queue is empty, publish the final state so the UI catches up.

ztanner (approved):

PR #90901, opened 2026-03-04
Add experimental.lightningCssFeatures config option

sokra:
### What?

Adds a new `experimental.lightningCssFeatures` config option that lets users control which CSS features lightningcss should always transpile (`include`) or never transpile (`exclude`), regardless of browserslist targets.

```js
// next.config.js
module.exports = {
  experimental: {
    useLightningcss: true,
    lightningCssFeatures: {
      include: ['light-dark', 'oklab-colors'],
      exclude: ['nesting'],
    },
  },
}
```

### Why?

Currently, lightningcss feature transpilation is determined solely by browserslist targets. There's no way to force transpilation of a specific feature (e.g., `light-dark()`) when targeting modern browsers that already support it, or to skip transpilation of a feature that the user wants to preserve as-is.

This is useful for:
- Forcing polyfill-style transpilation for features with incomplete browser support
- Testing transpiled output for specific features
- Opting out of specific transforms that may interfere with CSS tooling downstream

### How?

**TypeScript config & validation:**
- `config-shared.ts`: `LIGHTNINGCSS_FEATURE_NAMES` const array (single source of truth) with `LightningCssFeature` type derived from it, and `LightningCssFeatures` interface
- `config-schema.ts`: Zod validation using `z.enum(LIGHTNINGCSS_FEATURE_NAMES)` for both include/exclude arrays

**Feature name → bitmask mapping (Rust, shared via NAPI):**
- `crates/next-core/src/next_config.rs`: `lightningcss_feature_names_to_mask()` maps feature name strings to `lightningcss::targets::Features` bitflag constants and returns a `Result<u32>` bitmask. Unknown feature names produce an error via `bail!`.
- Exposed to JS via NAPI as `lightningcssFeatureNamesToMaskNapi`, callable through `bindings.css.lightning.featureNamesToMask(names)`
- Single source of truth for feature name resolution — both webpack and Turbopack paths use this Rust function

**Webpack path:**
- `global.ts` / `modules.ts` pass config through to loader options
- `loader.ts` calls `featureNamesToMask()` via the native bindings to compute include/exclude masks, then passes them to the SWC `transform()` call
- The `lightningCssFeatures` without `useLightningcss` warning only shows when using webpack (Turbopack always uses lightningcss)

**Turbopack path (Rust):**
- `next_config.rs`: `LightningCssFeatures` struct for deserialization, accessor methods converting names → u32 bitmask via `lightningcss_feature_names_to_mask()`
[...]

PR #95444, opened 2026-07-03
Add experimental.lightningCssModules config option

UditDewan:
### What?

Adds a new experimental config option, `experimental.lightningCssModules`, that exposes Lightning CSS's [CSS Modules options](https://lightningcss.dev/css-modules.html) when bundling with Turbopack:

```ts
// next.config.ts
const nextConfig = {
  experimental: {
    lightningCssModules: {
      // custom naming pattern for scoped class names
      pattern: 'my-company-[hash]-[local]',
      // keep authored @keyframes names instead of scoping them
      animation: false,
    },
  },
}
```

Supported options: `pattern` ([custom naming patterns](https://lightningcss.dev/css-modules.html#custom-naming-patterns)) and the feature-scoping toggles `animation`, `grid`, `customIdents`, `dashedIdents`, `container` ([turning off feature scoping](https://lightningcss.dev/css-modules.html#turning-off-feature-scoping)).

### Why?

Turbopack always processes CSS Modules with Lightning CSS using a hardcoded configuration (`[name]__[hash]__[local]` pattern, animation scoping on). Users repeatedly asked in discussion #64580 for a way to:

- specify the CSS Modules class naming pattern / hash names
- disable feature scoping, particularly `@keyframes` renaming, to match webpack's `css-loader` behavior when migrating to Turbopack

`experimental.lightningCssFeatures` (#90901) only controls feature *transpilation* flags and does not cover CSS Modules options, so there is currently no escape hatch at all on Turbopack.

### How?

- New `CssModulesOptions` struct in `turbopack-css`, threaded `NextConfig` → `CssOptionsContext` → `ModuleType::Css` → `CssModule` → `parse_css`/`process_content`, following the same path as `lightningcss_features` (#90901).
- All fields are optional; unset fields fall back to the previously hardcoded values (`[name]__[hash]__[local]`, `animation: true`, `grid: false`, `customIdents: true`, `dashedIdents: false`, `container: false`), so behavior is unchanged unless configured.
- A custom `pattern` is validated once at config level so a typo produces a single clear error naming the config key instead of one error per CSS file. The pattern string is interned so the parsed `Pattern` can borrow into the `'static` parser options stored in `ParseCssResult`.
- webpack does not use Lightning CSS for CSS Modules name scoping (`css-loader` handles that and is already configurable), so this option is Turbopack-only; a warning is logged when it is set with another bundler.
- Turbopack-only e2e test covering the custom pattern and `animation: false`, plus documentation in `useLightningcss.mdx`.

Addresses the CSS Modules configuration requests in #64580.


puopg:
🙌

PR #95451, opened 2026-07-03
Fix standalone output symlinks resolving outside `.next/standalone` on Windows with pnpm

### LINKED ISSUE #95450: Windows - Next.js standalone output (pnpm) may contain symlinks resolving outside `.next/standalone`

https://github.com/timheerwagen/next-standalone-pnpm-symlink-repro

### To Reproduce

`pnpm install`
`pnpm build`
`pnpm test` - The script shows the resolved path of the standalone Next Node module.

When using Next.js with output: "standalone" together with pnpm, the generated standalone output contains symlinks inside `.next/standalone/node_modules` that resolve to the root `node_modules/.pnpm` dir instead of the `.next/standalone/node_modules/.pnpm` directory.

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Home
  Available memory (MB): 16294
  Available CPU cores: 12
Binaries:
  Node: 24.11.1
  npm: 11.16.0
  Yarn: N/A
  pnpm: 10.23.0
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: N/A
Next.js Config:
  output: standalone
```

Module Resolution

next build (local)

### Additional context

This behaviour only occurs on Windows and not on Linux, as demonstrated here: https://github.com/timheerwagen/next-standalone-pnpm-symlink-repro/actions/runs/28657904104/job/84991341147.

heyparth1:
When using Next.js with `output: "standalone"` together with pnpm on Windows, the generated standalone output contains symlinks inside `.next/standalone/node_modules` that resolve to the root `node_modules/.pnpm` directory instead of the standalone directory. This happens because pnpm on Windows installs packages as junctions with absolute targets.

This PR fixes the issue by remapping absolute symlink targets that resolve inside the tracing root to their equivalent path in the standalone output during the `copyTracedFiles` process. It also explicitly sets the symlink type based on the original target's stat to handle cases where the remapped target hasn't been copied yet. A regression test has been added to verify the fix.

Closes #95450

### DIFF
--- a/packages/next/src/build/utils.ts
+++ b/packages/next/src/build/utils.ts
@@ -1291,8 +1291,31 @@ export async function copyTracedFiles(
           const symlink = await fs.readlink(tracedFilePath).catch(() => null)
 
           if (symlink) {
+            // pnpm on Windows installs packages as junctions with absolute
+            // targets (e.g. <project>\node_modules\.pnpm\...). Remap absolute
+            // targets that resolve inside the tracing root to the equivalent
+            // path in the standalone output so the copied link doesn't
+            // resolve back into the original node_modules.
+            let symlinkTarget = symlink
+            if (path.isAbsolute(symlink)) {
+              const relativeTarget = path.relative(tracingRoot, symlink)
+              if (
+                !relativeTarget.startsWith('..') &&
+                !path.isAbsolute(relativeTarget)
+              ) {
+                symlinkTarget = path.join(outputPath, relativeTarget)
+              }
+            }
+            // A remapped target may not have been copied yet, so Windows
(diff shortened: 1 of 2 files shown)


bgw:
I'm making pretty significant changes to how symlinks are processed in https://github.com/vercel/next.js/pull/98003. There's a good chance that that PR ends up fixing this issue by making Turbopack resolve the symlinks (which already has a better view of this information) and store those resolved targets in the `*.nft.json` files, rather than trying to convert symlink targets to relative paths in JS.

I'll take another look at this PR/issue after that's done.

PR #95461, opened 2026-07-03
Add lightningcss CSS-modules passthrough config for Turbopack

puopg:
Addresses:
- https://github.com/vercel/next.js/discussions/64580
- https://github.com/vercel/next.js/discussions/85869

## Summary

Adds an experimental config surface for customizing how Turbopack invokes lightningcss. Two things Turbopack previously did not expose:

1. **CSS-modules class names** — the class-name pattern was hard-coded as `[name]__[hash]__[local]`.
2. **A single passthrough home** — feature flags now live under `lightningCss` alongside CSS-modules config.

```js
// next.config.js
module.exports = {
  experimental: {
    lightningCss: {
      features: { include: ["nesting"], exclude: [] },
      cssModules: {
        pattern: "[name]__[hash]__[local]", // lightningcss placeholder syntax
        dashedIdents: false,
      },
    },
  },
}
```

- `cssModules.pattern` accepts lightningcss placeholders (`[name]`, `[local]`, `[hash]`, `[content-hash]`) plus literals, validated at config load (invalid values fail the build with a clear message).
- `lightningCss.features` (include/exclude) **takes precedence over the legacy top-level `experimental.lightningCssFeatures`** when both are set; `lightningCssFeatures` is kept for backwards compatibility. A warning is emitted if both are set.
- Turbopack-only for now. `useLightningcss` is unchanged — it remains the webpack-only switch (in Turbopack lightningcss always processes CSS, so it is a no-op there). A warning is emitted if `lightningCss` is set with the webpack bundler. The `lightningCss` object leaves room for webpack parity and broader passthrough (drafts / nonStandard / pseudoClasses) later.

Internally, the existing `LightningCssFeatureFlags` becomes the `features` field of a new `LightningCssOptions` (features + CSS-modules config) threaded through the CSS module pipeline.

## Backwards compatibility

Fully backwards compatible. The config is additive and optional at every layer (`Option` in Rust, `.optional()` in the zod schema). When `experimental.lightningCss` is unset, the code falls back to the exact previous defaults — the hard-coded `[name]__[hash]__[local]` pattern, `dashedIdents: false`, the same `grid`/`container` settings, and the legacy `lightningCssFeatures` for feature flags — so existing projects produce byte-identical class names and CSS output. No migration needed.

## Verification

- `cargo check -p turbopack-css -p turbopack -p next-core` — clean
- `pnpm build-all` (JS + native) — succeeds
- `pnpm test-dev-turbo test/e2e/app-dir/experimental-lightningcss-css-modules/...` — passes (custom pattern renders `.box` as `box__custom__<hash>`)
- `pnpm test-dev-turbo test/e2e/app-dir/experimental-lightningcss-features/...` — passes, including a new case asserting `lightningCss.features` overrides `lightningCssFeatures`
- Backwards-compat check on a real app (~176 `.module.scss` files) with no config: class names remain the default `[name]__[hash]__[local]` (e.g. `styles-module-scss-module__jMmjOa__root`), unchanged from before

## Verification screenshots
Screenshot 2026-07-03 at 9 59 37 AM
Screenshot 2026-07-03 at 9 59 10 AM

### DIFF
--- a/crates/next-core/src/next_client/context.rs
+++ b/crates/next-core/src/next_client/context.rs
@@ -373,7 +373,7 @@ pub async fn get_client_module_options_context(
         css: CssOptionsContext {
             source_maps,
             module_css_condition: Some(module_styles_rule_condition()),
-            lightningcss_features: *next_config.lightningcss_feature_flags().await?,
+            lightningcss: next_config.lightningcss_options().owned().await?,
             ..Default::default()
         },
         static_url_tag: Some(rcstr!("client")),
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -1415,6 +1415,8 @@ pub struct ExperimentalConfig {
     // turbopack_file_system_cache_for_dev: Option<bool>,
     // turbopack_file_system_cache_for_build: Option<bool>,
     lightning_css_features: Option<LightningCssFeatures>,
+    /// Passthrough configuration for lightningcss (Turbopack only for now).
+    lightning_css: Option<LightningCss>,
 }
 
 #[derive(
@@ -1451,6 +1453,52 @@ pub struct SubResourceIntegrity {
 pub struct LightningCssFeatures {
     pub include: Option<Vec<RcStr>>,
     pub exclude: Option<Vec<RcStr>>,
+}
+
+/// Passthrough configuration for lightningcss, mirroring the subset of
+/// lightningcss options that Next.js exposes via `experimental.lightningCss`.
+#[derive(
+    Clone,
+    Debug,
+    Default,
+    PartialEq,
+    Eq,
+    Deserialize,
+    TraceRawVcs,
+    NonLocalValue,
+    OperationValue,
+    Encode,
+    Decode,
+)]
+#[serde(rename_all = "camelCase")]
+pub struct LightningCss {
+    /// Feature `include`/`exclude` lists. Takes precedence over the top-level
+    /// `experimental.lightningCssFeatures` when set.
+    pub features: Option<LightningCssFeatures>,
+    /// CSS-modules configuration (class-name renaming).
+    pub css_modules: Option<LightningCssModules>,
+}
+
+/// User-facing subset of lightningcss' `css_modules::Config`.
+#[derive(
+    Clone,
+    Debug,
+    Default,
+    PartialEq,
+    Eq,
+    Deserialize,
+    TraceRawVcs,
+    NonLocalValue,
+    OperationValue,
+    Encode,
+    Decode,
+)]
+#[serde(rename_all = "camelCase")]
+pub struct LightningCssModules {
+    /// Custom class-name pattern, e.g. `[name]__[hash]__[local]`.
+    pub pattern: Option<RcStr>,
+    /// Whether to rename dashed identifiers (e.g. CSS custom properties).
+    pub dashed_idents: Option<bool>,
 }
 
 #[derive(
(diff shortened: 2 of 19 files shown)


Belyash:
This would unblock a migration on our end as well. We have a lot of `.module.scss` files generating classes like `styles-module-scss-module__<hash>__root` in Turbopack. The duplicated `scss-module` suffix comes from the `rename_as: "*.module.css"` + `file_stem().replace('.', '-')` interaction, and without this PR, the only way I'm aware of to clean this up is forking — happy to be corrected if there's a config option I missed.

A `[hash]__[local]` (or similar) pattern would let us keep the readable `[local]` keys in `import styles from ...` while producing ~10-char class names in the bundle, which is a big readability win in DevTools. Glad to test against canary once it's mergeable.

x5hwuk7nc6jsudcc:
Thanks for putting this up @puopg - this is the config surface we need, and it's the only route we found that doesn't force a much larger tradeoff (detail below).

@Belyash's diagnosis matches what we see exactly: the duplicated `scss-module` segment comes from the `rename_as: "*.module.css"` + `file_stem().replace('.', '-')` interaction, so a `.module.scss` file pays for it twice. Every one of our classes looks like this:

```
Header-module-scss-module__wtz7yW__c-header__nav-link
└──────── 38 bytes of machine ─────────┘└─ 17 of ours ─┘
```

Adding production numbers, since the thread so far is mostly qualitative. This is a static-export (`output: 'export'`) content site on Next 16.2.11, ~113 `.module.scss` files, 170 prerendered pages:

```
raw HTML:        34.99 MB       brotli: 3.34 MB
visible text:     2.28 MB   →   text/HTML 6.5%
RSC flight:      20.93 MB (59.8%)   - required for hydration
class prefixes:   8.88 MB (25.4%)   - 177,589 occurrences
  if emitted as [local]:  raw −8.88 MB, brotli −235 KB, text/HTML → 8.7%
```

Two things that aren't obvious until you measure:

1. **The prefix is paid twice per element.** Once in `class=""`, and again escaped inside the inline RSC flight payload as `className\":\"…\"`. On our homepage, **41.6% of the flight payload is classname strings** (65.5 KB of 157 KB). So in App Router the cost is roughly double what inspecting the DOM suggests.
2. **Compression absorbs a lot of it**, so the honest headline isn't bandwidth - it's raw HTML weight, parse cost, and token cost for LLM crawlers that fetch un-rendered HTML. Still, 25% of raw HTML being bundler bookkeeping is hard to justify when `[local]` would be correct output for us.

This lines up with what @ik-southpole reported in #85869 ("our stylesheets blew up in size because of all the metadata"), and @FeelGoodInc's point there is worth repeating on the PR: it isn't only length. Because the hash sits *before* the local name, `[class^="ui-table_root"]`-style selectors are simply unwritable now - a `pattern` config fixes that class of breakage too, not just the bytes.

**Why we think this config is the only real fix, rather than a nice-to-have.** We tested the workarounds against a real Turbopack build instead of reasoning about them, and each one removes `composes` - which our CSS architecture depends on (57 of 113 modules, 176 declarations):

| Attempt | Result |
|---|---|
| `composes` on a `:global()` selector | Turbopack **panic**: `The 'composes' property cannot be used with a simple class selector` |
| Sass `&__el` nesting inside `:global()` | Throws in the Sass compiler (can't suffix a pseudo-selector) |
| `:global { … }` block form | `Ambiguous CSS module class not supported` |
| `turbopack.rules` + `as: '*.css'` | Works, emits unprefixed CSS - but disables CSS Modules entirely; the `styles` object comes back empty |
| `@nimpl/classnames-minifier` | `peerDependencies: {"css-loader": ">=4.0.0"}` - webpack-only |

So `cssModules.pattern` is the only path that leaves CSS Modules fully intact and just renames. Reverting to webpack isn't neutral for us either - we moved the production build to Turbopack specifically to fix a bundle-size regression.

Happy to build a canary against our repo and report back once this is mergeable (it's showing conflicts against `canary` at the moment). We'd set `pattern: "[local]"`; we've already audited our 1,285 local names for cross-module collisions and added a build gate so that stays true.

@icyJoseph - you weighed in on the webpack-era version of this in #76887; is there a current owner for Turbopack's CSS Modules path who could take a look? This PR has been open since 3 July with no triage, and it's the third thread on the same request (#76887, #85869).

PR #95468, opened 2026-07-03
Fix top-level await tracking for routes and metadata

gaearon:
This adds a failing test and Claude's fix for a problem I encountered on my blog.

The issue was that await readFile with a big file at the top level, as suggested in [vercel/next.js#95246](https://github.com/vercel/next.js/pull/95246) for fonts in opengraph-image, actually subtly breaks the cache and prints nonsensical error messages about unstable cache key.

Claude then discovered three separate problems:

- Cache warming didn't wait for the metadata image module with TLA, so the cache didn't actually get warmed yet by the time of the render, and this caused a false positive cache key warning (it thought the *wrong* cache was warmed).
- The segmentor collector was reading exports before the modules were ready. So it didn't "find" `generateStaticParams` if the module was still waiting ro TLA, and the route silently became static.
- Errors during init in `output: 'export'` mode were incorrectly reported as unhandled rejections instead of failing the build.

This is supposed to fix these, with regression tests.

## Test Plan

New tests. They fail before the fixes.


packages/next/src/server/app-render/module-loading/track-module-loading.instance.ts
```diff
@@ -4,9 +4,15 @@ import { isThenable } from '../../../shared/lib/is-thenable'
...
+ *
+ * Returns null in the edge runtime: `CacheSignal` can't be constructed there,
+ * and its only consumer (`trackPendingModules`) has no edge callers.
  */
 let _moduleLoadingSignal: CacheSignal | null
-function getModuleLoadingSignal() {
+function getModuleLoadingSignal(): CacheSignal | null {
+  if (process.env.NEXT_RUNTIME === 'edge') {
+    return null
+  }
```
lubieowoce: hm. i don't love making this nullable. none of this should run in edge ever, so it'd be better if this threw an `InvariantError` if called there, and then we do a conditional `require` of this module in any callers that might run in edge

the `its only consumer (`trackPendingModules`) has no edge callers` assertion is kinda baseless, it's just what the robot saw at a point in time, not a guarantee, i'd drop this comment

(also, maybe my knowledge here is outdated, but IIRC our DCE doesn't like early returns, you gotta do an else instead. idk why)

crates/next-core/src/app_page_loader_tree.rs
```diff
@@ -216,11 +216,10 @@ impl AppPageLoaderTreeBuilder {
...
                     .insert(inner_module_id.into(), module);
 
+                // The module is emitted as a loader (instead of a bound call) so
+                // that its evaluation can be tracked. See `MetadataImageEntry`.
                 let s = "      ";
-                writeln!(
-                    self.loader_tree_code,
-                    "{s}async (props) => interopDefault(await {identifier}())(props),"
-                )?;
+                writeln!(self.loader_tree_code, "{s}{{ loadModule: {identifier} }},")?;
```
lubieowoce: hm. i wonder if it wouldn't make more sense to just codegen a wrapper that does `trackPendingImport` around this rather than having to do it in every place that loads a module? it's easy to add another callsite that loads things and forget a `trackPendingImport` for it

packages/next/src/server/route-modules/app-route/module.ts
```diff
@@ -235,6 +235,9 @@ export class AppRouteRouteModule extends RouteModule<
   // Non-null while an async userland module (top-level await) is loading.
   // ensureUserland() awaits this before the first request is handled.
   private _pendingUserland: Promise<void> | null = null
+  // Init error from an async userland module, rethrown by ensureUserland().
+  private _initError: unknown = null
+  private _hasInitError = false
```
lubieowoce: nit: don't love this. now we have 4 independently varying properties to represent what's happening with the userland. something like react's `promise.status` and `promise.value`/`promise.reason` would be cleaner although we don't have to use that exact shape. some kinda discriminated union thing would be ideal (including the `_userlandFactory` thing) bc really it's a state machine that represents potentially async loading

packages/next/src/build/segment-config/app/app-segments.ts
```diff
@@ -141,9 +150,13 @@ async function collectAppPageSegments(routeModule: AppPageRouteModule) {
...
  * @returns the segments for the app route module
  */
-function collectAppRouteSegments(
+async function collectAppRouteSegments(
   routeModule: AppRouteRouteModule
-): AppSegment[] {
+): Promise<AppSegment[]> {
+  // The route file may be an async module (top-level await), so the userland
+  // module must be resolved before its exports can be inspected.
+  await routeModule.ensureUserland()
```
lubieowoce, line 149: hm. i wonder if we wanna make the `routeModule.userland` getter throw an InvariantError if the module is async and wasn't ensured beforehand, who knows where else we're not checking this

really ideally it shouldn't be possible to access this without having it properly initialized, i guess this is a problem with it being a class because constructors can't be async or sth

crates/next-core/src/app_page_loader_tree.rs
lubieowoce: FWIW i'd do this via a helper function like
```ts
type ModuleGetter = () => unknown;

export function instrumentModuleGetter(getter: ModuleGetter): ModuleGetter {
  if (
    process.env.NEXT_RUNTIME === "edge" ||
    !process.env.__NEXT_CACHE_COMPONENTS
  ) {
    return getter;
  } else {
    return () => {
      const {
        trackPendingImport,
      } = require(/* track-module-loading.external, possibly aliased */);
      const result = getter();
      trackPendingImport(result);
      return result;
    };
  }
}

```
and import it in the loader tree code. avoid codegening all of this inline, it sucks to read and won't be typechecked

lubieowoce: (typechecking is useful, as evidenced by the fact that it took me like 3 tries to write `instrumentModuleGetter` correctly without it)

gaearon:
OK this is maybe worth another look, tried to address the review.

packages/next/src/server/app-render/module-loading/instrument-module-getter.ts
```diff
@@ -0,0 +1,34 @@
...
+      const { trackPendingImport } =
+        require('./track-module-loading.external') as typeof import('./track-module-loading.external')
+
+      const exportsOrPromise = getter()
+
+      // Not a thenable check — a module may legitimately export a function
+      // named `then`. An async module yields a real promise.
+      if (exportsOrPromise instanceof Promise) {
+        trackPendingImport(exportsOrPromise)
+      }
```
lubieowoce: hm i'm not sure if we actually care about a module exporting `then`. this is meant to instrument a next.js page or route (not an arbitrary module), and `then` is not a supported export for those. we don't error for that rn but have plans to

OTOH if we really really do want to support this for some reason, it feels like this change should be done inside `trackPendingImport` itself which uses `isThenable`

lubieowoce: also worth checking if this works for client reference proxies (which we can get here for a `"use client"` page). i have a comment about that in `trackPendingImport` but i can't remember the details, i guess the comment wasn't detailed enough...

test/e2e/app-dir/use-cache-og-image-top-level-await/og/gate.ts
```diff
@@ -0,0 +1,25 @@
...
+import { join } from 'node:path'
+
+// Blocks module evaluation until the test opens the corresponding gate file.
+// Gates are only armed when the test that orchestrates its own build enables
+// them; by default (dev, deployments) this module resolves immediately and
+// the fixture is a plain app with a top-level await. Gates are per-process,
+// so that every evaluation of the module (during page data collection, and
+// in each prerender worker) stalls until the test allows it to proceed.
+// Deliberately avoids timers, which are treated as runtime data during
+// prerendering.
```
lubieowoce: this doesn't sound right? timers behave like dynamic data if anything, they wouldn't resolve in the runtime stage (and also there's no runtime stage in static prerenders)

but also this is a TLA and not in a component so why does this matter at all? `for await (const _event of watch(process.cwd())) { ... }` is definitely IO anyway

test/e2e/app-dir/use-cache-og-image-top-level-await/use-cache-og-image-top-level-await.test.ts
```diff
@@ -0,0 +1,93 @@
...
+            buildSettled = true
+          })
+
+        // Open each gate only after its module evaluation has been waiting for
+        // a while, so that all other work of the build has settled by then. In
+        // a correct build, the prerender waits for the pending module
+        // evaluation, no matter how long the gate stays closed. In a broken
+        // build, the prerender concludes without the module evaluation ever
+        // finishing, and fails on its own.
+        const deadline = Date.now() + 5 * 60 * 1000
```
lubieowoce: why are we waiting 5 minutes for something. if the build hung for 5 minutes then jest would kill the test

(also there's a NEXT_E2E_TEST_TIMEOUT that's set to `240_000`ms so the ten minute thing below won't work, but why do we need it to work?)

packages/next/src/server/app-render/module-loading/instrument-module-getter.ts
gaearon: right but without this it would silently start hanging if someone does have an export like this. if we don't disallow arbitrary exports, seems like we should guard? i'll see about moving the check

gaearon:
this was merged via stack with https://github.com/vercel/next.js/pull/95791 as the tip

PR #95471, opened 2026-07-03
fix(create-next-app): render filenames in Geist Mono so the preloaded font is used

aurorascharff:
## What?

Render the filename in the "edit the … file" line using **Geist Mono** across all `create-next-app` templates, so the preloaded font is actually used.

## Why?

Every fresh app logged a browser warning: the Geist Mono `.woff2` is preloaded but never used, because the redesign in #81945 dropped the last monospace element (the old `<code>app/page.tsx</code>`). Rather than stop preloading the font (this PR's original approach), we _use_ it — the filename renders in Geist Mono. That clears the warning at its source and keeps the template showcasing both flagship fonts (two fonts, no flash of unstyled text).

## How?

- Reverted the `preload: false` change.
- Wrapped the filename in the intro line in a `<code>` rendered in Geist Mono, styled as a small inline code chip (slightly smaller size, subtle rounded background, padding) so it reads as a code token rather than plain heading text:
  - Tailwind templates → utility classes (`font-mono` maps to `--font-geist-mono` via `@theme inline`).
  - CSS-module templates → a `.code` class using `var(--font-geist-mono)`; the chip background uses `color-mix(currentColor …)` so it adapts to light/dark automatically.
- Applied across all four families (`app`, `app-tw`, `default`, `default-tw`), JS + TS.

## Screenshots

|           | Before                                              | After                                        |
| --------- | --------------------------------------------------- | -------------------------------------------- |
| **Light** | cna-geist-mono-before              | cna-geist-mono-after     |
| **Dark**  | cna-geist-mono-before-dark           | cna-geist-mono-after-dark    |

Before: filename in Geist Sans, preloaded-but-unused warning in the console. After: filename in Geist Mono, no warning.


gaearon:
If templates don't use it, why not just remove it?

aurorascharff:
> If templates don't use it, why not just remove it?

was thinking about this, not sure if is intentional, need to investiage

feedthejim (approved):

icyJoseph (approved):

PR #95486, opened 2026-07-04
Cancel a superseded Server Components HMR refresh's server-side work

unstubbable:
When rapid edits overlap Server Components HMR refreshes in `next dev`, only the newest refresh can commit. The client already aborts a superseded refresh's fetch, which closes its response. We use that response-close to stop the server work the superseded refresh started, so the dev server no longer runs a render (and, under Cache Components, a validation) whose result is discarded.

When the `serverComponentsHmrCancellation` flag is enabled and the request is an HMR refresh over a Node response, we derive an abort signal from `signalFromNodeResponse(ctx.res.originalResponse)` and thread it into the Flight render. This covers both dev RSC render paths: the Cache Components staged render in `generateDynamicFlightRenderResultWithStagesInDev` and the non-Cache-Components render in `generateDynamicFlightRenderResult`.

The render is aborted through `renderToNodeFlightStream`, which calls `abort()` on the pipeable that `renderToPipeableStream` returns when the signal fires. React's Node Flight API has no `signal` option, unlike the Web `renderToReadableStream`, aborting the returned pipeable is the intended way to stop the render. This also removes the incorrect `signal` field from the `renderToPipeableStream` type declaration and derives `FlightRenderOptions` from `renderToReadableStream`'s options, so the render wrappers are no longer typed as `any`.

Under Cache Components the superseded refresh additionally skips its detached validation, which only runs on the staged render, so it never validates a discarded tree, and any error thrown while its background renders tear down is swallowed. The aborted Flight renders surface as abort errors that `createReactServerErrorHandler` already ignores, so nothing reaches the error overlay or the CLI.

The `hmr-rsc-cancellation` suite exercises both render paths, with Cache Components controlled by the `__NEXT_CACHE_COMPONENTS` test shard rather than the fixture config. A child component logs when React renders it, and the test asserts the superseded refresh never logs, since its render is aborted before reaching the child, while the committed refresh does. The detached-validation assertions run only when Cache Components is enabled.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -174,7 +174,8 @@ import {
   getClientComponentLoaderMetrics,
   wrapClientComponentLoader,
 } from '../client-component-renderer-logger'
-import { isNodeNextRequest } from '../base-http/helpers'
+import { isNodeNextRequest, isNodeNextResponse } from '../base-http/helpers'
+import { signalFromNodeResponse } from '../web/spec-extension/adapters/next-request'
 import {
   parseRelativeUrl,
   type ParsedRelativeUrl,
@@ -843,6 +844,19 @@ async function generateDynamicFlightRenderResult(
     onFlightDataRenderError
   )
 
+  // With Server Components HMR cancellation enabled, a superseded HMR refresh
+  // aborts its own client fetch, which closes this response. We use that
+  // response-close to abort the discarded render. This is the
+  // non-Cache-Components dev RSC path; unlike the Cache Components staged path
+  // it has no detached validation, so the render is the only work to cancel.
+  const abortSignal =
+    process.env.__NEXT_DEV_SERVER &&
+    renderOpts.experimental.serverComponentsHmrCancellation === true &&
+    requestStore.isHmrRefresh === true &&
+    isNodeNextResponse(ctx.res)
+      ? signalFromNodeResponse(ctx.res.originalResponse)
+      : undefined
+
   if (process.env.__NEXT_USE_NODE_STREAMS) {
     const debugChannel = setReactDebugChannel && createNodeDebugChannel()
 
@@ -870,6 +884,7 @@ async function generateDynamicFlightRenderResult(
         temporaryReferences: options?.temporaryReferences,
         filterStackFrame,
         debugChannel: debugChannel?.serverSide,
+        signal: abortSignal,
       }
     )
 
@@ -905,6 +920,7 @@ async function generateDynamicFlightRenderResult(
         temporaryReferences: options?.temporaryReferences,
         filterStackFrame,
         debugChannel: debugChannel?.serverSide,
+        signal: abortSignal,
       }
     )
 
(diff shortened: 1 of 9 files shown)


packages/next/src/server/app-render/app-render.tsx
```diff
@@ -843,6 +844,19 @@ async function generateDynamicFlightRenderResult(
     onFlightDataRenderError
   )
 
+  // With Server Components HMR cancellation enabled, a superseded HMR refresh
+  // aborts its own client fetch, which closes this response. We use that
+  // response-close to abort the discarded render. This is the
+  // non-Cache-Components dev RSC path; unlike the Cache Components staged path
+  // it has no detached validation, so the render is the only work to cancel.
+  const abortSignal =
```
lubieowoce: can we name this something more descriptive, like `requestAbortSignal`? we have a lot of abort signals around this file, usually used for something related prerendering, it's useful to disambiguate

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -1378,6 +1394,34 @@ async function generateDynamicFlightRenderResultWithStagesInDev(
...
+    // The detached validation is Cache Components only: it runs only on this
+    // staged dev render, so there's nothing to skip on the non-Cache Components
+    // dev RSC path. That path (`generateDynamicFlightRenderResult`) aborts its
+    // superseded render the same way; it just has no validation to skip.
+    //
+    // TODO: The gate is `isHmrRefresh` for now because that's the only case we
+    // cancel today. The response-close signal itself is general, so this could
+    // later be relaxed to also cover a browser stop or a devtools "cancel
+    // render" button.
+    const abortSignal =
```
lubieowoce: same as https://github.com/vercel/next.js/pull/95486/changes#r3530340225

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -3475,6 +3520,8 @@ async function renderToStream(
               navigationKind: {
                 type: 'initial-load',
               },
+              // The initial load has no request-abort signal to tear it down.
```
lubieowoce:
```suggestion
              // We currently only abort HMR refresh requests.
```
we could very well have an abort signal for these, we're just currently choosing not to because for now we only want to abort HMR ones.

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -5061,13 +5149,15 @@ async function streamStagedRenderInDev({
...
-  // the response is the only reader outside this function.
+  // The render runs to completion unless `abortSignal` fires, in which case
+  // React aborts the Flight render and the accumulation tears down. The first
+  // task starts the render in the `ShellEarlyStatic` stage and creates the
+  // stream (one replay for the response, one to accumulate the chunks). The
+  // later tasks advance the stages, settle `hadCacheMiss`, and reveal the shell
+  // – as soon as a cache miss is seen, or once the render reaches
+  // `revealAfterStage` – then advance into the dynamic stage a task later. The
+  // replayable stays local: the response is the only reader outside this
+  // function.
```
lubieowoce: this comment feels pretty verbose and mostly unnecessary tbh, it's just restating what the code does with little extra value. i'd just delete this (in fact i thought i did, but it must've gotten lost somewhere).

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -5095,7 +5186,10 @@ async function streamStagedRenderInDev({
         accumulatedChunksPromise: accumulateStreamChunks(
           replayable.createReplayStream(),
           stageController,
-          null
+          // Tear down the chunk accumulation on abort too, so `resultPromise`
+          // settles promptly (instead of waiting out the aborted render) and
+          // the detached validation can see the abort and skip.
```
lubieowoce: verbose
```suggestion
          // Abort accumulation as soon as the signal aborts instead of waiting for the stream to close
```

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -5284,6 +5385,12 @@ async function prerenderWithWarmCachesForStaticValidationInDev(
   const finalReactController = new AbortController()
   const finalDataController = new AbortController()
 
+  // Aborting `finalReactController` (the prerender's own controller) on the
+  // signal tears the render down the same way its runtime-stage abort does.
```
lubieowoce: aborting the signal we pass to the render aborts the render. why does this need explanation?

packages/next/src/server/app-render/stream-ops.node.ts
```diff
@@ -536,21 +541,38 @@ export { renderToWebFlightStream } from './stream-ops.web'
...
+  opts: FlightRenderOptions
 ): AnyStream {
   if (!ComponentMod.renderToPipeableStream) {
     throw new Error('renderToPipeableStream is not implemented')
   }
 
+  // `renderToPipeableStream` has no `signal` option (unlike the Web
+  // `renderToReadableStream`), so pull `signal` out of the options and abort
+  // the returned pipeable ourselves when it fires. The listener is removed once
+  // the passthrough closes.
```
lubieowoce: > The listener is removed once the passthrough closes.

i can see that it is, is it important that we do this? maybe explain why then

lubieowoce (approved):

PR #95493, opened 2026-07-04
Normalize and validate `expire` and `revalidate` values to handle Infinity and surface mistakes early

gaearon:
Supersedes https://github.com/vercel/next.js/pull/93475

## Summary

There's a few fixes bundled in here.

We're fixing `Infinity` getting serialized to `null`, which caused cache handlers and build workers to misinterpret it as "don't cache" instead of "cache forever". We use the existing infinity sentinel for that.

We're also making the checks a bit more specific from `typeof !== 'number'` to forbidding `NaN` and `-Infinity`. The checks are applied consistently for profiles, inline cache life, and arguments to `fetch` and `unstable_cache`.

Now that configs are also validated, we need to make sure that we only do `revalidate > expire` validation when both values are specified. There's a regression test for that, which https://github.com/vercel/next.js/pull/93475 didn't do correctly.

Also, per review below:

- Now actually using returned value for `unstable_cache`
- Added validation + infinity fix for `revalidateTag` profile
- Added validation + infinity fix in `config.experimental.staleTimes` and `config.expireTime`
- Added deep clone for `defaultConfig.cacheLife`

## Test Plan

New e2e tests that fail before the fixes.


packages/next/src/server/lib/patch-fetch.ts
```diff
@@ -55,7 +55,9 @@ export function validateRevalidate(
   try {
     let normalizedRevalidate: number | undefined = undefined
 
-    if (revalidateVal === false) {
+    if (revalidateVal === false || revalidateVal === Infinity) {
+      // Unlike Infinity, INFINITE_CACHE survives JSON serialization (e.g. in
+      // the fetch cache).
```
gnoff, line 60: This may be a bit a drive by but this function is typed to return a new validated value but it's only actually used in this file and not in the two other places that do this validation

Unclear if we should be consuming the result everywhere if it doesn't throw or we should change it's type to not return anything to signal it's for throwing only.

In it's current form it doesn't seem suitable for converting Infinity to INFINITE_CACHE

gnoff: I guess we just need to adopt the modified revalidate for unstable_cache

packages/next/src/server/use-cache/cache-life-profile.ts
```diff
@@ -0,0 +1,96 @@
...
+  return value
+}
+
+export function validateAndNormalizeCacheLifeProfile(
+  profile: CacheLife,
+  context: CacheLifeProfileContext
+): CacheLife {
+  // Don't mutate the profile; it may be shared, e.g. as part of the user's
+  // config object.
+  const normalizedProfile = { ...profile }
```
gnoff, line 73: Also, in reviewing this PR with fable it found that the default config object can potentially be mutated and in a multi-next() custom server situation this coudl lead to unexpected results since one instance can mutate the process-local shared config that is then read by the other instance.

This clone here is protecting us from this in the current setup. Seems like we should tackle holistically by always cloning the shared config before merging it into the user config. It's not meant to be a mutable target anyway. Doesn't block this PR but maybe we can PR that next

gnoff (approved):
Seems like revalidateTag should also get this normalization for expire

Also the new config validation will start to fail builds. sorta breaking but I guess it's a bugfix so it's fine.

Seems like config.experimental.staleTimes and config.expireTime also don't consistently handle Infinity.

But we don't need to block on that.

gaearon:
Yea I feel like failing builds isn't the worst here since it happens upfront and is a bugfix. Will look at other issues

Issue #95496, opened 2026-07-05
Turbopack import.meta.glob fails for parent-relative patterns in server modules


chitwitgit:
### Link to the code that reproduces this issue

https://github.com/chitwitgit/turbopack-import-meta-glob

### To Reproduce

1. Clone https://github.com/chitwitgit/turbopack-import-meta-glob
2. Run `npm install`
3. Run `npm run build` — build succeeds
4. Run `npm run start`
5. Visit `/` and `GET /api/hello`

Alternatively:

```bash
npm run dev
curl http://localhost:3000/api/hello
```

### Current vs. Expected behavior

**Expected:** `import.meta.glob` is officially supported by Turbopack in Next.js 16.3 and should follow Vite-compatible importer-relative path semantics.

In this repro:

- the Server Component page should resolve `./page-modules/*.js` to `["page-local-module"]`
- the Server Component page should resolve `../lib/modules/*.js` to `["example-module"]`
- the App Route should resolve `./modules/*.js` to `["local-module"]`
- the App Route should resolve `../../../lib/modules/*.js` to `["example-module"]`

`/api/hello` should return:

```json
{
  "localValues": ["local-module"],
  "parentValues": ["example-module"]
}
```

The page should render:

- `page-local-values`: `["page-local-module"]`
- `page-parent-values`: `["example-module"]`

**Actual on Next.js 16.3 canary (`16.3.0-canary.78`):**

- `next build` and `next dev` succeed
- the local control glob works in both the Server Component page and the App Route
- the parent-relative glob matches nothing in both places
- `/api/hello` returns:

```json
{
  "localValues": ["local-module"],
  "parentValues": []
}
```

- the page renders:
  - `page-local-values`: `["page-local-module"]`
  - `page-parent-values`: `[]`

So this looks specific to parent-directory traversal in server-side `import.meta.glob` resolution, not a generic `import.meta.glob` failure.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Wed Nov  5 21:34:00 PST 2025; root:xnu-11417.140.69.705.2~1/RELEASE_ARM64_T8132
  Available memory (MB): 24576
  Available CPU cores: 10
Binaries:
  Node: 22.19.0
  npm: 11.17.0
  Yarn: 1.22.22
  pnpm: 10.33.4
Relevant Packages:
[...]

PR #95500, opened 2026-07-05
fix(turbopack): support parent-directory traversal in import.meta.glob

sammyZi:
## Summary

`import.meta.glob` in Turbopack silently returned no matches for any pattern
that traversed above the calling file's directory (`../foo/*.js`,
`../../lib/**/*.ts`, etc.). Local patterns (`./foo/*.js`) worked, so this only
affected parent-relative imports, which Vite documents as supported.

Root cause: in `packages/next/../import_meta_glob.rs`, `map()` compiled each
pattern into a `Glob` after stripping only a leading `./` and then called
`base_dir.read_glob(...)` on the origin directory. `read_glob` walks
*downward*, so a leading `../` never gets resolved and the pattern can't match
anything.

Fix: rewrite each pattern to be relative to a common `scan_base`. `map()` now:

1. Computes `max_parent` — the highest `../` count across all positive and
   negative patterns.
2. Walks the effective origin directory up by that many levels (clamped at the
   filesystem root by `FileSystemPath::parent()`).
3. Rewrites every pattern to be relative to that `scan_base`, prepending the
   appropriate slice of the origin's inner segments so `Glob::alternatives`
   still finds matches.

Result mapping (`origin_path.get_relative_path_to(path)`) was already correct,
so the JS keys and resolved imports come out as `../../lib/modules/foo.js`
without further changes.

## Verification

- `cargo test -p turbopack-tests --test execution -- test_tests__execution__turbopack__resolving__import_meta_glob`
  (3 execution fixtures pass, including four new `../`-traversal cases:
  parent-only, mixed local + parent, `import: 'default'` with `../`, and
  negatives applied to `../` matches).

New fixture files:

- `turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob/input/nested/parent-glob.js`
- `turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob/input/nested/sibling/local.js`

Fixes the reproduction at
https://github.com/chitwitgit/turbopack-import-meta-glob.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
@@ -316,11 +316,32 @@ pub fn parse_import_meta_glob(
 // Helpers for collecting files from ReadGlobResult
 // ---------------------------------------------------------------------------
 
-/// Strip the `./` prefix from a Vite-style glob pattern to produce a pattern
-/// compatible with Turbopack's `Glob` (which operates relative to the scan
-/// directory, without a leading `./`).
-fn strip_relative_prefix(pattern: &str) -> &str {
-    pattern.strip_prefix("./").unwrap_or(pattern)
+/// Consume leading `./` and `../` segments from a Vite-style glob pattern.
+///
+/// Returns the number of `..` (parent-directory traversals) and the remaining
+/// pattern, which is relative to the origin directory walked up by that many
+/// parents.
+///
+/// # Examples
+///
+/// - `"./foo/*.js"` → `(0, "foo/*.js")`
+/// - `"../lib/**/*.js"` → `(1, "lib/**/*.js")`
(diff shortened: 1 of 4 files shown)


chitwitgit:
Hi @sammyZi. I opened a duplicate in #95498 after filing #95496 and the [repro repo](https://github.com/chitwitgit/turbopack-import-meta-glob).

Your `scan_base` + pattern rewrite is cleaner than my original approach, so I updated #95498 to use the same idea instead of an originally larger refactor.

The main difference now is test coverage:

#95500 — turbopack execution tests for parent-only, mixed local+parent, import: 'default', and negatives
#95498 — e2e tests for parent-relative globs in a Server Component and App Route (including deeper ../../../ traversal), plus a few Rust unit tests

Happy for either PR to land — leaving it to maintainers to pick whichever fits best. I’m fine closing #95498 if this one merges, or contributing the e2e tests here if that’s preferred.

PR #95507, opened 2026-07-06
chore(deps): bump sharp@0.35.3

styfle:
## Notable changes
- No more install script needed 
- No more dynamic require/import tracing workaround
- AVIF uses [tune iq](https://aomedia.org/blog%20posts/Libavif_v1_4_0-Boasts-Major-Updates-to-Encoder-Technology/#the-tune-iq-image-quality-tuning-mode) to better match human perception

## All changes
- https://github.com/lovell/sharp/releases/tag/v0.35.0
- https://github.com/lovell/sharp/releases/tag/v0.35.1
- https://github.com/lovell/sharp/releases/tag/v0.35.2
- https://github.com/lovell/sharp/releases/tag/v0.35.3


packages/next/src/server/image-optimizer.ts
```diff
@@ -308,28 +309,25 @@ export async function detectContentType(
...
     const sharp = getSharp(concurrency, operationCache)
     const meta = await sharp(buffer)
       .metadata()
       .catch((_) => null)
-    format = meta?.format
+
+    const t = meta?.mediaType
+    if (t && [AVIF, WEBP, JPEG, PNG, GIF, SVG, TIFF].includes(t)) {
+      return t
+    }
```
styfle: The new version of sharp changed the metadata response.

- Before`{ "format": "avif" }`
- After `{ "format": "heic", "mediaType": "image/avif" }`

Since this code is after the mediaType, we can just return it here instead of needing to merge with the other detector response.

Also we can probably remove this sharp metadata call in a future PR anyway.

packages/next/src/server/image-optimizer.ts
```diff
@@ -372,8 +356,10 @@ export async function detectContentType(
     case 'psd':
     case 'tga':
     case undefined:
+      return null // unsupported formats
     default:
-      return null
+      format satisfies never // exhaustive check
+      return null // impossible to reach
```
styfle, line 345: The `satisfies` makes it safer in the future since we'll get a build error if we forget to add a `case`

packages/next/src/server/image-optimizer.ts
```diff
@@ -849,7 +835,7 @@ export async function optimizeImage({
 
   if (contentType === AVIF) {
     transformer.avif({
-      quality: Math.max(quality - 20, 1),
+      quality: Math.max(quality - 30, 1),
```
styfle: This was necessary since the avif output was larger than the webp output and we have a test to catch this problem.

I checked what sharp recommends and they have a default quality of 80 for webp and 50 for avif so a difference of 30 is expected here.

I also used my human eye balls to verify there is no discernable difference when converting mountains.jpg to webp or avif at this new quality.

test/production/next-server-nft/next-server-nft.test.ts
```diff
@@ -33,15 +36,8 @@ function normalizeNFT(base: string, files: string[]): string[] {
         })
         .map((file: string) => {
           // Normalize sharp, different architectures have different files
-          if (file.includes('/node_modules/@img/sharp-libvips-')) {
-            return '/node_modules/@img/sharp-libvips-*'
-          }
-          if (
-            file.match(
-              /\/node_modules\/@img\/sharp-\w+-\w+\/lib\/sharp-\w+-\w+.node$/
```
styfle, line 41: This regex was too brittle before so I changed it to be more explicit which is simply replace the platform and arch so this test will pass on windows/mac/linux/etc but still fail if its missing the files sharp needs to run.

mischnic: can you still remove the version from the filename though? i.e. the `0.35.3`

styfle: @mischnic Fixed in [3a3bc00](https://github.com/vercel/next.js/pull/95507/commits/3a3bc007d1f6f78e5b9cfd88d1df2f8e3b517355)

packages/next/src/server/image-optimizer.ts
styfle: Also here's an anaylsis for the output comparing before/after

- [avif-quality-comparison-q75.html.zip](https://github.com/user-attachments/files/29718560/avif-quality-comparison-q75.html.zip)

bgw: > I checked what sharp recommends and they have a default quality of 80 for webp and 50 for avif so a difference of 30 is expected here.

It'd be good to add a comment here with a link to the docs.

Also, unclear if just subtracting 30 is better, or if multiplying by `50/80` is better. The second option would probably scale better to really low quality settings?

Looking around I found this:
https://www.industrialempathy.com/posts/avif-webp-quality-settings/#quality-settings-for-a-range-of-jpeg-qualities
https://docs.google.com/spreadsheets/d/1E29kPLR5_0PThsw6SVbco7HvMU0aynBLfVN0RfIXgPk/edit?gid=1245219304#gid=1245219304

That implies something closer to a coefficient than a constant offset.

bgw (approved):

packages/next/src/server/image-optimizer.ts
styfle: @bgw Docs here:

- https://sharp.pixelplumbing.com/api-output/#avif
- https://sharp.pixelplumbing.com/api-output/#webp

styfle: I applied your suggestion for the 50/80 ratio in [78d062f](https://github.com/vercel/next.js/pull/95507/commits/78d062f6132a933bf4eb21b5f71faca2599a916f) since it makes avif better match webp at an value of quality 👍

See results here: [avif-quality-comparison-ratio-w1024.html.zip](https://github.com/user-attachments/files/29761147/avif-quality-comparison-ratio-w1024.html.zip)

styfle:
Changing to draft per @eps1lon so we can merge another PR first and avoid merge conflicts

bgw (approved):

packages/next/src/server/image-optimizer.ts
```diff
@@ -819,7 +820,10 @@ export async function optimizeImage({
 
   if (contentType === AVIF) {
     transformer.avif({
-      quality: Math.max(quality - 20, 1),
+      // Scale the quality to try and match webp. This ratio was derived
+      // from sharp's default 80 (webp) and 50 (avif), and then verified
+      // using DSSIM visual quality tests.
```
styfle:
```suggestion
      // using dssim and ssimulacra2 visual quality tests.
```

packages/next/src/server/image-optimizer.ts
```diff
@@ -92,7 +92,8 @@ export function getSharp(
     return _sharp
   }
   try {
-    _sharp = require('sharp') as typeof import('sharp')
+    // eslint-disable-next-line @next/internal/typechecked-require -- sharp is an optional peer dependency loaded lazily at runtime
+    _sharp = require('sharp') as SharpConstructor
```
eps1lon: `@next/internal/typechecked-require` should not be disabled. Why can't this stay as-is?

styfle: @eps1lon Because it assumes default export which this is not using. This was the only way I could get it to work. Let me know if you have a better solution.

eps1lon: [`07cb155` (this PR)](https://github.com/vercel/next.js/pull/95507/commits/07cb15590bdeea9d0c59ad67398a988400577e02) should do it

eps1lon (approved):

packages/next/src/server/image-optimizer.ts
eps1lon: or not.

eps1lon: [`e754027` (this PR)](https://github.com/vercel/next.js/pull/95507/changes/e754027916732ada51cf3f96cc9f8d9319f435e2)

styfle: Sounds great, thanks!

PR #95516, opened 2026-07-06
Turbopack: fix empty tree-shake chunks desyncing part ids (#93424)

jimmyhmiller:
split_module skipped chunks whose body ended up empty, but the group index is used directly as a part id by outputs (Key::Export/Key::ModuleEvaluation), part_deps, and the PartId::Internal asserts embedded in the emitted chunk ASTs. Skipping shifted every later module's position and desynced those indices, causing out-of-bounds panics in split_module and wrong-fragment selection in part_of_module. Always push a module for every group; use a well-formed side-effect-free `export {}` placeholder when the body is empty.

Update tree-shaker analyzer snapshots to include the placeholder parts.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/tests/tree-shaker/analyzer/next-response/output.md
+++ b/turbopack/crates/turbopack-ecmascript/tests/tree-shaker/analyzer/next-response/output.md
@@ -423,7 +423,7 @@ graph TD
     Export(
         "NextResponse",
     ): 0,
-    Exports: 1,
+    Exports: 7,
 }
 ```
 
@@ -569,6 +569,36 @@ export { };
 
 ```
 ## Part 1
+```js
+export { };
+
+```
+## Part 2
+```js
+export { };
+
+```
+## Part 3
+```js
+export { };
+
+```
+## Part 4
+```js
+export { };
+
+```
+## Part 5
+```js
+export { };
+
+```
+## Part 6
+```js
+export { };
+
+```
+## Part 7
 ```js
 export { NextResponse } from "__TURBOPACK_PART__" assert {
     __turbopack_part__: "export NextResponse"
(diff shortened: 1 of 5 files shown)


bgw (approved):
Nice! I assume this is pretty rare, so there shouldn't be any real-world performance or bundle size impacts (e.g. due to larger ids)?

lukesandberg:
no impact really since this only affects advanced tree shaking which is not enabled anywhere outside of tests


@jimmyhmiller wasn't there a github issue for this one?

lukesandberg (approved):

jimmyhmiller:
@bgw Yeah, not sure how rare it is. But since this is all behind the flag, shouldn't have any impact in practice. There are a series of these problems in the advanced tree shaking that I'm fixing to get this working on v0. (at least not crashing as it currently does). Once I have gotten past all the crashes, we can reassess what we need for this feature.

@lukesandberg  Yeah mentioned in title but should have done a better job. This closes https://github.com/vercel/next.js/issues/93424

There is an alternative fix https://github.com/vercel/next.js/pull/93582 But I've tested this locally and because of the assumption of indexes lining up with ids, that fix causes issues later. I will leave a message there about that.

bgw:
@jimmyhmiller ah, I didn't pick up that this was in the advanced tree shaking codepath, that makes a lot more sense.

turbopack/crates/turbopack-ecmascript/tests/tree-shaker/analyzer/next-response/output.md
```diff
@@ -570,6 +570,36 @@ export { };
...
+```
+## Part 5
+```js
+export { };
+
+```
+## Part 6
+```js
+export { };
+
```
mischnic, line 599: These certainly blow up the module graph unnecessarily...

PR #95521, opened 2026-07-06
Disable `supportsImmutableAssets` with `config.output`

mischnic:
`output=export` and `output=standalone` don't use adapters.
But for `supportsImmutableAssets`, you really want the adapter to get all necessary data.

Disable immutable assets for now in this case. Once `output=export` runs the adapter (as it should), we can remove this override again

### DIFF
--- a/packages/next/src/server/config.ts
+++ b/packages/next/src/server/config.ts
@@ -1624,6 +1624,17 @@ function finalizeConfig(config: NextConfigComplete): NextConfigComplete {
     config.experimental.supportsImmutableAssets = true
   }
 
+  if (
+    config.experimental.supportsImmutableAssets &&
+    (config.output === 'export' || config.output === 'standalone')
+  ) {
+    // supportsImmutableAssets is designed to work with adapters. Disable it for output=export and
+    // output=standalone, which are currently using a non-adapter codepath.
+    // Particularly output=export should just run through the adapter, with only static assets.
+    // TODO remove again once output=export (and output=standalone) are using adapters.
+    config.experimental.supportsImmutableAssets = false
+  }
+
   return config
 }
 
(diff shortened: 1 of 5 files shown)


mischnic:
* **#95521** <a href="https://app.graphite.com/github/pr/vercel/next.js/95521?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95521?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

bgw (approved):

packages/next/src/server/config.ts
```diff
@@ -1624,6 +1624,17 @@ function finalizeConfig(config: NextConfigComplete): NextConfigComplete {
...
 
+  if (
+    config.experimental.supportsImmutableAssets &&
+    (config.output === 'export' || config.output === 'standalone')
+  ) {
+    // supportsImmutableAssets is designed to work with adapters. Disable it for output=export and
+    // output=standalone, which are currently using a non-adapter codepath.
+    // Particularly output=export should just run through the adapter, with only static assets.
+    // TODO remove again once output=export (and output=standalone) are using adapters.
+    config.experimental.supportsImmutableAssets = false
```
janory, line 1635: Just asking to be sure, this won't be override by the "platform flag", right?

mischnic: No, that happens before.
```
/**
 * Post-processing applied by `loadConfig` after `applyModifyConfig`, so that
...
function finalizeConfig(config: NextConfigComplete): NextConfigComplete {
```

`applyModifyConfig` is "platform"

PR #95522, opened 2026-07-06
Turbopack: fix split_module correctness for app-router

jimmyhmiller:
Turbopack: fix split_module correctness for app-router

- Don't split the generated __nextjs-internal-proxy. client-reference
  proxy module. Its exports are all used, so splitting yields no benefit
  and duplicates its <exports> part in the module graph, tripping the
  duplicate-ident check in ModuleGraph::from_graphs.
- Pass the original comments to each fragment so magic-comment
  annotations (turbopackChunkingType, turbopackExports) are still
  parsed. Fragment ASTs keep their original spans, so comments resolve
  at the same byte positions. Without this, references fall back to
  defaults (ChunkingType::Shared -> Parallel), breaking app-router
  layout-segment chunk groups.

### DIFF
--- a/crates/next-core/src/next_client_reference/ecmascript_client_reference/ecmascript_client_reference_module.rs
+++ b/crates/next-core/src/next_client_reference/ecmascript_client_reference/ecmascript_client_reference_module.rs
@@ -20,7 +20,7 @@ use turbopack_core::{
     output::OutputAssetsReference,
     reference::{ModuleReference, ModuleReferences},
     reference_type::ReferenceType,
-    resolve::ModuleResolveResult,
+    resolve::{ModulePart, ModuleResolveResult},
     source::OptionSource,
     virtual_source::VirtualSource,
 };
@@ -152,20 +152,29 @@ impl EcmascriptClientReferenceModule {
         let proxy_module_content =
             AssetContent::file(FileContent::Content(File::from(code.source_code().clone())).cell());
 
-        let proxy_source = VirtualSource::new(
-            self.server_ident.await?.path.join(
-                // We choose the extension based on the original file because we're placing the
-                // virtual module next to the original code, so its parsing will be
-                // affected by `type` fields in package.json -- a bare `proxy.js`
-                // may end up being unexpectedly parsed as the wrong format.
-                // The name special cased later to always ignore-list this module.
-                &format!(
-                    "__nextjs-internal-proxy.{}",
-                    if is_esm { "mjs" } else { "cjs" }
-                ),
-            )?,
-            proxy_module_content,
-        );
+        let proxy_path = self.server_ident.await?.path.join(
+            // We choose the extension based on the original file because we're placing the
+            // virtual module next to the original code, so its parsing will be
+            // affected by `type` fields in package.json -- a bare `proxy.js`
+            // may end up being unexpectedly parsed as the wrong format.
+            // The name special cased later to always ignore-list this module.
+            &format!(
+                "__nextjs-internal-proxy.{}",
+                if is_esm { "mjs" } else { "cjs" }
+            ),
+        )?;
+
+        // Tag the ident with a `ModulePart::Exports` part so tree-shaking's
+        // `split_module` opt-out (`!ident.parts.is_empty()`) skips this module.
+        // The proxy is a codegen delegate whose exports are all used, so
+        // splitting yields no benefit; splitting it also duplicates its
+        // `<exports>` part in the module graph (the proxy is reached via two
+        // paths), tripping the duplicate-ident check in
+        // `ModuleGraph::from_graphs`.
+        let proxy_ident = AssetIdent::from_path(proxy_path)
+            .with_part(ModulePart::Exports)
+            .into_vc();
+        let proxy_source = VirtualSource::new_with_ident(proxy_ident, proxy_module_content);
 
         let proxy_module = self
             .server_asset_context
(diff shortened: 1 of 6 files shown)


jimmyhmiller:
Not the most elegant approach. But I found that this internal module was causing these problems. I saw that there were a number of other places we added an exception. So I followed suit.

turbopack/crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -491,6 +491,20 @@ pub(super) async fn split_module(asset: Vc<EcmascriptModuleAsset>) -> Result<Vc<
         .cell());
     }
```
lukesandberg, line 492: <p class='graphite__hidden'><i>[Re: lines +486 to +492]</i></p>
i wonder if this bug is still real?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/vercel/next.js/95522?utm_source=unchanged-line-comment">Graphite</a>.</p>

turbopack/crates/turbopack-ecmascript/src/tree_shake/mod.rs
```diff
@@ -491,6 +491,20 @@ pub(super) async fn split_module(asset: Vc<EcmascriptModuleAsset>) -> Result<Vc<
...
 
+    // Don't split the generated client-reference proxy module. It is a codegen
+    // delegate for `EcmascriptClientReferenceModule` whose exports are all used,
+    // so splitting yields no benefit. Splitting it duplicates its `<exports>`
+    // part in the module graph (the proxy is used for codegen but reached via
+    // two paths), which trips the duplicate-ident check in
+    // `ModuleGraph::from_graphs`. See `BindingUsageInfo::used_exports` in
+    // `binding_usage_info.rs` for a similar workaround.
+    if name.contains("__nextjs-internal-proxy.") {
```
lukesandberg: 2 thoughts

we could use a directive perhaps `"use turbopack: no split"` 

since we construct the module ident we could add a part `ModulePart::Exports` and thus hit the earlier opt out.  At least that way the opt out would be closer to the source.

jimmyhmiller: I like the second second suggestion. Not sure I want to commit to something public like the directive yet. I will make that change.

jimmyhmiller: @lukesandberg Fixed now

crates/next-core/src/next_client_reference/ecmascript_client_reference/ecmascript_client_reference_module.rs
```diff
@@ -152,20 +152,29 @@ impl EcmascriptClientReferenceModule {
...
+            ),
+        )?;
+
+        // Tag the ident with a `ModulePart::Exports` part so tree-shaking's
+        // `split_module` opt-out (`!ident.parts.is_empty()`) skips this module.
+        // The proxy is a codegen delegate whose exports are all used, so
+        // splitting yields no benefit; splitting it also duplicates its
+        // `<exports>` part in the module graph (the proxy is reached via two
+        // paths), tripping the duplicate-ident check in
+        // `ModuleGraph::from_graphs`.
```
lukesandberg: as discussed on slack. this explanation appears incomplete.  drop a TODO to revisit

lukesandberg (approved):

turbopack/crates/turbopack-ecmascript/src/tree_shake/mod.rs
afriadishariff-lab: Now

crates/next-core/src/next_client_reference/ecmascript_client_reference/ecmascript_client_reference_module.rs
jimmyhmiller: Updated to remove incomplete explanation with a proper todo.

crates/next-core/src/next_client_reference/ecmascript_client_reference/ecmascript_client_reference_module.rs
```diff
@@ -152,20 +152,29 @@ impl EcmascriptClientReferenceModule {
...
+                "__nextjs-internal-proxy.{}",
+                if is_esm { "mjs" } else { "cjs" }
+            ),
+        )?;
+
+        // Tag the ident with a `ModulePart::Exports` part so tree-shaking's
+        // `split_module` opt-out (`!ident.parts.is_empty()`) skips this module.
+        // Because of the state of this feature, it is unclear if this is competely
+        // the right course of action. Not skipping this causes "Duplicate module idents"
+        // errors.
```
mischnic: It would actually be nice to split this up (though the savings are marginaly(

Why do you get duplicate module idents? Shouldn't the module splitting add `(internal part X)` to all modules?

mischnic: This feels like a more fundamental bug in `split_module`. 
You might have your own custom Module via a module rule that doesn't have a ModulePart in the ident either. Then you'd also run into this bug.

mischnic (approved):

mischnic:
(This is a workaround to facilitate benchmarking. Not a fix and should be revisited)

PR #95534, opened 2026-07-06
Emit prerender classification metadata on adapter build outputs

agadzik:
### What?

Emits classification metadata on adapter `PRERENDER` outputs so platforms can tell what each build prerender produced, without re-deriving it from file paths or manifest heuristics.

New fields on `AdapterOutput.PRERENDER`:

- `route: string` — the source route matcher, aligned with the filesystem route and keeping dynamic segments (e.g. `/blog/[slug]` for the prerendered path `/blog/first`). Grouping outputs by `route` yields one matcher with prerenders at varying depths of specificity, including partial fallbacks (`/prefix/b/[two]`).
- `routeType?: 'fallback' | 'shell' | 'page'` — which URLs the prerender serves:
  - `fallback`: at least one prerenderable param (fillable by `generateStaticParams`/`getStaticPaths`) is missing — the generic UI for a class of URLs that can still be prerendered into something more specific
  - `shell`: all prerenderable params present; the remaining params are dynamic and resolve per request — the most specific prerender possible for its class of URLs
  - `page`: no params missing — the prerender for one specific URL
- `ui?: 'empty' | 'partial' | 'complete'` — visual completeness of the prerendered HTML: no initial UI at all / pending UI that appears after load (server- or client-resolved) / visually complete on first load
- `compute?: 'blocking' | 'resuming' | 'static'` — per-request compute involved in serving it: server must render before any UI is ready / prerendered UI is served and the server resumes the postponed parts / no per-request compute
- `htmlSize?: number` — byte size of the prerendered HTML shell (`0` = empty shell). Set only on the HTML output; RSC/data/segment variants mirror `route`/`routeType`/`ui`/`compute` but never `htmlSize`.

Semantics:

- All three dimensions are point-in-time snapshots of the build artifact — revalidation can change them; nothing here is intrinsic to the route.
- `undefined` = does not apply: route handlers are unclassified for now; `compute` is also `undefined` for templates that are never served (`fallback: false` / `dynamicParams = false`, where unmatched paths 404).
- The dimensions are correlated but independent, which is the point: a fully-`'use client'` page with dynamic params is a `shell` that is `static`; a page whose only hole is a client-resolving Suspense boundary (e.g. `useSearchParams`) is `partial` + `static` — no resume, CDN cost only.

Supporting change: `RouteMetadata` (`.meta` files) gains `hasPendingUi`, detected from React's pending (`<!--$?-->`) and client-render (`<!--$!-->`) Suspense boundary markers in the prerendered HTML. This is what separates client-resolving holes from complete pages — the postponed state alone cannot. Follow-up (per review discussion): source this from an augmented `RenderResult` instead of the HTML, since marker sniffing can false-positive on large boundary outlining.

All fields are documented in the adapter output-types reference (`docs/01-app/03-api-reference/07-adapters/09-output-types.mdx`).

### Why?

Deployment platforms want to display and act on prerender classification (e.g. a deployment summary grouping prerenders per route and tagging them). Today that requires build-outputs heuristics that break across Next.js versions. Shipping low-level orthogonal dimensions — rather than a single pre-baked label — lets each platform derive its own taxonomy, and keeps serving decisions Next.js can't know about (e.g. choosing to upgrade instead of serving a blocking fallback) on the platform side. Shape agreed in adapter API design review.

Example, grouped by `route` (from the test fixture: `[one]` has `generateStaticParams`, `[two]` doesn't):

```
route: /prefix/[one]/[two]   pathname: /prefix/[one]/[two]   -> fallback / partial / resuming
route: /prefix/[one]/[two]   pathname: /prefix/b/[two]       -> shell    / partial / resuming
route: /static               pathname: /static               -> page     / complete / static
```

### How?

- Derived once in `handleBuildComplete`: `routeType` from `remainingPrerenderableParams` (shell-bearing templates) or concrete-vs-template; `ui`/`compute` via a shared `classifyAppShell` helper over shell size + `.meta` (`postponed`, `hasPendingUi`); `htmlSize` via `fs.statSync`
- Pages router mapping: GSP page → `page`/`complete`/`static`; `fallback: true` template → `fallback`/`partial`/`static` (serving the loading shell is static; generation happens on the data-route request); `fallback: 'blocking'` → `fallback`/`empty`/`blocking`; `fallback: false` → `fallback`/`empty`/`undefined`
- `htmlSize` is attached at the HTML prerender push rather than on the shared `initialOutput`, so spread-derived RSC/data outputs can't inherit it
- e2e coverage in `test/production/app-dir/adapter-prerender-metadata` (7 tests): the fallback/shell matrix via partial `generateStaticParams`, the client-hole `partial`+`static` case, postponed concrete pages, pages router mapping, and variant/segment mirroring

### Verification

- `pnpm test-start-turbo test/production/app-dir/adapter-prerender-metadata/`
- Existing adapter suites unchanged: `adapter-config`, `adapter-config-export`, `adapter-partial-fallback`, `adapter-config-i18n-api`, `adapter-cache-handlers`, `empty-shell-route-cache`
- Note: suites verified with `NEXT_SKIP_ISOLATE=1` — isolated-mode installs are currently broken repo-wide by `typescript@7.0.2` being tagged `latest` (reproduced at clean HEAD, unrelated to this change)


packages/next/src/build/adapter/build-complete.ts
```diff
@@ -1258,6 +1295,14 @@ export async function handleBuildComplete({
         return newCheck
       }
 
+      // A missing file yields `undefined` ("no signal"), never 0 — an empty
+      // shell that exists on disk is a meaningful 0.
+      const statHtmlSize = (filePath: string): Promise<number | undefined> =>
+        fs
+          .stat(filePath)
```
timneutkens: Sync filesystem operations are generally faster for these cases.

agadzik: Good call — switched to `statSync` in 688f74d4ee.

PR #95539, opened 2026-07-06
[turbopack] Don't SSR on pages only navigated to through a soft nav

sampoder:
This PR essentially implements: https://github.com/vercel/next.js/pull/95539#discussion_r3555050929. Will add more a description soon.

### DIFF
--- a/crates/next-api/src/app.rs
+++ b/crates/next-api/src/app.rs
@@ -1,4 +1,4 @@
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, bail};
 use bincode::{Decode, Encode};
 use next_core::{
     app_structure::{
@@ -39,8 +39,8 @@ use next_core::{
 use tracing::{Instrument, field::Empty};
 use turbo_rcstr::{RcStr, rcstr};
 use turbo_tasks::{
-    Completion, FxIndexMap, NonLocalValue, ResolvedVc, TryJoinIterExt, ValueToString, Vc,
-    fxindexset, trace::TraceRawVcs,
+    Completion, FxIndexMap, NonLocalValue, OperationVc, ResolvedVc, TryJoinIterExt, ValueToString,
+    Vc, fxindexset, trace::TraceRawVcs,
 };
 use turbo_tasks_fs::{File, FileContent, FileSystemPath};
 use turbopack::{
@@ -78,6 +78,8 @@ use crate::{
     module_graph::{ClientReferencesGraphs, NextDynamicGraphs, ServerActionsGraphs},
     nft::{EndpointTraceResult, trace_endpoint},
     nft_json::NftJsonAsset,
+    operation::OptionEndpoint,
+    output_mode::{OptionSsrMarkTarget, SsrMarkTarget},
     paths::{
         all_asset_paths, all_paths_in_root, get_asset_paths_from_root, get_js_paths_from_root,
         get_wasm_paths_from_root, paths_to_bindings, wasm_paths_to_bindings,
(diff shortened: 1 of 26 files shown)


packages/next/src/server/dev/turbopack-utils.ts
```diff
@@ -346,7 +349,10 @@ export async function handleRouteType({
     case 'app-page': {
       const key = getEntryKey('app', 'server', page)
 
-      const writtenEndpoint = await route.htmlEndpoint.writeToDisk()
+      const endpoint =
+        rscOnly && route.rscEndpoint ? route.rscEndpoint : route.htmlEndpoint
+
+      const writtenEndpoint = await endpoint.writeToDisk()
```
lukesandberg: the risk with this approach is memory and overhead

each new endpoint means a new root turbopack task  which can be non-trivial overhead.   The alternative approach would be to manage the state inside of turbopack itself.

so isntead of `const htmlBuiltAppPages: Set<string>` in javascript it would be a similar structure inside of turbopack.  Managed as a `State<T>` parameter. (but maybe intver

this could work by passing a boolean `rscOnly` to `writeToDisk` (probably the `key` as well?)

then we would store this on the `ProjectContainer`,  then `writeToDisk` could insert the key into the state

then instead of splitting into separate endpoints, we would just read the state to see if the html only work should be done

this works better since instead of 2 endpoints it is 1 and we simply re-evaluate it when it changes.

crates/next-api/src/app.rs
```diff
@@ -1106,9 +1108,49 @@ pub fn app_entry_point_to_route(
...
+/// where there is no output mode state.
+#[turbo_tasks::function(operation, root)]
+pub(crate) async fn mark_as_ssr_operation(
+    endpoint_op: OperationVc<OptionEndpoint>,
+) -> Result<Vc<OptionSsrMarkTarget>> {
+    let Some(endpoint) = *endpoint_op.connect().await? else {
+        return Ok(Vc::cell(None));
+    };
+    let Some(app_endpoint) = ResolvedVc::try_downcast_type::<AppEndpoint>(endpoint) else {
+        return Ok(Vc::cell(None));
```
lukesandberg: error condition?

crates/next-api/src/app.rs
```diff
@@ -1106,9 +1108,49 @@ pub fn app_entry_point_to_route(
...
+    };
+    let app_endpoint = app_endpoint.await?;
+    if !matches!(
+        app_endpoint.ty,
+        AppEndpointType::Page {
+            ty: AppPageEndpointType::Html,
+            ..
+        }
+    ) {
+        return Ok(Vc::cell(None));
```
lukesandberg: ditto, error condition?

crates/next-api/src/app.rs
```diff
@@ -1280,21 +1334,17 @@ impl AppEndpoint {
...
+        let server_chunking_context = match runtime {
+            NextRuntime::NodeJs => Vc::upcast(project.server_chunking_context(true)),
+            NextRuntime::Edge => this
+                .app_project
+                .project()
+                .edge_chunking_context(process_client_assets),
+        }
+        .to_resolved()
+        .await?;
+
```
lukesandberg, line 1346: why do we always resolve the chunking context if we might not need it?

crates/next-api/src/app.rs
```diff
@@ -1106,9 +1108,49 @@ pub fn app_entry_point_to_route(
...
+    ) {
+        return Ok(Vc::cell(None));
+    }
+    let Some(state) = *app_endpoint
+        .app_project
+        .project()
+        .output_mode_state()
+        .await?
+    else {
+        return Ok(Vc::cell(None));
```
lukesandberg: yeah i think this is an error, mark_as_ssr should never be called outside of a dev session

packages/next/src/server/next-server.ts
```diff
@@ -1110,6 +1110,13 @@ export default class NextNodeServer extends BaseServer<
...
       const options: MatchOptions = {
         i18n: this.i18nProvider?.fromRequest(req, pathname),
+        // Only actual navigations compile the SSR-free output — never
+        // prefetches. With cache components, a prefetch triggers a dev
+        // validation render that consumes the Flight payload through
+        // `ssrModuleMapping`, which the SSR-free output doesn't emit.
+        rscOnly:
+          !!getRequestMeta(req, 'isRSCRequest') &&
+          !getRequestMeta(req, 'isPrefetchRSCRequest'),
       }
```
lukesandberg: maybe base-server could expose a method that computes this

and then we call it from both places....

`super.matchOptions(req, pathname)`

test/development/app-dir/skip-ssr-on-navigation/skip-ssr-on-navigation.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  // Reads the client-reference manifest that the most recent compile of
+  // `/other` wrote to disk, to distinguish the SSR-free `rscEndpoint` output
+  // (empty `ssrModuleMapping`) from the full `htmlEndpoint` output.
+  async function loadOtherPageClientReferenceManifest() {
+    const source = await next.readFile(
+      '.next/dev/server/app/other/page_client-reference-manifest.js'
+    )
+    const scope = { __RSC_MANIFEST: {} as Record<string, any> }
+    // eslint-disable-next-line no-new-func
+    new Function('globalThis', source)(scope)
```
lukesandberg: i think in other places we use `require`?  (resetting the global before and after)

i guess this is fine but `eval` always gives me the 'ick'

lukesandberg (approved):

crates/next-api/src/app.rs
sampoder: we're still using this for RSC

test/development/app-dir/skip-ssr-on-navigation/skip-ssr-on-navigation.test.ts
sampoder: yeah I can clean this up

PR #95548, opened 2026-07-07
test: Fix service worker error test

mischnic:
- One of them tests that an error is thrown. Don't run the deployment test there
- The other tests are still broken when deployed, but we will fix that separately


sampoder (approved):
thank you!

PR #95554, opened 2026-07-07
[turbopack] Output service workers to `/_next/static/`

sampoder:
This should fix deployment tests because this now re-uses Vercel's support for static JS files


packages/next/src/build/index.ts
```diff
@@ -2710,6 +2711,24 @@ export default async function build(
...
+      // `Service-Worker-Allowed` header.
+      const serviceWorkerDir = path.join(distDir, 'static', 'service-worker')
+      if (
+        existsSync(serviceWorkerDir) &&
+        (await fs.readdir(serviceWorkerDir)).length > 0
+      ) {
+        routesManifest.headers.push(
+          buildCustomRoute('header', {
+            source: '/_next/static/service-worker/:path*',
+            headers: [{ key: 'Service-Worker-Allowed', value: '/' }],
```
mischnic: what happens (or should happen) if you have a `assetPrefix` or a `basePath` configured in nexgt.config.js?

sampoder: If you were using an `assetPrefix`, I don't think you would want to serve it from there because service workers should be served from the same origin. I can patch the `basePath` issue.

Though merging this PR would make these docs on `assetPrefix` wrong: https://nextjs.org/docs/app/api-reference/config/next-config-js/assetPrefix

> The exact configuration for uploading your files to a given CDN will depend on your CDN of choice. The only folder you need to host on your CDN is the contents of .next/static/, which should be uploaded as _next/static/ as the above URL request indicates. Do not upload the rest of your .next/ folder, as you should not expose your server code and other configuration to the public.

I've pushed a patch to `basePath`

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -3077,11 +3077,20 @@ where
...
                         _ => rcstr!("/"),
                     };
+                    let base_path = compile_time_info
+                        .await?
+                        .defines
+                        .read_process_env(rcstr!("__NEXT_ROUTER_BASEPATH"))
+                        .await?
+                        .as_ref()
+                        .cloned()
+                        .unwrap_or_default();
```
mischnic: We don't want Next.js specific code in the `turbopack/` crates

packages/next/src/build/index.ts
mischnic: But don't we want `Service-Worker-Allowed: ${basepath}` then?

sampoder: Yep, sorry, forgot to patch that

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
sampoder: Moved this into the ChunkingContext to avoid this

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -315,6 +320,7 @@ pub struct BrowserChunkingContext {
     /// them.
     #[bincode(with = "turbo_bincode::indexmap")]
     asset_base_paths: FxIndexMap<RcStr, RcStr>,
+    base_path: Option<RcStr>,
```
mischnic: Add docs, in particular that this is only used for the service worker scope, not for output subdirectory logic

sampoder: Added a comment with this, maybe this is best to rename to `service_worker_scope_base_path` as we have many of these

crates/next-core/src/next_client/context.rs
```diff
@@ -530,6 +532,10 @@ pub async fn get_client_chunking_context(
 
     let next_mode = mode.await?;
     let asset_prefix = asset_prefix.owned().await?;
+    let base_path = base_path
+        .owned()
+        .await?
+        .filter(|base_path| !base_path.is_empty());
```
bgw: Why is this filter here? `base_path` should be an `Option` both before and after?

test/e2e/app-dir/service-worker-basepath/service-worker-basepath.test.ts
```diff
@@ -0,0 +1,36 @@
...
+import { retry } from 'next-test-utils'
+
+// Compiling `navigator.serviceWorker.register(new URL(...))` is a
+// Turbopack-only feature.
+;(process.env.IS_TURBOPACK_TEST ? describe : describe.skip)(
+  'app dir - service worker with basePath',
+  () => {
+    const { next } = nextTestSetup({
+      files: __dirname,
+    })
```
bgw: You can probably just run the other test that you have, but with two `describe` blocks that call `nextTestSetup` with a different next config (can be declared inline, as long as it's JSON-serializable). The actual test parts (the `it` blocks) can be in a helper function used by both `describe` blocks.

turbopack/crates/turbopack-ecmascript/src/references/service_worker.rs
```diff
@@ -247,20 +247,54 @@ pub struct ServiceWorkerAssetReferenceCodeGen {
...
+        let base_path = chunking_context.base_path().await?;
+        let base_path = base_path.trim_end_matches('/');
+        let url = format!(
+            "{base_path}/_next/static/service-worker/{}",
+            service_worker_chunk_filename(&self.scope)
+        );
+        let scope = match self.scope.as_str() {
+            "/" if base_path.is_empty() => "/".to_string(),
+            "/" => base_path.to_string(),
+            s => format!("{base_path}{s}"),
```
bgw, line 270: This is assuming that `self.scope` always begins with `/`, I'm not totally sure that's true? MDN says "usually": https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/register#scope

If we're not sure what's right to do here, let's just treat it as an error in the analyzer if it doesn't start with `/`.

bgw (approved):
I agree that we should rename the field used in the chunking context.

crates/next-core/src/next_client/context.rs
sampoder: removed this

test/e2e/app-dir/service-worker-basepath/service-worker-basepath.test.ts
sampoder: done!

turbopack/crates/turbopack-ecmascript/src/references/service_worker.rs
sampoder: added an error for it

PR #95564, opened 2026-07-07
docs: runtime prefetching update

icyJoseph:
Sync the runtime prefetching guide with latest App Shell changes


docs/01-app/02-guides/incremental-static-regeneration-cache-components.mdx
```diff
@@ -83,7 +83,7 @@ export default function CategoryLayout(props: LayoutProps<'/[category]'>) {
 }
 ```
 
-Notice that `CategoryLayout` does not `await props.params` itself. Instead, it passes the `params` promise to `CategoryHeader` inside `<Suspense>`. The `await` happens inside the boundary, so for unknown categories Next.js can still generate the App Shell.
+Notice that `CategoryLayout` does not `await props.params` itself. Instead, it passes the `params` promise to `CategoryHeader` inside `<Suspense>`. The `await` happens inside the boundary, so for unknown categories Next.js can still generate the App Shell. Keep the read inside the boundary even for the categories `generateStaticParams` covers, since a statically known param still belongs to one URL.
```
aurorascharff: known param still belongs to one URL.

maybe not entirely clear why this matters? should we link to something?

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,101 @@
...
 
-## What runtime prefetching does
+[Cache Components](/docs/app/getting-started/caching) takes a different approach and prefetches an [**App Shell**](/docs/app/glossary#app-shell) for every route. The shell contains the route's static output, and for a route that reads [`cookies()`](/docs/app/api-reference/functions/cookies) or [`headers()`](/docs/app/api-reference/functions/headers) (session data), it can also contain UI gated behind that session data.
 
-A user on `/` sees a [`<Link>`](/docs/app/api-reference/components/link) to `/courses`. The destination mixes four kinds of content: a static heading, an `<EnrolledBadge>` that reads the session cookie, a cached `<FeaturedCourses>` list, and a `<LiveEnrollment>` count that has to be fresh on every request.
+Prefetching one reusable shell per route, rather than a separate prefetch per link, is [**Partial Prefetching**](/docs/app/api-reference/config/next-config-js/partialPrefetching). Any number of links to the same route share that shell, so extra links to it are effectively free, and each route is fetched as its [`<Link>`](/docs/app/api-reference/components/link) enters the viewport.
```
aurorascharff: Hmm well PPF is actually another flag..? it kinda sounds like the app shell is true for CC in general, which I thiiiink andrew mentioned tying to the ppf flag only? not sure

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,101 @@
...
+
+export default nextConfig
 ```
 
-```tsx filename="app/courses/page.tsx"
+It also assumes your route is already structured for instant navigation. If it isn't, start with the [Instant navigation guide](/docs/app/guides/instant-navigation) to validate its caching structure first.
+
+## What runtime prefetching does
+
+Runtime prefetching takes two opt-ins: the destination route sets [`prefetch = 'allow-runtime'`](/docs/app/api-reference/file-conventions/route-segment-config/prefetch#allow-runtime), and links to it use `<Link prefetch={true}>`.
```
aurorascharff: Should this be reordered to mention link prefetch true first since its the first step youd do?

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,101 @@
...
+With [`prefetch = 'allow-runtime'`](/docs/app/api-reference/file-conventions/route-segment-config/prefetch) on the route, the router prefetches a prerender that resolves `<Results>` before the click. The `q` value comes from the link's URL, known at prefetch time, and the cached `search(q)` fills in from there. On the click, the results render immediately, with no fallback.
 
-The prerender advances through anything that's static or cached, then stops at uncached reads and falls back to the surrounding `<Suspense>` boundary. The boundary is already in place from your **instant-nav validation**.
+The prerender advances through anything static or cached, then stops at uncached reads and falls back to the surrounding `<Suspense>` boundary. That boundary is already in place from structuring the route for instant navigation.
 
-The result is a **runtime prerender**: more of the page is already rendered before the user clicks, with fewer loading spinners.
+The result is a **runtime prerender**: more of the page is rendered before the user clicks, with fewer loading states.
```
aurorascharff: Is this terminology used elsewhere/defined?  _runtime prerender_

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,101 @@
...
-Take a dashboard layout with a nav that depends on request data. Without runtime prefetching, `<UserNav>` always streams in behind a `<Suspense>` fallback after navigation, even though the cookie that determines its content is already known when the prefetch fires. With runtime prefetching, the router prefetches a prerender that resolves `<UserNav>` before the click:
+## Session data resolves in the shell
 
-```tsx filename="app/dashboard/layout.tsx"
-export const prefetch = 'allow-runtime'
-```
+Runtime prefetching is for URL data. Session data is handled separately: a route that reads `cookies()` or `headers()` gets an App Shell variant that includes its session data, cached per session on the client. That content is ready on navigation without `allow-runtime`.
 
-The route still needs a valid prerender, so `<UserNav>` stays behind a `<Suspense>` boundary:
+The wrinkle is derived data. Take a dashboard nav that reads a cookie, then looks up content based on it:
```
aurorascharff: Hmm this sounds AI!!

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -99,11 +114,11 @@ export default function DashboardLayout({
 }
 ```
 
-`<UserNav>` reads a cookie, then looks up data based on it. The challenge is that `"use cache"` can't read `cookies()` inside the cached function. Two patterns handle this: **extract and pass** when the lookup result is shared across many users, and `"use cache: private"` when it's tied to a specific user.
+The cookie itself is session data, which the session shell already knows. But the lookup it drives is a data read, and for that to land in the shell instead of streaming in after navigation, it needs a cache lifetime. Because `"use cache"` can't read `cookies()` inside the cached function, two patterns bridge it: **extract and pass** when the lookup result is shared across many sessions, and `"use cache: private"` when it is tied to one.
```
aurorascharff: "session shell"? is this the app shell?

aurorascharff: Could be nice to use a list here for the two patterns possibly

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -128,13 +143,13 @@ async function getTopics(team: string | undefined) {
 }
 ```
```
aurorascharff, line 168: "covering it until then"

a little vague maybe

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -151,55 +166,32 @@ async function getUser() {
...
 
-- A useful chunk of the page depends on request data: cookies, headers, the full URL, `searchParams`, or `params` not resolved by [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params)
+- A useful chunk of the page depends on URL data: the full URL, `searchParams`, or `params` not resolved by [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params)
 - That chunk has a known cache lifetime (it can be expressed with `"use cache"` or `"use cache: private"`)
 - The traffic justifies the per-link server invocation
 
 Skip it when the prefetch can't produce a better UI than the App Shell. Each visible `<Link>` to a route with `'allow-runtime'` wakes a server, and that cost only pays off if more of the page is ready before the click:
 
-- The route has little or no runtime-data dependency. The App Shell already gets you instant.
+- The route has little or no URL-data dependency. The App Shell, including its per-session variant, already gets you instant.
```
aurorascharff: , already gets you instant.


this is a bit unclear, what does "get you instant" mean?

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -151,55 +166,32 @@ async function getUser() {
...
-
-export default nextConfig
-```
-
-N links to the same route share one prefetched App Shell, so rendering a `<Link>` is effectively free. To prefetch more than the shell (including request data like cookies, headers, params, and `searchParams`), opt the destination into [`prefetch = 'allow-runtime'`](/docs/app/api-reference/file-conventions/route-segment-config/prefetch#allow-runtime) and use `<Link prefetch={true}>` on the link side.
-
-> **Good to know**: Routes that read `cookies()` or `headers()` produce an App Shell that includes session data. The framework auto-detects this, and the shell is cached per session on the client, not shared across users.
-
-App Shells vs runtime prefetches:
+A per-link runtime prefetch is best-effort. It only helps the navigations where it fires before the click and completes before the click. On a slow connection, on a feed of many links, or on a direct visit, it may not exist yet when the user navigates, and the navigation falls back to the App Shell. The shell is the reliable floor. Runtime prefetching is the upgrade when it lands in time.
```
aurorascharff: Good overview but maybe could be written a little more docs-educationally. or maybe it just reads AI-y to me with these two fullstops 

The shell is the reliable floor. Runtime prefetching is the upgrade when it lands in time.

docs/01-app/02-guides/instant-navigation.mdx
```diff
@@ -50,7 +50,7 @@ With Cache Components, **caching directives** (`"use cache"` and its variants) a
 
 > **Good to know:** A fallback may access `cookies()`, `headers()`, or the full URL. At build time, the fallback itself suspends, and a `<Suspense>` boundary further up the tree is needed. With [runtime prefetching](/docs/app/guides/runtime-prefetching), the information is available and such a fallback becomes part of the instant UI. Cached values like timestamps or data fetches can sit directly inside the fallback.
 
-Next.js can also generate an [**App Shell**](/docs/app/glossary#app-shell) per route: a fallback that renders instantly during client navigations when nothing else is ready. See [App Shells](/docs/app/guides/runtime-prefetching#app-shells) for how to enable them and how they pair with runtime prefetching.
+Next.js can also generate an [**App Shell**](/docs/app/glossary#app-shell) per route: a fallback that renders instantly during client navigations when nothing else is ready. See the [Runtime prefetching](/docs/app/guides/runtime-prefetching) guide for how to enable them and how they pair with per-link prefetching.
```
aurorascharff: this seems a bit contradictory. Wouldnt the app shell all ready be available without this guides runtime prefetch?

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,121 @@
...
+
+[Cache Components](/docs/app/getting-started/caching) takes a different approach and produces an [**App Shell**](/docs/app/glossary#app-shell) for every route. The shell contains the route's static output, and for a route that reads [`cookies()`](/docs/app/api-reference/functions/cookies) or [`headers()`](/docs/app/api-reference/functions/headers) (session data), it can also contain UI gated behind that session data.
+
+Prefetching one reusable shell per route, rather than a separate prefetch per link, is [**Partial Prefetching**](/docs/app/api-reference/config/next-config-js/partialPrefetching). Any number of links to the same route share that one shell, fetched once as a [`<Link>`](/docs/app/api-reference/components/link) enters the viewport and reused for the rest.
+
+The shell cannot carry data that varies per link: [`searchParams`](/docs/app/api-reference/file-conventions/page#searchparams-optional) and [`params`](/docs/app/api-reference/file-conventions/page#params-optional), the route's URL data. **Runtime prefetching** resolves that URL data at prefetch time, ready before the click rather than streaming in after it.
```
aurorascharff: do we want to link to URL data glossary here?

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,121 @@
...
+With [`prefetch = 'allow-runtime'`](/docs/app/api-reference/file-conventions/route-segment-config/prefetch) on the route, the router prefetches a prerender that resolves `<Results>` before the click. The `q` value comes from the link's URL, known at prefetch time, and the cached `search(q)` fills in from there. On the click, the results render immediately, with no fallback.
 
-The prerender advances through anything that's static or cached, then stops at uncached reads and falls back to the surrounding `<Suspense>` boundary. The boundary is already in place from your **instant-nav validation**.
+The prerender advances through anything static or cached, then stops at uncached reads and falls back to the surrounding `<Suspense>` boundary. That boundary is already in place from structuring the route for instant navigation.
```
aurorascharff: from structuring the route for instant navigation.

a little unclear what youre referring to here. I remember because I read this carefully

might be better to link back or re-state the setup requirement maybe?

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,121 @@
...
-The result is a **runtime prerender**: more of the page is already rendered before the user clicks, with fewer loading spinners.
+More of the page is rendered before the user clicks, with fewer loading states.
 
-Generating the runtime prerender costs **a server invocation per prefetchable link**, so it is opt-in per route.
+Generating it costs **a server invocation per prefetchable link**, so it is opt-in per route.
 
 > **Good to know:** A cold cache (first visit, or after expiration) means the server still has to compute the cached result. Users may see a loading spinner on that first navigation. Subsequent navigations are instant as long as the cache is warm.
 
-## Example: a dashboard layout
+Like `searchParams`, `params` needs a `<Suspense>` boundary, even when the values come from [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params). A statically known param still belongs to one URL. Runtime prefetching resolves the values `generateStaticParams` does not cover.
```
aurorascharff: even when the values come from [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params

maybe, has been predefined? "comes from" is not super clear

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,121 @@
...
 
-Take a dashboard layout with a nav that depends on request data. Without runtime prefetching, `<UserNav>` always streams in behind a `<Suspense>` fallback after navigation, even though the cookie that determines its content is already known when the prefetch fires. With runtime prefetching, the router prefetches a prerender that resolves `<UserNav>` before the click:
+## Session data resolves in the shell
 
-```tsx filename="app/dashboard/layout.tsx"
-export const prefetch = 'allow-runtime'
-```
+Runtime prefetching is for URL data. Session data is handled separately: a route that reads `cookies()` or `headers()` gets an App Shell that includes its session data, cached per session on the client. That content is ready on navigation without `allow-runtime`.
```
aurorascharff: Well, its also for private caches?

"Runtime prefetching is for URL data."

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -1,86 +1,121 @@
...
 
-The route still needs a valid prerender, so `<UserNav>` stays behind a `<Suspense>` boundary:
+Derived data needs one more step. Take a dashboard nav that reads a cookie, then looks up content based on it:
```
aurorascharff: Hmmm, can we write a better bridge here?

It's not clear what you mean by derived data. What is the thing we're missing with the previous examples? In both of them, you need to cache whatever is below the session or URL data. But if it is depending on cookies, that's when you would need to use cache private.

aurorascharff (approved):

PR #95568, opened 2026-07-07
docs: add React performance profiling guide and react-performance-tracks skills

icyJoseph:
Profiling an App with React Dev Tools

- Step 1, fixing a data waterfall
- Step 2, narrow down Suspense boundaries
- Explain how both can be fixed with cache components
- Step 3, blocking keystroke

TODO:

- [ ] write an eval for the Skill
- [ ] rename guide, it is not about performance necessarily
- [ ] all the screenshots
- [ ] sync with repository for the walkthrough app
- [ ] dxagent run

Closes: https://linear.app/vercel/issue/DOC-6459/using-performance-tracks-and-suspense-dev-tools-in-nextjs-apps

### DIFF
--- a/docs/01-app/02-guides/profiling-performance.mdx
+++ b/docs/01-app/02-guides/profiling-performance.mdx
@@ -211,25 +211,30 @@ The Suspense view shows the same single late reveal wherever the slow child live
 The problem is that **one boundary wraps several pieces that load at different speeds**, so React can only reveal them together. Give each independently-paced piece its own boundary so each reveals as soon as its own data is ready:
 
 ```tsx filename="app/dashboard/page.tsx"
-<Suspense fallback={<StatsSkeleton />}>
-  <Stats />
-</Suspense>
-<Suspense fallback={<ActivitySkeleton />}>
-  <ActivityFeed teamIds={user.teamIds} />
-</Suspense>
-<Suspense fallback={<WhoIsOnlineSkeleton />}>
-  <WhoIsOnline teamIds={user.teamIds} />
-</Suspense>
+return (
+  <>
+    <Header user={user} teams={teams} notifications={notifications} />
+
+    <Suspense fallback={<StatsSkeleton />}>
+      <Stats />
+    </Suspense>
+    <Suspense fallback={<ActivitySkeleton />}>
+      <ActivityFeed teamIds={user.teamIds} />
+    </Suspense>
+    <Suspense fallback={<WhoIsOnlineSkeleton />}>
+      <WhoIsOnline teamIds={user.teamIds} />
+    </Suspense>
+  </>
+)
 ```
 
 See [Parallel streaming with sibling boundaries](/docs/app/guides/streaming#parallel-streaming-with-sibling-boundaries).
 
 Reload again. The Suspense view now shows three reveal events at different times. The `Stats` and `ActivityFeed` sections paint as soon as their own data is ready, and only the `WhoIsOnline` region keeps its fallback until it resolves.
 
-Splitting has two things to get right,
+The split turned one fallback into three. The `Header` renders from the shell data, so it is on screen at once, and only `Stats`, `ActivityFeed`, and `WhoIsOnline` wait behind a boundary.
 
-- keep everything that does not need the slow data (headings, layout, chrome) outside the boundaries, so only the data-dependent parts ever show a fallback, and
-- size each fallback to match the content it replaces, so a reveal swaps in place instead of shifting the page as sections pop in.
+Push each boundary down to the part that actually awaits data, and leave the rest outside. Size each fallback to match its content, so a reveal swaps in place instead of shifting the page as sections arrive.
 
 <details>
   <summary>Nesting boundaries does not serialize them</summary>
@@ -259,8 +264,6 @@ Both `RandomRoll` reads start together. On the **Server Components** track they 
 </details>
 
 {/* TODO: screenshot, Suspense view: three staggered reveals (after) */}
-
-One reveal is now several. Content shows sooner, but a page that fills in piece by piece can read as busier, even slower, than the single reveal it replaced. That is what the two rules above prevent: chrome stays in the static frame and fallbacks hold their space, so the reveals do not reshuffle the layout.
 
 The more of the page that is already there on load, the less has to stream behind a fallback at all.
 
--- a/docs/01-app/02-guides/profiling-performance.mdx
+++ b/docs/01-app/02-guides/profiling-performance.mdx
@@ -254,7 +254,7 @@ async function ActivityList({ teamIds }: { teamIds: string[] }) {
 
 Now the heading is on screen with the rest of the shell, and only the list streams. The page stops wrapping `ActivityFeed` in a boundary, because the boundary lives inside it.
 
-Splitting is only worth it when you can show more up front, the chrome and frames, and scope the skeleton to just the data that is still loading. Start with one boundary, lower it as far as the page benefits, and size each fallback to minimize layout shift.
+Splitting is only worth it when you can show more up front, the chrome and frames, and scope the skeleton to just the data that is still loading. Start with one boundary and push it down only as far as what you reveal still makes sense on its own. The heading reads fine without the list, but do not surface UI that means nothing until its data lands. Size each fallback to minimize layout shift.
 
 <details>
   <summary>Nesting boundaries does not serialize them</summary>
--- a/docs/01-app/02-guides/profiling-performance.mdx
+++ b/docs/01-app/02-guides/profiling-performance.mdx
@@ -228,11 +228,62 @@ Reload again. The Suspense view now shows three reveal events at different times
 
 Splitting has two things to get right. Keep everything that does not need the slow data (headings, layout, chrome) outside the boundaries, so only the data-dependent parts ever show a fallback. And size each fallback to match the content it replaces, so a reveal swaps in place instead of shifting the page as sections pop in.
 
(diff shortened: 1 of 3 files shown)


docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,367 @@
...
+}
+```
+
+Record a page load in the Performance panel and open the **Server Components** track. Observe the three spans: `getTeams` starts only when `getUser` resolves, and `getNotifications` only when `getTeams` does. Each `await` pauses the page render until it resolves. This is a **data waterfall**.
+
+Reads that do not depend on one another could be started at the same time instead, which would cut the total down to just the slowest read.
+
+> **Good to know:** Running independent reads at the same time is concurrency, not parallelism. These reads are I/O-bound, so each `await` mostly waits on a network or database response, and a single thread can keep several requests in flight at once. They overlap their waiting instead of taking turns.
+
+The Performance panel can save the recording as JSON, where every span carries its own start and duration, the same data the [`react-performance-tracks`](#automate-the-loop-with-a-skill) skill parses to find the waterfall.
```
aurorascharff: Wondering if this skill should be in agent-skills if its not nextjs specific? potentially with a nextjs reference page

aurorascharff: If it is indeed specific to Next since it uses agent browser, Next browser, and such. Then we should maybe call it next- something, maybe with what you are trying to do so it fits the other skills.

docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,427 @@
...
+Reads that do not depend on one another could be started at the same time instead, which would cut the total down to just the slowest read.
+
+> **Good to know:** Running independent reads at the same time is concurrency, not parallelism. These reads are I/O-bound, so each `await` mostly waits on a network or database response, and a single thread can keep several requests in flight at once. They overlap their waiting instead of taking turns.
+
+The Performance panel can save the recording as JSON, where every span carries its own start and duration, so the tracks can be read programmatically, not only on screen.
+
+<details>
+  <summary>Read the Server Components track as data</summary>
+
+In the Performance panel, **Save profile** writes the whole recording to JSON, and each bar you saw is an event in it. A single read looks like this:
```
aurorascharff: Feeling some AI with the amount of "each" and "every" so far

docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,427 @@
...
+<Suspense fallback={<WhoIsOnlineSkeleton />}>
+  <WhoIsOnline teamIds={user.teamIds} />
+</Suspense>
+```
+
+See [Parallel streaming with sibling boundaries](/docs/app/guides/streaming#parallel-streaming-with-sibling-boundaries).
+
+Reload again. The Suspense view now shows three reveal events at different times. The `Stats` and `ActivityFeed` sections paint as soon as their own data is ready, and only the `WhoIsOnline` region keeps its fallback until it resolves.
+
+Splitting has two things to get right,
```
aurorascharff: should this be a colon: 

"Splitting has two things to get right:"

docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,427 @@
...
+}
+```
+
+Both `RandomRoll` reads start together. On the **Server Components** track they run in parallel and the pair finishes at 2s, not 3s, even though one boundary sits inside the other. A `<Suspense>` orders when its content is revealed, not when its data is fetched. Nest boundaries for a staggered reveal, not to sequence the work behind them.
+
+</details>
+
+{/* TODO: screenshot, Suspense view: three staggered reveals (after) */}
+
+One reveal is now several. Content shows sooner, but a page that fills in piece by piece can read as busier, even slower, than the single reveal it replaced. That is what the two rules above prevent: chrome stays in the static frame and fallbacks hold their space, so the reveals do not reshuffle the layout.
```
aurorascharff: One reveal is now several. Content shows sooner, 
kinda LLMy

docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,427 @@
...
+
+Meanwhile the Scheduler track shows a single `Update` and **no** `Cascading Update`. So this is not a cascade. It is one correct render at the top dragging a synchronous re-render of unmemoized children below it, which is why the keystroke lands as one long `EventDispatch` on the main thread.
+
+</details>
+
+This is a **blocking interaction**. Keep the typed value urgent and let the expensive filter run at a lower priority with `useDeferredValue`, passing the deferred value to the list:
+
+```tsx filename="app/explorer/explorer.tsx"
+export function Explorer() {
+  const issues = useMemo(() => generateIssues(5000), [])
```
aurorascharff, line 381: Do we need usememo? Or should we promote react compiler?

docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,427 @@
...
+})
+```
+
+On its own, `useDeferredValue` is not enough here. Typing updates `query`, which re-renders `Explorer` on every keystroke, and that re-renders `Results` with it, filter and all. Wrapping `Results` in [`memo`](https://react.dev/reference/react/memo) lets it skip the urgent keystroke render, because its props (`issues` and the lagging `deferredQuery`) have not changed yet. The filter then runs once `deferredQuery` catches up, as a deferred, interruptible task. You need both: `useDeferredValue` to defer the work, and `memo` to stop the keystroke render from doing it anyway.
+
+Record again. The keystroke task is short (the input and its paint fit in one frame), and the filtering runs afterward as a separate, lower-priority task on the main thread.
+
+{/* TODO: screenshot, main thread: short keystroke, deferred filter (after) */}
+
+[Building interactive apps](/docs/app/guides/interactive-apps) goes deeper on keeping interactions responsive: transitions, optimistic updates, and memoizing the list so it does not re-render on every keystroke.
```
aurorascharff: and memoizing the list so it does not re-render on every keystroke.

we dont do this in interactive apps

docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,427 @@
...
+
+## Automate the loop with a skill
+
+Every step above is the same loop: one capture, one track, recognize the pattern. It does not need a human watching, so an agent can run it.
```
aurorascharff: Every step above is the same loop: one capture, one track, recognize the pattern.

llmy!

docs/01-app/02-guides/profiling-performance.mdx
```diff
@@ -0,0 +1,427 @@
...
+
+[`agent-browser`](https://www.npmjs.com/package/agent-browser) drives a real Chrome with React DevTools from the command line. It runs the same captures you did by hand (start and stop a performance trace, read the Suspense boundary classifier, record a render profile) and returns them as JSON an agent can parse.
+
+This repository ships a [`next-performance-profiling`](https://github.com/vercel/next.js/tree/canary/skills/next-performance-profiling) skill that runs this loop for an agent. It builds on [`next-dev-loop`](https://github.com/vercel/next.js/tree/canary/skills/next-dev-loop), which opens the browser and connects to `/_next/mcp`.
+
+Loaded against a running app, the skill:
```
aurorascharff, line 428: I looove that we end with a skill? Potentially, we could introduce it earlier as well...? They might not ever find out we have a skill for this!

also, when https://github.com/vercel/next.js/pull/95861 merges we need to remember to add this new skill there

skills/next-performance-profiling/SKILL.md
```diff
@@ -0,0 +1,113 @@
...
+- A `Promise.all` group is recorded on the track as only the read that gates it (the slowest to resolve); members that resolve earlier get no span at all, even on the Server Requests track. So after a parallel fix, expect fewer spans, not the same set overlapping. A missing span is not proof nothing ran. Never verify a parallel fix by counting spans; verify it by timing: the gating read now starts right after its prerequisite, and dependent work starts earlier.
+- The `react renders stop` "recording Xs" header is wall-clock since `start`, not your interaction. Confirm the capture from the change-reason column.
+- `[agent-browser] restore: missing` is expected on first run.
+
+## when you get lost
+
+An empty or malformed capture (no Server Components ⚛ track, an empty render profile, a trace of the wrong navigation) is almost always preflight: React DevTools not enabled at launch, a drifted page, or a keystroke that did not register. Re-run `next-dev-loop` preflight before touching app code. A capture that merely lacks a `Cascading Update` is not a failure; see mode B.
+
+## references
```
icyJoseph, line 111: hoist up

skills/next-performance-profiling/SKILL.md
```diff
@@ -0,0 +1,113 @@
+---
```
icyJoseph, line 1: Reference from the ai-agents guide

skills/next-performance-profiling/SKILL.md
```diff
@@ -0,0 +1,113 @@
...
+name: next-performance-profiling
+description: >
```
icyJoseph, line 3: Agent needs to have a goal 

- Suggest a series of refactors
- or rewrite the feature w/ tests to prevent regressions

skills/next-performance-profiling/SKILL.md
icyJoseph, line 1: x-ref to the guide for instructions on how we expect the steps to be performed

PR #95574, opened 2026-07-07
Add experimental WebSocket route handlers

timneutkens:
### What?

Add experimental Node.js App Route WebSocket handlers with `NextResponse.upgrade()` backed by bundled CrossWS hooks.

### Why?

This gives Route Handlers a first-class WebSocket upgrade path while preserving the normal request model: the route `GET` runs once, can inspect the request before acceptance, can return ordinary rejection responses, and can attach headers/cookies to the handshake response.

### How?

The implementation adds the experimental `webSocketRouteHandlers` flag, bundles CrossWS, extends generated App Route modules with a namespaced Adapter-facing `upgradeHandler(context, { node })`, wires Node server upgrade routing, manages accepted peer lifecycle for shutdown and dev invalidation, and records build/dev feature telemetry from the final Adapter-modified config.

### Verification

- `pnpm --filter=next types`
- `pnpm --filter=next build`
- `pnpm exec jest --runInBand test/unit/web-runtime/websocket-response.test.ts`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-webpack test/e2e/app-dir/websocket-route-handlers/websocket-route-handlers.test.ts -t "executes GET once|Adapter-facing upgradeHandler"`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-webpack test/e2e/telemetry/config.test.ts -t "webSocketRouteHandlers|Adapter-modified WebSocket"`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-turbo test/e2e/telemetry/config.test.ts -t "webSocketRouteHandlers|Adapter-modified WebSocket"`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-webpack test/e2e/telemetry/config.test.ts -t "next dev starts"`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-turbo test/e2e/telemetry/config.test.ts -t "next dev starts"`
- Not run: full WebSocket e2e matrix after the final namespaced Adapter signature change. The changed call paths were covered by the focused production test above.

### Stack status (2026-08-24)

Rebased onto canary `a1cc1eafb3`. CrossWS remains replaced by the vendored-ws bounded transport. This layer now also carries: previewProps threaded through the Node App Route request context (canary #96004), the merge-marker repair, registryScope dropped from the call site, user-facing message polish (`experimental.webSocketRouteHandlers`, generic 501 body), hooks-object validation cleanup, dev telemetry emission, and new unit coverage for the defensive 501/short-circuit paths plus shared e2e upgrade helpers. Types + route-handler unit suites pass at this tip (66 tests).


packages/next/src/server/websocket-upgrade.ts
```diff
@@ -0,0 +1,430 @@
...
+        const connection = pendingRequests.get(request)
+        if (!connection) {
+          throw new Error(
+            'Invariant: CrossWS upgrade request is missing its App Route response.'
+          )
+        }
+
+        return {
+          headers: connection.response.headers,
+          context: {
```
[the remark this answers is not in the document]
timneutkens: :robot: Done -- the CrossWS upgrade path now clones and filters handshake headers with getUpgradeResponseHeaders(), which strips x-middleware-set-cookie while preserving ordinary response headers and cookies. Added unit coverage for the filtered handshake headers.

packages/next/src/server/lib/render-server.ts
```diff
@@ -178,7 +178,7 @@ async function initializeImpl(opts: {
     requestHandler,
     upgradeHandler,
     server,
-    closeUpgraded() {
+    async closeUpgraded() {
```
timneutkens: Made async but there is no await used and `.close` is not returned so this doesn't do anything as far as I can tell.

packages/next/src/server/route-modules/app-route/module.ts
```diff
@@ -764,6 +779,26 @@ export class AppRouteRouteModule extends RouteModule<
...
     }
 
+    if (isWebSocketUpgradeResponse(res)) {
+      let message: string | undefined
+      if (this.nextConfigOutput === 'export') {
+        message = 'NextResponse.upgrade() cannot be used with output: "export".'
+      } else if (workStore.forceStatic) {
+        message =
+          'NextResponse.upgrade() cannot be used in a route configured with dynamic = "force-static".'
+      } else if (workStore.isStaticGeneration) {
```
timneutkens: Are tests added to cover these cases?

packages/next/src/server/websocket-upgrade.ts
```diff
@@ -0,0 +1,438 @@
...
+}
+
+interface ConnectionContext {
+  metadata: WebSocketUpgradeMetadata
+  response: Response
+  transportContext: WebSocketUpgradeTransportContext
+}
+
+function validateHeaderPart(value: string, name: string): void {
+  if (/\r|\n/.test(value)) {
```
timneutkens: This validation looks incorrect/excessive. Probably not needed.

packages/next/src/server/websocket-upgrade.ts
```diff
@@ -0,0 +1,438 @@
...
+      lines.push(`${name}: ${value}`)
+    }
+  })
+
+  return lines
+}
+
+export function getUpgradeResponseHeaders(response: Response): Headers {
+  const headers = new Headers(response.headers)
+  headers.delete('x-middleware-set-cookie')
```
timneutkens: Is this expected? Is there a case where this is hit?

packages/next/src/server/lib/render-server.ts
DavidIlie: :robot: Fixed in the split stack. render-server now awaits settleWebSocketShutdownStages(), which closes both its WebSocket registry scope and the bundler service. router-server likewise awaits its own scope and the nested renderer cleanup, and shutdown callers await closeUpgraded().

packages/next/src/server/route-modules/app-route/module.ts
DavidIlie: :robot: Yes. The stack now covers each boundary: output: export is rejected in packages/next/src/server/config.test.ts, force-static is covered by the WebSocket Route Handler static-contract E2E, and prerendered/ISR runtime upgrades are rejected by the standalone suite. Cache Components also has a dedicated asynchronously returned upgrade-marker contract.

packages/next/src/server/websocket-upgrade.ts
DavidIlie: :robot: Reworked. The ad-hoc CR/LF helper is gone. Ordinary raw responses now use Node's validateHeaderName()/validateHeaderValue(). A 101 application header additionally has to be visible ASCII because the handshake lines are serialized as raw header bytes, and protocol/framing headers are rejected because the transport owns them. Tests cover control and non-ASCII values and assert that no handshake bytes are committed.

packages/next/src/server/websocket-upgrade.ts
DavidIlie: :robot: Yes. A raw fallback or handshake write can return false when the socket reaches its writable high-water mark. The current writeSocket() waits for drain while racing close/error and listener-installation reentrancy. Focused tests cover the backpressured drain path, client disconnects, synchronous write failures, and listener-cleanup failures.

PR #95578, opened 2026-07-07
turbo-tasks-backend: parent_count garbage collection for the persistent engine

lukesandberg:
### What?

Adds a garbage collector to the persistent (`turbo-tasks-backend`) engine that deletes tasks which have become unreachable from the live task graph, from both memory and the on-disk cache. Opt-in behind `TURBO_ENGINE_GC=1` for now; off by default.

### Why?

When a task is invalidated it can become **disconnected** from the live graph, but today it is never deleted. Canonical case: task `A` reads `read_file("old.txt")` via a child task `T`; the file is renamed, `A` re-executes and now reads `read_file("new.txt")` via a new task `T'`, and `CleanupOldEdges` drops `T` from `A`'s children. `T` is now unreachable from any root — but it still sits in the in-memory map and gets **re-persisted to disk on every snapshot**, living forever.

In practice this causes the persisted cache (and memory) to grow without bound whenever a project's common dependencies are reorganized — e.g. re-rooting a project, or churning dependencies — because each churn leaves a whole generation of tasks behind. GC reclaims them.

### How?

The core idea is an incrementally-maintained **`parent_count`**: the number of persistent parents that list a task as a child. A count can only drop to 0 while the parent is *in memory* (a parent re-executed and dropped the child), so a task becoming collectible is detectable at that exact moment — no whole-DB scan, no restoring all metadata.

- **`parent_count` maintenance.** A persistent `parent_count: u32` (lazy meta field, to keep `TaskStorage` within its 128-byte budget) is maintained at the only two child-edge sites: `+1` per genuinely-new persistent child in `connect_children`, `-1` per actually-removed persistent child in `CleanupOldEdges`. The delta rides a new durable `AggregationUpdateJob::AdjustParentCount`, so it is captured mid-flight by a snapshot and replayed to completion on restart (crash-consistent). A separate transient `transient_ref_count` covers in-session references that aren't persistent parent edges (a transient parent, or an out-of-graph pin).

- **Collection.** When a count reaches 0 the task is recorded as a GC candidate. A GC pass (idle-gated in the background job, immediately before the snapshot, and at `stop()`) runs under an execution-exclusion phase: it re-validates each candidate, scrubs its reverse-dependency edges off its targets, removes it from the in-memory map, cascades `-1` to its children (collecting the whole unreachable subtree), and tombstones the on-disk copy in the same atomic snapshot commit. A task resurrected in the small window between the GC phase and the snapshot is re-validated and not wrongly tombstoned.

- **Pinning escape hatch.** Values that escape the tracked graph — a `Vc` handed to JS across the NAPI boundary via `DetachedVc`, or anything guarded by `prevent_gc()` — are pinned (counted `transient_ref_count`) so they are never collected while the handle is alive. `DetachedVc` pins its `OperationVc`'s task on creation/clone and unpins on drop. `prevent_gc()` (previously an empty stub) now pins the current task.

- **Persistence plumbing.** `save_snapshot` gained a delete list that tombstones `TaskMeta`/`TaskData`/`TaskCache` in the same commit as the snapshot puts, so removal from disk is atomic with the rest of the cycle.

**Verification.** New backend tests cover `parent_count` maintenance, durability across a DB reopen (increment and decrement), collection of a disconnected subtree via the cascade, pinned/detached tasks surviving collection, and a re-rooting stress test that asserts the persistent task set stays flat across 20 dependency swaps (vs. growing ~linearly without GC) while GC does the collecting. The full backend suite is green with `TURBO_ENGINE_GC=1`, including under `--features verify_aggregation_graph`.

**Not in this PR / follow-ups.** GC is opt-in until it has had a clean CI run and dogfooding on trusted apps. The collection pass is sequential for now — parallelizing it hit a pre-existing fragility in `scope_and_block` on thread-limited runtimes (documented inline); making that robust would also benefit the snapshot path and is deferred. Transient-root (`run_once`/`Once`) references are not released on disposal (bounded, harmless — those roots aren't collected anyway), to revisit before enabling by default.

### DIFF
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -41,8 +41,12 @@ use crate::next_api::turbopack_ctx::NextTurbopackContext;
 /// [`turbo_tasks::OperationValue`] and should be dereferenced to an [`OperationVc`] before being
 /// passed to a [`turbo_tasks::function`].
 //
-// TODO: If we add a tracing garbage collector to turbo-tasks, this should be tracked as a GC root.
-#[derive(Clone)]
+/// A `DetachedVc` holds its operation's task alive against garbage collection for as long as the
+/// handle (or any clone) exists: it is a reference that escapes the tracked task graph, so no
+/// persistent parent lists the task as a child and GC would otherwise collect it. Each live
+/// `DetachedVc` counts as one transient reference (pin) on the task; the pin is released when the
+/// last clone is dropped. An `OperationVc` is statically a `TaskOutput`, so the pinned `TaskId` is
+/// recovered synchronously from the `RawVc`.
 pub struct DetachedVc<T> {
     turbopack_ctx: NextTurbopackContext,
     /// The Vc. Must be unresolved, otherwise you are referencing an inactive operation.
@@ -51,11 +55,36 @@ pub struct DetachedVc<T> {
 
 impl<T> DetachedVc<T> {
     pub fn new(turbopack_ctx: NextTurbopackContext, vc: OperationVc<T>) -> Self {
+        // Pin the operation's task so GC treats this out-of-graph handle as a root.
+        turbopack_ctx.turbo_tasks().pin_task_for_gc(vc.task_id());
+
         Self { turbopack_ctx, vc }
     }
 
     pub fn turbopack_ctx(&self) -> &NextTurbopackContext {
         &self.turbopack_ctx
+    }
+}
+
+impl<T> Clone for DetachedVc<T> {
+    fn clone(&self) -> Self {
+        // Each live handle holds its own pin; balanced by this clone's `Drop`.
+        self.turbopack_ctx
+            .turbo_tasks()
+            .pin_task_for_gc(self.task_id());
+
+        Self {
+            turbopack_ctx: self.turbopack_ctx.clone(),
+            vc: self.vc,
+        }
+    }
+}
+
+impl<T> Drop for DetachedVc<T> {
+    fn drop(&mut self) {
+        self.turbopack_ctx
+            .turbo_tasks()
+            .unpin_task_for_gc(self.task_id());
     }
 }
 
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -68,15 +68,7 @@ impl<T> DetachedVc<T> {
 
 impl<T> Clone for DetachedVc<T> {
     fn clone(&self) -> Self {
-        // Each live handle holds its own pin; balanced by this clone's `Drop`.
-        self.turbopack_ctx
-            .turbo_tasks()
-            .pin_task_for_gc(self.task_id());
-
-        Self {
-            turbopack_ctx: self.turbopack_ctx.clone(),
-            vc: self.vc,
-        }
+        Self::new(self.turbopack_ctx.clone(), self.vc)
     }
 }
 
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -63,6 +63,8 @@ impl<T> DetachedVc<T> {
     }
 }
 
+// Manual clone and Drop routines are required for proper reference counting to prevent GC
+
 impl<T> Clone for DetachedVc<T> {
     fn clone(&self) -> Self {
         Self::new(self.turbopack_ctx.clone(), self.vc)
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -42,11 +42,8 @@ use crate::next_api::turbopack_ctx::NextTurbopackContext;
 /// passed to a [`turbo_tasks::function`].
 //
 /// A `DetachedVc` holds its operation's task alive against garbage collection for as long as the
-/// handle (or any clone) exists: it is a reference that escapes the tracked task graph, so no
-/// persistent parent lists the task as a child and GC would otherwise collect it. Each live
-/// `DetachedVc` counts as one transient reference (pin) on the task; the pin is released when the
-/// last clone is dropped. An `OperationVc` is statically a `TaskOutput`, so the pinned `TaskId` is
-/// recovered synchronously from the `RawVc`.
+/// handle exists: it is a reference that escapes the tracked task graph, so no
+/// persistent parent lists the task as a child and GC would otherwise collect it.
 pub struct DetachedVc<T> {
     turbopack_ctx: NextTurbopackContext,
     /// The Vc. Must be unresolved, otherwise you are referencing an inactive operation.
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -358,7 +358,10 @@ impl TurboTasksBackend {
     /// The GC phase guarantees no operation is executing (see [`SnapshotCoordinator::begin_gc`]),
     /// so the cross-task scrub and the map removals cannot race a concurrent connect/read.
     /// Returns the number of tasks collected.
-    fn gc_collect(&self, turbo_tasks: &TurboTasks<TurboTasksBackend>) -> usize {
+    fn gc_collect(
+        &self,
+        turbo_tasks: &TurboTasks<TurboTasksBackend>,
+    ) -> (usize, Vec<TaskDeletion>) {
         // Seed the worklist from the candidates observed since the last pass. Re-validation happens
         // per task below (a candidate may have been revived by a concurrent re-connect before the
         // phase was acquired, or may not be Data-resident).
@@ -402,15 +405,12 @@ impl TurboTasksBackend {
             }
         }
 
-        if !deletions.is_empty() {
-            self.pending_gc_deletes.lock().extend(deletions);
-        }
         // Re-queue candidates that are still parentless but were not collectible this pass, so a
         // later pass retries once their transient activeness clears (or their Data is restored).
         if !retain.is_empty() {
             self.gc_candidates.lock().extend(retain);
         }
-        collected
+        (collected, deletions)
     }
 
     /// Scrubs all references from `task_id` off its dependency targets, removes it from the
@@ -522,13 +522,20 @@ impl TurboTasksBackend {
         (children, deletion)
     }
 
-    /// Runs a full GC pass under the GC phase and returns the number of tasks collected. Test-only
-    /// hook. Callers must be idle (no task executing).
+    /// Runs a full GC pass under the GC phase and returns the number of tasks collected. Buffers
+    /// the resulting tombstones into `pending_gc_deletes` so a subsequent
+    /// `snapshot_and_evict_for_testing` commits them (production runs GC inline in
+    /// `snapshot_and_persist` instead). Test-only hook; callers must be idle (no task
+    /// executing).
     #[doc(hidden)]
     pub fn gc_for_testing(&self, turbo_tasks: &TurboTasks<TurboTasksBackend>) -> usize {
         let _serialize = self.snapshot_in_progress.lock();
         let _gc_phase = self.snapshot_coord.begin_gc();
-        self.gc_collect(turbo_tasks)
+        let (collected, deletes) = self.gc_collect(turbo_tasks);
+        if !deletes.is_empty() {
+            self.pending_gc_deletes.lock().extend(deletes);
+        }
+        collected
     }
 
     fn operation_suspend_point(&self, suspend: impl FnOnce() -> AnyOperation) {
@@ -596,11 +603,11 @@ impl TurboTasksBackend {
             .unwrap_or(0)
     }
 
-    /// The transient `transient_parent_count` of a resident task (0 if absent or not resident).
+    /// The transient `transient_ref_count` of a resident task (0 if absent or not resident).
     #[doc(hidden)]
-    pub fn transient_parent_count_for_testing(&self, task: TaskId) -> u32 {
+    pub fn transient_ref_count_for_testing(&self, task: TaskId) -> u32 {
         self.storage
-            .with_task(task, |t| t.gc_transient_parent_count())
+            .with_task(task, |t| t.gc_transient_ref_count())
             .unwrap_or(0)
     }
 
@@ -1225,11 +1232,30 @@ impl TurboTasksBackend {
         // request bit, suspended_operations) assumes only one snapshot runs at
         // a time. Held for the entire snapshot lifecycle.
         let _snapshot_in_progress = self.snapshot_in_progress.lock();
-        // Drain GC tombstones buffered since the last snapshot (tasks the GC pass removed from
-        // memory whose on-disk copies must now be tombstoned). Drained before the early-return
-        // checks below so a GC-only cycle (deletes but no modifications) still commits its
-        // tombstones.
-        let deletes = std::mem::take(&mut *self.pending_gc_deletes.lock());
+
+        // Garbage-collection pass runs immediately before the snapshot, under the same held
+        // `snapshot_in_progress` lock: it tears down tasks whose persistent parent_count reached 0
+        // (removing them from the map) and returns their on-disk tombstones, which this snapshot
+        // then commits atomically. Doing it here — rather than as a separate background step —
+        // means a collected task is neither re-persisted nor left on disk: it is tombstoned in the
+        // very commit that would otherwise have re-written it, and the tombstone is re-validated
+        // against the post-drain map below so a task resurrected in the brief window between the GC
+        // phase and the snapshot drain is not wrongly tombstoned. Opt-in via TURBO_ENGINE_GC.
+        let gc_deletes = if *GC_ENABLED {
+            let _gc_span = tracing::info_span!(parent: parent_span.clone(), "gc").entered();
+            let _gc_phase = self.snapshot_coord.begin_gc();
+            let (collected, deletes) = self.gc_collect(turbo_tasks);
+            if collected > 0 {
+                tracing::info!(collected, "gc pass collected tasks");
+            }
+            deletes
+        } else {
+            Vec::new()
+        };
+        // Also drain any tombstones buffered by the test-only `gc_for_testing` hook (production
+        // takes the inline `gc_deletes` path above; both feed the same commit).
+        let mut deletes = gc_deletes;
+        deletes.append(&mut self.pending_gc_deletes.lock());
         let start = Instant::now();
         // SystemTime for wall-clock timestamps in trace events (milliseconds
         // since epoch). Instant is monotonic but has no defined epoch, so it
@@ -1244,6 +1270,14 @@ impl TurboTasksBackend {
         // Enter snapshot mode, which atomically reads and resets the modified count.
         // Checking after start_snapshot ensures no concurrent increments can race.
         let (snapshot_guard, has_modifications) = self.storage.start_snapshot();
+
+        // Re-validate GC tombstones against the now-drained map: `begin_snapshot` has drained/
+        // suspended all operations, so the map is stable. GC removed each collected task from the
+        // map; if one is resident again it was resurrected by an operation in the brief window
+        // between the GC phase ending and this drain (a concurrent read re-connected it, restoring
+        // it from its not-yet-tombstoned on-disk copy). Such a task is live again — drop its
+        // tombstone so we don't delete a task the snapshot is about to (re-)persist.
+        deletes.retain(|d| self.storage.with_task(d.task_id, |_| ()).is_none());
 
         let suspended_operations = snapshot_phase.take_suspended_operations();
 
@@ -1692,30 +1726,19 @@ impl TurboTasksBackend {
             self.is_idle.store(false, Ordering::Release);
             self.verify_aggregation_graph(turbo_tasks, false);
         }
+        // The task_cache is a pure perf cache backed by the DB and isn't read during the stop
+        // snapshot (no task creation runs concurrently with stop). Drop it before persisting to
+        // lower peak memory during the serialization/write.
+        self.storage.drop_task_cache();
         if self.should_persist() {
-            // Run a final GC pass before the stop snapshot so builds emit tombstones for
-            // disconnected tasks in the final commit, keeping the persisted DB tight for the next
-            // build. At shutdown the system is already quiescing, so this runs to completion; its
-            // tombstones are buffered and drained by the stop snapshot below. Must run before
-            // `drop_task_cache` so teardown's task_cache scrubbing stays consistent.
-            if *GC_ENABLED {
-                let _serialize = self.snapshot_in_progress.lock();
-                let _gc_phase = self.snapshot_coord.begin_gc();
-                let collected = self.gc_collect(turbo_tasks);
-                tracing::info!(collected, "stop GC pass");
-            }
-            // The task_cache is a pure perf cache backed by the DB and isn't read during the
-            // stop snapshot (no task creation runs concurrently with stop). Drop it before
-            // persisting to lower peak memory during the serialization/write.
-            self.storage.drop_task_cache();
+            // The stop snapshot runs a final GC pass internally (see `snapshot_and_persist`), so
+            // builds emit tombstones for disconnected tasks in the final commit — keeping the
+            // persisted DB tight for the next build.
             if let Err(err) =
                 self.snapshot_and_persist(Span::current().into(), SnapshotReason::Stop, turbo_tasks)
             {
                 eprintln!("Persisting failed during shutdown: {err:?}");
             }
-        } else {
-            // eagerly drop the task cache
-            self.storage.drop_task_cache();
         }
         self.storage.drop_contents();
         if let Err(err) = self.backing_storage.shutdown() {
@@ -3225,6 +3248,7 @@ impl TurboTasksBackend {
                         // grouped together in trace viewers.
                         let background_span =
                             tracing::info_span!(parent: None, "background snapshot");
+
                         match self.snapshot_and_persist(background_span.id(), reason, turbo_tasks) {
                             Err(err) => {
                                 // save_snapshot consumed persisted_task_cache_log entries;
@@ -3261,27 +3285,6 @@ impl TurboTasksBackend {
                                             _ = std::future::ready(()) => false,
                                         }
                                     }};
-                                }
-
-                                // Garbage-collection pass: tear down tasks whose persistent
-                                // parent_count reached 0. Runs BEFORE eviction so freshly-collected
-                                // garbage is removed from the map before it could be evicted to
-                                // disk-only. Holds the GC phase (excludes execution) via
-                                // `gc_for_testing`'s shape, and buffers on-disk tombstones for the
-                                // *next* snapshot commit. Opt-in via TURBO_ENGINE_GC.
-                                if *GC_ENABLED {
-                                    if check_idle_ended!() {
-                                        continue 'outer;
-                                    }
-                                    let _gc_span =
-                                        tracing::info_span!(parent: background_span.id(), "gc")
-                                            .entered();
-                                    let _serialize = self.snapshot_in_progress.lock();
-                                    let _gc_phase = self.snapshot_coord.begin_gc();
-                                    let collected = self.gc_collect(turbo_tasks);
-                                    if collected > 0 {
-                                        tracing::info!(collected, "gc pass collected tasks");
-                                    }
                                 }
 
                                 // Evict persisted tasks from memory to reclaim space.
@@ -4018,16 +4021,22 @@ impl Backend for TurboTasksBackend {
     }
 
     fn pin_task_for_gc(&self, task: TaskId, _turbo_tasks: &TurboTasks<Self>) {
-        // Set the transient `pinned` flag. Pinned tasks are unevictable (see `evictability`) and
-        // are treated as GC roots (see `is_gc_root`), so `prevent_gc` keeps the task — and the
-        // escaping references it anchors — alive. `pinned` is transient and never persisted; on
-        // restart the code that escaped the value re-establishes the pin (or the value is simply
-        // recomputed).
-        self.storage.access_mut(task).flags.set_pinned(true);
+        // A pin is an in-session reference from outside the tracked graph (an explicit
+        // `prevent_gc`, or a detached handle like `DetachedVc` holding the task's
+        // `OperationVc` across NAPI). It is counted the same way a transient parent's edge
+        // is: bump `transient_ref_count`. While that count is > 0 the task is uncollectible
+        // (see `is_gc_collectible`) and unevictable (see `evictability`, so the transient
+        // count can't be lost). Counting (rather than a bool flag) makes nested/cloned pins
+        // correct — each pin is balanced by its own unpin.
+        self.storage
+            .access_mut(task)
+            .gc_increment_transient_ref_count();
     }
 
     fn unpin_task_for_gc(&self, task: TaskId, _turbo_tasks: &TurboTasks<Self>) {
-        self.storage.access_mut(task).flags.set_pinned(false);
+        self.storage
+            .access_mut(task)
+            .gc_decrement_transient_ref_count();
     }
 
     fn connect_task(
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -1,6 +1,7 @@
 mod cell_data;
 mod counter_map;
 mod eviction;
+mod gc;
 mod operation;
 mod snapshot_coordinator;
 mod storage;
@@ -15,7 +16,7 @@ use std::{
     pin::Pin,
     sync::{
         Arc, LazyLock,
-        atomic::{AtomicBool, AtomicUsize, Ordering},
+        atomic::{AtomicBool, Ordering},
     },
     time::SystemTime,
 };
@@ -44,7 +45,7 @@ use turbo_tasks::{
     macro_helpers::NativeFunction,
     message_queue::{TimingEvent, TraceEvent},
     registry::get_value_type,
-    scope::{scope_and_block, scope_self_feeding},
+    scope::scope_and_block,
     task_statistics::TaskStatisticsApi,
     trace::TraceRawVcs,
     util::{IdFactoryWithReuse, good_chunk_size, into_chunks},
@@ -90,15 +91,6 @@ use crate::{
 /// If the number of dependent tasks exceeds this threshold,
 /// the operation will be parallelized.
 const DEPENDENT_TASKS_DIRTY_PARALLELIZATION_THRESHOLD: usize = 10000;
-
-/// When `TURBO_ENGINE_GC` is set to a truthy value, the background job runs a `parent_count`-driven
-/// garbage-collection pass (tearing down tasks whose persistent parent count reached 0). Opt-in
-/// until it has been proven on trusted apps; the eventual default-on flip gets its own escape
-/// hatch.
-static GC_ENABLED: LazyLock<bool> = LazyLock::new(|| {
-    std::env::var_os("TURBO_ENGINE_GC")
-        .is_some_and(|v| matches!(v.to_str(), Some("1" | "true" | "yes")))
-});
 
 /// Priority used to re-schedule a task that became stale during execution.
 ///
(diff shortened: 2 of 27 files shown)


lukesandberg:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/vercel/next.js/95578?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#95578** <a href="https://app.graphite.com/github/pr/vercel/next.js/95578?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95578?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#95695** <a href="https://app.graphite.com/github/pr/vercel/next.js/95695?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#95974](https://github.com/vercel/next.js/pull/95974) <a href="https://app.graphite.com/github/pr/vercel/next.js/95974?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `canary`


<h2></h2>

crates/next-napi-bindings/src/next_api/utils.rs
```diff
@@ -59,6 +66,28 @@ impl<T> DetachedVc<T> {
...
+    fn clone(&self) -> Self {
+        // Each live handle holds its own pin; balanced by this clone's `Drop`.
+        self.turbopack_ctx
+            .turbo_tasks()
+            .pin_task_for_gc(self.task_id());
+
+        Self {
+            turbopack_ctx: self.turbopack_ctx.clone(),
+            vc: self.vc,
+        }
```
lukesandberg: can we just call `new` here?

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -96,6 +96,15 @@ static IDLE_TIMEOUT: LazyLock<Duration> = LazyLock::new(|| {
         .unwrap_or(Duration::from_secs(2))
 });
 
+/// When `TURBO_ENGINE_GC` is set to a truthy value, the background job runs a `parent_count`-driven
+/// garbage-collection pass (tearing down tasks whose persistent parent count reached 0). Opt-in
+/// until it has been proven on trusted apps; the eventual default-on flip gets its own escape
+/// hatch.
+static GC_ENABLED: LazyLock<bool> = LazyLock::new(|| {
```
lukesandberg: we should expose this via an experimental next-config.js option  `turbopackEnableGc`

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -218,6 +227,21 @@ pub struct TurboTasksBackend {
...
+    /// have revived the task). Never collected directly from here.
+    gc_candidates: Mutex<FxHashSet<TaskId>>,
+
+    /// Tombstones for tasks the GC pass has removed from memory, awaiting the next persistence
+    /// commit. GC removes a collected task from the in-memory map immediately (under the GC
+    /// phase), but its already-persisted on-disk copy must be tombstoned in a `save_snapshot`
+    /// commit; each collected task's [`TaskDeletion`] is buffered here and drained into the
+    /// next snapshot. A crash before that commit just leaves the on-disk copy to be
+    /// re-collected — never a dangling reference.
+    pending_gc_deletes: Mutex<Vec<TaskDeletion>>,
```
lukesandberg: ideally we would _pass_ this directly to the take_snapshot instead of stashing on a new field here

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -286,6 +331,267 @@ impl TurboTasksBackend {
...
+        turbo_tasks: &'a TurboTasks<TurboTasksBackend>,
+    ) -> impl ExecuteContext<'a> {
+        ExecuteContextImpl::new_for_gc(self, turbo_tasks)
+    }
+
+    /// Records `task_id` as a garbage-collection candidate (its persistent `parent_count` reached
+    /// 0). Only non-transient tasks are tracked — transient tasks are never collected. Candidacy is
+    /// re-validated under exclusion before the task is actually collected, so a false positive here
+    /// (e.g. the count is bumped back up by a concurrent re-connect) is harmless.
+    fn note_gc_candidate(&self, task_id: TaskId) {
```
lukesandberg: add_gc_candidate or `record_gc_candidate` would be better

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -286,6 +331,267 @@ impl TurboTasksBackend {
...
+    fn gc_is_collectible(&self, task_id: TaskId) -> bool {
+        if task_id.is_transient() {
+            return false;
+        }
+        self.storage
+            .with_task(task_id, |t| t.is_gc_collectible())
+            .unwrap_or(false)
+    }
+
+    /// The persistent `parent_count` of a resident task (0 if absent or not resident).
```
lukesandberg: is zero accurate if not resident, it could be on disk with a non-zero value

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -286,6 +331,267 @@ impl TurboTasksBackend {
...
+    /// (tasks observed reaching `parent_count == 0`), re-validates each under the exclusion, and
+    /// tears down the ones that are still collectible: scrubbing their reverse-dependency edges,
+    /// decrementing their children's `parent_count` (cascading to any child that reaches 0),
+    /// removing them from the in-memory map + task_cache, and buffering an on-disk tombstone for
+    /// the next persistence commit.
+    ///
+    /// The GC phase guarantees no operation is executing (see [`SnapshotCoordinator::begin_gc`]),
+    /// so the cross-task scrub and the map removals cannot race a concurrent connect/read.
+    /// Returns the number of tasks collected.
+    fn gc_collect(
```
lukesandberg: consider moving this to a new parallel module

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
```diff
@@ -96,10 +96,38 @@ impl CleanupOldEdgesOperation {
...
+                                // A persistent parent losing a child drops that child's durable
+                                // `parent_count`; push it on the durable queue for
+                                // crash-consistency
+                                // (the transient-parent case is handled after `task` is dropped, to
+                                // avoid holding two task guards at once). Empty job is skipped.
+                                if !removed_persistent_children.is_empty()
+                                    && !task_id.is_transient()
+                                {
+                                    queue.push(AggregationUpdateJob::AdjustParentCount {
+                                        task_ids: removed_persistent_children.clone(),
```
lukesandberg: do we need this clone?

lukesandberg: ok, i see, `AdjustParentCount` should simply handle transient parents as well.

it shouldn't be handled inline below

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -183,6 +187,26 @@ impl<'e> ExecuteContextImpl<'e> {
...
+    /// The GC phase already excludes all concurrent operations and task execution, so taking an
+    /// operation guard here would be redundant — and would in fact deadlock, because
+    /// [`start_operation`](TurboTasksBackend::start_operation) blocks on the GC request bit that
+    /// the caller itself set. Only call this while a
+    /// [`GcPhase`](crate::backend::snapshot_coordinator::GcPhase) is held.
+    pub(super) fn new_for_gc(
+        backend: &'e TurboTasksBackend,
+        turbo_tasks: &'e TurboTasks<TurboTasksBackend>,
+    ) -> Self {
+        Self {
```
lukesandberg, line 239: should debug assert that we are in GC mode

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -218,6 +227,13 @@ pub struct TurboTasksBackend {
     /// `stop_and_wait`).
     snapshot_in_progress: Mutex<()>,
 
+    /// Tasks observed reaching `parent_count == 0` (no persistent parent) during operation
+    /// execution — candidates for garbage collection. Populated cheaply (a side-set write) in the
+    /// `AdjustParentCount` handler; the actual teardown/removal happens later under the GC phase's
+    /// exclusion (see `gc_collect`), where candidacy is re-validated (a concurrent re-connect may
+    /// have revived the task). Never collected directly from here.
+    gc_candidates: Mutex<FxHashSet<TaskId>>,
```
lukesandberg: perhaps an FxDashSet? a RoaringBitmap? a Vec?

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -286,6 +322,266 @@ impl TurboTasksBackend {
...
+        let mut worklist: Vec<TaskId> = self.gc_candidates.lock().drain().collect();
+        let mut collected = 0;
+        let mut deletions = Vec::new();
+        // Candidates that are still parentless but not *yet* collectible (e.g. residual transient
+        // activeness) are re-retained for a later pass.
+        let mut retain: Vec<TaskId> = Vec::new();
+
+        // Sequential worklist drain. This runs under the GC phase (execution is excluded) so it
+        // could in principle be parallelized, but `gc_delete_one` restores Data on demand, which
+        // uses `block_in_place`; fanning the worklist out with `scope_and_block` nests
```
lukesandberg: this comment is not accurate

restore_data does not use block-in-place

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -286,6 +322,266 @@ impl TurboTasksBackend {
...
+    /// 0. See [`GcOutcome`] for the cases.
+    fn gc_process_one(
+        &self,
+        task_id: TaskId,
+        turbo_tasks: &TurboTasks<TurboTasksBackend>,
+    ) -> GcOutcome {
+        if !self.gc_is_collectible(task_id) {
+            // Not collectible: keep retrying only while it is still parentless (and persistent);
+            // otherwise a concurrent re-connect revived it and it is no longer garbage.
+            return if self.gc_parent_count(task_id) == 0 && !task_id.is_transient() {
```
lukesandberg: both gc_is_collective and gc_parent_count peform dashmap lookups, then so does gc_delete_one

it would be better to access the task exactly once, grabbing a lock to do so is not concerning

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -286,6 +322,266 @@ impl TurboTasksBackend {
...
+            } else {
+                GcOutcome::Skip
+            };
+        }
+        let (children, deletion) = self.gc_delete_one(task_id, turbo_tasks);
+
+        // Cascade: each child loses a persistent parent; collect those that reach 0.
+        let mut ctx = self.execute_context_gc(turbo_tasks);
+        let mut newly_garbage = Vec::new();
+        for child in children {
```
lukesandberg: when we do parallelize it would make sense to parallelize here as well.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -286,6 +322,266 @@ impl TurboTasksBackend {
...
+
+        // Remove the task from the map. Its children set is dropped with it; we returned the child
+        // ids above so the caller can decrement their parent_count.
+        self.storage.remove_task(task_id);
+
+        // `surviving_task_ids` is left empty: it would carry any *other* live persistent task whose
+        // task type xxh3-collides with this one (the whole hash bucket is tombstoned). An 8-byte
+        // collision between two simultaneously-live tasks is astronomically rare (~2^32 tasks for
+        // one expected collision); if it ever occurred the survivor's mapping would be dropped and
+        // it would be transparently re-created on next lookup — a performance cost, not corruption.
```
lukesandberg: it would actually leak the tasks.  If we delete the hash mappings then a future lookup for the surviving task will fail and we will create a new task + task_id.  meaning the old task_id will be lost in the database

we need to implement surviving_task_ids, but it makes sense to consider alternate designs when we do

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -960,29 +1297,70 @@ impl TurboTasksBackend {
...
+        // the cascade and the snapshot, a task the cascade decremented cannot be resurrected before
+        // it is persisted — so the decremented children counts and the snapshot are consistent, and
+        // GC-collected tasks' tombstones ride this same commit.
+        //
+        // Without GC we just begin the snapshot directly. `deletes` are the GC tombstones for this
+        // commit: in the production (GC-enabled) path they are produced right here under the *same
+        // continuous exclusion* as the snapshot (the atomic `into_snapshot` hand-off), so nothing
+        // could have resurrected a collected task.
+        let (mut snapshot_phase, mut deletes) = if *GC_ENABLED {
+            let _gc_span = tracing::info_span!(parent: parent_span.clone(), "gc").entered();
```
lukesandberg: we should add a field for the number of deleted tasks

rather than an `info!` it should be a field on the span

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -960,29 +1297,70 @@ impl TurboTasksBackend {
...
+            let gc_phase = self.snapshot_coord.begin_gc();
+            let (collected, deletes) = self.gc_collect(turbo_tasks);
+            if collected > 0 {
+                tracing::info!(collected, "gc pass collected tasks");
+            }
+            let _span = tracing::info_span!("blocking").entered();
```
lukesandberg: blocking isn't a great name.... 

is this just to track the `waiting for operations to settle`?

lukesandberg: maybe the snapshot_coord struct should handle the tracing

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -960,29 +1297,70 @@ impl TurboTasksBackend {
...
-        };
         // Enter snapshot mode, which atomically reads and resets the modified count.
         // Checking after start_snapshot ensures no concurrent increments can race.
         let (snapshot_guard, has_modifications) = self.storage.start_snapshot();
 
+        // Belt-and-suspenders: drop the tombstone for any task that is resident again (would only
+        // happen via the non-atomic `gc_for_testing` path; the production path can't resurrect).
+        if !deletes.is_empty() {
+            deletes.retain(|d| self.storage.with_task(d.task_id, |_| ()).is_none());
+        }
```
lukesandberg: i don't understand this comment

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -562,6 +601,16 @@ impl TaskStorage {
...
+        // Tasks with a live transient reference (a `prevent_gc` pin, a detached handle, or a
+        // transient parent — see `transient_ref_count`) must stay fully resident: the count is a
+        // transient, session-only field, so evicting the task would silently lose it and expose the
+        // still-referenced task to collection.
+        if self.gc_transient_ref_count() > 0 {
+            return (
+                key_evictability,
+                ValueEvictability::Unevictable(UnevictableReason::Pinned),
+            );
+        }
```
lukesandberg: we could always evict everything but the transient count, which is the typical approach, drop_partial retains non-default transient fields

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -828,6 +877,80 @@ impl TaskStorage {
...
+    /// retain them; rather than tear down the aggregation overlay here (a miscount corrupts
+    /// activeness), we decline to collect and let a later pass take it once the edges clear.
+    /// Under-collecting is always safe.
+    pub fn is_gc_collectible(&self) -> bool {
+        self.gc_parent_count() == 0
+            && self.gc_transient_ref_count() == 0
+            && !self.is_gc_root()
+            && !self.flags.meta_restoring()
+            && !self.flags.data_restoring()
+            && self.flags.is_restored(TaskDataCategory::Meta)
```
lukesandberg: i don't think we need this condition, do we? i suppose if it isn't then `gc_parent_count` might be wrong?

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -294,13 +302,48 @@ impl TurboBackingStorage {
...
                 .unwrap_or_default();
 
+            // Apply GC tombstones after the parallel put phase (so no put/delete on the same key
+            // space is in-flight when we flush below) and after puts (so a delete wins over a
+            // racing stale put for the same id, though GC removes tasks from the map before the
+            // snapshot serializes so overlap should not occur). Applied sequentially: the delete
+            // list is small relative to the snapshot and ordering matters more than parallelism.
+            if !deletes.is_empty() {
+                let _span = tracing::trace_span!("delete tasks", count = deletes.len()).entered();
+                for TaskDeletion {
```
lukesandberg: should we map_collect across all task_ids

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -3645,6 +4013,25 @@ impl Backend for TurboTasksBackend {
...
+        // is: bump `transient_ref_count`. While that count is > 0 the task is uncollectible
+        // (see `is_gc_collectible`) and unevictable (see `evictability`, so the transient
+        // count can't be lost). Counting (rather than a bool flag) makes nested/cloned pins
+        // correct — each pin is balanced by its own unpin.
+        self.storage
+            .access_mut(task)
+            .gc_increment_transient_ref_count();
+    }
+
+    fn unpin_task_for_gc(&self, task: TaskId, _turbo_tasks: &TurboTasks<Self>) {
```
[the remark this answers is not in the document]
lukesandberg: :robot: Fixed in 0e21b1f01f. pin/unpin now take a lightweight operation guard (start_operation) so they run strictly before/after a GC pass, never interleaved with is_gc_collectible / task removal. Deadlock-free: neither caller holds a guard already (prevent_gc runs in the unguarded user-future region; DetachedVc pins from a NAPI thread) and the GC pass never calls pin/unpin. Also switched from access_mut to a non-inserting with_task_mut so a missing entry (a zombie OperationVc referencing an already-collected task) is surfaced with a debug_assert instead of resurrecting a blank zombie entry.

crates/next-napi-bindings/src/next_api/utils.rs
lukesandberg: :robot: Done — DetachedVc::clone now delegates to Self::new (single source of truth for the pin).

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done in 9819bab3e6 — dropped the pending_gc_deletes field. The production path keeps the GC-produced deletes as a local and threads them straight into save_snapshot; the test hook (gc_for_testing) now returns the deletes and snapshot_and_evict_with_deletes_for_testing forwards them.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Renamed to add_gc_candidate (trait + impls + caller) in e0832b5759.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Addressed in 6a80406266. gc_process_one now reads parent_count through the single restored ctx.task(All) guard, so the unwrap_or(0) can no longer mask a non-zero on-disk value for a non-resident task — the guard has restored Meta before the read. (And is_gc_collectible's Meta reads now go through check_access, which enforces the task was opened with Meta restored — see thread on line 950.)

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done in 6a80406266. gc_process_one takes a single ctx.task(id, All) guard that both decides collectibility and captures the edges to tear down (into a GcDeletePlan); the guard is dropped before gc_scrub_and_remove opens the target guards, and the same ExecuteContext is reused for the scrub and the cascade. Down from the previous 3 lookups (gc_is_collectible + gc_parent_count + gc_delete_one) to 1.

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
lukesandberg: :robot: Resolved in 6a80406266 + 55001920d5. The is_restored(Meta) check IS load-bearing (an evicted-Meta candidate reads parent_count/upper/followers as absent → 0/empty → falsely collectible), so rather than drop it we made it enforced-by-construction: is_gc_collectible moved onto TaskGuard and reads the Meta fields through the guard accessors, whose check_access asserts the task was opened with Meta restored. The hand-rolled is_restored/meta_restoring/data_restoring checks are gone.

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
lukesandberg: :robot: Done in e0832b5759. Folded transient-parent handling into the queue: added a transient (skip-serialized) AdjustTransientRefCount job that mirrors AdjustParentCount, so both connect_children and CleanupOldEdges just push one job and the handler adjusts the right counter. The inline second-guard loop (and the clone it needed) are gone.

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
lukesandberg: :robot: Done in e0832b5759 — new_for_gc now debug_asserts snapshot_coord.gc_in_progress() (added a gc_in_progress() accessor that reads the GC_REQUESTED_BIT).

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Fixed in e0832b5759. Corrected the comment: the parallelization deadlock isn't from block_in_place nesting (restore_data doesn't use it) — it's that scope_and_block spawns available_parallelism()-1 workers blocking on a Condvar, which starves thread-limited runtimes. Reworded to describe that and point at making scope_and_block robust as the real follow-up.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Fixed in 775920be4e. surviving_task_ids is dropped from TaskDeletion; survivors are now resolved at apply time in save_snapshot, which reads the authoritative on-disk bucket (get_multiple), tombstones it, and re-puts every id except the ones being deleted. The in-memory task_cache can't serve this (lazily populated, keyed by full task type not the hash). Added an end-to-end test that a disk-only colliding survivor is re-inserted.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done in e0832b5759 — the GC-collected count is now recorded as a field on the 'gc' span (collected = tracing::field::Empty, recorded after the pass) instead of a separate info! line.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done in e0832b5759 — renamed the span to 'await operations settle' (it's the wait for in-flight operations to drain before the snapshot/GC phase can begin), on both branches.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Reworded in e0832b5759. The re-validation only matters for the test-only gc_for_testing path, which runs GC in a separate exclusion window that has ended by snapshot time, so a task it collected could in principle be resurrected before the snapshot; the deletes.retain(...) after start_snapshot drops the tombstone for any such now-resident task. The production path runs GC under the same continuous exclusion as the snapshot, so it never resurrects.

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
lukesandberg: :robot: Done in 970316952c. Removed the pinned-task early return in evictability(); a task with transient_ref_count > 0 now falls through to normal Meta/Data eviction. drop_partial retains the transient count as residue (HasResidue keeps the map entry) so the pin can't be lost, while Meta/Data are reclaimed. Removed the now-unused UnevictableReason::Pinned. Covered by drop_partial_retains_transient_residue_{data,meta} + gc_does_not_collect_pinned_task.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Kept as Mutex<FxHashSet>. add_gc_candidate fires ~once per disconnected task and only when GC is enabled, so the single mutex isn't a real contention point, and FxHashSet gives free dedup (avoiding wasted re-validation of the same id). FxDashSet/Vec/RoaringBitmap didn't clear the cost/benefit bar for this access pattern.

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
lukesandberg: :robot: Reworked in 775920be4e (grouping deletes by hash, reading each bucket, re-putting survivors), but kept the delete phase sequential rather than map_collect/parallel: it runs after the parallel put phase and before the flush, the delete set is tiny relative to the snapshot, and ordering (tombstone-then-reput within a bucket) is easier to reason about serially. Happy to parallelize later if it ever shows up in a profile.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Agreed this would read better in its own module — leaving it as a follow-up rather than adding a file move to this PR. The GC methods are impl TurboTasksBackend touching private fields (gc_candidates, storage, snapshot_coord), so a split adds visibility plumbing that's better done once GC is proven and stable. Leaving this thread open to track it.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Agreed — when we parallelize the worklist drain the cascade should parallelize here too. Both are deferred together (see the gc_collect comment): parallelizing needs scope_and_block to be robust on thread-limited runtimes first. Leaving open to track.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Agreed — exposing this as an experimental next.config option (turbopackEnableGc) instead of only the TURBO_ENGINE_GC env var is a good follow-up. It's gated behind the env var for now while GC is opt-in; leaving open to track the config wiring.

crates/next-napi-bindings/src/next_api/utils.rs
```diff
@@ -41,8 +41,12 @@ use crate::next_api::turbopack_ctx::NextTurbopackContext;
 /// [`turbo_tasks::OperationValue`] and should be dereferenced to an [`OperationVc`] before being
 /// passed to a [`turbo_tasks::function`].
 //
-// TODO: If we add a tracing garbage collector to turbo-tasks, this should be tracked as a GC root.
-#[derive(Clone)]
+/// A `DetachedVc` holds its operation's task alive against garbage collection for as long as the
```
lukesandberg, line 44: simplify comment

turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
```diff
@@ -280,6 +280,27 @@ pub enum AggregationUpdateJob {
         lost_follower_ids: TaskIdVec,
         retry: u16,
     },
+    /// Adjust the persistent `parent_count` of each task in `task_ids` by `delta` (+1 when a
```
lukesandberg: simplfiy comments

turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
```diff
@@ -1445,6 +1467,35 @@ impl AggregationUpdateQueue {
...
+                    }
+                }
+                AggregationUpdateJob::AdjustTransientRefCount { task_ids, delta } => {
+                    // Session-only sibling of AdjustParentCount for edges from a transient parent.
+                    // Not persisted/replayed, and reaching 0 is not a collection trigger (only
+                    // losing a persistent parent is).
+                    for task_id in task_ids {
+                        let mut task = ctx.task(task_id, TaskDataCategory::Meta);
+                        task.update_and_get_transient_ref_count(delta);
+                    }
```
lukesandberg: we should use `for_each_meta` to optimize the updates

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
```diff
@@ -96,10 +96,42 @@ impl CleanupOldEdgesOperation {
...
+                                    } else {
+                                        AggregationUpdateJob::AdjustParentCount {
+                                            task_ids: removed_persistent_children,
+                                            delta: -1,
+                                        }
+                                    };
+                                    queue.push(job);
                                 }
                                 if is_aggregating_node(get_aggregation_number(&task)) {
+                                    drop(task);
```
lukesandberg, line 134: why add this here?

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_children.rs
```diff
@@ -33,13 +33,44 @@ pub fn connect_children(
...
+
     let new_follower_ids: SmallVec<_> = new_children.into_iter().collect();
 
     let aggregating_node = is_aggregating_node(parent_aggregation);
     let upper_ids = (!aggregating_node).then(|| get_uppers(&parent_task));
 
     drop(parent_task);
 
+    // Maintain the child-side parent reference count now that the parent guard is released. Each
+    // newly-connected persistent child gains a parent: a persistent parent bumps the durable
```
lukesandberg: is this the best place?  could we instead perform the increment when launching the child?

i suppose this is complex due to cache hits?  perhaps as part of the activeness adjustment

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -94,6 +94,10 @@ pub trait ExecuteContext<'e>: Sized {
         T: Clone + Into<AnyOperation>;
     fn should_track_dependencies(&self) -> bool;
     fn should_track_activeness(&self) -> bool;
+    /// Record that `task_id`'s persistent `parent_count` reached 0, making it a garbage-collection
+    /// candidate. This only notes candidacy (a cheap side-set write); the actual teardown/removal
+    /// happens later under the GC phase's exclusion, where candidacy is re-validated.
+    fn add_gc_candidate(&self, task_id: TaskId);
```
lukesandberg: as an alternate design we _could_ not maintain a set and instead just perform a parallel scan of the task storage table.

While this would be O(N) work instead of O(garbage) it is fast in practice and would be a natural way to parallize at least the initial part of the GC path

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -183,6 +187,30 @@ impl<'e> ExecuteContextImpl<'e> {
         }
     }
 
+    /// Constructs a context that does NOT take an operation guard, for use by the garbage
+    /// collector while it holds the coordinator's GC phase.
+    ///
+    /// The GC phase already excludes all concurrent operations and task execution, so taking an
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -1104,6 +1136,70 @@ pub trait TaskGuard: Debug + TaskStorageAccessors {
         new_value
     }
 
+    /// Adjust the count of **persistent** parents referencing this task by `delta` (+1 when a
+    /// persistent parent connects it as a child, -1 when one disconnects it) and return the new
+    /// value. A value of 0 means no persistent parent lists this task — a prerequisite for
+    /// collection. Panics in debug on underflow (a decrement below 0 means the count drifted from
+    /// the true number of `children`-edge references — the invariant this field must uphold).
+    fn update_and_get_parent_count(&mut self, delta: i32) -> u32 {
+        let current = self.get_parent_count().copied().unwrap_or(0);
```
lukesandberg, line 1221: as an inline default-ed field why is this an Option?

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -1104,6 +1136,70 @@ pub trait TaskGuard: Debug + TaskStorageAccessors {
...
+        // collected.
+        !self.id().is_transient()
+            && self.get_parent_count().copied().unwrap_or(0) == 0
+            && self.get_transient_ref_count().copied().unwrap_or(0) == 0
+            // GC root: active (a root/once task, positive active counter, or active-until-clean) or
+            // in progress (about to connect children).
+            && self.get_activeness().is_none()
+            && self.get_in_progress().is_none()
+            && self.iter_upper().next().is_none()
+            && self.iter_followers().next().is_none()
```
lukesandberg: the comment about the aggregation graph should really just be moved down here

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -951,29 +1001,81 @@ impl TurboTasksBackend {
...
+        // Run the GC pass and transition to the snapshot **atomically**, without ever releasing
+        // operation exclusion in between. `begin_gc` drains all operations; `gc_collect` tears down
+        // collectible tasks and cascades `parent_count` decrements to their children; then
+        // `GcPhase::into_snapshot` hands the exclusion straight to the snapshot phase (swapping the
+        // GC request bit for the snapshot bit under one lock). Because no operation can run between
+        // the cascade and the snapshot, a task the cascade decremented cannot be resurrected before
+        // it is persisted — so the decremented children counts and the snapshot are consistent, and
+        // GC-collected tasks' tombstones ride this same commit.
+        //
+        // Without GC we just begin the snapshot directly. `deletes` are the GC tombstones for this
```
lukesandberg: this comment could be massively simplified

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -951,29 +1001,81 @@ impl TurboTasksBackend {
...
         let wall_start = SystemTime::now();
         debug_assert!(self.should_persist());
 
-        let mut snapshot_phase = {
-            let _span = tracing::info_span!("blocking").entered();
-            self.snapshot_coord.begin_snapshot()
-        };
         // Enter snapshot mode, which atomically reads and resets the modified count.
         // Checking after start_snapshot ensures no concurrent increments can race.
         let (snapshot_guard, has_modifications) = self.storage.start_snapshot();
```
lukesandberg, line 1069: it is a little confusing that this isn't the responsibility of the snapshot coordinator

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1411,15 +1515,20 @@ impl TurboTasksBackend {
...
+        // lower peak memory during the serialization/write.
         self.storage.drop_task_cache();
         if self.should_persist() {
-            // The task_cache is a pure perf cache backed by the DB and isn't read during the
-            // stop snapshot (no task creation runs concurrently with stop). Drop it before
-            // persisting to lower peak memory during the serialization/write.
-            if let Err(err) =
-                self.snapshot_and_persist(Span::current().into(), SnapshotReason::Stop, turbo_tasks)
-            {
+            // The stop snapshot runs a final GC pass internally (see `snapshot_and_persist`), so
```
lukesandberg: final is weird, this would be the only GC pass

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -3701,6 +3817,15 @@ impl Backend for TurboTasksBackend {
         self.mark_own_task_as_finished(task_id, turbo_tasks);
     }
 
+    fn pin_task_for_gc(&self, task: TaskId, _turbo_tasks: &TurboTasks<Self>) {
+        // GC pin/unpin bookkeeping lives in the `gc` module (see `gc_pin`).
```
lukesandberg: drop comment

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -142,6 +142,39 @@ struct TaskStorageSchema {
...
+    /// - a **detached handle** that holds this task's `OperationVc` outside the tracked graph
+    ///   (e.g. a `DetachedVc` passed to JS across the NAPI boundary), which pins it like a GC
+    ///   root.
+    ///
+    /// A persistent task with `parent_count == 0` but `transient_ref_count > 0` is kept alive
+    /// in-session (uncollectible) — it is still referenced, just not by a persistent parent. On
+    /// restart these references are re-established (transient parents re-execute; detached handles
+    /// are re-created), so the count is never persisted.
+    #[field(storage = "direct", category = "transient", inline, default)]
+    transient_ref_count: u32,
```
lukesandberg, line 176: we should consider more efficient storage

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -7,12 +7,53 @@ use crate::{
...
+/// `get_or_create_task`) flows, so no `deleted` task can be returned to live use without reviving.
+fn resurrect_if_deleted(task_id: TaskId, ctx: &mut impl ExecuteContext<'_>) {
+    // Cheap Meta-only check first; only escalate on a hit. `deleted` is a Meta-category flag, so a
+    // Meta guard suffices to test/clear it — `immutable()` (a Data-category flag) is NOT checked
+    // here to avoid forcing a Data restore on the common not-deleted path.
+    let deleted = {
+        let mut task = ctx.task(task_id, TaskDataCategory::Meta);
+        let deleted = task.gc_is_deleted();
+        if deleted {
+            task.gc_clear_deleted();
```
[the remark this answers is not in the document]
lukesandberg: this is not possible

if we commit successfully then the task_cache entry sticks arround until eviction at which point it is removed under the same lock as the task itself. so there is no such race

lukesandberg:
Superseded by a stacked decomposition of this work into 4 reviewable PRs (each builds and tests standalone; the top of the stack reconstructs this branch exactly):

- #95974 — `scope_self_feeding` (parallelism primitive)
- #95975 — persistence delete/tombstone plumbing (Stage 0)
- #95976 — parent_count reference counting + GC collection (core)
- #95977 — GC roots, Data-residency gate, and app integration

Closing in favor of the stack.

PR #95579, opened 2026-07-07
Turbopack: order CSS modules by chunk-group co-occurrence in linearize

sokra:
## Summary

The graph-based CSS chunker's `linearize` step produces the global module order that `split_into_chunks` then cuts into chunks. It previously broke ties between ready modules using a weight-sorted stack.

It now prefers the ready module that shares the most chunk groups with the previously placed module, so modules that are loaded together end up adjacent in the global order and split into better-aligned chunks. On a real world example this measurably lowers the modeled chunk-loading cost, with fewer requests and less over-fetched CSS.

The shared-group count is read from edge weights (how many chunk groups two modules co-occur in). Since `make_acyclic` deletes edges to break cycles, it now returns the edges it cut so `linearize` can still count those conflicting-order co-occurrences — reconstructing the full co-occurrence without cloning the graph.

## Verification

- `cargo test -p turbopack-core style_groups_graph::tests` (all pass)
- Validated on a real world example that the resulting global order is deterministic and the modeled chunk-loading cost decreases.

### DIFF
--- a/turbopack/crates/turbopack-core/src/module_graph/style_groups_graph/algorithm.rs
+++ b/turbopack/crates/turbopack-core/src/module_graph/style_groups_graph/algorithm.rs
@@ -388,7 +388,14 @@ fn reconstruct_path(
 
 /// Mutate `graph` in place to remove all multi-node cycles by repeatedly cutting the
 /// lowest-weight edge of a short cycle in each SCC.
-pub(super) fn make_acyclic<N>(graph: &mut DiGraph<N, u32>) {
+///
+/// Returns the edges that were cut as `(from, to) -> weight`. [`linearize`] needs them: it derives
+/// its ordering heuristic from edge weights (chunk-group co-occurrence), and the cut edges are
+/// real co-occurrences that would otherwise be invisible in the now-acyclic graph.
+pub(super) fn make_acyclic<N>(
+    graph: &mut DiGraph<N, u32>,
+) -> FxHashMap<(NodeIndex, NodeIndex), u32> {
+    let mut cut_edges: FxHashMap<(NodeIndex, NodeIndex), u32> = FxHashMap::default();
     let mut queue: Vec<FxHashSet<NodeIndex>> = Vec::new();
     for scc in strongly_connected_components(&*graph) {
         if scc.len() > 1 {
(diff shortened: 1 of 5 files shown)


turbopack/crates/turbopack-core/src/module_graph/style_groups_graph/algorithm.rs
```diff
@@ -49,7 +61,35 @@ pub(super) fn create_graph(chunk_groups: &[Vec<usize>], node_count: usize) -> Di
...
-    graph
+    (graph, module_to_groups)
+}
+
+/// Count the chunk groups that both module `a` and module `b` belong to.
+///
+/// `module_to_groups` maps each module id to its sorted list of chunk-group indices (built by
+/// [`create_graph`]). The merge runs in O(min(|groups_a|, |groups_b|)).
+pub(super) fn shared_chunk_groups(
+    module_to_groups: &[Vec<usize>],
```
mischnic: nit/followup: Maybe we should introduce a newtype for this (vec is sorted, usize is a chunk group index)

turbopack/crates/turbopack-core/src/module_graph/style_groups_graph/algorithm.rs
```diff
@@ -450,56 +490,91 @@ pub(super) fn make_acyclic<N>(graph: &mut DiGraph<N, u32>) {
...
+pub(super) fn linearize<'a, G>(graph: G, module_to_groups: &[Vec<usize>]) -> Vec<NodeIndex>
 where
     G: ReadonlyGraph<'a>,
 {
-    let mut remaining_deps: FxHashMap<NodeIndex, usize> = FxHashMap::default();
+    // `remaining_deps` is an insertion-ordered map (matching `graph.nodes()`), so each node's
+    // position is a stable "seniority" index. Candidates carry that index as their tie-break key,
+    // so the `swap_remove` below — which scrambles positions within `candidates` — cannot disturb
+    // the "earliest remaining candidate" tie-break.
+    let mut remaining_deps: FxIndexMap<NodeIndex, usize> = FxIndexMap::default();
```
mischnic: this map should be presized with `graph.node().len()`?

mischnic (approved):

PR #95580, opened 2026-07-07
Add OpenTelemetry spans for API/middleware module loading and instrumentation register

lukesandberg:
## Summary

Adds OpenTelemetry spans for work that happens on the critical path before a request's handler runs but was previously untraced, leaving a gap in traces (especially visible on cold starts).

- **`NextNodeServer.runApi`** — wraps the pages API route module load (`load api route module`). On a cold module this reads and evaluates the bundled route chunk, which can be a meaningful share of request latency. Previously only the handler execution (`Node.runHandler`) was traced.
- **`NextNodeServer.runMiddleware`** — wraps middleware module resolution/load (`load middleware module`), i.e. compiling the edge bundle and warming the sandbox on first use. Previously only `Middleware.execute` (the middleware function itself) was traced.
- **`Instrumentation.register`** — a span covering the instrumentation `register()` hook, which runs once per server instance and must complete before the server serves requests. Because `register()` is typically where the user installs their OTEL provider, no live span can wrap it (the tracer is a no-op until it returns), so the span is **backdated** to cover the hook duration and emitted once the provider is in place.

All three span types are added to `NextVanillaSpanAllowlist` so they reach user OTEL exporters.

The `register()` span maybe be root trace or run part of an initial request. `expectTrace` in the OTel e2e suite now excludes it so it does not pollute request-trace assertions.

## Verification

- `pnpm test-start-turbo test/e2e/opentelemetry/instrumentation/opentelemetry.test.ts` (83 passed) — covers the new `runApi` / `runMiddleware` span assertions and the `register`-span filter across the default and direct-entrypoint variants
- `pnpm test-dev-turbo test/e2e/opentelemetry/instrumentation/opentelemetry.test.ts -t "middleware"` (dev-mode middleware path)
- Manually confirmed via the collector that the backdated `Instrumentation.register` span is exported with a non-zero duration. Not asserted in the suite: it fires once per server lifetime while a per-test collector is connected, so a stable assertion is not feasible in this harness.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -2627,7 +2627,6 @@ async function renderToHTMLOrFlightImpl(
               attributes: {
                 'next.clientComponentLoadCount':
                   metrics.clientComponentLoadCount,
-                'next.span_type': NextNodeServerSpan.clientComponentLoading,
               },
             })
             .end(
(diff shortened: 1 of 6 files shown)


lukesandberg:
* **#95580** <a href="https://app.graphite.com/github/pr/vercel/next.js/95580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95580?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

PR #95586, opened 2026-07-07
telemetry: add agentName to anonymous metadata

andrewimm:
Add agentName to anonymous telemetry metadata

### What

Adds an agentName: string | null field to the anonymous telemetry metadata block that accompanies every payload, alongside existing fields like ciName, nextVersion, and isCI. The value is the detected AI coding agent driving the process (e.g. claude, cursor, codex), or null when none is detected.

### Why

We currently have no visibility into how much Next.js development happens under AI coding agents. This field lets us understand that split in the same anonymous way we already track other CLI usage.

### How

Agent detection via @vercel/detect-agent
- Vendored @vercel/detect-agent (from the Vercel CLI) into `next/dist/compiled/@vercel/detect-agent` (new taskfile ncc task, externals entry, and type declaration), matching how I *understand* other third-party deps are bundled.
- The vendored detection method is async (it does a filesystem probe for Devin in addition to env-var checks) and potentially slower than we'd want for logging, so it's wrapped in a memoized helper: `telemetry/agent-name.ts`. `getAgentName()` caches the promise from `determineAgent()`: the first call runs detection once, concurrent callers share that in-flight promise, and every later call resolves against the already-settled result. The agent shouldn't change over the process lifetime, so this is safe. Detection failures resolve to null so telemetry never throws on this path.

Wiring into the payload
- `getAnonymousMeta()` is now async and includes `agentName: await getAgentName()`. `storage.ts` awaits it on the send path (the record method was thankfully already async).
- The NEXT_TELEMETRY_DEBUG=1 output now includes the meta block alongside each event, so the full payload — including agentName — is observable locally without sending anything. We didn't output this before, so it was harder to debug these fields.

Consolidating agent detection
- There was a second, synchronous, hand-vendored copy of agent detection (telemetry/detect-agent.ts) used only by the AGENTS.md/CLAUDE.md auto-generation hook. It's been removed and that hook (ensureAgentRulesForDev) now uses the same vendored `getAgentName()`. Now we have a single source of truth and can easily get upstream changes from the vercel CLI package. `ensureAgentRulesForDev` became async, its only caller in start-server.ts was already in an async scope, so this is easy.

### Testing

- NEXT_TELEMETRY_DEBUG=1 next build shows agentName in the payload meta block next to ciName/nextVersion.
- Verified the memoized helper: agent present -> name; cleared env -> null; repeated calls return the same promise instantly.
- agent-rules-auto-generate dev tests pass (agent-detected generates files; cleared-env does not)
- next-server-nft trace snapshot updated: adds @vercel/detect-agent + agent-name.js, removes detect-agent.js.
- ts, build, existing telemetry unit test, lint


andrewimm:
heads up @gaojude this replaces the manually copied agent detection logic with a vendored copy, like you suggested. I integrated that into your AGENT.md generation code, so any time we change the upstream package you can just update the vendor version.

packages/next/src/telemetry/agent-name.ts
```diff
@@ -0,0 +1,25 @@
...
+
+let agentNamePromise: Promise<string | null> | undefined
+
+/**
+ * Detects the AI coding agent (if any) driving the current process and resolves
+ * to its name, or `null` when no agent is detected.
+ *
+ * Detection is delegated to `@vercel/detect-agent`, which performs an async
+ * filesystem probe (e.g. for Devin) in addition to environment-variable checks,
+ * so it can be relatively expensive. The detected agent cannot change over the
```
bgw, line 11: It seems a little silly that `detectAgent` is async at all: https://github.com/vercel/vercel/blob/e12b1bd68c6e69ab1065dda89c9209ae1bdfb549/packages/detect-agent/src/index.ts#L146

It could just trust that if `DEVIN_LOCAL_PATH` is set, that we're in Devin, I don't really understand why it needs to also look for the file.

The fact that it's async makes this code a bit more annoying, and is also requiring the callsites to become async, which is all unfortunate.

packages/next/src/telemetry/storage.ts
```diff
@@ -318,7 +322,7 @@ export class Telemetry {
           projectId: await this.getProjectId(),
           sessionId: this.sessionId,
         },
-        meta: getAnonymousMeta(),
+        meta: await getAnonymousMeta(),
```
bgw, line 325: If `NEXT_TELEMETRY_DEBUG` is set, we're now computing this twice? I guess that's okay because `NEXT_TELEMETRY_DEBUG` isn't normally set, but it's gross.

packages/next/src/telemetry/detect-agent.ts
```diff
@@ -1,83 +0,0 @@
-// NOTE: This file duplicates the detection helper added in
-// https://github.com/vercel/next.js/pull/91854 (branch: imm/agent-telemetry).
-// When that PR lands, whichever branch merges second should collapse the
-// duplicate — the contents are intentionally identical so the merge is a
-// no-op. The env variable checks mirror the same functionality in the
-// `vercel` CLI.
```
bgw, line 6: Nice, thanks for cleaning up.

test/development/app-dir/agent-rules-auto-generate/agent-rules-auto-generate.test.ts
bgw, line 70: Can we use the `KNOWN_AGENTS` export from the detect-agent package? https://github.com/vercel/vercel/blob/e12b1bd68c6e69ab1065dda89c9209ae1bdfb549/packages/detect-agent/src/index.ts#L50

bgw (approved):

packages/next/src/telemetry/agent-name.ts
andrewimm: Nod. I'm just trusting the CLI team knows what they're doing here. It's frustrating that only one agent causes the whole thing to be async, but we're really lucky that all callsites were async / writing to the FS anyways.

packages/next/src/telemetry/storage.ts
andrewimm: I think we're the only people who run NEXT_TELEMETRY_DEBUG

gaojude (approved):

PR #95587, opened 2026-07-07
Report router transition end via a user-placed marker component

gaojude:
## Summary

Stacked on #95531.

Adds the third lifecycle event to the instrumentation-client router transition hooks: `unstable_onRouterTransitionEnd`, reporting when the destination page **declares itself loaded**. The declaration is a user-placed marker component:

```tsx
import { unstable_RouterTransitionEndMarker as RouterTransitionEndMarker } from 'next/navigation'

<Suspense fallback={<Skeleton />}>
  <Feed />
  <RouterTransitionEndMarker />
</Suspense>
```

When the first marker rendered for a navigation's destination is committed to the screen, the transition's `end` fires with that commit's timestamp. Together with the existing events, the latency story is complete:

| span | measures |
|---|---|
| `commit - start` | how long the navigation blocked before it was applied |
| `end - commit` | streaming / client-rendering cost paid after it was applied |

read against `cacheHit` (#95531), which attributes the first span to prefetching.

### Why a user-placed marker

The framework can only observe segment-granularity reveals; the user's own `<Suspense>` boundaries — where the content that actually matters streams in — are invisible to it. Inverting the responsibility makes React's reveal semantics do the measurement: a boundary only commits once its content is ready, so *marker mounted* = *the content the app declared meaningful is on screen*. It also gives `end` app-defined meaning (element-timing style) that no heuristic could infer, and composes: a marker in `error.tsx` reports an end for failed loads too.

Semantics:

- **First marker wins.** At most one `end` per transition, always after its `commit`, reported by whichever marker shows first. Later markers, StrictMode replays, and re-renders are no-ops (the transition's phase latch gains an `ended` state).
- **Same-commit degenerate case.** When the marker is part of the content the navigation itself commits (fully prefetched page, hash-only navigation, BFCache traversal), `end` reports in the same React commit as `commit`.
- **Not guaranteed.** No marker on the route, content that never streams in, or a newer navigation replacing the page all leave a transition with a `commit` but no `end` — the same "no terminal event" contract failed navigations already have.
- A marker that stays mounted across a navigation (shared layout) does not re-report — nothing new was shown. Markers belong in page-level content.

### How the marker attributes itself

The transition's identity lives on the router state (`AppRouterState.instrumentationTransition`, introduced in #95328): this PR adds the render-time consumer. `RouterTransitionEndContext` — provided by AppRouter, DCE'd when the flag is off — exposes the current state's transition to markers. Because markers read it at **render time**, a marker mounting in the very commit that applies the navigation attributes correctly even if its insertion effect runs before HistoryUpdater's; that ordering only affects *when* the event is emitted, which the `markerDidShow` handoff resolves (the commit emission reports `end` right after `commit`, with the same timestamp). A marker revealed by streamed content later reports immediately, against the page the user is actually on.

Note on naming: like React's own `unstable_`-prefixed components, the export must be aliased to a capitalized name at the import site — JSX treats lowercase-first tags as host elements (this is called out in the component's JSDoc; an unaliased `<unstable_RouterTransitionEndMarker />` silently renders a custom element).

## Verification

- `pnpm test-dev-turbo test/e2e/instrumentation-client-hook/instrumentation-client-hook.test.ts` — 29 passed.
- `pnpm test-start-turbo test/e2e/instrumentation-client-hook/instrumentation-client-hook.test.ts` — 28 passed (pre-restack; re-verified in dev after restacking onto the state-borne base).
- Manual smoke against a minimal app (`next dev`): `/end-marker` (marker in prefetched static content) reported `end` 0.6ms after `commit` in the same React commit; `/streaming` (marker inside a Suspense boundary with a 1s server delay) reported `end` ~997ms after `commit`, correlated by id, payload exactly `{id, timestamp}`.

### DIFF
--- a/packages/next/src/client/components/app-router-instance.ts
+++ b/packages/next/src/client/components/app-router-instance.ts
@@ -337,8 +337,14 @@ export function dispatchNavigateAction(
   // queued, so the hook runs outside React's render phase. (A user hook that
   // throws during the reducer would otherwise break error isolation between
   // hooks.) The transition object is threaded on the action so the reducer can
-  // attach the destination tree to it once it exists.
-  const instrumentationTransition = startRouterTransition(href, navigateType)
+  // attach the destination tree to it once it exists. The current state is
+  // passed in because router-transition must not import this module back
+  // (see startRouterTransition).
+  const instrumentationTransition = startRouterTransition(
+    href,
+    navigateType,
+    getCurrentAppRouterState()
+  )
 
   dispatchAppRouterAction({
     type: ACTION_NAVIGATE,
@@ -355,7 +361,11 @@ export function dispatchTraverseAction(
   href: string,
   historyState: AppHistoryState | undefined
 ) {
-  const instrumentationTransition = startRouterTransition(href, 'traverse')
+  const instrumentationTransition = startRouterTransition(
+    href,
+    'traverse',
+    getCurrentAppRouterState()
+  )
   dispatchAppRouterAction({
     type: ACTION_RESTORE,
     url: new URL(href),
--- a/packages/next/src/client/components/app-router-instance.ts
+++ b/packages/next/src/client/components/app-router-instance.ts
@@ -29,6 +29,7 @@ import { resetKnownRoutes } from './segment-cache/optimistic-routes'
 import { FreshnessPolicy } from './router-reducer/ppr-navigations'
 import { addBasePath } from '../add-base-path'
 import { isExternalURL } from './app-router-utils'
+import { createHrefFromUrl } from './router-reducer/create-href-from-url'
 import type {
   AppRouterInstance,
   NavigateOptions,
@@ -361,14 +362,20 @@ export function dispatchTraverseAction(
   href: string,
   historyState: AppHistoryState | undefined
 ) {
+  const url = new URL(href)
+  // A traversal has no caller-provided destination string (the popstate
+  // handler only has the absolute `location.href`), so report the canonical
+  // relative form — pathname + search + hash, the shape navigations
+  // conventionally receive — instead of leaking the absolute URL and making
+  // the hooks' `url` argument shape depend on the navigation type.
   const instrumentationTransition = startRouterTransition(
-    href,
+    createHrefFromUrl(url),
     'traverse',
     getCurrentAppRouterState()
   )
   dispatchAppRouterAction({
     type: ACTION_RESTORE,
-    url: new URL(href),
+    url,
     historyState,
     instrumentationTransition,
   })
--- a/packages/next/src/client/components/app-router.tsx
+++ b/packages/next/src/client/components/app-router.tsx
@@ -27,6 +27,7 @@ import {
 import { dispatchAppRouterAction, useActionQueue } from './use-action-queue'
 import { setLastCommittedTree } from './router-reducer/reducers/committed-state'
 import { commitRouterTransition } from './router-transition'
+import { RouterTransitionEndContext } from './router-transition-end-marker'
 import { AppRouterAnnouncer } from './app-router-announcer'
 import { RedirectBoundary } from './redirect-boundary'
 import { findHeadInCache } from './router-reducer/reducers/find-head-in-cache'
@@ -569,6 +570,22 @@ function Router({
     const { OfflineProvider } =
       require('./use-offline') as typeof import('./use-offline')
     content = <OfflineProvider>{content}</OfflineProvider>
+  }
+
+  // Positive flag check so the provider is removed by DCE when the
+  // experimental lifecycle is disabled (the marker then reads the context's
+  // default `null` and reports nothing). The provider binds each render to
+  // the transition of the state being rendered, which is what lets an
+  // `unstable_RouterTransitionEndMarker` attribute itself correctly even
+  // when it mounts in the same React commit that applies the navigation.
+  if (process.env.__NEXT_INSTRUMENTATION_CLIENT_ROUTER_TRANSITION_EVENTS) {
+    content = (
+      <RouterTransitionEndContext.Provider
+        value={state.instrumentationTransition}
+      >
+        {content}
+      </RouterTransitionEndContext.Provider>
+    )
   }
 
   return (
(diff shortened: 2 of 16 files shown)


gaojude:
Evaluated this stack against a real cacheComponents app — `vercel/front`'s templates gallery on an actual preview deployment, with the hooks wired into `instrumentation-client.ts` and markers in the page content (vercel/front#76850). Push navigations report the full lifecycle correctly (`start → commit ~3.1s (cacheHit: false) → end +8ms`, shared id, correct `from`/`to`). Two gaps surfaced, both fixed in 8d6cef5296:

| Finding | Cause | Fix |
|---|---|---|
| Traversals never fire `end` under cacheComponents | back/forward reveals the page preserved in a hidden `<Activity>` — nothing remounts, and the marker reported once per instance mount from an insertion effect (which doesn't reconnect on reveal) | marker reports from a dependency-free `useLayoutEffect` (disconnects/reconnects across Activity hide/reveal), transition read via an always-current ref; context swaps under a still-visible marker still don't re-report |
| Hooks' `url` arg shape depends on navigation type: pushes report the caller's href (`/templates/other/…`), traversals report absolute `window.location.href` | popstate handler only has `location.href` | traversals report `createHrefFromUrl(url)` — canonical relative form |

Real-world payloads that surfaced this: traverse commit arrived 23ms after start with `cacheHit: true` and no `end` across retries ([front e2e diagnostics](https://github.com/vercel/front/actions/runs/29253323381)).

Suite green in all four modes (dev/start × default/`__NEXT_CACHE_COMPONENTS=true`), including two new tests; the traverse-end test fails against the previous marker implementation under cacheComponents (verified by reverting the marker only).

Also nice in practice: `renderedPathname` (internal `/flg~-empty-/…` microfrontend rewrite) vs `canonicalUrl` (external URL) separation is exactly what telemetry wants.

**Update:** validated end-to-end on a fresh preview deployment of the fixed tarball (`16.3.0-preview-8d6cef52`): traverse now reports `start → commit (+11.9ms, cacheHit: true) → end (+5.7ms)` with one shared id and a relative `url` — [validation run](https://github.com/vercel/front/actions/runs/29277518074).

---

**Follow-up (continued dogfooding):** one more doc/behavior mismatch, fixed in 5424e5aa49.

The `RouterTransitionEndEvent` doc claimed hash-only navigations get `end` "in the same commit". They never have, under either marker implementation: a hash-only navigation commits the tree already on screen, the marker never leaves the screen, and both the old per-instance `didReport` latch and the new no-deps layout effect deliberately don't re-fire when nothing newly shows. Proven with a new fixture test (hash push on the marked page → `commit` with `to.canonicalUrl: "/end-marker#section"`, no `end`; fenced by a later marked navigation's `end`), passing in dev default + dev/start cacheComponents. I corrected the doc rather than the behavior — "nothing newly shown → no end" is consistent with the shared-layout-marker rule — but if you'd rather emit `end` at commit for same-tree navigations, that needs a mounted-marker registry; flagging as a design decision.

Same commit also documents the public `timestamp` timebase (`performance.timeOrigin + performance.now()`, epoch ms — previously only stated in the private impl).

Probe data for the same-tree design question (same-URL push on the marked page, all three modes): dev default/CC report `commit` with `cacheHit: false` — the fresh destination tree was fetched and rendered in place — and no `end` (re-render, not remount); prod reports `cacheHit: true` (content identical). Since a same-URL navigation's commit blocks on its fetch, `start → commit` already captures the load; the only unmeasurable tail is content that streams in after commit, which reconciles into existing boundaries without a marker remount.

---

**Follow-up 2: shallow routing strands `end` — fixed in cff418056a; plus a framework-computed-`end` design note.**

Docs-canonical shallow routing (`window.history.pushState(null, '', url)`) between a navigation's `commit` and its marker reveal silently drops the `end`. Empirical repro on the fixture (probe, 3s streamed page): `start → commit`, then nothing — while `replaceState(window.history.state, …)` survives only by accident (passing Next's state back hits the `data?.__NA` guard at `app-router.tsx:366`, skipping the router sync entirely). Mechanism: the patched history methods dispatch `ACTION_RESTORE` with `instrumentationTransition: null`; `restore-reducer.ts:105-113` stamps that null onto the produced state — unlike `refresh-reducer.ts:110-114`, which carries the transition forward — so `RouterTransitionEndContext` flips to null and the marker reveals reading nothing.

Fix: the shallow sync path now carries `getCurrentAppRouterState()?.instrumentationTransition` forward, mirroring the refresh semantics `settleRouterTransition` already documents for destination-preserving actions; the `pageshow`/BFCache restore stays null on purpose. Regression test: shallow pushState between commit and a ~3s streamed reveal → `end` fires with the navigation's id, no extra start/commit, no bogus abort. The test fails (0 ends) against the pre-fix build. Suite green: dev-default 33, start-cacheComponents 32.

Remaining hazard, not fixed here (queue-level): a shallow update landing while a navigation's action is still queued discards it (`app-router-instance.ts:222-237`), and that transition is later reported aborted with `replacedBy` pointing at an unrelated future commit.

**Design note.** A structured audit (traces + independent designs + judged synthesis) concluded the marker has three pathologies that are structural, not patchable: attribution rides React context/state, so anything deriving router state can drop it (the class behind this fix); it fires on *any* React commit that newly shows it — a client component mounting a marker on SWR resolve or a tab click closes the navigation span, fully decoupled from RSC (and `state.instrumentationTransition` is never cleared; only the phase latch guards a stale late `end`); and remount-keyed firing misses same-tree navigations (same-URL push probe earlier in this thread). The synthesized alternative: delete the marker; redefine `end` (same event shape) as framework-computed — navigation task drained AND every freshly-created segment's content committed (layout effect keyed by CacheNode identity, so shallow routing can't touch attribution and same-URL pushes are covered; fully-local navs report `end == commit`) — plus an `unstable_onRouterTransitionSettled` carrying a per-fresh-segment source breakdown (`settled − commit` = the network tail prefetching could remove; name collides with #94737 and needs reconciliation). Honest loss: app-defined "loaded" is gone — Suspense holes nested inside a segment's own payload are not awaited. Open questions before any of this ships: emit-vs-omit `end == commit`, hidden-`<Activity>`/inactive-parallel-slot reveal starvation, error.tsx terminals, retry coverage, flag-on overhead.

PR #95590, opened 2026-07-08
[experiment][ci] Use larger build machines for blocking build and build-native jobs

bgw:
This is counter to our goals of saving money in CI, but these two jobs tend to be the slowest jobs that run before we're able to start running tests, so any time these jobs take delays the rest of our CI, and we don't actually spend much money on these two jobs compared to all of the test shards.

So logically we should run these with the fastest machine types possible so that we can start running the tests that depend on them sooner.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -117,6 +117,7 @@ jobs:
       skipInstallBuild: 'yes'
       stepName: 'build-native'
       uploadNativeArtifact: true
+      runs_on_labels: '["ubuntu-latest-32-core-arm-oss"]'
     secrets: inherit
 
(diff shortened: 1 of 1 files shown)


bgw:
Apparently `ubuntu-latest-32-core-arm-oss` refers to a runner image that's a vanilla "partner" ubuntu image, and not a proper github runner image, so it's missing tons of dependencies. I'll need to figure out how to fix this before we can shift these to larger runners. I think it might be possible to define runners via terraform (https://registry.terraform.io/providers/integrations/github/latest/docs/resources/actions_hosted_runner), and we should probably move away from the "latest" label and towards more explicit (24.04, 26.04) versioned labels anyways.

PR #95591, opened 2026-07-08
Add `prefetch: 'app-shell'` option to the instant() navigation testing helper

gaojude:
By default, instant() simulates a fully warm cache: navigations render
everything the link would have prefetched, including per-link and
runtime-prefetched data. This adds a way to assert on just the route's
shared App Shell:

  await instant(page, fn, { prefetch: 'app-shell' })

The option is carried on the instant cookie's pending value and preserved
on captured values, so it survives MPA page loads within a scope. While
active:

- the locked-navigation prefetch stops after the App Shell phases, so no
  per-link concrete-param or runtime-prefetch data is fetched, and
- navigation reads are restricted to shell entries regardless of the
  link's own fetch strategy (e.g. prefetch={true}).

Cookie parsers now discriminate pending vs. captured values by the first
element instead of array length, since pending values can carry options.

### DIFF
--- a/packages/next-playwright/README.md
+++ b/packages/next-playwright/README.md
@@ -61,6 +61,36 @@ test('navigates to profile instantly', async ({ page }) => {
 })
 ```
 
+**Asserting on the App Shell only** (`prefetch: 'app-shell'`):
+
+By default, `instant()` assumes everything the link would have prefetched is
+cached — including per-link data and runtime-prefetched data on routes that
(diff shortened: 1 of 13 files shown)


acdlite:
I'm not totally opposed to adding an option like this but your description of the current behavior isn't correct: if you have Partial Prefetching on, and the link doesn't have a `prefetch` prop, then it only allows the shell to render. If you _do_ have a prefetch prop, then it allows the speculative content to render.

PR #95602, opened 2026-07-08
[fragment-scroll] Remove `config.experimental.appNewScrollHandler`

eps1lon:
The flag has been enabled by default for 16.3.

It could've been removed before 16.3 but I just missed the window.


scripts/run-jest.sh
icyJoseph: Clanker calling out that this is never read, only assigned to in line 44

icyJoseph (approved):

dmitrc:
There seem to be open issues concerning the new default, eg #96855. 

We repro the variation of this ourselves where the scroll is not correctly reset on the soft navigation, even though we do not use the parallel routes (seems the root cause has to do with removed `sticky`/`fixed` checks). 

Currently the only mitigation is to restore the `experimental.appNewScrollHandler = false` as was the default in 16.2.x. With the flag going away, seems like this will become proper broken without any recourse, @eps1lon / @icyJoseph / @marcoshernanz .

PR #95610, opened 2026-07-08
Static generation: batched concurrency → rolling concurrency

gaojude:
Adds rolling concurrency for static generation page export, gated behind a new experimental flag. Default behavior is unchanged (batched).

### DIFF
--- a/packages/next/src/lib/run-with-concurrency.ts
+++ b/packages/next/src/lib/run-with-concurrency.ts
@@ -1,0 +1,55 @@
+/**
+ * Maps over `items` while limiting the number of concurrently running tasks.
+ * Results are returned in the same order as the input items.
+ */
+export async function runWithConcurrency<T, R>(
+  items: readonly T[],
+  maxConcurrency: number,
+  fn: (item: T) => Promise<R>
+): Promise<R[]> {
+  if (!Number.isInteger(maxConcurrency) || maxConcurrency < 1) {
+    throw new RangeError('maxConcurrency must be a positive integer')
+  }
+
+  if (items.length === 0) {
+    return []
+  }
+
+  // Fill the array up front so that out-of-order completions do not make it
+  // sparse. Every entry is replaced before this function returns.
+  const results: Array<R | undefined> = []
+  for (let i = 0; i < items.length; i++) {
+    results.push(undefined)
+  }
+
+  let nextIndex = 0
+  let stopped = false
+
+  async function runWorker(): Promise<void> {
+    while (!stopped) {
+      const index = nextIndex
+      if (index >= items.length) {
+        return
+      }
+      nextIndex = index + 1
+
+      try {
+        results[index] = await fn(items[index])
+      } catch (error) {
+        // Other workers can finish their current task, but must not start new
+        // work after an unexpected failure.
+        stopped = true
+        throw error
+      }
+    }
+  }
+
+  const workerCount = Math.min(maxConcurrency, items.length)
+  const workers: Promise<void>[] = []
+  for (let i = 0; i < workerCount; i++) {
+    workers.push(runWorker())
+  }
+
+  await Promise.all(workers)
+  return results as R[]
+}
--- a/packages/next/src/lib/run-with-concurrency.ts
+++ b/packages/next/src/lib/run-with-concurrency.ts
@@ -34,12 +34,11 @@ export async function runWithConcurrency<T, R>(
     return []
   }
 
-  // Fill the array up front so that out-of-order completions do not make it
-  // sparse. Every entry is replaced before this function returns.
-  const results: Array<R | undefined> = []
-  for (let i = 0; i < items.length; i++) {
-    results.push(undefined)
-  }
+  // Workers finish out of order and write to `results[index]` directly, so
+  // create the array at its final length instead of pushing. On the success
+  // path every slot is overwritten with a real result, which is what makes
+  // the `as R[]` cast at the end safe.
+  const results: Array<R | undefined> = new Array(items.length).fill(undefined)
 
   let nextIndex = 0
   let stopped = false
--- a/packages/next/src/lib/run-with-concurrency.ts
+++ b/packages/next/src/lib/run-with-concurrency.ts
@@ -1,6 +1,10 @@
 /**
  * Maps over `items` while limiting the number of concurrently running tasks.
  * Results are returned in the same order as the input items.
+ *
+ * After the first rejection is observed, no new tasks are started. Tasks that
+ * are already running are not cancelled, and the returned promise rejects
+ * without waiting for them to finish.
  */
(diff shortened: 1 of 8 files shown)


gaojude:
CI triage update: this is a global TypeScript 7.0.2 break, not a scheduler regression. The last canary run before TS 7 became npm `latest` was green; the first canary run afterward failed 41 jobs, current canary has the same pattern, and a no-op control PR failed 40 jobs. This PR failed 37 jobs with the same missing compiler/extensionless `tsc` errors.

The checks closest to this change are green: the concurrency tests passed on Linux and Windows with Node 20/22, the updated NFT test passed three production runs, and build, lint, types, native builds, and unit suites passed. No persistent scheduler-related failure was found.

I’m holding off on rerunning or rebasing because `test/lib/next-modes/base.ts` still installs `typescript: latest`, so either action would reproduce the outage. I’ll rebase and rerun after the global TypeScript pin/compatibility fix lands on canary.

packages/next/src/lib/run-with-concurrency.ts
```diff
@@ -0,0 +1,59 @@
...
+      }
+      nextIndex = index + 1
+
+      try {
+        results[index] = await fn(items[index])
+      } catch (error) {
+        // Other workers can finish their current task, but must not start new
+        // work after an unexpected failure.
+        stopped = true
+        throw error
```
lubieowoce: the comment says "Other workers can finish their current task", but isn't this `throw` gonna error out of the `Promise.all()` below and make us exit from the function with some workers still running (i.e. haven't finished)? maybe we should wait for `Promise.allSettled()` instead or something?

gaojude: Good catch — the throw did reject `Promise.all()` while workers were still running. Fixed in f8e194f061: workers no longer rethrow; the first error is recorded, all in-flight tasks settle, then the promise rejects with that error. Added a test asserting the promise stays pending until the in-flight item finishes.

gaojude:
## A/B results: SSG phase, vercel-site preview builds

All 6 builds: identical PR tarball (`16.3.0-preview-c3675b2c`), 30-core/60GB build machines, 29 export workers, `staticGenerationMaxConcurrency: 15`, `vercel-site:build` = `cache bypass, force executing` (no turbo replay). Arm identity verified per build via the experiments preamble (`✓ staticGenerationRollingConcurrency` present ⇔ flag on). Metric = Next.js's own `✓ Generating static pages … in Ns` line (SSG phase only; compile/install excluded).

| Sample | Flag          | Pages | SSG time | Deployment                        |
| ------ | ------------- | ----- | -------- | --------------------------------- |
| A1     | off (batched) | 987   | 33.3s    | vercel-site-lu4ohko1b.vercel.sh   |
| B1     | on (rolling)  | 987   | 26.8s    | vercel-site-ezmusogxb.vercel.sh   |
| A2     | off (batched) | 987   | 29.1s    | vercel-site-cdbwnnqyq.vercel.sh   |
| B2     | on (rolling)  | 989   | 28.8s    | vercel-site-8iigq6iwv.vercel.sh   |
| A3     | off (batched) | 989   | 28.1s    | vercel-site-iq3qmas5o.vercel.sh   |
| B3     | on (rolling)  | 989   | 27.7s    | vercel-site-7zed1hbc2.vercel.sh   |

- Median: batched 29.1s vs rolling 27.7s (−4.8%). Mean: 30.2s vs 27.8s (−8.0%).
- Adjacent-pair deltas (drift-controlled): A1→B1 −6.5s (−19.5%), A3→B3 −0.4s (−1.4%). A2/B2 were hours apart (content drifted 987→989 pages) and aren't a clean pair.
- Rolling never lost a comparison; its worst sample (28.8s) beats the batched median. Batched spread 5.2s vs rolling 2.0s — lower variance is the expected signature of removing batch-straggler sensitivity: rolling wins big when a batch has slow-tail pages (pair 1), and converges to parity when pages are uniform (pair 3).
- Caveats: n=3 per arm on shared infra with a live CMS backend; batched arm drifted faster over the session (33.3→28.1), so the pairwise view is the defensible read.
- Playwright preview failures on this branch are uncorrelated with the flag (flag-off A2 failed the same `ai-chat-botid`/`instant-navigation` specs; flag-off A1 passed E2E).

### Quartile decomposition (from timestamped progress lines)

| Build | Flag    | Pages | Q1 (0–25%) | Q2 | Q3 | Q4 (75–100%) | Total | Q4 share |
| ----- | ------- | ----- | ---------- | ---- | ---- | ------------ | ----- | -------- |
| A1    | batched | 987   | 3.3s       | 2.9s | 4.9s | **22.2s**    | 33.3s | **67%**  |
| B1    | rolling | 987   | 4.8s       | 3.9s | 6.1s | 12.1s        | 26.8s | 45%      |
| A2    | batched | 987   | 4.7s       | 4.6s | 7.1s | 12.7s        | 29.1s | 44%      |
| B2    | rolling | 989   | 4.4s       | 4.4s | 6.3s | 13.7s        | 28.8s | 48%      |
| A3    | batched | 989   | 4.6s       | 4.7s | 7.5s | 11.3s        | 28.1s | 40%      |
| B3    | rolling | 989   | 4.6s       | 4.1s | 6.9s | 12.1s        | 27.7s | 44%      |

Q1–Q3 are scheduler-independent (slots never idle while the queue is deep). The entire effect is in the tail: under batching, a slow page freezes up to 14 of the 15 slots in its worker's wave, so cold-upstream builds blow up (A1's Q4 = 2× every other build's). Rolling backfills idle slots and caps the tail — worst case −13.5%, within-arm spread halved, zero retries in all six logs.

### Update: forced-cold pair (build cache disabled via `VERCEL_FORCE_NO_BUILD_CACHE`)

| Build | Flag    | Env          | Collect | SSG   | SSG Q4 |
| ----- | ------- | ------------ | ------- | ----- | ------ |
| A4    | batched | cold (forced)| 19.5s   | 23.7s | 12.8s  |
| B4    | rolling | cold (forced)| 19.3s   | 24.3s | 13.9s  |

This **falsifies the build-cache hypothesis** for A1's 22.2s tail: forced-cold builds reproduce A1's slow compile (87s) and collect (19.5s) but the SSG tail stays normal (~13s) — so A1's blowup was a transient upstream-latency event that batching amplified, not the cache. Cold pair itself is parity (±0.6s), consistent with the mechanism: the scheduler only matters when a straggler event occurs.

Final read across all 8 builds: steady-state batched ≈ rolling (warm and cold); the one straggler event observed (1 in 8 builds) ran under batching and doubled the SSG tail; rolling structurally caps that mode (deterministic fixture: next batch item starts +18ms after a slot frees vs +1521ms waiting for the batch). Rolling never regressed in any comparison. Verdict: neutral in the median, insurance against the slow-upstream tail mode.

PR #95611, opened 2026-07-08
Fork navigation-testing-lock module

eps1lon:
Ensures the module graph is not bundled. Right now this has no effect since all the modules imported by `navigation-test-lock` have legit uses in prod builds.

Also nice to not have to fork every single implementation inline.


mischnic (approved):
Are we asserting anywhere that this exclusion is working properly?

PR #95614, opened 2026-07-08
Reduce the size of OperationVc from 8 bytes to 4

lukesandberg:
Reduce the size, i don't really expect much of a win, but this is low hanging fruit and removes a few branches as well.

Consolidate the implementation of CollectibleSource, no need for OperationVc to delegate to RawVc here.

### DIFF
--- a/turbopack/crates/turbo-tasks/src/vc/operation.rs
+++ b/turbopack/crates/turbo-tasks/src/vc/operation.rs
@@ -2,6 +2,7 @@ use std::{
     fmt::Debug,
     future::Future,
     hash::Hash,
+    marker::PhantomData,
     pin::Pin,
     task::{Context, Poll},
 };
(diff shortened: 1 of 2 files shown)


lukesandberg:
* **#95614** <a href="https://app.graphite.com/github/pr/vercel/next.js/95614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/vc/operation.rs
bgw: Doesn't need `repr(transparent)` anymore

bgw (approved):
This makes sense. `ResolvedVc` is forced to have `repr(transparent)` and contain a `Vc` because we need it to be able to transmute to/from `Vc` and because `ResolvedVc` needs to be able to implement `Deref<Vc>`, which wouldn't be possible if it didn't contain the exact memory representation of `Vc`.

But `OperationVc` is always converted via `.connect()`, so there aren't the same restrictions.

PR #95616, opened 2026-07-08
Only compile edge instrumentation when an edge consumer exists (next dev)

### LINKED ISSUE #86479: `next dev`: Instrumentation hook is still analyzed for the edge runtime despite never running in the edge runtime

https://github.com/blairmcalpine/instrumentation-errors-repro

### To Reproduce

1. npm install
2. npm run dev
3. Visit http://localhost:3000
4. See the error in your console:
```
Ecmascript file had an error
  1 | export const register = async () => {
  2 |   console.log("registering instrumentation", process.env.NEXT_RUNTIME);
> 3 |   process.on("unhandledRejection", (reason, promise) => {
    |   ^^^^^^^^^^
  4 |     console.error("Unhandled rejection:", reason);
  5 |     console.error("Promise:", promise);
  6 |   });`
```

Current: There is an error logged from analysis of the instrumentation hook for edge runtimes, despite the instrumentation hook never running in the edge runtime. This error doesn't actually break anything at runtime it seems.

Expected: No error should appear. With the new proxy.ts enabling a full nodejs runtime version of Next.js, the instrumentation hook never runs its "edge" version (verified via the console log in the minimal repro, the register hook only runs with process.env.NEXT_RUNTIME set to "nodejs"). Despite this, it seems like there is still some checking for nodejs-only APIs running for the instrumentation hook, which produce console warnings.

If this helps debug:
1. notably the warnings only appear after you visit a page, and not when the initial instrumentation hook is run. 
2. It doesn't seem to produce the same warnings when using `fs`, but accessing the `process` global seems to cause problems.
3. This only happens in dev mode, not production.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 22.15.1
  npm: 10.9.2
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 16.0.3 // Latest available version is detected (16.0.3).
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Instrumentation, Middleware

next dev (local)

### Additional context

_No response_

webdevelopersrinu:
Fixes #86479

### What?

In Turbopack dev mode, the edge variant of the instrumentation hook is no longer compiled unconditionally. It is compiled lazily, when the first edge-runtime consumer is discovered: edge middleware (checked when the middleware endpoint is written) or an edge route (checked in `handleRouteType`). Once an edge consumer has been seen, later `handleEntrypoints` passes (e.g. after the instrumentation file changes) rebuild it eagerly again.

### Why?

`handleEntrypoints` compiled both `instrumentation.nodeJs` and `instrumentation.edge` whenever an `instrumentation` file existed — even when the app had no edge-runtime consumer at all (Node.js `proxy.ts`, no `runtime = 'edge'` routes). The edge variant's output is only ever loaded by edge functions (its files are merged into edge function definitions in the middleware manifest), so the eager compilation only surfaced irrelevant edge-compat issues like:

```
A Node.js API is used (process.on at line: 4) which is not supported in the Edge Runtime.
Ecmascript file had an error
```

Since 16.2.x these surface as errors that block local dev (see https://github.com/vercel/next.js/issues/86479#issuecomment).

### How?

- Added `ensureEdgeInstrumentationBuilt()` which compiles the edge variant, reports its issues, and loads the instrumentation middleware manifest. It's invoked from the middleware path (when `writtenEndpoint.type === 'edge'`) and the four edge-route branches of `handleRouteType`, before manifests are written, so instrumentation files are always merged into edge function definitions before they execute.
- A `WeakSet<Entrypoints>` remembers that an edge consumer was seen so instrumentation changes keep rebuilding the edge variant.
- Added `test/development/instrumentation-hook-node-apis` verifying no edge-compat issues are reported for a Node-API-using `instrumentation.js` when no edge consumer exists. Existing `test/e2e/instrumentation-hook` edge cases (`with-middleware`, `with-edge-api`, `with-edge-page`, `with-async-edge-page`) verify the lazy path still delivers edge instrumentation, and pass.

Note: production builds (`next build`) compile the edge variant unconditionally on the Rust side (`endpoint_groups` in `crates/next-api/src/project.rs`) — that path is out of scope here and could be a follow-up.

### DIFF
--- a/packages/next/src/server/dev/turbopack-utils.ts
+++ b/packages/next/src/server/dev/turbopack-utils.ts
@@ -143,6 +143,58 @@ type HandleRouteTypeHooks = {
   subscribeToChanges: StartChangeSubscription
 }
 
+/**
+ * Tracks whether an edge-runtime consumer (edge middleware or a route using
+ * `runtime = 'edge'`) has been seen for the given entrypoints.
+ *
+ * The edge variant of the instrumentation hook is only ever loaded by edge
+ * functions (its files are merged into edge function definitions in the
+ * middleware manifest), so it's only compiled once an edge consumer is
+ * discovered. Otherwise apps without any edge-runtime code would get
+ * irrelevant edge-compat issues reported for `instrumentation.ts` (e.g. when
+ * it uses Node.js APIs). See https://github.com/vercel/next.js/issues/86479
+ */
+const seenEdgeRuntimeConsumer = new WeakSet<Entrypoints>()
+
+export async function ensureEdgeInstrumentationBuilt({
+  entrypoints,
+  currentEntryIssues,
+  manifestLoader,
+  logErrors,
+  handleWrittenEndpoint,
+  startBuilding,
+}: {
+  entrypoints: Entrypoints
+  currentEntryIssues: EntryIssuesMap
+  manifestLoader: TurbopackManifestLoader
+  logErrors: boolean
+  handleWrittenEndpoint?: HandleWrittenEndpoint
+  startBuilding?: StartBuilding
+}): Promise<void> {
+  seenEdgeRuntimeConsumer.add(entrypoints)
+
+  const instrumentation = entrypoints.global.instrumentation
+  if (!instrumentation) {
+    return
+  }
+
+  const finishBuilding = startBuilding?.(
+    'instrumentation Edge',
+    undefined,
+    true
+  )
+  const key = getEntryKey('root', 'server', 'instrumentation.edge')
+
+  const writtenEndpoint = await instrumentation.edge.writeToDisk()
+  handleWrittenEndpoint?.(key, writtenEndpoint, false)
+  processIssues(currentEntryIssues, key, writtenEndpoint, false, logErrors)
+  await manifestLoader.loadMiddlewareManifest(
+    'instrumentation',
+    'instrumentation'
+  )
+  finishBuilding?.()
+}
+
 export async function handleRouteType({
   dev,
   page,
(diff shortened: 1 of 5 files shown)


apostolos:
Does that mean that we'll no longer have to guard with `process.env.NEXT_RUNTIME === 'nodejs'` anymore? Will it be safe to switch to static imports instead of `await import()`ing Node.js specific packages?

webdevelopersrinu:
mostly no, the guard is still the right pattern in the general case. What this PR changes is narrower:

- **If your app has no edge-runtime consumer at all** (Node.js middleware/`proxy.ts`, no `runtime = 'edge'` routes): with this PR the edge variant of `instrumentation.ts` is never compiled in `next dev`, so Node.js APIs and static imports of Node-specific packages won't surface edge-compat errors there.
- **If your app has any edge consumer** (edge middleware or an edge route): the edge variant is still compiled and `register()` still runs inside the edge runtime for those functions — so the `NEXT_RUNTIME` guard and dynamic `await import()` remain necessary.
- **Production builds are unchanged**: `next build` still compiles the edge variant unconditionally (that happens on the Rust side and is out of scope here, noted in the PR description as a possible follow-up). So static Node-only imports could still cause warnings/errors at build time even with zero edge consumers.

So: this PR removes the *irrelevant* dev errors for apps that never run edge code, but it doesn't change the contract of `register()` running per-runtime when edge consumers exist.

PR #95620, opened 2026-07-08
Add React sync development skill

timneutkens:
### What?

Add a repository skill for building local React stable and experimental bundles, syncing them into Next.js, and verifying the integration.

### Why?

Testing upstream React changes in Next.js requires a precise bundle list, both React release channels, and the repository-pinned pnpm version. Capturing the workflow as a skill makes it repeatable and avoids global pnpm versions bypassing workspace linking.

### How?

Add `$react-sync` with a reusable build helper and document it in `AGENTS.md`. The sync command puts the Corepack shim first on `PATH` so nested pnpm commands use the version pinned by Next.js.

### Verification

- `bash -n .agents/skills/react-sync/scripts/build-for-next.sh`
- `python3 ~/.codex/skills/.system/skill-creator/scripts/quick_validate.py .agents/skills/react-sync`
- Prettier check for `AGENTS.md` and `SKILL.md`
- Full local React build and `pnpm sync-react` workflow completed successfully
- `pnpm --filter=next build`


.agents/skills/react-sync/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+   checkouts. Rebuild Next.js when required, then run the focused test command
+   matching the changed behavior.
+
+## Build variants
+
+The script builds stable first and moves its `node_modules` to
+`build/oss-stable`. It then performs an experimental partial build and moves
+its `node_modules` to `build/oss-experimental`. Do not replace it with a
+generic React build: Next.js needs both release channels and the listed server,
+client, compiler, scheduler, webpack, and Turbopack bundles.
```
eps1lon: Does the skill need that information? `build-for-next.sh` and `pnpm sync-react` take care of these details.

timneutkens: :robot: Done — removed the redundant Build variants section. The skill now leaves those implementation details to build-for-next.sh and pnpm sync-react.

.agents/skills/react-sync/SKILL.md
```diff
@@ -0,0 +1,50 @@
...
+   cd <next-repo>
+   PATH="$(dirname "$(command -v corepack)"):$PATH" \
+     pnpm sync-react --version <react-repo>
+   ```
+
+   Keep the Corepack shim first on `PATH`. The repository pins its pnpm version
+   in `package.json`, and `sync-react` starts a nested `pnpm install`. Without
+   this override, a bundled or global pnpm can be selected for either command,
+   ignore the repository's pnpm configuration, and attempt to download local
+   workspace packages such as `@next/font` from npm.
```
eps1lon: This sounds like a halucination. It picks the right version just fine locally. Otherwise any `pnpm sync-react` we call would be wrong.

timneutkens: This did happen when testing the skill unfortunately. The shell it booted somehow did not run with corepack causing the wrong pnpm to be used.

I'll do another run to make sure this is correct.

eps1lon: But that's a bug by the agent in the way it spawns the shell. Not an issue with `pnpm sync-react`. If that's a general issue, it should be an actual skill "here's how you invoke pnpm".

timneutkens: I've created a reproduction and reported it to OpenAI. Updated the skill to account for this, but let's not block merging on that bug being fixed.

eps1lon: This doesn't spell out it's a agent bug. If we check this in without it, it gets picked up by other agents as well and proliferates into other skills.

timneutkens: Updated to say it's a bug!

eps1lon (approved):

PR #95621, opened 2026-07-08
Add attribute rendering benchmark

timneutkens:
### What?

Add two dynamic routes to the render-pipeline benchmark and include both in the default route suite:

- `/attributes` renders 1,000 rows with string, data, ARIA, title, and inline-style attributes.
- `/tailwind` renders a realistic dashboard with dense Tailwind-style utility class names across navigation, metrics, project cards, and an activity table.

### Why?

The existing benchmark routes focus on lightweight rendering or Suspense and Flight payload stress. They do not provide focused integration coverage for React Fizz attribute serialization.

`/attributes` provides a synthetic stress case modeled after the upstream React benchmark for facebook/react#36899. `/tailwind` complements it with a production-like component tree and realistic class strings, including responsive, state, dark-mode, and arbitrary-value utilities.

### How?

Both fixtures run through the real Next.js production-server render-pipeline harness and are forced dynamic so every request exercises server rendering.

The Tailwind-style fixture contains 18 project cards, 36 activity rows, and 101 `className` attributes. It intentionally does not install or configure Tailwind because the benchmark targets React serialization of the class strings, not generated CSS or visual styling.

The new routes are registered in the default stress suite and documented in the benchmark playbook and render-pipeline README.

### Verification

- Prettier for all changed files
- ESLint for both benchmark fixtures and `bench/render-pipeline/benchmark.ts`
- Production `next build` + `next start` smoke benchmark for `/tailwind`; single-client and under-load phases completed with zero request errors
- Three interleaved base/head `/attributes` runs with 500 serial and 5,000 loaded requests
- `/attributes` PR-head median throughput: +8.26% single-client and +10.85% under load
- `/attributes` PR-head median p95: 7.91% better single-client and 9.61% better under load
- `git diff --check`

### DIFF
--- a/bench/BENCHMARKING.md
+++ b/bench/BENCHMARKING.md
@@ -77,6 +77,7 @@ pnpm bench:render-pipeline \
 Default stress routes currently include:
 
 - `/`
+- `/attributes`
 - `/streaming/light`
 - `/streaming/medium`
 - `/streaming/heavy`
--- a/bench/BENCHMARKING.md
+++ b/bench/BENCHMARKING.md
@@ -78,6 +78,7 @@ Default stress routes currently include:
 
 - `/`
 - `/attributes`
+- `/tailwind`
 - `/streaming/light`
 - `/streaming/medium`
 - `/streaming/heavy`
--- a/bench/basic-app/app/attributes/page.js
+++ b/bench/basic-app/app/attributes/page.js
@@ -1,0 +1,34 @@
+import React from 'react'
+
+export const dynamic = 'force-dynamic'
+
+const ROWS = 1000
+
+function Row({ index }) {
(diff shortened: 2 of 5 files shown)


gaearon:
Does it make sense to make it a bit more varied or do you want to solve this through more routes? E.g. I doubt a ton of Next apps today use inline styles for everything, whereas it would be great to test shapes like "every node has a `className`" (Tailwind style) with some other attributes here and there but not everywhere.

timneutkens:
Plan is to add more real benchmark applications separate from this one page 👍

gaearon (approved):

awingfield916-tech:
I think some of that stuff will work!

PR #95628, opened 2026-07-09
[ci] Enforce minimumReleaseAge in e2e tests that install external dependencies

bgw:
We also mitigate the security impact of e2e tests by running everything in VMs, and by not reading caches in release jobs, but we should still be using `minimumReleaseAge` in e2e tests.

Ideally we'd also use lockfiles here too, and I'll work on that next, but that is likely a lot more complicated because we probably need some way to merge lockfiles, since we can't pin the next.js and react versions in every test, and we need a reasonable/efficient way to bulk-update all the lockfiles for tests that need it.

Depends on https://github.com/vercel/next.js/pull/95668 for a bunch of windows path representation fixes that this ends up exposing in CI, because adding workspace files causes pnpm to change the absolute path representation used in the Windows junction points that it creates.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -29,8 +29,11 @@ ignoredBuiltDependencies:
   - ssh2
   - unrs-resolver
   - wasm-pack
-# Keep security related settings in sync with pnpm-workspace.yaml written to
-# temporary fixture dirs in file://./scripts/install-native.mjs
+
+# - Keep security related settings in sync with pnpm-workspace.yaml written to
+#   temporary fixture dirs in `file://./scripts/install-native.mjs`
+# - `file://./test/lib/pnpm-security-settings.js` reads and copies these values for
+#   isolated test dirs
 blockExoticSubdeps: true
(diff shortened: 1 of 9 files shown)


pnpm-workspace.yaml
```diff
@@ -29,8 +29,11 @@ ignoredBuiltDependencies:
   - ssh2
   - unrs-resolver
   - wasm-pack
-# Keep security related settings in sync with pnpm-workspace.yaml written to
-# temporary fixture dirs in file://./scripts/install-native.mjs
+
+# - Keep security related settings in sync with pnpm-workspace.yaml written to
+#   temporary fixture dirs in `scripts/install-native.mjs`
+# - `test/lib/pnpm-security-settings.js` reads and copies these values for
```
eps1lon: file URLs pls like in the previous comment for ctrl+click in vscode

bgw: Okay, I removed the file URLs because I thought it was very strange.

eps1lon (approved):

PR #95631, opened 2026-07-09
fix(cms-contentful): await draftMode() and use Promise<> params type

manoraj:
### What?
Fix the `cms-contentful` example to use the async APIs introduced in Next.js 15:
  - Await `draftMode()` in `app/page.tsx`, `app/posts/[slug]/page.tsx`, and
  `app/api/disable-draft/route.ts`
  - Update `params` type to `Promise<{ slug: string }>` and await it in
  `app/posts/[slug]/page.tsx`

### Why?
In Next.js 15, `draftMode()` from `next/headers` became async and must be awaited.
  Page `params` are also now typed as a `Promise`. The example was using the old
  synchronous patterns, causing the build to fail and the app to not work when running against the latest
  version of Next.js -- this example project uses latest version of next.js in the package.json.

### How?
Awaited `draftMode()` at each call site and updated the `params` type signature to
  `Promise<{ slug: string }>` with a corresponding `await`.

Closes NEXT-
Fixes #


examples/cms-contentful/app/posts/[slug]/page.tsx
```diff
@@ -20,10 +20,11 @@ export async function generateStaticParams() {
 export default async function PostPage({
   params,
 }: {
-  params: { slug: string };
+  params: Promise<{ slug: string }>;
```
Comment (icyJoseph) [inline examples/cms-contentful/app/posts/[slug]/page.tsx]: We can replace this altogether with:

```tsx
export default async function PostPage({ params }: PageProps<'/posts/[slug]'>) 
```

Comment (manoraj) [reply examples/cms-contentful/app/posts/[slug]/page.tsx]: I guess PageProps is only available after Next.js generates route types? Using the same inline Props pattern as the other CMS examples

Comment (icyJoseph) [reply examples/cms-contentful/app/posts/[slug]/page.tsx]: Yes, they are auto generated on `next typegen`, `next dev` and `next build`. If you don't mind we should update the example to use them (LayoutProps too)

Comment (manoraj) [reply examples/cms-contentful/app/posts/[slug]/page.tsx]: @icyJoseph Done. Verified the types with the changes within the standalone app after `next typegen`

icyJoseph (approved):

PR #95639, opened 2026-07-09
(TypeScript 7 Support) Add experimental TypeScript CLI backend

timneutkens:
## Summary

- add `experimental.useTypeScriptCli` so projects can explicitly run their local `tsc` during `next build`, including TypeScript 7 while the legacy JavaScript API is unavailable
- preserve the TypeScript API backend as the default, with TypeScript 6-compatible dependency installation and actionable TypeScript 7 migration guidance
- load effective compiler metadata through `tsc --showConfig`, preserve inherited path origins, stream native diagnostics, redirect incremental state, and forward worker termination to the compiler process tree
- document the experimental behavior and cover TypeScript 6/7, both bundlers, raw diagnostics, full-project checking, dependency selection, and cleanup

## Demo

```
pnpm build

> cna@0.1.0 build /Users/timneutkens/projects/sandbox/cna
> next build

▲ Next.js 16.3.0-canary.81 (Turbopack)
✓ Running next.config.ts took 23ms
- Cache Components enabled
- Experiments (use with caution):
  ✓ useTypeScriptCli

  Creating an optimized production build ...
✓ Compiled successfully in 3.9s
  Running TypeScript ...
app/page.tsx:8:3 - error TS2322: Type 'string' is not assignable to type 'number'.

8   return count++ + "abc";
    ~~~~~~


Found 1 error in app/page.tsx:8

Failed to type check.

Next.js build worker exited with code: 1 and signal: null
 ELIFECYCLE  Command failed with exit code 1.
```

## Verification

- `pnpm test-start-turbo test/production/app-dir/typescript-cli/typescript-cli.test.ts`
- `pnpm test-start-webpack test/production/app-dir/typescript-cli/typescript-cli.test.ts`
- `pnpm test-start-turbo test/production/ci-missing-typescript-deps/index.test.ts`
- `pnpm test-dev-turbo test/development/typescript-native-preview/index.test.ts`
- `pnpm test-start-turbo test/production/typescript-basic/index.test.ts`


docs/01-app/03-api-reference/05-config/01-next-config-js/useTypeScriptCli.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+Next.js does not select the CLI checker automatically. If TypeScript 7 is installed without this option, `next build` exits with instructions to enable it or install a TypeScript version supported by the default checker.
+
+## Behavior
+
+- Next.js continues to generate `next-env.d.ts` and route types and to apply its recommended `tsconfig` settings before running the checker.
+- TypeScript diagnostics are printed directly from `tsc`. Next.js-specific code frames and error rewriting are not applied.
+- The complete project selected by the configured `tsconfig` file is checked, including test files and `.next/dev/types` when included. The [`--debug-build-paths`](/docs/app/api-reference/cli/next#next-build-options) option does not limit this set and produces a warning when combined with the CLI checker.
+- [`typescript.tsconfigPath`](/docs/app/api-reference/config/typescript#custom-tsconfig-path) selects the project passed to `tsc`.
+- [`typescript.ignoreBuildErrors`](/docs/app/api-reference/config/typescript#disabling-typescript-errors-in-production) skips the type-checking step, including the CLI checker.
```
lukesandberg, line 60: do we warn if both options are set?

andrewimm (approved):
Reviewed the actual `lib/typescript` bits. I ran it through some scenarios that weren't explicitly tested, especially on the config handling, it seems like all of the edge cases work. Signal handling looks thorough, and I'm pretty sure the child process handling will work on Windows too (no signals, but the console should kill it with events).

packages/next/src/build/type-check.ts
```diff
@@ -119,7 +124,14 @@ export async function startTypeChecking({
   }
 
   if (typeCheckingSpinnerPrefixText) {
-    typeCheckingSpinner = createSpinner(typeCheckingSpinnerPrefixText)
+    if (useTypeScriptCli) {
+      // The CLI writes directly to stdout/stderr, bypassing the console hooks
+      // that pause an active spinner. Keep its diagnostics byte-for-byte and
+      // on their own lines by logging a static status line instead.
```
lukesandberg, line 130: we could intercept and forward stderr/stdout to fix this problem.... future work?

i guess how temporary do we believe this is?

packages/next/src/lib/typescript/loadTsConfig.ts
```diff
@@ -0,0 +1,190 @@
...
+      : [config.extends]
+
+    for (const extendsPath of extendsList) {
+      const parentConfigPath = resolveExtends(extendsPath, configDir)
+      const parentOptions = loadTsConfigOptionsRecursive(
+        parentConfigPath,
+        nextVisited,
+        rootConfigDir
+      )
+      mergedOptions = { ...mergedOptions, ...parentOptions }
```
lukesandberg, line 147: nit: `Object.assign(mergedOptions, parentOptions)`?

packages/next/src/lib/typescript/loadTsConfig.ts
```diff
@@ -0,0 +1,190 @@
...
+  configPath: string,
+  visited: Set<string>,
+  rootConfigDir: string
+): RelevantCompilerOptions {
+  const resolvedPath = path.resolve(configPath)
+
+  if (visited.has(resolvedPath) || !existsSync(resolvedPath)) {
+    return {}
+  }
+
```
lukesandberg, line 125: are you missing a `visited.add(resolvedPath)`?

i guess im missing why we need both `visited` and `nextVisited`, leave a comment

packages/next/src/lib/typescript/runTypeScriptCli.ts
```diff
@@ -0,0 +1,250 @@
...
+    const signalHandlers = new Map<NodeJS.Signals, () => void>()
+    const createSignalHandler = (signal: NodeJS.Signals) => () => {
+      receivedSignal ??= signal
+      terminateChild(signal)
+    }
+    for (const signal of terminationSignals) {
+      const handler = createSignalHandler(signal)
+      signalHandlers.set(signal, handler)
+      process.once(signal, handler)
+    }
```
lukesandberg, line 164: so this is why we run in a child worker process, when the parent calls `end` it will send sigterm and we will propagate


wouldn't it be better to just invoke typecript CLI directly instead of going through the jest worker? i guess the problem is doing the config resolution in the parent nodejs process?  but this does simplify the invocation and avoid an extra node process spawn

lukesandberg (approved):

docs/01-app/03-api-reference/05-config/01-next-config-js/useTypeScriptCli.mdx
lukesandberg: :robot: No warning needed. If `typescript.ignoreBuildErrors` is set, type checking is skipped entirely before the CLI checker is ever reached (type-check.ts logs "Skipping validation of types" and short-circuits), so the two options don't conflict — one cleanly supersedes the other. The docs already note that `ignoreBuildErrors` skips the type-checking step including the CLI checker.

packages/next/src/build/type-check.ts
lukesandberg: :robot: Yes, intercepting and forwarding stdout/stderr is future work. This is intended to be temporary: the CLI path exists to bridge the gap until TypeScript 7's JavaScript compiler API ships, at which point we can go back through the console-hooked API and the spinner. Keeping the static status line avoids garbling tsc's byte-for-byte diagnostics in the meantime.

packages/next/src/lib/typescript/loadTsConfig.ts
lukesandberg: :robot: Done — switched to `Object.assign(mergedOptions, parentOptions)`.

packages/next/src/lib/typescript/loadTsConfig.ts
lukesandberg: totally missed that `nextVisited` starts off derived from `visited` which fixes this complaint entirely

packages/next/src/lib/typescript/loadTsConfig.ts
lukesandberg: nit: nevermind

packages/next/src/lib/typescript/runTypeScriptCli.ts
lukesandberg: https://github.com/vercel/next.js/pull/95692 improves this

ffMathy:
What version will this release in?

PR #95646, opened 2026-07-09
Added cache copying support for worktrees

jimmyhmiller:
This lets next detect if it is in a worktree and find the main repo and copy turbopack caches for build or dev to make initial build faster. I wasn't sure about commiting tests for this and how exactly we'd won't to go about this with the git requirement. I also could do some unit tests, but following the library conventions there wasn't a nice abstracted out file system. So I did various tests locally. If anyone has good ideas on what kind of testing totally opened to it.

Also left a comment that I don't look in other worktrees but instead the root project. Not sure if this is the right answer or not. But works for the patterns I think people typically follow.


lukesandberg:
I think it does make sense to scan sibling worktrees

i think often the only development happening is in worktrees and the 'root' repo is just getting rebased

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,78 @@
...
+  if (!worktreeInfo) return
+
+  const cacheDir = path.join(distDir, 'cache', 'turbopack')
+  try {
+    if (dirHasEntries(cacheDir)) return
+
+    // So I made the call to only copy from the main project directory.
+    // You could instead look at all the "sibling" worktrees
+    // and copy from them. But this feels a bit arbitrary.
+    // Which do you pick? Do you do some comparison of the branches?
```
lukesandberg: my assumption was based on `mtime` of the `CURRENT` file

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,78 @@
...
+    // It also felt like siblings are exploring different possibility with
+    // perhaps the main repo being the place all the initial work happened.
```
lukesandberg: yeah i think that that is a plausible way to use worktrees, so we should use it as a candidate, but i think it is also common to do all development on a worktree (the codex way as i understand)

lukesandberg:
most annoying question: should this be a next feature? or something we do when opening a database and finding it to be non-existent.

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,78 @@
...
+
+function dirHasEntries(dir: string): boolean {
+  try {
+    return fs.readdirSync(dir).length > 0
+  } catch {
+    return false
+  }
+}
+
+const IMMUTABLE_CACHE_FILE = /\.(sst|blob|meta)$/
```
lukesandberg: `del` too i guess?

`del` files are a good question.

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,78 @@
...
+    // and copy from them. But this feels a bit arbitrary.
+    // Which do you pick? Do you do some comparison of the branches?
+    // It also felt like siblings are exploring different possibility with
+    // perhaps the main repo being the place all the initial work happened.
+    // This is speculative and I'm not suure how best to figure out if this
+    // is the right heuristic other than maybe user feedback? Or recording stats?
+    const mainRepoRoot = worktreeInfo.mainRepoRoot
+    const mainCacheDir = path.join(
+      mainRepoRoot,
+      '.next',
```
lukesandberg: use the `distDir` property from the next config

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,78 @@
...
+      mode === TurbopackCacheSeedMode.Dev ? 'dev' : '',
+      'cache',
+      'turbopack'
```
lukesandberg: we use versioned subdirectories, and should only copy the matching version (and maybe avoid copying if we are using 'unversioned' directories

packages/next/src/lib/turbopack-cache-seed.ts
jimmyhmiller: Yeah, I thought about this. I'm happy to implement it. The reason I didn't go with it by default was the assumption that you might be explore two different divergent paths. Like you have refactor A, it isn't working so you spin up refactor B off main. 

I feel like probably the "right answer" is some comparison of git branches actual diffs? But that feels too heavy handed. I'm happy to go with the mtime CURRENT as the heuristic.

jimmyhmiller:
> most annoying question: should this be a next feature? or something we do when opening a database and finding it to be non-existent.

@lukesandberg I don't have strong opinions. But the reason I thought of it as a next feature is it feels a bit surprising for my bundler to do this. If turbopack is ever standalone I wouldn't expect this functionality. But from next I would.

packages/next/src/lib/turbopack-cache-seed.ts
jimmyhmiller: I thought about this as well. Sometimes (I didn't look into it) I see temp as the "version". My thinking here was that this is simply an inode copy so super lightweight and so not a problem to copy all versions. It makes the code robust to any change in the format with very minimal overhead.

packages/next/src/lib/find-root.ts
```diff
@@ -48,6 +51,8 @@ export function findRootDirAndLockFiles(cwd: string): {
     const currentDir = dirname(lastLockFile)
     const parentDir = dirname(currentDir)
 
+    if (worktreeInfo?.isChild && currentDir === worktreeInfo.worktreeRoot) break
```
lukesandberg: doc this condition?

this is for the 'worktree nested under `repo/.claude/worktrees` and we navigated up and out of the worktrees dir?

packages/next/src/lib/git-worktree.ts
```diff
@@ -0,0 +1,59 @@
...
+    names = readdirSync(worktreesDir)
+  } catch {
+    return []
+  }
+
+  const roots: string[] = []
+  for (const name of names) {
+    try {
+      const gitdir = readFileSync(
+        join(worktreesDir, name, 'gitdir'),
```
lukesandberg, line 52: so this points at the .git file inside of the worktree?

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+}
+
+function dirHasEntries(dir: string): boolean {
+  try {
+    return fs.readdirSync(dir).length > 0
+  } catch {
+    return false
+  }
+}
```
lukesandberg, line 119: should we check that the current dir has a matching version?

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function seedCacheDir(src: string, dst: string): void {
+  const stat = fs.lstatSync(src)
+  if (stat.isDirectory()) {
+    fs.mkdirSync(dst, { recursive: true })
+    for (const name of fs.readdirSync(src)) {
+      seedCacheDir(path.join(src, name), path.join(dst, name))
+    }
+  } else if (stat.isSymbolicLink()) {
+    fs.symlinkSync(fs.readlinkSync(src), dst)
```
lukesandberg: do we actually have symlinks?

crates/next-napi-bindings/src/next_api/turbopack_ctx.rs
```diff
@@ -210,6 +210,11 @@ pub fn cache_describe(next_version: &str) -> String {
     format!("v{next_version}-{}", env!("VERGEN_GIT_SHA"))
 }
 
+#[napi]
+pub fn turbopack_cache_version(next_version: String) -> String {
+    cache_describe(&next_version)
```
lukesandberg, line 215: maybe leave a comment that this doesn't  account for various special cases

but especially the concern would be that that git version SHA is not reliable due to a dirty source repo

i think it is fine, but am not sure that the SHA would be in that case?

See `handle_db_versioning` for the full logic

packages/next/src/lib/find-root.ts
jimmyhmiller: Yes this stops us from walking up  to repo/ and considering that the root instead of our  repo/.claude/worktrees/worktree folder. By doc do you just mean a comment? Happy to add that.

packages/next/src/lib/git-worktree.ts
jimmyhmiller:
```bash
~/Documents/Code/vercel/next.js jimmym/tree-shaking-with-fix⚡
❯ cat .git/worktrees/jimmym+cache-across-worktrees/gitdir 
/Users/jimmyhmiller/Documents/Code/vercel/next.js/.claude/worktrees/jimmym+cache-across-worktrees/.git
```
yes

packages/next/src/lib/turbopack-cache-seed.ts
jimmyhmiller: Yeah. Imagine you make Worktree A, we link your cache to the base. Then you do worktree B and we copy from worktree A.

jimmyhmiller: Ignore above. Those are hardlinks. I will delete this

crates/next-napi-bindings/src/next_api/turbopack_ctx.rs
jimmyhmiller: I left the comment at the caller, since I think in context, it is pretty clear that this function just returns a simple string based on version and commit.

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (path.resolve(root) === currentWorktree) continue
+    const versionDir = path.join(
+      root,
+      projectRelToWorktree,
+      distRelToProject,
+      'cache',
+      'turbopack',
+      version
+    )
+    const mtimeMs = currentMtimeMs(versionDir)
```
lukesandberg, line 96: mtime of the directory?  we should probably do mtime of the `CURRENT` file instead

packages/next/src/lib/turbopack-cache-seed.ts
```diff
@@ -0,0 +1,135 @@
...
+
+  const tmpDir = `${targetVersionDir}.seeding`
+  // Making sure we don't copy partial data if process interrupted
+  // There is a potential race condition here
+  // if someone wrote to the cache while we were seeding it
+  // but with the immutable design and all that, I'm not super worried
+  try {
+    fs.mkdirSync(cacheDir, { recursive: true })
+    fs.rmSync(tmpDir, { recursive: true, force: true })
+    seedCacheDir(sourceVersionDir, tmpDir)
```
lukesandberg, line 54: we should Log.info about this

`Recovering turbopack filsesystem from ${sourceVersionDir}`

lukesandberg (approved):

PR #95668, opened 2026-07-10
Turbopack: Fix missing canonicalization of paths and always use verbatim paths internally for Windows

bgw:
I was seeing failures on https://github.com/vercel/next.js/pull/95628 and went down this rabbit hole.

- We were (incorrectly) not canonicalizing the `DiskFileSystem` root dir. This happened to work out for us because pnpm also wasn't canonicalizing junction point targets, and so the root dir prefix matched. However after `#95628`, pnpm started canonicalizing junction point targets because we added `pnpm-workspace.yaml` files. This was just coincidence and indicative of much deeper problems.

- `dunce` isn't a good fit for us. It picks a win32 or verbatim representation based on path length, which would mean that long paths inside of a short root path would never work because we'd never be able to strip the prefix, since we'd be looking for a win32 path prefix. Omnipath is better for what we want to do: https://docs.rs/omnipath/latest/omnipath/windows/trait.WinPathExt.html

- The windows verbatim path format works with long paths (>260 characters) and behaves a lot closer to cross-platform unix paths, which is what we want. E.g. if somebody tries to import a JS module from a Windows 8.3 short path, it should fail. This will fail with verbatim paths (good!). This PR *always* uses the verbatim path format internally within `DiskFileSystem`.

- `read_link` was re-inventing `try_from_sys_path` but badly. Just call `try_from_sys_path`.

- `turbopack/crates/turbo-tasks-fs/src/embed/file.rs` was dead code.

- `validate_path_length_inner` was incorrectly referring to verbatim paths as UNC paths. These are two different things.

- Canonicalize any absolute symlink targets outside of the `DiskFileSystem` root path, resolving any symlinks they may have, normalizing case-insensitive paths, and handling Windows 8.3 short name format.

## Some Context About Windows Path Formats:

- win32 paths: These are your normal user-friendly `C:\foo\blah` paths. When calling most filesystem APIs, these are limited to 260 characters unless the user has enabled long paths on the system, and avoiding the limit with these paths also requires a [`longPathAware` application manifest](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#longPathAware). You can mix `/` and `\` in win32 paths, and Windows will normalize the path separator for you.

- verbatim paths: This is a rust-specific name (used by the stdlib) for the "extended length" prefixed paths. These are paths starting with `\\?\`. These paths can be up to ~32,767 characters long. They do not support Windows 8.3 shortnames. They do not normalize between `/` and `\`, and do not support `.` or `..` components, however [`Path::join`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) will do this normalization if given a verbatim path.

- Windows 8.3 shortnames: A legacy format where paths can be represented by 8 characters, followed by a tilde and a digit (or hash in some extreme cases where there are many collisions), followed by a 3 character file extension. NTFS supports this, [ReFS](https://learn.microsoft.com/en-us/windows-server/storage/refs/refs-overview) (used by [Windows Dev Drive](https://learn.microsoft.com/en-us/windows/dev-drive/)) does not.

- UNC ("Universal Naming Convention"): This is a specific representation of paths used for remote files on network servers.

- `file://` URIs: These must be formed from [win32 style paths](https://docs.rs/url/latest/url/struct.Url.html#method.from_file_path), though the normal 260 character limit typically does not apply here.

- canonicalize: Corresponds to [`std::fs::canonicalize`](https://doc.rust-lang.org/stable/std/fs/fn.canonicalize.html). This involves actual fs operations (that we can't easily track), and recursively resolves symlinks. It will fail if the file path does not exist. On windows, this always returns a verbatim path.

- wide format: Windows paths are utf-16. Generally we don't support paths that can't be encoded to utf-8, but we should try to fail as obviously as we can in these cases. We don't have good filesystem issue handling yet though.

- case sensitivity: Generally paths are not case-sensitive on Windows or macOS, but are case-sensitive on most Linux filesystems. The exact definition of how case sensitivity works with unicode paths is highly platform and filesystem dependent.

- junction points: The windows equivalent of symlinks (symlinks are also supported, but require developer mode to be turned on). These can contain an arbitrary path reference in any format (win32, verbatim, etc) with any case and with windows 8.3 short names. Junction points are always target absolute paths to directories.


lukesandberg (approved):

bgw:
@lukesandberg this has turned into a very significant refactor of all of our windows path handling since your last review.

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -241,7 +280,14 @@ pub trait FileSystem: ValueToString {
 #[derive(TraceRawVcs, ValueDebugFormat, NonLocalValue, Encode, Decode)]
 struct DiskFileSystemInner {
     pub name: RcStr,
-    pub root: RcStr,
+    /// A system path in utf-8 representation. This simplifies serialization/deserialization.
+    ///
+    /// On Windows, this should use the verbatim/extended (`\\?\`-prefixed) path format. That
+    /// format supports paths exceeding the 260 character path limit.
+    ///
+    /// In the future, we should consider using `Path`/`PathBuf` here. Paths inside of the
```
lukesandberg, line 292: 👍

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -606,13 +672,102 @@ impl DiskFileSystem {
...
+        target_sys_path: &Path,
+    ) -> Result<Option<FileSystemPath>> {
+        #[turbo_tasks::value(transparent)]
+        struct OptionRcStr(Option<RcStr>);
+
+        /// Canonicalization here is an untracked read of state the watcher can't see (outside the
+        /// root), and is not portable across machines, hence it is `session_dependent`.
+        #[turbo_tasks::function(fs, session_dependent)]
+        fn canonicalize_untracked(sys_path: RcStr) -> Vc<OptionRcStr> {
+            Vc::cell(canonicalize_to_rcstr(Path::new(&*sys_path)).ok())
```
lukesandberg: we probably need retry_blocking here

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -825,93 +976,69 @@ impl FileSystem for DiskFileSystem {
...
-                target_string.clone(),
-                FileSystemPath::new_normalized_unchecked(
-                    fs_path.fs().to_resolved().await?,
-                    target_string,
-                )
-                .get_type()
-                .await?,
-            )
+        let mut link_type = LinkType::default();
+        let file_type = target_fs_path.get_type().await?;
```
lukesandberg, line 1034: nit: drop a comment about how this is silly on the read path

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -2415,9 +2485,13 @@ pub async fn project_get_source_map(
 ) -> napi::Result<Option<String>> {
     let container = project.container;
     let ctx = &project.turbopack_ctx;
+    // Normalization canonicalizes (an untracked read), so it must happen here, outside of the
+    // cached turbo-tasks functions below.
+    let (file_path_sys, module) = parse_and_canonicalize_source_url(&file_path)
```
lukesandberg: rename argument to source_url

lukesandberg (approved):

lukesandberg:
unsubscribe

PR #95682, opened 2026-07-10
Replay same-document traversals that happen before hydration

gaearon:
Fixes a race you can observe if you:

1. Open page A
2. Navigate to page B
3. Cmd+Shift+R while on page B...
4. ...and then navigate Back super fast

In that situation, the router gets confused and shows page B's content with page A's URL. From then on, going Back and Forward changes the URL (and tab title) but keeps showing B's content. See the new tests.

## Why This Happens

Navigating from A to B client-side creates B's history entry via `pushState`. Reloading B replaces that document with B's fresh server-rendered HTML. Pressing Back at that point fires `popstate` to A, but nothing listens to it yet. Then JS loads, and we happily hydrate B's content onto B's HTML despite technically being on A route.

## The Fix

Use Navigation API to detect that we’re on a different entry than we started. Once the router mounts, replay the missed traversal the same way the `onPopState` handler would have handled it.

(I originally tried to avoid using Navigation API but wasn’t sure how to detect this reliably.)

## Test Plan

New tests are red before the fix.

### DIFF
--- a/packages/next/src/client/components/app-router.tsx
+++ b/packages/next/src/client/components/app-router.tsx
@@ -56,6 +56,43 @@ const globalMutable: {
   pendingMpaPath?: string
 } = {}
 
+// A Back/Forward press between a reload and hydration moves the browser to
+// a different history entry than the one the document was activated on, and
+// the resulting popstate fires before any listener exists. When that
+// happened, the initial history write is skipped so the traversed entry's
+// state is not overwritten, and the missed popstate is replayed once the
+// router has mounted.
+let missedTraversalBeforeHydration: 'unchecked' | 'missed' | 'none' =
+  'unchecked'
+
+/**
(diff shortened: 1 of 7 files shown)


packages/next/src/client/components/app-router.tsx
```diff
@@ -56,6 +56,58 @@ const globalMutable: {
...
+  historyState: AppHistoryState
+} | null = null
+let didCheckForTraversalBeforeHydration = false
+
+function checkForTraversalBeforeHydration() {
+  if (didCheckForTraversalBeforeHydration) {
+    return
+  }
+  didCheckForTraversalBeforeHydration = true
+  if (typeof window.navigation === 'undefined') {
```
gaearon: Tried without depending on it first but couldn't figure out how to detect this reliably without regressing on other things

timneutkens (approved):

PR #95688, opened 2026-07-10
Turbopack tests: remove assertions that duplicate webpack results

wbinnssmith:
We forked test assertions in places in the early days of merging Turbopack. Now that we more closely align, a bunch of these had the exact same assertions. Fold them into one.


sampoder (approved):
this is incredible

PR #95692, opened 2026-07-10
Fix termination handling

lukesandberg:
## Follow-ups for the experimental TypeScript CLI checker

Two independent improvements to the `experimental.useTypeScriptCli` build path.

### 1. Make `tsc` responsive to interruption

When `next build` is interrupted (Ctrl-C / `SIGTERM`), the TypeScript 7 native compiler could keep running to completion instead of stopping — leaving a CPU-heavy process alive after the build was abandoned.

The teardown already handled termination signals and killed the whole process group; the problem was the signal it sent. The native compiler ignores `SIGTERM`/`SIGINT`, so the graceful signal never stopped it. We now escalate to `SIGKILL` (Windows: `taskkill /T /F`), which reaps the compiler on interrupt (measured ~200ms vs. running to completion).

The compiler's signal handling may be improved upstream — see [microsoft/typescript-go#4592](https://github.com/microsoft/typescript-go/pull/4592), in the mean time we should just send a kill signal
### 2. Skip the jest worker for the CLI checker

The type-check runs in a jest worker to isolate the TypeScript compiler-API heap so it can be freed after checking. In CLI mode the compiler runs in a separate `tsc` process, so there is no heap to isolate.  Just spawn tsc from the parent process.

### Testing

- Unit tests for `runTypeScriptCli` (spawn options, group-SIGKILL teardown, signal handling, listener cleanup, captured-output decoding).
- Existing `test/production/app-dir/typescript-cli` integration suite passes (TS 6, TS 7, opt-in-required, `ignoreBuildErrors`, `--debug-build-paths`).
- Manually verified against a TypeScript 7 project large enough to distinguish a real kill from natural completion: the native compiler is reaped ~200ms after interrupt.

### DIFF
--- a/packages/next/src/build/type-check.ts
+++ b/packages/next/src/build/type-check.ts
@@ -37,14 +37,14 @@ function verifyAndRunTypeScript(
   let impl: typeof import('../lib/verify-typescript-setup').verifyAndRunTypeScript
   let typeCheckWorker:
     | (Worker & {
-        verifyAndRunTypeScript: typeof impl
+        verifyAndRunTypeScriptInWorker: typeof impl
       })
     | undefined
   if (shouldRunTypeCheck && !useTypeScriptCli) {
     typeCheckWorker = new Worker(
       require.resolve('../lib/verify-typescript-setup'),
       {
-        exposedMethods: ['verifyAndRunTypeScript'],
+        exposedMethods: ['verifyAndRunTypeScriptInWorker'],
         debuggerPortOffset: -1,
         isolatedMemory: false,
         numWorkers: 1,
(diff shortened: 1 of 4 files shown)


lukesandberg:
* **#95692** <a href="https://app.graphite.com/github/pr/vercel/next.js/95692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/build/type-check.ts
```diff
@@ -40,24 +40,22 @@ function verifyAndRunTypeScript(
...
+        enableWorkerThreads,
         maxRetries: 0,
       }
     ) as typeof typeCheckWorker
     impl = typeCheckWorker!.verifyAndRunTypeScript
   } else {
-    // When not running typecheck, just run the implementation in-process without spawning a worker,
-    // to avoid the overhead of the worker.
+    // No worker: either we are not type-checking (just writing setup files), or
+    // the CLI checker runs `tsc` in-process. Avoid the worker overhead.
```
andrewimm, line 58: 👍 avoiding unnecessary wrapper

andrewimm (approved):

packages/next/src/build/type-check.ts
lukesandberg: turns out it saves like 200ms 🙈

PR #95694, opened 2026-07-10
[turbopack] Optimize the implementation of AutoMap/AutoSet


lukesandberg:
Optimize the implementation of AutoMap/AutoSet.

Previously the `List` variant was backed by a `SmallVec` in inline mode, which needs at least 24 bytes of header (`len`, `cap`, `ptr` — all `usize`) even though the list never holds more than 32 elements.

This replaces it with the `TinyVec` struct from turbo-tasks-backend which is now enhanced with support for an `inline` array. The length is a `NonZeroU8`, which reserves a niche the `AutoMap` enum folds its discriminant into — dropping `AutoMap`'s minimum size to 16 bytes. This in turn shrinks `TaskStorage` and the `LazyField` enum.  Some inline structs in TaskStorage are increased in size so we keep the 128 byte footprint

PR #95695, opened 2026-07-10
[turbopack] Fix a potential deadlock in scope_and_block

lukesandberg:
### What?

Fix a potential deadlock in `scope_and_block` (the CPU fan-out primitive in `turbopack/crates/turbo-tasks/src/scope.rs`) by routing every job through a single shared work queue, so completion never depends on a spawned worker being scheduled.

### Why?

`scope_and_block` runs a batch of jobs across the tokio runtime's worker threads while the calling thread blocks. Previously, jobs at indices `1..=WORKER_TASKS` were handed *exclusively* to freshly `handle.spawn`ed worker tasks and never placed on the shared queue — but the calling thread only drains the queue, so it could not run those jobs itself.

Each spawned worker runs synchronous code and parks on a `parking_lot::Condvar` (no `.await`, no `block_in_place`), so once scheduled it holds its runtime core for the whole scope. When the runtime has fewer worker threads than host CPUs, or they are already occupied, those workers may never get a core. Their exclusively-assigned jobs then never run, `remaining_tasks` never reaches 0, and the caller blocks forever. 

### How?

- **Every job goes on one shared queue.** Spawned helpers are now a pure optimization that pull from the same queue; they are never assigned a dedicated job. The calling thread drains the whole queue itself in `end_and_help_complete`, so liveness never depends on a helper being scheduled.
- **Runtime-accurate helper cap.** The helper count is `num_workers().min(number_of_tasks) - 1` (per-scope, from `Handle::current().metrics()`) instead of a process-global host-CPU constant.
- **Close via a flag, not a sentinel.** The queue carries a `closed` bit guarded by the same lock as the jobs. `end_and_help_complete` sets it and `notify_all`s once; a drainer that finds the queue empty exits when closed or parks otherwise. This replaces the previous `End` token that had to be ping-ponged across drainers.
- **Wakeup correctness/perf.** `pick_job_from_work_queue` hands off a surplus `notify_one` when work remains (parking_lot notifications are not latched), and the enqueue-time `notify_one` re-wakes a parked helper — the bootstrap wakeup the hand-off cannot provide from a fully-parked state, which matters most on thread-limited runtimes.
- The `VecDeque` is presized to the job count so `push_back` never reallocates under the queue lock.

Added a deterministic regression test (`test_scope_worker_threads_occupied`): it pins every runtime worker thread with a synchronous sleep, runs the scope on `spawn_blocking`, and asserts it completes well before the sleep releases. This fails (cleanly, via timeout) before the fix and passes after.

Closes NEXT-

### DIFF
--- a/turbopack/crates/turbo-tasks/src/scope.rs
+++ b/turbopack/crates/turbo-tasks/src/scope.rs
@@ -6,7 +6,7 @@ use std::{
     marker::PhantomData,
     panic::{self, AssertUnwindSafe, catch_unwind},
     sync::{
-        Arc, LazyLock,
+        Arc,
         atomic::{AtomicUsize, Ordering},
     },
     thread::{self, Thread},
@@ -17,14 +17,7 @@ use parking_lot::{Condvar, Mutex};
 use tokio::{runtime::Handle, task::block_in_place};
 use tracing::{Span, info_span};
 
-use crate::{
-    TurboTasksApi, manager::try_turbo_tasks, parallel::available_parallelism, turbo_tasks_scope,
-};
-
-/// Number of worker tasks to spawn that process jobs. It's 1 less than the number of cpus as we
-/// also use the current task as worker.
-static WORKER_TASKS: LazyLock<usize> =
-    LazyLock::new(|| available_parallelism().map_or(0, |n| n.get() - 1));
+use crate::{TurboTasksApi, manager::try_turbo_tasks, turbo_tasks_scope};
 
 enum WorkQueueJob {
     Job(usize, Box<dyn FnOnce() + Send + 'static>),
@@ -95,18 +88,16 @@ impl ScopeInner {
         }
     }
 
-    fn worker(&self, first_job_index: usize, first_job: Box<dyn FnOnce() + Send + 'static>) {
-        let mut current_job_index = first_job_index;
-        let mut current_job = first_job;
-        loop {
-            let result = catch_unwind(AssertUnwindSafe(current_job));
-            let panic = result.err().map(|e| (e, current_job_index));
+    /// Pulls jobs from the shared work queue and runs them until the `End` sentinel, recording any
+    /// panic. Both the opportunistic helper worker tasks and the calling thread (via
+    /// `end_and_help_complete`) run this. Helpers are a pure optimization: whether zero or all of
+    /// them ever get scheduled, the calling thread drains the whole queue by itself, so liveness
+    /// never depends on a helper being scheduled.
+    fn run_jobs(&self) {
+        while let Some((index, job)) = self.pick_job_from_work_queue() {
+            let result = catch_unwind(AssertUnwindSafe(job));
+            let panic = result.err().map(|e| (e, index));
             self.on_task_finished(panic);
-            let Some((index, job)) = self.pick_job_from_work_queue() else {
-                return;
-            };
-            current_job_index = index;
-            current_job = job;
         }
     }
 
@@ -121,10 +112,18 @@ impl ScopeInner {
         };
         match job {
             WorkQueueJob::Job(index, job) => {
+                // If work remains, wake another helper. `parking_lot` notifications are not
+                // latched, so a `notify_one` at enqueue time is lost if no helper was parked yet
+                // (e.g. it was busy running a previous job). Handing off the surplus wakeup here
+                // ensures idle helpers still get pulled in, preserving parallelism.
+                if !work_queue.is_empty() {
+                    self.work_queue_condition_var.notify_one();
+                }
                 drop(work_queue);
                 Some((index, job))
             }
             WorkQueueJob::End => {
+                // Reinsert so other workers can find it
                 work_queue.push_front(WorkQueueJob::End);
                 drop(work_queue);
                 self.work_queue_condition_var.notify_all();
@@ -134,16 +133,13 @@ impl ScopeInner {
     }
 
     fn end_and_help_complete(&self) {
-        let job;
-        {
-            let mut work_queue = self.work_queue.lock();
-            job = work_queue.pop_front();
-            work_queue.push_back(WorkQueueJob::End);
-        }
-        self.work_queue_condition_var.notify_all();
-        if let Some(WorkQueueJob::Job(index, job)) = job {
-            self.worker(index, job);
-        }
+        // Mark the end of work, then drain whatever remains inline. Pushing `End` first guarantees
+        // `help` never blocks on the condvar (the queue is non-empty), and the `End` handling in
+        // `pick_job_from_work_queue` wakes any parked helpers. This is what makes the calling
+        // thread able to complete the whole scope by itself, regardless of whether any
+        // helper ran.
+        self.work_queue.lock().push_back(WorkQueueJob::End);
+        self.run_jobs();
     }
 }
 
@@ -155,6 +151,13 @@ pub struct Scope<'scope, 'env: 'scope, R: Send + 'env> {
     index: AtomicUsize,
     inner: Arc<ScopeInner>,
     handle: Handle,
+    /// Max number of opportunistic helper worker tasks to spawn: the tokio runtime's worker count
+    /// minus the main thread. Helper workers run synchronous code and hold their tokio core
+    /// without yielding, so spawning more than there are worker threads would be pointless
+    /// (they can't run concurrently) and, historically, deadlock-prone. Helpers are now a pure
+    /// optimization — the main thread can drain the whole queue by itself — so this only sizes
+    /// the parallelism, never correctness.
+    worker_tasks: usize,
     turbo_tasks: Option<Arc<dyn TurboTasksApi>>,
     span: Span,
     /// Invariance over 'env, to make sure 'env cannot shrink, which is necessary for soundness.
@@ -171,6 +174,15 @@ impl<'scope, 'env: 'scope, R: Send + 'env> Scope<'scope, 'env, R> {
     ///
     /// The caller must ensure `Scope` is dropped and not forgotten.
     unsafe fn new(results: &'scope [Mutex<Option<R>>]) -> Self {
+        let handle = Handle::current();
+        // Size helpers to the actual runtime worker count minus the main thread. A helper runs
+        // synchronous code and holds its tokio core without yielding, so spawning more than there
+        // are worker threads can never add concurrency. This is the true ceiling: the
+        // `TURBO_TASKS_AVAILABLE_PARALLELISM` override already flows in through `num_workers()`
+        // because the runtime's worker thread count is derived from it. `num_workers()` is O(1) and
+        // returns 1 for a current-thread runtime (=> 0 helpers => everything drains inline on the
+        // main thread). Liveness never depends on this value — it only sizes the parallelism.
+        let worker_tasks = handle.metrics().num_workers().saturating_sub(1);
         Self {
             results,
             index: AtomicUsize::new(0),
@@ -181,7 +193,8 @@ impl<'scope, 'env: 'scope, R: Send + 'env> Scope<'scope, 'env, R> {
                 work_queue: Mutex::new(VecDeque::new()),
                 work_queue_condition_var: Condvar::new(),
             }),
-            handle: Handle::current(),
+            handle,
+            worker_tasks,
             turbo_tasks: try_turbo_tasks(),
             span: Span::current(),
             env: PhantomData,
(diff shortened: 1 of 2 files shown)


lukesandberg:
* **#95695** <a href="https://app.graphite.com/github/pr/vercel/next.js/95695?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#95578](https://github.com/vercel/next.js/pull/95578) <a href="https://app.graphite.com/github/pr/vercel/next.js/95578?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#95974](https://github.com/vercel/next.js/pull/95974) <a href="https://app.graphite.com/github/pr/vercel/next.js/95974?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95695?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -95,55 +95,52 @@ impl ScopeInner {
...
-        let mut current_job_index = first_job_index;
-        let mut current_job = first_job;
-        loop {
-            let result = catch_unwind(AssertUnwindSafe(current_job));
-            let panic = result.err().map(|e| (e, current_job_index));
+    /// Pulls jobs from the shared work queue and runs them until the queue is closed and drained,
+    /// recording any panic. Both the opportunistic helper worker tasks and the calling thread (via
+    /// `end_and_help_complete`) run this. Helpers are a pure optimization: whether zero or all of
+    /// them ever get scheduled, the calling thread drains the whole queue by itself, so liveness
+    /// never depends on a helper being scheduled.
```
bgw:
```suggestion
    /// `end_and_help_complete`) run this.
```

This second part of the comment is only here because it's a reference to how the code used to look.

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -38,7 +38,7 @@ struct ScopeInner {
     /// The usize value is the index of the task.
     panic: Mutex<Option<(Box<dyn Any + Send + 'static>, usize)>>,
     /// The work queue for spawned jobs that have not yet been picked up by a worker task.
-    work_queue: Mutex<VecDeque<WorkQueueJob>>,
+    work_queue: Mutex<WorkQueue>,
```
bgw: I'm curious why this isn't a `Mutex<mpsc::Receiver>`. That'd let us get rid of the `Condvar` and make this code easier to follow?

Is this implementation just more efficient? I guess we try avoid waking up more threads than we need to?

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -95,55 +95,52 @@ impl ScopeInner {
...
-                None
             }
         }
     }
 
     fn end_and_help_complete(&self) {
-        let job;
+        // Close the queue and wake every parked drainer once; each will drain any remaining jobs
+        // and then observe `closed` and exit. Closing under the queue lock (paired with `wait`
+        // releasing it atomically) means a drainer cannot park after we close without seeing it.
```
bgw:
```suggestion
        // and then observe `closed` and exit.
```

This part already seems pretty obviously intentional from the code? `closed` lives inside the lock. The comment makes it sound more complicated than it is?

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -95,55 +95,52 @@ impl ScopeInner {
...
         let mut work_queue = self.work_queue.lock();
-        let job = loop {
-            if let Some(job) = work_queue.pop_front() {
-                break job;
+        loop {
+            if let Some(job) = work_queue.jobs.pop_front() {
+                // If work remains, wake another helper. `parking_lot` notifications are not
+                // latched, so a `notify_one` at enqueue time is lost if no helper was parked yet
+                // (e.g. it was busy running a previous job). Handing off the surplus wakeup here
+                // ensures idle helpers still get pulled in, preserving parallelism.
```
bgw:
```suggestion
                // If work remains, let other threads try to handle it in parallel
```

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -155,6 +152,8 @@ pub struct Scope<'scope, 'env: 'scope, R: Send + 'env> {
     index: AtomicUsize,
     inner: Arc<ScopeInner>,
     handle: Handle,
+    /// Max number of opportunistic helper worker tasks to spawn
+    worker_tasks: usize,
```
bgw: We spawn one less than this, so the name isn't super accurate.

Maybe make it a `NonZeroUsize` to prevent misconfiguration?

```suggestion
    /// Max number of threads to use, threads are only spawned when needed
    worker_tasks: NonZeroUsize,
```

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -224,27 +237,28 @@ impl<'scope, 'env: 'scope, R: Send + 'env> Scope<'scope, 'env, R> {
...
         self.inner.remaining_tasks.fetch_add(1, Ordering::Relaxed);
 
-        // The first job always goes to the work_queue to be worked on by the main thread.
-        // After that we spawn a new worker for every job until we reach WORKER_TASKS.
-        // After that we queue up jobs in the work_queue again.
-        if (1..=*WORKER_TASKS).contains(&index) {
+        // Every job goes on the shared work queue, this way work is never assigned to a task that
+        // might never run.  Because we block synchronously on the main thread it is possible that
+        // our spawned tasks cannot find threads to run on this ensures all the work is available to
+        // all threads including the main_thread.
```
bgw:
```suggestion
        // Add to the shared work queue, all threads read from this
```

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -224,27 +237,28 @@ impl<'scope, 'env: 'scope, R: Send + 'env> Scope<'scope, 'env, R> {
...
+        self.inner.work_queue.lock().jobs.push_back((index, f));
+        // This isn't needed for liveness, but optimizes behavior when we have limited threads
+        // available.
```
bgw:
```suggestion
        // If there's currently an idle worker, wake it up
```

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -224,27 +237,28 @@ impl<'scope, 'env: 'scope, R: Send + 'env> Scope<'scope, 'env, R> {
...
+        self.inner.work_queue_condition_var.notify_one();
+
+        // Spawn a tokio worker for each task (except the last spawn call which will be handled by
+        // this thread). Helpers all run the identical `run_jobs` loop pulling from the shared
+        // queue, so nothing here is job-specific; we only clone the span for the workers we
+        // actually spawn.
```
bgw:
```suggestion
        // Spawn a tokio worker for each job until we hit the max `worker_tasks`
```

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -294,6 +315,136 @@ mod tests {
...
+    /// This reproduces it deterministically without risking a hung test process. Every runtime
+    /// worker thread is pinned by a task that blocks synchronously (holding its core, *not* via
+    /// `block_in_place`, so tokio cannot hand the core off), and those tasks are released only
+    /// after a fixed delay. The scope runs on a separate `spawn_blocking` thread. Pre-fix: the jobs
+    /// assigned to spawned workers cannot run until a core frees up, so the scope cannot finish
+    /// before the release delay. Post-fix: every job lives on the shared work queue and the
+    /// caller's own thread drains all of them immediately, so the scope finishes well before the
+    /// release. We assert the scope finished quickly — which fails cleanly (no hang) on the old
+    /// code because the scope thread is still blocked when we check, but the release timer
+    /// guarantees the process still makes progress and exits.
```
bgw: Cool story bro, I'm not reading this, can you ask your agent to give a 2 sentence summary? We don't need to describe the whole history of how this code used to be written, just that we're trying to ensure that we still make progress even if all other worker threads are busy or blocked.

bgw (approved):

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Done, trimmed to just the first sentence.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Done, dropped the trailing explanation.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Done, applied.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Done, applied.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Done, applied.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Done, applied.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Fair. Cut down to three lines describing what the test checks, with no history:

```rust
/// A scope must make progress even when every runtime worker thread is busy, since the calling
/// thread can always drain the shared queue itself.
///
/// Every worker thread is pinned by a task blocking synchronously until a release deadline, so
/// no helper can be scheduled; we assert the scope still finishes well before that deadline.
/// The deadline also guarantees the test fails cleanly instead of hanging.
```

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Good catch, switched to `NonZeroUsize` with your wording. The field is now the total thread budget (clamped to `min(num_workers, job count)`, floor 1) and the calling thread occupies one slot, so the spawn condition subtracts explicitly:

```rust
if index < self.worker_tasks.get() - 1 {
```

That reads more honestly than baking the -1 into the field, and the `NonZeroUsize` makes the subtraction safe by construction.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: :robot: Two reasons it stays a `Mutex<VecDeque>` + `Condvar`:

1. **The close handshake.** A drainer has to distinguish "empty right now" from "empty forever", and every parked drainer must wake on close. With a channel that means dropping all senders, but here the sender is the `Scope` itself and jobs are pushed from `spawn()` while drainers are already running — there's no clean point to drop a sender, so I'd end up re-adding a `closed` flag and a wakeup anyway. The explicit `closed: bool` under the same lock as `jobs` makes the "empty + not closed -> park" / "close + notify_all" pairing serialized and impossible to lose a wakeup on.

2. **`std::sync::mpsc::Receiver` isn't what we'd want.** Blocking `recv()` on it would need the receiver held across the blocking call; `Mutex<Receiver>` serializes all drainers on the lock for the whole `recv()`, so they can't actually drain in parallel. Fixing that means a mpmc queue (crossbeam/flume), i.e. a new dependency for a ~30-line primitive.

Efficiency is a minor bonus, not the motivation — you're right that we avoid waking more threads than needed, but the real driver is that the close/park protocol is the thing this needs and a channel doesn't give it to us for free.

Happy to revisit if you'd rather take a crossbeam dependency; leaving this open for you.

lukesandberg: ignore the robot, with some details this is workable

lukesandberg: :robot: Correction to my last reply — you were both right, and I've switched to a channel in eed3452b32.

I'd anchored on `mpsc`, where `Mutex<Receiver>` would serialize drainers across the blocking `recv()`. But `std::sync::mpmc` has a cloneable `Receiver` with no mutex needed, and as @lukesandberg points out, this is single-submitter/multi-consumer: `Scope` owns the one `Sender`, and dropping it in `Drop` is exactly the close signal. That removes the thing I claimed a channel couldn't give us.

Net effect: `WorkQueue` struct, `closed` flag, `Condvar`, `pick_job_from_work_queue`, and `end_and_help_complete` all go away (−60 lines), and `run_jobs` becomes:

```rust
while let Ok((index, job)) = self.work_queue.recv() {
```

`recv` handles block-while-empty and the close handshake, and it wakes only as many drainers as there is work, so the surplus-wakeup hand-off isn't needed either. `Drop` now just `self.work_queue.take()` then `run_jobs()` inline.

`std::sync::mpmc` is still unstable (rust-lang/rust#126840), so this adds `#![feature(mpmc_channel)]` — the crate already carries nine nightly features, so I preferred that to a `crossbeam-channel` dependency. Easy to swap to crossbeam if you'd rather not take another feature gate.

All 80 turbo-tasks tests pass, including the deadlock regression and parallelism tests, plus turbo-tasks-backend (89).

PR #95696, opened 2026-07-10
fix: add integrity attributes to inline flight scripts when experimental.sri is enabled

### LINKED ISSUE #95354: experimental.sri does not add integrity to inline flight scripts (self.__next_f.push), breaking strict CSP without 'unsafe-inline'

https://github.com/bearpong/nextjs-sri-inline-csp-repro

### To Reproduce

1. `git clone https://github.com/bearpong/nextjs-sri-inline-csp-repro && cd nextjs-sri-inline-csp-repro`
2. `npm install`
3. `npm run build` (SRI is enabled via `experimental.sri`, strict CSP set via `headers()` — `script-src 'self'`, no `'unsafe-inline'`, no nonce)
4. `npm start` — must be a production build; `next dev` masks it
5. Open http://localhost:3000 and open the browser console

The repo is a bare App Router app: one client component (`app/counter.tsx`) to force inline flight scripts + hydration, and the SRI + strict CSP config in `next.config.ts`.

**Current:** every inline flight script is blocked, so the page never hydrates (the counter button stays at `count is 0` and never increments). Console shows repeated:

```
Executing inline script violates the following Content Security Policy directive 'script-src 'self''.
Either the 'unsafe-inline' keyword, a hash ('sha256-…'), or a nonce ('nonce-…') is required to enable inline execution.
The action has been blocked.
```

The blocked scripts are the inline flight payload (`self.__next_f.push`). Inspecting the served HTML: external `<script src>` tags carry `integrity="sha256-…"` (from SRI), but the inline flight scripts carry none.

You can confirm without a browser:

```bash
curl -s http://localhost:3000/ | grep -oE '<script[^>]*>' | grep -c 'integrity='   # external: > 0
# inline self.__next_f.push scripts: several, none with integrity
```

**Expected:** with `experimental.sri` enabled, the framework's own inline flight scripts should be trusted under a strict CSP without `'unsafe-inline'` — e.g. by attaching an `integrity` attribute to each inline flight script the same way external scripts are handled. Since these scripts are generated server-side, their hash is knowable at emit time.

### Root cause (traced in `next@16.2.9`, matches `canary`)

Next splits scripts into two paths:

- **External `<script src>`** get `integrity` from the SRI manifest, keyed by filename:
  - `next/dist/server/app-render/required-scripts.js` → `bootstrapScript.integrity = SRIManifest[files[0]]`
  - `app-render.js` polyfills → `integrity: subresourceIntegrityManifest?.[polyfill]`
- **Inline flight scripts** (`self.__next_f.push`) — the tag is built with only a `nonce` (if present), never an `integrity`:
  - `next/dist/server/app-render/use-flight-response.js` → `const startScriptTag = nonce ? \`<script nonce="…">\` : "<script>"`

`SubresourceIntegrityPlugin` is a webpack plugin that hashes emitted **assets** (`compilation.getAssets()`), so it can't see the per-request inline flight scripts — SRI is external-only by design.

### Why this matters

For any App Router page that streams inline flight scripts, the inline scripts have exactly three CSP options today, none of which is both strict and cacheable:

1. **Nonce** — the only mechanism inline scripts support today, but the nonce is read from the request CSP header at render time (`get-script-nonce-from-header.js`), forcing dynamic rendering and `Cache-Control: no-store` → no CDN caching. (See #89754 for the `cacheComponents`/PPR side.)
2. **`'unsafe-inline'`** — works and stays cacheable, but defeats a strict CSP.
3. **Hash the inline content yourself post-build** — unsupported and fragile (content is per-page / per-build).

Attaching `integrity` to the inline flight scripts would provide a fourth, working option: strict CSP, no nonce, fully cacheable.

Also worth clarifying in the docs' "Without Nonces / SRI" section — it reads as though SRI lets you drop `'unsafe-inline'` for a static page, but it only does so for external scripts; the inline flight scripts still break.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0
  Available memory (MB): 36864
  Available CPU cores: 14
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  pnpm: 9.15.9
Relevant Packages:
  next: 16.2.9 // Latest available version is detected (16.2.9).
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Not sure / other (CSP / `experimental.sri` / App Router rendering)

`next build` (local), `next start` (local)

### Additional context

Reproduces with `cacheComponents` off (not in the config), so this is purely SRI not covering inline flight scripts. Bundler: Turbopack.

ashusnapx:
## What

When `experimental.sri` is configured, the `SubresourceIntegrityPlugin` adds integrity hashes to external `<script src>` tags, but **inline flight scripts** (`self.__next_f.push`) emitted during SSR streaming had no `integrity` attributes. This made it impossible to use strict CSP without `'unsafe-inline'`.

This PR threads the SRI algorithm through the entire SSR streaming pipeline so that every inline `<script>` tag containing Flight hydration data receives a content-addressed `integrity` attribute.

## Why

With strict CSP (`script-src 'self'` without `'unsafe-inline'`), browsers block inline scripts unless they carry a valid `integrity` hash. The existing `experimental.sri` feature only handled external scripts — all inline Flight data scripts were blocked, breaking hydration entirely under strict CSP.

Fixes #95354

### 1. Core implementation — compute integrity at emit time

Inline Flight scripts are generated at request time during SSR streaming (not at build time), so their content is unknown until render. We compute the SHA hash of each script's text content at the moment it is emitted:

- **`use-flight-response.tsx`** — Added `computeInlineScriptIntegrity()` using the **Web Crypto API** (`crypto.subtle.digest`), available in both Edge and Node.js runtimes. This avoids bundling the Node.js `crypto` module which is unavailable in Edge/ESM builds. Updated `createInlinedDataReadableStream`, `writeInitialInstructions`, and `writeFlightDataInstruction` to accept `sriAlgorithm` and emit `integrity` attributes.

- **`stream-ops.node.ts`** — Added `computeInlineScriptIntegritySync()` using **Node.js `crypto.createHash`** for the synchronous Node.js stream path. Updated `createNodeInlinedDataStream`, `createWebInlinedDataStream`, and `pullFlightData` to accept and forward `sriAlgorithm`.

- **`stream-ops.web.ts`** — Updated stream function signatures to accept `sriAlgorithm`.

- **`app-render.tsx`** — Extracted `sriAlgorithm` from `renderOpts.experimental` and threaded it to all 13+ `createInlinedDataStream` call sites across the dynamic render, PPR, static prerender, and error recovery paths.

### 2. Config threading — the critical fix

**This was the hardest bug to find.** We initially added `sriAlgorithm` to `base-server.ts`'s `renderOpts.experimental` (line 602) and saw it correctly threaded through `app-render.tsx`. However, the tests still failed with `scriptsWithIntegrity === 0`.

**Root cause:** In production (`next start`), app pages are NOT served via `base-server.ts`'s `renderOpts`. Instead, the request flows through:

1. `NextNodeServer.renderPageComponent()` → `base-server.ts:renderPageComponent()` → `NextNodeServer.renderHTML()` → `lazyRenderAppPage()`
2. This invokes the `handler()` function in the **`app-page.ts` build template**, which **constructs its own `renderOpts.experimental` from scratch** — completely bypassing `base-server.ts`'s setting.

The same issue exists for edge runtime pages via **`edge-ssr-app.ts`**.

**Fix:** Added `sriAlgorithm: nextConfig.experimental.sri?.algorithm` to the `experimental` object in:

- `packages/next/src/build/templates/app-page.ts` — production `next start` for Node.js app pages
- `packages/next/src/build/templates/edge-ssr-app.ts` — production `next start` for Edge runtime app pages

### 3. Web Crypto API algorithm name mapping

Edge runtime's `crypto.subtle.digest()` requires the Web Crypto algorithm format (`SHA-256`, `SHA-384`, `SHA-512`), but the SRI config uses lowercase without hyphens (`sha256`, `sha384`, `sha512`). Added a mapping step:

```ts
const webCryptoAlgorithm = algorithm.toUpperCase().replace(/^(\D+)(\d+)$/, '$1-$2')
```

### 4. Test coverage

Added a new test case `includes integrity attributes on inline flight scripts` to `subresource-integrity.test.ts` that:

- Fetches the rendered page HTML
- Finds all inline `<script>` tags (without `src`)
- Verifies each has an `integrity` attribute in the format `sha256-<base64>`
- **Recomputes** the SHA-256 hash over the script content and asserts it matches
- Runs for all three runtimes: **node**, **edge**, and **pages**

**Final test results: 21/21 passing** (7 per runtime × 3 runtimes).

## Files changed

| File | Change |
|------|--------|
| `packages/next/src/server/app-render/use-flight-response.tsx` | Core: `computeInlineScriptIntegrity()` + updated stream functions |
| `packages/next/src/server/app-render/stream-ops.node.ts` | Core: `computeInlineScriptIntegritySync()` + updated node stream functions |
| `packages/next/src/server/app-render/stream-ops.web.ts` | Updated web stream function signatures |
| `packages/next/src/server/app-render/app-render.tsx` | Threaded `sriAlgorithm` to all inline data stream creation sites |
| `packages/next/src/server/app-render/types.ts` | Added `sriAlgorithm?` to `RenderOptsPartial.experimental` |
| `packages/next/src/server/base-server.ts` | Added `sriAlgorithm` to `this.renderOpts.experimental` |
| `packages/next/src/build/templates/app-page.ts` | **Key fix**: Added `sriAlgorithm` to build template's `experimental` |
| `packages/next/src/build/templates/edge-ssr-app.ts` | **Key fix**: Added `sriAlgorithm` to edge build template's `experimental` |
| `test/production/app-dir/subresource-integrity/subresource-integrity.test.ts` | New test for inline script integrity |

## Challenges encountered during development

1. **`require('crypto')` breaks Edge/ESM builds** — The Node.js `crypto` module cannot be `require()`'d in `use-flight-response.tsx` because it is bundled for Edge/ESM paths where `crypto` does not exist. Fixed by using the Web Crypto API exclusively.

2. **Optional `sriAlgorithm?` type** — Initially added `sriAlgorithm` as required, which caused TypeScript errors in `app-page.ts`, `edge-ssr-app.ts`, and `export/index.ts` where not all experimental fields were being passed. Fixed by making it optional (`sriAlgorithm?`).

3. **`base-server.ts` is a dead path for app pages** — The most time-consuming issue. `base-server.ts:602` correctly sets `sriAlgorithm` on `this.renderOpts.experimental`, but in production, the `app-page.ts` build template constructs its own `renderOpts` from `nextConfig` directly, never reading `base-server.ts`'s value. Same for `edge-ssr-app.ts` and edge runtime pages. This required deep tracing through the entire request pipeline: `NextNodeServer.renderPageComponent()` → `base-server.ts:renderPageComponent()` → `renderToResponseWithComponents()` → `renderHTML()` → `lazyRenderAppPage()` → `handler()` in the build template.

4. **Edge runtime `crypto.subtle.digest` algorithm format** — The Web Crypto API requires `SHA-256` format while SRI uses `sha256`. The edge runtime threw `Unrecognized algorithm name` on the first pass. Fixed with a mapping regex.

5. **`app-route.ts` and `export/index.ts` are not needed** — App routes (API routes) do not render HTML with inline Flight scripts, and `export/index.ts` is used for static export which does not generate inline scripts at request time. These do not need `sriAlgorithm`.

### DIFF
--- a/packages/next/src/build/templates/app-page.ts
+++ b/packages/next/src/build/templates/app-page.ts
@@ -937,6 +937,7 @@ export async function handler(
             maxPostponedStateSizeBytes: parseMaxPostponedStateSize(
               nextConfig.experimental.maxPostponedStateSize
             ),
+            sriAlgorithm: nextConfig.experimental.sri?.algorithm,
             exposeTestingApi,
           },
 
--- a/packages/next/src/build/templates/edge-ssr-app.ts
+++ b/packages/next/src/build/templates/edge-ssr-app.ts
@@ -187,6 +187,7 @@ async function requestHandler(
         maxPostponedStateSizeBytes: parseMaxPostponedStateSize(
           nextConfig.experimental.maxPostponedStateSize
         ),
+        sriAlgorithm: nextConfig.experimental.sri?.algorithm,
         exposeTestingApi:
           pageRouteModule.isDev === true ||
           nextConfig.experimental.exposeTestingApiInProductionBuild === true,
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -3267,6 +3267,7 @@ async function renderToStream(
   } = renderOpts
 
   const { cachedNavigations, appShells } = renderOpts.experimental
+  const { sriAlgorithm } = renderOpts.experimental
 
   const { ServerInsertedHTMLProvider, renderServerInsertedHTML } =
     createServerInsertedHTML()
@@ -3775,7 +3776,8 @@ async function renderToStream(
             const inlinedDataStream = createNodeInlinedDataStream(
               reactServerResult.tee(),
               nonce,
-              formState
+              formState,
+              sriAlgorithm
             )
 
             // End the span since there's no async rendering in this path
@@ -3829,7 +3831,8 @@ async function renderToStream(
               inlinedDataStream: createNodeInlinedDataStream(
                 reactServerResult.consume(),
                 nonce,
-                formState
+                formState,
+                sriAlgorithm
               ),
               getServerInsertedHTML,
               getServerInsertedMetadata,
@@ -3898,7 +3901,8 @@ async function renderToStream(
           inlinedDataStream: createNodeInlinedDataStream(
             reactServerResult.consume(),
             nonce,
-            formState
+            formState,
+            sriAlgorithm
           ),
           isStaticGeneration: generateStaticHTML,
           allReady,
@@ -3919,7 +3923,8 @@ async function renderToStream(
             const inlinedDataStream = createWebInlinedDataStream(
               reactServerResult.tee(),
               nonce,
-              formState
+              formState,
+              sriAlgorithm
             )
 
             // End the span since there's no async rendering in this path
@@ -3973,7 +3978,8 @@ async function renderToStream(
               inlinedDataStream: createWebInlinedDataStream(
                 reactServerResult.consume(),
                 nonce,
-                formState
+                formState,
+                sriAlgorithm
               ),
               getServerInsertedHTML,
               getServerInsertedMetadata,
@@ -4036,7 +4042,8 @@ async function renderToStream(
           inlinedDataStream: createWebInlinedDataStream(
             reactServerResult.consume(),
             nonce,
-            formState
+            formState,
+            sriAlgorithm
           ),
           isStaticGeneration: generateStaticHTML,
           allReady,
@@ -4189,7 +4196,8 @@ async function renderToStream(
               // render
               reactServerResult.consume(),
               nonce,
-              formState
+              formState,
+              sriAlgorithm
             ),
             isStaticGeneration: generateStaticHTML,
             deploymentId: ctx.sharedContext.deploymentId,
@@ -4287,7 +4295,8 @@ async function renderToStream(
               // render
               reactServerResult.consume(),
               nonce,
-              formState
+              formState,
+              sriAlgorithm
             ),
             isStaticGeneration: generateStaticHTML,
             deploymentId: ctx.sharedContext.deploymentId,
@@ -7619,7 +7628,8 @@ async function continueStaticPrerenderWithInlinedData(
   renderFlightStream: typeof renderToWebFlightStream,
   clientModules: Parameters<typeof renderToWebFlightStream>[2],
   filterStackFrameForError: typeof filterStackFrame,
-  serverComponentsErrorHandler: (err: unknown) => string | undefined
+  serverComponentsErrorHandler: (err: unknown) => string | undefined,
+  sriAlgorithm?: import('../../build/webpack/plugins/subresource-integrity-plugin').SubresourceIntegrityAlgorithm
 ): Promise<AnyStream> {
   const hasFallbackRouteParams =
     fallbackRouteParams && fallbackRouteParams.size > 0
@@ -7650,7 +7660,8 @@ async function continueStaticPrerenderWithInlinedData(
     const inlinedDataStream = createInlinedDataStream(
       emptyReactServerResult.consumeAsStream(),
       nonce,
-      formState
+      formState,
+      sriAlgorithm
     )
     return continueStaticFallbackPrerender(htmlStream, {
       inlinedDataStream,
@@ -7663,7 +7674,8 @@ async function continueStaticPrerenderWithInlinedData(
   const inlinedDataStream = createInlinedDataStream(
     reactServerResult.consumeAsStream(),
     nonce,
-    formState
+    formState,
+    sriAlgorithm
   )
   return continueStaticPrerender(htmlStream, {
     inlinedDataStream,
@@ -7712,6 +7724,7 @@ async function prerenderToStream(
   } = renderOpts
 
   const { cachedNavigations, appShells } = renderOpts.experimental
+  const { sriAlgorithm } = renderOpts.experimental
 
   const renderFlightStream = process.env.__NEXT_USE_NODE_STREAMS
     ? renderToNodeFlightStream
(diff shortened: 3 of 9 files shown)


ashusnapx:
cc @timneutkens @unstubbable @gnoff @eps1lon @lubieowoce @ztanner — touching the core app-render pipeline, SRI, flight streaming, and build templates. Would appreciate a review!

ashusnapx:
Good catch! Fixed — moved `writeInitialInstructions` from `ReadableStream.start()` (synchronous, fire-and-forget) into the first `pull()` call (async). This ensures:

1. The integrity hash computation (via `crypto.subtle.digest`) is properly awaited
2. Errors are caught and forwarded to the stream, not swallowed
3. The bootstrap script is always enqueued before any flight data scripts

All 21 tests still pass after this change.

AdzerKI:
Before this takes more reviewer time on the app-render pipeline: the premise in "Why" does not hold. CSP never consults `integrity` when deciding whether an inline `<script>` may execute — inline scripts are admitted only by `'unsafe-inline'`, a matching nonce, or a `hash-source` in the policy matching the script's content. `integrity` is Subresource Integrity and applies to fetched subresources; an inline script fetches nothing.

So with this change applied, the repro in #95354 still fails under `script-src 'self'`: the tags gain an attribute the browser ignores for the CSP check, and hydration stays blocked.

The hashing work here is not wasted, though — the digest is computed in the right place, it just needs a different destination. I wrote up the split in https://github.com/vercel/next.js/issues/95354#issuecomment-5316006663: for streaming SSR only a nonce can work, while for prerendered routes the digests are knowable at build time and belong in the manifest `experimental.sri` already emits, so the app can compose `script-src 'self' 'sha384-…'` for those routes.

PR #95716, opened 2026-07-12
[turbopack] Drop unused exports from a CJS module

sampoder:
This PR introduces two new things: a pass of the analyzer that recognises exports from CJS modules (in the form of `exports.foo`, see `analyze_cjs_exports`) and then a CodeGen that reads `export_usage_info` to remove CJS exports that are not used.

This PR does not add anything additionally to `export_usage_info`, so this means that we only trim from modules that are imported with `import { foo } from "bar"`. The CodeGen depends on the output of `analyze_cjs_exports` for the locations of exports, whether it is safe to drop exports (its pretty conservative), and also whether `__esModule` is present.


turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
```diff
@@ -0,0 +1,251 @@
...
+/// - the `__esModule` interop marker (`exports.__esModule = true` / `module.exports.__esModule =
+///   true`)
+///
+/// Anything else touching `exports` / `module` — aliasing, escapes, computed
+/// keys, `module.exports = …` reassignment, top-level `this`, or an export write
+/// in any other position (nested in a function, behind `if`, inside another
+/// expression) — sets `is_unsafe`.
+///
+/// TODO (@sampoder): handle `Object.defineProperty(exports, "__esModule", { value: true })`
+pub fn analyze_cjs_exports(program: &Program, unresolved_mark: Mark) -> CommonJsExportsAnalysis {
```
mischnic: why does this need to be a separate AST pass?

sampoder: I can try moving it into a pre-existing one. Might be a little less clean but it will probably run a lot faster. I also was thinking we might want to put this behind a flag so this method would make that easier. Is there a pass you had in mind that I could combine it with?

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -277,7 +277,7 @@ fn prop_is(prop: &MemberProp, name: &str) -> bool {
 /// binding — e.g. the real CommonJS `module`/`exports`/`require`, not a local
 /// shadow. Prevents `let exports = {}; exports.foo = 'a'` from being treated as
 /// a write to the global `exports`.
-fn is_global(identifier: &Ident, name: &str, unresolved_mark: Mark) -> bool {
+pub(crate) fn is_global(identifier: &Ident, name: &str, unresolved_mark: Mark) -> bool {
```
lukesandberg: these helpers should probably just move to a new module.... `cjs_ast.rs`?

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/code_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+#[derive(
+    PartialEq, Eq, TraceRawVcs, ValueDebugFormat, NonLocalValue, Hash, Debug, Encode, Decode,
+)]
+struct CjsExportDrop {
+    name: RcStr,
+    /// Source position (`Stmt::span().lo`) of the `exports.NAME = …` statement,
+    /// used to locate it in the module body at code-gen time.
+    span_lo: u32,
```
lukesandberg: typically we capture an AST path, is that possible in your traversal

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/code_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+    ) -> Result<CodeGeneration> {
+        let export_usage_info = chunking_context
+            .module_export_usage(*ResolvedVc::upcast(module))
+            .await?;
+        let export_usage_info = export_usage_info.export_usage.await?;
+
+        // `All` means the module is (or may be) consumed wholesale — e.g. via
+        // `require()` or a namespace import — so nothing can be dropped.
+        if matches!(*export_usage_info, ModuleExportUsageInfo::All) {
+            return Ok(CodeGeneration::empty());
```
lukesandberg: maybe this is more efficient... but this would fall out of the `is_export_used` checks below so you could just drop this special case

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/code_gen.rs
```diff
@@ -0,0 +1,148 @@
...
+/// Blanks the top-level `exports.NAME = …` statements whose source position is in
+/// `unused`, replacing each with an empty statement.
+struct DropExportWrites {
+    unused: FxHashSet<BytePos>,
+}
+
+impl DropExportWrites {
+    fn drop_matching(&self, stmt: &mut Stmt) {
+        if self.unused.contains(&stmt.span().lo) {
+            *stmt = Stmt::Empty(EmptyStmt { span: DUMMY_SP });
```
lukesandberg: i don't think we can drop the whole statement

`exports.foo = bar()` should become `bar()`

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
```diff
@@ -0,0 +1,251 @@
...
+#[derive(Debug, Default)]
+pub struct CommonJsExportsAnalysis {
+    /// Statically known export names, in source order.
+    pub names: FxIndexSet<RcStr>,
+    /// Whether the transpiled-ESM `__esModule` marker is set.
+    pub has_es_module: bool,
+    /// Whether the module touches `exports` / `module` outside the recognized
+    /// safe forms. When set, `names` may be incomplete and the module must be
+    /// treated as opaque CommonJS (no exports dropped).
+    pub is_unsafe: bool,
```
lukesandberg: `has_dynamic_exports` would be a better name

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
```diff
@@ -0,0 +1,251 @@
...
+        if !analysis.names.insert(name.clone()) {
+            analysis.is_unsafe = true;
+            break;
+        }
+        // The RHS may still leak the exports object (`exports.a = [exports]`);
+        // the write's own target is the recognized form itself.
+        assign.right.visit_with(&mut visitor);
+        // Only side-effect-free values are droppable: removing the statement
+        // removes the evaluation of its RHS, so an impure RHS must be kept.
+        if !assign.right.may_have_side_effects(expr_ctx) {
```
lukesandberg: i don't think we need this condition as long as our rewrite preserves the LHS

we can let SWC decide the expression is deletable

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
lukesandberg: the natural place would be do it as part of the main `Analyzer` and then use the Effects system to report a droppable effect for every export

or really follow the pattern we have for ESM items: see add_esm_module_item

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
```diff
@@ -0,0 +1,251 @@
...
+        n.visit_children_with(self);
+    }
+
+    fn visit_constructor(&mut self, n: &swc_core::ecma::ast::Constructor) {
+        self.fn_depth += 1;
+        n.visit_children_with(self);
+        self.fn_depth -= 1;
+    }
+
+    fn visit_this_expr(&mut self, _: &swc_core::ecma::ast::ThisExpr) {
```
lukesandberg: this should be a reference to a `this` that is not rebound by another function

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
sampoder: moved it into there!

turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
sampoder: moved them

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/code_gen.rs
sampoder: sgtm!

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/code_gen.rs
sampoder: done!

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
sampoder: yep this is true

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/recognize.rs
sampoder: fixed by moving into the main analyzer

turbopack/crates/turbopack-ecmascript/src/references/cjs_exports/code_gen.rs
sampoder: this is being done now!

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/mod.rs
```diff
@@ -70,6 +71,12 @@ pub fn create_graph<'a>(
         supports_block_scoping,
     };
 
+    // CommonJS export recognition runs only for a CommonJS module that emits code
+    // (deny-by-default; see the analyzer's `CjsExportsCollector`).
+    if analyze_mode.is_code_gen() && eval_context.is_cjs(specified_module_type) {
```
lukesandberg: these conditions make sense, but we shoudl also have a nextjs experimental option to control it

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -498,6 +522,59 @@ mod analyzer_state {
...
+
+        /// Records the `exports.__esModule = true` interop marker.
+        pub(super) fn set_cjs_has_es_module(&mut self) {
+            if let Some(c) = &mut self.state.cjs_exports {
+                c.has_es_module = true;
+            }
+        }
+
+        /// Records a recognized `exports.NAME = …` write. A duplicate name taints
+        /// the module (ambiguous which write to remove).
```
lukesandberg: why is it ambiguous which one to remove...

exports.foo = 1
exports.foo = 2

just remove both?

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -498,6 +522,59 @@ mod analyzer_state {
...
+        pub(in crate::analyzer::graph) fn enable_cjs_exports(&mut self) {
+            self.state.cjs_exports = Some(CjsExportsCollector::default());
+        }
+
+        pub(super) fn cjs_exports_enabled(&self) -> bool {
+            self.state.cjs_exports.is_some()
+        }
+
+        pub(super) fn taint_cjs_exports(&mut self) {
+            if let Some(c) = &mut self.state.cjs_exports {
```
lukesandberg: instead of storing a boolean, you could just rewrite the `Option` to `None` at his point. disabling tracking

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1092,6 +1169,42 @@ impl<'a> Analyzer<'a, '_> {
...
+        }
+        let AssignTarget::Simple(SimpleAssignTarget::Member(member)) = &n.left else {
+            return;
+        };
+        let MemberProp::Ident(name) = &member.prop else {
+            return;
+        };
+
+        if name.sym.as_ref() == "__esModule" {
+            // Only a literal `true` is the interop marker.
```
lukesandberg, line 1227:
```suggestion
        // Transpilers use this to signal that this is transpiled esm.
        // Which in turn changes the behavior of default exports. such that `import foo  from 'transpiled-esm-cjs' gets the `default` export instead of the namespace
        if name.sym.as_ref() == "__esModule" {
            // Only a literal `true` is the interop marker.
```

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1092,6 +1169,42 @@ impl<'a> Analyzer<'a, '_> {
...
+        &mut self,
+        n: &AssignExpr,
+        ast_path: &AstNodePath<AstParentNodeRef<'_>>,
+    ) {
+        if n.op != AssignOp::Assign {
+            return;
+        }
+        // Only top-level writes can be dropped.
+        if self.is_in_fn() || self.is_in_nested_block_scope() {
+            return;
```
lukesandberg, line 1214: if it is `exports.foo = ...` inside a function should we taint?

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1891,7 +2020,10 @@ impl VisitAstPath for Analyzer<'_, '_> {
                 var: atom!("this"),
                 ast_path: as_parent_path_in(self.arena, ast_path),
                 span: node.span(),
-            })
+            });
+            // In CommonJS, top-level `this` aliases `exports`, so the module can
+            // reach its exports through it — leave it opaque.
+            self.taint_cjs_exports();
```
lukesandberg, line 2099: should we expand the `is_named_cjs_export_target` to match an unbound `this`?

should be simple enough?  then we leave this here to taint if that doesn't match

turbopack/crates/turbopack-tests/tests/execution/turbopack/cjs-remove-unused-exports/named-import/input/index.js
```diff
@@ -0,0 +1,11 @@
+import { used } from './lib.js'
+
+it('keeps a used named CommonJS export', () => {
+  expect(used).toBe('used-value')
+})
+
+it('preserves the side effects of an unused-but-impure export write', () => {
+  // `exports.impure` is never imported, but its RHS has a side effect, so the
+  // write must not be dropped.
+  expect(globalThis.__cjs_impure_ran).toBe(true)
```
lukesandberg, line 10: using __webpack_exports_info__ you might be able to check that the `impure` export actually got removed

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: :sob: yeah - this is an old version of the feature that made this a pain to do. we can just remove both, I'm going to get rid of this.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: yes - going to add this.

turbopack/crates/turbopack-tests/tests/execution/turbopack/cjs-remove-unused-exports/named-import/input/index.js
sampoder: `webpack_exports_info` only supports ESM :/

Screenshot 2026-07-15 at 12 40 58 am

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: added this!

turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -524,6 +524,18 @@ impl ImportMap {
...
+        if self.has_exports {
+            return false;
+        }
+
+        match specified_type {
+            SpecifiedModuleType::CommonJs => true,
+            SpecifiedModuleType::EcmaScript => false,
+            SpecifiedModuleType::Automatic => !self.has_imports && !self.has_top_level_await,
+        }
+    }
```
mischnic, line 529: isn't this basically `is_cjs = !is_esm()`?
which would make sense because ESM has a clear definition to detect. everything else is varying levels of CJS

turbopack/crates/turbopack-tests/tests/execution/turbopack/cjs-remove-unused-exports/named-import/input/index.js
mischnic: you will be able to also support CJS in there with https://github.com/vercel/next.js/pull/95826

turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
sampoder: yeah - I thought this was a little more readable but I can switch it back to that

turbopack/crates/turbopack-tests/tests/execution/turbopack/cjs-remove-unused-exports/named-import/input/index.js
sampoder: oh yes this is true!

crates/next-core/src/next_config.rs
```diff
@@ -1409,6 +1409,8 @@ pub struct ExperimentalConfig {
     turbopack_remove_unused_exports: Option<bool>,
```
lukesandberg, line 1409: should we split across 2 just like esm

e.g. name the new one `turbopack_cjs_remove_unused_export_assignments` for brutal clarity

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -86,6 +91,18 @@ pub(super) struct Analyzer<'arena, 'eval> {
...
 }
 
+/// Collects a static CommonJS module's droppable named exports during the main
+/// analyzer walk. Any `exports` / `module` use that isn't a recognized
+/// `exports.NAME = …` write taints the module by dropping the collector,
+/// leaving the module opaque.
+#[derive(Default)]
+struct CjsExportsCollector {
+    /// Recognized `exports.NAME = …` writes, each removable if `NAME` is unused.
+    writes: Vec<CjsExportDrop>,
```
lukesandberg: `Drop` reads a little odd in the name  maybe just `CjsExportAssignment`? or `DroppableCjsExportAssignment`

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1885,13 +2055,25 @@ impl VisitAstPath for Analyzer<'_, '_> {
...
                 span: node.span(),
-            })
+            });
+            if !self.in_cjs_export_target() {
+                self.taint_cjs_exports();
+            }
+        } else if self.in_cjs_export_value() && self.this_binding_depth() == 1 {
+            // `this` at the top level of an exported function value is `exports` when the
+            // function is called as a method of the exports (`require(m).fn()`), so the
+            // exports escape — bail out.
```
lukesandberg:
```suggestion
            // `this` at the top level of an exported function value may be `exports` when the
            // function is called as a method of the exports (`require(m).fn()`), so the
```

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1885,13 +2055,25 @@ impl VisitAstPath for Analyzer<'_, '_> {
...
             self.add_effect(Effect::FreeVar {
                 var: atom!("this"),
                 ast_path: as_parent_path_in(self.arena, ast_path),
                 span: node.span(),
-            })
+            });
+            if !self.in_cjs_export_target() {
+                self.taint_cjs_exports();
+            }
+        } else if self.in_cjs_export_value() && self.this_binding_depth() == 1 {
```
lukesandberg, line 2074: i guess if i had `function foo() {... this ...}` and then `exports.foo = foo` would we miss it because we wouldn't set the `in_cjs_export_value` bit

but this goes on and on
```
const a = {b() { ...this ....}
exports.foo = a.b
```

so... ouch

i think we could do it using the linker to infer more information about the value assigned to the export.... we could track a bit on JsValue::Function about whether or not it references `this`?

but that seems like overkill

do you know how do other bundlers handle this?

turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
mischnic: the question is whether we want to have a state of "neither CJS nor ESM"
Otherwise, this seems like code duplication/a place where you can have a slight divergence in behavior 

e.g. because the condition in `SpecifiedModuleType::Automatic =>` is slightly different (but again, question is whether that's intentional)

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: Webpack does not attempt to handle this, for example, with:

```javascript
// module evil
function evil() {
  if (this.bar === undefined) throw new Error();
}

exports.foo = evil

exports.bar = ''


// module B
const a = require("a")

a.foo();
```

Screenshot 2026-07-17 at 2 10 51 pm

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: Going with `DroppableCjsExportAssignment` even though it is a mouthful

turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
sampoder: they should be the exact same, the condition in `SpecifiedModuleType::Automatic` is just because of:

```rust
if self.has_exports {
    return false;
}
```

I don't think we have a state of neither CJS nor ESM

crates/next-core/src/next_config.rs
sampoder: Going to go with one as both the new flag + `turbopack_remove_unused_exports` / `turbopack_remove_unused_imports` have to be true for this to do anything

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
lukesandberg: discussed IRL/slack

this is broadly considered to be both unfixable and not worth fixing

evil code is evil and breeds its own destruction.

lukesandberg (approved):

PR #95717, opened 2026-07-12
[turbopack] Track usage of modules imported with `require()`

sampoder:
This PR introduces a new method `analyze_require_usage` that analyzes how something imported with `require()` is used. It has two passes: one catches simple things like `const { a } = require(...)` and `require(...).foo`, the next one then handles if you import using `const x = require(...)`.

This then gets populated into `export_usage_info`, this will expand the set of exports we can drop through https://github.com/vercel/next.js/pull/95716.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/mod.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/mod.rs
@@ -1,8 +1,10 @@
 use rustc_hash::FxHashMap;
 use swc_core::{
     atoms::Atom,
+    common::BytePos,
     ecma::{ast::*, visit::VisitWithAstPath},
 };
+use turbopack_core::resolve::ExportUsage;
 
 pub use crate::analyzer::graph::{
     effects::{
@@ -18,7 +20,7 @@ use crate::{
 
 mod effects;
 mod eval_context;
-mod visitor;
+pub(crate) mod visitor;
 
 #[derive(Debug)]
 pub struct VarGraph<'a> {
@@ -33,6 +35,10 @@ pub struct VarGraph<'a> {
     pub effects: Vec<Effect<'a>>,
     // Some unconditional codegens, usually for ESM items.
     pub code_gens: Vec<CodeGen>,
+
+    /// [`ExportUsage`] per `require("…")` call, keyed by call position; absent
+    /// calls fall back to `ExportUsage::All`.
+    pub require_usage: FxHashMap<BytePos, ExportUsage>,
 }
 
 impl<'a> VarGraph<'a> {
@@ -63,6 +69,7 @@ pub fn create_graph<'a>(
             free_var_ids: Default::default(),
             effects: Default::default(),
             code_gens: Default::default(),
+            require_usage: Default::default(),
         },
         eval_context,
         state: Default::default(),
@@ -77,6 +84,10 @@ pub fn create_graph<'a>(
     if cjs_tree_shaking && analyze_mode.is_code_gen() && eval_context.is_cjs(specified_module_type)
     {
         analyzer.enable_cjs_exports();
+    }
+
+    if cjs_tree_shaking && analyze_mode.is_code_gen() {
+        analyzer.enable_require_usage(&eval_context.imports);
     }
 
     m.visit_with_ast_path(&mut analyzer, &mut Default::default());
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
@@ -4,9 +4,10 @@ use std::{
 };
 
 use bumpalo::boxed::Box as BumpBox;
+use rustc_hash::FxHashMap;
 use smallvec::SmallVec;
 use swc_core::{
-    common::{Span, Spanned, SyntaxContext, pass::AstNodePath},
+    common::{BytePos, Span, Spanned, SyntaxContext, pass::AstNodePath},
     ecma::{
         ast::*,
         atoms::atom,
@@ -20,7 +21,7 @@ use turbopack_core::resolve::ExportUsage;
 use crate::{
     AnalyzeMode,
     analyzer::{
-        Bump, BumpVec, ConstantValue, JsValue, WellKnownFunctionKind,
+        Bump, BumpVec, ConstantValue, ImportMap, JsValue, WellKnownFunctionKind,
         cjs_ast::{
             as_exports_define_property, define_property_sets_es_module, is_exports_object,
             is_global,
@@ -34,7 +35,7 @@ use crate::{
         cjs::{CjsExportsDropCodeGen, DroppableCjsExportAssignment},
         esm::EsmModuleItem,
     },
-    utils::{AstPathRange, unparen},
+    utils::{AstPathRange, extract_name_from_member_prop, extract_names_from_object_pat, unparen},
 };
 
 enum EarlyReturn<'a> {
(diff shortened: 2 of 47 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -624,7 +624,7 @@ fn extract_names_from_then_callback(call: &CallExpr) -> Option<SmallVec<[RcStr;
     }
 }
 
-fn extract_name_from_member_prop(prop: &MemberProp) -> Option<SmallVec<[RcStr; 1]>> {
+pub(crate) fn extract_name_from_member_prop(prop: &MemberProp) -> Option<SmallVec<[RcStr; 1]>> {
```
lukesandberg: nit: consider a new home?

sampoder: moved into turbopack/crates/turbopack-ecmascript/src/utils.rs!

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/mod.rs
```diff
@@ -77,6 +84,10 @@ pub fn create_graph<'a>(
         analyzer.enable_cjs_exports();
     }
 
+    if analyze_mode.is_code_gen() {
```
lukesandberg: need a configuration bit `turbopackTreeShakeCjsRequires` to control

turbopack/crates/turbopack-ecmascript/src/analyzer/require_usage.rs
```diff
@@ -0,0 +1,156 @@
...
+    let Expr::Call(call) = unparen(expr) else {
+        return None;
+    };
+    let Callee::Expr(callee) = &call.callee else {
+        return None;
+    };
+    let Expr::Ident(f) = &**callee else {
+        return None;
+    };
+    if !is_global(f, "require", unresolved_mark) {
```
lukesandberg: i think this is 'fine' but you can alias require

const myrequire = require;

myrequire('foo').bar

the analyzer supports this (will detect the dependency on `foo`)  and so i guess we won't have any 'require usage' data for it? as long as that failure fails open (retain all of 'foo') then it is fine.

turbopack/crates/turbopack-ecmascript/src/analyzer/require_usage.rs
```diff
@@ -0,0 +1,156 @@
...
+    };
+    if arg.spread.is_some() || !matches!(unparen(&arg.expr), Expr::Lit(Lit::Str(_))) {
+        return None;
+    }
+    Some(call)
+}
+
+/// For each `const x = require(...)` binding, scans every use of `x` and records
+/// the resulting [`ExportUsage`] under the call's position in `resolved`.
+pub(crate) fn resolve_namespace_bindings(
```
lukesandberg: for esm don't we do this when processing member effects? or is it also a separate traversal?

sampoder: with ESM, you're able to figure out what is used from the import statement eg. `import { foo } from "bar"`. if i added this second pass because I'm trying to do `const x = require(...)` and then finding `x.foo`. 

```javascript
function bar() {
  return x.foo
}
const x = require("foo")
bar()
```
problem here is we'd hit `x.foo` before we know that it `x` is tied to a require. now I could keep track of the usage of every variable to avoid the second pass? but that felt kinda crazy.

turbopack/crates/turbopack-ecmascript/src/analyzer/require_usage.rs
sampoder: yeah, nothing gets drop with smth like this.

made a test like this:

```javascript
const myRequire = require
console.log(myRequire('./lib.js').used)
```

the module is:

```javascript
exports.used = 'used-value'
exports.unused = 'unused-value'
```

and it becomes:

```
(globalThis["TURBOPACK"] || (globalThis["TURBOPACK"] = [])).push(["output/1jsg_tests_snapshot_cjs-remove-unused-exports_require-alias_input_1le4zip._.js",
"[project]/turbopack/crates/turbopack-tests/tests/snapshot/cjs-remove-unused-exports/require-alias/input/index.js [test] (ecmascript)", ((__turbopack_context__, module, exports) => {

// `require` is aliased, so the require-usage analysis can't recognize the call.
// No usage is recorded for it, which must fail open: lib.js keeps *all* its
// exports (including `unused`) rather than being narrowed to just `.used`.
const myRequire = /*TURBOPACK member replacement*/ __turbopack_context__.t;
console.log(__turbopack_context__.r("[project]/turbopack/crates/turbopack-tests/tests/snapshot/cjs-remove-unused-exports/require-alias/input/lib.js [test] (ecmascript)").used);
}),
"[project]/turbopack/crates/turbopack-tests/tests/snapshot/cjs-remove-unused-exports/require-alias/input/lib.js [test] (ecmascript)", ((__turbopack_context__, module, exports) => {

exports.used = 'used-value';
exports.unused = 'unused-value';
}),
]);
```

beautiful am I right. `unused` is still there though.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/mod.rs
```diff
@@ -33,6 +35,10 @@ pub struct VarGraph<'a> {
     pub effects: Vec<Effect<'a>>,
     // Some unconditional codegens, usually for ESM items.
     pub code_gens: Vec<CodeGen>,
+
+    /// [`ExportUsage`] per `require("…")` call, keyed by call position; absent
+    /// calls fall back to `ExportUsage::All`.
+    pub require_usage: FxHashMap<BytePos, ExportUsage>,
```
mischnic, line 41: Can't we do this in the `ImportMap` visitor?  Then we would have one visitor that does all import/export collection

sampoder: yes! this is actually perfect because it means we don't add any extra traversals. moved it in there and then moved the traversal that figures out how `x` in `const x = require(...)` is used into the main analyzer

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -2000,6 +2045,29 @@ impl VisitAstPath for Analyzer<'_, '_> {
             }
         }
 
+        // How a `const x = require(...)` binding is consumed: a static member read
+        // `x.foo` observes that export; anything else observes the whole namespace.
+        if self.require_usage_enabled() {
```
lukesandberg: we can probably remove the explicit check for require_usage_enabled since `is_tracked_require_binding` will just always return false

ideally we would also make this more parallel with the import path below

actually should we be using Effects?

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -2416,6 +2484,7 @@ impl VisitAstPath for Analyzer<'_, '_> {
         node: &'ast ExportDecl,
         ast_path: &mut swc_core::ecma::visit::AstNodePath<'r>,
     ) {
+        self.escape_exported_require_bindings(node);
```
lukesandberg, line 2560: this is for `const x = require(...); .... export x`?

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: talked about using effects irl, we don't need to as we aren't doing rewrites. shared some code with the import path below + dropped `require_usage_enabled()`

lukesandberg (approved):

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: yep!

PR #95718, opened 2026-07-12
[turbopack] Drop dead `require()` calls

sampoder:
This PR adds support for `turbopackRemoveUnusedImports` in CJS modules. If modules are side-effect free and their imports are unused we do the following re-writes:

```
require('./pure')          →   0;
console.log('x')               console.log('x')
```

```
const x = require('./pure')   →   const x = 0
```

```
const dead = require('./pure'), keep = 'v'   →   const dead = 0, keep = 'v'
```

```
const {} = require('./pure')   →   const {} = 0
```

```
if (cond) require('./pure')   →   if (cond) 0; 
```

SWC should then drop the 0.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
@@ -943,6 +943,8 @@ impl Analyzer<'_> {
             }
             Pat::Object(_) => {
                 let usage = match extract_names_from_object_pat(&n.name) {
+                    // `const {} = require(...)`: no members read → evaluation only.
+                    Some(names) if names.is_empty() => ExportUsage::Evaluation,
                     Some(names) => ExportUsage::PartialNamespaceObject(names),
                     None => ExportUsage::All,
                 };
@@ -1452,6 +1454,17 @@ impl Visit for Analyzer<'_> {
 
     fn visit_var_declarator(&mut self, node: &VarDeclarator) {
         self.record_require_usage_var(node);
+        node.visit_children_with(self);
+    }
+
+    fn visit_expr_stmt(&mut self, node: &ExprStmt) {
+        // A bare `require("…")` statement discards its result → evaluation only.
+        if let Some(call) = as_require_call(&node.expr, self.unresolved_mark) {
+            self.data
+                .cjs_imports
+                .resolved
+                .insert(call.span.lo, ExportUsage::Evaluation);
+        }
         node.visit_children_with(self);
     }
 
--- a/turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
@@ -102,6 +102,7 @@ pub struct CjsRequireAssetReference {
     chunking_type_attribute: Option<SpecifiedChunkingType>,
     resolve_override: Option<ResolvedVc<Box<dyn Module>>>,
     usage: ExportUsage,
+    cjs_tree_shaking: bool,
 }
 
 impl CjsRequireAssetReference {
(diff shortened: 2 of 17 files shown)


turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -195,6 +195,20 @@ impl CjsRequireAssetReferenceCodeGen {
...
+        // This `require()` is in `unused_references`: its result is unused and
+        // the target is side-effect free. Replace the call with a placeholder
+        // at the expression level.
+        if chunking_context
+            .unused_references()
+            .contains_key(&ResolvedVc::upcast(self.reference))
+            .await?
+        {
+            let visitor = create_visitor!(self.path, visit_mut_expr, |expr: &mut Expr| {
+                *expr = quote!("0" as Expr);
```
lukesandberg, line 215:
```suggestion
                // `const {a,b} = 0;` is a clever way to assign undefined to all the variables
                *expr = quote!("0" as Expr);
```

turbopack/crates/turbopack-tests/tests/execution/turbopack/cjs-remove-unused-imports/sibling-declarator/input/pure.js
```diff
@@ -0,0 +1,3 @@
+'use turbopack no side effects'
```
lukesandberg: do we not infer this as side-effect free anyway?

lukesandberg:
as discussed in office hours, we should steal the webpack tests for treeshaking and commonjs

turbopack/crates/turbopack-tests/tests/execution/turbopack/cjs-remove-unused-imports/sibling-declarator/input/pure.js
sampoder: lol yes

sampoder:
> as discussed in office hours, we should steal the webpack tests for treeshaking and commonjs 

added in https://github.com/vercel/next.js/pull/95811 because there was a lot!

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -195,6 +198,22 @@ impl CjsRequireAssetReferenceCodeGen {
...
+        // at the expression level.
+        if reference.cjs_tree_shaking
+            && chunking_context
+                .unused_references()
+                .contains_key(&ResolvedVc::upcast(self.reference))
+                .await?
+        {
+            let visitor = create_visitor!(self.path, visit_mut_expr, |expr: &mut Expr| {
+                // `const {a,b} = 0;` is a clever way to assign undefined to all the variables
+                *expr = quote!("0" as Expr);
```
mischnic, line 215: Is this the only thing this PR does?
removing dead requires is certainly a necessary step
but the really important step is also not chunking the subgraph that belongs to that require
removing the require might save you a couple of bytes
but removing the subgraph can save you an unbounded number of bytes.

sampoder: They should be skipped because they flow through the same code that unused ESM imports flow through: https://github.com/vercel/next.js/blob/42ccb2eedea561f5fac9e2b26a9a5527e9d7a647/turbopack/crates/turbopack-core/src/module_graph/binding_usage_info.rs#L163

In the snapshot tests I have in this PR it appears like we aren't chunking the unused imports

turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -1423,6 +1425,17 @@ impl Visit for Analyzer<'_> {
         node.visit_children_with(self);
     }
 
+    fn visit_expr_stmt(&mut self, node: &ExprStmt) {
+        // A bare `require("…")` statement discards its result → evaluation only.
+        if let Some(call) = as_require_call(&node.expr, self.unresolved_mark) {
+            self.data
+                .cjs_imports
+                .resolved
+                .insert(call.span.lo, ExportUsage::Evaluation);
```
lukesandberg, line 1466: at some point i would like to start distinguishing `require` dependencies that are reachable from module evaluation vs not

this could allow us to trim more side effect edges

lukesandberg (approved):

PR #95721, opened 2026-07-12
fix(turbopack): prevent tree-shaking index out of bounds panic

### LINKED ISSUE #95698: Turbopack bundling issue (webpack behaviour is correct, turbopack does not tree shake correctly)

### To Reproduce

1. Run `pnpm install`.
2. Run `pnpm build`.

```bash
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 32054
  Available CPU cores: 24
Binaries:
  Node: 24.17.0
  npm: 11.18.0
  Yarn: N/A
  pnpm: 11.9.0
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 6.0.3
Next.js Config:
  output: N/A
```

Turbopack, Performance

### Additional context

SparshGarg999:
### What?
Prevent an index out-of-bounds compiler panic in Turbopack when `experimental.turbopackTreeShaking` is enabled.

### Why?
During module splitting in `split_module`, empty chunks (where the AST body was empty) were skipped to optimize bundle structure. However, the graph analysis still registered and mapped these indices in graph structures (`groups.graph_ix`). When Turbopack later resolved module parts or dependencies, it indexed into the `modules` array using the group indices. Since empty modules were skipped, the index was out of bounds, triggering a compiler panic (`index out of bounds: the len is X but the index is Y`).

### How?
Always push the generated chunk (even if empty) to the `modules` vector so that the 1-to-1 index mapping is preserved.

Fixes #95698


sampoder (changes requested):
Hi there, `turbopackTreeShaking` is still (very) experimental. Turbopack's tree-shaking is mainly `turbopackRemoveUnusedExports` and `turbopackRemoveUnusedImports` which default to true. `turbopackTreeShaking` is an alternative version of tree-shaking that unfortunately has performance issues. I think it'd be best to focus efforts on improving `turbopackRemoveUnusedExports` and `turbopackRemoveUnusedImports` to tree-shake the module you referenced.

PR #95723, opened 2026-07-12
docs: clarify client cache behavior for dynamic routes

### LINKED ISSUE #74012: Docs: Contradiction in caching

In the image in section https://nextjs.org/docs/app/building-your-application/caching#static-and-dynamic-rendering,
It shows that for dynamic routes during subsequent visits the page will come from router cache, but this is not the case, by default dynamic pages are not cached by router cache. 

see above

_No response_

SparshGarg999:
### What?
Update prefetching, glossary, and `staleTimes` config documentation to accurately state the default Client Cache behavior for dynamic routes in Next.js 15+.

### Why?
By default, dynamic pages are not cached in the Client Cache (the default stale time is 0 seconds). Some diagrams or older documentation might show that dynamic pages are cached on subsequent visits, which is contradictory.

### How?
Added warning/clarification notes to `prefetching.mdx`, `staleTimes.mdx`, and `glossary.mdx` stating that dynamic pages are not cached by default on subsequent client-side visits.

Fixes #74012

### DIFF
--- a/packages/next/src/server/lib/trace/tracer.ts
+++ b/packages/next/src/server/lib/trace/tracer.ts
@@ -242,6 +242,19 @@ class NextTracerImpl implements NextTracer {
     return trace.getSpan(context?.active())
   }
 
+  /**
+   * Run `fn` with the active span cleared, so spans created inside become new
+   * roots (or parent to incoming propagated context). Used for in-process
+   * Node.js middleware so its span is a sibling of the request span, matching
+   * edge middleware which runs in a detached sandbox.
+   */
+  public runWithDetachedContext<T>(fn: () => T): T {
+    if (!NEXT_OTEL_PERFORMANCE_PREFIX && !this.isTracingEnabled()) {
+      return fn()
+    }
+    return context.with(ROOT_CONTEXT, fn)
+  }
+
   public withPropagatedContext<T, C>(
     carrier: C,
     fn: () => T,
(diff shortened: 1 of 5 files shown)


icyJoseph:
I agree with the behavior being described, but this PR does not fix the actual issue.

#74012 is about the caching diagram implying that a subsequent navigation to a dynamic route is served from the Client Cache. That's because the diagram was made when the staleTime value for dynamic pages was 30 seconds, rather than zero.

Adding repeated notes to `prefetching.mdx`, `staleTimes.mdx`, and `glossary.mdx` feels a bit broad and partially redundant, especially since `staleTimes.mdx` already states that `dynamic` defaults to `0 seconds (not cached)`.

SparshGarg999:
Hi @icyJoseph,

Thank you for your feedback!

After reviewing the current Next.js `canary`/latest documentation, we found that:
1. The caching documentation has been completely restructured around the new **Cache Components** model and `use cache` (e.g. `docs/01-app/01-getting-started/08-caching.mdx`).
2. The legacy caching guide that explained the 4 caching mechanisms (Request Memoization, Data Cache, Full Route Cache, Router Cache) and included the caching diagram showing the 30-second staleTime for dynamic routes has been removed from the latest codebase (it is now replaced by `caching-without-cache-components.mdx`, which does not contain any images or diagrams).
3. The image/diagram asset files themselves are hosted on Vercel's assets CDN and their sources are in Vercel's private repository (e.g. `vercel/next-site`), rather than in the `vercel/next.js` repository.

Since the old caching diagram containing the contradiction is no longer present in the latest App Router documentation, the contradiction is obsolete in the latest docs. We have reverted the redundant textual notes from `prefetching.mdx`, `staleTimes.mdx`, and `glossary.mdx` to keep the codebase clean.

This PR can be closed now as the contradiction is resolved.

PR #88551, opened 2026-01-14
Turbopack: only include needed polyfills

mischnic:
Closes https://github.com/vercel/next.js/issues/86785
Closes https://github.com/vercel/next.js/pull/87270
Closes PACK-6133

For Turbpack, split up `polyfills-module` into separate files and include only what is necessary (and only load `polyfills-nomodule` if a `browserslist` that requires it was configured)

The output looks something like this: 
Bildschirmfoto 2026-01-14 um 18 35 18

PR #95726, opened 2026-07-13
Skip polyfill-module when browserslist targets support polyfilled APIs natively

### LINKED ISSUE #86785: Polyfills for unsupported browsers are enabled by default

https://codesandbox.io/p/devbox/lucid-kirch-qm6y6g?workspaceId=ws_YGJRKdEKng1kurcxFGU5DV

### To Reproduce

1. Start the application (in any mode)
2. Run the Lighthouse tool in Chrome DevTools
3. Note in the Insights section that there is a report about a number of Legacy JavaScript polyfills being sent

Image

Note: the replication attached is just an unchanged copy of the provided replication template repository. You can see this issue on all nextjs applications running any version since these polyfills were added.

The polyfills noted should only be included if the user has requested them. Even in that case it could be argued that the user should be the one providing polyfills needed to support a wider ranger of browser versions, as would be the case if they wanted to handle cases requiring additional polyfills on top of the ones listed.

This claim is based on reviewing the code for these polyfills over in [packages/next-polyfill-module/src/index.js](https://github.com/vercel/next.js/blob/canary/packages/next-polyfill-module/src/index.js) and noting that the ones that Lighthouse is warning about all require browser versions older than any item in the current list of supported browsers:

```json
{
  "browserslist": ["chrome 111", "edge 111", "firefox 111", "safari 16.4"]
}
```

This is making the assumption that the current recommendation for loading polyfills would allow for loading them early enough to ensure nothing breaks, but if that isn't the case then that is a problem in and of itself.

```bash
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 4102
  Available CPU cores: 2
Binaries:
  Node: 20.12.1
  npm: 10.5.0
  Yarn: 1.22.19
  pnpm: 8.15.6
Relevant Packages:
  next: 16.1.0-canary.12 // Latest available version is detected (16.1.0-canary.12).
  eslint-config-next: N/A
  react: 19.2.1
  react-dom: 19.2.1
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Performance, Turbopack, Webpack

### Additional context

Note that of the polyfills currently in `next-polyfill-module` the only one that should be required by the current supported browsers is also the only one that Lighthouse doesn't mention, which is `URL.canParse`.

This is taking into account that a few of the comments stating `Edge: never` don't match the values on the linked caniuse.com pages.

I imagine that this will continue being a problem in the future as other polyfills go out of date, but I won't suggest at this point that substantial effort be made to have the build configuration checking the `browserslist` when choosing which polyfills to include rather than the current solution building them in.

See https://github.com/vercel/next.js/discussions/64330 for examples of people noticing this issue in production.

islamhafez0:
## What?

When all browsers in the project's browserslist already support the APIs shipped by `@next/polyfill-module`, skip bundling the polyfill module entirely. This saves ~14 KiB from the client bundle and silences the Lighthouse "Legacy JavaScript" audit.

Fixes #86785

## How?

- Add `needsPolyfill()` utility (`packages/next/src/build/polyfills/needs-polyfill.ts`) that checks browserslist targets against known minimum browser versions for each polyfill shipped by `@next/polyfill-module` (String.trimStart/trimEnd, Symbol.description, Array.flat/flatMap, Promise.finally, Object.fromEntries, Array.at, Object.hasOwn, URL.canParse).
- Use `NormalModuleReplacementPlugin` to replace the polyfill module with a noop when polyfills aren't needed. This approach matches on the resolved resource path, so it works regardless of whether the import uses a relative path (`../build/polyfills/polyfill-module`) or a module-style path.
- Also skip the `@next/polyfill-nomodule` bundle when polyfills aren't needed, since that bundle is only for browsers without ES module support (IE 11).

## Why `NormalModuleReplacementPlugin` instead of `resolve.alias`?

The polyfill-module is imported via a relative path (`import '../build/polyfills/polyfill-module'`) in the client entry files. `resolve.alias` matches against the import request string, so an alias key like `'next/dist/build/polyfills/polyfill-module'` may not reliably match the relative import. `NormalModuleReplacementPlugin` hooks into both `beforeResolve` and `afterResolve`, matching against both the request string and the resolved resource path.

## Testing

- Added unit tests for `needsPolyfill()` covering edge cases (modern browsers, legacy browsers, missing browsers, mixed browser lists).
- The existing production test suite (`test/production/polyfills/`) and browserslist tests (`test/production/css-features/browserslist.test.ts`) provide additional coverage.

### DIFF
--- a/packages/next/src/build/polyfills/needs-polyfill.ts
+++ b/packages/next/src/build/polyfills/needs-polyfill.ts
@@ -1,0 +1,81 @@
+/**
+ * Determines whether the project's browserslist targets need any of the
+ * module-level polyfills shipped by `@next/polyfill-module`.
+ *
+ * Each entry maps a polyfill to the **minimum** browser versions that
+ * support it natively (sourced from the inline comments in
+ * `packages/next-polyfill-module/src/index.js`).
+ *
+ * If every resolved browser is at or above the minimum for every
+ * polyfill, the caller can safely skip bundling the polyfill module —
+ * saving ~14 KiB from the client bundle and silencing the Lighthouse
+ * "Legacy JavaScript" audit.
+ */
+
+type BrowserFamily = 'chrome' | 'firefox' | 'safari' | 'edge'
+
+// Minimum (major) version that supports each polyfill, per browser family.
+// "0" means "never supported natively" — the polyfill is always needed
+// for that browser family.
+const POLYFILL_MIN_VERSIONS: Record<
+  string,
+  Partial<Record<BrowserFamily, number>>
+> = {
+  'string-trimstart': { chrome: 66, firefox: 61, safari: 12, edge: 79 },
+  'string-trimend': { chrome: 66, firefox: 61, safari: 12, edge: 79 },
+  'symbol-description': { chrome: 70, firefox: 63, safari: 12, edge: 79 },
+  'array-flat': { chrome: 69, firefox: 62, safari: 12, edge: 79 },
+  'array-flatmap': { chrome: 69, firefox: 62, safari: 12, edge: 79 },
+  'promise-finally': { chrome: 63, firefox: 58, safari: 11, edge: 15 },
+  'object-fromentries': { chrome: 73, firefox: 63, safari: 12, edge: 79 },
+  'array-at': { chrome: 92, firefox: 90, safari: 15, edge: 92 },
+  'object-hasown': { chrome: 93, firefox: 92, safari: 15, edge: 93 },
+  'url-canparse': { chrome: 120, firefox: 115, safari: 17, edge: 120 },
+}
+
+/**
+ * Parse a browserslist entry like `"chrome 131.0.0"` into
+ * `{ family: "chrome", major: 131 }`.
+ */
+function parseBrowserslistEntry(
+  entry: string
+): { family: BrowserFamily; major: number } | null {
+  // Browserslist output format: "family version"
+  // e.g. "chrome 131.0.0", "firefox 133.0", "safari 18.2"
+  const match = entry.match(
+    /^(chrome|firefox|safari|edge)\s+(\d+)/i
+  )
+  if (!match) return null
+  return {
+    family: match[1].toLowerCase() as BrowserFamily,
+    major: parseInt(match[2], 10),
+  }
+}
+
+/**
+ * Returns `true` when at least one polyfill is needed for the given
+ * browserslist targets, `false` when all targets support the APIs
+ * natively and the polyfill module can be skipped.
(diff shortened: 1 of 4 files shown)


packages/next/src/build/polyfills/needs-polyfill.ts
```diff
@@ -0,0 +1,81 @@
...
+  'string-trimend': { chrome: 66, firefox: 61, safari: 12, edge: 79 },
+  'symbol-description': { chrome: 70, firefox: 63, safari: 12, edge: 79 },
+  'array-flat': { chrome: 69, firefox: 62, safari: 12, edge: 79 },
+  'array-flatmap': { chrome: 69, firefox: 62, safari: 12, edge: 79 },
+  'promise-finally': { chrome: 63, firefox: 58, safari: 11, edge: 15 },
+  'object-fromentries': { chrome: 73, firefox: 63, safari: 12, edge: 79 },
+  'array-at': { chrome: 92, firefox: 90, safari: 15, edge: 92 },
+  'object-hasown': { chrome: 93, firefox: 92, safari: 15, edge: 93 },
+  'url-canparse': { chrome: 120, firefox: 115, safari: 17, edge: 120 },
+}
```
SukkaW, line 34: This is not true. On paper, Next.js still allows custom browserlist, which means there could be `op_mini` (Opera Mini), `ie`, and even more browser types, your code just ignore all those browsers and may skip vital polyfills.

packages/next/src/build/polyfills/needs-polyfill.ts
```diff
@@ -0,0 +1,81 @@
...
+ * `{ family: "chrome", major: 131 }`.
+ */
+function parseBrowserslistEntry(
+  entry: string
+): { family: BrowserFamily; major: number } | null {
+  // Browserslist output format: "family version"
+  // e.g. "chrome 131.0.0", "firefox 133.0", "safari 18.2"
+  const match = entry.match(
+    /^(chrome|firefox|safari|edge)\s+(\d+)/i
+  )
```
SukkaW, line 47: This is not the proper way to parse browserlist.

`not chrome < 130` is also a valid browserlist.
`not dead` is also a valid browserlist.
`last 2 safari versions` is also a valid browserlist.
`safari > 18` is also a valid browserlist.
`not firefox esr` is also a valid browserlist.

Just try to support them, and every time you even try to do that, I will give you another list of edge cases. Good luck catching up.

SukkaW:
Also, I noticed the https://github.com/vercel/next.js/pull/88551, and IMHO that's the proper way forward.

PR #95732, opened 2026-07-13
Turbopack: loader change invalidation

mischnic:
We apparently never invalidated when the loader source code changed. This now uses the same mechanism we already use for watching postcss/tailwind config files

This also required changing the tracing mode of configs (and now loaders) to only care about imports, not FS file references to prevent infinite loops.


mischnic:
* **#95732** <a href="https://app.graphite.com/github/pr/vercel/next.js/95732?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95732?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

sokra (approved):

turbopack/crates/turbopack-node/src/transforms/webpack.rs
```diff
@@ -222,6 +226,52 @@ async fn webpack_loaders_executor(
...
+/// the previous (non-invalidating) behavior instead of breaking the build.
+#[turbo_tasks::function]
+async fn loaders_changed(
+    loaders: Vc<WebpackLoaderItems>,
+    project_path: FileSystemPath,
+    asset_context: Vc<Box<dyn AssetContext>>,
+    resolve_options_context: Vc<ResolveOptionsContext>,
+) -> Result<Vc<Completion>> {
+    let options = resolve_options(project_path.clone(), resolve_options_context);
+    let origin_path = project_path.clone().join("_")?;
```
mischnic, line 245: We should be resolving the loader to a `FileSystemPath` at the very start and not handing around un-anchored `loader: RcStr` where we don't know where to resolve from.

packages/next/src/build/swc/index.ts
```diff
@@ -947,6 +947,49 @@ function bindingToApi(
...
+        } else {
+          return {
+            ...rule,
+            loaders: rule.loaders?.map(
+              (v) => transformRule(v) as TurbopackLoaderItem
+            ),
+          }
+        }
+      }
+
```
mischnic: I had to add this

mischnic:
I  had to change quite a lot. please take another look @sokra

turbopack/crates/turbopack-node/src/transforms/webpack.rs
```diff
@@ -222,6 +226,52 @@ async fn webpack_loaders_executor(
...
+        .map(async |loader| {
+            let result = asset_context
+                .resolve_asset(
+                    origin_path.clone(),
+                    Request::parse(Pattern::Constant(loader.loader.clone())),
+                    options,
+                    ReferenceType::Loader,
+                )
+                .await?;
+            let Some(module) = result.first_module().await? else {
```
sokra: This might resolve to multiple modules (e.g. due to unspecified export conditions). Better loop over all modules.

sokra (approved):

packages/next/src/build/swc/index.ts
```diff
@@ -1172,20 +1173,37 @@ function bindingToApi(
...
+          ...loaderItem,
+          loader: serializeLoader(loaderItem.loader),
+        }
       }
     }
+
+    // Webpack loader specifiers can be absolute paths, we need it to be relative for turbopack.
+    function serializeLoader(loader: string) {
+      return path.isAbsolute(loader)
+        ? normalizePathOnWindows('./' + path.relative(projectPath, loader))
```
sokra, line 1204: `path.relative` can return an absolute path too, e.g. when the `loader` is on a different drive on windows

sokra (approved):

PR #95743, opened 2026-07-13
Fix OpenTelemetry API singleton in server bundles

bryan-hunter:
## Summary

- Externalize `@opentelemetry/api` from server bundles so route code and instrumentation use the same process-wide singleton.
- Add production direct-entrypoint coverage for a tracer acquired at module scope before instrumentation registration.
- Sync the documented default `serverExternalPackages` list.

## Problem

A production route can acquire its tracer when the route module is evaluated:

```ts
// route.ts — evaluated first
import { trace } from '@opentelemetry/api'

const tracer = trace.getTracer('route')
```

Instrumentation registers the provider later:

```ts
// instrumentation.ts
import { trace } from '@opentelemetry/api'

trace.setGlobalTracerProvider(provider)
```

This ordering is supported by OpenTelemetry's proxy API. It broke when the route bundle contained API copy A while instrumentation loaded API copy B: registering the provider on B never activated the tracer from A, so custom spans stayed no-op.

Externalizing `@opentelemetry/api` makes both imports resolve to the same singleton, allowing an early tracer to delegate to the provider after registration.

## Reproduction

The E2E fixture reproduces the production sequence: preload the generated route, acquire its module-scope tracer, register OpenTelemetry, then invoke the route.

```bash
pnpm test-start-turbo test/e2e/opentelemetry/instrumentation/opentelemetry.test.ts \
  -t "records spans from a tracer acquired before instrumentation registration"
```

- Without the new external-package entry: the route and loader receive different API objects, the request returns 500, and no custom span is exported.
- With the entry: the API objects match, the request returns 200, and the collector receives `module-scope-tracer`.

## Observed versions

| Next.js | Production observation |
| --- | --- |
| `16.1.6` | Known good: 223 sampled application turn spans alongside 3,979 AI SDK spans |
| `16.2.10` | Affected: 0 application turn spans while 342 AI SDK spans confirmed traffic and OTel ingestion continued |

These sampled counts establish presence versus absence, not exact traffic volume. The application upgrade also changed its instrumentation integration, so the production comparison identifies the observed boundary but does not isolate Next.js by itself; the E2E negative control above isolates the server bundling behavior.

Target: canary, with a backport to the maintained `16.2.x` line.

## Testing

- Production Turbopack direct-entrypoint OpenTelemetry tests: 9 passed
- `pnpm --filter=next types`
- `pnpm validate-externals-doc`

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/serverExternalPackages.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/serverExternalPackages.mdx
@@ -31,6 +31,7 @@ Next.js includes a [short list of popular packages](https://github.com/vercel/ne
 - `@mikro-orm/knex`
 - `@node-rs/argon2`
 - `@node-rs/bcrypt`
+- `@opentelemetry/api`
 - `@prisma/client`
 - `@react-pdf/renderer`
 - `@sentry/profiling-node`
--- a/docs/02-pages/04-api-reference/04-config/01-next-config-js/serverExternalPackages.mdx
+++ b/docs/02-pages/04-api-reference/04-config/01-next-config-js/serverExternalPackages.mdx
@@ -31,6 +31,7 @@ Next.js includes a [short list of popular packages](https://github.com/vercel/ne
 - `@mikro-orm/knex`
 - `@node-rs/argon2`
 - `@node-rs/bcrypt`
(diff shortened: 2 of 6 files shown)


bryan-hunter:
this seems to solve it, so will close!

https://github.com/vercel/next.js/pull/94306

PR #95746, opened 2026-07-13
fix(turbopack): correct tree-shaking for re-exports (fixes #95698)

### LINKED ISSUE #95698: Turbopack bundling issue (webpack behaviour is correct, turbopack does not tree shake correctly)

Harshalj2108:
Closes #95698

This PR resolves an issue where Turbopack failed to properly tree-shake libraries that heavily rely on star re-exports and named re-exports (such as `viem` and `wagmi`), resulting in severe bundle bloat (e.g., bundling all 712 chains instead of just the 1 used chain).

### Root Cause
There were two main issues in how Turbopack assigned `ImportUsage` for re-exports during binding-usage analysis:
1. **Star re-exports** (`export * from '...'`) were defaulting to `ImportUsage::TopLevel`, which tells the optimizer that the target module must be unconditionally evaluated, effectively opting the whole branch out of tree-shaking.
2. **Named re-exports** (`export { named } from './module.js'`) were being modeled exclusively as export bindings rather than local declarations. Because `ProgramDeclUsage` only tracks local identifier bindings, these re-exports lacked `ImportUsage` metadata entirely and consequently fell back to `ImportUsage::TopLevel` as well.

### Changes Made
- **`turbopack_core`**: Added a new `ImportUsage::StarReexport` variant.
- **`binding_usage_info.rs`**: Added logic to handle `ImportUsage::StarReexport` during graph traversal. It now correctly propagates only the specifically-used exports (via `ExportUsage::PartialNamespaceObject`) to the target module, or skips the edge entirely if the parent has no used exports.
- **`references/exports.rs`**: Updated reference generation to identify star re-export namespaces and assign them `ImportUsage::StarReexport`.
- **`analyzer/imports.rs`**: Modified the import analyzer to explicitly collect `Export::ImportedBinding` and `Export::ImportedNamespace` names and map them as `ImportUsage::Exports` for their respective import references.

### Testing
- Validated via `cargo test -p turbopack-tests -- tree-shaking` (all 30 execution and snapshot tree-shaking tests pass successfully).
- Verified `next-swc` native build integrity.

### DIFF
--- a/turbopack/crates/turbopack-core/src/module_graph/binding_usage_info.rs
+++ b/turbopack/crates/turbopack-core/src/module_graph/binding_usage_info.rs
@@ -156,6 +156,7 @@ pub async fn compute_binding_usage_info(
                     return Ok(GraphTraversalAction::Continue);
                 };
 
+                let mut propagated_export_usage = ref_data.binding_usage.export.clone();
                 if remove_unused_imports {
                     // If this is an evaluation reference and the target has no side effects
                     // then we can drop it. NOTE: many `imports` create parallel Evaluation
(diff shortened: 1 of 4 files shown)


sampoder (commented):
Thanks for the PR! However, I think to solve this issue we will go with a different format for a fix: https://github.com/vercel/next.js/pull/95989

PR #95749, opened 2026-07-13
[turbopack] Switch `make_production_chunks` to use floats

sampoder:
We've talked about making this switch, making this switch because without it the diff for https://github.com/vercel/next.js/pull/95157 was very difficult to read and follow.

### DIFF
--- a/turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
+++ b/turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
@@ -234,14 +234,15 @@ pub async fn make_production_chunks(
                     .min(1_000_000) as i64;
 
                 // Default `P(N = 1)`: the probability that we request exactly 1 chunk group.
-                // An integer percentage (0..=100). `firstPageLoadPriority` maps directly to
-                // it; the default is 67 (~2/3).
-                let default_p1 = first_page_load_priority.map_or(67, |percent| percent as i64);
+                // `firstPageLoadPriority` (a config percentage) maps to it; the default is 0.67
+                // (~2/3).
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
```diff
@@ -234,14 +234,15 @@ pub async fn make_production_chunks(
                     .min(1_000_000) as i64;
 
                 // Default `P(N = 1)`: the probability that we request exactly 1 chunk group.
-                // An integer percentage (0..=100). `firstPageLoadPriority` maps directly to
-                // it; the default is 67 (~2/3).
-                let default_p1 = first_page_load_priority.map_or(67, |percent| percent as i64);
+                // `firstPageLoadPriority` (a config percentage) maps to it; the default is 0.67
+                // (~2/3).
+                let default_p1 =
+                    first_page_load_priority.map_or(0.67, |percent| percent as f64 / 100.0);
```
bgw, line 240: Now that this is internally a float, should we modify the `next.config.js` to accept a floating point proportion (`0..=1`) instead of an integer percent? We still have time to fix this before this goes into a real release and people start depending on it...

bgw (approved):
Much nicer!

turbopack/crates/turbopack-core/src/chunk/chunking/production.rs
sampoder: luckily this is already a floating point in the next config! sadly as it gets put in a turbo task value it has to be an integer for that (for `Eq` and `Hash` I think)

PR #95753, opened 2026-07-14
Better support the CLI spinner when running the TSC CLI

lukesandberg:
Currently if you enable `useTscCli` we render a `Running Typescript ...` but then never remove it.  This is simply because our normal approach of pausing/resuming the spinner doesn't work with the subprocess. 

Instead run the spinner in the normal way and stop it as soon as the subprocess  produces any output.  TSC produces all output in a small sequence of writes at the end, so this works as expected.  If it changes to start streaming data this will also work but the spinner will never restart, but this is fine.

### DIFF
--- a/packages/next/src/build/type-check.ts
+++ b/packages/next/src/build/type-check.ts
@@ -151,10 +151,10 @@ export async function startTypeChecking({
           pagesDir,
           debugBuildPaths,
           useTypeScriptCli,
-          // Stop the spinner before the CLI checker writes its diagnostics
-          // (which it does directly to stdout/stderr, bypassing the console
-          // hooks that would otherwise pause the spinner).
-          () => typeCheckingSpinner?.stop()
+          // Stop the spinner before as soon as the subprocess reports output.
+          useTypeScriptCli && typeCheckingSpinner
+            ? () => typeCheckingSpinner.stop()
+            : undefined
         ).then((resolved) => {
           const checkEnd = process.hrtime(typeCheckAndLintStart)
           return [resolved, checkEnd] as const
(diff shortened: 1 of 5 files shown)


lukesandberg:
* **#95754** <a href="https://app.graphite.com/github/pr/vercel/next.js/95754?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#95753** <a href="https://app.graphite.com/github/pr/vercel/next.js/95753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95753?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/lib/typescript/runTypeScriptCli.ts
```diff
@@ -106,24 +116,47 @@ export function runTypeScriptCli({
...
+      child.stderr?.on('data', (chunk: string) => {
         stderr += chunk
       })
+    } else {
+      let sawOutput = false
+      const forward = (dest: NodeJS.WriteStream, chunk: string) => {
+        if (!sawOutput) {
+          sawOutput = true
+          onFirstOutput?.()
+        }
```
lukesandberg: instead of a boolean we should just reassign onFirstOutput to undefined

packages/next/src/build/type-check.ts
```diff
@@ -155,7 +150,11 @@ export async function startTypeChecking({
           appDir,
           pagesDir,
           debugBuildPaths,
-          useTypeScriptCli
+          useTypeScriptCli,
+          // Stop the spinner before the CLI checker writes its diagnostics
+          // (which it does directly to stdout/stderr, bypassing the console
+          // hooks that would otherwise pause the spinner).
+          () => typeCheckingSpinner?.stop()
```
lukesandberg: instead of this we should pass `undefined` if we don't have a spinner

bgw (approved):

Trustinbtc999-hue (commented):
[]()

PR #95773, opened 2026-07-14
Add OpenTelemetry span for Server Action execution

timneutkens:
## What?

Adds an OpenTelemetry span around Server Action execution, including the resolved function name and source filename in development.

The new `AppRender.executeServerAction` span measures only the application function invocation. It is emitted through the standard Next.js tracer and is available anywhere OpenTelemetry tracing is enabled, independently of Request Insights. In development it resolves the action through the Server Actions manifest and records a readable function name and normalized source filename.

## Why?

Server Action execution was previously uninstrumented application work inside an action request. This made action latency invisible to OpenTelemetry users and made the span difficult to investigate without knowing which application function ran.

Keeping the complete span instrumentation in its own PR separates OpenTelemetry concerns from the Request Insights snapshot and UI handling added by the next PR in the stack.

## How?

- Adds `AppRender.executeServerAction` to the framework span constants and vanilla OpenTelemetry allowlist.
- Resolves development-only action metadata from the Server Actions manifest after validating the action through the existing production validation path.
- Wraps the actual `action.apply()` invocation while preserving the existing request async-storage context.
- Records a human-readable action label, application category, normalized action name, and source filename in development.
- Records genuine action failures as errors while treating redirects and HTTP access fallbacks as successful control flow.
- Adds development coverage for the resolved inline function and filename, plus production coverage for the generic span.

## Stack

1. **#95773 — Server Action OpenTelemetry span (this PR)**
2. #95770 — Server Action metadata in Request Insights
3. #95765 — Missing OpenTelemetry spans
4. #95766 — Human-readable OpenTelemetry span names

## Verification

- `pnpm --filter=next build`
- `pnpm test-dev-turbo test/e2e/opentelemetry/instrumentation/opentelemetry.test.ts -t "should trace Server Action execution"`
- `pnpm test-start-turbo test/e2e/opentelemetry/instrumentation/opentelemetry.test.ts -t "should trace Server Action execution"`

### DIFF
--- a/packages/next/src/server/app-render/action-handler.ts
+++ b/packages/next/src/server/app-render/action-handler.ts
@@ -1522,48 +1522,33 @@ export function getValidatedMPAActionId(
   return actionId
 }
 
-const ACTION_DESCRIPTOR_ID_PREFIX = '{"id":"'
-function isInvalidStringActionDescriptor(
-  actionDescriptor: string,
-  serverModuleMap: ServerModuleMap
-): unknown {
-  if (actionDescriptor.startsWith(ACTION_DESCRIPTOR_ID_PREFIX) === false) {
-    return true
-  }
-
-  const from = ACTION_DESCRIPTOR_ID_PREFIX.length
-  const to = from + ACTION_ID_EXPECTED_LENGTH
-
-  // We expect actionDescriptor to be '{"id":"<actionId>",...}'
-  const actionId = actionDescriptor.slice(from, to)
-  if (
-    actionId.length !== ACTION_ID_EXPECTED_LENGTH ||
-    actionDescriptor[to] !== '"'
-  ) {
-    return true
-  }
-
-  const entry = serverModuleMap[actionId]
-
-  if (entry == null) {
-    return true
-  }
-
-  return false
-}
-
 function getActionIdFromStringDescriptor(
   actionDescriptor: string,
   serverModuleMap: ServerModuleMap
 ): string | null {
-  if (isInvalidStringActionDescriptor(actionDescriptor, serverModuleMap)) {
+  let parsedActionDescriptor: unknown
+  try {
+    // React also parses this descriptor as JSON before reading its `id`. In
+    // particular, this makes the last of any duplicate `id` properties win.
+    parsedActionDescriptor = JSON.parse(actionDescriptor)
+  } catch {
     return null
   }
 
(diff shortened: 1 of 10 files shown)


test/e2e/opentelemetry/instrumentation/opentelemetry.test.ts
```diff
@@ -88,6 +88,8 @@ describe.each(
 
   // Edge runtime is currently not implemented in custom-entrypoint-server.ts
   const itEdge = useDirectEntrypointHandler ? it.skip : it
+  // This fixture is not wired into custom-entrypoint-server.ts.
+  const itServerAction = useDirectEntrypointHandler ? it.skip : it
```
mischnic: Add that route to the regex matcher list in custom-entrypoint-server.ts and it should just work

timneutkens: :robot: Done — added the Server Action route to custom-entrypoint-server.ts and enabled both Server Action tracing tests for direct entrypoints. The focused production/Turbopack run passes both tests.

packages/next/src/server/app-render/action-handler.ts
```diff
@@ -788,7 +830,11 @@ export async function handleAction({
             } else {
               // Multipart POST, but not a fetch action.
               // Potentially an MPA action, we have to try decoding it to check.
-              if (areAllActionIdsValid(formData, serverModuleMap) === false) {
+              const mpaActionId = getValidatedMPAActionId(
```
gnoff, line 833: This works for MPA actions because they by definition need to be singular but any given request can (in the future) contain many actions. Additionally there are forms like action composition where you pass one action into another HoC style and only the outermost action would show up.

packages/next/src/server/lib/trace/constants.ts
```diff
@@ -89,6 +89,7 @@ enum AppRenderSpan {
   fetch = 'AppRender.fetch',
   waitShellReady = 'AppRender.waitShellReady',
   renderToNodeFizzStream = 'AppRender.renderToNodeFizzStream',
+  executeServerAction = 'AppRender.executeServerAction',
```
gnoff, line 92: We might want to name it so it makes sense even when server actions are batched. We'll then likely want an outer span for the whole batch and one per inner action execution

PR #95774, opened 2026-07-14
Micro-optimization for string format

jimmyhmiller:
I get this is a bit of a silly change in some ways. It isn't going to have some massive impact. But using dhat I got these stats on a smallish sample nextjs project

| metric                  | A (parse.rs change) | B (baseline) | delta              |
| ------------------------ | -------------------- | ------------ | ------------------- |
| total bytes allocated    | 438.4 MiB            | 442.6 MiB    | -4.3 MiB (-0.94%)   |
| total allocation count   | 1,672,688            | 1,872,347    | -199,659 (-10.66%)

Eliminating ~200k allocations from this small change just felt worth it.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/parse.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/parse.rs
@@ -253,7 +253,7 @@ impl SourceMapGenConfig for InlineSourcesContentConfig {
         match f {
             FileName::Custom(s) => {
                 // format! here is suboptimal and allocates over and over again.
-                // On a random test next test project this one spot accounted for 
+                // On a random test next test project this one spot accounted for
                 // 10% of allocations, hence the more verbose approach.
                 let proto: &str = &SOURCE_URL_PROTOCOL;
                 let mut out = String::with_capacity(proto.len() + 3 + s.len());
--- a/turbopack/crates/turbopack-ecmascript/src/parse.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/parse.rs
@@ -252,7 +252,15 @@ impl SourceMapGenConfig for InlineSourcesContentConfig {
     fn file_name_to_source(&self, f: &FileName) -> String {
         match f {
             FileName::Custom(s) => {
-                format!("{SOURCE_URL_PROTOCOL}///{s}")
+                // format! here is suboptimal and allocates over and over again.
+                // On a random test next test project this one spot accounted for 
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -252,7 +252,15 @@ impl SourceMapGenConfig for InlineSourcesContentConfig {
     fn file_name_to_source(&self, f: &FileName) -> String {
         match f {
             FileName::Custom(s) => {
-                format!("{SOURCE_URL_PROTOCOL}///{s}")
```
jimmyhmiller, line 255: Yes I was surprised by format not doing the optimal thing here as well. My understanding is the builtin only tries to find size of the constant values.

turbopack/crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -252,7 +252,15 @@ impl SourceMapGenConfig for InlineSourcesContentConfig {
...
+                // format! here is suboptimal and allocates over and over again.
+                // On a random test next test project this one spot accounted for
+                // 10% of allocations, hence the more verbose approach.
+                let proto: &str = &SOURCE_URL_PROTOCOL;
```
lukesandberg: nit use `SOURCE_URL_PROTOCOL_STR` in all locations this is already a 'str' (and probably makes it possible to statically determine the length)

turbopack/crates/turbopack-ecmascript/src/parse.rs
lukesandberg: yeah and `SOURCE_URL_PROTOCOL` is a constant rcstr which probably doesn't have a const `len` so it doesn't work well i guess? im not sure where it fails

perhaps using SOURCE_URL_PROTOCOL_STR directly would fix this

lukesandberg (approved):

turbopack/crates/turbopack-ecmascript/src/parse.rs
jimmyhmiller: Good point, didn't notice that we had that split.

turbopack/crates/turbopack-ecmascript/src/parse.rs
bgw: `format!` has to go through the `Display` trait. There's no way for implementers of the `Display` trait to provide size hints. It may be possible that rustc has specialization for `&str`.

PR #95781, opened 2026-07-14
Turbopack: skip trimmed side-effect imports in SideEffectsModule

jimmyhmiller:
A star re-export barrel with imports but no top-level side effects is
classified ModuleEvaluationIsSideEffectFree. follow_reexports_with_side_effects
treats that as effectful and builds a SideEffectsModule that emits
TURBOPACK_IMPORT(<eval part id>), but compute_side_effect_free_module_info
classifies the same module as side-effect-free, so binding-usage prunes the
Evaluation edge and the eval part never gets a module id. The build then
panicked with 'ModuleId not found for ident: .../barrel.ts <internal part 0>'
during middleware chunking.

SideEffectsModule is designed so graph tree-shaking may trim these imports, so
the consumer must tolerate a trimmed side effect. Add
ModuleIdStrategy::try_get_id_from_module (returns None instead of erroring when
an explicit id map lacks the module) and have chunk_item_content skip side
effects with no id -- a pruned side-effect-free module has nothing to evaluate.

Adds an e2e regression test (treeshake-mw).

### DIFF
--- a/test/e2e/app-dir/treeshake-mw/app/layout.tsx
+++ b/test/e2e/app-dir/treeshake-mw/app/layout.tsx
@@ -1,0 +1,11 @@
+export default function RootLayout({
+  children,
+}: {
+  children: React.ReactNode
+}) {
+  return (
+    <html>
+      <body>{children}</body>
+    </html>
+  )
+}
--- a/test/e2e/app-dir/treeshake-mw/app/page.tsx
+++ b/test/e2e/app-dir/treeshake-mw/app/page.tsx
@@ -1,0 +1,3 @@
+export default function Page() {
+  return <h1>home</h1>
+}
--- a/test/e2e/app-dir/treeshake-mw/lib/barrel.ts
+++ b/test/e2e/app-dir/treeshake-mw/lib/barrel.ts
@@ -1,0 +1,6 @@
+// Star re-export barrel. Because `getKey` is not a static named entrypoint of
+// this module, resolving `import { getKey } from './lib/barrel'` goes through
+// `follow_reexports_with_side_effects`, which builds a `SideEffectsModule`.
+// This barrel has an import but no top-level side effects, so it is classified
+// `ModuleEvaluationIsSideEffectFree` — the status that triggers the desync.
+export * from './keys'
--- a/test/e2e/app-dir/treeshake-mw/lib/keys.ts
+++ b/test/e2e/app-dir/treeshake-mw/lib/keys.ts
@@ -1,0 +1,3 @@
+export function getKey() {
+  return 'k'
+}
--- a/test/e2e/app-dir/treeshake-mw/middleware.ts
+++ b/test/e2e/app-dir/treeshake-mw/middleware.ts
@@ -1,0 +1,14 @@
+import { NextResponse } from 'next/server'
+import { getKey } from './lib/barrel'
+
+export function middleware() {
+  const k = getKey()
+  const res = NextResponse.next()
+  res.headers.set('x-key', k)
+  return res
+}
+
+export const config = {
+  runtime: 'nodejs',
+  matcher: '/((?!_next/static|_next/image|favicon.ico).*)',
+}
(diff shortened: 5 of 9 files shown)


turbopack/crates/turbopack-ecmascript/src/tree_shake/side_effects/module.rs
```diff
@@ -149,7 +149,21 @@ impl EcmascriptChunkPlaceable for SideEffectsModule {
...
         let mut has_top_level_await = false;
 
+        let id_strategy = chunking_context.chunk_item_id_strategy().await?;
+
         for &side_effect in module.side_effects.iter() {
+            // These side effects were collected conservatively (see
+            // `follow_reexports_with_side_effects`). Inner-graph tree shaking may have
+            // since determined a "side effect" to be side-effect-free and pruned its
+            // evaluation edge, in which case it has no chunk item id.
+            // So we are going to skip it.
```
mischnic: The way this works with the existing non-splitting tree shaking already is different:

If some module was excluded, e.g. a `ModuleReference(ModulePart::SideEffects)` that points to a side-effect-free module, then `module_reference.resolve_reference()` would return "empty"
https://github.com/vercel/next.js/blob/b7ab553862a13a8a0d6de39d3467f86658031e30/turbopack/crates/turbopack/src/lib.rs#L191-L201

I think we shouldn't add `id_strategy.try_get_id_from_module` at all.

jimmyhmiller: @mischnic I think I'm not quite understanding. The module in question is not marked SideEffectFree it is marked ModuleSideEffects::ModuleEvaluationIsSideEffectFree.

```rust
#[derive(Hash, Debug, Copy, Clone)]
#[turbo_tasks::value(shared)]
pub enum ModuleSideEffects {
    /// Analysis determined that the module evaluation is free of side effects. The module may still
    /// have side effects based on its imports.
    ///
    /// This module might not be chunked after Turbopack performed a global analysis on the module
    /// graph.
    ModuleEvaluationIsSideEffectFree,
    /// Is known to be free of side effects either due to static analysis or some kind of
    /// configuration.
    ///
    /// ```js
    /// "use turbopack no side effects"
    /// ```
    ///
    /// This module might not even be parsed (and thus chunked) if no other module depends on any of
    /// its exports.
    SideEffectFree,
    // Neither of the above, so we should assume it has side effects.
    SideEffectful,
}
```

My best understanding of this is that ModuleEvaluationIsSideEffectFree is a conservative guess and may or may not be included in the final chunk. In this case, it was excluded, but we try to look it up anyways and fails. That's why I added this method

But I'm certain I don't understand this whole process, so if I just misunderstood the suggestion, let me know.

lukesandberg: the delta is simply about _when_ we know it is side effect free

we can sometimes figure out that modules are side effect free very early (based on a package.json annotation or based on "use turbopack no side effects" annotation)

`ModuleEvaluationIsSideEffectFree` means we couldn't find any _immediate_ side effects from statically analyzing the module, but it _might_ have side effects due to imports.  This can be 'upgraded' to SideEffectFree later on when we trim unused transitive imports.

(i.o.w. not during apply_module_type)

But yes we can remove the module later on which would make it so we don't compute a module id for it....

it would be better to query the global 'side effect free' modules set from `compute_side_effect_free_module_info` which can tell you the whole truth of the situation.

or really, we need to filter the iteration of `side_effects` through the unused references on the module graph.  That would be the more canonical way to do this

To put it another way

SideEffectsmodule::references returned some references that global analysis determined were dead, so we need to respect that when referencing them here

jimmyhmiller: @lukesandberg and @mischnic I *think* I understand now. Made a new commit that filters on unused. Is that correct?

mischnic: yes, exactly!

mischnic (approved):

turbopack/crates/turbopack-ecmascript/src/tree_shake/side_effects/module.rs
mischnic: (I didn't connect the dots that the core problem was that you were not skipping `unused_references` here, otherwise I would have said that directly of course)

PR #95786, opened 2026-07-14
[ci] Skip JS build when running all e2e deploy tests

eps1lon:
The `test-e2e-deploy-release` workflow ran `pnpm build` in all jobs even though deploy tests install the published `NEXT_TEST_VERSION` remotely on Vercel, so nothing that gets deployed uses the local build. The only local consumer of build output is the Jest test harness, which has load-time value imports from `next/dist` (e.g. `test/lib/e2e-utils/index.ts`, `test/lib/next-test-utils.ts`) and, transitively via `next/dist/server/next`, from `@next/env`'s dist.

Instead, we're now manually creating `next/dist` and `next-env/dist` from the configured Next.js version we're testing with. So this not only a perf fix but also correctness because tests didn't necessarily get the same `next` the deployed app was using.

## Test plan

- [ ] [e2e deploy release job from this branch](https://github.com/vercel/next.js/actions/runs/29360224144)

### DIFF
--- a/.github/workflows/build_reusable.yml
+++ b/.github/workflows/build_reusable.yml
@@ -11,6 +11,15 @@ on:
         required: false
         description: 'whether to skip pnpm install && pnpm build'
         type: string
+      hydrateNextDistVersion:
+        required: false
+        description: >
+          When set, skip `pnpm build` and populate packages/next/dist from the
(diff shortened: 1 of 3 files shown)


eps1lon:
Also need `@next/playwright`. I'll leave that as-is until we get Turborepo cache back.

PR #95789, opened 2026-07-14
Only track pending imports for real promises

gaearon:
Extracted from https://github.com/vercel/next.js/pull/95468.

Fixes a module with `then` export hanging the module tracking during the build.

---

`trackPendingImport` needs to track actually async modules (top-level await).

Before this change, this codepath was happy to adopt any thenable. But since thenables may not be Promises, this could break things downstream, hence why we had `Promise.resolve` there.

However, we don't *want* to adopt arbitrary thenables here. **Tracking is for top-level awaited modules.** For example, `module.exports = new Promise(() => {})` has a `.then` but we don't want to track that. It's not TLA.

This is why, to catch real top-level await modules, we narrow the check to `instanceof Promise`.

We add tests for:

- A client module with a `then` function export (**hangs** on Turbopack before this PR with CC on)
- Exporting a literal Promise (works before and after, but worth testing for future regressions)

### DIFF
--- a/packages/next/src/server/app-render/module-loading/track-module-loading.instance.ts
+++ b/packages/next/src/server/app-render/module-loading/track-module-loading.instance.ts
@@ -1,5 +1,4 @@
 import { CacheSignal } from '../cache-signal'
-import { isThenable } from '../../../shared/lib/is-thenable'
 
 /**
  * Tracks all in-flight async imports and chunk loads.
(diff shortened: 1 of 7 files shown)


packages/next/src/server/app-render/module-loading/track-module-loading.instance.ts
```diff
@@ -23,11 +22,8 @@ export function trackPendingImport(exportsOrPromise: unknown) {
 
   // requiring an async module returns a promise.
   // if it's sync, there's nothing to track.
-  if (isThenable(exportsOrPromise)) {
-    // A client reference proxy might look like a promise, but we can only call `.then()` on it, not e.g. `.finally()`.
```
timneutkens, line 27: This comment mentions `instanceof Promise` is not good enough 🤔

gaearon: I think what it's saying is — if we get a thenable here, passing it to tracking won't work because it'll crash on missing finally later. So the old code first turned thenables into Promises.

But we shouldn't be getting thenables through in the first place because this codepath is for tracking TLA async modules which are always real Promises. So we narrow the check to the actual case this codepath is for, and don't need to transform invalid input for code downstream.

timneutkens (approved):

PR #95795, opened 2026-07-14
Turbopack HMR: extract deferring build messages and use it in server hmr

wbinnssmith:
Churn in the graph can lead to a `building` message quickly followed by a `built` message. The underlying issue should be addressed, but for now, extract this debounce from the client HMR implementation and use it in server hmr.

### DIFF
--- a/packages/next/src/server/dev/hot-reloader-turbopack.ts
+++ b/packages/next/src/server/dev/hot-reloader-turbopack.ts
@@ -104,6 +104,7 @@ import { devIndicatorServerState } from './dev-indicator-server-state'
 import { getDisableDevIndicatorMiddleware } from '../../next-devtools/server/dev-indicator-middleware'
 import { getRestartDevServerMiddleware } from '../../next-devtools/server/restart-dev-server-middleware'
 import { backgroundLogCompilationEvents } from '../../shared/lib/turbopack/compilation-events'
+import { DeferredEmit } from '../../shared/lib/turbopack/deferred-emit'
 import { getSupportedBrowsers } from '../../build/get-supported-browsers'
 import { printBuildErrors } from '../../build/print-build-errors'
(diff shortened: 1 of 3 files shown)


packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -1837,10 +1846,13 @@ export async function createHotReloaderTurbopack(
   })
 
   async function handleProjectUpdates() {
+    const BUILDING_MESSAGE_DEFER_MS = 100
     for await (const updateMessage of project.updateInfoSubscribe(30)) {
       switch (updateMessage.updateType) {
         case 'start': {
-          hotReloader.send({ type: HMR_MESSAGE_SENT_TO_BROWSER.BUILDING })
+          pendingBuilding.schedule(BUILDING_MESSAGE_DEFER_MS, () => {
```
lukesandberg, line 1920: should the delay param be a constructor property?

lukesandberg (approved):

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -675,6 +676,12 @@ export async function createHotReloaderTurbopack(
   let hmrEventHappened = false
   let hmrHash = 0
 
+  // HACK: Defer sending `building` messages. Turbopack emits a compile pass for every
```
lukesandberg, line 716: is this really a 'hack'?

lukesandberg (approved):

packages/next/src/server/dev/hot-reloader-turbopack.ts
wbinnssmith: Kind of. I think it's something that should be kept around regardless but we're far chattier than we need to be when it comes to actually reporting changes (the thing that causes messages to be scheduled in the first place). I'll leave a comment/file a linear for the underlying problem.

PR #95802, opened 2026-07-15
docs: add authentication with Cache Components guide and iron-session example

icyJoseph:
- new guide, Auth w/ Cache Components
- iron-session example w/ Cache Components

TODO:

- [x] opt-out from the instant requirements
- [x] e2e for the example, using instant helper


docs/01-app/02-guides/authentication-with-cache-components.mdx
```diff
@@ -0,0 +1,299 @@
...
+
+## Migrating an existing app
+
+With Cache Components enabled, instant navigation validation flags every route that reads the session, because a request read can't be prerendered into the static shell. You don't have to resolve them all before shipping. Set [`export const instant = false`](/docs/app/guides/instant-navigation#opting-out) on the page or layout to let it keep blocking on the server, then adopt the patterns below one route at a time. For the full migration workflow, see [Migrating to Cache Components](/docs/app/guides/migrating-to-cache-components#following-validation).
+
+## Step 1: Read the current user
+
+Reading the current user reads the session cookie, then looks the user up. Because a request read can't be part of the static shell, it always sits behind a [`<Suspense>`](/docs/app/api-reference/file-conventions/loading) boundary, where it streams in on demand.
+
+That works, but it means streaming the section to the client on every navigation. Giving the read a cache lifetime lets the framework prefetch this UI ahead of time instead.
```
aurorascharff: well, even with a cache, wouldnt you still want to wrap it in suspense?

docs/01-app/02-guides/authentication.mdx
```diff
@@ -1132,6 +1134,8 @@ We recommend creating a DAL to centralize your data requests and authorization l
 
 The DAL should include a function that verifies the user's session as they interact with your application. At the very least, the function should check if the session is valid, then redirect or return the user information needed to make further requests.
```
aurorascharff, line 1135: We could potentially remove some of the stuff about auth and streaming from this guide and link out to the new CC auth gudie rather, since it covers it better, and that existing content was mainly for cc anyway

aurorascharff: And the auth provider oattern is properly described in this new guide!

docs/01-app/02-guides/authentication-with-cache-components.mdx
```diff
@@ -0,0 +1,299 @@
...
+      <h1>Welcome, {user.name}</h1>
+      <Suspense fallback={<span>Loading…</span>}>
+        <UserBadge />
+      </Suspense>
+    </UserProvider>
+  )
+}
+```
+
+Client Components read the user with `useUser()`, no prop drilling and no second session read. Keep the read behind a `<Suspense>` boundary, since `use()` suspends until the promise resolves:
```
aurorascharff: User()`, no prop drilling and no second session read

some ai language here

docs/01-app/02-guides/authentication.mdx
icyJoseph: Yeah - there's one ITG task here, to maybe separate this guide into smaller bits. We are at line 1k+ here...

docs/01-app/02-guides/authentication-with-cache-components.mdx
```diff
@@ -0,0 +1,297 @@
...
+}
+```
+
+Keep the session read out of a layout's top level, too. A top-level `await` on the session in a layout holds the whole segment, including `{children}`, behind that request, so push it into a component inside a boundary. See [Push dynamic access down](/docs/app/guides/streaming#push-dynamic-access-down).
+
+## Step 3: Share the user across components
+
+You don't have to read the session again in every component that needs the user. Read it once, then hand it to as many Server and Client Components as you like from inside the same boundary.
+
+Server Components can call `getCurrentUser()` directly. To reach Client Components without prop drilling, create the promise once and pass it through context, then unwrap it with [`use()`](https://react.dev/reference/react/use). This is the [context and `React.cache` pattern](/docs/app/getting-started/fetching-data#sharing-data-with-context-and-reactcache) from data fetching, with one auth-specific adjustment: because `getCurrentUser` reads the request, create its promise inside the Suspense boundary, not at the top of a layout.
```
aurorascharff: Didnt we move this link?

aurorascharff: https://github.com/vercel/next.js/pull/95860

Ah, we should merge this first

aurorascharff: "This is the [context and `React.cache` pattern](/docs/app/getting-started/fetching-data#sharing-data-with-context-and-reactcache) from data fetching, with one auth-specific adjustment:"

Maybe we dont need to say this, more like

"we can bla bla context, learn more about it in #"

docs/01-app/02-guides/authentication-with-cache-components.mdx
```diff
@@ -0,0 +1,297 @@
...
+  return use(userPromise)
+}
+```
+
+The Server Component behind the boundary creates the promise, awaits it for its own rendering, and passes the same promise to the provider:
+
+```tsx filename="app/page.tsx"
+async function Dashboard() {
+  const userPromise = getCurrentUser()
+  const user = await userPromise
```
aurorascharff: Hm, why is it like this?

docs/01-app/02-guides/authentication-with-cache-components.mdx
```diff
@@ -0,0 +1,297 @@
...
+
+On the server, plain `use cache` keeps the entry in memory as a best effort: it's evicted under pressure and, in serverless, doesn't persist across instances. If that data must survive across instances and requests, opt into [`use cache: remote`](/docs/app/api-reference/directives/use-cache-remote) for durable, shared storage, where [the cache key you choose](/docs/app/api-reference/directives/use-cache-remote#cache-key-considerations) drives your hit rate.
+
+## Step 5: Update session-derived data
+
+When a [Server Action](/docs/app/getting-started/mutating-data) changes a user's data, call [`updateTag`](/docs/app/api-reference/functions/updateTag) with the same tag to refresh the cached entry. Re-read the session inside the action so it authorizes itself rather than trusting the client.
```
aurorascharff: Maybe worth linking to the security guide or similar where we talk about why this is important

docs/01-app/02-guides/authentication-with-cache-components.mdx
```diff
@@ -0,0 +1,297 @@
...
+    <ul>
+      {announcements.map((announcement) => (
+        <li key={announcement}>{announcement}</li>
+      ))}
+    </ul>
+  )
+}
+
+async function Dashboard() {
+  const user = await getCurrentUser()
```
aurorascharff, line 151: Should we add another data fetching function with verifySession?

docs/01-app/02-guides/authentication.mdx
aurorascharff: Would be ideal do do in this PR and move copy over some of the DAL considerations too, and link out to the new one

docs/01-app/02-guides/authentication.mdx
```diff
@@ -1361,6 +1365,8 @@ Session and user data often appear in shell UI (header, nav) that repeats across
 
 If only part of the shell needs session data (for example, a user menu), move the `await` into a nested Server Component and wrap it in `<Suspense>` so the rest of the page streams first. See [Push dynamic access down](/docs/app/guides/streaming#push-dynamic-access-down) for the pattern.
 
+> **With Cache Components:** the same applies. Keep the session read behind `<Suspense>` so the static shell streams first. See [Show the user without blocking the page](/docs/app/guides/authentication-with-cache-components#step-2-show-the-user-without-blocking-the-page).
```
aurorascharff: Maybe all of these good to knows could be consolidates into a "cache components "section rather or somewhere where its not as noisy, linking out

aurorascharff (approved):

docs/01-app/02-guides/authentication-with-cache-components.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+Because a user session is valid for a period of time, adding a cache lifetime lets the framework prefetch that content ahead of time.
+
+The server-side directives can't give it that lifetime, though: neither a plain [`use cache`](/docs/app/api-reference/directives/use-cache) nor [`use cache: remote`](/docs/app/api-reference/directives/use-cache-remote) can call `cookies()`, and you can't [extract the value and pass it in](/docs/app/getting-started/caching#passing-runtime-values-to-cached-functions) either, because:
+
+- A session helper reads the cookie deep inside its own code, so there's nothing to lift out.
+- Validating it compares a token's expiry against the current time (iron-session's `unsealData` rejects an expired seal), so the read is request- and time-dependent.
+
+That's what [`use cache: private`](/docs/app/api-reference/directives/use-cache-private) is for: it reads `cookies()` and `headers()` directly, keeping the result in the browser only, never on the server.
```
icyJoseph, line 58: In this and the next paragraph, we say, never on the server  — the correct statement should be, that the result is scoped to the current context request, and won't persist to any other system. 

So it is possible to dedupe these functions too.

PR #95803, opened 2026-07-15
docs: cover direct visits and client navigations in the instant() e2e example

icyJoseph:
Reworks the instant-navigation guide's e2e testing example to cover both ways a route is reached: a direct visit (SSR document load) and a client navigation. 

Each test wraps the navigation in `instant()` and asserts the shell renders while the streamed content is held back, then appears after. 

Also documents where the testing API is available: automatically under `next dev`, and in a production build via `experimental.exposeTestingApiInProductionBuild`.

Closes: https://linear.app/vercel/issue/DOC-6477/instant-e2e-helper-for-ssr-and-soft-navs-snippets-show-difference


docs/01-app/02-guides/instant-navigation.mdx
```diff
@@ -295,28 +295,54 @@ yarn add -D @next/playwright @playwright/test
...
+      await page.goto('/store/hats')
+      await expect(page.locator('h1')).toContainText('Baseball Cap') // in the shell
+      await expect(page.getByText('In stock')).toHaveCount(0) // held back
+    },
+    { baseURL }
+  )
+  await expect(page.getByText('In stock')).toBeVisible() // streamed in after
+})
 
+// Client navigation: only the tree below the shared /store layout re-renders.
```
aurorascharff: Would be nice to have @gnoff verify that this is the intentional usage. It worked when Jude and I were testing!

aurorascharff (approved):

PR #95805, opened 2026-07-15
Fix ?dpl=undefined chunk URLs when deployment id is missing at manifest eval time

### LINKED ISSUE #95397: Dynamic routes emit `?dpl=undefined` chunk URLs on standard Vercel builds — every chunk downloaded twice (auto-enabled `runtimeServerDeploymentId`)

https://github.com/Adophlidu/next-dpl-undefined-repro

### To Reproduce

Skew Protection asset suffixes are wired up by the Vercel build environment, so this requires deploying to Vercel (any plan — reproduces on a Hobby project with no special settings). It does not reproduce locally; full steps are in the repro README.

1. Deploy the repro (one static page `/`, one `force-dynamic` page `/dyn`, `deploymentId: process.env.VERCEL_DEPLOYMENT_ID` in `next.config.mjs`): `npx vercel deploy --prod`
2. Compare the served HTML:

```bash
curl -s https://<deployment-url>/    | grep -o 'dpl=[A-Za-z0-9_]*' | sort | uniq -c
curl -s https://<deployment-url>/dyn | grep -o 'dpl=[A-Za-z0-9_]*' | sort | uniq -c
```

Observed (next 16.2.6, also 16.2.7; Vercel CLI 54.18.7):

```
/     ->  20 dpl=dpl_9yDhqpKfbm76Rb5knDDfGm9WX5HR      # static: all correct
/dyn  ->  10 dpl=dpl_9yDhqpKfbm76Rb5knDDfGm9WX5HR      # dynamic: mixed
          12 dpl=undefined
```

**Current:**

On dynamically rendered routes, the same chunk is referenced twice in one document — `<script src=".../chunk.js?dpl=dpl_xxx" async>` (entry scripts, correct) and `<script src=".../chunk.js?dpl=undefined" async>` (manifest-driven flight/preinit refs) — so browsers download **every chunk twice**. The RSC flight payload also carries the `dpl=undefined` URLs, so lazily loaded chunks use the broken suffix as well: those requests are not pinned by Skew Protection, and the `?dpl=undefined` query never changes across deployments, defeating the cache-busting semantics. Statically prerendered routes emit correct suffixes only.

On a mid-sized production app this doubled 67 chunks (~1.9MB extra compressed transfer and ~80 extra requests per cold visit), and left ~1250 `dpl=undefined` chunk references in the flight payload of a single page.

The mechanism appears to be the same auto-enable reported in #94734 for the prebuilt flow: when Vercel runs `next build` (`hasNextSupport` true, `NEXT_DEPLOYMENT_ID` present in the build env), `packages/next/src/server/config.ts` silently sets `experimental.runtimeServerDeploymentId = true`, deferring chunk-URL deployment ids to a runtime `process.env.NEXT_DEPLOYMENT_ID` / `globalThis.NEXT_DEPLOYMENT_ID` read. On the deployed serverless functions' dynamic-render path that value resolves to `undefined` and is stringified straight into the URL (no guard on this path, unlike `getDeploymentIdQuery`).

Notes from debugging (details in the repro README):

- `next build` + `next start` locally: `/dyn` is all-correct — the auto-enable never happens without `hasNextSupport`.
- `vercel build` locally: zero `dpl=undefined` anywhere in `.vercel/output`, so the broken value is produced at request time, not baked.
- A probe route on the deployed app shows `process.env.VERCEL_DEPLOYMENT_ID` set, while `globalThis.NEXT_DEPLOYMENT_ID` and `globalThis.NEXT_CLIENT_ASSET_SUFFIX` are `null` in the function runtime.
- Setting the env var or the globals from `instrumentation.ts` does not reach the rendering context.
- `experimental.useSkewCookie: true` sets the `__vdpl` cookie but does not remove the `?dpl=` suffixes; the `undefined` variants remain.

**Expected:**

Dynamic and static routes should emit the same, correct `?dpl=` suffix exactly once per chunk — or no suffix at all — but never a stringified `undefined`, and never both variants of the same chunk in one document.

**Verified workaround:** `experimental: { runtimeServerDeploymentId: false }` — with it, `/dyn` emits all-correct suffixes (`20 dpl=dpl_xxx, 0 undefined`) and the `x-nextjs-deployment-id` header on RSC responses stays intact.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.3.0
Binaries:
  Node: 24.13.0
  npm: 11.6.2
Relevant Packages:
  next: 16.2.6 (also reproduced on 16.2.7; 16.2.8-16.2.10 contain no code changes)
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 6.0.3
Next.js Config:
  output: N/A
Deployment: Vercel (standard build — Vercel runs `next build`; NOT the prebuilt flow)
```

Runtime, Turbopack

Vercel (Deployed)

UditDewan:
### What?

Fixes dynamically rendered App Router pages emitting `?dpl=undefined` chunk URLs (in both the HTML and the RSC flight payload) when `experimental.runtimeServerDeploymentId` is enabled but `process.env.NEXT_DEPLOYMENT_ID` is not visible at the time the client reference manifest is evaluated.

### Why?

When `runtimeServerDeploymentId` is on (auto-enabled on standard Vercel builds), the Turbopack-generated client reference manifest defers the deployment id to a runtime read and appends it to every chunk URL:

```js
val.chunks = val.chunks.map((c) => `${c}?dpl=${process.env.NEXT_DEPLOYMENT_ID}`)
```

Unlike every other `dpl` producer (`getDeploymentIdQuery`, `getAssetQueryString`, the Turbopack runtime's `globalThis[...] || ''` templates), this interpolation has no guard. If the env var is missing when the manifest is evaluated (`evalManifest` runs it in a `vm` sandbox that only exposes `process.env.NEXT_DEPLOYMENT_ID`), the literal string `undefined` is baked into every chunk URL. The same chunk then gets referenced with two different suffixes in one document, so browsers download every chunk twice, and the `?dpl=undefined` query never changes across deployments, defeating both Skew Protection pinning and cache busting. See #95397 for a deployed reproduction and measurements.

### How?

Wrap the generated suffix code in an `if (process.env.NEXT_DEPLOYMENT_ID)` guard so the query is only appended when the deployment id is actually available, aligning this path with the other asset-suffix code paths. When the id is missing the chunk URLs are left untouched instead of being poisoned.

Added a regression test that evaluates the built manifest the same way `evalManifest` does — once without the env var (chunks must not contain `dpl=`) and once with it (chunks must carry the correct `?dpl=` suffix).

Fixes #95397

### DIFF
--- a/crates/next-core/src/next_manifests/client_reference_manifest.rs
+++ b/crates/next-core/src/next_manifests/client_reference_manifest.rs
@@ -487,12 +487,17 @@ async fn build_manifest(
                     entry_name = StringifyJs(&normalized_manifest_entry),
                     manifest = &client_reference_manifest_json,
                     suffix = if add_deployment_id_at_runtime {
+                        // Only append the suffix when the deployment id is actually available at
+                        // runtime, otherwise a missing env var would be stringified into every
+                        // chunk URL as `?dpl=undefined`.
                         formatdoc!{
                             r#"
-                            for (const key in globalThis.__RSC_MANIFEST[{entry_name}].clientModules) {{
-                                const val = {{ ...globalThis.__RSC_MANIFEST[{entry_name}].clientModules[key] }}
-                                globalThis.__RSC_MANIFEST[{entry_name}].clientModules[key] = val
-                                val.chunks = val.chunks.map((c) => `${{c}}?dpl=${{process.env.NEXT_DEPLOYMENT_ID}}`)
+                            if (process.env.NEXT_DEPLOYMENT_ID) {{
+                                for (const key in globalThis.__RSC_MANIFEST[{entry_name}].clientModules) {{
+                                    const val = {{ ...globalThis.__RSC_MANIFEST[{entry_name}].clientModules[key] }}
+                                    globalThis.__RSC_MANIFEST[{entry_name}].clientModules[key] = val
+                                    val.chunks = val.chunks.map((c) => `${{c}}?dpl=${{process.env.NEXT_DEPLOYMENT_ID}}`)
+                                }}
                             }}
                             "#,
                             entry_name = StringifyJs(&normalized_manifest_entry),
--- a/test/production/deployment-id-handling/deployment-id-handling.test.ts
+++ b/test/production/deployment-id-handling/deployment-id-handling.test.ts
@@ -126,6 +126,41 @@ describe.each([
       }
     )
 
+    if (runtimeServerDeploymentId && process.env.IS_TURBOPACK_TEST) {
+      // The client reference manifest defers the deployment id to a runtime
+      // process.env.NEXT_DEPLOYMENT_ID read. If that value is missing when the
+      // manifest is evaluated, it must not be stringified into the chunk URLs
+      // as `?dpl=undefined`.
+      it('should not emit dpl=undefined chunk URLs when the deployment id is missing at manifest eval time', async () => {
+        const vm = require('node:vm')
+        const manifestSource = await next.readFile(
+          '.next/server/app/from-app/page_client-reference-manifest.js'
+        )
+
+        const getChunks = (env: Record<string, string | undefined>) => {
+          const context: any = { process: { env } }
+          vm.runInNewContext(manifestSource, context)
+          const clientModules =
+            context.__RSC_MANIFEST['/from-app/page'].clientModules
+          return Object.values(clientModules).flatMap(
+            (mod: any) => mod.chunks as string[]
+          )
+        }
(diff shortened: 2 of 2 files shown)


mischnic:
This is wrong.
> experimental.runtimeServerDeploymentId is enabled but process.env.NEXT_DEPLOYMENT_ID is not visible at the time the client reference manifest is evaluated.

You shouldn't have runtimeServerDeploymentId enabled in this setup.

PR #95817, opened 2026-07-15
Refine Cache Components and Partial Prefetching adoption skills

aurorascharff:
Refinements to the Cache Components and Partial Prefetching adoption skills from running them cold against real apps, plus the doc changes they lean on.

**Skills**
- PPF: URL-data sweep reworked to feature-by-feature with a resumable hand-off, and the requires section now says unfinished Cache Components adoption doesn't gate the skill. Prerender insights are non-blocking dev signals expected on a fresh branch; only a build-blocking failure stops the run.
- PPF: install `next-dev-loop` without asking, and verify a real blocker before falling back.
- CC: dropped the hard-coded `agent-browser` version pin; updated the PPF cross-reference to the current enable-first flow.

**Docs**
- `runtime-prefetching`: the per-card server cost of `allow-runtime` grids is specific to `prefetch={true}`, with hover-triggered prefetch as the mitigation.
- `use-cache`: the request-API restriction follows the call stack (`next-request-in-use-cache`, can pass build and fail under `next start`), and the nested-`cacheLife` build error with its fix.
- `instant` route segment config: removed a stale `version: draft` so the page publishes.


gaojude (approved):

docs/01-app/03-api-reference/03-file-conventions/02-route-segment-config/instant.mdx
```diff
@@ -1,7 +1,6 @@
 ---
 title: instant
 description: API reference for the instant route segment config.
-version: draft
```
icyJoseph, line 4: Good catch there

docs/01-app/02-guides/runtime-prefetching.mdx
```diff
@@ -201,14 +201,16 @@ Use it on routes where:
...
 - The dependent content has to be fresh on every request. The prerender stops at the same `<Suspense>` fallback, so the user sees the same UI either way.
 - The route is rarely navigated to. You pay per visible link, regardless of click-through.
 
 A per-link runtime prefetch is best-effort. It only helps the navigations where it completes before the click. On a slow connection, on a feed of many links, or on a direct visit, it may not be ready when the user navigates, and the navigation falls back to the App Shell. That shell is the reliable baseline, and runtime prefetching only layers on top when it arrives in time.
 
+When many links to an `allow-runtime` route are visible at once, such as a grid of cards, each `<Link prefetch={true}>` prefetches that link's content as it enters the viewport, so the grid wakes a server per card. Prefetch on intent instead: a [hover-triggered prefetch](/docs/app/guides/prefetching#hover-triggered-prefetch) fetches only the links the user is likely to click. The default `<Link>` (without `prefetch={true}`) prefetches just the App Shell, so it doesn't carry this cost.
```
icyJoseph: This "wakes a server" - re-reading it now, and specially when it says "per card", doesn't spark joy.

Maybe both in line 204 and here we can be more specific, like:

> Each visible <Link prefetch={true}> to a route with 'allow-runtime' makes its own request to the server to render that link, and that cost only pays off if more of the page is ready before the click.

And

> ...so the grid makes one such server request per card.

wydt?

aurorascharff: yes, agreed

docs/01-app/03-api-reference/01-directives/use-cache.mdx
```diff
@@ -314,6 +314,8 @@ async function getData() {
 }
 ```
 
+> **Good to know:** A `use cache` with a zero `revalidate` or a short `expire` (under 5 minutes), nested inside another `use cache` that has no explicit `cacheLife`, is not allowed during prerendering and fails the build with [`nested-use-cache-no-explicit-cachelife`](/docs/messages/nested-use-cache-no-explicit-cachelife). Add [`cacheLife()`](/docs/app/api-reference/functions/cacheLife) to the **outer** `use cache` to choose whether it prerenders or stays dynamic.
```
icyJoseph: Maybe this doesn't need to be a gtk:

```suggestion
Nesting a short-lived use cache inside one without an explicit [cacheLife](/docs/app/api-reference/functions/cacheLife) fails the build during prerendering. See [Nested short-lived caches](/docs/app/api-reference/functions/cacheLife#nested-short-lived-caches) for the rule and fix.
```

PR #95818, opened 2026-07-15
Fix Request Insights span collection

timneutkens:
## Summary

Fix Request Insights collection so the dev server records all non-hidden Next.js spans locally, including non-vanilla trace and wrapped spans, while preserving the existing OpenTelemetry export allowlist unless verbose tracing is enabled.

Keep reporting timelines correct by propagating request identity before work async storage exists, honoring explicit span timestamps, and anchoring request duration to the BaseServer request span. Standardize span categories on `next.span_category`, retain presentation-only UI filtering, and filter internal request identity headers at ingress.

## Verification

- `pnpm test-dev-turbo test/development/app-dir/request-insights/request-insights.test.ts`

### DIFF
--- a/packages/next/src/server/lib/trace/local-span-recorder.ts
+++ b/packages/next/src/server/lib/trace/local-span-recorder.ts
@@ -7,6 +7,7 @@ import type { AsyncLocalStorage } from 'async_hooks'
 import { SpanStatusCode } from 'next/dist/compiled/@opentelemetry/api'
 import {
   isLocalSpanRecordingEnabled,
+  isRequestInsightsEnabled,
   recordSpan,
   type SpanStoreAttributes,
   type SpanStoreEvent,
@@ -34,6 +35,7 @@ export function createLocalSpan({
   name,
   attributes,
   links,
+  startTime,
   traceId,
   spanId,
   parentSpanId,
@@ -42,6 +44,7 @@ export function createLocalSpan({
   name: string
   attributes?: LocalSpanAttributes
   links?: SpanOptions['links']
+  startTime?: SpanOptions['startTime']
   traceId?: string
   spanId?: string
   parentSpanId?: string
@@ -51,6 +54,7 @@ export function createLocalSpan({
     name,
     attributes,
     links,
+    startTime,
     delegateSpan,
     traceId: traceId ?? getLocalTraceId(),
     spanId: spanId ?? getLocalSpanId(),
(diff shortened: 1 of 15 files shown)


packages/next/src/server/lib/trace/tracer.ts
```diff
@@ -364,11 +366,15 @@ class NextTracerImpl implements NextTracer {
...
-    ) {
+    const isVanillaSpan =
+      NextVanillaSpanAllowlist.has(type) ||
+      process.env.NEXT_OTEL_VERBOSE === '1'
+    const shouldRecordLocalSpan =
+      localSpanRecordingEnabled &&
+      (isVanillaSpan ||
+        (localSpanRecorder?.isRequestInsightsEnabled() ?? false))
+
+    if ((!isVanillaSpan && !shouldRecordLocalSpan) || options.hideSpan) {
```
mischnic: This expression is way too complex.
Can we expression this somehow based on
```js
    const shouldTraceSpan =
      // Default list
      NextVanillaSpanAllowlist.has(type) ||
      // User opt in
      process.env.NEXT_OTEL_VERBOSE === '1' ||
      // Always record all with request insights
      (localSpanRecorder?.isRequestInsightsEnabled() ?? false)
```

packages/next/src/server/lib/trace/local-span-recorder.ts
```diff
@@ -452,20 +481,24 @@ function getSpanStoreExceptionAttributes(
...
+    const requestInsightsIdentity = getRequestInsightsIdentity()
     const url =
       workUnitStore && 'url' in workUnitStore ? workUnitStore.url : undefined
 
     return {
-      requestId: workStore?.requestId,
-      htmlRequestId: workStore?.htmlRequestId,
+      requestId: workStore?.requestId ?? requestInsightsIdentity?.requestId,
+      htmlRequestId:
+        workStore?.htmlRequestId ?? requestInsightsIdentity?.htmlRequestId,
```
mischnic, line 499: We already have `workStore.htmlRequestId`. Why do we need another async local storage (packages/next/src/server/lib/trace/request-insights-identity.ts)?

packages/next/src/server/lib/trace/tracer.ts
timneutkens: :robot: Done — I simplified this to a shouldTraceSpan predicate covering the default allowlist, verbose opt-in, and Request Insights. External delegation is kept as a separate shouldDelegateSpan gate so non-vanilla Request Insights spans remain local.

packages/next/src/server/lib/trace/local-span-recorder.ts
timneutkens: :robot: Kept the outer request identity scope and documented why it is needed: BaseServer.handleRequest and route-matching spans start before App Render creates workAsyncStorage, so workStore.htmlRequestId is not available for those spans. App Render copies the outer identity into workStore, which remains the primary source afterward. The integration test “keeps outer server and app render spans on the same request” verifies both span layers use one request.

mischnic (approved):

PR #95819, opened 2026-07-15
docs: use a valid src in Image example

JucieltonSantos:
Replace the empty src in the introductory next/image examples with a valid local image source.

### DIFF
--- a/docs/01-app/01-getting-started/12-images.mdx
+++ b/docs/01-app/01-getting-started/12-images.mdx
@@ -21,7 +21,7 @@ To start using `<Image>`, import it from `next/image` and render it within your 
 import Image from 'next/image'
 
 export default function Page() {
-  return <Image src="" alt="" />
+  return <Image src="/profile.png" alt="Picture of the author" width={500} height={500} />
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jucieltonsantos-orbia:
@icyJoseph This is about the issue we discussed. https://github.com/vercel/next.js/discussions/95780

docs/01-app/01-getting-started/12-images.mdx
```diff
@@ -21,15 +21,15 @@ To start using `<Image>`, import it from `next/image` and render it within your
 import Image from 'next/image'
 
 export default function Page() {
-  return <Image src="" alt="" />
+  return <Image src="/profile.png" alt="Picture of the author" width={500} height={500} />
```
icyJoseph, line 24: Mmm the intention is right. I think I'll approve, but I think the flow of this doc, was literally, just import the component, and give it this props, next we'll see what types of src you can have. 

While this prevents the "empty src" situation, it makes a choice for local images. 

What was the reasoning, or issue you found with `""` do agents, think it does anything?

JucieltonSantos: Thanks, that makes sense.

My reasoning was that `src=""` is not a usable image source and `next/image` reports it as a missing required `src` in development. It may also be interpreted by readers or coding agents as a valid copy-pasteable example.

However, I agree that replacing it with `/profile.png` makes an early choice for local images before the documentation introduces local and remote sources.

Would it be more appropriate to remove this introductory snippet and let the following local and remote sections provide the first concrete examples? That would avoid both the empty `src` and the early preference for a specific source type.

Issue #95815, opened 2026-07-15
Turbopack 16.3: nft.json drops store target files for an ESM serverExternalPackage reached via a 'use client' component in a shared SSR-layer chunk → dangling alias → ERR_MODULE_NOT_FOUND on Vercel


cipriancaba:
## Link to the code that reproduces this issue

https://github.com/cipriancaba/next-turbopack-nft-ssr-chunk-repro

`bun install && bun run build && node check-nft.mjs` — reproduces the mechanism up to the ssr-only chunk placement; see "Minimal reproduction — partial" below for why the final cross-route shared-chunk step needs an app-scale module graph, and `analysis/` for the chunk-level evidence extracted from a real production build.

> Related: #87737 — that issue describes a "hash mismatch after reinstall" variant. This report is a **distinct trigger** sharing the same alias/store machinery, isolated at chunk level with a healthy-vs-dangling diff from a single production build. It refines the bun/Vercel findings I posted in https://github.com/vercel/next.js/issues/87737#issuecomment-4897366831 (2026-07-06) — the root cause turned out to be narrower and more precise than the "all ESM externals" framing there.

## Summary

Since Next 16.1, Turbopack externalizes server packages behind content-hashed aliases
(`.next/node_modules/<pkg>-<hash>` → a symlink into the package-manager store). On 16.2.9 each
route's `*.nft.json` lists **both** the alias symlink and the store target files, so the packed
lambda resolves. On **16.3** (preview + canary), for a specific class of endpoints the trace lists
the alias symlink **but not the store target files**, so Vercel packs a dangling symlink and every
import of that package throws `ERR_MODULE_NOT_FOUND` (HTTP 500).

The affected endpoints are precisely those whose nft includes a **shared SSR-layer chunk**
(`.next/server/chunks/ssr/…`) that references an externalized package the endpoint does **not**
import in its own module graph.

## Environment

- `next@16.3.0-preview.6` (also reproduced on `16.3.0-canary.86`); 16.2.9 is unaffected.
- Turbopack (default). App Router, `cacheComponents: true`, partial-prerendered dynamic routes.
- Package manager: bun with the **isolated** linker (store at `node_modules/.bun/<pkg>@<ver>/…`);
  the mechanism is not bun-specific — any linker that produces store symlinks behind the alias
  should show it.

## Root cause / mechanism

1. **The discriminator is the chunk layer.** An externalized package whose alias appears only in
[...]

PR #92901, opened 2026-04-16
Turbopack: use module graph for NFT

mischnic:
- [x] skip traced edges properly in all graph traversals
- [x] iter_nodes and enumerate_nodes include the traced modules
    - [x] Don't propagate is_self_async inside of traced references as a perf optimization
- [x] fix performance
- [x] validate that NFT JSON files are identical
- [x] ~~validate/adjust bundle analyzer output~~ will be in followup
- [x] audit `fn chunking_type()` changes

Benchmark:
```
* ced85b3108 - (2 weeks ago) refactor - Niklas Mischkulnig (HEAD -> mischnic/nft-module-graph, origin/mischnic/nft-module-graph)

TURBOPACK_PERSISTENT_CACHE=0 NEXT_TURBOPACK_TRACING= pnpm next build   334.96s user 44.57s system 788% cpu 48.152 total
TURBOPACK_PERSISTENT_CACHE=0 NEXT_TURBOPACK_TRACING= pnpm next build   343.39s user 42.61s system 809% cpu 47.659 total

* 5edf48238f - (77 minutes ago) v16.3.0-canary.32 - next-js-bot[bot] (HEAD, tag: v16.3.0-canary.32)

TURBOPACK_PERSISTENT_CACHE=0 NEXT_TURBOPACK_TRACING= pnpm next build   333.42s user 43.57s system 712% cpu 52.877 total
TURBOPACK_PERSISTENT_CACHE=0 NEXT_TURBOPACK_TRACING= pnpm next build   332.95s user 43.15s system 791% cpu 47.515 total
TURBOPACK_PERSISTENT_CACHE=0 NEXT_TURBOPACK_TRACING= pnpm next build   337.59s user 41.28s system 771% cpu 49.117 total
TURBOPACK_PERSISTENT_CACHE=0 NEXT_TURBOPACK_TRACING= pnpm next build   335.25s user 45.53s system 718% cpu 52.975 total
```

<details>
<summary>Previous benchmarks</summary>

```
52212418699f362904d64c08e544d4994e8d27fa mischnic/nft-module-graph

pnpm next build   371.41s user 66.39s system 737% cpu 59.353 total
pnpm next build   380.13s user 59.62s system 751% cpu 58.548 total
pnpm next build   374.25s user 60.55s system 761% cpu 57.102 total

8e4cfc5062 canary

pnpm next build   370.77s user 53.89s system 762% cpu 55.725 total
pnpm next build   383.16s user 55.61s system 744% cpu 58.921 total
[...]

Issue #87737, opened 2025-12-23
Turbopack generates external module references with hashes that don't match installed packages when node_modules structure differs


yWilliamLima:
### Link to the code that reproduces this issue

https://github.com/Aluisio/test-styled-components

### To Reproduce

1. Create a Next.js 16.1.1+ project with OpenTelemetry instrumentation
2. Add `instrumentation.ts` file that imports `@opentelemetry/auto-instrumentations-node`
3. Build the project in a CI/CD environment (e.g., GitHub Actions) using `pnpm install --frozen-lockfile` followed by `next build`
4. Create a tar.gz archive **excluding** `node_modules` directory
5. Extract the tar.gz in a different environment
6. Run `pnpm install` to install dependencies
7. Run `next start` to start the production server

### Expected Behavior

The application should start successfully, loading the instrumentation hook without errors.

### Current Behavior

The application fails to start with the following error:

```
Error: An error occurred while loading instrumentation hook: Failed to load external module require-in-the-middle-a99415fa67232f7f: Error: Cannot find module 'require-in-the-middle-a99415fa67232f7f'
Require stack:
- /path/to/.next/server/chunks/[root-of-the-server]__39a9a9e1._.js
- /path/to/.next/server/chunks/[turbopack]_runtime.js
- /path/to/.next/server/instrumentation.js
```

The Turbopack bundler generates external module references with hashes (e.g., `require-in-the-middle-a99415fa67232f7f`) based on the `node_modules` structure during the build. When dependencies are reinstalled in a different environment, the `node_modules` structure may differ (especially with pnpm's symlink/hoisting behavior), causing the hash to not match the actual installed package.

### Workaround

Using `--webpack` flag in the build script resolves the issue:

```json
{
  "scripts": {
    "build": "next build --webpack"
  }
}
```

However, this defeats the purpose of using Turbopack as the default bundler in Next.js 16.

### Provide environment information

```bash
**Operating System:**

- Platform: linux
- Arch: x64
- Version: Ubuntu (GitHub Actions runner)

**Binaries:**

- Node: 24.11.1
- pnpm: 10.25.0

**Relevant Packages:**

- next: 16.1.1
- react: 19.2.3
- react-dom: 19.2.3
- @opentelemetry/auto-instrumentations-node: 0.67.3
- require-in-the-middle: 8.0.1
- import-in-the-middle: 2.0.1
```

### Which area(s) are affected? (Select all that apply)

Output

[...]

PR #95824, opened 2026-07-15
Turbopack: trace externals imported only by server actions

gaearon:
#95815 refactored the NFT flow and ended up only tracing the RSC template and it's subgraph. Not the server actions loader module whose module graph is completely separate.

---

Clauded.

Attempts to fix https://github.com/vercel/next.js/discussions/95130#discussioncomment-17652155.

---

<details>

## Summary

Fixes the NFT tracing regression behind #95815 and the deploy failures reported in https://github.com/vercel/next.js/issues/87737#issuecomment-4897366831.

Since the 16.3 NFT rewrite (#94224 / #92901), a `serverExternalPackages` entry that is imported **only from a `'use server'` action referenced by a client component** ends up in the route's trace as the content-hashed alias symlink (`.next/node_modules/<pkg>-<hash>`) **without the store files the symlink points to**. Any output assembled from the traces (Vercel lambdas, `output: 'standalone'`) then contains a dangling symlink, and invoking the action throws `Failed to load external module <pkg>-<hash>` (HTTP 500). 16.2.9 traces both. Both CJS and ESM externals are affected, with any package manager — the ESM/bun/monorepo framing in the issue reports was a coincidence of which packages were only reachable through actions.

The cause: `AppEndpoint::trace_result` started the trace DFS only from `rsc_entry`, but the server actions loader is a separate module-graph entry (`additional_entries`). It is chunked into the endpoint output (which emits the alias symlink via the chunk's output assets), yet its subgraph — including the externals' traced target files — was never visited by the tracer. `trace_endpoint` now accepts multiple entry modules and the app endpoint passes the actions loader alongside `rsc_entry`.

The test observes the user-visible failure: it builds with `output: 'standalone'`, deletes everything except the standalone output (so only traced files are available, like a deployed lambda), runs `server.js`, and invokes the server action in a browser. Without the fix the action fails with `Failed to load external module lodash-<hash>`; on 16.2.9 the same fixture responds with the action result. Note the fixture must not import the externals from any other route: the standalone output is the union of all route traces, so a route that traces them correctly would mask the missing entries.

## Verification

- `pnpm test-start-turbo test/production/standalone-mode/server-action-externals/standalone-mode-server-action-externals.test.ts` — fails without the fix (`Failed to load external module lodash-fb78ad0536ffcdab` from the standalone server), passes with it. The equivalent standalone flow on `next@16.2.9` returns the action result with HTTP 200.
- `pnpm test-start-turbo test/e2e/externals-transitive/externals-transitive.test.ts`
- `pnpm test-start-turbo test/production/standalone-mode/server-actions/standalone-mode-server-actions.test.ts`

</details>


mischnic (approved):
Thankfully these annoying additional_entries will be removed soon anyway.

PR #95826, opened 2026-07-15
[turbopack] Do the CJS analysis needed for scope hoisting

sampoder:
This PR does two things: adds a flag `turbopackCjsScopeHoisting` and plumbs it through and adds the extra stuff we need in `turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs` (just the names of the things exported by CJS) for PRs down the stack to work.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
@@ -29,6 +29,7 @@ use crate::{
         graph::{ConditionalKind, Effect, EffectArg, EffectsBlock, EvalContext, VarGraph},
         is_unresolved_id,
     },
+    chunk::CjsStaticExports,
     code_gen::CodeGen,
     references::{
         AstPath,
@@ -105,6 +106,11 @@ struct CjsExportsCollector {
     writes: Vec<DroppableCjsExportAssignment>,
     /// Whether the `exports.__esModule = true` interop marker is set.
     has_es_module: bool,
+    /// Whether the module body contains a top-level `return`, which skips every
+    /// write that follows it.
+    has_top_level_return: bool,
+    /// Whether any export is defined with `Object.defineProperty`.
+    has_define_property_export: bool,
 }
 
 trait FunctionLike {
@@ -658,6 +664,20 @@ mod analyzer_state {
             }
         }
 
+        /// Records a `return` in the module body, which exits the module early.
+        pub(super) fn set_cjs_has_top_level_return(&mut self) {
+            if let Some(c) = &mut self.state.cjs_exports {
+                c.has_top_level_return = true;
+            }
+        }
+
+        /// Records that an export is defined with `Object.defineProperty`.
+        pub(super) fn set_cjs_has_define_property_export(&mut self) {
+            if let Some(c) = &mut self.state.cjs_exports {
+                c.has_define_property_export = true;
+            }
+        }
+
         pub(super) fn record_cjs_export(&mut self, name: RcStr, path: AstPath) {
             self.push_cjs_export(DroppableCjsExportAssignment::Write { name, path });
         }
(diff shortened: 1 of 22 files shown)


turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -32,6 +32,12 @@ use crate::{
 pub trait EcmascriptChunkPlaceable: ChunkableModule + Module {
     #[turbo_tasks::function]
     fn get_exports(self: Vc<Self>) -> Vc<EcmascriptExports>;
+    /// The module's static CommonJS exports for scope hoisting; `None` unless it is
+    /// a statically-analyzable CommonJS module.
+    #[turbo_tasks::function]
+    fn get_cjs_static_exports(self: Vc<Self>) -> Vc<OptionCjsStaticExports> {
+        Vc::cell(None)
+    }
```
mischnic: We shouldn't have a separate method here.

This should just be an enum variant in `get_exports(self: Vc<Self>) -> Vc<EcmascriptExports>`

sampoder: oh yes, i'll do that!

sampoder: made the switch ✅

turbopack/crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -875,7 +876,14 @@ impl ChunkableModule for EcmascriptModuleAsset {
...
     #[turbo_tasks::function]
     async fn get_exports(self: Vc<Self>) -> Result<Vc<EcmascriptExports>> {
-        Ok(*compute_ecmascript_module_exports(self, None).await?.exports)
+        let exports = compute_ecmascript_module_exports(self, None).await?.exports;
+        if let EcmascriptExports::CommonJs(_) = &*exports.await? {
+            return Ok(EcmascriptExports::CommonJs(
+                self.analyze().await?.cjs_static_exports.clone(),
+            )
+            .cell());
+        }
```
mischnic, line 863: just move this into `compute_ecmascript_module_exports`?

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -166,6 +167,9 @@ pub struct AnalyzeEcmascriptModuleResult {
     /// `true` when the analysis was successful.
     pub successful: bool,
     pub source_map: Option<ResolvedVc<Box<dyn GenerateSourceMap>>>,
+    /// Present when the module is a statically-analyzable CommonJS module;
+    /// carries its named exports for scope hoisting.
+    pub cjs_static_exports: Option<CjsStaticExports>,
```
mischnic, line 172: make this a field in `EcmascriptExportsAnalysis` instead?

turbopack/crates/turbopack-ecmascript/src/lib.rs
sampoder: I don't think `compute_ecmascript_module_exports` is doing a full pass, and we're depending on the already existing analyzer visitor so we'd have to keep calling `analyze()` but `compute_ecmascript_module_exports()` is called inside of `analyze()`:

Image

So I think doing this would cause us to run the analyzer twice?

lukesandberg: yeah the goal of this function was to avoid running full analysis (which for ESM unblocked the ability to query exports of dependencies)

so could we compute static exports without the full analyzer too?

sampoder: I don't think that's possible because we need check for all taints of `exports`

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -678,15 +698,44 @@ mod analyzer_state {
...
+        ) -> (
+            Option<(Vec<DroppableCjsExportAssignment>, bool)>,
+            Option<CjsStaticExports>,
+        ) {
+            let Some(c) = self.state.cjs_exports.take() else {
+                return (None, None);
+            };
+            // A top-level `return` skips the writes after it, and would abandon the rest
+            // of a merged factory. An `Object.defineProperty` export keeps its value in
+            // the descriptor, which the merge has no local to bind it to.
```
lukesandberg, line 745: we would typially add a code gen effect to remove the dead code following the early return....

defineProperty does seem like a reasonable but managable opt out.... we could of course still generate the exports 'object' and in the scope hoisted world everyone would dereference through that?

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
sampoder: (see above for why I think this has to stay in the analyzer)

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: I added returns as a bail-out because I didn't really know how to handle them. They open up a bit of can of worms, for example conditional returns (you can't just remove the dead code afterwards). I'm bailing on `...defineProperty` because I didn't know how to rewrite them and there are a lot of options you can pass into an `Object.defineProperty` that I didn't really want to try handling. The basic `__esModule = true` write is fine.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -2214,6 +2264,8 @@ impl VisitAstPath for Analyzer<'_, '_> {
                 .unwrap_or(JsValue::Constant(ConstantValue::Undefined));
 
             self.add_return_value(return_value);
+        } else {
+            self.set_cjs_has_top_level_return();
         }
 
         self.add_early_return_always(ast_path);
```
lukesandberg, line 2328: maybe set the cjs early return from inside here? or something similar

turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -272,7 +272,9 @@ pub enum EcmascriptExports {
     /// A module using `__turbopack_export_namespace__`, used by custom module types.
     DynamicNamespace,
     /// A module using CommonJS exports.
-    CommonJs,
+    ///
+    /// Carries the static export names when statically analyzable, for scope hoisting.
```
lukesandberg, line 276: nit: clarify that `None` simply means the exports were not analyzeable for our scope hoising purposes

lukesandberg (approved):

PR #95833, opened 2026-07-15
Exclude stale under 5 minutes from app shells

gnoff:
App Shells are meant to be useful for handling navigations that are not prefetched. If the stale time is very low then they must be evicted often from the client router and thus their utility is undermined. We now exclude caches that have a stale time under 5 minutes from app shells.

For the default cacheLife profiles this means only the `seconds` profile is excluded becasue every other profile has a stale of exactly 5 minutes at this time.

In the future we may want to allow for this threshhold to be configured. This shouldn't be too disruptive since generally content that is changing often is likely not in your app shell to begin with.


acdlite (approved):

janusfnielsen:
Could I be allowed to decide my own stale times and what I want to have above my Suspense boundaries? Also - this should be in the docs - or even nicer - give us an error, when we did not set the stale >= 300. @gnoff

Issue #95380, opened 2026-07-01
Invalid routes for dynamic pages return 200 response on first render


lmounsey:
### Link to the code that reproduces this issue

https://github.com/lmounsey/cache-components-dynamic-routes

### To Reproduce

1. Build the application using `next build`
2. Start application server using `next start`
3. Visit invalid page route, e.g. http://localhost:3000/this-page-does-not-exist
4. See 200 response

<img width="562" height="344" alt="Image" src="https://github.com/user-attachments/assets/e351fee5-bfad-4545-8c10-11b4468950f7" />

### Current vs. Expected behavior

In `16.2.x` dynamic routes that are invalid correctly return a 404 response on first render when using `cacheComponents`.

There seems to be a change/regression with `16.3.x`, where it looks like it first returns a static shell (200 response) on first load. Once cached, the page then correctly returns a 404 response.

I would expect the page to return a 404 on first render, otherwise this could have an impact on SEO, reporting and monitoring. 

We rely on dynamic routes so new content can be published from the CMS instantly.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jan 19 22:01:13 PST 2026; root:xnu-11417.140.69.708.3~1/RELEASE_ARM64_T8122
  Available memory (MB): 24576
  Available CPU cores: 8
Binaries:
  Node: 25.2.1
  npm: 11.6.2
  Yarn: 1.22.22
  pnpm: 10.17.1
Relevant Packages:
  next: 16.3.0-canary.73 // Latest available version is detected (16.3.0-canary.73).
  eslint-config-next: N/A
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

cacheComponents, Error Handling

### Which stage(s) are affected? (Select all that apply)

next start (local)

### Additional context

This can only be replicated using `next build` + `next start`.

`next dev` returns the correct status code on first render.

I have not tested this deployed to Vercel, it only came up as it's not passing our internal e2e tests.

marcoshernanz:
Thanks for the clear repro. I can reproduce the first-request 200 on 16.3.0-canary.73, but not on 16.3.0-canary.103 or newer after #96297. Fresh invalid routes return 404 on both the first and subsequent requests in 16.3.0, 16.3.2, and 16.4.0-canary.5. Closing as fixed.

PR #95851, opened 2026-07-16
docs: navigation interrupts after streaming has started (notFound, forbidden, unauthorized)

aurorascharff:
Adds a "Calling `x()` after streaming has started" example to the `notFound`, `forbidden`, and `unauthorized` references.

- The check lives in the data-access function that loads the data, inside a `<Suspense>` boundary, so the page shell keeps streaming (blog post 404, account 401, admin projects 403).
- Documents the status-code trade-off: the interrupt fires after headers are sent, so the response is a `200`; a `loading.tsx` is part of the stream, not a way around it; `instant = false` opts the route out of the shell for a real status code.
- Links the authentication guide's DAL section instead of re-teaching it, plus a good-to-know on floating promises never rendering the UI.

Docs-only.

### DIFF
--- a/docs/01-app/03-api-reference/04-functions/not-found.mdx
+++ b/docs/01-app/03-api-reference/04-functions/not-found.mdx
@@ -32,6 +32,96 @@ export default async function Profile({ params }) {
 
 > **Good to know**: `notFound()` does not require you to use `return notFound()` due to using the TypeScript [`never`](https://www.typescriptlang.org/docs/handbook/2/functions.html#never) type.
 
+## Examples
+
+### Calling `notFound()` after streaming has started
+
+To keep a page's shell and loading UI visible while data loads, do the existence check inside a component wrapped in [`<Suspense>`](https://react.dev/reference/react/Suspense) instead of blocking the whole route. The idiomatic place for the check is the data-access function itself, awaited by the component that needs the data:
+
+```tsx filename="app/blog/[slug]/page.tsx" switcher highlight={8}
+import { Suspense } from 'react'
+import Link from 'next/link'
+import { notFound } from 'next/navigation'
+
+async function getPost(slug: string) {
+  const res = await fetch(`https://api.example.com/posts/${slug}`)
+  if (!res.ok) {
+    notFound()
+  }
+  return res.json()
+}
+
+async function Article({ slug }: { slug: string }) {
+  const post = await getPost(slug)
+  return (
+    <article>
+      <h1>{post.title}</h1>
+      <p>{post.content}</p>
+    </article>
+  )
+}
+
+export default async function PostPage({
+  params,
+}: {
+  params: Promise<{ slug: string }>
+}) {
+  const { slug } = await params
+
+  return (
+    <section>
+      <Link href="/blog">Blog</Link>
+      <Suspense fallback={<p>Loading...</p>}>
+        <Article slug={slug} />
+      </Suspense>
+    </section>
+  )
+}
+```
+
+```jsx filename="app/blog/[slug]/page.js" switcher highlight={8}
+import { Suspense } from 'react'
+import Link from 'next/link'
+import { notFound } from 'next/navigation'
+
+async function getPost(slug) {
+  const res = await fetch(`https://api.example.com/posts/${slug}`)
+  if (!res.ok) {
+    notFound()
+  }
+  return res.json()
+}
+
+async function Article({ slug }) {
+  const post = await getPost(slug)
+  return (
+    <article>
+      <h1>{post.title}</h1>
+      <p>{post.content}</p>
+    </article>
+  )
+}
+
+export default async function PostPage({ params }) {
+  const { slug } = await params
+
+  return (
+    <section>
+      <Link href="/blog">Blog</Link>
+      <Suspense fallback={<p>Loading...</p>}>
+        <Article slug={slug} />
+      </Suspense>
+    </section>
+  )
+}
+```
+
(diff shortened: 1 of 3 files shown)


docs/01-app/03-api-reference/04-functions/not-found.mdx
```diff
@@ -30,7 +60,156 @@ export default async function Profile({ params }) {
...
+      <h1>{post.title}</h1>
+      <p>{post.content}</p>
+    </article>
+  )
+}
+
+export default async function PostPage({
+  params,
+}: {
+  params: Promise<{ slug: string }>
```
icyJoseph: `PageProps`

docs/01-app/03-api-reference/04-functions/not-found.mdx
```diff
@@ -30,7 +60,156 @@ export default async function Profile({ params }) {
...
+
+`notFound()` also works in a [Route Handler](/docs/app/api-reference/file-conventions/route), where it serves a `404` to the caller.
+
+```tsx filename="app/api/posts/[slug]/route.ts" switcher
+import { NextResponse } from 'next/server'
+import { notFound } from 'next/navigation'
+
+export async function GET(
+  request: Request,
+  { params }: { params: Promise<{ slug: string }> }
```
icyJoseph: `RouteContext` - https://nextjs.org/docs/app/api-reference/file-conventions/route#route-context-helper

docs/01-app/03-api-reference/04-functions/not-found.mdx
```diff
@@ -30,7 +60,156 @@ export default async function Profile({ params }) {
 }
 ```
 
-> **Good to know**: `notFound()` does not require you to use `return notFound()` due to using the TypeScript [`never`](https://www.typescriptlang.org/docs/handbook/2/functions.html#never) type.
+## Good to know
+
+- `notFound()` does not require `return notFound()`, because of its TypeScript [`never`](https://www.typescriptlang.org/docs/handbook/2/functions.html#never) return type.
```
icyJoseph: I've always felt that this is backwards, let's correct it:

```suggestion
## Good to know

You do not need to write `return notFound()`. Calling it is enough, because it throws an exception that stops function execution. TypeScript understands this from its [`never`](https://www.typescriptlang.org/docs/handbook/2/functions.html#never) return type, so a value you check first stays narrowed afterward:

```tsx
// fetchUser resolves to a user object, or undefined
const user = await fetchUser(id)

if (!user) {
  notFound()
}

// user is defined here
return <Profile user={user} />
```

Like any exception, it travels up the call stack until something catches it. A `try/catch` around the call swallows it, and the not-found UI won't render. If you need to catch errors near the call, use [`unstable_rethrow`](/docs/app/api-reference/functions/unstable_rethrow) to let the interrupt through first.
```

File: docs/01-app/03-api-reference/04-functions/not-found.mdx [hunk 4]
```diff
@@ -30,7 +60,156 @@ export default async function Profile({ params }) {
...
+        <Article slug={slug} />
+      </Suspense>
+    </section>
+  )
+}
+```
+
+Because `notFound()` is thrown during rendering, it propagates up to the nearest [`not-found`](/docs/app/api-reference/file-conventions/not-found) boundary and replaces the UI, even after the shell has started streaming.
+
+Add a `not-found.tsx` alongside the route to define that UI:
```
icyJoseph: Otherwise, the default `/_not_found` is used right? Not sure we should say it that way, but that's the behavior IIRC.

docs/01-app/03-api-reference/04-functions/not-found.mdx
```diff
@@ -30,7 +60,156 @@ export default async function Profile({ params }) {
...
+      <Link href="/blog">Blog</Link>
+      <Suspense fallback={<p>Loading...</p>}>
+        <Article slug={slug} />
+      </Suspense>
+    </section>
+  )
+}
+```
+
+Because `notFound()` is thrown during rendering, it propagates up to the nearest [`not-found`](/docs/app/api-reference/file-conventions/not-found) boundary and replaces the UI, even after the shell has started streaming.
```
icyJoseph:
```suggestion
When the post doesn't exist, `getPost` calls `notFound()`, which throws. Because this happens during rendering, the exception propagates to the nearest [`not-found`](/docs/app/api-reference/file-conventions/not-found) boundary, which renders in place of the streamed-in content, even though the page shell has already been sent.
```

docs/01-app/03-api-reference/04-functions/not-found.mdx
```diff
@@ -30,7 +60,156 @@ export default async function Profile({ params }) {
...
+import Link from 'next/link'
+import { notFound } from 'next/navigation'
+
+async function getPost(slug: string) {
+  const res = await fetch(`https://api.example.com/posts/${slug}`)
+  if (!res.ok) {
+    notFound()
+  }
+  return res.json()
+}
```
icyJoseph, line 101: Should we be more formal here 

````suggestion
async function getPost(slug: string) {
  const res = await fetch(`https://api.example.com/posts/${slug}`)
  if (res.status === 404) {
    notFound()
  }
  if (!res.ok) {
    throw new Error(`Failed to load post: ${res.status}`)
  }
  return res.json()
}
````

Just thinking that, the !res.ok would have to be catched by an error.tsx UI - too much detail maybe

icyJoseph:
Let's not forget to check

- redirect and permanentRedirect
- if we do use instant=false, does that have any effect on the 200/404 situation? x-ref: https://github.com/vercel/next.js/issues/95380

icyJoseph (approved):

Issue #95848, opened 2026-07-16
Blank page when pressing Back between a reload and hydration with cacheComponents (regression in 16.3.0-canary.87)


franky47:
### Link to the code that reproduces this issue

https://github.com/franky47/next-repro-back-before-hydration

### To Reproduce

1. `pnpm install && pnpm test` — a Playwright test reproduces the race deterministically by stalling static scripts (same technique as `test/e2e/app-dir/back-before-hydration` in this repo), and fails with a list of blank frames.
2. Or manually: `pnpm build && pnpm start`, click **To post**, hard-reload (⌘⇧R), then press Back quickly before hydration finishes. On fast machines, enable network throttling in devtools (e.g. Slow 4G) to widen the pre-hydration window: the HTML streams in quickly but the static chunks take a while, so the Back press can happen at leisure.

The setup needs three ingredients:

- `cacheComponents: true`
- a `<Suspense>` boundary around `children` in the root layout
- a Back press landing between a reload and hydration

### Current vs. Expected behavior

**Current:** about 400ms after hydration, the reloaded page's content is hidden (`display: none`), but the page the browser traversed back to never renders. The URL bar shows the traversed-to URL over a blank page, and it usually never recovers (the test observes for 5s; I've watched it stay blank for 20s+). The failure is racy on fast machines — the router occasionally recovers cleanly — so the test runs the scenario 3 times.

**Expected:** the traversal replay introduced in #95682 recovers, as it does without `cacheComponents`: the traversed-to page renders, and the reloaded page's content stays visible until then.

This is a regression introduced by #95682 (16.3.0-canary.87):

- On 16.3.0-canary.86, the same scenario shows the reloaded page's content under the traversed-to URL — the URL/content desync that #95682 fixes — but never a blank page.
- On 16.3.0-canary.87 without `cacheComponents`, the replay works correctly.
- The `back-before-hydration` e2e suite added in #95682 doesn't run under `cacheComponents`, which is presumably why this wasn't caught.

### Provide environment information

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0
  Available memory (MB): 131072
  Available CPU cores: 16
Binaries:
  Node: 24.6.0
  npm: 11.16.0
  Yarn: 1.22.22
  pnpm: 11.10.0
Relevant Packages:
  next: 16.3.0-canary.87 // Latest available version is detected (16.3.0-canary.87).
  eslint-config-next: N/A
[...]

PR #95853, opened 2026-07-16
Revert "Replay same-document traversals that happen before hydration"

gaearon:
Reverts vercel/next.js#95682

This breaks CC (https://github.com/vercel/next.js/issues/95848) due to PPR codepath doing something different. Although Claude says https://github.com/vercel/next.js/commit/da4d63d759a9bffd3e8c586bac1e4dabffd5b35d can work around it, I think it's safest to revert this given that I don't know what the right full fix is.

I also think we need more test coverage because the navigation test suite is off in CC mode.

I'll send a follow-up with all the failing tests we know of (from #95682 and new) so we can fix this for good after a closer look.


timneutkens (approved):

gaearon:
Turns out it's a React bug. https://github.com/react/react/pull/37135

PR #95854, opened 2026-07-16
Test async modules deployment

devjiwonchoi:
### Why?

Check whether the async modules deployment test still fails on the latest canary.

### How?

Replay the existing async modules deployment diagnostic on the latest canary.


devjiwonchoi:

PR #95855, opened 2026-07-16
Test middleware-general deployment

devjiwonchoi:
### Why?

Check whether the middleware-general deployment test still fails on the latest canary.

### How?

Replay the existing middleware-general deployment diagnostic on the latest canary.


devjiwonchoi:

PR #95856, opened 2026-07-16
Test skip-trailing-slash-redirect deployment

devjiwonchoi:
### Why?

Check whether the skip-trailing-slash-redirect deployment test still fails on the latest canary.

### How?

Replay the existing skip-trailing-slash-redirect deployment diagnostic on the latest canary.


devjiwonchoi:

PR #95877, opened 2026-07-16
[test] Run the navigation e2e suite under cacheComponents

gaearon:
This test suite used to be disabled in CC runs, which feels like a bit of a hazard for developing CC.

Let's enable it.

PR #95878, opened 2026-07-16
Run more test suites under cacheComponents flag

gaearon:
Similar to https://github.com/vercel/next.js/pull/95877.

Some of these need loading boundaries so that CC doesn't complain about being blocking.


gnoff (approved):
Should we add loading states or just disable instant by default?

gaearon:
I tried adding `export const instant = false` but it's a build failure pre-CC? Not sure if there's a nice way to do this in tests that run for both modes.

PR #95896, opened 2026-07-17
docs: reframe prefetching around automatic vs. controllable behavior

aurorascharff:
## Why

Follow-up to feedback on the 16.3 partial-prefetching docs. The blog's "Fewer requests and lower bandwidth from visible Links" is an automatic, framework-level benefit, not a dev-facing knob, yet the prefetching docs still lead with the pre-Cache-Components model and never explain the shared-shell behavior that produces it.

## What changed

- **`getting-started/linking-and-navigating`**: added an opt-in Cache Components callout to the prefetching section. `<Link>` prefetches a per-route App Shell, and many links to the same route reuse one shell, so a page with many links makes fewer prefetch requests. The default (non-Cache-Components) model description is unchanged.
- **`guides/prefetching`**: restructured into two halves:
  - **What Next.js prefetches automatically** (static vs. dynamic, automatic prefetch, the App Shell model, scheduling, client cache)
  - **Controlling prefetching** (manual, hover, extend/eject, disable)

  Promoted and expanded the Cache Components / App Shell section as the home for the shared-shell / fewer-requests story, and scoped the legacy `staleTimes` tables to non-Cache-Components apps. All content-bearing anchors are preserved (including `#extending-or-ejecting-link`, the only one linked from other docs). Prose swept against the Vercel technical writing style guide.

Docs-only. Related: DOC-6496.


docs/01-app/01-getting-started/04-linking-and-navigating.mdx
icyJoseph, line 97: After this image and before the closing sentence, we should add:

> Good to know: See the Prefetching guide (/docs/app/guides/prefetching) for the full behavior, including how to configure the default prefetching. Prefetching also changes when you opt in to Cache Components (/docs/app/api-reference/config/next-config-js/cacheComponents), an improved rendering mode for the App Router. To understand that first, start with Caching (/docs/app/getting-started/caching).

And remove the callout from before the Image (Line 90).

docs/01-app/02-guides/prefetching.mdx
```diff
@@ -23,7 +22,13 @@ Next.js automatically splits your application into smaller JavaScript chunks bas
 
 When navigating to the new page, there's no full page reload or browser loading spinner. Instead, Next.js performs a [client-side transition](/docs/app/getting-started/linking-and-navigating#client-side-transitions), making the page navigation feel instant.
 
-## Prefetching static vs. dynamic routes
+## What Next.js prefetches automatically
+
+Next.js prefetches automatically in production. As each `<Link>` enters the viewport, Next.js prefetches the route behind it, then schedules and reuses those requests so a page full of links doesn't flood the network. What gets prefetched depends on whether [Cache Components](/docs/app/getting-started/caching) is enabled.
```
icyJoseph:
```suggestion
Next.js prefetches automatically in production. As each `<Link>` enters the viewport, Next.js prefetches the route behind it and schedules the work so a page full of links doesn't flood the network. How much of each route it prefetches depends on whether the route is static or dynamic, and changes when [Cache Components](/docs/app/getting-started/caching) is enabled.
```

Although now it is about `partialPrefetching` (which reads more naturally too), right?

docs/01-app/02-guides/prefetching.mdx
```diff
@@ -58,9 +63,38 @@ export default function NavLink() {
 
 Automatic prefetching runs only in production. Disable with `prefetch={false}` or use the wrapper in [Disabled Prefetch](#disabled-prefetch).
 
-## Manual prefetch
+### Cache Components
```
icyJoseph: Could we re-organize like from:

```
static vs. dynamic  (default)
automatic prefetch  (default)
Cache Components    (opt-in)
prefetch scheduling (general)
client cache        (general)
```

To:

```
static vs. dynamic
automatic prefetch
prefetch scheduling
client cache
Cache Components     ← opt-in, at the end
```

icyJoseph (approved):

PR #95902, opened 2026-07-17
[turbopack] Reduce macOS HMR latency

marcoshernanz:
## Summary

- supplement macOS FSEvents with a bounded per-file `kqueue` watcher for project files that Turbo Tasks has actually read, while keeping FSEvents as the authoritative recursive watcher
- reduce intrinsic Turbo Tasks overhead by reusing the existing worker for sequential follow-up work on single-worker runtimes and by skipping redundant child-connection bookkeeping for materialized top-level cached roots
- add scheduler, cached-root invalidation, and watcher coverage; make all active `turbo-tasks` doctests compile as part of the exact crate test

The macOS watcher sidecar excludes `.git`, `.next`, and `node_modules`, is capped at 2,048 files, and feeds the existing invalidation channel. The scheduler grace period is also deliberately bounded: after one normal self-wake it performs at most 16 unlocked OS yields, and only when the runtime has one worker. Multi-worker runtimes retain the immediate queue-pop path.

## Findings

macOS sampling of the unchanged `task_overhead` benchmark showed that minimal task latency was dominated by orchestration rather than task-body work: `Runtime::block_on`, worker creation and handoff, Tokio remote-queue pushes, parked-worker notification, wake/unpark, cache lookup, and connect-child aggregation were the recurring hotspots.

The retained implementation addresses two of those costs without weakening bookkeeping:

1. Sequential single-worker chains get a short opportunity to enqueue their next task onto the same pinned `WorkerFuture`, avoiding a teardown/spawn/wake cycle.
2. A cache hit for a top-level task that is already an aggregation root and already has output returns directly instead of running a redundant `ConnectChildOperation`. Dirty roots still take the normal execution path.

Fifty measured candidates covered scheduling/wakes, allocation and initialization, registration/lookup, hashing/maps, locks/refcounts, aggregation, and completion state. Two tempting alternatives were rejected: an atomic busy-spin occupied a Tokio worker, and applying the yield grace to multi-worker runtimes regressed `turbo-uncached-parallel/1` by 6.372% with a 95% CI of `[+5.410%, +11.502%]`. Moving work onto a coordinator was faster in isolation but unsafe because it changed task-local, `LocalSet`, cancellation, and lifetime behavior.

## Benchmark results

The benchmark source and configuration were unchanged. The primary metric was Criterion slope `E = turbo-uncached/1 - direct/1`; lower is better. The frozen baseline executable was compared adjacently with the candidate in alternating order in independent Linux sandbox sessions.

| Validation | Baseline E | Candidate E | Reduction | 95% bootstrap CI for C/B | Candidate U wins | Max control drift |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Batch A | 18,121.549 ns | 14,276.488 ns | 21.218% | `[0.785446, 0.869037]` | 5/5 | D 0.603%, Tokio 0.714% |
| Batch B | 21,421.499 ns | 15,333.384 ns | 28.421% | `[0.701792, 0.746936]` | 5/5 | D 0.931%, Tokio 0.861% |
| Combined median | 19,418.916 ns | 14,903.762 ns | 23.251% | `[0.711160, 0.802896]` | 10/10 | D 0.931%, Tokio 0.861% |

Combined medians were:

- baseline: D 1,056.735 ns, U 20,474.332 ns, E 19,418.916 ns
- candidate: D 1,058.013 ns, U 15,963.206 ns, E 14,903.762 ns

The complete guardrail benchmark showed `turbo-cached-same-keys` improving by 35-49% and `turbo-cached-different-keys` improving or remaining neutral across durations. `turbo-uncached-parallel` was noisy, but no fixed duration produced a reproducible statistically significant regression larger than 3%. Focused repeats were -0.319% for duration 1 with CI `[-2.795%, +2.232%]` and +2.735% for duration 10 with CI `[-0.875%, +6.372%]`.

These are production paths, not benchmark special cases: the benchmark source, runtime construction, task bodies, worker counts, toolchain, profiles, and feature flags are unchanged.

## Verification

- `cargo test -p turbo-tasks` (79 unit tests and 31 active doctests passed; 12 pre-existing intentional doctest ignores remain)
- `cargo test -p turbo-tasks-fs` (111 tests passed)
- `cargo bench -p turbo-tasks-backend --bench mod --no-run --message-format=json-render-diagnostics`
- two independent five-pair final validation batches plus the complete `task_overhead` guardrail benchmark
- pre-commit `prettier` and `rustfmt` checks
- independent pre-ship review by two reviewers, with no findings
- Not run: `pnpm --filter=next dev` (Rust-only Turbo Tasks scheduler/backend/filesystem changes; Cargo tests and benchmarks are the authoritative checks)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10193,6 +10193,8 @@ dependencies = [
  "include_dir",
  "indexmap 2.13.0",
  "jsonc-parser",
+ "kqueue-sys",
+ "libc",
  "mime",
  "notify",
  "parking_lot",
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -65,8 +65,8 @@ use crate::{
             AggregationUpdateJob, AggregationUpdateQueue, ChildExecuteContext,
             CleanupOldEdgesOperation, ConnectChildOperation, ExecuteContext, ExecuteContextImpl,
             LeafDistanceUpdateQueue, Operation, OutdatedEdge, TaskGuard, TaskType, TaskTypeRef,
-            connect_children, get_aggregation_number, get_uppers, make_task_dirty_internal,
-            prepare_new_children,
+            connect_children, get_aggregation_number, get_uppers, is_root_node,
+            make_task_dirty_internal, prepare_new_children,
         },
         snapshot_coordinator::{OperationGuard, SnapshotCoordinator},
         storage::Storage,
@@ -1516,6 +1516,12 @@ impl TurboTasksBackend {
             raw_get_in_shard(shard, hash, |k| k.eq_components(native_fn, this, arg_ref))
         {
             self.track_cache_hit_by_fn(native_fn);
+            if parent_task.is_none() {
+                let task = ctx.task(task_id, TaskDataCategory::Meta);
+                if is_root_node(get_aggregation_number(&task)) && task.has_output() {
+                    return task_id;
+                }
+            }
             operation::ConnectChildOperation::run(parent_task, task_id, ctx);
             return task_id;
         }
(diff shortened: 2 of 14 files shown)


marcoshernanz:
Superseded by #95903, which contains only the unique Turbo Tasks intrinsic-overhead change. The macOS watcher change and its findings remain in #95839.

PR #95903, opened 2026-07-17
[turbopack] Skip redundant top-level root updates

marcoshernanz:
## Summary

- skip the redundant aggregation-root update when `ConnectChildOperation` reconnects an already constructed top-level task that is already an aggregation root
- avoid entering the aggregation operation loop when no update was queued

The fast path is intentionally narrow. It applies only when the connection has no parent, the task has materialized output, and the task's aggregation number is already the root sentinel. The task may still be dirty and need to run; this only skips an idempotent aggregation update. A task that is not yet a root still goes through the normal update on its first top-level connection. Tasks without output are still scheduled, parented connections retain their activeness and aggregation updates, and transient-state boundaries are unchanged.

## Findings

The earlier scheduler experiment has been removed completely.

The apparent benefit from keeping a worker alive with `std::thread::yield_now()` was specific to the benchmark topology: Criterion's `Runtime::block_on` caller runs outside the single-worker Tokio runtime, so that external caller can enqueue the next root task while the runtime worker is yielding at the OS level. The sole Tokio worker itself cannot produce work during that loop. In normal scheduling, tasks typically enqueue their children before the current job exits, and a multi-worker runtime can start another worker when it is below its target. That does not justify burning CPU in an idle worker, nor replacing the loop with an async-yield retry loop.

The retained optimization is independent of that artifact. An already constructed top-level task that is already rooted has no new root update to perform; re-enqueuing the idempotent update only forces the operation executor through unnecessary suspend points.

This PR now contains no changes to `priority_runner`, worker lifetime, busy waiting, OS-thread yielding, async yielding, HMR, filesystem watching, or documentation.

## Benchmark results

The benchmark source, runtime construction, profiles, allocator, toolchain, and worker count were unchanged. Frozen baseline and candidate executables were run adjacently in alternating order. Three focused pairs met the control-drift threshold; one pair was excluded before analysis because its Tokio control moved by more than 2%.

Median candidate changes across the valid pairs:

| Case | Change |
| --- | ---: |
| direct control | +0.071% |
| Tokio control | -0.177% |
| uncached task | +1.266% |
| cached, same key | -37.868% |
| cached, different keys | -26.783% |

The complete all-duration local sweeps did not meet the control thresholds, so their task measurements are deliberately not cited. Fresh remote CodeSpeed and Stats results are the authoritative guardrail for the full benchmark.

## Verification

- `cargo clippy -p turbo-tasks-backend --all-targets -- -D warnings -A deprecated`
- `pnpm build-all`
- two-engine autoreview after the final changes, with no findings
- Not run: `pnpm --filter=next dev` (the final diff is Rust-only and was covered by the full bootstrap build and Cargo suites)

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -65,8 +65,8 @@ use crate::{
             AggregationUpdateJob, AggregationUpdateQueue, ChildExecuteContext,
             CleanupOldEdgesOperation, ConnectChildOperation, ExecuteContext, ExecuteContextImpl,
             LeafDistanceUpdateQueue, Operation, OutdatedEdge, TaskGuard, TaskType, TaskTypeRef,
-            connect_children, get_aggregation_number, get_uppers, is_root_node,
-            make_task_dirty_internal, prepare_new_children,
+            connect_children, get_aggregation_number, get_uppers, make_task_dirty_internal,
+            prepare_new_children,
         },
         snapshot_coordinator::{OperationGuard, SnapshotCoordinator},
         storage::Storage,
@@ -1516,12 +1516,6 @@ impl TurboTasksBackend {
             raw_get_in_shard(shard, hash, |k| k.eq_components(native_fn, this, arg_ref))
         {
             self.track_cache_hit_by_fn(native_fn);
-            if parent_task.is_none() {
-                let task = ctx.task(task_id, TaskDataCategory::Meta);
-                if is_root_node(get_aggregation_number(&task)) && task.has_output() {
-                    return task_id;
-                }
-            }
             operation::ConnectChildOperation::run(parent_task, task_id, ctx);
             return task_id;
         }
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -65,8 +65,8 @@ use crate::{
             AggregationUpdateJob, AggregationUpdateQueue, ChildExecuteContext,
             CleanupOldEdgesOperation, ConnectChildOperation, ExecuteContext, ExecuteContextImpl,
             LeafDistanceUpdateQueue, Operation, OutdatedEdge, TaskGuard, TaskType, TaskTypeRef,
-            connect_children, get_aggregation_number, get_uppers, make_task_dirty_internal,
-            prepare_new_children,
+            connect_children, get_aggregation_number, get_uppers, is_root_node,
+            make_task_dirty_internal, prepare_new_children,
         },
         snapshot_coordinator::{OperationGuard, SnapshotCoordinator},
         storage::Storage,
@@ -1516,6 +1516,12 @@ impl TurboTasksBackend {
             raw_get_in_shard(shard, hash, |k| k.eq_components(native_fn, this, arg_ref))
         {
             self.track_cache_hit_by_fn(native_fn);
+            if parent_task.is_none() {
+                let task = ctx.task(task_id, TaskDataCategory::Meta);
+                if is_root_node(get_aggregation_number(&task)) && task.has_output() {
+                    return task_id;
+                }
+            }
             operation::ConnectChildOperation::run(parent_task, task_id, ctx);
             return task_id;
         }
--- a/turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
@@ -159,3 +159,73 @@ impl Operation for ConnectChildOperation {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::sync::atomic::{AtomicUsize, Ordering};
+
+    use turbo_tasks::{TaskId, TurboTasks};
+
+    use super::*;
+    use crate::{
+        BackendOptions, TurboTasksBackend, backend::operation::ExecuteContextImpl,
+        data::OutputValue, noop_backing_storage,
+    };
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn completed_top_level_root_skips_aggregation_operation() {
+        let tt = TurboTasks::new(TurboTasksBackend::new(
+            BackendOptions {
+                num_workers: Some(1),
+                small_preallocation: true,
+                storage_mode: None,
+                ..Default::default()
+            },
+            noop_backing_storage(),
+        ));
+        let backend = tt.backend();
+        let child_task_id = TaskId::new(1).unwrap();
+        backend.storage.initialize_new_task(child_task_id, None);
+        backend
+            .storage
+            .access_mut(child_task_id)
+            .set_output(OutputValue::Output(child_task_id));
+
+        // The first top-level connection must promote the non-root task.
+        ConnectChildOperation::run(None, child_task_id, ExecuteContextImpl::new(backend, &tt));
+        let mut ctx = ExecuteContextImpl::new(backend, &tt);
+        let child_task = ctx.task(child_task_id, TaskDataCategory::Meta);
+        assert!(is_root_node(get_aggregation_number(&child_task)));
+        drop(child_task);
+        drop(ctx);
+
+        // The subsequent completed-root connection must not dispatch an empty aggregation
+        // operation, whose execute loop would otherwise visit at least one suspend point.
+        let suspend_points = AtomicUsize::new(0);
+        ConnectChildOperation::run(
+            None,
+            child_task_id,
+            ExecuteContextImpl::new(backend, &tt)
+                .with_operation_suspend_point_counter(&suspend_points),
+        );
+        assert_eq!(suspend_points.load(Ordering::Relaxed), 0);
+
+        // Prove the observer is active by explicitly executing the redundant update we skipped.
+        let mut queue = AggregationUpdateQueue::new();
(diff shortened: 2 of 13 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1516,6 +1516,12 @@ impl TurboTasksBackend {
             raw_get_in_shard(shard, hash, |k| k.eq_components(native_fn, this, arg_ref))
         {
             self.track_cache_hit_by_fn(native_fn);
+            if parent_task.is_none() {
+                let task = ctx.task(task_id, TaskDataCategory::Meta);
+                if is_root_node(get_aggregation_number(&task)) && task.has_output() {
+                    return task_id;
+                }
+            }
```
lukesandberg: ConnectChildOperation almost does this but i perceive a problem

ConnectChildOperation::execute always suspends before processing any part of the queue, but executing with an empty queue is common.  So ConnectChildOperation::run should probably just skip running execute if the queue is empty

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -314,11 +357,15 @@ impl<
...
                         *this.state = WorkerState::UnfinishedFuture;
                     } else {
-                        // No more tasks to execute
-                        // This worker ends here
-                        this.runner.decrease_active_workers(this.execute_context);
-                        *this.state = WorkerState::Closed;
-                        return Poll::Ready(());
+                        if matches!(this.state, WorkerState::Done) {
+                            cx.waker().wake_by_ref();
+                            *this.state = WorkerState::Yielded;
```
lukesandberg: is the yielded state really necessary?

why not just always do the busy loop?

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
marcoshernanz: :robot: Done -- I moved the completed top-level-root fast path into ConnectChildOperation::run and now skip execute when the aggregation queue is empty, so an empty queue no longer reaches the unconditional suspension point. The skip is deliberately limited to a top-level task that is already a root and has output; dirty/incomplete roots and transient/active boundaries retain the normal aggregation-update path. The cached-root invalidation regression still passes, and I also fixed the Clippy failure in that test.

turbopack/crates/turbo-tasks/src/priority_runner.rs
marcoshernanz: :robot: I dug into this and kept Yielded because the first handoff must be to Tokio, not only to the OS scheduler. On a one-worker runtime, std::thread::yield_now() cannot poll a follow-up producer that is waiting on that same Tokio worker; entering the grace loop immediately makes the worker close and forces each sequential follow-up to spawn a new WorkerFuture. I documented that invariant and added test_single_worker_yields_to_runtime_for_sequential_follow_up, which runs the producer on the same one-worker runtime and asserts the yielded worker remains available. The existing test still covers a concurrent producer on another OS thread.

marcoshernanz:
Update after Luke's follow-up review:\n\nThe scheduler/yield experiment described in the original version of this comment has been removed completely. The measured reuse came from the benchmark topology: an external Criterion block_on caller could enqueue work while the sole Tokio runtime worker performed OS-thread yields. That does not generalize to work produced by the same single-worker runtime, and normal task execution generally queues children before the current job exits. The PR no longer changes priority_runner, worker lifetime, or yielding behavior.\n\nThe remaining change is the narrowly scoped ConnectChild already-rooted fast path. It preserves first-time root promotion and dirty/incomplete behavior, while avoiding an idempotent aggregation update and its operation suspend points for an already constructed top-level task. At Luke's request, the added tests and suspend-point instrumentation were also removed; the effective PR diff is now one production file.\n\nCurrent implementation: af4724b64a.\n\n

turbopack/crates/turbo-tasks/src/priority_runner.rs
lukesandberg: on a single worker runtime the yielding approach will also just burn cpu waiting for nothing

in such a situation the thread::yields are always harmful (since no work can arrive)

so we could either structure our yields to just be async friendly?

but i guess i am suspicious that this optimization is _generally_ useful maybe this is an odd artifact of the benchmark timing loop.  typically tasks spawn each other so tasks are pushed before the job is exited, so the worker can essentially always find work when it completes a job

marcoshernanz: :robot: You were right — my previous reply established why the first handoff had to go back to Tokio, but it did not establish that the following OS-yield loop was generally useful. I traced the exact benchmark topology: Criterion Runtime::block_on is called from outside the one-worker Tokio runtime, so that external caller can enqueue the next root task while the runtime worker is in std::thread::yield_now(). The sole Tokio worker itself cannot produce work during that loop. In normal scheduling, tasks generally enqueue children before the current job exits, and a multi-worker runtime can start another worker when it is below its target. The improvement was therefore an artifact of this sequential benchmark handoff, not a sound general scheduler optimization.\n\nI removed the scheduler change completely, including WorkerState::Yielded, the OS-yield loop, and its tests. I did not replace it with async yields, because retrying an empty queue would still need an arbitrary lifetime policy and is not justified by the production scheduling path.\n\nThe PR now contains only the ConnectChild completed-root fast path. It skips the root update only for parent=None plus materialized output plus an aggregation number already at the root sentinel; first-time root promotion, incomplete scheduling, parent activeness, and dirty reexecution retain the normal paths. I added a direct test that first promotes a completed non-root task, then observes zero operation suspend points on its second top-level connection. Forcing the old unconditional update makes that test fail with three visits, and the strong-consistency integration test separately covers clean reuse and dirty reexecution.\n\nImplemented in ea827fb539 and 0f1afa83cd.\n\n

turbopack/crates/turbo-tasks-backend/tests/basic.rs
```diff
@@ -20,6 +23,60 @@ async fn test_basic() {
     .unwrap()
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_cached_top_level_root_fast_path() {
```
lukesandberg: drop this test and the corresponding suspend point instrumentation

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -62,22 +64,39 @@ impl ConnectChildOperation {
...
+            }
             queue.push(AggregationUpdateJob::IncreaseActiveCount {
                 task: child_task_id,
             });
         } else {
             let mut child_task = ctx.task(child_task_id, TaskDataCategory::Meta);
+            let has_output = child_task.has_output();
+            // A completed top-level task was already made a root when it was first connected.
+            // Repeating that idempotent update would only make the queue non-empty and force an
+            // otherwise unnecessary operation suspend point on every cache hit.
```
lukesandberg: nit, not completed but already constructed. it is only _possibly_ completed

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -138,3 +159,73 @@ impl Operation for ConnectChildOperation {
...
+    use turbo_tasks::{TaskId, TurboTasks};
+
+    use super::*;
+    use crate::{
+        BackendOptions, TurboTasksBackend, backend::operation::ExecuteContextImpl,
+        data::OutputValue, noop_backing_storage,
+    };
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn completed_top_level_root_skips_aggregation_operation() {
```
lukesandberg: i don't really think this test is too valuable

lukesandberg (approved):
Address the remaining feedback

turbopack/crates/turbo-tasks-backend/tests/basic.rs
marcoshernanz: :robot: Done in af4724b64a. I removed this integration test together with the direct unit test and all suspend-point counter instrumentation. The effective PR diff is now only the production ConnectChildOperation change.\n\n

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
marcoshernanz: :robot: Fixed in af4724b64a. The comment and local name now say already constructed rather than completed, and the comment explicitly notes that the task may still be dirty and need to run; the fast path only skips the idempotent aggregation update.\n\n

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
marcoshernanz: :robot: Removed in af4724b64a, along with the integration test and the ExecuteContext suspend-point instrumentation. The final PR has no added test-only machinery.\n\n

marcoshernanz:
@lukesandberg The remaining feedback is now applied on the current head, `af4724b64a`:

- the scheduler/yield experiment is removed completely
- `ConnectChildOperation` skips execution when no aggregation work was queued
- the added tests and suspend-point instrumentation are removed
- the implementation and comment use “already constructed” rather than “completed” and explicitly preserve dirty-task execution
- the effective diff is limited to the production `connect_child.rs` change

All review threads are resolved and CI is green on this head. When you have a moment, could you give the current head one final look/ack before merge?

lukesandberg (approved):

lukesandberg:
looks good

marcoshernanz:
:D

PR #95906, opened 2026-07-17
[Dev] Detect module-scoped state leaking across requests

gaojude:
Prototype/RFC for the guardrail asked for in [this thread](https://vercel.slack.com/archives/C0BGEL6UDPW/p1783638263690289?thread_ts=1783624971.544149&cid=C0BGEL6UDPW). A reused server function instance that keeps a top-level mutable container (`const cache = new Map()`) holding a per-request `Promise`, keyed deterministically (e.g. `useId()`), can serve one request's data to another. This is the v0 incident: a `useId()`-keyed module `Map` returned one user's sidebar chats to another (writeup: `chat/app/lib/server-query-swr/state-lifetime-history.md`).

The unwrap can't be intercepted, so the detector watches the container instead. `await` on a settled promise bypasses a patched `Promise.prototype.then`, and async_hooks exposes no cross-store consume event:

```
patch Promise.prototype.then; await Promise.resolve(42)
  → then() call count: 0            # await bypasses the JS-level then
B awaits A's already-settled promise, under async_hooks
  → before / promiseResolve fire only for B's own promises; no A→B signal
```

So `installCrossRequestStateDetector` (dev-only) patches `Map`/`Set`/`WeakMap`/`WeakSet`: a `Promise` stored under request A's `workUnitAsyncStorage` store and later read under request B's store warns. Only thenable-valued entries are ever tracked, so ordinary module caches of plain data never warn. Trace against the v0 shape:

```
const m = new Map()                                        # module scope
request A (/chat/alice): m.set('_r_0_', fetchSidebar('alice'))  # entry owner = A
request B (/chat/bob):   m.get('_r_0_') → alice's promise       # owner A != reader B -> WARN
```

Opt-in via `__NEXT_DETECT_CROSS_REQUEST_STATE` and gated to `__NEXT_DEV_SERVER`, because it patches global prototypes — never installed during `next build`, prerender, or `next start`. Detection is wrapped so a detector bug can never break user code, and internal bookkeeping uses pre-patch prototype methods to avoid re-entering the patches. The colocated unit test covers the leak (Map + `WeakSet<Promise>`), the `React.cache()`-style per-request container, a plain-data module cache, and same-request re-read.

Not yet validated against a real `next dev` process or burned in on a large app; global `Map`/`Set` patching in the server has a perf and false-positive surface worth measuring before this moves off opt-in.

### DIFF
--- a/packages/next/src/server/node-environment-extensions/cross-request-state-detector.test.ts
+++ b/packages/next/src/server/node-environment-extensions/cross-request-state-detector.test.ts
@@ -1,0 +1,121 @@
+import { AsyncLocalStorage } from 'node:async_hooks'
+import {
+  installCrossRequestStateDetector,
+  type WorkUnitStoreLike,
+} from './cross-request-state-detector'
+
+type FakeStore = { type: string; url: { pathname: string } }
+
+function setup() {
+  const als = new AsyncLocalStorage<FakeStore>()
+  const warnings: string[] = []
+  const uninstall = installCrossRequestStateDetector({
+    getStore: () => als.getStore() as WorkUnitStoreLike | undefined,
+    warn: (m) => warnings.push(m),
+  })
+  // fresh object per request => distinct identity, like a real RequestStore
+  const request = <T>(pathname: string, fn: () => Promise<T>): Promise<T> =>
+    als.run({ type: 'request', url: { pathname } }, fn)
+  return { warnings, uninstall, request }
+}
+
+async function fetchUserSidebar(user: string): Promise<string> {
+  return `sidebar-data-for-${user}`
+}
+
+describe('cross-request state detector', () => {
+  it('warns AND reproduces the leak: a module-scoped Map<useId, Promise>', async () => {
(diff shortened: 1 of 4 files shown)


gaojude:
Closing this. The discussion moved toward catching this class statically rather than at runtime.

The runtime detector here only covers `Map`/`Set`/`WeakMap`/`WeakSet`, because it hooks their prototype methods. It cannot see the other shapes of the same bug: a top-level `let` that gets reassigned, or a plain object or array used as a cache (`obj[key] = ...`). Those have no callable seam to patch, as the `async_hooks` and `Promise.then` probes showed.

A static AST pass sees all of those at author time, including the ones the runtime patch is blind to, and it runs in CI without patching globals. Moving the work to a small standalone tool (next-doctor) that pairs an ast-grep candidate pass with an AI judge for the per-request-data and key-collision call.

The one thing the runtime detector did that static cannot is confirm an actual cross-request read with both request URLs. That can return later as an optional confirmation tier that feeds off the static findings. Keeping the branch for reference.

PR #95910, opened 2026-07-17
Fix the queue

lukesandberg:
## What

Fixes duplicate compilation events in the dev/build trace — most visibly the `turbopack-persistence` `TraceEvent`, which showed up as two byte-identical spans in `.next/trace`.

## Why

`CompilationEventQueue` had a register/replay race. `send` and `subscribe` touched `history` and the `subscribers` map under **separate** locks, in opposite orders. A subscriber caught between "registered" and "history replayed" could receive the same event **both live and via replay**. Because both paths forward the same `Arc`, the two spans were identical. Persistence events are the common victim: they fire unconditionally on every snapshot, including shortly after startup (idle-timeout snapshot), right in the subscribe window.

## How

Put `history` and `subscribers` under a single mutex and do each side's pair of operations in one critical section:

- `send`: push to history **and** snapshot the recipient channels, then deliver live after unlocking.
- `subscribe`: register the channel **and** snapshot history to replay, then replay after unlocking.

An event is in the history snapshot iff it is *not* in the subscriber snapshot → **exactly-once** delivery per subscriber.

## Notes

- Delivery order between events is **not** guaranteed (each `send` runs on its own spawned task) — this was already the case and nothing depends on it (trace spans self-order by embedded wall-clock timestamps). Documented on `send`.
- Adds unit tests, including a multi-thread stress test that reproduces the original double-delivery.

### DIFF
--- a/turbopack/crates/turbo-tasks/src/message_queue.rs
+++ b/turbopack/crates/turbo-tasks/src/message_queue.rs
@@ -1,6 +1,11 @@
-use std::{any::Any, collections::VecDeque, fmt::Display, sync::Arc, time::Duration};
+use std::{
+    any::Any,
+    collections::{HashMap, VecDeque},
+    fmt::Display,
+    sync::Arc,
+    time::Duration,
+};
 
-use dashmap::DashMap;
 use serde::Serialize;
 use tokio::sync::{Mutex, mpsc};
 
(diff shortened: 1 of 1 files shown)


lukesandberg:
* **#95910** <a href="https://app.graphite.com/github/pr/vercel/next.js/95910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

PR #95911, opened 2026-07-17
Run Rust doctests in CI

marcoshernanz:
## Summary

Run all Rust workspace doctests as part of the existing Turbo-cached cargo unit test task so future documentation regressions fail CI without adding an uncached workspace build.

## Verification

- `cargo test --workspace --doc --profile=release-with-assertions --no-fail-fast`
- Not run: `pnpm --filter=@next/swc test-cargo-unit` (`cargo-nextest` is installed in CI but unavailable locally; the new doctest suffix passed independently)

### DIFF
--- a/packages/next-swc/package.json
+++ b/packages/next-swc/package.json
@@ -19,7 +19,7 @@
     "rust-check-doc": "RUSTDOCFLAGS='-Zunstable-options --output-format=json' cargo doc --no-deps --workspace",
     "rust-check-fmt": "cd ../..; cargo fmt -- --check",
     "rust-check-napi": "cargo check -p next-napi-bindings",
-    "test-cargo-unit": "cargo nextest run --workspace --exclude next-napi-bindings --exclude turbo-tasks-macros --cargo-profile release-with-assertions --no-fail-fast"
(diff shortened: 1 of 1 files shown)


packages/next-swc/package.json
```diff
@@ -19,7 +19,7 @@
     "rust-check-doc": "RUSTDOCFLAGS='-Zunstable-options --output-format=json' cargo doc --no-deps --workspace",
     "rust-check-fmt": "cd ../..; cargo fmt -- --check",
     "rust-check-napi": "cargo check -p next-napi-bindings",
-    "test-cargo-unit": "cargo nextest run --workspace --exclude next-napi-bindings --exclude turbo-tasks-macros --cargo-profile release-with-assertions --no-fail-fast"
+    "test-cargo-unit": "cargo nextest run --workspace --exclude next-napi-bindings --exclude turbo-tasks-macros --cargo-profile release-with-assertions --no-fail-fast && cargo test --workspace --doc --profile=release-with-assertions --no-fail-fast"
```
bgw, line 22: I was going to ask why we don't run the doctests with `nextest` so that they can run in parallel with all the other tests, and found this:
https://nexte.st/docs/running/?h=doctest#fn:doctest

> Doctests are currently [not supported](https://github.com/nextest-rs/nextest/issues/16) because of limitations in stable Rust. For now, run doctests in a separate step with `cargo test --doc`.

😢

bgw (approved):

PR #95913, opened 2026-07-17
Avoid duplicate cached redirect headers

### LINKED ISSUE #82117: Duplicate Location and x-nextjs-stale-time in Header using redirect/permanentRedirect alongside dynamic 'force-static'

https://github.com/d-mon-/test-redirect

### To Reproduce

1. Build and run the server with `npm run build && npm run start`
2. Go to http://localhost:3000/test with the network console opened
3. You shall notice the Location header duplicated

Currently you will notice the `Location` and `x-nextjs-stale-time` fields to be duplicated in the header
<details>
  <summary>screenshot</summary>

Image

</details>

I'd expect to see only a single field for each of them. This is problematic as the user will be redirected to `/redirect,/redirect` instead of `/redirect` depending on how the `Location` value is processed in production

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6031
  Available memory (MB): 65536
  Available CPU cores: 16
Binaries:
  Node: 22.16.0
  npm: 11.4.2
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 15.4.4 // Latest available version is detected (15.4.4).
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.3
Next.js Config:
  output: N/A
```

Headers

next start (local)

### Additional context

The problem seems to happen since the **15.4.1** release (previous versions seem fine)

I identified these two blocks adding the headers:
the first call is doing a setHeader location (using [NodeNextResponse](https://github.com/vercel/next.js/blob/ad1e3f209b87b6cc4829b125a667e94f7193f62f/packages/next/src/server/base-http/node.ts#L114))
https://github.com/vercel/next.js/blob/ad1e3f209b87b6cc4829b125a667e94f7193f62f/packages/next/src/server/app-render/action-handler.ts#L1061
the second call is doing an appendHeader of `{location, x-nextjs-stale-time}`, duplicating the value. This one seems to use the native function this time instead of NodeNextResponse which doesn't check for the existing values.
https://github.com/vercel/next.js/blob/ad1e3f209b87b6cc4829b125a667e94f7193f62f/packages/next/src/build/templates/app-page.ts#L1010-L1031

A potential candidate: https://github.com/vercel/next.js/pull/73431

<sub>[NEXT-4749](https://linear.app/vercel/issue/NEXT-4749/duplicate-location-and-x-nextjs-stale-time-in-header-using)</sub>

Austin1serb:
**Summary**

Fix duplicate headers when a cached App Router redirect response is replayed.

**Root cause**

Cached response headers were appended directly to the native Node response. This bypassed the duplicate-value protection already implemented by `NodeNextResponse.appendHeader`.

As a result, headers such as `Location` and `x-nextjs-stale-time` could be emitted twice. Intermediaries that combine repeated `Location` headers could then produce an invalid value such as `/destination, /destination`.

**Changes**

- Replay cached headers through `NodeNextResponse.appendHeader`.
- Avoid appending identical values that are already present.
- Preserve distinct values for legitimate multi-value headers.

**Regression coverage**

The production fixture inspects raw Node response headers and covers:

- `dynamic = 'force-static'` with `permanentRedirect()` (`308`).
- `dynamic = 'force-static'` with `redirect()` (`307`).
- ISR fallback generation for a dynamic route whose requested parameter was not returned by `generateStaticParams()`.
- The ISR response on initial generation, cache reuse, and after the revalidation window.
- A single `Location` value and a single `x-nextjs-stale-time` value for every cached redirect response.

The expanded tests fail against the previous implementation with two raw `Location` values and pass with this fix.

**Additional validation**

- Reproduced the original issue's catch-all `force-static` example.
- Reproduced the reported Next.js 16.1.6 `cacheComponents: true` case using a cached, revalidatable dynamic fallback route. Its first response emitted two `Location` headers before this change and one after applying the same header-replay fix.
- Verified the response directly from `next start`, without a proxy. Emitting one raw `Location` header also prevents Cloudflare/Datadome-style comma merging and the malformed redirect observed by crawlers.

**Validation completed on current canary**

- Next.js package build and type generation.
- Webpack production test.
- Turbopack production test.
- Focused ESLint.
- Prettier.
- Signed commit and pre-commit hooks.

Fixes #82117

### DIFF
--- a/packages/next/src/build/templates/app-page-runtime.ts
+++ b/packages/next/src/build/templates/app-page-runtime.ts
@@ -1856,18 +1856,19 @@ export function createAppPageEntrypoint({
             delete headers[NEXT_CACHE_TAGS_HEADER]
           }
 
+          const nextRes = new NodeNextResponse(res)
           for (let [key, value] of Object.entries(headers)) {
             if (typeof value === 'undefined') continue
 
             if (Array.isArray(value)) {
               for (const v of value) {
-                res.appendHeader(key, v)
+                nextRes.appendHeader(key, v)
               }
             } else if (typeof value === 'number') {
               value = value.toString()
-              res.appendHeader(key, value)
+              nextRes.appendHeader(key, value)
             } else {
-              res.appendHeader(key, value)
+              nextRes.appendHeader(key, value)
             }
           }
         }
--- a/test/production/app-dir/cached-redirect-headers/app/[...slug]/page.tsx
+++ b/test/production/app-dir/cached-redirect-headers/app/[...slug]/page.tsx
@@ -1,7 +1,17 @@
-import { permanentRedirect } from 'next/navigation'
+import { permanentRedirect, redirect } from 'next/navigation'
 
 export const dynamic = 'force-static'
 
-export default function Page() {
+export default async function Page({
+  params,
+}: {
+  params: Promise<{ slug: string[] }>
+}) {
+  const { slug } = await params
+
+  if (slug[0] === 'temporary') {
+    redirect('/destination')
+  }
+
   permanentRedirect('/destination')
 }
--- a/test/production/app-dir/cached-redirect-headers/app/[...slug]/page.tsx
+++ b/test/production/app-dir/cached-redirect-headers/app/[...slug]/page.tsx
@@ -1,0 +1,7 @@
+import { permanentRedirect } from 'next/navigation'
+
+export const dynamic = 'force-static'
+
+export default function Page() {
+  permanentRedirect('/destination')
+}
--- a/test/production/app-dir/cached-redirect-headers/app/destination/page.tsx
+++ b/test/production/app-dir/cached-redirect-headers/app/destination/page.tsx
@@ -1,0 +1,3 @@
+export default function Page() {
+  return <p>destination</p>
+}
--- a/test/production/app-dir/cached-redirect-headers/app/isr/[slug]/page.tsx
+++ b/test/production/app-dir/cached-redirect-headers/app/isr/[slug]/page.tsx
@@ -1,0 +1,11 @@
+import { permanentRedirect } from 'next/navigation'
+
+export const revalidate = 1
+
+export function generateStaticParams() {
+  return [{ slug: 'prerendered' }]
+}
+
+export default function Page() {
+  permanentRedirect('/destination')
+}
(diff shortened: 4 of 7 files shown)


bobrosset:
Confirming this is still reproducible, with some production impact data and a detail that makes the bug easy to miss.

**Still present on:** `15.5.16`, `15.5.23`, `16.3.0` and `16.3.1-canary.14`.

**Minimal reproduction** — the only difference between the two routes is `generateStaticParams`:

```tsx
// app/isr/[id]/page.tsx  -> emits Location TWICE on a cache MISS
import { permanentRedirect } from 'next/navigation';
export const revalidate = 3600;
export function generateStaticParams(): { id: string }[] { return []; }
export default async function Page({ params }: { params: Promise<{ id: string }> }) {
  const { id } = await params;
  permanentRedirect(`/target/${id}`);
}

// app/dynamic/[id]/page.tsx -> emits it once (same file, no generateStaticParams)
```

```
$ curl -sD - -o /dev/null localhost:3000/isr/cold-1 | grep -ci '^location:'
2      # x-nextjs-cache: MISS

$ curl -sD - -o /dev/null localhost:3000/isr/cold-1 | grep -ci '^location:'
1      # x-nextjs-cache: HIT
```

**Why this is easy to miss:** it only breaks over HTTP/2. `next start` speaks HTTP/1.1, where the two headers stay separate and the client just picks one — the redirect appears to work. Behind HTTP/2 (or any proxy that folds repeated headers) the values collapse into `/target/x,/target/x` and the request 404s. So local testing looks clean, and only the **first** request per URL is affected — every later one is a cache HIT and redirects correctly, which erases the evidence as soon as you retry.

**Production impact**, from a marketplace running 15.5.16 where every listing redirect goes through a prerendered route:

- **~1,500–2,300 404s/day** on comma-joined paths, sustained.
- **~46% from real browser user agents** — these are visitors arriving from search results, shared links and mobile deep links, i.e. cold URLs by definition, which is exactly the traffic that hits the MISS path.

**Note for anyone looking for a workaround:** patching a published `next` package does not work. Porting this fix by hand into both `dist/build/templates/app-page.js` and `dist/esm/build/templates/app-page.js` changed nothing — tracing `setHeader` showed the code that actually runs comes from the precompiled `dist/compiled/next-server/app-page.runtime.prod.js`. The only workarounds we found are at the application level: drop `generateStaticParams` from routes that redirect, or move the redirect into middleware so it runs before the render.

Since the 15.5.x line is affected too, a backport would help a lot of people who cannot move to 16 yet.

PR #95920, opened 2026-07-18
fix: wire experimental.webVitalsAttribution into useReportWebVitals

### LINKED ISSUE #95919: experimental.webVitalsAttribution has no effect on useReportWebVitals

https://github.com/dayongkr/webvitals-attribution-repro

### To Reproduce

1. Clone https://github.com/dayongkr/webvitals-attribution-repro, then run `npm install && npm run dev`
2. Open http://localhost:3000 with the browser console open
3. Click anywhere on the page (flushes LCP/INP), then switch tabs once (flushes CLS)
4. Compare the `[useReportWebVitals]` logs with the `[web-vitals/attribution]` logs

Current: `experimental.webVitalsAttribution: ["CLS", "LCP", "FCP", "TTFB", "INP"]` is set in `next.config.ts`, but every metric reported by `useReportWebVitals` has `attribution: undefined`. On the same page, the `web-vitals/attribution` build used directly does populate it.

`next dev` also lists the option under "Experiments (use with caution)" on boot, so the config itself is accepted.

Expected: metrics listed in the option should include the `attribution` property.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:38:56 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 25.9.0
  npm: 11.12.1
  Yarn: 1.22.22
  pnpm: 11.7.0
Relevant Packages:
  next: 16.3.0-canary.89 // Latest available version is detected (16.3.0-canary.89).
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Performance

### Additional context

- Last version where the option had an effect: v14.2.x (only through the Pages Router `reportWebVitals` export in `_app`; `useReportWebVitals` has never used it)
- The code consuming the option (`performance-relayer.ts` / `performance-relayer-app.ts`) was removed in #64199 (e7a8645c, first shipped in v15.0.0) together with the deprecated `analyticsId`
- On current canary, the only remaining reference is the env injection in `packages/next/src/build/define-env.ts`; nothing reads `__NEXT_WEB_VITALS_ATTRIBUTION` anymore
- Previously reported in #65334 (closed as stale), original proposal: #39241

dayongkr:
### What?

Pass `experimental.webVitalsAttribution` through to `useReportWebVitals`.

### Why?

The option currently has no effect on the hook. Fixes #95919

### How?

Made the hook behave like the previous v14 `performance-relayer` pattern ([reference](https://github.com/vercel/next.js/blob/v14.2.15/packages/next/src/client/performance-relayer.ts#L71-L102)): the `web-vitals` bundle is selected per metric, using the attribution build for metrics listed in the option.

Added an e2e test covering metrics listed and not listed in the option.

### DIFF
--- a/packages/next/src/client/web-vitals.ts
+++ b/packages/next/src/client/web-vitals.ts
@@ -1,23 +1,40 @@
 import { useEffect } from 'react'
-import {
-  onLCP,
-  onFID,
-  onCLS,
-  onINP,
-  onFCP,
-  onTTFB,
-} from 'next/dist/compiled/web-vitals'
 import type { Metric } from 'next/dist/compiled/web-vitals'
 
-export function useReportWebVitals(
-  reportWebVitalsFn: (metric: Metric) => void
-) {
+// copied to prevent pulling in un-necessary utils
+const WEB_VITALS = ['CLS', 'FID', 'LCP', 'INP', 'FCP', 'TTFB'] as const
+
+type ReportWebVitalsCallback = (
+  metric: Metric & { attribution?: Record<string, unknown> }
+) => void
+
+function trackWebVitals(reportWebVitalsFn: ReportWebVitalsCallback) {
+  const attributions: string[] | undefined = process.env
+    .__NEXT_WEB_VITALS_ATTRIBUTION as any
+
+  for (const webVital of WEB_VITALS) {
+    try {
+      let mod: any
+      if (process.env.__NEXT_HAS_WEB_VITALS_ATTRIBUTION) {
+        if (attributions?.includes(webVital)) {
+          mod =
+            require('next/dist/compiled/web-vitals-attribution') as typeof import('next/dist/compiled/web-vitals-attribution')
+        }
+      }
+      if (!mod) {
+        mod =
+          require('next/dist/compiled/web-vitals') as typeof import('next/dist/compiled/web-vitals')
+      }
+      mod[`on${webVital}`](reportWebVitalsFn)
+    } catch (err) {
+      // Do nothing if the module fails to load
+      console.warn(`Failed to track ${webVital} web-vital`, err)
+    }
+  }
+}
+
+export function useReportWebVitals(reportWebVitalsFn: ReportWebVitalsCallback) {
   useEffect(() => {
-    onCLS(reportWebVitalsFn)
-    onFID(reportWebVitalsFn)
-    onLCP(reportWebVitalsFn)
-    onINP(reportWebVitalsFn)
-    onFCP(reportWebVitalsFn)
-    onTTFB(reportWebVitalsFn)
+    trackWebVitals(reportWebVitalsFn)
   }, [reportWebVitalsFn])
 }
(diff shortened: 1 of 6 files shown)


dayongkr:
Open question: with per-metric selection, listing only some metrics ships both web-vitals builds, even though the attribution build is a superset of the regular one. An all-or-nothing approach, e.g. a separate subpath, would always ship just one build. Happy to rework in that direction if preferred.

andrezani:
# This pull request is one approval away from merging a null-dereference crash.

[vercel/next.js](https://github.com/vercel/next.js) · [PR #95920](https://github.com/vercel/next.js/pull/95920) — fix: wire experimental.webVitalsAttribution into useReportWebVitals — was **open right now**. We ran PR Quorum's review panel over the same diff, read-only, live. **Total cost: $0.0020.** Here is what it caught:

---

### 🟠 process.env.__NEXT_HAS_WEB_VITALS_ATTRIBUTION is always a string, not a boolean

`packages/next/src/client/web-vitals.ts:22` · Architecture reviewer · High · 95% confidence

In [vercel/next.js · fix: wire experimental.webVitalsAttribution into useReportWebVitals](https://github.com/vercel/next.js/pull/95920).

The condition `if (process.env.__NEXT_HAS_WEB_VITALS_ATTRIBUTION)` checks a string environment variable for truthiness. In Next.js, compile-time env vars like `__NEXT_HAS_WEB_VITALS_ATTRIBUTION` are replaced with the string `'true'` or `'false'` (or are undefined). The string `'false'` is truthy in JavaScript, so this condition will always be true when the env var is defined, even when attribution is disabled. This means the attribution bundle will always be loaded and checked against the `attributions` list, even when `experimental.webVitalsAttribution` is not configured. This is a bug that causes unnecessary bundle loading and potential runtime overhead.

**Suggested fix**

```
if (attributions?.length) {
```

> **This review runs on your PRs *before* they merge.** → [Install PR Quorum](https://prquorum.com?utm_source=capture&utm_medium=report&utm_campaign=next.js-95920&utm_content=after-catch)

---

## The receipts

- **Read-only.** These PRs were open at capture time; we reviewed the live diff without posting anything. Maintainers can still catch these the ordinary way — installed, this review would already be sitting on the PR.
- **Unedited.** These are raw outputs from the default panel (Correctness, Security, Architecture) on the default model. Some findings will be wrong — that's the honest state of AI review, shown rather than hidden.

---

**vercel/next.js merges hundreds of pull requests a year. At about $0.0020 a review, checking every one of them costs a few dollars.**

> ### Run this panel on your own pull requests
> Three reviewers + a fact-checking verifier, one combined review, posted **before** merge — never after.
> **[Install PR Quorum](https://prquorum.com?utm_source=capture&utm_medium=report&utm_campaign=next.js-95920&utm_content=footer)** · free tier · 2-minute setup · advisory only, your team keeps merge control

PR #95922, opened 2026-07-18
Add next/nav-link for active-aware navigation links

aurorascharff:
### What?

Adds a `NavLink` component, similar to [React Router's `NavLink`](https://reactrouter.com/api/components/NavLink), to simplify a common task.

```tsx
import NavLink from 'next/nav-link'

// 1. String className, styled via the data-active variant — no render prop, works from a Server Component
<NavLink href="/dashboard" className="nav-item data-active:font-bold">
  Dashboard
</NavLink>

// 2. Render-prop className — reads isActive AND isPending (Client Component)
<NavLink
  href="/inbox"
  className={({ isActive, isPending }) =>
    cn(isActive && 'text-blue-600', isPending && 'opacity-50')
  }
>
  Inbox
</NavLink>

// 3. Render-prop children — swap content when active (Client Component)
<NavLink href="/profile">
  {({ isActive }) => (isActive ? <ProfileFilled /> : <ProfileOutline />)}
</NavLink>
```

### Why?

There's a lot of gotchas when doing this yourself, which I have documented here, so it would be great if Next.js could have this built in.  https://aurorascharff.no/posts/building-an-active-navlink-component-in-nextjs/

### How?

- A `<NavLink>` for the App Router that knows when it points at the current route. It sets `data-active` on the active link — including an ancestor matched by prefix — for styling, and `aria-current="page"` only on the exact page, so an ancestor is never announced as the current page.
- Style the active link in CSS via the `data-active:` Tailwind variant (or a `[data-active]` rule), or take full control by passing `className`/`children` as a function of `{ isActive, isPending }`.
- Matches nested routes by default (`/blog` is active on `/blog/post`); add `exact` for an exact-only match. Matching uses the pathname; for query-string state such as tabs, read `useSearchParams()` yourself.
- Active state is worked out on the server, so it's right on first paint and a nav in a static layout stays static under `cacheComponents` — no extra `Suspense`.
- On a route with an unknown dynamic param it can't be resolved at build time, so the active class fills in after hydration (a small flicker); the docs show an inline-script fix.
- A clear, `NavLink`-specific error if you use the function form from a Server Component.
- Ships with a docs page and e2e tests.

<details>
<summary>A `next/nav-link` entry that reuses `Link`, reads the pathname untracked, and resolves the render props inside the anchor.</summary>

**Its own module.** `next/nav-link` is a separate entry (like `next/form`) rather than an overload of `next/link`. Webpack aliases (`create-compiler-aliases.ts`) and the Turbopack import map (`next_import_map.rs`) mirror `next/link` across the base / app-client / react-server conditions; the react-server variant throws the Server-Component error, and the Pages Router entry throws "App Router only".

**Two things a userland wrapper can't do:**

- _Stays in the static shell under `cacheComponents`._ The active read uses `useUntrackedPathname()`, which reads `PathnameContext` directly and skips the dynamic-API guard that makes a hand-written `usePathname()` nav hang on fallback-param prerenders. The context is SSR-seeded, so the active class is already in the initial HTML for static routes.
- _`isPending` reaches `className`, not just `children`._ Both function props are resolved inside the anchor render, where the optimistic link status lives. A userland wrapper sets `className` on `Link` from the outside, so it can only observe pending via `useLinkStatus` inside `children` — never in `className`.

**Two attributes, two meanings.** `data-active` is present on any match (exact or prefix) for styling; `aria-current="page"` is set only on an exact match, so a prefix-matched ancestor gets `data-active` but not `aria-current`.

**Matching.** `exact` → `pathname === href`; default → `pathname === href || pathname.startsWith(href + '/')`, with `/` exact. Trailing slashes are normalized so matching is stable under `trailingSlash`.

</details>

### DIFF
--- a/crates/next-core/src/next_import_map.rs
+++ b/crates/next-core/src/next_import_map.rs
@@ -198,6 +198,14 @@ pub async fn get_next_client_import_map(
             );
             insert_exact_alias_or_js(
                 &mut import_map,
+                rcstr!("next/nav-link"),
+                request_to_import_mapping(
+                    project_path.clone(),
+                    rcstr!("next/dist/client/app-dir/nav-link"),
+                ),
+            );
+            insert_exact_alias_or_js(
+                &mut import_map,
                 rcstr!("next/form"),
                 request_to_import_mapping(
                     project_path.clone(),
@@ -352,6 +360,14 @@ pub async fn get_next_server_import_map(
             );
             insert_exact_alias_or_js(
                 &mut import_map,
+                rcstr!("next/nav-link"),
+                request_to_import_mapping(
+                    project_path.clone(),
+                    rcstr!("next/dist/client/app-dir/nav-link"),
+                ),
+            );
+            insert_exact_alias_or_js(
+                &mut import_map,
                 rcstr!("next/form"),
                 request_to_import_mapping(
                     project_path.clone(),
@@ -415,6 +431,7 @@ pub async fn get_next_edge_import_map(
         rcstr!("next/headers") => rcstr!("next/dist/api/headers"),
         rcstr!("next/image") => rcstr!("next/dist/api/image"),
         rcstr!("next/link") => rcstr!("next/dist/api/link"),
+        rcstr!("next/nav-link") => rcstr!("next/dist/api/nav-link"),
         rcstr!("next/navigation") => rcstr!("next/dist/api/navigation"),
         rcstr!("next/router") => rcstr!("next/dist/api/router"),
         rcstr!("next/script") => rcstr!("next/dist/api/script"),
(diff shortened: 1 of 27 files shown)


stefanprobst:
great! two questions:
- is `aria-current=page` applied when matching nested routes (i think it should not)
- how to differentiate `/path?tab=one` vs. `/path?tab=two` with `exact`

aurorascharff:
> great! two questions:
> 
> * is `aria-current=page` applied when matching nested routes (i think it should not)
> * how to differentiate `/path?tab=one` vs. `/path?tab=two` with `exact`

Great questions! 

1. Might be good to also do a data-pending rather in addition to aria currentpage 🤔 otherwise we break the css only styling story for non-exact.
2. This should probably be a custom useSearchParam tab component, ill add a link to an example!

Committed the changes here:
https://github.com/vercel/next.js/pull/95922/changes/e47a1f2a3a73c5fc6dd00eddd44fb073774c8327

stefanprobst:
is this still being considered?

PR #95924, opened 2026-07-18
Fix Flight payload chunk list duplication

### LINKED ISSUE #95559: Flight payload repeats the full chunk URL list for every client module reference (~29% of SSR HTML is duplicated chunk URLs)

https://github.com/Adophlidu/next-flight-chunklist-repro

### To Reproduce

1. `npm install`
2. `npm run build` (default Turbopack production build)
3. `npm run start`
4. `node analyze.mjs http://localhost:3000/` — the script curls the SSR HTML, decodes the `self.__next_f` flight payload, and counts chunk-URL strings in the client module reference rows.

The page is an ordinary shape: 30 small statically imported `'use client'` components plus 3 `next/dynamic` client components.

Output on `16.3.0-canary.79` (identical on latest stable `16.2.10`):

```
HTML bytes (raw)               15,869
flight payload bytes (decoded) 10,302 (64.9% of HTML)
client module reference rows   41
chunk-URL strings in flight    117
unique chunk files             5
duplication factor             23.4x
bytes of chunk-URL text        4,568 (28.8% of HTML)
```

**Current:** Every client-module reference row in the flight payload —

```
N:I[moduleId, ["/_next/static/chunks/a.js", "/_next/static/chunks/b.js", ...], exportName]
```

— serializes the **entire chunk URL list of its chunk group**. Because Turbopack merges a page's client modules into a shared chunk group, every reference row on the page carries an identical copy of the same URL list. The redundant bytes grow multiplicatively as `(number of client references) × (chunk group size)`, so on real applications this dominates the document.

Measured on a real production page, https://staging.firefly.social/profile/farcaster/3 (Next.js 16.2.x + Turbopack, deployed on Vercel; publicly accessible, `node analyze.mjs <url>` from the repro works against it):

```
HTML bytes (raw)               473,786
flight payload bytes (decoded) 357,251 (75.4% of HTML)
client module reference rows   74
chunk-URL strings in flight    2,770
unique chunk files             64
duplication factor             43.3x
bytes of chunk-URL text        135,749 (28.7% of HTML)
```

74 reference rows each repeat a ~38-URL chunk list: 2,770 chunk-URL strings of which only 64 are unique, ≈136 KB of pure URL text — about 29% of the whole document (the same ~29% ratio the minimal repro shows). Before this app switched to hash-only immutable asset paths, every URL additionally carried a `?dpl=` skew-protection suffix, which pushed URL text past 200 KB per page.

Gzip absorbs most of the wire cost, but the raw bytes still have to be streamed, parsed, and held in memory on every SSR request, and they inflate TTFB-to-last-byte for the streamed document.

**Expected:** Chunk lists are deduplicated in the flight payload — e.g. each distinct chunk list is emitted once and referenced by id from the `I[...]` rows (a dictionary encoding, analogous to how the flight protocol already deduplicates repeated rows by row id). That would collapse ~136 KB of URL text to roughly the size of one chunk list.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:59 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6030
  Available memory (MB): 36864
  Available CPU cores: 11
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: 4.12.0
  pnpm: 11.6.0
Relevant Packages:
  next: 16.3.0-canary.79 // Latest available version is detected (16.3.0-canary.79).
  eslint-config-next: N/A
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Performance, Turbopack

next build (local), next start (local), Vercel (Deployed)

### Additional context

- Affects App Router production builds with Turbopack (Turbopack's chunk-group merging makes the per-reference list large and identical across references; webpack builds emit the same encoding but with smaller per-module chunk lists).
- Reproduces identically on latest stable `16.2.10` and `16.3.0-canary.79`, locally and deployed on Vercel.
- This is a scalability issue rather than a functional bug: payload overhead grows as `client references × chunk-group size`, so it gets quadratically worse exactly on the pages that are already the largest.

agarwalrahul2702:
### What?

Fixes duplicate Flight payload chunk URL lists in App Router by dictionary-encoding repeated chunk lists and decoding them wherever Flight/RSC responses are consumed.

### Why?

Issue #95559 reports that App Router Flight payloads repeat the full list of chunk URLs for every client module reference row, which significantly increases SSR HTML size.

### How?

- Adds shared Flight chunk-list dictionary decoding that preserves streamed Flight framing, binary rows, and Cache Components completeness markers.
- Emits chunk-list dictionaries for inline Flight payloads and standalone RSC responses.
- Decodes deduplicated payloads before server React consumption, client inline processing, router fetches, resume fetches, and server action RSC fetches.
- Adds focused tests for split streamed dictionaries and binary row preservation.

Verification against `Adophlidu/next-flight-chunklist-repro`:

- Duplication factor dropped to `1.6x`.
- Chunk URL text dropped to `2.7%` of raw HTML.
- Raw HTML size dropped from `15,869` bytes to `11,696` bytes, a `26.3%` reduction.
- Production hydration passed with no browser errors.
- Standalone `text/x-component` RSC navigation decoded and rendered successfully with no browser errors.

Fixes #95559

### DIFF
--- a/packages/next/src/client/app-index.tsx
+++ b/packages/next/src/client/app-index.tsx
@@ -62,6 +62,7 @@ type FlightSegment =
   | [isNotBootstrap: 1, responsePartial: string]
   | [isFormState: 2, formState: any]
   | [isBinary: 3, responseBase64Partial: string]
+  | [isChunksDict: 4, chunksDict: Record<string, string[]>]
 
 type NextFlight = Omit<Array<FlightSegment>, 'push'> & {
   push: (seg: FlightSegment) => void
@@ -108,6 +109,11 @@ function nextServerDataCallback(seg: FlightSegment): void {
     } else {
       initialServerDataBuffer.push(decodedChunk)
     }
+  } else if (seg[0] === 4) {
+    ;(self as any).__next_f_chunks_dict = Object.assign(
+      (self as any).__next_f_chunks_dict || {},
+      seg[1]
+    )
   }
 }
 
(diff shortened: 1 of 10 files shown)


gaearon:
This is not a good way to fix it. I agree we need to fix this, but the fix should be at its source, and this is way downstream from that.

PR #95928, opened 2026-07-18
docs: add 'Set up your editor' to the installation guide

aurorascharff:
## Summary

CleanShot 2026-07-19 at 01 07 57@2x

Adds a "Set up your editor" section to the App Router installation guide. The App Router names files by convention (`page.tsx`, `layout.tsx`, `route.ts`), so an editor quickly fills with same-named tabs. This section shows how to tell them apart with VS Code / Cursor custom editor labels, labeled two folders deep so dynamic routes like `blog/[id]/page.tsx` do not all collapse to the same `[id]` label. It includes:

- A copy-paste `.vscode/settings.json` config (`workbench.editor.customLabels.patterns`) for setting it up by hand.
- A copyable prompt that points a coding agent back at this section to apply the config, including adapting it to other editors.
- A good-to-know that JetBrains IDEs (WebStorm, IntelliJ) do this automatically.

## Verification

- Previewed locally against the docs site: the section renders (HTTP 200), with the config block, the prompt card, and the good-to-know all rendering correctly.

### DIFF
--- a/docs/01-app/01-getting-started/01-installation.mdx
+++ b/docs/01-app/01-getting-started/01-installation.mdx
@@ -348,6 +348,39 @@ You can enable the plugin in VS Code by:
 
 See the [TypeScript reference](/docs/app/api-reference/config/typescript) page for more information.
 
+<AppOnly>
+
+## Set up your editor
+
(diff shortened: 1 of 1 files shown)


icyJoseph (approved):
Looks good. 

Most of the review here was about going through the first 3 pages of getting started, thinking how can we make it all better, but for now, let's take this in.

PR #95930, opened 2026-07-18
Return plain text 404 for non-document requests to unknown paths

gaearon:
Not sure if there's any downside to this.

Currently, there are situations when a browser requests something in a particular format:

- Favicons
- Manifest icons
- `` with broken `src`
- Wrong `<link>`s
- `new Worker()`

These requests have a `Sec-Fetch-Dest` header which can definitely exclude things that can't be HTML. My proposal is to not return the HTML for them (even in dev) so that this doesn't waste time in case rendering the root layout is slow or expensive. Less work for the server, whether dev or prod.

I believe this was happening in dev, and in prod if rendering `not-found` requires something dynamic (e.g. in the root layout). And for static self-hosted, I believe it would respond with static HTML.

After this change, we don't render the HTML.

## Before

Image in manifest hits the real `/not-found` route and runs application code:

Screenshot 2026-07-19 at 00 23 01

Notice 404 at the end with 2 seconds in application code.

## After

Image in manifest is still 404 but doesn't hit any app code.

Screenshot 2026-07-19 at 00 21 17

It's maybe a bit weird it doesn't show in the terminal log. In that sense it's like `/_next/static/` because it's handled at an earlier layer. Maybe we could still log it but it would have to happen from a different place, and also breaks consistency, so I opted with keeping it hidden.

---

I can't see any coherent arguments for why this should reach the app render so this should be OK?

### DIFF
--- a/packages/next/src/server/base-server.ts
+++ b/packages/next/src/server/base-server.ts
@@ -49,6 +49,7 @@ import * as path from 'path'
 import { format as formatUrl } from 'url'
 import { formatHostname } from './lib/format-hostname'
 import { isRSCRequestHeader } from './lib/is-rsc-request'
+import { isNonHtmlSecFetchDest } from './lib/is-non-html-sec-fetch-dest'
 import {
   APP_PATHS_MANIFEST,
   NEXT_BUILTIN_DOCUMENT,
@@ -2354,6 +2355,18 @@ export default abstract class Server<
     // we need to ensure the status code if /404 is visited directly
     if (is404Page && !isNextDataRequest && !isRSCRequest) {
       res.statusCode = 404
+
+      // For subresource requests (e.g. images or fonts), return plain text
+      // 404 instead of rendering the not-found route.
+      if (isNonHtmlSecFetchDest(req.headers['sec-fetch-dest'])) {
+        res.setHeader(
+          'Cache-Control',
+          'private, no-cache, no-store, max-age=0, must-revalidate'
+        )
+        res.setHeader('Content-Type', 'text/plain; charset=utf-8')
+        res.body('Not Found').send()
+        return null
+      }
     }
 
     // ensure correct status is set when visiting a status page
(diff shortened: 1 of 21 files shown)


packages/next/src/server/base-server.ts
```diff
@@ -2331,6 +2332,18 @@ export default abstract class Server<
...
+      // For subresource requests (e.g. images or fonts), return plain text
+      // 404 instead of rendering the not-found route.
+      if (isNonHtmlSecFetchDest(req.headers['sec-fetch-dest'])) {
+        res.setHeader(
+          'Cache-Control',
+          'private, no-cache, no-store, max-age=0, must-revalidate'
+        )
+        res.setHeader('Content-Type', 'text/plain; charset=utf-8')
+        res.body('Not Found').send()
+        return null
```
eps1lon, line 2371: Supposedly we need to include these request headers in the `Vary` response header

> Important: For these protections to take effect on GET requests, your server must include Sec-Fetch-Dest, Sec-Fetch-Mode, Sec-Fetch-Site in its Vary header.

-- https://web.dev/articles/fetch-metadata#sec-fetch-dest

I guess you're attempting a similar thing with this specific `Cache-Control` header instead?

Also worth asserting on the response headers to make sure this works as expected when deployed to Vercel

gaearon: Yeah, this is what we do in the existing 404 (and generally error) path, for example here:

https://github.com/vercel/next.js/blob/eb2fc4b521813628c7b93f09763c197ea35c84cc/packages/next/src/server/base-server.ts#L2877-L2890

It seems like if we say it's no-store then there shouldn't be a need for Vary?

I'll add a test for the headers.

packages/next/src/server/lib/is-non-html-sec-fetch-dest.ts
```diff
@@ -0,0 +1,30 @@
...
+  'video',
+  'webidentity',
+  'worker',
+  'xslt',
+])
+
+export function isNonHtmlSecFetchDest(
+  value: string | string[] | null | undefined
+): boolean {
+  return typeof value === 'string' && NON_HTML_SEC_FETCH_DESTS.has(value)
```
gaearon, line 29: This could also be an allowlist instead like Vite does. https://github.com/vitejs/vite/pull/20866

Blocklist felt a bit safer but maybe I'm being overly cautious

gnoff (approved):

PR #95934, opened 2026-07-19
Turbopack: stop copying sourcesContent into every serialized source map

gaearon:
This reduces peak memory consumption with the original repro case in [vercel/next.js#81161](https://github.com/vercel/next.js/issues/81161).

If you walk the 15 routes in a quick succession and capture vmmap --summary at the end of each fast sweep (before memory is released), with this change the number reliably goes down from 3 GB to 2.6 GB:

```
before	1	3.0G	17
after	1	2.6G	16
before	2	3.0G	16
after	2	2.6G	16
before	3	3.1G	18
after	3	2.6G	17
before	4	3.1G	17
after	4	2.6G	16
before	5	3.1G	17
after	5	2.6G	17
before	6	3.0G	17
after	6	2.5G	17
before	7	3.1G	17
after	7	2.6G	17
before	8	3.0G	25
after	8	2.6G	24
before: [3.0, 3.0, 3.1, 3.1, 3.1, 3.0, 3.1, 3.0] 
after:  [2.6, 2.6, 2.6, 2.6, 2.6, 2.5, 2.6, 2.6]
mean before=3.050G after=2.587G  diff=0.462G (-15.2%)
exact permutation test: p = 2/12870 = 0.00016
```

More runs on this app:

Screenshot 2026-07-22 at 20 59 12

And here's an internal app showing ~20% dev memory win at peak (although reclaiming memory settles it):

Screenshot 2026-07-22 at 16 45 34

---

<details>

Module source maps are serialized to a JSON rope right after code generation and treated as opaque bytes from then on. Since `sourcesContent` — the full original source text — is inlined into that JSON, every later step that modifies the map has to copy all of it: rewriting the `sources` URLs for each chunking context copies the entire map per context, and embedding the module map into a chunk's sectioned map copies it again, once per layer that includes the module. In a dev session on a dependency-heavy app, each module's source text ends up resident ~5 times over, and source maps end up dominating the memory the task store holds for codegen output (in one measured app, maps were 57% of module factory bytes, and `sourcesContent` alone was 36%).

This PR stops copying the text. `StructuredSourceMap` keeps a map in field form until the moment it is actually emitted:

- Fields we never touch (`mappings`, `names`, `version`, ...) are stored as verbatim raw JSON snippets, so producers' bytes round-trip untouched and re-emission is rope sharing rather than re-serialization.
- `sources` is the one field rewrites modify, so it is stored typed, and the per-chunking-context "fileify" rewrites become a plain string rewrite of that field instead of a parse–modify–reserialize of the whole map.
- `sourcesContent` entries are individual, already-JSON-escaped ropes. Rewritten maps and chunk maps share them; the source text is escaped once per module and never copied again.

Two construction paths, chosen by where the bytes come from: maps we generate ourselves are built directly from the `swc_sourcemap` object (`from_swc_map` takes the contents out, serializes the small remainder, and never round-trips source text through JSON), while maps that arrive as bytes (prebuilt maps next to external code, the single-complete-original fast path) go through `from_json`. The emitter writes fields in exactly `swc_sourcemap`'s order and format, pinned by golden tests that assert byte-identity with `to_writer` — so emitted chunks and maps are byte-for-byte unchanged, and no snapshot fixtures change. The CSS pipeline gets the same treatment; the single-file output path is left alone since it serializes straight into the final artifact and there is nothing to share.

The win is proportional to how much duplicated source-map text an app holds, so results range accordingly. On a dependency-heavy reproduction app (13 routes sharing large wallet SDKs; rapid route sweep in `next dev`, macOS physical footprint, 8 interleaved cold runs per side): 3.05 GB → 2.59 GB mean (−15%), with zero overlap between the two distributions (exact permutation test p ≈ 0.00016) and per-route latencies unchanged. On `bench/basic-app` (small first-party modules): a full sweep of all 42 routes drops the post-sweep footprint 3.4–3.5 GB → 2.4–2.5 GB (−29%, interleaved runs); cold compile −4%, idle footprint identical (eviction reclaims both sides equally at idle). On `bench/heavy-npm-deps` (heavy deps disabled in its default config) and on very large first-party apps whose memory is dominated by the JS heap and application state rather than map text, the change measures flat — as expected, since there is little duplicated text for sharing to recover. `small_apps` build benchmarks are flat except `date-fns-all` at +1–2% (~2.5 ms): that app is 1000+ tiny modules and `from_swc_map` adds one small-skeleton JSON parse per module, with no redundancy to win back in a single-context one-shot build. The single-complete-original fast path (per prebuilt module with a shipped map, and per minified chunk) is cost-neutral: the structured map is built by splitting the raw map's serde fields directly (`from_serialize`), so the serialized document is never materialized or re-parsed — 93 ms per 5 iterations on a synthetic 21.7 MB map, matching the pre-PR baseline exactly, with byte-identical output pinned by a property test. Serialized task-store size is unchanged (bincode expands sharing back out), so this is purely a live-memory change.

Verified: snapshot suite passes with zero fixture changes, execution suite matches canary (the single pre-existing failure reproduces on the base tag), unit suites pass including new golden byte-identity tests, and emitted `.map` files were validated end-to-end in a running dev server (multi-section chunk maps, fileified `file://` sources, `sourcesContent` present and byte-identical between base and this branch).

</details>

### DIFF
--- a/turbopack/crates/turbopack-core/src/code_builder.rs
+++ b/turbopack/crates/turbopack-core/src/code_builder.rs
@@ -9,7 +9,7 @@ use anyhow::Result;
 use bincode::{Decode, Encode};
 use tracing::instrument;
 use turbo_rcstr::RcStr;
-use turbo_tasks::{ResolvedVc, Vc};
+use turbo_tasks::{NonLocalValue, ResolvedVc, Vc, trace::TraceRawVcs};
 use turbo_tasks_fs::{
     File, FileContent,
     rope::{Rope, RopeBuilder},
@@ -17,14 +17,47 @@ use turbo_tasks_fs::{
 use turbo_tasks_hash::hash_xxh3_hash128;
 
 use crate::{
+    source_map::structured::StructuredSourceMap,
     debug_id::generate_debug_id,
     output::OutputAsset,
     source_map::{GenerateSourceMap, SourceMap, SourceMapAsset},
     source_pos::SourcePos,
 };
 
+/// A per-section source map: either an opaque serialized map or a structured one whose
+/// `sourcesContent` is shared rather than copied when the section is embedded.
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TraceRawVcs, NonLocalValue)]
+pub enum SectionMap {
+    Raw(Rope),
+    Structured(StructuredSourceMap),
+}
+
+impl turbo_tasks_hash::DeterministicHash for SectionMap {
+    fn deterministic_hash<H: turbo_tasks_hash::DeterministicHasher>(&self, state: &mut H) {
+        match self {
+            SectionMap::Raw(map) => {
+                state.write_u8(0);
+                map.deterministic_hash(state);
+            }
+            SectionMap::Structured(map) => {
+                state.write_u8(1);
+                map.deterministic_hash(state);
+            }
+        }
+    }
+}
+
+impl SectionMap {
+    pub fn to_rope(&self) -> Rope {
+        match self {
+            SectionMap::Raw(map) => map.clone(),
+            SectionMap::Structured(map) => map.to_rope(),
+        }
+    }
+}
+
 /// A mapping of byte-offset in the code string to an associated source map.
-pub type Mapping = (usize, Option<Rope>);
+pub type Mapping = (usize, Option<SectionMap>);
 
 /// Code stores combined output code and the source map of that output code.
 #[turbo_tasks::value(shared, serialization = "hash")]
@@ -167,7 +200,14 @@ impl CodeBuilder {
     /// available. If it's not, this is no different than pushing Synthetic
     /// code.
     pub fn push_source(&mut self, code: &Rope, map: Option<Rope>) {
-        self.push_map(map);
+        self.push_map(map.map(SectionMap::Raw));
+        self.code += code;
+    }
+
+    /// Like [`CodeBuilder::push_source`] for a structured map whose `sourcesContent` stays
+    /// shared when the resulting code's map is emitted.
+    pub fn push_structured_source(&mut self, code: &Rope, map: Option<StructuredSourceMap>) {
+        self.push_map(map.map(SectionMap::Structured));
         self.code += code;
     }
 
@@ -205,7 +245,7 @@ impl CodeBuilder {
     /// original code section. By inserting an empty source map when reaching a
     /// synthetic section directly after an original section, we tell Chrome
     /// that the previous map ended at this point.
-    fn push_map(&mut self, map: Option<Rope>) {
+    fn push_map(&mut self, map: Option<SectionMap>) {
         let Some(mappings) = self.mappings.as_mut() else {
             return;
         };
@@ -339,7 +379,7 @@ impl Code {
             last_byte_pos = *byte_pos;
 
             if let Some(map) = map {
-                sections.push((pos, map.clone()))
+                sections.push((pos, map.to_rope()))
             } else {
                 // We don't need an empty source map when column is 0 or the next char is a newline.
                 if pos.column != 0
--- a/turbopack/crates/turbopack-core/src/code_builder.rs
+++ b/turbopack/crates/turbopack-core/src/code_builder.rs
@@ -31,6 +31,18 @@ pub enum SectionMap {
     Structured(Box<StructuredSourceMap>),
 }
 
+impl From<Rope> for SectionMap {
+    fn from(map: Rope) -> Self {
+        SectionMap::Raw(map)
+    }
+}
+
+impl From<StructuredSourceMap> for SectionMap {
+    fn from(map: StructuredSourceMap) -> Self {
+        SectionMap::Structured(Box::new(map))
+    }
+}
+
 impl DeterministicHash for SectionMap {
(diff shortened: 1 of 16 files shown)


turbopack/crates/turbopack-core/src/source_map/structured.rs
```diff
@@ -0,0 +1,333 @@
...
+    x_facebook_offsets: Option<Box<RawValue>>,
+    x_metro_module_paths: Option<Box<RawValue>>,
+    x_facebook_sources: Option<Box<RawValue>>,
+    #[serde(rename = "debug_id")]
+    debug_id_old: Option<Box<RawValue>>,
+    #[serde(rename = "debugId")]
+    debug_id: Option<Box<RawValue>>,
+}
+
+fn snippet(value: Option<Box<RawValue>>) -> Option<Rope> {
```
lukesandberg: kindof an odd name

as_rope or into_rope might be more intuitive

turbopack/crates/turbopack-core/src/source_map/structured.rs
```diff
@@ -0,0 +1,333 @@
...
+        ) {
+            if let Some(value) = value {
+                if !*first {
+                    *builder += ",";
+                }
+                *first = false;
+                *builder += "\"";
+                *builder += key;
+                *builder += "\":";
+                *builder += value;
```
lukesandberg, line 384: for key i could buy that we don't need escaping, but value is less clear

turbopack/crates/turbopack-core/src/code_builder.rs
```diff
@@ -9,22 +9,55 @@ use anyhow::Result;
...
 
+/// A per-section source map: either an opaque serialized map or a structured one whose
+/// `sourcesContent` is shared rather than copied when the section is embedded.
+#[derive(Clone, Debug, PartialEq, Eq, Encode, Decode, TraceRawVcs, NonLocalValue)]
+pub enum SectionMap {
+    Raw(Rope),
+    Structured(StructuredSourceMap),
+}
+
+impl turbo_tasks_hash::DeterministicHash for SectionMap {
```
lukesandberg: nit: import the name with `use`

turbopack/crates/turbopack-core/src/source_map/structured.rs
```diff
@@ -0,0 +1,333 @@
...
+    value.map(|v| Rope::from(v.get().as_bytes().to_vec()))
+}
+
+/// JSON-escapes a source's content into the string-literal form stored in `sources_content`.
+fn escape_content(content: Option<&str>) -> Rope {
+    match content {
+        Some(text) => {
+            Rope::from(serde_json::to_vec(text).expect("string serialization is infallible"))
+        }
+        None => Rope::from("null"),
```
lukesandberg: could store this in a Lazy if we expect it to be common. would save a few allocations (at the cost of some ref-counts)

turbopack/crates/turbopack-ecmascript/src/chunk/item.rs
```diff
@@ -165,22 +166,25 @@ impl EcmascriptChunkItemContent {
             code += " try {\n";
         }
 
-        let source_map = match &self.rewrite_source_path {
-            RewriteSourcePath::AbsoluteFilePath(path) => {
-                absolute_fileify_source_map(self.source_map.as_ref(), path.clone()).await?
```
lukesandberg, line 170: is this function dead now? can we delete it?

turbopack/crates/turbopack-ecmascript/src/parse.rs
```diff
@@ -240,6 +241,62 @@ pub fn generate_js_source_map<'a>(
     }
 }
 
+/// Like [`generate_js_source_map`], but returns a [`StructuredSourceMap`] whose
```
lukesandberg: migrate the last caller to this method (and have it call to_rope).  that way we can delete the old path

turbopack/crates/turbopack-core/src/source_map/structured.rs
```diff
@@ -0,0 +1,639 @@
+//! A source map kept in structured form until it is actually emitted.
+//!
+//! Historically module source maps were serialized to a JSON [`Rope`] immediately after code
+//! generation and treated as opaque bytes from then on. Because `sourcesContent` (the full
+//! original source text) is inlined into that JSON, every subsequent per-chunking-context
+//! rewrite of the `sources` URLs had to copy the entire map — duplicating each module's source
+//! text once per layer and again per chunking context — and every embedding of the map into a
+//! chunk's source map copied it again.
+//!
+//! [`StructuredSourceMap`] instead stores the map's fields:
```
lukesandberg: instead of describing the previous approach simply describe that this is an optimization to reduce redundant rope copies

turbopack/crates/turbopack-ecmascript/src/static_code.rs
```diff
@@ -55,7 +55,7 @@ impl StaticEcmascriptCode {
             .module_content_without_analysis(self.generate_source_map)
             .await?;
         let mut code = CodeBuilder::default();
-        code.push_source(
+        code.push_structured_source(
```
lukesandberg: consider whether we can just keep the old names

lukesandberg (approved):
🥇

eluup15-web (commented):
Block

Issue #77733, opened 2025-04-02
Next.js v15.2 creates a lot of fake scripts for server components with the same source map which breaks JS debuggers


Soarex16:
### Link to the code that reproduces this issue

https://github.com/Soarex16/next-15-server-components-reproduction-app

### To Reproduce

1. Start the application in dev mode (either `next dev` or `next dev --turbopack`)
2. Set a breakpoint on `app/page.tsx:3` (on the `console.log`)
3. Start debugging in a browser

### Expected behavior

The debugger will stop in `app/page.tsx` only once (at the breakpoint on line 3).
After resuming the code, the debugger will not stop.

### Actual behavior

The debugger stops at `app/page.tsx:3`, but after resume it stops at random locations in the same file.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:23 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T6020
  Available memory (MB): 65536
  Available CPU cores: 12
Binaries:
  Node: 20.9.0
  npm: 10.1.0
  Yarn: 1.22.21
  pnpm: 8.10.5
Relevant Packages:
  next: 15.3.0-canary.29 // Latest available version is detected (15.3.0-canary.29).
  eslint-config-next: N/A
  react: 19.1.0
  react-dom: 19.1.0
  typescript: 5.8.2
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Not sure

### Which stage(s) are affected? (Select all that apply)

next dev (local)

#### TLDR
- React Server components machinery creates fake scripts with the same source map as a real user code to preserve source mapped traces in the browser.
  This confuses js debuggers because they think that this piece of code was bundled in different places multiple times.
  So when the user put a breakpoint, the debugger put breakpoints in all these scripts causing redundant breaks.
- This issue affects server components debugging both Webpack and Turbopack
- This bug was introduced somewhere between 15.1.7 and 15.2. Sorry, I didn't have enough time to bisect it.
- Reproducible with all JS popular debuggers (Chrome DevTools, WebStorm, VS Code). What's interesting is, in Firefox it works as expected. 
- Video with debugging in Chrome DevTools:

https://github.com/user-attachments/assets/fc10b498-6972-4077-8a52-70e62f6275ab

#### Detailed investigation

[...]

PR #95936, opened 2026-07-19
[dev] Serve React fake-frame source maps by URL instead of inlining data: URLs

gaearon:
(This doesn't actually work)

---

This should fix some OOMs on devserver for large apps.

The issue is due to `data:` sourcemaps, as foreseen in https://github.com/vercel/next.js/pull/81904. Node.js forever retains a parsed copy of a sourcemap per eval'd fake function, keyed by each function's unique `sourceURL`.

I'm using the same endpoint that #71042 uses instead, and then resolving fake frames by devirtualizing them. 

## Test Plan

Run Vercel docs in front.

```
rm -rf apps/vercel-docs/.next
```

Then in one tab:

```
while sleep 1; do
  pid=$(lsof -ti tcp:3332 -sTCP:LISTEN | head -1)
  if [ -n "$pid" ]; then
    ps -o rss= -p "$pid" | awk -v p="$pid" '{printf "pid %s  %5.2f GB\n", p, $1/1048576}'
  else
    echo "nothing listening on :3332"
  fi
```

In another tab:

```
PORT=3332
for r in /docs /docs/functions /docs/cli /docs/rest-api /docs/edge-config /docs/analytics /docs/speed-insights /docs/project-configuration/vercel-json /docs/deployments /docs/git /docs/domains /docs/storage /docs/observability; do curl -so /dev/null -w "$r  %{http_code}  %{time_total}s\n" --max-time 600 "http://localhost:$PORT$r"; done
```

### Before

Before the fix, we can only render a few routes before the crash:

```
/docs  200  11.587641s
/docs/functions  200  7.714000s
/docs/cli  200  18.758103s
/docs/rest-api  200  18.791006s
```

They're also *super* slow.

At this point it is killed by the memory watcher:

```
⚠ Server is approaching the used memory threshold, restarting...
```

I didn't measure after restart.

Memory until it's killed:

```
pid 45076   0.90 GB
pid 45076   0.90 GB
pid 45076   0.90 GB
pid 45076   0.98 GB
pid 45076   1.11 GB
pid 45076   1.71 GB
pid 45076   1.75 GB
pid 45076   1.98 GB
pid 45076   2.32 GB
pid 45076   2.54 GB
pid 45076   2.59 GB
pid 45076   2.86 GB
pid 45076   4.08 GB
pid 45076   4.45 GB
pid 45076   4.59 GB
pid 45076   4.97 GB
pid 45076   5.36 GB
pid 45076   5.72 GB
pid 45076   5.99 GB
pid 45076   6.01 GB
pid 45076   6.01 GB
pid 45076   5.98 GB
pid 45076   6.17 GB
pid 45076   6.35 GB
pid 45076   6.53 GB
pid 45076   6.75 GB
pid 45076   7.08 GB
pid 45076   7.11 GB
pid 45076   7.28 GB
pid 45076   7.11 GB
pid 45076   7.25 GB
pid 45076   7.40 GB
pid 45076   7.55 GB
pid 45076   7.71 GB
pid 45076   7.86 GB
pid 45076   8.02 GB
pid 45076   8.18 GB
pid 45076   8.35 GB
pid 45076   8.52 GB
pid 45076   8.68 GB
pid 45076   8.87 GB
pid 45076   9.06 GB
pid 45076   9.22 GB
pid 45076   9.16 GB
pid 45076   9.32 GB
pid 45076   9.47 GB
pid 45076   9.63 GB
pid 45076   9.77 GB
pid 45076   9.94 GB
pid 45076  10.09 GB
pid 45076  10.25 GB
pid 45076  10.42 GB
pid 45076  10.19 GB
pid 45076  10.35 GB
pid 45076  10.54 GB
pid 45076  10.71 GB
pid 45076  10.90 GB
pid 45076  10.90 GB
nothing listening on :3332 [OOM]
```

### After

After the fix, we can get through all the routes.

```
/docs  200  12.391647s
/docs/functions  200  3.426163s
/docs/cli  200  3.078210s
/docs/rest-api  200  4.144664s
/docs/edge-config  200  2.470627s
/docs/analytics  200  2.083005s
/docs/speed-insights  200  2.069460s
/docs/project-configuration/vercel-json  200  6.318112s
/docs/deployments  200  6.636235s
/docs/git  200  2.687616s
/docs/domains  200  2.237283s
/docs/storage  200  2.600683s
/docs/observability  200  3.324540s
```

```
pid 37866   0.87 GB
pid 37866   0.87 GB
pid 37866   0.87 GB
pid 37866   0.87 GB
pid 37866   0.90 GB
pid 37866   1.09 GB
pid 37866   1.61 GB
pid 37866   1.71 GB
pid 37866   1.75 GB
pid 37866   1.75 GB
pid 37866   1.75 GB
pid 37866   2.00 GB
pid 37866   2.30 GB
pid 37866   2.40 GB
pid 37866   2.53 GB
pid 37866   2.68 GB
pid 37866   2.85 GB
pid 37866   2.91 GB
pid 37866   2.99 GB
pid 37866   3.03 GB
pid 37866   3.22 GB
pid 37866   3.30 GB
pid 37866   3.41 GB
pid 37866   3.93 GB
pid 37866   4.19 GB
pid 37866   4.27 GB
pid 37866   4.48 GB
pid 37866   4.61 GB
pid 37866   4.69 GB
pid 37866   4.74 GB
pid 37866   4.91 GB
pid 37866   4.96 GB
pid 37866   5.13 GB
pid 37866   5.16 GB
pid 37866   5.21 GB
pid 37866   5.38 GB
pid 37866   5.38 GB
pid 37866   5.40 GB
pid 37866   6.95 GB
pid 37866   7.06 GB
pid 37866   7.54 GB
pid 37866   7.94 GB
pid 37866   7.96 GB
pid 37866   8.22 GB
pid 37866   8.67 GB
pid 37866   8.73 GB
pid 37866   8.80 GB
pid 37866   8.81 GB
pid 37866   8.81 GB
pid 37866   8.82 GB
pid 37866   8.84 GB
pid 37866   8.83 GB
pid 37866   8.85 GB
pid 37866   8.85 GB
pid 37866   8.85 GB
pid 37866   8.86 GB
pid 37866   8.87 GB
pid 37866   8.87 GB
pid 37866   8.87 GB
pid 37866   8.87 GB
pid 37866   8.87 GB
pid 37866   8.08 GB
```

---

<details>

### What

In development, the server-side Flight clients (SSR consumption of the RSC stream, `"use cache"` deserialization, cacheComponents validation, segment prefetch generation, server action argument decryption) pass `findSourceMapURLDEV` to React. React embeds the returned string into every eval'd fake stack-frame function it creates for Server Components debug info. Today that string is the **entire source map of the referenced chunk, inlined as a base64 `data:` URL** (often several MB).

Because `next dev` force-enables Node.js source maps, Node extracts and parses that data URL for **every eval'd fake function** and retains the parsed payload in `generatedSourceMapCache` — a plain `SafeMap` that is never evicted — keyed by each fake function's globally unique `sourceURL` (`about://React/<env>/<file>?<idx>`), so nothing ever dedupes.

Retained memory ≈ unique stack frames × chunk source map size. On a component-heavy app this OOMs the dev server within a few page loads.

Measured on vercel.com/docs (large MDX app, `cacheComponents: true`, next 16.3.0-canary.87), sweeping 13 doc routes with a CDP harness (heap sampling + `Debugger.scriptParsed` census):

| | before | after this PR |
|---|---|---|
| sweep outcome | memory watchdog restart at route 3, then V8 OOM (8GB limit) | all 13 routes, 0 restarts |
| `heapUsed` | 3.1GB after 2 routes, monotonic | ~2.4GB peak, GC-stable |
| source-map bytes attached to fake scripts | 1.3GB after 2 routes (would be ~19GB for the full sweep) | **8MB** for 28,210 fake scripts |
| dominant allocations | `sourceMapFromDataUrl` + magic-comment `RegExp.exec` ≈ 2.4GB of 2.8GB sampled | gone |

~96% of the fake functions came from the cacheComponents validation pass (`environmentName: "Prerender"`), the rest from SSR/`use cache` — all go through this one function.

### How

**`source-maps.ts`:** when the server origin is known (`process.env.__NEXT_PRIVATE_ORIGIN`, set by `start-server`), return a URL to the already-existing `/__nextjs_source-map` endpoint — exactly what the browser-side `findSourceMapURL` (`app-find-source-map-url.ts`) has always done — instead of inlining the payload. Falls back to the previous `data:` URL behavior when the origin is unknown (custom servers).

For non-`data:` source map URLs on eval'd scripts, Node stores a tiny no-parse cache entry (`dataFromUrl` → unknown protocol → `null`), so the per-fake-function cost drops from a parsed multi-MB payload to a ~150-byte string. Debugger frontends (Chrome DevTools attached to the Node process, VS Code js-debug) fetch `http:` source map URLs themselves, and can now cache the map per URL — impossible before, since every fake script carried its own identical multi-MB `data:` blob.

**`patch-error-inspect.ts`:** devirtualize `about://React/<env>/<file>?<idx>` → `<file>` *before* the `findSourceMap`/bundler lookups (the dev overlay's `getOriginalStackFrames` already does this). React's fake functions are whitespace-padded so their frame positions match the underlying chunk, so the real chunk's source map produces identical results. This keeps fake frames in printed stacks sourcemappable now that they no longer carry their own inlined copy, and dedupes the `SourceMapConsumer` cache across the many fake scripts per chunk (it was previously keyed by each unique fake URL).

### Behavior verification

Consumers of these source maps, traced and tested against the vercel-docs workload:

- **Terminal error output** (`patch-error-inspect`): thrown-error probe produces byte-identical output before/after — original path, line:col, and code frame.
- **Dev overlay** (`__nextjs_original-stack-frames`): already devirtualizes and uses `project.traceSource`; never consumed the inlined maps.
- **Browser devtools**: already served by HTTP endpoint; unchanged.
- **Attached debuggers on the server process**: every fake script's `sourceMapURL` now points at `/__nextjs_source-map?filename=…`; verified the endpoint returns the full map (HTTP 200) for server chunk filenames (Node cache first, `project.getSourceMap` fallback). One manual smoke test with Chrome DevTools attached to the Node process is still worth doing before landing — marking this PR draft partly for that.
- **React unchanged**: unique fake sourceURLs, eager task/stack materialization, and `fakeFunctionCache` semantics are untouched; React never inspects the returned URL string.

### Notes

- The 512MB `sourceMapURLs` LRU is kept for the fallback path; if the fallback is deemed unnecessary it can be deleted in a follow-up.
- Complementary upstream issues worth filing separately: Node's `generatedSourceMapCache` never evicts, and identical `sourceMappingURL` strings are parsed/retained once per eval'd script (content-keyed dedupe would make any producer's duplication survivable).
- Related: #77733, discussion #62198.

## Debugger support

Chrome DevTools fetches source maps for attached Node.js targets from its privileged `devtools://devtools` origin ([docs](https://developer.chrome.com/docs/devtools/developer-resources)), which `blockCrossSiteDEV` rejected with a 403, so the served URLs were unusable by the primary consumer this PR needs to keep working. This PR allows `devtools://` origins through the dev cross-origin protection. This admits no requester that couldn't already reach these endpoints: the `Origin` header is browser-controlled and web content can never run under the `devtools://` scheme, while requests without an `Origin` header (e.g. non-browser clients) are already allowed today.

## Tests

- `fake-frame-source-map-urls.test.ts` (new): attaches to the dev server over CDP like a debugger, asserts every `about://React/` fake script references a `/__nextjs_source-map` URL instead of a `data:` URL, fetches the maps with `Origin: devtools://devtools`, and asserts a padded fake frame position maps back to the original source file. Fails without the fix (data: URLs), and fails if the endpoint rejects DevTools' origin.
- `server-source-maps.test.ts` (new case): an error thrown inside `"use cache"` crosses a Flight boundary, so the logged error's stack is rebuilt through fake frames (`resolveErrorDev`); asserts it still prints a source-mapped stack with a codeframe. Fails if fake frames are no longer devirtualized to the underlying chunk's map.
- `allowed-dev-origins.test.ts` (new case): `devtools://devtools` origin requests to internal dev endpoints are not blocked, for both `basePath` variants.

</details>

### DIFF
--- a/packages/next/src/server/lib/source-maps.ts
+++ b/packages/next/src/server/lib/source-maps.ts
@@ -172,6 +172,11 @@ const sourceMapURLs = new LRUCache<string | typeof invalidSourceMap>(
 export function findSourceMapURLDEV(
   scriptNameOrSourceURL: string
 ): string | null {
+  if (scriptNameOrSourceURL === '') {
+    // Anonymous frames don't have a source map. Bail explicitly since
+    // Node.js' `findSourceMap('')` can return an unrelated script's map.
+    return null
+  }
   let sourceMapURL = sourceMapURLs.get(scriptNameOrSourceURL)
   if (sourceMapURL === undefined) {
     let sourceMapPayload: ModernSourceMapPayload | undefined
@@ -183,9 +188,25 @@ export function findSourceMapURLDEV(
       )
     }
 
+    const origin = process.env.__NEXT_PRIVATE_ORIGIN
     if (sourceMapPayload === undefined) {
       sourceMapURL = invalidSourceMap
+    } else if (origin !== undefined) {
+      // Prefer returning a URL to the dev server's source map endpoint (the
+      // same one the browser-side `findSourceMapURL` uses) over inlining the
+      // whole payload as a `data:` URL. React embeds the returned URL into
+      // every eval'd fake stack frame function, and each fake function has a
+      // unique sourceURL. With `data:` URLs, Node.js (with source maps
+      // enabled, as `next dev` enables them) parses and retains a separate
+      // copy of the full multi-MB source map payload per fake function in a
+      // never-evicted cache, which can consume gigabytes of heap on
+      // component-heavy apps in development. A URL reference costs ~100 bytes
+      // per fake function instead, can be cached per file by debugger
+      // frontends, and Node.js stores it without parsing anything.
+      sourceMapURL = `${origin.replace(/\/$/, '')}/__nextjs_source-map?filename=${encodeURIComponent(scriptNameOrSourceURL)}`
     } else {
+      // Fallback for custom servers where the origin is unknown: inline the
+      // payload.
       // TODO: Might be more efficient to extract the relevant section from Index Maps.
       // Unclear if that search is worth the smaller payload we have to stringify.
       const sourceMapJSON = JSON.stringify(sourceMapPayload)
--- a/packages/next/src/server/patch-error-inspect.ts
+++ b/packages/next/src/server/patch-error-inspect.ts
@@ -177,11 +177,18 @@ function getSourcemappedFrameIfPossible(
   stack: IgnorableStackFrame
   code: string | null
 } {
-  const sourceMapCacheEntry = sourceMapCache.get(frame.file)
+  // React creates fake eval'd scripts for Server Component stack frames with
+  // virtual URLs like `about://React/Server/file:///path/to/chunk.js?42`.
+  // Their frame positions are padded to match the underlying chunk, so we can
+  // resolve them via the real chunk's source map. This keeps fake frames
+  // sourcemappable when they don't carry their own inlined source map, and
+  // dedupes the consumer cache across the many fake scripts per chunk.
+  const fileKey = devirtualizeReactServerURL(frame.file)
+  const sourceMapCacheEntry = sourceMapCache.get(fileKey)
   let sourceMapConsumer: SyncSourceMapConsumer
   let sourceMapPayload: ModernSourceMapPayload
   if (sourceMapCacheEntry === undefined) {
-    let sourceURL = frame.file
+    let sourceURL = fileKey
     // e.g. "/Users/foo/APP/.next/server/chunks/ssr/[root-of-the-server]__2934a0._.js"
     // or "C:\Users\foo\APP\.next\server\chunks\ssr\[root-of-the-server]__2934a0._.js"
     // will be keyed by Node.js as "file:///APP/.next/server/chunks/ssr/[root-of-the-server]__2934a0._.js".
@@ -189,8 +196,8 @@ function getSourcemappedFrameIfPossible(
     //
     // But frame.file might also be "webpack-internal:///(rsc)/./app/bad-sourcemap/page.js" or
     // "<anonymous>" or "node:internal/process/task_queues" here
-    if (path.isAbsolute(frame.file)) {
-      sourceURL = url.pathToFileURL(frame.file).toString()
+    if (path.isAbsolute(sourceURL)) {
+      sourceURL = url.pathToFileURL(sourceURL).toString()
     }
     let maybeSourceMapPayload: ModernSourceMapPayload | undefined
     try {
@@ -204,7 +211,7 @@ function getSourcemappedFrameIfPossible(
       )
       // If loading fails once, it'll fail every time.
       // So set the cache to avoid duplicate errors.
-      sourceMapCache.set(frame.file, null)
+      sourceMapCache.set(fileKey, null)
       // Don't even fall back to the bundler because it might be not as strict
       // with regards to parsing and then we fail later once we consume the
       // source map payload.
@@ -228,10 +235,9 @@ function getSourcemappedFrameIfPossible(
       // is sufficient to compute relative paths but is actually wrong (the
       // chunk and sourcemap have different content hashes). We are using the
       // node API to read the sourcemap and it doesn't give us access to the
-      // URI. Devirtualize `about://React/Server/file:///path/to/chunk.js?4` to
-      // `file:///path/to/chunk.js` so that relative `sources` in the source map
-      // resolve against the real chunk URL, not the virtual one.
-      const sourceMapURL = devirtualizeReactServerURL(sourceURL) + '.map'
+      // URI. `sourceURL` is already devirtualized so that relative `sources`
+      // resolve against the real chunk URL, not React's virtual one.
+      const sourceMapURL = sourceURL + '.map'
       sourceMapConsumer = new SyncSourceMapConsumer(
         sourceMapPayload,
         // @ts-expect-error: our typings don't include this parameter but it is here.
@@ -245,10 +251,10 @@ function getSourcemappedFrameIfPossible(
       )
       // If creating the consumer fails once, it'll fail every time.
       // So set the cache to avoid duplicate errors.
-      sourceMapCache.set(frame.file, null)
+      sourceMapCache.set(fileKey, null)
       return createUnsourcemappedFrame(frame)
     }
-    sourceMapCache.set(frame.file, {
+    sourceMapCache.set(fileKey, {
       map: sourceMapConsumer,
       payload: sourceMapPayload,
     })
--- a/test/e2e/app-dir/server-source-maps/server-source-maps.test.ts
+++ b/test/e2e/app-dir/server-source-maps/server-source-maps.test.ts
@@ -537,6 +537,62 @@ describe('app-dir - server source maps', () => {
     }
   })
 
+  it('thrown errors from "use cache" have a sourcemapped stack with a codeframe', async () => {
+    // Errors thrown inside "use cache" functions cross a Flight boundary:
+    // the logged error object is deserialized by the consuming Flight client,
+    // and its stack is rebuilt through React's eval'd fake frame functions
+    // (`about://React/Cache/...` sourceURLs). This covers sourcemapping of
+    // fake stack frames, which resolve via the underlying chunk's source map.
+    if (isNextDev) {
+      const outputIndex = next.cliOutput.length
+      await next.render('/rsc-error-throw-cached')
+
+      await retry(() => {
+        expect(next.cliOutput.slice(outputIndex)).toContain(
+          'Error: rsc-error-throw-cached'
+        )
+      })
+      const cliOutput = normalizeCliOutput(next.cliOutput.slice(outputIndex))
+      // TODO(veil): Fake frames resolve their original source to a `file:`
+      // URL that's not normalized to a path relative to the project root
+      // like regular frames are.
+      expect(cliOutput).toMatch(
+        /at throwsInCache \([^)]*app\/rsc-error-throw-cached\/page\.js:5:9\)/
+      )
+      expect(cliOutput).toContain(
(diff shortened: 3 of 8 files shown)


packages/next/src/server/lib/source-maps.ts
```diff
@@ -183,9 +188,25 @@ export function findSourceMapURLDEV(
       )
     }
 
+    const origin = process.env.__NEXT_PRIVATE_ORIGIN
     if (sourceMapPayload === undefined) {
       sourceMapURL = invalidSourceMap
+    } else if (origin !== undefined) {
```
eps1lon, line 194: If we're using http: URLs, we should bail out earlier before we consult Node.js at all.

packages/next/src/server/lib/source-maps.ts
```diff
@@ -183,9 +188,25 @@ export function findSourceMapURLDEV(
...
+      // whole payload as a `data:` URL. React embeds the returned URL into
+      // every eval'd fake stack frame function, and each fake function has a
+      // unique sourceURL. With `data:` URLs, Node.js (with source maps
+      // enabled, as `next dev` enables them) parses and retains a separate
+      // copy of the full multi-MB source map payload per fake function in a
+      // never-evicted cache, which can consume gigabytes of heap on
+      // component-heavy apps in development. A URL reference costs ~100 bytes
+      // per fake function instead, can be cached per file by debugger
+      // frontends, and Node.js stores it without parsing anything.
+      sourceMapURL = `${origin.replace(/\/$/, '')}/__nextjs_source-map?filename=${encodeURIComponent(scriptNameOrSourceURL)}`
```
eps1lon, line 206: This needs to be gated on the dev server running. 

We should also double check how this works in Node.js. I suspect most of the sourcemapping is now race-y depending on how fast Node.js gets the sourcemap via http. `data:` and `file:` URLs are sync.

packages/next/src/server/patch-error-inspect.ts
```diff
@@ -177,20 +177,27 @@ function getSourcemappedFrameIfPossible(
...
 } {
-  const sourceMapCacheEntry = sourceMapCache.get(frame.file)
+  // React creates fake eval'd scripts for Server Component stack frames with
+  // virtual URLs like `about://React/Server/file:///path/to/chunk.js?42`.
+  // Their frame positions are padded to match the underlying chunk, so we can
+  // resolve them via the real chunk's source map. This keeps fake frames
+  // sourcemappable when they don't carry their own inlined source map, and
+  // dedupes the consumer cache across the many fake scripts per chunk.
+  const fileKey = devirtualizeReactServerURL(frame.file)
+  const sourceMapCacheEntry = sourceMapCache.get(fileKey)
```
eps1lon, line 187: Conceptually each environment could have been compiled differently and therefore would need a different sourcemap. This isn't an issue for top level directives but may break with inline directives.

packages/next/src/server/lib/source-maps.ts
```diff
@@ -172,6 +172,11 @@ const sourceMapURLs = new LRUCache<string | typeof invalidSourceMap>(
 export function findSourceMapURLDEV(
   scriptNameOrSourceURL: string
 ): string | null {
+  if (scriptNameOrSourceURL === '') {
+    // Anonymous frames don't have a source map. Bail explicitly since
+    // Node.js' `findSourceMap('')` can return an unrelated script's map.
```
eps1lon, line 177: This seems like a fix we should land separately if that's true. But sounds like a Node.js bug to me.

test/e2e/app-dir/server-source-maps/server-source-maps.test.ts
```diff
@@ -537,6 +537,62 @@ describe('app-dir - server source maps', () => {
...
+          'Error: rsc-error-throw-cached'
+        )
+      })
+      const cliOutput = normalizeCliOutput(next.cliOutput.slice(outputIndex))
+      // TODO(veil): Fake frames resolve their original source to a `file:`
+      // URL that's not normalized to a path relative to the project root
+      // like regular frames are.
+      expect(cliOutput).toMatch(
+        /at throwsInCache \([^)]*app\/rsc-error-throw-cached\/page\.js:5:9\)/
+      )
```
eps1lon: What's the TODO here?

eps1lon (changes requested):
Breaks stacks of errors in `use cache` when inspecting the Node.js process.

packages/next/src/server/lib/source-maps.ts
eps1lon, line 206: I was confused for a bit why this works because I thought Node.js doesn't support http: URL. The reason this works is because we fallback to the bundler for sourcemapping. So for any external this disables sourcemapping.

test/e2e/app-dir/server-source-maps/server-source-maps.test.ts
eps1lon: I think that's the assertion that shows the breakage:
```
⨯ Error: rsc-error-throw-cached
    at throwsInCache (test/e2e/app-dir/hello-world/file:/Users/sebbie/repos/next3/test/e2e/app-dir/hello-world/app/page.tsx:5:9)
```

Can you assert on the full stack here?

We also need to restore the assertions that https://github.com/vercel/next.js/commit/782d04db1f7c62892a8b76374c07b639ac50c13f removed. These were testing this specific behavior.

gaearon: Extracting this in https://github.com/vercel/next.js/pull/95938 so we have coverage

PR #95938, opened 2026-07-19
[test] Add source-mapping coverage of React's fake stack frame scripts in `use cache`

gaearon:
Moved to https://github.com/vercel/next.js/pull/95945

### DIFF
--- a/test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
+++ b/test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
@@ -1,0 +1,368 @@
+import * as fs from 'fs'
+import * as path from 'path'
+import * as url from 'url'
+import { nextTestSetup } from 'e2e-utils'
+import { retry } from 'next-test-utils'
+import WebSocket from 'ws'
+// Next.js' vendored synchronous source-map fork. Its type declarations
+// forward to the `source-map` package, which only resolves from within
+// packages/next, not from the test tsconfig — hence `require`.
+const {
+  SourceMapConsumer: SyncSourceMapConsumer,
+} = require('next/dist/compiled/source-map')
+
+interface InspectorTarget {
+  title: string
+  url: string
+  webSocketDebuggerUrl: string
+}
+
+/** Minimal Chrome DevTools Protocol client on top of `ws`. */
+class CDPSession {
+  private ws: WebSocket
+  private lastId = 0
+  private pending = new Map<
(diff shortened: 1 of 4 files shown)


test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      NODE_OPTIONS: '--inspect=0',
+    },
+  })
+
+  if (skipped) return
+
+  if (!isNextDev) {
+    it('is only relevant in development', () => {})
+    return
+  }
```
eps1lon: I think we can test prod behavior here as well. Valid sourcemaps are equally interesting for inspecting next build and next start. 

You can move it to test/production to cover dev and start but not deployment automatically.

test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
```diff
@@ -0,0 +1,262 @@
...
+  const { skipped, next, isNextDev } = nextTestSetup({
+    dependencies,
+    files: path.join(__dirname, 'fixtures/default'),
+    skipDeployment: true,
+    env: {
+      // Expose the inspector on a random port so that the test can observe
+      // the scripts React evals for fake Server Component stack frames the
+      // same way an attached debugger would.
+      NODE_OPTIONS: '--inspect=0',
+    },
```
eps1lon: next-start and next-dev have 1st class support for `--inspect` as a CLI arg e.g. `next start --inspect=0`.

next-build is currently missing since that needs some consideration with regards to each prerender worker and what other works might be interesting.

test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
```diff
@@ -0,0 +1,262 @@
...
+        }
+      }
+
+      // The rendered page's own Server Component frames must resolve to its
+      // original source file.
+      expect(
+        mappedSources.filter((source) =>
+          source.includes('rsc-error-log/page.js')
+        )
+      ).not.toEqual([])
```
eps1lon: Can we tighten this down a bit e.g. with some normalization based on `next.testDir`? We had issues in the past where paths were incorrectly concatenated which was hidden by assertions like this.

gaearon:
Moved to #95945 (same content, branch in-repo).

PR #95939, opened 2026-07-19
Fix instant validation blocking navigations

timneutkens:
### What?

- Run development instant validation only after the navigation response has finished.
- Let a newer request supersede stale validation work and yield between validation render attempts.
- Forward the cancellation signal into React validation prerenders so asynchronous validation work can stop promptly.
- Add focused scheduler/response tests and assert that development validation starts after the response finishes.

### Why?

Instant validation was detached from the response promise, but its additional React renders still ran on the same Node.js event loop. A subsequent navigation could therefore wait hundreds of milliseconds for stale diagnostic work from the previous request.

Instant validation still runs when the server is idle. When another navigation arrives, foreground request work now takes priority and the obsolete validation is discarded.

### How?

Each Cache Components development request receives a validation generation signal. Starting a newer request aborts the previous generation. Validation waits for the Node response `finish` event, yields through the event-loop poll phase between attempts, and combines the generation signal with React's existing prerender abort signals.

Tests:

- `pnpm --filter=next build`
- `pnpm --filter=next types`
- `pnpm exec jest --runTestsByPath packages/next/src/server/app-render/dev-validation-scheduler.test.ts packages/next/src/server/app-render/wait-for-response.test.ts`
- `NEXT_SKIP_ISOLATE=1 NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-turbo test/e2e/app-dir/instant-validation/suspense-boundaries.test.ts -t "valid - static prefetch - suspense around runtime and dynamic"`

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -179,7 +179,10 @@ import {
 } from '../client-component-renderer-logger'
 import { isNodeNextRequest, isNodeNextResponse } from '../base-http/helpers'
 import { waitForResponseToFinish } from './wait-for-response'
-import { yieldToForegroundRequest } from './dev-validation-scheduler'
+import {
+  beginDevValidation,
+  yieldToForegroundRequest,
+} from './dev-validation-scheduler'
 import { signalFromNodeResponse } from '../web/spec-extension/adapters/next-request'
 import {
   parseRelativeUrl,
@@ -1262,8 +1265,7 @@ async function generateDynamicFlightRenderResultWithStagesInDev(
   ctx: AppRenderContext,
   initialRequestStore: RequestStore,
   createRequestStore: (() => RequestStore) | undefined,
-  fallbackParams: OpaqueFallbackRouteParams | null,
-  devValidationSignal: AbortSignal | undefined
+  fallbackParams: OpaqueFallbackRouteParams | null
 ): Promise<RenderResult> {
   const {
     htmlRequestId,
@@ -1308,6 +1310,7 @@ async function generateDynamicFlightRenderResultWithStagesInDev(
   // instant configs exist, since we render all the layouts necessary to perform
   // the validation in those cases.
   const shouldValidate =
+    !ctx.isPrefetch &&
     !isBypassingCachesInDev(initialRequestStore, workStore) &&
     (initialRequestStore.isHmrRefresh === true ||
       (await anySegmentNeedsInstantValidationInDev(loaderTree)))
@@ -1425,7 +1428,6 @@ async function generateDynamicFlightRenderResultWithStagesInDev(
         prefetchStage,
       },
       requestAbortSignal,
-      devValidationSignal,
     })
     stream = result.stream
     debugChannel = result.debugChannel
@@ -2555,7 +2557,6 @@ async function renderToHTMLOrFlightImpl(
     enableTainting,
     cacheComponents,
     setIsrStatus,
-    devValidationSignal,
   } = renderOpts
 
   const { cachedNavigations } = renderOpts.experimental
@@ -2911,8 +2912,7 @@ async function renderToHTMLOrFlightImpl(
             ctx,
             requestStore,
             createRequestStore,
-            fallbackParams,
-            devValidationSignal
+            fallbackParams
           )
         } else if (cacheComponents && cachedNavigations) {
           // MARK: RSC cacheComponents
@@ -2959,8 +2959,7 @@ async function renderToHTMLOrFlightImpl(
             postponedState,
             metadata,
             undefined, // Prevent restartable-render behavior in dev + Cache Components mode
-            fallbackParams,
-            devValidationSignal
+            fallbackParams
           )
 
           return new RenderResult(stream, {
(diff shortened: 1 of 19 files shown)


unstubbable (commented):
The defer and yield pieces look right to me, and gating validation on `waitForResponseToFinish` seems like a solid way to keep it off the response's critical path. The part I think still needs work is the supersession.

`beginDevValidationRequest(this)` runs in `pipeImpl`, which as far as I can tell means any request through `pipe` cancels the previous validation, not just a follow-up navigation. That includes a second browser tab, a `curl` or agent probe, incidental requests like a favicon 404, pages-router or error renders, and even server actions (the POST still hits `pipeImpl`, but the action re-renders through `generateDynamicFlightRenderResult`, which doesn't run validation, so we'd cancel the current page's diagnostics without replacing them). Most of those are cases where we'd want the validation to finish rather than be cancelled.

Since the case we primarily care about is the fast navigation clicks, which all happen in one tab, I was wondering whether we could scope this per browser document instead of per dev server. Everything from one tab shares a single `htmlRequestId` (it starts as the document's `requestId`, is stored on `self.__next_r`, and gets echoed on every RSC, HMR, and prefetch request), so keying the generation on that (a `Map<string, AbortController>` instead of the per-server `WeakMap`) would still abort the previous validation on each fast click while leaving other tabs and probes alone. It might also let us move the call out of `pipeImpl` into the validation path, where we already know both the `htmlRequestId` and whether we're going to validate, so we'd only open a generation for foreground navigations and HMR refreshes that actually validate, and skip prefetches and server actions. We'd probably want to drop the map entry once the validation settles so it stays roughly one per open tab. Was the broader scope intentional, or does per-document seem reasonable?

We should also add an e2e that drives the fast successive navigation case directly, since that's what this is really about and there's no dedicated coverage for it yet. The `hmr-rsc-cancellation` suite might be a good pattern to borrow: it reads the `<VALIDATION_MESSAGE>` markers via `parseValidationMessages` to check the validation lifecycle, so we could fire a few quick navigations and assert the earlier ones abort. It's worth covering the opposite case too, where a validation should run to completion: loading a second page in another tab while the first page's validation is still in flight would be a good candidate, asserting the first one still reaches `validation_end` instead of `validation_aborted`.

Finally, `requestAbortSignal` might now be redundant in the validation path. I tried removing it from `validationAbortSignal` locally and the HMR cancellation test still passed, with the abort coming through the `waitForResponseToFinish` gate instead (it keys off the same early-`close` event that `requestAbortSignal` does). It's still needed for aborting the render, but for validation we could probably simplify to `validationAbortSignal = devValidationSignal` and note in a comment that `waitForResponseToFinish` covers the client-abort case.

timneutkens:
Addressed in fbe26d61a8.

- Validation generations are now created only in the staged validation path and keyed by `htmlRequestId`, so same-document navigations supersede stale work while other browser documents remain independent. Prefetches do not create generations; Server Actions continue to bypass this path.
- The insertion-ordered registry is hard-capped at 100 active document validations. Same-document replacement does not grow it, capacity evicts and aborts the oldest generation, and identity-guarded completion prevents stale cleanup from deleting a replacement. Setup failures, response close, completion, errors, aborts, supersession, and eviction all release their registry entry.
- The per-document generation signal is the validation signal. `requestAbortSignal` remains limited to the foreground HMR render, while `waitForResponseToFinish` handles response closure before detached validation begins.
- Test lifecycle markers now report `validation_aborted` when cancellation happens after `validation_start`. Unit coverage includes document isolation, stale callbacks, success/failure cleanup, replacement cardinality, and the >100-document bound. Cache Components E2E coverage verifies same-tab supersession, cross-document isolation, final completion, and the existing HMR response-close behavior in both Turbopack and Webpack where applicable.

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -4427,18 +4433,35 @@ function runDevValidationInBackground(
   createRequestStore: () => RequestStore,
   getPayload: (requestStore: RequestStore) => Promise<RSCPayload>,
   onError: (error: unknown) => void,
-  requestAbortSignal: AbortSignal | undefined
+  validationAbortSignal: AbortSignal,
```
unstubbable, line 4436: Some of the downstream function signatures can now be changed from `validationAbortSignal: AbortSignal | undefined` to `validationAbortSignal: AbortSignal` and we can drop the optional chaining.

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -5510,78 +5538,92 @@ async function stagedRenderWithCachesInDev({
...
+        cacheSignal,
+        ctx,
+        fallbackRouteParams,
+        prerenderResumeDataCache,
+        getDevRenderDidError,
+        createRequestStore,
+        getPayload,
+        onError,
+        validationGeneration.signal,
+        validationGeneration.finish
```
unstubbable, line 5600: Per the declaration this is a method so technically it's not correct to pass in the reference, although the implementation uses a function property so in practice it doesn't matter today. But it's a bit fragile. Should we just pass in the whole `validationGeneration` instance as a single argument instead?

unstubbable (approved):

packages/next/src/server/app-render/app-render.tsx
timneutkens: :robot: Addressed in #95998: validation-only helpers now require an AbortSignal and no longer use redundant optional checks. Helpers shared with build-time validation remain optional.

packages/next/src/server/app-render/app-render.tsx
timneutkens: :robot: Addressed in #95998: runDevValidationInBackground now receives the complete DevValidationGeneration and calls validationGeneration.finish() during finalization.

PR #95945, opened 2026-07-20
[test] Add source-mapping coverage of React's fake stack frame scripts in `use cache`

gaearon:
Adds tests verifying that`use cache` stacks get symbolicated.

We noticed insufficient coverage in https://github.com/vercel/next.js/pull/95936, which should have failed the tests (it broke symbolication) but tests passed. So this restores the missing stack assertions. This also adds some new coverage that specifically verifies that symbolicating via CDP also works.

---

<details>

### What

Test-only. Server Components debug stacks are materialized as eval'd fake frame functions (`about://React/<env>/<file>?<idx>` scripts) whose `//# sourceMappingURL=` is consumed by several parties: Node.js' source map cache, `patch-error-inspect` for terminal stacks, and attached debuggers. Nothing asserted that these maps stay resolvable or that fake frames keep printing source-mapped stacks — regressions here only surfaced when manually attaching a debugger (see the discussion in #95936).

- **`fake-frame-source-maps.test.ts` (new):** starts the server with the first-class `--inspect=0` flag (works for `next dev` and `next start`), attaches over CDP, and resolves script source maps the way a debugger frontend can — self-contained `data:` URL, a `file:`/relative reference read from disk, or served by the debugged target via `Network.loadNetworkResource`. In dev it maps each fake frame script's padded position back to its original source (exact `file://` URL match against the test dir, with a well-formedness check on every mapped source to catch mangled path concatenations). In `next start` it verifies the build output's declared source map references all resolve and that the rendered page's original source is resolvable from them (per-bundler exact match).
- **`rsc-error-throw-cached` fixture + test:** an error thrown inside `"use cache"` crosses a Flight boundary, so the logged error's stack is rebuilt through fake frames (`resolveErrorDev`); asserts the full source-mapped stack with a codeframe, plus that prod logs don't leak `about://React/` URLs.
- **`cache-components-dev-errors.test.ts`:** restores the CLI stack + codeframe assertion for blocking-route validation errors (reduced to a message-only check in 782d04db1f). Those stacks flow through Prerender-environment fake frames.

### Verification

Green on this branch in all four local mode/bundler combinations (`next dev` and `next start` × Turbopack and webpack). Ran against #95936's branch to confirm they catch the breakage discussed there:

- the CDP test fails with `Unable to load source map through the debugged target` / `Network resource loading is not enabled…` — the same failure an attached debugger hits,
- the `use cache` test fails on unnormalized fake frame paths (`at throwsInCache (file:/private/var/...)`, sometimes `../`-prefixed),
- the restored validation-error assertion fails on frames in bracket-named chunks leaking raw `about://React/...%5Broot-of-the-server%5D...` URLs.

</details>

### DIFF
--- a/test/development/app-dir/cache-components-dev-errors/cache-components-dev-errors.test.ts
+++ b/test/development/app-dir/cache-components-dev-errors/cache-components-dev-errors.test.ts
@@ -95,6 +95,17 @@ describe('Cache Components Dev Errors', () => {
     expect(stripAnsi(next.cliOutput.slice(outputIndex))).toContain(
       'https://nextjs.org/docs/messages/blocking-prerender-dynamic'
     )
+    // The logged error's stack is built from React's fake stack frame
+    // functions for the Prerender environment, and must be source-mapped
+    // with a codeframe like any other frame.
+    expect(stripAnsi(next.cliOutput.slice(outputIndex))).toContain(
+      '\n    at Page (app/no-accessed-data/page.js:2:9)' +
+        '\n  1 | export default async function Page() {' +
+        '\n> 2 |   await new Promise((r) => setTimeout(r, 200))' +
+        '\n    |         ^' +
+        '\n  3 |   return <p>Page</p>' +
+        '\n  4 | }'
+    )
 
     await expect(browser).toDisplayCollapsedRedbox(`
      {
--- a/test/development/app-dir/cache-components-dev-errors/cache-components-dev-errors.test.ts
+++ b/test/development/app-dir/cache-components-dev-errors/cache-components-dev-errors.test.ts
@@ -95,9 +95,6 @@ describe('Cache Components Dev Errors', () => {
     expect(stripAnsi(next.cliOutput.slice(outputIndex))).toContain(
       'https://nextjs.org/docs/messages/blocking-prerender-dynamic'
     )
-    // The logged error's stack is built from React's fake stack frame
-    // functions for the Prerender environment, and must be source-mapped
-    // with a codeframe like any other frame.
     expect(stripAnsi(next.cliOutput.slice(outputIndex))).toContain(
       '\n    at Page (app/no-accessed-data/page.js:2:9)' +
         '\n  1 | export default async function Page() {' +
--- a/test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
+++ b/test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
@@ -1,0 +1,368 @@
+import * as fs from 'fs'
+import * as path from 'path'
+import * as url from 'url'
+import { nextTestSetup } from 'e2e-utils'
+import { retry } from 'next-test-utils'
+import WebSocket from 'ws'
+// Next.js' vendored synchronous source-map fork. Its type declarations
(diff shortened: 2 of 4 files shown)


test/development/app-dir/cache-components-dev-errors/cache-components-dev-errors.test.ts
```diff
@@ -95,6 +95,17 @@ describe('Cache Components Dev Errors', () => {
     expect(stripAnsi(next.cliOutput.slice(outputIndex))).toContain(
       'https://nextjs.org/docs/messages/blocking-prerender-dynamic'
     )
+    // The logged error's stack is built from React's fake stack frame
+    // functions for the Prerender environment, and must be source-mapped
+    // with a codeframe like any other frame.
```
eps1lon:
```suggestion
```
comment provides no value.

test/e2e/app-dir/server-source-maps/fake-frame-source-maps.test.ts
```diff
@@ -0,0 +1,368 @@
+import * as fs from 'fs'
+import * as path from 'path'
+import * as url from 'url'
+import { nextTestSetup } from 'e2e-utils'
+import { retry } from 'next-test-utils'
+import WebSocket from 'ws'
+// Next.js' vendored synchronous source-map fork. Its type declarations
+// forward to the `source-map` package, which only resolves from within
+// packages/next, not from the test tsconfig — hence `require`.
```
eps1lon: Why do we need the vendored version? The less `next/dist` imports in tests the better since these are technically internals and tests are written against the public interface.

Can you use the `SourceMap` from `node:module` instead?

test/e2e/app-dir/server-source-maps/server-source-maps.test.ts
```diff
@@ -537,6 +537,53 @@ describe('app-dir - server source maps', () => {
     }
   })
 
+  it('thrown errors from "use cache" have a sourcemapped stack with a codeframe', async () => {
+    // Errors thrown inside "use cache" functions cross a Flight boundary:
+    // the logged error object is deserialized by the consuming Flight client,
+    // and its stack is rebuilt through React's eval'd fake frame functions
+    // (`about://React/Cache/...` sourceURLs). This covers sourcemapping of
+    // fake stack frames.
```
eps1lon:
```suggestion
```
Same here. This feature is covered in a lot of existing tests.

test/e2e/app-dir/server-source-maps/server-source-maps.test.ts
```diff
@@ -537,6 +537,48 @@ describe('app-dir - server source maps', () => {
...
+      const outputIndex = next.cliOutput.length
+      await next.render('/rsc-error-throw-cached')
+
+      await retry(() => {
+        expect(next.cliOutput.slice(outputIndex)).toContain(
+          'Error: rsc-error-throw-cached'
+        )
+      })
+      // Runtime stacks are not sourcemapped in production, but fake frames
+      // must not leak virtual `about://React/` URLs into logged stacks.
```
eps1lon: React will never produce fake frames in production either. Let's clarify this in this comment so that we don't give off the wrong impression.

eps1lon (approved):

PR #95951, opened 2026-07-20
Turbopack: Use `Arc<PathBuf>` for keys in turbo-task-fs's per-path `MutexMap`

bgw:
We clone these `PathBuf`s a bunch of times, so use `Arc`s

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/lib.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/lib.rs
@@ -244,7 +244,7 @@ struct DiskFileSystemInner {
     pub root: RcStr,
     #[turbo_tasks(debug_ignore, trace_ignore)]
     #[bincode(skip)]
-    mutex_map: MutexMap<PathBuf>,
+    mutex_map: MutexMap<Arc<PathBuf>>,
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -830,12 +830,12 @@ impl FileSystem for DiskFileSystem {
         if self.inner.is_path_denied(&fs_path) {
             return Ok(LinkContent::NotFound.cell());
         }
-        let full_path = self.to_sys_path(&fs_path);
+        let full_path = Arc::new(self.to_sys_path(&fs_path));
 
         self.inner.register_read_invalidator(&full_path).await?;
```
lukesandberg, line 835: the invalidator map should probably also take the Arc since it stores it in a map

lukesandberg (approved):

turbopack/crates/turbo-tasks-fs/src/lib.rs
bgw: https://github.com/vercel/next.js/pull/95987

PR #95960, opened 2026-07-20
examples: add with-next-i18next (App Router)

adrai:
### What?

App Router example for next-i18next v16: en/de with `proxy.ts` detection.
next-i18next v16 works with both the App Router and the Pages Router approach.

It shows:
- locale-prefixed routing (`/en`, `/de`) with detection in `proxy.ts`
- `getT` in Server Components incl. `generateMetadata`
- `useT` in Client Components incl. pluralization
- a `Trans` example
- a language switcher that keeps the current page
- bundler-traceable dynamic imports, so the example deploys to Vercel as-is

### Why?

Until v16, next-i18next supported only the Pages Router approach, because it
was relatively easy to use i18next and react-i18next directly — but it seems
this confused developers, so v16 fixed that. It shows everything important to
know about how to use next-i18next.

The examples directory covers next-intl, lingui, react-intl, next-translate
and rosetta, but none of the i18next family (i18next ~19.9 M, react-i18next
~13.9 M weekly npm downloads).

Related docs PR: #94744.

### How?

I built next-i18next on my own and tested it across various examples and also
on real Next.js projects.

This example: structure modeled on `with-ably`; `next build` passes on
Next.js 16.2.10; `next start` smoke test: `Accept-Language: de`
redirects `/` → `/de/`, server-translated content and metadata in both
locales; prettier-clean against the repo config.

### DIFF
--- a/examples/with-next-i18next/.gitignore
+++ b/examples/with-next-i18next/.gitignore
@@ -1,0 +1,40 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
+/.pnp
+.pnp.*
+.yarn/*
+!.yarn/patches
+!.yarn/plugins
+!.yarn/releases
+!.yarn/versions
+
+# testing
+/coverage
+
+# next.js
+/.next/
+/out/
+
+# production
+/build
+
+# misc
+.DS_Store
+*.pem
+
+# debug
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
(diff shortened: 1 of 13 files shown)


adrai:
Re-verified today against Next.js 16.3.0 and next-i18next 16.0.10: builds clean, locale routing, SSR translations and metadata all pass, and prettier matches `examples/.prettierrc.json`. The branch is rebased onto current canary.

Happy to adjust anything. This would be the first i18next-family example in the directory.

PR #95963, opened 2026-07-20
Add shell offset to static segment prefetches

acdlite:
Updates the per-segment, static prefetch response type to include a bytes offset for the App Shell. This will allow the client to extract a shell response from a more complete response, like we already do for page-level prerender responses.

All of the significant changes are confined to the server, during the step when a full page prerender is used to create per-segment responses (collect-segment-data). The overall strategy is the same one used at the page level. The input data is unblocked in distinct stages, separated by macrotasks. At each stage, we record the number of bytes that have been emitted to the output stream. Then, before closing the final stream, we embed the byte offsets into the response.

During this phase, no actual user code runs: all we're doing is decoding and reencoding an existing Flight response. So the timing logic is more straightforward. For example, we don't need to worry about dynamic data leaking into the result.

This commit is purely an in-place refactor. Subsequent steps will update the client to take advantage of the new offsets.

### DIFF
--- a/packages/next/src/server/app-render/collect-segment-data.tsx
+++ b/packages/next/src/server/app-render/collect-segment-data.tsx
@@ -11,16 +11,15 @@ import {
   PrefetchHint,
   StaticPrefetchDisabled,
 } from '../../shared/lib/app-router-types'
-import {
-  readVaryParams,
-  type VaryParamsIterable,
-} from '../../shared/lib/segment-cache/vary-params-decoding'
+import type { VaryParamsIterable } from '../../shared/lib/segment-cache/vary-params-decoding'
 import type { ManifestNode } from '../../build/webpack/plugins/flight-manifest-plugin'
 
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { createFromReadableStream } from 'react-server-dom-webpack/client'
 // eslint-disable-next-line import/no-extraneous-dependencies
 import { prerender } from 'react-server-dom-webpack/static'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { renderToReadableStream } from 'react-server-dom-webpack/server'
 
 import {
   streamFromBuffer,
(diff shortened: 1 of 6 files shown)


packages/next/src/server/app-render/collect-segment-data.tsx
```diff
@@ -107,33 +107,70 @@ export type SegmentPrefetchResponse = {
...
   isUpgradeableISRFallback: boolean
+  /**
+   * Shell byte boundary — the segment-level analogue of the route-level `a`.
+   * A promise because the value is only known mid-stream; its resolution row
+   * flushes past the boundary, so a truncated shell decode reads it as
+   * pending, which is harmless.
+   *
+   * - `> 0`: byte offset such that re-decoding the response truncated there
+   *   yields the shell variant of every segment — param-dependent content
+   *   reduced to still-pending references (which render as the param fallback).
```
unstubbable, line 118: same for cached content with a stale time below 5 minutes

packages/next/src/server/app-render/collect-segment-data.tsx
```diff
@@ -975,100 +1109,197 @@ function collectSegmentDataImpl(
...
+  // Wait for the metadata rows to flush before halting — two macrotasks,
+  // each a distinct hop:
+  //   1. streamInfoStage unblocks the isPartial probe renders; a static
+  //      segment's probe resolves within this task (a partial one never does,
+  //      which is how it stays pending → read as partial).
+  //   2. those resolutions (and the already-resolved `a`) ping the render,
+  //      which emits their rows; the consumer reads that chunk here.
+  // Halting after only the first hop would drop the not-yet-flushed metadata.
+  await waitAtLeastOneReactRenderTask()
+  await waitAtLeastOneReactRenderTask()
```
unstubbable, line 1285: Not blocking, but this makes me slightly uneasy: the hop count is empirical and a miss would degrade silently rather than fail. `runInSequentialTasks` isn't an option here with the concurrent segment renders, so I don't have a better suggestion right now.

unstubbable (approved):

packages/next/src/server/app-render/collect-segment-data.tsx
acdlite: Yeah I tried my best to make it more structural, this was the best I could come up with for now. Hopefully there's enough test coverage that if something degrades we'll catch it.

PR #95964, opened 2026-07-20
Track whether runtime data is accessed during prefetch

acdlite:
Adds a server-computed signal to static per-segment prefetch responses that tells the client whether a runtime prefetch request would return more content than the static response already contains.

The signal is computed by tracking whether the prerender accessed any data source that hangs during a static prerender but would resolve during a runtime prerender. For example: cookies, headers, fallback params, and search params.

The information is encoded two places:

- As a prefetch hint called ShouldAttemptStaticPrefetch. This tells the client that it's worth attempting to do a static prefetch instead of a runtime one. It's only an optimization, though: if the static response ends up being insufficient, the client will follow it up with a runtime request. It's semantically OK if there are false positives.
- Embedded in the static segment response. This tells the client whether the static response is missing data that would have been included in a runtime response. Unlike the prefetch hint, this value must never falsely claim that no runtime request is needed.

A follow-up will update the client to skip the runtime request using this information. This commit only encodes the signal into the response.

### DIFF
--- a/packages/next/src/client/components/segment-cache/scheduler.ts
+++ b/packages/next/src/client/components/segment-cache/scheduler.ts
@@ -855,10 +855,15 @@ function pingRootRouteTree(
             // opt-in because the Shell for a given route is reusable across
             // all given params, by definition. So it does not lead to an
             // explosion in prefetching costs.
-            // TODO: In the future, the server could emit a hint to tell us
-            // *not* to prefetch via a runtime request, via build-time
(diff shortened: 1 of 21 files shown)


unstubbable (approved):
Any reason `params` records the access at creation while `searchParams` records it in the proxy traps? Since the tracking is route-wide it probably works out the same in practice, just wondering whether the difference is intentional.

acdlite:
Good call, should do it the same for both

PR #95388, opened 2026-07-02
Better error message for invalid `generateStaticParams`

SukkaW:
The PR fixes https://github.com/vercel/next.js/discussions/56731

Currently, invalid `generateStaticParams` is being treated as missing `generateStaticParams`, resulting in a misleading error message and bad DX.

The PR introduced better error messages for invalid `generateStaticParams`.

e2e tests added, documentation added and referenced.

PR #95968, opened 2026-07-20
Throw when generateStaticParams returns invalid values

devjiwonchoi:
> [!TIP]
> This PR is best reviewed commit by commit.

### Why?

`generateStaticParams` must return an array of objects, but Next.js does not currently provide a clear error when it returns something else.

### How?

- Throw when `generateStaticParams` does not return an array.
- Throw when an item in the returned array is not an object.
- Keep allowing `{}`. The follow-up rejects it with `output: 'export'` when it leaves a dynamic route parameter missing.

Follow-up: #95969

Supersedes #95388.

Co-authored-by: SukkaW <isukkaw@gmail.com>

### DIFF
--- a/errors/generate-static-params.mdx
+++ b/errors/generate-static-params.mdx
@@ -1,0 +1,23 @@
+---
+title: 'Invalid `generateStaticParams` Value'
+---
+
+## Why This Error Occurred
+
+Your `generateStaticParams` function returned a value that is not an array. Next.js expects an array containing one params object for each route to generate.
+
+## Possible Ways to Fix It
+
+Return an array from `generateStaticParams`:
+
+```tsx filename="app/blog/[slug]/page.tsx"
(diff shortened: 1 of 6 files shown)


errors/generate-static-params.mdx
```diff
@@ -0,0 +1,28 @@
...
+---
+
+## Why This Error Occurred
+
+Next.js expects `generateStaticParams` to return an array containing one params object for each route to generate.
+
+This error occurs when:
+
+- `generateStaticParams` returns a value that is not an array, or
+- an item in the returned array is not an object. This includes `null`, arrays, primitive values, and `undefined`.
```
icyJoseph:
```suggestion
- an item in the returned array is not a plain object. This includes `null`, arrays, primitive values, and `undefined`.
```

icyJoseph (approved):

PR #95974, opened 2026-07-20
turbo-tasks: add scope_unbounded, a scoped execution primitive that allows more work to be discovered

lukesandberg:
## What

Adds `scope_unbounded `, a parallel scope to turbo-tasks: jobs run on a shared work-queue, any job may enqueue more, and the pool is drained by the runtime worker threads plus the calling thread until empty.

## Why

The garbage collector (later in this stack) needs to process a pool of work that *discovers more work as it runs* — collecting a task cascades into collecting newly-parentless children. A fixed `parallel::for_each` can't express that. `scope_self_feeding` is the general primitive for "parallel pool where jobs spawn jobs".

## Design notes

- **Deadlock-safe on thread-limited / contended runtimes.** Helper workers are a pure optimization: the calling thread always makes progress on the shared queue on its own, so the pool completes even when no worker threads are available. 
- Supports growing/shrinking the set of workers pulling from the queue based on the amount of work available.
- Supports a simple mostly lock free way to aggregate data from the tasks
- Supports a way for tasks to abort the whole queue, which is important for making GC interruptible

## Testing

There are a fair number of new unit tests and i have done some runs under miri


lukesandberg:
* **#95974** <a href="https://app.graphite.com/github/pr/vercel/next.js/95974?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95974?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+/// Shared, NON-generic state for a [`scope_self_feeding`] run: the counter, panic slot, and the
+/// work queue of erased per-item jobs.
+///
+/// The work queue stores one boxed closure per item — exactly like [`ScopeInner`] — rather than raw
+/// `T` values. This is a deliberate choice: it reuses `Scope::spawn`'s *proven* `'scope`->`'static`
+/// transmute (erasing a boxed closure, not a `T`), so `FeedingInner` needs no `T` generic and we
+/// never touch `dyn Any` / `T: 'static`. Each job closure captures its `T` and an `Arc` to the
+/// generic [`Feeding`] context (which owns the shared `run`), and when invoked it reconstructs a
+/// [`Spawner`] so `run` can enqueue children.
+struct FeedingInner {
```
lukesandberg: I don't care for the 'feeding' metaphore

this is really just a non-fixed work queue

`unbounded`? might be a better term

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+    main_thread: Thread,
+    /// Number of items that have been enqueued but not yet finished processing. An item is counted
+    /// from the moment it is pushed onto `work_queue` until its job closure returns (or panics).
+    /// The scope is complete exactly when this reaches zero. See `Spawner::spawn` /
+    /// `on_item_finished` for the ordering that makes zero a reliable "all done" signal.
+    remaining_tasks: AtomicUsize,
+    /// The first panic that occurred while processing an item. Unlike the fixed-set scope there is
+    /// no meaningful per-item index to order panics by, so we keep the *first* panic to be
+    /// recorded (ties broken by lock order).
```
lukesandberg: don't describe how this differs, describe how we use it

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+    panic: Mutex<Option<Box<dyn Any + Send + 'static>>>,
+    /// Queue of jobs awaiting processing. Each is a self-contained closure erased to `'static`
+    /// (see `Spawner::spawn`).
+    work_queue: Mutex<VecDeque<Box<dyn FnOnce() + Send + 'static>>>,
+    /// Notifies parked workers of newly-enqueued work, and of termination (via `notify_all` once
+    /// `remaining_tasks` hits zero).
+    work_queue_condition_var: Condvar,
+    /// Set once `remaining_tasks` reaches zero, so workers parked on the condvar with an empty
+    /// queue can distinguish "more work coming" from "we are done". Checked under the queue lock.
+    done: std::sync::atomic::AtomicBool,
```
lukesandberg: import the type

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+    }
+
+    /// Latches `done` and wakes everyone. Called when the counter is (or has just become) zero.
+    ///
+    /// `done` is set and the notifications issued while holding the queue lock, so it is impossible
+    /// for a worker to observe an empty queue, decide to wait, and only then miss the
+    /// `done`/`notify_all` — a classic lost-wakeup. Holding the lock serializes against
+    /// `pick_item`'s empty-queue-check-then-wait sequence. Idempotent: safe to call more than once
+    /// and from more than one thread (e.g. a finishing worker and the post-fill idle check racing).
+    fn latch_done_and_wake(&self) {
```
lukesandberg: simplify doc comment

also the name should probably be something like `close`

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+        }
+        // `Release` pairs with the `Acquire` loads in `wait()` and `pick_item`, so that once a
+        // thread observes zero / `done` it also observes every prior queue and panic write.
+        if self.remaining_tasks.fetch_sub(1, Ordering::Release) == 1 {
+            // This decrement brought the count to zero: the scope is done.
+            self.latch_done_and_wake();
+        }
+    }
+
+    /// Called by the calling thread once the initial fill is complete. If no work is (or is still)
```
lukesandberg: Called by the calling thread, is odd.

simplify this comment

is it called by the Main thread? if it is.... is this necessary wouldn't the worker thread have already called `on_item_finished`? 

is this about a race betwween `fetch_sub` and `latch_done`?

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+    fn finalize_if_idle(&self) {
+        if self.remaining_tasks.load(Ordering::Acquire) == 0 {
+            self.latch_done_and_wake();
+        }
+    }
+
+    /// Worker/main-thread drain loop. Pulls jobs and runs them until the scope terminates
+    /// (`remaining_tasks == 0`, surfaced via `done`). Both helpers and the calling thread run this;
+    /// helpers are a pure optimization, exactly as in the fixed-set scope.
+    fn run_items(&self) {
```
lukesandberg: `work` or `run` are reasonable simpler names

the doc comment can be reduced to just the first sentence

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+                drop(work_queue);
+                return Some(job);
+            }
+            // Queue is empty. If the scope is done, stop; otherwise wait for new work or the final
+            // wakeup. `done` is read under the queue lock, so it cannot be set between our
+            // emptiness check and going to sleep (see `on_item_finished`).
+            if self.done.load(Ordering::Acquire) {
+                return None;
+            }
+            self.work_queue_condition_var.wait(&mut work_queue);
```
lukesandberg: should we time out? a long running tail task can cause all workers to get stuck while it runs...

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+/// into the same scope.
+///
+/// `'scope` ties the borrowed enqueue callback to a single `run` invocation; `'env` is the
+/// caller-visible lifetime of the borrowed data captured by `run` and the items.
+///
+/// The concrete `run` closure type is hidden behind a single type-erased "enqueue one item"
+/// callback (`dyn Fn(T) + 'env`), so callers only ever name `Spawner<'_, 'env, T>`. Crucially this
+/// callback is `'env`-typed (not `'static`), which is what lets the internal `Spawner`s built by
+/// worker threads type-check against the user's `run: Fn(&Spawner<'_, 'env, T>, T)` without any
+/// lifetime transmute of the `Spawner` itself.
```
lukesandberg: simplify comment

also given how the usecase has evolved, we could perhaps demand a single concrete type for the enqueued task

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+/// Liveness matches `scope_and_block`: the calling thread participates in draining and will drain
+/// the entire (growing) queue by itself if no helper is ever scheduled, so this does not deadlock
+/// on a thread-limited or otherwise-contended runtime. The `block_in_place` fallback in `wait()` is
+/// reached only after the calling thread can make no further progress on its own.
+///
+/// As with `scope_and_block`, prefer calling this from within `spawn_blocking` if other work is
+/// running concurrently in the same task, since `block_in_place` will suspend it.
+///
+/// The first panic raised by any `run` invocation is propagated to the caller after all in-flight
+/// work has been joined.
```
lukesandberg: again the `feeding` name needs thought

surely rayon has a similar api maybe we can steal a name

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,490 @@ where
...
+        inner: Arc<FeedingInner>,
+        helper_handles: Vec<tokio::task::JoinHandle<()>>,
+    }
+    impl Drop for Joiner {
+        fn drop(&mut self) {
+            // The initial fill is complete (normally or via panic-unwind). If nothing is in flight
+            // — empty initial set, or helpers already drained everything — latch `done` now so the
+            // drain loops below do not block forever waiting for an `on_item_finished` that will
+            // never come.
+            self.inner.finalize_if_idle();
```
lukesandberg: check if this is actually needed, run_items should take care of it

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Renamed to `scope_unbounded` (the queue is unbounded/growing vs the fixed-set `scope_and_block`). `FeedingInner`→`UnboundedInner`.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Left as-is for now with a TODO in `wait()` — a timeout/steal strategy to preempt a long tail task is a real design change; flagging rather than solving it in this PR.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Investigated: the per-item boxing isn't strictly necessary. It can be a `VecDeque<T>` with a single shared `run` if we bound `T: 'static` (which the only consumer, `GcJob`, satisfies) — that drops the per-item `Box`. Without the `'static` bound it needs HRTBs + an extra spawner-lifetime transmute, which is *more* subtle, not less. Kept the boxed-closure form here; happy to add the `T: 'static` simplification as a follow-up if you'd prefer it.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Renamed to `close_if_idle` and trimmed. It IS needed, for the empty-initial (or already-fully-drained-by-helpers) case: there `on_item_finished` never fires (or already fired), so nothing else latches `done` and the drain loop would park forever. `drain`/`pick` can't cover it because an empty queue is not termination while `done` is unset — that's the whole point of the `done` flag. It's the race between `fetch_sub` reaching zero and this post-fill idle check, resolved by both running `close` under the queue lock (idempotent).

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Renamed `latch_done_and_wake`→`close` and simplified the doc.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Renamed `run_items`→`drain` and cut the doc to one line. (Kept `drain` rather than `run` to avoid colliding with the user's `run` closure.)

lukesandberg:
Pushed the review pass:
- Renamed `scope_self_feeding`→`scope_unbounded` and the internals (`UnboundedInner`, `drain`, `pick`, `close`, `close_if_idle`).
- Trimmed the doc comments substantially (−148 lines net), keeping only the load-bearing termination-ordering and transmute-safety notes.
- TODO added on the long-tail-task starvation question.
- Concrete-type feasibility answered inline (doable with a `T: 'static` bound; deferred).

Replied inline to each thread. 16 scope tests still pass.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Went ahead and did it — the item type (`GcJob`) is `'static`, so bounding `T: 'static` on `scope_unbounded` is free for the consumer and unlocks the clean design: the queue now holds raw `T` (no per-item `Box<dyn FnOnce>`), the whole `Ctx`/`make_job`/`enqueue_item` machinery is gone, `Spawner` drops its `'env` param, and there's a single `'env`->`'static` transmute on the shared `run` instead of one per item. Net ~−190 lines in scope.rs and one `unsafe` block instead of per-item. (`Copy` itself wasn't the lever — `'static` was — but your pointing at the job type is what surfaced it.)

lukesandberg:
Follow-up: took the concrete-type simplification (bound `T: 'static`, which the `GcJob` consumer satisfies). scope.rs is now ~190 lines lighter — raw-`T` queue, no per-item boxing, single `run` erasure, `Spawner<'scope, T>`. 16 tests still green; the stack's `gc.rs` updated for the 2-param `Spawner`.

lukesandberg:
Fixed the `clippy::type_complexity` warnings with a `RunFn<'run, T>` type alias for the boxed closure.

Re: making `UnboundedInner` generic over the closure to capture `run` by value and drop the box — I looked at it, but it doesn't pay off:
- It doesn't remove the lifetime laundering, it just moves it: storing `run: F` where `F: 'env` makes `UnboundedInner<T, F>: 'env`, so the helper tasks (which need `'static`) still require an `'env`->`'static` transmute — now on the whole `Inner` instead of on `run`.
- Worse, `F` then infects every type that names `Inner` — including the **public** `Spawner`, which would become `Spawner<'scope, T, F>` (leaking an unnameable closure type into the API). Avoiding that needs a `Shared`/`run` struct split, which is more machinery than the box removes.
- The box is a single once-per-scope allocation (`run` is boxed once, not per item), so the cost is negligible.

So: kept the box + alias.

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -309,9 +309,295 @@ where
...
+    /// Spin-park briefly, then `block_in_place` so tokio can reuse this core. Reached by the main
+    /// thread only after it has drained everything it could.
+    fn wait(&self) {
+        if self.remaining_tasks.load(Ordering::Acquire) == 0 {
+            return;
+        }
+
+        let _span = info_span!("blocking").entered();
+
+        // TODO: a single long-running tail item can leave other drainers parked here while it runs.
```
lukesandberg: this TODO is in the wrong location, it should be on the `pick` site

lukesandberg: :robot: Done — moved the TODO to the `pick()` park site (where an idle helper actually waits with no work to steal); the main-thread `wait()` just blocks on the counter, so it was the wrong spot.

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -264,9 +265,319 @@ where
...
+                thread::park();
+            }
+        });
+    }
+}
+
+/// Handle passed to the `run` closure of [`scope_unbounded`], used to enqueue additional items into
+/// the same scope. Only borrows the scope's shared state; the items it enqueues are `T: 'static`,
+/// so it carries no `'env` lifetime of its own.
+pub struct Spawner<'scope, T: Send + 'static> {
```
bgw: Bikeshedding: `std::thread` calls this object `Scope`: https://doc.rust-lang.org/stable/std/thread/struct.Scope.html

But I guess we've already got a type here with the same name. Maybe this should be something like `ScopeUnbounded` and the other one should be `ScopeBounded` (e.g. see the names on https://docs.rs/tokio/latest/tokio/sync/mpsc/index.html for inspiration)? Or maybe these types should live in separate modules and just have the same name?

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -264,9 +265,319 @@ where
...
+            thread::park_timeout(timeout_remaining);
+            if self.remaining_tasks.load(Ordering::Acquire) == 0 {
+                return;
+            }
+            let elapsed = beginning_park.elapsed();
+            if elapsed >= TIMEOUT {
+                break;
+            }
+            timeout_remaining = TIMEOUT - elapsed;
+        }
```
bgw: I don't think the complexity of trying to avoid calling `block_in_place` is worth it. If the caller is using the `scope` API, presumably it's because they intend to do a significant amount of blocking work that will take >1ms.

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -264,9 +265,319 @@ where
...
+    /// `'static` here; see the `SAFETY` comment there.
+    run: RunFn<'static, T>,
+}
+
+impl<T: Send + 'static> UnboundedInner<T> {
+    /// Counts a newly-enqueued item. MUST run before the push; see [`enqueue`].
+    #[inline]
+    fn account_new_item(&self) {
+        self.remaining_tasks.fetch_add(1, Ordering::Relaxed);
+    }
```
bgw: Don't need the helper function, just inline this at the two callsites?

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -264,9 +265,319 @@ where
...
+/// scope returns as soon as the currently-running jobs finish. Jobs already in flight on other
+/// threads are **not** interrupted. Use this when the remaining work is discardable (it can be
+/// recomputed on a later run) and finishing it is not worth the latency.
+///
+/// TODO: this occupies every worker for the whole call duration. If work turns out to be bursty,
+/// let workers time out when there is not enough work and spawn more when new work is produced.
+pub fn scope_unbounded<'env, T, F>(initial: impl IntoIterator<Item = T>, run: F)
+where
+    T: Send + 'static,
+    F: Fn(&Spawner<'_, T>, T) -> ControlFlow<()> + Send + Sync + 'env,
```
bgw: Would it make more sense to more closely match the `std::thread::scope` API? https://doc.rust-lang.org/stable/std/thread/fn.scope.html

That API doesn't take `initial`, just just takes a closure that receives the `Scope`/`Spawner` object. If the caller wants to perform control flow, it just does it directly in that closure instead of returning a `ControlFlow` object.

That would also bring us closer in line with the bounded version of this API (`scope_and_block`).

turbopack/crates/turbo-tasks/src/scope.rs
bgw: Should this be something like `scope_bounded` now?

turbopack/crates/turbo-tasks/src/scope.rs
```diff
@@ -264,9 +265,319 @@ where
     })
 }
 
+// ---------------------------------------------------------------------------
+// Unbounded scope
+// ---------------------------------------------------------------------------
```
bgw: Maybe we should split this into `scope/bounded.rs` and `scope/unbounded.rs`

turbopack/crates/turbo-tasks/src/scope.rs
bgw: Do we also need this on `Spawner`?

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Changed the object name to Scope and moved to a new module 

also renamed `scope_and_block` to `scope_bounded` to keep things parallel

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Done

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: Done.

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: The Control flow type is being used for aborts which is a little distinct

This API we have does simplify the implementation quite a bit since we have one closure and a simple `'static` job token

so i think a separate name would be the best solution here, for the time being it is in a new module, but rather than mirroring the thread::scope api i guess i would prefer a `loop` terminology

This is kind of like a `while` loop or a generator, im not sure what a better name would be though

turbopack/crates/turbo-tasks/src/scope.rs
lukesandberg: i think the answer is no because the 'spawn' primitive doesn't accept closures that might infer new smaller scopes

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+///
+/// This is both the spawn budget and the join set: a slot is occupied from before its helper task
+/// exists until after that helper's last access to the caller's frame, so an empty `occupied` is
+/// exactly the condition [`Joiner::drop`] needs. See [`UnboundedInner::helper_slots`] for why one
+/// field can answer both questions.
+struct HelperSlots {
+    /// `Some` for a live helper, `None` for a free slot or one whose handle has been aborted and
+    /// dropped. Length is the cap and never changes, so a slot index stays valid for the whole
+    /// scope.
+    handles: Vec<Option<tokio::task::AbortHandle>>,
```
bgw: Import `tokio::task::AbortHandle`

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+//! `initial` iterator, and [`ControlFlow::Break`] rather than local control flow for early
+//! termination: aborting has to reach the *shared* queue, not just return from one invocation.
+
+use std::{
+    any::Any,
+    ops::ControlFlow,
+    panic::{self, AssertUnwindSafe, catch_unwind},
+    sync::{
+        atomic::{AtomicBool, AtomicUsize, Ordering},
+        mpmc::{self, Receiver, Sender},
```
bgw: Very nitpicky: I prefer not to import `mpsc::Receiver` and `mpsc::Sender`, and only import `mpsc` instead, because otherwise it's horribly ambiguous what `Receiver` and `Sender` is referring to (e.g. it could be `oneshot` or any number of other channel-like APIs)

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+/// `'run` is the lifetime of the borrows held by the `run`/`init`/`merge` closures (`'env` at the
+/// call site). It stays a real lifetime here rather than being pinned to `'static` so the fields
+/// don't each force `R: 'static`; the single erasure to `'static` happens at the `Drainable`
+/// hand-off to tokio, where it is justified by the join in `Joiner::drop`.
+struct UnboundedInner<'run, T: Send + 'static, R> {
+    /// Items enqueued but not yet finished. The scope is done exactly when this reaches zero; see
+    /// [`enqueue`] for the increment-before-finish ordering that makes zero reliable.
+    remaining_tasks: AtomicUsize,
+    /// First panic raised while processing an item; propagated to the caller after the join.
+    panic: Mutex<Option<Box<dyn Any + Send + 'static>>>,
```
bgw: This could be a `OnceLock`?

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+    slot: usize,
+}
+
+impl Drop for HelperGuard<'_> {
+    fn drop(&mut self) {
+        let mut slots = self.slots.lock();
+        slots.release(self.slot);
+        if slots.is_idle() {
+            // Still holding the lock: the predicate the joiner waits on is `!is_idle()`, read
+            // under this same lock, so it cannot go true between its check and its `wait`.
```
bgw: Just put `guard` in the variable name and then we don't need this comment

```suggestion
        let mut slots_guard = self.slots.lock();
        slots_guard.release(self.slot);
        if slots_guard.is_idle() {
```

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+        }
+
+        // Fold this drainer's accumulator into the shared results slot. This is the last access to
+        // the caller's frame, and a helper's `HelperGuard` releases its slot just after — which is
+        // what publishes this merge to the joiner, since both happen under `helper_slots`.
+        //
+        // `merge` is caught because an escaping panic has nowhere good to go: on a helper it would
+        // be swallowed at the task boundary, silently dropping these results, and on the
+        // calling thread `drain` runs from `Joiner::drop`, where unwinding during an unwind
+        // aborts the process.
```
bgw: I think this is all pretty clear from the code

```suggestion

```

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+            }
+            let spawner = Scope { inner: self };
+            // `init` must be inside the `catch_unwind`: it is user code, and it runs before this
+            // item's `on_item_finished`. An unwind past that decrement would strand the item —
+            // `remaining_tasks` never reaches zero, the queue never closes, and the calling
+            // thread's `recv()` blocks forever.
+            let result = catch_unwind(AssertUnwindSafe(|| {
+                let acc = acc.get_or_insert_with(self.init);
+                (self.run)(&spawner, item, acc)
+            }));
```
bgw:
```suggestion
            let result = catch_unwind(AssertUnwindSafe(|| {
                let acc = acc.get_or_insert_with(self.init);
                (self.run)(&spawner, item, acc)
            }));
```

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+            .enable_time()
+            .build()
+            .unwrap();
+        runtime.block_on(async { tokio::task::spawn_blocking(body).await.unwrap() })
+    }
+
+    // -----------------------------------------------------------------------
+    // scope_unbounded tests
+    // -----------------------------------------------------------------------
+
```
bgw: What else would this be? We're in a `mod tests` block inside of the `scope_unbounded` module

```suggestion
```

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -0,0 +1,1393 @@
...
+        time::Duration,
+    };
+
+    use super::*;
+
+    /// Runs `body` on a runtime with the I/O driver disabled, so the test also works under Miri.
+    ///
+    /// `#[tokio::test]` hardcodes `Builder::enable_all()`, whose I/O driver calls
+    /// `kqueue()`/`epoll_create1()` — Miri has no shim for either, so such a test aborts at runtime
+    /// construction before reaching any code under test. Nothing here needs I/O.
```
bgw, line 492: Should we have a way to run unit tests in CI under Miri? That would be neat. Doesn't need to be this PR.

bgw (approved):

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
lukesandberg: @mischnic is on the case!

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
lukesandberg: good idea

PR #95975, opened 2026-07-20
turbo-tasks-backend: add persistence delete/tombstone plumbing for GC

lukesandberg:
## What

The persistence-layer mechanism to tombstone (delete) tasks from the on-disk cache. A GC-collected task is represented as a `SnapshotItem::Delete` that rides the same streaming iterator `save_snapshot` already consumes for puts, so tombstones are applied in the same commit/batch as the snapshot with no side-channel.

- `SnapshotItem` becomes an enum: `Put { … }` (the existing path) and `Delete(TaskDeletion)`.
- `TurboWriteBatch::delete/value_delete` passthrough to `turbo_persistence`.
- `save_snapshot` applies deletes inline in the parallel put phase

## Note

`SnapshotItem::Delete` is `#[allow(dead_code)]` here: this PR lands the persistence-side *handling* and its tests, but nothing *emits* a `Delete` yet. The GC pass that soft-deletes tasks and serializes them to `Delete` lands in the next PR in the stack. So this PR is a no-op at runtime.

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
@@ -8,15 +8,50 @@ use turbo_tasks_hash::Xxh3Hash64Hasher;
 
 pub type TaskTypeHash = [u8; 8];
 
-/// A single item yielded by the snapshot iterator during persistence.
-pub struct SnapshotItem {
+/// A single item yielded by the snapshot iterator during persistence: either a put (persist a
+/// modified task's meta/data + optionally register a new task's type) or a delete (tombstone a
+/// GC-collected task's on-disk copy). Both ride the one streaming iterator `save_snapshot`
+/// consumes, so GC tombstones are applied in the same commit and batch as the puts without a
+/// side-channel that would have to be fully materialized before the put loop.
+pub enum SnapshotItem {
+    Put {
+        task_id: TaskId,
+        /// Serialized task meta data, if modified
+        meta: Option<TurboBincodeBuffer>,
+        /// Serialized task data, if modified
+        data: Option<TurboBincodeBuffer>,
+        /// Task type for new tasks that need to be added to the task cache
+        task_type_hash: Option<TaskTypeHash>,
+    },
+    /// Tombstone a GC-collected task (see [`TaskDeletion`]).
+    ///
+    /// Not yet constructed: this PR lands the persistence-side delete mechanism (the
+    /// `save_snapshot` handling below + its tests). The GC pass that emits `Delete` for
+    /// soft-deleted tasks lands in a later PR in the stack.
+    #[allow(dead_code)]
+    Delete(TaskDeletion),
+}
+
+impl SnapshotItem {
+    /// The task this item persists or tombstones. (Currently only used by tests, which assert on
+    /// the id of items yielded by the snapshot iterator.)
+    #[cfg(test)]
+    pub fn task_id(&self) -> TaskId {
+        match self {
+            SnapshotItem::Put { task_id, .. } => *task_id,
+            SnapshotItem::Delete(TaskDeletion { task_id, .. }) => *task_id,
+        }
+    }
+}
+
+/// A GC-collected task to tombstone from persistent storage, applied in the same commit as the
+/// snapshot. See `save_snapshot` for how the tombstone is applied (the `TaskCache` bucket
+/// re-insertion of hash-colliding survivors lives there).
+pub struct TaskDeletion {
     pub task_id: TaskId,
-    /// Serialized task meta data, if modified
-    pub meta: Option<TurboBincodeBuffer>,
-    /// Serialized task data, if modified
-    pub data: Option<TurboBincodeBuffer>,
-    /// Task type for new tasks that need to be added to the task cache
-    pub task_type_hash: Option<TaskTypeHash>,
+    /// The deleted task's `TaskCache` key. Always present: only persistent tasks are collected,
+    /// and those always have a task type.
+    pub task_type_hash: TaskTypeHash,
 }
 
 /// Computes a deterministic 64-bit hash of a CachedTaskType for use as a TaskCache key.
--- a/turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
@@ -10,9 +10,8 @@ pub type TaskTypeHash = [u8; 8];
 
 /// A single item yielded by the snapshot iterator during persistence: either a put (persist a
 /// modified task's meta/data + optionally register a new task's type) or a delete (tombstone a
-/// GC-collected task's on-disk copy). Both ride the one streaming iterator `save_snapshot`
-/// consumes, so GC tombstones are applied in the same commit and batch as the puts without a
-/// side-channel that would have to be fully materialized before the put loop.
+/// GC-collected task's on-disk copy). Both ride the one iterator `save_snapshot` consumes, so
+/// tombstones are applied in the same commit and batch as the puts.
 pub enum SnapshotItem {
     Put {
         task_id: TaskId,
@@ -23,23 +22,8 @@ pub enum SnapshotItem {
         /// Task type for new tasks that need to be added to the task cache
         task_type_hash: Option<TaskTypeHash>,
     },
-    /// Tombstone a GC-collected task, applied in the same commit as the surrounding puts.
-    ///
-    /// This is identity-only — no copy of the `TaskMeta`/`TaskData` buffers or the full
-    /// `CachedTaskType` behind the hash — because that is all the deletion needs:
-    ///
-    /// - `TaskMeta` and `TaskData` are `SingleValue`, so a key-granular delete is exact.
-    /// - `TaskCache` is `MultiValue` and keyed by the type hash, so deleting the key would also
-    ///   drop any task type that xxh3-collides with this one. The task id is the value in that
-    ///   bucket, so `delete_value` names exactly the one mapping to remove.
-    ///
-    /// Every part of this is a buffered write; nothing is read back at commit time. Compaction
-    /// drops the tombstoned entry the next time it rewrites the key group, and reads filter it out
-    /// until then.
-    ///
-    /// Not yet constructed: this PR lands the persistence-side delete mechanism (the
-    /// `save_snapshot` handling below + its tests). The GC pass that emits `Delete` for
-    /// soft-deleted tasks lands in a later PR in the stack.
+    // Constructed by the GC pass that emits `Delete` for soft-deleted tasks, which lands in a
+    // later PR in the stack.
     #[allow(dead_code)]
     Delete {
         task_id: TaskId,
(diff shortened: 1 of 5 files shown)


lukesandberg:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/vercel/next.js/95975?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#95975** <a href="https://app.graphite.com/github/pr/vercel/next.js/95975?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95975?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
```diff
@@ -8,15 +8,59 @@ use turbo_tasks_hash::Xxh3Hash64Hasher;
...
+///
+/// `TaskMeta` and `TaskData` are `SingleValue` families, so the task id key is tombstoned directly.
+/// `TaskCache` is `MultiValue` keyed by the 8-byte task-type hash, where a tombstone erases the
+/// *whole* hash bucket. If another live task's type xxh3-collides with this one it shares the
+/// bucket, so the apply path must re-insert those survivors after the tombstone. Rather than carry
+/// them on this struct, the survivors are resolved at apply time in `save_snapshot`, which reads
+/// the authoritative on-disk bucket (`get_multiple`) and re-puts everything except the ids being
+/// deleted — the in-memory task cache is not a reliable source (it is lazily populated and keyed by
+/// the full task type, not the hash). Collisions between two simultaneously-live tasks are
+/// astronomically rare (~2^32 tasks for one expected collision), so this path almost never fires.
```
lukesandberg: this comment is very redundant

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -236,20 +240,48 @@ impl TurboBackingStorage {
...
-            let mut snapshot_meta =
+            let span = tracing::trace_span!("update task data");
+            // GC-soft-deleted tasks serialize to a `SnapshotItem::Delete` rather than a put. The
+            // parallel put phase applies their SingleValue TaskMeta/TaskData tombstones inline (see
+            // the `Delete` arm) and returns each `TaskDeletion` in `shard_deletes` so the
+            // sequential phase below can apply the TaskCache tombstone — that one
+            // erases a whole MultiValue hash bucket and re-inserts survivors, a
+            // read-modify-write that must run after all puts. These merge with the
+            // caller-supplied deletes (from the test-only `gc_for_testing` hook), which
+            // arrive as their own shard.
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -236,20 +240,48 @@ impl TurboBackingStorage {
...
+                            } => (task_id, meta, data, task_type_hash),
+                            SnapshotItem::Delete(deletion) => {
+                                // TaskMeta/TaskData are SingleValue, so tombstoning is a direct
+                                // per-id delete with no cross-shard read-modify-write — issue it
+                                // inline here, on the same thread-local collector the puts use (a
+                                // deleted id is never also emitted as a put in the same commit: GC
+                                // removes tasks from the map before the snapshot serializes). Only
+                                // the TaskCache tombstone must be deferred to the sequential phase
+                                // below (it erases a whole MultiValue hash bucket and re-inserts
+                                // survivors), so we still carry `deletion` out for that.
```
lukesandberg: simplify

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -279,28 +311,101 @@ impl TurboBackingStorage {
...
+                    |(task_type_hash, deleted_ids)| {
+                        let _span = span.clone().entered();
+                        // Read the authoritative on-disk bucket so we can keep every id we are NOT
+                        // deleting. (A survivor that is a *new* task added in this same commit is
+                        // covered by its own put above — new tasks aren't on disk yet, so they
+                        // can't appear here and won't be double-inserted.)
+                        let bucket = database
+                            .get_multiple(KeySpace::TaskCache, &task_type_hash)
+                            .with_context(|| {
+                                format!(
```
lukesandberg: consider moving this out into another method for readability

turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
lukesandberg: :robot: Done — trimmed the `TaskDeletion` doc; the bucket-rewrite mechanics now live at the `save_snapshot` apply site.

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
lukesandberg: :robot: Done — extracted the parallel TaskCache bucket-rewrite into a dedicated `apply_task_cache_deletions` helper.

turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
```diff
@@ -8,15 +8,50 @@ use turbo_tasks_hash::Xxh3Hash64Hasher;
...
+        match self {
+            SnapshotItem::Put { task_id, .. } => *task_id,
+            SnapshotItem::Delete(TaskDeletion { task_id, .. }) => *task_id,
+        }
+    }
+}
+
+/// A GC-collected task to tombstone from persistent storage, applied in the same commit as the
+/// snapshot. See `save_snapshot` for how the tombstone is applied (the `TaskCache` bucket
+/// re-insertion of hash-colliding survivors lives there).
```
bgw: Document here why we decided not to include value information in the tombstone (complexity? possible future work)

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -236,20 +240,40 @@ impl TurboBackingStorage {
...
+                            SnapshotItem::Delete(deletion) => {
+                                // `TaskMeta`/`TaskData` are SingleValue: tombstone them inline
+                                // here. The `TaskCache` tombstone
+                                // (a MultiValue read-modify-write) is
+                                // carried out in `deletion` for the sequential phase below.
+                                let key = IntKey::new(*deletion.task_id);
+                                let key = key.as_ref();
+                                batch.delete(KeySpace::TaskMeta, WriteBuffer::Borrowed(key))?;
+                                batch.delete(KeySpace::TaskData, WriteBuffer::Borrowed(key))?;
+                                shard_deletes.push(deletion);
```
bgw:
```suggestion
            let span = tracing::trace_span!("update task data");
            let per_shard =
parallel::map_collect_owned::<_, _, Result<Vec<_>>>(snapshots, |shard: I| {
                    let _span = span.clone().entered();
let mut max_new_task_id = 0;
let mut data_items = 0;
let mut meta_items = 0;
let mut task_cache_items = 0;
                    let mut shard_deletes: Vec<TaskDeletion> = Vec::new();
                    for item in shard {
                        let (task_id, meta, data, task_type_hash) = match item {
                            SnapshotItem::Put {
                                task_id,
                                meta,
                                data,
                                task_type_hash,
                            } => (task_id, meta, data, task_type_hash),
                            SnapshotItem::Delete(deletion) => {
                                let key = IntKey::new(*deletion.task_id);
                                let key = key.as_ref();
                                // apply these deletions immediately
                                batch.delete(KeySpace::TaskMeta, WriteBuffer::Borrowed(key))?;
                                batch.delete(KeySpace::TaskData, WriteBuffer::Borrowed(key))?;
                                // defer these until after the parallel block
                                shard_deletes.push(deletion);
```

bgw (approved):

turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
lukesandberg: :robot: Documented on `TaskDeletion` in 996f572. Summary: the tombstone is identity-only (task id + type hash) with no copy of the value being deleted.

For `TaskMeta`/`TaskData` that's exact — they're `SingleValue`, so a key-granular delete is all you need. The cost is on `TaskCache`, which is `MultiValue`: a delete tombstones the whole hash bucket, so we read the bucket back and re-insert any hash-colliding survivors.

The reason carrying the value wouldn't help is that the write batch has no delete-one-value-from-a-bucket op, so removing a single entry from a `MultiValue` bucket is a read-modify-write regardless of what the caller hands us. Getting rid of the read needs a new persistence primitive (value-granular `MultiValue` delete), not a fatter tombstone — so possible future work, but on the persistence layer rather than here. In practice the read is cheap since collisions are rare and the survivor set is almost always empty.

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
lukesandberg: :robot: Applied in 996f572 — dropped the block comment above `per_shard` and moved the explanation to the two points of action, keeping a short note on each about why one is immediate and the other deferred.

turbopack/crates/turbo-tasks-backend/src/backing_storage.rs
lukesandberg: :robot: Correcting my reply above — Luke pointed out a better design and my rationale was wrong.

I claimed a value-granular `MultiValue` delete would need a new persistence primitive. It doesn't. The deletion in `turbo-persistence` is already lazy: a tombstone is just an entry, and the actual removal happens in the compaction merge (`MultiValue` arm sets `skip_remaining_for_this_key`) and in `get_impl` (stops at the first tombstone). Both already inspect tombstones and already have the entry in hand.

So if the tombstone carried the *value* being deleted, both would narrow from "skip the whole key group" to "skip entries matching this value", and `apply_task_cache_deletions` goes away entirely — GC emits the delete optimistically, compaction drops the entry the next time it rewrites the group, and reads filter it out until then. No synchronous bucket read on the GC commit path at all. Two things already line up for this: tombstones sort last within a key group, so a filtering pass sees every value before the tombstone that kills it, and `TaskCache` values are 4-byte task ids, small enough to ride inline.

The real cost is a persistence format change (`KEY_BLOCK_ENTRY_TYPE_DELETED` carries no value bytes today, so old value-less tombstones would need to coexist behind a version bump) plus a query change (reads could no longer return early at the first tombstone — they'd need to accumulate tombstones across all older layers). That's `turbo-persistence` work and out of scope for this PR, and notably it wouldn't change `TaskDeletion`'s shape: it already carries exactly the identity such a tombstone needs.

Rewrote the doc comment to describe that instead in 6f022da.

PR #95976, opened 2026-07-20
turbo-tasks-backend: parent_count reference counting

lukesandberg:
Adds the reference counting that lets the backend know when a task has become unreachable. 

### What this adds

Two inline counts on `TaskStorage`, maintained incrementally so a task's reachability is known the moment its last parent drops — with no DB scan:

- **`parent_count`** (meta, persisted) — how many persistent parents list this task in their `children` set. Reaching 0 means the task is unreachable from persistent roots.
- **`transient_ref_count`** (transient) — session-only references that aren't persistent parent edges: a transient parent's child edge, or a detached handle held outside the tracked graph. Never persisted; re-established on restart.

Maintenance happens at the two `children`-edge mutation sites:

- `connect_children` applies the `+1` directly under the child guard it already takes. It has to land before the first `operation_suspend_point` so the new child edge and the `+1` are atomically observable.
- `CleanupOldEdges` queues the `-1` as an aggregation-update job, which takes the child guards itself rather than nesting a second guard under the parent's. 

`AdjustParentCount` is durably encoded so an in-flight delta replays after a crash and the persisted count stays consistent. `AdjustTransientRefCount` is `bincode(skip)` and collapses to `Noop`, like the other transient active-count jobs.

### Two things worth a closer look

**`output_dependent`'s inline capacity drops from 6 to 4.** `test_schema_size` asserts `TaskStorage` is exactly 128 bytes, and the two new `u32`s add 8. `AutoSet<TaskId, 6>` is 32 B and `AutoSet<TaskId, 4>` is 24 B, freeing exactly the 8 needed (capacity 5 is still 32 B, so 4 is the first that fits). This is a real behavior change, not just a layout tweak: a task with 5 or 6 output dependents now spills to the heap where it previously stayed inline.

**`task_execution_completed_unfinished_children_dirty` is gone.** Since `connect_children` now takes a guard on every new child anyway, the "mark children with no output dirty" pass folds into that same loop — one guard acquisition instead of two. Behavior-preserving relocation, but it does mean the dirty-marking now reads as part of the connect path.

### Testing

`tests/parent_count.rs` covers the count reaching the right value through connect/disconnect, the parallel wide-fanout `connect_children` path, revalidation not double-counting, survival across a snapshot + eviction, and survival across a full DB reopen (the count is persisted meta, and an in-flight `AdjustParentCount` replays via the durable queue).

Each of the four commits builds and tests green on its own.


lukesandberg:
* **#95976** <a href="https://app.graphite.com/github/pr/vercel/next.js/95976?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95976?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+    },
+    storage::{SpecificTaskDataCategory, TaskDataCategory},
+    storage_schema::TaskStorageAccessors,
+};
+
+/// When `TURBO_ENGINE_GC` is set to a truthy value, the background job runs a `parent_count`-driven
+/// garbage-collection pass (tearing down tasks whose persistent parent count reached 0). Opt-in
+/// until it has been proven on trusted apps; the eventual default-on flip gets its own escape
+/// hatch.
+static GC_ENABLED: LazyLock<bool> = LazyLock::new(|| {
```
lukesandberg: nit: move into `gc_enabled`

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+/// A unit of garbage-collection work fed through the self-feeding parallel pool in
+/// [`TurboTasksBackend::gc_collect`]: collecting one task. Parallelism is *across* tasks — the pool
+/// runs many `Collect` jobs on different workers — while each task's own teardown (its
+/// `CleanupOldEdges` run) is sequential. A `Collect` discovers more `Collect` jobs (each child the
+/// cleanup drives to `parent_count == 0` and that is itself collectible), which flow straight back
+/// into the pool until it drains.
+///
+/// (A single-variant enum today; kept as an enum so the self-feeding pool's job type has room to
+/// grow — e.g. if the per-task teardown is later split to fan a huge task's edges across workers.)
+enum GcJob {
```
lukesandberg: simplify, this doesn't need to be an enum

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+
+/// A unit of garbage-collection work fed through the self-feeding parallel pool in
+/// [`TurboTasksBackend::gc_collect`]: collecting one task. Parallelism is *across* tasks — the pool
+/// runs many `Collect` jobs on different workers — while each task's own teardown (its
+/// `CleanupOldEdges` run) is sequential. A `Collect` discovers more `Collect` jobs (each child the
+/// cleanup drives to `parent_count == 0` and that is itself collectible), which flow straight back
+/// into the pool until it drains.
+///
+/// (A single-variant enum today; kept as an enum so the self-feeding pool's job type has room to
+/// grow — e.g. if the per-task teardown is later split to fan a huge task's edges across workers.)
```
lukesandberg: comment could just be

A Request to delete a given task. Once deleted Cleans up old edges and enqueues new jobs as needed

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+        //
+        // TODO(perf): recycle the task ids of collected tasks. `persisted_task_id_factory`
+        // (`IdFactoryWithReuse`) can hand out freed ids, and the persisted `next_free_task_id`
+        // high-water mark only grows today, so the id space grows unboundedly across churn even
+        // though the task set stays flat. Reuse must happen only AFTER the `save_snapshot` that
+        // tombstoned the id has committed (a crash before commit leaves the task on disk — reusing
+        // its id would alias it), and must be guarded against resurrection: between removal and
+        // commit a `get_or_create_task` for the same type could re-mint the id, and the id must not
+        // be handed out while any live `OperationVc`/`DetachedVc` still references it. Feed the
+        // recycled ids into `persisted_task_id_factory` so the high-water mark can stop growing.
```
lukesandberg: we should accumulate deleted tasks in a RoaringBitmap that we

1. persist
2. use to make accessing deleted tasks a hard error

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+            .gc_collectible_candidates()
+            .into_iter()
+            .map(GcJob::Collect)
+            .collect();
+        if seeds.is_empty() {
+            return 0;
+        }
+
+        // Written once per collected task (not per child/dep), so the atomic is not a hot path.
+        let collected = AtomicUsize::new(0);
```
lukesandberg: we should consider other stats

edges deleted

roots detected

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+            self.gc_run_job(job, spawner, turbo_tasks, &collected);
+        });
+
+        collected.into_inner()
+    }
+
+    /// Runs one [`GcJob`], possibly spawning follow-up jobs into the same pool. See
+    /// [`Self::gc_collect`] for the concurrency argument. Each job builds its own GC
+    /// [`ExecuteContext`].
+    fn gc_run_job(
```
lukesandberg: just inline this into the `scope_self_feeding closure`?

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+
+                // Capture ALL of this task's edges as `OutdatedEdge`s, then hand them to the same
+                // `CleanupOldEdges` operation a re-executing task uses. This is what makes GC
+                // teardown correct: alongside dropping each child's `parent_count` and scrubbing
+                // the forward-dep reverse edges, it PROPAGATES THE AGGREGATION
+                // REBALANCE (removes this task from its children's `upper` sets via
+                // `InnerOfUppersLostFollowers`). Without that, collected children
+                // would be left with a dangling upper edge and never
+                // become collectible themselves. The op opens `ctx.task(task_id)` to remove the
+                // child edges, so it must run while `task_id` is still resident.
```
lukesandberg: simplify the comment

just collect all edges and delete them before tagging the task itself as deleted

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+                    old_edges,
+                    AggregationUpdateQueue::new(),
+                    &mut ctx,
+                );
+
+                // The edges are gone and the children are rebalanced. Instead of removing the task
+                // from the map now (which would let a later `ctx.task` on it — e.g. a sibling's
+                // `CleanupOldEdges` forward-dep scrub — resurrect it from disk as a zombie), mark
+                // it soft-deleted and keep it resident. The next snapshot tombstones its on-disk
+                // copy; a later step hard-deletes it from memory once the tombstone has committed.
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+                // pinned, a root, or still hold aggregation edges) and spawn a `Collect` for the
+                // ones that are collectible. Each child reaches 0 exactly once, so there is no
+                // double-queueing.
+                let newly_parentless = ctx.take_gc_parent_count_zeroed();
+                for child in newly_parentless {
+                    debug_assert!(
+                        !child.is_transient(),
+                        "gc: a transient task should never have a persistent parent_count to zero"
+                    );
+                    if ctx.task(child, TaskDataCategory::Meta).is_gc_collectible() {
```
lukesandberg: CleanupOldEdges must have fetched `All` and the next job will do the same... not sure if it matters but should we do the same here?

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+                        .map(OutdatedEdge::CollectiblesDependency),
+                );
+                drop(task);
+
+                CleanupOldEdgesOperation::run(
+                    task_id,
+                    old_edges,
+                    AggregationUpdateQueue::new(),
+                    &mut ctx,
+                );
```
lukesandberg: does the relative order of `task_deleted` and CleanupOldEdges actually matter?

if we mark deleted first, we could save a lock acquisition

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+        //      and
+        //   2. a task persisted parentless in a prior session, restored this cycle (restore sets no
+        //      modified flags) and collected with no intervening mutation.
+        // In those cases this call is what carries the tombstone into the snapshot (it flips
+        // `has_modifications`, so the modified scan runs). When meta is already dirty (the common
+        // cascade case) it's a cheap no-op. Runs before `start_snapshot`, so it takes the
+        // not-in-snapshot-mode path and bumps the shard modified count. The returned `TrackOutcome`
+        // is only for undo; GC's mark is permanent, so discard it.
+        let _ = task.track_modification(SpecificTaskDataCategory::Meta, "gc_deleted");
+    }
```
lukesandberg: all the comments here can be massively simplified

soft deletion is necessary because of racing edge deletions and track_modification is required to ensure that snapshotting sees it

actually i wonder if `deleted` should be a `meta` category if only to trigger implicit `track_modification` behavior

it would never actually be persisted which is slightly amusing but it would simplify reasoning

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -0,0 +1,357 @@
...
+        // `AdjustTransientRefCount` queue job uses.
+        let existed = ctx.resident_task(task);
+        debug_assert!(
+            existed.is_some(),
+            "pin_task_for_gc: task {task} has no resident entry (pinned an already-collected \
+             task?)"
+        );
+        if let Some(mut guard) = existed {
+            guard.update_and_get_transient_ref_count(1);
+        }
```
lukesandberg: comments are too volumninous

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -247,8 +253,29 @@ impl TurboTasksBackend {
...
             .next_free_task_id()
             .expect("Failed to get task id");
+
+        // GC leaves collected tasks resident (soft-deleted) until a reclaim step removes them. The
+        // background `ReadWrite` loop reclaims them in `evict_after_snapshot`, so GC there REQUIRES
+        // eviction to be on; with eviction off, soft-deleted tasks would accumulate forever. The
+        // `ReadWriteOnShutdown` drain path drops the whole map wholesale (no per-cycle eviction
+        // needed), and `ReadOnly` never persists/GCs. In debug builds, refuse the unsafe combo by
+        // forcing GC off with a warning; release builds trust the caller's configuration.
+        let mut gc_enabled = gc::gc_enabled();
```
lukesandberg: should the `gc_enabled` function/lazylock just be moved here?

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -306,6 +333,10 @@ impl TurboTasksBackend {
     /// This is exposed for integration tests that need to verify the
     /// snapshot → evict → restore cycle works correctly.
     ///
+    /// A snapshot run after a [`Self::gc_for_testing`] pass commits that pass's tombstones
+    /// automatically: GC leaves collected tasks resident with their `deleted` flag set, and the
+    /// snapshot derives the on-disk tombstones from that flag (no tombstones need threading in).
+    ///
```
lukesandberg: pointless comment, drop it

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -951,18 +1020,51 @@ impl TurboTasksBackend {
...
+        // `GcPhase::into_snapshot` hands the exclusion straight to the snapshot phase (swapping the
+        // GC request bit for the snapshot bit under one lock). Because no operation can run between
+        // the cascade and the snapshot, a task the cascade decremented cannot be resurrected before
+        // it is persisted — so the decremented children counts and the snapshot are consistent, and
+        // GC-collected tasks' tombstones ride this same commit.
+        //
+        // Without GC we just begin the snapshot directly. `deletes` are the GC tombstones for this
+        // commit: in the production (GC-enabled) path they are produced right here under the *same
+        // continuous exclusion* as the snapshot (the atomic `into_snapshot` hand-off), so nothing
+        // could have resurrected a collected task.
```
lukesandberg: simplify the comment

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -972,8 +1074,8 @@ impl TurboTasksBackend {
         drop(snapshot_phase);
 
         if !has_modifications {
-            // No tasks modified since the last snapshot — drop the guard (which
-            // calls end_snapshot) and skip the expensive O(N) scan.
+            // No tasks modified since the last snapshot (GC-collected tasks count as modified) —
+            // drop the guard (which calls end_snapshot) and skip the expensive O(N) scan.
```
lukesandberg: revert diff

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1182,6 +1284,26 @@ impl TurboTasksBackend {
...
             }
 
+            // A GC-soft-deleted task is not persisted; instead its on-disk copy (task meta/data +
+            // its `TaskCache` entry) is tombstoned in this same commit. We represent that as a
+            // `SnapshotItem::Delete` so it rides the streaming shard iterator that `save_snapshot`
+            // consumes (rather than a side-channel that would have to be fully populated before the
+            // put loop runs). GC marks such tasks modified so the per-shard scan visits them here,
+            // and the iterator clears their modified flags, so a still-`deleted` task not
+            // hard-deleted this cycle won't be re-tombstoned next snapshot. Only persistent tasks
+            // are collected, so a persistent task type (the `TaskCache` key) is always present.
```
lukesandberg: for tasks that are `new` and delete we should skip persisting them altogether, there is nothing to tombstone

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2060,9 +2180,8 @@ impl TurboTasksBackend {
         let has_new_children = !new_children.is_empty();
         span.record("new_children", new_children.len());
 
-        if has_new_children {
-            self.task_execution_completed_unfinished_children_dirty(&mut ctx, &new_children)
-        }
+        // Note: dirtying any unfinished new children is folded into `connect_children`'s
+        // parent_count pass (a single guard per child), so there is no separate dirty pass here.
```
lukesandberg: drop comment

turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
```diff
@@ -294,6 +303,14 @@ pub enum AggregationUpdateJob {
...
         #[bincode(skip, default = "unreachable_decode")]
         task: TaskId,
+        /// Whether this increase is for a task being connected as a genuinely-new child (the
+        /// direct-connect chokepoint in `ConnectChildOperation::run`). When set, the handler also
+        /// resurrects the task if GC had soft-deleted it — riding the child guard it takes anyway,
+        /// so the common (not-deleted) case is a free Meta flag read. `false` for the aggregation
+        /// propagation producers of this job, which never target a `deleted` task (a collected
+        /// task has no `upper`/`followers` to be propagated to).
+        #[bincode(skip, default = "Default::default")]
+        resurrect: bool,
```
lukesandberg: i wonder if we need this bit? couldn't we just always ressurrect on activation, it would be redundant sometimes, but checkined a deleted flag on a guard we already have is fine

turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
```diff
@@ -3153,6 +3221,20 @@ impl AggregationUpdateQueue {
             // persistent_task_type is now set eagerly in initialize_new_task.
             AGGREGATION_UPDATE_CATEGORY,
         );
+        // Revive a GC-soft-deleted child before touching activeness, so the rest of this function
+        // (and the scheduling it drives) sees a live, re-dirtied task — matching the original
+        // resurrect-first ordering. Peek on the guard we already hold; only on the connect path and
+        // only when actually deleted does `resurrect_deleted` consume it, re-acquire `All` to
+        // double-check and atomically clear+re-dirty, and hand back a fresh guard.
+        if resurrect && task.deleted() {
+            task = crate::backend::operation::connect_child::resurrect_deleted(
```
lukesandberg: import the function

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
```diff
@@ -96,10 +96,42 @@ impl CleanupOldEdgesOperation {
...
                                 });
                                 let mut task = ctx.task(task_id, TaskDataCategory::All);
-                                for task_id in children.iter() {
-                                    task.remove_children(task_id);
+                                // Track which persistent children were *actually* removed (the edge
+                                // was present) so we adjust the child-side count exactly once per
+                                // real edge removal. Gating on the `remove_children` return makes
+                                // this replay-safe: a resumed CleanupOldEdges re-removes the same
+                                // edges, and edges already gone (returning false) are skipped.
+                                let mut removed_persistent_children: SmallVec<[TaskId; 4]> =
```
lukesandberg: we probably don't need the explicit type declaration

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
```diff
@@ -96,10 +96,42 @@ impl CleanupOldEdgesOperation {
...
+                                    }
+                                }
+                                // Each removed persistent child loses a parent. A persistent parent
+                                // drops the child's durable `parent_count` (queued so it is
+                                // crash-consistent); a transient parent drops the session-only
+                                // `transient_ref_count` (skipped during serialization, never
+                                // replayed). Both ride the queue via the same job family, so the
+                                // handler — not this site — adjusts the right counter and takes the
+                                // child guards (avoiding a second guard while `task` is still
+                                // held).
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
```diff
@@ -172,6 +204,19 @@ impl CleanupOldEdgesOperation {
                                     task: cell_task_id,
                                     cell,
                                 } = forward;
+                                // Under the GC phase, a collected task stays resident
+                                // (soft-deleted) so this scrub
+                                // finds a live entry; a non-resident target would mean
+                                // `ctx.task` is about to resurrect an already-collected task from
+                                // disk (the bug soft-deletion fixes). `gc_target_resident` is
+                                // `None` outside GC, where
+                                // restoring a missing target is legitimate.
```
lukesandberg: simplify comment, we should also generalize this check since it generic for gc corrupts state

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -7,12 +7,66 @@ use crate::{
...
+/// in which case there is nothing to do. When still deleted, the clear and the re-dirty happen
+/// **under that single `All` guard without an intervening drop**, so no operation can observe the
+/// intermediate `!deleted && !dirty` state (a task that looks live but still holds the stale/empty
+/// edges GC scrubbed). A **mutable** task is cleared and made dirty so it re-executes and rebuilds
+/// those edges; an **immutable** task cannot be made dirty (invariant in `make_task_dirty`) and
+/// does not need to be — its output is deterministic and its edges self-contained — so clearing the
+/// marker alone suffices, leaving it immediately valid for any concurrent observer.
+///
+/// GC collection is the single producer of the `deleted` flag and runs under an exclusion that
+/// stops all operations, so once cleared here the flag cannot be re-set concurrently.
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -29,6 +83,16 @@ impl ConnectChildOperation {
...
     ) {
+        // Resurrection of a GC-soft-deleted child (clearing the `deleted` marker + re-dirtying it)
+        // is folded into the child guard the connect handshake takes below — the activeness path
+        // does it in `increase_active_count`, the non-activeness path inline — so the common
+        // (not-deleted) case costs only a Meta flag read on a guard we hold anyway, with no extra
+        // acquisition. A `deleted` child is always in NEITHER the parent's `children` nor its
+        // `new_children` when first connected (GC only collects `parent_count == 0` tasks, so it is
+        // in no live parent's children; and a prior connect in this execution would have already
+        // staged+resurrected it), so the dedup early-returns below never skip a needed
+        // resurrection.
```
lukesandberg: drop comment

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -71,12 +135,30 @@ impl ConnectChildOperation {
...
             });
         } else {
             let mut child_task = ctx.task(child_task_id, TaskDataCategory::Meta);
 
+            // Peek for GC soft-deletion on the guard we already hold — common path: one Meta flag
+            // read, no extra lock. Rare (was deleted): `resurrect_deleted` consumes this guard,
+            // re-acquires `All` to double-check and atomically clear+re-dirty, and hands back a
+            // fresh Meta guard for the schedule decision below. Resurrection thus completes before
+            // scheduling, matching the original resurrect-first order.
+            if child_task.deleted() {
```
lukesandberg: now that ressurrect_deleted takes a guard

it should really include the deleted check insternally and can just be `child_task = resurrect_delete` and it can sometimes be a no-op

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_children.rs
```diff
@@ -52,6 +58,71 @@ pub fn connect_children(
...
+        // *under*-collect for one pass,    self-correcting — so it stays on the durable
+        // queue there.)
+        //
+        // 2. Make any child that has not produced output yet dirty, so it gets scheduled and
+        //    computes. This was formerly a separate `for_each_task_all` pass over the same children
+        //    (`task_execution_completed_unfinished_children_dirty`) before the connect; folding it
+        //    here reuses this guard and, on the parallelized path, chunks it with the rest. It only
+        //    runs on the successful connect (not the stale/cancel early-returns in
+        //    `task_execution_completed_connect`), which is consistent with those paths undoing the
+        //    children's temporarily-increased active count rather than keeping them.
```
lukesandberg: simplify comment

we don't need to discuss the various alternatives that were explored

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -7,12 +7,66 @@ use crate::{
...
+            // Clear + re-dirty atomically under this single guard so no observer sees `!deleted`
+            // before the task has been re-validated.
+            task.set_deleted(false);
+            if !task.immutable() {
+                make_task_dirty_internal(
+                    task,
+                    task_id,
+                    true,
+                    #[cfg(feature = "task_dirty_cause")]
+                    crate::backend::operation::invalidate::TaskDirtyCause::Unknown,
```
lukesandberg: Create a new TaskDirtyCause `Resurrected`

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -168,6 +200,11 @@ pub struct ExecuteContextImpl<'e> {
     turbo_tasks: &'e TurboTasks<TurboTasksBackend>,
     _operation_guard: Option<OperationGuard<'e, AnyOperation>>,
     task_lock_counter: TaskLockCounter,
+    /// GC-only: ids whose persistent `parent_count` reached 0 during this context's operations
+    /// (recorded by `note_gc_parent_count_zeroed`). `None` for normal operation contexts (the
+    /// recording hook is a no-op there); `Some` only for the GC context, drained by the collector
+    /// via [`Self::take_gc_zeroed`] to discover cascade-collectible tasks.
+    gc_zeroed: Option<Vec<TaskId>>,
```
lukesandberg: we should consider having a dedicated trait implementation for the GC context instead of these optional fields

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -300,7 +413,84 @@ impl<O> Drop for SnapshotPhase<'_, O> {
...
+    /// Atomically transitions from the GC phase directly into a snapshot phase **without ever
+    /// releasing operation exclusion**. Under the state lock it clears the GC request bit and sets
+    /// the snapshot request bit in one critical section, so no operation can start in between (an
+    /// operation blocks while *either* bit is set). This closes the race where a mutation could
+    /// resurrect a just-collected task in the gap between the GC pass and the snapshot: with an
+    /// atomic hand-off there is no such gap, so the GC cascade's `parent_count` decrements and the
+    /// snapshot see a consistent graph.
+    ///
+    /// Because operations were already drained to zero by `begin_gc`, no further draining is
+    /// needed.
```
lukesandberg: simplfiy comment

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -846,6 +909,12 @@ impl TurboTasksBackend {
         } else {
             (ctx.task(task_id, TaskDataCategory::All), None)
         };
+        // See the matching assert in `try_read_task_output`: a GC-deleted task must be resurrected
+        // (at connect) before any read; reaching a read with the flag still set is a missed path.
+        debug_assert!(
+            !task.deleted(),
+            "read_task_cell on a GC-deleted task {task_id} — a resurrection path was missed"
```
lukesandberg: we need to find better locations for this

if a task 'must exist' then it cannot be deleted

but on the maybe-create path we could resurrect

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -966,18 +1035,43 @@ impl TurboTasksBackend {
...
         let _snapshot_in_progress = self.snapshot_in_progress.lock();
+
+        // When GC is enabled, run the pass and transition to the snapshot **atomically**, without
+        // releasing operation exclusion in between: `begin_gc` drains all operations, `gc_collect`
+        // marks collectible tasks soft-deleted (cascading `parent_count` decrements to children),
+        // and `into_snapshot` swaps the GC request bit for the snapshot bit under one lock. Since
+        // no operation can run in between, a decremented task can't be resurrected before
+        // it is persisted, so the children counts and the snapshot stay consistent and the
+        // tombstones (derived from the `deleted` flag) ride this same commit. Without GC,
+        // begin the snapshot directly. Opt-in via `TURBO_ENGINE_GC`.
```
lukesandberg: simplify comment

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1197,6 +1291,31 @@ impl TurboTasksBackend {
...
+            // its `TaskCache` entry) is tombstoned in this same commit, riding the same streaming
+            // shard iterator `save_snapshot` consumes. GC marks such tasks modified so the scan
+            // visits them here, and the iterator clears their flags so a still-`deleted` task not
+            // hard-deleted this cycle isn't re-tombstoned next snapshot.
+            if inner.flags.deleted() {
+                // A task collected before it was ever persisted (`new_task`) has no on-disk copy —
+                // emit nothing (just let the scan clear its flags) rather than tombstone a
+                // key that was never written.
+                if inner.flags.new_task() {
+                    return SnapshotItem::Skip;
```
lukesandberg: this combination is interesting and should be handled in the gc instead of in the snapshot

we can drop 'skip' and instead change the 'mark_deleted' path in the gc to detect this

if we delete a new task we can clear the modified bits and decrement the shard modified count for the task, then snapshotting will naturally skip it but eviction will still delete

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1268,6 +1387,8 @@ impl TurboTasksBackend {
         // Tasks were already consumed by take_snapshot, so a future snapshot
         // would not re-persist them — returning an error signals to the caller
         // that further persist attempts would corrupt the task graph in storage.
+        // GC tombstones are carried inline as `SnapshotItem::Delete` entries in `task_snapshots`
+        // (derived from the `deleted` flag during the scan) and applied in the same atomic commit.
```
lukesandberg: drop comment

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Done — the `TURBO_ENGINE_GC` env read is inlined into the `TurboTasksBackend` constructor; the static + `gc_enabled` fn were removed from gc.rs.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Done — `GcJob` is now a newtype (`struct GcJob(TaskId)`), not an enum.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Done — the `GcJob` doc is simplified to a one-liner.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Added `edges_deleted` to `GcStats`. "roots detected" isn't a meaningful per-pass counter: GC roots are filtered out by `gc_maybe_collectible` before seeding, so they never enter the pool — the count would always be 0.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Done — `gc_run_job` is inlined into the `scope_unbounded` closure.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Kept `Meta` for the cascade re-check (`is_gc_collectible` reads only Meta fields); a child that turns out collectible re-opens with `All` in its own collect job, so fetching `All` here would only waste a Data restore on non-collectible children. Justified in a comment.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Order doesn't matter, so mark-deleted now happens first, on the guard already held for edge capture — saving the separate lock acquisition. Documented.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Done — comments condensed.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: The persisted RoaringBitmap deleted-set (hard-error on accessing a deleted task) is addressed up-stack: the TaskAccess work makes `task(.., MustExist)` refuse to fabricate a task that exists in neither memory nor disk (the non-fabrication guarantee), which is the "accessing a deleted task is a hard error" property. A separate persisted bitmap is unnecessary given that; leaving as-is.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: :robot: Simplified the comments. Evaluated `deleted` as a `meta` category: it would auto-track modification (dropping the explicit `track_modification`), but meta flags are in the persisted bitmask — making a soft-delete marker nominally persistable. Kept it transient + explicit `track_modification` to preserve the never-persisted contract.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — the env read is inlined into the constructor here; the gc.rs static + fn are gone.

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
```diff
@@ -125,95 +153,180 @@ impl<O> SnapshotCoordinator<O> {
...
-            state.suspended_operations.remove(&PtrEqArc::from(op));
+            if let Some(op) = recorded_op {
+                state.suspended_operations.remove(&PtrEqArc::from(op));
+            }
         }
         suspend_point_cold(self, suspend);
     }
 
-    /// Begin a snapshot. Sets the snapshot bit, blocks until all in-flight
-    /// operations have drained or suspended, and returns a [`SnapshotPhase`]
```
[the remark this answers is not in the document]
lukesandberg: :robot: Fixed. `suspend_point` now records the suspended operation unconditionally (even during a GC-only suspension), so a GC phase that hands off to a snapshot via `into_snapshot` carries the operation into the uncompleted-operations log. A standalone GC pass (ended via `GcPhase::Drop`, no snapshot) simply never reads the recorded set. Updated the coordinator test to the new contract.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Addressed by the TaskAccess refactor up-stack: the `!deleted()` assertion now lives only on the consumer read paths (`try_read_task_output`/`try_read_task_cell`), not on every `task()`/`MustExist` open — GC/aggregation bookkeeping legitimately opens soft-deleted tasks mid-resurrection. `MustExist` guarantees existence (no fabrication); the stale-read invariant is asserted where a stale read would escape.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — dropped `SnapshotItem::Skip`. GC's mark-deleted path now detects a never-persisted (`new_task`) collected task and calls `discard_modifications_for_gc_new_task` (clears modified bits + decrements the shard modified count), so the snapshot naturally skips it while eviction still deletes it.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — dropped.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — reverted.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — implemented (see the thread on the `process` deleted branch): GC now clears the modified bits for a collected `new_task`, so it's skipped by the snapshot entirely.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — dropped.

turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
lukesandberg: :robot: Done — dropped the `resurrect` flag; `increase_active_count` always calls `resurrect_deleted` (a no-op on the common not-deleted path).

turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
lukesandberg: :robot: Done — `resurrect_deleted` is imported.

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
lukesandberg: :robot: Done — dropped the explicit type declaration.

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
lukesandberg: :robot: Done — dropped.

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
lukesandberg: :robot: Done — `resurrect_deleted` now internalizes the `deleted()` check, so the call site is `child_task = resurrect_deleted(...)` and is a no-op on the common not-deleted path.

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_children.rs
lukesandberg: :robot: Done — simplified; dropped the alternatives discussion.

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
lukesandberg: :robot: Done — added `TaskDirtyCause::Resurrected`, used on the resurrection re-dirty.

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
lukesandberg: :robot: Addressed up-stack by the TaskAccess work — see the discussion; the dedicated GC-context concern is subsumed by the `MustExist`/existence machinery. Leaving the optional-field shape as-is for now.

turbopack/crates/turbo-tasks-backend/src/backend/snapshot_coordinator.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — simplified.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Done — dropped.

lukesandberg:
TODO needs some perfomance testing

turbopack/crates/turbo-tasks-backend/benches/gc.rs
```diff
@@ -0,0 +1,179 @@
...
+    std::fs::create_dir_all(&parent).unwrap();
+    let dir = tempfile::Builder::new()
+        .prefix("gc-bench-")
+        .tempdir_in(&parent)
+        .unwrap();
+    let tt = TurboTasks::new(TurboTasksBackend::new(
+        BackendOptions {
+            num_workers: None,
+            small_preallocation: false,
+            storage_mode: Some(turbo_tasks_backend::StorageMode::ReadWriteOnShutdown),
```
bgw: Import

turbopack/crates/turbo-tasks-backend/benches/gc.rs
```diff
@@ -0,0 +1,179 @@
...
+        },
+        turbo_tasks_backend::turbo_backing_storage(
+            dir.path(),
+            &GitVersionInfo {
+                describe: "bench-unversioned",
+                dirty: false,
+            },
+            false,
+            true,
+            true,
```
bgw: We gotta do something about this (in a later PR)

lukesandberg:
Heads up for anyone with review threads here: this PR now covers only the `parent_count` / `transient_ref_count` reference counting. The garbage collector that consumes those counts moved to #97282, directly above this one in the stack.

Comments on GC internals — `gc.rs`, the snapshot-coordinator GC phase, soft-deletion and resurrection, tombstoning, the pin API — now apply to #97282. Comments on the counts themselves, their maintenance sites, and the schema change still apply here.

The two branches together are byte-identical to what this PR contained before, apart from GC tests moving into their own file.

turbopack/crates/turbo-tasks-backend/benches/gc.rs
lukesandberg: https://github.com/vercel/next.js/pull/97284

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -563,6 +592,13 @@ impl TaskStorage {
             Some(arc) if arc.count() == 1 => KeyEvictability::AlreadyEvicted,
             Some(_) => KeyEvictability::Evictable,
         };
+        // A task with a live transient reference (`transient_ref_count`: a `prevent_gc` pin, a
+        // detached handle, or a transient parent) is NOT forced fully resident — it falls through
+        // to the normal Meta/Data evictability below. That is safe because `drop_partial`
+        // retains non-default *transient* fields, so `transient_ref_count` survives as
+        // residue and the map entry is kept (`DropPartialOutcome::HasResidue`): the count
+        // can never be lost to eviction and make a still-referenced task look unreferenced.
+
```
bgw:
```suggestion
```

bgw (approved):

PR #95977, opened 2026-07-20
turbo-tasks-backend: anchor GC roots, gate on Data residency, wire up the app

lukesandberg:
## What

Makes GC correct and safe on a real application: anchors GC roots, refuses to collect tasks whose data isn't resident, and wires the anchoring into the app.

- **`gc_root`**: a task created with **no parent** (`parent == None`) — spawned from outside the tracked graph (a `turbo_tasks.run`/NAPI entry, `run_once`, or `OperationVc::connect` outside a task) — is marked a GC root and is never collected. Without this, root operations reach `parent_count == 0` once their transient creator finishes and GC collects them, cascading into their whole subgraph. 
- **Data-resident gate**: `gc_maybe_collectible` also requires Data restored, so GC never selects a task whose type-bearing Data was evicted (which crashed at snapshot). Interim stop-gap with a `TODO` — the real fix is restore-Data-in-GC or moving the task-type-hash into Meta.
- **`DetachedVc`** pins its `OperationVc` across the NAPI boundary; pinned tasks are partially evicted rather than forced fully resident.
- **`VersionedContentMap`**: `serialization = "skip"` + `evict = "never"` + pin/unpin of the `OperationVc`s it stores, so the ops held in its `State` are anchored (they live outside the task graph and would otherwise be collected).
- **`keys_in_path` / `raw_get` are `session_dependent`**: these are the two tasks that read the map's non-persisted `State`. Because the map is `serialization = "skip"`, its `State` (and the read-subscription invalidators recorded inside it) is not persisted — but the reader tasks *are* persisted, so on a cross-session restore their cached output would be served against a fresh, empty map with no edge to invalidate it. `session_dependent` forces both to re-execute on restore instead of reusing cache, so they recompute against the empty map (`keys_in_path → []`, `raw_get → None`). Without this, a route deleted between sessions was still keyed by the stale restored `keys_in_path` output, causing the next session to re-subscribe and re-drive the deleted route's endpoint (`AppEndpoint::output` against a tree with no such route → `AppPageLoaderTree` cell miss → a burst of non-fatal "Failed to write app endpoint" errors at startup). This is a no-op without dependency tracking (non-persistent runs), so it's inert exactly where it's irrelevant.

## Why

The collection pass (PR below) is correct for graph-reachable tasks but over-collects anything held *outside* the tracked graph — the project container, per-request operations, and ops stashed in long-lived `State`. This PR anchors those.

## Known tradeoff (interim)

`gc_root` is a permanent pin, so request-scoped roots (e.g. a fresh source-map/trace operation per navigation) accumulate uncollected within a session. This is the conservative correctness-over-reclamation choice; a real root lifecycle is a follow-up.

## Verification

Backend GC suites (incl. `gc_does_not_collect_parentless_root_task` and the flag-layout tests), and a dev-app run with `TURBO_ENGINE_GC` enabled through navigation/HMR cycles — no crashes, dead leaves collected.

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
@@ -250,6 +250,10 @@ struct TaskStorageSchema {
     #[field(storage = "flag", category = "transient")]
     deleted: bool,
 
+    /// Marks the task as a GC root meaning it can never be collected via reference counting.
+    #[field(storage = "flag", category = "meta")]
+    gc_root: bool,
+
     // =========================================================================
     // CHILDREN & AGGREGATION (meta)
     // =========================================================================
@@ -898,15 +902,28 @@ impl TaskStorage {
     /// The aggregation-edges check is conservative: a disconnected task is typically removed from
     /// the aggregation graph, but that can lag and race GC, so we back off rather than collect.
     pub fn gc_maybe_collectible(&self) -> bool {
+        // Note: this intentionally does NOT gate on `Data` being resident. Data holds
+        // `persistent_task_type` (needed to tombstone the `TaskCache` entry), but the collect path
+        // opens the task with `All`, restoring Data before it reads the type. Gating here would
+        // instead strand a truly-unreferenced Data-evicted task forever (it would never be
+        // selected, so its Data would never be restored) — a disk-space leak. Accepting the
+        // extra Data restore for a task about to be deleted is the better trade.
         self.flags.is_restored(TaskDataCategory::Meta)
             // Already collected this session (soft-deleted, awaiting tombstone + hard-delete):
             // don't re-select it, or a second pass would collect it again while it is still
             // resident.
             && !self.flags.deleted()
+            // A GC root (a task spawned with no parent, see the `gc_root` flag) is held from
+            // outside the tracked graph and has no persistent parent — never collect it.
+            && !self.flags.gc_root()
+            // Refcounts must be 0
             && self.gc_parent_count() == 0
             && self.gc_transient_ref_count() == 0
+            // Must not be in progress or being connected
             && self.get_activeness().is_none()
             && self.get_in_progress().is_none()
+            // Must not be participating in the aggregation graph.  Getting disconnected from parents
+            // should clear these but that process isn't atomic.
             && self.upper().is_empty()
             && self.followers().is_none_or(|f| f.is_empty())
     }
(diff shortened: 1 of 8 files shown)


lukesandberg:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/vercel/next.js/95977?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#95977** <a href="https://app.graphite.com/github/pr/vercel/next.js/95977?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/95977?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -920,14 +924,29 @@ impl TaskStorage {
     /// the aggregation graph, but that can lag and race GC, so we back off rather than collect.
     pub fn gc_maybe_collectible(&self) -> bool {
         self.flags.is_restored(TaskDataCategory::Meta)
+            // TODO: gating on Data being resident is a stop-gap for the crash where GC selects a
+            // task whose Data (holding `persistent_task_type`) was evicted, then `snapshot_and_persist`
+            // reads the absent type and panics. It creates a standoff: a truly-unreferenced task
+            // whose Data is never restored can never be collected (can't-delete-because-not-restored
+            // / won't-restore-because-unreferenced) — a disk-space leak, not memory. Real fix:
+            // restore Data during the GC pass before deciding, or move the task-type-hash into Meta
+            // so collection needs only Meta.
```
lukesandberg: i think it is reasonable for the gc to restore data for tasks that are about to be deleted

so this should really skip this check, and the caller will need to update their guard to All to actually delete.

this is kind of lame but the over-read is better than a leak

lukesandberg: :robot: Done — removed the `is_restored(Data)` gate from `gc_maybe_collectible`. The collect path opens `All` (restoring Data before it reads `persistent_task_type` for the tombstone), so the gate was over-conservative and created the disk-leak standoff; the extra Data restore for a to-be-deleted task is the better trade.

PR #95978, opened 2026-07-20
[turbopack] Rename `turbopackTreeShaking` to `turbopackModuleFragments`

sampoder:
This has been causing a lot of confusion when contributors have been debugging tree shaking, eg. https://github.com/vercel/next.js/issues/95698 and the linked issues. Turbopack's standard tree-shaking features are controlled through the `turbopackRemoveUnusedExports` and `turbopackRemoveUnusedImports` options which default to true.


packages/next/src/server/config-shared.ts
```diff
@@ -879,6 +879,11 @@ export interface ExperimentalConfig {
   /**
    * Enable tree shaking for the turbopack dev server and build.
```
sampoder: todo: update this

crates/next-core/src/next_config.rs
```diff
@@ -2421,7 +2421,7 @@ impl NextConfig {
 
     #[turbo_tasks::function]
     pub fn tree_shaking_mode_for_user_code(&self, _is_development: bool) -> Vc<OptionTreeShaking> {
-        OptionTreeShaking(match self.experimental.turbopack_tree_shaking {
+        OptionTreeShaking(match self.experimental.turbopack_module_fragments {
```
lukesandberg: rename these functions and the value type/enum as well?

docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -376,7 +376,7 @@ Additionally, the following experimental options are available under `experiment
...
 | `turbopackInputSourceMaps`                                                                                   | Enable extraction of source maps from input files.                                                                                           | `true`        | `true`                        |
-| `turbopackTreeShaking`                                                                                       | Use advanced module-fragments tree shaking instead of the default reexports-only mode.                                                       | `false`       | `false`                       |
+| `turbopackModuleFragments`                                                                                   | Use advanced module-fragments tree shaking instead of the default reexports-only mode.                                                       | `false`       | `false`                       |
```
lukesandberg: clarify in the comment that this is under active development

packages/next/src/server/config-shared.ts
lukesandberg: fix the doc

clarify that this is very much under development

docs/01-app/03-api-reference/08-turbopack.mdx
sampoder: Probably can drop the mention of reexports

packages/next-codemod/transforms/next-experimental-turbo-to-turbopack.ts
lukesandberg, line 38: you can do something similar here where we either remove or rename it in the codemod

packages/next/src/server/config-shared.ts
```diff
@@ -879,6 +879,11 @@ export interface ExperimentalConfig {
...
    */
+  turbopackModuleFragments?: boolean
+
+  /**
+   * @deprecated Use `turbopackModuleFragments` instead.
+   */
   turbopackTreeShaking?: boolean
```
lukesandberg: we can just delete it imho.

it isn't breaking to set a non-existent experimental property

packages/next/src/server/config.ts
```diff
@@ -1051,6 +1072,13 @@ function assignDefaultsAndValidate(
   ) {
     result.skipMiddlewareUrlNormalize = userConfig.skipProxyUrlNormalize
   }
+  if (
+    userConfig.experimental?.turbopackModuleFragments !== undefined &&
+    userConfig.experimental?.turbopackTreeShaking === undefined
+  ) {
+    result.experimental.turbopackTreeShaking =
+      userConfig.experimental.turbopackModuleFragments
+  }
```
lukesandberg: why copy back?

lukesandberg (approved):

packages/next/src/server/config.ts
sampoder: Removed this

Issue #92301, opened 2026-04-03
`next build --webpack` (v16): `/_not-found` executed at build time + Google Font `Retrying 1/3...` visible — both absent in `next build` (Turbopack) — not seen in v14.2 webpack


chandra2607:
### Link to the code that reproduces this issue

https://github.com/chandra2607/nextjs-webpack-turbopack-issue.git

### To Reproduce

```bash
git clone https://github.com/chandra2607/nextjs-webpack-turbopack-issue.git
cd nextjs-webpack-turbopack-issue

# Turbopack (next build)
docker build --build-arg BUILD_CMD=build -t repro-turbo . 2>&1 | tee out-turbo.txt

# Webpack (next build --webpack)
docker build --build-arg BUILD_CMD=build:webpack -t repro-webpack . 2>&1 | tee out-webpack.txt

# Compare
grep "not-found\|Retrying\|Generating static\|ƒ" out-webpack.txt
grep "not-found\|Retrying\|Generating static\|ƒ" out-turbo.txt
```

### Current vs. Expected behavior

**Current behavior:**

- `next build --webpack`: webpack's static generation worker **executes not-found.tsx** at build time (side effects fire, fetches run) and logs `Retrying 1/3...` for transient Google Font download failures
- `next build` (Turbopack): static generation worker **never enters not-found.tsx** and emits no retry logs — the component is silently skipped
- Both produce the same final output: `ƒ /_not-found (Dynamic)`

**Expected behavior:**

Both bundlers should behave identically for `/_not-found` during the build. If the route is dynamic, neither should probe-render it. If one does, both should — so build-time side effects are predictable regardless of which bundler is used. 

### Provide environment information

```bash
Next.js:  16.2.1
Node:     22-alpine3.20 (Docker)
React:    19.0.0
OS:       Linux x86_64
```

### Which area(s) are affected? (Select all that apply)

Webpack, Turbopack, Font (next/font)

### Which stage(s) are affected? (Select all that apply)

next build (local)

### Additional context

Hi everyone — I've been digging into some weird build log differences between `next build --webpack` and `next build` (Turbopack) and wanted to write up what I found in case others have run into the same thing or can add context.

There are two separate observations below. I'm not 100% sure whether either of these is a bug, a regression, or intentional — just sharing the data.

---

### What I noticed

[...]

PR #95981, opened 2026-07-20
fix(next/font/google): bound Google Fonts fetch timeout on Turbopack

aurorascharff:
### What?

Bound the compile-time Google Fonts fetch on the Turbopack path so `next dev` / `next build` can't hang forever on a stuck network.

### Why?

`next/font/google` fetches from Google Fonts at compile time. When the connection *hangs* (captive portal, packet-dropping proxy, broken IPv6), the fetch had no timeout, so compilation blocked until the OS connect timeout (~75s) or indefinitely.

### How?

- `FetchClientConfig` gets a `connect_timeout` (10s) and total `timeout` (60s); Google Fonts overrides them to 5s / 30s.
- Transient failures retry up to 3×, each attempt its own `duration_span!`.
- On failure: `next build` errors (fails the build), `next dev` warns and uses the fallback font. Both messages report the attempt count and suggest `next/font/local`.

### Follow-ups

- [ ] Reduce the dev-mode timeout (build keeps the longer value).
- [ ] Port the fix to the webpack/JS path (`packages/font/src/google`).

 Related: #92301.  #76473.

### DIFF
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -2711,7 +2711,7 @@ impl NextConfig {
         // TODO: Use a less aggressive timeout for build than dev.
         FetchClientConfig {
             connect_timeout: std::time::Duration::from_secs(5),
-            timeout: std::time::Duration::from_secs(30),
+            timeout: Duration::from_secs(30),
             ..Default::default()
         }
         .cell()
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -2705,19 +2705,21 @@ impl NextConfig {
     }
 
     #[turbo_tasks::function]
-    pub fn fetch_client(&self) -> Vc<FetchClientConfig> {
-        // Used for the Google Fonts fetch.
-        // Use timeouts to prevent indefinite/long hangs.
-        // `next dev` will fall back to system fonts if this fails.
-        // `next build` will fail if this times out.
-        // TODO: Use a less aggressive timeout for build than dev.
-        FetchClientConfig {
-            connect_timeout: Duration::from_secs(10),
-            timeout: Duration::from_secs(30),
-            max_retries: 3,
+    pub async fn fetch_client(&self, next_mode: Vc<NextMode>) -> Result<Vc<FetchClientConfig>> {
+        // Bounds the Google Fonts fetch. Dev fails fast and falls back to a system font;
+        // build tolerates a slower network since a missing font fails the build.
+        let (connect_timeout, timeout) = if matches!(*next_mode.await?, NextMode::Development) {
+            (Duration::from_secs(5), Duration::from_secs(10))
+        } else {
+            (Duration::from_secs(10), Duration::from_secs(30))
+        };
+        Ok(FetchClientConfig {
+            connect_timeout,
+            timeout,
+            max_retries: 1,
             ..Default::default()
         }
-        .cell()
+        .cell())
     }
 
     #[turbo_tasks::function]
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use anyhow::{Context, Result, bail};
 use async_trait::async_trait;
 use bincode::{Decode, Encode};
@@ -2712,6 +2714,7 @@ impl NextConfig {
         FetchClientConfig {
             connect_timeout: Duration::from_secs(10),
             timeout: Duration::from_secs(30),
+            max_retries: 3,
             ..Default::default()
         }
         .cell()
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -2710,7 +2710,7 @@ impl NextConfig {
         // `next build` will fail if this times out.
         // TODO: Use a less aggressive timeout for build than dev.
         FetchClientConfig {
-            connect_timeout: std::time::Duration::from_secs(5),
+            connect_timeout: Duration::from_secs(10),
             timeout: Duration::from_secs(30),
             ..Default::default()
         }
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -2704,10 +2704,11 @@ impl NextConfig {
 
     #[turbo_tasks::function]
     pub fn fetch_client(&self) -> Vc<FetchClientConfig> {
-        // Used for the Google Fonts fetch, which downloads from a fast CDN at compile time. Bound
-        // it tightly: a short connect timeout fails fast when a proxy or captive portal hangs the
-        // connection (so `next build`/`next dev` fall back instead of blocking indefinitely), while
-        // the total timeout stays generous enough for a slow-but-working network during a build.
+        // Used for the Google Fonts fetch.
+        // Use timeouts to prevent indefinite/long hangs.
+        // `next dev` will fall back to system fonts if this fails.
+        // `next build` will fail if this times out.
+        // TODO: Use a less aggressive timeout for build than dev.
         FetchClientConfig {
             connect_timeout: std::time::Duration::from_secs(5),
             timeout: std::time::Duration::from_secs(30),
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -2704,7 +2704,16 @@ impl NextConfig {
 
     #[turbo_tasks::function]
     pub fn fetch_client(&self) -> Vc<FetchClientConfig> {
-        FetchClientConfig::default().cell()
+        // Used for the Google Fonts fetch, which downloads from a fast CDN at compile time. Bound
+        // it tightly: a short connect timeout fails fast when a proxy or captive portal hangs the
+        // connection (so `next build`/`next dev` fall back instead of blocking indefinitely), while
+        // the total timeout stays generous enough for a slow-but-working network during a build.
+        FetchClientConfig {
+            connect_timeout: std::time::Duration::from_secs(5),
+            timeout: std::time::Duration::from_secs(30),
+            ..Default::default()
+        }
+        .cell()
     }
 
     #[turbo_tasks::function]
--- a/crates/next-core/src/next_font/google/mod.rs
+++ b/crates/next-core/src/next_font/google/mod.rs
@@ -20,7 +20,7 @@ use turbopack_core::{
     asset::AssetContent,
     context::AssetContext,
     ident::Layer,
-    issue::{IssueExt, IssueSeverity, StyledString},
+    issue::{IssueExt, IssueSeverity},
     module_graph::{ModuleGraph, SingleModuleGraph},
     reference_type::{InnerAssets, ReferenceType},
     resolve::{
@@ -49,7 +49,7 @@ use crate::{
             stylesheet::build_stylesheet,
             util::{get_font_axes, get_stylesheet_url},
         },
-        issue::NextFontIssue,
+        issue::GoogleFontsFetchIssue,
         util::{
             FontCssProperties, FontFamilyType, can_use_next_font, get_request_hash, get_request_id,
             get_scoped_font_family,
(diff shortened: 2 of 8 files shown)


packages/font/src/google/fetch-resource.ts
```diff
@@ -15,12 +15,21 @@ export function fetchResource(
...
+    const timeout = isDev ? 3000 : 6000
+
+    // A proxy agent establishes its CONNECT tunnel before the request socket is
+    // assigned, so `req.setTimeout` below never fires against a hanging proxy
+    // (e.g. a captive portal whose ToS hasn't been accepted). Bound the agent's
+    // own timeout so that case is covered too.
+    const agent = getProxyAgent()
+    if (agent && timeout) {
+      ;(agent as unknown as { timeout?: number }).timeout = timeout
+    }
```
bgw: This is nasty.

Can we update `http-proxy-agent`? It looks like this is a hack to work around the constructor function not exposing the options object that accepts `timeout`.

Newer versions of the package take an `options` argument:
- https://npmx.dev/package/http-proxy-agent/v/9.1.0#user-content-new-httpproxyagentproxy-string-url-options-httpproxyagentoptions
- https://nodejs.org/api/http.html#new-agentoptions

packages/font/src/google/loader.test.ts
```diff
@@ -148,4 +151,42 @@ describe('next/font/google loader', () => {
...
+          emitFontFile: jest.fn(),
+          resolve: jest.fn(),
+          loaderContext: {} as any,
+          isDev: false,
+          isServer: true,
+          variableName: 'myFont',
+        })
+      ).rejects.toThrow()
+    })
+  })
```
bgw: This isn't really testing a timeout though? It's testing an immediate failure, which isn't really a complicated edge case.

The test in `packages/font/src/google/fetch-resource.test.ts` seems good, I'd remove this one, it doesn't seem to provide much value.

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -17,6 +17,9 @@ use turbo_tasks::{
 use crate::{FetchError, FetchResult, HttpResponse, HttpResponseBody};
 
 const MAX_CLIENTS: usize = 16;
+/// Number of times a transient fetch failure (connection error / timeout) is retried before giving
+/// up and surfacing the error to the caller.
+const MAX_FETCH_RETRIES: u32 = 3;
```
bgw: Let's make this part of `FetchClientConfig`?

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -34,12 +37,17 @@ pub struct FetchClientConfig {
...
+    /// (offline, a captive portal that drops packets, a blocked endpoint) blocks compilation until
+    /// the OS-level connect timeout fires (~75s). Defaults to 3 seconds.
+    pub connect_timeout: Duration,
 }
 
 impl Default for FetchClientConfig {
     fn default() -> Self {
         Self {
             min_cache_control: Duration::from_secs(60 * 60),
+            connect_timeout: Duration::from_secs(3),
```
bgw: If we want to match the JS behavior, we should use a longer timeout for builds, which implies that we should pass down this value from the callsites (there's no global value we can check for build vs dev).

IMO the default value here should be very large (e.g. 60 seconds) because this specific crate is trying to be very generic, but we should certainly override it for google fonts.

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -169,9 +177,26 @@ impl FetchClientConfig {
...
+                        Ok(response) => break response,
+                        Err(err)
+                            if attempt < MAX_FETCH_RETRIES
+                                && (err.is_connect() || err.is_timeout() || err.is_request()) =>
+                        {
+                            attempt += 1;
+                            tracing::warn!(
+                                "Failed to fetch {url_ref}, retrying \
+                                 ({attempt}/{MAX_FETCH_RETRIES})..."
+                            );
```
bgw: This is fine, but FYI `tracing` events are not shown to the user, they only show up in our trace files that we view with https://trace.nextjs.org/ (and we prefer spans for things like retries, not events)

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -68,7 +76,7 @@ impl FetchClientConfig {
 
     fn try_build_uncached_reqwest_client(&self) -> reqwest::Result<reqwest::Client> {
         #[allow(unused_mut)]
-        let mut builder = reqwest::Client::builder();
+        let mut builder = reqwest::Client::builder().connect_timeout(self.connect_timeout);
```
bgw: It looks like the JS code has both a normal timeout *and* a connect timeout. This only sets the connect timeout.

- https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout
- https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout

I think it would be reasonable for us to just set `timeout` (the "total timeout") and not `connect_timeout`. If we set both `connect_timeout` should be some smaller value than `timeout`.

packages/font/src/google/fetch-resource.ts
icyJoseph: Mmm yeah, I asked about this, but I had only thought of plain JS. 

The reason is that I think there's a bug, the agent doesn't pick up on the `req.setTimeout`, so the request hangs ignoring the 3 seconds timeout. This 'hack', does the trick, but if bumping also fixes it, we should do that indeed.

bgw: I'm not saying that just updating the package fixes this. What I am saying is that updating the package lets the constructor take a proper options object, so that you don't have to bypass type safety and mutate an undocumented field on an object.

crates/next-core/src/next_config.rs
```diff
@@ -2704,7 +2704,16 @@ impl NextConfig {
 
     #[turbo_tasks::function]
     pub fn fetch_client(&self) -> Vc<FetchClientConfig> {
-        FetchClientConfig::default().cell()
+        // Used for the Google Fonts fetch, which downloads from a fast CDN at compile time. Bound
+        // it tightly: a short connect timeout fails fast when a proxy or captive portal hangs the
+        // connection (so `next build`/`next dev` fall back instead of blocking indefinitely), while
+        // the total timeout stays generous enough for a slow-but-working network during a build.
+        FetchClientConfig {
+            connect_timeout: std::time::Duration::from_secs(5),
```
bgw: - Let's be really conservative, because this could cause a build to fail.
- Import `std::time::Duration` at the top of the file

```suggestion
            connect_timeout: Duration::from_secs(10),
```

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -169,9 +188,28 @@ impl FetchClientConfig {
...
+                    // nested, timed spans in the trace viewer (https://trace.nextjs.org/).
+                    let result = {
+                        let _span = duration_span!("fetch attempt", url = url_ref, attempt);
+                        request.send().await.and_then(|r| r.error_for_status())
+                    };
+                    match result {
+                        Ok(response) => break response,
+                        Err(err)
+                            if attempt < MAX_FETCH_RETRIES
+                                && (err.is_connect() || err.is_timeout() || err.is_request()) =>
```
bgw: It may also be reasonable to retry 5xx error codes here

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -17,6 +17,10 @@ use turbo_tasks::{
 use crate::{FetchError, FetchResult, HttpResponse, HttpResponseBody};
 
 const MAX_CLIENTS: usize = 16;
+/// Number of times a transient fetch failure (connection error / timeout) is retried before giving
+/// up and surfacing the error to the caller. The total number of attempts made is
+/// `MAX_FETCH_RETRIES + 1`.
+pub const MAX_FETCH_RETRIES: u32 = 3;
```
bgw: Move this constant into the `FetchClientConfig` object, and let's default to 0 retries (but pass down a option from google fonts).

crates/next-core/src/next_config.rs
```diff
@@ -2704,7 +2704,16 @@ impl NextConfig {
...
+            timeout: std::time::Duration::from_secs(30),
```
bgw: FYI with a 30 second timeout + 4 attempts, this can still hang up to 2 minutes. I think that's okay, but just pointing this out.

bgw:
```suggestion
            timeout: Duration::from_secs(30),
```

crates/next-core/src/next_font/google/mod.rs
```diff
@@ -281,8 +281,13 @@ impl NextFontGoogleCssModuleReplacer {
...
                                 format!(
-                                    "Failed to fetch `{}` from Google Fonts.",
-                                    options.await?.font_family
+                                    "Failed to fetch `{}` from Google Fonts after {} \
+                                     attempts.\nIf you are offline or behind a proxy that blocks \
+                                     `fonts.googleapis.com`, self-host the font with \
+                                     `next/font/local`, or make sure the endpoint is reachable \
+                                     during the build.",
+                                    options.await?.font_family,
+                                    MAX_FETCH_RETRIES + 1,
```
bgw: Maybe also mention the possibility of configuring a proxy with `HTTP_PROXY`/`HTTPS_PROXY`

crates/next-core/src/next_font/google/mod.rs
```diff
@@ -281,8 +281,13 @@ impl NextFontGoogleCssModuleReplacer {
                             .resolved_cell(),
                             description: StyledString::Text(
                                 format!(
-                                    "Failed to fetch `{}` from Google Fonts.",
-                                    options.await?.font_family
+                                    "Failed to fetch `{}` from Google Fonts after {} \
```
bgw: Don't use format + `{}` interpolation with `StyledString`. Prefer to compose a `StyledString::Line`: https://turbopack-rust-docs.vercel.sh/rustdoc/turbopack_core/issue/enum.StyledString.html

crates/next-core/src/next_font/google/mod.rs
```diff
@@ -281,8 +281,13 @@ impl NextFontGoogleCssModuleReplacer {
...
+                                     attempts.\nIf you are offline or behind a proxy that blocks \
```
bgw: Don't put `\n` in `StyledString::Text`, prefer `StyledString::Stack`: https://turbopack-rust-docs.vercel.sh/rustdoc/turbopack_core/issue/enum.StyledString.html

crates/next-core/src/next_font/google/mod.rs
```diff
@@ -305,9 +310,12 @@ impl NextFontGoogleCssModuleReplacer {
...
                                 format!(
-                                    "Failed to download `{}` from Google Fonts. Using fallback \
-                                     font instead.",
-                                    options.await?.font_family
+                                    "Failed to download `{}` from Google Fonts after {} attempts. \
+                                     Using a fallback font instead.\nIf you are offline or behind \
+                                     a proxy that blocks `fonts.googleapis.com`, self-host the \
+                                     font with `next/font/local`.",
+                                    options.await?.font_family,
+                                    MAX_FETCH_RETRIES + 1,
```
bgw: Can we share this description between both dev and prod?

Maybe make `NextFontIssue::description()` on the trait impl generate the description instead of storing the `StyledString` on the struct. That's better anyways, as we want description formatting to be lazy, as issues can get filtered out before they're displayed.

turbopack/crates/turbo-tasks-fetch/src/client.rs
```diff
@@ -169,9 +188,28 @@ impl FetchClientConfig {
 
             let response = {
                 let _span = duration_span!("fetch request", url = url_ref);
-                builder.send().await
-            }
-            .and_then(|r| r.error_for_status())?;
+                let mut attempt = 0;
+                loop {
+                    // A GET request without a body is always cloneable.
```
bgw:
```suggestion
```

Your `expect` string should basically be your comment, don't need a comment + `expect`.

test/e2e/next-font/google-fetch-error.test.ts
```diff
@@ -7,6 +7,9 @@ const mockedGoogleFontResponses = require.resolve(
 
 describe('next/font/google fetch error', () => {
   const isDev = (global as any).isNextDev
+  // Turbopack bounds the fetch with retries and an enriched message; the
+  // webpack/JS path still emits the original wording (updated in a follow-up).
```
bgw:
```suggestion
```

crates/next-core/src/next_config.rs
```diff
@@ -2704,7 +2704,16 @@ impl NextConfig {
 
     #[turbo_tasks::function]
     pub fn fetch_client(&self) -> Vc<FetchClientConfig> {
-        FetchClientConfig::default().cell()
+        // Used for the Google Fonts fetch, which downloads from a fast CDN at compile time. Bound
+        // it tightly: a short connect timeout fails fast when a proxy or captive portal hangs the
+        // connection (so `next build`/`next dev` fall back instead of blocking indefinitely), while
+        // the total timeout stays generous enough for a slow-but-working network during a build.
```
bgw: Many word. Caveman do good less word.

```suggestion
        // Used for the Google Fonts fetch.
        // Use timeouts to prevent indefinite/long hangs.
        // `next dev` will fall back to system fonts if this fails.
        // `next build` will fail if this times out.
        // TODO: Use a less aggressive timeout for build than dev.
```

bgw (changes requested):

crates/next-core/src/next_config.rs
aurorascharff: hm yeah i think thats still too long, will fix

aurorascharff: I did the dev/prod split instead and also made it shorter

crates/next-core/src/next_font/issue.rs
```diff
@@ -35,3 +36,73 @@ impl Issue for NextFontIssue {
...
+            StyledString::Code(rcstr!("HTTP_PROXY")),
+            StyledString::Text(rcstr!("/")),
+            StyledString::Code(rcstr!("HTTPS_PROXY")),
+            StyledString::Text(rcstr!(" so Next.js can reach ")),
+            StyledString::Code(rcstr!("fonts.googleapis.com")),
+            StyledString::Text(rcstr!(".")),
+        ]);
+        Ok(Some(StyledString::Stack(vec![summary, guidance])))
+    }
+}
```
bgw, line 108: Awesome. This is much nicer, thanks.

crates/next-core/src/next_config.rs
```diff
@@ -2703,8 +2705,21 @@ impl NextConfig {
...
-    pub fn fetch_client(&self) -> Vc<FetchClientConfig> {
-        FetchClientConfig::default().cell()
+    pub async fn fetch_client(&self, next_mode: Vc<NextMode>) -> Result<Vc<FetchClientConfig>> {
+        // Bounds the Google Fonts fetch. Dev fails fast and falls back to a system font;
+        // build tolerates a slower network since a missing font fails the build.
+        let (connect_timeout, timeout) = if matches!(*next_mode.await?, NextMode::Development) {
+            (Duration::from_secs(5), Duration::from_secs(10))
+        } else {
+            (Duration::from_secs(10), Duration::from_secs(30))
+        };
```
bgw, line 2715: 👍

bgw (approved):
Thanks for doing this @aurorascharff

PR #95983, opened 2026-07-20
fix: add missing export default to HoverPrefetchLink and .Provider to…

Syed-Umer-Ali:
### What?

Two broken code examples in the App Router docs that fail when copy-pasted.

### Why?

1. **`04-linking-and-navigating.mdx`** — The `HoverPrefetchLink` component example is missing `export default`, so it cannot be imported from its file.
2. **`06-fetching-data.mdx`** — The `UserProvider` example uses `<UserContext>` instead of `<UserContext.Provider>`, so `useContext(UserContext)` always returns the default value (`null`) instead of the provided promise.

### How?

- Added `export default` to both TSX and JSX variants of `HoverPrefetchLink`
- Changed `<UserContext>` to `<UserContext.Provider>` in both TSX and JSX variants of `UserProvider`


icyJoseph:
Starting in React 19, after you create context, like:

```tsx
const MyContext = createContext()
```

Then you can render `<MyContext>` as a provider. In older versions of React you still need to do `<MyContext.Provider>`, but App Router uses React 19, so the `.Provider` is not needed.

docs/01-app/01-getting-started/06-fetching-data.mdx
```diff
@@ -595,7 +598,7 @@ import { createContext } from 'react'
 export const UserContext = createContext(null)
 
 export default function UserProvider({ children, userPromise }) {
-  return <UserContext value={userPromise}>{children}</UserContext>
+  return <UserContext.Provider value={userPromise}>{children}</UserContext.Provider>
```
icyJoseph, line 601: Reposting from: https://github.com/vercel/next.js/pull/95983#issuecomment-5027101862

Starting in React 19, after you create context, like:

```tsx
const MyContext = createContext()
```

Then you can render `<MyContext>` as a provider. In older versions of React you still need to do `<MyContext.Provider>`, but App Router uses React 19, so the `.Provider` is not needed.

PR #95984, opened 2026-07-20
[test] Add coverage for throwing custom "use cache" cache handler

eps1lon:
This is relevant if the cache is in a broken state or the keys or entries are not suitable (e.g. an entry that's to large to be stored).

Revealed several oddities:
1. cache handler have no no sources content (not too bad since that means the codeframe points into the problem `use cache` function which is usually the cause)
2. A throwing `set` is swallowed in dev
3. A throwing `get` is logged as a caught error and an unhandled rejection. This is spammy.

### DIFF
--- a/test/e2e/app-dir/app-custom-cache-handler-errors/app-custom-cache-handler-errors.test.ts
+++ b/test/e2e/app-dir/app-custom-cache-handler-errors/app-custom-cache-handler-errors.test.ts
@@ -1,0 +1,136 @@
+import { nextTestSetup } from 'e2e-utils'
+import { hasErrorToast, retry, waitForNoRedbox } from 'next-test-utils'
+import stripAnsi from 'strip-ansi'
+
+describe('app-custom-cache-handler-errors - get throws', () => {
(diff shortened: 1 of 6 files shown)


test/e2e/app-dir/app-custom-cache-handler-errors/app-custom-cache-handler-errors.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (isNextDev) {
+      if (isTurbopack) {
+        await expect(browser).toDisplayRedbox(`
+         {
+           "description": "CustomCacheHandler.get failed",
+           "environmentLabel": "Prefetch",
+           "label": "Runtime Error",
+           "source": "app/page.tsx (19:14) @ CachedData
+         > 19 |   return <p>{await getCachedData(input)}</p>
+              |              ^",
```
eps1lon, line 29: This isn't too bad in the case where the input key being too large is an issue.

unstubbable (approved):

Issue #83905, opened 2025-09-17
Docs: Polyfills are not awaited


r34son:
### What is the documentation issue?

https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation-client#polyfills

### Is there any context that might help us understand?

polyfill import is not awaited. How can i be sure that any code that contains unsupported features will not be executed?

My example (instrumentation-client.ts):

```ts
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-nocheck
delete Intl.Locale;
delete Intl.getCanonicalLocales;
console.time('IntlLocale');
// https://formatjs.github.io/docs/polyfills/intl-locale#dynamic-import--capability-detection
import { shouldPolyfill as shouldPolyfillLocale } from '@formatjs/intl-locale/should-polyfill';
// https://formatjs.github.io/docs/polyfills/intl-getcanonicallocales#dynamic-import--capability-detection
import { shouldPolyfill as shouldPolyfillGetcanonicallocales } from '@formatjs/intl-getcanonicallocales/should-polyfill';

if (shouldPolyfillLocale()) {
    if (shouldPolyfillGetcanonicallocales()) {
        await import('@formatjs/intl-getcanonicallocales/polyfill');
        console.log('IntlGetcanonicallocales is polyfilled');
    } else {
        console.warn('IntlGetcanonicallocales is not polyfilled');
    }
    await import('@formatjs/intl-locale/polyfill');
    console.log('IntlLocale is polyfilled');
} else {
    console.warn('IntlLocale is not polyfilled');
}
console.timeEnd('IntlLocale');
```

Locally it logs around 100ms and warning printed in dev by https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation-client#performance-considerations

### Does the docs page already exist? Please link to it.

https://nextjs.org/docs/app/api-reference/file-conventions/instrumentation-client#polyfills

r34son:
@icyJoseph Hi Joseph! I am sure that you can help me with this 🙏

r34son:
also warns in terminal:

```
 ⚠ ./instrumentation-client.ts
The generated code contains 'async/await' because this module is using "topLevelAwait".
However, your target environment does not appear to support 'async/await'.
As a result, the code may not run as expected or may cause runtime errors.

Import trace for requested module:
./instrumentation-client.ts
[...]

PR #95988, opened 2026-07-21
docs: clarify instrumentation-client polyfill loading and hydration timing

icyJoseph:
The `instrumentation-client` Polyfills docs recommended a conditional dynamic `import().then()` to load a polyfill. That import is fire-and-forget: Next.js loads this file synchronously and does not await it before hydration, so the polyfill can install after components have already run (see #83905). 

- In "Execution timing" mention that only synchronous top-level code completes before hydration. Call out that a conditional `import()` here is not awaited, and cross-reference `supported-browsers` for load-on-demand strategies and the built-in baseline.

- In **`supported-browsers.mdx`**, "fix" the `Conditionally loading polyfills` example to `await import()` instead of a floating `.then()`, otherwise, a reader needs to put their code within the `.then` chain of the polyfill load.

Closes: https://linear.app/vercel/issue/DOC-6506/revisit-polyfills-with-instrumentation-client


docs/01-app/03-api-reference/03-file-conventions/instrumentation-client.mdx
```diff
@@ -236,27 +238,21 @@ export function onRouterTransitionStart(url) {
...
-  import('./lib/polyfills/resize-observer').then((mod) => {
-    window.ResizeObserver = mod.default
-  })
+  window.ResizeObserver = ResizeObserverPolyfill
 }
 ```
 
-```js filename="instrumentation-client.js" switcher
-import './lib/polyfills'
+A conditional `import()` here is fire-and-forget and may land after hydration, so for anything loaded on demand, prefer polyfilling in the code that uses the feature. See [Custom Polyfills](/docs/architecture/supported-browsers#custom-polyfills) for that and other strategies.
```
SukkaW: Should we add a failing example here to better teach people to avoid this pattern?

```js filename="instrumentation-client.js" switcher
// DON'T DO THIS, it has race condition implications
if (!window.ResizeObserver) {
  import('./lib/polyfills/resize-observer').then((mod) => {
	  // By the time './lib/polyfills/resize-observer'
    // is loaded and executed by the browser, Next.js
    // may already have begun hydration and could be
    // too late.
    window.ResizeObserver = mod.default
  })
}
```

icyJoseph: Added 👍 

I think the rest of the edit looks good at this point

ztanner (approved):

PR #95989, opened 2026-07-21
[turbopack] Track re-exports in `import_usage` inside of `compute_import_usage`

### LINKED ISSUE #95698: Turbopack bundling issue (webpack behaviour is correct, turbopack does not tree shake correctly)

sampoder:
Previously, code such as:

```
export { foo } from "bar"
```

Would contribute to a module's exports (we export foo).

However, it would not contribute to our tracking of imports from "bar".  Therefore we would fall back to the default import usage:

https://github.com/vercel/next.js/blob/c0a7b91becb33490bf412d02bbf09ebaa7216083/turbopack/crates/turbopack-ecmascript/src/references/exports.rs#L140

```rust
pub enum ImportUsage {
    #[default]
    TopLevel,
    Exports(FrozenSet<RcStr>),
}
```

We would therefore not tree-shake. This is rarely a problem with normal imports / exports because they support following reexports and this is enabled by default in Turbopack. However, async imports do not support following reexports.

In https://github.com/vercel/next.js/issues/95698, the import of the Viem chain definitions is an async-imported module:

Screenshot 2026-07-20 at 5 47 43 pm

Closes https://github.com/vercel/next.js/issues/95698

I looked at the reproduction from the original issue, and this was the result:

```
Client graph contains 11/712 viem chain definition files.
Not reproduced; missing 701 definitions.
```

Yay!

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
@@ -6,6 +6,7 @@ use std::{
 };
 
 use anyhow::{Context, Result};
+use auto_hash_map::AutoSet;
 use rustc_hash::{FxHashMap, FxHashSet};
 use smallvec::SmallVec;
 use swc_core::{
@@ -284,6 +285,8 @@ pub(crate) struct ProgramDeclUsage {
     pub(crate) decl_usages: FxHashMap<Id, DeclUsage>,
     // import -> immediate usage (top level decl)
     pub(crate) import_usages: FxHashMap<usize, DeclUsage>,
+    // import reference -> names it is directly re-exported as (`export { x } from '...'`)
+    pub(crate) named_reexports: FxHashMap<usize, AutoSet<RcStr>>,
     // export name -> top level decl
     pub(crate) exports: FxHashMap<RcStr, Id>,
 }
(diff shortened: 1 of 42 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/imports.rs
```diff
@@ -284,6 +284,8 @@ pub(crate) struct ProgramDeclUsage {
     pub(crate) decl_usages: FxHashMap<Id, DeclUsage>,
     // import -> immediate usage (top level decl)
     pub(crate) import_usages: FxHashMap<usize, DeclUsage>,
+    // import reference -> names it is directly re-exported as (`export { x } from '...'`)
+    pub(crate) reexports: FxHashMap<usize, FxHashSet<RcStr>>,
```
mischnic: What about `export {x as y} from "..."` though?

lukesandberg: maybe called these `named_reexports`?

i am also thinking about `export * from ....` and `export * as foo from ....`

finally consider an `AutoSet` instead of a FxHashSet since we expect these to be small (or even an ImmutableSet i guess)

sampoder: We don't do anything with `export * from ....` at the moment. `export * as foo from ....` gets added to reexports as "foo". Switching to `named_reexports` makes sense, I'll do that.

sampoder: This gets added as `y`

sampoder: Added a test for this

sampoder: Switched to AutoSet and change the name

lukesandberg (approved):

PR #95993, opened 2026-07-21
[turbopack] Follow re-exports for side-effect free async modules

sampoder:
This adds a very basic follow re-exports support to async-imported modules that are side effect free. For example, it supports:

```javascript
// a.js
export const a = 'A'

// b.js
export const b = 'B'        // unused

// barrel.js  (pure re-export barrel)
export { a } from './a.js'
export { b } from './b.js'

// index.js
const { a } = await import('./barrel.js')
console.log(a) 
```

`apply_reexport_tree_shaking` was moved into `turbopack-ecmascript` as part of this.


turbopack/crates/turbopack-ecmascript/src/references/esm/dynamic.rs
```diff
@@ -91,14 +96,54 @@ impl ModuleReference for EsmAsyncAssetReference {
...
             *self.origin,
             *self.request,
             EcmaScriptModulesReferenceSubType::DynamicImport,
             self.error_mode,
             Some(self.issue_source),
         )
-        .await
+        .await?;
+
+        // `const { x } = await import(...)`: follow the potential re-export to its origin.
```
bgw, line 108: Some comment here, I don't know if my wording is good or not

```suggestion
        // `const { x } = await import(...)`: follow the potential re-export to its origin.
        // only follow the re-export if there's a single member to avoid splitting and introducing new dynamic entrypoints
```

bgw (approved):

PR #95996, opened 2026-07-21
[turbopack] Tree-shake `module.exports = {...}` (CJS)

sampoder:
This PR builds on https://github.com/vercel/next.js/pull/95716 by adding support tree-shaking exports exported using the `module.exports = {...}` syntax

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
@@ -1,7 +1,10 @@
 use swc_core::{
     common::Mark,
     ecma::{
-        ast::{CallExpr, Callee, Expr, Ident, Lit, MemberExpr, MemberProp, Prop, PropOrSpread},
+        ast::{
+            AssignExpr, AssignOp, AssignTarget, CallExpr, Callee, Expr, Ident, Lit, MemberExpr,
+            MemberProp, ObjectLit, Prop, PropOrSpread, SimpleAssignTarget,
+        },
         utils::prop_name_eq,
     },
 };
@@ -96,6 +99,27 @@ pub(crate) fn define_property_sets_es_module(call: &CallExpr) -> bool {
                 if prop_name_eq(&kv.key, "value")
                     && matches!(unparen(&kv.value), Expr::Lit(Lit::Bool(b)) if b.value)))
     })
+}
+
+/// The object literal of a `module.exports = { … }` whole-exports assignment,
+/// whose properties are the module's named exports.
+pub(crate) fn as_module_exports_object_literal(
+    n: &AssignExpr,
+    unresolved_mark: Mark,
+) -> Option<&ObjectLit> {
+    if n.op != AssignOp::Assign {
+        return None;
+    }
+    let AssignTarget::Simple(SimpleAssignTarget::Member(member)) = &n.left else {
+        return None;
+    };
+    if !is_module_dot_exports(member, unresolved_mark) {
+        return None;
+    }
+    match unparen(&n.right) {
+        Expr::Object(obj) => Some(obj),
+        _ => None,
+    }
 }
 
 /// Whether `member` writes the module's own CommonJS exports: `exports.<x>`,
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
@@ -22,8 +22,8 @@ use crate::{
     analyzer::{
         Bump, BumpVec, ConstantValue, JsValue, WellKnownFunctionKind,
         cjs_ast::{
-            as_exports_define_property, define_property_sets_es_module, is_exports_object,
-            is_global,
+            as_exports_define_property, as_module_exports_object_literal,
+            define_property_sets_es_module, is_exports_object, is_global,
         },
         graph::{ConditionalKind, Effect, EffectArg, EffectsBlock, EvalContext, VarGraph},
         is_unresolved_id,
@@ -575,8 +575,22 @@ mod analyzer_state {
         }
 
         pub(super) fn record_cjs_export(&mut self, name: RcStr, path: AstPath) {
+            self.push_cjs_export(DroppableCjsExportAssignment::Write { name, path });
+        }
+
+        /// Records the named exports of a `module.exports = { … }` object literal. They
+        /// share `path` (the assignment); the code-gen removes each unused property.
+        pub(super) fn record_cjs_object_literal_exports(
+            &mut self,
+            names: Vec<RcStr>,
+            path: AstPath,
+        ) {
+            self.push_cjs_export(DroppableCjsExportAssignment::ObjectLiteral { names, path });
+        }
+
+        fn push_cjs_export(&mut self, drop: DroppableCjsExportAssignment) {
             if let Some(c) = &mut self.state.cjs_exports {
-                c.writes.push(DroppableCjsExportAssignment { name, path });
+                c.writes.push(drop);
             }
         }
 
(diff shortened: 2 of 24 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -572,8 +575,20 @@ mod analyzer_state {
         }
 
         pub(super) fn record_cjs_export(&mut self, name: RcStr, path: AstPath) {
+            self.push_cjs_export(name, path, CjsExportDropKind::Write);
+        }
+
+        /// Records a named export of a `module.exports = { … }` object literal. All
+        /// such drops for one object share `path` (the assignment); the code-gen
+        /// removes each unused property from the object literal.
+        pub(super) fn record_cjs_object_export(&mut self, name: RcStr, path: AstPath) {
```
lukesandberg: nit maybe 'record_cjs_object_literal_export`

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1237,6 +1252,42 @@ impl<'a> Analyzer<'a, '_> {
             as_parent_path(ast_path).into(),
         );
     }
+
+    /// Records the named exports of a top-level `module.exports = { … }` object
+    /// literal. A spread, method, getter/setter, or computed key makes the export
```
lukesandberg: i think supporting `getter`s is reasonable (setters too i guess?)

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1237,6 +1252,42 @@ impl<'a> Analyzer<'a, '_> {
...
+        // flag and never drop it (below).
+        if object_literal_sets_es_module(obj) {
+            self.set_cjs_has_es_module();
+        }
+        match static_prop_names(obj) {
+            Some(names) => {
+                let path: AstPath = as_parent_path(ast_path).into();
+                for name in names {
+                    if &*name == "__esModule" {
+                        continue;
```
lukesandberg: nit: just call `set_cjs_has_es_module` here

unless it is important to call that even if we are tainting?

turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
```diff
@@ -44,6 +51,71 @@ pub(crate) fn is_module_exports_chain(expr: &Expr, unresolved_mark: Mark) -> boo
...
+/// The droppable named exports of a `module.exports = { … }` literal. `None` if a
+/// spread or computed key makes the set unknown.
+pub(crate) fn static_prop_names(obj: &ObjectLit) -> Option<Vec<RcStr>> {
+    let mut names = Vec::with_capacity(obj.props.len());
+    for prop in &obj.props {
+        let PropOrSpread::Prop(prop) = prop else {
+            return None; // spread → unknown export set
+        };
+        match &**prop {
+            Prop::Shorthand(id) => names.push(RcStr::from(id.sym.as_str())),
```
lukesandberg: i think you need the `is_pure_object_value` here as well

turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
```diff
@@ -44,6 +51,71 @@ pub(crate) fn is_module_exports_chain(expr: &Expr, unresolved_mark: Mark) -> boo
...
+            Prop::KeyValue(kv) => {
+                let name = match &kv.key {
+                    PropName::Ident(i) => RcStr::from(i.sym.as_str()),
+                    PropName::Str(s) => RcStr::from(s.value.to_string_lossy().into_owned()),
+                    _ => return None, // computed / numeric / bigint key
+                };
+                if is_pure_object_value(&kv.value) {
```
lukesandberg: do we need this?

i think our rewrite just needs to preserve the RHS... but i guess that gets tricky?

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: this comment is wrong, we don't taint on getters & setters. we just can't drop them. we only taint if the getter / setter contains a reference to `module.exports` / `this` in a way that can taint.

turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
sampoder: This is a limitation with rewriting `module.exports`:

```
module.exports = { a: 1, b: sideEffect() }
→ module.exports = { a: 1 }
```

it's a bit messy to try to preserve the RHS. You'd have to transform the whole `module.exports` into:

```
const A = 1
const B = sideEffect()
module.exports = { a: A }
```

turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
sampoder: This is for syntax `{ a }` - which I think has to be pure?

Image

Unless I'm missing something.

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
sampoder: This has been moved around to get rid of the static names helper so now we just cleanly call `set_cjs_has_es_module` in the same traversal

turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
lukesandberg: good point.

it has to be an ident

turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
lukesandberg: `module.exports = {...(void sideEffect())`

is the solution

ill upstream an optimization to SWC to fold it all the way, but even now it is better than the alternatives

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
lukesandberg: discussed IRL: could be a TODO, but i don't see any problem with supporting this

turbopack/crates/turbopack-ecmascript/src/analyzer/cjs_ast.rs
sampoder: added this! + getters & setters

lukesandberg: https://github.com/swc-project/swc/pull/12068 merged so we should pick it up in our next SWC bump

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1293,6 +1305,64 @@ impl<'a> Analyzer<'a, '_> {
...
+                | Prop::Method(MethodProp { key, .. }) => match key {
+                    PropName::Ident(i) => RcStr::from(i.sym.as_str()),
+                    PropName::Str(s) => RcStr::from(s.value.to_string_lossy().into_owned()),
+                    // computed / numeric / bigint key → unknowable
+                    _ => {
+                        self.taint_cjs_exports();
+                        return;
+                    }
+                },
+                Prop::Assign(_) => continue,
```
lukesandberg:
```suggestion
                Prop::Assign(_) => continue,  // should never happen for a literal
```

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
lukesandberg, line 473: we should merge all the ObjectProperty export drops together, so we only visit the assignment once

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -516,6 +552,34 @@ fn take_define_property_value(call: &mut CallExpr) -> Option<Box<Expr>> {
...
+            let PropOrSpread::Prop(p) = &prop else {
+                return Some(prop);
+            };
+            let value = match &**p {
+                Prop::Shorthand(id) if id.sym.as_ref() == name => Box::new(Expr::Ident(id.clone())),
+                Prop::KeyValue(kv) if prop_name_eq(&kv.key, name) => kv.value.clone(),
+                Prop::Getter(g) if prop_name_eq(&g.key, name) => return None,
+                Prop::Setter(s) if prop_name_eq(&s.key, name) => return None,
+                Prop::Method(m) if prop_name_eq(&m.key, name) => return None,
+                _ => return Some(prop),
```
lukesandberg: unreachable? implies a mismatch in behavior between the analyzer and this function

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -516,6 +552,34 @@ fn take_define_property_value(call: &mut CallExpr) -> Option<Box<Expr>> {
...
+    obj.props = obj
+        .props
+        .take()
+        .into_iter()
+        .filter_map(|prop| {
+            let PropOrSpread::Prop(p) = &prop else {
+                return Some(prop);
+            };
+            let value = match &**p {
+                Prop::Shorthand(id) if id.sym.as_ref() == name => Box::new(Expr::Ident(id.clone())),
```
lukesandberg: for shorthand i think we can just `return None` since it must be an ident

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -516,6 +552,34 @@ fn take_define_property_value(call: &mut CallExpr) -> Option<Box<Expr>> {
...
+                Prop::KeyValue(kv) if prop_name_eq(&kv.key, name) => kv.value.clone(),
+                Prop::Getter(g) if prop_name_eq(&g.key, name) => return None,
+                Prop::Setter(s) if prop_name_eq(&s.key, name) => return None,
+                Prop::Method(m) if prop_name_eq(&m.key, name) => return None,
+                _ => return Some(prop),
+            };
+            Some(PropOrSpread::Spread(SpreadElement {
```
lukesandberg:
```suggestion
            // The value might have a side effect, preserve it by generating `...void (expr)`
            // which preserves relative position and side effects, if there are none SWC can fold it to nothing.
            Some(PropOrSpread::Spread(SpreadElement {
```

lukesandberg (approved):

PR #96017, opened 2026-07-21
Block prefetch task until sufficient response is received

acdlite:
This is a refactor of the prefetching algorithm to block the prefetch task from completing until the response is received.

Currently, a prefetch task is fire-and-forget: once a prefetch's requests are spawned, the scheduler marks the originating task as complete regardless of what the server ends up returning. However, certain planned optimizations require us to verify the response from the server to determine whether additional work is necessary. For example, we may attempt to do a static prefetch instead of a runtime prefetch if we optimistically assume that the static prefetch does not vary on runtime data. We need to verify that assumption is correct before marking the task as complete.

Now, whenever a pass over the task spawns a segment request, or encounters an entry whose response hasn't arrived yet, the task registers itself on the entry and exits as Blocked instead of Done. When the entry resolves, the blocked task is pinged and the pass re-runs against the received data. Only a pass that observes every response it cares about advances the task to the next phase or completes it.

Rejected entries are handled per segment: a rejection counts as an observed response, so the pass skips that segment and keeps prefetching the rest. The task never registers on a rejected entry, since nothing ever pings one. Retries are governed by the existing per-entry logic.

Also fixes a cache bug this exposed: when a revalidation response is keyed at a more generic vary path than the stale partial entry that prompted it (because the segment turned out not to vary on some param), the stale entry shadowed the result on every lookup — the upgraded data was unreachable and the revalidation was wasted. With blocking, that dead end became an infinite loop: each fulfillment re-pinged the blocked task, which re-read the same stale entry and spawned another revalidation. Upserting (or re-keying) a segment now deletes settled entries at more specific vary paths that the incoming entry supersedes.

The router-act test helper's "at least one request initiated" watchdog is adjusted to account for the new blocking behavior: countable requests can now be legitimately gated on an in-flight App Shell response (which the watchdog intentionally doesn't count). On expiry it now proceeds to response processing if the act scope intercepted a shell request, and keeps waiting while shell requests are in flight elsewhere on the page, instead of timing out.

### DIFF
--- a/test/lib/router-act.ts
+++ b/test/lib/router-act.ts
@@ -6,23 +6,25 @@ import { equals } from '@jest/expect-utils'
 // prefetches carry the value '3' (FetchStrategy.RuntimeShell). The App Shell is
 // the param/searchParam-independent chrome of a route — conceptually part of the
 // route itself, not prefetch data in the way we normally think of it. By default
-// we therefore exclude App Shell requests from all `act` assertion logic
-// (`includes` matching, `no-requests`, `block: 'reject'`, and the "at least one
-// request" check). They are still intercepted, fulfilled, and awaited so that the
-// browser caches the shell and no requests are left in flight. Pass
-// `includeAppShellRequests: true` to `createRouterAct` to assert on them directly
-// (e.g. when testing App Shell behavior specifically).
+// we therefore exclude App Shell requests from `act` assertion logic
+// (`includes` matching, `no-requests`, and `block: 'reject'`). They are still
+// intercepted, fulfilled, and awaited so that the browser caches the shell and
+// no requests are left in flight — and they do satisfy the "at least one
+// request" check, since they prove the router reacted to the scope. Pass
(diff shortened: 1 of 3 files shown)


test/lib/router-act.ts
```diff
@@ -354,17 +380,26 @@ export function createRouterAct(
...
+          let timerId: ReturnType<typeof setTimeout>
+          const onExpiry = () => {
+            // Before timing out, check for App Shell requests in flight
+            // elsewhere on the page (e.g. spawned by a viewport prefetch
+            // during page load, before this `act` scope began, so it was
+            // never intercepted here). The prefetch scheduler's Shell phase
+            // doesn't complete until its shell responses arrive, so the
+            // first observable request may be legitimately gated on one.
+            // Keep the watchdog alive until it settles.
+            if (inFlightAppShellRequests.size > 0) {
```
unstubbable, line 406: Should the set be cleared in `hardNavigationHandler` to avoid a hang here?

unstubbable (approved):

PR #96029, opened 2026-07-21
Add failing test for allow-runtime routes under the Instant Navigation lock

gaojude:
Adds a failing test demonstrating a regression from #95415: a route that opts into runtime prefetching (`export const prefetch = 'allow-runtime'`) can no longer commit a navigation from a full link prefetch while the Instant Navigation lock is held.

The new fixture page is a copy of `full-prefetch-target` plus that single export, and the new test is a copy of the sibling full-prefetch test pointed at it. The sibling passes and the new test times out; under the lock the navigation commits nothing at all: not the prefetched content, not `loading.tsx`, and the source page stays on screen.

```
✓ renders full prefetch content instantly when prefetch={true} (122 ms)
✕ renders full prefetch content instantly when the target also allows runtime prefetching (20093 ms)
```

Reproduces in both dev and start modes. Bisected against published packages in an external repro: passes through `16.3.0-canary.87`, fails from `16.3.0-canary.88` (the release containing #95415), under webpack as well as Turbopack. Request logs there show the router re-issuing prefetch requests it had already completed (identical `_rsc` cache-bust hashes) and then waiting on the dynamic request the lock withholds. v0 hit this on its canary bump and skipped the affected app test until this is fixed (vercel/v0#27128).

### DIFF
--- a/test/e2e/app-dir/instant-navigation-testing-api/fixtures/default/app/full-prefetch-runtime-target/page.tsx
+++ b/test/e2e/app-dir/instant-navigation-testing-api/fixtures/default/app/full-prefetch-runtime-target/page.tsx
@@ -1,0 +1,20 @@
+import { connection } from 'next/server'
+
+// Identical to full-prefetch-target except for the added runtime-prefetch
+// opt-in. A full link prefetch (prefetch={true}) should still resolve the
+// dynamic content ahead of navigation, exactly as it does for the sibling
+// page without this export.
+export const prefetch = 'allow-runtime'
+
+export default async function FullPrefetchRuntimeTargetPage() {
+  await connection()
+
+  return (
+    <div>
+      <h1>Full Prefetch Runtime Target</h1>
+      <div data-testid="full-prefetch-runtime-content">
+        Full prefetch runtime content loaded
+      </div>
+    </div>
+  )
+}
--- a/test/e2e/app-dir/instant-navigation-testing-api/fixtures/default/app/full-prefetch-runtime-target/loading.tsx
+++ b/test/e2e/app-dir/instant-navigation-testing-api/fixtures/default/app/full-prefetch-runtime-target/loading.tsx
@@ -1,0 +1,7 @@
+export default function Loading() {
+  return (
+    <div data-testid="full-prefetch-runtime-loading">
+      Loading full prefetch runtime target...
+    </div>
+  )
+}
--- a/test/e2e/app-dir/instant-navigation-testing-api/fixtures/default/app/page.tsx
+++ b/test/e2e/app-dir/instant-navigation-testing-api/fixtures/default/app/page.tsx
@@ -29,6 +29,15 @@ export default function HomePage() {
           </Link>
         </li>
         <li>
+          <Link
+            href="/full-prefetch-runtime-target"
+            prefetch={true}
+            id="link-to-full-prefetch-runtime"
+          >
+            Go to full prefetch runtime target
+          </Link>
+        </li>
+        <li>
           <Link href="/cookies-page" id="link-to-cookies-page">
             Go to cookies page
           </Link>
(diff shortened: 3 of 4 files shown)


gaojude:
Traced the mechanism by instrumenting the published canary.92 client in the external repro (dev mode, console checkpoints). Chain:

1. #95415 widened `PrefetchHint.SubtreeHasPartialPrefetching` to include `prefetch: 'allow-runtime'` segments (`app-router-types.ts`; previously only truthy `instant` configs set it).
2. For hinted routes the scheduler ignores the link's Full strategy (`scheduler.ts`, "a 'full' prefetch (<Link prefetch={true}>) has no effect. You're meant to use Runtime Prefetching instead"). That downgrade now applies to allow-runtime routes, whose pattern relies on the full link prefetch. Instrumentation confirms: `strategy downgraded to PPR from 4` on every pass.
3. The downgraded flow spawns a `PPRRuntime` request. With Cached Navigations metadata present (auto-enabled by `cacheComponents`), `fetchSegmentPrefetchesUsingDynamicRequest` calls `resolveShellStageData`, which does `const shellByteLength = await flightResponse.a`. On this runtime-prefetch response the `a` row exists as a promise but never resolves, so response processing hangs forever and the spawned entries never fulfill or reject. Checkpoint log ends at `shellStage: a is a promise, awaiting...`; the sibling RuntimeShell request resolves `null` and completes normally.
4. Under the Instant Navigation lock, `ensurePrefetchThenNavigate` awaits every spawned entry before navigating, so the navigation never starts and nothing commits, not even `loading.tsx`. The lock-prefetch refcount sticks at 2, exactly the two PPRRuntime entries. Without the lock the navigation falls back to its own dynamic request, so production loses the prefetched instant nav silently instead of hanging.

So there are two layers: an intended semantic shift (full prefetch superseded on partial-prefetching routes) and a bug in the replacement path (the unresolved `flightResponse.a` await deadlocks runtime-prefetch response processing for these routes; the entries also stay pending in the cache indefinitely). Config ablations that do not help: `experimental.partialPrefetching: true`, disabling `cachedNavigations`, `instant: true`, and `instant: { unstable_samples: [{}] }` on the page.

gaojude:
Production impact without the lock, same repro, measured by counting dynamic RSC requests (rsc header, no prefetch header) between the link click and the marker paint after the prefetch settled: canary.85 paints with 0 dynamic requests (served from the prefetched cache), canary.92 paints only after 1 fresh dynamic request. So outside the testing API the page still renders and nothing errors, but the navigation is no longer instant: the full link prefetch on an allow-runtime route no longer produces data the navigation can use. The instant() hang is the loud version of this silent regression.

test/e2e/app-dir/instant-navigation-testing-api/fixtures/default/app/full-prefetch-runtime-target/page.tsx
```diff
@@ -0,0 +1,20 @@
+import { connection } from 'next/server'
+
+// Identical to full-prefetch-target except for the added runtime-prefetch
+// opt-in. A full link prefetch (prefetch={true}) should still resolve the
+// dynamic content ahead of navigation, exactly as it does for the sibling
+// page without this export.
+export const prefetch = 'allow-runtime'
+
+export default async function FullPrefetchRuntimeTargetPage() {
+  await connection()
```
acdlite, line 10: This is a dynamic component. It's intentionally not cacheable in the new model. If you do a dynamic fetch, you must mark it as cacheable with "use cache".

"allow-runtime" does not make dynamic things cacheable, it just permits us to invoke the server during a prefetch as opposed to only fetching from the CDN cache.

acdlite (changes requested):
Left a comment. The behavior described in the fixture is correct: dynamic content cannot be prefetched under the new Partial Prefetching model. You must mark it with `"use cache"`.

PR #96035, opened 2026-07-22
Turbopack: Use `swc_core::ecma::utils::prop_name_eq` for a couple of the `next-custom-transforms`

bgw:
Found this helper while looking through one of @sampoder's PRs, noticed that these places could benefit from it.


sampoder (approved):
lgtm! nice

Issue #95894, opened 2026-07-17
OpenTelemetry HttpInstrumentation set up via instrumentation.ts never instruments incoming requests — http is loaded before register() and no later require passes the hook


r34son:
### Link to the code that reproduces this issue

https://github.com/r34son/next-otel-http-instrumentation-repro

### To Reproduce

1. npm install && npm run build && npm run start (standalone output, node .next/standalone/server.js)
2. curl -sI http://localhost:3000/ | grep -i x-otel-trace-id

The repro uses the OpenTelemetry setup from the [Next.js OpenTelemetry guide (manual configuration)](https://nextjs.org/docs/app/guides/open-telemetry#manual-opentelemetry-configuration) — NodeSDK initialized from instrumentation.ts — plus @opentelemetry/instrumentation-http@0.219.0 with a responseHook that sets an x-otel-trace-id response header, used as an externally observable marker of whether the instrumentation handles incoming requests.

Control experiment in the same repo: npm run start:preload loads the exact same SDK setup via NODE_OPTIONS="--require ./otel-preload.cjs" instead of instrumentation.ts.

### Current vs. Expected behavior

**Current:**

| Setup | `x-otel-trace-id` header | HTTP server spans |
|---|---|---|
| via `instrumentation.ts`, next@16.2.10 (Turbopack build) | 0/3 requests | none — every exported span has `instrumentationScope: next.js` |
| via `instrumentation.ts`, next@15.5.20 (webpack build, same files) | 0/3 requests | none |
| same code via `NODE_OPTIONS="--require"`, either version | 3/3 requests | present |

The failure is completely silent: the SDK starts fine (`[repro] OTel NodeSDK started` is printed), Next's built-in `next.js`-scoped spans flow, and `trace.getSpanContext(context.active())` returns valid contexts during SSR — which masks the fact that `@opentelemetry/instrumentation-http` never attached. There is no warning (the `"Module http has been loaded before…"` diag warning is only visible if a diag logger is configured).

**Expected:** instrumentations registered in `instrumentation.ts` `register()` — the place the OpenTelemetry guide recommends — instrument incoming HTTP requests, or at least the limitation is documented in the OpenTelemetry guide.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin 25.5.0
  Available memory (MB): 32768
  Available CPU cores: 12
Binaries:
  Node: 22.18.0
  npm: 10.9.3
Relevant Packages:
[...]

PR #96047, opened 2026-07-22
fix(instrumentation): re-trigger require hooks for builtins after reg…

tabrejansary:
…ister (#95894)


.husky/pre-commit
```diff
@@ -1 +1,2 @@
+export PATH="$PATH:/c/Users/Tabrej ansari/AppData/Roaming/npm"
```
Sanjays2402: this looks like it got committed by accident — a hardcoded local npm path with your username. drop it from the PR.

PR #96054, opened 2026-07-22
Replace vendored `http-proxy` with `httpxy`

eps1lon:
[`httpxy`](https://github.com/unjs/httpxy) is a faster and actively maintained variant of `http-proxy`. It also doesn't need any manual patches to avoid security vulnerabilities in Next.js.

`httpxy` is used by Nuxt and maintained by the Nitro team.


eps1lon:
Replaced by https://github.com/vercel/next.js/pull/96060. httpxy revealed some bugs in the old version of msw we vendor so we had to update msw first.

PR #96060, opened 2026-07-22
Replace vendored `http-proxy` with `httpxy`

eps1lon:
[`httpxy`](https://github.com/unjs/httpxy) is a faster and actively maintained variant of `http-proxy`. It also doesn't need any manual patches to avoid security vulnerabilities in Next.js.

`httpxy` is used by Nuxt and maintained by the Nitro team.


packages/next/src/server/lib/router-utils/proxy-request.ts
```diff
@@ -108,7 +107,9 @@ export async function proxyRequest(
         }
       })
     })
-    proxy.ws(req, res, upgradeHead)
+    // The returned promise rejects on errors of the client socket (e.g. an
+    // abrupt client disconnect) even when an `error` listener is registered.
+    proxy.ws(req, res as Socket, {}, upgradeHead).catch(onProxyError)
```
eps1lon, line 116: The cast doesn't introduce new unsoundness. The `http-proxy` types just hid it because they typed `socket: any`.

gnoff (approved):

PR #96069, opened 2026-07-22
navigation(): Omit content from runtime prefetches

acdlite:
Adds a new server API, `await navigation()`. (Currently prefixed as `unstable_navigation`). Its purpose is to save on prefetching costs by preventing content from being rendered at runtime, except during an actual navigation (or during build/ISR). Unlike `await connection()`, it does not prevent the content from being cacheable.

For example, when prefetching a product page, you might want to include only the above-the-fold content.

```ts
function ProductPage() {
  return (
    <>
      <ItemHeader />
      <Suspense fallback={<Loading />}>
        <FullItemContent />
      </Suspense>
    </>
  )
}

async function FullItemContent() {
  await navigation()
  ...
}
```

In this example, `<ItemHeader>` will be prefetched, but not `<FullItemContent>`.

Calling `navigation()` inside "use cache" or "use cache: private" is an error for now. The supported pattern is to await `navigation()` in an uncached wrapper and put the "use cache" directive on an inner function below the gate. Direct support inside cache scopes — omitting the entire cache entry from runtime prefetches — will be added in a subsequent PR.

### DIFF
--- a/packages/next/src/server/request/stages.ts
+++ b/packages/next/src/server/request/stages.ts
@@ -1,0 +1,135 @@
+import { workAsyncStorage } from '../app-render/work-async-storage.external'
+import {
+  throwForMissingRequestStore,
+  workUnitAsyncStorage,
+} from '../app-render/work-unit-async-storage.external'
+import {
+  applyOwnerStack,
+  makeStageHangingPromise,
+} from '../dynamic-rendering-utils'
+import { isRequestApiAllowedInCurrentPhase } from './utils'
+import { InvariantError } from '../../shared/lib/invariant-error'
+
+/**
+ * This function allows you to indicate that the subsequent code should be
+ * deferred to the actual navigation instead of rendering during a runtime
+ * prefetch. Runtime prefetches are rendered per-user, per-link, so deferring
+ * content below `await unstable_navigation()` saves that per-request
+ * rendering cost.
+ *
+ * It has no effect during static prerendering — static output is computed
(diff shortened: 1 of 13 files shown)


packages/next/src/server/request/stages.ts
```diff
@@ -0,0 +1,127 @@
...
+          // unstable_navigation() relies on the staged rendering model, which
+          // only exists with Cache Components.
+          throw new Error(
+            `Route ${workStore.route} used \`unstable_navigation()\`, which requires Cache Components to be enabled. Learn more: https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheComponents`
+          )
+        case 'request':
+          // Invariant: an actual navigation or initial load never defers
+          // content — in dev or prod. unstable_navigation() only defers
+          // content out of runtime prefetches.
+          return Promise.resolve(undefined)
```
lubieowoce, line 123: this needs a TODO to make this work in instant validation (i can work on that)

packages/next/stages.js
```diff
@@ -0,0 +1,26 @@
+let stagesExports
+
+if (process.env.NEXT_RUNTIME === '') {
+  const notAvailableInClient = (name) => {
+    return function notAvailable() {
+      throw new Error(`\`${name}\` is only available in a Server Component.`)
+    }
+  }
```
lubieowoce: this should also be a compile time error, should be added it to `invalid_client_imports` in `react_server_components.rs`

PR #96071, opened 2026-07-22
turbopack: serve dev source content on demand (experimental)

lukesandberg:
## Summary

Adds an experimental Turbopack **dev-only** option, `experimental.turbopackServeSourceContent`, that stops inlining the full text of source files into browser source maps and instead serves that content on demand.

Today, `next dev` (Turbopack) inlines every source file's contents into `sourcesContent` for each generated `.js.map`. This bloats the maps shipped to the browser and costs server work copying bytes around, even though the browser only needs a handful of files when you actually open DevTools.

With the flag enabled:

- Browser source maps omit `sourcesContent` for first-party project files and set `sourceRoot` to an on-demand dev endpoint (`/__nextjs_source-content/[project]/`) with clean project-relative `sources`. Chrome DevTools shows a tidy Sources tree and fetches each file's content lazily. URLs are host-relative, so this works both on `localhost` and in a remote dev environment (e.g. a sandbox).
- `node_modules` and framework-internal (`[next]`/`[turbopack]`) sources keep their existing URIs and inlined content, so they stay ignore-listed and traceable.
- Server-runtime code frames read `file://` sources from disk when a map omits content, so terminal stack traces still render.

Security: reads through the content endpoint are gated by an admission filter of paths that a source map actually referenced, on top of Turbopack's existing filesystem-root sandbox — no weaker than today, and path traversal / non-project reads are rejected.

Also fuses the error-overlay code-frame path into a single native call (`project.getCodeFrameForAsset`) so the (potentially large) source string is read and rendered entirely in Rust instead of being laundered through JavaScript only to be handed back to the native renderer.

The flag defaults to **off**; behavior is unchanged unless opted in.

Incidentally renames the `SOURCE_URL_PROTOCOL` constant to `SOURCE_URL_PROTOCOL_STR`.

## Verification

- New dev test `test/development/app-dir/source-content-on-demand/` (Turbopack): asserts maps omit `sourcesContent` + set `sourceRoot`, the content endpoint serves referenced files, unreferenced files and path-traversal are rejected, and error frames trace back to project source.
- `pnpm test-dev-turbo test/development/app-dir/error-overlay/` — passes with the flag both off and on (verified no `file://` frames leak and framework frames stay ignore-listed).
- `pnpm test-dev-turbo test/development/acceptance-app/ReactRefreshLogBox.test.ts` — used to confirm the default (flag-off) redbox behavior is unchanged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -318,6 +318,8 @@ petgraph = "0.8.3"
 pin-project-lite = "0.2.17"
 postcard = "1.0.4"
 proc-macro2 = "1.0.79"
+# See https://github.com/vercel/next.js/issues/91708 for why we need the legacy_x86_64_support feature
+qfilter = { version = "=0.3.0-alpha.4", features = ["serde", "legacy_x86_64_support"] }
 qstring = "0.7.2"
 quick_cache = { version = "0.6.14" }
 quote = "1.0.45"
--- a/crates/next-api/src/analyze.rs
+++ b/crates/next-api/src/analyze.rs
@@ -16,7 +16,7 @@ use turbo_tasks_fs::{
 };
 use turbopack_analyze::split_chunk::{split_output_asset_into_parts, split_traced_file_into_parts};
 use turbopack_core::{
-    SOURCE_URL_PROTOCOL,
+    SOURCE_URL_PROTOCOL_STR,
     asset::{Asset, AssetContent},
     chunk::{ChunkingType, TracedMode},
     module::Module,
@@ -410,7 +410,7 @@ pub async fn analyze_output_assets(
 
     let mut builder = AnalyzeDataBuilder::new();
 
-    let prefix = format!("{SOURCE_URL_PROTOCOL}///");
+    let prefix = format!("{SOURCE_URL_PROTOCOL_STR}///");
 
     // Process the output assets and extract chunk parts.
     // Also creates sources for the chunk parts.
--- a/crates/next-api/src/versioned_content_map.rs
+++ b/crates/next-api/src/versioned_content_map.rs
@@ -9,6 +9,7 @@ use turbo_tasks::{
 };
 use turbo_tasks_fs::{FileContent, FileSystemPath};
 use turbopack_core::{
+    PROJECT_FILESYSTEM_NAME_STR, SOURCE_URL_PROTOCOL_STR,
     asset::{Asset, AssetContent},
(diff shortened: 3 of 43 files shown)


lukesandberg:
* **#96071** <a href="https://app.graphite.com/github/pr/vercel/next.js/96071?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/96071?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

Cargo.toml
```diff
@@ -318,6 +318,8 @@ petgraph = "0.8.3"
 pin-project-lite = "0.2.17"
 postcard = "1.0.4"
 proc-macro2 = "1.0.79"
+# See https://github.com/vercel/next.js/issues/91708 for why we need the legacy_x86_64_support feature
+qfilter = { version = "=0.3.0-alpha.4", features = ["serde", "legacy_x86_64_support"] }
```
lukesandberg, line 322: do we still need this?

revert?

crates/next-api/src/analyze.rs
```diff
@@ -410,7 +410,7 @@ pub async fn analyze_output_assets(
 
     let mut builder = AnalyzeDataBuilder::new();
 
-    let prefix = format!("{SOURCE_URL_PROTOCOL}///");
+    let prefix = format!("{SOURCE_URL_PROTOCOL_STR}///");
```
lukesandberg, line 413: now that these are constants.... can this be a `concat!` call?

crates/next-api/src/versioned_content_map.rs
```diff
@@ -267,6 +298,133 @@ type GetEntriesResultT = Vec<(FileSystemPath, ResolvedVc<Box<dyn OutputAsset>>)>
...
+    let assets_ref = assets.await?;
+    let per_asset = assets_ref
+        .iter()
+        .map(|&asset| async move {
+            let Some(generate_source_map) =
+                ResolvedVc::try_sidecast::<Box<dyn GenerateSourceMap>>(asset)
+            else {
+                return Ok(Vec::new());
+            };
+            let map = generate_source_map.generate_source_map().await?;
```
lukesandberg, line 395: it should be possible to access the StructuredSourceMap and walk the sources

Alternatively we could consider accessing the module graph for the route and walking the modules

PR #96076, opened 2026-07-22
<Link prefetch="navigation">

acdlite:
Adds the ability to opt into a more aggressive prefetch that simulates the response you'd get during an actual navigation: `<Link prefetch="navigation">`. When this option is set, `navigation()` no longer blocks content from being prefetched.

This can be set on Links that have a high probability of being clicked, to prefetch more of the content would otherwise be worth the cost or performance trade off. For example, when rendering a list of N links, it's might be too costly to prefetch the entirety of every target page. However, if one of those same links is rendered in an above-the-fold treatment, you could afford to prefetch it more aggressively.


packages/next/src/server/app-render/work-unit-async-storage.external.ts
```diff
@@ -216,6 +216,28 @@ export interface PrerenderStoreModernRuntime
...
+   * The deepest stage this runtime prerender is allowed to advance conceptual
+   * content through. Content belonging to a deeper stage is deferred — e.g.
+   * with a ceiling of `RenderStage.Runtime`, `unstable_navigation()` (which
+   * gates `RenderStage.Dynamic` content) hangs, while a ceiling of
+   * `RenderStage.Dynamic` renders through it. This is data on the store
+   * consulted via `checkAndRecordStageDeferral`; it deliberately does not affect
+   * the StagedRenderingController, which continues to gate dynamic APIs like
+   * `connection()` in all cases.
+   */
+  readonly allowedRenderStage: RenderStage
```
lubieowoce: we have `finalStage` for this now

packages/next/src/server/request/stages.ts
```diff
@@ -92,12 +94,20 @@ export function unstable_navigation(): Promise<void> {
...
-            workStore.route,
-            '`unstable_navigation()`',
-            workUnitStore
-          )
+          //
+          // A navigation-depth runtime prefetch, however, is allowed to
+          // render through the navigation gate (real dynamic APIs like
+          // `connection()` still hang), so we resolve immediately in that
+          // case.
+          if (checkAndRecordStageDeferral(workUnitStore, RenderStage.Dynamic)) {
```
lubieowoce: i don't think i get this, navigation should resolve in `NavigationStatic` or `NavigationRuntime`, why are we referencing `Dynamic` here?

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -1539,28 +1565,39 @@ async function generateRuntimePrefetchResult(
...
         shellByteLengthDeferred: createPromiseWithResolvers(),
+        earliestDeferredRenderStageDeferred: createPromiseWithResolvers(),
       }
 
+  // A navigation-depth prefetch renders through the navigation gate
+  // (`unstable_navigation()` resolves instead of hanging). Every other
+  // runtime prefetch's ceiling is the final stage its mode renders to.
+  const allowedRenderStage: PrerenderStoreModernRuntime['allowedRenderStage'] =
+    isNavigationPrefetch
+      ? RenderStage.Dynamic
```
lubieowoce: same question as in `navigation()`: why `Dynamic`? we don't actually include it in runtime prerenders so i'm confused why we're not using `NavigationRuntime` instead

PR #96079, opened 2026-07-22
[turbopack] Fix deployment skew protection for component chunks

sampoder:
Noticed this bug where we were assuming chunks were plain strings, which may not always be the case given #95261. When putting together https://github.com/react/react/pull/37095, I noticed this was also a spot we were missing so this fixes that. 

Almost all of this diff is a regression test.


lukesandberg (approved):
i think you can just drop the test

PR #90271, opened 2026-02-20
Turbopack: add on-demand memory reporting for dev server

lukesandberg:
## Summary

Adds an on-demand memory reporting system for Turbopack dev servers, enabling developers to inspect memory usage without restarting or attaching external profilers. This implements the design described in PACK-6917.

### Layers

- **Rust core** (`turbo-tasks-backend`): `collect_memory_report()` iterates all in-memory tasks and cells, grouping by function name and value type. Size estimation uses bincode encode-and-discard with a reusable scratch buffer. Allocator stats come from `TurboMalloc::memory_usage()`.

- **NAPI binding**: `Project.getMemoryReport(topN?)` exposes the Rust report as a JSON string to Node.js.

- **HTTP endpoint**: `GET /__nextjs_turbopack-memory` on the dev server. Supports `format=json|markdown` and `top_n` query parameters. The Node.js layer augments the report with process-level stats (RSS, heap, PID, Node version).

- **CLI command**: `next internal turbopack-memory [dir] [--format json|markdown] [--server host:por]` discovers the running dev server via the `.next/dev/lock` file and fetches the report.

- **Documentation**: Added a "Turbopack memory reporting" section to the local development guide.

### Output formats

- **JSON** (default): Machine-readable, includes version, timestamps, task/cell/allocator stats
- **Markdown**: Human-readable tables

### Key design decisions

We estimate memory size via encoding, which is trivial to do but horrifically inaccurate.  This means we both overcount and undercount memory (because persistent formats are sometimes often compact than memory, think slack space in hash tables) and overcount because we don't respect reference counted datastructures like `RcStr`/`Rope`.  A 'correct' approach would be very complex.

## Test plan

- [x] Integration tests for fetch API (JSON schema, markdown, top_n)
- [x] Integration tests for CLI (`next internal turbopack-memory` with JSON, markdown, top_n)
- [x] Manual testing with a real app to verify report accuracy

Fixes PACK-6917

PR #96086, opened 2026-07-22
[turbopack] On-demand memory report for the dev server

lukesandberg:
### What?

Adds an on-demand memory report for the Turbopack dev server. While `next dev` is running you can ask the backend what it is holding in memory, without restarting or attaching an external profiler:

- **HTTP:** `GET /__nextjs_turbopack-memory?format=json|markdown`
- **CLI:** `next internal turbopack-memory [--format json|markdown] [--url <devServerUrl>]`

The report enumerates every cell of every task and groups it by value type, ranked by **`strong_count`** — the number of live `triomphe::Arc` references to the cell's data, a cheap and real proxy for retention. It is split into two sections:

- **Transient** — tasks eviction skips entirely (where per-route/per-request accumulation shows up). For the top types it also includes a sample dependency chain so you can see which computation is holding the data.
- **Persistent** — cells that survived eviction (unevictable cells, or tasks blocked from eviction).

The JSON is additionally augmented with Node process stats (RSS, heap, pid, uptime) by the dev-server layer.

### Why?

We needed a way to diagnose what Turbopack retains after eviction on a live dev server.

This supersedes #90271, which is the original attempt at this feature. That PR estimated per-item memory via bincode encode-and-discard, which both over- and under-counts badly.

### How?

- **Rust (`turbo-tasks-backend`):** `TurboTasksBackend::collect_memory_report()` builds a structured `MemoryReport { transient, persistent }` from `audit_all_cells()`, reusing `debug_trace_transient_task` for the transient sample chains. The previous behavior — dumping a text report to `/tmp` on every eviction cycle — has been removed; the report is now on-demand only.
- **NAPI:** `project_get_memory_report` exposes it as `projectGetMemoryReport` (runs on a blocking thread since it walks all tasks).
- **Dev server:** `getMemoryReportMiddleware` serves the endpoint and renders the markdown; wired into the Turbopack hot-reloader middleware chain.
- **CLI:** `next internal turbopack-memory` discovers the running dev server from the `.next` lock file (or `--url`) and prints the report.
- **Docs + e2e test** added; error `1423` registered for the failure path.

This is a dev-only diagnostic tool, so no telemetry is added.

Closes PACK-6917
Supersedes #90271


lukesandberg:
* **#96086** <a href="https://app.graphite.com/github/pr/vercel/next.js/96086?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/96086?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/server/dev/middleware-turbopack.ts
```diff
@@ -503,6 +503,124 @@ export function getSourceMapMiddleware(project: Project) {
...
+}
+const formatCount = (n: number) => n.toLocaleString('en-US')
+
+/**
+ * Render the strong-count memory audit as markdown: a Transient and a
+ * Persistent section, each with a per-value-type table ranked by
+ * `strongCountSum`, plus dependency-chain samples for the top transient
+ * offenders.
+ */
+function renderMemoryReportMarkdown(report: any): string {
```
lukesandberg: isn't there a type we can use here

scripts/benchmark-memory.js
```diff
@@ -0,0 +1,397 @@
+#!/usr/bin/env node
```
lukesandberg: drop this file from the pr

scripts/visualize-memory.js
```diff
@@ -0,0 +1,406 @@
+#!/usr/bin/env node
```
lukesandberg: drop from the PR

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -194,6 +195,64 @@ impl SnapshotReason {
...
+#[derive(Debug, serde::Serialize)]
+pub struct AuditSection {
+    /// Number of tasks (of this transient/persistent kind) in storage.
+    pub task_count: u64,
+    /// Number of cells contributed by those tasks.
+    pub cell_count: u64,
+    /// Number of distinct value types across those cells.
+    pub distinct_value_types: u64,
+    /// Sum of `strong_count` across all cells in this section.
+    pub total_strong_count_sum: u64,
```
lukesandberg: i don't think this total is useful

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -194,6 +195,64 @@ impl SnapshotReason {
...
+    /// Per-value-type aggregates, ranked by `strong_count_sum` descending.
+    pub by_type: Vec<TypeAudit>,
+    /// Sample tasks for the top-ranked types. For the transient section each
+    /// carries a dependency `chain`; for the persistent section `chain` is
+    /// empty and `task` carries a human-readable description instead.
+    pub sample_chains: Vec<SampleChain>,
+}
+
+/// Per-value-type aggregate within an [`AuditSection`].
+#[derive(Debug, serde::Serialize)]
```
lukesandberg: import Serialize

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1698,6 +1757,129 @@ impl TurboTasksBackend {
...
+        #[derive(Default)]
+        struct TypeAggregate {
+            cells: usize,
+            strong_count_sum: u64,
+            strong_count_max: usize,
+            distinct_tasks: FxHashSet<TaskId>,
+            sample_task: Option<TaskId>,
+        }
+
+        let mut by_type_map: FxHashMap<ValueTypeId, TypeAggregate> = FxHashMap::default();
```
lukesandberg: the type declaration is unneeded

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -194,6 +195,64 @@ impl SnapshotReason {
...
+/// A sampled task for one of the top-ranked types in an [`AuditSection`].
+#[derive(Debug, serde::Serialize)]
+pub struct SampleChain {
+    pub rank: u64,
+    #[serde(rename = "type")]
+    pub type_name: &'static str,
+    /// Debug/description string identifying the sampled task.
+    pub task: String,
+    /// Dependency-chain trace (transient section only; empty otherwise).
+    pub chain: String,
```
lukesandberg: we should drop this

it isn't fit for purpose

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -679,6 +681,76 @@ impl Storage {
...
+    /// Debug-only audit: enumerate every cell of every task (transient and
+    /// persistent) and record `(task_id, value_type_id, strong_count)`.
+    ///
+    /// Transient vs. persistent is recovered from `task_id.is_transient()`
+    /// at aggregation time. Aggregation and formatting live in
+    /// `TurboTasksBackendInner::dump_memory_audit` because the human-readable
+    /// chain trace needs backend-level APIs.
+    ///
+    /// Read-only: takes a read lock per shard.
+    pub fn audit_all_cells(&self) -> AuditRawData {
```
lukesandberg: we should also collect information on tasks with no cells

organized by task type

the problem is that we can end up with hundreds of persistent tasks retained but only a handful of cells.  so we need to report on the tasks with only transient or non-cell data retained (e.g. meta)

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1698,6 +1757,129 @@ impl TurboTasksBackend {
...
+    }
+
+    /// Aggregate one set of cells (transient or persistent) into a ranked
+    /// [`AuditSection`].
+    ///
+    /// If `include_chain` is true, a `DebugTraceTransientTask` trace is attached
+    /// to each top-10 sample — only meaningful for transient tasks. For
+    /// persistent samples the chain is left empty and the sample carries only
+    /// `debug_get_task_description`, which is enough to identify the holder.
+    fn audit_section(
```
lukesandberg: should this logic just be merged with the parallel loop?

lukesandberg: tricky to do given the aggregation logic but worth considering at least if some can be moved

PR #96095, opened 2026-07-23
Attempt static prefetch before resorting to runtime

acdlite:
If we're reasonably confident that a segment can be prefetched statically without omitting data that would have been included during a runtime prefetch (e.g. cookies), the client should attempt prefetch the segment statically instead of going straight to a runtime request.

The decision for whether to do this is based on the ShouldAttemptStaticPrefetch added in previous steps.

If it turns out the static response is not sufficient, then it will fall back to a runtime request.

This makes prefetching cheaper for pages that are fully statically renderable. We can make the optimization better in the future with more reliable per-segment computation of the ShouldAttemptStaticPrefetch, but the current approach should at least work for fully static pages, which is what's most important.

This optimization applies during both the Shell phase and the Speculative phase of the prefetching algorithm.

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -27,6 +27,7 @@ import {
 import {
   createFetch,
   createFromNextReadableStream,
+  decodeBufferedStage,
   resolveShellStageData,
   type RSCResponse,
   type RequestHeaders,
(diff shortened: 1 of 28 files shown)


packages/next/src/client/components/segment-cache/cache.ts
```diff
@@ -2452,18 +2517,151 @@ async function fetchSegmentsOnCacheMissImpl(
...
+  // (positive shell byte offset): decode the buffered bytes a SECOND time,
+  // truncated at the boundary. The truncation is what produces the shell
+  // variant: each segment's param-dependent rows land past the boundary and
+  // decode as still-pending, which renders as the param fallback. It also
+  // rewinds the response's signals — `needsRuntimeRequest` and `isPartial`
+  // fulfillments past the boundary read as pending in this decode, so a
+  // post-shell runtime-data access doesn't mark the shell variant itself as
+  // needing a runtime request.
+  // (An offset that somehow isn't visible in the decode reads as 0 — no
+  // shell — the conservative direction: it can waste a runtime request but
```
unstubbable: This comment is a bit contradictory with: https://github.com/vercel/next.js/blob/cdffd14903e86cf7ae3dc21126c0326ef91327b0/packages/next/src/client/components/segment-cache/cache.ts#L2599-L2600

unstubbable (approved):

packages/next/src/client/components/segment-cache/cache.ts
```diff
@@ -2529,47 +2759,110 @@ function writeSegmentBundleResponse(
...
+    // lets the scheduler decide "would a runtime request return more?" by
+    // comparing tiers alone, with no separate signal to consult.
+    //
+    // Otherwise the content is only as complete as the static tier it was
+    // requested at, so a follow-up runtime request can still supersede it.
+    const recordedFetchStrategy = !needsRuntimeRequest
+      ? payloadFetchStrategy === FetchStrategy.StaticShell
+        ? FetchStrategy.RuntimeShell
+        : FetchStrategy.PPRRuntime
+      : fetchStrategy
```
lubieowoce, line 2789: hmm, this part feels a little weird. having to pretend that we used a different strategy makes me feel like there's some nicer way to model this. although i suppose it makes sense if we think about the strategies as "level of completeness"

lubieowoce (approved):

packages/next/src/client/components/segment-cache/cache.ts
acdlite: yeah my plan is to rename the field something like "completeness" but have it represent both sides

PR #96098, opened 2026-07-23
Remove inert experimental.viewTransition flag

aurorascharff:
## What

Removes `experimental.viewTransition` — the flag is inert. Code only; all docs changes ship in #96097 so the two can be reviewed independently.

## Why

Nothing in the runtime reads it: it appears only in config validation (`config-schema.ts`) and its type/default (`config-shared.ts`). `ViewTransition` ships in the bundled React canary, and every `<Link>` navigation already runs inside `React.startTransition`, so view transitions work with no configuration.

The flag was reserved for Next.js automatically adding transition types to navigations, which never landed (confirmed with the team — a new flag will gate that work when it's done properly). Historically it also switched the bundled React to the experimental channel, but that stopped being necessary when React released `ViewTransition` to canary; the experimental channel is now selected by other flags (`gestureTransition`, `blockingSSR`, `taint`, `transitionIndicator`).

## Changes

- `config-schema.ts` / `config-shared.ts`: schema entry, type + JSDoc, and default removed
- `test/e2e/app-dir/view-transitions` fixture: `next.config.js` deleted (it contained only the flag; the test never references it)

### DIFF
--- a/packages/next/src/server/config-schema.ts
+++ b/packages/next/src/server/config-schema.ts
@@ -336,7 +336,6 @@ export const experimentalSchema = {
   swcTraceProfiling: z.boolean().optional(),
   // NonNullable<webpack.Configuration['experiments']>['buildHttp']
   urlImports: z.any().optional(),
-  viewTransition: z.boolean().optional(),
   workerThreads: z.boolean().optional(),
   webVitalsAttribution: z
     .array(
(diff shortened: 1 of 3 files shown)


eps1lon (commented):
What about https://nextjs.org/docs/app/api-reference/config/next-config-js/viewTransition?

aurorascharff:
> What about https://nextjs.org/docs/app/api-reference/config/next-config-js/viewTransition?

https://github.com/vercel/next.js/pull/96097/changes#diff-48feb912ea21846b01294bd1a486cd9b087ef7f9c633bf4f9ee31b271a38c2c6

Did all the docs changes here @eps1lon

eps1lon (approved):

PR #96103, opened 2026-07-23
Turbopack: Very minor improvements for watcher loop

bgw:
Noticed these things while reading through the code.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/watcher.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/watcher.rs
@@ -508,6 +508,10 @@ impl DiskWatcher {
                         if event.need_rescan() {
                             let _lock = fs_inner.invalidation_lock.blocking_write();
 
+                            // flush the whole mpsc queue, we're about to rescan, we don't need to
+                            // process any other update events that have already happened
+                            while rx.try_recv().is_ok() {}
+
                             if let State::NonRecursive(non_recursive) = &self.state {
                                 // we can't narrow this down to a smaller set of paths: Rescan
                                 // events (at least when tested on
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbo-tasks-fs/src/watcher.rs
```diff
@@ -508,6 +508,10 @@ impl DiskWatcher {
                         if event.need_rescan() {
                             let _lock = fs_inner.invalidation_lock.blocking_write();
 
+                            // flush the whole mpsc queue, we're about to rescan, we don't need to
+                            // process any other update events that have already happened
+                            while rx.try_recv().is_ok() {}
```
bgw, line 513: Minor perf improvement if we're forced to do a full rescan. Might be good because a rescan event already likely indicates that the system is overloaded.

turbopack/crates/turbo-tasks-fs/src/watcher.rs
```diff
@@ -746,7 +750,7 @@ impl BatchedInvalidations {
             EventKind::Modify(ModifyKind::Name(RenameMode::Both)) => {
                 // For the rename::both, notify provides an array of paths
                 // in given order
-                if let [source, destination, ..] = &paths[..] {
+                if let [source, destination] = &paths[..] {
```
bgw, line 753: This event should always be exactly two paths.
https://docs.rs/notify/latest/notify/event/enum.RenameMode.html#variant.Both

> This event should be emitted when both source and target are known. The paths should be provided in this exact order (from, to).

turbopack/crates/turbo-tasks-fs/src/watcher.rs
```diff
@@ -767,7 +771,6 @@ impl BatchedInvalidations {
             // but we also check other RenameModes to cover cases where notify couldn't match the
             // two rename events.
             EventKind::Any | EventKind::Modify(ModifyKind::Any | ModifyKind::Name(..)) => {
-                self.path.extend(paths.clone());
```
bgw, line 770: `path` is already covered by `path_and_children`.

turbopack/crates/turbo-tasks-fs/src/watcher.rs
lukesandberg: thanks, i was wondering about this in the other PR

lukesandberg (approved):

gdoogamal37-prog (commented):
Ok

PR #96107, opened 2026-07-23
Bump postcss to 8.5.23

### LINKED ISSUE #96349: postcss bundled by next is vulnerable to arbitrary file read (GHSA-r28c-9q8g-f849) — please update to >=8.5.18

https://github.com/Tello-Wharton/postcss-sourcemap-repro

### To Reproduce

Create a new Next.js project (latest stable version).
Run npm audit --omit=dev.
Observe 3 high severity advisories referencing postcss <= 8.5.17.

Expected Behavior
next should bundle postcss >= 8.5.18, which contains the fix for GHSA-r28c-9q8g-f849.

Actual Behavior
next ships with a nested postcss version below 8.5.18, causing npm audit to flag it.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 25.9.0
  npm: 11.12.1
  Yarn: N/A
  pnpm: 10.33.0
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Not sure

Other (Deployed)

### Additional context

https://github.com/advisories/GHSA-r28c-9q8g-f849
https://github.com/advisories/GHSA-6g55-p6wh-862q

icyJoseph:
Note that https://github.com/advisories/GHSA-6g55-p6wh-862q does not affect Next.js users unless they build from untrusted source code (which would have more severe security implications). We're merging this to reduce noise from security scanners.

- 8.5.21 is 50h+ old so it can be installed
- 8.5.22 is still under 48h - valid at 2026-07-24 08:48:15 UTC 

Closes: https://github.com/vercel/next.js/issues/96349


eps1lon (approved):
Adjusted PR description to make it clear we're not fixing a vulnerability. We're doing this to reduce GHSA reporting noise.

packages/next/package.json
```diff
@@ -104,7 +104,7 @@
     "@swc/helpers": "0.5.15",
     "baseline-browser-mapping": "^2.9.19",
     "caniuse-lite": "^1.0.30001579",
-    "postcss": "8.5.10",
```
akegvd, line 107: just question why we can't change it to be like `^8.5.21`?
because now lasted release was `8.5.23`

is it because supply chain attack?

package.json
```diff
@@ -309,6 +309,7 @@
   "packageManager": "pnpm@10.33.0",
   "pnpm": {
     "overrides": {
+      "postcss": "8.5.21",
```
akegvd: why we need this?

packages/next/package.json
lorand-horvath: The version is pinned to a certain tested version, such as 8.5.21. Developers can't know right now whether a future 8.x version will or will not be compatible with the rest of the framework.
If you check, it's the same situation with the version of `next` and `eslint-config-next` being fixed/pinned down to a specific version in a template generated with create-next-app.

haimuhaimu (commented):
Blocking finding: the PostCSS bump is not self-contained yet. On the current head, the `types and precompiled / build` check runs `pnpm types-and-precompiled`, detects uncommitted output in 10 checked-in compiled CSS/PostCSS bundles, and exits 1. The affected paths include `packages/next/src/compiled/cssnano-simple/index.js`, `postcss-preset-env/index.cjs`, `postcss-safe-parser/safe-parse.js`, and seven other compiled PostCSS helpers listed in the job log.

Because these compiled files are part of the published Next.js package, leaving them generated from the previous dependency graph would make the declared PostCSS 8.5.23 baseline differ from the shipped precompiled output. Please run `pnpm types-and-precompiled` on this head and commit the resulting 10 files (or otherwise update the checked-in compiled artifacts so the reproducibility check is clean).

CI evidence: https://github.com/vercel/next.js/actions/runs/30295024102/job/90074269616

haimuhaimu (commented):
Follow-up review on `abf23bf52b11244255c778deaa636b290e8992a6`: the previously missing precompiled output is now committed. I verified that the final file list contains the same 10 compiled CSS/PostCSS bundles reported by `pnpm types-and-precompiled`, including `cssnano-simple`, `postcss-preset-env`, `postcss-safe-parser`, and the seven other helpers. The current check rollup is 114 passed, 0 failed, and 15 skipped, and the test report says all tests passed, so my earlier blocking finding is resolved.

The PR is currently conflicting with the moving `canary` branch; after resolving that conflict, please rerun the precompiled reproducibility check to ensure the generated bundles still match the rebased dependency graph. No further issue found in the PostCSS bump itself.

terrymun:
Something worth noting: postcss 8.5.24 seems to cause Next.js build failures. I've reported it to the postcss repo but they pointed out it might be a downstream issue with lightningcss or turbopack.

https://github.com/postcss/postcss/issues/2122

> [CSS Syntax Level 3 §3.2 "Determine the fallback encoding"](https://www.w3.org/TR/css-syntax-3/#input-byte-stream)
> specifies that a leading UTF-8 BOM is to be consumed as an encoding signal, not treated as content. A
> stylesheet beginning with `EF BB BF` is valid CSS, so failing to parse it is a spec-conformance gap. Stripping
> a leading `U+FEFF` before tokenization would fix this class of failure regardless of what produced the BOM.
> 
> This also matters beyond our setup: **any** `compressed` dart-sass output containing non-ASCII will carry a
> BOM, so this is reachable by any Sass + Lightning CSS pipeline, with or without PostCSS in between.

eps1lon (approved):

PR #96115, opened 2026-07-23
fix: preserve CSS shorthand declaration order

tylergibbs1:


tylergibbs1:
Rebased onto current `canary` and regenerated the `cssnano-simple` bundle to resolve the generated-artifact conflict. Verified with `pnpm build-all`, the two focused cssnano Jest files (8 tests passed), and Prettier checks on the source and regression test. The PR is mergeable again; Vercel Agent Review is still running on the new head.

PR #96116, opened 2026-07-24
Turbopack: More aggressively debounce filesystem watch events if we detected changes to node_modules

bgw:
Previously, we were debouncing update by sleeping 1ms at a time on macos and windows, and 10ms at a time on Linux.

During a slow `pnpm install`, or a `git checkout`, this could cause us to do a bunch of extra throwaway work.

Changes:
- Increase the debounce interval to a consistent 10ms everywhere. This should still be small enough that it's not noticable on macos or windows.
- If an event touches `node_modules`, there's a good chance that a package manager is running and many other files will be modified, so extend the batch deadline by 200ms instead of 10ms.
- Because there's a chance that the batch deadline could get extended indefinitely (this was always possible, just more likely now) include a compilation event that gets logged after 5 seconds.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
@@ -118,7 +118,7 @@ impl PendingBatch {
         let _guard = fs.tokio_handle().enter();
         if let Some(turbo_tasks) = fs.turbo_tasks() {
             turbo_tasks.send_compilation_event(Arc::new(FilesystemSettlingEvent {
-                elapsed_secs: (now - self.started).as_secs(),
+                elapsed_secs: (now - self.started).as_secs_f64(),
             }));
         }
         self.event_interval = self.event_interval.saturating_mul(2).min(max_event_delay);
@@ -136,7 +136,7 @@ impl PendingBatch {
 #[derive(Debug, Clone, Serialize)]
 pub struct FilesystemSettlingEvent {
     /// How long the current batch has been held open, in seconds.
-    pub elapsed_secs: u64,
+    pub elapsed_secs: f64,
 }
 
 impl CompilationEvent for FilesystemSettlingEvent {
@@ -151,7 +151,7 @@ impl CompilationEvent for FilesystemSettlingEvent {
     fn message(&self) -> String {
         format!(
             "Turbopack has seen frequent file updates and is waiting for the filesystem to settle \
-             ({}s elapsed).",
+             ({:.1}s elapsed so far).",
             self.elapsed_secs
         )
     }
--- a/turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
@@ -1,0 +1,171 @@
+use std::{
+    sync::{
+        Arc,
+        mpsc::{Receiver, RecvTimeoutError},
+    },
+    time::{Duration, Instant},
+};
+
+use serde::Serialize;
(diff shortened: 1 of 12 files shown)


turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+}
+
+/// Emitted when frequent filesystem updates cause us to keep a batch open for an extended period of
+/// time. Informing the user when this happens may help them understand what's happening, and that
+/// Turbopack is not stalled.
+#[derive(Debug, Clone, Serialize)]
+pub struct FilesystemSettlingEvent {
+    /// How long the current batch has been held open, in seconds.
+    pub elapsed_secs: u64,
```
lukesandberg: should this be milliseconds?  rounding up to seconds seems like a lot

turbopack/crates/turbo-tasks-fs/src/watcher/mod.rs
```diff
@@ -613,13 +648,23 @@ impl DiskWatcher {
...
-                        // `BATCH_DELAY`.
+                        // Any event that contributes to the batch keeps it open for another
+                        // `batch_delay`. A path matching `extended_batch_delay_matcher` (e.g. a
+                        // package-manager install target) keeps it open for
+                        // `extended_batch_delay_duration` instead.
+                        let mut delay = config.batch_delay;
+                        if let Some(matcher) = &extended_batch_delay_matcher
+                            && event.paths.iter().any(|path| matcher.match_path(path))
+                        {
+                            delay = delay.max(config.extended_batch_delay_duration);
```
lukesandberg, line 667: nit: we should reject configs with an assert/debug_assert where the extended delay is not >= the regular delay

turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    fn severity(&self) -> Severity {
+        Severity::Info
+    }
+
+    fn message(&self) -> String {
+        format!(
+            "Turbopack has seen frequent file updates and is waiting for the filesystem to settle \
+             ({}s elapsed).",
```
lukesandberg: maybe 5s elapsed so far?

lukesandberg: i wonder if we could use this to drive a Spinner on the js side?

bgw: In theory we could, but doing that well would require a lot of work on the compiler events side, and we don't do a good job of locking stdout/stderr on the JS side (lots of stuff write directly to these) so trying to drive an animation is going to have some janky behavior if (e.g.) a webpack loader decides to write something to stdout while we're rending the spinner.

turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
bgw: I didn't want to deal with rendering of fractional seconds, and we only show you this after 5 seconds.

IRL discussion: Display as fractional seconds and reduce the reporting threshold to `500ms`.

turbopack/crates/turbo-tasks-fs/src/watcher/batch_schedule.rs
lukesandberg: yeah agreed, risky and not so beneficial

lukesandberg (approved):
discussed IRL: add the most recently changed file to the message to aid debugging

bgw:
Follow-up: Include the last-modified file in the warning message.

PR #96117, opened 2026-07-24
fix: Uncontrolled command line commandString Use non-shell execution

dev-tmg:
Code that passes untrusted user input directly to `child_process.exec` or similar APIs that execute shell commands allows the user to execute malicious code. fixes avoid `exec(commandString)`/`bash -c` for operations containing variable input. Use non-shell execution (`spawn`/`execFile`) with argument arrays and `cwd`, so user-controlled values are not interpreted by a shell.

Best fix here without changing functionality:
- In `.github/actions/next-stats-action/src/prepare/repo-setup.js`, replace the vulnerable `exec(...)` usages in:
  - `cloneRepo` (line region around current `git clone ...`)
  - `getCommitId`
  - `resetToRef`
  - `mergeBranch`
- Use `exec.spawnPromise` with:
  - command: `'git'`
  - args array: exact git args (e.g., `['clone', url, '--single-branch', '--branch', branch, ...]`)
  - options: `{ cwd: <dir when needed>, stdio: 'pipe' }`
- For `getCommitId`, use `exec.spawn` (not promise wrapper) with `stdio: ['ignore', 'pipe', 'pipe']`, collect stdout, await exit code, and return trimmed hash.

### DIFF
--- a/.github/actions/next-stats-action/src/prepare/repo-setup.js
+++ b/.github/actions/next-stats-action/src/prepare/repo-setup.js
@@ -8,9 +8,18 @@ module.exports = (actionInfo) => {
   return {
     async cloneRepo(repoPath = '', dest = '', branch = '', depth = '20') {
       await fs.promises.rm(dest, { recursive: true, force: true })
-      await exec(
-        `git clone ${actionInfo.gitRoot}${repoPath} --single-branch --branch ${branch} --depth=${depth} ${dest}`
-      )
+      await exec.spawnPromise('git', {
+        args: [
+          'clone',
+          `${actionInfo.gitRoot}${repoPath}`,
+          '--single-branch',
+          '--branch',
+          branch,
+          `--depth=${depth}`,
+          dest,
+        ],
+        stdio: 'pipe',
+      })
     },
     async getLastStable() {
       const res = await fetch(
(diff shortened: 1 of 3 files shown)


examples/cms-agilitycms/lib/preview.ts
```diff
@@ -87,6 +87,13 @@ export async function validatePreview({ agilityPreviewKey, slug, contentID }) {
     };
   }
 
+  if (typeof agilityPreviewKey !== "string") {
+    return {
+      error: true,
+      message: `Invalid agilitypreviewkey.`,
+    };
+  }
+
```
dev-tmg, line 96: fix is to enforce runtime type checks for all user-controlled query parameters before applying string operations or relying on string comparison behavior. For this specific issue, the best minimal fix is in `examples/cms-agilitycms/lib/preview.ts` inside `validatePreview`: reject requests unless `agilityPreviewKey` is a string, and only then run `.includes/.split` and equality checks. This preserves current functionality for valid inputs and safely rejects tampered array inputs.

Concretely:
- In `validatePreview`, replace the current `if (!agilityPreviewKey)` block with a stricter check:
  - reject when missing
  - reject when `typeof agilityPreviewKey !== "string"`
- Keep existing normalization (`space` → `+`) and comparison logic unchanged after type check.

turbopack/packages/devlow-bench/src/shell.ts
```diff
@@ -191,7 +191,6 @@ export function command(
   } = {}
 ): Command {
   const process = spawn(command, args, {
-    shell: true,
```
dev-tmg, line 194: fix is to stop invoking a shell for argument-based command execution. In Node.js, use `spawn(command, args, { shell: false, ... })` (or omit `shell` since false is default). This preserves functionality for normal executable + argv execution while preventing shell metacharacter interpretation of caller-provided `args`.

In `turbopack/packages/devlow-bench/src/shell.ts`, update the `spawn` options in `command(...)` so `shell` is not enabled. Keep `args` as an array and pass through existing `env`, `cwd`, and `stdio` unchanged

.github/actions/next-stats-action/src/prepare/repo-setup.js
```diff
@@ -8,9 +8,18 @@ module.exports = (actionInfo) => {
   return {
     async cloneRepo(repoPath = '', dest = '', branch = '', depth = '20') {
       await fs.promises.rm(dest, { recursive: true, force: true })
-      await exec(
-        `git clone ${actionInfo.gitRoot}${repoPath} --single-branch --branch ${branch} --depth=${depth} ${dest}`
-      )
+      await exec.spawnPromise('git', {
```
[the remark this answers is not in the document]
dev-tmg: `exec(commandString)`/`bash -c` for operations containing variable input. Use non-shell execution (`spawn`/`execFile`) with argument arrays and `cwd`, so user-controlled values are not interpreted by a shell.

- In `.github/actions/next-stats-action/src/prepare/repo-setup.js`, replace the vulnerable `exec(...)` usages in:
  - `cloneRepo` (line region around current `git clone ...`)
  - `getCommitId`
  - `resetToRef`
  - `mergeBranch`
- Use `exec.spawnPromise` with:
  - command: `'git'`
  - args array: exact git args (e.g., `['clone', url, '--single-branch', '--branch', branch, ...]`)
  - options: `{ cwd: <dir when needed>, stdio: 'pipe' }`
- For `getCommitId`, use `exec.spawn` (not promise wrapper) with `stdio: ['ignore', 'pipe', 'pipe']`, collect stdout, await exit code, and return trimmed hash.

This preserves behavior (same git operations), removes shell parsing, and closes injection vectors from `branch`, `ref`, and directory values.

PR #96120, opened 2026-07-24
Optimize implicit cache tag derivation

marcoshernanz:
## What

Derive implicit cache tags by scanning pathname slash boundaries incrementally instead of repeatedly rebuilding prefixes with `split`, `slice`, and `join`.

The new loop emits exactly the same derived-tag array as the old helper, including the full-path prefix. The caller's existing `Set` continues to deduplicate that tag against `pathname` when they are equal. This PR's optimization is specifically the removal of repeated intermediate path arrays and prefix joins; it does not rely on changing deduplication semantics or tag order.

Coverage includes root paths, `page` and `route` suffixes, missing leading slashes, repeated/trailing slashes, Unicode, malformed UTF-16, and already-encoded pathnames.

## Correctness

The candidate was compared element-by-element with the original `split('/').slice(0, i).join('/')` implementation over **1,713,653 paths** on each of Node 20.9.0, Node 22.23.1, and Node 24.14.1:

- 349,525 exhaustive strings of length 0–9 over a slash/ASCII alphabet;
- every UTF-16 code unit: 65,536 cases;
- every valid surrogate pair: 1,048,576 cases;
- 250,000 deterministic randomized paths with Unicode, malformed surrogates, controls, reserved characters, suffix-like content, and arbitrary slash placement;
- 16 targeted cases, including a one-million-character segment and a 1,024-segment path.

All three runtime runs completed with **0 mismatches**.

## Performance

Pinned CPU 4 on a 16-vCPU / 32-GiB Vercel DevBox. Each of 35 scenarios used 31 interleaved rounds, adaptive iterations, and two identical baseline lanes (A/A control). Scenarios cover depth 1–256, slash-dense paths, segment widths 1–512, suffix cases, no-leading-slash inputs, Unicode/malformed code units, and seeded shallow/medium/deep corpora.

| Runtime | Scenarios improved | Seeded shallow | Seeded medium | Seeded deep |
| Node 20.9.0 / V8 11.3 | 35/35 | **-77.11%** | **-82.59%** | **-92.69%** |
| Node 22.23.1 / V8 12.4 | 35/35 | **-81.07%** | **-88.60%** | **-95.47%** |
| Node 24.14.1 / V8 13.6 | 35/35 | **-80.98%** | **-89.00%** | **-95.88%** |

No candidate p05–p95 interval crosses zero on any runtime. Selected Node 24 results:

| Scenario | Median delta | p05–p95 |
| --- | ---: | ---: |
| Realistic root/suffix corpus | **-77.70%** | -78.02% to -77.22% |
| Repeated/trailing slashes | **-77.88%** | -79.24% to -77.07% |
| Unicode and malformed UTF-16 | **-63.49%** | -64.59% to -62.31% |
| Depth 8, 4-char segments | **-83.14%** | -83.98% to -82.74% |
| Depth 32, 4-char segments | **-92.48%** | -92.60% to -92.22% |
| Depth 256, 4-char segments | **-98.77%** | -98.80% to -98.75% |
| Depth 32, 512-char segments | **-97.02%** | -97.09% to -96.98% |

Inspector heap-allocation sampling, five repeats (median sampled bytes):

| Corpus | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| Realistic, 500k calls | 489,583,800 | 206,253,144 | **-57.87%** |
| Shallow, 500k calls | 493,465,152 | 203,994,592 | **-58.66%** |
| Deep, 20k calls | 843,885,312 | 112,216,736 | **-86.70%** |

### Alternative tested

A split-once implementation that accumulated prefix strings also passed all 1.71M cases, but a direct 35-scenario interleaved comparison found the shipped `indexOf` implementation faster in **35/35** cases. Accumulation was 26.49% to 1,810.86% slower and sampled 47.5%/70.3%/280.8% more bytes on realistic/shallow/deep corpora. It was rejected.

### Production profile

The original matched render-pipeline profile used 1,000 serial plus 10,000 concurrent-load requests:

| Attribution | Baseline | Candidate |
| --- | ---: | ---: |
| Total sampled CPU | 20,534.6 ms | 20,461.6 ms |
| Next runtime sampled CPU | 7,786.1 ms | 7,856.8 ms |
| `implicit-tags.ts` | 54.1 ms (0.69%) | 41.4 ms (0.53%) |

That is a **23.48% reduction** in source-attributed production CPU. A single load-throughput run improved 3.64%, but that is inside the workload's established 6.76% full range, so this PR does not claim a separately resolvable total-throughput improvement.

## Validation

- `pnpm exec jest packages/next/src/server/lib/implicit-tags.test.ts --runInBand` — 12/12 passed
- `pnpm prettier --check packages/next/src/server/lib/implicit-tags.ts packages/next/src/server/lib/implicit-tags.test.ts`
- `pnpm eslint packages/next/src/server/lib/implicit-tags.ts packages/next/src/server/lib/implicit-tags.test.ts`
- `pnpm --filter next types`
- `pnpm --filter next build`
- `pnpm build-all` — 18/18 tasks successful, including `@vercel/turbopack-next` checks
- Correctly scoped Codex GPT-5.6 Sol xhigh review: clean, 0.98 correctness confidence
- Correctly scoped Claude Opus 4.8 xhigh review: clean, 0.90 correctness confidence

GitHub CI has 111 passing checks. The Windows integration job failed after `curl` received a connection reset while downloading Datadog tooling (exit 35); the aggregate `thank you, next` check reflects that infrastructure failure. No changed-file or implicit-tag test failed.

### DIFF
--- a/packages/next/src/server/lib/implicit-tags.test.ts
+++ b/packages/next/src/server/lib/implicit-tags.test.ts
@@ -74,6 +74,23 @@ describe('getImplicitTags()', () => {
       ],
     },
     {
+      page: '/api/route',
+      pathname: '/api/route',
+      fallbackRouteParams: null,
+      expectedTags: ['_N_T_/layout', '_N_T_/api/layout', '_N_T_/api/route'],
+    },
+    {
+      page: '/foo//bar/',
+      pathname: '/foo//bar/',
+      fallbackRouteParams: null,
+      expectedTags: [
+        '_N_T_/layout',
+        '_N_T_/foo/layout',
+        '_N_T_/foo//bar/layout',
+        '_N_T_/foo//bar/',
+      ],
+    },
+    {
       // Non-ASCII pathname must be percent-encoded so it can be safely
       // serialized into the `x-next-cache-tags` HTTP header. Surrogate-pair
       // emoji exercises run-based replacement (a per-code-unit regex would
--- a/packages/next/src/server/lib/implicit-tags.test.ts
+++ b/packages/next/src/server/lib/implicit-tags.test.ts
@@ -80,17 +80,6 @@ describe('getImplicitTags()', () => {
       expectedTags: ['_N_T_/layout', '_N_T_/api/layout', '_N_T_/api/route'],
     },
     {
-      page: '/foo//bar/',
-      pathname: '/foo//bar/',
-      fallbackRouteParams: null,
-      expectedTags: [
-        '_N_T_/layout',
-        '_N_T_/foo/layout',
-        '_N_T_/foo//bar/layout',
-        '_N_T_/foo//bar/',
-      ],
-    },
-    {
       // Non-ASCII pathname must be percent-encoded so it can be safely
       // serialized into the `x-next-cache-tags` HTTP header. Surrogate-pair
       // emoji exercises run-based replacement (a per-code-unit regex would
--- a/packages/next/src/server/lib/implicit-tags.ts
+++ b/packages/next/src/server/lib/implicit-tags.ts
@@ -29,11 +29,14 @@ const getDerivedTags = (pathname: string): string[] => {
   // we automatically add the current path segments as tags
   // for revalidatePath handling
   if (pathname.startsWith('/')) {
-    const pathnameParts = pathname.split('/')
+    let end = pathname.indexOf('/', 1)
 
-    for (let i = 1; i < pathnameParts.length + 1; i++) {
-      let curPathname = pathnameParts.slice(0, i).join('/')
+    while (true) {
+      if (end === -1) {
+        end = pathname.length
+      }
 
+      let curPathname = pathname.slice(0, end)
       if (curPathname) {
         // all derived tags other than the page are layout tags
         if (!curPathname.endsWith('/page') && !curPathname.endsWith('/route')) {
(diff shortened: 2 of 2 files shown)


packages/next/src/server/lib/implicit-tags.test.ts
```diff
@@ -73,6 +73,23 @@ describe('getImplicitTags()', () => {
...
       ],
     },
+    {
+      page: '/api/route',
+      pathname: '/api/route',
+      fallbackRouteParams: null,
+      expectedTags: ['_N_T_/layout', '_N_T_/api/layout', '_N_T_/api/route'],
+    },
+    {
+      page: '/foo//bar/',
```
bgw: Not sure this test case is useful? It looks like `page` refers to the next.js route? Is it even valid for a next.js route to have double slashes?

bgw (approved):
I doubt this is a particularly hot codepath in any real application (it looks like codex mostly benchmarked this function in isolation?), but it might help with some framework microbenchmarks? It doesn't significantly increase code complexity, so sure?

packages/next/src/server/lib/implicit-tags.test.ts
marcoshernanz: You’re right — `page` is the normalized internal route identifier, and call-site inspection did not find a reachable double-slash form. I removed this synthetic fixture in 453bbe71c4. The valid nested page and `/api/route` fixtures still cover the boundary scan and suffix handling; the focused suite passes 12/12.

PR #96131, opened 2026-07-24
perf: index static route matchers by pathname

marcoshernanz:
### What?

Index non-dynamic route matchers by their definition pathname when the matcher manager reloads, then validate only the exact-path bucket for a static request.

This also adds focused tests showing that unrelated static matchers are not validated and that duplicate static matchers retain provider order.

### Why?

The static request path currently scans every static matcher before finding a match or returning a miss. That makes static route matching O(number of static routes) per request.

A focused warm-process benchmark used an A1/B1/B2/A2 sequence on three identical 8-vCPU/16-GiB Vercel Sandbox workers at the same commit, toolchain, and cache state. At 10,000 static routes, median matcher-call time changed as follows:

| Lookup | Baseline | Candidate | Improvement |
| --- | ---: | ---: | ---: |
| middle | 183,647 ns | 792 ns | 99.557% (229.49x) |
| last | 364,223 ns | 771 ns | 99.783% (461.58x) |
| miss | 336,203 ns | 412 ns | 99.878% (817.35x) |

All three workers agreed directionally for those scenarios. First-route lookup was unchanged within A/A noise and is not claimed as a win.

The tradeoff is additional work during matcher-manager reload: at 10,000 static routes, median reload time increased from 1.667 ms to 2.496 ms (+0.829 ms). This is paid once per reload rather than on each request.

### How?

The matcher manager now builds a `Map<string, RouteMatcher[]>` for static routes in the existing reload pass while collecting dynamic routes separately. Matching uses the locale-normalized pathname, preserving the existing i18n behavior, and keeps arrays in provider order so duplicate handling remains stable. Dynamic route sorting and matching are unchanged.

Validation:

- focused matcher-manager unit tests: 9/9
- `pnpm --filter=next types`
- targeted Prettier and ESLint checks
- `pnpm --filter=next build`
- independent GPT-5.6 Sol xhigh and Claude Opus 4.8 xhigh review: no findings

### DIFF
--- a/packages/next/src/server/route-matcher-managers/default-route-matcher-manager.test.ts
+++ b/packages/next/src/server/route-matcher-managers/default-route-matcher-manager.test.ts
@@ -168,6 +168,62 @@ describe('DefaultRouteMatcherManager', () => {
     const match = await manager.match('/en-US/some/path', options)
     expect(match?.definition).toBe(definition)
   })
+
+  it('only validates static matchers for the requested pathname', async () => {
+    const manager = new DefaultRouteMatcherManager()
+    const unrelated = new RouteMatcher({
+      kind: RouteKind.PAGES,
+      filename: 'unrelated.js',
+      bundlePath: 'unrelated',
+      page: '/unrelated',
+      pathname: '/unrelated',
+    })
+    const expected = new RouteMatcher({
+      kind: RouteKind.PAGES,
+      filename: 'expected.js',
+      bundlePath: 'expected',
+      page: '/expected',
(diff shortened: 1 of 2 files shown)


bgw:
I think you should find somebody from the Next team who's more familiar with the routing to review this. My concerns are:
- This might change the order in which routes are matched.
- This might be over-optimizing for a case that isn't likely to occur in practice.
- This codepath is only hit once per request. Even in your contrived worst-case that only adds 0.364ms/request.

PR #96145, opened 2026-07-24
docs: server client and directives

icyJoseph:
- Enrich the directives entry point
- New guide for Server and Client components. Conceptual backing for the getting started page.


docs/01-app/02-guides/server-and-client-boundary.mdx
```diff
@@ -0,0 +1,201 @@
...
+```txt
+Server Component
+├─ Server Component
+└─ Client Component
+   └─ Client Component
+```
+
+Each component in this tree belongs to the server module graph or the client module graph, and a module shared by both is compiled into each, one copy per graph. The server graph holds the references to the Client Components and the data it passes them at the boundary. The client graph has no knowledge of the server graph, only the data it receives there.
+
+That boundary is where most confusion about Server and Client Components starts. Things like "server vs client rendering," or why you cannot call `useState` in some files, feel like separate rules to memorize. Each is really a consequence of the same boundary, and this guide takes them one at a time.
```
aurorascharff: ules to memorize. Each is really a consequence of the same boundary, and this guide takes them one at a time.

dont love this outro of the intro! Maybe "in this guide, let's go through them. "

docs/01-app/02-guides/server-and-client-boundary.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+Together they become the page's HTML. This server pass runs at build time or per request, and some Client Components may not be reached in it, because they sit behind user interaction, effects, or state that has not happened yet.
+
+On a direct visit, the browser receives that HTML along with the RSC Payload, the serialized server output with its client references, and script tags for the Client Components' JavaScript. During hydration that JavaScript runs again, and React settles until state changes or a [promise passed from the server](/docs/app/guides/streaming) resolves.
+
+A client-side navigation works differently. There is no HTML for the new route: the server sends only the RSC Payload, and the browser builds the Client Components itself, from the JavaScript it already has (or downloads now) and the references and props the payload carries. The Server Components still run on the server to produce that payload, but this time the Client Components run only in the browser.
+
+So "where does it run" is really two questions, and each kind answers them differently:
```
aurorascharff: this explanation felt a little bit long and dragged out when the table is pretty simple

docs/01-app/02-guides/server-and-client-boundary.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+_When_ a component renders, at build time or per request, statically or dynamically is covered in [Rendering Philosophy](/docs/app/guides/rendering-philosophy).
+
+## State and interactivity
+
+A Server Component runs on the server and produces output. Its code never reaches the browser, so on the client it holds no state, runs no effects, and never re-renders.
+
+A Client Component's code does reach the browser. React hydrates it, and from then on it runs on the client: it can hold state, run effects, respond to events, and re-render when its state or props change.
+
+> **Good to know:** Even though a Server Component never runs on the client, its output is not yours to mutate directly. The RSC Payload ships inside the HTML, and React keeps the DOM in sync with it, so hand-editing server-rendered nodes puts them out of sync and React reconciles them away. See [Streaming](/docs/app/guides/streaming).
```
aurorascharff: See [Streaming](/docs/app/guides/streaming).

should there be an anchor here?

docs/01-app/02-guides/server-and-client-boundary.mdx
```diff
@@ -0,0 +1,201 @@
...
+data
+  ↓
+loader or API
+  ↓
+props
+  ↓
+component tree
+```
+
+A Server Component runs on the server, so it can reach server-only resources directly: a database, the filesystem, an internal service, a secret. It reads them in the body of the component, during its own render, with no API route in between to expose the data to the client first.
```
aurorascharff: Feels like this part came very late into the guide

docs/01-app/02-guides/server-and-client-boundary.mdx
```diff
@@ -0,0 +1,201 @@
...
+  const posts = await getPosts() // runs on the server, during render
+  return <PostList posts={posts} />
+}
+```
+
+Getting the data is no longer a separate step that feeds the tree. It is part of evaluating a node in the tree.
+
+> **Good to know:** Because a Server Component can read secrets and server-only data directly, be deliberate about what you pass to Client Components, props are serialized and sent to the browser. See [Data Security](/docs/app/guides/data-security#passing-data-from-server-to-client).
+
+Client Components fetch too, so this is not reserved for the server. What differs is when, and who starts it. A Client Component can fetch after it mounts, in an effect or a data library. Or a Server Component can start the fetch during its own render and pass the pending promise down, and the Client Component reads it with [`use`](https://react.dev/reference/react/use) as it resolves, so the request is in flight before the client even runs.
```
aurorascharff: Maybe worth linking to the new section of the use() docs about this

docs/01-app/02-guides/server-and-client-boundary.mdx
```diff
@@ -0,0 +1,201 @@
...
+
+You only need `'use client'` at the entry to a client subtree, not on every file inside it. Everything below the entry is already client, so a component that calls `useState` or `useEffect` does not need its own directive, you can add it at the import site instead. That lets you leave a shared primitive untouched and put the boundary on a module closer to your application code. If client code ends up in the server graph without a boundary, the compiler tells you, so you can add the directive where it is needed rather than everywhere.
+
+For more patterns, see [Interleaving Server and Client Components](/docs/app/getting-started/server-and-client-components#interleaving-server-and-client-components).
+
+More on other directives used in Next.js's, see the [directives](/docs/app/api-reference/directives) documentation.
+
+## Putting it together
+
+It is all one boundary, seen from a few angles:
```
aurorascharff: I dont really like the "main conclusion" of the guide. The explanations and concepts are very useful, but what are we trying to teach?

docs/01-app/03-api-reference/01-directives/index.mdx
```diff
@@ -3,4 +3,57 @@ title: Directives
 description: Directives are used to modify the behavior of your Next.js application.
 ---
 
-The following directives are available:
+Directives are string literals that tell the compiler and bundler how to treat the code around them. On their own they look like a plain string, but Next.js and React read them as an instruction to transform the code they cover: into a client entry point, a server function, or a cached function.
+
+| Directive      | Marks the code it covers to...                              | Reference                                                     |
+| -------------- | ----------------------------------------------------------- | ------------------------------------------------------------- |
```
aurorascharff: So good docs page!

aurorascharff: But: would be great to do another column about whats nextjs and whats react

docs/01-app/03-api-reference/01-directives/index.mdx
```diff
@@ -3,4 +3,57 @@ title: Directives
...
+// Function level: applies to this function only
+export async function getUsers() {
+  'use cache'
+  // ...
+}
+```
+
+Each directive marks one side of a boundary, and a module sits on a single side. You define `'use server'` and `'use cache'` functions in server modules, never inside a `'use client'` module. A Client Component can import and call them; it cannot declare them.
+
+For `'use server'` and `'use cache'`, put the directive where the code that needs it lives. Add it to the function or component that needs it, or wherever a build error points you. From there you can move it up to a parent or down into a utility function as your code changes.
```
aurorascharff: or wherever a build error points you

what is the context here? We need some sort of link or explanation, or just remove this

docs/01-app/03-api-reference/01-directives/index.mdx
```diff
@@ -3,4 +3,57 @@ title: Directives
...
+- **`'use server'`** marks its exports as [Server Functions](/docs/app/glossary#server-function). They must be `async`, and their arguments and return values are serialized across the network. A Client Component never receives the function's code, only a reference it uses to invoke the server.
+- **`'use cache'`** marks its exports as cacheable by their inputs. They must be `async`, their arguments and return values must be serializable, and they cannot read request-time APIs like `cookies()`, `headers()`, or `searchParams` directly.
+
+At the file level these rules cover every export, framework functions included. A `'use cache'` at the top of a `page` or `layout`, for example, requires `generateMetadata`, `generateStaticParams`, and any other exports to be `async` too, even the ones that would otherwise be synchronous.
+
+## Placement decides what you cache
+
+With `'use cache'`, placement also decides how much you store. The directive on a page caches everything the page renders; the same directive on a single data function caches only that function's result. When the expensive part is one small request, cache that function instead of the whole page.
```
aurorascharff: I feel like some of this is being repeated many times accross the page in different ways

docs/01-app/03-api-reference/01-directives/index.mdx
```diff
@@ -3,4 +3,57 @@ title: Directives
...
+| `'use cache'`  | Next.js    | be cached and reused based on its inputs                    | [`use cache`](/docs/app/api-reference/directives/use-cache)   |
+
+`'use cache'` also has [`'use cache: remote'`](/docs/app/api-reference/directives/use-cache-remote) and [`'use cache: private'`](/docs/app/api-reference/directives/use-cache-private) variants, and you can control where cached entries are stored by configuring your own [cache handlers](/docs/app/api-reference/config/next-config-js/cacheHandlers).
+
+> **Good to know:** Directives transform your code when it compiles, in development and in production alike. The error overlay, indicators, and stack traces still point at the file and line you wrote, not the generated output.
+
+## Where to place a directive
+
+`'use client'` came first, and it set the original shape: a directive at the top of a file, before any imports, that applies to the whole module. It only goes there. The whole module is bundled and shipped to the browser, so the boundary is drawn at the file, not partway through it; there is no inline `'use client'`.
```
aurorascharff: can we start with not code block and say something else than "came first"

aurorascharff: double "whole"

docs/01-app/03-api-reference/01-directives/index.mdx
```diff
@@ -3,4 +3,57 @@ title: Directives
...
+}
+
+// Function level: applies to this function only
+export async function getUsers() {
+  'use cache'
+  // ...
+}
+```
+
+Each directive marks one side of a boundary, and a module sits on a single side. You define `'use server'` and `'use cache'` functions in server modules, never inside a `'use client'` module. A Client Component can import and call them; it cannot declare them.
```
aurorascharff: This section still reads a bit strange to me

docs/01-app/03-api-reference/01-directives/index.mdx
```diff
@@ -3,4 +3,57 @@ title: Directives
...
+
+For `'use server'` and `'use cache'`, put the directive where the code that needs it lives. Add it to the function or component that needs it. From there you can move it up to a parent or down into a utility function as your code changes.
```
aurorascharff: double "needs it"

docs/01-app/02-guides/server-and-client-boundary.mdx
```diff
@@ -0,0 +1,182 @@
+---
+title: The Server and Client Boundary
+description: A mental model for Server and Client Components in the App Router. What runs on the server, what runs in the browser, why "server vs client rendering" is two questions, and why the same React component now comes in two kinds.
```
aurorascharff: "two two" idk why the agent loves pointless numbers, lets simplify

aurorascharff (approved):

PR #96146, opened 2026-07-24
docs(preventing-flash): cover reconciling state 

icyJoseph:
We got feedback for one case where, in dev mode, Strict Mode caused removed the attributes added to the HTML tag by the injection script.

### DIFF
--- a/docs/01-app/02-guides/preventing-flash-before-hydration.mdx
+++ b/docs/01-app/02-guides/preventing-flash-before-hydration.mdx
@@ -479,6 +479,35 @@ export function Accordion() {
 
 The inline script and the lazy `useState` initializer both read from `localStorage`. They always agree, so React's initial state matches the DOM.
 
+## Re-applying attributes in development
+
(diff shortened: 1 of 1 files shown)


docs/01-app/02-guides/preventing-flash-before-hydration.mdx
```diff
@@ -479,6 +479,35 @@ export function Accordion() {
...
+
+One way to fix this is to do what you would do without the inline script at all. Read the stored value on the client and apply it, in the component that owns the theme, in a [`useLayoutEffect`](https://react.dev/reference/react/useLayoutEffect) that runs [before paint](#why-not-useeffect):
+
+```tsx filename="app/components/theme-toggle.tsx"
+'use client'
+
+import { useLayoutEffect } from 'react'
+
+export function ThemeToggle() {
+  // Re-apply after React clears it on the dev remount; a no-op in production.
```
aurorascharff: remove semicolon

aurorascharff (approved):

docs/01-app/02-guides/preventing-flash-before-hydration.mdx
icyJoseph:
```suggestion
  // Re-apply after React clears it on the dev remount. This is a no-op in production.
```

PR #96153, opened 2026-07-24
Run Cache Components dev validation on a worker thread

unstubbable:
This moves the dev-mode Cache Components validation renders off the dev server's main thread onto a worker thread, so rapid navigation no longer starves the event loop. The worker crosses into the app-page bundle exactly once, calling a new `ComponentMod.routeModule.runValidationInDev` entry that rebuilds the render context, work store, and request store from a serializable snapshot and runs the whole validation there, so the client prerender and the user's client components resolve the single app-page React instance rather than a second copy.

A thin worker shell (`dev-validation-worker.ts`) plus a single-worker pool (`dev-validation-worker-pool.ts`) load the user bundle via `loadComponents`, install code-frame support for CLI output, and forward the validation errors back as Flight bytes; the main thread only delivers them to the overlay through `sendErrorsToBrowser`. The snapshot, the globalThis-symbol handoff, and the shared error-delivery helpers live in their own modules. The dev server installs the worker when `experimental.devValidationWorker` is not `false`, and `runDevValidationInBackground` uses it when present, falling back to the in-process path otherwise. A one-slot `SharedArrayBuffer` propagates a supersede abort into the worker so a newer navigation cancels an in-flight validation.

The runtime bundle gains an `app-worker` entry for the worker with the `build/swc` boundary externalized for the code-frame native binding, and `patch-error-inspect.ts` now backs its code-frame renderer with a globalThis symbol so all copies of the module share it across the thread.

The synthetic `bench/dev-validation` benchmark navigates back-to-back, so every click lands inside the validation window — the worst case for main-thread contention. Browser-observed navigation TTFB in that window, worker vs in-process:

| Route  | Worker (p50/p95/max) | In-process (p50/p95/max) |
| ------ | -------------------- | ------------------------ |
| client | 19 / 24 / 27 ms      | 40 / 66 / 7762 ms        |
| server | 42 / 45 / 46 ms      | 122 / 158 / 252 ms       |
| sprite | 109 / 117 / 169 ms   | 196 / 208 / 299 ms       |

The steady-state difference is only tens of milliseconds; the effect that matters is the tail. In-process, a navigation that collides with an in-flight validation render can stall for seconds (this run peaked at ~7.8s on the client route, and the peak varies run to run) because a staged render does not yield until it finishes; off-thread the main thread stays free and that stall disappears.

Read these as an upper bound, not a speedup that generalizes. The work moved off-thread is the validation render's CPU — bounded, route-dependent, and free of IO — so it does not grow with the main render's cost. In an app whose main render is dominated by IO the same absolute saving is a small fraction of the request, and it only appears when a navigation lands in the brief validation window; at ordinary click speed it is largely invisible. This is a dev-only responsiveness improvement whose benefit varies widely with the app and the navigation pattern.

As a follow-up, the validation work that still runs on the main thread could move to the worker as well. When the main render can't be reused for validation (for example after a cache miss), the validation re-renders on the main thread to produce its inputs, resuming from the Resume Data Cache (RDC) the main render already filled, and only then hands the resulting Flight chunks to the worker. A later iteration could run those renders inside the worker too, transporting the (serializable) RDC so the worker can resume from the filled caches rather than reading them back on the main thread.

closes NAR-895

### DIFF
--- a/packages/next/src/server/app-render/dev-validation-worker-snapshot.ts
+++ b/packages/next/src/server/app-render/dev-validation-worker-snapshot.ts
@@ -1,0 +1,101 @@
+import type { AppRenderContext, ResolvedValidationInputs } from './app-render'
+import type { PrefetchingMode } from './app-render'
+import type { OpaqueFallbackRouteParams } from '../request/fallback-params'
+import type { AnyStream } from './stream-ops'
+import type {
+  DevValidationSnapshot,
+  SerializedValidationInputs,
+} from './dev-validation-worker-globals'
+
+import { isNodeNextResponse } from '../base-http/helpers'
+
+/**
+ * Builds the serializable snapshot the dev validation worker needs to rebuild
+ * the render context and inputs and run `runValidationInDev` on the worker
+ * thread. Reads only serializable data from the settled render; the live
+ * objects (`componentMod`, the async-storage stores,
+ * `getDynamicParamFromSegment`) are reconstructed on the worker from this seed.
+ * Draining the debug channels is async, so this returns a promise.
+ */
+export async function buildDevValidationSnapshot(
+  ctx: AppRenderContext,
+  instantInputs: ResolvedValidationInputs | null,
+  staticInputs: ResolvedValidationInputs,
+  prefetchMode: PrefetchingMode,
+  fallbackRouteParams: OpaqueFallbackRouteParams | null,
+  devRenderDidError: boolean
+): Promise<DevValidationSnapshot> {
+  const requestStore = staticInputs.requestStore
+  const responseFinished =
+    !isNodeNextResponse(ctx.res) || ctx.res.originalResponse.writableFinished
+
+  // The instant and static inputs may share one debug channel, so drain each
+  // channel only once.
+  const chunksByChannel = new WeakMap<AnyStream, Uint8Array[]>()
+  const getDebugChunksOnce = async (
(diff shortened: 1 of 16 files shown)


packages/next/src/server/app-render/dev-validation-worker-snapshot.ts
```diff
@@ -0,0 +1,105 @@
...
+      urlPathname: requestStore.url.pathname,
+      urlSearch: requestStore.url.search,
+      rootParams: requestStore.rootParams,
+      isDraftMode: requestStore.draftMode.isEnabled,
+      isHmrRefresh: requestStore.isHmrRefresh ?? false,
+      hmrRefreshHash: requestStore.hmrRefreshHash,
+    },
+    interpolatedParams: ctx.interpolatedParams,
+    requestFallbackRouteParams: ctx.fallbackRouteParams
+      ? [...ctx.fallbackRouteParams.entries()]
```
timneutkens, line 82: I think we can avoid the spread here and same for fallbackRouteParams. Calling `entries()` would be enough 🤔

timneutkens (approved):

packages/next/src/server/app-render/dev-validation-worker-snapshot.ts
unstubbable: Addressed in #96210

PR #96157, opened 2026-07-24
Cache filesystem operations during output tracing

marcoshernanz:
## What changed

Cache the `readFile`, `readlink`, and `stat` promises used by the final `nodeFileTrace` pass for the lifetime of that pass.

The custom callbacks and their tolerated errors are unchanged. Storing the promise before returning it also deduplicates concurrent operations for the same path.

## Why

`@vercel/nft` normally caches these filesystem operations, but `collectBuildTraces` replaces NFT's methods with custom callbacks so Next can tolerate `EISDIR` reads and `ENOTDIR` stats. That replacement bypassed NFT's cache and repeated the same physical operations many times.

On a representative trace replay, caching changed:

- `readFile`: 3,754 → 1,314 calls
- `readlink`: 148,988 → 6,615 calls
- `stat`: 26,311 → 4,262 calls

The normalized 1,680,526-byte trace graph remained byte-identical across all 20 direct A/A and A/B observations.

## Performance

Exact commit `0950e8ae...`, Node 24.14.1, pnpm 10.33.0, 16 vCPU / 32 GiB, cold `.next`, 100 page groups / 300 static leaf routes, explicit webpack builds.

Four identical-baseline builds established a 0.92% arbitrary-label wall difference. Eight ABBA–BAAB builds then measured:

| Metric | Baseline | Candidate | Reduction |
|---|---:|---:|---:|
| Whole build wall | 90.945 s | 82.790 s | 8.97% |
| `collect-build-traces` | 28.352 s | 19.467 s | 31.34% |
| `node-file-trace-build` | 28.204 s | 19.319 s | 31.50% |
| User CPU | 156.305 s | 146.508 s | 6.27% |
| System CPU | 14.723 s | 12.575 s | 14.59% |

Both four-run blocks agreed (8.88% and 9.05% whole-build reductions). All four candidate walls were below all four baselines; exact two-sided permutation p = 0.0286. Peak RSS was neutral (-0.23%, p = 0.514).

## Validation

- `pnpm --filter=next build`
- `pnpm --filter=next types`
- focused Prettier and ESLint
- webpack `build-trace-extra-entries` production integration test
- webpack `build-trace-extra-entries-monorepo` production integration test
- 4/4 A/A and 8/8 A/B production builds, each with all 302 static pages
- GPT-5.6 Sol xhigh + Claude Opus 4.8 xhigh review: zero findings; patch classified correct at 0.97 confidence

### DIFF
--- a/packages/next/src/build/collect-build-traces.ts
+++ b/packages/next/src/build/collect-build-traces.ts
@@ -42,6 +42,19 @@ export const makeIgnoreFn = (root: string, ignores: string[]) => {
   }
 }
 
+function cacheFileOperation<T>(
+  cache: Map<string, Promise<T>>,
+  file: string,
+  operation: () => Promise<T>
+): Promise<T> {
+  let promise = cache.get(file)
+  if (!promise) {
+    promise = operation()
+    cache.set(file, promise)
+  }
+  return promise
+}
+
 function shouldIgnore(
   file: string,
   serverIgnoreFn: (file: string) => boolean,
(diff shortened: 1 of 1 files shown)


bgw (commented):
I think this is only in the webpack codepath? We implement NFT (node file trace) in rust when you're using turbopack, and that uses turbo-tasks-fs which already caches disk IO.

I'm fine with caching readlink and stat since that probably won't consume much memory.

I think that caching readFile is risky because it could lead to tons of memory usage in the case that NFT follows excess numbers of paths that it shouldn't (which it often does).

However, just caching readlink and stat likely isn't worth it? The majority of your wins are probably from caching readFile?

coreyward:
@bgw:

> I think that caching readFile is risky because it could lead to tons of memory usage in the case that NFT follows excess numbers of paths that it shouldn't (which it often does).

Not sure if this would matter since `@vercel/nft` already caches `readFile`. The peak memory from Marcos' tests didn't really change either.

> However, just caching readlink and stat likely isn't worth it? The majority of your wins are probably from caching readFile?

Even if `readFile` were reducing the majority of the bytes read, dropping 93% of the IO calls would still be a significant win. From Marcos' post:

- readFile: 3,754 → 1,314 calls
- readlink: 148,988 → 6,615 calls
- stat: 26,311 → 4,262 calls

Somewhat anecdotal, but my own work trying to speed up builds after upgrading to Next 15 led me to a similar optimization, and my measurements showed `readlink` was about 40% of the issue.

PR #96159, opened 2026-07-24
Improve root detection to handle worktrees, more workspaces & stray lockfiles

### LINKED ISSUE #92978: Turbopack: misdetected workspace root (from stray parent lockfile) causes RAM exhaustion (40+ GB) and infinite module-resolution retry on first request

https://github.com/MrMyNameMaster/next-turbopack-ram-repro

### To Reproduce

**Environment setup (the trigger):**

1. Create an empty/orphan `package-lock.json` in a parent directory of your project. This can happen accidentally by running `npm install` in your home folder. Example content:
   ```json
   {
     "name": "BobCleaner",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {}
   }
   ```
   Location: `C:\Users\<name>\package-lock.json` (no [package.json](cci:7://file:///c:/Users/BobCleaner/Desktop/CoreLab/package.json:0:0-0:0) beside it)

2. Create a Next.js 16 project in a subdirectory, e.g. `C:\Users\<name>\Desktop\myapp`:
   ```
   npx create-next-app@latest myapp
   ```
   Use App Router + Tailwind v4 (`@tailwindcss/postcss`).

3. Project structure:
   ```
   C:\Users\<name>\package-lock.json   <-- orphan
   C:\Users\<name>\Desktop\myapp\
       package.json
       package-lock.json
       postcss.config.mjs
       node_modules\
       ...
   ```

**Steps:**

1. `cd` into the project folder and run `npm run dev`.
2. Next.js prints the expected warning about multiple lockfiles and picks `C:\Users\<name>\` as the inferred workspace root.
3. Open `http://localhost:3000` in a browser.
4. Observe:
   - Compiler throws `Can't resolve 'tailwindcss' in 'C:\Users\<name>\Desktop'` repeatedly
   - Node workers start spawning and RAM usage grows unbounded (observed: one parent Node process reaching **43,516 MB**, dozens of child workers, see screenshot)
   - The machine becomes unresponsive if not killed

**Mitigation that works:**
- Deleting the orphan `C:\Users\<name>\package-lock.json` — OR —
- Setting `turbopack.root: __dirname` in [next.config.ts](cci:7://file:///c:/Users/BobCleaner/Desktop/CoreLab/next.config.ts:0:0-0:0)

Either fix makes the dev server behave normally.

**Current behavior:**

When Next.js infers a workspace root from a stray parent `package-lock.json`, module resolution for CSS/PostCSS-related imports (e.g. `@import "tailwindcss"`) is performed relative to the wrong root. The resolver (`enhanced-resolve`) walks up from `C:\Users\<name>\Desktop` → `C:\Users\<name>` → `C:\Users` → `C:\` looking for `node_modules`, never finding one, and throws:

```
Error: Can't resolve 'tailwindcss' in 'C:\Users\BobCleaner\Desktop'
    at finishWithoutResolve (...\enhanced-resolve\lib\Resolver.js:587:18)
    ...
  details: "resolve 'tailwindcss' in 'C:\\Users\\BobCleaner\\Desktop'\n" +
    '  Parsed request is a module\n' +
    '  No description file found in C:\\Users\\BobCleaner\\Desktop or above\n' +
    '  resolve as module\n' +
    "    C:\\Users\\BobCleaner\\Desktop\\node_modules doesn't exist or is not a directory\n" +
    "    C:\\Users\\BobCleaner\\node_modules doesn't exist or is not a directory\n" +
    "    C:\\Users\\node_modules doesn't exist or is not a directory\n" +
    "    C:\\node_modules doesn't exist or is not a directory"
```

Instead of failing once with a clear, actionable error, the dev server:

1. Retries / respawns worker processes continuously.
2. The file watcher appears to scan the entire inferred parent tree (`C:\Users\<name>\`), which on a normal Windows user profile contains tens of GB of files (Documents, Downloads, AppData, OneDrive, etc.).
3. RAM usage grows without any cap. In my case, one parent Node process reached **43.5 GB** within seconds, with 20+ child workers. This can OOM-crash or freeze a machine with less RAM.

Screenshot of Task Manager during the incident is attached below.
Image

**Expected behavior:**

1. **Fail fast, fail once.** If a module cannot be resolved during compilation, emit a single clear error and stop — do not respawn workers or retry in a hot loop.
2. **Memory safeguard.** Turbopack / Next.js workers should have an upper bound on RAM usage; a single dev session should never be able to allocate 40+ GB.
3. **Respect `turbopack.root`.** When `turbopack.root` is explicitly configured, all downstream resolvers (PostCSS / `enhanced-resolve` / file watcher scope) should honor it.
4. **Stricter root detection.** An empty / orphan `package-lock.json` with no sibling `package.json` should not be treated as a workspace root. At minimum, the warning about multiple lockfiles should be promoted to an error, or the detection should prefer the lockfile nearest to the current working directory rather than the outermost one.

```bash
Operating System:
  Platform: win32
  Arch: x64
  Available memory (MB): 65297
  Available CPU cores: 32
Binaries:
  Node: 24.14.1
  npm: 11.11.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

CSS, Turbopack

next dev (local)

### Additional context

**Configuration details:**

- Tailwind v4 via `@tailwindcss/postcss` (the default setup from `create-next-app`)
- [postcss.config.mjs](cci:7://file:///c:/Users/BobCleaner/Desktop/CoreLab/postcss.config.mjs:0:0-0:0):
  ```js
  const config = {
    plugins: {
      "@tailwindcss/postcss": {},
    },
  };
  export default config;
  ```
- [next.config.ts](cci:7://file:///c:/Users/BobCleaner/Desktop/CoreLab/next.config.ts:0:0-0:0) (after mitigation):
  ```ts
    turbopack: { root: __dirname },
    experimental: {
      serverActions: { bodySizeLimit: "6mb" },
    },
    // ...
  };
  ```

**Why this matters for Windows users specifically:**

On Windows it is extremely common to accidentally create `package-lock.json` in `C:\Users\<name>\` by running `npm install` in the default PowerShell/CMD starting directory. The user has no obvious reason to suspect this file exists, and the symptom (Tailwind can't resolve + RAM explosion) is very far from the actual cause (orphan lockfile in home folder). The error message contains no hint pointing to the lockfile.

**Severity:**

This is not just a confusing error — it is an uncontrolled resource exhaustion that can make a developer machine unusable. On machines with less RAM than mine (I have 64 GB), this would OOM-kill the session or trigger a BSOD/hang before the user can react.

**Suggested fixes (in order of impact):**

1. Hard cap on Turbopack worker memory / stop respawning on unresolvable modules.
2. When `turbopack.root` is set explicitly in config, skip the multi-lockfile inference entirely.
3. Detect and ignore "empty" `package-lock.json` files (no `packages` / no [package.json](cci:7://file:///c:/Users/BobCleaner/Desktop/CoreLab/package.json:0:0-0:0) sibling) during root inference, or at least warn loudly.
4. In the multi-lockfile warning, prefer the lockfile closest to the CWD instead of the outermost one.

I have not yet tested against `next@canary` — will update if I find the exact canary that introduced the behavior.

sampoder:
Watching too many files can hurt performance, and impacts how much work Turbopack is doing. See for example the conversation around https://github.com/vercel/next.js/pull/94597.

This PR handles the following four cases:

- Git worktrees: we can include a guard to prevent exiting a worktree (by looking at the `.git` file).
- Git repository boundaries: we can include a guard to prevent us from exiting a checked-out repository and watching the repository’s parent directory.
    - We will log a warning if this happens.
- Empty / bogus `package.lock` files & `package-lock.json` files (and equivalents using other package managers): if we spot a `package-lock.json` file but it has no packages or a `package-lock.json` w/o an associated `package.json` file. We can ignore them. This handles the case of accidentally running `npm install` in any directory.
- Home directory detection: we can detect if we’re going to start watching your home dir and rollback up if this is the case.
    - We will log a warning if this happens.
    - If your Next config is in your home dir, we will still run your app there as Next config is the furthest we can rollback
- NPM, Yarn, Lerna & Bun workspaces: based on the `workspaces` field in the `package.json` or `lerna.json`

I would recommend using split-view to review. I also cleaned up the `warnDuplicatedLockFiles()` function because it was a bit unnecessary to repeat the full message.

Closes https://github.com/vercel/next.js/issues/92978

### DIFF
--- a/packages/next/src/lib/find-root.ts
+++ b/packages/next/src/lib/find-root.ts
@@ -1,94 +1,279 @@
-import { dirname } from 'path'
+import { basename, dirname, isAbsolute, join, relative, resolve } from 'path'
+import { existsSync, readFileSync, realpathSync, statSync } from 'fs'
+import { homedir } from 'os'
 import findUp from 'next/dist/compiled/find-up'
 import * as Log from '../build/output/log'
 
-function findWorkRoot(cwd: string) {
-  // Find-up evaluates the list of files at each level.
-  // For pnpm-workspace.yaml we first want to look up before searching for lockfiles as those can be included in the application directory by accident.
-  const pnpmWorkspaceFile = findUp.sync(
-    'pnpm-workspace.yaml',
+export type RootBoundary = {
+  boundary: string
+  marker: string
+  root: string
+  type: 'repository' | 'home'
+}
 
-    {
-      cwd,
-    }
-  )
+// folders that are git checkouts / git worktrees are a good indicator
+// of the folder that someone's project is contained in.
+type GitBoundary =
+  | { type: 'repository'; dir: string }
+  | { type: 'worktree'; dir: string }
 
-  if (pnpmWorkspaceFile) {
-    return pnpmWorkspaceFile
+function toRealPath(path: string): string {
+  try {
+    return realpathSync.native(path)
+  } catch {
+    return path
+  }
+}
+
+// `homedir()` throws when it cannot determine a home directory, and returns an
+// empty string on some platforms. an empty value would match every directory,
+// so treat both as "no home directory".
+function findHomeDir(): string | undefined {
+  try {
+    return toRealPath(homedir()) || undefined
+  } catch {
+    return undefined
+  }
+}
+
+// if `descendant` is inside of `ancestor` or `descendant` == `ancestor`
+function isAncestorOrSelf(ancestor: string, descendant: string): boolean {
+  const rel = relative(ancestor, descendant)
+  return rel === '' || (!rel.startsWith('..') && !isAbsolute(rel))
+}
+
+// if `descendant` is inside of `ancestor`
+function isStrictAncestor(ancestor: string, descendant: string): boolean {
+  const rel = relative(ancestor, descendant)
+  return rel !== '' && !rel.startsWith('..') && !isAbsolute(rel)
+}
+
+// true for a git worktree (its git dir has a `commondir` file);
+// this method returns false for a submodule.
+function isGitWorktree(dir: string, gitFile: string): boolean {
+  let contents: string
+  try {
+    contents = readFileSync(gitFile, 'utf8')
+  } catch {
+    return false
   }
 
-  return findUp.sync(
-    [
-      'pnpm-lock.yaml',
-      'package-lock.json',
-      'yarn.lock',
-      'bun.lock',
-      'bun.lockb',
-    ],
-    {
-      cwd,
-    }
+  // Match the behavior in git:
+  // https://github.com/git/git/blob/13c7afec212fc97ce257d15601659314c6673d6c/setup.c#L1007-L1018
+  const match = /^gitdir: (.+?)[\r\n]*$/s.exec(contents)
+  if (!match) return false
+
+  const gitDir = resolve(dir, match[1])
+  return existsSync(join(gitDir, 'commondir'))
+}
+
+// finds the first git repository or git worktree above the
+// current working directory. we limit the project root to
+// inside a worktree or git repository. this is because if
+// you use worktrees, you may have lock files from the
+// parent git repo that are in folders that are not relevant
+// to watch.
+function findGitBoundary(cwd: string): GitBoundary | undefined {
+  let isWorktree = false
+  const gitDir = findUp.sync(
+    (dir) => {
+      const candidate = join(dir, '.git')
+      const isDirectory = statSync(candidate, {
+        throwIfNoEntry: false,
+      })?.isDirectory()
+
+      isWorktree = !isDirectory && isGitWorktree(dir, candidate)
+      return isDirectory || isWorktree ? dir : undefined
+    },
+    { cwd, type: 'directory' }
   )
+  if (!gitDir) return undefined
+
+  return {
+    type: isWorktree ? 'worktree' : 'repository',
+    dir: toRealPath(gitDir),
+  }
+}
+
+const LOCK_FILES = [
+  'pnpm-lock.yaml',
+  'package-lock.json',
+  'yarn.lock',
+  'bun.lock',
+  'bun.lockb',
+]
+
+// Dedicated workspace-root files. pnpm and Lerna declare the root this way.
+const WORKSPACE_MARKERS = ['pnpm-workspace.yaml', 'lerna.json']
+
+// npm, Yarn, and Bun declare the root via a package.json `workspaces` field,
+// which may be an array or an object (e.g. Yarn's `{ packages: [...] }`).
+function hasWorkspacesField(packageJson: string): boolean {
+  try {
+    return JSON.parse(readFileSync(packageJson, 'utf8')).workspaces != null
+  } catch {
+    return false
+  }
+}
+
+function findWorkspaceMarker(directory: string): string | undefined {
+  for (const name of WORKSPACE_MARKERS) {
+    const markerPath = join(directory, name)
+    if (existsSync(markerPath)) {
+      return markerPath
+    }
+  }
+
+  const packageJson = join(directory, 'package.json')
+  return hasWorkspacesField(packageJson) ? packageJson : undefined
+}
+
+type RootCandidate = { path: string; isWorkspaceMarker: boolean }
+
+function findNextRootCandidate(cwd: string): RootCandidate | undefined {
+  // a workspace marker takes precedence over lockfiles, which can be included in
+  // the application directory by accident.
+  const marker = findUp.sync(findWorkspaceMarker, { cwd })
+  if (marker) {
+    return { path: marker, isWorkspaceMarker: true }
+  }
+
+  const lockFile = findUp.sync(LOCK_FILES, { cwd })
+  return lockFile ? { path: lockFile, isWorkspaceMarker: false } : undefined
 }
 
 export function findRootDirAndLockFiles(cwd: string): {
   lockFiles: string[]
   rootDir: string
+  boundary?: RootBoundary
 } {
-  const lockFile = findWorkRoot(cwd)
-  if (!lockFile)
+  const initialCandidate = findNextRootCandidate(cwd)
+  if (!initialCandidate)
     return {
       lockFiles: [],
       rootDir: cwd,
     }
 
-  const lockFiles = [lockFile]
+  const candidates = [initialCandidate.path]
+  let foundWorkspaceMarker = initialCandidate.isWorkspaceMarker
   while (true) {
-    const lastLockFile = lockFiles[lockFiles.length - 1]
-    const currentDir = dirname(lastLockFile)
+    const lastCandidate = candidates[candidates.length - 1]
+    const currentDir = dirname(lastCandidate)
     const parentDir = dirname(currentDir)
 
     // dirname('/')==='/' so if we happen to reach the FS root (as might happen in a container we need to quit to avoid looping forever
     if (parentDir === currentDir) break
 
-    const newLockFile = findWorkRoot(parentDir)
+    const nextCandidate = findNextRootCandidate(parentDir)
 
-    if (!newLockFile) break
+    if (!nextCandidate) break
 
-    lockFiles.push(newLockFile)
+    // findNextRootCandidate searches for markers all the way up, so a non-marker
+    // here means there is no higher marker: nothing above can extend the root.
+    if (foundWorkspaceMarker && !nextCandidate.isWorkspaceMarker) break
+
+    candidates.push(nextCandidate.path)
+    foundWorkspaceMarker ||= nextCandidate.isWorkspaceMarker
+  }
+
+  // filter out candidates that would push the root past a
+  // git boundary or into the home directory.
+  const homeDir = findHomeDir()
+  const gitBoundary = findGitBoundary(cwd)
+
+  const acceptedCandidates: string[] = []
+  let boundary: RootBoundary | undefined
+
+  for (const candidate of candidates) {
+    const dir = dirname(candidate)
+
+    // the app's own directory is always a valid root. even if it
+    // is the home directory.
+    if (dir !== cwd) {
+      const resolvedDir = toRealPath(dir)
+
+      // `dir` sits above the repo or worktree that contains the app.
+      if (gitBoundary && isStrictAncestor(resolvedDir, gitBoundary.dir)) {
+        // add a warning for going beyond a repository
+        if (gitBoundary.type === 'repository') {
+          boundary = {
+            boundary: gitBoundary.dir,
+            marker: candidate,
+            root: dir,
+            type: 'repository',
+          }
+        }
+        break
+      }
+
+      // never treat the home directory (or above) as the root.
+      if (homeDir && isAncestorOrSelf(resolvedDir, homeDir)) {
+        boundary = {
+          boundary: homeDir,
+          marker: candidate,
+          root: dir,
+          type: 'home',
+        }
+        break
+      }
+    }
+
+    acceptedCandidates.push(candidate)
   }
 
   return {
-    lockFiles,
-    rootDir: dirname(lockFiles[lockFiles.length - 1]),
+    lockFiles: acceptedCandidates.filter((candidate) =>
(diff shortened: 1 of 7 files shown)


packages/next/src/lib/find-root.ts
```diff
@@ -58,37 +203,85 @@ export function findRootDirAndLockFiles(cwd: string): {
...
+  // filter out lockfiles that would push the root past a 
+  // git boundary or into the home directory.
+  const homeDir = homedir()
+  const gitBoundary = findGitBoundary(cwd)
+
+  const acceptedLockFiles: string[] = []
+  let boundary: RootBoundary | undefined
+
+  for (const candidate of lockFiles) {
+    const dir = dirname(candidate)
```
bgw: All the stuff on stackoverflow suggests just using `relative` for this, which is frustrating because that's incomplete advice because it only compares paths lexically, and you might think that you're not inside of a directory when you actually are.

Use `fs.realpathSync.native` on `homeDir` and on the result of `dirname` before doing comparisons: https://nodejs.org/api/fs.html#fsrealpathnativepath-options-callback

Caveats:
- This will throw if the path does not exist (e.g. `HOME` can be set to arbitrary values) or if there's an invalid symlink, in these cases, fall back to the original path name.
- This performs some IO, so do it outside of the `isStrictAncestor`/`isAncestorOrSelf` helper functions. We don't need to repeatedly call `realpath` on the homedir inside of the loop.
- Use the sync API, it's faster than the async one, and it's not worth trying to do what effectively amounts to a few `stat()` calls outside of the main thread, especially when this logic would be blocking startup anyways.

packages/next/src/lib/find-root.ts
```diff
@@ -1,30 +1,174 @@
...
+  const gitDir = resolve(dir, match[1].trim())
+  return existsSync(join(gitDir, 'commondir'))
+}
+
+// finds the first git repository or git worktree above the 
+// current working directory. we limit the project root to 
+// inside a worktree or git repository. this is because if
+// you use worktrees, you will have lock files from the 
+// parent git repo that are in folders that are not relevant
+// to watch.
```
lukesandberg: nit: only true if the worktree is in a gitignored directory inside of the repo

so i might soften it to 'might' have a lockfile above

packages/next/src/lib/find-root.ts
```diff
@@ -1,30 +1,174 @@
...
+function findGitBoundary(cwd: string): GitBoundary | undefined {
+  let dir = cwd
+  while (true) {
+    const gitPath = join(dir, '.git')
+    let isDirectory: boolean | undefined
+    try {
+      isDirectory = statSync(gitPath).isDirectory()
+    } catch {
+      isDirectory = undefined
```
lukesandberg: nit: pass `throwIfNoEntry:false` to `statSync` instead of the catch

lukesandberg: then i think it would  just be `const isDirectory = statSync(gitPath, {throwIfNoEntry:false})?.isDirectory()`

packages/next/src/lib/find-root.ts
```diff
@@ -1,30 +1,174 @@
...
+const LOCK_FILES = [
+  'pnpm-lock.yaml',
+  'package-lock.json',
+  'yarn.lock',
+  'bun.lock',
+  'bun.lockb',
+]
+
+// this function is for checking that a lockfile makes a folder
+// worthwhile watching. if the lockfile is empty, we don't
```
lukesandberg: i see why we are doing the empty check, but how useful is this in reality?    reading a giant lockfile to a string seems potentially quite slow

a `stat` call to check the size would be better but doesn't solve the 'only whitespace' case but do we really care?  size<2 might just work

packages/next/src/lib/find-root.ts
```diff
@@ -1,30 +1,174 @@
...
+    if (existsSync(join(directory, name))) {
+      return join(directory, name)
+    }
+  }
+
+  const packageJson = join(directory, 'package.json')
+  return hasWorkspacesField(packageJson) ? packageJson : undefined
+}
+
+function findWorkRoot(cwd: string) {
```
bgw: Maybe `findWorkRoot` isn't the best name here anymore? It's more like `findNextRootCandidate`?

packages/next/src/lib/find-root.ts
lukesandberg: i guess this is about `npm i` in random directory always writes a lockfile? but wouldn't the package.json check fix that?

packages/next/src/lib/find-root.ts
```diff
@@ -1,30 +1,174 @@
...
+    if (isDirectory === true) {
+      return { type: 'repository', dir }
+    }
+    if (isDirectory === false && isGitWorktree(dir, gitPath)) {
+      return { type: 'worktree', dir }
+    }
+
+    const parentDir = dirname(dir)
+    if (parentDir === dir) return undefined
+    dir = parentDir
```
lukesandberg: can we use `findUp` here as well

packages/next/src/lib/find-root.ts
```diff
@@ -34,6 +178,7 @@ function findWorkRoot(cwd: string) {
 export function findRootDirAndLockFiles(cwd: string): {
   lockFiles: string[]
   rootDir: string
+  boundary?: RootBoundary
 } {
   const lockFile = findWorkRoot(cwd)
```
bgw: I think we need to figure out how to better differentiate between lockfiles and workspace markers here.

If we encounter a workspace marker, do we even want to keep looking for more lockfiles in parent directories above that?

packages/next/src/lib/find-root.ts
sampoder:
```suggestion
// current working directory. we limit the project root to 
// inside a worktree or git repository. this is because if
// you use worktrees, you may have lock files from the 
// parent git repo that are in folders that are not relevant
// to watch.
```

packages/next/src/lib/find-root.ts
sampoder: we can drop it, or i could guard it with a quick stat call so if it is large we just assume it is not empty

sampoder: dropped it

packages/next/src/lib/find-root.ts
sampoder: Added a helper to do this, and called it on `homeDir` and the others being passed into `isStrictAncestor` and `isAncestorOrSelf`!

packages/next/src/lib/find-root.ts
sampoder: switched to `findNextRootCandidate`

packages/next/src/lib/find-root.ts
sampoder: Updated it so if we encounter a workspace marker, we stop there.

packages/next/src/lib/find-root.ts
```diff
@@ -1,48 +1,146 @@
...
-      'yarn.lock',
-      'bun.lock',
-      'bun.lockb',
-    ],
-    {
-      cwd,
+function findWorkspaceMarker(directory: string): string | undefined {
+  for (const name of WORKSPACE_MARKERS) {
+    if (existsSync(join(directory, name))) {
+      return join(directory, name)
```
bgw: nit
```suggestion
    const markerPath = join(directory, name)
    if (existsSync(markerPath)) {
      return markerPath
```

packages/next/src/lib/find-root.ts
```diff
@@ -1,48 +1,146 @@
...
+  try {
+    contents = readFileSync(gitFile, 'utf8')
+  } catch {
+    return false
+  }
+
+  const match = /^gitdir:\s*(.*)$/m.exec(contents)
+  if (!match) return false
+
+  const gitDir = resolve(dir, match[1].trim())
```
bgw: Extreme nitpick:
https://github.com/git/git/blob/13c7afec212fc97ce257d15601659314c6673d6c/setup.c#L1011-L1012

- It looks like `git` only expects exactly one leading space
- It looks like `git` only trims off trailing `\r` and `\n` characters from the end. A path could have trailing whitespace that's not stripped.
- Filenames can contain newline characters (use the `s` flag and not `m`)
- We should fail if the file path is empty

```suggestion
  // Match the behavior in git:
  // https://github.com/git/git/blob/13c7afec212fc97ce257d15601659314c6673d6c/setup.c#L1007-L1018
  const match = /^gitdir: (.+?)[\r\n]*$/s.exec(contents)
  if (!match) return false

  const gitDir = resolve(dir, match[1])
```

packages/next/src/lib/find-root.ts
```diff
@@ -51,44 +149,103 @@ export function findRootDirAndLockFiles(cwd: string): {
...
+    // here means there is no higher marker: nothing above can extend the root.
+    if (foundWorkspaceMarker && !nextCandidate.isWorkspaceMarker) break
+
+    lockFiles.push(nextCandidate.path)
+    foundWorkspaceMarker ||= nextCandidate.isWorkspaceMarker
+  }
+
+  // filter out lockfiles that would push the root past a
+  // git boundary or into the home directory.
+  const homeDir = toRealPath(homedir())
```
bgw: - Handle for if `os.homedir()` throws
- Ignore this value if it's an empty string, because that would make us think that the cwd is the homedir, which would make us emit a not-very-useful error claiming this.

packages/next/src/lib/find-root.ts
```diff
@@ -1,48 +1,146 @@
...
+function findGitBoundary(cwd: string): GitBoundary | undefined {
+  const gitDir = findUp.sync(
+    (dir) => {
+      const candidate = join(dir, '.git')
+      const isDirectory = statSync(candidate, {
+        throwIfNoEntry: false,
+      })?.isDirectory()
+
+      // a `.git` file is a worktree or a submodule; only a worktree is a boundary.
+      return isDirectory || isGitWorktree(dir, candidate) ? dir : undefined
```
bgw: Store `isGitWorktree` in a variable outside of this closure so that below we don't have to call `statSync(join(gitDir, '.git'))` a second time.

packages/next/src/lib/find-root.ts
```diff
@@ -1,48 +1,146 @@
...
-  if (!lockFile)
+  const initialCandidate = findNextRootCandidate(cwd)
+  if (!initialCandidate)
     return {
       lockFiles: [],
       rootDir: cwd,
     }
 
-  const lockFiles = [lockFile]
+  const lockFiles = [initialCandidate.path]
```
bgw:
```suggestion
  const candidates = [initialCandidate.path]
```

packages/next/src/lib/find-root.ts
```diff
@@ -51,44 +149,103 @@ export function findRootDirAndLockFiles(cwd: string): {
...
+    lockFiles.push(nextCandidate.path)
+    foundWorkspaceMarker ||= nextCandidate.isWorkspaceMarker
+  }
+
+  // filter out lockfiles that would push the root past a
+  // git boundary or into the home directory.
+  const homeDir = toRealPath(homedir())
+  const gitBoundary = findGitBoundary(cwd)
+
+  const acceptedLockFiles: string[] = []
```
bgw:
```suggestion
  const acceptedCandidates: string[] = []
```

packages/next/src/lib/find-root.ts
```diff
@@ -51,44 +149,103 @@ export function findRootDirAndLockFiles(cwd: string): {
...
+  const config = process.env.TURBOPACK
+    ? 'turbopack.root'
+    : 'outputFileTracingRoot'
+  const reason =
+    type === 'repository'
+      ? `it is outside the current Git repository (${boundary})`
+      : `it would include your home directory (${boundary})`
+
+  Log.warnOnce(
+    `Warning: Next.js ignored a workspace marker in ${root} because ${reason}.\n` +
```
bgw: Could we say what the workspace marker was? The filename?

bgw (approved):

PR #96170, opened 2026-07-24
[turbopack] Scope-hoist CommonJS modules that require CommonJS

sampoder:
The goal of this PR is to support scope-hoisting with CJS modules. The goal is to take:

```javascript
// math.js
exports.add = (a, b) => a + b
exports.PI = 3.14

// index.js
const { add } = require('./math')
add(1, 2)
```

And transform it into one file:

```javascript
var _add, _PI;
_add = (a, b) => a + b;
_PI = 3.14;

const { add } = { add: _add };
add(1, 2);
```

To be considered `hoistable` the require has to be a top-level require. This is because if it is was conditional, we may not always want to be fetching that module or if it was in a function, the intention may not be for that module to be loaded when this module is loaded.

And then this PR does the following rewrite so the import uses `__turbopack_merged_module__` (previously `__turbopack_merged_esm__`, yay!):

```javascript
_add = (a, b)=>a + b;
_PI = 3.14;
__turbopack_context__.v({ add: _add, PI: _PI }, "math.js");
const { add: add1, PI: PI1 } = __turbopack_merged_module__(0, {
    add: _add,
    PI: _P
});
```


turbopack/crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -177,6 +177,13 @@ pub trait MergeableModule: Module {
         Vc::cell(true)
     }
 
+    /// Whether the module is ESM or CommonJS. Only modules of the same kind are merged
+    /// into a group, as mixing them requires interop between the two export shapes.
+    #[turbo_tasks::function]
+    fn merge_kind(self: Vc<Self>) -> Vc<MergeableModuleKind> {
+        MergeableModuleKind::EcmaScript.cell()
+    }
```
lukesandberg, line 179: maybe merge these two as an `OptionMergableModuleKind`?

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
lukesandberg, line 1860: we propogate `in_try` from the visitor

perhaps we should also propagate a `top_level` which is easy for the visitor to compute?

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -201,9 +290,62 @@ impl CjsRequireAssetReferenceCodeGen {
...
+        {
+            // A bare `require(...)` statement (not a binding initializer) discards its
+            // result, so inline the body only (one-arg marker); a value-position require
+            // also substitutes a `<value>` for the call.
+            let is_bare_statement = !self
+                .path
+                .iter()
+                .any(|kind| matches!(kind, AstParentKind::VarDeclarator(_)));
+
+            let id: Expr = Lit::Num(index.into()).into();
```
lukesandberg:
```suggestion
            let id = Expr::from(Lit::Num(index.into()));
```

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -128,10 +136,91 @@ impl CjsRequireAssetReference {
...
+        .export_names
+        .iter()
+        .filter(|name| match usage {
+            ExportUsage::All => true,
+            ExportUsage::Named(used) => *name == used,
+            ExportUsage::PartialNamespaceObject(used) => used.contains(name),
+            ExportUsage::Evaluation => false,
+        })
+        .cloned()
+        .collect())
```
lukesandberg, line 171: if i use a name that isn't actually exported, today i get undefined, will that still work?

critically we don't report errors if we do not see the export in a target

turbopack/crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -1878,6 +2051,223 @@ async fn process_parse_result(
...
+                impl CjsExportRewriter {
+                    /// Rewrites `module.exports = { a, b: c }` into `(_a = a, _b = c)`. A
+                    /// property with no local keeps only its value, for the side effects.
+                    fn rewrite_whole_exports_object(
+                        &self,
+                        assign: &mut AssignExpr,
+                    ) -> Option<Expr> {
+                        let obj = as_module_exports_object_literal(assign, self.unresolved_mark)?;
+                        // Each property must name one export with an eager value, or be the
+                        // `...void <expr>` spread a dropped export leaves behind.
```
lukesandberg, line 2185: 😹

turbopack/crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -1878,6 +2051,223 @@ async fn process_parse_result(
...
+                                span: DUMMY_SP,
+                                exprs,
+                            }),
+                        })
+                    }
+                }
+
+                /// `var _A, _B, …;` declaring a module's per-export locals for the writes to
+                /// assign.
+                fn cjs_locals_declaration(locals: &[(RcStr, Ident)]) -> Stmt {
```
lukesandberg, line 2254: why do this in a separate declaration instead of just rewriting `exports.foo = 1` => `var _foo = 1`  and doing the declaration there?

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
sampoder: yes that'll still work.

```
const { a, nope } = require('./m')
const ns = require('./m')
const viaNamespace = ns.alsoMissing
```

becomes:

```
var _a;
_a = 1;
__turbopack_context__.v({ a: _a }, "…/m.js");
const { a: a1, nope } = { a: _a };
const ns = { a: _a };
const viaNamespace = ns.alsoMissing;
```

where `nope` is the name that isn't actually exported

turbopack/crates/turbopack-ecmascript/src/lib.rs
sampoder: it's not always safe to rewrite `exports.foo = 1` => `var _foo = 1` directly, for example, `const x = (exports.d = 3)` can't be rewritten to `const x = (var _d = 3)`

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -128,10 +136,91 @@ impl CjsRequireAssetReference {
...
+                })))
+            })
+            .collect(),
+    })
+}
+
+/// Whether a `require("…")` at this AST path may be scope-hoisted. To be scope
+/// hoisted, the require must be top-level.
+pub(crate) fn is_hoistable_require_path(path: &[AstParentKind]) -> bool {
+    path.iter().all(|kind| {
```
lukesandberg: this probably needs to be positional

we say it can be an Expr as long as it is immediately in an ExprStatement or the rhs of a var decl

Maybe it just has to be Expr(ExprField::Call)?

lukesandberg: you might also need to match on AstParentKind::CallExpr? or maybe not

so maybe just Expr is wrong

turbopack/crates/turbopack-ecmascript/src/lib.rs
lukesandberg: fair

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
sampoder: Just updated this to be stricter and also a little more readable imo!

turbopack/crates/turbopack-core/src/module_graph/merged_modules.rs
```diff
@@ -131,22 +131,27 @@ pub async fn compute_merged_modules(module_graph: Vc<ModuleGraph>) -> Result<Vc<
...
-            .await?
+            .await?;
+
+        let merge_kinds: FxHashMap<_, _> = mergeable_modules.iter().copied().collect();
+
+        let mergeable: FxHashSet<_> = mergeable_modules
             .into_iter()
-            .collect::<FxHashSet<_>>();
+            .map(|(module, _)| module)
+            .collect();
```
lukesandberg, line 154: i don't think we need two datastrucures here

`mergable` can be a Map from module -> kind

and the uses below can either iterate the keys() collection or call contains_key on the map

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -742,25 +742,26 @@ mod analyzer_state {
...
-                                    names.as_slice()
-                                }
-                            })
-                            .cloned()
-                            .collect(),
-                        has_es_module: c.has_es_module,
-                    }
+            // the descriptor, which the merge has no local to bind it to. A sloppy-mode
+            // module can't be inlined into the always-strict merged factory.
+            let static_exports = (!c.has_top_level_return
```
lukesandberg, line 747:
```suggestion
            // the descriptor, which the merge has no local to bind it to. A sloppy-mode
            // module can't be inlined into the always-strict merged factory.  Merged modules
            // could allow for non-strict merge groups but we don't believe this is important so
            // it isn't currently supported
            let static_exports = (!c.has_top_level_return
```

turbopack/crates/turbopack-ecmascript/src/references/cjs.rs
```diff
@@ -106,9 +112,13 @@ pub struct CjsRequireAssetReference {
     resolve_override: Option<ResolvedVc<Box<dyn Module>>>,
     usage: ExportUsage,
     cjs_tree_shaking: bool,
+    /// Whether this `require(...)` is a top-level, unconditional call, so its
+    /// target may be scope-hoisted into the requiring module.
```
lukesandberg, line 116: discussed IRL

the condition on this needs to be tigher to prevent hoisting

```
const a = require(...);
exports.foo = ....
const b = require(...);
````

we could only consider adding b to the hoistable requires if its target is side effect free,

the simpler approach for now would be to only consider requires in prefix position in the file

sampoder: we replace the require with the module in place so this is ok

PR #96174, opened 2026-07-24
Fulfill route cache entries with fresh objects to avoid map deprecation

acdlite:
## Root cause

`fulfillRouteCacheEntry` fulfilled a pending `RouteCacheEntry` by mutating its null-initialized fields in place (`tree`, `canonicalUrl`, `metadata`, `renderedSearch`, ...) and adding a new `hasDynamicRewrite` field. The first fulfillment in an isolate generalized the pending shape's field representations (null → object/string), deprecating the hidden class shared by all route cache entries. Every optimized reader of route entries then hit a one-time wrong-map deopt, and readers of fulfilled entries stayed permanently polymorphic because entries fulfilled in place had a different hidden class than the fulfilled-entry literals in `matchKnownRoute` and `deprecated_requestOptimisticRouteCacheEntry`.

Observed with `pnpm bench:deopt --scenario segment-cache`: `map,Deprecate` of the route entry chain fires mid-workload, immediately followed by the `hasDynamicRewrite` LoadIC going polymorphic (`1->P`), plus polymorphic ICs on `tree`, `metadata`, `status`, `canonicalUrl`, `renderedSearch`, `staleAt`, `couldBeIntercepted`, `supportsPerSegmentPrefetching`, and `version` across `navigation.ts`, `scheduler.ts`, and `optimistic-routes.ts`.

## Fix

- `fulfillRouteCacheEntry` now constructs the fulfilled entry as a fresh object and swaps it into the pending entry's cache map slot via a new `replaceCacheMapValue` helper (transfers the `ref` back-pointer and LRU slot; no-op if the entry was detached or lazily evicted). The pending shape is never generalized.
- All fulfilled entries are allocated through a single factory, `createFulfilledRouteCacheEntry` (also used by `matchKnownRoute` and `deprecated_requestOptimisticRouteCacheEntry`), so every fulfilled entry shares one hidden class regardless of how it was created.
- The `InliningHintsStale` already-expired timestamp is `now - 1` instead of `-1`, so `staleAt` is always a double and never flips the field's representation from Smi.
- Scope is route entries only. Segment entries still fulfill in place and are a possible follow-up.

## Identity-swap safety audit

Every holder of a pending route entry across the fulfillment boundary was audited:

- **Cache map slot + `ref` back-pointer** (`cache-map.ts`): the swap choke point; handled by `replaceCacheMapValue`.
- **`fetchRouteOnCacheMiss` local `entry`**: the only caller holding the pending entry across the boundary. Its single post-fulfillment use (the `!couldBeIntercepted` re-key) now uses the entry returned from `discoverKnownRoute` / `writeDynamicTreeResponseIntoCache` (which now returns the latest entry; on rejection it returns the original entry, matching the previous re-key behavior).
- **`blockedTasks` waiters**: tasks never hold the entry; `pingBlockedTasks` consumes the old entry's set and re-queued tasks re-read via `readOrCreateRouteCacheEntry` on the next ping.
- **Scheduler locals** (`pingRoute`, `pingRootRouteTree`): used synchronously within a single ping; fulfillment only happens in async continuations, so no interleaving. Status mutations in the Empty branch stay on the pending shape (Smi/number stores, no generalization).
- **Navigation** (`navigation.ts`): reads the map synchronously at navigation time; pending entries are not retained.
- **`knownRoutePart.pattern`** (`optimistic-routes.ts`): only ever assigned fulfilled entries, always from the factory's return value.
- Rejection (`rejectRouteCacheEntry`) keeps identity: it only writes `status`/`staleAt` (Smi/double into existing fields), which does not change the hidden class.
- No WeakMap/WeakSet keys, no `===` identity comparisons on route entries, no bfcache or navigation-testing-lock references (those track segment entries only), and no consumers outside `segment-cache/` + `router-reducer/` handle route entry objects.

## Before/after (bench:deopt, segment-cache scenario)

```
BEFORE (canary + in-place fulfillment)
  Findings: 102 matching filters (ic-polymorphic: 88)
  hasDynamicRewrite LoadIC: 0->1, then 1->P (two maps) right after map,Deprecate of the route entry chain mid-workload
  Polymorphic ICs on fulfilled-entry fields in navigation.ts, scheduler.ts, optimistic-routes.ts:
    canonicalUrl, tree, metadata, status, renderedSearch, hasDynamicRewrite,
    couldBeIntercepted, supportsPerSegmentPrefetching, staleAt, size, version

AFTER (this PR)
  Findings: 80 matching filters (ic-polymorphic: 66-69 across runs)
  hasDynamicRewrite LoadIC: 0->1 only (single map, monomorphic for the whole workload)
  All fulfilled-entry field ICs above: cleared (monomorphic)
  Route-entry map deprecations during the workload: none
    (only the init-time literal boilerplate setup at first allocation, before any code is optimized)
```

The `matchKnownRoutePart wrong map` line still appears in the findings snapshot, but its cause changed: the V8 log shows the IC stays monomorphic and no route-entry map is deprecated during the workload. The residual 1x deopt is the first-allocated instance lazily migrating from the literal boilerplate's initial map chain (an unavoidable per-isolate warm-up artifact of object literals), not the recurring deprecation + polymorphism this PR removes. `isValueExpired wrong map` also remains: that site is inherently polymorphic across route/segment/bfcache entry kinds and was not triggered by route entry deprecation alone. No new high-severity findings.

## Tests

- `test/e2e/app-dir/segment-cache/basic` (test-start-turbo): 10 passed, 1 skipped
- `test/e2e/app-dir/segment-cache/revalidation` (test-start-turbo): 6 passed
- `pnpm --filter=next types`: pass

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache-map.ts
+++ b/packages/next/src/client/components/segment-cache/cache-map.ts
@@ -418,6 +418,38 @@ function setMapEntryValue(entry: UnknownMapEntry, value: MapValue): void {
   }
 }
 
+/**
+ * Replaces the value stored in the cache map with a new value, in place. The
+ * new value takes over the old value's slot (and its position in the LRU),
+ * and the `ref` back-pointer is transferred.
+ *
+ * This is used when fulfilling a pending entry: rather than mutating the
+ * pending object's fields in place (which would change the pending object's
+ * hidden class), the fulfilled entry is constructed as a fresh object and
+ * swapped into the map slot.
+ *
+ * The caller is responsible for carrying over `size` to the new value so the
+ * LRU accounting stays consistent.
+ *
+ * No-op if the old value is not (or is no longer) in the map, e.g. because it
+ * was lazily evicted while a request was in flight.
+ */
+export function replaceCacheMapValue(
+  oldValue: MapValue,
+  newValue: MapValue
+): void {
+  const entry = oldValue.ref
+  oldValue.ref = null
+  if (entry === null || entry.value !== oldValue) {
+    // The old value is not a member of any map. (The second check handles a
+    // stale back-pointer: lazy eviction clears the map entry's value without
+    // clearing the value's `ref`.)
+    return
+  }
+  entry.value = newValue
+  newValue.ref = entry
+}
+
 export function deleteFromCacheMap(value: MapValue): void {
   const entry = value.ref
   if (entry === null) {
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -80,6 +80,7 @@ import {
   setInCacheMap,
   setSizeInCacheMap,
   deleteFromCacheMap,
+  replaceCacheMapValue,
   isValueExpired,
   EntryStatus,
   type CacheMap,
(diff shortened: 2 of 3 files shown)


acdlite:
Closing after cost/benefit discussion. The deopt this fixes is a one-time-per-isolate warmup event: the pending-entry map is deprecated on the first fulfillment, one optimized function takes a single wrong-map bailout, and after re-optimization the steady state is effectively monomorphic again (all later entries are born on the migrated map). The residual cost is a couple of nanoseconds of extra map comparison on paths that run at prefetch/navigation frequency — not measurable in any user-facing metric.

Against that, the fix introduces a permanent identity-swap invariant: a route entry's object identity changes at fulfillment, and any future code that holds an entry reference across an await must know to re-read it. The audit in this PR shows current code is safe, but that's a subtle correctness constraint to carry forever for an unmeasurable win. The shape-unification benefits largely overlap #96164, which achieves them without changing entry identity.

The identity-swap audit and the raw V8 log analysis here remain useful reference material if fulfillment-related deopt churn ever shows up as a recurring (rather than one-time) event in bench/deopt runs.

PR #96179, opened 2026-07-24
Skip reader task locking for immutable Turbo Tasks reads

marcoshernanz:
## What changed

Tracked output and cell reads now acquire the target task first. If the target is already immutable, the read skips locking the reader task because no dependency edge can be added and the target cannot be invalidated. Mutable targets drop the speculative target lock and reacquire the target/reader pair together, preserving the existing invalidation-race guarantee.

`verify_immutable` deliberately retains the pair-lock path. Listener diagnostics also preserve the tracked reader ID when the fast path skips its guard.

The separate current-task ID cache is in #96180. This PR is now backend-only so the lock-ordering change can receive focused backend review.

## Why

The previous path locked both the target and reader task for every tracked read. That synchronization is necessary for mutable targets but redundant once the target is immutable. Instrumentation found that 1,541,880 of 4,501,364 sampled tracked reads (34.254%) could use the bypass.

## Performance evidence

Measured at `0950e8ae05db20c33302f02257114f5b857edabc` on a dedicated 8-vCPU/16-GiB Vercel Sandbox with frozen side-by-side binaries.

These measurements are directional evidence, not precise estimates. Small percentage changes are difficult to resolve reliably. This campaign and #96180's incremental campaign used separate run sets: the absolute endpoint here does not match #96180's starting point, which demonstrates cross-campaign noise. The values must not be compared across PRs, and their percentages must not be added.

### Browser-driven HMR, 30-second B/C/C/B

- Evaluation: 20.771360 → 19.176645 ms (**-7.677%**)
- Commit: 21.714832 → 20.371425 ms (**-6.187%**)

### Production builds

- First 12-campaign matrix: **-1.931%**
- Complementary 12-campaign matrix: **-1.990%**
- Combined 24-campaign aggregate: **-1.959%**

Five of six balanced blocks favored the candidate. The conservative six-block 95% interval was -4.042% to +0.248%, so the production estimate is supporting evidence rather than a precise claim.

### Causal profiles

- HMR `Storage::access_pair_mut` self time fell from 3.19%/3.19% to 1.21%/1.18%.
- HMR voluntary context switches fell 32.898%.
- Production pair-access self time and voluntary context switches also fell; production voluntary switches changed -14.552%.
- All accepted perf captures reported zero lost samples.

## Validation

On the exact backend-only head (`9ccfd6753526f716466a9584fba8c8753b7d83bf`):

- Complete `turbo-tasks-backend` package tests, integrations, and doctests
- Complete backend package matrix with `--features verify_immutable`
- `cargo fmt --check -p turbo-tasks-backend`
- `git diff --check`
- Conflict-free merge-tree check against current `canary` (`5cdcbbb05b`)

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -679,7 +679,12 @@ impl TurboTasksBackend {
 
         let reader_description = reader_task
             .as_ref()
-            .map(|r| EventDescription::new(|| r.get_task_desc_fn()));
+            .map(|r| EventDescription::new(|| r.get_task_desc_fn()))
+            .or_else(|| {
+                need_reader_task.map(|reader_id| {
+                    EventDescription::new(move || move || format!("{reader_id:?}"))
+                })
+            });
         if let Some(value) = check_in_progress(&task, reader_description.clone(), options.tracking)
         {
             return value;
@@ -825,11 +830,16 @@ impl TurboTasksBackend {
         } = options;
 
         let mut ctx = self.execute_context(turbo_tasks);
-        let (mut task, reader_task) = if self.should_track_dependencies()
+        let need_reader_task = if self.should_track_dependencies()
             && !matches!(tracking, ReadCellTracking::Untracked)
             && let Some(reader_id) = reader
             && reader_id != task_id
         {
+            Some(reader_id)
+        } else {
+            None
+        };
+        let (mut task, reader_task) = if let Some(reader_id) = need_reader_task {
             // Immutable tasks never need dependency edges and can never be invalidated. Avoid
             // locking the reader too in that common case. When the task is still mutable, drop
             // the speculative lock and reacquire both locks together to preserve the invalidation
@@ -867,7 +877,7 @@ impl TurboTasksBackend {
             Some(InProgressState::InProgress(..) | InProgressState::Scheduled { .. })
         ) {
             return Ok(Err(self
-                .listen_to_cell(&mut task, task_id, &reader_task, cell)
+                .listen_to_cell(&mut task, task_id, need_reader_task, &reader_task, cell)
                 .0));
         }
         let is_cancelled = matches!(in_progress, Some(InProgressState::Canceled));
@@ -901,7 +911,8 @@ impl TurboTasksBackend {
         }
 
         // Listen to the cell and potentially schedule the task
-        let (listener, new_listener) = self.listen_to_cell(&mut task, task_id, &reader_task, cell);
+        let (listener, new_listener) =
+            self.listen_to_cell(&mut task, task_id, need_reader_task, &reader_task, cell);
         drop(reader_task);
         if !new_listener {
             return Ok(Err(listener));
@@ -927,6 +938,7 @@ impl TurboTasksBackend {
         &self,
         task: &mut impl TaskGuard,
         task_id: TaskId,
+        reader: Option<TaskId>,
         reader_task: &Option<impl TaskGuard>,
         cell: CellId,
     ) -> (EventListener, bool) {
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -452,11 +452,18 @@ impl TurboTasksBackend {
             None
         };
         let (mut task, mut reader_task) = if let Some(reader_id) = need_reader_task {
-            // Having a task_pair here is not optimal, but otherwise this would lead to a race
-            // condition. See below.
-            // TODO(sokra): solve that in a more performant way.
```
bgw, line 457: You can keep the TODO here, it's still relevant, just only for the mutable case.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -818,16 +830,28 @@ impl TurboTasksBackend {
...
+            if task.immutable() && !cfg!(feature = "verify_immutable") {
+                (task, None)
+            } else {
+                drop(task);
+                let (task, reader) = ctx.task_pair(task_id, reader_id, TaskDataCategory::All);
+                (task, Some(reader))
+            }
         } else {
             (ctx.task(task_id, TaskDataCategory::All), None)
         };
```
bgw: Hmm... This is the same block of code? Maybe we should follow up by refactoring into a helper?

bgw:
I have less direct experience with the backend, but the changes to `turbopack/crates/turbo-tasks/src/manager.rs` LGTM, and could be split out into a separate PR.

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -576,12 +576,38 @@ impl CurrentTaskState {
...
 }
 
+/// A shareable current-task state handle with the immutable task ID cached
+/// outside the lock. The rest of the state is mutated by global and local
+/// tasks, but the task ID is fixed for the lifetime of an execution.
+#[derive(Clone)]
+struct CurrentTaskStateHandle {
+    current_task_id: Option<TaskId>,
+    state: Arc<RwLock<CurrentTaskState>>,
+}
```
bgw: *Very* minor improvement: Move the `current_task_id` inside the `Arc`, but still outside of the `RwLock`.

marcoshernanz:
Addressed @bgw's feedback:

- Split the `turbo-tasks/src/manager.rs` change back into #96180 and requested @bgw there.
- This PR is backend-only again, with @sokra requested for the lock-ordering review.
- Rewrote both performance sections to make the ~1–2% measurement uncertainty, cross-campaign endpoint mismatch, and non-additive percentages explicit.
- Revalidated both standalone branches with their full package tests, doctests, formatting, clippy, diff checks, and current-canary merge checks.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -818,19 +842,17 @@ impl TurboTasksBackend {
...
-            // Having a task_pair here is not optimal, but otherwise this would lead to a race
-            // condition. See below.
-            // TODO(sokra): solve that in a more performant way.
-            let (task, reader) = ctx.task_pair(task_id, reader_id, TaskDataCategory::All);
-            (task, Some(reader))
+            Some(reader_id)
         } else {
-            (ctx.task(task_id, TaskDataCategory::All), None)
+            None
         };
```
sokra: `(...).then_some(reader_id)`

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
sokra: You can drop this check now, as reader_task will be None for immutable tasks

sokra (approved):

PR #96180, opened 2026-07-24
Cache immutable current task IDs outside the task-state lock

marcoshernanz:
## What changed

Introduce a cloneable `CurrentTaskStateHandle` that keeps the execution's immutable `Option<TaskId>` beside the existing shared `Arc<RwLock<CurrentTaskState>>`.

Global tasks, local tasks, detached futures, and top-level runs continue to share the same mutable state and task-local scope. Only current-task ID reads bypass the lock; all other state remains behind the existing `RwLock`.

This was split from #96179 following review feedback. It is source-independent and changes only `turbopack/crates/turbo-tasks/src/manager.rs`.

## Why

`current_task_if_available` and `current_task` acquired a shared standard-library `RwLock` on every native function call and tracked read, even though the task ID is fixed for the entire execution. The lock and its contention path remained visible in both HMR and production profiles.

## Performance evidence

Measured at `0950e8ae05db20c33302f02257114f5b857edabc` as an incremental change on top of the independently proposed immutable-read bypass in #96179.

These measurements are directional evidence, not a precise standalone estimate. Changes around 1–2% are difficult to measure reliably. The #96179 and #96180 campaigns used separate run sets, and #96179's absolute endpoint does not match this campaign's starting point; that mismatch demonstrates cross-campaign noise. The percentages must not be added, and the absolute values must not be compared across the two PRs. The standalone GitHub aggregate benchmark also moved in the opposite direction, reinforcing that build-level percentages are inconclusive at this scale.

### Browser-driven HMR, 30-second B/C/C/B

- Evaluation: 19.772807 → 19.356650 ms (**-2.105%**)
- Commit: 20.632962 → 20.167726 ms (**-2.255%**)
- Both candidate runs beat both controls for both metrics.

### Production builds

- Matrix 1: **-1.804%**
- Complementary matrix 2: **-1.474%**
- Combined 24 campaigns: 1,641.839610 → 1,614.918617 ms (**-1.640%**)
- All six balanced blocks favored the candidate.
- Six-block mean: -1.636%, with 95% interval **[-2.718%, -0.555%]**.

### Causal profiles

- The `current_task_if_available` helper and contended read-lock symbols disappeared from candidate profiles.
- HMR voluntary context switches fell 2.700%.
- Production voluntary context switches fell 2.576%.
- All eight accepted perf captures reported zero lost samples.

## Validation

On the exact standalone head (`bce272743ff44021fda204d98924bf8519cd8c1d`):

- Complete `turbo-tasks` package tests and doctests
- Complete `turbo-tasks-backend` package tests, integrations, and doctests
- `cargo fmt --check -p turbo-tasks -p turbo-tasks-backend`
- `cargo clippy -p turbo-tasks -p turbo-tasks-backend --all-targets -- -D warnings`
- `git diff --check`
- Conflict-free merge-tree check against current `canary` (`91c6309c52`)

### DIFF
--- a/turbopack/crates/turbo-tasks/src/manager.rs
+++ b/turbopack/crates/turbo-tasks/src/manager.rs
@@ -465,7 +465,7 @@ enum ScheduledTask {
         ty: LocalTaskSpec,
         persistence: TaskPersistence,
         local_task_id: LocalTaskId,
-        global_task_state: Arc<RwLock<CurrentTaskState>>,
+        global_task_state: CurrentTaskStateHandle,
         span: Span,
     },
 }
@@ -576,12 +576,38 @@ impl CurrentTaskState {
     }
 }
 
+/// A shareable current-task state handle with the immutable task ID cached
+/// outside the lock. The rest of the state is mutated by global and local
+/// tasks, but the task ID is fixed for the lifetime of an execution.
+#[derive(Clone)]
+struct CurrentTaskStateHandle {
+    current_task_id: Option<TaskId>,
+    state: Arc<RwLock<CurrentTaskState>>,
+}
+
+impl CurrentTaskStateHandle {
+    fn new(state: CurrentTaskState) -> Self {
+        Self {
+            current_task_id: state.task_id,
+            state: Arc::new(RwLock::new(state)),
+        }
+    }
+}
+
+impl std::ops::Deref for CurrentTaskStateHandle {
+    type Target = RwLock<CurrentTaskState>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.state
+    }
+}
+
 // TODO implement our own thread pool and make these thread locals instead
 task_local! {
     /// The current TurboTasks instance
     static TURBO_TASKS: Arc<dyn TurboTasksApi>;
 
-    static CURRENT_TASK_STATE: Arc<RwLock<CurrentTaskState>>;
+    static CURRENT_TASK_STATE: CurrentTaskStateHandle;
 
     /// Temporarily suppresses the eventual consistency check in top-level tasks.
     /// This is used by strongly consistent reads to allow them to succeed in top-level tasks.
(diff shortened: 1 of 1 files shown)


marcoshernanz:
Consolidated into #96179. That PR now contains both this immutable current-task ID cache and the immutable-read reader-lock bypass, plus the tracked-reader diagnostics fix. It also preserves this PR's incremental benchmark results and validation context.

marcoshernanz:
Reopened after splitting the combined change on review feedback. This PR is again the standalone `turbo-tasks/src/manager.rs` current-task ID cache; #96179 now contains only the backend reader-lock bypass.

The description now explicitly treats the benchmark percentages as directional and noisy, notes that the two campaign endpoints are not directly comparable, and avoids any additive performance claim.

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -576,12 +576,38 @@ impl CurrentTaskState {
...
+impl CurrentTaskStateHandle {
+    fn new(state: CurrentTaskState) -> Self {
+        Self {
+            current_task_id: state.task_id,
+            state: Arc::new(RwLock::new(state)),
+        }
+    }
+}
+
+impl std::ops::Deref for CurrentTaskStateHandle {
```
bgw: Import `std::ops::Deref` at the top of the file.

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -576,12 +576,38 @@ impl CurrentTaskState {
...
 }
 
+/// A shareable current-task state handle with the immutable task ID cached
+/// outside the lock. The rest of the state is mutated by global and local
+/// tasks, but the task ID is fixed for the lifetime of an execution.
+#[derive(Clone)]
+struct CurrentTaskStateHandle {
+    current_task_id: Option<TaskId>,
+    state: Arc<RwLock<CurrentTaskState>>,
+}
```
bgw, line 591: (comment copied from other PR) *Very* minor improvement: Move the `current_task_id` inside the `Arc`, but still outside of the `RwLock`.

bgw (approved):

PR #96221, opened 2026-07-25
Fix dev overlay symbolication for project paths needing percent-encoding

gaearon:
In a project whose absolute path contains characters that percent-encode in URLs — e.g. a space — the dev overlay showed no code frame and raw `file://` URLs instead of source locations for server frames.

Two encoding mismatches caused this:

- Turbopack's `trace_source` percent-decoded the source map's original file URL before comparing it against the still-encoded project root URI, so the containment check failed ("Original file ... outside project") for any project path with such characters. The comparisons now stay in the encoded domain and only the outputs are decoded back into paths.

- React synthesizes stack frame `file:` URLs by prepending the scheme to a filesystem path, so the overlay receives URLs with e.g. raw spaces. These aren't well-formed, and they don't match the `url.pathToFileURL` form Node.js keys its source map cache by. The overlay middleware now re-encodes them through WHATWG URL parsing, which tolerates such input.

## Test Plan

New tests.

Known cases that remain broken are commented out with a TODO.
These could be resolved if https://github.com/react/react/pull/37105 or equivalent lands.

## Before

Screenshot 2026-07-25 at 14 39 18

## After

Screenshot 2026-07-25 at 14 58 14

### DIFF
--- a/crates/next-napi-bindings/src/next_api/project.rs
+++ b/crates/next-napi-bindings/src/next_api/project.rs
@@ -2343,10 +2343,8 @@ async fn project_trace_source_operation(
 
     let (original_file, line, column, method_name, is_ignored) = match token {
         Token::Original(token) => (
-            match urlencoding::decode(&token.original_file)? {
-                Cow::Borrowed(_) => token.original_file,
-                Cow::Owned(original_file) => RcStr::from(original_file),
-            },
+            // Still percent-encoded, like the URIs it's compared against.
+            token.original_file,
             // JS stack frames are 1-indexed, source map tokens are 0-indexed
             Some(token.original_line + 1),
             Some(token.original_column + 1),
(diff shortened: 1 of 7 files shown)


test/e2e/app-dir/server-source-maps/special-project-paths.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  })
+}
+
+createSuite('space dir')
+createSuite('ünïcode-dir')
+createSuite('bracket [dir]')
+// TODO(veil): Make these pass. Both need React's fake stack frame filename
+// virtualization to round-trip losslessly.
+// createSuite('percent%20dir')
+// createSuite('hash#dir')
```
unstubbable: Can we instead mark them with `it.failing` so CI will remind us to enable them in the PR for the React sync?

unstubbable (approved):

PR #96226, opened 2026-07-25
Turbopack: support import.meta.glob caseSensitive option

timneutkens:
## Summary

Add Vite-compatible `caseSensitive` support to `import.meta.glob()`. Matching remains case-sensitive by default, while `caseSensitive: false` applies case-insensitive matching to directory traversal, filenames, positive patterns, and negative patterns.

The option is included in matcher equality and non-default virtual module identity, and is exposed through diagnostics, TypeScript declarations, and the Turbopack options reference.

## Verification

- `cargo test -p turbo-tasks-fs glob`
- `cargo nextest run -p turbopack-tests -E "test(import_meta_glob) and not test(import_meta_glob_errors)"`
- `pnpm build-all` and `pnpm --filter=next types` are blocked by current-canary `sharp.default` type errors in `packages/next/src/server/image-optimizer.ts`

### DIFF
--- a/docs/01-app/03-api-reference/08-turbopack.mdx
+++ b/docs/01-app/03-api-reference/08-turbopack.mdx
@@ -234,12 +234,13 @@ const eager = import.meta.glob('./dir/*.ts', { eager: true })
 
 ### Options reference
 
-| Option   | Type                                          | Default     | Description                                                            |
-| -------- | --------------------------------------------- | ----------- | ---------------------------------------------------------------------- |
-| `eager`  | `boolean`                                     | `false`     | Import modules synchronously instead of returning thunks.              |
-| `import` | `string`                                      | `undefined` | Named export to select from each matched module (e.g. `'default'`).    |
-| `query`  | `string \| Record<string, string \| boolean>` | `undefined` | Query string (or object) to append to each import.                     |
-| `base`   | `string`                                      | `undefined` | Override the base path used for resolving patterns and keying results. |
+| Option          | Type                                          | Default     | Description                                                            |
+| --------------- | --------------------------------------------- | ----------- | ---------------------------------------------------------------------- |
+| `eager`         | `boolean`                                     | `false`     | Import modules synchronously instead of returning thunks.              |
+| `import`        | `string`                                      | `undefined` | Named export to select from each matched module (e.g. `'default'`).    |
+| `query`         | `string \| Record<string, string \| boolean>` | `undefined` | Query string (or object) to append to each import.                     |
+| `base`          | `string`                                      | `undefined` | Override the base path used for resolving patterns and keying results. |
+| `caseSensitive` | `boolean`                                     | `true`      | Match glob patterns case-sensitively. Set to `false` to ignore case.   |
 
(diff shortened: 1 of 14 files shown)


turbopack/crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -40,7 +40,7 @@ pub struct Glob {
 
 impl PartialEq for Glob {
     fn eq(&self, other: &Self) -> bool {
-        self.glob == other.glob
+        self.glob == other.glob && self.opts == other.opts
```
lukesandberg, line 43: 🤦‍♂️

docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -234,12 +234,13 @@ const eager = import.meta.glob('./dir/*.ts', { eager: true })
...
+| Option          | Type                                          | Default     | Description                                                            |
+| --------------- | --------------------------------------------- | ----------- | ---------------------------------------------------------------------- |
+| `eager`         | `boolean`                                     | `false`     | Import modules synchronously instead of returning thunks.              |
+| `import`        | `string`                                      | `undefined` | Named export to select from each matched module (e.g. `'default'`).    |
+| `query`         | `string \| Record<string, string \| boolean>` | `undefined` | Query string (or object) to append to each import.                     |
+| `base`          | `string`                                      | `undefined` | Override the base path used for resolving patterns and keying results. |
+| `caseSensitive` | `boolean`                                     | `true`      | Match glob patterns case-sensitively. Set to `false` to ignore case.   |
```
lukesandberg: specify that this is only 'ascii' case insentive

turbopack/crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -152,8 +154,9 @@ impl Glob {
     }
 }
 
-fn new_regex(pattern: &str) -> Regex {
+fn new_regex(pattern: &str, opts: GlobOptions) -> Regex {
     RegexBuilder::new(pattern)
+        .case_insensitive(opts.case_insensitive)
```
lukesandberg, line 160:
```suggestion
        // Because we aren't setting the `unicode` flag, this is only `ASCII` case insensitive
        .case_insensitive(opts.case_insensitive)
```

lukesandberg (approved):
nice!

turbopack/crates/turbo-tasks-fs/src/glob.rs
timneutkens: :robot: The glob equality now includes GlobOptions, so case-sensitive and case-insensitive matchers cannot be treated as equivalent. Thanks for catching the original omission.

docs/01-app/03-api-reference/08-turbopack.mdx
timneutkens: :robot: Done — the options reference now states that caseSensitive: false ignores ASCII case only.

turbopack/crates/turbo-tasks-fs/src/glob.rs
timneutkens: :robot: Done — added an inline comment explaining that matching is ASCII case-insensitive because the regex Unicode flag is not enabled.

PR #96228, opened 2026-07-25
Port React's @gate test directive to the e2e harness

acdlite:
The test suite has accumulated a bunch of patterns for disabling tests that are known to fail under some configuration: `it.skip`, `if (isNextDev) { test('skipped in dev mode', () => {}); return }`, whole describes toggled off by checking `process.env.__NEXT_CACHE_COMPONENTS`. These all have the same flaw: nothing tells you when the thing you skipped starts working. The test stays disabled forever, and the workaround it was guarding rots along with it.

React solves this with the `@gate` pragma, and this PR ports it to the Next.js e2e harness:

```ts
// Blocked on the optimization that marks a route as fully static when
// no dynamic params are referenced in Server Components.
// @gate !cacheComponents
it('navigates to a page with a lazily-generated static param', async () => {
  // body unchanged
})
```

The test still runs. If the condition is false and the test fails, the failure is absorbed and the suite stays green. If it _passes_, the suite fails: the gate is stale, delete it. So instead of a skip that hides a fixed bug indefinitely, you get a CI failure the day the fix lands.

When the condition is static, the inversion is Jest's own `test.failing` under the hood. A lazy condition isn't known until the fixture's resolved config is read inside the body, so those tests invert at runtime instead.

`// @force-gate <condition>` skips for real — for tests that can't even be attempted (prefetching is disabled in dev, deploy has no local build output, the fixture won't build under the condition), and for tests of a new API, where the disabled state can only throw and running it proves nothing:

```ts
// Prefetching is disabled in dev, so this suite has nothing to test.
// @force-gate prefetching
describe('segment cache prefetch scheduling', () => {
  // ...
})
```

There's no staleness check in that case, so this is a judgment call: prefer `@gate` when the off state fails for a meaningful reason — the flag changes behavior that already exists — and `@force-gate` when the body can only throw because the API doesn't exist. A static condition (mode, bundler) resolves at collection time into a normal Jest skip. A lazy condition resolves at runtime, and when a lazy force-gate on a describe is false, we skip the fixture build entirely — that's what makes it usable for suites whose fixtures are build-incompatible with the condition. (One caveat: Jest has no way to skip a test that's already running, so these report as passing with a warning in the log, not as skipped.)

Conditions live in a hand-written registry. I considered deriving the lazy ones from the config schema automatically, but a gate is a claim about which dimension of the test matrix explains a failure, and I'd rather each of those claims be spelled out with a description. Referencing an undeclared name fails the suite at collection time, so a typo can't silently disable a gate.

The important design decision for lazy conditions is that they read the fixture's _resolved_ config, never `process.env`. The env var isn't the truth: `__NEXT_CACHE_COMPONENTS=true` only applies when the fixture doesn't set `cacheComponents` itself, and config resolution implies flags the fixture never mentions (`cacheComponents: true` alone turns on `experimental.ppr`). Resolution happens in a child process, because in-process `loadConfig` would leak the fixture's `.env` files into the Jest worker. Suites with no lazy gate never pay for any of this.

The condition expression is parsed using a small grammar (also ported from the React repo). An expression that doesn't parse fails the suite:

```ts
// @gate mode === 'start' && !cacheComponents
// @gate !(turbopack || rspack)
```

There's also a runtime version, mirroring React's `gate(flags => flags.enableFoo)`, for tests that run under both states but assert differently (and for `it.each`, where the pragma can't attach):

```ts
import { gate } from 'next-test-utils'

it('renders the fallback', async () => {
  if (await gate((conditions) => conditions.cacheComponents)) {
    // PPR: the fallback is part of the static shell
  } else {
    // fully dynamic: the fallback streams in
  }
})
```

It also accepts the pragma expression language as a string: `await gate('cacheComponents && !dev')`.

Docs are in `test/lib/gate/README.md`; `test/unit/gate/` covers the transform, the expression language, and the runtime.

### DIFF
--- a/.agents/skills/gate-tests/SKILL.md
+++ b/.agents/skills/gate-tests/SKILL.md
@@ -1,0 +1,194 @@
+---
+name: gate-tests
+description: >
+  How to use the `@gate` / `@force-gate` test directives instead of `it.skip`
+  or fake-green skip patterns. Use when a test is known-failing under some
+  test-matrix dimension (dev mode, a bundler, an experimental flag like
+  cacheComponents), when converting `if (isNextDev) return` guards or
+  env-var `describe.skip` branches, when adding a condition to
+  test/lib/gate/conditions.ts, or when keying a fixture's experimental flag
+  on a __NEXT_TEST_AXIS letter. Covers directive choice, condition tiers,
+  the test-axis fixture pattern, pitfalls, and verification commands.
+user-invocable: false
+metadata:
+  internal: true
+---
+
+# Gating tests with `@gate` / `@force-gate`
+
+Full reference: [`test/lib/gate/README.md`](../../../test/lib/gate/README.md).
+This skill is the decision guide: which directive to reach for, the standard
+conversion patterns, and how to verify.
+
+## Never write these — gate instead
+
+| Anti-pattern                                                           | Replacement                                                                 |
+| ---------------------------------------------------------------------- | --------------------------------------------------------------------------- |
+| `it.skip('...')` for a known failure                                   | `// @gate <cond>` (or `@gate FIXME` if no condition explains it)            |
+| `if (isNextDev) { test('skipped in dev mode', () => {}); return }`     | `// @force-gate prefetching` (or `!dev`) on the `describe`                  |
+| `(flagEnabled ? describe.skip : describe)(...)` keyed on `process.env` | `// @force-gate <cond>` (lazy) on the `describe`                            |
+| Duplicating a fixture directory per flag state                         | one fixture keyed on `__NEXT_TEST_AXIS` + a `@gate`/`@force-gate`           |
+| Branching expectations on `process.env.__NEXT_CACHE_COMPONENTS`        | `if (await gate((c) => c.cacheComponents))` (`gate` from `next-test-utils`) |
+
+The skip patterns are fake-greens: nothing tells you when the bug they hide is
+fixed. `@gate` still runs the body and fails the suite the day the "known
+failure" starts passing, so stale workarounds get deleted instead of rotting.
+
+## Choosing the directive
+
+Ask what kind of difference you're encoding:
+
+1. **A behavior change — both states assert something meaningful.** Don't
+   gate the test at all: fork inside the body with the runtime `gate()` —
+   same condition registry, no inversion — which pinpoints exactly what
+   differs, and also covers `it.each`, where a pragma cannot attach:
+   `if (await gate((c) => c.cacheComponents)) { ... } else { ... }`. It
+   mirrors React's `gate(flags => ...)`; a pragma expression string works
+   too (`await gate('cacheComponents && !dev')`). A suite-level pragma is
+   too coarse here — it hides _what_ is different between the states.
+2. **A flag that changes the behavior of existing surface**
+   (`cacheComponents`, `optimisticRouting`) **and the suite is written for one
+   state.** `// @gate <cond>` on the test or `describe`. The body runs; a
+   false condition inverts the expectation (failure absorbed, a pass fails as
+   "stale gate"). The off state fails for a meaningful reason — the behavior
+   differs — so a pass is real information: the gate is stale, delete it.
+3. **A new API — the off state proves nothing.** Typically
+   `// @force-gate <cond>` (lazy) on the `describe`. An API that throws when
+   its flag is off — or is inert, like `useOffline()`, which compiles to a
+   hook that always reports online — can only fail vacuously (often slowly,
+   by timing out), and browser e2e time is considerable, so skip the run
+   (and the fixture build) instead of paying for it. Working example:
+   `test/e2e/app-dir/use-offline/`. This is discretion, not a rule: when the
+   flag changes behavior the suite can observe, the off state is meaningful
+   and `@gate` buys the staleness check.
+4. **`@force-gate <cond>` also when running the body is impossible**, not
+   merely failing: prefetching is off in dev, deploy has no local build
+   output, the fixture cannot even build under the condition.
+   - Static condition (`!dev`, `bundler`…) → real Jest `○ skipped` at
+     collection.
+   - Lazy condition on a `describe` → the fixture **build is skipped** when
+     false; tests report passed-with-`⚠ skipped by @force-gate` (Jest cannot
+     skip at runtime). Build-skipping covers `start`/`dev` suites where
+     `nextTestSetup` owns the build — not `skipStart` suites, not deploy.
+5. Pragmas stack: a common pair is a static `// @force-gate prefetching` plus
+   a lazy `// @gate <flag>` on the same `describe`.
+
+### Is the off-state run worth its cost?
+
+Browser e2e time is not free, so weigh what the gated-off run buys. For a
+behavior flag it usually replaces a run that was already being paid for — a
+fixture that pins its flags runs identically with and without the axis set,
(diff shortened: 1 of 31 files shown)


.agents/skills/gate-tests/SKILL.md
```diff
@@ -0,0 +1,159 @@
...
+   output, the fixture cannot even build under the condition.
+   - Static condition (`!dev`, `bundler`…) → real Jest `○ skipped` at
+     collection.
+   - Lazy condition on a `describe` → the fixture **build is skipped** when
+     false; tests report passed-with-`⚠ skipped by @force-gate` (Jest cannot
+     skip at runtime). Build-skipping covers `start`/`dev` suites where
+     `nextTestSetup` owns the build — not `skipStart` suites, not deploy.
+3. Pragmas stack: a common pair is a static `// @force-gate prefetching` plus
+   a lazy `// @gate <flag>` on the same `describe`.
+4. **A body that runs under both states but asserts differently is not a
```
eps1lon: Wdyt about moving this up a bit? I oftentimes feel a top level gate is too coarse. A top-level gates makes sense for new APIs but for behavior changes an inline gate helps understand what specifically is different.

.agents/skills/gate-tests/SKILL.md
```diff
@@ -0,0 +1,159 @@
...
+  cacheComponents: true,
+  experimental: {
+    useOffline: process.env.__NEXT_TEST_VARIANT !== 'true',
+  },
+}
+```
+
+```ts
+// @force-gate prefetching
+// @gate useOffline
```
eps1lon: For React this is fine because tests can be run quickly so the overhead is minimal. For browser e2e test the cost of running a test is considerable. For new APIs there's very little point of running them at all so we might as well skip instead to avoid cost.

.agents/skills/gate-tests/SKILL.md
```diff
@@ -0,0 +1,159 @@
...
+describe('useOffline', () => { ... })
+```
+
+The plain shard exercises the feature; the variant shard asserts it is inert
```
eps1lon: We wanted to avoid "variant" to disambiguate with the "variants" feature

test/lib/gate/conditions.ts
```diff
@@ -0,0 +1,203 @@
...
+ * ## The two tiers
+ *
+ * **`staticCondition`** — the value is known before any test runs (run mode,
+ * bundler, React version). These are the only conditions `@force-gate` accepts,
+ * because a real Jest skip has to be decided while tests are being collected.
+ *
+ * **`lazyCondition`** — a predicate over the *resolved* `next.config` of the
+ * fixture the suite booted (`NextInstance.getResolvedConfig()`). The value is
+ * read the first time a gate asks for it, which is inside the test body,
+ * because nothing about the fixture exists at collection time.
```
eps1lon, line 22: This feels pretty expensive. While the env variable wasn't pretty, it was at least cheap. Setting up a fixture and resolving the config isn't.

test/lib/gate/pragma-transform.js
```diff
@@ -0,0 +1,175 @@
...
+ * This is deliberately a regex rewrite and not an AST transform: the only
+ * edited bytes are on the `it(` / `describe(` line itself, so the line numbers
+ * of every other line are preserved exactly. That matters because Jest writes
+ * `toMatchInlineSnapshot()` results back by line/column, and because stack
+ * traces should still point at the original source.
+ */
+
+/**
+ * A pragma block is one or more consecutive `// @gate` lines *immediately*
+ * followed by an `it` / `test` / `fit` / `describe` call. Anything else (a
```
eps1lon, line 31: missing `xit` which is an alias for `it.skip`

.agents/skills/gate-tests/SKILL.md
acdlite: That's what the "force" version does, it prevents it from running at all

acdlite: However I would dispute there's no reason running the tests for experimental flags. It prevents vacuous results (test that passes doesn't require the new feature to pass).

Also consider that many of our fixtures hardcode experimental flags to true today. Maybe they shouldn't, but that's a wasted CI run that this new pattern now reclaims.

.agents/skills/gate-tests/SKILL.md
acdlite: ah that's a good call, I'll rename

test/lib/gate/conditions.ts
acdlite: This is up to the discretion of the test author to choose between `@force-gate` and `@gate`.

acdlite: Oh you mean the lazy conditions thing. Yes but I think most of those cases that we would prefer to skip are statically known (e.g. we frequently disable things during deploy runs, or in dev) and can be a `@force-gate` if desired.

test/lib/gate/pragma-transform.js
acdlite: I made it error if you combine gate with `.skip` since it's ambiguous how that's meant to work.

.agents/skills/gate-tests/SKILL.md
acdlite: Added an earlier exit + updated the skill with advice

.agents/skills/gate-tests/SKILL.md
acdlite: Renamed to `__NEXT_EXPERIMENTAL_CI_SHARD`

.agents/skills/gate-tests/SKILL.md
eps1lon: I'm not seeing what value there is to run a test with an API that doesn't exist. It just throws.

.agents/skills/gate-tests/SKILL.md
```diff
@@ -0,0 +1,180 @@
...
+an experimental CI shard number and gate the suite. Key it so the flag is
+**enabled by default** — then the suite exercises the feature in plain local
+runs with no special env, and the shard run covers the off state:
+
+```js
+// next.config.js — pin every dimension except the one under test
+const nextConfig = {
+  cacheComponents: true,
+  experimental: {
+    useOffline: process.env.__NEXT_EXPERIMENTAL_CI_SHARD !== '1',
```
eps1lon: is "1" here a number or a boolean? It's a bit ambiguous because sometimes we also refer to the groups as shards. telling tests to only run if the test is in a specific bucket out of the 7 we split tests into would basically mean you want to test it randomly. If it's a boolean, we can use `'true'` instead matching existing flags we set in CI.

Also let's avoid saying "both CI runs" because the same test runs more than twice. We also have bundler, React version and Adapter as a test dimension.

acdlite: "shard" is perhaps wrong word for this concept but the idea is that there would be a fixed number of variants and you should choose which one you want it to run in. Maybe it's clearer with A, B, C instead of numbers.

acdlite: going to rename to __NEXT_TEXT_AXIS instead. I imagine we'll start out with just A and B but the reason I chose not to use a boolean is in case we want to add more later (even just temporarily).

.agents/skills/gate-tests/SKILL.md
acdlite: Switched the example to `concurrentRouterQueue` and added more guidance

eps1lon (approved):
Exciting!

PR #96242, opened 2026-07-26
Keep the dev server alive when its stdout pipe breaks

### LINKED ISSUE #96216: Turbopack dev server hangs all requests forever with sustained native CPU spin after its stdout pipe reader exits (`next dev | head -N`)
### Verify canary release

- [x] I verified that the issue exists in the latest Next.js canary release (`16.3.0-canary.96`)

https://github.com/kando19804-oss/next-turbopack-broken-pipe-hang-repro

### To Reproduce

1. `pnpm install`
2. Start the dev server with an early-exiting stdout reader (bash / Git Bash):
   ```bash
   npx next dev -p 3111 2>&1 | head -15 &
   ```
3. Request the page repeatedly:
   ```bash
   for i in $(seq 1 30); do curl -s -o /dev/null -m 10 -w "#$i -> %{http_code}\n" http://localhost:3111/; done
   ```

**Current:** the server works normally for the first ~11 requests (~140 ms each), then — once its cumulative log output exceeds the pipe budget (the reader's `-N` lines + ~64 KB pipe buffer) — **every subsequent request hangs forever**:

```
#1  -> 200
...
#11 -> 200
#12 -> 000   (curl timeout)  ← hangs forever from here
#13 -> 000
...
```

While frozen (measured on the minimal repro):

- the process stays alive and keeps **LISTENing**; TCP connections are accepted but never answered (client sockets pile up in `CLOSE_WAIT`),
- **CPU time grows ~+26 s per 20 s wall clock (~1.3 cores) indefinitely while serving nothing**, memory grows slowly (530 → 567 MB in 20 s),
- a JS CPU profile (`--experimental-cpu-prof`, taken on a real app in the same state) shows the JS thread **99.9 % idle** — the stall and the CPU burn are on the native (Turbopack) side,
- the server never recovers; only killing the process helps.

**Control:** the identical server started with a file redirect instead of the pipe (`npx next dev -p 3111 > dev.log 2>&1`) served 1,000+ requests (cold start, 15-way parallel bursts, HMR churn) without a single failure.

**Expected:** anything but an unbounded silent hang + CPU spin. Either

- keep serving requests when stdout writes fail (drop log output), or
- fail fast with a visible error (e.g. exit on EPIPE).

```
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 65278
  Available CPU cores: 20
Binaries:
  Node: 20.20.1
  npm: 10.8.2
  pnpm: 10.32.1
Relevant Packages:
  next: 16.2.11 (also reproduced on 16.3.0-canary.96)
  react: 19.2.3
  react-dom: 19.2.3
Next.js Config:
  output: N/A (repro), also reproduced with output: "standalone" on a large app
```

Turbopack

next dev (local)

### Additional context

- Real-world impact: piping a dev server's output through `head` to capture startup logs is common in CI scripts, tmux panes, and agent/automation harnesses (`pnpm dev 2>&1 | head -40 &`). On a large production app this exact pattern produced "the dev server dies after ~20–30 requests" with **no error output whatsoever** (stdout is gone), which took us two days and a CPU profile to trace back to the launch command. The freeze threshold scales with *log lines written*, not requests — a browser page that fires 15 API calls reaches it in one navigation, which made it look like an app bug.
- The repro's `next.config.js` sets `turbopack.root` only to keep workspace-root inference stable; it is unrelated.
- If the pipe budget is exhausted *during startup* (e.g. `| head -5`, smaller than the banner), the process dies before ever listening — the pathological hang + CPU spin happens when the budget is exhausted *after* startup.
- Reproduced on Windows 11 (pipe created by bash/Git Bash). The mechanism (writes to a broken/undrained stdout pipe stalling the native layer) is presumably platform-generic, but I could only verify on Windows.

petehunt:
### What?

`next dev` becomes permanently unresponsive when its stdout is piped into a reader that exits early — `next dev | head -20`, a detached tmux pane, a CI log tailer. The server keeps accepting connections but never answers them, and prints nothing to explain why.

Fixes #96216

### Why?

Once the read end of the pipe closes, every log line raises an `EPIPE` error event on `process.stdout`/`process.stderr`. With no listener attached, Node re-throws it as an `uncaughtException`, which the dev server's handler in `router-server.ts` reports via `Log.error` — over the same broken pipe. That raises another `EPIPE`, and so on. The loop pins a CPU core, allocates steadily, and starves incoming requests.

### How?

New `setupBrokenPipeHandling()` attaches an `error` listener to both stdio streams. On a fatal code (`EPIPE`, `ECONNRESET`, `EBADF`, `ERR_STREAM_DESTROYED`) it replaces `stream.write` with a no-op so nothing retries the dead destination; other errors are swallowed rather than crashing the process. When only one of the two streams is broken, a one-time notice goes to the other.

It is called from two entrypoints: `bin/next.ts` (before anything writes output) and `server/lib/start-server.ts`, which is the forked dev server inheriting the CLI's stdio.

The regression test drives the CLI directly rather than `nextTestSetup` — the harness keeps stdout drained for its own log assertions, and this test needs to close the read end. It destroys both pipes, then asserts 25 subsequent requests still return 200 and the process is alive.

Closes NEXT-


packages/next/src/lib/setup-broken-pipe-handling.ts
```diff
@@ -0,0 +1,78 @@
...
+      process.nextTick(cb as () => void)
+    }
+    return true
+  } as NodeJS.WriteStream['write']
+
+  // If the other stream is still alive (`next dev | head -20` leaves stderr
+  // attached to the terminal) say why the output stopped. If it shares the same
+  // broken pipe, this write is a no-op or silences it too.
+  const sibling = name === 'stdout' ? process.stderr : process.stdout
+  if (sibling && typeof sibling.write === 'function') {
```
lukesandberg, line 35: nit:

```suggestion
  if (sibling && typeof sibling.write === 'function' && silenced.size < 2`) {
```

lukesandberg (approved):

PR #96245, opened 2026-07-26
Only advance the Turbopack HMR hash when compiled output actually changed

gaearon:
Stacked on #96235.

Addresses [this TODO](https://github.com/vercel/next.js/pull/96245#discussion_r3659708308).

The Turbopack HMR refresh hash that dev folds into every `"use cache"` key was a counter, incremented once per change-subscription emission:

```ts
// TODO: Get an actual content hash from Turbopack.
const message = await createMessage(change, String(++hmrHash))
```

A subscription emission does not always mean code changed: a newly added page's subscription emits once for the entry's own initial build, and a removed page's subscription emits one final time on the way out. Before this fix, those emissions incremented the counter like any other — and one stray increment discards every cached entry, because the counter is part of every key.

Whether those emissions actually got counted depends on timing. On the July canary this stack started from, they did — visiting a newly added page reliably threw all cached data away under CPU load:

```
value: 0.485620
value: 0.485620
$ echo '...' > app/unrelated/page.tsx
$ curl localhost:3000/unrelated          # first visit builds the page
value: 0.890276                          # cached entry gone
```

On the current base I could not reproduce this: the extra emissions now arrive early enough to be folded into the initial emission that the subscription already skips, and the page add/remove test from #96235 passes on Turbopack as-is (13 runs under CPU load, plus a scripted loop of 15 add/remove cycles, all clean). But that depends entirely on emission timing, which has already shifted once between July and now. The counter still counts emissions rather than changes, so a future shift can bring the bug back. This PR removes the dependence on timing: the hash advances only when the compiled output actually changed. The content comparison it introduces is also what the next PR builds on.

## Fix

On each emission, digest the content hashes of the endpoint's written server files (`writeToDisk()` returns them), and increment only when the digest differs from the last one. An empty digest means the entry was removed, which can't affect other entries' cached data, so it doesn't increment either. If the digest can't be computed, err on the side of invalidating.

This is cheaper than it looks: the subscription itself is computed from the endpoint's output (`server_changed` walks the output assets' contents to decide whether to emit), and `writeToDisk()` returns that same memoized output. The middleware subscription already calls `writeToDisk()` on every emission today.

Skipping the first emission by position instead would be wrong: emissions coalesce, so an initial-build emission that arrives late can already include a real change. I observed this with an `.env` edit folded into the first emission — skipping by position served stale cached entries for code that had changed.

## Test plan

The page add/remove test from #96235 runs on Turbopack and keeps guarding this: it catches the counter bug whenever timing exposes it, as it did on the July base. Green with the fix, idle and under CPU load: `cache-components-spurious-cache-invalidation` (6), `server-components-hmr-cache` (17 — edits and env changes must still invalidate), `cache-components-dev-warmup`, and the basic `hmr` suites, which cover the subscription loop this changes.

### DIFF
--- a/packages/next/src/server/dev/hot-reloader-turbopack.ts
+++ b/packages/next/src/server/dev/hot-reloader-turbopack.ts
@@ -1619,13 +1619,13 @@ export async function createHotReloaderTurbopack(
     },
 
     getServerComponentsHmrRefreshHash() {
-      // The current server-components generation. Only the change subscription
-      // (an actual recompile) advances `hmrHash`; reloads and config
-      // invalidations don't, so the value stays stable across requests until a
-      // real edit. Returned unconditionally (`"0"` before the first edit) so
-      // `"use cache"` keys are present and consistent for every request,
-      // mirroring webpack's always-present `stats.hash`.
-      return String(hmrHash)
+      // Only the change subscription (an actual recompile) advances `hmrHash`;
+      // reloads and config invalidations don't, so the value stays stable
+      // across requests until a real edit. `sessionId` stands in for a key
+      // derived from the compiled implementation, which would let entries
+      // outlive a restart when the code didn't change (see the note on Action
+      // IDs in `use-cache-wrapper.ts`).
+      return `${sessionId}-${hmrHash}`
     },
 
     sendToLegacyClients(action) {
@@ -1762,10 +1762,7 @@ export async function createHotReloaderTurbopack(
       }
       return errors
     },
-    async invalidate({
-      // .env files or tsconfig/jsconfig change
-      reloadAfterInvalidation,
-    }) {
+    async invalidate({ reloadAfterInvalidation }) {
       if (reloadAfterInvalidation) {
         for (const [key, entrypoint] of currentWrittenEntrypoints) {
           clearRequireCache(key, entrypoint, { force: true })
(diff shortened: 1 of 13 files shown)


packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -863,10 +883,30 @@ export async function createHotReloaderTurbopack(
...
+      // entry *may* have changed — including once for the entry's own
+      // initial build, and possibly coalescing several changes into one
+      // emission — so emissions can't be counted directly. Instead, compare
+      // the entry's compiled output and only advance the hash when it
+      // actually changed.
+      let contentDigest = await getServerContentDigest(endpoint)
+
       for await (const change of changed) {
         processIssues(currentEntryIssues, key, change, false, true)
-        // TODO: Get an actual content hash from Turbopack.
```
gaearon, line 943: This and the next PR are to address this TODO.

wbinnssmith (approved):

gaearon:
Folded into #96248, which now implements this natively from the start and carries the test. The JS-side digest here called `writeToDisk()` once per change-subscription emission; that's a strongly-consistent, effect-applying operation, so every emission waited for a settled graph and successive rapid edits stalled HMR — `acceptance-app/app-hmr-changes` timed out deterministically on this commit and was green again on #96248, which deletes the function.

PR #96248, opened 2026-07-26
Only advance the Turbopack HMR hash when compiled output actually changed

gaearon:
Stacked on #96235. Supersedes #96245, which did the same thing with a JS-side digest.

Addresses [this TODO](https://github.com/vercel/next.js/pull/96245#discussion_r3659708308).

The Turbopack HMR refresh hash that dev folds into every `"use cache"` key was a counter, incremented once per change-subscription emission:

```ts
// TODO: Get an actual content hash from Turbopack.
const message = await createMessage(change, String(++hmrHash))
```

A subscription emission does not always mean code changed: a newly added page's subscription emits once for the entry's own initial build, and a removed page's subscription emits one final time on the way out. Those emissions increment the counter like any other — and one stray increment discards every cached entry, because the counter is part of every key.

Whether those emissions actually get counted depends on timing. On the July canary this stack started from, they did — visiting a newly added page reliably threw all cached data away under CPU load:

```
value: 0.485620
value: 0.485620
$ echo '...' > app/unrelated/page.tsx
$ curl localhost:3000/unrelated          # first visit builds the page
value: 0.890276                          # cached entry gone
```

On the current base it's intermittent instead: the extra emissions usually arrive early enough to be folded into the initial emission that the subscription already skips. The first commit adds a test for it, which failed one run in five in CI's flake detection on the base and passes on every run here.

The counter counts emissions rather than changes, so how often it's wrong is a scheduling detail, and it has already shifted once between July and now.

## Fix

The hash advances only when the compiled output actually changed. Turbopack already reads every output asset's content just to decide whether to emit, so the hash is computed right there:

- `Project::server_content_hash` hashes the contents of the same assets that `server_changed` watches. The filter is a shared helper (`watched_output_assets`) that `any_output_changed` now also uses, so the two can't drift apart.
- Each endpoint gets a `server_content_hash()` method that mirrors its `server_changed`. It returns None when the endpoint has no output anymore (the entry was removed). The subscriber treats that as "not a change".
- Every emission now carries the hash, and a new `endpointServerContentHash()` binding returns it on demand. The hot reloader calls that once at subscribe time and uses the result as the baseline, so a first emission that arrives late and already contains a real change is still detected as a change.
- The JS side compares two strings.

Skipping the first emission by position instead would be wrong: emissions coalesce, so an initial-build emission that arrives late can already include a real change. I observed this with an `.env` edit folded into the first emission — skipping by position served stale cached entries for code that had changed.

The same change is applied to middleware for consistency.

A last commit removes the hash argument that change-subscription callbacks were passed: no callback ever read it, and the messages that need the current hash read `hmrHash` where they're built.

## Test plan

Verified with a locally built binding, idle and under CPU load (`for i in $(seq 1 12); do yes > /dev/null & done`):

- `cache-components-spurious-cache-invalidation`: 6 of 6 on Turbopack, 5 of 5 on webpack, idle and in repeated loaded runs. This includes the page add/remove test this PR adds, and the env test that catches a real change folded into a subscription's first emission.
- `server-components-hmr-cache` (17 tests — edits and env changes must still invalidate), `middleware-dev-update` (4 tests — middleware edits must still reprocess and apply), `cache-components-dev-warmup`, the basic `hmr` suites, and `route-change-refetch` (10 tests, plus 10 consecutive full-suite Turbopack runs) all pass.

To confirm emissions behave the same before and after, I logged every subscription emission while scripting the same sequence against both builds on the original base — boot, add a page and visit it, edit the page, edit `.env`, remove the page. The logs came out identical: one emission for the page edit, one for the env edit, nothing else.

### DIFF
--- a/crates/next-api/src/project.rs
+++ b/crates/next-api/src/project.rs
@@ -48,6 +48,7 @@ use turbo_tasks_fs::{
     DiskFileSystem, FileContent, FileSystem, FileSystemPath, VirtualFileSystem,
     canonicalize_to_rcstr, invalidation,
 };
+use turbo_tasks_hash::{HashAlgorithm, hash_xxh3_hash64};
 use turbo_unix_path::join_path;
 use turbopack::{
     ModuleAssetContext, evaluate_context::node_build_environment, externals_tracing_module_context,
@@ -55,6 +56,7 @@ use turbopack::{
 };
 use turbopack_core::{
     PROJECT_FILESYSTEM_NAME,
+    asset::{Asset, no_hash_salt},
     changed::content_changed,
     chunk::{
         ChunkingContext, EvaluatableAssets, UnusedReferences,
@@ -2716,6 +2718,17 @@ impl Project {
         Ok(any_output_changed(roots, path, true))
     }
 
+    /// A hash of the contents of the same output assets that `server_changed`
+    /// watches. None when there are no assets (e.g. the entry was removed).
+    #[turbo_tasks::function]
+    pub async fn server_content_hash(
+        self: Vc<Self>,
+        roots: Vc<OutputAssets>,
+    ) -> Result<Vc<OptionContentHash>> {
+        let path = self.node_root().owned().await?;
+        Ok(output_assets_content_hash(roots, path, true))
+    }
+
     /// Completion when client side changes are detected in output assets
     /// referenced from the roots
     #[turbo_tasks::function]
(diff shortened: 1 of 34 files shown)


crates/next-api/src/project.rs
```diff
@@ -2848,6 +2861,63 @@ async fn any_output_changed(
     Ok(Vc::<Completions>::cell(completions).completed())
 }
 
+#[turbo_tasks::value(transparent)]
+pub struct OptionContentHash(Option<RcStr>);
+
+/// Hashes the contents of the output assets referenced from the roots,
+/// with the same asset filtering as `any_output_changed`. Returns None when
```
wbinnssmith: > same asset filtering as `any_output_changed`

this is a fairly non-trivial set of conditions. can it be shared at all with that implementation?

wbinnssmith (approved):

crates/next-api/src/project.rs
gaearon: Extracted the filter into `watched_output_assets`, shared by `any_output_changed` and `output_assets_content_hash`.

gaearon:
I can't reliably reproduce the original issue anymore so I'm going to drop this for now and maybe revive later.

PR #96252, opened 2026-07-27
Fix race when navigating Back before hydration

gaearon:
Relands #95682 (reverted in #95853 due to #95848).

This contains:

- A bunch of tests (some failing and some passing before this PR; all passing after this PR).
- The fix from #95682.

In short, we use Navigation API to detect during hydration that we're on a different history entry than we started, and in that case, replay the missed traversal with similar logic to what we do for `onPopState`.

The issue with the original PR was that under Activity, React would sometimes hang applying the state update. This was fixed in https://github.com/react/react/pull/37135 and synced, so we can do this fix again now.

## Test Plan

Old and new tests.

### DIFF
--- a/packages/next/src/client/components/app-router.tsx
+++ b/packages/next/src/client/components/app-router.tsx
@@ -56,14 +56,30 @@ const globalMutable: {
   pendingMpaPath?: string
 } = {}
 
-// A Back/Forward press between a reload and hydration moves the browser to
-// a different history entry than the one the document was activated on, and
-// the resulting popstate fires before any listener exists. When that
-// happened, the initial history write is skipped so the traversed entry's
-// state is not overwritten, and the missed popstate is replayed once the
-// router has mounted.
-let missedTraversalBeforeHydration: 'unchecked' | 'missed' | 'none' =
-  'unchecked'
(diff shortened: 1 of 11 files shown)


packages/next/src/client/components/app-router.tsx
```diff
@@ -70,6 +107,28 @@ function HistoryUpdater({
 
     const { tree, pushRef, canonicalUrl, renderedSearch } = appRouterState
 
+    if (missedTraversalBeforeHydration === 'unchecked') {
+      missedTraversalBeforeHydration = 'none'
```
ztanner: Could there still be a small race between this insertion effect and the `useEffect` that installs the popstate listener, where this is set to none but the user navigates Back or Forward before the listener is attached, so the event is missed and never replayed?

gaearon: Yeah. Pushed a slightly different fix. Not sure how to capture in tests, but tried to verify it by telling the agent to artificially delay the later code block, and to confirm that it wasn't confused by a navigation between them.

ztanner (approved):

franky47:
Confirmed working fine with the initial nuqs regression test in #95848, thanks!

PR #96267, opened 2026-07-27
Use a safe clock for Request Insights bookkeeping

DavidIlie:
## Summary

- Request Insights was using `Date.now()` to keep track of spans, but Cache Components patches the `Date` constructor during prerendering to detect unstable wall-clock reads.
- This meant the Request Insights call could be mistaken for a `Date.now()` call from user code and trigger an error.
- The fix now uses `performance.timeOrigin + performance.now()` for implicit Request Insights timestamps.
- This is a focused core change.

## Verification

- `pnpm exec jest --runTestsByPath packages/next/src/server/lib/trace/local-span-recorder.test.ts packages/next/src/server/lib/trace/request-insights.test.ts packages/next/src/server/lib/trace/span-store.test.ts`
- `HEADLESS=true pnpm test-dev-turbo test/development/app-dir/request-insights/request-insights.test.ts`
- `HEADLESS=true pnpm test-dev-webpack test/development/app-dir/request-insights/request-insights.test.ts`
- `pnpm --filter=next build`

### DIFF
--- a/packages/next/src/server/lib/trace/local-span-recorder.test.ts
+++ b/packages/next/src/server/lib/trace/local-span-recorder.test.ts
@@ -197,51 +197,6 @@ describe('local recording span', () => {
     ])
   })
 
-  it('uses performance timing for implicit span bookkeeping', () => {
-    const before = performance.timeOrigin + performance.now()
-    const dateNow = jest.spyOn(Date, 'now').mockImplementation(() => {
-      throw new Error('Date.now should not be used for span bookkeeping')
-    })
-
-    try {
-      const span = createLocalSpan({ name: 'test.local-span.safe-clock' })
-      span.addEvent('event')
-      span.recordException(new TypeError('boom'))
-      span.end()
-    } finally {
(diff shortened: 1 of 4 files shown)


gaearon:
For tests, would it make sense to test the user-observable behavior instead if possible? 

That is, instead of a unit test, we could have an E2E test that:

- Triggers the problematic behavior before the fix (and is confirmed to fail) by turning on Request Insights in CC mode
- Then, the fix

This would cover a broader “I turned on Request Insights and CC works” instead of a narrow “this particular file should use this particular API”. So this catches other similar regressions.

gaearon:
In general E2E tests are often better because they can assert the thing the user actually cares about, encodes what was being fixed, and doesn’t slow down refactorings that move code around.

DavidIlie:
hi @gaearon, you're right

i have modified the test to be more end-user oriented. removed the mocking of `Date.now()` and kept only the explicit OpenTelemetry timestamp-input coverage

PR #96272, opened 2026-07-27
[DOCS] Clarify server actions usecase

anay-208:
### What?
In the Next.js Documentation page for mutations, I realised that it is really unclear that server actions are supposed to be used for "mutations" only. While mutations are explicitly mentioned, it might get overlooked by some people, and they might assume it's a primary use case, and it could have secondary use cases too

### Why?
So this issue originated with my AI agent; it is really common for AI agents to hallucinate, but this could potentially reduce it.
In addition, it'll improve clarify for anyone reading it

### How?
Update the relevant parts of the docs.


anay-208:
can someone review it

anay-208:
can someone review these changes

PR #96273, opened 2026-07-27
[turbopack] Treat `Object.defineProperty(exports, …, { value: … })` as side-effect free

sampoder:
Treat code like this:

```
Object.defineProperty(exports, '__esModule', { value: true });
exports.foo = 1;
```

As side-effect free. This is an extension of the work on https://github.com/vercel/next.js/pull/94293 and will allow https://github.com/vercel/next.js/pull/95718 to drop more dead requires.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
@@ -439,10 +439,11 @@ fn module_exports_has_accessor(program: &Program, unresolved_mark: Mark) -> bool
                 found = true;
             }
         }
-        // A `get`/`set` in the descriptor installs an accessor, and a spread or a
-        // computed key could carry one.
+        // look for `Object.defineProperty(exports, …)`
         Expr::Call(call) => {
             let is_accessor_key = |key: &PropName| {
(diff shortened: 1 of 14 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/side_effects.rs
```diff
@@ -417,22 +421,51 @@ fn module_exports_is_tainted(program: &Program, unresolved_mark: Mark) -> bool {
...
+                && contains_getters_or_setters(&assign.right)
+            {
+                found = true;
+            }
+        }
+        // A `get`/`set` in the descriptor installs an accessor, and a spread or a
+        // computed key could carry one.
+        Expr::Call(call) => {
+            let is_accessor_key = |key: &PropName| {
+                matches!(key, PropName::Computed(_))
```
bgw: Move this comment down a few lines where it makes more sense, add a comment about `Object.defineProperty` here.

```suggestion
        // look for `Object.defineProperty(exports, …)`
        Expr::Call(call) => {
            let is_accessor_key = |key: &PropName| {
                // A `get`/`set` in the descriptor installs an accessor, and a spread or a
                // computed key could carry one.
                matches!(key, PropName::Computed(_))
```

bgw (approved):

PR #91507, opened 2026-03-17
block disallowed dev origins by default

ztanner:
This removes the warn-only default behavior and enforces the dev-origin guard by default. Cross-origin requests to internal dev resources now block unless they match the built-in local allowlist or an explicit `allowedDevOrigins` entry. The tests are expanded to cover default blocking, configured-but-not-allowlisted hosts, missing Referer in the no-cors path, and same-site requests without an Origin, and the docs are updated to match the new behavior.

PR #96283, opened 2026-07-28
Fix for networkHostname should be in allowedDevOrigins by default

### LINKED ISSUE #96241: networkHostname should be in allowedDevOrigins by default

https://github.com/mhverbakel/nextjs-reproduction-template

### To Reproduce

⚠️ Note that the reproduction URL is an empty reproduction template. This issue occurs in every Next.JS project without `allowedDevOrigins` in its config.

1. Start the development server (`next dev`)
2. Note that the start-up script produces two URLs (IP address may be different on your local machine):
```
▲ Next.js 16.3.0-canary.97 (Turbopack)
- Local:         http://localhost:3000
- Network:       http://192.168.1.2:3000
✓ Ready in 1137ms
```
3. Open the network URL. The page will fail to load and hot reloading will not be enabled. The console shows the following warning:
```
⚠ Blocked cross-origin request to Next.js dev resource /_next/hmr from "192.168.1.2".
Cross-origin access to Next.js dev resources is blocked by default for safety.

To allow this host in development, add it to "allowedDevOrigins" in next.config.js and restart the dev server:

// next.config.js
module.exports = {
  allowedDevOrigins: ['192.168.1.2'],
}

Read more: https://nextjs.org/docs/app/api-reference/config/next-config-js/allowedDevOrigins
```

Current behavior blocks the provided URL from working, because the resolved [`networkHostname`](https://github.com/vercel/next.js/blob/367e5215bf6fbf7a894d940df61e664b08538e22/packages/next/src/server/lib/start-server.ts#L350) is not included in the default [`allowedDevOrigins`](https://nextjs.org/docs/app/api-reference/config/next-config-js/allowedDevOrigins).

Expected behavior would be that the URL works out of the box, and hence gets added to the default `allowedDevOrigins`.

Alternatively, expected behavior could be dropping the network URL or marking it as unusable in the startup log when the allowedDevOrigins does not contain it, to avoid a developer thinking that it works.

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32409
  Available CPU cores: 12
Binaries:
  Node: 24.18.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.3.0-canary.97 // Latest available version is detected (16.3.0-canary.97).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Not sure

next dev (local)

### Additional context

Blocking behavior was enabled in https://github.com/vercel/next.js/pull/91507.

mei-tate:
**Title:**
`dev: warn at startup when network URL is printed that cross-origin access is blocked by default`

---

**What is the current behavior?**

When `next dev` starts, it prints both a Local URL and a Network URL in the startup output:

```
▲ Next.js 16.3.0-canary.97 (Turbopack)
- Local:         http://localhost:3000
- Network:       http://192.168.1.2:3000
✓ Ready in 1137ms
```

Opening the Network URL causes the page to fail to load and HMR to be blocked, with a warning only appearing in the browser console after the developer has already opened the URL and encountered the problem. There is no indication at startup that the Network URL requires additional configuration to work.

Fixes #96241

---

**What is the new behavior?**

When the dev server starts and a Network URL is present, a message is now printed immediately in the startup output explaining that cross-origin access is blocked by default and how to enable it:

```
▲ Next.js 16.3.0-canary.97 (Turbopack)
- Local:         http://localhost:3000
- Network:       http://192.168.1.2:3000
Cross-origin access to Next.js dev resources is blocked by default for safety.
If you need to access dev resources from another host, add it to "allowedDevOrigins" in next.config.js:

// next.config.js
module.exports = {
  allowedDevOrigins: ['your-host'],
}

Read more: https://nextjs.org/docs/app/api-reference/config/next-config-js/allowedDevOrigins
✓ Ready in 1137ms
```

The developer now sees the restriction and the fix in the same place they see the Network URL, before they open the browser.

---

**How was this implemented?**

A `logAllowedDevOriginsInfo` function was added inside `logStartInfo` in `app-info-log.ts`. It is called immediately after the Network URL is logged. The message reuses the existing `allowedDevOriginsDocs` URL already referenced elsewhere in the codebase and formats the output consistently with the surrounding bootstrap log style.

---

**Files changed:**
- `packages/next/src/server/lib/app-info-log.ts` — adds `logAllowedDevOriginsInfo` and calls it after the Network URL is printed

---

**Does this PR introduce a breaking change?**
No.

---

**Additional context**

The cross-origin blocking behavior was introduced in #91507. This PR does not change the blocking logic — it only ensures that the startup output informs developers of the restriction at the point where they first see the Network URL, rather than leaving them to discover it after opening a broken browser tab. No config reading is required since the message is unconditional — if there is a Network URL, the warning is always relevant.


packages/next/src/server/lib/app-info-log.ts
```diff
@@ -45,6 +64,9 @@ export function logStartInfo({
   if (networkUrl) {
     Log.bootstrap(`- Network:       ${networkUrl}`)
   }
+
+  logAllowedDevOriginsInfo()
+
```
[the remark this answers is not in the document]
mhverbakel: It even shows when the configuration is set up properly, and the message would be irrelevant as well.

packages/next/src/server/lib/app-info-log.ts
```diff
@@ -34,6 +34,25 @@ export function logStartInfo({
     ? ` (${bundlerName(getBundlerFromEnv())})`
     : ''
 
+  const allowedDevOriginsDocs =
+    'https://nextjs.org/docs/app/api-reference/config/next-config-js/allowedDevOrigins'
+
+  function logAllowedDevOriginsInfo() {
```
mhverbakel, line 40: This function is duplicated from another place where it prints a similar/the same message.

mei-tate: Thanks for the comments, I'll fix what I have and update the PR so I use existing code instead and make the message conditional so it doesn't show up at irrelevant instances.

PR #96284, opened 2026-07-28
[turbopack] Remove loader calls for dropped dynamic imports

sampoder:
For modules that we were not using, that we're asynchronously loading, we would still generate the loader but the file would not exist. This meant building a Next.js app with:

```javascript
// pure.js — side-effect free: nothing observable happens on evaluation
export const version = '1.0.0'

// app/page.tsx
export default async function Page() {
  const {} = await import('../pure')   // ← nothing destructured out
  return <p>hello world</p>
}
```

Would error with:

```
▲ Next.js 16.3.0-canary.87 (Turbopack)
  Creating an optimized production build ...

FATAL: An unexpected Turbopack error occurred.

> Build error occurred
Error [TurbopackInternalError]: Failed to write app endpoint /page

Caused by:
- ModuleId not found for ident: [project]/pure.js [app-rsc] (ecmascript, async loader)

Debug info:
...
- Execution of PatternMapping::resolve_request failed
- ModuleId not found for ident: [project]/pure.js [app-rsc] (ecmascript, async loader)
```  

Now, we re-write those loaders to empty promises:

```diff
 async function evaluationOnly() {
-    const {} = await __turbopack_context__.A("…/pure.js [test] (ecmascript, async loader)");
+    const {} = await Promise.resolve({});
 }
 async function named() {
     const { used } = await __turbopack_context__.A("…/named.js [test] (ecmascript, async loader)");
     console.log(used);
 }
 async function pattern(locale) {
     const {} = await __turbopack_context__.f({
         "./locales/effectful.js": {
             id: ()=>"…/locales/effectful.js [test] (ecmascript, async loader)",
             module: ()=>__turbopack_context__.A("…/locales/effectful.js …")
         },
         "./locales/pure.js": {
-            id: ()=>"…/locales/pure.js [test] (ecmascript, async loader)",
-            module: ()=>__turbopack_context__.A("…/locales/pure.js …")
+            id: ()=>undefined,
+            module: ()=>Promise.resolve({})
         }
     }).import(`./locales/${locale}.js`);
 }
```

It turns `UnusedReferences` into a map between references and target that can be dropped, where the target IDs are the unused targets. It then passes the reference through to `pattern_mapping.rs` so it can determine the `dropped_targets`. It needs to be target-specific as you can have a situation (like above) where some targets are pure and some aren't.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs
@@ -2,6 +2,7 @@ use std::{borrow::Cow, collections::HashSet};
 
 use anyhow::Result;
 use bincode::{Decode, Encode};
+use rustc_hash::FxHashSet;
 use swc_core::{
     common::DUMMY_SP,
     ecma::ast::{
@@ -21,6 +22,8 @@ use turbopack_core::{
         IssueExt, IssueSeverity, StyledString, code_gen::CodeGenerationIssue,
         module::emit_unknown_module_type_error,
     },
+    module::Module,
+    reference::ModuleReference,
     resolve::{
         ExternalType, ModuleResolveResult, ModuleResolveResultItem, origin::ResolveOrigin,
         parse::Request,
@@ -65,6 +68,8 @@ pub(crate) enum SinglePatternMapping {
     ModuleLoader(ModuleId),
     /// External reference with request and type
     External(RcStr, ExternalType),
+    /// The target was unused and dropped from the module graph, so nothing is loaded for it.
+    Dropped,
 }
 
 /// A mapping from a request pattern (e.g. "./module", `./images/${name}.png`)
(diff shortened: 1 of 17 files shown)


turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs
```diff
@@ -118,6 +123,8 @@ impl SinglePatternMapping {
             Self::Invalid => self.create_id(key_expr),
             Self::Unresolvable(request) => throw_module_not_found_expr(request),
             Self::Ignored => quote!("{}" as Expr),
+            // `const {a,b} = 0` assigns undefined to every binding.
+            Self::Dropped => quote!("0" as Expr),
```
bgw: nope

```
const {toString} = 0
```

Image

bgw (approved):

turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs
sampoder: oh wow - TIL - I guess so

PR #96286, opened 2026-07-28
Add browserOnly for client-only rendering

devjiwonchoi:
### Why?

Client Components sometimes need to opt a Suspense subtree out of server rendering while preserving the surrounding server-rendered shell. Existing APIs do not expose a promise-shaped primitive that can be consumed with `use()` and reused safely during client renders.

The Pages Router also did not pass a custom Fizz `onError` handler for page-shell renders. As a result, an expected `BailoutToCSRError` lost its digest, was logged on the server, and surfaced as a recoverable hydration error.

### How?

- Add `browserOnly()` to `next/navigation`. It is client-only, returns a shared fulfilled promise in the browser, and an immediately rejected CSR-bailout promise during server rendering.
- Cover Suspense boundaries, missing boundaries, RSC import restrictions, repeated renders, static export, Edge runtime, instant navigation, and instant validation.
- Preserve CSR bailout digests in Pages Router page-shell rendering while retaining React's default reporting for unexpected errors.
- Keep `_document` and other internal string renders unchanged by scoping the handler to the user page shell.

The first commit introduces the API and App Router behavior. The second adds Pages Router support independently.

### Test plan

- Dev and production with Turbopack and Webpack.
- Static export and Edge runtime with Turbopack and Webpack.
- Pages static optimization, GSSP, multiple Suspense boundaries, and a custom `_document`.
- Verify expected bailouts do not reach `onRequestError`, while real SSR errors remain reported.
- Verify existing Pages `next/dynamic({ ssr: false })` behavior remains unchanged.
- Verify instant navigation and instant validation behavior.

### DIFF
--- a/crates/next-core/src/next_import_map.rs
+++ b/crates/next-core/src/next_import_map.rs
@@ -985,9 +985,13 @@ async fn apply_vendored_react_aliases_server(
     if react_condition == "server" {
         // This is used in the server runtime to import React Server Components.
         alias.extend(
-            fxindexmap! {rcstr!("next/error") => rcstr!("next/dist/api/error.react-server"),
+            fxindexmap! {// ESM projects may add a `.js` suffix, which must use the same RSC entrypoints.
+            rcstr!("next/error") => rcstr!("next/dist/api/error.react-server"),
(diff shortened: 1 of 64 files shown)


gaearon:
Does this need a lower-level React counterpart? @acdlite

PR #96288, opened 2026-07-28
Fix Turbopack polling watcher write-time invalidation

### LINKED ISSUE #80665: Docs: how to enable turbopack's polling feature for the file watcher?

Based on #69684 / #68255, I believe that the Turbopack incremental builder should be able to detect file changes in Docker containers + WSL2.

However, I couldn't find [any doc explaining how to actually use it.](https://nextjs.org/docs/app/api-reference/turbopack#module-resolution) :-(

Is it ready for public use?

I'm using next.js in Docker + WSL2 and tried to use `next dev --turbopack` but noticed that it won't detect source file changes unless I remove the `--turbopack` option.
See issue #68255 for more context about this particular issue when watching files in this scenario.

https://nextjs.org/docs/app/api-reference/turbopack#module-resolution

shravan2x:
Fixes #80665, where users reported that the watch function didn't work in Turbopack.

Polling mode represents ordinary file edits as `ModifyKind::Metadata(MetadataKind::WriteTime)`. The code previously only handled `MetadataKind::Any`, preventing affected paths from being invalidated.

I tested this issue locally on a VM with a Windows host and a Linux guest. It fixes the issue.

I've added a regression test as well.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/watcher.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/watcher.rs
@@ -744,7 +744,12 @@ impl BatchedInvalidations {
             // invalidation, it's the way to reliably detect file content changes.
             // ref other implementation, i.e libuv does same thing to trigger `UV_CHANGES`
             // https://github.com/libuv/libuv/commit/73cf3600d75a5884b890a1a94048b8f3f9c66876
-            EventKind::Modify(ModifyKind::Data(_) | ModifyKind::Metadata(MetadataKind::Any)) => {
+            // The polling backend reports file changes as `MetadataKind::WriteTime` when the
+            // modification time changes, so it must be handled alongside `MetadataKind::Any`.
+            EventKind::Modify(
+                ModifyKind::Data(_)
+                | ModifyKind::Metadata(MetadataKind::Any | MetadataKind::WriteTime),
+            ) => {
                 for path in paths {
                     self.mark(path.into_boxed_path(), InvalidationFlags::PATH);
                 }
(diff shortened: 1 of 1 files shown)


bgw:
I'm going to take a pass on this:
- The watcher needs to be a bit more unit testable than it currently is.
- We should always use the non-recursive watcher when polling is enabled.
- We should only subscribe to `MetadataKind::All` on macos, and we should only subscribe to `MetadataKind::WriteTime` if polling watcher is enabled.

FWIW, you should really avoid using the polling watcher. You're going to get horrible CPU usage and slow HMRs because of it. It's much better to let the VM own the project filesystem if you can instead of trying to mount it in the VM, so that `inotify` works properly.

shravan2x:
@bgw I appreciate the quick review!

We're running our development stacks in a VM, in this case. So our IDEs, coding agents, etc are still on the host. The VM we use (Docker Sandboxes) doesn't support FS event forwarding. I figured that getting the VM vendor to fix those modules would be much harder than submitting a fix to Turbopack myself and accepting the polling cost.

In practice, I'm seeing that compile times are a much bigger problem (many seconds) than the time to detect the file changes alone (100s of ms). Per cold page, we see ~1.5s on the host with turbopack+events, ~9s with webpack+polling, and ~3.5s with turbopack 16.3+polling. The 3.5s is _just barely acceptable_ for daily use, given the other benefits the dev sandbox brings. So I'm spending my time on the compile perf right now rather than polling.

> I'm going to take a pass on this:

To clarify, are you saying you want to take over the PR and redo it? Or that your pass was the code review itself?

If it was the latter, I'm happy to address the issues you mentioned.

PR #96297, opened 2026-07-28
Gate partial fallback shell upgrades behind `partialPrefetching` for `next start`

acdlite:
In #96074 I put the `partialFallback` behavior behind the `partialPrefetching` flag, so that we don't risk an explosion in ISR costs for apps that haven't opted into Partial Prefetching. But it turns out I only handled this for deploy mode — i.e. the build output consumed by the Vercel adapter. The behavior was still unconditionally on for `next start`.

The reason it was easy to miss is that `next start` and the adapter express partial fallback shells through completely separate mechanisms. The adapter emits a `partialFallback` flag into the build output and lets the platform's ISR layer perform the shell upgrade. `next start` has no such flag; the server performs the upgrade itself, in the compiled page runtime. So the gate I added to the adapter output had no effect on the self-hosted path.

To understand how the `next start` path was configured, I looked at the PR that removed the original top-level `partialFallbacks` config flag (#93859). Before that PR, the behavior was gated in the runtime; when the flag was removed those gates were deleted and the behavior became always-on.

The important subtlety is that only the *upgrade* should be gated, not the shell machinery as a whole. On `next start`, the value that decides whether a shell can be specialized (`remainingPrerenderableParams`) also drives core Cache Components behavior — build-time partial prerendering of params and serving build-time sub-shells — which must stay on regardless of `partialPrefetching`. So this gates only the two things that make up the "upgrade on first request" cost risk: the background ISR revalidation that specializes a shell per request, and the client-facing `isFallbackUpgradeable` signal that tells the client to retry a prefetch waiting for that upgrade.

Adds coverage with a `partialPrefetching`-disabled fixture asserting that fallback shells stay shared rather than specializing per request.

### DIFF
--- a/packages/next/src/build/index.ts
+++ b/packages/next/src/build/index.ts
@@ -3135,8 +3135,12 @@ export default async function build(
                     _isRoutePPREnabled: isRoutePPREnabled,
                     _allowEmptyStaticShell: !route.throwOnEmptyStaticShell,
                     // A fallback shell can only be upgraded if at least one of
-                    // its fallback params is a `generateStaticParams` candidate.
+                    // its fallback params is a `generateStaticParams` candidate,
+                    // and only when Partial Prefetching is enabled. Otherwise
+                    // nothing ever performs the upgrade, so flagging the shell
+                    // would only cause the client to retry the prefetch.
                     _isFallbackUpgradeable:
+                      Boolean(config.partialPrefetching) &&
                       (route.remainingPrerenderableParams?.length ?? 0) > 0,
                   }
                 })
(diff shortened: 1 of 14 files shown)


test/e2e/app-dir/segment-cache/prefetch-fallback-retry/prefetch-fallback-retry.test.ts
```diff
@@ -67,54 +67,40 @@ describe('segment cache prefetch fallback retry', () => {
...
+    })
+    expect(await res.text()).not.toContain('Post body for recovery')
+
+    // No retry fires even after advancing well past the retry interval.
+    await act(async () => {
+      await page.clock.fastForward(5000)
+    }, 'no-requests')
   })
 
   it('retries a static fallback prefetch a bounded number of times, then stops', async () => {
```
ztanner, line 120: Should this test run under a fixture with `partialPrefetching: true`? With the current config, `isUpgradeableISRFallback` should always be false, so the retry loop never starts and the `< 5` assertion can pass with only the initial request.

ztanner (approved):

PR #96301, opened 2026-07-28
fix(next/image): improve test coverage for unsupported image types

styfle:
We were missing some test coverage for image optimization. This PR adds tests for many missing image formats.

Additionally, this PR also ensures the sharp encoders match that allowlist from detectContentType() such that future versions of sharp that enable new encoders by default are not automatically enabled in next/image.

Practically, there is no change to the end user because detectContentType() runs before sharp so that is the true allowlist, but this gives us a little more control to align the sharp allowlist as well (since sharp can release new features in semver minor).


timneutkens (approved):

icyJoseph (approved):
Putting a note on my work queue here, we should be more explicit about formats are supported in the docs too. Today it is implicitly done through the patterns we show.

ceolinwill:
@styfle I think we need to allow SVG too. Right now, it seems to break `next/og`. See #96612 and #96621

PR #96302, opened 2026-07-28
docs(interactive-apps): update prefetching for unified runtime prefetching

aurorascharff:
### What?

Update the interactive apps guide's prefetching section for unified runtime prefetching.

### Why?

#96106 removed the `allow-runtime` segment config, and the guide's section still described the old model after the mechanical rename: it told readers session data (cookies, headers) needs the link opt-in, when the App Shell already carries session-gated UI. Only URL data needs `prefetch={true}`.

### How?

Rework the section around the per-link model: the shell carries static, cached, and session output, `prefetch={true}` resolves URL data ahead of the click, and the static prefetch attempt (#96095) makes the server cost an internal optimization. Includes a technical-writing pass on the section (full-noun subjects, split splices) and aligns the Next steps table row.


docs/01-app/02-guides/interactive-apps.mdx
```diff
@@ -738,17 +738,17 @@ Every Server Function follows the same shape: run the write, then call `updateTa
...
-
-If the destination's cached reads depend on request data (cookies, headers, `searchParams`, or `params` not resolved by [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params)), opt the link into [runtime prefetching](/docs/app/guides/runtime-prefetching) with `prefetch={true}` so those reads resolve at prefetch time:
+Under [Partial Prefetching](/docs/app/guides/adopting-partial-prefetching), [`<Link>`](/docs/app/api-reference/components/link) prefetches the destination's [App Shell](/docs/app/glossary#app-shell) by default, which carries the route's static, cached, and session output. [URL data](/docs/app/glossary#url-data) like [`searchParams`](/docs/app/api-reference/file-conventions/page#searchparams-optional) and [`params`](/docs/app/api-reference/file-conventions/page#params-optional) varies per link, so the shell can't carry it. To resolve URL data ahead of the click, opt the link into [runtime prefetching](/docs/app/guides/runtime-prefetching) with `prefetch={true}`:
```
icyJoseph: > which carries the route's static, cached, and session output.

I think "session output", or even (static, cached and session -> output), is a bit too vague. 

The App Shell is the RSC payload rendered with session data included, and the contrast is that the App Shell does not use URL Data to render that payload, by definition, because it makes the result link specific

icyJoseph (approved):

PR #96304, opened 2026-07-28
[cd] Send Slack notification if publish fails

eps1lon:
Failing to publish (especially a partial publish) leaves Next.js in a corrupted state that needs to be addressed immediately. However, we only created alerts when deploy tests failed which requires a successful publish. 

Now we send special message to the same channel as broken canary and deploy tests go.

This won't catch a failing deploy-target job. We'll see if deploy-target also frequently fails/flakes.

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -660,6 +660,42 @@ jobs:
           path: pnpm-publish-summary.json
           if-no-files-found: ignore
 
+  reportPublishFailure:
+    name: report publish failure to slack
+    needs:
+      - deploy-target
+      - publishRelease
+    # Only alert for runs that would publish (production == v* tag push) where
+    # the publish did not succeed. In a production run publishRelease's own `if`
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -660,6 +660,42 @@ jobs:
...
+    steps:
+      - name: send webhook
+        uses: slackapi/slack-github-action@6c661ce58804a1a20f6dc5fbee7f0381b469e001 # v1.25.0
+        with:
+          # The url is intentionally missing the protocol,
+          # allowing it to be transformed into an actual link in the Slack workflow
+          # (through slightly hacky means).
+          payload: |
+            {
+              "commit_title": ${{ toJSON(github.event.head_commit.message) }},
```
ztanner: Would `github.event.head_commit.message` be empty in the case where this workflow gets triggered through `workflow_dispatch`?

ztanner (approved):

PR #96322, opened 2026-07-29
Share one Ecmascript chunk version type between browser and node HMR

wbinnssmith:
## Summary

`turbopack-nodejs` and `turbopack-browser` each carry a full copy of the Ecmascript chunk-content and HMR-versioning stack — roughly 770 near-verbatim duplicated lines. `turbopack-ecmascript/src/chunk_list/merged_update.rs` already documents the intent to share this, noting the turbo-tasks value types "cannot be generic and therefore remain per-runtime".

This is **phase 1 of 3** in unifying that stack. Since turbo-tasks value types cannot be generic, the approach is one shared concrete type per duplicated pair, with per-platform differences carried by fields rather than type parameters.

Planned sequence, each phase independently shippable:

1. **This PR** — collapse `EcmascriptBrowserChunkVersion` and `EcmascriptBuildNodeChunkVersion` into a single `EcmascriptHmrChunkVersion` in a new `turbopack-ecmascript/src/chunk_hmr/`.
2. A `value_trait` over chunk content, plus the shared single-chunk diff. Both existing content types implement it in a few lines, so no struct or `ChunkingContext` changes are needed.
3. The `merged/` quartet (merger, content, version, diff), unblocked by 1 and 2.

Phases 1–3 remove ~530 of the duplicated lines without widening any struct fields. Unifying the chunk-content structs themselves is deliberately deferred and optional: it requires widening to `Box<dyn ChunkingContext>`, and both chunking contexts define `minify_type` twice (an inherent method and a `ChunkingContext` method whose trait default is `NoMinify`), so that widening needs its own verification. The trait in phase 2 means the two content structs can coexist indefinitely at no architectural cost.

The runtime, evaluate, and entry chunk types are out of scope — `get_browser_runtime_code` takes eight more parameters than `get_node_runtime_code`, so those are genuine behavioral divergence rather than copy-paste.

### Behavior changes

Three intentional changes, all in HMR bookkeeping that never reaches emitted output:

- Browser chunk version IDs now hash `minify_type`, matching node. Version IDs are opaque and recomputed in-memory on both sides of an HMR connection, so this cannot desync a client; browser dev never minifies.
- Node error text said "chunk path ... is not in client root", a copy-paste artifact. Now says "output root".
- `#[turbo_tasks(trace_ignore)]` is now applied to node’s merged chunk version, matching the browser copy. Landed here because the field type changed anyway.

## Verification

- `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets`, `cargo fmt --check` — all clean
- `cargo test -p turbopack-tests` — 112 snapshot tests pass with **zero fixture changes**, confirming emitted chunk bytes are unchanged
- `pnpm test-dev-turbo test/development/app-dir/server-hmr/server-hmr.test.ts` — 16/16, the primary gate for the node HMR path
- `pnpm test-dev-turbo test/development/app-hmr/hmr.test.ts test/development/app-dir/hmr-dep-accept/hmr-dep-accept.test.ts test/development/app-dir/hmr-shared-css/hmr-shared-css.test.ts` — 11/11, client HMR path


wbinnssmith:
Folded into #96325, which now contains both the shared chunk version type and the shared diffing/merging machinery as a single commit.

PR #96323, opened 2026-07-29
test: reproduce empty parallel route scroll regression

ztanner:
## Summary

Adds a minimal App Router regression test for a soft navigation where an empty parallel route is rendered after persistent page content.

The test runs with `appNewScrollHandler` both disabled and enabled. The legacy-handler case passes normally, while the new-handler case uses `it.failing` to highlight the regression.


DavidIlie:
made a fix here #96342.

eps1lon (approved):

PR #96325, opened 2026-07-29
Share Ecmascript HMR chunk versioning, diffing, and merging between browser and node

wbinnssmith:
## Summary

`turbopack-nodejs` and `turbopack-browser` each carried their own copy of the Ecmascript chunk versioning and HMR diffing stack. `chunk_list/merged_update.rs` already documented the intent to share this, noting the turbo-tasks value types "cannot be generic and therefore remain per-runtime". A `value_trait` resolves that: the shared machinery is written against `EcmascriptHmrChunkContent`, which each runtime implements in a few lines to expose `entries()` and `own_version()`.

The shared code now lives in `turbopack-ecmascript/src/hmr/` (`version.rs`, `update.rs`, `content.rs`, `merger.rs`). Both runtimes' `ecmascript/version.rs`, `ecmascript/update.rs`, and `ecmascript/**/merged/` directories are deleted.

No struct fields were widened and no `ChunkingContext` methods changed.

### Behavior changes

- Browser chunk version IDs now hash `minify_type`, matching node. IDs are opaque and recomputed in-memory on both ends of an HMR connection, so this cannot desync a client.
- The shared diff uses node's lazy `entries()` materialization, so chunks with only deletions no longer materialize any `Vc<Code>`. This now applies to the browser path too.
- Node error text said "chunk path ... is not in client root", a copy-paste artifact. Now says "output root".
- `#[turbo_tasks(trace_ignore)]` is applied consistently to the merged chunk version; previously only the browser copy had it.

Browser chunk content keeps the default `VersionedContent::update`. Overriding it bypasses the `ChunkListUpdate` envelope the client runtime expects.

### Not included

Unifying the chunk-content structs. That requires widening to `Box<dyn ChunkingContext>`, and both chunking contexts define `minify_type` twice — an inherent method and a `ChunkingContext` method whose trait default is `NoMinify` — so it needs separate verification. The trait added here lets the two structs coexist at no cost.

The runtime, evaluate, and entry chunk types are also out of scope: `get_browser_runtime_code` takes eight more parameters than `get_node_runtime_code`, so those differ behaviorally rather than by duplication.


turbopack/crates/turbopack-ecmascript/src/hmr/content.rs
```diff
@@ -5,26 +5,30 @@ use turbopack_core::{
...
+/// Composite [`VersionedContent`] that is the result of merging multiple
+/// Ecmascript chunk contents together through the
+/// [`super::merger::EcmascriptChunkContentMerger`]. This allows a chunk
+/// list to produce a single `EcmascriptMergedUpdate` for multiple chunks updating
+/// at the same time.
 ///
-/// [`EcmascriptChunkContentMerger`]: super::merger::EcmascriptChunkContentMerger
+/// "Merged" here refers to [`turbopack_core::version::VersionedContentMerger`]
+/// combining several chunks into one update payload. It has nothing to do with
+/// sharing this code between the browser and node runtimes.
```
bgw: This comment seems overly specific.

```suggestion
/// "Merged" here refers to [`turbopack_core::version::VersionedContentMerger`]
/// combining several chunks into one update payload.
```

turbopack/crates/turbopack-ecmascript/src/hmr/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+/// to expose the two things the shared machinery needs.
+#[turbo_tasks::value_trait]
+pub trait EcmascriptHmrChunkContent: VersionedContent {
+    /// The per-module code and hashes making up this chunk.
+    #[turbo_tasks::function]
+    fn entries(self: Vc<Self>) -> Vc<EcmascriptChunkContentEntries>;
+
+    /// This chunk's own version, used as the `to` side of a diff.
+    #[turbo_tasks::function]
+    fn own_version(self: Vc<Self>) -> Vc<EcmascriptChunkVersion>;
```
bgw: These methods are stubs that mostly just call other turbo-task functions? Maybe better for these to just be normal async methods on `Vc<Self>` that aren't cached `#[turbo_tasks::function]`s.

bgw: Given that this is just a ecmascript-specific variant of `VersionedContent::version`, maybe `ecmascript_chunk_version` is a better name than `own_version`? `own_version` seems very vague, and it's unclear how it's supposed to be different than `version`.

bgw (approved):
Thanks for deduping all of this!

PR #96327, opened 2026-07-29
fix: preserve telemetry timeout when an abort signal is provided

### LINKED ISSUE #96291: next build can wait far beyond the built-in telemetry timeout when a caller signal is provided

https://github.com/cyhforlight/next-telemetry-timeout-repro

### To Reproduce

1. Clone the reproduction repository:

```bash
git clone https://github.com/cyhforlight/next-telemetry-timeout-repro.git
cd next-telemetry-timeout-repro
```

2. Build the Docker image:

```bash
docker build -t next-telemetry-timeout-repro .
```

3. Run the reproduction:

```bash
docker run --rm \
  --add-host telemetry.nextjs.org:127.0.0.1 \
  next-telemetry-timeout-repro
```

4. Observe that the telemetry-disabled build finishes normally, while the telemetry-enabled build takes significantly longer than the built-in five-second timeout.

The script prints `BUG REPRODUCED` and exits with code 0 when the telemetry-attributable delay exceeds the eight-second reproduction threshold.

### Current behavior

The same unmodified project is built twice in the same container, with `.next` removed between runs. The only relevant difference is whether `NEXT_TELEMETRY_DISABLED=1` is set.

With Next.js 16.2.6:

```text
telemetry disabled: 2.2s (exit 0)
telemetry enabled:  23.4s (exit 0)
attributable to telemetry: 21.2s
```

The telemetry request's built-in five-second timeout is not installed when the caller provides a signal. The production build path passes `postController.signal`, so a stalled request falls back to the HTTP client's substantially longer timeouts.

### Expected behavior

The caller-provided cancellation signal should not replace the telemetry request timeout. A stalled telemetry endpoint should stop delaying `next build` after approximately five seconds, while caller-initiated cancellation should continue to work.

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #1 SMP PREEMPT_DYNAMIC Fri Jun 5 01:12:21 UTC 2026
  Available memory (MB): 15746
  Available CPU cores: 16
Binaries:
  Node: 20.20.2
  npm: 10.8.2
  Yarn: 1.22.22
  pnpm: N/A
Relevant Packages:
  next: 16.3.0-canary.97
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: N/A
Next.js Config:
  output: N/A
```

Runtime, Performance

next build (local)

### Additional context

I originally encountered this while running `next build` behind a TUN/Fake-IP proxy, where the telemetry connection stopped making progress. The reproduction removes the proxy dependency and replaces it with a deterministic local TCP endpoint.

I also tested the same reproduction against `next@16.3.0-canary.97` without changing or instrumenting the Next.js runtime:

```text
telemetry disabled: 3.0s (exit 0)
telemetry enabled:  45.1s (exit 0)
attributable to telemetry: 42.2s
```

The relevant condition is still present on canary:

https://github.com/vercel/next.js/blob/canary/packages/next/src/telemetry/post-telemetry-payload.ts

A related report previously described `telemetry.flush()` waiting indefinitely, but maintainers could not reproduce it consistently:

https://github.com/vercel/next.js/discussions/71063

This repository provides a deterministic reproduction and an A/B comparison with telemetry disabled.

I opened #96327 with a candidate fix and regression tests.

cyhforlight:
## What?

This change preserves the built-in five-second telemetry timeout regardless of whether the caller provides a `signal`, while still allowing the caller to cancel the request early.

## Why?

The current implementation only creates the built-in five-second timeout signal when the caller does not provide a `signal`. However, the production caller used by the build path always passes `postController.signal`, so the timeout signal is never created.

When a telemetry request stalls, it can therefore remain pending until the underlying fetch implementation or network stack times it out, causing `next build` to wait far longer than five seconds.

Fixes #96291

## How?

The previous implementation was:

```typescript
if (!signal && 'timeout' in AbortSignal) {
  signal = AbortSignal.timeout(5000)
}
```

This only created the timeout signal when the caller did not provide a `signal`. Since the production caller always passes `postController.signal`, the build path never used the timeout.

The implementation now combines both signals:

```typescript
const timeoutSignal = AbortSignal.timeout(5000)
signal = signal ? AbortSignal.any([signal, timeoutSignal]) : timeoutSignal
```

This allows both cancellation conditions to remain effective:

- The timeout signal aborts a stalled telemetry request after five seconds.
- The caller can still abort the request earlier when Next.js no longer needs to wait for it.

This change also removes the Edge Runtime `AbortController` ponyfill import from `storage.ts`, allowing the production caller to use Node.js's native `AbortController`. This ensures that both signals passed to `AbortSignal.any()` use the native Node.js implementation and that caller cancellation propagates correctly to the combined signal.

Regression tests were added to verify that:

- The five-second timeout is preserved when the caller provides a signal.
- Caller cancellation propagates to the combined signal.

The standalone file-tracing snapshot was also updated because the Edge Runtime ponyfill is no longer referenced.

## Testing

- Telemetry unit tests
- Next.js type checking
- `pnpm build-all`
- `next-server-nft` production tests


gesposito:
The ponyfill removal here fixes more than the timeout. On Node 24+ undici brand-checks `signal`, and the `@edge-runtime` ponyfill fails it, so telemetry silently sends nothing:

`TypeError: RequestInit: Expected signal ("AbortSignal {}") to be an instance of AbortSignal.`

Same class name, different class — `signal instanceof AbortSignal` is `false`. It's a single chokepoint: one `new AbortController()` at `storage.ts:327` and one `postNextTelemetryPayload` call at `:328`, so every producer (build, dev, MCP, SWC load failure, trace reporting) goes through it.

Real `next build` on stock 16.3.1, telemetry enabled, same build twice, only the signal changed:

| signal | attempts | delivered |
|---|---|---|
| ponyfill (as shipped) | 14 | 0 |
| native | 14 | 7 × 200 |

Lost: `NEXT_CLI_SESSION_STARTED`, `NEXT_PACKAGE_DETECTED`, `NEXT_BUILD_FEATURE_USAGE`, `NEXT_BUILD_COMPLETED`, `NEXT_TYPE_CHECK_COMPLETED`, `NEXT_BUILD_OPTIMIZED`. Build-failure telemetry too.

Version boundary, one ponyfill signal vs one native signal against a local server:

| Node | undici | ponyfill | native |
|---|---|---|---|
| 18.20.8 | 5.29.0 | 200 | 200 |
| 20.20.2 | 6.24.1 | 200 | 200 |
| 22.23.2 | 6.28.0 | 200 | 200 |
| 23.11.1 | 6.21.2 | 200 | 200 |
| 24.0.0 | 7.8.0 | TypeError | 200 |
| 25.8.0 | 7.22.0 | TypeError | 200 |

`instanceof` is false on every row — undici 7 just started enforcing it, so this is latent breakage surfacing, not a new regression. All of Node 24 is affected, and it is the active LTS while `next` declares `engines: >=20.9.0`.

This may also explain #96291's 21s stall: that needs an open socket, so it must be Node ≤23. On 24+ the request cannot stall because it never starts — two symptoms, one cause, split by undici version.

`post-telemetry-payload.test.ts` misses it because it mocks `global.fetch` and passes no `signal`, which is the one branch production never takes.

cyhforlight:
@gesposito Thanks for the detailed investigation and version matrix. This is valuable additional context and provides further justification for removing the ponyfill, especially on Node 24+.

cyhforlight:
Hi! This PR has been open for over a month. The latest build-and-test and build-and-deploy workflows are still awaiting maintainer approval, while the available checks are passing. The fix includes focused regression coverage, along with the additional Node 24 validation above. Could someone approve the workflows when convenient? Happy to adjust the approach if needed.

PR #96330, opened 2026-07-29
Force no-store on prerendered responses that carry a Set-Cookie header

### LINKED ISSUE #96266: Middleware Set-Cookie on prerendered static pages produces cacheable responses with s-maxage=31536000 - CDN caches and serves auth tokens to other users

https://github.com/uiYzzi/nextjs-middleware-setcookie-cache-leak

### To Reproduce

1. Clone the reproduction repo above
2. `npm install && npm run build && npm start`
3. Run the following curl command:

```bash
# Request 1: User A (has rt cookie, middleware sets Set-Cookie)
curl -sI -b "rt=USER_A" http://localhost:3000/
```

**Observed response:**
```
HTTP/1.1 200 OK
set-cookie: at=access_token_for_USER_A; Path=/; HttpOnly; SameSite=lax
set-cookie: rt=refresh_token_for_USER_A; Path=/; HttpOnly; SameSite=lax
x-nextjs-cache: HIT
x-nextjs-prerender: 1
Cache-Control: s-maxage=31536000
```

The response has **both** `Set-Cookie` (sensitive tokens) and `Cache-Control: s-maxage=31536000` (cacheable for 1 year).

**Current:** When middleware adds `Set-Cookie` headers to a `NextResponse.next()` response for a prerendered (static) page, Next.js does not override the page's `Cache-Control: s-maxage=31536000` to `no-store`. The response contains both `Set-Cookie` and `s-maxage=31536000`.

**Expected:** When middleware adds `Set-Cookie` to a response, Next.js should set `Cache-Control: no-store` (or `private, no-cache`), because:
- RFC 6265 (HTTP State Management Mechanism) Section 7.3: shared caches must not store or distribute responses containing `Set-Cookie` headers to other clients, to prevent credential leakage
- A response with `Set-Cookie` contains user-specific data and is unsafe for shared CDN caching

```
Operating System:
  Platform: darwin
  Arch: arm64
Binaries:
  Node: 22.0.0
  npm: 10.0.0
Relevant Packages:
  next: 16.2.12
  react: 19.1.0
  react-dom: 19.1.0
Next.js Config:
  output: N/A
```

Also verified on Next.js 15.5.9 with the same result.

Middleware, Cookies

### Additional context

**Impact: Critical (Security)**

When deployed behind a non-Vercel CDN (e.g., AWS Amplify Hosting's managed CloudFront), this causes **cross-user authentication token leakage**: the CDN caches the response including `Set-Cookie` headers, and serves User A's tokens to User B on cache hits.

Vercel's [caching docs](https://vercel.com/docs/edge-network/caching) explicitly state that responses with `Set-Cookie` headers are NOT cached on Vercel's CDN. However, other CDNs (CloudFront, Cloudflare, Netlify) may cache responses with `Set-Cookie` when `s-maxage` is present, because the response explicitly declares itself cacheable.

This affects a very common pattern: BFF/OAuth architectures where middleware refreshes expired access tokens via `Set-Cookie`. Real-world incident: users were logged in as a different user they had never authenticated as.

**Suggested fix:** When `Set-Cookie` headers are present on a middleware `NextResponse.next()` response, automatically override `Cache-Control` to `no-store`:

```typescript
if (response.headers.has('set-cookie')) {
  response.headers.set('cache-control', 'no-store, must-revalidate');
}
```

Related issues:
- Amplify Hosting: https://github.com/aws-amplify/amplify-hosting/issues/4100
- Supabase similar incident: https://github.com/supabase/supabase-js/issues/1682
- Vercel CDN caching docs: https://vercel.com/docs/edge-network/caching
- RFC 6265 Section 7.3 (Cookie security): shared caches must not distribute responses with `Set-Cookie` to other clients

lazerg:
When middleware adds a `Set-Cookie` header to a `NextResponse.next()` response on a prerendered static page, Next.js leaves the page's `Cache-Control: s-maxage=31536000` untouched. The response then advertises itself as cacheable for a year while carrying user-specific cookies, so shared CDNs that honor `s-maxage` (CloudFront, Cloudflare, Netlify) can store one user's tokens and serve them to the next visitor.

`sendRenderResult` now swaps the cacheable header for `no-store` whenever the response already carries a `Set-Cookie`, in line with RFC 6265 §7.3. Responses without cookies keep their normal caching. Added a production test covering both the cookie and no-cookie cases.

Fixes #96266

### DIFF
--- a/test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
+++ b/test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
@@ -3,14 +3,12 @@ import { nextTestSetup } from 'e2e-utils'
 describe('middleware-set-cookie-cache', () => {
   const { next, skipped } = nextTestSetup({
     files: __dirname,
-    skipStart: true,
+    // Cache-Control is applied outside of the Next.js server when deployed, so
+    // these assertions only hold for `next start`.
     skipDeployment: true,
   })
 
   if (skipped) return
-
-  beforeAll(() => next.start())
-  afterAll(() => next.stop())
 
   it('replaces a prerendered page cache-control with no-store when middleware sets a cookie', async () => {
     const res = await next.fetch('/with-cookie')
--- a/test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
+++ b/test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
@@ -1,0 +1,28 @@
+import { nextTestSetup } from 'e2e-utils'
+
+describe('middleware-set-cookie-cache', () => {
+  const { next, skipped } = nextTestSetup({
+    files: __dirname,
+    skipStart: true,
+    skipDeployment: true,
+  })
+
+  if (skipped) return
+
+  beforeAll(() => next.start())
+  afterAll(() => next.stop())
+
+  it('replaces a prerendered page cache-control with no-store when middleware sets a cookie', async () => {
+    const res = await next.fetch('/with-cookie')
+    expect(res.headers.get('set-cookie')).toContain('token=secret-user-token')
+    expect(res.headers.get('cache-control')).toBe(
+      'private, no-cache, no-store, max-age=0, must-revalidate'
+    )
+  })
+
+  it('keeps the prerendered page cache-control when middleware sets no cookie', async () => {
+    const res = await next.fetch('/no-cookie')
+    expect(res.headers.get('set-cookie')).toBeNull()
+    expect(res.headers.get('cache-control')).toContain('s-maxage')
+  })
+})
--- a/packages/next/src/server/send-payload.ts
+++ b/packages/next/src/server/send-payload.ts
@@ -58,9 +58,9 @@ export async function sendRenderResult({
   // If cache control is already set on the response we don't
   // override it to allow users to customize it via next.config
   if (cacheControl && !res.getHeader('Cache-Control')) {
-    // A response carrying `Set-Cookie` holds user-specific data and must not be
-    // stored by shared caches (RFC 6265 §7.3), so never let a prerendered page's
-    // cacheable `s-maxage` header leak cookies across users on non-Vercel CDNs.
+    // `Set-Cookie` does not stop a shared cache from storing and reusing a
+    // response (RFC 9111 §7.3), so a prerendered page's cacheable `s-maxage`
+    // would otherwise leak one user's cookies to the next visitor.
     if (res.getHeader('Set-Cookie')) {
       res.setHeader(
         'Cache-Control',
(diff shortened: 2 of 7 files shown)


test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
```diff
@@ -0,0 +1,28 @@
...
+  const { next, skipped } = nextTestSetup({
+    files: __dirname,
+    skipStart: true,
+    skipDeployment: true,
+  })
+
+  if (skipped) return
+
+  beforeAll(() => next.start())
+  afterAll(() => next.stop())
```
eps1lon: It automatically stops/starts once you remove skipStart.

test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
```diff
@@ -0,0 +1,28 @@
+import { nextTestSetup } from 'e2e-utils'
+
+describe('middleware-set-cookie-cache', () => {
+  const { next, skipped } = nextTestSetup({
+    files: __dirname,
+    skipStart: true,
+    skipDeployment: true,
```
eps1lon, line 8: Why does this not work when deployed on Vercel?

lazerg:
None of the red checks come from this change. The turbopack dev shard failed on a 502 while downloading fnm before any test ran, and the webpack stats job was cancelled; both are transient CI infra flakes that a re-run should clear.

lazerg:
The turbopack dev (6/7) failure is a transient 502 fetching the Node binary during FNM setup, unrelated to this diff. Stats (webpack) just got cancelled as fallout from that. I don't have permission to re-run the job myself.

eps1lon (commented):
Is this for CDNs that don't implement the RFC by ignoring `Cache-Control` on `Set-Cookie` responses?

test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
lazerg: You are right, removed both. The two cases still pass locally with `test-start-turbo`.

test/production/app-dir/middleware-set-cookie-cache/middleware-set-cookie-cache.test.ts
lazerg: Prerenders go to the platform with their `initialHeaders`, so on Vercel the client-facing `Cache-Control` comes from the proxy rather than from `sendRenderResult`, same reason `custom-cache-control` skips deploy. Added a comment explaining that. Didn't get a chance to verify it against an actual Vercel deployment though, happy to drop the skip if you'd rather see it run there.

lazerg:
Actually it's the opposite: `RFC 9111 §7.3` says `Set-Cookie` doesn't inhibit caching, and the storage conditions in `§3` never mention it either. A spec-compliant shared cache is free to store `s-maxage=31536000` alongside the cookie and replay it, so Cache-Control is the only thing stopping that.

Also the `RFC 6265 §7.3` reference in the description was wrong, that section is about cookie expiration dates, not caching. Fixed the code comment to point at `RFC 9111 §7.3` instead.

lazerg:

lazerg:
This has gone quiet for a few weeks, so I'll close it for now. Happy for it to be reopened or merged whenever it's convenient.

PR #96334, opened 2026-07-29
fix: add accessible label to icon-only link

niketchandivade:
## Summary

This PR improves the accessibility of the Ant Design example by adding an accessible label to the icon-only `Link`.

### Changes

- Added `aria-label="Home"` to the icon-only link.

### Why

The link currently renders only an icon, so it does not provide an accessible name for assistive technologies. Adding an `aria-label` makes its purpose clear to screen reader users without changing the visual appearance.

### Before

Screen readers announce the element as an unlabeled link.

### After

Screen readers announce the link as **"Home"**.


marcoshernanz (approved):

marcoshernanz:
Thanks for your contribution!

PR #96342, opened 2026-07-29
Fix empty Fragment scroll ownership

DavidIlie:
## Summary

- treat Fragment instances without client rects as unavailable route scroll targets
- leave the shared scroll signal unconsumed so another changed segment can handle an empty route Fragment
- avoid focus and scroll side effects when an empty segment cannot claim ownership
- consume and clear an explicit hash intent when no matching target exists, without falling back to route content
- cover empty and visible intercepted routes across both scroll handlers, including focus preservation and real hash scrolling

This PR is stacked on #96323, which provides the minimal reproduction.

## Verification

- `HEADLESS=true pnpm test-dev-turbo test/e2e/app-dir/parallel-routes-scroll-owner/parallel-routes-scroll-owner.test.ts` (8/8)
- `HEADLESS=true pnpm test-dev-webpack test/e2e/app-dir/parallel-routes-scroll-owner/parallel-routes-scroll-owner.test.ts` (8/8)
- `HEADLESS=true pnpm test-dev-turbo test/e2e/app-dir/navigation-focus/navigation-focus.test.ts` (5/5)

### DIFF
--- a/test/e2e/app-dir/parallel-routes-scroll-owner/parallel-routes-scroll-owner.test.ts
+++ b/test/e2e/app-dir/parallel-routes-scroll-owner/parallel-routes-scroll-owner.test.ts
@@ -51,18 +51,23 @@ describe.each([false, true])(
       )
     })
 
-    it('preserves scroll and focus when the hash target is missing', async () => {
+    it('consumes a missing hash without scrolling or blurring', async () => {
       const browser = await next.browser('/modal')
 
       const initialScroll = await browser.eval(`
           window.scrollTo(0, 1200)
           document.getElementById('focus-target').focus({ preventScroll: true })
-          document.getElementById('open-empty-modal-missing-hash').click()
+          document.getElementById('open-modal-missing-hash').click()
           window.scrollY
         `)
       await retry(async () => {
         expect(await browser.url()).toBe(
-          `${next.url}/modal/open#missing-target`
+          `${next.url}/modal/visible#missing-target`
+        )
+      })
+      await retry(async () => {
+        expect(await browser.elementByCss('#visible-modal').text()).toBe(
+          'Visible modal'
         )
       })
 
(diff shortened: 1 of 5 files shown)


test/e2e/app-dir/parallel-routes-scroll-owner/parallel-routes-scroll-owner.test.ts
```diff
@@ -13,26 +13,82 @@ describe.each([false, true])(
...
+
+      await browser.eval(
+        'new Promise((resolve) => requestAnimationFrame(() => requestAnimationFrame(resolve)))'
+      )
+      expect(await browser.eval('document.activeElement.id')).toBe(
+        'focus-target'
+      )
+    })
+
+    it('preserves scroll and focus when the hash target is missing', async () => {
```
ztanner: I think there is a subtle bug and this is passing incidentally because it's flowing into the new NoClientRects path. 

I think we need to distinguish between two cases: 

- No hash + empty fragment: leave the signal in place for the changed segment
- URL has a hash, but no matching target exists: consume and clear the hash scroll intent without falling back to the fragment or blurring focus.

DavidIlie: :robot: Good catch — the missing-hash case was passing incidentally through the empty Fragment path. I updated both handlers to consume and clear a missing hash intent before any route fallback, and changed the regression to use a non-empty intercepted Fragment. The test now proves that route content rendered while scroll and focus remain unchanged; it passes in both Turbopack and webpack (8/8).

ztanner (approved):
This looks good to me, probably worth getting @eps1lon to take a look as well.

philwolstenholme:
Thanks for looking at this so quickly @DavidIlie and @ztanner , we appreciate it! (I was the Vercel customer that raised it via Slack)

DavidIlie:
that's what we're here for! :D

PR #96352, opened 2026-07-29
Bound Turbopack server HMR to active App routes

marcoshernanz:
## What

Bound Turbopack's global server Fast Refresh working set to App Router pages
that are active under the existing `onDemandEntries` policy.

Today, every App page visited during a dev-server process continues to
participate in `Project::all_hmr_update`. As a session navigates through more
routes, later shared-module edits enumerate and version an ever-growing set of
unrelated entries.

This change:

- sends the existing App Router tree over the dev WebSocket on connection,
  tree changes, and a 2.5-second heartbeat;
- keeps connected, recently active, and buffered pages in the aggregate-HMR
  working set;
- retires older page chunks from the global aggregate;
- keeps their endpoint output graph and change iterator connected so inactive
  source edits are not lost;
- drains inactive endpoint events without refreshing the browser;
- transactionally writes and applies current changed chunks to the existing
  server runtime before the first reentry response;
- advances an explicit native reactivation epoch only after that catch-up write
  and waits for the exact epoch before rendering;
- bounds that catch-up wait and falls back to the existing safe full
  re-evaluation path after one second;
- handles concurrent HTML/RSC requests, syntax-error recovery, route
  deletion/re-addition, failed-catch-up retries, and empty aggregate working
  sets.

This is intentionally an App Router / Node dev-server change. It does not
change production builds, Pages Router, Edge routes, or route handlers.

## Why

The existing global server-HMR subscription accumulates every written App page
for the process lifetime. The source already called out the missing lifecycle.
Large, long-running sessions—especially agent workflows that navigate across
many previews—therefore pay for old routes on every later edit.

The patch separates two requirements:

1. an inactive endpoint's output must stay current; and
2. that endpoint does not need to participate in every global HMR version scan.

Rejected faster prototypes disconnected one of the source-observing roots and
returned stale first responses. The final design leaves those roots live,
removes only aggregate membership, and publishes reentry transactionally.

## Performance

Exact reviewed candidate, Node 24.14.1, 16 vCPU / 32 GB Internal Playground
DevBox, fresh `.next` and process per run, eight edits, alternating A/B order.
The primary estimate is the geometric mean of six within-block ratios; intervals
are deterministic block-bootstrap 95% intervals.

### 100 previously visited routes

| Metric | Baseline | Candidate | Paired effect |
| --- | ---: | ---: | ---: |
| Mean edit completion | 204.96 ms | 169.38 ms | **-17.37%** [-21.09, -14.13] |
| Warm edits 2–8 | 165.97 ms | 135.51 ms | **-18.34%** [-21.76, -15.17] |
| First edit | 477.87 ms | 406.49 ms | **-14.97%** [-20.13, -9.53] |
| Route-warming navigation | 18.95 s | 18.90 s | -0.22% [-3.38, +3.21] |
| RSS after eight edits | 1.78 GiB | 1.91 GiB | **+6.80%** [+5.60, +8.00] |

All six latency pairs favored the candidate and all 12 processes reached every
exact revision. Exact-baseline A/A was +3.18% [-1.23, +7.79].

### Route-count boundary

| Visited pages | Paired mean-edit effect |
| ---: | ---: |
| 1 | +1.01% |
| 10 | -1.64% |
| 25 | -5.27% |
| 50 | **-18.47%** |
| 100 | **-17.37%** (six-block primary panel) |

The benefit is for long sessions around 50+ visited pages, not startup or small
apps.

### Reentry cost

Four alternating blocks, 100 visited routes, four inactive edits per process:

| Metric | Baseline | Candidate |
| --- | ---: | ---: |
| No-edit retired-route request | 51.03 ms | 106.03 ms |
| First request after inactive edit | 42.10 ms | 121.60 ms |
| Following warm request | 31.57 ms | 31.51 ms |

All first responses contained the current revision. The one-time ~80 ms dirty
reentry cost is the transactional catch-up that prevents stale previews.

That powered panel used the canary.102 performance runtime. The final
canary.104 landing build adds the explicit post-write epoch required by review.
Two fresh 100-route correctness smokes measured:

| Final landing smoke | Process 1 | Process 2 |
| --- | ---: | ---: |
| No-edit retired-route request | 294.33 ms | 308.58 ms |
| Mean first request after inactive edit | 108.03 ms | 112.18 ms |
| Mean following warm request | 26.04 ms | 25.89 ms |

All eight dirty reentries returned the exact marker. The two-process smoke is
not a powered estimate, but it makes the final tradeoff explicit: a no-edit
return to a retired route can cost roughly 250 ms more than the original
baseline.

The memory regression is real and is not claimed as an OOM improvement.

### Real-v0 negative control

Private v0 `/projects`, 22 warmed production routes, 12 files written 50 ms
apart, cancellation enabled, six A/B process pairs, 72 measured edits:

| Metric | Baseline | Candidate | Paired effect |
| --- | ---: | ---: | ---: |
| Final visible from first write | 1,432.20 ms | 1,491.60 ms | +3.88% [-1.99, +10.11] |
| RSC requests / server renders | 1.00 / 1.00 | 1.00 / 1.00 | 0.00% / 0.00% |
| Target-route navigation | 16.97 s | 18.13 s | **+6.93%** [+3.63, +10.45] |
| Mean of 22 warm navigations | 11.21 s | 11.60 s | **+3.42%** [+1.59, +5.53] |
| Peak RSS | 14.68 GiB | 15.25 GiB | +3.59% [-7.58, +17.03] |

All 72 revisions were exact; all 12 processes restored their source/config and
had zero non-OK responses or page errors. This result deliberately narrows the
claim: the patch is not a v0 edit-latency improvement at 22 warmed routes. Its
measured benefit starts around the 50–100-route accumulated-session regime.

## Correctness

The new regression uses a one-second inactive age and exercises:

- connected-route retention;
- a shared server edit while the route is retired;
- no-edit reentry without unrelated module re-evaluation;
- a retired two-file server + client-component edit whose single first response
  contains both changes;
- a visible sibling dynamic-import route and later in-place lazy-chunk HMR;
- unrelated server state preservation;
- retired-route syntax error, redbox, first-response recovery after source fix,
  and automatic browser recovery;
- route delete, a fresh first successful re-add response, and later HMR.

The route remains excluded from aggregate HMR while its endpoint is written.
After the write, native code advances a monotonic reactivation epoch and
includes it in both the aggregate version identity and the update resource
header. The request waits for that exact epoch before rendering. Pure epoch
updates survive an empty aggregate, and retries left by a failed catch-up own
the same post-write acknowledgement.

The normal path never globally force-evicts the shared runtime. A one-second
timeout deduplicates through the existing full re-evaluation promise so a route
cannot wedge.

## Validation

- `pnpm test-dev-turbo test/development/app-dir/server-hmr/server-hmr.test.ts`
  with the exact native binary: **17/17 pass** on canary.102; after rebasing
  onto canary.104, the strict-route-types + filesystem-cache run also passed
  **17/17** in 101.57 s
- `pnpm --filter next build`: pass
- `cargo check -p next-api -p next-napi-bindings`: pass
- repository lint-staged hook: pass
- Prettier / ESLint / Rustfmt / `git diff --check`: pass
- Claude Opus 5 xhigh final review: clean on exact `c49038b32b`, no
  accepted/actionable findings, `patch is correct` (0.72)

The bundled Codex reviewer could not authenticate in the clean review
environment, so the final review is explicitly a one-reviewer panel.

Runtime fingerprints:

```text
hot-reloader-turbopack.js
395178d6e5a1eb4f32690e4810796bd61ca46d1382152bec3d4f9609b4c34462

next-swc.linux-x64-gnu.node
e9f634bfd4621b528b7e24b3895a9c7315486f09a5e507849aa693cd0c471c17
```

## Follow-ups / tradeoffs

- Exact-final RSS is 6%–8% higher in the 100-route fixture.
- The route-tree heartbeat is a small message every 2.5 seconds.
- Reentry moves work from every unrelated edit to the first later use.
- Route activity/inactive-path entries that never regain an emitted chunk can
  remain for the dev-server lifetime.
- This does not make multi-file filesystem edits atomic; it composes with a
  future acknowledged edit transaction.
- The real-v0 22-route control did not reproduce the synthetic edit-latency
  win and made route warming 3%–7% slower.

### DIFF
--- a/crates/next-api/src/aggregate_hmr.rs
+++ b/crates/next-api/src/aggregate_hmr.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use anyhow::Result;
-use rustc_hash::FxHashMap;
+use rustc_hash::{FxHashMap, FxHashSet};
 use turbo_rcstr::RcStr;
 use turbo_tasks::{FxIndexMap, FxIndexSet, ReadRef, ResolvedVc, TraitRef, TryJoinIterExt, Vc};
 use turbo_tasks_fs::FileSystemPath;
@@ -70,12 +70,54 @@ impl AggregateHmrVersion {
     ) -> Result<Vc<Box<dyn Version>>> {
         // An empty `versions` map behaves the same as `NotFoundVersion` would in
         // `diff_chunks_against`, so no special case is needed here.
-        let chunks = map.hmr_chunks_in_path(root).await?;
-        Ok(Vc::upcast(Self::from_chunks(&chunks).await?))
+        let snapshot = map.hmr_chunks_in_path(root).await?;
+        Ok(Vc::upcast(Self::from_chunks(&snapshot.active).await?))
     }
 
     pub async fn from_chunks(chunks: &[HmrChunkWithContent]) -> Result<Vc<Self>> {
-        let versions = chunks
+        let versions = Self::versions_from_chunks(chunks)
+            .await?
+            .into_iter()
+            .collect();
+        Ok(Self { versions }.cell())
+    }
+
+    /// Builds the next aggregate version while retaining the last version of
+    /// chunks outside the current working set. Inactive entries therefore do
+    /// not need to be evaluated on every update, but can still be diffed against
+    /// their last active version when they re-enter the working set.
+    pub async fn from_chunks_with_previous(
+        chunks: &[HmrChunkWithContent],
+        inactive_paths: &FxHashSet<RcStr>,
+        previous: Vc<VersionState>,
+    ) -> Result<Vc<Self>> {
+        let mut versions = Self::versions_from_chunks(chunks).await?;
+
+        let previous = previous.get().to_resolved().await?;
+        if let Some(previous) = ResolvedVc::try_downcast_type::<AggregateHmrVersion>(previous) {
+            versions.extend(
+                previous
+                    .await?
+                    .versions
+                    .iter()
+                    .filter(|(path, _)| inactive_paths.contains(*path))
+                    .map(|(path, version)| (path.clone(), version.clone())),
+            );
+        }
+
+        // Aggregate version IDs depend on iteration order. Keep the union sorted
+        // so merely activating or deactivating an unchanged entry is a no-op.
+        versions.sort_by(|(a, _), (b, _)| a.cmp(b));
+        Ok(Self {
+            versions: versions.into_iter().collect(),
+        }
+        .cell())
+    }
+
+    async fn versions_from_chunks(
+        chunks: &[HmrChunkWithContent],
+    ) -> Result<Vec<(RcStr, TraitRef<Box<dyn Version>>)>> {
+        let mut versions = chunks
             .iter()
             .map(|HmrChunkWithContent { path, content }| {
                 let path = path.clone();
@@ -86,10 +128,9 @@ impl AggregateHmrVersion {
                 }
             })
             .try_join()
-            .await?
-            .into_iter()
-            .collect();
-        Ok(Self { versions }.cell())
+            .await?;
+        versions.sort_by(|(a, _), (b, _)| a.cmp(b));
+        Ok(versions)
     }
 }
 
--- a/crates/next-api/src/aggregate_hmr.rs
+++ b/crates/next-api/src/aggregate_hmr.rs
@@ -33,6 +33,8 @@ pub fn is_entry_chunk_list_content(content: ResolvedVc<Box<dyn VersionedContent>
 pub struct AggregateHmrVersion {
     #[turbo_tasks(trace_ignore)]
     pub versions: FxIndexMap<RcStr, TraitRef<Box<dyn Version>>>,
+    /// Monotonic token identifying the active-route aggregate transition.
+    pub reactivation_epoch: u32,
 }
 
 #[turbo_tasks::value_impl]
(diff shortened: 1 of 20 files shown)


marcoshernanz:
Superseded by #96433. Exact private-v0 benchmarking showed this debounce-oriented direction did not materially improve the real multi-file agent-edit hot path. The replacement adds acknowledged semantic edit transactions: it cut ordinary real-v0 recompilation/RSC work by about 49%, and collapsed the long staggered 24-file case from 12 requests to 1. Closing this draft so there is one active proposal.

marcoshernanz:
Closing this explored alternative so the investigation leaves one draft PR.

The exact synthetic working-set panel showed a real long-session effect at 50–100 accumulated App routes (17.37% faster edit completion at 100 routes), but the benefit was neutral below 25 routes. The private-v0 `/projects` control at 22 warmed routes did not reproduce an edit win, route warming was 3.42% slower, exact synthetic RSS increased 6%–8%, and the final epoch-hardened no-edit reentry smoke added roughly 250 ms versus the original baseline.

Those tradeoffs make this less meaningful than #96433, which directly addresses the agent multi-file complaint on real v0: ordinary 12-file edits cut requests/renders about 49% and sampled CPU 35.9%, while the complaint-scale 24-file case went from 12 RSC waves to exactly one and improved final-write-to-preview latency 40.9%. #96433 is now rebased onto current canary, mergeable, signed, and fully revalidated.

The full #96352 description and benchmark artifacts remain preserved as useful negative/working-set evidence; this close is a product-significance decision, not a claim that the synthetic result was false.

PR #95546, opened 2026-07-07
Turbopack server hmr: avoid complete `clear` on graph changes

wbinnssmith:
Currently, importing a new module causes a complete eviction (`clear()`) and re-evaluation of server chunks, both in the Turbopack runtime's module cache and Node's `require.cache`.

Here's what happens today:

- A new module is imported into a chunk's graph
- This changes its `availability_info`, which in dev for non-entry chunks is encoded into the chunk's filepath
- `VersionedContentMap` works entirely on chunk paths. When we construct instructions to transition a chunk into its new state, we fail to find the prior state and fall back to `clear()` as described above.

An ideal version of this is a refactor that implements `VersionedContentMap` on a per-module basis, not a per-chunk one. This PR doesn't do that, but achieves consistent module-level updates when a chunk's availability info changes.

It does this by implementing chunk lists for server entry chunks the same way the client hmr implementation does: an entry chunk's version aggregates the versions of its dependent chunks (including dynamically imported ones), keyed by merger rather than path. Entry chunks do not encode availability info in their paths, so they are insulated from missing versions. Updates are applied in Node.js through the same shared merged-update machinery in the unified hmr runtime that the client already uses, plus a new `ChunkListUpdate` branch in the server hmr client to unwrap the merged updates.

Details:

- The merged-update wire format (`EcmascriptMergedUpdate` etc.) moves from turbopack-browser into shared `turbopack_ecmascript::chunk_list`, along with runtime-agnostic `ChunkListVersion` and `update_chunk_list` implementations.
- `turbopack-nodejs` gains a chunk content merger mirroring the browser one. `EcmascriptBuildNodeEntryChunk`'s versioned content is now a chunk-list content over its sync and async chunks. It is version-only and never emitted; the entry chunk still inlines its loader calls.
- The aggregate server hmr subscription now tracks only entry chunks. Shared chunks under `server/chunks/` ride the entry's `ChunkListUpdate` as module deltas, so their content-hash paths no longer matter.
- The Node hmr client applies `ChunkListUpdate` by feeding each nested merged update through the shared apply path. Modules that appear in an "added" chunk but already exist in the module cache were moved by a chunk rename and are treated as modified.
- On a successful partial apply, the hot reloader clears the manifest cache for updated chunks and notifies browsers to refetch RSC, without clearing `require.cache`.

### Test Plan

Adds a series of e2e tests to catch the described manual test plan in both entry and dynamic chunks.

PR #96357, opened 2026-07-29
Deliver client HMR events to a callback instead of an async iterator

gaearon:
The napi binding delivers HMR events to a callback. `project.hmrEvents()` wrapped that callback in an async generator — a buffer, a parked promise, a cancel sentinel — so the consumer could drive it with `for await`.

An async iterator is an awkward fit here. Every update pays for a promise and a generator step, every idle subscription holds a promise parked on the next event (one per connected browser per entrypoint), and unsubscribing isn't an operation you can call — it means breaking out of a loop elsewhere. Handing the callback through is cleaner and slightly cheaper:

```ts
const unsubscribe = project.hmrEvents(id, HmrTarget.Client, (err, update) => …)
```

`projectHmrEvents` returns the task handle synchronously, so one boolean makes unsubscribing idempotent and drops an event that was already queued at dispose. The target picks the update type, replacing two overloads and a cast at the call site.

It also fixes a latent one-way unsubscribe: `unsubscribeFromClientHmrEvents` left the entry in `state.subscriptions`, which is the map the subscribe path checks, so subscribing again to the same chunk on one connection did nothing. No app produces that sequence today.

Server side is unaffected — #95546 left `hmrEvents`' `HmrTarget.Server` overload with no callers. Supersedes #91704 and the client half of #96182.

<details>
<summary>Verification</summary>

`test/development/basic/next-rs-api.test.ts`, `test/development/basic/hmr/*` and `test/development/app-hmr/*` match the base commit. A new RS API test unsubscribes from inside the first update, while that rebuild's events are still queued for sibling subscriptions: 37 events with the guard, 78 without it.

Both consumers were also run against a simulated napi layer (synchronous task handle, events a tick late, an event queued before `rootTaskDispose` still landing) over 412 randomized subscribe / unsubscribe / update / error / close programs. Differences: the map entry above, plus the new consumer clearing a chunk's issue key on the error path just before it closes the client.

</details>

### DIFF
--- a/packages/next/src/build/swc/index.ts
+++ b/packages/next/src/build/swc/index.ts
@@ -27,6 +27,7 @@ import type {
   DefineEnv,
   Endpoint,
   HmrChunkNames,
+  HmrEvent,
   Lockfile,
   NodeJsHmrUpdate,
   PartialProjectOptions,
@@ -37,7 +38,6 @@ import type {
   TurboEngineOptions,
   TurbopackResult,
   TurbopackStackFrame,
-  Update,
   UpdateMessage,
   WrittenEndpoint,
 } from './types'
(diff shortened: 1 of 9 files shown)


gaearon:
meh, this doesn't seem important

PR #96359, opened 2026-07-29
Add automated code review workflow

marcoshernanz:
## Summary

Adds one protected `pull_request_target` workflow for advisory automated review of non-draft, member-authored PRs.

The existing Vercel reviewer runs Codex, Claude, and synthesis with read-only GitHub access. This base-owned workflow treats that result as untrusted, validates it, and owns the only GitHub mutation.

## Write boundary

- no checkout, PR code execution, dependency installation, Actions secrets, or shell interpolation
- globally empty permissions; the job has only OIDC and pull-request read
- exact repository, event, author association, PR state, base SHA, and head SHA validation
- final read-only PR revalidation immediately before publishing
- HTML comments stripped, mentions neutralized, token patterns redacted, and output capped at 48 KB
- a write token is minted only after validation
- exactly one top-level comment `POST` or `PATCH`
- existing comments are eligible only when attributed to one of gh-sts's immutable `general` GitHub App IDs and begin with the fixed marker
- both actions are pinned to immutable commit SHAs

The workflow intentionally runs once on `opened` or `ready_for_review`; it does not run on `synchronize`.

## Validation

- `actionlint`
- Prettier
- syntax validation of all three embedded `github-script` programs
- upstream commit verification for both pinned actions
- independent Codex and Claude autoreview

The workflow remains inert until the matching gh-sts policy is approved.

### DIFF
--- a/.github/workflows/automated_code_review.yml
+++ b/.github/workflows/automated_code_review.yml
@@ -1,10 +1,9 @@
 name: Automated Code Review
 
-# This privileged workflow runs from the protected canary branch. It never
-# checks out or executes pull-request code, installs dependencies, reads
-# Actions secrets, or interpolates pull-request text into a shell. Vercel only
-# computes an untrusted, bounded review; this workflow revalidates the PR and
-# owns the sole GitHub comment mutation.
+# This privileged workflow runs only from protected canary. It checks out the
+# base commit that triggered the workflow, never pull-request code. Vercel
+# computes an untrusted review; this workflow validates it and owns the sole
+# GitHub mutation.
 on:
   pull_request_target:
     types: [opened, ready_for_review]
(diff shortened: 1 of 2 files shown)


.github/workflows/automated_code_review.yml
```diff
@@ -0,0 +1,318 @@
...
+              await wait(30_000);
+            }
+            throw new Error("Automated review timed out after 50 minutes.");
+
+      - name: Validate review and select comment
+        id: validate
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const fs = require("node:fs/promises");
```
eps1lon: Can we put this into a JS file? Makes it easier to manually run and review

marcoshernanz: Moved both action scripts into `.github/scripts/automated_code_review.js`. The workflow now checks out only the trusted base commit and calls the exported functions, so the JS can be reviewed and run independently.

eps1lon (approved):

PR #96366, opened 2026-07-29
test: expand Cache Components metadata streaming coverage

ztanner:
## Summary

Expand Cache Components metadata streaming coverage across the default, custom, and wildcard `htmlLimitedBots` configurations.

The active matrix now covers regular user agents, configured HTML-limited bots, built-in bot classifications, metadata placement for static and dynamic routes, PPR shell behavior, streaming versus buffered responses, and hydration after revalidation. It replaces the skipped legacy PPR metadata suite with active Cache Components coverage.

PR #75366, opened 2025-01-27
[metadata] new metadata insertion API and support PPR

huozhi:
### What

- We're introducing a new non-public API hook `useServerInsertedMetadata` internally for only inserting the metadata to rendered page.
- Add PPR test cases

### Why

We currently use the public API `useServerInsertedHTML` to insert streamed metadata, however that API has few flaws with PPR and contains too much logic for handling CSS-in-JS inserted html. For PPR it's not easy to get it work with it since there're server situatios in PPR such as different phases like "prerender" and "resume", then the different modes such as "static" and "partial static rendering". Not all of these cases we need to insert the metadata in the `<head>` which `useServerInsertedHTML` can do and has a lot of rendering tricks for it. After discussing with @gnoff that we'll leverage this new API to insert the streamed metadata, and then append them in the stream.

Another discussion about a case we're concerning it could happen but seems not possible, which is the metadata is inserted in **prerender** and get re-inserted during **resume**. In reality the metadata is next to the page element and react will still render the sibling metadata while the page itself is suspended. So that we always see it's included in prerender when metadata is static. We came up with a solution of saving "wether it's inserted in prerender" into postpone state, but it seems not necessary now. We have test cases covered. We'll gonna land this and see what's going on later. If there's any missing case of double inserted metadata, then we can revisit the postpone solution.

Closes NDX-705

PR #96367, opened 2026-07-29
fix: respect `htmlLimitedBots` in cache components without buffering the full response

ztanner:
## Summary

DOM-capable crawlers can consume streamed metadata, but with cacheComponents enabled, it treated them like HTML-limited bots and forced a full dynamic render. This meant capable crawlers could not use the prerendered shell even when they were not matched by `htmlLimitedBots`.

This change lets DOM-capable crawlers follow the normal PPR shell and streaming-metadata path. `htmlLimitedBots` remains the source of truth for requests that need blocking metadata, including when a custom pattern explicitly matches a DOM-capable crawler.

## Test Plan
Tests were added that contained a Suspense boundary that intentionally never resolves. The test reads the response stream until it receives the fallback, then verifies that blocking metadata has already been emitted while the unresolved page content has not. A fully buffered response would never return, so this avoids relying on timing or arbitrary delays.

This PR is stacked on #96366.

### DIFF
--- a/docs/01-app/02-guides/streaming.mdx
+++ b/docs/01-app/02-guides/streaming.mdx
@@ -668,15 +668,15 @@ export default async function PostPage({ params }) {
 
 ### Bots and crawlers
 
-Bots and crawlers consume a complete, fully formed HTML document rather than rendering chunks as they arrive. Next.js detects them by their user agent and waits for the render to finish, then sends the whole document in a single response instead of streaming it progressively. [`generateMetadata`](/docs/app/api-reference/functions/generate-metadata) resolves before the response is sent for bots that only scrape static HTML (such as Twitterbot or Slackbot), so metadata is present in the `<head>` of that HTML. Full browsers and capable crawlers can instead receive [streaming metadata](/docs/app/api-reference/functions/generate-metadata#streaming-metadata) alongside the page content.
+HTML-limited bots and crawlers need metadata to be available in the `<head>` of the initial HTML. Next.js detects them by their user agent and waits for [`generateMetadata`](/docs/app/api-reference/functions/generate-metadata) to resolve before streaming the page content. Full browsers and DOM-capable crawlers can instead receive [streaming metadata](/docs/app/api-reference/functions/generate-metadata#streaming-metadata) alongside the page content.
 
 You can customize which bots receive blocking metadata with the [`htmlLimitedBots`](/docs/app/api-reference/config/next-config-js/htmlLimitedBots) configuration option. See the [`loading.js` SEO section](/docs/app/api-reference/file-conventions/loading#seo) for more details.
 
 #### Cache Components
 
-With [Cache Components](/docs/app/getting-started/caching), a visitor receives the prerendered shell immediately and dynamic content streams in as it resolves. A bot is served differently: because it needs a complete document, Next.js skips the prerendered shell and renders the entire page dynamically at request time, waiting for the full render to complete before sending the finished HTML.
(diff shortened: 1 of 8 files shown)


packages/next/src/build/templates/app-page-runtime.ts
```diff
@@ -539,12 +535,14 @@ export function createAppPageEntrypoint({
...
-          ? true
-          : shouldServeStreamingMetadata(userAgent, nextConfig.htmlLimitedBots)
+    const serveStreamingMetadata = !userAgent
+      ? true
+      : shouldServeStreamingMetadata(userAgent, nextConfig.htmlLimitedBots)
+
+    // A custom htmlLimitedBots pattern replaces the default HTML-limited bot
+    // pattern. Preserve the built-in bot behavior only when that bot is still
+    // included in the effective pattern.
+    const isHtmlBot = botType === 'html' && !serveStreamingMetadata
```
gnoff: I don't think i understand the taxonomy. what is isHtmlBot that isn't botType html? I get that this excludes serveStreamingMetdata cases but why does that define it as an htmlBot? shouldn't htmlLimited not get streaming metadata anyway?

packages/next/src/build/templates/app-page-runtime.ts
```diff
@@ -588,10 +586,10 @@ export function createAppPageEntrypoint({
...
           isDynamicRSCRequest)
 
-    // The fixed bot list historically waits for the entire PPR render. A bot
-    // configured only through htmlLimitedBots may stream after its metadata
-    // resolves.
-    const shouldWaitOnAllReady = Boolean(botType) && isRoutePPREnabled
+    // HTML-limited bots from the built-in classification historically wait
+    // for the entire PPR render. A bot configured only through htmlLimitedBots
+    // may stream after its metadata resolves.
+    const shouldWaitOnAllReady = isRoutePPREnabled && isHtmlBot
```
gnoff: the old logic seems wrong but the new logic is odd too, there used to be historical pathways where we would block the stream on allReady but to my knowledge no PPR enabled route needed special additional blocking. Is this because the set of historical builtin html limited bots got fully blocking renders and we are trying to preserve that? I think we need to go back to that original justifcation b/c there really should be no bots afaik that get fully blocking renders.

If we do want to support this maybe it becomes it's own config and the default set is empty.

Additionally, React will outline Suspense boundaries so even in a fully waited on stream that is not a guarantee that there will be no inline scripts emitted. If we want to make the document complete without executing any javascript we need to also set the SSR chunk size to infinity.

We don't necessarily need to block this PR on resolving these things but this still feels wrong to me

ztanner: Yeah, I traced this back to #75366. The full wait was added to make sure metadata was available in the head, but seems that we don’t need to block the rest of the render for that. I've added tests that show the metadata blocking while the body continues streaming.

I'll remove the bot-specific `allReady` wait

packages/next/src/build/templates/app-page-runtime.ts
ztanner: `isHtmlBot` wasn’t meant to be a separate taxonomy. It was trying to represent `botType === 'html'` plus whether the user agent still matched the effective `htmlLimitedBots` pattern, since a custom pattern can replace the default.

Removed it entirely so now `htmlLimitedBots` directly controls whether metadata streams, while `botType` remains the bot classification.

docs/01-app/02-guides/streaming.mdx
eps1lon, line 1: Worth pointing out that none of these consider whether the bot supports JS or not. There's a common misconception where bot = assume no JS execution but that's never something Next.js/React guaranteed. Even with waiting on `allReady` to flush, we could've streamed in large boundaries which would need JS.

iirc we do have a regex for bots that don't support JS but it's not followed-through entirely (bot type "html" or something?).
For UAs without JS we need
1. wait to flush until `onAllReady`
2. Set `progressiveChunkSize` to +infinity (ensures large boundaries are not outlined)
3. (optional) don't emit Flight payload in `<script>` tags

gnoff (approved):

PR #96372, opened 2026-07-30
fix: Route group layout leaks into root not-found page

### LINKED ISSUE #96361: Route group layout leaks into root `not-found` page

https://github.com/lsvs/route-group-not-found-bug-reproduction

### To Reproduce

1. Start the app with `pnpm dev` or `pnpm build && pnpm start`.
2. Visit `/route-segment/error` — the `[slug]` page calls `notFound()`
3. Observe that the `not-found` page is rendered inside the `(route-group)` layout.

Current: The `not-found` page is rendered inside the parent `(route-group)` layout.

Expected: Since there is no `not-found` page within that route group, Next.js falls back to the root `not-found` page, which should render only within the root layout.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Tue Apr 21 20:13:48 PDT 2026; root:xnu-11417.140.69.710.16~1/RELEASE_ARM64_T8112
  Available memory (MB): 24576
  Available CPU cores: 8
Binaries:
  Node: 24.18.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: 11.17.0
Relevant Packages:
  next: 16.3.0-preview.10 // Latest available version is detected (16.3.0-preview.10).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Not Found

### Additional context

The layout hierarchy appears to be inconsistent:

- `/` renders only the root layout.
- `/route-segment` renders all three layouts.
- `/route-segment/error` (which calls `notFound()`) renders the root layout and the `(route-group)` layout, but not the nested route layout. ← the bug
- `/route-segment/error/error` renders only the root layout again.

This suggests the `(route-group)` layout leaks into the `not-found` boundary when `notFound()` is called.

icyJoseph:
Fixes: https://github.com/vercel/next.js/issues/96361

Turbopack copied the root's `not-found.js` / `forbidden.js` / `unauthorized.js` into first-layer group routes, so an interrupt thrown below `app/(group)/` rendered the root boundary nested inside the group layout instead of bubbling up to the root layout. The parent module is now only propagated when it is the built-in default, matching `next-app-loader`. Webpack was already correct.

### DIFF
--- a/crates/next-core/src/app_structure.rs
+++ b/crates/next-core/src/app_structure.rs
@@ -1088,8 +1088,8 @@ async fn directory_tree_to_loader_tree(
 ///   set.
 /// * `file_path` - The file path to the default page if neither the current module nor the parent
 ///   module is set.
-/// * `is_first_layer_group_route` - If true, the module will be overridden with the parent module
-///   if it is not set.
+/// * `is_first_layer_group_route` - If true and the module is not set, the module will be
+///   overridden with the parent module, but only when the parent module is the built-in default.
 async fn check_and_update_module_references(
     app_dir: FileSystemPath,
     module: &mut Option<FileSystemPath>,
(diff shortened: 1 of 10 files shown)


lsvs:
@icyJoseph Thanks for working on that! Just checking — should the next step be assigning a reviewer, or is there anything else needed to proceed?

PR #96383, opened 2026-07-30
[turbopack] Respect `sideEffects` in a `package.json` file with `optimizePackageImports`

### LINKED ISSUE #96333: Turbopack + `optimizePackageImports` tree-shakes a `sideEffects`-allowlisted, side-effect-only module (breaks `react-focus-lock`)

https://github.com/cultureamp/turbopack-sidecar-treeshaking-repro

### To Reproduce

1. `npm install`
2. `npm run build` – ❌ Turbopack (Next's default bundler) — build FAILS at prerender: `"Error: Sidecar medium not found"`
3. `npm run build:webpack` – ✅ webpack — same config, builds fine

With `experimental.optimizePackageImports` containing **both** `react-focus-on` and `react-focus-lock`, a production **Turbopack** build tree-shakes react-focus-lock's side-effect-only `sidecar` module out of the bundle. That module's sole purpose is a registration side-effect (`exportSidecar(mediumSidecar, FocusTrap)`); with it gone, rendering any focus trap throws:

```
Error: Sidecar medium not found
```

react-focus-lock's `package.json` marks the module side-effectful via a `sideEffects` allowlist (`["**/sidecar.js", "**/index.js"]`), so it should be retained. **webpack retains it; Turbopack drops it** when the import is rewritten by `optimizePackageImports`.

### Expected

`optimizePackageImports` should not cause a module in the target package's `sideEffects` allowlist to be dropped. Turbopack should honour `sideEffects` for side-effect-only (bare) imports even when the barrel import is rewritten by the optimization.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 26.5.0
  npm: 11.17.0
  Yarn: N/A
  pnpm: 10.34.5
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: N/A
Next.js Config:
  output: N/A


Note: Above-linked repro repo is on Next 16.2.11, but I've verified it still reproduces with the current canary.
```

Turbopack

next build (local)

### Additional context

- Reproduces only in a **production Turbopack build**; `next dev`, `next build --webpack`, unit tests, and Storybook all pass — so it typically surfaces only in deployed apps.
- Impact is broad: `optimizePackageImports` is widely used (Next enables it by default for many packages; design systems add their own), and most modal/popover/menu components transitively use `react-focus-on` / `react-focus-lock`.
- `react-remove-scroll` uses the same sidecar pattern and is likely affected the same way.

sampoder:
Closes https://github.com/vercel/next.js/issues/96333 by re-ordering the statements in `get_side_effect_free_declaration`. The rest of this PR is then a test + adding to a test fixture. This seems like a reasonable change to me as we'd match Webpack's behaviour.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
@@ -228,12 +228,6 @@ pub async fn get_side_effect_free_declaration(
     path: FileSystemPath,
     side_effect_free_packages: Option<Vc<Glob>>,
 ) -> Result<Vc<SideEffectsDeclaration>> {
-    if let Some(side_effect_free_packages) = side_effect_free_packages
-        && side_effect_free_packages.await?.matches(&path.path)
-    {
-        return Ok(SideEffectsDeclaration::SideEffectFree.cell());
-    }
-
     let find_package_json = find_context_file(path.parent(), package_json(), false).await?;
 
     if let FindContextFileResult::Found(package_json, _) = &*find_package_json {
@@ -259,6 +253,12 @@ pub async fn get_side_effect_free_declaration(
                 }
             }
         }
+    }
+
+    if let Some(side_effect_free_packages) = side_effect_free_packages
+        && side_effect_free_packages.await?.matches(&path.path)
+    {
+        return Ok(SideEffectsDeclaration::SideEffectFree.cell());
     }
 
     Ok(SideEffectsDeclaration::None.cell())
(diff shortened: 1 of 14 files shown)


turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -228,12 +228,6 @@ pub async fn get_side_effect_free_declaration(
     path: FileSystemPath,
     side_effect_free_packages: Option<Vc<Glob>>,
 ) -> Result<Vc<SideEffectsDeclaration>> {
-    if let Some(side_effect_free_packages) = side_effect_free_packages
-        && side_effect_free_packages.await?.matches(&path.path)
-    {
-        return Ok(SideEffectsDeclaration::SideEffectFree.cell());
-    }
-
     let find_package_json = find_context_file(path.parent(), package_json(), false).await?;
```
lukesandberg:
```suggestion
// Always respect the package.json over the global side_effect_free_packages by checking it first
// See #96333
```

lukesandberg (approved):

PR #96005, opened 2026-07-21
Per-route prerender-manifest.json

mischnic:
- have per route manifests containing only the needed information for rendering (i.e. `.next/server/preview-props.json` plus `.next/server/app/foo/page/prerender-manifest.json`)
   - this also provides more isolation (if function /foo has a change in the manifest, function /bar doesn't necessarily)
- the global legacy `.next/prerender-manifest.json` is only for the builder (i.e. CDN configuration, etc)

<br>

- [x] list manifests in NFT file
- [ ] cleanup

PR #96384, opened 2026-07-30
fix: serialize prerender manifest updates in dev to prevent corruption

### LINKED ISSUE #96259: Dev server 500s on every request: concurrent getStaticPaths corrupts prerender-manifest.json

https://github.com/LouisTsang-jk/nextjs-prerender-manifest-race

### To Reproduce

```
1. git clone https://github.com/LouisTsang-jk/nextjs-prerender-manifest-race
2. npm install
3. ./repro.sh
```

`repro.sh` starts `next dev`, requests 10 pages under `app/[locale]/` concurrently, then checks `.next/dev/prerender-manifest.json` and issues one more request.

The app is minimal: a root `app/[locale]/layout.tsx` with `generateStaticParams` returning 9 locales, 10 trivial pages under it, and one route handler.

Since the failure is a race, a run can come back clean — rerunning hits it within the first couple of attempts on an M-series Mac. `race-unit.js` in the same repo reproduces the underlying read-modify-write deterministically (consistently ~180 corrupt manifests out of 300 rounds) with no Next.js involved.

**Current:** once two pages under the same dynamic segment resolve their static paths at the same time, `.next/dev/prerender-manifest.json` stops being valid JSON and **every subsequent request 500s**, including requests to completely unrelated routes:

```
⨯ SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>) {
  page: '/en/page-9'
}
 GET /en/page-9 500 in 4.9s
```

Two distinct corruption shapes show up, depending on how the writes interleave:

- `SyntaxError: Unexpected end of JSON input` — a read observes a file that is still being written.
- `SyntaxError: Unexpected non-whitespace character after JSON at position N` — valid JSON followed by leftover bytes.

From a real i18n app (9 locales, ~25 pages under `app/[locale]/`), the on-disk manifest after a collision was 1264 bytes with valid JSON only up to 1192:

```
..."previewModeEncryptionKey":"<64-hex value X>"}}ey":"<the same value X>"}}
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
                              72 bytes left over from a previous, longer write
```

The trailing fragment repeats the *same* `previewModeEncryptionKey` as the valid portion, so both writes came from a single dev server process within one session — not stale cache, not two processes.

**Expected:** the manifest stays valid JSON regardless of how many pages resolve their static paths concurrently.

### Root cause

`packages/next/src/server/dev/next-dev-server.ts` (`getStaticPaths`) performs an unsynchronized read-modify-write on that file:

```ts
const rawExistingManifest = await fs.promises.readFile(
  pathJoin(this.distDir, PRERENDER_MANIFEST),
  'utf8'
)
const existingManifest: PrerenderManifest = JSON.parse(rawExistingManifest)
// ...mutate routes / dynamicRoutes...
const updatedManifest = JSON.stringify(existingManifest)

if (updatedManifest !== rawExistingManifest) {
  await fs.promises.writeFile(
    pathJoin(this.distDir, PRERENDER_MANIFEST),
    updatedManifest
  )
}
```

There is no lock, and the write is not atomic. Both `writeFile` calls open with `'w'` and write from offset 0, so when the shorter payload finishes last it only overwrites a prefix of the longer one and leaves the tail in place.

The blast radius comes from `base-server.ts` calling `getPrerenderManifest()` on every request (for `previewProps` and the `isSSG` check) — so one bad file takes down the whole dev server and the error surfaces on routes that have nothing to do with the pages that raced. In our case it consistently appeared on `/api/auth/session`, which sent us looking in entirely the wrong place.

> **Correction** (thanks to [@zoujimmy82-boop](https://github.com/vercel/next.js/issues/96259#issuecomment-5128644737); verified on `canary`): the attribution above is wrong, though the user-visible effect is the same. `next-server.ts` memoizes into `_cachedPreviewManifest` and the dev server never overrides it, so it normally reads from disk once per process — but the assignment happens *after* `loadManifest()` returns, so while the file is corrupt the cache never populates and every request re-reads the bad file and re-throws. That is what makes it "every subsequent request", and why a later, longer write silently heals it. Of the two call sites cited, `previewProps` is in the constructor and the `isSSG` one goes through the memo. The path that keeps re-reading the file in dev is `route-modules/route-module.ts`, via `shouldCache: !this.isDev`.

A `generateStaticParams` on a root layout (the standard i18n setup) amplifies this: every page under `[locale]` runs its own read-modify-write cycle on first compile, so collision probability grows with page count. We only started seeing it regularly after crossing roughly 20 pages under `app/[locale]/`.

This is also why the failure looks intermittent and why `rm -rf .next` "sometimes helps": a later write whose payload is at least as long as the corrupt file silently repairs it, until the next collision.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 27.0.0: Mon Jun 29 21:25:16 PDT 2026; root:xnu-13432.0.50.501.3~1/RELEASE_ARM64_T8112
  Available memory (MB): 24576
  Available CPU cores: 8
Binaries:
  Node: 22.21.1
  npm: 10.9.4
  Yarn: N/A
  pnpm: 10.15.0
Relevant Packages:
  next: 16.3.0-canary.97 // Latest available version is detected (16.3.0-canary.97).
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 6.0.3
Next.js Config:
  output: N/A
```

Dynamic Routes, Internationalization (i18n)

next dev (local)

### Additional context

Also reproduces on `16.2.7`. Not bundler specific — Turbopack and `--webpack` both hit it, since the code path is in the dev server rather than the bundler.

Workaround that does not require patching `node_modules` (included in the repro repo as `workaround-shim.cjs`) — makes writes to that one file atomic:

```jsonc
"dev": "NODE_OPTIONS='--require ./workaround-shim.cjs' next dev"
```

LouisTsang-jk:
### What

`getStaticPaths` in the dev server performs an unsynchronized read-modify-write on `.next/dev/prerender-manifest.json`:

```ts
const rawExistingManifest = await fs.promises.readFile(manifestPath, 'utf8')
const existingManifest: PrerenderManifest = JSON.parse(rawExistingManifest)
// ...mutate routes / dynamicRoutes...
if (updatedManifest !== rawExistingManifest) {
  await fs.promises.writeFile(manifestPath, updatedManifest)
}
```

When two pages under the same dynamic segment resolve their static paths at the same time, the cycles interleave. Both `writeFile` calls open with `'w'` and write from offset 0, so a shorter payload finishing last only overwrites a prefix of a longer one and leaves its tail behind — the file becomes valid JSON followed by garbage. A reader landing mid-write sees a truncated document instead. Interleaving also loses updates, since the later cycle read the file before the earlier one wrote it.

A `generateStaticParams` on a root layout (the standard i18n `app/[locale]` setup) amplifies it: every page under the segment runs its own cycle on first compile, so collision probability grows with page count.

**Why one corrupt file takes down the whole dev server.** The two read paths behave differently, and the correction here is [@zoujimmy82-boop's](https://github.com/vercel/next.js/issues/96259#issuecomment-5128644737) — the original issue text attributed this to `base-server.ts` reading the manifest per request, which is not accurate:

- `next-server.ts` memoizes into `_cachedPreviewManifest`, and the dev server does not override it — so normally this reads from disk once per process. But the assignment happens *after* `loadManifest` returns, so while the file is corrupt the cache never gets populated and every request re-reads the bad file and re-throws. That is why the symptom is "every subsequent request 500s" rather than one bad request, and why a later, longer write silently heals it.
- `route-modules/route-module.ts` loads the same manifest with `shouldCache: !this.isDev`, so in dev it is re-read and re-parsed on route module loads rather than cached. That is what keeps the read window wide enough to collide with the unsynchronized write in the first place.

Either way the `SyntaxError` surfaces on routes unrelated to the pages that raced, which makes it very hard to trace back.

### How

- Add `PrerenderManifestUpdater`, which routes the read-modify-write through a promise queue so cycles cannot interleave. A failed cycle is surfaced to the caller but does not poison the queue.
- Write via a temporary file and `rename`, so a concurrent reader can never observe a partially written manifest.

The queue also removes lost updates, which the atomic write alone would not fix.

### Tests

`prerender-manifest-updater.test.ts` covers this without depending on `next dev` compile timing, so it is not flaky:

- **`runs concurrent cycles one at a time, in call order`** — 25 concurrent cycles each record how many routes they observed. Serialized, that sequence is exactly `[0, 1, ..., 24]`, and all 25 routes survive. Against the current implementation this fails deterministically: every cycle observes `0` and only one route survives.
- **`never exposes a partially written manifest to a reader`** — a reader parses the file in a loop while ten large cycles run. Against the current implementation this reproduces the exact reported failure, `SyntaxError: Unexpected end of JSON input`. It also asserts the observed route count never goes backwards, which is what a lost update looks like from the outside.
- Three smaller cases: a failed cycle rejects without poisoning the queue, an unchanged manifest is not rewritten, and the temporary file is cleaned up when the write fails.

### Reproduction

https://github.com/LouisTsang-jk/nextjs-prerender-manifest-race

`./repro.sh` starts the dev server and requests 10 `[locale]` pages concurrently:

```
==> .next/dev/prerender-manifest.json
    CORRUPT: Unexpected end of JSON input
==> GET /api/ping
    HTTP 500
⨯ SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>) { page: '/en/page-9' }
```

Independently confirmed on `16.2.10` in [this comment](https://github.com/vercel/next.js/issues/96259#issuecomment-5128644737), where it presented as a Playwright suite failing roughly 1 run in 3 on a test unrelated to the pages that raced.

### Notes

This replaces #96260, which I closed myself shortly after opening because the branch was based wrong — the diff showed 30k files. The change itself was never reviewed. This one is based on `canary` and touches three files.

Fixes #96259

### DIFF
--- a/packages/next/src/server/dev/next-dev-server.ts
+++ b/packages/next/src/server/dev/next-dev-server.ts
@@ -27,6 +27,7 @@ import fs from 'fs'
 import { Worker } from 'next/dist/compiled/jest-worker'
 import { installUseCacheProbe } from './use-cache-probe-pool'
 import { installDevValidationWorker } from './dev-validation-worker-pool'
+import { PrerenderManifestUpdater } from './prerender-manifest-updater'
 import { join as pathJoin } from 'path'
 import { PUBLIC_DIR_MIDDLEWARE_CONFLICT } from '../../lib/constants'
 import { findPagesDir } from '../../lib/find-pages-dir'
@@ -86,7 +87,6 @@ import {
   ensureInstrumentationRegistered,
   getInstrumentationModule,
 } from '../lib/router-utils/instrumentation-globals.external'
-import type { PrerenderManifest } from '../../build'
 import { getRouteRegex } from '../../shared/lib/router/utils/route-regex'
 import type { PrerenderedRoute } from '../../build/static-paths/types'
 import { HMR_MESSAGE_SENT_TO_BROWSER } from './hot-reloader-types'
@@ -143,6 +143,7 @@ export default class DevServer extends Server {
   private staticPathsCache: LRUCache<
     UnwrapPromise<ReturnType<DevServer['getStaticPaths']>>
   >
+  private readonly prerenderManifestUpdater: PrerenderManifestUpdater
   private startServerSpan: Span
   private readonly serverComponentsHmrCache:
     | ServerComponentsHmrCache
@@ -221,6 +222,10 @@ export default class DevServer extends Server {
         }
       )
     }
+
+    this.prerenderManifestUpdater = new PrerenderManifestUpdater(
+      pathJoin(this.distDir, PRERENDER_MANIFEST)
+    )
 
     installUseCacheProbe({
       distDir: this.distDir,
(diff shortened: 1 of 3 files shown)


sirrryasir (commented):
The use of a queue coupled with atomic `rename` via a PID-suffixed tmp file is a very clean way to solve the dev manifest corruption. 

One small detail: since `this.queue = cycle.catch(() => {})` guarantees the queue won't hang if a cycle throws (e.g., if `readFile` hits `ENOENT` or `writeFile` fails due to disk space), recovering is graceful and the `unlink` in the catch block correctly cleans up the stale tmp file. Looks solid.

LouisTsang-jk:
Status for triage, since a few things have moved since this was opened:

- **Still unfixed on `16.3.1-canary.4`** — `packages/next/src/server/dev/next-dev-server.ts` L933-976 is unchanged.
- **Two more reports/PRs landed on the same bug.** #96664 (08/04) is an independent third report of the corruption, and #96695 (08/05, draft) is a second PR using the same serialize-and-write-atomically strategy, differing mainly in reusing the vendored `write-file-atomic` and covering it with an e2e test rather than unit tests. I've commented there proposing the two be consolidated instead of reviewed in parallel — I have no preference about which one lands, and I'm happy to close this in favour of that one if it's the better base.
- **CI is green and this is still cleanly mergeable**, though now 116 commits behind `canary`. Glad to rebase on request.

One note on scope: #96005 ("Per-route prerender-manifest.json") doesn't touch the dev write path, so it doesn't subsume this — but if that stack lands first, this will need a rebase onto the new manifest layout.

PR #96388, opened 2026-07-30
Fallback to worker threads when loopback binding is denied

DavidIlie:
## What?

When the default Turbopack Node.js plugin runtime cannot bind its
`127.0.0.1:0` IPC listener because a sandbox returns `PermissionDenied`, select
worker threads before backend-specific runtime code is emitted.

Explicit `turbopackPluginRuntimeStrategy` choices remain strict. Other probe
errors preserve the child-process default and its existing lazy startup
behavior. The capability probe is session-dependent so a persistent Turbo
Tasks cache cannot carry a backend decision into a differently sandboxed
session.

## Why?

The default Codex CLI macOS Seatbelt profile denies local socket binding.
Production builds that exercise PostCSS or webpack loaders therefore fail
after child-process startup retries with `binding to a port` / `EPERM`, while
webpack and Turbopack's worker-thread backend succeed.

This is the underlying compatibility issue reported in
https://github.com/openai/codex/issues/18895.

## Verification

- `cargo test -p next-core --lib loopback`
- `cargo test -p next-core --lib keeps_the_child_process_default_for_unexpected_probe_errors`
- `cargo check -p next-core`
- `cargo check -p next-core --features worker_pool`
- `cargo check -p next-core --no-default-features --features worker_pool`
- `pnpm exec jest packages/next/src/server/config.test.ts --runInBand`
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/production/turbopack-node-backend/turbopack-node-backend.test.ts`
- `CI=1 pnpm build-all`
- Direct Codex Seatbelt control: a fresh local-canary Turbopack production
  build completed, and the loader PID matched the build PID, confirming the
  automatic worker-thread fallback.
- Warm-cache control: an unsandboxed build populated the filesystem cache with
  different build/loader PIDs (child process); restoring that same cache under
  Codex Seatbelt completed with matching PIDs (worker thread), confirming the
  capability probe re-ran for the new session.

### DIFF
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -1,3 +1,5 @@
+#[cfg(any(all(feature = "process_pool", feature = "worker_pool"), test))]
+use std::future::Future;
 use std::time::Duration;
 
 use anyhow::{Context, Result, bail};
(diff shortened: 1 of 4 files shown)


crates/next-core/src/next_config.rs
sampoder, line 1: It feels a bit odd for this to live in the `next-core` crate, maybe we should put it in the Turbopack crate?

bgw:
I don't want to do this because the plan is to kill the TCP socket codepath in Next.js ASAP. That may come as soon as a patch release on 16.3 (e.g. 16.3.1 or 16.3.2).

PR #96390, opened 2026-07-30
Fix adapter outputs for not-found routes when used with cache components

ztanner:
## Summary

Fix the build output for App Router `not-found` routes that get partially prerendered when cacheComponents is enabled.

Previously, Next.js treated any statically generated `/_not-found` route as a complete static 404, even when it contained postponed state. It copied the partial HTML shell to `server/pages/404.html` and omitted the corresponding `/_not-found` prerender from adapter outputs.

This change distinguishes fully static not-found routes from partially prerendered ones. Fully static routes retain the existing `404.html` behavior, while PPR routes remain resumable prerender outputs.

This allows a `not-found.tsx` that suspends under Cache Components to render its dynamic content when deployed.

Closes https://github.com/vercel/next.js/pull/96399

## Testing

Expanded `not-found-non-document-dynamic` coverage to verify that:

- dynamic not-found content renders for ordinary missing paths
- dynamic not-found content renders when `/_not-found` is selected by a rewrite
- a PPR not-found does not emit `server/pages/404.html`
- its adapter output includes `PARTIALLY_STATIC` metadata, postponed state, and a resume chain

Also verified that fully static App Router not-found routes continue to emit and serve `pages/404.html`.

- `__NEXT_CACHE_COMPONENTS=true pnpm test-start-turbo test/e2e/app-dir/not-found-non-document-dynamic/not-found-non-document-dynamic.test.ts`
- `pnpm test-start-turbo test/e2e/app-dir/not-found/basic/index.test.ts`

### DIFF
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -1387,9 +1387,14 @@ export async function handleBuildComplete({
         const isAppPage =
           Boolean(appOutputMap[srcRoute]) || srcRoute === '/_not-found'
 
-        // if we already have 404.html favor that instead of
-        // _not-found prerender
-        if (srcRoute === '/_not-found' && hasStatic404) {
+        // If we already have a complete 404.html, favor that instead of the
+        // _not-found prerender. A partially static route only produced a shell,
+        // so preserve its prerender output in order to resume it.
+        if (
+          srcRoute === '/_not-found' &&
+          hasStatic404 &&
+          renderingMode !== RenderingMode.PARTIALLY_STATIC
+        ) {
           continue
         }
 
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -1387,17 +1387,6 @@ export async function handleBuildComplete({
         const isAppPage =
           Boolean(appOutputMap[srcRoute]) || srcRoute === '/_not-found'
 
-        // If we already have a complete 404.html, favor that instead of the
-        // _not-found prerender. A partially static route only produced a shell,
-        // so preserve its prerender output in order to resume it.
-        if (
-          srcRoute === '/_not-found' &&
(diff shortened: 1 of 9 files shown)


eps1lon:
This also fixes https://github.com/vercel/next.js/pull/96399 I assume. It was just broken on the build output and wasn't specific to which segment the resume was in?

ztanner:
@eps1lon yeah - same failure case there afaict

eps1lon (approved):

packages/next/src/build/adapter/build-complete.ts
```diff
@@ -1387,9 +1387,14 @@ export async function handleBuildComplete({
...
-        // if we already have 404.html favor that instead of
-        // _not-found prerender
-        if (srcRoute === '/_not-found' && hasStatic404) {
+        // If we already have a complete 404.html, favor that instead of the
+        // _not-found prerender. A partially static route only produced a shell,
+        // so preserve its prerender output in order to resume it.
+        if (
+          srcRoute === '/_not-found' &&
+          hasStatic404 &&
+          renderingMode !== RenderingMode.PARTIALLY_STATIC
```
gnoff: the rendering mode for a fully static PPR route will also be PARTIALLY_STATIC. I thin you need to consider the postponed state here instead or in addition to?

gnoff (dismissed):

packages/next/src/build/adapter/build-complete.ts
ztanner: Good catch. I used hasPostponed for the corresponding build decision but missed that distinction here

ztanner: Fixed and added a test case that would have caught this in [2f0543d](https://github.com/vercel/next.js/pull/96390/commits/2f0543d5b192444fb2709ab98768f168f7c450fa)

gnoff (approved):

PR #96392, opened 2026-07-30
Fix hybrid Pages/App Router not-found rendering with adapters

ztanner:
## Summary

When Pages Router and App Router routes coexist, a Pages Router route that returns `notFound` will internally hand rendering to the App Router not-found page. Next.js [prefers the App Router `/_not-found` entry for 404 responses](https://github.com/vercel/next.js/blob/5f5e4773bbb54f19697063ea093e8d89ca7d081a/packages/next/src/server/base-server.ts#L2965-L2980) before falling back to a Pages Router `/404` page.

When deployed via an adapter, this means it would have attempted to resume the Pages entry, which wouldn't contain the relevant prerendered shell or postponed state. If the not-found page suspends, its response cannot be resumed using the metadata associated with the Pages Router output.

For adapter requests, this marks the handoff with an empty postponed state. This uses the existing resume protocol to perform a complete, non-cacheable dynamic render. `next start` continues to use the normal PPR path, where it can load and resume the App Router not-found output directly.

This trades PPR shell reuse for correctness only during this cross-router not-found handoff. Longer term, adapters should have an explicit contract for cross-output handoffs so they can select and resume the App Router not-found output normally, without necessarily declaring the pages router output PPR-capable.

The change also documents the empty postponed-state contract and adds deployment regression coverage.

## Verification

- Added an end-to-end test that verifies a Pages Router `notFound` response includes the dynamic App Router not-found content when deployed.
- `pnpm testheadless test/e2e/app-dir/pages-router-app-not-found`
- `pnpm --filter=next types`


packages/next/src/server/route-modules/pages/pages-handler.ts
```diff
@@ -116,6 +116,15 @@ export const getHandler = ({
...
       // TODO: should route-module itself handle rendering the 404
       if (routerServerContext?.render404) {
+        // When Pages Router and App Router apps coexist, a Pages Router 404
+        // renders the App Router not-found page. If that page suspends when
+        // Cache Components is enabled, the renderer needs postponed state to
+        // resume it, but that information does not exist on a Pages Router
+        // entry. Mark the postponed state as empty to signal that the renderer
+        // should perform a full render instead of attempting to resume.
+        if (nextConfig.adapterPath && nextConfig.cacheComponents) {
+          addRequestMeta(req, 'postponed', '')
```
gnoff, line 132: I don't think we can do this unconditionally since an adapter could have set postponed data already. maybe this is better scoped to minimal mode only. Or we only set it if it doesn't already exist

gnoff: I guess you can't provide this meta through the pages entrypoint anyway so this is maybe wrong

gnoff: Also, adapterPath ever set when running in `next start` mode. Like some strange combo of configuring an adapter and then doing next build, next start? Is that the intended behavior?

gnoff (approved):

PR #96395, opened 2026-07-30
Enable turbopackFileSystemCacheForBuild by default

sokra:
### What?

Makes `experimental.turbopackFileSystemCacheForBuild` (Turbopack's on-disk
filesystem cache for `next build`) enabled by default, matching the dev-mode
flag. It stays disabled by default in non-Vercel CI environments, and can
always be turned off explicitly with
`experimental.turbopackFileSystemCacheForBuild: false`.

### Why?

The build filesystem cache greatly speeds up warm builds, but was previously
opt-in (only auto-enabled for canary builds on the Vercel builder). Defaulting
it on gives everyone faster subsequent builds, while keeping it off in
non-Vercel CI where the cache is unlikely to persist between builds.

### How?

- `turbopackFileSystemCacheForBuildDefault()` now returns `true` except when
  `isCI && !NOW_BUILDER` (removed the `isStableBuild()` gate).
- The explicit `false` opt-out already works via the existing
  `config.experimental?.turbopackFileSystemCacheForBuild || false` read.
- Adds an e2e test (`test/e2e/app-dir/turbopack-fs-cache-build-default`)
  asserting `.next/cache/turbopack` is written by default, empty when the flag
  is `false`, and empty in non-Vercel CI.
- Updates the `turbopackFileSystemCache` / Turbopack docs.

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackFileSystemCache.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackFileSystemCache.mdx
@@ -10,7 +10,7 @@ Turbopack FileSystem Cache enables Turbopack to reduce work across `next dev` or
 
 > **Good to know:** The FileSystem Cache is **enabled by default** for `next dev`. It is enabled by default when running `next build` locally or on Vercel, where the cache is likely to persist between builds. You can explicitly enable or disable it with the flags below.
 
-> **For deployment providers:** Other providers can enable the build cache by default on their platform by setting `experimental.turbopackFileSystemCacheForBuild: true` from their adapter's [`modifyConfig`](/docs/app/api-reference/adapters/creating-an-adapter) hook, when their infrastructure preserves the `.next/cache` directory between builds.
+> **For deployment providers:** Other providers can enable the build cache by default on their platform by setting `experimental.turbopackFileSystemCacheForBuild: true` from their adapter's [`modifyConfig`](/docs/app/api-reference/adapters/creating-an-adapter) hook. Do this when your infrastructure preserves the `.next/cache` directory between builds, and only when the user has not set an explicit value for the flag.
 
 ```ts filename="next.config.ts" switcher
 import type { NextConfig } from 'next'
@@ -41,9 +41,9 @@ module.exports = nextConfig
 
 ## Version Changes
 
-| Version   | Changes                                                                                    |
-| --------- | ------------------------------------------------------------------------------------------ |
-| `v16.3.0` | FileSystem caching is enabled by default for builds (disabled by default in non-Vercel CI) |
-| `v16.1.0` | FileSystem caching is enabled by default for development                                   |
-| `v16.0.0` | Beta release with separate flags for build and dev                                         |
-| `v15.5.0` | Persistent caching released as experimental on canary releases                             |
+| Version   | Changes                                                                              |
+| --------- | ------------------------------------------------------------------------------------ |
+| `v16.3.0` | FileSystem caching is enabled by default for builds (depends on CI platform support) |
+| `v16.1.0` | FileSystem caching is enabled by default for development                             |
+| `v16.0.0` | Beta release with separate flags for build and dev                                   |
+| `v15.5.0` | Persistent caching released as experimental on canary releases                       |
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackFileSystemCache.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackFileSystemCache.mdx
@@ -8,16 +8,15 @@ description: Learn how to enable FileSystem Caching for Turbopack builds
 
 Turbopack FileSystem Cache enables Turbopack to reduce work across `next dev` or `next build` commands. When enabled, Turbopack will save and restore data to the `.next` folder between builds, which can greatly speed up subsequent builds and dev sessions.
 
-> **Good to know:** The FileSystem Cache feature is considered stable for development and experimental for production builds
+> **Good to know:** The FileSystem Cache is **enabled by default** for both `next dev` and `next build`. For `next build` it is disabled by default in non-Vercel CI environments, where the cache is unlikely to persist between builds. You can explicitly enable or disable it with the flags below.
 
 ```ts filename="next.config.ts" switcher
 import type { NextConfig } from 'next'
 
 const nextConfig: NextConfig = {
   experimental: {
-    // Enable filesystem caching for `next dev`
+    // Filesystem caching is enabled by default; set to `false` to disable it.
     turbopackFileSystemCacheForDev: true,
-    // Enable filesystem caching for `next build`
     turbopackFileSystemCacheForBuild: true,
   },
 }
@@ -29,9 +28,8 @@ export default nextConfig
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   experimental: {
-    // Enable filesystem caching for `next dev`
+    // Filesystem caching is enabled by default; set to `false` to disable it.
     turbopackFileSystemCacheForDev: true,
-    // Enable filesystem caching for `next build`
     turbopackFileSystemCacheForBuild: true,
   },
 }
@@ -41,8 +39,9 @@ module.exports = nextConfig
 
 ## Version Changes
 
-| Version   | Changes                                                        |
-| --------- | -------------------------------------------------------------- |
-| `v16.1.0` | FileSystem caching is enabled by default for development       |
-| `v16.0.0` | Beta release with separate flags for build and dev             |
-| `v15.5.0` | Persistent caching released as experimental on canary releases |
+| Version   | Changes                                                                                    |
+| --------- | ------------------------------------------------------------------------------------------ |
+| `v16.3.0` | FileSystem caching is enabled by default for builds (disabled by default in non-Vercel CI) |
+| `v16.1.0` | FileSystem caching is enabled by default for development                                   |
+| `v16.0.0` | Beta release with separate flags for build and dev                                         |
+| `v15.5.0` | Persistent caching released as experimental on canary releases                             |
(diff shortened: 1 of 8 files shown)


docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackFileSystemCache.mdx
```diff
@@ -8,16 +8,15 @@ description: Learn how to enable FileSystem Caching for Turbopack builds
 
 Turbopack FileSystem Cache enables Turbopack to reduce work across `next dev` or `next build` commands. When enabled, Turbopack will save and restore data to the `.next` folder between builds, which can greatly speed up subsequent builds and dev sessions.
 
-> **Good to know:** The FileSystem Cache feature is considered stable for development and experimental for production builds
+> **Good to know:** The FileSystem Cache is **enabled by default** for both `next dev` and `next build`. For `next build` it is disabled by default in non-Vercel CI environments, where the cache is unlikely to persist between builds. You can explicitly enable or disable it with the flags below.
```
bgw: Avoid the negatives.

```suggestion
> **Good to know:** The FileSystem Cache is **enabled by default** for `next dev`. It is enabled by default when running `next build` locally or on Vercel, where the cache is likely to persist between builds. You can explicitly enable or disable it with the flags below.
```

Can we also put a note here about what environment variables other providers can set to opt-into enabling this by default?

cc @icyJoseph

docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -346,10 +346,10 @@ This can lead to subtle rendering differences when migrating from webpack to Tur
 
 ### Build Caching
 
-Webpack supports [disk build caching](https://webpack.js.org/configuration/cache/#cache) to improve build performance. Turbopack provides a similar feature, currently in beta. Starting with Next 16, you can enable Turbopack’s filesystem cache by setting the following experimental flags:
+Webpack supports [disk build caching](https://webpack.js.org/configuration/cache/#cache) to improve build performance. Turbopack provides a similar feature, currently in beta. Starting with Next 16, Turbopack's filesystem cache is controlled by the following experimental flags:
 
 - [`experimental.turbopackFileSystemCacheForDev`](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is enabled by default
-- [`experimental.turbopackFileSystemCacheForBuild`](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is currently opt-in
+- [`experimental.turbopackFileSystemCacheForBuild`](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is enabled by default (except in non-Vercel CI environments, where it is disabled by default)
```
bgw:
```suggestion
- [`experimental.turbopackFileSystemCacheForBuild`](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is enabled by default for local builds and on Vercel, where the build cache is likely to persist.
```

test/e2e/filesystem-cache/build-cache-default.test.ts
```diff
@@ -0,0 +1,139 @@
+import { nextTestSetup, isNextStart } from 'e2e-utils'
```
bgw, line 1: Do we need a new test for this? Don't we already have some e2e test coverage of the build cache? Should we just update/extend some other test suite?

bgw (changes requested):
We need to be very clear about the messaging here cc @icyJoseph

This isn't a Vercel-only feature, but we do need providers to opt in somehow (via an environment variable?) to indicate that they support the Turbopack build cache.

We don't do this by default on other providers because it the cache isn't preserved, or if the cache is slow to restore, it can be significantly worse than building with no cache at all.

test/e2e/filesystem-cache/build-cache-default.test.ts
sokra: I don't think have an existing test for the non-vercel-CI behavoir yet

andrewimm:
re: env variables... my immediate reaction was "we do have the env variable that enables this"
But I realize what you're saying is, we need an env variable that enables the default, but does not override an explicit `false` (which the current env var would do).

Otherwise I'd expect users who confirm their CI actually re-uses build artifacts to explicitly enable it in their config file.
In the preview blog post we mentioned this:
> CI setups can take advantage of this by copying the generated .next directory from one run to the next. When Turbopack sees the cache at the start of a build, it reads entries from disk before compiling any new changes.

We should be explicit in the docs about how to make this work:
1) Save the .next dir from your previous run
2) Restore it on the next run
3) Make sure you've enabled it in config, or set NEW_ENV_VARIABLE_TBD

bgw:
@andrewimm 

> I'd expect users who confirm their CI actually re-uses build artifacts to explicitly enable it in their config file

Oh, right, I think I forgot this conversation.

It's not just users. We need a way for providers to be able to override the default (unconfigured behavior) on behalf of their users.

The expectation was for other providers to do this in their adapters if they've confirmed they support the cache?

Either way, we should document both what users can configure, and *also* what providers can configure on behalf of their users (e.g. "providers should override the default in their adapters if the user has not set an explicit configuration").

docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -349,7 +349,9 @@ This can lead to subtle rendering differences when migrating from webpack to Tur
 Webpack supports [disk build caching](https://webpack.js.org/configuration/cache/#cache) to improve build performance. Turbopack provides a similar feature, currently in beta. Starting with Next 16, Turbopack's filesystem cache is controlled by the following experimental flags:
 
 - [`experimental.turbopackFileSystemCacheForDev`](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is enabled by default
-- [`experimental.turbopackFileSystemCacheForBuild`](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is enabled by default (except in non-Vercel CI environments, where it is disabled by default)
+- [`experimental.turbopackFileSystemCacheForBuild`](/docs/app/api-reference/config/next-config-js/turbopackFileSystemCache) is enabled by default for local builds and on Vercel, where the build cache is likely to persist.
+
+Other deployment providers can enable the build cache by default on their platform by setting `experimental.turbopackFileSystemCacheForBuild: true` from their adapter's [`modifyConfig`](/docs/app/api-reference/adapters/creating-an-adapter) hook, when their infrastructure preserves the `.next/cache` directory between builds.
```
andrewimm, line 354: Nice. Adapters is the right place for this.

docs/01-app/03-api-reference/05-config/01-next-config-js/turbopackFileSystemCache.mdx
```diff
@@ -41,8 +41,9 @@ module.exports = nextConfig
...
 ## Version Changes
 
-| Version   | Changes                                                        |
-| --------- | -------------------------------------------------------------- |
-| `v16.1.0` | FileSystem caching is enabled by default for development       |
-| `v16.0.0` | Beta release with separate flags for build and dev             |
-| `v15.5.0` | Persistent caching released as experimental on canary releases |
+| Version   | Changes                                                                                    |
+| --------- | ------------------------------------------------------------------------------------------ |
+| `v16.3.0` | FileSystem caching is enabled by default for builds (disabled by default in non-Vercel CI) |
```
bgw: Make this less negative and less Vercel-specific

```suggestion
| `v16.3.0` | FileSystem caching is enabled by default for builds (depends on CI platform support) |
```

docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -420,6 +422,9 @@ Additionally, the following experimental options are available under `experiment
 | [`turbopackLocalPostcssConfig`](/docs/app/api-reference/config/next-config-js/turbopackLocalPostcssConfig)   | Resolve `postcss.config.js` from the CSS file's directory first, then the project root.                                                      | `false`       | `false`                       |
 | `turbopackWorkerAssetPrefix`                                                                                 | Custom asset prefix for Web Worker URLs (entrypoint + module chunks), overriding `assetPrefix`. Mirrors webpack's `output.workerPublicPath`. | `undefined`   | `undefined`                   |
 
+<sup>1</sup> Enabled by default for local builds and on Vercel, where the build
+cache is likely to persist. Disabled by default in other CI environments.
```
bgw:
```suggestion
<sup>1</sup> Enabled by default for local builds and on CI platforms (including Vercel) where the build cache is likely to persist. Disabled by default on CI platforms that have not yet opted into the feature.
```

packages/next/src/server/config-shared.ts
```diff
@@ -2241,12 +2240,13 @@ export const defaultConfig = Object.freeze({
...
-    // Assume caching is available on vercel
-    return true
+  // Enabled by default, except in non-Vercel CI environments where the cache
+  // is unlikely to persist between builds (on Vercel, indicated by
+  // `NOW_BUILDER`, caching is available).
+  if (isCI && !process.env.NOW_BUILDER) {
+    return false
   }
-  return false
+  return true
```
bgw:
```suggestion
  return !isCI || process.env.NOW_BUILDER
```

packages/next/src/server/config-shared.ts
```diff
@@ -2241,12 +2240,13 @@ export const defaultConfig = Object.freeze({
 } satisfies NextConfig)
 
 function turbopackFileSystemCacheForBuildDefault() {
-  if (isStableBuild()) return false
-  if (isCI && process.env.NOW_BUILDER) {
-    // Assume caching is available on vercel
-    return true
+  // Enabled by default, except in non-Vercel CI environments where the cache
+  // is unlikely to persist between builds (on Vercel, indicated by
+  // `NOW_BUILDER`, caching is available).
```
bgw:
```suggestion
  // Disable in most CI environments, because we don't know if the cache will persist across builds.
  // Providers: Override the default behavior using `modifyConfig` in your adapter.
```

bgw (changes requested):

bgw (approved):
I think you should also get a 👍 from @icyJoseph before merging if you can, but LGTM.

andrewimm (approved):
Let's merge. Agree it'd be good to get more eyes on the docs, but we've explicitly called out the important bits for now. Docs can be ITG'd and re-published anytime, we shouldn't block package release on this, and I really want to get this in for 16.3

icyJoseph:
Yes, I do have some comments, but I can totally follow up with a new PR.

PR #96396, opened 2026-07-30
[turbopack] Track members across namespace re-exports

sampoder:
## What?

This PR closes a missed tree-shaking opportunity for libraries that expose a module namespace as a named export. One real-world example is Zod, whose public API has a structure similar to the following (and this pattern also appears in v0's codebase):

```js
// user.js
import { z } from './lib.js'

console.log(z.object(), z.string(), z.defaultLocale)
```

```js
// lib.js — mirrors zod/index.js
import * as z from './external.js'

export * from './external.js'
export { z }
export default z

// external.js — mirrors zod/v4/classic/external.js
import en from './locales/en.js'

export * as locales from './locales/index.js'

export const object = () => 'object'
export const string = () => 'string'
export const defaultLocale = en
```

The analyzer can now preserve the information that only `object`, `string`, and `defaultLocale` are read from `z`, and use it after resolving that named export to the namespace of `external.js`.

This PR is now standalone on `canary`. Its two former prerequisites have landed: #98680 introduced the independently reviewed `EsmAssetReferenceOptions` constructor refactor, and #96282 integrated CommonJS and ESM whole-namespace forwarding with canary's generic `ExportUsage::Passthrough` fixed point.

## Why?

Previously, Turbopack knew that the named export `z` was used, but lost the static member information from expressions such as `z.object()`. Once `z` resolved to a namespace object, the optimizer therefore kept every export of `external.js` — including `locales` and all of `./locales/index.js` in this example.

The optimization must remain conservative in several cases:

- a computed property whose name cannot be determined statically, such as `z[key]`;
- a namespace binding that escapes to unknown code, such as passing `z` to another function;
- static and opaque uses of the same binding, where the opaque use must dominate narrowing;
- live exports and method calls, whose binding and receiver semantics must not change.

It must also avoid changing ordinary named-import behavior. A property read on an imported value is not proof that the value is a module namespace; narrowing and namespace provenance are valid only after re-export resolution proves that fact.

### Represent namespace-member requests explicitly

The module-part model now has explicit states for this operation:

- `ExportedNamespaceMember { export, member }` represents a static member read on a named export before re-export resolution;
- `RenamedNamespaceMember { export, member }` represents the case after that export resolves to a module namespace.

This avoids overloading an ordinary `Export` or an optional `member` field with two meanings. It also keeps existing `Export` and `RenamedNamespace` identities unchanged; the new variants have distinct hashes only for the new states.

Multiple static reads and multiple aliases are tracked as independently keyed references, so `z.object`, `z.string`, and `z.defaultLocale` all survive rather than overwriting one another.

### Narrow only after proving the export is a namespace

The analyzer records a statically known member on an imported binding while leaving the emitted member expression intact. Re-export following first resolves the outer named export:

1. If it resolves to a namespace object, Turbopack synthesizes a namespace-member reference to the resolved module and contributes `PartialNamespaceObject({ member })` there.
2. If it resolves to an ordinary named export, it continues to behave like a normal named read of the outer module. It does not mark the module as namespace-observable or unnecessarily disable export-name mangling.

Dynamic property access and escaping bindings do not produce a narrowed member part, so they continue to retain the full namespace. When static and opaque uses are joined, the full-namespace use wins.

The explicit module-part variants are threaded through both fragment and non-fragment infrastructure so they remain valid graph identities. The new `ExportedNamespaceMember` rewrite itself currently runs only when module fragments are disabled. A dedicated `moduleFragmentsEnabled: true` fixture verifies the supported existing split behavior without claiming that this new rewrite is active there.

While rebasing, canary moved follow-reexport canonicalization into `turbopack-ecmascript` so lazy dynamic imports can carry the same canonicalization policy. Namespace-member handling now lives in that canonicalization helper as well, preserving both the new lazy-compilation path and this PR's namespace rewrite.

## Testing

Enabled coverage includes:

- the Zod-shaped snapshot showing only `object`, `string`, and `defaultLocale` retained while the unused locale namespace is removed;
- multiple static members on one namespace-valued export;
- dynamic access, escaping use, and mixed static/opaque use retaining the full namespace;
- direct `export * as`, including an enabled used/unused precision assertion;
- computed-string, parenthesized, optional, and missing-property reads;
- `this`-sensitive method calls;
- default-exported and renamed namespace values;
- live bindings read and mutated through multiple aliases, with single evaluation;
- nested namespace runtime behavior and top-level-await targets;
- scope-hoisted, non-scope-hoisted, and module-fragment configurations;
- export-name mangling safety and non-pessimization across #96282's forwarding edges.

Development coverage updates both a value behind the CommonJS forwarding edge and a statically read namespace member, verifying HMR observes each new value without stale exports or runtime errors.

One optimization-only limitation is documented under `__skipped__/namespace-value-reexport-nested`: member narrowing currently stops after the first namespace level for a read such as `z.locales.en`. The full nested runtime behavior is covered by the enabled `namespace-value-reexport` fixture.

Local verification on the standalone canary rebase passes affected-crate and `next-core` checks, 14 focused execution/snapshot tests, the merged passthrough/annotation unit tests, 60 analyzer graph tests, `build-all`, and the development/HMR test. A throwaway ordinary object-valued named-import fixture also verified that the pre-resolution path preserves property values and `this` for getters and methods; it was removed after review.

### DIFF
--- a/turbopack/crates/turbopack-core/src/ident.rs
+++ b/turbopack/crates/turbopack-core/src/ident.rs
@@ -247,9 +247,14 @@ impl AssetIdent {
                 ModulePart::Evaluation => {
                     1_u8.deterministic_hash(&mut hasher);
                 }
-                ModulePart::Export(export) => {
+                ModulePart::Export { name, member } => {
                     2_u8.deterministic_hash(&mut hasher);
-                    export.deterministic_hash(&mut hasher);
+                    name.deterministic_hash(&mut hasher);
+                    // Only hashed when present, so plain exports keep their existing idents.
+                    if let Some(member) = member {
+                        9_u8.deterministic_hash(&mut hasher);
+                        member.deterministic_hash(&mut hasher);
+                    }
                 }
                 ModulePart::RenamedExport {
                     original_export,
@@ -259,9 +264,14 @@ impl AssetIdent {
                     original_export.deterministic_hash(&mut hasher);
                     export.deterministic_hash(&mut hasher);
                 }
-                ModulePart::RenamedNamespace { export } => {
+                ModulePart::RenamedNamespace { export, members } => {
                     4_u8.deterministic_hash(&mut hasher);
                     export.deterministic_hash(&mut hasher);
+                    if let Some(members) = members {
+                        for member in members {
+                            member.deterministic_hash(&mut hasher);
+                        }
+                    }
                 }
                 ModulePart::Internal(id) => {
                     5_u8.deterministic_hash(&mut hasher);
--- a/turbopack/crates/turbopack-core/src/resolve/mod.rs
+++ b/turbopack/crates/turbopack-core/src/resolve/mod.rs
@@ -217,6 +217,11 @@ impl ExportUsage {
     pub fn named(name: RcStr) -> Vc<Self> {
         Self::Named(name).cell()
     }
+
+    #[turbo_tasks::function]
+    pub fn partial_namespace_object(names: Vec<RcStr>) -> Vc<Self> {
+        Self::PartialNamespaceObject(names.into_iter().collect()).cell()
+    }
 }
 
 #[turbo_tasks::value(shared)]
@@ -3358,15 +3363,20 @@ pub enum ModulePart {
     /// Represents the side effects of a module. This part is evaluated even if
     /// all exports are unused.
     Evaluation,
-    /// Represents an export of a module.
-    Export(RcStr),
+    /// Represents an export of a module. `member` is a hint that only that member is read
+    /// (`ns.member`); it narrows the export only if the export is a re-exported namespace.
+    Export { name: RcStr, member: Option<RcStr> },
     /// Represents a renamed export of a module.
     RenamedExport {
         original_export: RcStr,
         export: RcStr,
     },
-    /// Represents a namespace object of a module exported as named export.
-    RenamedNamespace { export: RcStr },
+    /// Represents a namespace object of a module exported as named export. `None` members means
+    /// the whole namespace object.
+    RenamedNamespace {
+        export: RcStr,
+        members: Option<Vec<RcStr>>,
+    },
     /// A pointer to a specific part.
     Internal(u32),
     /// The local declarations of a module.
@@ -3384,7 +3394,10 @@ impl ModulePart {
     }
 
     pub fn export(export: RcStr) -> Self {
-        ModulePart::Export(export)
+        ModulePart::Export {
+            name: export,
+            member: None,
+        }
     }
 
     pub fn renamed_export(original_export: RcStr, export: RcStr) -> Self {
@@ -3395,7 +3408,17 @@ impl ModulePart {
     }
 
     pub fn renamed_namespace(export: RcStr) -> Self {
-        ModulePart::RenamedNamespace { export }
+        ModulePart::RenamedNamespace {
+            export,
+            members: None,
+        }
+    }
+
+    pub fn renamed_namespace_members(export: RcStr, members: Vec<RcStr>) -> Self {
+        ModulePart::RenamedNamespace {
+            export,
+            members: Some(members),
+        }
     }
 
     pub fn internal(id: u32) -> Self {
@@ -3419,13 +3442,33 @@ impl Display for ModulePart {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
             ModulePart::Evaluation => f.write_str("module evaluation"),
-            ModulePart::Export(export) => write!(f, "export {export}"),
+            ModulePart::Export { name, member: None } => write!(f, "export {name}"),
+            ModulePart::Export {
+                name,
+                member: Some(member),
+            } => write!(f, "export {name} member {member}"),
             ModulePart::RenamedExport {
                 original_export,
                 export,
             } => write!(f, "export {original_export} as {export}"),
-            ModulePart::RenamedNamespace { export } => {
+            ModulePart::RenamedNamespace {
+                export,
+                members: None,
+            } => {
                 write!(f, "export * as {export}")
+            }
+            ModulePart::RenamedNamespace {
+                export,
+                members: Some(members),
+            } => {
+                write!(f, "export * as {export} members ")?;
+                for (i, member) in members.iter().enumerate() {
+                    if i > 0 {
+                        write!(f, ", ")?;
+                    }
+                    write!(f, "{member}")?;
+                }
+                Ok(())
             }
             ModulePart::Internal(id) => write!(f, "internal part {id}"),
             ModulePart::Locals => f.write_str("locals"),
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
@@ -2188,14 +2188,25 @@ impl VisitAstPath for Analyzer<'_, '_> {
                 self.add_effect(Effect::ImportedBinding {
                     esm_reference_index,
                     export: Some(prop_str.into()),
+                    member: None,
                     // point to the MemberExpression instead
                     ast_path: as_parent_path_skip_in(self.arena, ast_path, 1),
                     span: member.span(),
                 });
             } else {
+                // ast_path stays on the ident, so `.exportName` remains in the emitted code.
+                let mut member = None;
+                if export.is_some()
+                    && let Some(access) = member_access_parent(ast_path)
+                    && let Some(prop) = self.eval_context.eval_member_prop(self.arena, &access.prop)
+                    && let Some(prop_str) = prop.as_str()
+                {
+                    member = Some(RcStr::from(prop_str));
+                }
                 self.add_effect(Effect::ImportedBinding {
                     esm_reference_index,
                     export: export.map(|e| RcStr::from(e.as_str())),
+                    member,
                     ast_path: as_parent_path_in(self.arena, ast_path),
                     span: ident.span(),
                 })
--- a/turbopack/crates/turbopack-ecmascript/src/module_fragments/mod.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/module_fragments/mod.rs
@@ -383,7 +383,7 @@ async fn get_part_id(result: &SplitResult, part: &ModulePart) -> Result<u32> {
     // TODO implement ModulePart::Facade
     let key = match part {
         ModulePart::Evaluation => Key::ModuleEvaluation,
-        ModulePart::Export(export) => Key::Export(export.clone()),
+        ModulePart::Export { name, .. } => Key::Export(name.clone()),
         ModulePart::Exports => Key::Exports,
         ModulePart::Internal(part_id) => return Ok(*part_id),
         ModulePart::Locals
@@ -408,7 +408,7 @@ async fn get_part_id(result: &SplitResult, part: &ModulePart) -> Result<u32> {
     }
 
     // This is required to handle `export * from 'foo'`
-    if let ModulePart::Export(..) = part
+    if let ModulePart::Export { .. } = part
         && let Some(&v) = entrypoints
             .get(&Key::StarExports)
             .or_else(|| entrypoints.get(&Key::Exports))
--- a/turbopack/crates/turbopack-ecmascript/src/module_fragments/part/module.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/module_fragments/part/module.rs
@@ -158,12 +158,15 @@ impl EcmascriptModulePartAsset {
                 ));
             }
 
-            ModulePart::Export(export) => {
+            ModulePart::Export { name: export, .. } => {
                 if entrypoints.contains_key(&Key::Export(export.clone())) {
                     return Ok(Vc::upcast(
                         EcmascriptModulePartAsset::new_with_resolved_part(
                             module,
-                            ModulePart::Export(export),
+                            ModulePart::Export {
+                                name: export,
+                                member: None,
+                            },
                         ),
                     ));
                 }
@@ -205,7 +208,10 @@ impl EcmascriptModulePartAsset {
                 }
                 let side_effects_module = SideEffectsModule::new(
                     module,
-                    ModulePart::Export(export),
+                    ModulePart::Export {
+                        name: export,
+                        member: None,
+                    },
                     *final_module,
                     side_effects.iter().map(|v| **v).collect(),
                 );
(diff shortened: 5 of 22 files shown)


turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3358,15 +3363,20 @@ pub enum ModulePart {
...
         export: RcStr,
     },
-    /// Represents a namespace object of a module exported as named export.
-    RenamedNamespace { export: RcStr },
+    /// Represents a namespace object of a module exported as named export. `None` members means
+    /// the whole namespace object.
+    RenamedNamespace {
+        export: RcStr,
+        members: Option<Vec<RcStr>>,
+    },
```
mischnic: This should ideally never happen with multiple members.

The whole reexport-side-effect-optimization mechanism relies on the fact that basically all references only target a single export. And then those references can be resolved to different modules through a barrel file.
A single import with multiple statically known members here would prevent that.

sampoder: yeah this is true. I switched this to just `member`

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3358,15 +3363,20 @@ pub enum ModulePart {
...
         original_export: RcStr,
         export: RcStr,
     },
-    /// Represents a namespace object of a module exported as named export.
-    RenamedNamespace { export: RcStr },
+    /// Represents a namespace object of a module exported as named export. `None` member means
+    /// the whole namespace object.
+    RenamedNamespace {
+        export: RcStr,
+        member: Option<RcStr>,
```
mischnic, line 3378: Still confused what `member` means here.

sampoder: maybe `property` would be a better name?

it's `object` in:

```
import { z } from './lib.js'
z.object()
```

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3402,15 +3407,20 @@ pub enum ModulePart {
...
+    ExportedNamespaceMember { export: RcStr, member: RcStr },
     /// Represents a renamed export of a module.
     RenamedExport {
         original_export: RcStr,
         export: RcStr,
     },
-    /// Represents a namespace object of a module exported as named export.
+    /// Represents a namespace object of a module exported as a named export.
     RenamedNamespace { export: RcStr },
+    /// Represents one static member of a namespace object exported as a named export.
```
lukesandberg, line 3430: an example would be useful here

`import ns from '...'; export const foo = ns.bar`  is what i think this means?

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -2472,14 +2472,25 @@ impl VisitAstPath for Analyzer<'_, '_> {
...
+                    && let Some(access) = member_access_parent(ast_path)
+                    && let Some(prop) = self.eval_context.eval_member_prop(self.arena, &access.prop)
+                    && let Some(prop_str) = prop.as_str()
+                {
+                    member = Some(RcStr::from(prop_str));
+                }
                 self.add_effect(Effect::ImportedBinding {
                     esm_reference_index,
                     export: export.map(|e| RcStr::from(e.as_str())),
+                    member,
```
lukesandberg, line 2489: couldn't we support an arbitrary number of member accesses

import {ns} from ...

console.log(ns.foo.bar.baz)

where baz is actually an export?  follow_rexports could dect that if we passed all the idents i think

lukesandberg (approved):

turbopack/crates/turbopack-core/src/ident.rs
```diff
@@ -251,6 +251,13 @@ impl AssetIdent {
                     2_u8.deterministic_hash(&mut hasher);
                     export.deterministic_hash(&mut hasher);
                 }
+                ModulePart::ExportedNamespaceMember { export, member } => {
+                    // Keep the named-export prefix stable, then distinguish the namespace member.
+                    2_u8.deterministic_hash(&mut hasher);
+                    export.deterministic_hash(&mut hasher);
+                    9_u8.deterministic_hash(&mut hasher);
+                    member.deterministic_hash(&mut hasher);
```
sokra:
```suggestion
                    3_u8.deterministic_hash(&mut hasher);
                    export.deterministic_hash(&mut hasher);
                    member.deterministic_hash(&mut hasher);
```

Renumber all the cases

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3488,12 +3511,16 @@ impl Display for ModulePart {
         match self {
             ModulePart::Evaluation => f.write_str("module evaluation"),
             ModulePart::Export(export) => write!(f, "export {export}"),
+            ModulePart::ExportedNamespaceMember { export, member } => {
+                write!(f, "export {export} namespace member {member}")
```
sokra:
```suggestion
                write!(f, "export {export} .{member}")
```

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3488,12 +3511,16 @@ impl Display for ModulePart {
...
+            }
             ModulePart::RenamedExport {
                 original_export,
                 export,
             } => write!(f, "export {original_export} as {export}"),
-            ModulePart::RenamedNamespace { export } => {
-                write!(f, "export * as {export}")
+            ModulePart::RenamedNamespace { export } => write!(f, "export * as {export}"),
+            ModulePart::RenamedNamespaceMember { export, member } => {
+                write!(f, "export * as {export} namespace member {member}")
```
sokra:
```suggestion
                write!(f, "export * as {export} .{member}")
```

turbopack/crates/turbopack-core/src/ident.rs
sokra:
```suggestion
                    4_u8.deterministic_hash(&mut hasher);
```

turbopack/crates/turbopack-core/src/ident.rs
```diff
@@ -263,6 +270,12 @@ impl AssetIdent {
                     4_u8.deterministic_hash(&mut hasher);
```
sokra:
```suggestion
                    5_u8.deterministic_hash(&mut hasher);
```

turbopack/crates/turbopack-core/src/ident.rs
```diff
@@ -263,6 +270,12 @@ impl AssetIdent {
...
                     export.deterministic_hash(&mut hasher);
                 }
+                ModulePart::RenamedNamespaceMember { export, member } => {
+                    4_u8.deterministic_hash(&mut hasher);
+                    export.deterministic_hash(&mut hasher);
+                    9_u8.deterministic_hash(&mut hasher);
+                    member.deterministic_hash(&mut hasher);
```
sokra:
```suggestion
                    6_u8.deterministic_hash(&mut hasher);
                    export.deterministic_hash(&mut hasher);
                    member.deterministic_hash(&mut hasher);
```

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3426,15 +3431,25 @@ pub enum ModulePart {
...
+    ExportedNamespaceMember { export: RcStr, member: RcStr },
     /// Represents a renamed export of a module.
     RenamedExport {
         original_export: RcStr,
         export: RcStr,
     },
-    /// Represents a namespace object of a module exported as named export.
+    /// Represents a namespace object of a module exported as a named export.
     RenamedNamespace { export: RcStr },
+    /// Represents one static member of a namespace object exported as a named export.
```
sokra: Also include an example here. Best include examples for all variants of that enum.

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3426,15 +3431,25 @@ pub enum ModulePart {
...
+    /// Represents a named export of a module.
     Export(RcStr),
+    /// Represents a static member read from an export that resolves to a namespace object.
+    ///
+    /// For example, `import { ns } from "./lib"; ns.bar` is represented as
+    /// `{ export: "ns", member: "bar" }` when `lib` contains
+    /// `import * as ns from "./source"; export { ns }`.
+    ///
+    /// If the export does not resolve to a namespace, this behaves like [`ModulePart::Export`].
+    ExportedNamespaceMember { export: RcStr, member: RcStr },
```
sokra: I think `ExportMember` would be a better name. When using this we don't yet know if the export is a namespace object. We only know that we are accessing a member of an export.

turbopack/crates/turbopack-ecmascript/src/module_fragments/part/module.rs
```diff
@@ -152,6 +152,11 @@ impl EcmascriptModulePartAsset {
             return Ok(Vc::upcast(module));
         };
 
+        let namespace_member = match &part {
+            ModulePart::ExportedNamespaceMember { member, .. } => Some(member.clone()),
+            _ => None,
+        };
```
sokra: This should move into the `match` below into the branch with ExportedNamespaceMember

turbopack/crates/turbopack-core/src/resolve/mod.rs
sokra: Actually we should call it `PartialExport`, since it still represents the export, but only one member is used from it.

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3426,15 +3431,25 @@ pub enum ModulePart {
...
     /// Represents a renamed export of a module.
     RenamedExport {
         original_export: RcStr,
         export: RcStr,
     },
-    /// Represents a namespace object of a module exported as named export.
+    /// Represents a namespace object of a module exported as a named export.
     RenamedNamespace { export: RcStr },
+    /// Represents one static member of a namespace object exported as a named export.
+    RenamedNamespaceMember { export: RcStr, member: RcStr },
```
sokra: This should be called `RenamedPartialNamespace`. This exports a (partial) namespace object as a named exports. We know that only a single export (`member`) is used from the namespace object.

turbopack/crates/turbopack-ecmascript/src/references/esm/base.rs
```diff
@@ -738,7 +738,12 @@ impl ModuleReference for EsmAssetReference {
         )
         .await?;
 
-        if let Some(ModulePart::Export(export_name)) = &self.export_name {
+        if let Some(export_name) = self.export_name.as_ref().and_then(|part| match part {
+            ModulePart::Export(export) | ModulePart::ExportedNamespaceMember { export, .. } => {
+                Some(export)
+            }
+            _ => None,
+        }) {
```
sokra: We should add a helper method to `ModulePart`. `ModulePart::get_export(&self) -> Option<&RcStr>` should return the export name.

turbopack/crates/turbopack-ecmascript/src/module_fragments/mod.rs
```diff
@@ -408,10 +411,12 @@ async fn get_part_id(result: &SplitResult, part: &ModulePart) -> Result<u32> {
...
 
     // This is required to handle `export * from 'foo'`
-    if let ModulePart::Export(..) = part
-        && let Some(&v) = entrypoints
-            .get(&Key::StarExports)
-            .or_else(|| entrypoints.get(&Key::Exports))
+    if matches!(
+        part,
+        ModulePart::Export(_) | ModulePart::ExportedNamespaceMember { .. }
+    ) && let Some(&v) = entrypoints
```
sokra: This could use `part.get_export().is_some()`

turbopack/crates/turbopack-ecmascript/src/references/esm/base.rs
```diff
@@ -789,6 +794,12 @@ impl ModuleReference for EsmAssetReference {
                 (Some(ModulePart::Export(export_name)), _) => {
                     ExportUsage::Named(export_name.clone())
                 }
+                (Some(ModulePart::ExportedNamespaceMember { export, .. }), _) => {
```
sokra: This can be merged with the previous case

turbopack/crates/turbopack-ecmascript/src/references/esm/base.rs
```diff
@@ -879,9 +890,10 @@ impl EsmAssetReference {
...
-                                    this.export_name.as_ref().and_then(|e| match e {
-                                        ModulePart::Export(export_name) => {
-                                            Some(export_name.clone())
+                                    this.export_name.as_ref().and_then(|part| match part {
+                                        ModulePart::Export(export)
+                                        | ModulePart::ExportedNamespaceMember { export, .. } => {
+                                            Some(export.clone())
                                         }
                                         _ => None,
                                     }),
```
sokra: use `ModulePart::get_export()`

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3417,26 +3422,42 @@ async fn resolve_package_internal_with_imports_field(
 #[turbo_tasks::task_input]
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash, Encode, Decode)]
 pub enum ModulePart {
-    /// Represents the side effects of a module. This part is evaluated even if
-    /// all exports are unused.
+    /// Represents the side effects of a module, such as `import "./module"`.
+    /// This part is evaluated even if all exports are unused.
     Evaluation,
-    /// Represents an export of a module.
+    /// Represents a named export, such as `foo` in `import { foo } from "./module"`.
```
sokra:
```suggestion
    /// Represents a named export, such as `foo` in `export const foo = "bar";`.
```

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3417,26 +3422,42 @@ async fn resolve_package_internal_with_imports_field(
...
+    /// `{ export: "ns", member: "member" }` after resolving `ns` to the namespace object.
+    RenamedPartialNamespace { export: RcStr, member: RcStr },
+    /// Points to a generated module-fragment part by its numeric index, such as the part that
+    /// contains one local declaration.
     Internal(u32),
-    /// The local declarations of a module.
+    /// Represents the local declarations of a module, such as `const value = 1`.
     Locals,
-    /// The whole exports of a module.
+    /// Represents the module's export declarations, such as `export { value }`.
```
sokra:
```suggestion
    /// Represents the module's export declarations, such as `export { foo, bar }`.
```

PR #96400, opened 2026-07-30
Fix `isHeadPartial` when hydrating from a static fallback shell

acdlite:
When a static fallback shell is served as-is — e.g. from a CDN in a deployed environment — the client can't hydrate from inline Flight data; it fetches the full RSC payload separately and reconstructs the initial payload on the client. During that reconstruction, the head's partiality flag was accidentally set to the head itself. Since the head is always truthy, the head was unconditionally treated as partial.

This happens to be unobservable today: the only flows that reach the reconstruction are fallback-shell prerenders, where the server pessimistically marks the head partial anyway, and when Cache Components is enabled the client ignores the per-payload flag entirely in favor of response-level partiality. But we're about to rework the RSC response format so that head partiality has a single, load-bearing source of truth, so the flag needs to survive the reconstruction correctly rather than by coincidence.

While auditing the payload consumers, I also fixed the payload shape check in segment prefetch generation (it used `&&` where `||` was intended, so it accepted any shape) and deleted `should-hard-navigate.ts`, which no longer has any callers on the client.

Adds e2e coverage asserting that metadata survives hydrating a fallback shell page: the title is correct after hydration, the head is included in prefetched data, and a return navigation is served entirely from the cache that was populated during initial hydration. In deployed environments this exercises the client resume path, which is where the reconstruction runs.


packages/next/src/server/app-render/collect-segment-data.tsx
```diff
@@ -256,7 +258,10 @@ function extractFlightData(initialRSCPayload: InitialRSCPayload): {
 } | null {
   const flightDataPaths = initialRSCPayload.f
   // FlightDataPath is an unsound type, hence the additional checks.
-  if (flightDataPaths.length !== 1 && flightDataPaths[0].length !== 3) {
+  if (
+    flightDataPaths.length !== 1 ||
+    (flightDataPaths[0].length !== 3 && flightDataPaths[0].length !== 4)
```
acdlite, line 263: Reason I found this is because I'm about to refactor the over-the-wire protocol, so the array indexing will go away for good

ztanner: I think this logic bug is also [here](https://github.com/vercel/next.js/blob/71436d6a13bae94156113cc8fc1b4d58629e4995/packages/next/src/server/app-render/instant-validation/instant-validation.tsx#L670)

ztanner (approved):

packages/next/src/server/app-render/collect-segment-data.tsx
acdlite: Good catch — fixed in d059daa0cb. Same pattern, but this copy throws an InvariantError instead of logging, so the corrected check also has to accept both the 4-element shape from getRSCPayload and the 3-element shape from instant-validation's own payload reconstruction.

PR #96423, opened 2026-07-31
fix: throw when generateStaticParams returns incomplete params with output: export

deepshekhardas:
Rebased version of #95969.

Throws when generateStaticParams returns an empty array or incomplete params when using output: export, and documents the new errors.


deepshekhardas:
Superseded - original #95969 merged upstream

PR #96424, opened 2026-07-31
fix: recognize Meta's 2024 crawler UAs as HTML-limited bots

deepshekhardas:
Rebased version of #96033.

Adds Meta's 2024 crawler user agents to the HTML-limited bot list.


deepshekhardas:
Superseded - original #96033 already open

PR #96426, opened 2026-07-31
[Cache] Make caches error if called after prerender aborts

### LINKED ISSUE #96339: "use cache" fill aborted by renderSignal during a runtime prefetch is stored as a valid entry — truncated Flight prelude permanently breaks all readers ("Error: Connection closed.")

https://github.com/hypernomad/next-use-cache-halted-prelude-repro

### To Reproduce

1. `npm ci`
2. `npm run repro`

The driver builds the app (production), boots `next start`, and runs an A/B on a cold server per round:

- **Control**: ordinary `GET /slow` → the `"use cache"` entry fills correctly, all routes healthy.
- **Poison**: `GET /slow` with `RSC: 1` + `Next-Router-Prefetch: 2` (a runtime prefetch, i.e. a runtime prerender) → the fill is aborted by `renderSignal` mid-flight, and the **halted (truncated) Flight prelude is stored as a healthy cache entry**.

Two other routes (`/late-reader`, `/b`) that only ever *read* the entry then fail permanently with `Error: Connection closed.` — re-verified over 3 further reads. Reproduces **5/5 rounds deterministically** on `next@16.3.0-canary.102`. No DB, no network, no client abort involved.

The repro's trigger condition: the cached function is awaited behind ~2.5s of **uncached** async work, so its fill *starts after* `cacheSignal.cacheReady()` has already resolved. Fills that start before `cacheReady()` are waited for (that's why naive repros fail); fills that start after run under an already-aborted `renderSignal`.

**Current:** In `generateCacheEntryImpl` (`next/dist/server/use-cache/use-cache-wrapper.js`), the fill's abort signal is `AbortSignal.any([dynamicAccessAbortSignal, outerWorkUnitStore.renderSignal, timeout])`, but the post-`prerender()` handling only checks the timeout branch (errored stream — correct) and the dynamic-access branch (hanging promise — correct). An abort that came from **`renderSignal` matches neither branch and falls through to `stream = prelude`** (canary.102: `use-cache-wrapper.js:840`). React's `prerender()` halts pending tasks without emitting error rows, so the entry is persisted with `isErrored: false`. The reader at `use-cache-wrapper.js:2389` does not pass `unstable_allowPartialStream`, so every subsequent read of the entry rejects with `Error: Connection closed.` — permanently, until the entry expires or the process restarts. On a long-lived `next start` deployment this is a persistent outage of every route sharing that entry (e.g. a layout-level cached function), while static PPR shells keep returning HTTP 200.

Note the asymmetry: Next's own client-side readers pass `allowPartialStream: true` (`fetch-server-response.js:353,447`, `segment-cache/cache.js:1369`); only the server-side use-cache reader treats a partial stream as fatal.

**Regression note:** the defective branch is byte-identical in `16.2.12` and `16.3.0-preview.9`, but in preview.9 this trigger produced a zero-byte halted prelude which the LRU rejected (`calculateSize` returned 0) — an incidental protection. canary.102 changed the size function to `entry.size + cacheKey.length` (`cache-handlers/default.js:39`), so zero-byte halted entries are now stored: **latest canary is strictly worse**. (Mid-fill aborts can also produce non-empty truncated preludes, which are stored on all versions — that variant is what we originally hit in production-like testing on preview.9.)

**Expected:** Either (a) a fill aborted by `renderSignal` must not be persisted as a valid entry (mark errored / discard, as the timeout branch already does), or (b) the server-side use-cache reader should tolerate partial streams like the client readers do. (a) seems preferable — it also avoids serving silently truncated content.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin 25.5.0
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 24.15.0
  npm: 11.12.1
Relevant Packages:
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 5.9.3
Next.js Config:
  cacheComponents: true
  partialPrefetching: true
  output: N/A
```

Use Cache, Partial Prerendering (PPR)

next start (local), Other (self-hosted)

### Additional context

- Previously filed without a reproduction link as #96338 (auto-closed) — this issue supersedes it with a deterministic public reproduction.
- First observed in a production-like environment (`next start`, `16.3.0-preview.9`, `cacheComponents` + `partialPrefetching`) under sustained prefetch traffic mixed with Server Action mutations: a layout-shared cached entry got poisoned and every route reading it served `⨯ Error: Connection closed. digest: '1097405047'` until restart, while routes without shell-level cache reads stayed healthy. DB/Redis were verified healthy; the query layer wasn't even invoked on failing requests.
- A warm server can also be poisoned via `revalidateTag(tag, { expire: 0 })` followed by a runtime prefetch (verified manually in the repro app).
- Impact note: on serverless the blast radius is one instance and heals on expiry (appears as sporadic transient content loss); on self-hosted long-running servers it is a permanent outage that ordinary uptime monitoring misses (static shell stays HTTP 200). It also makes any shared/remote `"use cache"` handler dangerous, since one poisoned fill would propagate to all instances.

lubieowoce:
Closes #96339

Caches that started filling after a prerender was aborted would produce an empty stream, because an aborted `renderSignal` made [the signal we pass to `prerender()` inside `use-cache-wrapper`] be aborted immediately. So we produced an empty entry and saved it, essentially poisoning the cache.

Fixed by removing `renderSignal` from the `AbortSignal.any(...)` we pass in and returning a rejected promise before we get to the cache filling codepath. This should not be observable to userspace because the prerender is already aborted.

### DIFF
--- a/packages/next/src/server/dynamic-rendering-utils.ts
+++ b/packages/next/src/server/dynamic-rendering-utils.ts
@@ -91,6 +91,17 @@ export function makeDynamicHangingPromise<T>(
   )
 }
 
+export function makeUntrackedHangingPromise<T>(
+  signal: AbortSignal,
+  route: string,
+  expression: string
+): Promise<T> {
+  return makeHangingPromiseWithError(
+    signal,
+    new HangingPromiseRejectionError(route, expression)
+  )
+}
+
 /**
  * Constructs a promise that never resolves, standing in for *runtime* data:
  * data that hangs during a static prerender but is available during a runtime
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -42,6 +42,7 @@ import {
   makeDynamicHangingPromise,
   makeRuntimeHangingPromise,
   makeStageHangingPromise,
+  makeUntrackedHangingPromise,
   RENDER_STAGES_BY_DATA_KIND,
 } from '../dynamic-rendering-utils'
 
@@ -1283,7 +1284,7 @@ async function generateCacheEntryImpl(
 
   switch (outerWorkUnitStore.type) {
     case 'prerender-runtime':
-    case 'prerender':
+    case 'prerender': {
       const timeoutAbortController = new AbortController()
       const timer = setTimeout(
         () => {
@@ -1299,7 +1300,6 @@ async function generateCacheEntryImpl(
       const abortSignal = dynamicAccessAbortSignal
         ? AbortSignal.any([
             dynamicAccessAbortSignal,
-            outerWorkUnitStore.renderSignal,
             timeoutAbortController.signal,
           ])
         : timeoutAbortController.signal
@@ -1357,6 +1357,7 @@ async function generateCacheEntryImpl(
         stream = prelude
       }
       break
+    }
     case 'request':
       // TODO: We should just check if the render is abandonable. This is
       // relevant in restart-on-cache-miss in general, so when we implement that
@@ -1640,6 +1641,51 @@ export async function cache(
     )
   }
 
+  const workUnitStore = workUnitAsyncStorage.getStore()
+  if (workUnitStore === undefined) {
+    throw new InvariantError(
+      '"use cache" cannot be used outside of App Router. Expected a WorkUnitStore.'
+    )
+  }
+
+  // In a prerender, tasky IO may result in cache reads that start
+  // after the prerender has already been aborted:
+  //
+  //   await setTimeout(100)  // we can't abort this uncached IO...
+  //   await cachedData()     // ...so we'll still get here even though the prerender aborted
+  //
+  // The prerender is over, so we should just return an erroring promise.
+  // (NOTE: we also shouldn't fill this cache, because it's behind uncached IO,
+  // so semantically it is not part of the prerender)
+  switch (workUnitStore.type) {
+    case 'prerender':
+    case 'prerender-runtime': {
+      if (workUnitStore.renderSignal.aborted) {
+        // We don't know if the cache itself is dynamic or runtime data,
+        // but the prerender is over, so it doesn't need to participate
+        // in runtime data tracking at all.
+        return makeUntrackedHangingPromise<never>(
+          workUnitStore.renderSignal,
+          workStore.route,
+          '"use cache" called after prerender ended'
+        )
+      }
+      break
+    }
+    case 'prerender-ppr':
+    case 'prerender-legacy':
+    case 'prerender-client':
+    case 'validation-client':
+    case 'request':
+    case 'cache':
+    case 'private-cache':
+    case 'unstable-cache':
+    case 'generate-static-params':
+      break
+    default:
+      workUnitStore satisfies never
+  }
+
   // Probe re-executions (the dev-server's hang-detection worker) short-circuit
   // further down before any handler is consulted, so we skip handler selection
   // entirely and the worker can boot without registering handlers at all.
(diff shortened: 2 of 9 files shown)


packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -1299,7 +1300,6 @@ async function generateCacheEntryImpl(
       const abortSignal = dynamicAccessAbortSignal
         ? AbortSignal.any([
             dynamicAccessAbortSignal,
-            outerWorkUnitStore.renderSignal,
```
unstubbable, line 1302: Talked to @gnoff about it, let's keep it here and instead add that signal to this condition:
https://github.com/vercel/next.js/blob/0b56bf592bf171aa7d6e2b559152d0b546555b5a/packages/next/src/server/use-cache/use-cache-wrapper.ts#L1327
so that we error the stream, and the cache handler wouldn't silently save a partial/halted stream. We probably can't reach this case in an e2e test because we always await `cacheReady`, but it still feels semantically correct to do, despite the early hanging promise return at the top level we've added now.

packages/next/src/server/dynamic-rendering-utils.ts
```diff
@@ -91,6 +91,17 @@ export function makeDynamicHangingPromise<T>(
   )
 }
 
+export function makeUntrackedHangingPromise<T>(
```
unstubbable, line 94: Isn't that the same as `makeDynamicHangingPromise`?

lubieowoce: yes, but it's meant to express a different intent. felt weird to do use a dynamic one if we don't actually know

packages/next/src/server/use-cache/use-cache-wrapper.ts
lubieowoce: that's what i did originally, but i believe that it wouldn't be correct when u factor in leader/joiner stuff. if we did a bailout like this in `generateCacheEntryImpl`, and then had an ordering of events where:

1. the prerender tries to fill a cache entry, becomes leader
2. another prerender needs the same cache entry, becomes joiner
3. leader notices that its prerender is already aborted and resolves the shared result with `type: prerender-dynamic` 

then we'd do the wrong thing, because the joiner would also get `prerender-dynamic`, and that's incorrect, because we want to abort the cache fill, not omit it from every other prerender. in other words, i think that once we've started filling, we have to commit to it, because other prerenders might depend on the same cache fill, so it's correct to omit `renderSignal` (unless i'm missing something about how deduping works)

(also we'd add the entry to `dynamicCacheKeys` which also feels semantically incorrect: https://github.com/vercel/next.js/blob/b4e3fecc7943d75a3f9c407343091520afafeaac/packages/next/src/server/use-cache/use-cache-wrapper.ts#L3141-L3143)

packages/next/src/server/dynamic-rendering-utils.ts
lubieowoce: in other words: the tracking is currently implemented such that "dynamic" is the same as "untracked" because we only track runtime hanging promises. but it feels weird to implicitly rely on that

packages/next/src/server/use-cache/use-cache-wrapper.ts
lubieowoce: oh, sorry, i'm shadowboxing here, you were talking about the timeout branch, not the dynamicAccess branch, so my arguments about deduping might not apply. lemme see

unstubbable: You're right, that makes sense.

test/e2e/app-dir/use-cache-after-uncached-io/components/offline-status.tsx
unstubbable, line 1: Accidentally committed?

unstubbable (approved):

test/e2e/app-dir/use-cache-after-uncached-io/components/offline-status.tsx
lubieowoce: oops yeah, copy pasted helpers 😓

PR #96435, opened 2026-07-31
docs: proxy event argument

icyJoseph:
The event parameter (second position) passed to proxy was not documented in the Params section


aurorascharff (approved):

docs/01-app/03-api-reference/03-file-conventions/proxy.mdx
```diff
@@ -156,6 +158,36 @@ export function proxy(request) {
 }
 ```
 
+### `event`
+
+The second parameter is an instance of `NextFetchEvent`, which extends the native [`FetchEvent`](https://developer.mozilla.org/docs/Web/API/FetchEvent). It exposes a single method, `waitUntil(promise)`, which keeps the Proxy invocation alive until the promise settles, so background work like logging or analytics can finish after the response is sent. See [`waitUntil` and `NextFetchEvent`](#waituntil-and-nextfetchevent) for a full example.
```
icyJoseph:
```suggestion
The second parameter is an instance of `NextFetchEvent`. It exposes a single method, `waitUntil(promise)`, which keeps the Proxy invocation alive until the promise settles, so background work like logging or analytics can finish after the response is sent. See [`waitUntil` and `NextFetchEvent`](#waituntil-and-nextfetchevent) for a full example.
```

PR #96453, opened 2026-08-01
Trace development route preparation

DavidIlie:
## Summary

This PR introduces the first bounded route-preparation phases used by Request Insights. It instruments existing matcher and bundler boundaries without changing their behavior or making the spans default-public OpenTelemetry data.

- trace development route matching and route preparation while a route is ensured
- keep production matcher reload as a separate sibling phase so reload work is not charged to matching
- surface the internal phases as `match route`, `prepare route`, and `reload route matchers` in Request Insights
- cover cold and warm page, App Route/API, and dynamic-route behavior in both development bundlers
- preserve the existing matcher results, errors, and public OTel allowlist

## Verification

- `pnpm test-dev-turbo test/development/app-dir/request-insights-route-preparation/request-insights-route-preparation.test.ts`
- `pnpm test-dev-webpack test/development/app-dir/request-insights-route-preparation/request-insights-route-preparation.test.ts`
- `pnpm --filter=next types`

### DIFF
--- a/packages/next/src/server/route-matcher-managers/dev-route-matcher-manager.ts
+++ b/packages/next/src/server/route-matcher-managers/dev-route-matcher-manager.ts
@@ -7,6 +7,8 @@ import path from '../../shared/lib/isomorphic/path'
 import * as Log from '../../build/output/log'
 import { cyan } from '../../lib/picocolors'
 import type { RouteMatcher } from '../route-matchers/route-matcher'
+import { DevRouteMatcherManagerSpan } from '../lib/trace/constants'
+import { getTracer } from '../lib/trace/tracer'
 
 export interface RouteEnsurer {
   ensure(match: RouteMatch, pathname: string): Promise<void>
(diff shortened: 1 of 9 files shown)


packages/next/src/server/route-matcher-managers/dev-route-matcher-manager.ts
```diff
@@ -70,8 +72,16 @@ export class DevRouteMatcherManager extends DefaultRouteMatcherManager {
...
-      await this.ensurer.ensure(developmentMatch, pathname)
-      await this.production.reload()
+      await getTracer().trace(
+        DevRouteMatcherManagerSpan.ensureRoute,
+        {
+          spanName: 'prepare route',
+        },
+        async () => {
+          await this.ensurer.ensure(developmentMatch, pathname)
+          await this.production.reload()
```
timneutkens: This call should be a separate span

DavidIlie: Split `production.reload()` into its own `reload route matchers` span in the latest commit.

timneutkens (approved):

PR #96456, opened 2026-08-01
Trace App Router rendering and response readiness

DavidIlie:
## Summary

This PR completes the internal App Router render timeline that Request Insights consumes. The spans remain internal or verbose by default and do not expand the public OpenTelemetry allowlist.

- `Trace App Router render preparation`
  - records `prepare app page response` and `initialize app render` around the bounded setup work that happens before the route starts producing RSC or HTML
  - keeps the instrumentation close to the actual phase boundaries instead of deriving durations later in the Request Insights viewer
- `Trace App Router RSC rendering`
  - records RSC response production and the wait for the RSC render task
  - handles fulfilled, rejected, cancelled, and prematurely closed streams with one-shot span completion
- `Trace route matching and compilation phases`
  - adds bounded internal phases for route matching, route preparation, route entrypoint readiness, and development route building
  - keeps reload work separate so matcher refresh time is not incorrectly charged to route ensuring
- `Fix route phase allowlist test types`
  - keeps the allowlist regression typed without weakening the assertion that these new phases are not default-public OTel spans
- `Fix route compilation error propagation`
  - preserves the original compilation failure while ensuring its tracing span is stopped exactly once
- `Trace HTML completion`
  - records the wait between starting HTML production and the stream actually completing
  - covers Node and Web streams without double-stopping on close/error races
- `Trace time to the first response chunk`
  - records response-start readiness through the first committed chunk
  - preserves streaming semantics and does not treat response construction as response delivery
- `Stabilize first response chunk coverage`
  - makes the later stream failure begin only after the response has observed the first write
  - proves first-chunk success cannot be overwritten by a later error without relying on scheduler timing

Together these phases explain cold route preparation, compilation, App Router setup, RSC work, HTML completion, and first-byte readiness without importing the broad mutable BaseServer/Fizz instrumentation explored in earlier prototypes.

## Verification

- focused route-phase ownership, compilation-error, RSC completion, HTML completion, and first-response tests
- Request Insights development suite: 51/51 in Turbopack and 51/51 in Webpack at the final stack tree
- route-preparation development suite: 4/4 in Turbopack and 4/4 in Webpack
- public OTel audit confirms `NextVanillaSpanAllowlist` and `Middleware.execute` are unchanged
- focused first-chunk suite passes 13/13, including the deterministic late-error regression
- exact stack head is included in the final 43-commit linear-history and authorship audit

### DIFF
--- a/packages/next/src/build/templates/app-page-runtime.ts
+++ b/packages/next/src/build/templates/app-page-runtime.ts
@@ -22,7 +22,10 @@ import {
   getRequestMeta,
   setRequestMeta,
 } from '../../server/request-meta' with { 'turbopack-transition': 'next-server-utility' }
-import { BaseServerSpan } from '../../server/lib/trace/constants' with { 'turbopack-transition': 'next-server-utility' }
+import {
+  AppRenderSpan,
+  BaseServerSpan,
+} from '../../server/lib/trace/constants' with { 'turbopack-transition': 'next-server-utility' }
 import { stripFlightHeaders } from '../../server/app-render/strip-flight-headers' with { 'turbopack-transition': 'next-server-utility' }
 import {
   NodeNextRequest,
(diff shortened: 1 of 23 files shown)


packages/next/src/build/templates/app-page-runtime.ts
```diff
@@ -2220,6 +2229,29 @@ export function createAppPageEntrypoint({
...
+      requestMeta?: RequestMeta
+    }
+  ) {
+    const finishPrepareAppPageResponse = createOneShotTracePhase(
+      AppRenderSpan.prepareAppPageResponse,
+      'prepare app page response'
+    )
+
+    try {
+      return await handleRequest(req, res, ctx, finishPrepareAppPageResponse)
```
timneutkens: This looks off. It's passing the function but also adding `finally {`. Seems complicated. What kind of span is this? Can you show what it looks like in the request insights view?

DavidIlie: image

the `finally` stops double-stopping, but yes it makes the code look more complicated than it should be

PR #96482, opened 2026-08-02
Throw when generateStaticParams returns non-object/non-array entries

deepshekhardas:
Rebased and resubmitted version of original PR.


deepshekhardas:
Superseded - upstream merged as #95968 (81643544fca). Validation already in canary.

PR #96483, opened 2026-08-02
Throw when generateStaticParams returns incomplete params or empty arrays with output: export

deepshekhardas:
Rebased and resubmitted version of original PR.


deepshekhardas:
Superseded - upstream merged as #95969 (f2e6abb47eb). Validation already in canary.

PR #96487, opened 2026-08-02
fix: add precedence parameter to insertStylesheets for ReactDOM.preinit

deepshekhardas:
Adds precedence parameter to insertStylesheets for ReactDOM.preinit.

### DIFF
--- a/packages/next/src/client/script.tsx
+++ b/packages/next/src/client/script.tsx
@@ -26,7 +26,10 @@ export interface ScriptProps extends ScriptHTMLAttributes<HTMLScriptElement> {
  */
 export type Props = ScriptProps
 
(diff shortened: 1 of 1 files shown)


packages/next/src/client/script.tsx
```diff
@@ -26,7 +26,10 @@ export interface ScriptProps extends ScriptHTMLAttributes<HTMLScriptElement> {
  */
 export type Props = ScriptProps
 
-const insertStylesheets = (stylesheets: string[]) => {
+const insertStylesheets = (
+  stylesheets: string[],
+  precedence: 'reset' | 'low' | 'medium' | 'high' = 'medium'
```
eps1lon, line 31: Second parameter seems to be unused. Could you add an e2e showing what you're trying to fix?

PR #96489, opened 2026-08-02
fix: ignore stray parent lockfiles in find-root

deepshekhardas:
Ignores stray parent lockfiles in find-root.


deepshekhardas:
Superseded by upstream #96159 (root detection rewrite handles stray parent lockfiles via git boundary/worktree logic).

PR #96505, opened 2026-08-02
Flag newly disabled deploy tests

marcoshernanz:
## What changed

Added scoped review guidance under `test/` so Vercel Agent flags changes that newly disable deploy-mode tests or assertions.

## Why

Deploy coverage should only be disabled for assertions that are inherently self-hosted-specific. When deployed behavior differs, the test should prefer documenting those semantics instead of skipping deploy mode.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm build-all`
- Hosted Vercel Agent E2E: an intentional `skipDeployment: true` probe was flagged inline, then removed
- Hosted Vercel Agent Review passed on the final diff


marcoshernanz:
@vercel run a full code review

marcoshernanz:
@vercel run a full code review

timneutkens (approved):

PR #96506, opened 2026-08-02
turbo-tasks: explicit GC root anchoring + cross-session orphan reclamation

lukesandberg:
### What?

Reworks how the reference-counting GC decides what is a **root** (a task that must not be collected), and adds a durable path to reclaim cross-session orphans.

This PR stacks on #95977 (which added `parent_count`/`transient_ref_count` reference counting and wired GC into the app). It replaces the "every parent-less task is a permanent root" rule with explicit anchoring, and closes the two leaks that rule was papering over:

1. **In-session** — disposable top-level ops (the `initialize`/`update` stubs, per-request source-map ops) no longer leak for the whole session.
2. **Cross-session** — a persisted subtree whose root is never re-requested in a later session (a deleted route, or a root left pinned by the shutdown race) is now aged out and collected instead of leaking on disk forever.

### Why?

The previous strategy flagged **any task created with no parent** as a permanent, persisted `gc_root` that hard-vetoed collection. Two problems fell out of that:

- **Disposable roots never died.** Top-level stub ops (e.g. the `update`/`initialize` container hacks, per-request source-map ops) are read once and dropped, but being created parent-less flagged them uncollectible for the whole session — so each such op is a permanent leak. Being a root should mean "something is actually holding this," not "this happened to be created at the top level."
- **The GC only scans the resident map.** A route deleted between session A and session B — if its subtree is never brought back into memory — had its on-disk reference count decremented by nothing, so it (and everything transitively reachable from it) leaked on disk permanently. The resident-only scan structurally cannot reach that garbage.

### How?

**Anchor roots explicitly instead of by topology.** A task is now kept alive only while a *real* anchor holds it — a live parent (`parent_count > 0`) or an external hold (`transient_ref_count > 0`, i.e. a transient parent or an explicit pin). The blanket `parent_task.is_none()` flagging is removed. A new `GcRoot` RAII guard (pins on construct, unpins on drop, not `Clone`) anchors the one genuinely permanent root — the `ProjectContainer` operation — for the lifetime of the JS-held `ProjectInstance`. This is what lets the disposable stubs collect right after their single read: they're held only by activeness while the read is in flight, and GC only runs at quiescence.

**Persisted roots set + TTL age-out (the cross-session backstop).** A new Infra-keyspace `GcRoots` entry persists `Vec<(TaskId, last_anchored_ms)>`. Every GC pass refreshes the timestamp of each root that is still anchored, keeps the timestamp of roots that aren't, and collects any root un-anchored past a wall-clock TTL (default 24h) — re-validating collectibility under a guard before seeding the cascade, never blind-deleting. So "orphaned for 24h" = "not anchored in any GC pass for 24h"; a branch-switch-back re-anchors a root lazily on demand and resets its clock, while a genuinely-deleted route ages out. This also subsumes the shutdown race for free: a root left pinned at teardown persists, and if the next session never re-anchors it, it ages out on the same path as a deleted route.

**Dispose transient root tasks on drop.** `RootTask::Drop` was a no-op, so a subscription's transient root task leaked whenever JS skipped explicit disposal. `Drop` now disposes as an idempotent backstop, and disposing a clean root tears down its outgoing edges via `CleanupOldEdges` — shedding the `transient_ref_count` its child edges placed on the persistent tasks it read (and rebalancing the aggregation graph) — so the subscription's subgraph becomes collectible. Disposal is a no-op once `stopping`, mirroring `gc_pin`/`gc_unpin`, so a root finalized during Node teardown is safe.

**Remove the now-dead `gc_root` flag** from the task storage schema (the roots map fully replaces its veto role), shifting the persisted flag-bit layout. The on-disk cache is versioned by git-describe, so the layout change is safe across released binaries.

**Fixes found by dogfooding the cross-session path** (all unreachable by the small unit-test subtrees, only surfaced by a real deleted-route restart with GC on):
- `CleanupOldEdges` cell/output-dependency arms asserted the scrub target was resident. But a forward-dependency target is an upstream *producer* the task reads from, not a child; under a cross-session cascade that producer can be disk-only — a live persistent task, not a collected one. Restore it `MustExist` and scrub the stale reverse edge (correct), instead of asserting.
- Seeded-cascade collectibility was a `debug_assert!`, but jobs run concurrently under `scope_unbounded` and a sibling cascade can flip a task non-collectible between spawn and execution. Re-check `is_gc_collectible` under the held guard and skip instead of asserting — self-healing (still-garbage tasks are re-selected next pass).
- The `CollectiblesDependency` arm opened its dependent with `Data` access, but `collectibles_dependents` is a `Meta`-category field, so scrubbing that edge panicked `check_access`. Open it with `Meta`.

Also adds GC tracing (`NEXT_TURBOPACK_TRACING=turbo-tasks`) so the dogfood can confirm a deleted route's subtree is reclaimed in a *later* session, tagging cross-session-root collects distinctly from resident-scan collects.

### Testing

- Backend suites green with `TURBO_ENGINE_GC=1`: `parent_count` (incl. new `gc_collects_cross_session_orphan_root`, `gc_keeps_reanchored_cross_session_root`, `gc_collect_scrubs_disk_only_forward_dep_target`, and the root-disposal tests), `gc_stress`, `gc_resurrection`, `detached`.
- End-to-end dogfood: `next dev` on an example app with `TURBO_ENGINE_GC=1 TURBO_ENGINE_IGNORE_DIRTY=1` — all routes served, HMR works across edits, clean shutdown runs stop-time GC, and a cross-session restart with a route deleted serves the surviving routes, 404s the deleted route, and (with `TURBO_ENGINE_GC_ROOT_TTL_MS=0`) reclaims the deleted route's subtree — with zero panics.

### DIFF
--- a/crates/next-api/src/versioned_content_map.rs
+++ b/crates/next-api/src/versioned_content_map.rs
@@ -5,7 +5,7 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use turbo_rcstr::RcStr;
 use turbo_tasks::{
     FxIndexSet, NonLocalValue, OperationValue, OperationVc, ResolvedVc, State, TryFlatJoinIterExt,
-    TryJoinIterExt, Vc, debug::ValueDebugFormat, trace::TraceRawVcs, turbobail,
+    TryJoinIterExt, Vc, debug::ValueDebugFormat, trace::TraceRawVcs, turbo_tasks, turbobail,
 };
 use turbo_tasks_fs::{FileContent, FileSystemPath};
 use turbopack_core::{
@@ -68,10 +68,18 @@ unsafe impl OperationValue for PathToOutputOperation {}
 type OutputOperationToComputeEntry =
     FxHashMap<OperationVc<ExpandedOutputAssets>, OperationVc<OptionMapEntry>>;
 
-// TODO: Ideally this structure is never persisted, so new sessions start from scratch and don't
-// accumulate entries or force rebuilds of all chunks when a new session is only interested in some
-// of them. If this happens, this should have #[turbo_tasks::value(evict = "never")].
-#[turbo_tasks::value]
+// This map is a dev-only, rebuilt-every-session index over the current session's output assets
+// (populated via `insert_output_assets` during the write/emit phase), not a source of truth — the
+// underlying `ExpandedOutputAssets` operations are the persisted data. So we skip persisting it
+// (`serialization = "skip"`), which keeps the `OperationVc` GC pins it holds (see the pin/unpin
+// calls in the mutation methods below) purely in-session: no persist/restore re-pinning needed. This
+// also means new sessions start from scratch and don't accumulate entries or force rebuilds of all
+// chunks when a new session is only interested in some of them.
+//
+// `evict = "never"` because the op set and the record of what's pinned live in this cell's `State`;
+// evicting the cell would drop that state and (with serialization skipped) it could not be
+// restored, silently resetting the map and orphaning the pins.
+#[turbo_tasks::value(serialization = "skip", evict = "never")]
 pub struct VersionedContentMap {
     // TODO: turn into a bi-directional multimap, ExpandedOutputAssets ->
     // FxIndexSet<FileSystemPath>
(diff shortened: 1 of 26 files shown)


lukesandberg:
Superseded by #96857. The stack was rewritten and the head branch renamed `fewer_routes` -> `gc-05-roots-orphans`; GitHub cannot move a PR's head branch, so this PR is replaced by #96857 which carries the same content, title, and description.

PR #96517, opened 2026-08-03
docs(cache-handlers): explain updateTags durations, fix Redis EX for Infinity

leejpsd:
Two small gaps left on the `cacheHandlers` page after #98039 rewrote it.

**`updateTags()` doesn't say what `durations` means.** The table calls it "optional expiration duration in seconds", but its presence is the signal: without it the call is an immediate invalidation (`updateTag()`, single-argument `revalidateTag()`); with it the tags go stale and entries carrying them may keep serving until `durations.expire` (`revalidateTag(tag, 'max')`). `revalidation-utils.ts` only passes `durations` when a profile resolves, so a handler that treats both cases the same loses stale-while-revalidate for profile-based revalidation.

**The Redis example passes `entry.expire` straight to `EX`.** `EX` needs a positive integer and `expire` can be `Infinity` ("never expire"), so that call fails for such entries. The example now omits `EX` when `expire` isn't finite and rounds up otherwise.

One file, docs only.

---

History: this PR originally corrected the `get()` guidance that told handlers to drop entries past `revalidate`. #98039 fixed that as part of the page rewrite, so this PR was rebased and reduced to the two items above.

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
@@ -88,9 +88,7 @@ get(cacheKey: string, softTags: string[]): Promise<CacheEntry | undefined>
 
 Returns a `CacheEntry` object if found, or `undefined` if not found or expired.
 
-Your `get` method should retrieve the cache entry from storage, check if it has expired based on the `expire` time, and return `undefined` for missing or expired entries.
-
-Return the entry even when it is past its `revalidate` time. `revalidate` marks the entry as stale, not expired: Next.js serves the stale entry and schedules a background refresh. Only `expire` means the entry can no longer be used.
+Your `get` method should retrieve the cache entry from storage and return `undefined` only for missing entries or entries past their `expire` time. Return entries that are only past `revalidate`, since those are stale rather than expired, and returning them lets Next.js serve them while refreshing in the background.
 
 ```js
 const cacheHandler = {
@@ -98,9 +96,7 @@ const cacheHandler = {
     const entry = cache.get(cacheKey)
     if (!entry) return undefined
 
-    // Only `expire` means the entry is unusable. An entry past its
-    // `revalidate` time is stale, not expired: return it so Next.js can
-    // serve it while refreshing in the background.
+    // Check if expired
     const now = Date.now()
     if (now > entry.timestamp + entry.expire * 1000) {
       return undefined
@@ -281,8 +277,7 @@ module.exports = {
       return undefined
     }
 
-    // Only drop the entry once it is past `expire`. Past `revalidate` it is
-    // stale, and returning it lets Next.js serve it while it refreshes.
+    // Check if entry has expired
     const now = Date.now()
     if (now > entry.timestamp + entry.expire * 1000) {
       return undefined
@@ -350,6 +345,17 @@ module.exports = {
     // Deserialize the entry
     const data = JSON.parse(stored)
 
+    // Check tags: explicit tags travel inside the stored entry, while soft
+    // tags (implicit route tags, see [Soft Tags](#soft-tags)) arrive as the
+    // second argument. A tag revalidated after this entry was stored
+    // invalidates the entry.
+    const tags = [...data.tags, ...softTags]
+    if (tags.length > 0) {
+      const stamps = await client.mGet(tags.map((tag) => `tag:${tag}`))
+      const latest = Math.max(...stamps.map((stamp) => Number(stamp) || 0))
+      if (latest > data.timestamp) return undefined
+    }
+
     // Reconstruct the ReadableStream from stored data
     return {
       value: new ReadableStream({
@@ -396,9 +402,9 @@ module.exports = {
         expire: entry.expire,
         revalidate: entry.revalidate,
       }),
-      // Use Redis TTL for automatic expiration. `EX` requires a positive
-      // integer, while `expire` can be fractional or `Infinity` (the
-      // never-expire sentinel), so clamp before passing it through.
+      // `EX` needs a positive integer.
+      // By definition `expire: Infinity` ("never expire")
+      // maps to no TTL at all, so omit `EX` and let Redis persist the key.
       Number.isFinite(entry.expire)
         ? { EX: Math.max(1, Math.ceil(entry.expire)) }
         : {}
@@ -411,14 +417,16 @@ module.exports = {
   },
 
   async getExpiration(tags) {
-    // Return 0 to indicate no tags have been revalidated
-    // Could query Redis for tag expiration timestamps if tracking them
-    return 0
+    // `Infinity` tells Next.js that soft tags are checked in `get()`
+    // instead (see the tag check there), so no lookup is needed here
+    return Infinity
   },
 
-  async updateTags(tags, durations) {
-    // Implement tag-based invalidation if needed
-    // Could iterate over keys with matching tags and delete them
+  async updateTags(tags) {
+    // Record when each tag was revalidated; `get()` compares these
+    // stamps against each entry's `timestamp`
+    const now = String(Date.now())
+    await Promise.all(tags.map((tag) => client.set(`tag:${tag}`, now)))
   },
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
```diff
@@ -88,17 +88,21 @@ get(cacheKey: string, softTags: string[]): Promise<CacheEntry | undefined>
 
 Returns a `CacheEntry` object if found, or `undefined` if not found or expired.
 
-Your `get` method should retrieve the cache entry from storage, check if it has expired based on the `revalidate` time, and return `undefined` for missing or expired entries.
+Your `get` method should retrieve the cache entry from storage, check if it has expired based on the `expire` time, and return `undefined` for missing or expired entries.
+
+Return the entry even when it is past its `revalidate` time. `revalidate` marks the entry as stale, not expired: Next.js serves the stale entry and schedules a background refresh. Only `expire` means the entry can no longer be used.
```
icyJoseph:
```suggestion
Your `get` method should retrieve the cache entry from storage and return `undefined` only for missing entries or entries past their `expire` time. Return entries that are only past `revalidate`, since those are stale rather than expired, and returning them lets Next.js serve them while refreshing in the background.
```

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
```diff
@@ -88,17 +88,21 @@ get(cacheKey: string, softTags: string[]): Promise<CacheEntry | undefined>
...
 ```js
 const cacheHandler = {
   async get(cacheKey, softTags) {
     const entry = cache.get(cacheKey)
     if (!entry) return undefined
 
-    // Check if expired
+    // Only `expire` means the entry is unusable. An entry past its
+    // `revalidate` time is stale, not expired: return it so Next.js can
+    // serve it while refreshing in the background.
```
icyJoseph:
```suggestion
   // Check if expired
```

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
```diff
@@ -277,9 +281,10 @@ module.exports = {
       return undefined
     }
 
-    // Check if entry has expired
+    // Only drop the entry once it is past `expire`. Past `revalidate` it is
+    // stale, and returning it lets Next.js serve it while it refreshes.
```
icyJoseph:
```suggestion
    // Check if entry has expired
```

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
```diff
@@ -391,7 +396,12 @@ module.exports = {
         expire: entry.expire,
         revalidate: entry.revalidate,
       }),
-      { EX: entry.expire } // Use Redis TTL for automatic expiration
+      // Use Redis TTL for automatic expiration. `EX` requires a positive
+      // integer, while `expire` can be fractional or `Infinity` (the
+      // never-expire sentinel), so clamp before passing it through.
```
icyJoseph:
```suggestion
      // `EX` needs a positive integer.
      // By definition `expire: Infinity` ("never expire")
      // maps to no TTL at all, so omit `EX` and let Redis persist the key.
```

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
Svish: Not sure if it's even relevant, but would it be possible to add usage of `durations` to this example? Or to another example, if it's for example only relevant for an in-memory cache handler?

Svish, line 345: Same here, would it be possible to add example usage of `softTags` in whatever example it would make sense to use it?

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
leejpsd: Applied, thanks — reads better as one paragraph.

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
leejpsd: Applied.

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
leejpsd: Applied.

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
leejpsd: Applied. Kept the `Math.ceil` in the code itself since `expire` can also be fractional, comment as suggested.

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
leejpsd: Good call — added it to this example: `get()` now checks the entry's own tags plus the incoming `softTags` against `tag:*` stamps and drops the entry when any tag is newer. To make that coherent I also switched this example's `getExpiration` to the documented `Infinity` sentinel ("soft tags are checked in `get`") and made `updateTags` actually write the stamps — the runtime special-cases `Infinity` and keeps implicit tags unexpired on its side, so the `get`-side check is the single enforcement point in this variant.

leejpsd: Added — the distributed example's markers are now `{ stale, expired }` pairs, mirroring what the built-in handler stores: no `durations` (e.g. `updateTag()`) → hard-expire now; with `durations` (profile-based, e.g. `revalidateTag(tag, 'max')`) → stale now, hard-expire `durations.expire` seconds later (`Infinity` → never). `getExpiration` reports only the `expired` stamp, since that's the one the runtime discards on — the `stale` stamp is a `get()`-side concern, where the built-in handler serves the entry but returns it with `revalidate: -1` to force a background refresh.

Svish, line 439: When would one use this "Distributed Tag Coordination" setup? For the Redis example, with tracking invalidations in Redis and checking of `softTags` in `get()`, my understanding is that Redis is the single source of truth regardless of multiple Next.js instances?

leejpsd: Right — in both patterns Redis stays the single source of truth. The mirror isn't a second truth, it's a per-request cache of it, and the difference is purely where reads pay.

Checking tags inside `get()` (the external-storage example) costs an extra Redis round-trip on every cache read. Perfectly correct, simplest to write, and at moderate traffic it's all you need. But a single request can hit dozens of `'use cache'` reads, plus `getExpiration()` for the implicit tags — with in-`get()` checks that's dozens of extra round-trips per request. The mirror flips that: `refreshTags()` syncs tag state once per request (that's the reason this method exists in the interface at all), and every read in that request checks tags at memory speed. What you give up is a propagation window of at most one request boundary for invalidations issued by other instances.

So: start with in-`get()` checks, reach for the mirror when per-read latency or Redis load starts to matter. Added a paragraph to the section intro saying exactly this, since I think every reader will land on your question sooner or later.

Svish: Cool, that makes sense, thanks!

Svish:
Maybe out of scope for this PR, but these changes and improvements to the documentation on `cacheHandlers`, are any of those also relevant to the docs on the separate `cacheHandler` (no `s`)?

- [Guides: Self-Host, Configuring Caching](https://nextjs.org/docs/app/guides/self-hosting#configuring-caching)
- [Custom Next.js Cache Handler](https://nextjs.org/docs/app/api-reference/config/next-config-js/incrementalCacheHandlerPath)

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
```diff
@@ -438,29 +458,45 @@ module.exports = {
...
+    // `durations` (profile-based, for example `revalidateTag(tag, 'max')`),
+    // entries turn stale now — keep serving them while they refresh in the
+    // background — and only hard-expire `durations.expire` seconds later.
+    const expired =
+      durations === undefined
+        ? now
+        : Number.isFinite(durations.expire)
+          ? now + durations.expire * 1000
+          : 0 // `expire: Infinity` — never hard-expire
+    const marker = JSON.stringify({ stale: now, expired })
```
icyJoseph: This whole thing is scope creeping without a holistic view I think, for example now, the `stale` key, as it is, is not used right? We'd need to expand line 474... and compare timestamp to stale and expire? 

Right now the only other get method that could possible use this, is actually dropping them no? 

I know having a more concrete example is good, but having way too specific examples, might be better for an example, and even that one can go so far only. 

Wydt?

icyJoseph: @leejpsd anything work to say here? — Kind of want to merge this and start trying to get the example ready to merge too

leejpsd: Sorry for the slow reply — this sat far longer than it should have.

You're right, and I'd rather cut it than defend it. The `stale` stamp isn't read anywhere: that example's `get()` is elided, so anyone following it would store a field nothing consumes. It also left the two examples disagreeing — the external-storage `get()` drops on a newer stamp while this one implied serve-stale, without ever showing it.

So the distributed example is back to plain timestamps, and the `durations` semantics moved up into the `updateTags()` reference, where it's two bullets of contract instead of a half-built state machine: absent means invalidate now (`updateTag()`, single-arg `revalidateTag()`), present means stale now and still servable until `expire` passes (`revalidateTag(tag, 'max')`). That's what @Svish was asking for and it belongs there rather than in an example anyway.

Net effect is the page is smaller than my previous revision. Ready to merge from my side, and happy to take another pass if anything still reads as too much.

leejpsd:
> @Svish: Maybe out of scope for this PR, but these changes and improvements to the documentation on `cacheHandlers`, are any of those also relevant to the docs on the separate `cacheHandler` (no `s`)?

Sorry for the late reply. Good question — some of it carries over, but not as a copy-paste.

The singular `cacheHandler` is a different interface: entries are `html` / `rscData` rather than a stream, tags arrive through `x-next-cache-tags` on the entry instead of `updateTags`, and the stale-vs-fresh decision is made above the handler in `IncrementalCache` rather than by what the handler returns. So the `revalidate` / `expire` wording here doesn't transfer directly — it has to be re-derived against that path.

Worth doing, but as its own PR rather than growing this one. I'll take a look at those two pages once this lands.

docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
leejpsd: One more small push after that: dropped the now-unused `durations` parameter from the distributed example's `updateTags`, so both concrete examples take just what they use. Left the signature as-is in the API reference example, since showing the full shape belongs there.

That's everything from me.

leejpsd:
Rebased onto canary. #98039 rewrote this page and covers the `get()` point this PR started from (no age check; dropping at `revalidate` is the built-in handler's policy, not a rule), so I dropped that part and everything that had grown around it.

What's left is the two things the rewritten page still doesn't have:

- `updateTags()`: what the presence of `durations` means. Without it the call is an immediate invalidation (`updateTag()`, single-argument `revalidateTag()`); with it the tags go stale and entries may keep serving until `durations.expire` (`revalidateTag(tag, 'max')`). This matches `revalidation-utils.ts`, which only passes `durations` when a profile resolves.
- Redis example: `EX` needs a positive integer and `expire` can be `Infinity`, so omit `EX` in that case. This is the wording you suggested earlier.

One file, +15/-5. If you'd rather fold these into the example PR or close this, that's fine too.

PR #96525, opened 2026-08-03
[testmode] Fix infinite recursion in testmode passthrough fetch

### LINKED ISSUE #96521: Nextjs test mode issue

https://github.com/melkir/testproxy-h2-repro

### To Reproduce

```bash
git clone https://github.com/melkir/testproxy-h2-repro
cd testproxy-h2-repro
npm install
npm test
```

`experimental.testProxy: true` in `next.config.ts` breaks any outbound `fetch()` call to a server that negotiates HTTP/2 via ALPN — virtually every modern HTTPS server. No third-party library involved; the repro is a single plain `fetch()` call. A plain HTTP (no TLS) request succeeds even with `testProxy: true` — only TLS connections that negotiate `h2` are affected.

With `testProxy: true`, the request fails:

```json
{ "error": "TypeError: fetch failed", "cause": "SocketError: other side closed" }
```

(Sometimes instead: `HTTPParserError: Response does not match the HTTP/1.1 protocol`, with a raw HTTP/2 SETTINGS frame in the response bytes. Both are symptoms of the same mismatch, not two different bugs.)

Traced the mechanism in `@mswjs/interceptors`' compiled `ClientRequest` module (vendored by Next): a global `socket-interceptor` patches `net.Socket.prototype.connect` for every outbound socket in the process (not just `http.ClientRequest` traffic, so `fetch()`/undici is included). For requests with no active test context, it reconstructs the connection as a "passthrough": opens its own real socket using TLS options copied from the original connection attempt (including its ALPN list), then replays the request onto it. That machinery assumes HTTP/1.1 framing end-to-end. When the passthrough socket negotiates `h2`, the interceptor still writes plain HTTP/1.1 text and parses the response with its own vendored HTTP/1.1 parser (`llhttp`, WASM) — confirmed directly by instrumenting the compiled interceptor to log the negotiated ALPN protocol (`h2`) and the raw bytes written to the socket (`GET /zen HTTP/1.1\r\n...`).

**Current:** with `experimental.testProxy: true` set, any `fetch()` call made during rendering to an HTTP/2-capable host fails instead of resolving.

**Expected:** requests made outside a recognized test context should pass through once and resolve normally, regardless of what protocol the upstream negotiates.

```bash
Relevant Packages:
  next: 16.4.0-canary.1
```

Testing

next start (local)

lazerg:
With `experimental.testProxy` on, `@mswjs/interceptors` (switched to socket-level interception in #96059) also catches the passthrough `fetch` that `handleFetch` makes for non-test requests. That request has no `next-test-internal` marker, so the interceptor calls `handleFetch` again, takes the passthrough branch again, and loops until it errors. So any server-side request made outside a test context never resolves during a test run.

This adds the `next-test-internal` marker to the passthrough request, and to the `continue` case that has the same problem. It is the guard #96059 already added for the proxy-protocol request, so the request reaches the real server instead of being intercepted again.

Fixes #96521
Closes https://github.com/vercel/next.js/pull/97257

### DIFF
--- a/packages/next/src/experimental/testmode/fetch.ts
+++ b/packages/next/src/experimental/testmode/fetch.ts
@@ -82,6 +82,18 @@ function buildResponse(proxyResponse: ProxyFetchResponse): Response {
   })
 }
 
+function passthroughFetch(
+  originalFetch: Fetch,
+  request: Request
+): Promise<Response> {
+  // Mark the request so the ClientRequest interception in `httpget.ts` sends it
+  // to the real server instead of intercepting it again. @mswjs/interceptors
+  // intercepts at the TCP level, so an unmarked passthrough would re-enter the
+  // interception listener from within itself, recursing indefinitely.
+  request.headers.set('next-test-internal', '1')
+  return originalFetch(request)
+}
+
 export async function handleFetch(
   originalFetch: Fetch,
   request: Request
@@ -89,7 +101,7 @@ export async function handleFetch(
   const testInfo = getTestReqInfo(request, reader)
   if (!testInfo) {
     // Passthrough non-test requests.
-    return originalFetch(request)
+    return passthroughFetch(originalFetch, request)
   }
 
   const { testData, proxyPort } = testInfo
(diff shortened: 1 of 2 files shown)


lazerg:

eps1lon (approved):
Thank you!

kirponomarenko-hub:
@eps1lon 

Please backport this to 16.3 if possible. We really want to upgrade to 16.3, but can't unless we go with a canary build

mischnic:
-> https://github.com/vercel/next.js/pull/97691

PR #96527, opened 2026-08-03
[ci] Wait for the preview tarballs before running the deploy tests

unstubbable:
The "Test new and changed tests when deployed" jobs install Next.js from the preview tarballs built for the commit under test, but nothing made them wait for those tarballs to exist. Until #96438 the job was ordered behind `test-prod` and the other new-test jobs, which delayed it long enough that the tarballs had usually landed by the time it looked for them. Now that it starts immediately it frequently fails with "Artifacts not found for commit ...".

The ordering cannot be expressed as a `needs:` entry, because the tarballs are published by a different workflow and, for pull requests from branches in this repository, a different event than the `build-and-test` run that consumes them. We therefore add a `wait-for-preview-tarball` job that polls for the tarball instead, and make both deploy test jobs depend on it. Doing the waiting in one cheap job keeps the deploy test matrices off their 16-core runners until there is something for them to install.

The new job is also listed in `tests-pass`. A dependency failure skips the jobs that need it, and a skipped job does not count as a failure, so without that entry a tarball that never arrives would have let the required check pass with the deploy tests never having run.

`test-new-tests-deploy-cache-components` gets the same treatment, dropping the incidental ordering it still inherited from `test-cache-components-prod` and picking up the docs-only guard that `test-new-tests-deploy` already had.

### DIFF
--- a/scripts/test-new-tests.mjs
+++ b/scripts/test-new-tests.mjs
@@ -2,6 +2,10 @@
 import execa from 'execa'
 import yargs from 'yargs'
 import getChangedTests from './get-changed-tests.mjs'
+import {
+  previewTarballUrl,
+  waitForPreviewTarball,
+} from './wait-for-preview-tarball.mjs'
 
 /**
  * Run tests for added/changed tests in the current branch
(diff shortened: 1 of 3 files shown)


scripts/test-new-tests.mjs
```diff
@@ -95,50 +100,20 @@ async function main() {
...
-    try {
-      await fetchWithRetry(nextTestVersion)
-      console.log(`Artifacts verified for commit ${commitSha}`)
-    } catch (error) {
-      console.error(error.message)
-      throw error
-    }
+    // The `wait-for-preview-tarball` job in build_and_test.yml already blocks
+    // this job until the tarballs are published, so this only covers the short
+    // window where the blob is not yet readable from this runner.
```
eps1lon: Is this a regional thing?

unstubbable: I think the clanker overreached here. We should just assert here and not wait again. I've pushed a fix.

eps1lon (approved):

PR #96536, opened 2026-08-03
docs(adapters): document assetsHashes and routing.middlewareMatchers

leejpsd:
While writing an adapter against `16.3.0-canary.107` I noticed the adapter docs are missing two fields that the shipped `NextAdapter` types already have.

**`assetsHashes`** — every `PAGES` / `PAGES_API` / `APP_PAGE` / `APP_ROUTE` / `MIDDLEWARE` output carries it right next to `assets`, and it's declared with a doc comment in `build-complete.d.ts`, but none of the five shapes in Output Types mention it.

**`routing.middlewareMatchers`** — dumping the `routing` object from a real `onBuildComplete` call gives:

```
afterFiles, beforeFiles, beforeMiddleware, dynamicRoutes, fallback,
middlewareMatchers, onMatch, rsc, shouldNormalizeNextData
```

but the docs list eight of those nine, everywhere the interface appears: the Creating an Adapter snippet (which says "The interface is defined as follows"), the API Reference parameter list, and Routing Information. This one feels worth fixing soon — an adapter that does its own request matching from the documented fields alone has no way to decide when middleware should run. My guess for why nobody has hit it: if you pass `routes: routing` wholesale into `resolveRoutes` from `@next/routing`, everything works without ever looking at the field.

The wording I added comes from the doc comments in the shipped types, not my own descriptions. `middlewareMatchers` is inserted where the type puts it (right after `beforeMiddleware`).

For what it's worth: I checked the rest of the section against the same build while I was at it — the `output: 'export'` behavior, the prerender classification fields, `pprChain.headers`, the fallback fields, the immutable-assets flow, and the `@next/routing` params/result — and everything else matched the docs. These two were the only gaps I found.

### DIFF
--- a/docs/01-app/03-api-reference/07-adapters/02-creating-an-adapter.mdx
+++ b/docs/01-app/03-api-reference/07-adapters/02-creating-an-adapter.mdx
@@ -48,6 +48,7 @@ export interface NextAdapter {
   onBuildComplete?: (ctx: {
     routing: {
       beforeMiddleware: Array<Route>
+      middlewareMatchers: Array<Route>
       beforeFiles: Array<Route>
       afterFiles: Array<Route>
       dynamicRoutes: Array<Route>
--- a/docs/01-app/03-api-reference/07-adapters/03-api-reference.mdx
+++ b/docs/01-app/03-api-reference/07-adapters/03-api-reference.mdx
@@ -24,6 +24,7 @@ Called after the build process completes with detailed information about routes 
 
 - `context.routing`: Object containing Next.js routing phases and metadata
   - `routing.beforeMiddleware`: Routes executed before middleware (includes header and redirect handling)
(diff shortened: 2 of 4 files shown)


icyJoseph:
Looks like `assetHashes` is relatively new — but should be documented yeah. Not sure what happened to `middlewareMatchers`. 

You do need to have signed commits though. Add one signing key to Github and rewrite the history of this branch.

leejpsd:
Done — added a signing key and rewrote the branch, the commit shows as verified now.

On `middlewareMatchers`: it's been in the shipped types with its own doc comment (`build-complete.d.ts`, right after `beforeMiddleware`) for as long as I can tell, so it looks like the docs page just never picked it up rather than something getting removed.

icyJoseph (approved):

PR #96542, opened 2026-08-03
fix(next/image): don't hang cold transforms when the requester aborts

### LINKED ISSUE #96538: `next start`: client abort during a cold image transform permanently hangs that `/_next/image` URL for all future requests (coalesced generator never settles)

https://github.com/wine-hound/next-image-optimizer-abort-hang

### To Reproduce

The linked repository is a single blank page + `repro.sh`; `npm install && npm run build && ./repro.sh` reproduces end-to-end. Manually:

1. `next build && next start` with a large source image in `public/` (a ~12 MB JPEG makes the timing window easy to hit locally; we originally hit it with a 484 KB JPEG on a slow 2-core CI runner).
2. Abort a **cold** (not yet disk-cached) transform request mid-flight:
   ```bash
   curl -m 0.02 'http://localhost:3000/_next/image?url=%2Fbig.jpg&w=640&q=75'
   ```
3. Request the **same URL** again with a generous timeout:
   ```bash
   curl -o /dev/null -w '%{http_code} in %{time_total}s\n' -m 15 \
     'http://localhost:3000/_next/image?url=%2Fbig.jpg&w=640&q=75'
   ```

Output of the reproduction script (`next@16.3.0-canary.107`):

```
--- step 1: abort cold transform mid-flight: curl -m 0.02 …&w=640&q=75 ---
    curl exit: 28 (28 = client aborted, expected)
--- step 2: re-request the same URL with a 15 s timeout ---
    000 in 15.006627s
    curl exit: 28 (28 = HANG reproduced)
--- step 3: control, same image at w=828 ---
    200 in 0.268416s
    curl exit: 0 (0 = fine)
--- server.log (note: no error, no warning, nothing) ---
    ▲ Next.js 16.3.0-canary.107
    ✓ Ready in 115ms
--- step 4: restart next start, probe the poisoned URL again ---
    after restart: 200 in 0.166564s
```

**Current:** after one aborted request, every subsequent request for that exact transform URL (`url` + `w` + `q`) hangs indefinitely — for **all** clients — until the server process is restarted. Other widths of the same image keep working. Nothing is logged; there is no timeout and no error anywhere in the chain.

**Expected:** one client aborting its own request should affect only that request. Later requests for the same transform should be served normally (the transform should run again, or the coalesced entry should fail/expire).

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 8192
  Available CPU cores: 8
Binaries:
  Node: 26.5.1
  npm: 11.17.0
  pnpm: 11.5.3
Relevant Packages:
  next: 16.3.0-canary.107  (also reproduced on 15.5.19)
  react: 19.1.0
  react-dom: 19.1.0
Next.js Config:
  output: N/A
```

Also reproduced on Linux (GitHub Actions `ubuntu-latest`, 2-core) with `next@15.5.19` / Node 22.

Image (next/image)

next start (local), self-hosted production

### Additional context

We traced the mechanism; every step below refers to `canary` source (also present in 15.5.x dist):

1. `NextNodeServer.handleNextImageRequest` wraps each transform in `this.imageResponseCache.get(cacheKey, generator)` — [`packages/next/src/server/next-server.ts` ~L1018](https://github.com/vercel/next.js/blob/v16.3.0-canary.107/packages/next/src/server/next-server.ts#L1018).
2. `ResponseCache.get` coalesces concurrent requests per `cacheKey` through `Batcher.batch` ([`packages/next/src/lib/batcher.ts`](https://github.com/vercel/next.js/blob/v16.3.0-canary.107/packages/next/src/lib/batcher.ts#L76-L89)). The pending map entry is only deleted in a `finally` that runs **when the generator settles**.
3. For local images the generator reaches `fetchInternalImage` ([`packages/next/src/server/image-optimizer.ts` ~L966](https://github.com/vercel/next.js/blob/v16.3.0-canary.107/packages/next/src/server/image-optimizer.ts#L966)), which creates request/response mocks wired to the **first requester's real client socket** — `socket: _req.socket` (L984) — and then `await mocked.res.hasStreamed` (L989).
4. `MockedResponse.hasStreamed` ([`packages/next/src/server/lib/mock-request.ts` ~L200](https://github.com/vercel/next.js/blob/v16.3.0-canary.107/packages/next/src/server/lib/mock-request.ts#L200)) settles only on the mock's `finish` / `end` / `error` events.
5. Serving the `public/` file goes through `serveStatic` → the vendored `send` package. `SendStream.prototype.stream` registers `on-finished(res, cleanup)`, and `on-finished` watches **`res.socket`** — which is the real client's socket, per step 3. When the client disconnects mid-stream, the socket emits `close`, `on-finished` declares the mocked response finished, and `send` destroys the file read stream mid-pipe. `end()` is therefore never called on the mock, so no `finish`/`end`/`error` ever fires.
6. `hasStreamed` never settles → the generator never settles → `Batcher.batch`'s `finally` never runs → the unsettled promise stays in the pending map **forever**, and every subsequent request for that cache key awaits it.

Notes:

- Only **cold** transforms are vulnerable: once a transform has succeeded once, later requests are served from the disk cache and never re-enter the batcher generator. (Pre-warming every transform is the workaround we're using.)
- Vercel-hosted deployments are unaffected (different optimizer); this hits self-hosted `next start` only, where a single user on a flaky connection can permanently break an image variant for everyone.
- We found this because it made a Playwright e2e suite order-dependent: an early test navigating away from a page aborted a cold transform on a saturated CI runner, and a later test asserting "every image loads" then timed out on exactly that URL, with empty server logs.

- Previously filed as #96537, which the link-detection action closed as "invalid link" — the reproduction-link section contained a line of prose alongside the URL. The linked repository is public and unchanged.

Possible fix directions (happy to PR whichever you prefer):

- Don't wire the coalesced internal fetch to the first requester's socket. Request coalescing means the shared result must not depend on any single client's connection state; passing `_req.socket` into `createRequestResponseMocks` is what lets one client's disconnect wedge the shared generator.
- And/or make `fetchInternalImage` fail fast when the underlying pipe is abandoned (e.g. also settle `hasStreamed` on the mock being destroyed / the socket closing), so the batcher's `finally` can release the key and the next request retries the transform.

lazerg:
On self-hosted `next start`, a client aborting a cold `/_next/image` request mid-transform permanently hangs that exact transform URL for every later request, with nothing logged, until the server restarts.

The coalesced internal fetch wires its mock response to the aborting client's real socket, so `send`'s `on-finished` treats the response as finished when that socket closes and tears down the file stream without ever ending the mock. `hasStreamed` never settles, the generator never settles, and the batcher never releases the cache key.

This stops passing the requester's socket into the internal request, so the shared transform completes regardless of any single client's connection state.

Fixes #96538

### DIFF
--- a/packages/next/src/server/image-optimizer.ts
+++ b/packages/next/src/server/image-optimizer.ts
@@ -978,10 +978,12 @@ export async function fetchInternalImage(
     // Coerce HEAD to GET to avoid issues with the image optimizer
     const method = !_req.method || _req.method === 'HEAD' ? 'GET' : _req.method
 
+    // Don't wire the requester's socket into the coalesced internal request:
+    // one client aborting would tear down the shared stream and wedge the
+    // cache entry for every other waiter.
     const mocked = createRequestResponseMocks({
       url: href,
       method,
-      socket: _req.socket,
       maximumResponseBody,
     })
 
--- a/test/unit/image-optimizer/fetch-internal-image.test.ts
+++ b/test/unit/image-optimizer/fetch-internal-image.test.ts
@@ -3,7 +3,12 @@ import {
   fetchInternalImage,
   ImageError,
 } from 'next/dist/server/image-optimizer'
+import { serveStatic } from 'next/dist/server/serve-static'
 import type { IncomingMessage, ServerResponse } from 'http'
+import { EventEmitter } from 'events'
+import { promises as fs } from 'fs'
+import { tmpdir } from 'os'
+import { join } from 'path'
 
 describe('fetchInternalImage', () => {
   describe('response size limit', () => {
@@ -147,6 +152,38 @@ describe('fetchInternalImage', () => {
 
       expect(result.buffer).toBeInstanceOf(Buffer)
       expect(result.buffer.length).toBe(maximumResponseBody)
+    })
+  })
+
+  describe('requester disconnect', () => {
+    it('should still complete when the requester socket is already gone', async () => {
+      const size = 4 * 1024 * 1024
+      const filePath = join(tmpdir(), `fetch-internal-image-${Date.now()}.bin`)
+      await fs.writeFile(filePath, Buffer.alloc(size, 1))
+
+      // A disconnected client's socket reports `writable: false`, which is what
+      // `send` reads to decide the response is done and destroy the stream.
+      const socket = Object.assign(new EventEmitter(), { writable: false })
+      const mockReq = { method: 'GET', socket } as unknown as IncomingMessage
+      const mockRes = {} as ServerResponse
+
+      try {
+        const result = await Promise.race([
+          fetchInternalImage(
+            '/test-image.jpg',
+            mockReq,
+            mockRes,
+            50_000_000,
+            (req, res) => serveStatic(req, res, filePath)
+          ),
+          new Promise((resolve) => setTimeout(() => resolve('timeout'), 5_000)),
+        ])
+
+        expect(result).not.toBe('timeout')
+        expect((result as { buffer: Buffer }).buffer.length).toBe(size)
+      } finally {
+        await fs.unlink(filePath).catch(() => {})
+      }
     })
   })
 })


lazerg:

lazerg:

PR #96552, opened 2026-08-03
Preserve the original host on forwarded Server Action requests

### LINKED ISSUE #96344: Forwarded Server Actions expose the internal origin (localhost:PORT) via headers() — breaks host-based multi-tenancy when self-hosting
# Forwarded Server Actions expose the internal origin (`localhost:PORT`) via `headers()` — breaks host-based multi-tenancy when self-hosting

https://github.com/squaredd/next-forwarded-action-host-repro

### To Reproduce

1. Two pages and one shared server action that reads `headers()`:

```js
// app/actions.js
'use server';
import { headers } from 'next/headers';
export async function whoAmI() {
  const h = await headers();
  return {
    host: h.get('host'),
    xForwardedHost: h.get('x-forwarded-host'),
    actionWasForwarded: h.get('x-action-forwarded'),
  };
}
```

- `app/a/page.jsx` renders a client component that imports `whoAmI` → the action is bundled into `app/a/page`'s worker.
- `app/b/page.jsx` does **not** import the action.
- Both pages use `export const dynamic = 'force-dynamic'`.

2. `next build && next start -p 4111`

3. POST the action to each page (only the URL differs):

```bash
AID=$(node -e "const m=require('./.next/server/server-reference-manifest.json');console.log(Object.keys(m.node)[0])")

# control — POST to /a (page IS in the action's worker set)
curl -s http://127.0.0.1:4111/a \
  -H 'Host: repro.example' -H 'Origin: http://repro.example' \
  -H 'Content-Type: text/plain;charset=UTF-8' -H "next-action: $AID" \
  --data-raw '[]'
# → {"host":"repro.example","xForwardedHost":"repro.example","actionWasForwarded":null}

# bug — same POST to /b (page is NOT in the action's worker set)
curl -s http://127.0.0.1:4111/b \
  -H 'Host: repro.example' -H 'Origin: http://repro.example' \
  -H 'Content-Type: text/plain;charset=UTF-8' -H "next-action: $AID" \
  --data-raw '[]'
# → {"host":"localhost:4111","xForwardedHost":"repro.example","actionWasForwarded":"1"}   HTTP 200
```

Deterministic: same result 100% of runs.

**Current:** when a Server Action POST hits a route whose page is not in the action's `workers` set, `handleAction` forwards the request to a worker that has it (`selectWorkerForForwarding`, `packages/next/src/server/app-render/action-handler.ts`). The forward is an HTTP self-fetch to `process.env.__NEXT_PRIVATE_ORIGIN` — which `startServer` sets to `http://localhost:PORT` (`packages/next/src/server/lib/start-server.ts`). Inside the forwarded subrequest, `headers().get('host')` returns **`localhost:PORT`** instead of the original request's host.

**Expected:** the action should observe the original request's `Host` (or the docs should state that `host` is unreliable inside actions and `x-forwarded-host` must be used).

### Why this matters (production impact)

Any self-hosted multi-tenant app that resolves the tenant from `headers().get('host')` (a very common pattern) breaks *only* for forwarded actions, which makes it maddening to debug:

- The forward is invisible: it happens on loopback, the APM shows a POST to the **target worker's literal pathname** (e.g. `/[...slug]`, brackets included) with `Host: localhost:3000` and a duplicated `x-forwarded-for`.
- Which actions get forwarded depends on the build's static import graph vs. where the action is *actually invoked* from at runtime. Shared client components (a header search box, cards reused across routes) and intercepted/parallel routes routinely cause a POST to land on a URL whose page is not in the action's worker set.
- In our production app (9 domains, host-based tenancy) this produces a steady ~1,000 failed actions/day for real users, all `Error: Project not found: localhost:3000`.

Related but distinct: #84504 (forwarding loop, fixed by #93792 — the guard prevents a *second* forward but the first forward still rewrites the Host), #62903, #58295.

### Notes on the mechanism (from reading 16.2.2 sources)

- `createForwardedActionResponse` fetches `${__NEXT_PRIVATE_ORIGIN}${workerPathname}`; `fetch` derives `Host` from the URL, so the original `Host` header is lost (it is a forbidden header and cannot be preserved via `getForwardedHeaders`).
- `x-forwarded-host` *is* preserved and carries the original host — that is also why the CSRF origin check passes inside the subrequest (`parseHostHeader` prefers `x-forwarded-host`). So the information is available; it is `headers().get('host')` that silently changes.
- A possible fix: have the forwarded fetch set the original host into the subrequest (e.g. re-write the `host` header after `getForwardedHeaders`, or make `headers()` in forwarded actions surface the original host), or document `x-forwarded-host` as the canonical way to read the host inside Server Actions.

```
Next.js: 16.2.2 (also inspected canary sources — behavior unchanged)
Node.js: v22.22.2
OS: Linux (self-hosted, `next start` behind reverse proxies; repro is plain `next start`)
```

Server Actions

next start (local), Other (self-hosted production)

EnricoOrt:
### What

Restores the original `Host` header on a Server Action request that Next.js forwarded to another worker, so `headers().get('host')` inside the action reports the host the user requested instead of the server's internal origin.

Fixes #96344

### Why

When an action POST lands on a route whose page doesn't bundle that action, `handleAction` forwards it to a worker that does (`selectWorkerForForwarding` → `createForwardedActionResponse`). The forward is an HTTP self-fetch to the server's own origin, normally `http://localhost:PORT`.

`host` is a forbidden `fetch` header, so it can't be carried onto the subrequest; `fetch` derives `Host` from the URL it is given. The action ends up seeing:

    POST /a  (page bundles the action)   host: "repro.example"    x-forwarded-host: "repro.example"
    POST /b  (page does NOT bundle it)   host: "localhost:4111"   x-forwarded-host: "repro.example"

Reading the tenant from `headers().get('host')` is a common self-hosting pattern, and it breaks only for the subset of actions that happen to get forwarded, which depends on the build's static import graph versus where the action is actually invoked. Shared client components and intercepted or parallel routes routinely land a POST on a route that doesn't bundle the action. The reporter measures about 1,000 failed actions a day across 9 domains, all surfacing as "project not found: localhost:3000".

`x-forwarded-host` does survive the forward and still carries the original host, so the information is recoverable.

### How

`restoreForwardedActionHost` rewrites `host` from `x-forwarded-host`. It runs in `base-server.ts` immediately after `attachRequestMeta`, and the placement is load-bearing in both directions. It has to be after `attachRequestMeta`, because the origin it compares against can come from the `initURL` request meta, and that call is what sets it. It has to be before `i18nProvider.detectDomainLocale`, a few lines further down, which is the first thing in the request lifecycle to read `host`. Doing the rewrite inside `handleAction` would leave the locale lookup reading the internal origin.

The marker header is not authenticated, so it is treated as a hint rather than proof. The rewrite requires all four of:

1. `x-action-forwarded` is exactly `1`, the only value `createForwardedActionResponse` ever sends. A repeated header arrives comma-joined and fails this.
2. The request is a fetch action. `handleAction` only forwards a POST carrying an action id, so no other request shape can have come through the forwarding path.
3. `x-forwarded-host` is present.
4. `host` matches the host of the origin this server forwards to, port included.

The send and receive sides have to agree on that origin, so both now go through `getActionForwardingOrigin(req)`: `__NEXT_PRIVATE_ORIGIN` when it is set, otherwise the origin of `initURL`. That also replaces two identical copies of the resolution block, in `createForwardedActionResponse` and `createRedirectRenderResult`.

The `initURL` fallback needed a correction from the first version of this PR, which claimed `initURL` already carries the right host. It only does under `experimental.trustHostHeader`. When the server is started with a hostname and port, which is the self-hosted case in the issue, `attachRequestMeta` builds `initURL` as `${protocol}://${fetchHostname}:${port}${req.url}` and the bug is still live, so that path now gets the rewrite too. The `trustHostHeader` case is deliberately excluded: there `initURL` comes from the request's own `host` header, which would make guard 4 vacuously true. `base-server` passes `hasConfiguredOrigin: Boolean(this.fetchHostname && this.port)`, mirroring the condition in `attachRequestMeta`, and the helper bails when there is neither a private origin nor a configured one.

On trust. `parseHostHeader` already prefers `x-forwarded-host` over `host` for the action CSRF origin check, so the value itself isn't newly trusted. What is new is that it becomes visible as `headers().get('host')`, and that is the value host-based tenancy reads. So, plainly: the marker is a hint, not authentication. In the normal topology the forwarding origin is loopback and unreachable from outside, so guard 4 can't be satisfied over the network. A proxy that presents an internal `Host` to Next.js while also passing through a client-controlled `x-forwarded-host` would satisfy all four guards. Authenticating the marker with a per-server secret would close that. I left it out because it's a larger change than the issue calls for, and I'd rather have a maintainer opinion on it first.

The obvious alternative doesn't work: `x-action-forwarded` can't just be added to `INTERNAL_HEADERS`. `filterInternalHeaders` runs on every ingress in `router-server.ts`, including the genuine loopback request, so listing it there would strip the real marker.

`ACTION_FORWARDED_HEADER`, `ACTION_FORWARDED_VALUE`, `getForwardedHostValue`, `getActionForwardingOrigin` and `restoreForwardedActionHost` now live in `app-render/action-forwarding.ts`, shared between `base-server.ts` and `action-handler.ts`. `getForwardedHostValue` is the existing `x-forwarded-host` list parsing lifted out of `parseHostHeader` rather than duplicated.

### Tests

Unit, `action-forwarding.test.ts`, 25 cases. Both list forms of `x-forwarded-host`, both origin sources, and the rejection paths: a forged marker arriving at a host that isn't the internal one, marker values `""`, `"true"`, `"0"`, `"1, 1"` and `"yes"`, non fetch-action requests, missing `x-forwarded-host`, a malformed private origin, and the `trustHostHeader` shape where `initURL` is derived from the request.

e2e, `test/e2e/app-dir/action-forward-host`. `/with-action` bundles the action and `/without-action` doesn't, so a POST to the latter goes through the forwarding path. Both arms are asserted, so the test pins the invariant rather than only the symptom, and it checks `x-action-forwarded` to prove forwarding actually ran. A third case sends a forged marker together with a forged `x-forwarded-host` and asserts `host` comes through unchanged.

It uses raw `node:http` instead of `next.fetch`, because undici refuses to set `host`. Same forbidden-header rule that causes the bug.

`skipDeployment: true`. The test needs a local port to connect to, has to send a `Host` different from the one it connected to, and reads the action's output from the running server's stdout. A deploy instance has a remote URL, no local app port, and build logs rather than live server output. The behaviour under test is a loopback self-fetch to the server's own origin, which is inherently self-hosted.

Before and after:

    before   host=localhost:54034   actionForwarded="1"   ✗ Expected "example.test"
    after    host=example.test      actionForwarded="1"   ✓

Green after rebasing onto current canary: the new suite in dev and start for both webpack and Turbopack, `action-forward-loop`, `actions-allowed-origins` including "should error if x-forwarded-host does not match the origin", the new unit suite, and `tsc`.

### Out of scope

`createRedirectRenderResult` does a second self-fetch for app-relative redirect streaming and loses `Host` the same way. It shares the origin resolver now, but not the restoration: it sets `x-action-forwarded` only when the original action was itself forwarded, so it needs its own marker and the same provenance decision as above. Happy to follow up separately.

For an Edge action target, `runEdgeFunction` builds the request URL from `initURL`, so `request.url` still shows the internal origin even though `headers().get('host')` is restored.

### DIFF
--- a/packages/next/src/server/app-render/action-forwarding.test.ts
+++ b/packages/next/src/server/app-render/action-forwarding.test.ts
@@ -1,0 +1,267 @@
+import type { IncomingHttpHeaders } from 'http'
+import type { BaseNextRequest } from '../base-http'
+import { NEXT_REQUEST_META } from '../request-meta'
+import {
+  getActionForwardingOrigin,
+  getForwardedHostValue,
+  restoreForwardedActionHost,
+} from './action-forwarding'
+
+const ACTION_ID = '00' + 'a'.repeat(40)
+const PRIVATE_ORIGIN = 'http://localhost:3000'
+const INTERNAL_HOST = 'localhost:3000'
+const PUBLIC_HOST = 'tenant.example'
+
+function createRequest({
+  headers,
+  method = 'POST',
+  initURL,
+}: {
+  headers?: IncomingHttpHeaders
+  method?: string
+  initURL?: string
+}): BaseNextRequest {
+  return {
+    method,
+    headers: {
+      'next-action': ACTION_ID,
+      'content-type': 'text/plain;charset=UTF-8',
+      ...headers,
+    },
+    [NEXT_REQUEST_META]: { initURL },
+  } as unknown as BaseNextRequest
+}
+
+// The shape `createForwardedActionResponse` produces: it arrives at the origin
+// we forwarded to, and only `x-forwarded-host` still knows the original host.
+function createForwardedRequest(
+  headers?: IncomingHttpHeaders,
+  initURL?: string
+): BaseNextRequest {
+  return createRequest({
+    headers: {
+      host: INTERNAL_HOST,
+      'x-forwarded-host': PUBLIC_HOST,
+      'x-action-forwarded': '1',
+      ...headers,
+    },
+    initURL,
+  })
+}
+
+describe('getForwardedHostValue', () => {
+  it('reads a single value', () => {
+    expect(getForwardedHostValue({ 'x-forwarded-host': PUBLIC_HOST })).toBe(
+      PUBLIC_HOST
+    )
+  })
+
+  it('reads the first value of a comma-separated list', () => {
+    expect(
+      getForwardedHostValue({
+        'x-forwarded-host': `${PUBLIC_HOST}, proxy.example`,
+      })
+    ).toBe(PUBLIC_HOST)
+  })
+
+  it('reads the first value of a repeated header', () => {
+    expect(
+      getForwardedHostValue({
+        'x-forwarded-host': [PUBLIC_HOST, 'proxy.example'],
+      })
+    ).toBe(PUBLIC_HOST)
+  })
+
+  it('returns undefined when the header is missing or empty', () => {
+    expect(getForwardedHostValue({})).toBeUndefined()
+    expect(getForwardedHostValue({ 'x-forwarded-host': [] })).toBeUndefined()
+  })
+})
+
+describe('getActionForwardingOrigin', () => {
+  const originalPrivateOrigin = process.env.__NEXT_PRIVATE_ORIGIN
+
+  afterEach(() => {
+    process.env.__NEXT_PRIVATE_ORIGIN = originalPrivateOrigin
+  })
+
+  it('prefers the private origin', () => {
+    process.env.__NEXT_PRIVATE_ORIGIN = PRIVATE_ORIGIN
+
+    expect(
+      getActionForwardingOrigin(
+        createRequest({ initURL: 'http://127.0.0.1:4000/some/path' })
+      )
+    ).toBe(PRIVATE_ORIGIN)
+  })
+
+  it('falls back to the origin of initURL', () => {
+    delete process.env.__NEXT_PRIVATE_ORIGIN
+
+    expect(
+      getActionForwardingOrigin(
+        createRequest({ initURL: 'http://localhost:3000/some/path?a=b' })
+      )
+    ).toBe(PRIVATE_ORIGIN)
+  })
+
+  it('throws when initURL is missing', () => {
+    delete process.env.__NEXT_PRIVATE_ORIGIN
+
+    expect(() => getActionForwardingOrigin(createRequest({}))).toThrow(
+      'Missing initURL'
+    )
+  })
+
+  it('throws when initURL is not absolute', () => {
+    delete process.env.__NEXT_PRIVATE_ORIGIN
+
+    // What `attachRequestMeta` produces when the server has no configured
+    // hostname and port and does not trust the host header.
+    expect(() =>
+      getActionForwardingOrigin(createRequest({ initURL: '/some/path' }))
+    ).toThrow('Could not determine origin')
+  })
+})
+
+describe('restoreForwardedActionHost', () => {
+  const originalPrivateOrigin = process.env.__NEXT_PRIVATE_ORIGIN
+
+  beforeEach(() => {
+    process.env.__NEXT_PRIVATE_ORIGIN = PRIVATE_ORIGIN
+  })
+
+  afterEach(() => {
+    process.env.__NEXT_PRIVATE_ORIGIN = originalPrivateOrigin
+  })
+
+  function restore(req: BaseNextRequest, hasConfiguredOrigin = true) {
+    restoreForwardedActionHost(req, { hasConfiguredOrigin })
+    return req.headers['host']
+  }
+
+  it('restores the original host on a forwarded action', () => {
+    expect(restore(createForwardedRequest())).toBe(PUBLIC_HOST)
+  })
+
+  it('restores the first value of an x-forwarded-host list', () => {
+    expect(
+      restore(
+        createForwardedRequest({
+          'x-forwarded-host': `${PUBLIC_HOST}, proxy.example`,
+        })
+      )
+    ).toBe(PUBLIC_HOST)
+
+    expect(
+      restore(
+        createForwardedRequest({
+          'x-forwarded-host': [PUBLIC_HOST, 'proxy.example'],
+        })
+      )
+    ).toBe(PUBLIC_HOST)
+  })
+
+  it('restores a host that carries an explicit port', () => {
+    expect(
+      restore(
+        createForwardedRequest({ 'x-forwarded-host': `${PUBLIC_HOST}:8443` })
+      )
+    ).toBe(`${PUBLIC_HOST}:8443`)
+  })
+
+  it('compares the internal origin with its port', () => {
+    // Same hostname, different port: this request did not arrive at the origin
+    // we forward to.
+    expect(restore(createForwardedRequest({ host: 'localhost:3001' }))).toBe(
+      'localhost:3001'
+    )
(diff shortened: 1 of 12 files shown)


DavidIlie:
good start

squaredd:
Reporter of #96344 here — thank you for picking this up, and for the care you put into it. Reporting a self-hosting edge case usually means watching it sit, so seeing a fix with unit tests *and* a dedicated e2e suite show up a few days later is genuinely appreciated. Putting the restore in `base-server.ts` ahead of `detectDomainLocale` rather than inside `handleAction` is also a better call than what we had concluded internally — we'd have patched the action and left locale resolution still reading the internal origin.

We run this in production (self-hosted `output: standalone`, Next 16.2.2, 9 domains on one server that resolves tenant config from `headers().get('host')`), so I can confirm your diagnosis and add two things I don't think are in the thread yet. The ~1,000 failed actions/day you quote from the issue is now ~1,400/day.

**1. The failure is completely silent — that's why this goes unnoticed.**

`createForwardedActionResponse` has two exits when the forwarded subrequest doesn't come back as RSC:

- the `catch` logs `failed to forward action response`;
- the `else` branch (non-RSC `content-type`, i.e. the target worker errored *while rendering*) just cancels the body, **with no log at all**.

Both end at `RenderResult.fromStatic('{}', JSON_CONTENT_TYPE_HEADER)`, so the client gets **HTTP 200 with `{}`**. In our case the target worker throws `project not found: localhost:3000` while resolving the tenant, which means:

- the browser gets a 200 and just renders nothing where the section should be;
- the forwarding worker logs nothing;
- the only trace is the error in the *target* worker's log — we counted 14 of them in 25 minutes in a single pod while every client request returned 200.

None of our monitoring (error rate, status codes, RUM) shows this at all. We only found it by grepping pod logs. So I'd expect the real-world incidence to be much higher than the number of reports.

**2. The code already intends to preserve `host`.**

`host` is not in `actionsForbiddenHeaders`, so `getForwardedHeaders` does copy it into the forwarded `Headers` — undici then drops it per the Fetch spec's forbidden-header rules. So this isn't "Next chose not to forward the host"; it's an existing intent being silently defeated by the platform, which seems like an argument in favour of restoring it rather than treating the internal origin as authoritative.

**On the trust question you raised:** `handleAction` already calls `parseHostHeader(req.headers)`, which prefers `x-forwarded-host` over `host`, and uses that value for the CSRF origin comparison of the very same request. If `x-forwarded-host` were attacker-controlled, that check is already compromised. Restoring `host` from it doesn't appear to widen the trust surface — it just makes both sides agree on one value.

**Repro note, in case it helps the e2e case:** whether a given action forwards is fully deterministic, though it looks random until you read the manifests. It depends on the action's worker-set in `server-reference-manifest.json`: an action referenced by only one or two pages forwards whenever it's invoked from any other page, while an action referenced by many pages never forwards. In our build 5 of 15 actions fell in the first group, and those 5 are exactly the ones that fail in production.

Happy to run a canary build against production traffic if that's useful for validating the fix — we have the failing actions identified, so we can give you a before/after on real traffic rather than a synthetic case. Thanks again for the work on this.

PR #96557, opened 2026-08-03
fix(turbopack): resolve `../` and `/`-rooted `import.meta.glob` patterns

sokra:
### What?

`import.meta.glob()` now resolves `../`-relative and `/`-project-root-absolute
patterns (and `base`), validates every pattern (positive and negative) for
whether it escapes the filesystem root, and shares the underlying
"resolve leading `./`/`../` against a directory" primitive with
`next-core::relativize_glob` (used by `outputFileTracingIncludes`) instead of
duplicating it.

### Why?

`import.meta.glob()` handed its patterns straight to Turbopack's `Glob`, which
matches relative to the scanned directory and understands neither a leading
`/` nor `..` segments. `import.meta.glob('../../content/*.js')` and
`import.meta.glob('/content/*.js')` therefore silently returned `{}`, and a
`base` that walked above the filesystem root failed with an internal error
and a panic report instead of a normal diagnostic. A negative (exclusion)
pattern that escaped the root was silently dropped, which is the worst
outcome for an exclusion — it would silently include files the user asked to
exclude.

### How?

- Patterns are split into the directory they're rooted in and the remaining
  glob, matched relative to a single common scan directory covering every
  positive and negative pattern. `/` is resolved from the project root,
  matching Vite's rule that patterns are relative to the importing file or
  absolute from the project root.
- Map keys follow the pattern too: `base` now produces base-relative keys
  (`'./zeta.js'` instead of `'../../content/zeta.js'`, matching Vite), a
  project-root-absolute pattern produces project-root-absolute keys, and
  origin-relative keys are unchanged otherwise.
- Every pattern is validated up front; one that leaves the project (a `base`,
  a positive pattern, or a negative one) reports an error issue at the call
  site and yields an empty object, instead of crashing or being silently
  dropped.
- The `./`/`../`-resolution loop is extracted into
  `turbo_tasks_fs::glob::relativize_glob`, a crate both `next-core` and
  `turbopack-ecmascript` already depend on. `next-core::relativize_glob` is
  now a thin wrapper over it (same name, signature and error message); its 5
  unit tests are removed since the shared primitive's own tests cover the
  same cases.

Co-authored-by: Niklas Mischkulnig <4586894+mischnic@users.noreply.github.com>


turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -618,64 +697,150 @@ impl ImportMetaGlobAsset {
...
+            // Every root is an ancestor of `base_dir` (or the project root), so
+            // the shortest path is an ancestor of all of them.
+            if root_dir.path.len() < scan_dir.path.len() {
+                scan_dir = root_dir;
+            }
+        }
+        // A negative pattern can't add files, but it can be rooted above the
+        // positive patterns (e.g. `['../dir/*.js', '!/dir/skip.js']`), and it
+        // still has to be expressible relative to `scan_dir`. A negative pattern
+        // that leaves the filesystem can't exclude anything and is ignored.
```
sokra: There should be a error too when this leaves the filesystem

turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-escape-root/issues/import.meta.glob() cannot look outside of the proj-6b599f.txt
```diff
@@ -0,0 +1,21 @@
...
+       9 | it('should return an empty object for a base above the project root', () => {
+      10 |   expect(escapingBase).toEqual({})
+      11 | })
+  
+  the 'base' option "../../../../../../../../../../../.." of import.meta.glob("*.js") resolves to a directory above the project root, relative to turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-escape-root/input. Patterns are relative to the importing file, or absolute from the project root when they start with `/`.
+  
+  Import trace:
+    test:
+      ./turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-escape-root/input/index.js
+      ./turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-escape-root/input/index.js
```
sokra: Why is this duplicated here?

sokra: fixed in https://github.com/vercel/next.js/pull/96560

turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -618,64 +702,151 @@ impl ImportMetaGlobAsset {
...
+        // No pattern may point outside of the project, not even a negative one:
+        // silently ignoring it would include files the user asked to exclude.
+        // Report the pattern as it was written, so it can be found in the source.
+        for pattern in &self.patterns {
+            if pattern_root_dir(strip_negation(pattern)).is_none() {
+                emit_escapes_root_issue(self, &origin_dir, &format!("the pattern {pattern:?}"))
+                    .await?;
+                return Ok(Vc::cell(Default::default()));
+            }
+        }
```
mischnic: Luke added a `relativize_glob` util some time ago for `outputFileTracingIncludes`
You are doing exactly the same operation here.

turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-relative/input/index.js
```diff
@@ -0,0 +1,136 @@
...
+it('should apply negative patterns that walk up out of the importer directory', () => {
+  expect(Object.keys(parentNegative)).toEqual(['../outside/one.js'])
+})
+
+// Absolute from the project root. In this test suite the project root is the
+// repository root.
+const absolute = import.meta.glob(
+  '/turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-relative/outside/*.js',
+  { eager: true }
+)
```
mischnic: This makes me wonder, we do we support this for globs, but not for regular import and require?

sokra: I think we also support that for normal import and require. Fleet do you know?

sokra: Yeah I think that makes sense to implement as separate PR. Do that

turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -618,64 +716,164 @@ impl ImportMetaGlobAsset {
...
+        // No pattern may point outside of the project, not even a negative one:
+        // silently ignoring it would include files the user asked to exclude.
+        // Report the pattern as it was written, so it can be found in the source.
+        for pattern in &self.patterns {
+            if split_pattern(strip_negation(pattern), &base_dir, root.as_ref()).is_none() {
+                emit_escapes_root_issue(self, &origin_dir, &format!("the pattern {pattern:?}"))
+                    .await?;
+                return Ok(Vc::cell(Default::default()));
+            }
+        }
```
mischnic: move this into the `.context("every pattern was checked above")?` below
no need for this additional traversal

turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -618,64 +716,164 @@ impl ImportMetaGlobAsset {
...
-        // Build the positive Glob. Turbopack's Glob operates on paths relative
-        // to the scan directory (no leading `./`), so strip that prefix. For
-        // multiple patterns, use `Glob::alternatives` to combine them.
-        let positive_globs: Vec<Vc<Glob>> = positive_raw
-            .iter()
-            .map(|p| Glob::new(strip_relative_prefix(p).into(), glob_options))
-            .collect();
+        // Rewrite a pattern to be relative to `scan_dir`.
+        let normalize = |pattern: &str| -> Result<NormalizedPattern> {
+            let (pattern_root, rest, root_dir) = split_pattern(pattern, &base_dir, root.as_ref())
```
mischnic: this is the second time we call split_pattern for every single pattern again.

turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -618,64 +716,164 @@ impl ImportMetaGlobAsset {
...
-            Some(neg)
+        // Build the negative Glob (if any). Negative patterns are normalized the
+        // same way and combined into a single alternation glob.
+        let mut negative_globs: Vec<Vc<Glob>> = Vec::with_capacity(negative_raw.len());
+        for pattern in &negative_raw {
+            let normalized = normalize(pattern)?;
+            negative_globs.push(Glob::new(
+                normalized.relative_to_scan_dir.clone(),
+                glob_options,
+            ));
```
mischnic: What about `normalized.root` in this loop? That is just ignored?

turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -433,15 +499,20 @@ impl ImportMetaGlobMap {
         let origin_path = origin.into_trait_ref().await?.origin_path().parent();
 
         // Use read_glob for efficient directory-pruning file discovery.
-        let glob_result = base_dir.read_glob(positive_glob).await?;
+        let glob_result = scan_dir.read_glob(positive_glob).await?;
```
mischnic, line 502: Unrelated to this PR:
Do we really not have the ability to pass a negative glob here? it causes unnecessary invalidations to filter out the negative entries after the fact

turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -622,64 +727,150 @@ impl ImportMetaGlobAsset {
...
+                split_pattern(stripped, &base_dir, root.as_ref())
+            else {
+                if root.is_none() && stripped.starts_with('/') {
+                    emit_absolute_unsupported_issue(self, &format!("the pattern {written:?}"))
+                        .await?;
+                } else {
+                    emit_escapes_root_issue(self, &origin_dir, &format!("the pattern {written:?}"))
+                        .await?;
+                }
+                return Ok(Vc::cell(Default::default()));
```
mischnic: Would it be more incremental to just `continue;` this particular pattern while still processing the rest?
Somewhat similar to the keep_last_successful_parse optimization

turbopack/crates/turbopack-ecmascript/src/references/import_meta_glob.rs
```diff
@@ -622,64 +727,150 @@ impl ImportMetaGlobAsset {
...
+        // that one `read_glob` call covers all of them and each pattern can be
+        // rewritten relative to it. A negative pattern can't add files, but it
+        // can be rooted above the positive patterns (e.g. `['../dir/*.js',
+        // '!/dir/skip.js']`) and still has to be expressible relative to
+        // `scan_dir`, so both are considered here.
+        let mut scan_dir = base_dir.clone();
+        for pattern in &patterns {
+            // Every root is an ancestor of `base_dir` (or of the root a `/`-rooted
+            // pattern resolves from), so the shortest path is an ancestor of all.
+            if pattern.root_dir.path.len() < scan_dir.path.len() {
```
mischnic, line 804: Good that we fixed the symlink normalization already, otherwise this check would be completely broken.

mischnic (approved):

PR #96559, opened 2026-08-03
fix(turbopack): give eager `import.meta.glob` values the ESM namespace

sokra:
Eager glob values were built with a plain CommonJS require, so the value was a
module's `exports` object rather than its ESM namespace — for JSON (or any
CommonJS) module that meant no CommonJS interop ran, so `modules[path].default`
(the shape our own docs show) and `{ import: 'default' }` were `undefined`. A
hand-written `import * as ns from './data.json'` did have `default`.

Eager values now use the same namespace-import primitive a static
`import * as ns from '…'` produces.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs
@@ -141,6 +141,34 @@ impl SinglePatternMapping {
         }
     }
 
+    /// Like [`Self::create_require`], but evaluates to the ESM *namespace* of the
+    /// module instead of its CommonJS `exports` object, so that the CommonJS
+    /// interop applies. This is what a static `import * as ns from "..."`
+    /// produces, and it is the difference between a JSON (or CommonJS) module
+    /// having a `default` export and not having one.
+    pub fn create_esm_require(&self, key_expr: Cow<'_, Expr>) -> Expr {
+        match self {
+            Self::Invalid => self.create_id(key_expr),
+            Self::Unresolvable(request) => throw_module_not_found_expr(request),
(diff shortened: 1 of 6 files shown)


turbopack/crates/turbopack-ecmascript/src/references/pattern_mapping.rs
```diff
@@ -141,6 +141,34 @@ impl SinglePatternMapping {
         }
     }
 
+    /// Like [`Self::create_require`], but evaluates to the ESM *namespace* of the
+    /// module instead of its CommonJS `exports` object, so that the CommonJS
+    /// interop applies. This is what a static `import * as ns from "..."`
+    /// produces, and it is the difference between a JSON (or CommonJS) module
+    /// having a `default` export and not having one.
+    pub fn create_esm_require(&self, key_expr: Cow<'_, Expr>) -> Expr {
```
mischnic, line 149: Isn't this basically the same as `create_import` (below) though?

sokra: create_import returns a Promise. This is sync.

mischnic: there are some differences between the two though that seem wrong:
- create_esm_require for Dropped returns `0`. create_import returns `{}`
- create_esm_require for ModuleLoader uses turbopack_import, while create_import uses turbopack_async_loader

can't we make one of them call the other? like`create_import = wrap_with_promise(create_esm_require)`

sirgagahandino9-design: Go production

mischnic (approved):

PR #96560, opened 2026-08-03
fix(turbopack): name the module in the non-ESM-placeable error, and stop duplicating the importer in traces

sokra:
A build that matched a stylesheet through `import.meta.glob` failed with
"non-ecmascript placeable asset" naming neither CSS nor the stylesheet, and an
import trace listing the page twice. The message now names the offending
module; the duplicate came from the glob's virtual module ident having no
layer, which is now set to the importer's layer so trace simplification can
collapse it (and so the virtual module is distinct per layer).


turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-warnings/issues/__l___TP1008__ import.meta.glob() 'as' option is n-f58c91.txt
```diff
@@ -7,11 +7,4 @@ warning - [analysis] /turbopack/crates/turbopack-tests/tests/execution/turbopack
        3 | 
        4 | // Warning: non-constant eager — should default to lazy (eager: false)
        5 | const nonConstEager = import.meta.glob('./dir2/*.js', {
-       6 |   eager: Math.random() > -1,
-  
-  
-  
-  Import trace:
-    test:
-      ./turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/import-meta-glob-warnings/input/index.js
```
sokra, line 16: This is already in the header of the error. It was repeated 3 times.

mischnic (approved):

PR #96566, opened 2026-08-03
lazy server hmr

wbinnssmith:
## Summary

Make Turbopack server HMR demand-driven. Server updates are now compiled and applied when the next relevant request writes an endpoint, instead of eagerly evaluating changed server modules after every file change.

This replaces the aggregate server HMR subscription with an on-demand update API, while preserving incremental updates and falling back to full cache eviction when a restart is required.

Test Plan: added an e2e test


crates/next-api/src/aggregate_hmr.rs
```diff
@@ -55,11 +50,8 @@ impl Version for AggregateHmrVersion {
             .iter()
             .map(|(path, version)| {
                 let path = path.clone();
-                let version = TraitRef::cell(version.clone());
```
lukesandberg, line 58: nice, i always thought this `cell` call was lame

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -75,36 +67,29 @@ impl Version for AggregateHmrVersion {
...
             .map(|HmrChunkWithContent { path, content }| {
                 let path = path.clone();
                 let content = *content;
                 async move {
-                    let version = content.version().into_trait_ref().await?;
+                    let version = ResolvedVc::try_downcast_type::<ChunkListVersion>(
+                        content.version().to_resolved().await?,
+                    )
+                    .expect("server HMR entry chunks use chunk-list versions")
+                    .await?;
```
lukesandberg: nit: there is a try_downcast on TraitRef, so you could avoid one of these two awaits

`content.version().into_trait_ref().await?.try_downcast::<ChunkListVersion>().expect(...);

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -134,69 +136,100 @@ impl ChunkListUpdateBuilder {
...
+async fn diff_chunks_against(
+    chunks: &[HmrChunkWithContent],
+    from: &AggregateHmrVersion,
+) -> Result<DiffResult> {
     let mut has_new_chunks = false;
     let chunk_updates = chunks
         .iter()
         .filter_map(|HmrChunkWithContent { path, content }| {
-            let Some(prev) = from_aggregate.versions.get(path).cloned() else {
+            turbo_tasks().send_compilation_event(Arc::new(ServerHmrEntryDiffEvent {
```
lukesandberg: debugging code?

crates/next-napi-bindings/src/next_api/endpoint.rs
```diff
@@ -62,11 +63,16 @@ impl From<Option<EndpointOutputPaths>> for NapiWrittenEndpoint {
...
                 client_paths,
             }) => Self {
                 r#type: "nodejs".to_string(),
                 entry_path: Some(server_entry_path.into_owned()),
+                server_hmr_entry_paths: server_hmr_entry_paths
+                    .into_iter()
+                    .map(From::from)
+                    .collect(),
                 client_paths: client_paths.into_iter().map(From::from).collect(),
                 server_paths: server_paths.into_iter().map(From::from).collect(),
```
lukesandberg: From::from is wild, can you call a qualified `from` api.  String::from? perhaps? RcStr::from?

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -1850,124 +1853,126 @@ async fn hmr_update_with_issues_operation(
...
         .final_read_hint()
         .await?;
-    let filter = project.issue_filter().await?;
-    let issues = get_issues(update_op, &filter).await?;
-    let effects = Arc::new(take_effects(update_op).await?);
-    Ok(HmrUpdateWithIssues {
-        update,
-        issues,
+    let effects = Arc::new(take_effects(snapshot_op).await?);
+    Ok(ServerHmrSnapshotWithEffects {
```
lukesandberg, line 1908: no longer capturing/reporting issues here?

crates/next-api/src/aggregate_hmr.rs
wbinnssmith: I added this so we can look for it in an e2e test. For that test, we need to have some way of confirming that certain routes are not recompiled

crates/next-api/src/app.rs
```diff
@@ -2161,14 +2161,26 @@ impl Endpoint for AppEndpoint {
...
+                    let server_entry_path: RcStr = node_root
                         .get_path_to(&*rsc_chunk.path().await?)
                         .context("Node.js chunk entry path must be inside the node root")?
-                        .into(),
-                    server_paths,
-                    client_paths,
-                },
+                        .into();
+                    let hmr_entry_path = server_entry_path
+                        .strip_prefix("server/app/")
```
wbinnssmith, line 2177: this is a bit gross but is cleaned up in the next PR up the stack

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -123,92 +115,202 @@ impl ChunkListUpdateBuilder {
...
-        });
+struct DiffResult {
+    chunk_updates: Vec<ReadRef<Update>>,
+    has_new_chunk_lists: bool,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct ServerHmrChunkListDiffEvent {
+    #[serde(rename = "entryPath")]
```
wbinnssmith, line 176: why was this renamed? seems unnecessary, we should pick one or the other

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -138,77 +128,212 @@ impl ChunkListUpdateBuilder {
...
+        .any(|path| !current_chunk_list_paths.contains(path));
+    let mut has_new_chunk_lists = false;
+    let chunk_updates = chunk_lists
+        .iter()
+        .filter_map(
+            |ServerHmrChunkList {
+                 relative_path,
+                 versioned_content,
+             }| {
+                if *TRACE_DIFFING {
```
lukesandberg, line 224: guard with `debug_assertions` also?

that is on in CI, but not in prod

crates/next-api/src/versioned_content_map.rs
```diff
@@ -109,7 +108,7 @@ impl VersionedContentMap {
     pub async fn hmr_chunks_in_path(
```
lukesandberg: rename to `server_hmr_chunks_in_path`?

lukesandberg (approved):

PR #96570, opened 2026-08-03
Replace WorkStore isStaticGeneration with executionMode

ztanner:
## Summary

- replace `WorkStore.isStaticGeneration` with an explicit `'prerender' | 'request'` execution mode
- preserve the existing derivation and behavior while making the render/prerender distinction explicit
- document that prerendering includes both build-time generation and runtime revalidation
- establish groundwork for moving the render decision higher in the stack and replacing downstream mode checks with purpose-specific state

Stacked on #96564.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -791,7 +791,9 @@ async function generateDynamicRSCPayload(
       // With Cache Components, all routes support it. Without it, only fully
       // static pages do, because their per-segment prefetch responses are
       // generated during static generation (build or ISR).
-      S: workStore.isStaticGeneration || ctx.renderOpts.cacheComponents,
+      S:
+        workStore.executionMode === 'prerender' ||
+        ctx.renderOpts.cacheComponents,
       h: getMetadataVaryParamsAccumulator(),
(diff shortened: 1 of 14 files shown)


packages/next/src/server/app-render/app-render.tsx
```diff
@@ -842,7 +844,10 @@ function createErrorContext(
     // TODO: is this correct if `isPossibleServerAction` is a false positive?
     routeType: ctx.isPossibleServerAction ? 'action' : 'render',
     renderSource,
-    revalidateReason: getRevalidateReason(ctx.workStore),
+    revalidateReason: getRevalidateReason({
+      isOnDemandRevalidate: ctx.workStore.isOnDemandRevalidate,
+      isStaticGeneration: ctx.workStore.executionMode === 'prerender',
+    }),
```
gnoff, line 850: drive by but let's drop these option args for internal functions. just pass what you need. not point in allocating for stuff like this.

gnoff (approved):

packages/next/src/server/app-render/app-render.tsx
ztanner: Sounds good - since it's not new with the refactor I'll pull that out into a separate PR so I don't have to rebase the entire stack.

PR #96576, opened 2026-08-03
Pass explicit render capabilities through App Render

ztanner:
## Summary

`create-component-tree` currently combines `WorkStore.executionMode` with route-level PPR configuration at several leaf call sites to determine whether it can postpone dynamic subtrees and whether the resulting response may be partial.

This introduces `AppRenderCapabilities`, derived once at the App Render boundary, containing:

- `canPostpone`
- `isPossiblyPartialResponse`
- `supportsPerSegmentPrefetching`

Component tree construction consumes the explicit rendering capabilities instead of independently interpreting the PPR configuration. RSC payload generation likewise consumes `supportsPerSegmentPrefetching` instead of re-deriving the client `S` flag from execution mode and Cache Components, keeping the protocol decision consistent across dynamic, initial, and error payloads.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -354,10 +354,26 @@ export type AppSharedContext = {
   clientAssetToken: string
 }
 
+type AppRenderCapabilities = {
+  /** Whether this render may postpone dynamic subtrees. */
+  canPostpone: boolean
+  /**
+   * Whether the response may contain postponed holes. This is conservatively
+   * true for prerenders with PPR enabled, even when the response turns out to
+   * be fully static, because a false negative could cause the client to skip
+   * fetching dynamic data. Per-segment prefetch responses replace this with a
+   * more precise value during segment-data collection.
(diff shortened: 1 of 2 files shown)


packages/next/src/server/app-render/app-render.tsx
```diff
@@ -2779,6 +2787,14 @@ async function renderToHTMLOrFlightImpl(
       isStaticGeneration,
       renderOpts.cacheComponents
     ),
+    // These are distinct rendering and protocol properties even though they
+    // are both determined by route-level PPR support today.
+    renderCapabilities: {
+      canPostpone: isStaticGeneration && isRoutePPREnabled,
+      isPossiblyPartialResponse: isStaticGeneration && isRoutePPREnabled,
```
gnoff, line 2794: Can land for now but really we need to remove legacy PPR code which is dead code. canPostpone is (I think) only used in non-cc paths and if we cleaned that up we'd get to simplify a lot

gnoff (approved):

PR #96433, opened 2026-07-31
[turbopack] Add acknowledged edit transactions for agents

marcoshernanz:
## Summary

Add an experimental, acknowledged edit-transaction protocol for bounded
multi-file agent edits in Turbopack development mode.

When `__NEXT_EXPERIMENTAL_EDIT_TRANSACTIONS=true`, the Next development MCP
server exposes:

- `begin_edit_transaction`
- `renew_edit_transaction`
- `end_edit_transaction`

An agent controller declares the complete source-path set before writing. The
native watcher acknowledges that it has entered the hold, retains filesystem
invalidations while the semantic edit is incomplete, and submits one combined
invalidation batch after the final transaction ends.

This PR is deliberately opt-in. It does not change the default developer
workflow and is not intended as a general debounce.

## Why this matters for v0

v0 already knows which file operations belong to one agent action. Turbopack
currently sees only independently timed filesystem events:

1. the agent schedules a bounded set of file operations;
2. remote stat/write/stat and sandbox work make individual files finish at
   different times;
3. each completed file can start compilation, RSC rendering, HMR, and browser
   refetch work;
4. those intermediate revisions may already be obsolete or temporarily
   inconsistent;
5. the preview can render several times before reaching the revision the agent
   intended.

Filesystem timing cannot recover the missing semantic commit boundary. A short
debounce still splits slower edits, while a long debounce delays small edits
that already form one watcher wave.

## Protocol

```mermaid
sequenceDiagram
    participant Agent as v0 agent controller
    participant MCP as Next dev MCP
    participant Watcher as Turbopack watcher
    participant Graph as Turbopack graph

    Agent->>MCP: begin_edit_transaction(paths)
    MCP->>Watcher: begin(paths, token)
    Watcher-->>MCP: hold acknowledged
    MCP-->>Agent: opaque token + lease

    loop bounded file operations
        Agent->>Watcher: filesystem changes
        Watcher->>Watcher: retain invalidations
        Agent->>MCP: renew_edit_transaction(token)
    end

    Agent->>MCP: end_edit_transaction(token)
    MCP->>Watcher: end(token)
    Watcher->>Graph: one combined invalidation
    Watcher-->>MCP: flush acknowledged
    MCP-->>Agent: final status
```

The acknowledgement before the first write is essential. Without it, an edit
can race ahead of the watcher and recreate the partial-state problem.

## Real-v0 verdict

This is a large improvement for the specific multi-file agent-edit complaint:

[...]

PR #96577, opened 2026-08-03
Turbopack: scale down idle Node.js process pools

marcoshernanz:
## What

Scale each Turbopack Node.js child-process pool back to one warm process after two seconds of inactivity.

- active operations still use the configured concurrency;
- the hottest process stays warm for the next edit;
- a new successfully acquired operation cancels pending cleanup;
- normal, failed, non-reusable, killed, and cancelled operation paths all re-evaluate idle cleanup;
- per-pool worker statistics are decremented for removed idle processes.

## Why

Incremental HMR can expand webpack-loader/PostCSS evaluation pools to machine parallelism, but those expensive Node.js children remain in the idle heap afterward. The existing global scale-down hook runs after a whole-application graph operation and does not reliably cover incremental loader work.

On a large real application this makes the development-server working set history-dependent: identical warmed sessions retained 3-13 loader children, and a 12-file edit burst could add about 3.2 GB to the process tree. This directly contributes to development sessions crossing Next.js's memory threshold and restarting after repeated saves.

## Real-application benchmark

I tested optimized native binaries from the same base commit in the same 16-vCPU/32-GB Vercel environment against a real v0 `/projects` route. The paired matrix covers 1, 2, 4, 8, 12, and 24 distinct files plus 12-file bursts with 0, 5, 20, 50, 100, and 150 ms write gaps. Chromium had to observe the exact final server-rendered revision; the harness also verified RSC requests, renders, errors, source/config restoration, and the loaded native binary hash.

After five seconds idle, across 11 paired scenarios:

| Metric | Baseline | Candidate | Change | Paired bootstrap 95% CI |
|---|---:|---:|---:|---:|
| Process-tree RSS | 7.649 GB | 4.468 GB | **-3.181 GB / -41.6%** | -45.2% to -37.5% |
| Process-tree PSS | 6.850 GB | 4.336 GB | **-2.514 GB / -36.7%** | -40.8% to -31.9% |
| Private memory | 6.803 GB | 4.299 GB | **-2.505 GB / -36.8%** | -40.9% to -32.1% |
| Idle pool children | 12.64 | 1.00 | **-92.1%** | -92.6% to -91.5% |
| Browser-visible update | 1,164 ms | 1,137 ms | -27 ms / -2.3% | -8.2% to +2.9% |

Candidate memory was lower in all 11 pairs. The latency interval crosses zero, so the result is no measurable HMR latency regression. A separate 20-burst endurance run reduced mean RSS from 7.209 GB to 4.358 GB and maximum RSS from 8.461 GB to 5.167 GB. A second real route reduced post-idle RSS from 8.620 GB to 4.031 GB.

This is independent of #96433: that PR coalesces invalidations during acknowledged edit transactions; this PR releases expanded loader processes after work becomes idle and requires no controller protocol.

## Validation

- `cargo test -p turbopack-node --test pool --features process_pool` (8/8)
- `cargo clippy -p turbopack-node --tests --features process_pool -- -D warnings`
- `pnpm --dir packages/next-swc build-native-release`
- `pnpm build-all` (18/18)
- `git diff --check`

The new tests cover ordinary idle shrink/reuse, a non-reusable operation, and ensuring an active operation is never scaled down.

## Scope

This intentionally changes only the process-pool backend. It does not cap active parallelism or scale to zero. The pre-existing global `ChildProcessPool::scale_down()` statistics path is unchanged and can be corrected separately with a broader pool-registry redesign.

### DIFF
--- a/turbopack/crates/turbopack-node/src/process_pool/heap_queue.rs
+++ b/turbopack/crates/turbopack-node/src/process_pool/heap_queue.rs
@@ -46,12 +46,16 @@ impl<T: Ord> HeapQueue<T> {
         Ok(item)
     }
 
-    pub fn reduce_to_one(&self) {
+    pub fn has_multiple(&self) -> bool {
+        self.semaphore.available_permits() > 1
+    }
+
+    pub fn reduce_to_one(&self) -> usize {
         // Drain the semaphore permits
         let mut n = self.semaphore.forget_permits(usize::MAX);
         if n <= 1 {
             self.semaphore.add_permits(n);
-            return;
+            return 0;
         }
         let mut heap: parking_lot::lock_api::MutexGuard<'_, parking_lot::RawMutex, BinaryHeap<T>> =
             self.heap.lock();
@@ -63,6 +67,7 @@ impl<T: Ord> HeapQueue<T> {
         }
         heap.push(top);
         self.semaphore.add_permits(1);
+        n
     }
 
     pub fn reduce_to_zero(self: &Arc<Self>, active_queues: &Mutex<Vec<Arc<Self>>>) {
--- a/turbopack/crates/turbopack-node/src/process_pool/mod.rs
+++ b/turbopack/crates/turbopack-node/src/process_pool/mod.rs
@@ -23,6 +23,7 @@ use tokio::{
     process::{Child, ChildStderr, ChildStdout, Command},
     select,
     sync::Semaphore,
+    task::JoinHandle,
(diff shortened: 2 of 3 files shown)


marcoshernanz:
Closing because the optimized child-process pool backend is being replaced by the worker-thread backend. The measurements were real, but this is no longer the durable place to invest; I am continuing the v0 investigation against the worker-thread/current hot paths and will replace this with a different improvement.

PR #96578, opened 2026-08-03
[turbopack] Support `experimental.serverMinification` & expand `experimental.turbopackMinify`

### LINKED ISSUE #96574: Turbopack builds mangle server class names with no opt-out: experimental.serverMinification silently ignored, breaking MikroORM

https://github.com/andres-dejesus/turbopack-mikroorm-mangling-repro

### To Reproduce

The repro is fully self-contained (in-memory SQLite, no database server): four decorator-based MikroORM 7 entities, and a statically prerendered page that runs ORM discovery at build time (the same shape as a sitemap that queries the DB during `next build`).

1. `npm install`
2. `npm run build` → **fails**
3. `npm run build:unminified` (`experimental.turbopackMinify: false`) → passes, but client bundles ship unminified
4. `npm run build:webpack` (`--webpack`, honors `serverMinification: false`) → passes

**Current:** the Turbopack build compiles, then fails during prerendering:

```
Error [MetadataError]: Duplicate table names are not allowed: reactions
```

MikroORM's decorator metadata registry is keyed by constructor name; Turbopack's build minifier mangles the entity class names, so distinct entities collide in the registry. The exact message varies with entity count and chunk layout — in our production app (24 entities) the same root cause surfaces as:

```
Error [MetadataError]: Multiple property decorators used on 'd.user' property
    at <unknown> (src/entities/UsernameHistory.ts:21:39)
```

(the entity class `UsernameHistory` minified to `d`). Original class names survive only in `.js.map` files. `experimental.serverMinification: false` — which fixes exactly this on the webpack path — is silently ignored by Turbopack (no warning, no effect).

**Expected:** one of:

- `experimental.turbopackMinify` accepts per-environment granularity, e.g. `{ server: false, client: true }` (restoring `serverMinification` parity), or
- a `keep_classnames`/`keepNames`-style passthrough to Turbopack's minifier (SWC supports `keep_classnames`), or
- at minimum, a build warning that `experimental.serverMinification` has no effect under Turbopack.

### Why this matters

Every library that resolves runtime behavior from constructor names breaks the same way under Turbopack production builds: MikroORM, typegoose (#59594 from the webpack era), class-transformer/class-validator patterns, etc. With webpack de-prioritized, this class of app currently has no good Turbopack path: unminified client bundles are the only escape hatch, and server bundle size is a cost nobody was paying for — the server bundle's minification buys nothing that matters while breaking runtime class identity.

```
Operating System:
  Platform: linux
  Arch: x64
  Version: Ubuntu 24.04
  Available memory (MB): 23874
  Available CPU cores: 8
Binaries:
  Node: 24.16.0
  npm: 11.13.0
Relevant Packages:
  next: 16.3.0
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 6.0.3
```

Turbopack

next build (local), Vercel (Deployed)

### Additional context

Related prior art: #59594 (same problem on webpack, solved there via terser `keep_classnames` / `serverMinification: false`), discussion #77800 (notes `serverMinification` no-ops under alternative bundlers), mikro-orm/mikro-orm#6868 (MikroORM maintainer confirms mangling as root cause and recommends `keepClassNames`).

Tested on 16.3.0 stable. The reproduction matrix in the linked repo demonstrates that minification is the only blocker: the identical app builds and runs correctly with `turbopackMinify: false` and with `--webpack`.

sampoder:
Closes https://github.com/vercel/next.js/issues/96574.

Implements:

> experimental.turbopackMinify accepts per-environment granularity, e.g. { server: false, client: true } (restoring serverMinification parity)

I also added the edge runtime in there, even though it isn't too widely used.


crates/next-core/src/next_config.rs
```diff
@@ -2537,6 +2537,15 @@ impl NextConfig {
...
 
+    #[turbo_tasks::function]
+    pub async fn turbo_server_minify(&self, mode: Vc<NextMode>) -> Result<Vc<bool>> {
+        let minify = self.experimental.turbopack_minify;
+        Ok(Vc::cell(
+            minify.unwrap_or(matches!(*mode.await?, NextMode::Build))
+                && self.experimental.server_minification.unwrap_or(true),
+        ))
+    }
+
```
bgw, line 2611: It looks like this makes it possible to turn off server minification, or both client and server minification, but not just client minification.

Can we change the config instead like https://github.com/vercel/next.js/issues/96574 suggests?

> experimental.turbopackMinify accepts per-environment granularity, e.g. { server: false, client: true } (restoring serverMinification parity)

We can also respect `serverMinification`.

sampoder: sure!

bgw (approved):

PR #96580, opened 2026-08-04
Keep loaded env when a forced .env reload reads nothing

### LINKED ISSUE #96547: `@next/env`: `loadEnvConfig` with `forceReload` wipes `process.env` before reading, breaking FIFO-mounted `.env` files

https://github.com/jhb-dev/next-env-fifo-repro

### To Reproduce

1. `git clone https://github.com/jhb-dev/next-env-fifo-repro && cd next-env-fifo-repro`
2. `npm install`
3. `npm run repro`

The script creates a FIFO named `.env` that serves its contents once and empty
thereafter, as a drained pipe does. It then calls `loadEnvConfig` twice: normally,
then with `forceReload: true` — what the dev bundler does on any `.env` watcher
event.

**Current** — the value loads, then the forced reload deletes it:

```
1. before any load:                  SECRET=undefined
2. after initial load:               SECRET="from-secret-manager"
3. after forceReload:                SECRET=undefined
```

**Expected** — a reload that reads nothing should leave the previously-loaded
environment intact.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Tue Apr 21 20:19:12 PDT 2026; root:xnu-11417.140.69.710.16~1/RELEASE_ARM64_T6041
  Available memory (MB): 24576
  Available CPU cores: 14
Binaries:
  Node: 22.19.0
  npm: 10.9.3
  Yarn: N/A
  pnpm: 11.1.2
Relevant Packages:
  next: 16.2.12 // Latest available version is detected (16.2.12).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 6.0.3
Next.js Config:
  output: N/A
```

Runtime

next dev (local)

### Additional context

`packages/next-env/index.ts` resets `process.env` before it knows the re-read will
succeed, and swallows read failures:

```ts
replaceProcessEnv(initialEnv)   // deletes everything not in the pre-load snapshot
...
try {
  const stats = statSync(dotEnvPath)
  if (!stats.isFile() && !stats.isFIFO()) continue
  const contents = readFileSync(dotEnvPath, 'utf8')
  loadedEnvFiles.push({ path: envFile, contents, env: {} })
} catch (err) {
  if (err.code !== 'ENOENT') log.error(...)   // swallowed
}
```

`initialEnv` is snapshotted before the first load, so it holds no `.env` values.
If the re-read yields nothing, the environment is gone.

This is not FIFO-specific — any transient read failure loses the environment the
same way. But FIFOs make it reachable in normal use. #70725 added `isFIFO()` so
mounted `.env` files from secret managers (1Password, Doppler) would load, on the
premise that a pipe "otherwise acts exactly like a file". A pipe can only be
consumed once, and the watcher and reload path were not adapted:

```js
// server/lib/router-utils/setup-dev-bundler.js
if (envFiles.includes(fileName)) { if (fileChanged) { envChange = true } continue }
...
if (envChange) {
  await propagateServerField(opts, 'loadEnvConfig', [{ dev: true, forceReload: true }])
}
```

1Password [documents](https://www.1password.dev/environments/local-env-file) that
FIFO mounts emit filesystem events on open, close and read. So reading the pipe
causes the event that triggers the reload that wipes the environment and cannot
refill it. The app boots fine, then loses its environment on the first watcher
event.

**Suggested fix** — read first, swap only on success:

```ts
if (forceReload && nextLoadedEnvFiles.length === 0 && loadedEnvFiles.length > 0) {
  return { combinedEnv, parsedEnv, loadedEnvFiles }
}
replaceProcessEnv(initialEnv)
```

That alone makes FIFO-mounted `.env` files work: the first read populates, later
reads come back empty, and Next.js keeps what it has.

Reproduced on `@next/env@16.2.12`; both the `isFIFO()` acceptance and the
unconditional `replaceProcessEnv(initialEnv)` are also present on `canary`.
Versions ≤ 14.x are unaffected because FIFOs were skipped entirely.

lazerg:
On the dev server's forced env reload, `loadEnvConfig` calls `replaceProcessEnv(initialEnv)` before it reads the `.env` files, and read failures are swallowed. A FIFO-mounted `.env` (used by secret managers such as 1Password and Doppler) can only be read once, so the first watcher-triggered reload wipes `process.env` and then re-reads nothing, losing those values for the rest of the dev session.

This reads the env files first and, when a forced reload finds no content but a previous load did, keeps the already-loaded environment instead of replacing it.

Fixes #96547

### DIFF
--- a/packages/next-env/index.ts
+++ b/packages/next-env/index.ts
@@ -139,9 +139,6 @@ export function loadEnvConfig(
   if (combinedEnv && !forceReload) {
     return { combinedEnv, parsedEnv, loadedEnvFiles: cachedLoadedEnvFiles }
   }
-  replaceProcessEnv(initialEnv)
-  previousLoadedEnvFiles = cachedLoadedEnvFiles
-  cachedLoadedEnvFiles = []
 
   const isTest = process.env.NODE_ENV === 'test'
   const mode = isTest ? 'test' : dev ? 'development' : 'production'
@@ -155,6 +152,7 @@ export function loadEnvConfig(
     '.env',
   ].filter(Boolean) as string[]
 
+  const newLoadedEnvFiles: LoadedEnvFiles = []
   for (const envFile of dotenvFiles) {
     // only load .env if the user provided has an env config file
     const dotEnvPath = path.join(dir, envFile)
@@ -168,7 +166,7 @@ export function loadEnvConfig(
       }
 
       const contents = fs.readFileSync(dotEnvPath, 'utf8')
-      cachedLoadedEnvFiles.push({
+      newLoadedEnvFiles.push({
         path: envFile,
         contents,
         env: {}, // This will be populated in processEnv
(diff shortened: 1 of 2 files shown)


lazerg:

lazerg:

PR #96601, opened 2026-08-04
[turbopack] Collapse nested promises in the analyzer

### LINKED ISSUE #96595: [Turbopack] Dead-code elimination drops code after `if (x)` when `x` comes from an un-awaited async tail call

https://github.com/waleedlatif1/turbopack-dce-repro

### To Reproduce

1. `npm install`
2. `npm run build`
3. `npm run check` — greps the emitted server chunks for a marker string from each route

```
next 16.3.0

route                          code after the `if`
/api/broken                    deleted
/api/fixed                     kept
/api/control-no-wrapper        kept
/api/control-nonreturning-if   kept
```

Change `next` to `16.2.12` in `package.json`, reinstall, rebuild — every route keeps its code.

`app/api/broken/route.ts`:

```ts
async function findRow(key: string): Promise<Row | null> {
  const [row] = await query({ key })
  return row ?? null
}

// The wrapper returns the promise from `findRow` without awaiting it.
async function findRowWrapped(key: string): Promise<Row | null> {
  return findRow(key)
}

export async function GET() {
  const existing = await findRowWrapped('k')

  if (existing) {
    return NextResponse.json({ marker: 'BROKEN_INSIDE_IF', id: existing.id })
  }

  return NextResponse.json({ marker: 'BROKEN_AFTER_IF' }, { status: 201 })
}
```

Emitted by `next build` on 16.3.0:

```js
async function a(e) { let [t] = await (0, r.query)({ key: e }); return t ?? null }
async function n(e) { return a(e) }
async function s() {
  let e = await n("k")
  return t.NextResponse.json({ marker: "BROKEN_INSIDE_IF", id: e.id })
}
```

The `if` is gone, its body runs unconditionally, and the 201 branch no longer exists in the
bundle. `query()` returns `[]`, so `existing` is `null` and the route throws
`TypeError: null is not an object (evaluating 'e.id')` instead of returning 201.

Expected: the `if` and the code after it are preserved, as they are on 16.2.12. `next dev` is
correct on both versions; only `next build` is affected.

Three ingredients are each necessary — the repro isolates them one per route:

| Route | Difference from `/api/broken` | Code after the `if` |
| --- | --- | --- |
| `/api/broken` | — | **deleted** |
| `/api/fixed` | wrapper says `return await findRow(key)` | kept |
| `/api/control-no-wrapper` | no wrapper; `GET` awaits `findRow` directly | kept |
| `/api/control-nonreturning-if` | `if` body assigns instead of returning | kept |

So it needs an `async` wrapper returning a promise-returning call without `await`, the result
truthiness-tested by the caller, and an `if` body where every path returns. The shape reads as
if the wrapper's return value is modelled as the promise object — always truthy — and that fact
is carried through the caller's `await`, making the `if` look always-taken and everything after
it unreachable.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 22.19.0
  npm: 10.9.3
Relevant Packages:
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

Also reproduced with `output: 'standalone'` on `linux/amd64`.

Turbopack

next build (local), next start (local), Other (Deployed)

### Additional context

This reached production for us on the 16.2.12 -> 16.3.0 bump, with no build error or warning.
Two sites in one app were hit: an API route lost its entire create path and 500'd on every
insert, and a second function collapsed to `async () => await lookup(id)` with its insert
deleted, which failed silently — no exception, no failed request, just rows that were never
written.

Detection is the hard part. We ended up diffing string literals between two images built from
the same commit on the two versions, and the second site was only visible because the deleted
region happened to contain a log message. A dropped branch containing no distinctive string
leaves no trace at all.

Not reproducible with standalone `@swc/core` 1.15.47 using the compress config from
https://github.com/swc-project/swc/issues/11970, which suggests this one is Turbopack-side
rather than the bare minifier.

Same failure mode as https://github.com/vercel/next.js/issues/95257 (optimizer decides a
construct always returns, deletes what follows), different trigger.

sampoder:
`Promise<Promise<T>>` is not a possible type in JavaScript / TypeScript, see this in the TypeScript playground: https://www.typescriptlang.org/play/?#code/IYZwngdgxgBAZgV2gFwJYHsL1RAJgJXQHcAKAShgG8AoGGAJwFNkF6sIEAbTgbmoF9q1APTCYAFQAWjGEXrAADgsb0GzVhBAxk0mAvroAtqhAy4BwzAAGcHAWJXZqHegTIYwIsGc4A5jGcAOmpQSFhEFAwsWzxCIgB1eSVGXHIqWjUWNmxY4nI+QRDwaHgkKDRMHPsEpOVccSJUKEY0mjomLOi7OMTFOvyBIVEYAGVgQ0YAGhgAIzcYRmAoSVn5aBX0OG1dDo0UhYAPfUYQECiArWA9CxNGYNCSiPLzmOqAYUxcH0xgThIEUwAMVQ9BAyAAXLN0OhOIsIBQ2pkNDAAYxgaD3AB+KpxNKQ149WopAaFRhHdD0dxQTBgjxeZz7AC8dO87gJxF6yVSZD4ZIUFKpNPcnlZKQaTRkzJFzhxHKJ9UazQGfIFMGpmmF9OQKQ+eG+EF+MClWtlRF1XwqBr+yHoCEYPKGYgAcuhNaLcNMQOhtiYYJgZExfpwwBcPNcjLdgirKWqhTAIK6AIJapmmzn9HlAA.

A nested promise like `Promise<Promise<T>>` gets folded to `Promise<T>`. 

I believe this bug existed before 16.3 but https://github.com/vercel/next.js/pull/94186 made it obvious. 

This PR adds `resolve_promises` to handle this folding.

Closes https://github.com/vercel/next.js/issues/96595


sokra (approved):
A proper execution test would be great

mischnic:
>I believe this bug existed before 16.3 but https://github.com/vercel/next.js/pull/94186 made it obvious.

how did that make it obvious?

sampoder:
> how did that make it obvious?

well previously I'm not sure if we were using the information that something was a promise but now we began incorrectly folding

mischnic:
>we began incorrectly folding

as tobias said, please add a execution test for that

sampoder:
yep, in the newly amended commit, I added one based on the original issue

mischnic:
we already have `JsValue::normalize` though. shouldn't we use that mechanism?
https://github.com/vercel/next.js/blob/479397867c9d02ef97e03d7a154960408fd55bb1/turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/normalize.rs#L39-L43

sampoder:
> we already have `JsValue::normalize` though. shouldn't we use that mechanism?
> 
> https://github.com/vercel/next.js/blob/479397867c9d02ef97e03d7a154960408fd55bb1/turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/normalize.rs#L39-L43

oh, awesome! switched!

Issue #96063, opened 2026-07-22
Nonce-based CSP: no script gets a nonce on Turbopack + output:'standalone' (16.2.11), breaking hydration


jorel89:
### Link to the code that reproduces this issue

https://github.com/jorel89/next-nonce-turbopack-standalone-repro

### To Reproduce

```bash
git clone https://github.com/jorel89/next-nonce-turbopack-standalone-repro
cd next-nonce-turbopack-standalone-repro
npm install
npx next build
cp -r .next/static .next/standalone/.next/
cd .next/standalone && PORT=3111 node server.js
```

In another shell:

```bash
curl -s -D headers.txt http://localhost:3111/ -o page.html

grep -i content-security-policy headers.txt   # CSP contains a nonce
grep -c '<script'            page.html        # 12
grep -c '<script[^>]*nonce=' page.html        # 0
```

The whole app is 8 files / 68 lines. `src/proxy.js` follows the documented CSP
pattern: it generates a nonce, sets **both** `x-nonce` and a
`content-security-policy` **request** header through
`NextResponse.next({ request: { headers } })`, and mirrors the CSP on the response.

### Current vs. Expected behavior

**Current:** none of the `<script>` tags Next generates carry a `nonce` attribute.

Response header actually sent:

```
content-security-policy: default-src 'self'; script-src 'self' 'nonce-9I6l6Ik+LF00RS/p+TfpWA==' 'strict-dynamic'; style-src 'self' 'unsafe-inline'
```

Scripts actually rendered — **12 total, 0 with a nonce** (5 inline, 7 external):

```html
<script src="/_next/static/chunks/turbopack-1c556wmd80d39.js" async=""></script>
<script src="/_next/static/chunks/1jq4o6yq14o4c.js" id="_R_" async=""></script>
<script>self.__next_f.push([1,"1:\"$Sreact.fragment\"\n2:I[39756,[\"/_next/static...
```

The nonce value from the header appears **zero times** anywhere in the HTML, so it
is not propagated at all — this is not a partial/one-chunk gap.

Because the CSP carries a nonce, browsers must ignore `'unsafe-inline'`
([CSP3 §allow-all-inline](https://www.w3.org/TR/CSP3/#allow-all-inline)), so every
one of those 12 scripts is blocked. The page renders but **never hydrates**: the
`Counter` client component in the repro stays dead on click, and in a real app
every interactive element (login forms in particular) silently stops working.

[...]

PR #96608, opened 2026-08-04
fix(app-render): add nonce to loading/template/error segment script tags

### LINKED ISSUE #96605: App Router: loading/template/error segment chunks are emitted without `nonce` (createComponentStylesAndScripts), breaking nonce + 'strict-dynamic' CSP

I could not distill a small public reproduction, and I want to be upfront about why rather than link something that doesn't actually reproduce. The bug is in a single call site and is trivially verifiable by reading the source (below) or by a one-line `grep` on any installed copy (next section) — whether it *surfaces* at runtime depends on the app being large enough that Turbopack emits per-file chunks in dev. On a fresh app Turbopack merges the whole app into one chunk, so the affected code path never produces an element.

I tried 6 variants on a clean `next@16.3.0` app, all of which produced 0 violations: bare root `loading.tsx`; `loading.tsx` importing a server module used only by it; importing a client component used only by it; padding the graph with 61 client modules; nested route (`app/collections/[slug]/page.tsx`) with a root `loading.tsx`; and `turbopack.root` pointed above the app directory to mimic a monorepo.

Happy to put together a larger reproduction repo if the source comparison and `grep` below aren't enough.

### To Reproduce

`createComponentStylesAndScripts` creates the `<script src>` tags for the `template` / `error` / `loading` segments **without a `nonce`**, while its sibling `getLayerAssets` — which does the same job for the `layout` / `page` segments — passes `nonce: ctx.nonce`.

`packages/next/src/server/app-render/create-component-styles-and-scripts.tsx` (canary):

```js
const scripts = jsHrefs
  ? jsHrefs.map((href, scriptIndex) =>
      createElement('script', {
        src: `${ctx.assetPrefix}/_next/${encodeURIPath(href)}${getAssetQueryString(ctx, true)}`,
        async: true,
        key: `script-${scriptIndex}`,
        //  ^ no nonce
      })
    )
  : null
```

`packages/next/src/server/app-render/get-layer-assets.tsx` (canary), same construct:

```js
createElement('script', {
  src: fullSrc,
  async: true,
  key: `script-${scriptIndex}`,
  nonce: ctx.nonce,   // <-- present here
})
```

**Ten-second check on any installed copy.** The compiled runtime is what actually executes — editing `dist/server/...` has no effect, the code runs from `dist/compiled/next-server/*.runtime.*.js`:

```bash
# one hit: the missing-nonce site
grep -o 'key:`script-${[a-z]*}`}' \
  node_modules/next/dist/compiled/next-server/app-page-turbo.runtime.dev.js

# one hit: the sibling that does it correctly
grep -o 'key:`script-${[a-z]*}`,nonce:' \
  node_modules/next/dist/compiled/next-server/app-page-turbo.runtime.dev.js
```

Both hit on 16.2.7 and on 16.3.0 (`.prod.js` too), and the canary source above is unchanged.

**When it surfaces at runtime**, it needs the segment to own a chunk that isn't also registered with a nonce somewhere else:

- If the segment's chunks are already in `injectedJS`, nothing is emitted.
- If the chunk belongs to a client reference, React's Fizz hoistable-script dedup by `src` keeps whichever registration lands first — and the nonce-carrying `preinit` from the flight client usually wins, masking it.
- It reliably fires when the segment owns a chunk that is *not* a client reference — which happens once the app is large enough for Turbopack to emit per-file chunks in dev.

**Current.** A production App Router app (Next 16.2.7, Turbopack, nonce + `'strict-dynamic'` CSP generated in `middleware`/`proxy` and mirrored on both the request and the response headers, exactly as the CSP guide describes) serves this on every public route in dev:

```html
<script src="/_next/static/chunks/…_app_collections_%5Bslug%5D_page_tsx_….js" async nonce="NmEwZTllODQt…"></script>
<script src="/_next/static/chunks/…_app_not-found_tsx_….js"                  async nonce="NmEwZTllODQt…"></script>
<script src="/_next/static/chunks/…_app_loading_tsx_….js"                    async></script>   <!-- blocked -->
```

Browser console:

```
Loading the script '…_app_loading_tsx_….js' violates the following Content Security
Policy directive: "script-src 'self' 'nonce-…' 'strict-dynamic' 'unsafe-eval'".
Note that 'strict-dynamic' is present, so host-based allowlisting is disabled.
```

58 of the 62 `<script>` tags on that page carry the nonce; the only executable one that does not is the `loading` segment's chunk (the other 3 are `application/ld+json` data blocks, which aren't subject to `script-src`). So nonce propagation works everywhere except this one call site.

**Expected.** The tag carries the nonce, like every other framework-generated script tag.

Adding `nonce: ctx.nonce` to the call above — applied via `pnpm patch`, no other change — brings that page to 62/62, and a production build of the same app stays clean across every route I checked.

**Production is affected too, it's just usually masked.** I instrumented the compiled prod runtime to confirm: in a production build the buggy path *does* run and *does* produce a nonce-less element for `loading.tsx` / `error.tsx` / `not-found.tsx`; React's dedup is what hides it. That makes a production app one chunk-splitting change away from the same breakage, with a broken loading skeleton or error boundary shown to real users.

```
Next.js: 16.2.7 (also verified on 16.3.0, and against current canary source)
Bundler: Turbopack (dev and build)
React: 19.2.7
Platform: darwin arm64
Node: 22
```

Loading UI and Streaming, Turbopack

### Additional context

Suggested fix:

```diff
   createElement('script', {
     src: `${ctx.assetPrefix}/_next/${encodeURIPath(href)}${getAssetQueryString(ctx, true)}`,
     async: true,
     key: `script-${scriptIndex}`,
+    nonce: ctx.nonce,
   })
```

The same call site also covers `template.tsx` and `error.tsx`; `not-found` / `forbidden` / `unauthorized` discard the returned scripts at the destructuring site, so they aren't affected in practice.

This is distinct from #96063 (`output: 'standalone'`, where *no* script gets a nonce). Here nonce propagation works everywhere except this one call.

Henrynncc:
### What?

Add `nonce: ctx.nonce` to the `<script>` tags `createComponentStylesAndScripts` builds for the `template` / `error` / `loading` segments.

### Why?

Its sibling `get-layer-assets.tsx` — which builds the same kind of tag for the `layout` / `page` segments — already passes `nonce: ctx.nonce`:

```js
createElement('script', {
  src: fullSrc,
  async: true,
  key: `script-${scriptIndex}`,
  nonce: ctx.nonce,   // <-- present here
})
```

`create-component-styles-and-scripts.tsx` does not. One has it, the other doesn't, and both are reached from the same tree walk in `create-component-tree.tsx` — this reads as an omission rather than an intentional difference.

Under a nonce-based CSP with `'strict-dynamic'`, host-source allowlisting is disabled, so a parser-inserted `<script src>` with no nonce is blocked outright:

```
Loading the script '…_app_loading_tsx_….js' violates the following Content Security
Policy directive: "script-src 'self' 'nonce-…' 'strict-dynamic' 'unsafe-eval'".
Note that 'strict-dynamic' is present, so host-based allowlisting is disabled.
```

On the app where I hit this (Next 16.2.7, Turbopack, nonce + `'strict-dynamic'` generated in `proxy`/`middleware` and mirrored on both the request and response headers, per the CSP guide), 58 of the 62 `<script>` tags on the page carry the nonce; the only executable one that does not is the `loading` segment's chunk. The other 3 are `application/ld+json` data blocks, which aren't subject to `script-src`.

With this change applied locally (via `pnpm patch` on the compiled runtime), the same page goes to 62/62 with no other change, and a production build of the same app stays clean on every route I checked.

### Why it is rarely seen

Whether the missing nonce actually surfaces depends on whether that chunk *also* gets registered somewhere else with a nonce:

- If the segment's chunks are already in `injectedJS`, nothing is emitted at all.
- If the chunk belongs to a client reference, React's Fizz hoistable-script dedup by `src` keeps whichever registration lands first — and the nonce-carrying `preinit` from the flight client usually wins, masking it.
- It reliably fires when the segment owns a chunk that is **not** a client reference — e.g. the per-file chunk Turbopack emits in dev for a `loading.tsx` that is itself a Server Component.

That last condition is why I could not distill a small reproduction: on a fresh app Turbopack merges the whole app into one chunk and the code path never produces an element. I tried 7 variants on a clean `next@16.3.0` app (bare `loading.tsx`; loading-only server module; loading-only client component; padding the graph to ~1 MB of client modules; nested route with root `loading.tsx`; `turbopack.root` above the app dir) — all 0 violations.

Production is affected by the same code path, it's just usually masked. I instrumented the compiled prod runtime to check: in a production build the path *does* run and *does* produce a nonce-less element for `loading.tsx` / `error.tsx` / `not-found.tsx`; React's dedup is what hides it. That leaves a production app one chunk-splitting change away from the same breakage — a broken loading skeleton or error boundary shown to real users.

### How

One line. No behavior change when no CSP nonce is present: `ctx.nonce` is `undefined` there, and React omits the attribute — which is exactly what `get-layer-assets.tsx` already relies on.

The same call site also serves `template.tsx` and `error.tsx`. `not-found` / `forbidden` / `unauthorized` discard the returned scripts at the destructuring site, so they are unaffected in practice.

### Verifying on an installed copy

The compiled runtime is what executes (editing `dist/server/...` has no effect):

```bash
# before this change — one hit: the missing-nonce site
grep -o 'key:`script-${[a-z]*}`}' \
  node_modules/next/dist/compiled/next-server/app-page-turbo.runtime.dev.js

# the sibling that already does it correctly
grep -o 'key:`script-${[a-z]*}`,nonce:' \
  node_modules/next/dist/compiled/next-server/app-page-turbo.runtime.dev.js
```

Both hit on 16.2.7 and 16.3.0 (`.prod.js` too).

I did not add a test: the failure needs a module graph large enough for Turbopack to give the segment its own non-client-reference chunk, which I could not construct in a fixture. Happy to add one if you can point me at the right harness for asserting rendered `<script>` attributes for a segment.

Closes #96605

### DIFF
--- a/packages/next/src/server/app-render/create-component-styles-and-scripts.tsx
+++ b/packages/next/src/server/app-render/create-component-styles-and-scripts.tsx
@@ -37,6 +37,7 @@ export async function createComponentStylesAndScripts({
         src: `${ctx.assetPrefix}/_next/${encodeURIPath(href)}${getAssetQueryString(ctx, true)}`,
         async: true,
         key: `script-${scriptIndex}`,
+        nonce: ctx.nonce,
       })
     )
     scriptIndex++


eolhc:
Hi, I can't comment on the original issue #96605 but I just want to confirm this on 16.3.2 and 16.3.3 (I see you tested 16.2.7 and 16.3.0), from a large production App Router app.

We're currently blocked from upgrading.

## Setup
Turbopack, a nonce + 'strict-dynamic' CSP generated in proxy.ts exactly as the CSP guide describes, output: 'standalone', and every route dynamically rendered (our prerender manifest only contains the built-in /_global-error page), so this isn't #96063.

## What we are experiencing 
Everything in the PR description above matches what I'm experiencing but here are some additions from our side:
- Our trigger is a root template.tsx rather than loading.tsx, so it isn't specific to the loading segment.
- The blocked chunk belongs to a component nested inside a client component (the header's mobile nav). Notably the header is a client reference and its chunk list does include that chunk, so a nonce-carrying registration should exist but the tag that reaches the browser is still the nonce-less one. So the flight-client preinit doesn't reliably win the dedup, at least not at our scale

In dev, the scripts belonging to that template are the only ones on the page served without a nonce. Everything else (framework scripts, page bundles, inline flight data) is nonced correctly. On the home page that's 58 <script> tags: 41 nonced and 17 not, and 16 of those 17 are real scripts that get blocked (the other is a JSON-LD data block, which script-src doesn't apply to). As a result a bunch of components do not hydrate, so local development doesn't work.

Applying the patch in this PR locally takes that from 16 blocked scripts to 0!! (Thanks @Henrynncc !) But we would rather upgrade normally rather than apply a local patch.

## Reproduction issues

I couldn't build a small repro either. I tried, but I think this would need a massive app. The closest I got was a 150 component app with five big vendor packages (date-fns, lodash-es, micromark, @faker-js/faker, @floating-ui/react), which did split out per package and took the template's list from 3 files to 7, but all 7 still came out nonced. The repro app code always ended up in one shared file no matter how many components I added.

## The scale at which it does show up

We have ~4,000 source files, 206 component directories, 131 feature modules, 123 routes, and ~1,750 chunk files per dev session. The main template file owns 30 files while the page owns 14, and the 16 in that gap are exactly the ones served without a nonce.

PR #96621, opened 2026-08-04
fix(next/image): preserve image response after optimization

### LINKED ISSUE #96612: ImageResponse fails after an uncached next/image optimization

https://github.com/ceolinwill/next-image-response-repro

### To Reproduce

1. Clone the reproduction and start a production server:

```sh
pnpm install
pnpm build
pnpm start
```

2. In another terminal, request the `ImageResponse` route:

```sh
curl -o /dev/null -w 'OG before: %{http_code}\n' http://localhost:3000/og
```

This returns `200`.

3. Perform one uncached image optimization:

```sh
image_id="$(node -p 'crypto.randomUUID()')"
curl -D - -o /dev/null "http://localhost:3000/_next/image?url=%2Fpixel%2F${image_id}&w=640&q=75"
```

This returns `200` with `X-Nextjs-Cache: MISS`

4. Request the same `ImageResponse` route again:

```sh
curl -v -o /dev/null http://localhost:3000/og
```

The server closes the connection without returning a response.

On Next.js 16.3.0, the `/og` route initially returns a valid PNG with status `200`.

After the same server process handles one cache-miss request through the Next.js image optimizer, subsequent calls to `/og` fail with an empty response. The server logs:

```text
Error: failed to pipe response
[cause]: Error: Input buffer contains unsupported image format
```

I expected `ImageResponse` to continue returning the same valid PNG regardless of whether the server previously handled an image optimization request.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 16
Binaries:
  Node: 24.12.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 11.16.0
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: N/A
Next.js Config:
  output: N/A
```

Image (next/image), Metadata, Runtime

next start (local)

### Additional context

The failure is order-dependent and appears to affect the entire server process:

- Opening `/og` directly works before the image optimizer is invoked.
- Only an optimizer cache miss triggers the failure. A cached request with `X-Nextjs-Cache: HIT` does not reproduce it.
- Restarting the server resets the behavior until another uncached optimization occurs.
- The reproduction does not import `sharp` directly and has no custom Next.js configuration.

I also tested `next@canary` on August 4, 2026 and the issue does not reproduce there: both `/og` requests return `200`. The regression is present in stable `16.3.0`.

### Possible regression source

This appears related to #96301.

That PR added a process-wide Sharp loader policy that first blocks `VipsForeignLoad` and then unblocks the HEIF, JPEG, GIF, PNG, TIFF, and WebP loaders. It does not unblock `VipsForeignLoadSvg`.

Node.js `ImageResponse` generates SVG internally and then passes that SVG to the same Sharp module to produce the PNG. After an uncached `next/image` request initializes Sharp with the new raster-only allowlist, `ImageResponse` can no longer load its generated SVG.

This change is present in Next.js 16.3.0 but absent from 16.3.0-preview.10, where the reproduction continues returning `200`.

I also reproduced the mechanism directly with the bundled Sharp version:

- SVG rendering works before applying the loader policy.
- Applying the allowlist from #96301 causes `Input buffer contains unsupported image format`.
- Unblocking `VipsForeignLoadSvg` restores SVG rendering.

ceolinwill:
An uncached next/image optimization globally blocked Sharp’s SVG loader, causing subsequent Node.js ImageResponse rendering to fail

This change restores the SVG loader required for ImageResponse and adds coverage for the exact same-process sequence.

This issue was introduced by #96301 

Fixes #96612

### DIFF
--- a/packages/next/src/server/image-optimizer.ts
+++ b/packages/next/src/server/image-optimizer.ts
@@ -99,6 +99,7 @@ export function getSharp(
         'VipsForeignLoadJpeg',
         'VipsForeignLoadNsgif',
         'VipsForeignLoadPng',
+        'VipsForeignLoadSvg',
         'VipsForeignLoadTiff',
         'VipsForeignLoadWebp',
       ],
--- a/test/e2e/og-api/app/app/pixel/[id]/route.js
+++ b/test/e2e/og-api/app/app/pixel/[id]/route.js
@@ -1,0 +1,12 @@
+const pixel = Buffer.from(
+  'iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNk+M/wHwAF/gL+X8mN6QAAAABJRU5ErkJggg==',
+  'base64'
+)
+
+export async function GET() {
+  return new Response(pixel, {
+    headers: {
+      'Content-Type': 'image/png',
+    },
+  })
+}
--- a/test/e2e/og-api/index.test.ts
+++ b/test/e2e/og-api/index.test.ts
@@ -1,5 +1,6 @@
 import { FileRef, nextTestSetup } from 'e2e-utils'
 import { fetchViaHTTP, renderViaHTTP } from 'next-test-utils'
+import { randomUUID } from 'crypto'
 import fs from 'fs-extra'
 import { join } from 'path'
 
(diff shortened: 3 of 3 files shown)


styfle (approved):
Thanks for the PR and test! 🎉 

I filed an issue with sharp since this makes the blocklist nearly unusable for real applications:

- https://github.com/lovell/sharp/issues/4582

In an ideal world we could have a separate blocklist for next/image vs next/og but we'll see what they say.

In the meantime, this is a great solution, thanks!

lukesandberg:
Heads up on the CI: the failing "when deployed" job needs Vercel deployment secrets, which GitHub deliberately withholds from pull requests opened from forks. It can't pass here no matter how many times it's retriggered — nothing wrong with your change. (`thank you, next` is just the aggregate gate reporting that one failure.)

So I've re-opened this as a branch PR where that job can actually run: #96681. Your commit is cherry-picked unchanged and still authored by you, so the credit stays yours — GitHub attributes it to your account on the new PR.

I verified the fix locally and it holds up well. Confirmed the test genuinely catches the regression: without the one-line change it fails with `socket hang up`, caused by `Input buffer contains unsupported image format`. Worth noting the symptom is more severe than a broken image — it crashes the response, so any `ImageResponse` route hit after an uncached optimization in the same process goes down. Nice catch on the same-process ordering in the test, and on using a random URL so the optimization can't be served from cache.

I also checked the change doesn't loosen the SVG protections, since #96301 had deliberately tightened that allowlist: `dangerouslyAllowSVG` throws a 400 before Sharp is invoked, and `dangerously-allow-svg.test.ts` passes 94/94.

I'll leave this open for you to close in favor of #96681.

PR #96622, opened 2026-08-04
Add cache-handler-redis-turbo example (Next.js 16 Cache Components + Redis)

JannikZed:
**What**:

New example examples/cache-handler-redis-turbo demonstrating Redis as a shared cache with `@trieb.work/nextjs-turbo-redis-cache`
Shows both cacheHandler (singular, Pages Router ISR) and cacheHandlers (plural, 'use cache' / cacheComponents: true) — the gap the existing cache-handler-redis example doesn't cover (neshca is pinned to next < 15)
Includes /cet + /gmt time demo and /use-cache demo with cacheLife("minutes"), cacheTag, and revalidateTag(..., "max") server actions
Build-phase no-op handlers so next build works without Redis
docker compose with redis/redis-stack for local dev

**Why:**

The existing examples/cache-handler-redis uses @neshca/cache-handler (peerDependencies.next: ">= 13.5.1 < 15"), which doesn't support Next.js 16 Cache Components
No existing example in the repo shows cacheComponents: true + cacheHandlers plural with Redis
`@trieb.work/nextjs-turbo-redis-cache` supports both interfaces, is compatible with Next.js 15.0.3+ and 16.x, and has ~11K weekly npm downloads.
This is our well maintained project, that exists for a long time already and is better maintained than @neshca/cache-handler


icyJoseph:
Hi, we have an ongoing effort here: https://github.com/vercel/next.js/pull/95346, which actually drops dependencies. 

Edit: I did not see this was indeed a different example (GitHub UI cutting the important parts)
~Could we create this as a separate example?~ I think we could also have less docs surface changes too – the example above, with no dependencies is what should be the main content. We should list our examples in the docs, for example, there's also https://github.com/mrjasonroy/cache-components-cache-handler — which would be worth calling out.

Note that we also have: https://github.com/vercel/next.js/pull/96517

icyJoseph:
@JannikZed — I want to move forward with this PR but there are some points I'd like to resolve.

examples/cache-handler-redis-turbo/README.md
```diff
@@ -0,0 +1,132 @@
+# Next.js Redis Cache Integration Example (Turbo Redis Cache)
+
+This example demonstrates a self-hosted Next.js setup using Redis as a shared
+cache with [`@trieb.work/nextjs-turbo-redis-cache`](https://www.npmjs.com/package/@trieb.work/nextjs-turbo-redis-cache).
+
+It covers **both** Next.js cache handler interfaces side by side:
+
+| Interface | Used by | Handler file |
+| --- | --- | --- |
+| `cacheHandler` (singular) | Pages Router ISR, on-demand revalidation | `cache-handler.js` |
```
icyJoseph, line 10: This also applies to App Router's ISR — the example doesn't actually have pages anymore either

examples/cache-handler-redis-turbo/README.md
```diff
@@ -0,0 +1,132 @@
...
+| `cacheHandlers` (plural) | `'use cache'` directive, `cacheComponents: true` (Next.js 16+) | `cache-components-handler.js` |
+
+The existing [`cache-handler-redis`](../cache-handler-redis) example uses
```
icyJoseph, line 13: Not true anymore

examples/cache-handler-redis-turbo/app/[timezone]/page.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * `next build` without network access).
+ */
+async function getTimeData(timezone: string): Promise<TimeData> {
+  "use cache";
+  cacheLife("minutes");
+  cacheTag("time-data");
+
+  try {
+    const res = await fetch(
+      `https://worldtimeapi.org/api/timezone/${timezone}`,
```
Comment (icyJoseph) [inline examples/cache-handler-redis-turbo/app/[timezone]/page.tsx:30]: Visiting `https://worldtimeapi.org/api/timezone/cet` or `gmt` — 404

In the updated example, we moved to `timeapi.io` 

```js
  const response = await fetch(
    `https://timeapi.io/api/time/current/zone?timeZone=${ianaTimeZone}`,
  );
```

examples/cache-handler-redis-turbo/app/[timezone]/page.tsx
```diff
@@ -0,0 +1,110 @@
...
+      </Suspense>
+      <RevalidateFrom />
+    </>
+  );
+}
+
+export default async function Page({
+  params,
+}: {
+  params: Promise<{ timezone: string }>;
```
Comment (icyJoseph) [inline examples/cache-handler-redis-turbo/app/[timezone]/page.tsx:79]: nit — Let's use `PageProps<'/[timezone]'>`

examples/cache-handler-redis-turbo/app/cache-state-watcher.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { ReactNode, useEffect, useState } from "react";
+
+type CacheStateWatcherProps = { time: number; revalidateAfter: number };
+
+export function CacheStateWatcher({
+  time,
+  revalidateAfter,
+}: CacheStateWatcherProps): ReactNode {
+  const [cacheState, setCacheState] = useState("");
```
icyJoseph, line 11: I shall fix this in the redis example too, but in canary we can use `use(browser())` from React.

examples/cache-handler-redis-turbo/cache-handler.js
```diff
@@ -0,0 +1,60 @@
...
+      // redis://localhost:6379.
+      // redisUrl: process.env.REDIS_URL,
+      // Optional: isolate cache entries per deployment.
+      // keyPrefix: process.env.VERCEL_URL ?? "turbo-example:",
+    });
+
+    return cachedHandler;
+  }
+
+  get(...args) {
```
icyJoseph, line 46: These four are not used right?

AFAIK, Next.js will do:

```js
const handler = new CacheHandler(/* some arguments */) // from line 60
```

And at runtime that'll return an instance of `RedisStringsHandler` — and it is those methods that'll be called, not the ones defined here.

examples/cache-handler-redis-turbo/package.json
```diff
@@ -0,0 +1,21 @@
+{
```
icyJoseph, line 1: Adding the comment here for lack of a better place.

I wonder if we have to merge this package, `@mrjasonroy/cache-components-cache-handler`, and possibly others, as options into the main example now. 

We could have different `next.config.js` variants showing how to couple or adapt each handler.

JannikZed: Thanks for the review — and for wanting to move this forward.

Agreed that the no-dependency `cache-handler-redis` example should stay the canonical starting point. I don’t think we should merge this package, `@mrjasonroy/cache-components-cache-handler`, and the official handler into one example with several `next.config.js` variants: they solve different jobs (protocol demo vs production adapters).

What I’m doing instead:

- Keep this as a separate example (`create-next-app --example cache-handler-redis-turbo`).
- Add a short **Production community handlers** section to the official Redis example README, listing this package and `@mrjasonroy/cache-components-cache-handler` (the k8s community link there already sets that pattern).
- Shrink the self-hosting docs change to a single Good-to-know that points at that list, instead of presenting this as a second Redis guide.

examples/cache-handler-redis-turbo/next-env.d.ts
```diff
@@ -0,0 +1,7 @@
+/// <reference types="next" />
```
icyJoseph, line 1: Just a nit here, but we recommend not checking this into git — it might be that the other example has it though

examples/cache-handler-redis-turbo/next.config.js
```diff
@@ -0,0 +1,37 @@
...
+  // Next.js 16+ Cache Components: enables the `'use cache'` directive.
+  cacheComponents: true,
+  // ISR / Pages Router / on-demand revalidation handler.
+  cacheHandler:
+    process.env.NODE_ENV === "production"
+      ? require.resolve("./cache-handler.js")
+      : undefined,
+  // `'use cache'` directive handler (plural API, Next.js 16+).
+  cacheHandlers: {
+    default: require.resolve("./cache-components-handler.js"),
```
icyJoseph, line 29: Could we use remote here — or override both? Semantically remote is the one you'd couple to a remote store.

examples/cache-handler-redis-turbo/cache-handler.js
icyJoseph: @JannikZed no comment on this? or is it documented in your library? cuz I think people might put monitoring and such within these methods, and it would not work, right? or?

icyJoseph:
@JannikZed docs part looks OK to me — and for the example I just left a couple of additional notes. Let's try to address them and prepare for a merge 🚀

examples/cache-handler-redis-turbo/app/[timezone]/page.tsx
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * `'use cache'` with a short cacheLife so the cache-state watcher can show
+ * the fresh → stale transition. Tagged with "time-data" so the server action
+ * can invalidate it via `updateTag("time-data")`.
+ *
+ * Falls back to a placeholder when the time API is unreachable (e.g. during
+ * `next build` without network access).
+ */
+async function getTimeData(ianaTimeZone: string): Promise<CachedTime> {
+  "use cache";
```
Comment (icyJoseph) [inline examples/cache-handler-redis-turbo/app/[timezone]/page.tsx]: Let's make this `"use cache: remote"`
```suggestion
  "use cache: remote";
```

examples/cache-handler-redis-turbo/app/use-cache/page.tsx
```diff
@@ -0,0 +1,73 @@
+import { cacheLife, cacheTag } from "next/cache";
+import Link from "next/link";
+import { RevalidateUseCacheFrom } from "./revalidate-from";
+
+type UselessFact = {
+  id: string;
+  text: string;
+};
+
+/**
```
icyJoseph: Here I guess it is up to you — but right now the comment is not correct. 

The `cacheHandlers` key would need a `default` entry, together with the `remote` one, otherwise this `'use cache'` is using the in-memory handler.

PR #96627, opened 2026-08-04
Fix Navigation Inspector infinite RuntimeShell requests after Resume with prefetch={true}

### LINKED ISSUE #96611: Navigation inspector: infinite requests with <Link prefetch={true}>

https://github.com/mdj-uk/next-navigation-inspector-bug

### To Reproduce

With cacheCompoments and partialPrefetching enabled:

When running`next dev`
1. open nav inspector, set to pause on nav
2. click link to page 2 (has `prefetch={true}` prop)
3. click link to home
4. click resume
5. click link to page 2 - you'll see infinite requests to dev server

Note that it makes no difference if you click resume between steps 2 and 3.

https://github.com/user-attachments/assets/438f54a7-5415-43e6-b2a3-f81e99661c50

(With partialPrefetching disabled this issue doesn't occur, but there's an earlier error "Error: Next.js encountered dynamic data during prefetching for /page2" - which makes no sense, as there is no dynamic data access) 

Currently, the second time you click a link with `prefetch={true}` prop with the nav inspector open, if it's not in the 'paused' state, infinite requests are made to the dev server.

I would expect a single request to be made.

```bash
Operating System:
  Platform: darwin
  Arch: x64
  Version: Darwin Kernel Version 23.2.0: Wed Nov 15 21:54:10 PST 2023; root:xnu-10002.61.3~2/RELEASE_X86_64
  Available memory (MB): 16384
  Available CPU cores: 4
Binaries:
  Node: 22.18.0
  npm: 10.9.3
  Yarn: N/A
  pnpm: 10.30.3
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

cacheComponents

next dev (local)

### Additional context

tested on
- next@16.3.1-canary.0
- next 16.3.0

jasleenkaur-qed42:
## What

After Navigation Inspector Resume, a second click on `<Link prefetch={true}>` could spam `next-router-prefetch: 3` (RuntimeShell) requests and never commit the navigation (URL stuck on `/`).

## Why

Locked Instant Navigation ignores segment-cache entries that aren’t in the current lock’s `ownedEntries`. Soft refresh after Resume leaves older **concrete** vary-path entries in the cache. Shell strategies store at the **shell** vary path, so those leftover concrete entries shadowed the owned shell entry. Each scheduler pass treated that as a miss, created a new Empty at the shell path, and spawned another RuntimeShell forever.

## How

In `readOrCreateSegmentCacheEntry`, when a concrete-path hit isn’t owned and the request uses a different (shell) vary path, reuse an already-owned entry at the request/shell path instead of creating another Empty.

## Test plan

- [x] `HEADLESS=true NEXT_SKIP_ISOLATE=1 pnpm test-dev-turbo test/development/app-dir/instant-navs-devtools/instant-navs-devtools.test.ts -t "SPA captures"`
- [ ] Manual: `cacheComponents` + `partialPrefetching`, Navigation Inspector pause → `prefetch={true}` nav → home → Resume → second `prefetch={true}` click → navigates once, no request flood

Fixes #96611

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -903,6 +903,7 @@ export function readOrCreateSegmentCacheEntry(
   // Navigation Testing API only; always null in production). See below.
   navigationLockPrefetch: NavigationLockPrefetch | null
 ): SegmentCacheEntry {
+  const varyPathForRequest = getSegmentVaryPathForRequest(fetchStrategy, tree)
   const existingEntry = readSegmentCacheEntry(now, tree.varyPath)
   if (existingEntry !== null) {
     if (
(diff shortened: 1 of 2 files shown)


samselikoff:
Thanks for the report, looking into this!

PR #91100, opened 2026-03-09
Turbopack: emit/collect

mischnic:
```javascript
__turbopack_emit__('./c.js', {
  namespace: 'my-test',
  data: 'data-for-c', // optional
  exports: ['c'], // optional
  with: ... // optional, import attributes like with ESM
})

// -------------------------

const getList = __turbopack_collect__({
  namespace: 'my-test',
})
const list = getList();
```

- [ ] support data-only emit
- [ ] make it `data: Array<any>` instead
- [ ] what about stylegroups
- [ ] pages router is still a problem: it has two separate ChunkGroup::Entry, so you only see what was emitted on the current runtime, not both

PR #94889, opened 2026-06-17
Turbopack: reimplement server actions using emit/collect

mischnic:
Recreation of https://github.com/vercel/next.js/pull/91212

Closes https://github.com/vercel/next.js/issues/83186

Benefits:

1. Less special casing, removes the base/full dual module graph setup
2. you get proper import traces across server actions (for errors and for the bundle analyzer)
3. client reference --imports--> server action --imports–-> client reference works now
4. There is no "server action loader module" at all anymore. So server actions get lazily initialized now, as opposed to the first server action initializing everything

- [x] the "don't compile edge when using nodejs runtime" optimization
- [x] test cases for:
    - [x] import traces, and
    - [x] enable in app-action.test.ts: `client component imported action` client->server->client
	- [x] "don't compile edge when using nodejs runtime"
- [ ] edge overcompilation?

<details>
<summary>edge overcompilation</summary>

No Edge page directly imports that file.

The normal import chain is:

app/client-error/page.js → client.js → actions.js → actions-lib.js

/client-error uses the default Node.js runtime.

The warning occurs because:

- emit_server_action() emits both Node.js and Edge references for actions imported by client components: server_actions.rs:3290-3303.
- Production uses one whole-app module graph: project.rs:1211-1213, app.rs:1131-1133.
- That graph contains runtime-specific action collectors for every endpoint: app.rs:1239-1267, app.rs:2328-2364.
- Any Edge collector can therefore discover /client-error’s emitted browser-edge reference and compile actions-lib.js through next-edge-rsc.

Edge routes in this fixture include app/client/edge/page.js, app/server/edge/page.js, app/form-edge/page.js, and the other */edge pages. None imports client-error/actions-lib.js; their presence merely activates the Edge collector.

So this warning indicates cross-page collection leakage in the new emit/collect implementation, not that the middleware-rewrite page uses Edge runtime. Also, -t filters the test execution but still builds the entire fixture.
</details>

Benchmark:
```
mischnic/emit-collect-server-actions 6f1f85836031b9e19cd3baca659ce8dca01388fc
pnpm next build --experimental-build-mode=compile  398.21s user 65.46s system 731% cpu 1:03.39 total
pnpm next build --experimental-build-mode=compile  396.63s user 64.95s system 753% cpu 1:01.22 total
pnpm next build --experimental-build-mode=compile  397.53s user 57.59s system 780% cpu 58.294 total

canary 606c4ebb368ab155ae4d26937aecca7e85da9109
[...]

PR #96631, opened 2026-08-04
Turbopack: emit/collect

mischnic:
Only add this for using it for Server Actions for now
Undocumented, not API stable. But we can't really put it behind a flag, because the next-custom-transform is going to add it

New traits:
```rust
/// A module that can collect other modules during the collect phase.
#[turbo_tasks::value_trait]
pub trait CollectingModule: Module {
    /// The namespace that this module is interesed in
    #[turbo_tasks::function]
    fn namespace(self: Vc<Self>) -> Vc<RcStr>;

    #[turbo_tasks::function]
    fn as_chunk_item(
        self: Vc<Self>,
        chunking_context: Vc<Box<dyn ChunkingContext>>,
        entry_chunk_group: Vc<Modules>,
    ) -> Vc<Box<dyn ChunkItem>>;
}

#[turbo_tasks::value_trait]
pub trait EmittedModuleReference: ModuleReference + ValueToString {
    #[turbo_tasks::function]
    fn data(self: Vc<Self>) -> Vc<CompileTimeDefineValue>;
}

```

Example:

```javascript
__turbopack_emit__('./c.js', {
  namespace: 'my-test',
  data: 'data-for-c', // optional
  exports: ['c'], // optional
  with: ... // optional, import attributes like with ESM
})

// -------------------------

const getList = __turbopack_collect__({
  namespace: 'my-test',
})
const list = getList();
```

Shortcomings in the current version (which are not a blocker for now however):
- data-only emit not implemented
- stylegroups aren't properly implemented. emitting styles is probably broken
- pages router is still a problem: it has two separate ChunkGroup::Entry, so you only see what was emitted on the current runtime, not both

Recreation of https://github.com/vercel/next.js/pull/91100


mischnic:
* **#94889** <a href="https://app.graphite.com/github/pr/vercel/next.js/94889?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#96631** <a href="https://app.graphite.com/github/pr/vercel/next.js/96631?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/96631?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#97385** <a href="https://app.graphite.com/github/pr/vercel/next.js/97385?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

crates/next-core/src/next_shared/transforms/mod.rs
```diff
@@ -35,12 +35,23 @@ pub use next_track_dynamic_imports::get_next_track_dynamic_imports_transform_rul
...
 
 use crate::next_image::{StructuredImageModuleType, module::BlurPlaceholderMode};
 
+pub async fn get_collect_rule() -> Result<ModuleRule> {
+    Ok(ModuleRule::new(
+        RuleCondition::ReferenceType(ReferenceTypeCondition::Collect),
+        vec![ModuleRuleEffect::ModuleType(ModuleType::Custom(
+            ResolvedVc::upcast(CollectModuleType::new().to_resolved().await?),
+        ))],
+    ))
```
sokra: It's a bit unexpected and prone to misuse to require a module rule here.
Assuming emit/collect is a built-in Turbopack feature, this should be handled out of the box and doesn't require an extra rule in next.js.

This should better be implemented as built-in special case.

turbopack/crates/turbopack-tests/tests/execution.rs
```diff
@@ -489,6 +493,12 @@ async fn run_test_operation(prepared_test: ResolvedVc<PreparedTest>) -> Result<V
                 }
                 .resolved_cell(),
             )],
+            module_rules: vec![ModuleRule::new(
+                RuleCondition::ReferenceType(ReferenceTypeCondition::Collect),
+                vec![ModuleRuleEffect::ModuleType(ModuleType::Custom(
+                    ResolvedVc::upcast(CollectModuleType::new().to_resolved().await?),
+                ))],
+            )],
```
sokra: Should not be needed

turbopack/crates/turbopack/src/collect_module.rs
```diff
@@ -0,0 +1,328 @@
...
+#[turbo_tasks::value_impl]
+impl CollectModuleType {
+    #[turbo_tasks::function]
+    pub fn new() -> Vc<Self> {
+        CollectModuleType {}.cell()
+    }
+}
+
+#[turbo_tasks::value_impl]
+impl CustomModuleType for CollectModuleType {
```
sokra: This should no longer be needed. You can directly create the CollectModule when matching the ReferenceType::Collect

turbopack/crates/turbopack-cli/src/contexts.rs
```diff
@@ -102,6 +104,12 @@ async fn get_client_module_options_context(
         follow_reexports: true,
         module_fragments_enabled: false,
         keep_last_successful_parse: is_dev,
+        module_rules: vec![ModuleRule::new(
+            RuleCondition::ReferenceType(ReferenceTypeCondition::Collect),
+            vec![ModuleRuleEffect::ModuleType(ModuleType::Custom(
+                ResolvedVc::upcast(CollectModuleType::new().to_resolved().await?),
+            ))],
+        )],
```
sokra: Should not be needed

turbopack/crates/turbopack-core/src/chunk/chunk_group.rs
```diff
@@ -106,6 +109,38 @@ pub async fn make_chunk_group(
...
+                Ok((chunk_item, chunk_type))
+            })
+            .try_join()
+            .await?
+            .iter()
+            .map(|&(chunk_item, chunk_type)| {
+                ChunkItemOrBatchWithAsyncModuleInfo::ChunkItem(ChunkItemWithAsyncModuleInfo {
+                    chunk_item,
+                    chunk_type,
+                    module: None,
```
sokra: We have a module here, why isn't it set here?

turbopack/crates/turbopack-core/src/chunk/chunk_group.rs
```diff
@@ -328,6 +401,14 @@ async fn chunk_group_content_operation(
                         // handled above before the sidecast
                         unreachable!();
                     }
+                    ChunkingType::PerEntry => {
+                        // TODO currently not implemented
```
sokra, line 402: Should this use `todo!()` here?

turbopack/crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -364,6 +365,17 @@ pub enum ChunkingType {
         _ty: ChunkGroupType,
         merge_tag: Option<RcStr>,
     },
+    /// Declare an emitted module (corresponds to __turboack_emit__).
+    Emitted {
+        merge_tag: RcStr,
```
sokra: Does this `merge_tag` have the same behavior as the `merge_tag` in `Shared`?

Otherwise it might make sense to give it a different name (like `namespace`)

turbopack/crates/turbopack-core/src/chunk/mod.rs
```diff
@@ -489,6 +525,8 @@ pub struct ChunkGroupContentInner {
     pub batch_groups: Vec<ResolvedVc<ModuleBatchGroup>>,
     #[bincode(with = "turbo_bincode::indexset")]
     pub async_modules: FxIndexSet<ResolvedVc<Box<dyn ChunkableModule>>>,
+    #[bincode(with = "turbo_bincode::indexset")]
+    pub collecting_modules: FxIndexSet<ResolvedVc<Box<dyn CollectingModule>>>,
```
sokra, line 534: Add a comment here (and maybe to the other fields too)

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -651,6 +666,28 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
...
+                            // These only exist in the module_batches graph.
+                            unreachable!();
+                        }
+                        ChunkingType::PerEntry => {
+                            // A collect module belongs wherever its importer belongs. The
+                            // per-entry behavior is carried by the `chunk_group` argument of
+                            // `CollectingModule::as_chunk_item`, not by chunk group membership.
+                            // This matches `collect.rs`, which inherits the parent's entry
+                            // membership across this edge.
+                            ChunkGroupInheritance::Inherit(parent)
```
sokra: Do these edges exist in this graph at all? Looks like it's only used by CollectModules which are created later?

turbopack/crates/turbopack-core/src/module_graph/collect.rs
```diff
@@ -0,0 +1,294 @@
...
+    for (i, entries) in entry_groups.iter().enumerate() {
+        for entry in *entries {
+            module_entry_membership
+                .entry(*entry)
+                .or_default()
+                .insert(i as u32);
+        }
+    }
+
+    // First, compute the depth for each module in the graph
```
sokra, line 74: And initializes the `module_entry_membership` values

turbopack/crates/turbopack-core/src/module_graph/collect.rs
```diff
@@ -0,0 +1,294 @@
...
+            // BinaryHeap prioritizes high values
+
+            // Smaller depth has higher priority
+            let depth_order = self.depth.cmp(&other.depth).reverse();
+            // Smaller bitmap length has higher priority
+            let bitmap_len_order = self.bitmap_len.cmp(&other.bitmap_len).reverse();
+
+            depth_order.then(bitmap_len_order)
+        }
+    }
```
sokra: This could also be derived and you would use `Reverse<TraversalPriority>` in the traversal

turbopack/crates/turbopack-core/src/module_graph/collect.rs
```diff
@@ -0,0 +1,294 @@
...
+            if parent == node {
+                // A self-reference
+                Ok(GraphTraversalAction::Skip)
+            } else {
+                let [Some(parent_membership), Some(current_membership)] =
+                    module_entry_membership.get_disjoint_mut([&parent, &node])
+                else {
+                    // All modules are inserted in the previous iteration
+                    // Technically unreachable, but could be reached due to eventual
+                    // consistency
```
sokra, line 143: That looks really unreachable to me, even with eventual consistency

turbopack/crates/turbopack-core/src/module_graph/collect.rs
```diff
@@ -0,0 +1,294 @@
...
+            .await?
+        {
+            map.entry(namespace).or_default().push(m);
+        }
+        map
+    };
+
+    // Now we have all necessary information. List out all collected references for each (Entry
+    // Module, Collecting Module) pair they are contained in.
+    #[allow(clippy::type_complexity)]
```
sokra, line 215: I kind of agree with this lint. The type is really complex

turbopack/crates/turbopack-core/src/module_graph/collect.rs
```diff
@@ -0,0 +1,294 @@
...
+        };
+
+        for collecting_module in collecting_modules.get(merge_tag).into_iter().flatten() {
+            let collecting_membership = module_entry_membership
+                .get(&ResolvedVc::upcast(*collecting_module))
+                .context("Module entry membership not found")?;
+
+            if *emit_to_all_entries || !collecting_membership.is_disjoint(emitted_membership) {
+                for entry in collecting_membership.iter() {
+                    if *emit_to_all_entries || emitted_membership.contains(entry) {
```
sokra: For efficiency this could use collecting_membership.intersection(emitted_membership) and iterate that.

turbopack/crates/turbopack-core/src/module_graph/collect.rs
```diff
@@ -0,0 +1,294 @@
...
+                        collected_references
+                            .entry(ResolvedVc::upcast(*collecting_module))
+                            .or_default()
```
sokra: This can be pulled before the `for entry in collecting_membership.iter()` loop

turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
mischnic: We have the collecting module (loader inserted by `__turbopack_collect__`) which is a normal module in the graph (but with this special chunking type)

The collected modules are only created later on.

turbopack/crates/turbopack-core/src/module_graph/collect.rs
mischnic: We added this here: https://github.com/vercel/next.js/pull/79242

turbopack/crates/turbopack-core/src/module_graph/collect.rs
mischnic: This was copied 1:1 from turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs.
I've made that one public and reused it.
We can change that separately.

turbopack/crates/turbopack-core/src/chunk/chunk_group.rs
mischnic: We should indeed.
But it's a `Vc<Box<dyn CollectingModule>>` and that one doesn't implement the required `ChunkableModule` because it has it's own special `as_chunk_item` with the chunk group as additional parameter:
```
    #[turbo_tasks::function]
    fn as_chunk_item(
        self: Vc<Self>,
        chunking_context: Vc<Box<dyn ChunkingContext>>,
        entry_chunk_group: Vc<Modules>,
    ) -> Vc<Box<dyn ChunkItem>>;
```

turbopack/crates/turbopack-core/src/chunk/mod.rs
mischnic: renamed. it does behave very similar, but namespace is what's used everywhere else as well

turbopack/crates/turbopack-core/src/chunk/chunk_group.rs
mischnic: replaced with bail

turbopack/crates/turbopack-core/src/chunk/mod.rs
mischnic: Done

turbopack/crates/turbopack-core/src/module_graph/collect.rs
mischnic: Yes and no.
This should only really happen though if the for loop has a match then. And I haven't found a better way to do this lazy initialization.

turbopack/crates/turbopack-core/src/module_graph/collect.rs
mischnic: very good idea

turbopack/crates/turbopack-core/src/module_graph/collect.rs
mischnic: actually, I was able to fix this after the `intersection` change

turbopack/crates/turbopack-core/src/module_graph/collect.rs
mischnic: I added a comment here.
I don't know how to make the type better. It really is a `Module -> Chunk Group -> List<(reference, module)>` multi-map of lists, no way around that

mischnic:
I have fixed
> [P2] __turbopack_collect__ accepts unknown and malformed options; __turbopack_emit__ rejects them

This is intentional (unfortunately)
> [P2] scope: 'app' silently means "all entries in the current graph", which differs between dev and prod

turbopack/crates/turbopack-core/src/chunk/chunk_group.rs
mischnic: Fixed. I changed `ChunkItemWithAsyncModuleInfo.module` to be a `ResolvedVc<Box<dyn Module>>` instead of `ResolvedVc<Box<dyn ChunkableModule>>`

turbopack/crates/turbopack-core/src/module_graph/merged_modules.rs
```diff
@@ -185,63 +186,70 @@ pub async fn compute_merged_modules(module_graph: Vc<ModuleGraph>) -> Result<Vc<
...
-                        if module_merged_groups.entry(module).or_default().insert(idx) {
-                            // Mark and continue traversal because modified (or first visit)
-                            GraphTraversalAction::Continue
+                        // Either a non-hoisted reference or an incompatible parent or child module
+                        if entry_modules.insert(module) {
+                            // Not assigned a new group before, create a new one.
+                            let idx = next_index;
+                            next_index += 1;
+
+                            if module_merged_groups.entry(module).or_default().insert(idx) {
```
sokra: This is a functional change to merged modules?
There is probably a reason behind that. Which one?

turbopack/crates/turbopack-core/src/module_graph/mod.rs
```diff
@@ -1931,6 +1948,22 @@ impl Visit<SingleModuleGraphBuilderNode, RefData> for SingleModuleGraphBuilder<'
                     let _span = span.entered();
                     span = tracing::info_span!("async reference");
                 }
+                ChunkingType::PerEntry => {
+                    let _span = span.entered();
+                    span = tracing::info_span!("per-entry reference");
+                }
+                ChunkingType::Emitted {
+                    namespace: merge_tag,
```
sokra:
```suggestion
                    namespace,
```

turbopack/crates/turbopack-core/src/module_graph/module_batches.rs
```diff
@@ -888,8 +960,46 @@ pub async fn compute_module_batches(
             }
         }
 
-        debug_assert_eq!(graph.capacity().0, graph.node_count());
-        debug_assert_eq!(graph.capacity().1, graph.edge_count());
```
sokra, line 892: Why deleting the debug asserts?

turbopack/crates/turbopack-core/src/module_graph/module_batches.rs
```diff
@@ -838,6 +891,18 @@ pub async fn compute_module_batches(
         span.record("modules", modules_count);
         span.record("edges", edges_count);
 
+        // Build a module -> NodeIndex lookup for all modules in the graph
+        let mut module_to_node: FxHashMap<ResolvedVc<Box<dyn Module>>, NodeIndex> =
+            FxHashMap::default();
+        for (module, batches) in &parallel_module_to_pre_batch {
+            if let Some(&first_batch) = batches.first() {
```
sokra, line 900: Worth a comment that parallel_module_to_pre_batch will only have vecs of len = 1

turbopack/crates/turbopack-core/src/compile_time_info.rs
```diff
@@ -127,6 +128,39 @@ pub enum CompileTimeDefineValue {
...
+        match self {
+            CompileTimeDefineValue::Null => "null".to_string(),
+            CompileTimeDefineValue::Bool(false) => "false".to_string(),
+            CompileTimeDefineValue::Bool(true) => "true".to_string(),
+            CompileTimeDefineValue::Number(v) => serde_json::to_string(v).unwrap(),
+            CompileTimeDefineValue::String(v) => serde_json::to_string(v).unwrap(),
+            CompileTimeDefineValue::BigInt(b) => format!("{}n", b),
+            CompileTimeDefineValue::Array(a) => {
+                let elements: Vec<_> = a.iter().map(|v| v.as_ecmascript()).collect();
+                format!("[{}]", elements.join(", "))
```
sokra: That's not very efficient. Only create a single string and avoid the temp string

turbopack/crates/turbopack-core/src/compile_time_info.rs
```diff
@@ -127,6 +128,39 @@ pub enum CompileTimeDefineValue {
     Regex(RcStr, RcStr),
 }
 
+impl CompileTimeDefineValue {
+    pub fn as_ecmascript(&self) -> String {
```
sokra: I feel like this should use `Display` instead. A formatter pattern would be more efficient

sokra (approved):

turbopack/crates/turbopack-core/src/module_graph/module_batches.rs
mischnic: I don't remember. reverted

turbopack/crates/turbopack-core/src/module_graph/merged_modules.rs
mischnic: at one point, emitted or collected modules used self-references. 
somehow merged_modules didn'T properly handle these loops. but I can remove this again now

turbopack/crates/turbopack-core/src/compile_time_info.rs
mischnic: Done. Using `CompileTimeDefineValue.display_ecmascript()` similar to `Path.display()`

PR #96632, opened 2026-08-04
Fix missing styled-jsx styles in Pages Router SSR on adapter builds

sokra:
### What?

Fixes styled-jsx styles being missing from the server-rendered HTML of Pages
Router apps built through a build adapter (what a Vercel deployment uses),
when the app has its own `styled-jsx` dependency.

- `crates/next-api/src/next_server_nft.rs`: extracts
  `styled_jsx_require_hook_modules()` and adds an accurate comment to the
  `is_using_adapter` early return (it no longer claims no NFT tracing is
  needed - see "Why?").
- `crates/next-api/src/project.rs`: `Project::additional_traced_modules` now
  also returns those modules, alongside the existing `cacheHandler`/
  `cacheHandlers` entries.
- `packages/next/src/server/require-hook.ts`: extracts
  `styledJsxRequireHookEntries()` (the JS-side equivalent, used by webpack)
  and replaces the silent `catch (_) {}` around registering the aliases with
  a diagnostic warning (still never throws).
- `packages/next/src/build/adapter/build-complete.ts`,
  `packages/next/src/build/collect-build-traces.ts`: use that helper instead
  of re-deriving the same resolution inline. No behavior change for webpack.
- `test/production/adapter-styled-jsx/`: new regression test.

### Why?

The Pages Router renderer and user code have to share a single `styled-jsx`
module instance: `render.tsx` creates the style registry and hands it to
user code through a React context owned by that specific module instance. If
user code ends up with a *different* `styled-jsx` instance - which happens as
soon as the app depends on a `styled-jsx` version that doesn't dedupe with
Next.js' own pinned version - `JSXStyle` silently renders nothing during SSR
(`if (!registry) return null`). The `jsx-*` class names are still emitted by
the transform, so the only visible symptom is a flash of unstyled content:
the CSS only gets inserted client-side after hydration.

Turbopack keeps `styled-jsx`/`styled-jsx/style` external in the pages server
bundle, so the single-instance guarantee is established at **runtime** by
`next/dist/server/require-hook`, which resolves those requests to Next.js'
own copy from its own install location via a plain `require.resolve()`.
Nothing in any module graph references the files that resolves to (user code
only ever references its own copy), so output tracing has to add them
explicitly, or the deployment is missing the file, the hook's
`require.resolve()` throws, the aliases are (silently, since #89402) never
registered, and the two `styled-jsx` instances drift apart.

That explicit tracing existed only for the whole-app
`next-server.js.nft.json` / `next-minimal-server.js.nft.json` - and those are
not generated at all when a build adapter is used, because an adapter
assembles its output from each endpoint's own NFT instead
(`is_using_adapter` early return in `next_server_nft_assets`). Its comment
said adapters "don't need any server NFTs" - true for those two whole-app
files, but not for the styled-jsx entries they also carried, which had no
other way into an adapter build. Reproduced against a published
`next@16.3.0` release: a plain `next build` with an adapter never shipped
`node_modules/next/node_modules/styled-jsx/style.js`, nothing deleted or
hand-edited.

### How?

`Project::additional_traced_modules` is Turbopack's existing mechanism for
"trace this into every endpoint even though nothing references it" - it
already carries `nextConfig.cacheHandler`/`cacheHandlers` for exactly the
same reason (a runtime-only dependency invisible to static analysis).
Adding the require hook's styled-jsx modules there gets them into every
endpoint's own `*.nft.json` via the existing `trace_endpoint` plumbing, which
`build-complete.ts` already loads unconditionally for both bundlers - so no
TypeScript change is needed on the Turbopack side at all.
`styled_jsx_require_hook_modules()` centralizes the resolution (used by both
`additional_traced_modules` and the pre-existing whole-app NFT, which still
needs it independently for `output: 'standalone'`).

Webpack has no equivalent Rust-side tracing, so `build-complete.ts` keeps
tracing these itself via `nodeFileTrace`, now through the shared
`styledJsxRequireHookEntries()` TS helper instead of re-deriving the same
resolution inline.

The new test builds a Pages Router page with `<style jsx>` through a build
adapter fixture (same shape as `test/production/adapter-root`), with a
`styled-jsx` dependency version that doesn't dedupe with Next.js' pinned
version, and asserts the adapter's `onBuildComplete` output declares every
file the require hook resolves.

Verified end-to-end against a published `next@16.3.0` install (the sandbox
can't build the native Turbopack addon):
- an existing, unmodified mechanism (`cacheHandler`) was used to confirm the
  endpoint-NFT -> adapter-assets pipeline this fix relies on actually carries
  a traced module through to the deployment;
- the Rust change's exact output was simulated at the point `build-complete`
  loads each endpoint's NFT, and the SSR HTML went from 0 `<style
  id="__jsx-...">` tags (with `jsx-*` classes still present - the reported
  flash) to the expected styled-jsx CSS being present;
- reverting the simulation reproduces the failure again;
- webpack output (`next/node_modules/styled-jsx/{index,style}.js`,
  `node-environment`, `require-hook`) is unaffected.
- `cargo check` / `cargo clippy` / `rustfmt --check` pass on the Rust change.

- [x] Tests added (`test/production/adapter-styled-jsx/`)
- [x] Errors have a helpful link attached — n/a, no user-facing error added (warning only)

### DIFF
--- a/crates/next-api/src/next_server_nft.rs
+++ b/crates/next-api/src/next_server_nft.rs
@@ -23,19 +23,18 @@ use turbopack_resolve::ecmascript::cjs_resolve;
 
 use crate::{nft::traced_modules_for_entries, project::Project};
 
-/// The modules `next/dist/server/require-hook` resolves `styled-jsx` requests to at runtime so
-/// that the Pages Router renderer and user code share a single `styled-jsx` module instance (see
-/// that file for why they otherwise wouldn't). Nothing in any module graph references these -
-/// user code only ever imports its own copy of `styled-jsx` - so whatever assembles a
-/// deployment's file list has to include them explicitly, or the hook fails to register its
-/// aliases and all styled-jsx styles silently disappear from the server-rendered HTML.
+/// The modules `next/dist/server/require-hook` resolves its aliased requests to at runtime
+/// (currently all of styled-jsx), so that the Pages Router renderer and user code share a single
+/// `styled-jsx` module instance (see that file for why they otherwise wouldn't). Nothing in any
+/// module graph references these - user code only ever imports its own copy of `styled-jsx` - so
+/// whatever assembles a deployment's file list has to include them explicitly, or the hook fails
+/// to register its aliases and all styled-jsx styles silently disappear from the server-rendered
+/// HTML.
 ///
 /// Used by the server NFTs below and, so that they are part of every endpoint's trace regardless
 /// of how the output is assembled, by [`Project::additional_traced_modules`].
 #[turbo_tasks::function]
-pub(crate) async fn styled_jsx_require_hook_modules(
-    project_path: FileSystemPath,
-) -> Result<Vc<Modules>> {
+pub(crate) async fn require_hook_modules(project_path: FileSystemPath) -> Result<Vc<Modules>> {
     let asset_context = Vc::upcast(externals_tracing_module_context(
         get_tracing_compile_time_info(),
         false,
@@ -286,15 +285,13 @@ impl ServerNftJsonAsset {
             )),
         };
 
-        // The `styled-jsx` modules the require hook needs are part of every endpoint's trace (see
+        // The modules the require hook needs are part of every endpoint's trace (see
         // `Project::additional_traced_modules`), but `next-server.js` / `next-minimal-server.js`
         // are traced on their own for `output: 'standalone'`, so they have to be added here too.
-        let styled_jsx_modules = styled_jsx_require_hook_modules(project_path)
-            .owned()
-            .await?;
+        let hook_modules = require_hook_modules(project_path).owned().await?;
 
         Ok(Vc::cell(
-            styled_jsx_modules
+            hook_modules
                 .into_iter()
                 .chain(
                     entries
--- a/crates/next-api/src/next_server_nft.rs
+++ b/crates/next-api/src/next_server_nft.rs
@@ -23,6 +23,49 @@ use turbopack_resolve::ecmascript::cjs_resolve;
 
 use crate::{nft::traced_modules_for_entries, project::Project};
 
+/// The modules `next/dist/server/require-hook` resolves `styled-jsx` requests to at runtime so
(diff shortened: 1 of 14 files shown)


packages/next/src/build/adapter/build-complete.ts
```diff
@@ -51,7 +51,7 @@ import { getRedirectStatus, modifyRouteRegex } from '../../lib/redirect-status'
 import { getNamedRouteRegex } from '../../shared/lib/router/utils/route-regex'
 import { escapeStringRegexp } from '../../shared/lib/escape-regexp'
 import { sortSortableRoutes } from '../../shared/lib/router/utils/sortable-routes'
-import { defaultOverrides } from '../../server/require-hook'
+import { styledJsxRequireHookEntries } from '../../server/require-hook'
```
mischnic: rename this
```suggestion
import { requireHookEntries } from '../../server/require-hook'
```

crates/next-api/src/next_server_nft.rs
```diff
@@ -23,6 +23,49 @@ use turbopack_resolve::ecmascript::cjs_resolve;
...
+/// that the Pages Router renderer and user code share a single `styled-jsx` module instance (see
+/// that file for why they otherwise wouldn't). Nothing in any module graph references these -
+/// user code only ever imports its own copy of `styled-jsx` - so whatever assembles a
+/// deployment's file list has to include them explicitly, or the hook fails to register its
+/// aliases and all styled-jsx styles silently disappear from the server-rendered HTML.
+///
+/// Used by the server NFTs below and, so that they are part of every endpoint's trace regardless
+/// of how the output is assembled, by [`Project::additional_traced_modules`].
+#[turbo_tasks::function]
+pub(crate) async fn styled_jsx_require_hook_modules(
```
mischnic: rename to
```suggestion
pub(crate) async fn require_hook_modules(
```

mischnic (approved):

mischnic:
Ideally we'd have a proper behavioral test though. not just checking if some files exist

PR #96635, opened 2026-08-04
accidental

jimmyhmiller:
## Summary

Use the AST facade abstraction when detecting directive prologues in Turbopack ECMAScript transforms. Add coverage for directives within and after the directive prologue.

## Verification

- `git diff --check`
- `cargo fmt -- --check`
- Not run: Rust tests (`../../oxc-fascade` is missing locally)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -327,6 +327,20 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jimmyhmiller:
Please ignore. Complete mistake of something else

PR #96636, opened 2026-08-04
Fix Turbopack worker chunk loading with asset prefix

### LINKED ISSUE #96613: Turbopack: `experimental.turbopackWorkerAssetPrefix` makes Workers load but never execute when `assetPrefix` is a cross-origin CDN

https://github.com/Manitej66/turbopack-worker-asset-prefix-repro

### To Reproduce

```bash
git clone https://github.com/Manitej66/turbopack-worker-asset-prefix-repro
cd turbopack-worker-asset-prefix-repro
npm install
npm run build
npm run cdn &   # serves .next/static on :3101, standing in for a cross-origin CDN
npm start       # serves the app on :3100
```

Open http://localhost:3100 and click **run worker**.

The relevant configuration is two lines:

```js
// next.config.js
module.exports = {
  assetPrefix: 'http://localhost:3101',
  experimental: { turbopackWorkerAssetPrefix: '' },
}
```

and the worker is six lines:

```ts
// app/worker.ts
console.log('[worker] module evaluated')

self.onmessage = (event: MessageEvent<string>) => {
  console.log('[worker] message received')
  self.postMessage(`pong: ${event.data}`)
}

export {}
```

The repo also has `npm run verify`, which shows the cause straight from the build output with no browser involved, exiting `1` on the mismatch and `0` when the two prefixes agree.

**Expected:** the page logs `reply: pong: ping`.

**Current:** nothing happens — no message, no `onerror`, no page error, no failed request.

Verified in headless Chrome across three builds of identical application code.

**1. `assetPrefix: 'http://localhost:3101'` + `turbopackWorkerAssetPrefix: ''` — the bug**

```
creating worker
posted message, waiting for reply
5s elapsed
```

Every worker asset loads successfully, same-origin:

```
GET http://localhost:3100/_next/static/chunks/turbopack-worker-2gqdcwp7k90ea.js#params=...  200
GET http://localhost:3100/_next/static/chunks/turbopack-06-2_v5x_p6ki.js                    200
GET http://localhost:3100/_next/static/chunks/1a292u8thxfy7.js                              200
```

Neither `[worker] module evaluated` nor `[worker] message received` is ever logged. `self.onmessage` is never assigned, so the posted message is delivered to a worker with no handler and silently dropped. The failure is completely silent, which makes it hard to track down in production.

**2. `assetPrefix: ''` — control, same code**

```
console: [worker] module evaluated
console: [worker] message received
reply: pong: ping
```

**3. `assetPrefix: 'http://localhost:3101'`, option omitted — the error the option exists to prevent**

```
constructor threw: Failed to construct 'Worker': Script at
'http://localhost:3101/_next/static/chunks/turbopack-worker-....js#params=...'
cannot be accessed from origin 'http://localhost:3100'.
```

So with a cross-origin `assetPrefix` there is currently no working configuration for `new Worker(new URL(..., import.meta.url))`.

### Root cause

`turbopackWorkerAssetPrefix` (added in #93271) is applied to the URLs generated in the **parent** chunking context — the worker entrypoint URL and the chunk URLs handed to the bootstrap via `#params=` — but the worker's **own runtime chunk** is still emitted with `CHUNK_BASE_PATH` set from `assetPrefix`:

```js
// .next/static/chunks/turbopack-<hash>.js  (the worker's runtime chunk)
var CHUNK_BASE_PATH = "https://cdn.example.com/_next/";
```

Inside `registerChunk`, the two sides of the chunk-resolver map are then keyed with different base paths:

```ts
// turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
async registerChunk(chunk, params) {
  let chunkPath
  if (chunk != null) {
    chunkPath = getPathFromScript(chunk)
    const resolver = getOrCreateResolver(getUrlFromScript(chunk))  // ← key: chunk.src, i.e. WORKER base path
    resolver.resolve()
  }
  if (params == null) return

  for (const otherChunkData of params.otherChunks) {
    const otherChunkPath = getChunkPath(otherChunkData)
    const otherChunkUrl = getChunkRelativeUrl(otherChunkPath)      // ← key: CHUNK_BASE_PATH (the CDN)
    getOrCreateResolver(otherChunkUrl)
  }

  await Promise.all(
    params.otherChunks.map((d) => loadInitialChunk(chunkPath, d))  // ← waits on the CDN-keyed resolver
  )
  // runtime module ids are instantiated after this await — never reached
}
```

In a worker, `chunk.src` comes from the bootstrap:

```ts
// runtime-base.ts, getChunkFromRegistration
if (typeof TURBOPACK_NEXT_CHUNK_URLS !== "undefined") {
  return { src: TURBOPACK_NEXT_CHUNK_URLS.pop()! } as CurrentScript;
}
```

and `TURBOPACK_NEXT_CHUNK_URLS` is populated from `#params=`, which the parent built with the **worker** asset prefix (`/_next/static/chunks/x.js?dpl=…`). Meanwhile `getChunkRelativeUrl` defaults to `CHUNK_BASE_PATH`:

```ts
// runtime-base.ts
function getChunkRelativeUrl(
  chunkPath: ChunkPath | ChunkListPath,
  basePath: string = CHUNK_BASE_PATH   // ← the CDN, not the worker prefix
): ChunkUrl {
  …
  return `${basePath}${encodedPath}${ASSET_SUFFIX}` as ChunkUrl
}
```

So `registerChunk` resolves the resolver keyed `"/_next/static/chunks/x.js?dpl=…"` while awaiting the one keyed `"https://cdn.example.com/_next/static/chunks/x.js?dpl=…"`. Two different entries in the map.

The await never settles because for `SourceType.Runtime` the backend deliberately does not fetch — it assumes `importScripts` already ran (which it did, under the other key):

```ts
if (sourceType === SourceType.Runtime) {
  resolver.loadingStarted = true;
  if (isCss(chunkUrl)) resolver.resolve(); // JS: never resolved
  return resolver.promise;
}
```

Hence: no request, no error, `Promise.all` pending forever, `params.runtimeModuleIds` never instantiated, worker entry module never evaluated.

`getPathFromScript` has the same latent problem — it strips `CHUNK_BASE_PATH` from `src`, which in a worker no longer matches, so the derived `chunkPath` keeps its leading `/_next/`.

Note this affects **every** worker, not just ones with heavy dependencies: even a trivial dependency-free worker is emitted as a runtime chunk plus one module chunk, so `params.otherChunks` is never empty and the broken path is always taken. It may be worth checking whether `test/e2e/turbopack-worker-asset-prefix` asserts an actual message round-trip, since a URL-shape assertion would pass here.

### Proposed fix

Emit the worker's runtime chunk with `CHUNK_BASE_PATH` set to the worker asset prefix rather than the global one. Every URL a worker's runtime constructs has to be same-origin with the worker anyway (the bootstrap enforces exactly that with its `Refusing to load script from foreign origin` check), so within a worker chunking context the worker prefix _is_ the chunk base path. That fixes `getChunkRelativeUrl`, `getPathFromScript` and any subsequent dynamic chunk loads at once, with no call-site changes.

The narrower alternative is to thread `WORKER_BASE_PATH` into the worker runtime and use `WORKER_BASE_PATH ?? CHUNK_BASE_PATH` as the default `basePath` in `getChunkRelativeUrl` and as the prefix stripped in `getPathFromScript`.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 24.14.1
  npm: 11.11.0
Relevant Packages:
  next: 16.3.0
  react: 19.2.5
  react-dom: 19.2.5
  typescript: 6.0.3
Next.js Config:
  output: standalone
```

Turbopack

### Additional context

Confirmed by building a real app twice — once with `ASSET_PREFIX` unset and once with `ASSET_PREFIX=https://cdn.example.com` — and diffing the emitted worker runtime chunks. With no `assetPrefix` both keys are `/_next/…`, they match, and workers work; that is why this only reproduces in deployments that use a CDN.

Related: #93271 (which introduced `turbopackWorkerAssetPrefix`).

### Additional context

Supersedes #96610, which the reproduction bot closed before I had a repository link.

timneutkens:
## Summary

- propagate the effective worker asset prefix through the worker bootstrap so runtime chunk registration and loading use the same base path
- verify workers execute with both explicit and empty `turbopackWorkerAssetPrefix` overrides
- add coverage for the Resvg WASM worker reproduction

Fixes #96613

## Verification

- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-turbo test/e2e/turbopack-worker-asset-prefix/turbopack-worker-asset-prefix.test.ts`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-dev-turbo test/e2e/turbopack-worker-asset-prefix/turbopack-worker-asset-prefix.test.ts`
- Resvg WASM worker test in production and development Turbopack
- focused `turbopack-tests` worker and ignore-comments snapshots
- `pnpm swc-build-native`
- Full `pnpm build-all` did not complete locally because of unrelated existing `httpxy` and proxy-request type errors


turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -17,6 +17,9 @@ declare var TURBOPACK_ASSET_SUFFIX: string
 // can't be detected via `document.currentScript`. Note it's stored in reversed
 // order to use `push` and `pop`
 declare var TURBOPACK_NEXT_CHUNK_URLS: ChunkUrl[] | undefined
+// Used in WebWorkers to override the regular chunk base path with the base
+// used for the worker entrypoint and its initial chunks.
+declare var TURBOPACK_WORKER_CHUNK_BASE_PATH: string | undefined
```
sokra:
```suggestion
declare var TURBOPACK_CHUNK_BASE_PATH: string | undefined
```

That's similar to `TURBOPACK_ASSET_SUFFIX` and `TURBOPACK_NEXT_CHUNK_URLS` which also don't have WORKER in the name

sokra (approved):

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
timneutkens: :robot: Done -- renamed the bootstrap/runtime global to TURBOPACK_CHUNK_BASE_PATH, updated the generated bootstrap assignment and runtime reads, and regenerated all affected snapshots.

PR #96637, opened 2026-08-04
fix(dev-overlay): restore touch-action none on the draggable surface

### LINKED ISSUE #96634: Dev indicator dragging makes it stuck when using touch

https://github.com/klmkyo/nextjs-devtools-drag-bug-demo

### To Reproduce

This bug can be reproduced on any touch enabled device. Using Chrome Devtools to simulate a touch device works, but the same issue happens in Android WebView and Chromium, or iOS WebKit, including Safari. iOS examples were taken on an iOS simulator to visualize touches. Issue is present on devices with: iOS 26.5, iOS 27 (Public Beta 2), Android Chrome 150.0.7871.128 (Android 17).

To reproduce, a drag needs to be initiated, originating from where the dev tools indicator is, that diagonal or vertical, which will cause the browser to initiate a viewport scroll event instead, leaving the indicator in a broken state.

https://github.com/user-attachments/assets/27e25147-5fde-4861-8f81-9c2f8e39488b

https://github.com/user-attachments/assets/d41b44c5-16cd-42f0-a544-aaa607af8844

Current behaviour: devtools indicator gets stuck, reloading the page is needed to fix it (see videos listed above).

Expected behaviour: devtools indicator, similarly to how it would behave when dragged by a mouse, would move to the top of the screen.

Video demo of the expected behaviour (taken from the https://github.com/vercel/next.js/pull/96638 fix):

https://github.com/user-attachments/assets/377fa189-d99c-423e-ad9e-d7f340fd8256

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 24.18.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: 10.33.0
Relevant Packages:
  next: 16.3.1-canary.0 // Latest available version is detected (16.3.1-canary.0).
  eslint-config-next: N/A
  react: 19.3.0-canary-cbb046ab-20260731
  react-dom: 19.3.0-canary-cbb046ab-20260731
  typescript: 6.0.2
Next.js Config:
  output: N/A
```

Error Overlay, CSS, Not sure

next dev (local)

### Additional context

The regression was introduced in https://github.com/vercel/next.js/pull/86816. In the commit before (51459c6f8a) the issue did not occur, and in the commit after (907e1b671c) it can be observed.

A proposed solution, with the addition of graceful cancelled event handling, is implemented in https://github.com/vercel/next.js/pull/96638.

lazerg:
Fixes: https://github.com/vercel/next.js/issues/96634

On a touch device, a vertical or diagonal drag starting on the dev tools indicator makes the browser take the gesture over as a viewport scroll. That fires `pointercancel`, so the drag state machine never reaches `pointerup` and the indicator is left mid-drag with a stale `translate` until the page is reloaded.

#86816 moved `user-select` handling into the pointer handlers and dropped the whole inline style block from `<Draggable>`, which also took `touch-action: none` with it. That property is what kept the browser from claiming the gesture, so this puts it back and leaves the `user-select` change alone.

### DIFF
--- a/packages/next/src/next-devtools/dev-overlay/components/errors/dev-tools-indicator/draggable.tsx
+++ b/packages/next/src/next-devtools/dev-overlay/components/errors/dev-tools-indicator/draggable.tsx
@@ -143,7 +143,12 @@ export function Draggable({
   }
 
   return (
-    <div {...props} {...drag} ref={ref}>
+    <div
+      {...props}
+      {...drag}
+      ref={ref}
+      style={{ touchAction: 'none', ...props.style }}
+    >
       {children}
     </div>
   )
--- a/test/unit/dev-overlay-draggable.test.tsx
+++ b/test/unit/dev-overlay-draggable.test.tsx
@@ -1,0 +1,36 @@
+/**
+ * @jest-environment jsdom
+ */
+import { render } from '@testing-library/react'
+import '@testing-library/jest-dom'
+import { Draggable } from '../../packages/next/src/next-devtools/dev-overlay/components/errors/dev-tools-indicator/draggable'
+
+describe('Draggable', () => {
+  it('opts the drag surface out of browser touch gestures', () => {
+    const { container } = render(
+      <Draggable padding={20} position="bottom-left" setPosition={() => {}}>
+        <div>child</div>
+      </Draggable>
+    )
+
+    const surface = container.firstElementChild as HTMLElement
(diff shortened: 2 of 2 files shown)


lazerg:

lazerg:
Closing since the linked issue #96634 was already resolved elsewhere.

PR #96639, opened 2026-08-04
Turbopack: skip symlinks to directories when collecting traced includes

### LINKED ISSUE #96626: Turbopack build panics (EISDIR) in 16.3.0 when output tracing encounters a symlink to a directory (regression from 16.2)

https://github.com/rollokd/next-turbopack-nft-symlink-eisdir

### To Reproduce

1. `npm install`
2. `npm run build`

The repro is intentionally tiny: a fresh app-router app plus

- `data/dir-link` — a symlink pointing at the directory `data/real-dir` (committed as a symlink)
- an `outputFileTracingIncludes` glob that matches it:

```ts
// next.config.ts
outputFileTracingIncludes: {
  '/*': ['./data/**/*'],
}
```

Control: `npm install next@16.2.11 && rm -rf .next && npm run build` on the same project succeeds, so this is a 16.3.0 regression.

**Current:** `next build` crashes with a fatal Turbopack panic:

```
Error [TurbopackInternalError]: reading file "/private/tmp/nft-symlink-repro/data/dir-link"

Caused by:
- Is a directory (os error 21)

Debug info:
- Execution of emit_all_output_assets_once_with_issues_operation failed
- Execution of EntrypointsOperation::new failed
- Execution of emit_all_output_assets_once_operation failed
- Execution of Project::emit_all_output_assets failed
- Execution of emit_assets failed
- Execution of emit failed
- Execution of <NftJsonAsset as Asset>::content failed
- Execution of *FileContent::hash failed
- Execution of <DiskFileSystem as FileSystem>::read failed
- reading file "/private/tmp/nft-symlink-repro/data/dir-link"
- Is a directory (os error 21)
```

**Expected:** the build succeeds, as it does on 16.2.11 — a symlink to a directory inside a traced/included path should be traversed or skipped, not treated as a regular file.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 36864
  Available CPU cores: 12
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 11.9.0
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 6.0.2
Next.js Config:
  output: N/A
```

Turbopack, Output (export/standalone)

next build (local), Vercel (Deployed)

### Additional context

Root cause, from reading the v16.3.0 source:

1. `crates/next-api/src/nft.rs` (`get_glob_includes`, around L198) classifies glob results with:

   ```rust
   let (DirectoryEntry::File(file_path) | DirectoryEntry::Symlink(file_path)) = entry
   else { continue; };
   result.push(file_path.clone());
   ```

   Every symlink is pushed into the includes list as if it were a file — the symlink target's type is never checked.

2. `crates/next-api/src/nft_json.rs` (`NftJsonAsset::content`) — new in 16.3 — content-hashes every include for the new `fileHashes` field:

   ```rust
   file_path.read().hash(hash_salt, HashAlgorithm::Xxh3Hash128Hex)
   ```

   `read()` on a symlink-to-directory fails with EISDIR, which surfaces as a fatal `TurbopackInternalError`.

On 16.2 the misclassified entry was only *listed* in the `.nft.json`, so it was harmless; the unconditional hashing added in 16.3 turns it into a build-breaking panic. The same `DirectoryEntry::File | DirectoryEntry::Symlink` pattern is still present on current canary, so canary is affected too.

Notes:

- `outputFileTracingExcludes` does **not** work around this — the panic happens while hashing the includes collection, before excludes are relevant.
- In pnpm monorepos no unusual config is needed to hit this: any workspace package listed in `serverExternalPackages` whose `node_modules` contains a workspace link (pnpm creates `node_modules/@scope/pkg → ../../sibling` directory symlinks) crashes the same way. That's how we originally hit it in a production monorepo; the only workaround was removing the dependency so the symlink no longer exists.
- Previously reported in #96255 with the same diagnosis, but that issue was closed by the triage bot because the reproduction was inlined rather than linked. This issue provides the public repro repository.

lazerg:
When an `outputFileTracingIncludes` glob matches a symlink that points at a directory, `next build` dies with `Is a directory (os error 21)`. `get_glob_includes` treats every symlink as a file, and since 16.3 the `.nft.json` writer hashes the contents of every included file, so it ends up trying to read a directory.

The symlink is now resolved first and only kept if it points at a file. The glob walk already descends into symlinked directories and lists the files inside them, so nothing drops out of the trace. This also covers pnpm monorepos, where `node_modules` links to workspace packages hit the same path.

Fixes #96626

### DIFF
--- a/crates/next-api/src/nft.rs
+++ b/crates/next-api/src/nft.rs
@@ -195,9 +195,16 @@ async fn get_glob_includes(
     while let Some(glob_result) = stack.pop_back() {
         // Process direct results (files and directories at this level)
         for entry in glob_result.results.values() {
-            let (DirectoryEntry::File(file_path) | DirectoryEntry::Symlink(file_path)) = entry
-            else {
-                continue;
+            let file_path = match entry {
+                DirectoryEntry::File(file_path) => file_path,
+                DirectoryEntry::Symlink(symlink_path) => {
+                    match entry.clone().resolve_symlink().await? {
+                        DirectoryEntry::File(_) => symlink_path,
+                        // Symlinks to directories are traversed via `glob_result.inner`.
+                        _ => continue,
+                    }
+                }
+                _ => continue,
             };
 
             result.push(file_path.clone());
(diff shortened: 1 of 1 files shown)


lazerg:

lazerg:
Closing since the linked issue #96626 was already resolved elsewhere.

PR #96641, opened 2026-08-04
[ci] Fix create_release_branch for new repo permissions

lukesandberg:
### What

Fixes the **Create Release Branch** workflow, which currently fails on every run.

1. Drop `permission-environments: write` from the token request.
   - We can no longer request it, and due to repo policies we no longer need to mutate the environment
2. Drop the runtime `deployment-branch-policies` API call from `scripts/create-release-branch.js`.
   - No longer needed: the `release-stable` environment now statically allows branches conforming to `next-NN` / `next-NN-*`. This is what the permission above existed for.
3. Create the setup commit as a GitHub-signed commit via `scripts/github-utils/signed-commit.js` instead of `git push`.
   - Branch protection requires signed commits

### Why

The most recent run ([30923742579](https://github.com/vercel/next.js/actions/runs/30923742579/job/92040655448)) dies at the first step, `Create GitHub App token`, with a 422 — so the job never even clones the repo.

### DIFF
--- a/scripts/create-release-branch.js
+++ b/scripts/create-release-branch.js
@@ -6,15 +6,51 @@ const {
   configureGitHubAuth,
   getGitHubToken,
   getGitHubTokenMissingMessage,
-  verifyGitHubApiAccess,
 } = require('./release-github-auth')
 const {
+  githubRequest,
   createSignedCommit,
   upsertBranchRef,
 } = require('./github-utils/signed-commit')
 
 const REPO_OWNER = 'vercel'
 const REPO_NAME = 'next.js'
+
+/**
+ * Fail fast, before any file mutation, if the branch already exists.
+ *
+ * This doubles as the API preflight: it exercises the same git-data path the
+ * script later writes to (blobs/trees/commits/refs), so a token missing those
+ * grants fails here rather than midway through creating objects. A plain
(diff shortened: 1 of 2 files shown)


bgw (approved):

scripts/create-release-branch.js
```diff
@@ -89,44 +96,49 @@ async function main() {
...
+    baseSha: baseSha.trim(),
+    localCommitSha: localCommitSha.trim(),
+    message: commitMessage,
   })
 
-  console.log(`Waiting 5s before updating branch rules`)
-  await new Promise((resolve) => setTimeout(resolve, 5_000))
-
-  const updateEnvironmentRes = await fetch(
-    'https://api.github.com/repos/vercel/next.js/environments/release-stable/deployment-branch-policies',
```
eps1lon, line 110: We need that though.

eps1lon (changes requested):

scripts/create-release-branch.js
lukesandberg: @bgw changed the branch restrictions so the bot has access to `next-[0-9][0-9]` and `next-[0-9][0-9]-*` so we shouldn't need this

eps1lon (approved):

PR #96647, opened 2026-08-04
Warn when styled-jsx can't be deduplicated instead of failing silently

sokra:
### What?

`next/dist/server/require-hook` resolves `styled-jsx` requests to Next.js'
own copy so that the Pages Router renderer and user code share a single
`styled-jsx` module instance. If that resolution throws, the failure was
silently swallowed (`catch (_) {}`). This logs a warning with the underlying
error instead - the resolution still isn't fatal, it just isn't invisible
anymore.

### Why?

The renderer hands the styled-jsx style registry to user code through a
React context owned by that specific module instance - they have to share
one instance. If the aliases fail to register, user code loads its own
`styled-jsx` copy while the renderer loads a different one; `JSXStyle` then
finds no registry and silently renders nothing during SSR. The `jsx-*` class
names are still emitted, so the only symptom is a flash of unstyled content
on first load, with nothing in any log pointing at the cause.

### How?

Kept the `try/catch` non-fatal (an app that doesn't use styled-jsx doesn't
care if this fails), but log the caught error through `console.warn` so the
condition is diagnosable. Verified against a published Next.js install: the
normal (resolvable) case stays silent; an unresolvable case warns exactly
once with the real underlying error and does not throw.

### DIFF
--- a/errors/styled-jsx-not-resolvable.mdx
+++ b/errors/styled-jsx-not-resolvable.mdx
@@ -1,0 +1,49 @@
+---
+title: '`styled-jsx` is not resolvable'
+---
+
+## Why This Error Occurred
+
+Next.js could not resolve its own copy of `styled-jsx`.
+
+`styled-jsx` is a dependency of `next`, and Next.js resolves it eagerly at startup to make sure that
+your code and the Pages Router renderer use the _same_ `styled-jsx` instance: the renderer creates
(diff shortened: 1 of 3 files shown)


andrewimm:
What's the fix in this case? Force node to resolve a specific package, or next?
This should be resolved by adding a line to a config file somewhere -- is it possible to add that detail to the error message?

PR #96651, opened 2026-08-04
fix: preserve repeated --require/--import flags when forking workers

### LINKED ISSUE #96582: Repeated --require/--import in NODE_OPTIONS is corrupted when forking workers

https://github.com/colinhacks/nextjs-node-options-repro

### To Reproduce

```bash
git clone https://github.com/colinhacks/nextjs-node-options-repro
cd nextjs-node-options-repro
npm install
npm run repro
```

The script runs `next build` three times, varying only `NODE_OPTIONS`, and reports how many processes each preload actually ran in.

Node accepts a repeated `--require` and runs every one of them. Next reparses `NODE_OPTIONS` and reformats it for each forked worker, and that round-trip cannot represent two of the same flag.

Exact process counts vary slightly per run; the pattern does not.

| Case | `NODE_OPTIONS` | Outcome |
|---|---|---|
| 1 (control) | `--require=<a>` | Preload a runs in every process. Exit 0. |
| 2 | `--require=<a> --require=<b>` | Preload a is missing from the forked workers — b runs in all of them, a in only a few. **Exit 0.** |
| 3 | `--require <a> --require <b>` | Build fails, exit 1. |

Case 3 fails with both paths joined into one specifier:

```
Error: Cannot find module '/path/preload-a.cjs /path/preload-b.cjs'
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'internal/preload' ]
```

Case 2 is the more dangerous one, since the build succeeds and the missing preload is silent.

Expected in both: each preload runs in every process, as it does under plain `node`.

### Root cause

Parsing in `server/lib/utils.ts` stores Node options in a `Record` keyed by option name:

```ts
export type NodeOptions = Record<string, string | boolean | undefined>

const parseNodeArgs = (args: string[]): NodeOptions => {
  const { values, tokens } = parseArgs({ args, strict: false, tokens: true })
```

A `Record` cannot hold two `require` entries, and the two spellings lose the data differently.

The equals form keeps only the last value, because `parseArgs` with `strict: false` and no `options` config does not accumulate:

```js
parseArgs({ args: ["--require=A", "--require=B"], strict: false }).values
// { require: 'B' }
```

The space-separated form instead reaches the orphan branch, which concatenates:

```ts
values[orphan.name] += ` ${token.value}`
```

producing the single bogus specifier `A B`. `formatNodeOptions` then rebuilds the string from that `Record`, and `getFormattedNodeOptionsWithoutInspect` is what each forked worker receives.

### Suggested fix

Declaring the repeatable flags as arrays preserves them:

```js
parseArgs({
  args: ["--require=A", "--require=B"],
  strict: false,
  options: { require: { type: "string", multiple: true } },
}).values
// { require: [ 'A', 'B' ] }
```

`formatNodeOptions` would then emit one flag per array element. The same applies to `--import`, which reproduces identically.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 65536
  Available CPU cores: 10
Binaries:
  Node: 26.5.0
  npm: 11.17.0
  pnpm: 10.15.1
Relevant Packages:
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

Not sure

### Additional context

Affects any setup needing more than one preload — an APM agent alongside an env loader, or a runtime that injects its own `--require` and lets the user add their own.

Related, and the reason this is filed fresh rather than as a comment:

- #77550 is open and covers case 3 (the space-form crash) via New Relic plus dotenv. It does not cover case 2, the silent drop, and no root cause was identified there. I have also left this analysis on that thread.
- #67286 reported the same bug in 2024. It was closed by the stale bot with no human response and then locked; that bot message invites a new issue.

anujbolewar:
Repeated --require/--import flags in NODE_OPTIONS are corrupted when Next.js reparses and reformats the options for forked workers.

## What changed

- Declare `require` and `import` as repeatable options in `parseNodeArgs` so `parseArgs` accumulates every occurrence into an array instead of keeping only the last value (equals form) or joining the space form into one bogus specifier.
- Emit one flag per array element in `formatArg`/round-trip.
- Widened `NodeOptions` and `getParsedDebugAddress` to accept `string[]`.

## Why

Two preloads in `NODE_OPTIONS` (e.g. an APM agent plus an env loader) silently lose one in forked workers (`--require=a --require=b`), or crash the build (`--require a --require b`, joining paths into a single MODULE_NOT_FOUND). Node itself runs every occurrence; the round-trip through Next must preserve them.

Fixes #96582

## Validation

- Unit tests in `packages/next/src/server/lib/utils.test.ts` cover both spellings, repeated `--import`, quoted values, and array formatting (22 passing).
- `pnpm --filter=next types` clean.

### DIFF
--- a/packages/next/src/server/lib/utils.test.ts
+++ b/packages/next/src/server/lib/utils.test.ts
@@ -77,6 +77,19 @@ describe('formatNodeOptions', () => {
       ],
     })
   })
+
+  it('emits one flag per element for array values', () => {
+    const result = formatNodeOptions({
+      require: ['/path/a.cjs', '/path/b.cjs'],
+      import: '/path/c.mjs',
+    })
+
+    expect(result).toEqual({
+      execArgv: [],
+      nodeOptions:
+        '--require=/path/a.cjs --require=/path/b.cjs --import=/path/c.mjs',
+    })
+  })
 })
 
 describe('getMemoryRestartStats', () => {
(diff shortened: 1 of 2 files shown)


andrewimm (approved):
Because of the explicit `REPEATABLE_OPTIONS` check, this should be safe and not break anyone else's setup unintentionally

andrewimm:
@anujbolewar thanks for this contribution. Can you rebase off of canary and force-push to your PR branch? The Next.js requires all commits to be signed, so I can't automatically rebase from the GitHub UI.

anujbolewar:
Done — rebased off the latest canary and force-pushed. The commit is still signed. Let me know if anything else is needed.

PR #96656, opened 2026-08-04
[turbopack] Don't run Webpack tests on Turbopack-only changes

sampoder:
Hopefully this saves us money + makes it easier to merge Turbopack PRs due to flaky Webpack tests


bgw (approved):
Thanks for taking initiative and doing this!

PR #96660, opened 2026-08-04
Remove App Page execution mode reads

ztanner:
## Summary

App Page leaf logic still reads the Work Store execution mode after request rendering and prerendering have already been separated at the entrypoint.

This change removes those remaining reads by:

- Deriving component-tree behavior from the active Work Unit type.
- Relying on request-only boundaries for dynamic Flight payloads and Server Actions.
- Selecting request or prerender instrumentation context from the corresponding render pipeline.

After this change, App Page only supplies `executionMode` when constructing a Work Store; none of its downstream behavior reads it. Removing the field itself remains a follow-up once App Route and cache consumers have been migrated.

### DIFF
--- a/packages/next/src/server/app-render/create-component-tree.tsx
+++ b/packages/next/src/server/app-render/create-component-tree.tsx
@@ -22,7 +22,10 @@ import { getTracer } from '../lib/trace/tracer'
 import { NextNodeServerSpan } from '../lib/trace/constants'
 import { StaticGenBailoutError } from '../../client/components/static-generation-bailout'
 import type { Params } from '../request/params'
-import { workUnitAsyncStorage } from './work-unit-async-storage.external'
+import {
+  workUnitAsyncStorage,
+  type WorkUnitStore,
+} from './work-unit-async-storage.external'
 import {
   createVaryParamsAccumulator,
   emptyVaryParamsAccumulator,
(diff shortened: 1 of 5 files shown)


packages/next/src/server/app-render/create-component-tree.tsx
```diff
@@ -137,7 +170,9 @@ async function createComponentTreeInternal(
     query,
   } = ctx
 
-  const { executionMode } = workStore
+  // A WorkStore can span several request and prerender work units during
+  // staged rendering, so component-tree behavior follows the active work unit.
+  const isPrerendering = isPrerenderWorkUnit(workUnitStore)
```
gnoff: is the workUnitStore really optional? when would we not have it?

gnoff: shouldn't we just take in isPrerendering as an argument to the createComponentTree call?

ztanner: yeah good call, that simplifies it a bit. Pushed

gnoff (approved):
Another refactor we should consider is splitting createComponentTree by CC. There are a ton of legacy code paths in here that limit how tight we can make this function. the revalidate assignment for instance would just go away if we didn't have to consider non-CC. It would of course increase the surface area for common changes but I think there is benefit to narrowing the "context" of complex functions like these even if it means making double changes sometimes

PR #96670, opened 2026-08-04
Remove cache revalidation execution mode reads

ztanner:
## Summary

- derive foreground stale-cache revalidation from Work Unit state instead of `WorkStore.executionMode`
- propagate that specific policy through `"use cache"` and `unstable_cache` scopes so nested cache and fetch calls preserve their outer behavior
- add coverage for foreground ISR revalidation through nested `unstable_cache` scopes

This is intended to be a behavior-preserving refactor.

### DIFF
--- a/packages/next/src/server/app-render/work-unit-async-storage.external.ts
+++ b/packages/next/src/server/app-render/work-unit-async-storage.external.ts
@@ -342,6 +342,12 @@ export type StaticPrerenderStore = Exclude<
 export interface CommonCacheStore
   extends Omit<CommonWorkUnitStore, 'implicitTags'> {
   /**
+   * Whether a stale cache entry must be revalidated before the outer work unit
+   * can continue. This is copied from the outer work unit because entering a
+   * cache scope intentionally hides that store.
+   */
+  readonly shouldRevalidateStaleCacheEntryInForeground: boolean
+  /**
    * A cache work unit store might not always have an outer work unit store,
    * from which implicit tags could be inherited.
    */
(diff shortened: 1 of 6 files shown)


packages/next/src/server/app-render/work-unit-async-storage.external.ts
```diff
@@ -341,6 +341,12 @@ export type StaticPrerenderStore = Exclude<
 
 export interface CommonCacheStore
   extends Omit<CommonWorkUnitStore, 'implicitTags'> {
+  /**
+   * Whether a stale cache entry must be revalidated before the outer work unit
+   * can continue. This is copied from the outer work unit because entering a
+   * cache scope intentionally hides that store.
+   */
+  readonly shouldRevalidateStaleCacheEntryInForeground: boolean
```
gnoff, line 349: I'd prefer we model this as "will the result of this cache entry be consumed by another cache scope" because it explains why stale entries are revalidated eagerly (we don't want to persist a stale response into another cache that extends it's lifetime)

maybe `consumerWillCache`

To be clear the prerender store types must still set this to true because a prerender is a form of a cache

packages/next/src/server/app-render/work-unit-async-storage.external.ts
```diff
@@ -439,6 +445,33 @@ export type WorkUnitStore =
   | PrerenderStore
   | GenerateStaticParamsStore
 
+export function shouldRevalidateStaleCacheEntryInForeground(
```
gnoff, line 448: `willConsumerCache()`?

gnoff (approved):

packages/next/src/server/app-render/work-unit-async-storage.external.ts
ztanner: yea that framing makes more sense to me. one thing i want to make sure i understand: should this describe the immediate consumer rather than propagate through the whole outer chain?
for request -> cache A -> cache B, B’s consumer is A, so I think B should see `consumerWillCache: true` even though A’s consumer is a request. the current implementation propagates `false` there, which could allow A to persist B’s stale result. 

is that behavior you’re intending to change, or were you thinking of this as naming-only?

ztanner: I think it's worth pulling this into a separate PR since it's a correctness fix (from canary behavior) https://github.com/vercel/next.js/pull/96731

gnoff: >even though A’s consumer is a request. the current implementation propagates false there, which could allow A to persist B’s stale result.

it only needs to look up 1 level. if A was stale we would be producing a new entry for A in background and so B will have consumerWillCache: true. But while evaluating A we don't want to force it to revalidate eagerly because the request is dynamic. So A will have consumerWillCache: false

PR #96676, opened 2026-08-04
turbo-tasks-backend: let a GC pass wind down when it is blocking an operation

lukesandberg:
A GC pass holds total operation exclusion, so every invalidation, cell update and child connect
blocks for its whole duration. In dev that is an HMR edit stalled behind the cascade, and the pass
had no way to notice or yield.

This makes a gc pass interruptible so we are guaranteed to not stall too much work.  To ensure progress we have a minimum guaranteed execution time `100ms`

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
@@ -284,7 +284,6 @@ impl TurboTasksBackend {
 
         stats.gc_roots = roots.len();
         stats.aged_out_roots = aged_out_count;
-        // No budget means an uninterruptible pass, which by construction never interrupts.
         stats.interrupted = budget
             .as_ref()
             .is_some_and(|budget| budget.was_interrupted());
@@ -385,9 +384,7 @@ impl TurboTasksBackend {
         );
         let _serialize = self.snapshot_in_progress.lock();
         let phase = self.snapshot_coord.begin_snapshot();
-        // Uninterruptible: the exact-count tests this hook exists for need a full pass, and an
-        // interrupted one would silently collect less than they assert.
-        let (stats, roots) = self.gc_collect(turbo_tasks, &phase, false);
+        let (stats, roots) = self.gc_collect(turbo_tasks, &phase, /* interruptible= */ false);
 
         // Persist the roots map this pass produced. Some tests query the roots set and GC itself
         // does as well, this ensures it is available to the next cycle.
--- a/turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
@@ -122,26 +122,6 @@ pub(crate) struct GcStats {
     pub interrupted: bool,
 }
 
-/// Test-only snapshot of the most recent [`GcStats`]. Production reads these off the `gc` span; a
-/// unit test can't, so [`TurboTasksBackend`] stashes this after each pass. Kept as a small `Copy`
-/// struct so it can be cloned out from behind the mutex cheaply.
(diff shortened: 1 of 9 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -147,13 +190,30 @@ impl TurboTasksBackend {
...
             (0..self.storage.shard_count())
                 .map(GcJob::ScanShard)
                 .chain(aged_out.into_iter().map(GcJob::Collect)),
             GcStats::default,
             |spawner, job, stats| {
+                // Abort the gc loop if we are interrupted
+                if let Some(budget) = &budget
+                    && budget.should_stop()
+                {
+                    return ControlFlow::Break(());
```
lukesandberg, line 215: could we make this cuter using the `?` operator

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -384,28 +446,35 @@ impl TurboTasksBackend {
...
+                return TestSnapshotOutcome::default();
             }
         };
-        let counts = self.storage.evict_after_snapshot(None);
-        (had_new_data, counts)
+        let eviction_counts = self.storage.evict_after_snapshot(None);
+        TestSnapshotOutcome {
+            had_new_data,
+            eviction_counts,
+            gc: gc_stats.map(|s| (s.collected, s.interrupted)),
```
lukesandberg: just pass the GcStats object

turbopack/crates/turbo-tasks-backend/tests/gc_interrupt.rs
```diff
@@ -0,0 +1,129 @@
...
+            anyhow::Ok(())
+        })
+        .await
+        .unwrap();
+    });
+
+    let outcome = tt.backend().snapshot_and_evict_for_testing(&tt);
+    waiter.await.unwrap();
+
+    let (collected, interrupted) = outcome.gc_stats();
```
lukesandberg: assert the number of collected tasks

comment that this test is just triggering a race, since we don't know the operation will be running concurrently with gc

turbopack/crates/turbo-tasks-backend/tests/gc_interrupt.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    build_generation(&tt, 0).await;
+
+    // Phase 1: accumulate garbage under passes that keep interrupting.
+    const ROUNDS: u32 = 10;
+    let mut collected_while_interrupting = 0usize;
+    let mut interrupted_rounds = 0usize;
+    for gen_value in 1..=ROUNDS {
+        build_generation(&tt, gen_value).await;
+        let (collected, interrupted) = tt.backend().snapshot_and_evict_for_testing(&tt).gc_stats();
```
lukesandberg: not clear exactly how we are forcing the two phases to run concurrently

turbopack/crates/turbo-tasks-backend/tests/gc_interrupt.rs
```diff
@@ -0,0 +1,129 @@
...
+    println!(
+        "self-healing: collected_while_interrupting={collected_while_interrupting} \
+         healed={healed} total_collected={total_collected} produced={produced} \
+         interrupted_rounds={interrupted_rounds}/{ROUNDS}"
+    );
+
+    // Every generation but the live one is garbage, so most of it must be accounted for.
+    let expected_min = produced / 2;
+    assert!(
+        total_collected >= expected_min,
```
lukesandberg: if we expected everything to be collected, then this can be a tighter assert

bgw (approved):

PR #96681, opened 2026-08-04
fix(next/image): preserve image response after optimization

### LINKED ISSUE #96612: ImageResponse fails after an uncached next/image optimization

lukesandberg:
Fork PR #96621 by @ceolinwill, re-opened as a branch PR so the "when deployed" CI job can run — it requires Vercel deployment secrets that GitHub does not expose to pull requests from forks, so it can never pass on the original.

**The commit is unchanged and still authored by @ceolinwill.** Please credit them; #96621 should be closed in favor of this one.

### What?

`getSharp()` calls `_sharp.block({ operation: ['VipsForeignLoad'] })` to block every image loader, then unblocks a specific allowlist — and that allowlist omitted `VipsForeignLoadSvg`. Because `_sharp` is a module-level singleton, the block is process-wide and permanent: the first `/_next/image` request in a process permanently disables Sharp's SVG loader.

`ImageResponse` (`next/og`) rasterizes via resvg and then hands SVG to Sharp, so once that loader is blocked it fails. The symptom is worse than a bad image — the render throws `Input buffer contains unsupported image format`, which surfaces as a **socket hang up / crashed response** on any `ImageResponse` route requested after an uncached image optimization in the same process.

Introduced by #96301, which aligned the Sharp allowlist with `detectContentType()` but missed SVG.

Fixes #96612

### How?

Adds `'VipsForeignLoadSvg'` to the unblock list.

This does not weaken SVG protections for user-supplied images. `detectContentType()` already returns SVG, and untrusted SVG is gated separately by `dangerouslyAllowSVG`, which throws a 400 in `imageOptimizer()` before Sharp is ever invoked. Unblocking the loader only restores Sharp's ability to process SVG that Next.js itself generates.

### Tests

The existing `og-api` test is extended to reproduce the exact same-process ordering: render `/og-node`, perform an uncached `/_next/image` optimization (asserting `x-nextjs-cache: MISS`, with a random URL so it can't be served from cache), then render `/og-node` again.

Verified locally that this fails without the one-line fix (`socket hang up`, caused by `Input buffer contains unsupported image format`) and passes with it.

No regressions in the SVG security coverage — `test/e2e/image-optimizer/dangerously-allow-svg.test.ts` passes 94/94 including all `maintain vector svg` / blur-svg cases, and `test/e2e/image-optimizer/image-optimizer.test.ts` plus the `og-*` suites are green.

### DIFF
--- a/test/e2e/og-api/index.test.ts
+++ b/test/e2e/og-api/index.test.ts
@@ -45,7 +45,10 @@ describe('og-api', () => {
       `/_next/image?url=${imageUrl}&w=640&q=75`
     )
     expect(optimized.status).toBe(200)
-    expect(optimized.headers.get('x-vercel-cache') || optimized.headers.get('x-nextjs-cache')).toBe('MISS')
+    expect(
+      optimized.headers.get('x-vercel-cache') ||
+        optimized.headers.get('x-nextjs-cache')
+    ).toBe('MISS')
 
     const after = await fetchViaHTTP(next.url, '/og-node')
     expect(after.status).toBe(200)
--- a/test/e2e/og-api/index.test.ts
+++ b/test/e2e/og-api/index.test.ts
@@ -1,5 +1,6 @@
 import { FileRef, nextTestSetup } from 'e2e-utils'
 import { fetchViaHTTP, renderViaHTTP } from 'next-test-utils'
+import { randomUUID } from 'crypto'
 import fs from 'fs-extra'
 import { join } from 'path'
 
@@ -32,12 +33,24 @@ describe('og-api', () => {
     expect(body.size).toBeGreaterThan(0)
   })
 
-  it('should work in app route in node runtime', async () => {
-    const res = await fetchViaHTTP(next.url, '/og-node')
-    expect(res.status).toBe(200)
-    expect(res.headers.get('content-type')).toContain('image/png')
-    const body = await res.blob()
-    expect(body.size).toBeGreaterThan(0)
+  it('should work in app route in node runtime after image optimization', async () => {
+    const before = await fetchViaHTTP(next.url, '/og-node')
+    expect(before.status).toBe(200)
+    expect(before.headers.get('content-type')).toContain('image/png')
+    expect((await before.blob()).size).toBeGreaterThan(0)
+
+    const imageUrl = encodeURIComponent(`/pixel/${randomUUID()}`)
+    const optimized = await fetchViaHTTP(
+      next.url,
+      `/_next/image?url=${imageUrl}&w=640&q=75`
+    )
+    expect(optimized.status).toBe(200)
+    expect(optimized.headers.get('x-nextjs-cache')).toBe('MISS')
+
+    const after = await fetchViaHTTP(next.url, '/og-node')
+    expect(after.status).toBe(200)
+    expect(after.headers.get('content-type')).toContain('image/png')
+    expect((await after.blob()).size).toBeGreaterThan(0)
   })
 
   it('should work in middleware', async () => {
--- a/test/e2e/og-api/index.test.ts
+++ b/test/e2e/og-api/index.test.ts
@@ -45,7 +45,7 @@ describe('og-api', () => {
       `/_next/image?url=${imageUrl}&w=640&q=75`
     )
     expect(optimized.status).toBe(200)
-    expect(optimized.headers.get('x-nextjs-cache')).toBe('MISS')
+    expect(optimized.headers.get('x-vercel-cache') || optimized.headers.get('x-nextjs-cache')).toBe('MISS')
 
     const after = await fetchViaHTTP(next.url, '/og-node')
     expect(after.status).toBe(200)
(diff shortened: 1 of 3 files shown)


styfle (approved):

styfle (approved):
Approving per my previous approval here:

https://github.com/vercel/next.js/pull/96621#pullrequestreview-4855462666

test/e2e/og-api/index.test.ts
```diff
@@ -32,12 +33,24 @@ describe('og-api', () => {
...
+    expect(before.headers.get('content-type')).toContain('image/png')
+    expect((await before.blob()).size).toBeGreaterThan(0)
+
+    const imageUrl = encodeURIComponent(`/pixel/${randomUUID()}`)
+    const optimized = await fetchViaHTTP(
+      next.url,
+      `/_next/image?url=${imageUrl}&w=640&q=75`
+    )
+    expect(optimized.status).toBe(200)
+    expect(optimized.headers.get('x-nextjs-cache')).toBe('MISS')
```
styfle:
```suggestion
    expect(optimized.headers.get('x-vercel-cache') || optimized.headers.get('x-nextjs-cache')).toBe('MISS')
```

test/e2e/og-api/index.test.ts
```diff
@@ -32,12 +33,24 @@ describe('og-api', () => {
...
+    expect(before.headers.get('content-type')).toContain('image/png')
+    expect((await before.blob()).size).toBeGreaterThan(0)
+
+    const imageUrl = encodeURIComponent(`/pixel/${randomUUID()}`)
+    const optimized = await fetchViaHTTP(
+      next.url,
+      `/_next/image?url=${imageUrl}&w=640&q=75`
+    )
+    expect(optimized.status).toBe(200)
+    expect(optimized.headers.get('x-vercel-cache') || optimized.headers.get('x-nextjs-cache')).toBe('MISS')
```
styfle:
```suggestion
    expect(
      optimized.headers.get('x-vercel-cache') ||
        optimized.headers.get('x-nextjs-cache')
    ).toBe('MISS')
```

styfle (approved):

PR #96685, opened 2026-08-04
docs: add TruoCloud to the example loader configurations

ramphy:
Adds [TruoCloud](https://docs.truo.cloud/images) to the list of example loader
configurations in `images.mdx`, in the same shape as the entries around it: one
bullet in the index and one code block.

Docs-only, 12 lines added, nothing removed.

The one thing worth a comment in the snippet — and it is in there — is that the
parameters are set in alphabetical order deliberately. TruoCloud treats the
sorted form as canonical, so an unsorted loader would produce a second CDN cache
entry for every image without changing what is served.

There is also a published loader package, `@truocloud/img-next`, which adds
per-segment RFC 3986 encoding of the source path and optional URL signing. I
kept the example self-contained to match the other entries.

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/images.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/images.mdx
@@ -66,6 +66,7 @@ To learn more about configuring the behavior of the built-in [Image Optimization
 - [Sirv](#sirv)
 - [Supabase](#supabase)
 - [Thumbor](#thumbor)
+- [TruoCloud](#truocloud)
 - [Imagekit](#imagekitio)
 - [Nitrogen AIO](#nitrogen-aio)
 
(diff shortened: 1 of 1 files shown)


docs/01-app/03-api-reference/05-config/01-next-config-js/images.mdx
```diff
@@ -243,6 +244,22 @@ export default function thumborLoader({ src, width, quality }) {
 }
 ```
 
+### TruoCloud
+
+```js
+// Docs: https://docs.truo.cloud/images
```
sampoder, line 250: fyi: this 404-s at the moment

ramphy: Thanks — you were right, and it was worse than the one URL: the docs deploy for
that group had failed outright, so every page under it was 404, not just this
one. Fixed and deployed:

```
https://docs.truo.cloud/images  →  200
```

The snippet in this PR is unchanged.

PR #96686, opened 2026-08-04
Serialize frozen collections by value only

wbinnssmith:
## What?

Serialize `FrozenMap` as a map and `FrozenSet` as a sequence instead of exposing their internal fields.

Add JSON shape and round-trip coverage for both collection types.

## Why?

Keep serialized formats tied to collection semantics rather than their current storage representation. This matches Serde's implementations for `HashMap` and `HashSet`, which serialize through `collect_map` and `collect_seq` respectively.

### DIFF
--- a/turbopack/crates/turbo-frozenmap/src/map.rs
+++ b/turbopack/crates/turbo-frozenmap/src/map.rs
@@ -9,10 +9,7 @@ use std::{
 
 use bincode::{BorrowDecode, Decode, Encode};
 use indexmap::IndexMap;
-use serde::{
-    Deserialize, Serialize,
-    de::{MapAccess, Visitor},
-};
+use serde::{Deserialize, Serialize};
 
 /// A compact frozen (immutable) ordered map backed by a sorted boxed slice.
 ///
--- a/turbopack/crates/turbo-frozenmap/src/map.rs
+++ b/turbopack/crates/turbo-frozenmap/src/map.rs
@@ -4,12 +4,16 @@ use std::{
     fmt::{self, Debug},
     hash::BuildHasher,
     iter::FusedIterator,
+    marker::PhantomData,
     ops::{Bound, Index, RangeBounds},
 };
 
 use bincode::{BorrowDecode, Decode, Encode};
 use indexmap::IndexMap;
-use serde::{Deserialize, Serialize};
+use serde::{
+    Deserialize, Serialize,
+    de::{MapAccess, Visitor},
+};
 
 /// A compact frozen (immutable) ordered map backed by a sorted boxed slice.
 ///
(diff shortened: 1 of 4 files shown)


turbopack/crates/turbo-frozenmap/src/map.rs
```diff
@@ -49,6 +49,23 @@ pub struct FrozenMap<K, V> {
...
+}
+
+impl<'de, K, V> Deserialize<'de> for FrozenMap<K, V>
+where
+    K: Deserialize<'de> + Ord,
+    V: Deserialize<'de>,
+{
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        // Reuse BTreeMap's map visitor; its sorted representation converts without another sort.
+        BTreeMap::deserialize(deserializer).map(Into::into)
```
lukesandberg: why not just serialize/deserializing as a vec? the deserialize copy here will churn a lot of memory but collecting a vec and then moving it into a map is free

lukesandberg: oh its for json?  ok that makes sense, but still going through BTreemap is expensive, surely there is a more straightforward way to collect the entries

turbopack/crates/turbo-frozenmap/src/set.rs
```diff
@@ -46,6 +46,23 @@ pub struct FrozenSet<T> {
...
+        serializer.collect_seq(self.iter())
+    }
+}
+
+impl<'de, T> Deserialize<'de> for FrozenSet<T>
+where
+    T: Deserialize<'de> + Ord,
+{
+    fn deserialize<D: serde::Deserializer<'de>>(deserializer: D) -> Result<Self, D::Error> {
+        // Reuse BTreeSet's sequence visitor; its sorted representation converts without another
```
lukesandberg: same basic concern, BtreeSet will construct a tree and then we flatten it

it would be better to collect a vec and then into_boxed_slice it

lukesandberg (approved):

bgw (changes requested):
I agree with @lukesandberg's comments

lukesandberg (approved):

turbopack/crates/turbo-frozenmap/src/map.rs
```diff
@@ -812,6 +854,18 @@ impl<K, V> Clone for Range<'_, K, V> {
...
     use super::*;
 
+    #[test]
+    fn serde_uses_map_shape() {
+        let map = FrozenMap::from([("b", 2), ("a", 1)]);
+        let json = serde_json::to_string(&map).unwrap();
+
+        assert_eq!(json, r#"{"a":1,"b":2}"#);
+        assert_eq!(
+            serde_json::from_str::<FrozenMap<&str, i32>>(&json).unwrap(),
```
bgw: It'd be good to use the intentionally-wrong order here, like you do when testing `serde_json::to_string`.

```suggestion
            serde_json::from_str::<FrozenMap<&str, i32>>(r#"{"b":2, "a":1}"#).unwrap(),
```

turbopack/crates/turbo-frozenmap/src/set.rs
```diff
@@ -46,6 +50,42 @@ pub struct FrozenSet<T> {
...
+            fn expecting(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+                formatter.write_str("a sequence")
+            }
+
+            fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<Self::Value, A::Error> {
+                let mut items = Vec::with_capacity(seq.size_hint().unwrap_or(0));
+                while let Some(item) = seq.next_element()? {
+                    items.push(item);
+                }
+                Ok(items.into_iter().collect())
```
bgw: There's some awkwardness with `FrozenSet` because it's defined in terms of `FrozenMap`, for the sake of trying to keep the implementation a little simpler.

It'd be more efficient to build a `Vec<(T, ())>` instead of `Vec<T>`. Then directly construct the `FrozenSet` using `FrozenSet { map: FrozenMap::from(items) }`

turbopack/crates/turbo-frozenmap/src/set.rs
```diff
@@ -355,6 +395,15 @@ impl<T> Clone for Range<'_, T> {
 mod tests {
     use super::*;
 
+    #[test]
+    fn serde_uses_sequence_shape() {
+        let set = FrozenSet::from([2, 1]);
+        let json = serde_json::to_string(&set).unwrap();
+
+        assert_eq!(json, "[1,2]");
+        assert_eq!(serde_json::from_str::<FrozenSet<i32>>(&json).unwrap(), set);
```
bgw:
```suggestion
        assert_eq!(serde_json::from_str::<FrozenSet<i32>>("[2, 1]").unwrap(), set);
```

bgw (approved):

PR #96695, opened 2026-08-05
Serialize prerender manifest updates in dev

### LINKED ISSUE #96664: next dev: concurrent first requests to multiple generateStaticParams routes corrupt prerender-manifest.json (non-atomic read-modify-write), 500s persist until restart

https://github.com/adrianhRV/next-manifest-race-repro

### To Reproduce

1. `npm install`, then start a cold dev server: `rm -rf .next && npm run dev` (port 3100)
2. Once ready, run `./hammer.sh` — it fires 20 parallel requests at each of the 4 routes, then probes each route serially
3. The race is probabilistic: repeat from step 1 if probes return 200. It typically tears within 1–5 cold attempts (~30s each). Verify the manifest directly:

```bash
node -e "try{JSON.parse(require('fs').readFileSync('.next/dev/prerender-manifest.json','utf8'));console.log('VALID')}catch(e){console.log('CORRUPT:',e.message)}"
```

The app is intentionally minimal: four dynamic routes, each a static JSX page with a one-entry `generateStaticParams`, empty `next.config.ts`, no data fetching, no cache directives. Also reproduces with `cacheComponents: true` and `"use cache"` functions — neither is required.

**Current:** once the race lands, every route in the app returns 500 with:

```
⨯ SyntaxError: Unexpected non-whitespace character after JSON at position 7 (line 1 column 8)
    at JSON.parse (<anonymous>) {
  page: '/methodology/review-methodology'
}
```

and `.next/dev/prerender-manifest.json` on disk is durably invalid — a shorter document written over a longer one without truncation, leaving the old tail behind:

```
...previewModeEncryptionKey":"f1e0a...b49a"}}dc35111d9bb640e093f00c368fea9ef07fdb49a"}}
```

The 500s persist until the dev server restarts (boot rewrites the manifest). A milder variant occurs when a reader catches the file mid-write (`SyntaxError: Unexpected end of JSON input`); those requests 500 but the server recovers. The error position is stable per corruption (it equals the length of the winning document) and varies between corruptions.

**Expected:** concurrent first requests to dynamic routes render successfully and never corrupt dev state.

### Root cause (from reading `next/dist/server/dev/next-dev-server.js`)

`DevServer.loadStaticPaths` performs an unserialized, non-atomic read-modify-write of `PRERENDER_MANIFEST` when a dynamic route's static paths are first resolved (v16.3.1-canary.1, ~lines 658–691):

```js
const rawExistingManifest = await fs.promises.readFile(join(this.distDir, PRERENDER_MANIFEST), 'utf8');
const existingManifest = JSON.parse(rawExistingManifest);
// ...merge this route's staticPaths and dynamicRoutes entry...
const updatedManifest = JSON.stringify(existingManifest);
if (updatedManifest !== rawExistingManifest) {
  await fs.promises.writeFile(join(this.distDir, PRERENDER_MANIFEST), updatedManifest);
}
```

With N dynamic routes cold-rendering concurrently, N of these RMWs race on one file:

- **Torn file (sticky):** two writers' `open(O_TRUNC)` both happen before either write; the shorter document lands over the longer one, leaving the old tail. `loadManifest` (`load-manifest.external.js`) only caches on successful parse, so `AppPageRouteModule.prepare` re-reads and re-throws on every request, for every route.
- **Torn read (transient):** a concurrent reader parses the file between truncate and write-complete.
- **Lost update (silent):** two RMWs read the same base and the last write drops the other route's entries.

Suggested fix: write to a temp file and `rename()` into place (atomic on POSIX), and serialize manifest updates behind a single-writer queue. The read side could also tolerate a parse failure by falling back to in-memory state instead of failing the render.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 18432
  Available CPU cores: 11
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: 1.22.22
  pnpm: 10.28.2
Relevant Packages:
  next: 16.3.1-canary.1 // Latest available version is detected (16.3.1-canary.1).
  eslint-config-next: N/A
  react: 19.2.5
  react-dom: 19.2.5
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

Developer Experience

next dev (local)

### Additional context

- Found in a large app where Playwright workers fan out over a freshly started dev server; the same parallel-request pattern against 4 real routes corrupts the manifest on most cold runs.
- Reproduced on 16.2.6, 16.3.0, and 16.3.1-canary.1 (Turbopack dev). The racy code is unchanged on canary.
- The stable-per-corruption error position equals the length of the shorter (winning) document — a useful fingerprint for triaging duplicate reports.
- Workaround: warm each `generateStaticParams` route serially once after dev-server start before allowing parallel traffic; restart the dev server to clear an already-torn manifest.

lazerg:
Concurrent first requests to several `generateStaticParams` routes can corrupt `.next/dev/prerender-manifest.json`. `loadStaticPaths` reads the manifest, merges one route's entries, and writes the whole file back, and nothing serializes those read-modify-writes. When two writers truncate before either finishes, a short document lands on top of a long one. Every render after that throws `SyntaxError: Unexpected non-whitespace character after JSON`, so the dev server returns 500 for every route until it restarts. Even when the file stays parseable, routes that read the same base drop each other's entries.

This queues the updates behind a single promise chain and writes through `write-file-atomic`, so each update sees the previous one's result and readers never catch a half-written file.

Fixes #96664

### DIFF
--- a/packages/next/src/server/dev/next-dev-server.ts
+++ b/packages/next/src/server/dev/next-dev-server.ts
@@ -25,6 +25,7 @@ import type { PagesManifest } from '../../build/webpack/plugins/pages-manifest-p
 import * as React from 'react'
 import fs from 'fs'
 import { Worker } from 'next/dist/compiled/jest-worker'
+import writeFileAtomic from 'next/dist/compiled/write-file-atomic'
 import { installUseCacheProbe } from './use-cache-probe-pool'
 import { installDevValidationWorker } from './dev-validation-worker-pool'
 import { join as pathJoin } from 'path'
@@ -146,6 +147,7 @@ export default class DevServer extends Server {
   private staticPathsCache: LRUCache<
     UnwrapPromise<ReturnType<DevServer['getStaticPaths']>>
   >
+  private prerenderManifestUpdate: Promise<void> = Promise.resolve()
   private startServerSpan: Span
   private readonly serverComponentsHmrCache:
     | ServerComponentsHmrCache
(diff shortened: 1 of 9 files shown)


LouisTsang-jk:
This overlaps with #96384 (opened 07/30), which fixes the same read-modify-write with the same strategy — serialize the cycles behind a single promise chain, write atomically instead of in place. Flagging it so the two don't get reviewed independently.

Also worth linking: #96664, which this PR lists as the issue it fixes, is itself a duplicate of #96259 (07/27), where the mechanism and an independent third-party confirmation live.

The two PRs differ in ways that are a maintainer's call rather than something to race over:

- **This one** keeps the change inside `next-dev-server.ts`, reuses the vendored `next/dist/compiled/write-file-atomic`, and covers it with an e2e test that boots a real dev server against four concurrent `generateStaticParams` routes.
- **#96384** extracts a `PrerenderManifestUpdater` module holding the queue plus a PID-suffixed tmp file and `rename`, and covers it with 5 unit tests: call-order serialization, no partially written manifest ever visible to a reader, a failed cycle not poisoning the queue, no write when the manifest is unchanged, and tmp cleanup when the write fails.

The e2e test here covers something the unit tests don't (that the real dev server path is actually fixed), and the unit tests cover queue semantics an e2e test can't pin down deterministically. Happy to fold either set onto whichever PR a reviewer prefers, or to close #96384 if this one is the better base. The part that matters is that one of them lands — `16.3.1-canary.4` is still on the unsynchronized path.

lazerg:
> This overlaps with #96384 (opened 07/30), which fixes the same read-modify-write with the same strategy — serialize the cycles behind a single promise chain, write atomically instead of in place. Flagging it so the two don't get reviewed independently.
> 
> Also worth linking: #96664, which this PR lists as the issue it fixes, is itself a duplicate of #96259 (07/27), where the mechanism and an independent third-party confirmation live.
> 
> The two PRs differ in ways that are a maintainer's call rather than something to race over:

My PR is ready, but because of a limit, I cannot mark it as Open.

lazerg:

lazerg:
Closing since the linked issue #96664 was closed as a duplicate of #96259.

PR #96697, opened 2026-08-05
[turbopack] Raise registration calls in hoisted modules to the top

### LINKED ISSUE #96648: Turbopack scope hoisting breaks React context identity: useContext returns undefined for a provider mounted above the consumer

https://github.com/blockscout/frontend

### To Reproduce

Public repo, no signup or private packages needed. It is a large app rather than a minimal
reproduction — see "Additional context" for why we could not shrink it, and for what we did manage to
isolate instead.

```bash
git clone https://github.com/blockscout/frontend && cd frontend
pnpm install

# Fetch a live instance's public config; this one sets
# NEXT_PUBLIC_ACCOUNT_AUTH_PROVIDER=dynamic, which mounts @dynamic-labs' provider at the app root.
# (Needs outbound internet; writes .env.tmp.)
./tools/dev-server/fetch.sh rootstock_testnet

# Build with Turbopack. NOTE: `pnpm build` now passes --webpack (that is our workaround for this
# very bug), so invoke next build directly to get the Turbopack pipeline.
pnpm exec dotenv -e .env.tmp -- next build

pnpm exec dotenv -e .env.tmp -- bash -c './deploy/scripts/make_envs_script.sh && next start'
```

Open `http://localhost:3000/`. The page crashes during hydration, on every route, with the error
below in the console.

To confirm it is scope hoisting, add `experimental: { turbopackScopeHoisting: false }` to
`next.config.js` and rebuild — the page renders. Building with `next build --webpack` also renders.

**Current.** The build emits code in which a `useContext(...)` call no longer observes the provider
mounted above it. Inside `@dynamic-labs/sdk-react-core@4.91.4`, `useUpdateUserWithModal` does:

```js
const editorContext = useContext(UserFieldEditorContext);
if (!editorContext)
  throw new DynamicError('useUserUpdateRequest can only be used inside the context of DynamicContextProvider', 'missing_dynamic_context');
```

`editorContext` is `undefined` at runtime, so the SDK throws and the app dies:

```
Error: useUserUpdateRequest can only be used inside the context of DynamicContextProvider
    at <chunk>.js:28:106113
    at ap (<chunk>.js:25:46567)     <- the SDK's own SyncAuthFlow component
```

The consumer is **not** application code — it is the SDK's internal `SyncAuthFlow`, which the SDK's
own `DynamicContext` renders *inside* `UserFieldEditorContextProvider`
(`src/lib/context/DynamicContext/DynamicContext.js`, the provider tree around line 759). The provider
is therefore guaranteed to be an ancestor of the consumer, in the same React tree, from a single copy
of the package. `useContext` returning `undefined` there should not be reachable.

**Expected.** Hoisted output preserves context identity, so the consumer reads the value its ancestor
provider supplied — as it does in dev, and as it does with webpack.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin 25.5.0
Binaries:
  Node: 22.14.0
  pnpm: 11.5.1
Relevant Packages:
  next: 16.2.6 (also reproduced on 16.3.0)
  react: 19.1.4
  react-dom: 19.1.4
  @dynamic-labs/sdk-react-core: 4.91.4
```

Also reproduces in `linux/amd64` and `linux/arm64` container images built in CI, so it is not
platform-specific.

Turbopack

### Additional context

Pages Router.

**Isolation.** Three single-variable results, each verified by loading the built app in a real
browser and checking for the error, not by reading bundles. Same source, same lockfile, same runtime
config:

| Build | Result |
| --- | --- |
| `next build` (Turbopack) | crashes |
| `next build` + `experimental.turbopackScopeHoisting: false` | works |
| `next build --webpack` | works |

**Why there is no minimal reproduction.** We bisected this to one commit of ours that only reshaped
the module graph around that SDK — moving a provider behind `next/dynamic`, adding lazily
`import()`ed wrapper hooks, splitting some modules. No change to the provider tree itself; parent
commit good, that commit bad. We then tried to reduce the trigger and failed three times: reverting
the lazy `import()` wrappers, reverting the `_app.tsx` provider restructure, and removing a newly
added direct dependency each left the build broken. That is consistent with the bug depending on how
hoisting groups the graph as a whole rather than on any single import, which is also why we have not
been able to shrink it. We are aware a large app is not what the template asks for; we are filing
because the flag flip above is unambiguous and reproducible. Happy to invest in a smaller
reproduction if that is what unblocks a look.

We have not confirmed the precise defect in the emitted code — whether the context module ends up
duplicated across hoisted groups, or a binding is mis-resolved during renaming. We could not
establish that with confidence from minified output, so we are only claiming the observable: context
identity is not preserved.

**Why the flag is not a sufficient workaround for us.** `turbopackScopeHoisting: false` fixes the
crash but costs real performance — on our main page, medians of 3 traces: FCP 432 → 790 ms,
main-thread blocking 133 → 408 ms, with bytes-before-FCP essentially unchanged (1038 → 1064 KB). So
we have moved production builds to `--webpack`, which for us is both correct and a smaller bundle.
Filing because that leaves us off Turbopack for production indefinitely.

**Note on impact.** The crash is invisible in development (unminified, no hoisting), so it reached a
tagged release of ours and was caught only by running the release image by hand. A production-only
miscompile is not something an app's own test suite can catch.

Re-filed from #96645, which the triage bot closed and locked: I had put explanatory prose after the
URL in the reproduction section, and the bot uses that section verbatim as the link. My mistake — the
link is unchanged, this time on its own.

sampoder:
Closes https://github.com/vercel/next.js/issues/96648, take a look at code like this:

image

Line 26 of scope-hoisting group A enters scope-hoisting group B, then on line 95 we re-enter scope-hoisting group A. Because our first execution of group A hadn't reached Line 29 yet to register schemas.js (which B depends on schemas.js).

On non-scope hoisted modules with cycles we already raise the module registration call to the start of the factory. But when we scope hoist, we lose that. This PR changes it so we do the registration at the start of the scope hoisted module

Before:

```javascript
"shared.js", "library/src/schemas.js", "library/src/index.js <export * as z>",
((__turbopack_context__) => {
"use strict";

// MERGED MODULE: shared.js
;
var …errors = __turbopack_context__.i("library/src/errors.js");
// MERGED MODULE: library/src/index.js <export * as z>
;
// MERGED MODULE: library/src/index.js
;
var …index_locals = __turbopack_context__.i("library/src/index.js <locals>");  // ← leaves the group
// MERGED MODULE: library/src/schemas.js
;
__turbopack_context__.s([                        // ← schemas registers HERE, too late
    "any", ()=>any,
    "anyBase", ()=>anyBase,
    "instance", ()=>instance,
    "instanceFromClosure", ()=>instanceFromClosure
], "library/src/schemas.js");
var …iso = __turbopack_context__.i("library/src/iso.js");
```

After:

```javascript
"shared.js", "library/src/schemas.js", "library/src/index.js <export * as z>",
((__turbopack_context__) => {
"use strict";

__turbopack_context__.s([                        // ← raised to the top of the merged module
    "any", ()=>any,
    "anyBase", ()=>anyBase,
    "instance", ()=>instance,
    "instanceFromClosure", ()=>instanceFromClosure
], "library/src/schemas.js");
// MERGED MODULE: shared.js
;
var …errors = __turbopack_context__.i("library/src/errors.js");
// MERGED MODULE: library/src/index.js <export * as z>
;
// MERGED MODULE: library/src/index.js
;
var …index_locals = __turbopack_context__.i("library/src/index.js <locals>");  // ← re-entry is safe
// MERGED MODULE: library/src/schemas.js
;
var …iso = __turbopack_context__.i("library/src/iso.js");
```

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/lib.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/lib.rs
@@ -61,7 +61,7 @@ use swc_core::{
     base::SwcComments,
     common::{
         BytePos, DUMMY_SP, FileName, GLOBALS, Globals, Loc, Mark, SourceFile, SourceMap,
-        SourceMapper, Span, SpanSnippetError, SyntaxContext,
+        SourceMapper, Span, SpanSnippetError, Spanned, SyntaxContext,
         comments::{Comment, CommentKind, Comments},
         source_map::{FileLinesResult, Files, SourceMapLookupError},
         util::take::Take,
@@ -1235,6 +1235,41 @@ impl EcmascriptModuleContent {
     }
 }
 
+/// Comments delimiting the early hoisted statements, which [`merge_modules`] moves in front of the
+/// merged module so that a cyclic importer can't re-enter it before they ran.
+const EARLY_HOIST_START: &str = " TURBOPACK EARLY HOIST START";
+const EARLY_HOIST_END: &str = " TURBOPACK EARLY HOIST END";
+
+fn early_hoist_comment(text: &str) -> Comment {
+    Comment {
+        kind: CommentKind::Line,
+        span: DUMMY_SP,
+        text: text.into(),
+    }
+}
+
+/// Finds the statements delimited by [`EARLY_HOIST_START`] and [`EARLY_HOIST_END`], as an inclusive
+/// index range over `body` covering both delimiters. Must run before the spans are rewritten.
+fn early_hoist_range(
+    comments: &SwcComments,
+    body: impl Iterator<Item = Span>,
+) -> Option<(usize, usize)> {
+    let (mut start, mut end) = (None, None);
+    for (i, span) in body.enumerate() {
+        let Some(leading) = comments.get_leading(span.lo) else {
+            continue;
+        };
+        for comment in leading {
+            if comment.text == EARLY_HOIST_START {
+                start = Some(i);
+            } else if comment.text == EARLY_HOIST_END {
+                end = Some(i);
+            }
+        }
+    }
+    Some((start?, end?))
+}
+
 /// Merges multiple Ecmascript modules into a single AST, setting the syntax contexts correctly so
 /// that imports work.
 ///
@@ -1404,18 +1439,32 @@ async fn merge_modules(
         let mut unique_contexts_cache =
             FxHashMap::with_capacity_and_hasher(contents.len() * 5, Default::default());
 
+        let mut merged_prelude = Vec::new();
         let mut prepare_module =
             |module_count: usize,
              current_module_idx: usize,
              (module, content): &(ResolvedVc<Box<dyn EcmascriptChunkPlaceable>>, CodeGenResult),
              program: &mut Program,
+             merged_prelude: &mut Vec<ModuleItem>,
              lookup_table: &mut Vec<ModulePosition>| {
                 let _ = tracing::trace_span!("prepare module").entered();
                 if let CodeGenResult {
                     scope_hoisting_syntax_contexts: Some((module_contexts, _)),
+                    comments: CodeGenResultComments::Single { extra_comments, .. },
                     ..
                 } = content
                 {
+                    // The delimiter comments are keyed by the original spans, so this has to happen
+                    // before the visitor below rewrites them.
+                    let early_hoisted = match &*program {
+                        Program::Module(module) => {
+                            early_hoist_range(extra_comments, module.body.iter().map(|i| i.span()))
+                        }
+                        Program::Script(script) => {
+                            early_hoist_range(extra_comments, script.body.iter().map(|s| s.span()))
+                        }
+                    };
+
                     let modules_header_width = module_count.next_power_of_two().trailing_zeros();
                     GLOBALS.set(globals_merged, || {
                         let mut visitor = SetSyntaxContextVisitor {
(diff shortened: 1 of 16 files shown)


turbopack/crates/turbopack-ecmascript/src/lib.rs
```diff
@@ -1896,6 +1918,11 @@ async fn process_parse_result(
 
             Ok(CodeGenResult {
                 program,
+                merged_prelude: if retain_syntax_context.is_some() {
```
lukesandberg: why is this conditional? can't we always set this?

turbopack/crates/turbopack-ecmascript/src/lib.rs
lukesandberg, line 1986: the core risk is here

if any of the early hoisted statements get moved around by these passes then the index computations will be wrong

so this could happen if

* an early hoisted statement _is itself_ a directive  (only a theoretical risk really)


Though i do think that `remove_directives` just is non-functional since we add 'early hoisted' statements _first_ and remove_directives just look for directives in leading statement position... so weird.


ok i think quite reasonably non of these passes could remove/add a leading statement in a way that would corrupt the index computations but that is very non-obvious, so we should make it obviously correct.  Could we add an assertion that the prelude statements still exist?

could we add special marker statements and then compute the indices afterwards by scanning?

turbopack/crates/turbopack-ecmascript/src/lib.rs
sampoder: removed this, it wasn't needed

turbopack/crates/turbopack-ecmascript/src/lib.rs
sampoder: switched to adding special marker statements!

lukesandberg (approved):

PR #96698, opened 2026-08-05
Add a `turbopackChunking` documentation page for pages router

sampoder:
These options are also supported on pages router but were missing a documentation page. Component chunks are not supported.


bgw:
If the content here is identical to app router, we typically prefer to use the `source` frontmatter to just copy the content from the app router page: https://github.com/vercel/next.js/blob/canary/docs/02-pages/04-api-reference/04-config/01-next-config-js/env.mdx?plain=1#L4C1-L4C52

There's a `<PagesOnly>` component you can use if there are small modifications between app router and pages router.

bgw (changes requested):
See previous comment

sampoder:
today I learnt! added.

bgw (approved):

PR #96731, opened 2026-08-05
Derive foreground cache revalidation from the consumer

ztanner:
## Summary

Foreground revalidation is necessary when a stale result will be persisted by another server cache, not only when the request is prerendering. Otherwise, a cache created during a dynamic request can consume a stale inner value and extend its lifetime.

- derive foreground revalidation from whether the immediate work-unit consumer will persist the result in a server cache
- model that capability as willConsumerServerCache rather than inheriting it through the full outer scope chain
- treat server cache and static prerender work units as server-caching consumers, while runtime prerenders are not
- add regression coverage for an outer "use cache" scope consuming a stale unstable_cache entry

### DIFF
--- a/packages/next/src/server/app-render/work-unit-async-storage.external.ts
+++ b/packages/next/src/server/app-render/work-unit-async-storage.external.ts
@@ -342,11 +342,11 @@ export type StaticPrerenderStore = Exclude<
 export interface CommonCacheStore
   extends Omit<CommonWorkUnitStore, 'implicitTags'> {
   /**
-   * Whether a stale cache entry must be revalidated before the outer work unit
-   * can continue. This is copied from the outer work unit because entering a
-   * cache scope intentionally hides that store.
+   * Whether this work unit will persist the results it consumes in a server
+   * cache. This only describes the immediate consumer; it is not inherited
+   * from outer scopes.
    */
-  readonly shouldRevalidateStaleCacheEntryInForeground: boolean
+  readonly consumerWillServerCache: boolean
   /**
    * A cache work unit store might not always have an outer work unit store,
    * from which implicit tags could be inherited.
@@ -445,7 +445,7 @@ export type WorkUnitStore =
   | PrerenderStore
   | GenerateStaticParamsStore
 
-export function shouldRevalidateStaleCacheEntryInForeground(
+export function willConsumerServerCache(
   workUnitStore: WorkUnitStore | undefined
 ): boolean {
   if (!workUnitStore) {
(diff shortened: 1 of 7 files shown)


packages/next/src/server/app-render/work-unit-async-storage.external.ts
```diff
@@ -456,14 +450,13 @@ export function shouldRevalidateStaleCacheEntryInForeground(
     case 'cache':
     case 'private-cache':
     case 'unstable-cache':
-      return workUnitStore.shouldRevalidateStaleCacheEntryInForeground
     case 'prerender':
     case 'prerender-client':
```
gnoff: btw, prerender-client should be an error or something. you shouldn't be able to use a cache in this scope b/c it's an SSR scope not an RSC one.

packages/next/src/server/app-render/work-unit-async-storage.external.ts
```diff
@@ -456,14 +450,13 @@ export function shouldRevalidateStaleCacheEntryInForeground(
...
+    case 'prerender-runtime':
```
gnoff: This one is debatable. runtime prerenders are cached in the browser but from a Server Caching perspective they are more like a dynamic render. If we enter a "use cache" during a runtime prerender we should be able to serve stale results b/c they won't be persisted in some other cache on the server, only on the client where stale time rules and the times are usually much shorter.

So I think this should be false

packages/next/src/server/app-render/work-unit-async-storage.external.ts
```diff
@@ -445,7 +439,7 @@ export type WorkUnitStore =
   | PrerenderStore
   | GenerateStaticParamsStore
 
-export function shouldRevalidateStaleCacheEntryInForeground(
+export function willConsumerCache(
```
gnoff: The question isn't will this scope produce a cached result it's whether the parent scope will be cached. So I think you still need to propagate this value as a property of the store type itself and not just compare against the current store's type

packages/next/src/server/app-render/work-unit-async-storage.external.ts
ztanner: Agreed, though probably worth doing in a separate follow-up

packages/next/src/server/app-render/work-unit-async-storage.external.ts
ztanner: agree, though wonder if `willConsumerCache` is too generic. Perhaps `willConsumerPersistResultToServerCache`?

packages/next/src/server/app-render/work-unit-async-storage.external.ts
ztanner: Fixed the propagation in [182e798](https://github.com/vercel/next.js/pull/96731/commits/182e798b5a9916156800fd4163dd32ea932768cc)

packages/next/src/server/app-render/work-unit-async-storage.external.ts
ztanner: `willConsumerServerCache`

gnoff (approved):

PR #96737, opened 2026-08-05
fix: hash canonical invariant

### LINKED ISSUE #96714: router.replace/push restores stale hash from initial URL on dynamic param routes (segment cache seeds hash-ful canonical URL)

https://github.com/FelixK-Witt/next-hash-restore-repro

### To Reproduce

1. `npm install && npm run build && npm run start`
2. Open `http://localhost:3000/p/123#modal` in a **new tab** (direct entry, hash in the initial URL — `/p/[id]` is a dynamic param route).
3. Click the button, which runs:

   ```js
   router.replace(pathname, { scroll: false }) // pathname === '/p/123', no hash
   ```

4. Observe the URL.

**Current:** the URL stays `/p/123#modal`. Instrumenting `history.replaceState` shows the router itself calls `replaceState(..., "/p/123#modal")` although the app passed `/p/123`. Anything keyed on the hash (e.g. modals opened via `#hash` and closed with `router.replace(pathnameWithoutHash)`) can never be closed when the page was entered directly. This is a production incident for us: users cannot close modals on deep links.

**Expected:** `replaceState(..., "/p/123")`, a hash the app removed must not be restored. This worked in Next 15.x.

Notes:

- Only **dynamic param routes** are affected. The repo contains a control case: the identical component on the static root route (`/#modal`) drops the hash correctly.
- Verified on `next@16.3.0` and `next@16.3.1-canary.3` (repo pins `canary`).
- Reproduces with both Turbopack and webpack production builds, with `next start` and with a custom server; independent of `next.config`.
- Related: #91658 (closed), but neither of its linked fixes covers this path: #94144 (shipped in 16.3.0) fixes the search-param variant only, and #91670 (open) changes href resolution in `dispatchNavigateAction`, where the dispatched URL is already hash-less, that is not where the hash comes back.

### Root cause analysis

[`discoverKnownRoute`](https://github.com/vercel/next.js/blob/v16.3.0/packages/next/src/client/components/segment-cache/optimistic-routes.ts#L203-L215) takes the canonical URL to store on the route cache entry. It has two call sites, and they disagree about whether that URL may contain a hash.

**1. The navigation call site strips the hash, and documents why** — [`segment-cache/navigation.ts#L528-L541`](https://github.com/vercel/next.js/blob/v16.3.0/packages/next/src/client/components/segment-cache/navigation.ts#L528-L541):

```ts
discoverKnownRoute(
  ...
  // Store a hashless canonical URL: the entry is shared across hashes, and
  // a later same-route hash nav appends `url.hash` to it.
  createHrefFromUrl(canonicalUrl, false),
  ...
)
```

**2. The initial-load call site does not** — [`router-reducer/create-initial-router-state.ts#L107-L119`](https://github.com/vercel/next.js/blob/v16.3.0/packages/next/src/client/components/router-reducer/create-initial-router-state.ts#L107-L119):

```ts
discoverKnownRoute(
  ...
  canonicalUrl, // <- includes location.hash
  ...
)
```

That `canonicalUrl` is derived a few lines above at [`#L66-L72`](https://github.com/vercel/next.js/blob/v16.3.0/packages/next/src/client/components/router-reducer/create-initial-router-state.ts#L66-L72) as `createHrefFromUrl(location)`, and [`createHrefFromUrl`](https://github.com/vercel/next.js/blob/v16.3.0/packages/next/src/client/components/router-reducer/create-href-from-url.ts#L1-L6) includes the hash unless `includeHash` is explicitly `false`.

So entering `/p/123#modal` directly seeds the route cache entry with `"/p/123#modal"`. A later same-route navigation then rebuilds the target URL at [`segment-cache/navigation.ts#L398`](https://github.com/vercel/next.js/blob/v16.3.0/packages/next/src/client/components/segment-cache/navigation.ts#L398):

```ts
const canonicalUrl = route.canonicalUrl + url.hash
```

With the poisoned seed this yields `"/p/123#modal"` even though the dispatched `url.hash` is empty, and that string is what reaches `history.replaceState`.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.6.0: Mon Jul 29 21:14:30 PDT 2024; root:xnu-10063.141.2~1/RELEASE_ARM64_T6000
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 24.12.0
  npm: 11.6.2
  Yarn: 1.22.22
  pnpm: 10.29.2
Relevant Packages:
  next: 16.3.1-canary.3 // Latest available version is detected (16.3.1-canary.3).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Linking and Navigating

next build (local), next start (local), Other (Deployed)

### Additional context

_No response_

icyJoseph:
Fixes #96714

Route cache entries are keyed by pathname and search only, so one entry is shared across every hash of a route. A same-route navigation rebuilds the URL by appending the incoming hash to the entry's stored canonical URL:

- https://github.com/vercel/next.js/blob/6f2db21c74533dc839a5019c451f19ed7f197a88/packages/next/src/client/components/segment-cache/navigation.ts#L398

Which means that "the stored canonical URL is hashless". Every time we write to `RouteCacheEntry.canonicalUrl` it has to be without hash. These places don't do that:

- `create-initial-router-state.ts` passes `createHrefFromUrl(location)`, so a direct entry to `/p/123#modal` seeds the entry with the hash
- `server-action-reducer.ts` derives its URL from `state.canonicalUrl` when the action didn't redirect, so any revalidating action fired on a hash-ful URL poisons the entry
- `ppr-navigations.ts` passes a retry URL that `fetchServerResponse` returns as `originalUrl` when the response wasn't redirected

> `AppRouterState.canonicalUrl` is the opposite: it drives `history.pushState` and must keep the hash. Both are plain `string`, and #93132 already fixed one writer of the first kind, which is why this recurred.

Any client router navigation then conflicts with the stored hash. In #96714, `router.replace(pathname)` restores a hash the app just removed, and `router.push` to a new hash produces `#foo#bar`, once again.

Both happen on routes entered directly with a hash, and also after a server action revalidates.

Instead of going around patching `createHrefFromUrl`, we should normalize inside `discoverKnownRoute`, the single point all route cache entries are written through, rather than at each call site.

I got an agent to look at `createHrefFromUrl` usage further, and also noticed that `redirect('/foo#bar')` from a server action loses its fragment. 

- https://github.com/vercel/next.js/blob/6f2db21c74533dc839a5019c451f19ed7f197a88/packages/next/src/client/components/router-reducer/reducers/server-action-reducer.ts#L391-L394

That dropped a fragment the action explicitly asked for, while the external-redirect branch four lines above kept it. Now both keep it.

Also adding tests for the other places a fragment could plausibly end up on a `canonicalUrl`.

- `test/e2e/app-dir/hash-canonical-url` has six of them: the hash not being cleared, and the double fragment, each on initial load and after a server action, plus the action redirect and a guard on the prefetch path.

- `test/e2e/app-dir/hash-canonical-url-retry` has three more pinning the redirect-retry path.

On turning prefetch off for the `hash-cross-path-push` test, a prefetched entry is hashless, so as it stood it wouldn't have caught a regression.


ewah:
i just got here... thanks for the fix

PR #96739, opened 2026-08-05
[15.x] support sharp 0.34 and 0.35

styfle:
Backports https://github.com/vercel/next.js/pull/95507 (`chore(deps): bump sharp@0.35.3`) to `next-15-5`, widened to support both sharp lines.

## Why not a straight bump

sharp 0.35 requires Node `>=20.9.0`, but 15.x still supports Node 18 (`engines: ^18.18.0 || ^19.8.0 || >= 20.0.0`) and pins `nodeVersion: 18.18.2` across 10 CI jobs.

Because sharp is an **optionalDependency**, npm and pnpm both *silently* skip it when `engines` do not match — exit 0, no warning. Pinning 0.35 alone therefore left Node 18 with no sharp at all, which failed the type check in CI and would have broken image optimization for Node 18 users on upgrade.

Verified with identical `package.json` across package managers:

| Node | npm | pnpm 9.6 | pnpm 10 |
| --- | --- | --- | --- |
| 18 | 0.34.5 | not installed | not installed |
| 20.9+ | 0.35.3 | 0.35.3 | 0.35.3 |

sharp 0.35.3 itself runs fine on Node 18 (AVIF/WebP encode, `concurrency()`, `cache()` all verified on 18.17.1 and 18.20.6) — the `>=20.9.0` declaration is what blocks installation, not an actual incompatibility.

## Approach

Widen the range to `^0.34.3 || ^0.35.3` so npm resolves 0.34 on Node 18 and 0.35 on Node 20.9+.

- **`SharpModule` conditional type** resolves `.default` on 0.35 and the module namespace on 0.34. `typeof import('sharp').default` is a hard `TS2694` against 0.34, so a single annotation could not cover both. Type-checked against 0.34.3, 0.34.5, and 0.35.3.
- **Blur placeholders differ** between the two (libvips 8.17 vs 8.18), so the static image tests now read the expected bytes from the installed version via `getStaticPngBlurBase64()` rather than hard-coding one.
- **Lockfile stays on 0.35.3.** CI installs 0.34.3 for the Node 18 jobs before `pnpm build`, since the type check runs as the `generate_types` task inside that build. Net effect: 26 jobs exercise 0.35.3, 10 exercise 0.34.3.

## Behavior change to be aware of

Not purely a version bump — AVIF quality scaling changes:

```diff
- quality: Math.max(quality - 20, 1),
+ quality: Math.max(Math.round(quality * (50 / 80)), 1),
```

This alters generated AVIF bytes and invalidates existing image cache entries on upgrade. The ratio was tuned against sharp 0.35's `tune=iq`, but with this range it also applies to 0.34. This is necessary since we don't know the version of sharp running.

## Also adjusted for this branch

- `pnpm-lock.yaml` regenerated rather than taking canary's (diverged graph). Scoped to sharp: libvips 1.2.0 -> 1.3.2, `color` replaced by vendored `@img/colour`.
- `getSharp()` keeps its existing signature; canary's `operationCache` parameter does not exist here.
- Blur snapshots live under `test/integration/next-image-new/**` here, not `test/e2e/**`.
- Dropped the `test/production/next-server-nft` changes; that suite does not exist on this branch.

## Verification

- `pnpm --filter=next types` passes with **both** 0.34.3 and 0.35.3 installed
- `pnpm build` passes
- Image integration tests not run locally; relying on CI


KarimPwnz (approved):

packages/next/package.json
```diff
@@ -130,7 +130,7 @@
     }
   },
   "optionalDependencies": {
-    "sharp": "^0.34.3"
+    "sharp": "^0.34.3 || ^0.35.3"
```
styfle, line 133: This is necessary since both npm and pnpm skip `optionalDependencies` whose engines don't match and sharp@0.35 requires node@20 or newer.

Since next@15 supports node@18, we have to allow older sharp too.

ztanner (approved):

PR #96771, opened 2026-08-05
[Bench] Fixes for pure Fizz bench

gaearon:
## Summary

Harness fixes, all surfaced while benching a Fizz-only React PR (react/react#37022):

- **Profile pass now alternates arm order by VM index.** The timed runs are ABBA, but the CPU profile pass ran base-then-cand on every VM. The second arm runs warmer (page cache, governor), which deflated every function in the cand profile by a uniform few percent — untouched Flight frames "improved" ~7%. With balanced order the same untouched frames diff to ~1.5% across 14 paired VMs while real movers keep their magnitude. Each VM records its capture order in the profiles tarball (`prof-order.txt`) so analyses can check order drift directly.
- **The fingerprint now covers every server-side React layer.** `FP_FILES` hashed only the vendored Flight server files, which a Fizz-only React change leaves byte-identical — both arms of a valid A/B reported the same fingerprint. Added the vendored Fizz bundle and the shared react-server runtime (hooks/cache) in both suites. The VM boot loop also warns loudly when arms fingerprint identically, instead of leaving it for the analysis footer.
- **`--prepare` writes a terminal status phase.** It previously exited leaving `phase: "measuring"`, so `bench-status` prescribed collecting measurement data that never existed. `bench-status` now reports prepare-only runs as complete rather than "DEAD — relaunch".

## Verification

- Full 16-VM e2e run with the ordering + Fizz-fingerprint changes (`run-pr37022-e2e-prof2`): arms fingerprint distinctly (`18ea275e` vs `f9805be5`), profile order alternates per VM, untouched hot frames diff to ~zero in the cross-VM profile aggregate.
- `bench-status.mjs` against a real `--prepare` run dir reports `complete (--prepare: caches only)`.
- `--dry-run` smoke of the modified launcher; `node --check` on all three scripts.

### DIFF
--- a/.agents/skills/sandbox-bench/scripts/sandbox-e2e.mjs
+++ b/.agents/skills/sandbox-bench/scripts/sandbox-e2e.mjs
@@ -50,12 +50,9 @@ import {
 
 // Runtime provenance: the compiled server files prod app-page runtimes
 // are bundled from — BOTH bundlers, so changes touching only one still
-// move the fingerprint. Covers every server-side React layer the
-// runtime executes — Flight (react-server-dom-*), Fizz (react-dom
-// server), and the shared react-server runtime (hooks/cache) — because
-// a change confined to one layer leaves the other layers' files
-// byte-identical, and two genuinely different arms would fingerprint
-// the same.
+// move the fingerprint. One file per server-side React layer (Flight,
+// Fizz, shared react-server runtime): a change confined to one layer
+// leaves the other layers' files byte-identical.
 const FP_FILES = [
   'packages/next/dist/compiled/react-server-dom-turbopack-experimental/cjs/react-server-dom-turbopack-server.node.production.js',
   'packages/next/dist/compiled/react-server-dom-webpack-experimental/cjs/react-server-dom-webpack-server.node.production.js',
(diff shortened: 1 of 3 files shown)


.agents/skills/sandbox-bench/scripts/sandbox-e2e.mjs
```diff
@@ -955,18 +965,22 @@ try {
...
-  writeStatus({
-    phase: 'measuring',
-    expSnap,
-    rowsExpected:
-      cfg.vms * cfg.blocks * cfg.runs * 2 * cfg.routes.split(',').length * 2,
-  })
   if (cfg.prepare) {
+    // Terminal phase, not 'measuring': a prepare run launches no
+    // measurement VMs, and a stale 'measuring' status makes
+    // bench-status prescribe collecting data that never existed.
```
ztanner: I think `bench-status` needs to handle `prepared` since it only treats 'done' as successful. Otherwise I think `--prepared` gets reported as "DEAD before measurement"

ztanner (approved):

.agents/skills/sandbox-bench/scripts/sandbox-e2e.mjs
gaearon: should be in https://github.com/vercel/next.js/pull/96771/changes/b130468d434dab34a8171d65ce3ce59f99ca6eb7

PR #96773, opened 2026-08-05
[Bench] Run every remote exec detached

gaearon:
`vercel sandbox exec` drops its stream on multi-minute commands. It also doesn't reliably propagate the remote exit code, so the build fails while the work on the VM keeps going fine:

```
[expsnap] PHASE pack-pr37147 1785965258
[expsnap] hb mem=1367MB
Error: Stream ended before command finished
```

It hit three of four consecutive snapshot/arm builds.

`runDetached` already avoided this for the measurement loops. It nohups the script on the VM and polls the log with short execs, so nothing depends on a long-lived stream. This PR makes it the only way anything runs remotely. Kicks and polls are then the only things riding the stream, and both are safe to repeat — retrying a build is not.

- every remote command goes through `runDetached`: snapshot builds, arm builds, profile passes, upload concat, cleanup. `sbExec` is gone.
- polling starts at 3s and backs off to the 45s cap, so short commands stay cheap
- the kick writes a nonce to `loop.kick` before the `nohup`, so a kick retried after the transport died skips instead of starting a second copy alongside the first. Kicks retry up to 3 times.
- the java install moves into the react-snap script via `sudo` (the VM user has passwordless sudo, verified on a node24 VM)
- `sandbox-gate.mjs` uses the shared `runDetached` and `ensureReactBuildSnapshot` instead of its own copies; its same-lockfile memoization moves into the shared one
- `--scope` replaces the `--team` alias in the CLI invocations (unrelated to PR but this started failing)

After the first commit, the workload that had failed three times completed on the first attempt: arm build, snapshot, 16-VM e2e run, profile passes, no transport casualties. A second arm build and a gate run also completed under the detached path.

### DIFF
--- a/.agents/skills/sandbox-bench/scripts/bench-common.mjs
+++ b/.agents/skills/sandbox-bench/scripts/bench-common.mjs
@@ -4,7 +4,7 @@
 // CI-green gate, cached react arm builds, snapshots, the status.json
 // recovery record, and live interim estimates. Workload-specific code
 // (what runs on a measurement VM) stays in each launcher.
-import { execFile, spawn } from 'child_process'
+import { execFile } from 'child_process'
 import { promisify } from 'util'
 import fs from 'fs'
 import os from 'os'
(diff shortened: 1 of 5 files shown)


.agents/skills/sandbox-bench/scripts/bench-common.mjs
```diff
@@ -438,9 +370,13 @@ export async function rmVm(name) {
...
 }
 
-// Long-running remote work detached from the exec stream (streams drop
-// flakily on multi-minute silences): nohup the script on the VM, then
-// poll its log with short execs. Immune to transport hiccups.
+// Remote execution: nohup the script on the VM, then poll its log with
+// short execs. The exec stream drops flakily on multi-minute commands and
+// the CLI does not reliably propagate remote exit codes, so executions are
+// detached (they run exactly once and survive the transport) and the
+// stream is only used for kicks and polls, which are safe to repeat.
```
ztanner: wondering if this is actually safe to repeat? looking at the logic, it seems like retrying the kick would remove `loop.done`/`loop.log` and start another copy, while the first process could keep running with its output connected to the old unlinked log file.

.agents/skills/sandbox-bench/scripts/bench-common.mjs
```diff
@@ -361,75 +362,6 @@ export function pairedP(deltas) {
...
 
-export function sbExec(vm, timeout, script, tag, onRow) {
-  return new Promise((resolve, reject) => {
-    // The CLI does not reliably propagate the remote exit code (observed
-    // exit 0 after a remote `exit 1`, CLI 56.3.x), so the script reports
-    // its own exit through an EXIT-trap marker; no marker means the
-    // transport died mid-run. Both are failures.
-    const wrapped = `trap 'echo "@@EXEC_EXIT $?"' EXIT
-${script}`
-    const child = spawn(
```
ztanner, line 372: nit: `spawn` is now an unused import

.agents/skills/sandbox-bench/scripts/bench-common.mjs
gaearon: should be fixed by 3ae874d220e74cbf8e69fb810ca727ab2117fc67

.agents/skills/sandbox-bench/scripts/bench-common.mjs
```diff
@@ -453,21 +387,40 @@ export async function runDetached(vm, tag, script, onLine, deadlineMin) {
...
+    try {
+      await sb([
+        'exec',
+        vm,
+        '--timeout',
+        '2m',
+        '--',
+        'bash',
+        '-c',
+        `if [ "$(cat /vercel/sandbox/loop.kick 2>/dev/null)" = "${kickId}" ]; then echo already kicked; ` +
```
ztanner, line 403: the marker check/write isn't atomic, so if the first kick is still running when the retry starts, both shells could get past the check and launch loop.sh. 

I think you could do something like this to have an atomic check+write:
```
if mkdir "/vercel/sandbox/loop.kick-$kickId" 2>/dev/null; then
  # Only one process can reach here.
  nohup bash /vercel/sandbox/loop.sh ...
fi
```

ztanner: there also seems to be a small window after writing the marker but before nohup where a retry would skip even though the script never started, but that might be a bit of an edge case. just want to flag it since the comment describes this whole thing as being idempotent

PR #96775, opened 2026-08-05
Development: don't 404 routes whose files exist on disk

gaearon:
## Summary

In dev, the first request for a newly written page can return a 404 even though the file is on disk. Route knowledge lives in three places that each lag the filesystem on their own schedule: the router server's `fsChecker` (filled by Watchpack), the render server's route matchers (reloaded from Turbopack's entrypoints subscription), and Turbopack's entrypoints themselves. A request that arrives between the write and any of them catching up gets a 404. #83829 diagnosed the underlying race — "the watchpack aggregate event runs before Turbopack's watcher has emitted a change event" — and moved the browser `ADDED_PAGE` announcements out of harm's way, but requests can still observe the stale state directly.

This makes each layer reconcile with a more truthful source before returning a 404, on the miss path only:

- The router server checks the request path against the development route matchers (new `testRouteMatch`) before rendering a 404.
- `DevRouteMatcherManager` re-scans the filesystem once when it has no match for a path.
- Turbopack's `ensurePage` waits for the next entrypoints update instead of throwing `route not found` while the route's file exists on disk, with a 10s liveness backstop for states where no update will ever arrive (e.g. a broken watcher).

Requests for known routes stay on the same code path as before; the extra work only happens on misses.

Measured with the new test (write `app/x/page.tsx`, then make the first-ever request for `/x`):

- before: 12-18 of 30 first requests returned 404, at delays up to ~60ms after the write
- after: all 30 return 200, including a request issued immediately after the write

This also removes the 500ms sleep the test harness added after every `app/`/`pages/` write to paper over this race (`handleDevWatchDelayAfterChange`, whose TODO points at exactly this problem).

## Verification

- `pnpm test-dev-turbo test/development/app-dir/new-route-first-request` (new test; fails on canary)
- `pnpm test-dev-webpack test/development/app-dir/new-route-first-request`
- `pnpm test-dev-turbo test/development/basic test/development/app-hmr test/development/ondemand test/development/duplicate-pages test/development/dynamic-route-rename test/e2e/app-dir/not-found test/e2e/app-dir/not-found-default` (35 suites, 249 tests — run before the harness sleep removal; the matrix without the sleep is left to CI)

### DIFF
--- a/packages/next/src/server/dev/hot-reloader-turbopack.ts
+++ b/packages/next/src/server/dev/hot-reloader-turbopack.ts
@@ -1966,11 +1966,13 @@ export async function createHotReloaderTurbopack(
                   resolve(true)
                 })
               })
-              if (!updated) break
               pendingEntrypointsUpdate = nextEntrypointsUpdate.promise
               route = isInsideAppDir
                 ? currentEntrypoints.app.get(normalizedAppPage)
                 : currentEntrypoints.page.get(page)
+              // The route is checked once more even when the deadline was
+              // hit, in case an update finished at the same time.
+              if (!updated) break
             }
           }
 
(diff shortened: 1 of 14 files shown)


gaearon:

PR #96776, opened 2026-08-06
Detect sibling root layouts when collecting root params

### LINKED ISSUE #96756: next/root-params emits no types when an app has two root layouts and one serves "/"

https://github.com/CHC383/nextjs-root-params-two-root-layouts-repro

### To Reproduce

The repro has two root layouts, siblings rather than nested:

```
  app/(marketing)/layout.tsx   serves "/",           no dynamic segment
  app/[locale]/layout.tsx      serves "/[locale]/*", `locale` is a root param
```

1. npm install
2. npx next typegen
3. cat .next/types/root-params.d.ts => "// No root params detected." and `export {}`

Now remove the layout that serves "/":

4. mv "app/(marketing)" ../hidden && npx next typegen
5. cat .next/types/root-params.d.ts => export function locale(): Promise<string>

Put it back and the type disappears again.

I expected `locale` to be typed, since it is a root param of
app/[locale]/layout.tsx and resolves fine at runtime. Instead typegen writes
"No root params detected", and because next ships a shorthand ambient
declaration for the module, next/root-params silently becomes `any`. This
compiles with no error under tsc --strict:

    import * as rootParams from "next/root-params";
    export async function probe(): Promise<number> {
      return await rootParams.locale();   // locale() is `any`
    }

Adding my own `declare module "next/root-params"` doesn't help — the shipped
declaration is a resolved file, so it wins over an ambient one. tsc accepts the
local declaration while the module still resolves to `any`. With
typescript-eslint's strict type-checked rules the call reports no-unsafe-call,
so the only options left are suppressing the rule or dropping the strict rules.

Runtime is unaffected, which is what makes this easy to miss. `next build` on
the repro prerenders /en and /de, and each page renders the correct value:

    grep -o 'root param at runtime:.\{0,20\}' .next/server/app/en.html
    => root param at runtime: en

Since "/" has no `locale` segment, I'd expect
`locale(): Promise<string | undefined>` rather than nothing at all.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.6.0: Sat Jul 11 15:26:21 PDT 2026; root:xnu-12377.161.13~4/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 10
Binaries:
  Node: 24.19.0
  npm: 12.0.2
  Yarn: 1.22.22
  pnpm: 10.33.2
Relevant Packages:
  next: 16.3.1-canary.3 // Latest available version is detected (16.3.1-canary.3).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Route Groups, TypeScript

### Additional context

Hit this on 16.3.0 in a real app; the repro pins next@canary and still fails on
16.3.1-canary.3.

It looks like the root-layout detection in
packages/next/src/server/lib/router-utils/route-types-utils.ts,
collectRootParamsFromLayouts:

    const rootLayoutRoutes = routes.filter(
      (route) =>
        !routes.some(
          (other) =>
            other !== route && (other === '/' || route.startsWith(other + '/'))
        )
    )

`app/(marketing)/layout.tsx` normalizes to "/", so the `other === '/'` clause
treats it as an ancestor of "/[locale]" and that layout contributes no root
params. They're siblings, though. The predicate compares route strings, which
have already had the route group stripped and so no longer carry the nesting
information needed to tell an ancestor from a sibling — manifest.layoutRoutes
does keep a filePath per layout, which would.

The runtime doesn't share the bug, which is why only types break:
next-root-params-loader's findRootLayouts walks the app directory and stops at
the first layout per branch, so it finds both roots and collects `locale`.

Two root layouts with one serving "/" is how you split a localized subtree from
an unprefixed marketing or legal tree, so this isn't an exotic shape. It also
fails open rather than loudly — the import keeps working, just unchecked.

Claude Opus 5 helped with the investigation and built the linked reproduction.

lazerg:
When an app has two root layouts and one of them serves `/` through a route group, `next typegen` writes "No root params detected" and `next/root-params` silently resolves to `any`. Root layout detection compared route strings, and route groups are already stripped from those, so `app/(marketing)/layout.tsx` normalizes to `/` and looks like an ancestor of `app/[locale]/layout.tsx` instead of its sibling. Comparing the layout file paths keeps both roots, and `locale` is then typed as `Promise<string | undefined>` because it is missing from one of them.

Fixes #96756

### DIFF
--- a/packages/next/src/server/lib/router-utils/route-types-utils.test.ts
+++ b/packages/next/src/server/lib/router-utils/route-types-utils.test.ts
@@ -1,0 +1,35 @@
+import { createRouteTypesManifest } from './route-types-utils'
+
+describe('route types manifest root params', () => {
+  function createManifest(layoutRoutes: { route: string; filePath: string }[]) {
+    return createRouteTypesManifest({
+      dir: '/project',
+      pageRoutes: [],
+      appRoutes: [],
+      appRouteHandlers: [],
+      pageApiRoutes: [],
+      layoutRoutes,
+      slots: [],
+    })
+  }
+
+  it('should collect root params from sibling root layouts', async () => {
+    const manifest = await createManifest([
+      { route: '/', filePath: '/project/app/(marketing)/layout.tsx' },
+      { route: '/[locale]', filePath: '/project/app/[locale]/layout.tsx' },
+    ])
+
+    expect(manifest.rootParams).toEqual(
+      new Map([['locale', new Set(['string', 'undefined'])]])
+    )
+  })
+
+  it('should not collect params from layouts below the root layout', async () => {
+    const manifest = await createManifest([
+      { route: '/', filePath: '/project/app/layout.tsx' },
+      { route: '/[locale]', filePath: '/project/app/[locale]/layout.tsx' },
+    ])
+
+    expect(manifest.rootParams).toEqual(new Map())
+  })
+})
--- a/packages/next/src/server/lib/router-utils/route-types-utils.test.ts
+++ b/packages/next/src/server/lib/router-utils/route-types-utils.test.ts
@@ -1,7 +1,8 @@
+import type { RouteInfo } from '../../../build/file-classifier'
 import { createRouteTypesManifest } from './route-types-utils'
 
-describe('route types manifest root params', () => {
-  function createManifest(layoutRoutes: { route: string; filePath: string }[]) {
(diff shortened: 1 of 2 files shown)


lazerg:

lazerg:

PR #96778, opened 2026-08-06
[turbopack] Enable the shared runtime by default

sampoder:
16.3 is out so we will test this in canary!


bgw (approved):
10KiB win on every route navigation, here we come!

PR #96780, opened 2026-08-06
[Bench] Add timed browser phase to the e2e suite

gaearon:
Stacked on #96773.

`--browser` adds a timed headless-Chromium phase per (run, arm) against the built app, paired and interleaved like every other phase.

Every counter is read once the page goes quiet, not at a fixed point after load. Two pages with the same total JS cost (a 300ms busy loop), differing only in when it runs:

|  | during load | deferred 100ms past hydration |
| --- | --- | --- |
| fixed read point | 300ms | **0ms** |
| read at quiescence | 300ms | 300ms |

Deferred post-hydration work is the shape of streamed Suspense hydration and lazy Flight initialization. A fixed read point reports "no client cost" for exactly the changes most likely to have one. A sample that never quiesces within 10s is still counted, warned about on stderr, and marked in the row, so the analysis flags the cell instead of passing off a windowed read as a total.

Metrics, 10 iterations medianed into one row per run:

- **browser-load** — `hydrate` (the app's `bench:hydrated` mark), `hydrateClient` (`hydrate - ttfb`, the client share with server latency removed), `fcp`, `lcp`, `blocking` (long-task time over 50ms up to the hydration mark), `script`/`task` (CDP `Performance.getMetrics`), `heapMb` (read after a forced GC, so retained rather than transient)
- **browser-nav** — client navigation via `window.next.router.push`: `nav` (wall to URL commit), `navSettled` (wall to the last DOM mutation), `navScript`/`navTask` (Flight parse + render in the browser)

A/A on `run-browser-aa4` (both arms at the same react commit, 16 boots): 84 metric rows, zero significant at the p<0.01 claim bar. All 36 browser rows land at p>0.05, smallest p=0.084. At 16 boots the browser CIs are ±0.9-1.7% on `hydrate`/`hydrateClient`/`script`/`task`/`nav`/`navScript`, ±1.6-2.1% on `blocking`, ±2.8-4.1% on `lcp`.

Non-significant cells now report the largest effect their data still allows. A cell that could only have resolved a 12% change and one that resolved 1% both printed as "no difference". The analysis now ends with a summary splitting cells into claims (p < 0.01), flags, and unresolved, the last sorted tightest first. The `/dashboard` `fcp` band below is the case that motivated it.

<details>
<summary>Environment and measurement details</summary>

- A fixed CPU throttle (default 4x) amplifies JS cost over scheduling noise; localhost keeps the network out of the signal. One browser process serves all iterations, each in a fresh context (cold cache, fresh renderer).
- Server and browser are pinned to disjoint CPU sets (`taskset -c 0-3` and `-c 4-7`) so they don't contend for the same cores.
- `lcp` and `blocking` come from buffered `PerformanceObserver`s, so the pass still runs without CDP tracing — tracing perturbs the timings it collects.
- The URL commit is detected through page events over CDP. It replaces an in-page 5ms poll that was costing part of the number it was helping measure: `navScript` on `/blog` went from ±6.1% to ±1.4% between the two A/A runs, and its absolute value dropped ~130ms -> ~80ms.
- `heapMb` reads ±0.0% on every route now that the read follows a forced GC; one route had a ±0.7% band before, wide enough to have produced a spurious heap "regression".
- Not every metric discriminates on every route: `fcp` on `/dashboard` has a ±12.3% band (present before these changes too, ±10.5%), so it carries no signal there. It is still reported rather than dropped, since the CI says so directly.
- The pinned Chromium is installed into the experiment snapshot at build time, so both arms run the exact same binary and measurement VMs boot browser-ready. AL2023 has no apt, so `playwright install --with-deps` fails; the system libraries are installed explicitly via dnf and a launch smoke test gates the snapshot.
- The two A/A runs sit on different Next canaries (a canary shipped between them), so the CI comparisons above are indicative rather than controlled; the two called out have a known mechanism behind them.

</details>


.agents/skills/sandbox-bench/scripts/sandbox-e2e.mjs
```diff
@@ -630,6 +659,59 @@ for run in $(seq 1 ${total}); do
...
+        if (r.load.fcpMs>0) row.fcp=r.load.fcpMs;
+        if (r.load.lcpMs>0) row.lcp=r.load.lcpMs;
+        if (r.load.blockingMs>0) row.blocking=r.load.blockingMs;
+        if (r.load.scriptMs>0) row.script=r.load.scriptMs;
+        if (r.load.taskMs>0) row.task=r.load.taskMs;
+        if (r.load.heapMb>0) row.heapMb=r.load.heapMb;
+        // Samples that never quiesced make the counters windowed reads,
+        // not totals; surface through the errors channel so the analysis
+        // flags the cell.
+        if (r.unsettled>0) row.errors=r.unsettled;
```
ztanner, line 699: what is errors used for? doesn't seem like verification / analysis read it. Should `errors: 10` still produce a clean p=0.000 claim?

.agents/skills/sandbox-bench/scripts/vm-browser-bench.mjs
```diff
@@ -0,0 +1,338 @@
...
+            resolve({
+              ttfbMs: nav ? nav.responseStart : null,
+              dclMs: nav ? nav.domContentLoadedEventEnd : null,
+              loadMs: nav ? nav.loadEventEnd : null,
+              fcpMs: fcp ? fcp.startTime : null,
+              lcpMs,
+              hydratedMs: hydrated ? hydrated.startTime : null,
+              blockingMs,
+            })
+          }, 100)
```
ztanner, line 169: Would this miss other potential LCP candidates, since this disconnects after 100ms?

PR #96782, opened 2026-08-06
Turbopack: replace the ensurePage wait deadline with a route probe

gaearon:
## Summary

Stacked on #96775.

That PR made `ensurePage` wait for entrypoints updates when a route's file exists on disk but Turbopack hasn't reported it yet, bounded by a 10s deadline — needed because in some states (a broken watcher, a route key that never matches) no update ever arrives, and the request would otherwise hang forever. The deadline also meant those states stall every request for 10s before failing.

This replaces the deadline with the actual signal. A new `projectContainsRoute` binding invalidates the tracked directory reads for the file's directory and answers from a strongly consistent read of the entrypoints computation:

- If the route exists, the entrypoints subscription is guaranteed to deliver an update containing it under the same key (pages routes are keyed by pathname, app routes by original name, on both sides), so `ensurePage` waits for exactly that.
- If it doesn't, the request fails immediately — no constant, and states that previously stalled 10s per request now fail fast, including fatal build errors (the probe errors and the request 404s).

The invalidation goes through a new synced variant of `invalidate_path_and_children_with_reason` that takes the filesystem's invalidation lock, the same way watcher-driven invalidation batches do. Without it, a concurrent read can observe directory listings and file contents from different sides of the invalidation and bake a transient "file not found" stub into a compiled chunk — reproduced under delete/recreate churn before adding the lock (~7/40 cycles with a transient 500, vs 1/40 baseline with the probe disabled; with the lock it matches the baseline).

## Verification

- `pnpm test-dev-turbo test/development/app-dir/new-route-first-request` (3 consecutive runs)
- `pnpm test-dev-webpack test/development/app-dir/new-route-first-request`
- `pnpm test-dev-turbo test/development/app-hmr`
- Standalone delete/recreate stress (40 cycles with requests in flight across each transition): converges every cycle, transient 5xx rate equal to the probe-disabled baseline

### DIFF
--- a/crates/next-napi-bindings/src/next_api/project.rs
+++ b/crates/next-napi-bindings/src/next_api/project.rs
@@ -1031,6 +1031,105 @@ fn project_container_entrypoints_operation(
     container.entrypoints()
 }
 
+#[turbo_tasks::function(operation, root)]
+async fn project_contains_route_operation(
+    container: ResolvedVc<ProjectContainer>,
+    route_key: RcStr,
+) -> Result<Vc<bool>> {
+    let entrypoints = container.entrypoints().await?;
+    // Pages routes are keyed by pathname and app routes by original name,
+    // mirroring how the JS side keys the routes it receives from the
+    // entrypoints subscription (see `handleEntrypoints` in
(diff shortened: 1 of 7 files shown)


gaearon:

PR #96783, opened 2026-08-06
Only announce server component changes when compiled output changed

gaearon:
Stacked on #96248.

In dev with Turbopack, applying a server HMR update ends with telling every connected tab to refetch its RSC payload. But an applied update doesn't necessarily change anything a tab renders. Removing a page is the clearest case: the update just rewrites the removed page's module into a throwing stub inside the merged server chunks —

```
"code": "((__turbopack_context__, module, exports) => {
  var e = new Error(\"Could not parse module '[project]/app/zz-probe/page.tsx', file not found\");
  e.code = 'MODULE_UNPARSABLE';
  throw e;
})"
```

— and no other page's output changes. Every tab was still told to refetch. Since the removed-page announcement already makes tabs refetch, deleting one page made every tab refetch twice instead of once.

Messages the server sends for `rm app/zz-added/page.tsx` while a tab shows `/counted`, before the fix:

```
{"type":"removedPage","data":["/zz-added"]}
{"type":"serverComponentChanges"}            ← 2–4ms later, from the HMR apply
```

after:

```
{"type":"removedPage","data":["/zz-added"]}
```

Because the two messages arrive within milliseconds, the browser sometimes handles them as one refetch and sometimes as two, depending on the machine. This is why the refetch-count test could not pin down the removal: idle machines usually saw two fetches, CI machines usually saw one, and the test expected an exact number. That was the `Expected: 3, Received: 2` failure every Turbopack CI job showed for the PRs below this one. Since no exact refetch count is correct while the extra announcement exists, the test now also asserts on the announcement itself, which doesn't depend on timing: without this fix it fails every run, on every machine.

The apply loop now decides the same way the `"use cache"` refresh hash already does one PR down: compare each subscribed entry's compiled-output hash (`serverContentHash()` from #96248) against a baseline, and announce only when one differs.

- Baselines are seeded when an entry's change subscription starts, before the entry's first render is underway. An edit landing in the few milliseconds between the two can be absorbed into the baseline, losing one announcement. The refresh-hash baseline one PR down has the same window, and the next change corrects it.
- A null hash means the entry has no compiled output right now (removed, or failing to build). It keeps the previous baseline, so output that comes back different is still announced.
- A hash that can't be read counts as changed. The reads go through `Promise.allSettled` because a rejected read must not take down the apply loop — its failure path re-evaluates every server module.
- Env-change invalidations announce through a separate path and are unaffected.

## Test plan

The first commit extends the refetch-count test with a removal phase — exactly one refetch per addition and one per removal, on both bundlers — and asserts that no `serverComponentChanges` is announced for either. The test ends by checking the opposite direction too: editing the page the tab is showing must produce exactly one `serverComponentChanges` and exactly one more refetch. Besides covering the edit path, this proves the message capture works — if the capture silently broke, the "no announcements" assertions would pass without checking anything.

Without the fix, the test fails on Turbopack 3 of 3 runs; webpack passes it unchanged. With the fix, Turbopack passes 3 of 3, and an earlier revision of the same test (without the edit phase) passed 30 consecutive full-suite runs.

<details>
<summary>Suites covering the changed loop, all green with the fix (locally built binding)</summary>

- `route-change-refetch` — 30 consecutive Turbopack runs of the counted test green; webpack 3×10/10. The pre-existing "tab stuck on 404 after a page add" flake (the ensurePage/watcher race, being fixed separately) is unaffected: matched warm runs are 10/10 both with and without this fix.
- `server-components-hmr-cache` (17) — edits and env changes must still refetch and invalidate
- `middleware-dev-update` (4)
- `hmr` basic (4), error-recovery (12 + 1 skipped), full-reload (2)
- `cache-components-spurious-cache-invalidation` (6, idle and under CPU load)
- `dynamic-optional-routing` under experimental Turbopack (33)

</details>

### DIFF
--- a/packages/next/src/server/dev/hot-reloader-turbopack.ts
+++ b/packages/next/src/server/dev/hot-reloader-turbopack.ts
@@ -547,6 +547,11 @@ export async function createHotReloaderTurbopack(
 
   // Dev specific
   const changeSubscriptions: ChangeSubscriptions = new Map()
+  // Baselines for deciding whether an applied server HMR update changed any
+  // entry's compiled output. Seeded when an entry's change subscription
+  // starts; advanced only by the apply loop (`onApplied` below).
+  const serverContentHashEndpoints = new Map<EntryKey, Endpoint>()
+  const appliedServerContentHashes = new Map<EntryKey, string>()
   const serverPathState = new Map<string, string>()
   const readyIds: ReadyIds = new Set()
   let currentEntriesHandlingResolve: ((value?: unknown) => void) | undefined
@@ -927,6 +932,7 @@ export async function createHotReloaderTurbopack(
    * advance the hash when it actually changed.
    */
   async function* serverChanges(
+    key: EntryKey,
     changedPromise: Promise<
       AsyncIterableIterator<TurbopackResult<ServerChanged>>
     >,
@@ -934,6 +940,11 @@ export async function createHotReloaderTurbopack(
   ) {
     const changed = await changedPromise
     let contentHash = await endpoint.serverContentHash()
+
+    serverContentHashEndpoints.set(key, endpoint)
+    if (contentHash !== null && !appliedServerContentHashes.has(key)) {
+      appliedServerContentHashes.set(key, contentHash)
+    }
 
     for await (const change of changed) {
       // A null hash means the entry has no compiled output (it was removed,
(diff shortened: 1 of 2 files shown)


gaearon:
May revisit this in the future, for now don't want to spend more time on it

PR #96784, opened 2026-08-06
Serve a new dev route as soon as Turbopack announces it

gaearon:
## Summary

Stacked on #96782 (which stacks on #96775).

In dev the router works out what it can serve from three pieces of state: the exact path table (`appFiles` and `pageFiles`), the dynamic route list, and `appPathRoutes`, which says whether a path belongs to the App Router or the Pages Router. All three are written by the watchpack handler in setup-dev-bundler.ts.

With Turbopack, `ADDED_PAGE` and `REMOVED_PAGE` are sent from the entrypoints event instead, and Turbopack gets there before watchpack. So the browser hears about a route, refetches right away, and gets a 404 because the router can't serve it yet. Under load that gap grows to seconds.

Now the route table is written from the entrypoints event too, so the state the router serves from and the event that tells the browser to refetch come from the same place. Webpack still writes it from watchpack, where it already announces from inside the same handler and so never had the race.

#83829 moved the events but left the state behind. This is the other half.

On top of the two PRs below it, this also removes a whole class of skew: the router's tables, the `ADDED_PAGE` announcements, and the entrypoints the render server compiles from now all come from the same event. Requests arriving before that event are covered by the on-miss reconciliation from #96775/#96782 (dev matchers re-scan the filesystem; `ensurePage` asks Turbopack directly).

## Verification

- `pnpm test-dev-turbo test/development/app-dir/new-route-first-request` (2 consecutive runs)
- `pnpm test-dev-turbo test/development/app-hmr test/development/basic/gssp-ssr-change-reloading test/e2e/app-dir/not-found/index.test.ts test/development/duplicate-pages` (data routes, page add/remove, 404 rendering)

### DIFF
--- a/crates/next-core/src/app_structure.rs
+++ b/crates/next-core/src/app_structure.rs
@@ -19,8 +19,9 @@ use crate::{
     next_app::{
         AppPage, AppPath, PageSegment, PageType,
         metadata::{
-            GlobalMetadataFileMatch, MetadataFileMatch, match_global_metadata_file,
-            match_local_metadata_file, normalize_metadata_route,
+            GlobalMetadataFileMatch, MetadataFileMatch, is_multi_dynamic_metadata,
+            match_global_metadata_file, match_local_metadata_file, multi_dynamic_metadata_page,
+            normalize_metadata_route,
         },
     },
     next_import_map::get_next_package,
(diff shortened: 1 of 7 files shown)


gaearon:

PR #96785, opened 2026-08-06
Turbopack: derive the dev route matchers from the entrypoints

gaearon:
## Summary

Stacked on #96784 (→ #96782 → #96775). Exploratory: collapses the last independently-maintained route store in dev.

With Turbopack, the dev route matchers no longer scan the filesystem. They serve from route definitions derived from the compiled entrypoints, pushed on every entrypoints update alongside the route tables from #96784 — so the router's tables, the render server's matchers, and the entrypoints all describe the same state, updated by the same event. On a request path that doesn't match, the matchers pull a freshly recomputed route list from Turbopack (`projectGetRoutes`, which invalidates the tracked app/pages directory-listing reads first) instead of re-scanning the filesystem.

Because matched definitions now always originate from Turbopack's own routes, `ensurePage` stops consulting the filesystem entirely: the `containsRoute` probe from #96782 is the existence authority, and a deleted file terminates the wait as the route vanishing from a strongly consistent recompute.

Dynamic metadata routes are the one place the emission is thinner than a filesystem scan: sitemap.ts with `generateSitemaps` serves both `/sitemap.xml` and `/sitemap/[__metadata_id__]`, but Turbopack reports only the single form. The derivation adds the multi variant the way the filesystem-scanning provider used to — and also restores the `appPathRoutes` entry for it, which deriving the route table from the entrypoints (#96784) had silently dropped (caught by `test/e2e/app-dir/metadata-dynamic-routes`; that fix logically belongs one PR down and can be folded there if this one doesn't land).

Webpack keeps the filesystem-scanning providers.

## Verification

- `pnpm test-dev-turbo test/development/app-dir/new-route-first-request` (2 consecutive runs)
- `pnpm test-dev-turbo test/e2e/app-dir/metadata-dynamic-routes/index.test.ts` (red before the multi-variant derivation, green after)
- `pnpm test-dev-turbo test/development/app-hmr test/development/basic/gssp-ssr-change-reloading test/e2e/app-dir/not-found/index.test.ts test/development/duplicate-pages test/e2e/app-dir/not-found-with-pages-i18n`
- Broader coverage left to CI

### DIFF
--- a/crates/next-napi-bindings/src/next_api/project.rs
+++ b/crates/next-napi-bindings/src/next_api/project.rs
@@ -1073,58 +1073,156 @@ pub async fn project_contains_route(
         .turbopack_ctx
         .turbo_tasks()
         .run_once(async move {
-            if !invalidate_dirs.is_empty() {
-                #[turbo_tasks::value(cell = "new", eq = "manual")]
-                struct ProjectDirInfo(FileSystemPath, DiskFileSystem);
-
-                #[turbo_tasks::function(operation, root)]
-                async fn project_dir_info_operation(
-                    container: ResolvedVc<ProjectContainer>,
-                ) -> Result<Vc<ProjectDirInfo>> {
-                    let project = container.project();
-                    let project_path = project.project_path().owned().await?;
-                    let project_fs = project.project_fs().owned().await?;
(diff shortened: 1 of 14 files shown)


gaearon:

PR #96786, opened 2026-08-06
Give route-shaped dev server strings distinct types

gaearon:
Stacked on #96783.

Turbopack's entrypoints keep routes in maps keyed by string namespaces that look alike but rarely equal each other: the routes map is keyed by route pathname (`/blog`), while the app entrypoints map is keyed by entry name (`/blog/page`). Mixing them up compiles fine and silently never matches — which is how the dev server came to announce every app route as added on every update (fixed at the bottom of this stack, #96250).

This brands the two namespaces the way `EntryKey` already guards entry keys:

- `TurbopackRouteKey` — keys of `RawEntrypoints.routes`
- `AppEntryName` — `originalName` / keys of `AppEntrypoints`

A plain string can't be used where a branded key is expected, and the brands can't be used in each other's place. Reverting #96250's fix no longer compiles:

```
hot-reloader-turbopack.ts: error TS2345: Argument of type 'TurbopackRouteKey'
  is not assignable to parameter of type 'AppEntryName'.
hot-reloader-turbopack.ts: error TS2345: Argument of type 'string'
  is not assignable to parameter of type 'TurbopackRouteKey'.
```

Strings enter the namespaces at the NAPI boundary (route construction in `swc/index.ts`) and leave through message payloads, where the brands widen back to plain strings. `normalizedPageToTurbopackStructureRoute` returns an entry name — per its own comment it maps a normalized page path back to Turbopack's entry key — and is typed accordingly. Pages Router maps are left as plain strings: there the route pathname and the page name are genuinely the same string.

No runtime change; `route-change-refetch` passes 10/10 as a sanity check.

### DIFF
--- a/packages/next/src/build/swc/index.ts
+++ b/packages/next/src/build/swc/index.ts
@@ -21,6 +21,7 @@ import type {
   NapiCodeFrameOptions,
 } from './generated-native'
 import type {
+  AppEntryName,
   Binding,
   BuildFeatureUsage,
   CompilationEvent,
@@ -36,6 +37,7 @@ import type {
   Route,
   TurboEngineOptions,
   TurbopackResult,
+  TurbopackRouteKey,
   TurbopackStackFrame,
   Update,
   UpdateMessage,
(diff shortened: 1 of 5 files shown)


gaearon:
May revisit this in the future, for now don't want to spend more time on it

PR #96788, opened 2026-08-06
Convert tree prefetches to NavigationFlightResponse format

acdlite:
Part of a stack of changes to convert all RSC requests to a single unified response type, NavigationFlightResponse.

This step converts static route tree prefetches (i.e. `/_tree` requests).

The old format could omit a dynamic param's value from the response so that a statically served response is cacheable across param values; the client parses the value from the URL instead. That behavior moves into the shared decoder, where it's now permitted for any transport response.

A cached /_tree response from an older build decodes without a build id and falls back to an MPA navigation, like any other cross-build response. The staleTime field of the old format is not carried over because the client never read it.


packages/next/src/client/components/segment-cache/cache.ts
```diff
@@ -2097,129 +1891,89 @@ export async function fetchRouteOnCacheMiss(
...
-      // because all data is static in this mode.
-      isOutputExportMode
-
-    if (routeIsPPREnabled) {
-      const { stream: prefetchStream, size: responseSize } =
-        await createNonTaskyPrefetchResponseStream(response.body)
-      closed.resolve()
-      setSizeInCacheMap(entry, responseSize)
-      const serverData = await createFromNextReadableStream<RootTreePrefetch>(
+    const routeIsPPREnabled = wasServedFromPerSegmentCache(response)
```
ztanner: non-blocking nit but I find the naming here a bit confusing because fully static, non PPR routes can also serve from the per-segment cache. I don't think the postpone header necessarily means the render actually postponed or that PPR is enabled.

packages/next/src/shared/lib/app-router-types.ts
```diff
@@ -484,12 +484,25 @@ export type NavigationFlightResponse = {
...
-  /** renderedSearch */
-  q: string
-  /** couldBeIntercepted */
-  i: boolean
+  /**
+   * supportsPerSegmentPrefetching — present in live-render responses;
+   * absent in per-segment prefetch responses served from static storage
+   * (where it's implied by the response's existence).
+   */
+  S?: boolean
```
ztanner: Do you think it'd be worth modeling this as two response variants that share the same common base, rather than making fields optional on `NavigationFlightResponse`? I worry that making these fields optional forces `fetchServerResponse` to provide fallbacks that could mask a malformed navigation response.

Something like:

```ts
type NavigationResponse = ResponseBase & {
  q: string
  S: boolean
  i: boolean
}

type RouteTreePrefetchResponse = ResponseBase & {
  t: PartialTransportData
}
```

acdlite: Yeah I can create aliases to enforce which fields are required in which contexts

packages/next/src/client/components/segment-cache/cache.ts
acdlite: yeah I'll rename to something else

ztanner (approved):

PR #96802, opened 2026-08-06
fix(turbopack): validate WASM import/export names before emitting them as JS identifiers

marcoshernanz:
## What

`instantiating_loader_source()` (turbopack/crates/turbopack-wasm/src/loader.rs) generates the `_.loader.mjs` module for a `.wasm` import by interpolating the WebAssembly **import and export names directly as raw JavaScript identifiers**:

- `import { <items.join(", ")> } from <path>`
- the value side of the imports object (`"key": <item>`)
- `const { <exports.join(", ")> } = await instantiate(...)` and `export { <exports> };`

WASM import/export names are **arbitrary UTF-8 strings** (wasmparser copies them verbatim in analysis.rs — no validation anywhere). A crafted `.wasm` binary with an export name such as

```
a } = x; maliciousCode(); const { b
```

escapes the destructuring context and injects arbitrary JavaScript statements into the generated loader module, which is then compiled into the application bundle. Unlike a malicious `.js` dependency, a `.wasm` file is normally treated as sandboxed, inert data — the injected JavaScript runs outside the wasm sandbox (Node/edge server context for a server-side import, or the shipped client bundle).

## Fix

Validate that every import/export name is a strict-mode-bindable ECMAScript identifier before emitting it as a raw identifier; fail the build with a clear error otherwise.

The identifier grammar intentionally matches the JS parser exactly:

- `ID_Start`/`ID_Continue` via the `unicode-id-start` crate (the same crate SWC's `swc_ecma_ast` uses — `unicode-ident`'s XID properties are a strict subset and would have regressed valid modules),
- plus `_`/`$` and ZWNJ/ZWJ identifier-part handling per the ECMAScript spec,
- minus reserved words and strict-mode-unbindable names.

Names that produce valid loader code today — including non-ASCII identifiers like `café` or `関数`, and ID-but-not-XID chars like U+037A/U+0E33/U+309B (confirmed valid with V8) — are unaffected. Rejected names previously either failed with an opaque parser error or enabled the injection.

## Tests

Five unit tests in loader.rs, including malicious import/export names embedded in real binaries built with `wat` (parsed through the same wasmparser path as `analyze`), acceptance of common/unicode identifiers, and reserved-word rejection. `cargo check`/`clippy`/`fmt`/`test -p turbopack-wasm` all green.

---


mischnic:
we should instead just use the syntax
`export { foo as "🙂"}` which allows arbitrary unicode as the export name just fine

Issue #95798, opened 2026-07-14
Turbopack dev-server livelocks compiling monorepo — all tokio workers stuck at same address in next-swc.darwin-arm64.node (16.2.6, works on 16.2.5)


domelic:
### Verify canary release

- [x] I verified that the issue exists in the latest Next.js canary release

### Provide environment information

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin 25.5.0 (macOS 26.5)
Binaries:
  Node: v24.5.0
  pnpm: 10.x
Relevant Packages:
  next: 16.2.6 (broken) / 16.2.5 (works)
  eslint-config-next: 16.1.5
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.x
Next.js Config:
  output: N/A (dev)

### Which area(s) are affected? (Select all that apply)

Turbopack

### Which stage(s) are affected? (Select all that apply)

next dev (local)

### Additional context

**Symptom:** `next dev` on 16.2.6 accepts the TCP connection on :3000, prints `✓ Ready in ~300ms`, then never responds to any request. `next-server` process sits at ~1200% CPU with **no compile output whatsoever** — no "Compiling /" line, no error, no timeout. Curl requests hang until the client's own deadline (verified up to 5 minutes).

Downgrading to **16.2.5 restores compile in ~3.2s** (307 redirect from `/` → `/login`, then 200 in ~10s cold-cache).

**Reproduction environment:**
- Pnpm workspace monorepo: ~5 TS packages (web, api, core, logging, video)
- `packages/web` (Next.js) is one of them
- `packages/web/package.json` has `"type": "module"`
- `next.config.js` is ESM, imports one project-local module + uses `turbopack.root: path.resolve(__dirname, '../..')`
- Ubuntu-latest may be immune (all reproduction on darwin-arm64)

**Confirmed NOT the cause** (each ruled out independently):
- `.next` cache — wiped, reproduces
- `~/.cache/next-swc` — absent
- Fresh SWC binary — `rm -rf node_modules/.pnpm/@next+swc-darwin-arm64@16.2.6 && pnpm install --filter @bookstudio/web` refreshes the 122MB binary; reproduces
- Full monorepo reinstall — `pnpm install --force` from root; reproduces
- CPU contention — killed every unrelated Node process (Vitest workers, sibling worktrees, Playwright MCP servers); reproduces on a completely idle machine
[...]

PR #96808, opened 2026-08-06
turbo-tasks: execute scheduled tasks inline when they are read

sokra:
### What?

Two changes to how the async turbo-tasks engine handles a read whose value isn't ready yet:

1. When a read finds its task merely *scheduled* (queued, not started), it takes the task out of
   the scheduler queue and executes it on the reading thread instead of waiting for a worker to
   pick it up. If the execution completes without yielding, the read returns synchronously; if it
   yields, the partially-polled execution is handed to tokio and the read waits as before.
2. When a read finds its task's execution already **in progress** (a worker is running it), it now
   waits directly on the task's completion event instead of also attempting to take it over —
   avoiding a futile acquisition of the scheduler's queue lock, which is otherwise the most
   contended lock in the system.

An inline execution is visible in traces: the executed task's span carries
`inline_execution = "complete"` when the reader's poll finished it, or `"partial"` when it yielded
and was handed to the runtime. A task a worker executed leaves the field unset.

Optional diagnostics (`InlineExecutionStats`) count queue pushes, claim attempts and their outcomes,
and reads that waited for an already-running task. They are behind the `inline_execution_stats` Cargo
feature, off by default, so a normal build has no counters, no atomics and no extra fields.

### Why?

A large share of read misses target tasks that have merely been scheduled, not tasks actually
being computed elsewhere. Parking on those adds two avoidable thread hops (schedule → worker
pickup → wake) for work the reader could simply do itself. Conversely, a read of a task a worker
is already executing gains nothing from trying to claim it, so it shouldn't pay for trying.

### How?

- `PriorityRunner` gained a keyed claim: queued entries are indexed by a `ScheduleKey` (task or
  local task) and stored in a slot store, so one entry can be removed by key under the existing
  queue lock while the priority heap keeps a tombstone that a popping worker skips. Every scheduled
  execution still runs exactly once.
- The backend already distinguished a queued task from one already executing
  (`InProgressState::Scheduled` vs. `InProgress`) when building a listener for a read, and
  discarded that distinction. It's now threaded through as `ReadOutcome<T> = Value(T) |
  Scheduled(EventListener) | InProgress(EventListener)`; only `Scheduled` reads attempt a claim.
  The state is a hint — a worker pops a task off the queue before it marks it started, so a read can
  see `Scheduled` for a task that is no longer claimable. Acting on a stale hint costs one failed
  claim, never correctness.
- Inline execution nests (A reads B inline, B reads C inline, ...), so it's capped at 16 levels per
  thread to bound stack growth; at the cap, reads fall back to waiting for a worker as before. The
  cap is a thread-local counter because it guards the *thread* stack, not the task stack.
- Recording the outcome on the *executed task's* span needs a small handoff: when
  `poll_once_or_spawn` returns, that task's span has already been exited, so the executor registers
  the span it instruments the task body with in a slot that only exists while a claimed task is
  polled inline.
- Added `turbo-tasks-backend/tests/inline_read_execution.rs` (8 tests), `inline_execution_span.rs`
  and `inline_execution_span_worker.rs`, plus unit tests in `priority_runner.rs`/`manager.rs`
  covering keyed claim/exactly-once, inline completion for global and local tasks, the
  in-progress/scheduled distinction, nested inline executions, restoring from a persistent cache,
  and a deep dependency chain (stack-depth guard).

### Benchmarks

`bench/nested-deps-app-router-many-pages` (1000 pages, 3020 generated components, 3010 routes),
release builds of `@next/swc` for this branch and for its merge base, runs interleaved, first run of
each variant discarded as warm-up. Shared 8-vCPU VM.

**Cold `next build --turbopack`** — 5 measured runs per variant:

| metric | canary (median) | this branch (median) | delta |
| --- | ---: | ---: | ---: |
| total build | 96.70 s | 89.29 s | **−7.7 %** |
| turbopack compile phase | 68 s | 61 s | **−10.3 %** |

Raw totals (ms) — canary: 93566, 95354, 96702, 96717, 97545; branch: 89070, 89140, 89287, 89892,
90462. The distributions don't overlap (the branch is faster in all 25 pairwise comparisons, exact
two-sided Mann-Whitney p ≈ 0.008), and the 7.4 s median gap is about twice canary's own 4.0 s
run-to-run spread.

**Incremental build** (second build reusing the persistent cache) — 3 measured runs per variant:
36.14 s vs 37.29 s median, compile phase 10.4 s vs 10.5 s. The 1.2 s gap is *smaller* than canary's
own 1.2 s spread, so this is **no measurable difference** — the incremental case is dominated by
restore and by static generation in Node workers rather than by the scheduling this PR changes.

**Cache-hit reads** (`task_overhead/turbo-cached-*`, criterion): no regression. Measured before the
rebase, on the same read-path code — the hit path never enters this code, since the inline attempt
only happens on the branch where a read reports a miss.

Caveats: one app and one workload shape, on a shared VM, so only the relative comparison on the same
machine is meaningful, not the absolute seconds.


turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -33,42 +50,145 @@ where
...
+/// Items are ordered by priority in a [`BinaryHeap`], but they are stored out-of-line in `slots` so
+/// that a single item can be removed by key without disturbing the heap (a binary heap has no
+/// keyed removal). Claiming an item takes the value out of its slot and leaves the heap entry
+/// behind as a tombstone, which is skipped (and its slot recycled) when a worker pops it.
+struct Queue<P, T: Claimable> {
+    heap: BinaryHeap<HeapItem<P>>,
+    /// The queued items with their priority. A slot is `Some` while the item is queued, `None` if
+    /// it was claimed. The slot itself is only recycled once its (tombstone) heap entry has
+    /// been popped, so a slot index is never reused while it is still referenced by the heap.
+    slots: Vec<Option<(P, T)>>,
```
sokra, line 95: Why does the slot need to have P?

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -77,28 +197,39 @@ pub struct PriorityRunner<
...
             phantom: std::marker::PhantomData,
         }
     }
 
     pub fn schedule(self: &Arc<Self>, execute_context: &Arc<C>, task: T, priority: P) {
         let mut queue = self.queue.lock();
-        if !queue.is_empty() {
-            // If there is already work in the queue, we don't have any
-            // free capacity so we can just push the task to the queue.
+        if !queue.is_empty() && self.active_workers.load(Ordering::Relaxed) > 0 {
```
sokra: Why was this active_workers > 0 check added here?

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -465,12 +478,157 @@ enum ScheduledTask {
...
+/// Executing a task inline (see [`execute_read_target_inline`]) runs its body on the stack of the
+/// task that reads it, and that task might itself have been executed inline. Every level adds a
+/// task body to the stack, so the nesting is capped: at the cap, reads park and wait for a worker
+/// as they always did.
+const MAX_INLINE_EXECUTION_DEPTH: usize = 16;
+
+thread_local! {
+    /// How many task executions are currently nested inline on this thread.
+    static INLINE_EXECUTION_DEPTH: Cell<usize> = const { Cell::new(0) };
+}
```
lukesandberg, line 609: would it be possible to model this with the CurrentTaskStateInner? execute_inline could propogate a counter?

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -465,12 +478,157 @@ enum ScheduledTask {
...
+    fn drop(&mut self) {
+        INLINE_EXECUTION_DEPTH.set(INLINE_EXECUTION_DEPTH.get() - 1);
+    }
+}
+
+/// Polls `future` once on the current thread. When it doesn't complete, it is spawned so it is
+/// driven to completion as usual. Returns whether it completed.
+///
+/// The waker used for the inline poll never wakes anything, which is fine: a spawned task is always
+/// polled at least once, and that poll registers the real waker.
```
lukesandberg: this should be an inline comment on the `noop` waker not a doc comment

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -900,21 +1058,50 @@ impl<B: Backend + 'static> TurboTasks<B> {
...
+    fn try_execute_scheduled_task_inline(&self, key: ScheduleKey) -> bool {
+        let this = self.pin();
+        InlineExecutionCounters::bump(&self.inline_counters.claim_attempted);
+        // An execution that doesn't complete inline is handed over to the runtime, so there has to
+        // be one. Reads from outside a runtime keep waiting for a worker, as they always did.
+        if tokio::runtime::Handle::try_current().is_ok()
+            && let Some(future) = self.priority_runner.claim(&this, &key)
+        {
+            return if poll_once_or_spawn(future) {
+                InlineExecutionCounters::bump(&self.inline_counters.claim_completed_inline);
```
lukesandberg: nit:  i would prefer `claim_completed` and `claim_yielded` functions on InlineExecutionCounters instead of this generic `bump` function and accessing the underlying counter.

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -900,21 +1058,50 @@ impl<B: Backend + 'static> TurboTasks<B> {
...
+    ///
+    /// When the task is not in the queue (a worker already picked it up, it was never scheduled, or
+    /// it is already done) nothing is executed and `false` is returned. When the execution doesn't
+    /// complete on the first poll, it is spawned to be completed as usual.
+    fn try_execute_scheduled_task_inline(&self, key: ScheduleKey) -> bool {
+        let this = self.pin();
+        InlineExecutionCounters::bump(&self.inline_counters.claim_attempted);
+        // An execution that doesn't complete inline is handed over to the runtime, so there has to
+        // be one. Reads from outside a runtime keep waiting for a worker, as they always did.
+        if tokio::runtime::Handle::try_current().is_ok()
```
lukesandberg: when do we have reads from outside the runtime?  these would be top level calls from `run` closures?

i think it would be safe to always assume this is called in a tokio context?

turbopack/crates/turbo-tasks/src/manager.rs
lukesandberg: i guess it is also inconsistent with the `bump_inline_counter!` macro

but still dedicated boring functions are more readable imho

turbopack/crates/turbo-tasks/src/manager.rs
lukesandberg: it would be kind of neat if we could tag executions with a span attribute that said: 'inline executed: partial' or 'inline executed: complete'

i think storing on the CurrentTaskState would enable this

turbopack/crates/turbo-tasks/src/priority_runner.rs
lukesandberg: so claims can access the priority which is then passed into the CurrentTaskState

turbopack/crates/turbo-tasks/src/manager.rs
sokra: Storing this on a thread local makes more sense, than a task_local (which CurrentTaskState is). Inline execution depth is trying to avoid big (thread) stack and not big task stack. Also it can partially execute inline, which would mean the value of inline execution depth would change, therefore requiring a lock. Thread local doesn't require that.

turbopack/crates/turbo-tasks/src/manager.rs
lukesandberg: a panic from inside of the runtime kills the process, so this isn't a liveness concern, it is a clear bug signal given the constraints

so spawn panicing is the behavior we want

turbopack/crates/turbo-tasks-backend/tests/inline_read_execution.rs
```diff
@@ -0,0 +1,411 @@
...
+                .await
+        }
+    })
+    .await
+    .unwrap();
+}
+
+#[turbo_tasks::function(operation, root)]
+async fn read_scheduled_task_inline(nonce: u32) -> Result<Vc<()>> {
+    let _ = nonce; // ensure the nonce is part of our cache key
```
lukesandberg: is this necessary given that nonce is passed to leaf below?

turbopack/crates/turbo-tasks-backend/tests/inline_execution_span_collector/mod.rs
lukesandberg, line 1: nit: it is good to verify that the span annotations work but i don't believe these tests are worth commiting

just drop them, the

turbopack/crates/turbo-tasks-backend/tests/inline_read_execution.rs
```diff
@@ -0,0 +1,411 @@
...
+                .await
+        }
+    })
+    .await
+    .unwrap();
+}
+
+#[turbo_tasks::function(operation, root)]
+async fn read_yielding_task_operation(nonce: u32) -> Result<Vc<()>> {
+    let _ = nonce;
```
lukesandberg: ditto

turbopack/crates/turbo-tasks-backend/tests/inline_read_execution.rs
```diff
@@ -0,0 +1,411 @@
...
+    .await
+    .unwrap();
+    assert_eq!(
+        COUNTED_EXECUTIONS.load(Ordering::SeqCst),
+        1,
+        "the task must be executed exactly once, no matter how many readers race for it"
+    );
+}
+
+static COUNTED_EXECUTIONS: AtomicUsize = AtomicUsize::new(0);
```
lukesandberg, line 331: how does this work given parallel execution under `cargo nextest`

i guess a comment that this can only be used by one test case?  or introducing scoping via a test module

turbopack/crates/turbo-tasks/src/vc/raw.rs
```diff
@@ -502,6 +504,18 @@ fn suppress_top_level_task_check<R>(strongly_consistent: bool, f: impl FnOnce()
...
 }
 
+/// Executes the task a read is waiting for when it is only scheduled, so the read can continue
+/// without waiting for a worker.
+///
+/// The read loops in [`ResolveRawVcFuture`] / [`ReadRawVcFuture`] report the task to execute
+/// through an out parameter and this is called *outside* of [`with_turbo_tasks`]: the read borrows
+/// the `TURBO_TASKS` task-local, and executing a task enters a task-local scope of its own, which
+/// is not allowed while it is borrowed. Reporting it out-of-band (instead of returning it) keeps
+/// the value path of a read — the hot path — free of any extra work.
```
lukesandberg: shorten this comment

none of it is wrong really but it is verbose

turbopack/crates/turbo-tasks/src/vc/raw.rs
```diff
@@ -587,22 +618,39 @@ impl Future for ResolveRawVcFuture {
...
-        suppress_top_level_task_check(this.strongly_consistent, || with_turbo_tasks(poll_fn))
+        loop {
+            let mut execute_inline_key = None;
+            // HACK: Temporarily suppress top-level task check if doing strongly consistent read.
+            //
+            // This masks a bug: There's an unlikely TOCTOU race condition in `poll_fn`. Because the
+            // strongly consistent read isn't a single atomic operation, any inner `TaskOutput` or
+            // `TaskCell` could get mutated after the strongly consistent read of the outer
+            // `TaskOutput`.
+            let result = suppress_top_level_task_check(strongly_consistent, || {
```
lukesandberg, line 643: nit: defer `result` and the comment to the None block?

also why does execute_inline not need the  `suppress_top_level_task_check`?

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -161,6 +163,21 @@ pub trait TurboTasksApi: TurboTasksCallApi + Sync + Send {
...
+    ///
+    /// Used by the read paths; see `TurboTasks::try_execute_scheduled_task_inline` for the details.
+    fn try_execute_scheduled_task_inline(&self, key: ScheduleKey) -> bool;
+
+    /// Records that a read waited for a task that a worker was already executing, so it did not try
+    /// to claim it. Diagnostics only, see `TurboTasks::inline_execution_stats`.
+    ///
+    /// Without the `inline_execution_stats` feature this is an empty function. It stays on the
+    /// trait unconditionally because the path that calls it is about to park on an
+    /// [`EventListener`] anyway, so one no-op call is noise next to that.
```
lukesandberg: lets just cfg guard the definition which is more idiomatic and self documenting we will have to propogate the cfg guards to the callers but that is a trivial call site overhead which more than pays for itself by not having to write the justification comment

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -465,12 +482,209 @@ enum ScheduledTask {
...
+    });
+    *REQUESTED
+}
+
+/// Maximum number of task executions that may be nested inline on a single thread.
+///
+/// Executing a task inline (see [`execute_read_target_inline`]) runs its body on the stack of the
+/// task that reads it, and that task might itself have been executed inline. Every level adds a
+/// task body to the stack, so the nesting is capped: at the cap, reads park and wait for a worker
+/// as they always did.
```
lukesandberg: shorten this

the limitation is to conserve stack space, we don't need to describe where that happens

the alternative is automatically growing the stack like SWC

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -465,12 +482,209 @@ enum ScheduledTask {
...
+    fn drop(&mut self) {
+        INLINE_EXECUTION_DEPTH.set(INLINE_EXECUTION_DEPTH.get() - 1);
+    }
+}
+
+/// Polls `future` once on the current thread. When it doesn't complete, it is spawned so it is
+/// driven to completion as usual. Returns whether it completed.
+///
+/// The outcome is also recorded on the span of the task that was executed, see
+/// [`InlineExecutionSpanSlot`].
```
lukesandberg: the doc here is sloppy

`
/// Polls future once inline and then spawns it if it doesn't complete so tokio drives it

`

should be enough

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -465,12 +482,209 @@ enum ScheduledTask {
...
+/// Called by a read that found its task merely *scheduled*: executes it on the reading thread, so
+/// the read can go on without waiting for a worker. Returns whether the read should be retried
+/// because the task's execution completed.
+///
+/// Reads never get here for a task a worker is already executing ([`ReadOutcome::InProgress`]), so
+/// the scheduler's queue lock is not touched in that case.
+///
+/// This is the only place where reads pay for inline execution: a read that finds its value does no
+/// extra work at all, which is why this is kept out of the caller's hot path.
+#[inline(never)]
```
lukesandberg: why never inline?

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -465,12 +482,209 @@ enum ScheduledTask {
...
+
+/// Called by a read that found its task merely *scheduled*: executes it on the reading thread, so
+/// the read can go on without waiting for a worker. Returns whether the read should be retried
+/// because the task's execution completed.
+///
+/// Reads never get here for a task a worker is already executing ([`ReadOutcome::InProgress`]), so
+/// the scheduler's queue lock is not touched in that case.
+///
+/// This is the only place where reads pay for inline execution: a read that finds its value does no
+/// extra work at all, which is why this is kept out of the caller's hot path.
```
lukesandberg: shorten comment

Executes the task inline if possible, returns true if it executed to completion.

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -2284,7 +2618,122 @@ pub(crate) async fn read_local_output(
...
+                    continue;
+                }
+                event_listener.await
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lukesandberg, line 2608: some of these tests are low value and overlapping

e.g. asserting that the span slot is stored in a task_local

lukesandberg (approved):
Everything looks basically good but many of the comments are overly verbose and i question the value some of the tests

koding88:
Follow-up: this fix is confirmed working on canary (16.4.0-canary.6 — 0% CPU idle, 200 in ~2s) but **hasn't shipped in any stable release yet** (latest stable is 16.3.2, released 2026-08-21, before this merge).

The regression it fixes is still live on every stable from 16.2.6 onward: `next dev --turbo` in pnpm workspace monorepos on darwin-arm64 livelocks — `next-server` climbs to 603–777% CPU within ~30s of `✓ Ready`, RSS grows to 3.5GB, and no request ever resolves. `sample` shows all 11 `tokio-runtime-worker` threads 100% busy inside `next-swc.darwin-arm64.node` (~2900/6300 samples in `_pthread_cond_wait` on the scheduler queue lock), matching the "most contended lock in the system" this PR addresses.

The original report (#95798) was auto-closed by the stale-issue bot for lacking a public repro link and is now locked, so users hitting this on stable have no open thread to track. Is there a plan to backport this to a 16.3.x patch? A lot of macOS monorepo users are currently stuck on `--webpack` as a workaround.

Issue #12105, opened 2020-04-22
scroll to top after following a link with firefox


bpetitcollot:
# Bug report

## Describe the bug

Using Firefox, when i click on a Link near to the bottom of a page, window doesn't scroll to the top of the page after it is loaded.

## To Reproduce

This bug can be reproduced on nextjs.org

1. Go to https://nextjs.org/learn/basics/create-nextjs-app/setup
2. Scroll to the bottom of the page
2. Click on 'Next >' blue button
4. Page 3 of the lesson appears but window doesn't scroll back to the top

## Expected behavior

Window should scroll to the top of the page

## Screenshots

Here is an mp4 screencast

[linkNext.zip](https://github.com/zeit/next.js/files/4516295/linkNext.zip)

## System information

Browser: Firefox

bug reproduced with
- Firefox 68.6.0esr / Debian 10
- Firefox 74.0.1 / Windows

it seems this bug doesn't occur with 
- Firefox 74.0.1 / Ubuntu 18.04

loicbourg:
Hello,

I reproduced this bug in my app with next.js 9.4.1 .


Is this an expected behavior ?

loicbourg:
Hello,

For information, i tried to enable the new experimental option *scrollRestoration* but it does not fix the bug.

franky47:
I have the same issue, on:

- Firefox 80.0.1
- macOS 10.14.6
- Next.js 9.5.3

Demo on my website:
- Go to <https://preview.francoisbest.com>
- Scroll to the end of the page
- Click "More albums"
- The browser navigates to `/music` using a Next link (the `scroll` prop is not passed explicitly, it uses the default value), and should scroll to the top.

Scrolling occurs on Chrome (with an animation) and on Safari the page loads already at the top, but Firefox gets stuck to the bottom of the page.

mstruebing:
This is the case if we use `scroll-behavior: smooth;` in our CSS.
If we remove it, it is working also in firefox.

Timer:
Closing as high likelihood duplicate of #6462, as I cannot reproduce.

alextseitlin:
I have the issue on my end too. Works on chrome but not scrolling of firefox

albert-hajdu-mondd-io:
Thanks @mstruebing removing `scroll-behavior: smooth;` fixed the issue.

TBH smooth scrolling on navigation felt weird anyways (especially with page transitions) so I'm not missing it.

58bits:
[...]

PR #96820, opened 2026-08-06
[turbopack] Reduce native React Compiler work

marcoshernanz:
## What?

- Use the released `swc_ecma_react_compiler::fast_check::is_required` API to skip React Compiler work for modules that cannot change in native `infer` mode.
- Keep explicit `annotation` and `all` modes unconditional.
- Delete Next.js's duplicate React Compiler predicates and use the same upstream check from the native N-API binding.
- Fail open from the N-API check on unreadable files, fatal parse failures, and recovered parser errors.
- Keep the client-runtime-only compiler out of App SSR, matching the existing Babel integration.
- Move the workspace to the coherent SWC 76 dependency family, including the official `mdxjs-rs-turbopack` branch, with no duplicate SWC 75 stack.
- Keep the React Compiler dependency/module native-only; the WASM facade already deliberately fails open.

This consumes the implementation landed in [swc-project/swc#12105](https://github.com/swc-project/swc/pull/12105), released in `swc_ecma_react_compiler` 23.0.0. Its contract is conservative: false positives add compiler work, while a negative result guarantees that compilation cannot alter the program.

## Why?

The native compiler currently runs for every parsed module in infer mode and is configured for both browser and App SSR contexts. This adds avoidable AST conversion/compiler work and compiles hydrated client modules in a server context that the existing Turbopack Babel path excludes.

This matters for large applications such as v0. Their legacy Babel React Compiler path creates a separate machine-sized Node worker pool in addition to PostCSS. Moving to the native path removes that pool; this patch makes the native path cheaper while preserving the established browser-only scope.

### Upstream fast check on the v0 corpus

The final SWC predicate was checked against 1,816 real v0 modules (10.46 MB):

- selected 302 modules
- retained all 257 modules whose output actually changed
- **zero false negatives**
- scanned the full corpus in 20.624 ms
- reduced the measured full compiler pipeline from 2,371.956 ms to 1,906.180 ms: **-19.64%**

### This Next.js patch only

Real v0 homepage, cold `.next`, Chromium interaction gate, 16 vCPU / 32 GB, exact Next canary `7916855653`, exact v0 commit `1ab042a47c`, three samples per arm. Both arms use the native compiler.

| Median | Unpatched native | This PR | Change |
|---|---:|---:|---:|
| Visible interactive UI | 50.737 s | 48.574 s | **-4.26%** |
| Network quiet | 70.125 s | 67.546 s | **-3.68%** |
| Aggregate CPU | 450 s | 418 s | **-7.11%** |
| Peak process-tree RSS | 7,904,772 KB | 7,823,008 KB | -1.03% |

All three paired visible, quiet, and CPU deltas favored the patch. The RSS difference is inside noise, so this PR makes no patch-only memory claim.

### v0 legacy Babel path to native + this PR

Both arms use `TURBO_ENGINE_EVICT_AFTER_SNAPSHOT=1` and remain open for 120 seconds after network quiet.

| Median | Legacy Babel | Native + this PR | Change |
|---|---:|---:|---:|
| Visible interactive UI | 52.411 s | 48.372 s | **-7.71%** |
| Peak process-tree RSS | 12.19 GiB | 7.39 GiB | **-39.40%** |
| RSS at +120 s | 5.31 GiB | 3.02 GiB | **-43.07%** |
| Aggregate CPU | 480 s | 429 s | **-10.63%** |
| Processes at +120 s | 16 | 4 | **-75.00%** |

The large memory/process delta comes from switching v0 off the external Babel pool, not from this source diff alone. v0 still needs to enable the native compiler configuration separately.

Every accepted browser run required HTTP 200, a visible nonblank UI with at least ten controls, zero Next overlays/page errors, and a verified prompt interaction. The complete Next process session was sampled every 100 ms.

## Validation

- affected Rust `cargo check`
- affected Clippy with warnings denied, including Turbopack tests
- React Compiler unit tests: 2 passed
- React Compiler integration tests: 3 passed
- Turbopack snapshots: 114 passed
- direct native N-API/Node smoke: wrapped JSX and member hooks select; plain modules skip; recovered parse errors and missing files fail open
- dependency trees: native includes `swc_ecma_react_compiler`; WASI excludes it; no `swc_core` 75 remains
- official `mdxjs-rs-turbopack` SWC 76 bump: 188 tests passed
- final GPT-5.6 Sol + Claude Opus 5 xhigh autoreview: clean, 0.98 confidence from both reviewers
- real-v0 cold browser matrix, 120-second memory soak, and full production build: 44 Workflow steps / 10 workflows, 2,722/2,722 prerenders


marcoshernanz:
Upstream follow-up is now open as [swc-project/swc#12105](https://github.com/swc-project/swc/pull/12105). I compared SWC\047s existing \140is_required\140 against this PR\047s conservative predicate on the full v0 corpus. The existing SWC heuristic is not safe as a skip gate: it missed 10 of 257 actually transformed modules (nine \140forwardRef\140 modules plus one member-hook module). The new upstream \140may_require\140 had zero false negatives, selected the same 315 files as this PR, and was 1.52% faster than the local visitor across five detector-only runs. It also moves the Node binding to the conservative contract and treats parse failures conservatively. This Next.js draft should keep the local implementation until the SWC crate release containing that API is available, then switch both Turbopack and next-napi to the upstream predicate and delete the duplicate.

crates/next-core/src/next_server/context.rs
```diff
@@ -721,8 +708,9 @@ pub async fn get_server_module_options_context(
                     enable_jsx: Some(jsx_runtime_options),
                     enable_typescript_transform: Some(tsconfig),
                     enable_decorators: Some(decorators_options.to_resolved().await?),
-                    enable_rust_react_compiler,
-                    rust_react_compiler_target,
+                    // React Compiler optimizes client runtime semantics. Keep App SSR aligned with
+                    // the existing Babel integration, which only runs it for the browser context.
```
lukesandberg: shorten comment

// React compiler only optimizes the react client runtime, skip it.

crates/next-custom-transforms/src/react_compiler.rs
lukesandberg, line 1: per slack disucssion, leverage the fast-check api from SWC instead of these predicates.

crates/next-core/src/next_server/context.rs
marcoshernanz: Done in `3b98e0a965`: `// React Compiler only optimizes the React client runtime, so skip it.`

crates/next-custom-transforms/src/react_compiler.rs
marcoshernanz: Agreed. The upstream API is now proposed in swc-project/swc#12105. Next currently pins released `swc_ecma_react_compiler` 20.1.0, whose only fast check is unsafe as a skip gate here (10/257 false negatives on the v0 corpus). I am not switching this PR to that predicate or to a fork dependency. Once #12105 lands and a crate containing `fast_check::may_require` is published, I will bump the crate, switch both Next callers to the SWC API, delete this local predicate module, and rerun the v0 benchmark. Leaving this thread open until that upstream dependency is available.

crates/next-custom-transforms/src/react_compiler.rs
```diff
@@ -21,12 +31,81 @@ struct Finder {
...
+                | Expr::JSXEmpty(..)
+                | Expr::JSXElement(..)
+                | Expr::JSXFragment(..)
+        ) || matches!(
+            node,
+            Expr::Lit(Lit::Str(value))
+                if value
+                    .value
+                    .as_str()
+                    .is_some_and(|value| matches!(value, "use memo" | "use forget"))
```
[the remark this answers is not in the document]
marcoshernanz: This is a false positive: the string value here is SWC's `Wtf8Atom`, whose `as_str()` returns `Option<&str>`, so `is_some_and` is the intended API. The branch compiles locally and Next's `build`, `rust check`, Clippy, native, WASM, and unit-test CI jobs all passed on this exact code. The suggested dereference is for `Atom`, not `Wtf8Atom`.

crates/next-custom-transforms/src/react_compiler.rs
marcoshernanz: Update after SWC maintainer review: swc-project/swc#12105 now improves the existing `fast_check::is_required` instead of adding `may_require`. The final SWC implementation still has the conservative contract and the same v0 correctness result (302 selected, 257/257 transforms, zero false negatives). Once a release containing commit `6f72747ac0` is available, this PR can bump SWC, call `is_required` from both Next sites, and delete the local predicates. Keeping this thread open until that release-dependent migration is possible.

marcoshernanz: Addressed in c18d747464. This now consumes the released `swc_ecma_react_compiler::fast_check::is_required` from 23.0.0 in both Turbopack infer mode and the native N-API binding, deletes the duplicate Next.js predicate module, and keeps explicit annotation/all modes unconditional. I also aligned the workspace and the official MDX fork on SWC 76 so there is no duplicate SWC 75 stack. The final v0 corpus check retains 257/257 transforming modules with zero false negatives, and the complete affected Rust test/snapshot matrix plus final two-model autoreview is green.

marcoshernanz:
This is ready for final maintainer review/merge from the implementation and CI side.

- consumes the released SWC React Compiler `fast_check::is_required` API and removes the duplicate Next.js predicates
- preserves explicit annotation/all semantics and fail-open behavior on every native read/parse failure
- retains the browser-only App Router scope requested in review
- aligns Next.js and the official MDX fork on one SWC 76 stack, with the React Compiler dependency excluded from WASI
- final v0 corpus result: 257/257 transforming modules retained, zero false negatives, and a 19.64% reduction in the measured compiler pipeline
- all review threads resolved; final GPT-5.6 Sol + Claude Opus 5 xhigh autoreview clean at 0.98 confidence each
- fresh CI on `c18d747464`: 114 passed, 17 intentionally skipped, 0 failed (the unrelated navigation timing failure passed its isolated rerun)

The PR is mergeable with current canary. It remains a draft/review-required PR, so maintainer approval and the normal ready-for-review state transition are the only remaining gates.

crates/next-napi-bindings/src/react_compiler.rs
```diff
@@ -25,7 +24,7 @@ impl Task for CheckTask {
             //
             let cm = Arc::new(SourceMap::default());
             let Ok(fm) = cm.load_file(&self.filename) else {
-                return Ok(false);
+                return Ok(true);
```
lukesandberg: why switch these to true?  no need to run react-compiler on broken or missing files?

turbopack/crates/turbopack-ecmascript/src/transform/mod.rs
```diff
@@ -458,6 +466,13 @@ async fn apply_rust_react_compiler(
         return Ok(helpers);
     };
 
+    // Match the legacy React Compiler loader: in infer mode, avoid invoking the compiler for
+    // modules that cannot contain a component or hook. The detector runs on the SWC AST we
+    // already parsed, so this is cheaper than converting every module to the compiler AST.
+    if !should_run_rust_react_compiler(program, compilation_mode) {
```
lukesandberg, line 551: should we process all compilation_mode values? or just infer?

seems odd either way

it looks like the native compiler only checks for 'annotations' after the AST conversion.  so doing our own scan for `'use memo'` here would be much much faster... do it?

TODO is also fine

lukesandberg (approved):

crates/next-napi-bindings/src/react_compiler.rs
marcoshernanz: Fixed in 01a3e986b9. File-load failures and fatal parse failures now return false, so missing or broken files do not invoke React Compiler. I kept only recovered parser diagnostics fail-open: in that case SWC produced an AST and Babel may accept the source differently, so returning false could incorrectly skip required compilation because of a parser difference. The code now documents that distinction.

turbopack/crates/turbopack-ecmascript/src/transform/mod.rs
marcoshernanz: Implemented in 01a3e986b9. The pre-conversion check is now mode-specific: infer uses the upstream SWC fast check, annotation recursively scans function directive prologues for use memo / legacy use forget before AST conversion, and all remains unconditional. This matches the native compiler options used here, including dynamic_gating: None. I added coverage for nested annotations, misplaced and module-level directives, dynamic-gating directives, JSX/hook-only negatives, and all-mode behavior.

marcoshernanz:
Luke review follow-up is complete in signed commit 01a3e986b9.

Changes:
- Missing files and fatal SWC parse failures now skip React Compiler. Recovered parser diagnostics remain fail-open because SWC produced an AST and Babel parser behavior can differ.
- The pre-conversion gate now handles every compilation mode explicitly: infer uses SWC fast_check::is_required, annotation recursively scans real function directive prologues for use memo / legacy use forget, and all remains unconditional.
- Annotation coverage includes nested functions, misplaced and module-level directives, dynamic-gating directives under the current dynamic_gating: None options, JSX/hook-only negatives, and all-mode behavior.

Verification:
- cargo fmt --all -- --check
- cargo clippy -p turbopack-ecmascript --lib --tests -- -D warnings -A deprecated
- cargo test -p turbopack-ecmascript --lib react_compiler_tests: 4/4
- cargo test -p turbopack-tests --test react_compiler: 3/3
- cargo check -p next-napi-bindings
- GPT-5.6 Sol xhigh + Claude Opus 5 xhigh autoreview: 0 findings, patch correct at 0.98 confidence
- Fresh GitHub CI: 115 passed, 17 intentional skips, 0 pending/failing. One unrelated app-action test process hung until the fixed 30-minute job timeout; the preserved-cache isolated retry passed in 2m56s.

All review threads are resolved, Luke approval remains valid, and GitHub reports MERGEABLE / CLEAN. Ready to merge.

lukesandberg (approved):

PR #95111, opened 2026-06-24
Fix PPR not working with Cache Components and legacy builders

eps1lon:
Required until changes like https://github.com/vercel/vercel/pull/16706 land and we dropped support for using versions prior.

See https://github.com/vercel/next.js/pull/94955/changes#r3465285336

## Test plan

- [x] e2e release deploy tests using a version from this branch are green: https://github.com/vercel/next.js/actions/runs/28083525033

PR #96827, opened 2026-08-06
Use Cache Components as the internal PPR signal

ztanner:
## Summary

`experimental.ppr` is deprecated, and configuring it with a truthy value already produces an error. However, enabling Cache Components still backfilled this option internally, and several build and runtime paths continued treating it as the source of truth for PPR.

This change:

- uses `cacheComponents` directly for application and route PPR checks
- removes the internal `experimental.ppr` backfill and associated plumbing
- deletes the obsolete PPR configuration helpers
- removes legacy PPR telemetry and an unreachable non-Cache-Components workaround
- preserves the deprecated config type, schema, and migration error for user-provided `experimental.ppr`

This does not change supported behavior: Cache Components already implied `experimental.ppr: true` internally, so these checks resolve to the same value.


packages/next/src/server/config.ts
```diff
@@ -1600,11 +1600,6 @@ function assignDefaultsAndValidate(
     result.experimental.mcpServer = true
   }
 
-  if (result.cacheComponents) {
-    // TODO: remove once we've finished migrating internally to cacheComponents.
-    result.experimental.ppr = true
-  }
```
eps1lon, line 1606: I think we need https://github.com/vercel/next.js/pull/95111 for older versions of the builder. At least until we completely dropped support for legacy builder. Only newer versions of the builder use `config.cacheComponents` i.e. versions that have https://github.com/vercel/vercel/pull/16706

ztanner: Makes sense to me -- I cherrypicked that in

gnoff (approved):

PR #96834, opened 2026-08-06
Turbopack: serialize moduleLoading.crossOrigin as null instead of "none" in client reference manifest

### LINKED ISSUE #96831: 16.3.0: Turbopack serializes moduleLoading.crossOrigin as string "none", adding unexpected crossorigin="" to chunk scripts (breaks cross-origin assetPrefix CDNs)

https://github.com/banqinghe/next-16-3-crossorigin-none-repro

### To Reproduce

1. `npm install` (installs `next@16.3.0`), `npm run build`
2. `npm run static` — serves `.next/static` on `http://127.0.0.1:9999` **without** CORS headers (emulating a CDN / the cached-without-ACAO case)
3. `npm run start` — app on `http://localhost:3400` with `assetPrefix: "http://127.0.0.1:9999"` (different origin)
4. Open `http://localhost:3400` in a browser, or `curl -s http://localhost:3400/ | grep -o "<script[^>]*>"`

**Current (16.3.0):** SSR-preinited client component chunk scripts are rendered with an unexpected `crossorigin=""` attribute, even though `crossOrigin` is **not** set in `next.config.js`:

```html
<script src="http://127.0.0.1:9999/_next/static/chunks/3fntmmi971322.js" async="" crossorigin=""></script>
<script src="http://127.0.0.1:9999/_next/static/chunks/13srbf737j4ud.js" async="" crossorigin=""></script>
```

The browser therefore fetches these chunks in CORS mode and refuses to run them when the asset host does not reply with `Access-Control-Allow-Origin`:

```
Access to script at http://127.0.0.1:9999/_next/static/chunks/....js from origin
http://localhost:3400 has been blocked by CORS policy: No Access-Control-Allow-Origin
header is present on the requested resource.
```

Note only the *preinited* chunks get the attribute — the layout/page `<script async>` tags and the bootstrap script on the same page do not — so the same chunk URLs are requested in a mix of no-cors and CORS modes. On real CDNs whose cache key does not include the `Origin` header, the no-cors responses (cached without ACAO) poison the CORS-mode loads even when the origin/S3 CORS configuration is correct. This broke our production (www + CDN on different origins) after upgrading, and we had to roll back.

**Expected (16.2.12 behavior):** no `crossorigin` attribute unless `crossOrigin` is configured. Downgrading `next` to 16.2.12 in the same repro and rebuilding yields zero `crossorigin` attributes.

### Root cause (from bisecting the build output)

The Turbopack-generated client reference manifest changed how `moduleLoading.crossOrigin` is serialized:

```
# .next/server/app/page_client-reference-manifest.js
16.2.12: "moduleLoading":{"prefix":"","crossOrigin":null}
16.3.0:  "moduleLoading":{"prefix":"","crossOrigin":"none"}
```

The string `"none"` looks like a Rust `Option::None` / enum variant leaking into JSON. React Flight only checks `typeof crossOrigin === "string"` and normalizes any string other than `"use-credentials"` to `""` (anonymous), so `"none"` becomes `crossorigin=""` on every chunk script emitted through `prepareDestinationWithChunks` / `preinitScriptForSSR`.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin 25.2.0
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 22.23.1
  npm: 10.9.8
Relevant Packages:
  next: 16.3.0
  react: 19.2.0
  react-dom: 19.2.0
  typescript: N/A
Next.js Config:
  assetPrefix: http://127.0.0.1:9999 (cross-origin)
```

Turbopack, Runtime

next build (local), Other (production serving behind a cross-origin CDN assetPrefix)

banqinghe:
Fixes #96831

### What?

Since 16.3.0, the Turbopack-generated client reference manifest serializes `moduleLoading.crossOrigin` as the string `"none"` when `crossOrigin` is not configured:

```
16.2.12: "moduleLoading":{"prefix":"","crossOrigin":null}
16.3.0:  "moduleLoading":{"prefix":"","crossOrigin":"none"}
```

`CrossOrigin::None` (the `#[default]` variant) goes through the enum's kebab-case serde derive and comes out as `"none"`.

### Why is that a problem?

React Flight only checks `typeof crossOrigin === "string"` and normalizes any string other than `"use-credentials"` to `""` (anonymous). So every SSR-preinited client component chunk `<script>` is emitted with an unexpected `crossorigin=""`, and browsers fetch those chunks in CORS mode.

For apps with a cross-origin `assetPrefix` (CDN), this newly requires `Access-Control-Allow-Origin` on chunk responses. Worse, other references to the same chunk URLs (layout/page script tags, bootstrap, dynamic imports) still load in no-cors mode, so CDNs whose cache key does not include the `Origin` header end up serving no-cors-cached copies (without ACAO) to the CORS-mode loads, which the browser then blocks:

```
Access to script at 'https://cdn.example.com/_next/static/chunks/xxx.js' from origin
'https://www.example.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin'
header is present on the requested resource.
```

This took down our production frontend (www + CDN on separate origins) after upgrading to 16.3.0 and we had to roll back. Reproduction: https://github.com/banqinghe/next-16-3-crossorigin-none-repro

### How?

Serialize the field through the existing `CrossOrigin::as_str()` (`None` → `null`, `Anonymous` → `"anonymous"`, `UseCredentials` → `"use-credentials"`), restoring the 16.2.x manifest shape. Added a unit test covering all three variants.

### DIFF
--- a/crates/next-core/src/next_manifests/client_reference_manifest.rs
+++ b/crates/next-core/src/next_manifests/client_reference_manifest.rs
@@ -70,9 +70,8 @@ pub struct CssResource {
 #[serde(rename_all = "camelCase")]
 pub struct ModuleLoading {
     pub prefix: RcStr,
-    /// React Flight treats any string as CORS-enabled (normalizing everything
-    /// but "use-credentials" to anonymous), so `CrossOrigin::None` must
-    /// serialize as `null`, not as the string "none".
+    /// React treats any string here as CORS-enabled, so `CrossOrigin::None`
+    /// must serialize as `null`, not `"none"`.
     #[serde(serialize_with = "serialize_cross_origin")]
     pub cross_origin: CrossOrigin,
 }
--- a/crates/next-core/src/next_manifests/client_reference_manifest.rs
+++ b/crates/next-core/src/next_manifests/client_reference_manifest.rs
@@ -70,7 +70,21 @@ pub struct CssResource {
 #[serde(rename_all = "camelCase")]
 pub struct ModuleLoading {
     pub prefix: RcStr,
+    /// React Flight treats any string as CORS-enabled (normalizing everything
+    /// but "use-credentials" to anonymous), so `CrossOrigin::None` must
+    /// serialize as `null`, not as the string "none".
+    #[serde(serialize_with = "serialize_cross_origin")]
     pub cross_origin: CrossOrigin,
+}
+
+fn serialize_cross_origin<S>(
+    cross_origin: &CrossOrigin,
+    serializer: S,
+) -> std::result::Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    cross_origin.as_str().serialize(serializer)
 }
 
 #[derive(Serialize, Default, Debug, Clone)]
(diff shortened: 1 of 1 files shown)


mischnic:
This has been fixed by https://github.com/vercel/next.js/pull/97164

PR #96826, opened 2026-08-06
Replace CSRBailout error with `ReactDOM.browser` behind a flag

devjiwonchoi:
CSRBailout error is the use case for `ReactDOM.browser()`. Replace behind an experimental flag.
This PR replaces the bailout error used in next/dynamic ssr: false.

PR #96843, opened 2026-08-06
Replace useSearchParams bailout error with ReactDOM.browser behind a flag

devjiwonchoi:
Stacked on #96826.

This PR replaces the bailout error used by `useSearchParams` with `ReactDOM.browser()` behind the experimental flag.

### How?

Move the bailout behind a browser-safe module boundary, use `ReactDOM.browser()` when enabled, and preserve existing behavior otherwise.

### DIFF
--- a/packages/next/src/client/components/navigation-dynamic-rendering.ts
+++ b/packages/next/src/client/components/navigation-dynamic-rendering.ts
@@ -6,6 +6,7 @@
 // the server module is not bundled into the client. Callers use optional calls
 // (`useDynamicRouteParams?.(...)`), so the browser stub is a no-op.
 import React from 'react'
+import { browser } from 'react-dom'
 
 import { BailoutToCSRError } from '../../shared/lib/lazy-dynamic/bailout-to-csr'
 import { InvariantError } from '../../shared/lib/invariant-error'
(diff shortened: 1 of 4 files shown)


packages/next/src/client/components/navigation-dynamic-rendering.ts
```diff
@@ -5,7 +5,81 @@
...
+      // During instant validation we try to behave as close to client as possible,
+      // so this shouldn't hang during SSR.
+      return
+    case 'prerender-client': {
+      React.use(
+        makeClientHookHangingPromise(
+          workUnitStore.renderSignal,
+          new ClientHookDynamicError(workStore.route, expression)
+        )
+      )
```
lubieowoce: @gnoff is `browser()` meant to replace this one too? rn the PR only changes the `prerender-legacy` branch, i.e. only pre-`cacheComponents` code

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -9665,6 +9665,19 @@ async function prerenderToStream(
       throw err
     }
 
+    if (err instanceof Error) {
```
lubieowoce: hmm i'm not sure about this, feels like we should be tracking what `onBrowserBailout` was called with instead? then we can get rid of the `$$typeof` check altogether:

```ts
const browserBailouts = new Set<unknown>()

prerender(..., {
  onBrowserBailout(err) {
    browserBailoutReasons.add(err)
  }
})
```
and then
```ts
if (browserBailoutReasons.has(err)) {
  // log the error
}
```
although hmm, maybe that's not enough, as it'd trigger for userspace `browser()` as well

i guess we want to identify bailouts created by next itself? then i suppose having a special object is fine, but we should probably also use the same wrapper object inside `BailoutToCSR` (changed in the previous PR)?

devjiwonchoi: Yeah good catch, we should use that from PR 1

gnoff (approved):

PR #96844, opened 2026-08-06
Replace resumed render bailout error with ReactDOM.browser behind a flag

devjiwonchoi:
Stacked on #96843.

This PR replaces the bailout error used to abort resumed renders with `ReactDOM.browser()` behind the experimental flag.

### How?

Pass the flag to the App Router runtime, use `ReactDOM.browser()` as the abort reason when enabled, and preserve existing behavior otherwise.

### DIFF
--- a/packages/next/src/server/app-render/stream-ops.ts
+++ b/packages/next/src/server/app-render/stream-ops.ts
@@ -44,6 +44,8 @@ export const processPrelude = _m.processPrelude
 export const createWebInlinedDataStream = _m.createWebInlinedDataStream
 export const createNodeInlinedDataStream = _m.createNodeInlinedDataStream
 export const createPendingStream = _m.createPendingStream
+export const createRenderInBrowserAbortSignal =
+  _m.createRenderInBrowserAbortSignal
 export const createOnHeadersCallback = _m.createOnHeadersCallback
 export const resumeAndAbort = _m.resumeAndAbort
 export const renderToWebFlightStream = _m.renderToWebFlightStream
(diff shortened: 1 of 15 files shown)


packages/next/src/server/app-render/stream-ops.ts
```diff
@@ -44,6 +44,8 @@ export const processPrelude = _m.processPrelude
 export const createWebInlinedDataStream = _m.createWebInlinedDataStream
 export const createNodeInlinedDataStream = _m.createNodeInlinedDataStream
 export const createPendingStream = _m.createPendingStream
+export const createRenderInBrowserAbortSignal =
+  _m.createRenderInBrowserAbortSignal
```
lubieowoce: why does this need to live in stream-ops now? what's dependent on node/edge here?? both reexport the exact same code, so it's kinda pointless to put it here, unless i'm missing something

devjiwonchoi: Wasn't aware of that, moved to separate module

lubieowoce: why does it need to move at all? i'm not against relocating code for clarity but like is there a problem with its current location or what
(moving code makes the diff harder to read and breaks blame/history a bit)

devjiwonchoi: files like cookie.ts imports stuff like `trackDynamicDataInDynamicRender` from dynamic-rendering.ts, so cant pull in browser() to this module

gnoff (approved):

PR #96850, opened 2026-08-06
fix(client): bound the offline retry loops in fetchServerResponse and fetchServerAction

marcoshernanz:
## What

`checkOfflineError()` (client/components/offline.ts, behind `experimental.useOffline`) classifies by exclusion: every thrown value except `AbortError`/`TimeoutError` is treated as a network failure. But in `fetch-server-response.ts` the `try` block spans the **whole response pipeline** — `createFetch`, `new URL(res.url)`, Flight decoding, `resolveStaticStageData` — so any deterministic, non-network failure (a Flight decode error, a truncated RSC stream, a framework `TypeError`) is misclassified as offline, and the handler then:

```
checkOfflineError(err) → notifyOffline() → checkConnectivity() → HEAD succeeds
(the server was always reachable) → waitForConnection() resolves immediately
→ return fetchServerResponse(url, options)   // same error → repeat
```

— an **uncapped, no-backoff retry loop** of full RSC requests from every affected client (and of POST requests via `server-action-reducer.ts`): self-DoS in the browser, request amplification against the origin/CDN, the navigation never completes, and the existing `console.error` + MPA-navigation fallback is never reached.

Verified empirically: with fetch rejecting deterministically and the offline module mocked to report offline-then-online, the navigation retry **loops forever** (jest 60s timeout) before this fix.

## Fix

Both `fetchServerResponse` and `fetchServerAction` get a retry budget (`MAX_OFFLINE_RETRIES = 2`, threaded as an internal trailing parameter, default 0). Once exhausted:

- navigation fetches fall through to the pre-existing `console.error` + MPA-navigation fallback,
- action fetches rethrow.

Genuine offline recovery is unaffected: the retry only runs after `waitForConnection()` resolves, and a healthy retry succeeds on the first replay.

Also fixed along the way: the retry now starts from `originalUrl` — in `output: 'export'` mode the request URL was rewritten to the `.txt` asset before the fetch, and retrying from the rewritten URL would append `.txt` again on every attempt (`/foo.txt.txt`).

## Scope note

The segment-cache prefetch path (`cache.ts`, `rejectRouteCacheEntry(entry, -1)`) has an analogous ping-driven re-fetch loop, and the deeper fix — scoping offline classification to the network call itself rather than the whole pipeline — is left as follow-up. This PR is the bounded safety net that makes a deterministic failure terminate.

## Tests

New `fetch-server-response-offline.test.ts` (mocks `../segment-cache/fetch` with a deterministic rejection, `../offline` to report offline-then-online, and `react-server-dom-webpack/client` to emulate the Flight client's rejection handling):

- budget exhaustion → exactly 3 fetch calls (1 + 2 retries), then the MPA fallback — loops forever before the fix;
- control: with the flag disabled, exactly 1 call.

All 43 client unit tests pass; `tsc`/`eslint` green.

---

### DIFF
--- a/packages/next/src/client/components/router-reducer/fetch-server-response-offline.test.ts
+++ b/packages/next/src/client/components/router-reducer/fetch-server-response-offline.test.ts
@@ -1,0 +1,72 @@
+import { fetchServerResponse } from './fetch-server-response'
+import { fetch as segmentCacheFetch } from '../segment-cache/fetch'
+
+jest.mock('../segment-cache/fetch', () => ({
+  fetch: jest.fn(),
+}))
+
+// Mock the offline module so the test controls connectivity state without
+// touching the network or timers.
+jest.mock('../offline', () => ({
+  checkOfflineError: jest.fn(() => true),
+  getOffline: jest.fn(() => ({})),
+  waitForConnection: jest.fn(() => Promise.resolve()),
+}))
+
+// The real react-server-dom-webpack client needs Next's bundler-config aliasing
+// that only exists in built apps. The Flight client reports fetch rejections
+// through the response (not as unhandled rejections); emulate that here.
+jest.mock('react-server-dom-webpack/client', () => ({
+  createFromFetch: (promise: Promise<any>) => {
+    promise.catch(() => {})
+    return promise
+  },
+  createFromReadableStream: jest.fn(),
+}))
+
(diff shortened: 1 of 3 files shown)


acdlite:
What is the point/motivation of this PR? The whole point of the feature is that it detects automatically when the network is available again. Stopping the polling loop defeats the point.

PR #96856, opened 2026-08-06
fix(server): propagate errors/cancellation through the Node HTML transform pipeline

marcoshernanz:
## What

`continueFizzStream` and `continueDynamicHTMLResumeNode` (server/app-render/stream-ops.node.ts — the Node HTML pipeline for all non-edge builds) built their transform chains with `readable.pipe()`. Per Node's own docs, `pipe()` does **not** forward errors and does **not** destroy upstream stages on downstream teardown:

1. **Upstream error ⇒ hung response.** An erroring stage (flight-data injection when the inlined RSC stream fails, head/metadata insertion when `getServerInsertedHTML()`/`getServerInsertedMetadata()` rejects, a Fizz PassThrough error, …) destroyed only itself. The stream returned to `pipeNodeReadableToNodeResponse` then emitted neither `end` nor `error`, so the request promise and the socket were held open indefinitely — a socket/memory exhaustion vector on repeated render-time errors.

2. **Downstream destroy ⇒ upstream leak.** On client disconnect, `res.on('close')` destroyed only the last transform; the entire upstream chain (buffered transform, Fizz PassThrough, React render, RSC stream) kept running and buffering instead of being cancelled — unlike the web implementation's `pipeThrough` chain, which cancels back to the Fizz stream.

## Fix

- **Build the chains with `stream.pipeline()`** (stage order preserved exactly). `pipeline()` propagates errors downstream — each stage is destroyed with the error and the returned last stream emits `'error'`, which the existing `pipe-readable` handler turns into `res.destroy(err)` — and destroying any stage (e.g. the last one, on client disconnect) tears down the whole chain back to the render. Verified with direct Node experiments.
- **`pipeNodeReadableToNodeResponse`: handle streams that errored before piping.** `pipeline()` starts draining as soon as the chain is built, and Node does not replay an already-emitted `'error'` event (verified empirically). If that race window hits, the pipe would wait forever — now it checks `readable.errored` on entry and destroys the response with that error instead. This is a general fix for piping any pre-errored stream.
- **`createFlightDataInjectionTransform`: cancel the side stream on teardown.** The transform's async iterator over the inlined RSC stream kept pulling after the transform was destroyed (leaking the read loop); a `destroy(err, cb)` hook now destroys the data stream first.

## Tests

- New `pipe-readable.test.ts`: piping an already-errored stream **hangs forever before this fix** (jest timeout) and resolves with `res.destroy(error)` after; a healthy stream pipes normally.
- Live e2e with a real app: normal Suspense streaming and the React-recovered error path unchanged.
- 618 server unit tests pass (3 pre-existing environment failures in unrelated suites, identical without the change); `tsc`/`eslint` clean.

---

### DIFF
--- a/packages/next/src/server/app-render/stream-ops.node.ts
+++ b/packages/next/src/server/app-render/stream-ops.node.ts
@@ -14,7 +14,7 @@ import {
   resumeToPipeableStream,
 } from 'react-dom/server'
 import { prerender } from 'react-dom/static'
-import { PassThrough, Readable, Transform } from 'node:stream'
+import { PassThrough, Readable, Transform, pipeline } from 'node:stream'
 import { isUtf8 } from 'node:buffer'
 
 import {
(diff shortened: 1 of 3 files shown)


lobacz-limango:
Hi, is there any chance you'll be working on this soon? I'm hitting this bug in my app and it's blocking my upgrade from 16.2.x to 16.3.x.

PR #96857, opened 2026-08-06
turbo-tasks: explicit GC root anchoring + cross-session orphan reclamation

lukesandberg:
### What?

Reworks how the reference-counting GC decides what is a **root** (a task that must not be collected), and adds a durable path to reclaim cross-session orphaned roots

Currently GC roots are persisted and the only way to get collected is for another session to revive them, and then for their ref-counts to drop.  This provides a new mechanism.

1. **In-session** — Top level operations get reference counted so they can be dropped in session.
2. **Cross-session** — live roots are recorded along with a simple aging mechanism so they can still be collected even if a later session never restores it.

### Why?
This closes the final GC gap.  Top level operations are ambiguous currently and need tracking.  With cross session aging we close a basic gap where if you delete a route between sessions we never clean up the tasks.  Now it will eventually age out.

### How?

**Roots are anchored explicitly**.  Via pins and RAII guards like `GcRoot`

**Persisted roots set + TTL age-out** A new Infra-keyspace `GcRoots` entry persists `Vec<(TaskId, ttl)>`.  That allows us to detect when roots are stale and can be dropped.

**Dispose transient root tasks on drop.** `RootTask::Drop` now disposes tasks and cleans up references allowing the GC to work

### DIFF
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -47,20 +47,18 @@ use crate::next_api::turbopack_ctx::NextTurbopackContext;
 /// handle exists.
 pub struct DetachedVc<T> {
     turbopack_ctx: NextTurbopackContext,
-    /// The Vc. Must be unresolved, otherwise you are referencing an inactive operation.
-    vc: OperationVc<T>,
-    /// Holds a pin on this operation to prevent GC
-    _gc_root: GcRoot,
+    /// Pins the operation to prevent GC, and holds the `Vc` itself. Must be unresolved, otherwise
+    /// you are referencing an inactive operation.
+    gc_root: GcRoot<T>,
 }
 
 impl<T> DetachedVc<T> {
     pub fn new(turbopack_ctx: NextTurbopackContext, vc: OperationVc<T>) -> Self {
         // Pin the operation's task so GC treats this out-of-graph handle as a root.
-
+        let gc_root = GcRoot::pin(turbopack_ctx.turbo_tasks().clone(), vc);
         Self {
             turbopack_ctx,
-            vc,
-            _gc_root: GcRoot::pin(turbopack_ctx.turbo_tasks().clone(), vc.task_id()),
+            gc_root,
         }
     }
 
@@ -73,7 +71,7 @@ impl<T> Deref for DetachedVc<T> {
     type Target = OperationVc<T>;
 
     fn deref(&self) -> &Self::Target {
-        &self.vc
+        &self.gc_root
     }
 }
 
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -56,12 +56,11 @@ pub struct DetachedVc<T> {
 impl<T> DetachedVc<T> {
     pub fn new(turbopack_ctx: NextTurbopackContext, vc: OperationVc<T>) -> Self {
         // Pin the operation's task so GC treats this out-of-graph handle as a root.
-        let gc_root = GcRoot::pin(turbopack_ctx.turbo_tasks().clone(), vc.task_id());
 
         Self {
             turbopack_ctx,
             vc,
-            _gc_root: gc_root,
+            _gc_root: GcRoot::pin(turbopack_ctx.turbo_tasks().clone(), vc.task_id()),
         }
     }
 
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -85,30 +85,32 @@ impl<T> Deref for DetachedVc<T> {
 /// task promptly. If it doesn't, [`Drop`] disposes it as a backstop.
 ///
 /// This is used by [`subscribe`] to create a computation that re-executes when dependencies change.
-pub struct RootTask {
+pub struct SubscriptionTask {
     turbopack_ctx: NextTurbopackContext,
     task_id: Option<TaskId>,
 }
 
-impl Drop for RootTask {
-    fn drop(&mut self) {
-        // Tear it down now if `root_task_dispose` wasn't called.
+impl SubscriptionTask {
+    fn dispose(&mut self) {
         if let Some(task) = self.task_id.take() {
             self.turbopack_ctx.turbo_tasks().dispose_root_task(task);
         }
     }
 }
 
+impl Drop for SubscriptionTask {
+    fn drop(&mut self) {
+        self.dispose();
+    }
+}
+
 #[napi]
 pub fn root_task_dispose(
-    #[napi(ts_arg_type = "{ __napiType: \"RootTask\" }")] mut root_task: ExternalRef<RootTask>,
+    #[napi(ts_arg_type = "{ __napiType: \"RootTask\" }")] mut root_task: ExternalRef<
+        SubscriptionTask,
+    >,
 ) -> napi::Result<()> {
-    if let Some(task) = root_task.task_id.take() {
-        root_task
-            .turbopack_ctx
-            .turbo_tasks()
-            .dispose_root_task(task);
-    }
+    root_task.dispose();
     Ok(())
 }
 
@@ -444,7 +446,7 @@ pub fn subscribe<
     func: &FunctionRef<V, ()>,
     handler: impl 'static + Sync + Send + Clone + Fn() -> F,
     mapper: impl 'static + Sync + Send + FnMut(ThreadsafeCallContext<T>) -> napi::Result<V>,
-) -> napi::Result<External<RootTask>> {
+) -> napi::Result<External<SubscriptionTask>> {
     let js_func = func.borrow_back(env)?;
     let func: ThreadsafeFunction<T, (), V, Status, true> = js_func
         .build_threadsafe_function::<T>()
@@ -472,7 +474,7 @@ pub fn subscribe<
             }
         }
     });
-    Ok(External::new(RootTask {
+    Ok(External::new(SubscriptionTask {
         turbopack_ctx: ctx,
         task_id: Some(task_id),
     }))
--- a/crates/next-napi-bindings/src/next_api/utils.rs
+++ b/crates/next-napi-bindings/src/next_api/utils.rs
@@ -21,7 +21,7 @@ use regex::Regex;
 use rustc_hash::FxHashMap;
 use serde::Serialize;
 use turbo_rcstr::RcStr;
-use turbo_tasks::{Effects, OperationVc, ReadRef, TaskId, Vc, VcValueType, take_effects};
+use turbo_tasks::{Effects, GcRoot, OperationVc, ReadRef, TaskId, Vc, VcValueType, take_effects};
 use turbo_tasks_fs::FileContent;
 use turbopack_core::{
     issue::{
@@ -43,17 +43,26 @@ use crate::next_api::turbopack_ctx::NextTurbopackContext;
 /// [`turbo_tasks::OperationValue`] and should be dereferenced to an [`OperationVc`] before being
 /// passed to a [`turbo_tasks::function`].
 //
-// TODO: If we add a tracing garbage collector to turbo-tasks, this should be tracked as a GC root.
-#[derive(Clone)]
+/// A `DetachedVc` holds its operation's task alive against garbage collection for as long as the
+/// handle exists.
 pub struct DetachedVc<T> {
     turbopack_ctx: NextTurbopackContext,
     /// The Vc. Must be unresolved, otherwise you are referencing an inactive operation.
     vc: OperationVc<T>,
+    /// Holds a pin on this operation to prevent GC
+    _gc_root: GcRoot,
 }
 
 impl<T> DetachedVc<T> {
     pub fn new(turbopack_ctx: NextTurbopackContext, vc: OperationVc<T>) -> Self {
-        Self { turbopack_ctx, vc }
+        // Pin the operation's task so GC treats this out-of-graph handle as a root.
+        let gc_root = GcRoot::pin(turbopack_ctx.turbo_tasks().clone(), vc.task_id());
+
+        Self {
+            turbopack_ctx,
+            vc,
+            _gc_root: gc_root,
+        }
     }
 
     pub fn turbopack_ctx(&self) -> &NextTurbopackContext {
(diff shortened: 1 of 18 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/operation/cleanup_old_edges.rs
```diff
@@ -54,6 +54,34 @@ pub enum OutdatedEdge {
     CollectiblesDependency(CollectiblesRef),
 }
 
+/// Captures *all* of a task's outgoing edges — children and every forward-dependency kind — as
+/// [`OutdatedEdge`]s to hand to [`CleanupOldEdgesOperation::run`]. Used when a task is being fully
+/// torn down (GC collection, disposing a transient root task): removing these edges drops each
+/// child's parent reference count, scrubs the forward-dependency reverse edges on the targets, and
+/// rebalances the aggregation graph. This is the "remove every edge" case; a re-executing task
+/// instead captures only the *outdated* edges (those in the old child/dep set but not the new one).
+pub fn capture_all_outgoing_edges(task: &impl TaskStorageAccessors) -> Vec<OutdatedEdge> {
```
lukesandberg, line 58: move down to another pr

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -6,97 +6,345 @@
...
+    ///
+    /// Appends into whichever side is already longer so the fold does not repeatedly copy a large
+    /// accumulator into a small one.
+    fn merge(mut self, mut other: Self) -> Self {
         self.collected += other.collected;
         self.edges_deleted += other.edges_deleted;
+        self.gc_roots += other.gc_roots;
+        if self.scanned_roots.len() < other.scanned_roots.len() {
+            std::mem::swap(&mut self.scanned_roots, &mut other.scanned_roots);
+        }
```
lukesandberg: can we have a helper that does this instead of repeating it?

turbopack/crates/turbo-tasks-macros/src/derive/task_storage_macro.rs
lukesandberg, line 1: i think we can revert these changes we don't need the extra bit anymore

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -131,7 +190,95 @@ impl TurboTasksBackend {
...
+            return Duration::from_millis(override_ms);
+        }
+        match std::env::var("TURBO_ENGINE_GC_ROOT_TTL_MS") {
+            Ok(v) => match v.parse::<u64>() {
+                Ok(ms) => Duration::from_millis(ms),
+                Err(_) => GC_ROOT_TTL,
+            },
+            Err(_) => GC_ROOT_TTL,
+        }
+    }
```
lukesandberg: all this logic should be decided during backend construction, not via an atomic read

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -229,6 +242,25 @@ pub struct TurboTasksBackend {
...
+    gc_root_ttl_override_ms: AtomicU64,
+
+    /// `true` until the first GC pass of this session runs. Only that pass may demote a live root
+    /// to `TtlCounter::FirstStale` — GC runs many times per session (on the snapshot cadence), and
+    /// a single pass can easily miss a root that is still live (evicted, or not yet re-requested).
+    /// Demoting on every pass would make the TTL measure idleness within a session rather than
+    /// "was not live in a whole session". See `gc_roots_refresh_and_age_out`.
+    first_gc_pass_of_session: AtomicBool,
+
+    /// Test-only: number of GC passes that produced a roots map to write (i.e. did *not* skip the
```
lukesandberg: i don't think test coverage for this is critical as a small perf optimization

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -550,17 +550,57 @@ impl Storage {
...
+    /// tasks that are pinned for some reason — the pinned `ProjectContainer` op, a live endpoint,
+    /// an active task, or one still holding aggregation edges. GC uses them to maintain the
+    /// persisted roots map (mark still-pinned roots live / age out orphans).
+    ///
+    /// Separate from [`Self::gc_scan_shard_candidates`] because the two run in **different phases**
+    /// of a pass: candidates seed the collect pool one shard at a time, whereas roots are
+    /// classified in one shot *after* that pool drains — root-ness reads `upper`/`followers`,
+    /// which a running cascade is still moving around.
+    ///
+    /// Sees only resident tasks; disk-only roots ride the carried-forward roots map instead.
```
lukesandberg: this comment should be one line

`Return the set of all known live roots`

turbopack/crates/turbo-tasks-backend/tests/gc_collection.rs
```diff
@@ -198,3 +408,52 @@ async fn unpin_after_stop_does_not_panic() {
...
+///
+/// This also guards root classification against depending on anything ephemeral. Root-ness reads
+/// `upper`/`followers` and the `in_progress`/`activeness` flags — state a running cascade moves
+/// around — so if it were ever read mid-cascade the map would differ run to run with thread
+/// scheduling and this would fail.
+///
+/// Asserted via `gc_roots_writes_for_testing` rather than by comparing the persisted set, since an
+/// unchanged set looks identical whether or not it was rewritten.
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn steady_state_gc_passes_skip_the_roots_write() {
```
lukesandberg: drop this test

turbopack/crates/turbo-tasks-backend/tests/gc_collection.rs
```diff
@@ -177,6 +198,195 @@ async fn gc_does_not_collect_pinned_task() {
...
+    // worker teardown would be (the whole task map is dropped by `stop`).
+    tt.stop_and_wait().await;
+    tt.dispose_root_task(root_id);
+}
+
+/// Unpinning a task after the backend has started stopping must not panic. This mirrors the real
+/// shutdown ordering: a `DetachedVc` handed to JS across NAPI is dropped (which unpins) during Node
+/// worker teardown, which can run *after* `stop()` has dropped the whole task map. pin/unpin are
+/// gated on the `stopping` flag so a late unpin is a no-op rather than resurrecting a blank entry
+/// and underflowing `transient_ref_count`.
```
lukesandberg: drop comment, not need to comment things added by downstack prs

turbopack/crates/turbo-tasks-backend/tests/gc_collection.rs
```diff
@@ -177,6 +198,195 @@ async fn gc_does_not_collect_pinned_task() {
     tt.stop_and_wait().await;
 }
 
+/// A [`GcRoot`] guard pins a task for its lifetime and unpins it on drop. This is the anchor a
```
lukesandberg: do we need all 3 of these pinning tests?

RootTask::dispose and GcRoot are very similar mechanisms that don't require their own tests

turbopack/crates/turbo-tasks-backend/tests/gc_cross_session.rs
```diff
@@ -0,0 +1,264 @@
...
+async fn root_with_child(n: u32) -> Result<Vc<u32>> {
+    Ok(Vc::cell(*orphan_leaf(n).await? + 1))
+}
+
+/// Width of the diamond built for the disk-only forward-dep scrub test (see `gc_fixture`): the
+/// fanout gives many chances for the racing interleaving where an `A` scrubs a not-yet-restored
+/// `B`.
+const DIAMOND_FANOUT: u32 = 64;
+
+/// **A root is kept alive by being used, and reclaimed by not being used** — across process
```
lukesandberg: the two tests in this file are very similar and could be merged

turbopack/crates/turbo-tasks-backend/tests/gc_fixture.rs
```diff
@@ -0,0 +1,83 @@
+//! Shared graph shapes for the GC tests.
```
lukesandberg, line 1: simplify comments in this file

crates/next-napi-bindings/src/next_api/utils.rs
```diff
@@ -43,17 +43,26 @@ use crate::next_api::turbopack_ctx::NextTurbopackContext;
...
-// TODO: If we add a tracing garbage collector to turbo-tasks, this should be tracked as a GC root.
-#[derive(Clone)]
+/// A `DetachedVc` holds its operation's task alive against garbage collection for as long as the
+/// handle exists.
 pub struct DetachedVc<T> {
     turbopack_ctx: NextTurbopackContext,
     /// The Vc. Must be unresolved, otherwise you are referencing an inactive operation.
     vc: OperationVc<T>,
+    /// Holds a pin on this operation to prevent GC
+    _gc_root: GcRoot,
```
bgw: Would it be better for `GcRoot` to wrap `OperationVc` instead of `TaskId`?

crates/next-napi-bindings/src/next_api/utils.rs
```diff
@@ -73,20 +82,21 @@ impl<T> Deref for DetachedVc<T> {
...
 ///
-/// JavaScript code receiving this value **must** call [`root_task_dispose`] in a `try...finally`
-/// block to avoid leaking root tasks.
+/// JavaScript code should call [`root_task_dispose`] in a `try...finally` block to dispose the root
+/// task promptly. If it doesn't, [`Drop`] disposes it as a backstop.
 ///
 /// This is used by [`subscribe`] to create a computation that re-executes when dependencies change.
-//
-// TODO: If we add a tracing garbage collector to turbo-tasks, this should be tracked as a GC root.
 pub struct RootTask {
```
bgw: We could give this a less generic name, like `SubscriptionTask`, maybe that's less confusing with `GcRoot`.

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -132,7 +208,80 @@ impl TurboTasksBackend {
...
+        (stats, roots_to_persist)
+    }
+
+    /// Wall-clock now as millis since the Unix epoch.
+    fn now_ms() -> u64 {
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.as_millis() as u64)
+            .unwrap_or(0)
+    }
```
bgw: This has a single caller, inline it

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -78,10 +129,30 @@ impl TurboTasksBackend {
...
+    ) -> (GcStats, Option<Vec<(TaskId, TtlCounter)>>) {
+        let now = Self::now_ms();
+
+        let mut roots = self
+            .backing_storage
+            .roots()
+            .unwrap_or_else(|err| {
+                eprintln!("failed to read GC roots, treating as empty: {err:?}");
+                Vec::new()
+            })
```
bgw: Just do an `expect` or `unwrap` since this should only happen in the case of db corruption

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -123,6 +194,11 @@ impl TurboTasksBackend {
                 drop(task); // drop the lock so CleanupOldEdgesOperation can run
                 stats.collected += 1;
                 stats.edges_deleted += old_edges.len();
+                // If we happeend to delete a known root at this point record it so we can reconcile
```
bgw:
```suggestion
                // If we happened to delete a known root at this point record it so we can reconcile
```

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
```diff
@@ -132,7 +208,80 @@ impl TurboTasksBackend {
...
+        for id in self.storage.gc_scan_roots() {
+            roots.insert(id, TtlCounter::MostRecent);
+        }
+
+        stats.gc_roots = roots.len();
+        stats.aged_out_roots = aged_out_count;
+
+        // Only persist the roots map if it actually changed
+        let roots_to_persist: Option<Vec<_>> =
+            (roots != roots_before).then(|| roots.into_iter().collect());
```
bgw, line 230: We could likely track this more efficiently by storing a `modified` bool, but I guess it doesn't matter much if the roots list is sort.

turbopack/crates/turbo-tasks-backend/tests/gc_collection.rs
```diff
@@ -180,6 +180,61 @@ async fn gc_does_not_collect_pinned_task() {
...
+        tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+    }
+    assert_eq!(tt.backend().parent_count_for_testing(leaf_id), 0);
+    assert_eq!(tt.backend().transient_ref_count_for_testing(leaf_id), 1);
+    assert_eq!(tt.backend().gc_for_testing(&tt), 0);
+
+    // Disposing twice must not panic or underflow the child's count: JS may dispose explicitly and
+    // then drop.
+    tt.dispose_root_task(root_id);
+    tt.dispose_root_task(root_id);
```
bgw: The `Drop::drop` impl that's used doesn't call `dispose_root_task` a second time, but I do think it's reasonable for us to make sure we handle underflows okay.

```suggestion
    // Disposing twice must not panic or underflow the child's count: JS could dispose explicitly
    // multiple times
    tt.dispose_root_task(root_id);
    tt.dispose_root_task(root_id);
```

bgw (approved):

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: i write these typos so you know that I am writing the code

turbopack/crates/turbo-tasks-backend/src/backend/gc.rs
lukesandberg: pretty tricky i think

we basically kick out all resident tasks in the first traversal and then add them all back in the final scan

so even in a 'no-op' case we might touch the map a bunch of times.

im hopeful that the map stays small in which case who cares but if roots gets large then we will have a problem

crates/next-napi-bindings/src/next_api/utils.rs
lukesandberg: yes.  i thought there was a blocker but no. this is the correct API shape

PR #96860, opened 2026-08-06
Remove `turbopack/packages` and relocate devlow to `packages/`

wbinnssmith:
## Summary

`@vercel/devlow-bench` hasn't published since December 2024. The job that published it was removed in #93064; it was gated on the head commit message starting with `chore: release turbopack npm packages`, which squash-merge rewrites, so it couldn't fire through normal review — #93951 bumped the version correctly and still published nothing.

Rather than rebuild that machinery for `turbopack/packages/*`, this moves the one remaining package to `packages/`, where lerna and the existing recursive publish already handle everything. **`publish-release.js` is unchanged.**

The tradeoff is that the package joins the Next.js lockstep version, going from `0.4.0` to the current canary, and ships nightly instead of on manual bumps. At ~10k downloads/month, essentially all CI, nobody is reading its version for compatibility — not worth ~300 lines of bespoke release code in the script that gates Next.js releases.

`packages/` already holds unscoped, non-`@next/` packages (`create-next-app`, `eslint-config-next`, `next-rspack`), all lockstep.

The root eslint config covers `packages/` but never applied to `turbopack/`, so this exempts the three rules the source doesn't satisfy and prefixes six unused parameters, rather than reformatting the package as part of a move.

## Deletions

Three nft packages, last published in 2022, none with a build script or any consumer:

- `@vercel/experimental-nft` — a stub containing only a `package.json`
- `@vercel/webpack-nft` — resolved binaries built from a `node-file-trace` crate that no longer exists, so its core codepath couldn't work
- `@vercel/experimental-nft-next-plugin` — depends on the above

`turbopack/xtask/src/publish.rs` goes with them: its only caller was the workflow removed in #93064, its `npm` subcommand targeted that same missing crate, and nothing invokes `cargo xtask`.

`turbopack/packages/` no longer exists.

## Verification

- `publish-release.js --dry-run` — publishes `@vercel/devlow-bench` via the untouched recursive publish
- `lerna list` and `turbo run build` both resolve it at the new path
- `pnpm run --filter=devlow-bench test` — 3/3, and the CI jobs reference it by name, not path
- Root eslint and the package's own `lint` script both clean
- `cargo check`/`clippy`/`fmt -p xtask`
- Not run: a real publish, which only CI can exercise

`pnpm-lock.yaml` is hand-edited to repoint the workspace entry and drop the three removed ones. A full `pnpm install` rewrites ~1800 unrelated lines, and `--frozen-lockfile` fails identically on unmodified `canary` from pre-existing `apps/docs` drift.


eslint.config.mjs
```diff
@@ -601,4 +601,15 @@ export default defineConfig([
...
+    // Moved from turbopack/, where the root config never applied. Exempted from
+    // the rules it doesn't currently satisfy rather than reformatting it as
+    // part of the move.
+    files: ['packages/devlow-bench/**'],
+    rules: {
+      '@typescript-eslint/no-shadow': 'off',
+      '@typescript-eslint/consistent-type-imports': 'off',
+      'import/no-extraneous-dependencies': 'off',
+    },
+  },
```
lukesandberg, line 614: seems worth a fast-follow PR

lukesandberg (approved):

PR #96863, opened 2026-08-06
docs: link View Transitions skill on skills.sh and clarify the example prompt

aurorascharff:
## What

Updates the [Designing view transitions](/docs/app/guides/view-transitions) guide so the `vercel-react-view-transitions` skill points to its **skills.sh** page instead of the GitHub source, and reframes the sample prompt as an example.

## Changes

- Link the skill to `https://skills.sh/vercel-labs/agent-skills/vercel-react-view-transitions` instead of the `github.com/vercel-labs/agent-skills/...` source tree.
- Use the skills.sh install form (`npx skills add vercel-labs/agent-skills --skill ...`), consistent with the other skills referenced in the docs.
- Frame the prompt as "a prompt like this" and add a sentence making clear the example is not required, readers can prompt for whatever they want to achieve and the skill guides the agent through the relevant patterns.

## Why

The skills.sh page is the canonical place to view and install the skill. The old wording read as if the one prompt was the only way to invoke the skill.


docs/01-app/02-guides/view-transitions.mdx
```diff
@@ -14,20 +14,22 @@ This guide walks through four patterns that cover the most common cases: morphin
...
 
 Install the skill:
 
 ```bash filename="Terminal"
-npx skills add https://github.com/vercel-labs/agent-skills --skill vercel-react-view-transitions
+npx skills add vercel-labs/agent-skills --skill vercel-react-view-transitions
 ```
 
-Then give the agent this prompt:
+Then give the agent a prompt like this:
```
aurorascharff, line 26:
```suggestion
Then prompt the agent with something like:
```

icyJoseph (approved):

PR #96867, opened 2026-08-06
perf: overlap App type checking with Turbopack

marcoshernanz:
## What

Run the existing TypeScript checker concurrently with the default Turbopack
production build worker for App Router and hybrid projects.

The build now:

- completes TypeScript dependency/config/`next-env.d.ts` setup before either
  worker starts;
- starts the existing checker immediately before `turbopackBuild`;
- waits for both checking and Turbopack shutdown before static generation;
- preserves compiler-first diagnostics and the immediate
  `Compiled successfully` event;
- distinguishes user type errors from worker/infrastructure failures, including
  a checker-specific diagnostic for an unexpected child exit.

Pure Pages Router, Webpack/Rspack, compile/generate modes,
`ignoreBuildErrors`, `runAfterProductionCompile`, JavaScript-only projects,
and the private in-process Turbopack path keep their previous behavior.

## Why

App builds currently serialize compilation and TypeScript checking even though
Turbopack route declarations (`routes.d.ts`, `validator.ts`, cache-life and
root-parameter types) are generated before compilation. The preflight removes
the remaining config-file write race, allowing the two CPU-heavy phases to
overlap without changing emitted output.

## Performance

All A/B series used packages built from the same commit/toolchain, a stable
resolved `next` path, alternating AB/BA order, identical cache state, eight
pairs, process-tree RSS sampling, and paired bootstrap intervals.

| Workload | Shape/cache | Baseline | Candidate | Paired speedup (95% CI) |
|---|---|---:|---:|---:|
| Vercel AI Chatbot | 16 vCPU, cold | 70.381 s | 63.522 s | **9.62% (7.42–12.04%)** |
| NextBeats, exact PR package | 16 vCPU, cold | 26.251 s | 24.901 s | **5.12% (3.54–6.45%)** |
| NextBeats | 16 vCPU, fixed warm | 8.729 s | 7.646 s | **12.40% (11.15–13.55%)** |
| NextBeats | 8 vCPU, cold | 34.594 s | 33.269 s | **3.74% (1.08–6.43%)** |

The exact PR tarball SHA-256 is
`481bfe894058fdf95a3d0573296ab1303dedb8f943b26c5d5f60b98edff5477d`.

Four earlier scheduling-identical eight-pair cold batches measured 4.42%,
5.49%, 6.24%, and 8.52%; the exact package is near the conservative end of the
replicated range.

Every corresponding balanced same-package A/A interval crossed zero. The
effect was neutral at 4 vCPU (+0.88%, −0.56–2.25% CI) and 2 vCPU (−0.04%,
−1.44–1.32% CI), so this PR does not claim a benefit on small builders.

Fixed-ID NextBeats A/A and A/B builds emitted the same 462-file inventory and
the same seven nondeterministic paths; the other 455 files were byte-identical.

## Measured tradeoffs

- Exact cold NextBeats RSS moved −365,757 KiB with a
  −1,016,663–+213,909-KiB interval that crosses zero. Earlier
  scheduling-identical batches ranged from a significant −688,299 KiB to a
  significant +757,962 KiB (+740 MiB); pooled across two earlier batches the
  increase was +554 MiB / 14.0%. AI Chatbot cold, NextBeats warm, and the
  8-vCPU series crossed zero. A process-level follow-up found every global peak
  after `tsc` exited, in static-generation workers. Memory is therefore treated
  as workload- and batch-dependent, not as a reliable benefit.
- App builds with a TypeScript error were 3.82% faster end to end because the
  diagnostic is discovered during compilation, while still waiting for
  compiler shutdown.
- App builds with only a compiler error were 0.712 s / 3.29% slower
  (0.76–5.51% regression CI): baseline never reaches its sequential checker,
  while the candidate has already spent checker CPU.
- A prototype that also covered pure Pages builds made failed builds 17.57%
  slower and added 4.16 GiB peak RSS, so pure Pages is deliberately excluded.

## Validation

- `pnpm build-all` — 19/19 tasks, including native SWC and declarations
- Turbopack TypeScript build-output suite — 6/6
- TypeScript version/worker-failure suite — 5/5, including `SIGKILL`
- frozen compatibility batch — 28/28
- wider Pages/typegen suites — 9/9
- Webpack and Rspack closeout — 2/2 each
- hybrid App/Pages production navigation — 4/4
- fixed-ID output equivalence and `next start` route smoke


packages/next/src/build/index.ts
```diff
@@ -1774,6 +1794,25 @@ export default async function build(
...
+        if (shouldRunTypeChecking) {
+          typeCheckingPromise = startTypeChecking({
+            ...typeCheckingOptions,
+            typeCheckPreflightDone: true,
+            deferExitOnError: true,
+            showTypeCheckingSpinner: false,
+          })
+          // Compilation can fail before this promise is awaited below.
+          // Handle rejections immediately to avoid an unhandled rejection.
+          void typeCheckingPromise.catch(() => {})
```
lukesandberg, line 1812: drop void keyword

packages/next/src/build/index.ts
```diff
@@ -1687,8 +1691,24 @@ export default async function build(
...
+      // Keep pure Pages builds sequential: they historically type check before
+      // compilation, so overlapping would delay failures and increase their
+      // peak memory. App builds historically run the post-compile hook before
+      // type checking, so it may generate or modify files the checker observes.
+      const shouldOverlapTurbopackTypeChecking =
+        bundler === Bundler.Turbopack &&
+        !!appDir &&
+        useTurbopackBuildWorker &&
+        !config.typescript.ignoreBuildErrors &&
+        !config.compiler?.runAfterProductionCompile &&
```
lukesandberg, line 1706: we definitely document that `runAfterProductionCompile` runs before type checking, but i read that more as an example for when it runs than a guarantee, so maybe we should just skip the docs

this carve out just means certain common plugins disable the optimization (sentry, otel) overal defeating the goal of the optimization

so, imho, we either abandon the approach, or we change the docs on `runAfterProductionCompile`

packages/next/src/build/index.ts
```diff
@@ -1687,8 +1691,24 @@ export default async function build(
...
+        process.env.NEXT_TURBOPACK_USE_WORKER !== '0'
+      // Keep pure Pages builds sequential: they historically type check before
+      // compilation, so overlapping would delay failures and increase their
+      // peak memory. App builds historically run the post-compile hook before
+      // type checking, so it may generate or modify files the checker observes.
+      const shouldOverlapTurbopackTypeChecking =
+        bundler === Bundler.Turbopack &&
+        !!appDir &&
+        useTurbopackBuildWorker &&
+        !config.typescript.ignoreBuildErrors &&
```
lukesandberg, line 1705: why opt out for `ignoreBuildErrors`? i guess this option does confuse me generally

packages/next/src/build/index.ts
lukesandberg: @timneutkens do you think the `runAfterProductionCompile` constraint makes sense?

PR #96868, opened 2026-08-06
Remove legacy PPR code paths

ztanner:
## Summary

`experimental.ppr` is deprecated, and Cache Components is now the only internal path that enables partial prerendering. This makes the legacy PPR renderer and its `prerender-ppr` work unit unreachable.

This removes that rendering path, the associated work-unit branches, Next.js’s remaining integration with React’s deprecated `unstable_postpone` API, the dead `__NEXT_PPR` client flag, and obsolete legacy PPR tests.

Cache Components’ postponed-state and resume behavior remain unchanged.


laugharn:
Pour one out

packages/next/src/server/node-environment-extensions/process-error-handlers.ts
gnoff: can delete reference to React postpones

gnoff (approved):

PR #96871, opened 2026-08-06
Lint devlow-bench with the root eslint config

wbinnssmith:
Follow-up to #96860, addressing @lukesandberg's review.

That PR moved `devlow-bench` into `packages/`, where the root eslint config applies for the first time, and exempted the package from the three rules it didn't satisfy rather than fixing the violations.

This drops the exemption block:

- **`import/no-extraneous-dependencies`** — had no violations. The exemption was unnecessary.
- **`@typescript-eslint/consistent-type-imports`** — 13 violations, all autofixable.
- **`@typescript-eslint/no-shadow`** — 17 violations, renamed by hand.

The shadow fixes are mostly loop variables and callback parameters that restated their outer binding. Two worth a look:

- `table.ts` — `getValue(data, ...)` took a single entry but shadowed the `data` array it indexes into. Renamed the parameter to `entry`, and the `row`/`column` grouping keys to `rowKey`/`columnKey` so the loop variables can keep the shorter names.
- `shell.ts` — `command(command, args)` shadowed the exported function with its own first parameter. Renamed to `executable`, which also fixes the `[SHELL]` debug log that was interpolating the function rather than the argument.

## Verification

- `eslint` clean under both the root config and the package's own `lint` script
- `tsc` compiles, `pnpm run --filter=devlow-bench test` 3/3
- prettier clean


packages/devlow-bench/src/interfaces/compare.ts
```diff
@@ -1,5 +1,6 @@
-import { Interface } from '../index.js'
-import { SampleGroup, groupRows, makeKey, printComparison } from '../compare.js'
+import type { Interface } from '../index.js'
+import type { SampleGroup } from '../compare.js'
+import { groupRows, makeKey, printComparison } from '../compare.js'
 import { readSnapshot } from '../snapshot.js'
 import { formatVariantProps } from '../utils.js'
```
lukesandberg, line 10:
```suggestion
import type { Interface } from '../index.js'
import { type SampleGroup, groupRows, makeKey, printComparison } from '../compare.js'
import { readSnapshot } from '../snapshot.js'
import { formatVariantProps } from '../utils.js'

```

pretty sure that works

packages/devlow-bench/src/interfaces/snapshot.ts
```diff
@@ -1,8 +1,9 @@
 import { execFile } from 'node:child_process'
 import { promisify } from 'node:util'
 import picocolors from 'picocolors'
-import { Interface } from '../index.js'
-import { SnapshotRow, defaultSnapshotPath, writeSnapshot } from '../snapshot.js'
+import type { Interface } from '../index.js'
+import type { SnapshotRow } from '../snapshot.js'
+import { defaultSnapshotPath, writeSnapshot } from '../snapshot.js'
 import { formatVariantProps } from '../utils.js'
```
lukesandberg, line 10:
```suggestion
import type { Interface } from '../index.js'
import { type SnapshotRow, defaultSnapshotPath, writeSnapshot } from '../snapshot.js'
import { formatVariantProps } from '../utils.js'
```

packages/devlow-bench/src/cli.ts
```diff
@@ -1,7 +1,8 @@
 import minimist from 'minimist'
 import { setCurrentScenarios } from './describe.js'
 import { join } from 'path'
-import { Scenario, ScenarioVariant, runScenarios } from './index.js'
+import type { Scenario, ScenarioVariant } from './index.js'
+import { runScenarios } from './index.js'
 import compose from './interfaces/compose.js'
 import { groupRows, printComparison } from './compare.js'
```
lukesandberg:
```suggestion
import { type Scenario, type ScenarioVariant , runScenarios } from './index.js'
import compose from './interfaces/compose.js'
import { groupRows, printComparison } from './compare.js'
```

lukesandberg (approved):

PR #96873, opened 2026-08-07
Scope app entry export validation to the app directory

### LINKED ISSUE #96859: Turbopack build fails on pages-router files named `sitemap`/`robots`: `"getStaticProps" is not supported in app/` (no app directory)

https://github.com/rodrigo-arias/next-16-3-pages-sitemap-repro

### To Reproduce

1. `npm install` (next 16.3.0, react 19.2.8 — pages router only, **no `app/` directory**)
2. `npx next build`

**Current:** The Turbopack build fails compiling `pages/sitemap.js`, a plain pages-router page that exports `getStaticProps`:

```
Error: Turbopack build failed with 1 error:
./pages/sitemap.js:5:23
Error: "getStaticProps" is not supported in app/. Read more: https://nextjs.org/docs/app/building-your-application/data-fetching
```

The same happens for a page named `pages/robots.js`. The failure is purely filename-based: renaming the file (e.g. `sitemap-page.js`) with identical contents builds fine, so the app-router **metadata-route filename convention** (`sitemap`, `robots`, …) appears to be applied to root-level `pages/` files, which are then compiled under app-router rules where `getStaticProps` is rejected. `pages/api/robots.js` (API route) is unaffected.

**Expected:** Root-level `pages/` files named `sitemap`/`robots` are ordinary pages-router routes (`/sitemap`, `/robots`) and build normally — as they do on 16.2.x.

Version matrix (identical project):

| next version | result |
| --- | --- |
| 16.2.12 | ✅ builds; `/sitemap` prerenders as SSG |
| 16.3.0 | ❌ `"getStaticProps" is not supported in app/` |
| 16.3.1-canary.4 | ❌ same error |

Possibly related (same filename-collision class, Turbopack-only): #78609

```
Operating System:
  Platform: darwin
  Arch: arm64
Binaries:
  Node: 24.19.0
  npm: 11.17.0
  pnpm: 10.34.5
Relevant Packages:
  next: 16.3.0
  react: 19.2.8
  react-dom: 19.2.8
  typescript: N/A
Next.js Config:
  output: N/A
```

Turbopack, Pages Router

### Additional context

A human-readable `/sitemap` HTML page is a common pages-router pattern; this blocks upgrading to 16.3.0 for any such site. Reproduced on macOS arm64; first hit upgrading a production pages-router site from 16.2.10.

lazerg:
PR #94962 extended the app entry filename regex in the RSC transform to cover metadata conventions like `sitemap`, `robots` and `icon`, but that regex is matched against every file, not only files under `app/`. A Pages Router project with `pages/sitemap.js` now fails to build with `"getStaticProps" is not supported in app/`, even when there is no app directory at all.

This scopes the check to files inside the app directory, the same way the `error.js` check just above it already does. The existing fixtures that depend on the check moved under `app-dir/` so they keep triggering it.

Fixes #96859


lazerg:
Closing this, I'm not pursuing this fix right now. The branch stays up if anyone wants to pick it up.

PR #96878, opened 2026-08-07
Unify how a response's shell and full payloads are written

acdlite:
A response may contain two payloads: the full response, plus a shell stage that the client obtains by decoding the same bytes again, truncated at the shell byte offset. Each flow had its own copy of the logic that decides which payload fulfills the pending entries and what fetch strategy each write records. Those copies are replaced by a single function, writeResponsePayloadsIntoCache. The embedded runtime prefetch stream, which previously ignored the byte offset entirely, now goes through it too.

When a shell request receives more than it asked for, the extra content is at least as complete as what the corresponding non-shell request would have returned, so that's the strategy the write records: PPR for a StaticShell request, PPRRuntime for a RuntimeShell request. Previously some of these writes under-recorded, causing the scheduler to issue runtime requests for content it already had. A rejected needsRuntimeRequest row now also reads conservatively (as true) on the client, matching how the server reads the same flag.

The second commit fixes an oversight where a static prefetch attempt was mistakenly skipped: when the scheduler found a cached entry that was only a shell, it deopted straight to a runtime request even though the full static response contains more and the segment's hint says a static attempt is worthwhile. A shell entry with the hint now takes the normal static revalidation path first, and the response's own signals decide whether to escalate afterward.

### DIFF
--- a/packages/next/src/client/components/segment-cache/scheduler.ts
+++ b/packages/next/src/client/components/segment-cache/scheduler.ts
@@ -221,8 +221,10 @@ const enum PrefetchTaskExitStatus {
  * before per-link work:
  *
  * - RouteTree: fetch the route's tree structure.
- * - Shell: fetch the route's reusable App Shell (param-free loading state),
- *   if the route can produce one and the feature is enabled. Bounded by
+ * - Shell: fetch each segment's shell-stage variant, keyed to be reusable
+ *   across all params below the root (root params are kept) — the phase's
+ *   target is the conceptual App Shell. Runs if the route can produce a
+ *   shell and the feature is enabled. Bounded by
  *   filesystem-route count, not link count — so all Shell prefetches across
  *   queued tasks complete before any Speculative prefetch runs, because
  *   shell responses are shared across every navigation to the same route.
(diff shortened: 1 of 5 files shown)


packages/next/src/client/components/segment-cache/scheduler.ts
```diff
@@ -2222,10 +2228,40 @@ function pingSegmentBundle(
...
+        // static re-fetch would return the same bytes. So when the segment's
+        // hint says a static attempt is worthwhile (the build-time prerender
+        // accessed no runtime data), treat the entry like a less complete
+        // one: take the static attempt path below (the revalidation) instead
+        // of deopting straight to a runtime request, and let the response's
+        // own verdict decide — as for any static attempt — whether to
+        // escalate afterward. The attempt can't recur: its response records
+        // at least the concrete static tier, which fails the shell-tier
+        // check here on the re-run pass.
+        const shellEntryEligibleForStaticAttempt =
```
ztanner, line 2287: Should the `InlinedIntoChild` early-return in `accumulateSegmentBundle` apply this same exception? It still treats `wouldRuntimeRequestProvideMore` as an immediate runtime deopt, which seems like it could prevent an eligible shell entry from reaching the bundle's static request.

ztanner (approved):

PR #96882, opened 2026-08-07
fix(turbo-tasks-fs): parse glob character-class ranges correctly

marcoshernanz:
## What

Two bugs in `parse_class` (turbopack/crates/turbo-tasks-fs/src/globset.rs — the vendored globset fork used for `outputFileTracingIncludes`/`Excludes` keys, `import.meta.glob` patterns, and NFT rule matching):

1. **Range-end character discarded.** The range-end arm called `add_to_last_range(ranges.last_mut().unwrap(), '-')` with a hardcoded dash instead of the just-consumed character, closing every range at U+002D:
   - `[a-z]`, `[A-Z]`, `[0-9]`, `[a-f]`, … → `'-' < 'a'` → `InvalidRange` → the glob is **rejected** (build-breaking),
   - ranges starting at codepoints ≤ `-` (e.g. `[*-z]`) silently mis-parse as `('*','-')`, matching only `*`, `+`, `-`.
   
   Upstream globset 0.4.16 passes the consumed character (verified against the registry source).

2. **Negated-class `matches_slash` inverted.** The comment says "if all of the ranges exclude `/`, then the negation includes it" — but the code computed `!all(...)`. A negated class matches `/` iff **no** range contains it, i.e. `all(r.0 > '/' || r.1 < '/')`. With negated ranges parseable again after fix 1, the inversion made globs like `a[!b-z]x/**` mis-prune directories containing valid matches in `read_glob` (verified empirically: `can_match_in_directory("a")` was `false` before, `true` after; plain-class shapes are unchanged).

## Tests

- 6 new rstest matrix cases asserting the exact generated regexes: `[a-z]`, `[0-9]`, `[a-z0-9_]`, `[*-z]` (range contains `/` → directory-match regex fully optional), `[a-]` (legit literal trailing dash), plus a parse-success test for `[!a-z]`.
- 2 directory-match tests pinning the post-fix behavior: `a[!b-z]x/**` keeps directory `a`; `a[!-/]x/**` keeps the class required (identical to the plain-class behavior for that shape).
- Verified by reverting just the one line: all 5 matrix range cases fail pre-fix (`InvalidRange`); `[a-]` and all pre-existing cases pass either way. All 122 crate lib tests pass; `clippy`/`fmt` clean.

---

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/globset.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/globset.rs
@@ -585,7 +585,10 @@ impl<'a> Parser<'a> {
                     if in_range {
                         // invariant: in_range is only set when there is
                         // already at least one character seen.
-                        if let Some(kind) = add_to_last_range(ranges.last_mut().unwrap(), '-') {
+                        // Pass the consumed character — hardcoding '-' here
+                        // would close every range at U+002D, rejecting common
+                        // ranges ([a-z]) or silently mis-parsing ([*-z]).
+                        if let Some(kind) = add_to_last_range(ranges.last_mut().unwrap(), c) {
                             return Err(self.error(kind));
                         }
                     } else {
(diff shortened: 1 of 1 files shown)


mischnic:
Fixed by https://github.com/vercel/next.js/pull/97502

PR #96887, opened 2026-08-07
ci: stop publishing @vercel/devlow-bench

mischnic:
`@vercel/devlow-bench` was supposed to start being published again in #96860, but something's going wrong there and it broke the publishing workflow

I think our npm token currently doesn't allow publishing that package


lubieowoce (approved):

eps1lon (approved):
force merging to break the deadlock

PR #96895, opened 2026-08-07
[ci] Default deploy e2e tests to the repo next version

eps1lon:
This is mostly just to remove some confusion and clean up unused default values in some places.

When you run on a backport branch or older tag, you probably never want to use `canary` and just the version at that point. Running on `canary` branch will implicitly use the latest `@canary` from NPM.

Since our setup job already checks for the version declared in the checkout, we should use that everywhere so that we only have to check a single place to be sure about what version is/will be used.

I'll backport this especially because it'll fix a bug in 15.x where we ignore the release tag.

### DIFF
--- a/.github/workflows/test_e2e_deploy_release.yml
+++ b/.github/workflows/test_e2e_deploy_release.yml
@@ -95,6 +95,12 @@ jobs:
             echo 'value=${{ fromJson(steps.nextPackageInfo.outputs.value).version }}' >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Log resolved `next` version
+        env:
+          RESOLVED_NEXT_VERSION: ${{ steps.version.outputs.value }}
+        run: |
+          echo "Resolved next version: $RESOLVED_NEXT_VERSION"
+
       - name: Install dependencies
         run: pnpm install
 
--- a/.github/workflows/test_e2e_deploy_release.yml
+++ b/.github/workflows/test_e2e_deploy_release.yml
@@ -8,8 +8,11 @@ on:
   workflow_dispatch:
     inputs:
       nextVersion:
-        description: canary or custom tarball URL
-        default: canary
+        description: >-
+          Version of `next` to test against. Anything installable by npm:
+          a dist-tag (canary), an exact version (16.2.4), or a custom https
+          tarball URL. Leave empty to use the `next` version from
+          packages/next/package.json.
         type: string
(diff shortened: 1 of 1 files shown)


lubieowoce (approved):

.github/workflows/test_e2e_deploy_release.yml
```diff
@@ -126,7 +145,7 @@ jobs:
         IS_WEBPACK_TEST=1 \
         NEXT_ENABLE_ADAPTER=0 \
         NEXT_EXTERNAL_TESTS_FILTERS="test/deploy-tests-manifest.json" \
-        NEXT_TEST_VERSION="${{ github.event.inputs.nextVersion || needs.setup.outputs.next-version || 'canary' }}" \
+        NEXT_TEST_VERSION="${{ needs.setup.outputs.next-version }}" \
```
ztanner, line 148: I'm wondering if the signed tarball outputs might be silently masked by GitHub here: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idoutputs

> If an output is skipped because it may contain a secret, you will see the following warning message: "Skip output {output.Key} since it may contain secret."

eps1lon: checking.

eps1lon: Works. 

Between steps: https://github.com/vercel/next.js/actions/runs/31196722639/job/92926623131#step:7:15
Between jobs: https://github.com/vercel/next.js/actions/runs/31196722639/job/92927035198#step:34:8

Though I didn't know about this. "may contain a secret" is quite vague. But reasonable to assume a commit might get flagged due high entropy.

eps1lon: > signed tarball outputs

What do you mean with "signed"? They just have the commit in the URL. We're not signing these.

ztanner: I just couldn't remember if tarball URLs were signed or not. SGTM if they are not

ztanner (approved):

.github/workflows/test_e2e_deploy_release.yml
eps1lon: For vercel/next.js they're just public (because the repo is public) and for vercel/next-js-mirror they're on a route requiring authorization.

PR #95739, opened 2026-07-13
[test] Disable i18n-api-support deploy test for Turbopack with adapters

eps1lon:
Follow-up to https://github.com/vercel/next.js/pull/95680 fixing https://github.com/vercel/next.js/actions/runs/29214339280/job/86714009862#step:32:382

PR #96898, opened 2026-08-07
test: Reenable i18n-api-support deploy test for Turbopack with adapters

mischnic:
Reverts #95739
It's fixed in the adapter: https://github.com/nextjs/adapter-vercel/pull/81


mischnic:
There were some reverts here:
https://github.com/vercel/next.js/pull/97327
https://github.com/nextjs/adapter-vercel/pull/101

---

1. i18n-api-support was enabled
2. the next.js PR was merged. this broke the test, because it uncovered a adapter bug
3. the adapter PR fixed the bug. the test works again
4. revert 2 and 3. we are now back to 1. so the test passes again

timneutkens (approved):

PR #96908, opened 2026-08-07
[PPF] unstable_navigation()

lubieowoce:
`navigation()` is a new API that allows omitting contents from runtime shells and runtime prefetches. Conceptually, the point is to express that something is expensive to compute, so we shouldn't do it for requests that may not get used (shells and prefetches). Notably, this means that it's fine to include it in a static prerender -- it'll be computed once and used for many requests, so it doesn't make sense to exclude it.

## Implementation

The split in behavior across static and runtime prerenders is a departure from how most of our APIs behave -- usually, if something resolves statically, then it also resolves in "more complete" prerender. Departing from this leads to some implementation complexity.

We include three new stages, used by two facets of the implementation:

```diff
export enum RenderStage {
  Before = 1,
  //
  ShellStatic = 10,
+ PrefetchStatic = 11, <------- params, prefetch()  [static prerenders]
+ NavigationStatic = 12  <------navigation()        [static prerenders]
  Static = 13, <--------------- finish accumulators [static prerenders]
  //
  ShellRuntime = 20,
  Runtime = 21, <-------------- params, prefetch() [runtime prerenders]
+ NavigationRuntime = 22, <---- navigation()       [runtime prerenders]
  //
  Dynamic = 30,
  Abandoned = 40,
}
```

### NavigationRuntime

In runtime prerenders (or dev renders that simulate them), `navigation()` resolves in `NavigationRuntime`
We only reach this stage in 1. the embedded runtime prerender produced for Cached Navigations and 2. during dev/prod full staged renders -- runtime shells end in `ShellRuntime`, and runtime prefetches end in `Runtime`.

Notably, this means that content gated behind `navigation()` is included in the embedded runtime prefetch stream.

### PrefetchStatic & NavigationStatic

This is a helper stage added before `Static`. Static prefetches still use the `Static` stage for their output. This new stage exists so that we can resolve static `params` (and `prefetch()` when we implement it) which the stage is named after) separately from `navigation()`, which resolves in `NavigationStatic`, after which the prerender ends in `Static`. This separation is important, because during static prerenders we track whether or not runtime APIs are used (see `trackRuntimeDataAccessed`) to determine if a runtime shell (or runtime prefetch) might give us more content than the static ones. However, a runtime shell/prefethc **would not resolve navigation()**, so `await navigation(); await cookies()` would not reveal more content, and thus shouldn't count as a usage that prevents static optimization.
We achieve this by checking the stage inside `trackRuntimeDataAccessedImpl` and not tracking anything if we reached the `NavigationStatic` stage.

### Behavior of shells and validation

As noted before, `navigation()` has an incompatible resolution order between static and runtime prerenders. In #97040, we did some groundwork to deal with this in validation.

Static prerenders resolve `navigation()`, which means that static shells include content gated behind navigation(). This means that Static Shell Validation allows them.

On the other hand, App shells **do not** resolve `navigation()`. This leads to an inconsistency for Instant Validation -- a `await navigation()` might be fine if a page is prefetched statically, but would become blocking as soon as the page starts using runtime data and switches to a runtime shell. To avoid this pitfall, we pessimistically assume that any `navigation()` _might_ be part of a runtime shell/prefetch, so any `navigation()` unguarded by Suspense will error in IV.

In practice, this is handled analogously to static params: we do a dev render with `needsAppShell: true`, which makes `navigation()` resolve in `NavigationRuntime`, and then we use the `ShellRuntime` stage when validating, which means that `navigation()` will be a hole. Note that the discriminated error message logic currently only retries errors using the `Runtime` stage, which won't have `navigation()` resolved either, so it will be incorrectly reported as dynamic data. This will be improved in a follow up.

### DIFF
--- a/packages/next/src/server/request/cache-stages.ts
+++ b/packages/next/src/server/request/cache-stages.ts
@@ -1,3 +1,16 @@
+import { workAsyncStorage } from '../app-render/work-async-storage.external'
+import {
+  throwForMissingRequestStore,
+  workUnitAsyncStorage,
+} from '../app-render/work-unit-async-storage.external'
+import {
+  applyOwnerStack,
+  trackIncompatibleShellContent,
+} from '../dynamic-rendering-utils'
+import { isRequestApiAllowedInCurrentPhase } from './utils'
+import { InvariantError } from '../../shared/lib/invariant-error'
+import { RenderStage } from '../app-render/staged-rendering'
+
 /**
  * This function allows you to indicate that the subsequent code should be
  * deferred to the actual navigation instead of rendering during a runtime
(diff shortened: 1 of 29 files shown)


test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts
```diff
@@ -429,6 +429,226 @@ describe('App Shell prefetching', () => {
...
+          .click()
+      }, [
+        // Two runtime responses carry the shell text, in order: the Shell
+        // phase's runtime App Shell request, then the Speculative phase's
+        // batched per-link runtime prefetch. The route reads request data,
+        // so its static-attempt hint is unset and the prefetch deopts to
+        // runtime requests — the runtime-completeness contract of Partial
+        // Prefetching routes.
+        { includes: 'App shell for navigation', kind: 'runtime' },
+        { includes: 'App shell for navigation', kind: 'runtime' }, // TODO: wait, why are there two requests here?
```
lubieowoce: Why? Because https://github.com/vercel/next.js/pull/97524 hasn't landed yet

packages/next/src/server/request/cache-stages.ts
```diff
@@ -13,5 +28,125 @@
...
+    case 'request': {
+      const { stagedRendering } = workUnitStore
+      if (stagedRendering) {
+        // We can either recover a static shell or a runtime shell, but not both.
+        trackIncompatibleShellContent(workUnitStore)
+        const stage = workUnitStore.needsAppShell
+          ? RenderStage.NavigationRuntime // Match the timing of 'prerender-runtime'.
+          : RenderStage.Static // Match the timing of 'prerender'.
+
+        return stagedRendering.delayUntilStage(stage, 'navigation', undefined)
```
unstubbable: Should the display name match the current API name?
```suggestion
        return stagedRendering.delayUntilStage(stage, 'unstable_navigation', undefined)
```

test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts
```diff
@@ -429,6 +429,221 @@ describe('App Shell prefetching', () => {
...
+          .elementByCss('input[data-link-accordion="/static-navigation/1"]')
+          .click()
+      }, [{ includes: 'App shell for navigation', kind: 'static' }])
+
+      await act(async () => {
+        // Click the link to /static-navigation/2. This link is rendered with
+        // prefetch={false}, so it was never prefetched. The cached App Shell
+        // should render immediately, before any navigation response arrives.
+        await browser.elementByCss('a[href="/static-navigation/2"]').click()
+        await new Promise((resolve) => setTimeout(resolve, 10_000))
```
unstubbable: Leftover from debugging, or actually "load-bearing" for this test?

test/e2e/app-dir/segment-cache/prefetch-app-shell/next.config.ts
```diff
@@ -8,6 +8,7 @@ const nextConfig: NextConfig = {
     cachedNavigations: true,
     varyParams: true,
   },
+  productionBrowserSourceMaps: true,
```
unstubbable: Leftover from debugging? It might slow down the builds in CI if we keep it.

test/e2e/app-dir/segment-cache/prefetch-static-shell/app/uses-runtime-after-navigation/page.tsx
```diff
@@ -0,0 +1,52 @@
...
+        fallback={<p id="navigation-loading">Loading navigation content...</p>}
+      >
+        <Navigation />
+      </Suspense>
+    </main>
+  )
+}
+
+async function Navigation() {
+  // The runtime data tracking should ends here, so the prerender
```
unstubbable:
```suggestion
  // The runtime data tracking should end here, so the prerender
```

test/e2e/app-dir/segment-cache/prefetch-static-shell/app/page.tsx
```diff
@@ -25,6 +25,21 @@ export default function Page() {
...
+            Uses runtime APIs after navigation()
+          </LinkAccordion>
+        </li>
+        <li>
+          <LinkAccordion href="/uses-navigation-static">
+            Uses navigation() on a static page
+          </LinkAccordion>
+        </li>
+        <li>
+          <LinkAccordion href="/uses-navigation-runtime">
```
unstubbable: I don't think this route exists.

test/e2e/app-dir/segment-cache/prefetch-static-shell/prefetch-static-shell.test.ts
```diff
@@ -102,6 +102,48 @@ describe('static App Shell prefetch attempt', () => {
...
+  it('prefetches a fully static route that uses navigation() with static requests only, then navigates instantly from cache', async () => {
+    let page: Playwright.Page
+    const browser = await next.browser('/', {
+      beforePageLoad(p: Playwright.Page) {
+        page = p
+      },
+    })
+    const act = createRouterAct(page, { includeAppShellRequests: true })
+
+    // Reveal the LinkAccordion for /fully-static. The route accesses no
```
unstubbable:
```suggestion
    // Reveal the LinkAccordion for /uses-navigation-static. The route accesses no
```

unstubbable (commented):
Looks good overall, just a couple of minor things.

I wonder if these stage names wouldn't make more sense:

```diff
export enum RenderStage {
  Before = 1,
  //
  ShellStatic = 10,
  Static = 11, <--------------- params, prefetch() [static prerenders]
+ NavigationStatic = 12, <----- navigation()       [static prerenders]
  //
  ShellRuntime = 20,
  Runtime = 21, <-------------- params, prefetch() [runtime prerenders]
+ NavigationRuntime = 22, <---- navigation()       [runtime prerenders]
  //
  Dynamic = 30,
  Abandoned = 40,
}
```

That would also avoid the need for `STATIC_PRERENDER_NAVIGATION_STAGE` and this would look nicer:

```
const stage = workUnitStore.needsAppShell
  ? RenderStage.NavigationRuntime // Match the timing of 'prerender-runtime'.
  : RenderStage.NavigationStatic // Match the timing of 'prerender'.
```

And more importantly, the import of `STATIC_PRERENDER_NAVIGATION_STAGE` currently creates a cycle which is why the additional modules show up in `test/production/next-server-nft/next-server-nft.test.ts`. So that could be avoided.

test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts
lubieowoce: oof. thanks

unstubbable (approved):

PR #96910, opened 2026-08-07
test: cover dynamic APIs with i18n base path

anthonyshew:
### What?

Adds a test for i18n basepath fallbacks.

### Why?

We had this test in `vercel/vercel`: https://github.com/vercel/vercel/blob/main/packages/next/test/fixtures/00-i18n-basepath-fallback-false-404/index.test.js

We're getting rid of it in favor of moving it to here in Next.js. This new test is meant to be equivalent to the original.


mischnic:
Had to push this branch to the main repo so that all tests can run: https://github.com/vercel/next.js/pull/97367

PR #96914, opened 2026-08-07
Classify framework request activity

DavidIlie:
Stacked on #98677.

## Summary

- Assign server-owned Request Insights identities and classify page, API, image, asset, RSC, Proxy and Instant Insights activity from framework metadata.
- Classify streaming App Routes before their handlers finish. Join Proxy and page activity using the request identity and retain their full time range.
- Treat browser correlation IDs as validated debug hints, not storage identities. Skip this work when Request Insights is disabled.
- Keep public OpenTelemetry span names and parent relationships unchanged.

## Verification

- 75 focused tracing, store and request-label tests passed at this layer.
- The assembled stack passed Request Insights development tests under both bundlers and the Webpack standalone production suite.

### DIFF
--- a/packages/next/src/server/base-server.ts
+++ b/packages/next/src/server/base-server.ts
@@ -114,7 +114,10 @@ import {
   SpanStatusCode,
 } from './lib/trace/tracer'
 import { BaseServerSpan } from './lib/trace/constants'
-import { runWithRequestInsightsIdentity } from './lib/trace/request-insights-identity'
+import {
+  resolveRequestInsightsIdentity,
+  runWithRequestInsightsIdentity,
+} from './lib/trace/request-insights-identity'
 import { isRequestInsightsEnabled } from './lib/trace/span-store'
 import { I18NProvider } from './lib/i18n-provider'
 import { sendResponse } from './send-response'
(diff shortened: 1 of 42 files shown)


packages/next/src/server/base-server.ts
```diff
@@ -1001,25 +1004,33 @@ export default abstract class Server<
       return handleRequest()
     }
 
-    const requestIdHeader = req.headers[NEXT_REQUEST_ID_HEADER]
-    const requestId =
-      typeof requestIdHeader === 'string' ? requestIdHeader : nanoid()
-    const htmlRequestIdHeader = req.headers[NEXT_HTML_REQUEST_ID_HEADER]
+    const requestInsightsIdentity = resolveRequestInsightsIdentity({
```
timneutkens: The object should be `readonly` so that there aren't changes made to the object. I.e. below it shows `.source` being edited. That makes it a pain to refactor later.

DavidIlie: Done — made Request Insights identities readonly and replaced direct source and Proxy-status mutations with immutable snapshots. The focused unit tests and Request Insights dev suite pass with both Turbopack and Webpack.

timneutkens (approved):

PR #96915, opened 2026-08-07
Record Instant Insights validation worker spans

DavidIlie:
Continued in #98677. This PR was marked merged into a feature branch during a stack reorder, not into canary.

## Before this PR

When dev validation ran in a worker, Instant Insights opened `Prepare validation inputs` and `Run validation` on the main dev-server thread, but the component loading and validation renders happened in a worker. That worker could not write to the main thread's in-memory tracing state, so Request Insights showed the outer operation without its worker spans.

## After this PR

- Collect completed Next.js spans inside a per-validation AsyncLocalStorage collector in the worker.
- Return one bounded batch through the existing validation result message.
- Keep the real parent span and request identity on the main thread. Remap worker span IDs and attach worker roots beneath `Run validation`.
- Include validation warmup, static-shell validation, and instant-navigation validation in the trace.
- Finish the collector before worker cleanup and report any omitted spans.

There is no fake parent span, temporary global recorder swap, or second OpenTelemetry provider. This does not change cache behavior or WorkStore state, and it does not forward user OpenTelemetry spans across threads.

## Verification

- Recorder, tracer, and span-store unit tests covering collector isolation, batch import, and parent relationships.
- Request Insights development tests with Turbopack and Webpack, including separate Instant Insights records and validation phase spans.

These checks passed on the refreshed cumulative Request Insights stack. The full local verification matrix is in #96914; each PR's GitHub checks cover its own head.

### DIFF
--- a/packages/next/src/next-devtools/dev-overlay/components/request-insights/fetch-label.test.ts
+++ b/packages/next/src/next-devtools/dev-overlay/components/request-insights/fetch-label.test.ts
@@ -1,0 +1,60 @@
+import { getFetchUrlPresentation, truncateMiddle } from './fetch-label'
+
+describe('request insight fetch labels', () => {
+  it('classifies same-origin and external fetches by full web origin', () => {
+    expect(
+      getFetchUrlPresentation('/api/items?token=redacted', 'https://app.test')
+    ).toEqual({
+      fullUrl: 'https://app.test/api/items?token=redacted',
+      path: '/api/items?token=redacted',
+      originKind: 'same-origin',
+      originLabel: 'Same origin',
+    })
+    expect(
+      getFetchUrlPresentation('https://api.test:445/items', 'https://app.test')
+    ).toEqual({
+      fullUrl: 'https://api.test:445/items',
+      path: '/items',
+      originKind: 'external',
+      originLabel: 'External origin · api.test:445',
+    })
+    expect(
+      getFetchUrlPresentation('http://app.test/items', 'https://app.test')
+    ).toEqual({
+      fullUrl: 'http://app.test/items',
+      path: '/items',
+      originKind: 'external',
+      originLabel: 'External origin · http://app.test',
+    })
+  })
+
+  it('does not guess origin semantics without a usable browser origin', () => {
+    expect(
+      getFetchUrlPresentation('https://api.test/items', undefined)
+    ).toEqual({
+      fullUrl: 'https://api.test/items',
+      path: '/items',
+      originKind: 'unknown',
+      originLabel: 'https://api.test',
+    })
+    expect(getFetchUrlPresentation('/api/items', undefined)).toEqual({
+      fullUrl: '/api/items',
+      path: '/api/items',
+      originKind: 'unknown',
+      originLabel: 'Origin unavailable',
+    })
+  })
+
+  it('middle-truncates labels while retaining their distinguishing suffix', () => {
+    const value = '/api/a/very/long/path/to/tail-marker?token=redacted'
+    const truncated = truncateMiddle(value, 28)
+
+    expect(truncated).toHaveLength(28)
+    expect(truncated).toContain('…')
+    expect(truncated).toMatch(/token=redacted$/)
+
+    expect(truncateMiddle('😀abc', 3)).toBe('😀…c')
+    expect(truncateMiddle('secret', 1)).toBe('…')
+    expect(truncateMiddle('secret', 0)).toBe('')
+  })
+})
--- a/packages/next/src/next-devtools/dev-overlay/components/request-insights/fetch-label.ts
+++ b/packages/next/src/next-devtools/dev-overlay/components/request-insights/fetch-label.ts
@@ -1,0 +1,111 @@
+export type FetchOriginKind = 'same-origin' | 'external' | 'unknown'
+
+export type FetchUrlPresentation = {
+  fullUrl: string
+  path: string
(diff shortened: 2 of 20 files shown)


packages/next/src/next-devtools/server/devtools-config-middleware.ts
```diff
@@ -36,36 +38,63 @@ export function devToolsConfigMiddleware({
...
+    const updateCompleted = new Promise<void>((resolve) => {
+      releaseUpdate = resolve
+    })
+    const queueTail = previousUpdate.then(() => updateCompleted)
+    configUpdateQueues.set(configPath, queueTail)
+    void queueTail.then(() => {
+      if (configUpdateQueues.get(configPath) === queueTail) {
+        configUpdateQueues.delete(configPath)
+      }
+    })
```
timneutkens: This set of changes looks and a bit complicated. There must be a simpler approach

DavidIlie: Done — I removed the entire generic DevTools config persistence hardening commit. This PR now only isolates the Instant Insights validation work store and async context; it no longer changes /__nextjs_devtools_config or its middleware.

lubieowoce:
TLDR on the motivation: when instant validation runs in a worker thread, normal tracing doesn't work, because we're on a different thread and tracing is all in-memory. so we need to capture the tracing spans that happen in the worker and send them back over to the main thread to be stitched back into its tracing. as long as we set the parent span id correctly we don't even really need the Request Insights identity ALS, because really all that it currently holds is the request id

packages/next/src/server/lib/trace/local-span-recorder.ts
```diff
@@ -180,6 +222,55 @@ type RequestIdentity = {
   url?: string
 }
 
+class PropagatedLocalSpan implements Span {
```
lubieowoce: this code is weird, could we have some explanation for why it looks like it does? e.g. why its `traceId` is equal to its parent's traceId? i guess it's meant to be a passthrough no-op (ish) so most of the methods do nothing, but that's not obvious

(might also be worth admitting that this is *kind of a hack* lol)

packages/next/src/server/lib/trace/local-span-recorder.ts
```diff
@@ -137,37 +156,60 @@ export type LocalSpanRecorder = {
...
   }
 }
 
-function getLocalSpanAsyncStorage(): AsyncLocalStorage<Span> {
-  if (!localSpanAsyncStorage) {
-    const { createAsyncLocalStorage } =
-      require('../../app-render/async-local-storage') as typeof import('../../app-render/async-local-storage')
-    localSpanAsyncStorage = createAsyncLocalStorage()
+export function registerLocalSpanRecorder(): void {
+  ;(globalThis as GlobalWithLocalSpanRecorder)[LOCAL_SPAN_RECORDER_KEY] =
```
lubieowoce, line 188: using a global here seems sketchy, what if we're running two validations (from two different tabs) in parallel?

packages/next/src/server/lib/trace/local-span-recorder.ts
```diff
@@ -49,15 +52,16 @@ type TraceLocalSpanOptions = Omit<
...
-
-const getLocalTraceId = () =>
-  (++lastLocalTraceId).toString(16).padStart(TRACE_ID_HEX_LENGTH, '0')
+function getRandomHexId(length: number): string {
+  let id: string
+  do {
+    id = randomBytes(length / 2).toString('hex')
+  } while (/^0+$/.test(id))
+  return id
+}
```
lubieowoce, line 58: are we changing this so that we have consistently unique span IDs across threads? in that case idk if randomBytes is a good idea because it doesn't guarantee that. we might want a guid instead or something?

packages/next/src/server/lib/trace/local-span-recorder.ts
lubieowoce: ...or do workers nott actually need two local recorders and this is effectively a global? like maybe i'm pattern-matching too hard idk

packages/next/src/server/lib/trace/local-span-recorder.ts
DavidIlie: The fake parent span is gone. The worker returns its completed spans, and the main thread attaches their roots to the real Run validation span.

packages/next/src/server/lib/trace/local-span-recorder.ts
DavidIlie: The global now contains only a versioned stateless API. Each validation keeps its own collector in AsyncLocalStorage, so concurrent runs cannot mix spans.

packages/next/src/server/lib/trace/local-span-recorder.ts
DavidIlie: I removed randomBytes. Worker IDs are temporary counters, and the main thread replaces every span, parent, and link ID before recording the batch.

packages/next/src/server/lib/trace/local-span-recorder.ts
DavidIlie: Updated: #97793 externalizes the recorder and span store, the span ID counters are module-scoped again, and the remaining versioned global contains only the recorder functions needed by bundled tracer copies such as the development proxy; each validation’s collector stays isolated in AsyncLocalStorage.

PR #96917, opened 2026-08-07
Track Request Insights lifecycle and causality

DavidIlie:
## Summary

This PR completes Request Insights lifecycle, causality, Edge-runtime, and cross-bundle correctness on top of the bounded per-server controller.

- `Record Request Insights response delivery lifecycle`
  - records response start, active streaming, completion, abort, committed status, and late delivery errors
  - handles Node and Web response paths with one-shot terminal transitions
- `Link same-origin Request Insights causality`
  - issues bounded opaque single-use capabilities for same-origin server work
  - validates origin, method, expiry, and ownership before linking
  - keeps causal children as distinct request roots rather than creating an infinitely nested request list
- `Bridge Request Insights into the Edge sandbox`
  - carries the host controller and validated request identity into Edge execution
  - keeps Node and Edge work under the same retention, lifecycle, and causality semantics
- `Harden Request Insights lifecycle integration`
  - covers double-write, close/error races, aborted streams, committed errors, and cleanup ordering
  - preserves framework response behavior while making lifecycle publication deterministic
- `Stabilize nested causality coverage`
  - uses bounded same-origin fixtures and deterministic capability consumption
  - rejects replay, cross-origin, malformed, expired, and forged causal tokens
- `Expose Request Insights response delivery state`
  - adds Active/streaming, Finished, Aborted, Errored, and Unknown delivery filters and overview badges
  - keeps selection stable as an active response transitions to a terminal state
- `Harden Request Insights lifecycle and causality`
  - propagates exact server-owned identity through App Render, Router Server, BaseServer, NextServer, fetch patching, unstable cache, Proxy, Pages, and Edge paths
  - reads sanitized Route Handler test headers directly so webpack fixtures do not depend on ambient async request storage
  - treats non-standard request-ID headers as bounded dev-only hints, never trusted storage keys
- `Deduplicate cross-bundle Request Insights fetches`
  - shares fetch-index ownership through `Symbol.for('@next/request-insights-fetch-index-setter')`
  - avoids cross-bundle `instanceof` assumptions when separate `LocalRecordingSpan` constructors exist
  - proves one foreground fetch and one Instant fetch produce exactly one record each
  - stores the exact Instant identity on the cloned WorkStore so foreground and validation allocators cannot mix
- `Fix Request Insights CI regressions`
  - keeps strict test fixtures aligned with the final identity and timestamp contracts
  - skips the Edge-only bridge suite under Cache Components, which intentionally requires the Node.js runtime, while retaining normal Turbopack and Webpack Edge coverage
- `Keep Request Insights out of production traces`
  - loads the MCP Request Insights tool only inside the compile-time development branch
  - restores the existing standalone NFT snapshot instead of accepting DevTools code in production output
- `Handle unknown Request Insights values`
  - maps future or version-skewed request sources to the existing Unknown presentation
  - maps future response outcomes to the bounded Unknown delivery filter instead of returning an undefined UI model
- `Guard Request Insights identity loading`
  - places identity, WorkStore, and controller imports inside an explicit compile-time dev-server branch
  - preserves production dead-code elimination while retaining the same development identity precedence

Deliberate boundary: this does not add a public Server Action OTel span, retain raw action IDs, add per-component React-private RSC timings, or introduce per-use cache ownership UI.

## Verification

- focused lifecycle, stream, sandbox, and Instant suites: 28/28 tests
- controller ownership: 4/4 tests
- Request Insights development suite: 51/51 in Turbopack and 51/51 in Webpack
- Edge bridge: 8/8 in Turbopack and 8/8 in Webpack
- route preparation: 4/4 in Turbopack and 4/4 in Webpack
- `pnpm build-all`: 18/18 at the unchanged final tree
- exact layer head passes `pnpm --filter=next types`
- Cache Components correctly skips the unsupported Edge-runtime bridge matrix rather than producing fixture 500s
- production standalone NFT coverage retains its original snapshot; Request Insights DevTools code is eliminated
- changed-file Prettier, ESLint, `git diff --check`, authorship, subject-only commit message, forbidden-attribution, production-DCE, public-OTel, and 50-commit linear-history audits pass
- exact CI type/precompiled reproduction passes after moving package-internal tests to their owning package
- focused package-boundary, production-DCE, version-skew, rejected-chunk, and identity hardening coverage passes 47/47
- the previously frozen Fable 5 review returned `SHIP` with zero blockers; its concrete current-tree findings are fixed and a new exact-head review will run after GitHub CI is green

### DIFF
--- a/packages/next/errors.json
+++ b/packages/next/errors.json
@@ -1472,5 +1472,6 @@
   "1471": "Response stream failed before its first chunk",
   "1472": "Request Insights maxRequestGroupsPerBucket must be an integer from 1 to %s.",
   "1473": "Request Insights MCP output exceeded its byte limit.",
-  "1474": "Invariant: missing Request Insights runtime"
+  "1474": "Invariant: missing Request Insights runtime",
+  "1475": "Cannot call onResponseEnd on a WebNextResponse that is already closed"
 }
--- a/packages/next/src/cli/next-request-insights-schema.test.ts
+++ b/packages/next/src/cli/next-request-insights-schema.test.ts
@@ -1,0 +1,71 @@
+import type { RequestInsight } from '../next-devtools/shared/request-insights'
+import { isRequestInsightsSnapshot } from './next-request-insights'
+
+function createRequest(
+  requestId: string,
+  overrides: Partial<RequestInsight> = {}
+): RequestInsight {
+  return {
+    requestId,
+    source: 'page',
+    htmlRequestId: requestId,
+    route: `/${requestId}`,
+    startTime: 0,
+    status: 'ok',
+    spans: [],
+    fetches: [],
+    ...overrides,
+  }
+}
+
+describe('next experimental-request-insights schema', () => {
+  it('accepts the combined response lifecycle and causality fields', () => {
+    const snapshot = {
+      requests: [
+        createRequest('child', {
+          rootRequestId: 'child',
+          parentRootRequestId: 'parent',
+          parentFetchIndex: 2,
+          status: 'aborted',
+          response: {
+            trackingStartTime: 1,
+            endTime: 2,
+            statusCode: 200,
+            outcome: 'aborted',
+            error: { type: 'ResponseAborted' },
+          },
+        }),
+      ],
+    }
+
+    expect(isRequestInsightsSnapshot(snapshot)).toBe(true)
+    expect(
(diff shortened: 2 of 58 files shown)


DavidIlie:
Superseded by a fresh draft PR at the same verified branch head so GitHub can run the full opened-event CI matrix.

PR #96919, opened 2026-08-07
[ci] Use OIDC tokens to read private preview builds

eps1lon:
Replaces the shared, static `PREVIEW_BUILDS_READ_TOKEN` secret with OIDC tokens for reading auth-protected preview builds from vercel-packages (see https://github.com/vercel/vercel-packages/pull/96):

- CI jobs mint a GH oidc token for the `https://vercel-packages.vercel.app` audience 
- deploys have a `.npmrc` written that uses the Vercel OIDC token

Tarball polling now also fails fast on 401/403 (retrying can't change the authorization outcome) and failure messages include response headers so request IDs are available for debugging.

## Test plan

- [x] Vercel OIDC path: mirror deploy test installed `next` from the protected route via the build's `VERCEL_OIDC_TOKEN` using the preview deployment on https://github.com/vercel/vercel-packages/pull/96: [e2e deploy test job](https://github.com/vercel/next-js-mirror/actions/runs/31195027603/job/92936310701) -> [vercel deployment (ignore the version, it was a dummy upload](https://vercel.com/vtest314-next-e2e-tests/vtest314-e2e-tests/3hHME9D3A5LU9k8FTWNaQzk9Xesd) -> [authorized request](https://vercel.com/vtest314-next-e2e-tests/vercel-packages/7ixXx4zvwN47JLjREVtds3xdENm3/logs?selectedLogId=zg56v-1786122054430-415c6d26ef41&panelState=opened&search=requestId%3Azg56v-1786122054430-415c6d26ef41)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -752,7 +752,12 @@ jobs:
     # Outer bound only. The script gives up first, so a tarball that never
     # arrives is reported by it rather than by the runner killing the job.
     timeout-minutes: 35
-    permissions: {}
+    permissions:
+      # contents: read is needed to check out in private repos (the mirror).
+      contents: read
+      # Minting GitHub OIDC tokens to read private preview builds. Dormant in
+      # this repo since builds here are public.
+      id-token: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
@@ -766,7 +771,7 @@ jobs:
       - name: Wait for preview tarball
         env:
           PREVIEW_BUILDS_BASE_URL: ${{ vars.PREVIEW_BUILDS_BASE_URL }}
-          PREVIEW_BUILDS_READ_TOKEN: ${{ secrets.PREVIEW_BUILDS_READ_TOKEN }}
+          PREVIEW_BUILDS_ACCESS: ${{ vars.PREVIEW_BUILDS_ACCESS }}
         run: |
           set -euo pipefail
           node scripts/wait-for-preview-tarball.mjs \
@@ -785,6 +790,11 @@ jobs:
       matrix:
         group: [1/5, 2/5, 3/5, 4/5, 5/5]
 
+    permissions:
+      contents: read
+      # Minting GitHub OIDC tokens to read private preview builds. Dormant in
+      # this repo since builds here are public.
+      id-token: write
     uses: ./.github/workflows/build_reusable.yml
     with:
       # Keep Next.js related env variables in sync with additionalEnv in next-deploy.ts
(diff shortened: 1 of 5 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -735,7 +735,12 @@ jobs:
     # Outer bound only. The script gives up first, so a tarball that never
     # arrives is reported by it rather than by the runner killing the job.
     timeout-minutes: 35
-    permissions: {}
+    permissions:
+      # contents: read is needed to check out in private repos (the mirror).
+      contents: read
```
eps1lon, line 757: drive-by fix for https://github.com/vercel/next-js-mirror/actions/runs/31104879823/job/92627410537

scripts/wait-for-preview-tarball.mjs
```diff
@@ -231,7 +312,7 @@ async function main() {
     commitSha,
     previewBuildsBaseUrl: values['preview-builds-base-url'],
     timeoutMs: timeoutMinutes * 60_000,
-    readToken: process.env.PREVIEW_BUILDS_READ_TOKEN,
+    readToken: await getPreviewBuildsReadToken(),
```
ztanner: I think OIDC tokens only last like 5 minutes, I'm wondering if we need to worry about this potential expiring since we'll retry for up to 30 min?

.github/workflows/build_and_test.yml
```diff
@@ -735,7 +735,12 @@ jobs:
...
+      # Minting GitHub OIDC tokens to read private preview builds. Dormant in
+      # this repo since builds here are public.
+      id-token: write
```
ztanner, line 760: we'll also need this for `test-new-tests-deploy` and `test-new-tests-deploy-cache-components` right?

scripts/wait-for-preview-tarball.mjs
eps1lon: 0cac7fa4e77a9adea14e12e9d3e676e4bf92c7e6 should help

.github/workflows/build_and_test.yml
eps1lon: We don't install preview builds in the GH jobs outside of checking if the preview build is available. The GH jobs use the version built from source directly. And when building on Vercel we use Vercel's OIDC token

ztanner: I’m referring specifically to the availability assertion in `test-new-tests`: it passes `createPreviewBuildsReadTokenGetter()` to `assertPreviewTarballPublished()`, which invokes the getter to mint a GitHub OIDC token when preview builds are private

eps1lon: ....

eps1lon: Should handle that now. I'm not 100% about permission inheritance with reusable builds but we'll see if that works once synced.

ztanner (approved):

PR #96920, opened 2026-08-07
Turbopack: implement Eq, Decode, Encode for ConstantValue

mischnic:
I want to store `ConstantValue` in a cell.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/constants.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/constants.rs
@@ -5,6 +5,7 @@ use std::{
     sync::Arc,
 };
 
+use bincode::{Decode, Encode};
 use num_bigint::BigInt;
 use num_traits::Zero;
 use swc_core::{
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/constants.rs
```diff
@@ -52,6 +54,10 @@ impl ConstantNumber {
     }
 }
 
+/// Not strictly correct: `PartialEq` above uses `f64`'s equality, so `NaN != NaN` (violating
+/// reflexivity) and `0.0 == -0.0` even though the two hash differently below.
+impl Eq for ConstantNumber {}
```
mischnic: Either it's this, or using `eq = "manual"` for every consumer, right?

lukesandberg: should we implement Eq/PartialEq in terms of `to_ne_bytes` to have a correct if confusing implementation?

i guess no...

we could have some kind of rule where

* NaN and `-0` are never stored in ConstantNumber?

or our projection to store in a Cell could literally be the bincode encoded value???

mischnic: the real problem is that we use constantnumber equality for js analyzer semantic purposes: https://github.com/vercel/next.js/pull/94172

so the best course of action is probably adding a wrapper type that has byte-based eq+hash.

mischnic:
I ended up going with CompileTimeDefineValue after all

PR #96921, opened 2026-08-07
Bound and isolate Request Insights capture

DavidIlie:
## Summary

This PR makes Request Insights capture explicitly owned, bounded, observable, and resilient under high request volume. It also aligns the DevTools, CLI, and MCP projections around the same controller model.

- `Scope Request Insights to a per-server controller`
  - replaces process-global request capture with one controller owned by each development server
  - carries controller ownership through AsyncLocalStorage and prevents separate local servers from sharing records
- `Fix Request Insights controller validation`
  - hardens controller lookup and rejects mismatched or stale ownership rather than silently falling back to global state
- `Bound Request Insights capture`
  - caps logical roots, serialized bytes, records, spans, fetches, events, links, identifiers, metadata, and snapshots
  - raises the configurable per-type maximum to 200 while keeping a larger global byte quota to prevent unbounded memory growth
  - separates permanently omitted capture from publication-only projection loss
- `Add Request Insights capture controls`
  - adds compact usage progress, validated per-type limits, and Clear captured data
  - keeps capture-history presets and reset-to-200 controls out of the UI
  - exposes bounded CLI mutation for set/clear while MCP remains bounded read-only access
- `Make Request Insights HMR delivery authoritative and bounded`
  - adds generation, sequence, retention revision, bounded buffering, and authoritative resynchronization
  - prevents records from flashing in and out when HMR reconnects or the serialized store reaches its limit
- `Harden bounded Request Insights capture`
  - covers oversized identifiers, metadata, events, links, snapshots, malformed input, omission accounting, and deterministic retention order
- `Explain omitted Request Insights capture data`
  - distinguishes omitted logical groups from related activity without adding noisy retained-group copy to the DevTools UI
  - keeps the CLI's explicit retained-group context for bounded terminal output
- `Use authoritative snapshots in overlay regressions`
  - makes overlay tests consume the same typed authoritative snapshots as production HMR delivery
  - removes stale partial-update assumptions from UI regressions
- `Align Request Insights query projections`
  - shares bounded projection semantics across DevTools, CLI, and MCP reads
  - preserves terminal-control escaping and maximum-result enforcement
- `Update Request Insights test expectations`
  - uses `NEXT_OTEL_VERBOSE=1` only for tests that intentionally require internal phase data
  - aligns the CLI wording and bounded snapshot fixtures with the final controller contract
- `Harden Request Insights CI coverage`
  - runs every development cleanup callback before removing controller test fixtures, matching the real server lifecycle and avoiding Windows filesystem races
  - makes dev-only Request Insights loaders use explicit compile-time branches so production bundling can eliminate them
  - keeps strict fixture types aligned with the final identity, HMR, span, and response callback contracts
- `Keep Request Insights CLI tests package-local`
  - keeps bounded serialization coverage beside the CLI and shared snapshot modules it imports
  - prevents root-project typechecking from resolving the Next.js package through an incomplete compiled declaration surface
- `Keep Request Insights data out of DevTools traces`
  - moves the canonical non-UI capture limits, byte accounting, projection model, and protocol data into the existing `shared/lib` boundary
  - keeps the DevTools module as a compatibility re-export while server, HMR, CLI, and MCP code consume the shared implementation directly
  - preserves the original standalone NFT snapshot instead of accepting a DevTools-only module in production traces
- `Wait for Request Insights route matchers in tests`
  - waits for the router's initial asynchronous matcher reload before closing and deleting the ownership fixture
  - removes the Windows-only `pages` scan race without sleeps, retries, or a production lifecycle change

## Verification

- exact layer head passes `pnpm build-all` (18/18) and `pnpm --filter=next types`
- focused controller ownership, validation, retention, omission, query projection, CLI/MCP, terminal escaping, HMR backpressure, authoritative resync, and production-DCE tests
- Request Insights suite verifies configurable limits, clear, omission reporting, bounded CLI queries, malformed responses, secure URL handling, and no record flashing in both bundlers
- controller ownership passes 4/4, including two-server isolation and explicit async teardown
- production NFT coverage passes 5/5 with its original snapshot and no DevTools Request Insights module in the standalone trace
- controller ownership passes 4/4 after awaiting the exact initial matcher lifecycle used by the router
- predecessor Fable 5 production review returned `SHIP`; exact-head GitHub CI remains the merge gate


DavidIlie:
false

PR #96927, opened 2026-08-07
Complete App Router render observability

DavidIlie:
## Summary

This PR completes the internal App Router render timeline that Request Insights consumes. The spans remain internal or verbose by default and do not expand the public OpenTelemetry allowlist.

- `Trace App Router render preparation`
  - records `prepare app page response` and `initialize app render` around the bounded setup work that happens before the route starts producing RSC or HTML
  - keeps the instrumentation close to the actual phase boundaries instead of deriving durations later in the Request Insights viewer
- `Trace App Router RSC rendering`
  - records RSC response production and the wait for the RSC render task
  - handles fulfilled, rejected, cancelled, and prematurely closed streams with one-shot span completion
- `Trace route matching and compilation phases`
  - adds bounded internal phases for route matching, route preparation, route entrypoint readiness, and development route building
  - keeps reload work separate so matcher refresh time is not incorrectly charged to route ensuring
- `Fix route phase allowlist test types`
  - keeps the allowlist regression typed without weakening the assertion that these new phases are not default-public OTel spans
- `Fix route compilation error propagation`
  - preserves the original compilation failure while ensuring its tracing span is stopped exactly once
- `Trace HTML completion`
  - records the wait between starting HTML production and the stream actually completing
  - covers Node and Web streams without double-stopping on close/error races
- `Trace time to the first response chunk`
  - records response-start readiness through the first committed chunk
  - preserves streaming semantics and does not treat response construction as response delivery
- `Stabilize first response chunk coverage`
  - makes the later stream failure begin only after the response has observed the first write
  - proves first-chunk success cannot be overwritten by a later error without relying on scheduler timing
- `Verify first response chunk delivery`
  - asserts that the first payload reached the response writer before the synthetic late stream error begins
  - protects the regression test from passing on span state alone without proving real delivery ordering
- `Handle rejected client chunk loads`
  - observes fulfilled and rejected chunk promises without creating an unhandled rejected promise
  - preserves the exact promise returned by the bundler while still closing the client-loading measurement

Together these phases explain cold route preparation, compilation, App Router setup, RSC work, HTML completion, and first-byte readiness without importing the broad mutable BaseServer/Fizz instrumentation explored in earlier prototypes.

## Verification

- focused route-phase ownership, compilation-error, RSC completion, HTML completion, and first-response tests
- Request Insights development suite: 51/51 in Turbopack and 51/51 in Webpack at the final stack tree
- route-preparation development suite: 4/4 in Turbopack and 4/4 in Webpack
- public OTel audit confirms `NextVanillaSpanAllowlist` and `Middleware.execute` are unchanged
- focused first-chunk suite passes 13/13, including the deterministic late-error regression
- exact stack head is included in the final 53-commit linear-history and authorship audit

### DIFF
--- a/packages/next/errors.json
+++ b/packages/next/errors.json
@@ -1467,5 +1467,7 @@
   "1466": "TypeScript %s does not provide the compiler API required by Next.js. Set %s to true in your Next.js config to use the TypeScript CLI, or install TypeScript 6 instead.",
   "1467": "TypeScript %s does not provide the compiler API required by Next.js. Set %s back to true in your Next.js config to use the TypeScript CLI, or install TypeScript 6 instead.",
   "1468": "Missing workUnitStore in createComponentTree",
-  "1469": "%s stream closed before completion"
+  "1469": "%s stream closed before completion",
+  "1470": "Response closed before its first chunk",
+  "1471": "Response stream failed before its first chunk"
 }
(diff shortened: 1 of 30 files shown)


DavidIlie:
Closing after review and reducing the launch scope. The stack keeps development route preparation and compilation visibility, while removing the additional below-rendering RSC, HTML-completion, and first-response instrumentation because of its complexity and runtime overhead. The rejected client chunk-load correctness fix is now folded into #96455.

PR #96929, opened 2026-08-07
turbo-persistence: add key-value tombstones for MultiValue families

lukesandberg:
Add a new `tombstone` format to the persistence layer so we can delete key-value pairs out of MultiValued tables

This is in service of the upcoming GC support, but also fills a basic API gap in the db.

To delete a key-value-pair you need to call `value_delete` and currently the values are limited to only those that are able to be stored inline in key blocks.  This is a non-trivial limitation but it fits our current usecase, and makes the compaction/query algorithms a bit simpler (don't need to 'resolve' values)

One non-trivial complexity was maintaining the 'FixedLayout' block optimization for a mix of tombstones and values, so now we support a slightly different mode where all keys have the same length but possibly different types.

Finally, this branch solves a problem with deleting tombstones.  Tombstones 'shadow' older values and allow us to drop them during compaction.  With GC getting ready to start writing tombstones the risk becomes 'when can we delete a tombstone! This is solved probabilistically using the amqf filters, during compaction we drop tombstones if they could not possibly match anything in an older SST.    Without this, tombstones in the TaskCache table would fill up over time.

### DIFF
--- a/turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
+++ b/turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
@@ -6,8 +6,9 @@
 //! Entry types:
 //! - 0: Small value (stored in value block)
 //! - 1: Blob reference
-//! - 2: Deleted/tombstone
+//! - 2: Key tombstone (deletes all values for the key)
 //! - 3: Medium value
+//! - 4: Key-value tombstone (deletes one value from the key's group)
 //! - 8-255: Inline value where (type - 8) = value byte count
 
 use std::{
@@ -28,8 +29,9 @@ use turbo_persistence::{
     sst_filter::SstFilter,
     static_sorted_file::{
         BLOCK_TYPE_FIXED_KEY_NO_HASH, BLOCK_TYPE_FIXED_KEY_WITH_HASH, BLOCK_TYPE_KEY_NO_HASH,
-        BLOCK_TYPE_KEY_WITH_HASH, KEY_BLOCK_ENTRY_TYPE_BLOB, KEY_BLOCK_ENTRY_TYPE_DELETED,
-        KEY_BLOCK_ENTRY_TYPE_INLINE_MIN, KEY_BLOCK_ENTRY_TYPE_MEDIUM, KEY_BLOCK_ENTRY_TYPE_SMALL,
+        BLOCK_TYPE_KEY_WITH_HASH, KEY_BLOCK_ENTRY_TYPE_BLOB, KEY_BLOCK_ENTRY_TYPE_INLINE_MIN,
+        KEY_BLOCK_ENTRY_TYPE_KEY_DELETED, KEY_BLOCK_ENTRY_TYPE_KEY_VALUE_DELETED,
+        KEY_BLOCK_ENTRY_TYPE_MEDIUM, KEY_BLOCK_ENTRY_TYPE_SMALL,
     },
 };
 
@@ -96,10 +98,11 @@ struct SstStats {
 
     /// Value sizes by type (inline values track actual bytes)
     inline_value_bytes: u64,
-    small_value_refs: u64,  // Count of references to value blocks
-    medium_value_refs: u64, // Count of references to medium values
-    blob_refs: u64,         // Count of blob references
-    deleted_count: u64,     // Count of deleted entries
+    small_value_refs: u64,        // Count of references to value blocks
+    medium_value_refs: u64,       // Count of references to medium values
+    blob_refs: u64,               // Count of blob references
+    deleted_count: u64,           // Count of key tombstones
+    key_value_deleted_count: u64, // Count of key-value tombstones
 
     /// File size in bytes
     file_size: u64,
@@ -122,6 +125,7 @@ impl SstStats {
         self.medium_value_refs += other.medium_value_refs;
         self.blob_refs += other.blob_refs;
         self.deleted_count += other.deleted_count;
+        self.key_value_deleted_count += other.key_value_deleted_count;
         self.file_size += other.file_size;
     }
 }
@@ -144,8 +148,11 @@ fn track_entry_type(stats: &mut SstStats, entry_type: u8) {
         KEY_BLOCK_ENTRY_TYPE_BLOB => {
             stats.blob_refs += 1;
         }
-        KEY_BLOCK_ENTRY_TYPE_DELETED => {
+        KEY_BLOCK_ENTRY_TYPE_KEY_DELETED => {
             stats.deleted_count += 1;
+        }
+        KEY_BLOCK_ENTRY_TYPE_KEY_VALUE_DELETED => {
+            stats.key_value_deleted_count += 1;
         }
         KEY_BLOCK_ENTRY_TYPE_MEDIUM => {
             stats.medium_value_refs += 1;
@@ -162,7 +169,8 @@ fn entry_type_description(ty: u8) -> String {
     match ty {
         KEY_BLOCK_ENTRY_TYPE_SMALL => "small value (in value block)".to_string(),
         KEY_BLOCK_ENTRY_TYPE_BLOB => "blob reference".to_string(),
-        KEY_BLOCK_ENTRY_TYPE_DELETED => "deleted/tombstone".to_string(),
+        KEY_BLOCK_ENTRY_TYPE_KEY_DELETED => "key tombstone".to_string(),
+        KEY_BLOCK_ENTRY_TYPE_KEY_VALUE_DELETED => "key-value tombstone".to_string(),
         KEY_BLOCK_ENTRY_TYPE_MEDIUM => "medium value".to_string(),
         ty if ty >= KEY_BLOCK_ENTRY_TYPE_INLINE_MIN => {
             let inline_size = ty - KEY_BLOCK_ENTRY_TYPE_INLINE_MIN;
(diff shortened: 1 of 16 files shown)


turbopack/crates/turbo-persistence/src/collector_entry.rs
```diff
@@ -42,7 +48,8 @@ impl CollectorEntryValue {
             CollectorEntryValue::Small { value } => value.len(),
             CollectorEntryValue::Medium { value } => value.len(),
             CollectorEntryValue::Large { blob: _ } => 0,
-            CollectorEntryValue::Deleted => 0,
+            CollectorEntryValue::KeyDeleted => 0,
+            CollectorEntryValue::KeyValueDeleted { .. } => 0,
```
lukesandberg: why not counted

turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -95,10 +98,11 @@ struct SstStats {
...
     /// Value sizes by type (inline values track actual bytes)
     inline_value_bytes: u64,
-    small_value_refs: u64,  // Count of references to value blocks
-    medium_value_refs: u64, // Count of references to medium values
-    blob_refs: u64,         // Count of blob references
-    deleted_count: u64,     // Count of deleted entries
+    small_value_refs: u64,        // Count of references to value blocks
+    medium_value_refs: u64,       // Count of references to medium values
+    blob_refs: u64,               // Count of blob references
+    deleted_count: u64,           // Count of key tombstones
```
bgw: Should this be `key_deleted_count` now?

turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -383,38 +411,61 @@ fn parse_key_block_header(block: &[u8]) -> Result<KeyBlockHeader> {
...
+                    0
+                };
+                let key_size = block[4] as usize;
+                // +1 for the per-entry type byte.
+                let val_size = block[6] as usize + 1;
+                Ok(KeyBlockHeader::FixedMixedType {
+                    entry_count,
+                    hash_len,
+                    key_size,
+                    stride: hash_len + key_size + val_size,
```
bgw:
```suggestion
                let val_size = block[6] as usize;
                Ok(KeyBlockHeader::FixedMixedType {
                    entry_count,
                    hash_len,
                    key_size,
                    // +1 for the per-entry type byte.
                    stride: hash_len + key_size + val_size + 1,
```

turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -831,9 +889,10 @@ fn main() -> Result<()> {
             eprintln!("Entry types:");
             eprintln!("  0: Small value (stored in separate value block)");
             eprintln!("  1: Blob reference");
-            eprintln!("  2: Deleted/tombstone");
+            eprintln!("  2: Key tombstone (deletes all values for the key)");
             eprintln!("  3: Medium value");
-            eprintln!("  8+: Inline value (size = type - 8)");
+            eprintln!("  8-16: Inline value (size = type - 8)");
+            eprintln!("  17-25: Key-value tombstone (deleted value size = type - 17)");
```
bgw: Could we pull these numbers from the `const`s defined in `turbo_persistence::static_sorted_file` rather than hardcoding the values here and risking them going out of sync?

turbopack/crates/turbo-persistence/src/collector.rs
```diff
@@ -95,7 +96,32 @@ impl<K: StoreKey, const SIZE_SHIFT: usize> Collector<K, SIZE_SHIFT> {
...
+        });
+    }
+
+    /// Adds a key-value tombstone to the collector: deletes only the single `key` -> `value` pair,
+    /// leaving any other values for `key` intact.
+    ///
+    /// Only meaningful for [`FamilyKind::MultiValue`] families. Deleting a pair written in the
+    /// same batch is not supported; see [`WriteBatch::delete_value`][crate::WriteBatch].
+    ///
+    /// `value` must be at most [`MAX_INLINE_VALUE_SIZE`] bytes; callers validate this.
```
bgw: Explain the motivation for this.

turbopack/crates/turbo-persistence/src/collector_entry.rs
```diff
@@ -51,6 +58,20 @@ impl CollectorEntryValue {
...
+        match self {
+            CollectorEntryValue::KeyValueDeleted { value, len }
+            | CollectorEntryValue::Tiny { value, len } => Some(&value[..*len as usize]),
+            CollectorEntryValue::Small { value } | CollectorEntryValue::Medium { value } => {
+                Some(value)
+            }
+            CollectorEntryValue::Large { .. } | CollectorEntryValue::KeyDeleted => None,
+        }
+    }
+
```
bgw, line 76: There's no caller of `as_bytes` in this diff? Is this dead code?

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -1500,7 +1501,39 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
                                     meta,
                                 });
                             }
-
+                            let tombstone_is_dead = {
```
bgw, line 1558: Add some comment explaining that this is probabilistic, and may claim that a tombstone that is not dead is dead (false negative).

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -1629,15 +1662,38 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
...
                             for entry in iter {
                                 let entry = entry?;
                                 if current_key.as_ref() != Some(&entry.key) {
                                     // we changed keys so undo this flag
                                     skip_remaining_for_this_key = false;
+                                    deleted_values_for_this_key.clear();
                                     current_key = Some(entry.key.clone());
                                 }
+                                // Key-value tombstones sort first within a group, so each is
+                                // recorded before the values it might delete.
```
bgw, line 1733:
```suggestion
                                // Key-value tombstones sort first within a group, so each is
                                // recorded before the values it might delete.
                                // See: `crate::collector_entry::sort_rank`
```

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -1629,15 +1662,38 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
...
+                                    deleted_values_for_this_key.push(value.clone());
+                                    // Applied to this job's values above; keep it only if an SST
+                                    // outside the job could still hold a matching key.
+                                    if tombstone_is_dead(entry.hash) {
+                                        continue;
+                                    }
+                                } else if !deleted_values_for_this_key.is_empty()
+                                    // Deleted values cannot match blobs, just normal payloads.
+                                    && let IterValue::Slice { value } = &entry.value
+                                    // O(N) but in practice there is never more than 1 element.
```
bgw:
```suggestion
                                    // O(N) but in practice there is never more than 1 element.
                                    // because turbo-tasks-backend stores 64-bit hashes as keys
                                    // and it would require a 64-bit hash collision
```

turbopack/crates/turbo-persistence/src/meta_file.rs
```diff
@@ -49,6 +49,14 @@ impl Display for MetaEntryFlags {
     }
 }
 
+/// Magic number identifying a `.meta` file, and implicitly its format version.
+///
+/// Bump this whenever [`EntryHeader`]'s layout changes. The header is a fixed-size struct read at a
+/// computed stride, so an older file parsed by newer code would not fail — it would silently
+/// misinterpret every field after the change. Rejecting the file outright is the only safe
+/// outcome; callers handle the error by discarding and rebuilding the database.
```
bgw: slop

```suggestion
```

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -40,16 +40,33 @@ pub const BLOCK_TYPE_FIXED_KEY_WITH_HASH: u8 = 3;
 /// The block header for a fixed-size key block without hash.
 pub const BLOCK_TYPE_FIXED_KEY_NO_HASH: u8 = 4;
 
+/// Written in a fixed-size key block header's value type field when entries share a value size but
+/// not a value type. Each entry then carries its own type byte ahead of its value.
+///
+/// Tag 4 is safe as a sentinel because it is not a valid entry type: it was the old key-value
+/// tombstone tag, freed when tombstones moved to their own range above the inline range.
```
bgw: This comment is implying that we're trying to keep backwards compatibility

```suggestion
```

turbopack/crates/turbo-persistence/src/tests.rs
```diff
@@ -2261,6 +2261,566 @@ fn current_file_is_json_with_commit_time() -> Result<()> {
...
+    Ok(())
+}
+
+/// Counts tombstone entries (both kinds) across every live SST, by reading the files directly.
+/// Tombstone counts are not tracked in the meta file, so there is nothing cheaper to read.
+fn count_tombstones(
+    path: &std::path::Path,
+    db: &TurboPersistence<RayonParallelScheduler, 1>,
+) -> Result<usize> {
+    use crate::{lookup_entry::IterValue, static_sorted_file::StaticSortedFileIter};
```
bgw: Hoist these imports

bgw (approved):

turbopack/crates/turbo-persistence/src/collector_entry.rs
lukesandberg: it is called by the verify_sst_content feature, added the guard here as well

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: i guess on second inspection this is not really true, it is true in practice but not in theory lol

this is a solid usecase for an AutoSet

turbopack/crates/turbo-persistence/src/db.rs
lukesandberg: Done. though the reality is that we may claim a tombstone is _live_ when it is actually dead, if this says dead then it must be actually dead.

PR #96932, opened 2026-08-07
Handle Server Actions on dynamic PPR fallback routes

ztanner:
## Summary

A fetch action can be dispatched to a parameterized route fallback without concrete route params. When the deployment adapter also supplies that route’s postponed state, Next.js throws because postponed state and fallback params are present together.

These requests only need to execute the action and cannot safely render the fallback route. This change preserves the Resume Data Cache for cached reads, discards the React postponed state that cannot be resumed without concrete params, and consistently skips fallback-route rendering for successful and failed actions.

## Verification

- Added a `next start` integration test that synthesizes an action-only fallback request and fails without the fix.
- Added coverage for access fallback errors and direct Resume Data Cache extraction.
- Added a deployment test using a public browser interaction where an action is retained across navigation to a route that does not bundle it.


packages/next/src/server/app-render/action-handler.ts
```diff
@@ -1216,7 +1224,7 @@ export async function handleAction({
             boundActionArguments,
             workStore,
             requestStore,
-            actionWasForwarded
+            shouldSkipPageRendering
```
gnoff, line 1227: Maybe worth a follow up to make an entrypoint for actions that don't render and then hoist this to the app-page module

gnoff (approved):

PR #96936, opened 2026-08-07
[refactor] Rename `encodeCacheTag` to `encodeHeaderSafe`

unstubbable:
The helper percent-encodes anything outside Node's valid header-value class so a string can be serialized into an HTTP header. That is a property of the transport, not of cache tags, and the next caller is `unstable_cache`'s synthetic `fetchUrl`, which is a cache item name rather than a tag. Naming the function after one of its callers would make that call site read as though it were tagging something.

This change renames the module to `encode-header-safe.ts` and the function to `encodeHeaderSafe`, and generalizes the leading paragraph of the doc comment accordingly. The rationale specific to cache tags stays, since tag construction and invalidation remain the callers today. The doc now also names the `ByteString` conversion that `fetch` performs on header values, alongside `validateHeaderValue`; both already applied to tags, because the Vercel bridge sends them through `fetch`.

There is no behavior change. The function body is identical apart from the parameter name.

### DIFF
--- a/packages/next/src/server/lib/encode-header-safe.ts
+++ b/packages/next/src/server/lib/encode-header-safe.ts
@@ -1,30 +1,27 @@
 /**
- * Percent-encode every character outside printable ASCII so a value can be
- * safely serialized into an HTTP header.
+ * Percent-encode every character other than horizontal tab and printable ASCII
+ * so a value can be safely serialized into an HTTP header.
  *
- * Node's `validateHeaderValue` rejects any code unit outside `\t\x20-\x7e`,
- * and `fetch` rejects the same values when it converts a header to a
- * ByteString. A value containing a non-ASCII character (Hebrew, Arabic,
- * Chinese, emoji, …) would otherwise throw `ERR_INVALID_CHAR` or a
- * `TypeError`, which crashes ISR or fails a cache read on every affected
- * request.
+ * Node's `validateHeaderValue` and `fetch`'s ByteString conversion accept
+ * different character ranges. Restricting values to `\t\x20-\x7e` produces a
+ * conservative representation that both can serialize. Characters outside a
+ * transport's accepted range (Hebrew, Arabic, Chinese, emoji, …) would
+ * otherwise throw `ERR_INVALID_CHAR` or a `TypeError`, which crashes ISR or
+ * fails a cache read on every affected request.
  *
  * Cache tags are encoded at the public boundaries — tag construction
  * (`getImplicitTags`, `validateTags`) and invalidation input
  * (`revalidatePath`, `revalidateTag`, `updateTag`) — so storage, comparison,
  * and the wire all see the same canonical form.
  *
- * The character class `[\t\x20-\x7e]` mirrors Node's `validHdrChars` table —
- * `\t` plus printable ASCII through `~`. Anything outside that is rejected
- * by `validateHeaderValue`, so we encode runs of those characters and leave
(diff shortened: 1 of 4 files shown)


gnoff (approved):

packages/next/src/server/lib/encode-header-safe.ts
```diff
@@ -1,16 +1,18 @@
...
+ * request.
  *
- * This is applied at the public boundaries — tag construction
- * (`getImplicitTags`, `validateTags`) and invalidation input (`revalidatePath`,
- * `revalidateTag`, `updateTag`) — so storage, comparison, and the wire all see
- * the same canonical ASCII-safe form.
+ * Cache tags are encoded at the public boundaries — tag construction
+ * (`getImplicitTags`, `validateTags`) and invalidation input
+ * (`revalidatePath`, `revalidateTag`, `updateTag`) — so storage, comparison,
+ * and the wire all see the same canonical form.
```
gnoff:
```suggestion
 * Percent-encode every character other than horizontal tab and printable
 * ASCII so a value can be safely serialized into an HTTP header.
 *
 * Node's `validateHeaderValue` and `fetch`'s ByteString conversion accept
 * different character ranges. Restricting values to `\t\x20-\x7e` produces
 * a conservative representation that both can serialize. Characters outside
 * a transport's accepted range (Hebrew, Arabic, Chinese, emoji, …) would
 * otherwise throw `ERR_INVALID_CHAR` or a `TypeError`, which crashes ISR or
 * fails a cache read on every affected request.
 *
 * Cache tags are encoded at the public boundaries — tag construction
 * (`getImplicitTags`, `validateTags`) and invalidation input
 * (`revalidatePath`, `revalidateTag`, `updateTag`) — so storage, comparison,
 * and the wire all see the same canonical form.
```

packages/next/src/server/lib/encode-header-safe.ts
```diff
@@ -30,8 +32,8 @@
 const OUT_OF_CLASS_CHAR = /[^\t\x20-\x7e]/
```
gnoff, line 29: we maybe should consider encoding tabs. if a leading tab is stripped in fetch the upstream cache tag may not match the intended one

gnoff (approved):

packages/next/src/server/lib/encode-header-safe.ts
unstubbable: Will follow up with that, as this is supposed to be a pure refactor.

unstubbable: I looked into this, and having leading tabs in tags actually doesn't cause a mismatch. On a deployment they do get stripped (spaces too, and on every tag in the list, since the suspense cache API trims each one), but nothing ever compares an untrimmed tag against a trimmed one. Next.js is consistent: with a custom cache handler you can see `" spaced-tag "` in both `set` and `updateTags`, and `get`/`getExpiration` only ever receive the implicit `_N_T_/…` tags, which never have whitespace. The deployed side is consistent too, since registering a tag (header) and `revalidateTag` (invalidate query) both get trimmed.

So whitespace just isn't significant in tag identity, `' foo '` and `'foo'` are the same tag. Same class as the encoded/raw collision, and harmless. I'd skip encoding tab since it isn't fixing a reachable bug.

gnoff: yeah fair, trimming makes sense

PR #96937, opened 2026-08-07
Encode the cache item name built by `unstable_cache`

### LINKED ISSUE #76286: `unstable_cache` broken on Vercel when non english searchParams present in URL

https://github.com/cygnetdigital/next-unstable-cache-bug-repro

### To Reproduce

1. Deploy the application to Vercel (or visit the link below).
2. Visit the URL `https://unstable-cache-repro.vercel.app/?key=bar&extra=english`. The `utcTime` will remain static as you refresh the page.
3. Now add an unrelated non-english search param. `https://unstable-cache-repro.vercel.app/?key=bar&extra=中國人`. Notice how the `utcTime` is increasing for every page refresh.

The repro does not consume the `extra` searchParam anywhere in the application, yet when any non english searchParam is in the URL it can break `unstable_cache`. 

Looking at the vercel request logs, it's as if there is no attempt to visit the vercel data cache. Perhaps the `unstable_cache` is bailing and just invoking the callback on every request when this condition occurs.

```bash
Node.js v20.18.0

Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.2.0: Fri Dec  6 18:40:14 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 20.18.0
  npm: 10.8.2
  Yarn: 3.2.3
  pnpm: N/A
Relevant Packages:
  next: 14.2.24 // An outdated version detected (latest is 15.1.7), upgrade is highly recommended!
  eslint-config-next: N/A
  react: 18.3.1
  react-dom: 18.3.1
  typescript: 5.7.3
```

Not sure

Vercel (Deployed)

### Additional context

I have tested the following next.js versions.
- `14.1.4` - bug NOT present
- `14.2.15` - bug present
- `14.2.24` - bug present
- `15.1.7` - bug present

unstubbable:
A cache implementation may serialize cache metadata into HTTP request headers, whose values are limited to Latin-1. `unstable_cache` assembles a cache item name from the request URL and the name of the cached callback, and neither part was encoded. When that name holds a character above U+00FF the conversion throws before the request is dispatched, so the read never reaches the cache and the write that follows it fails the same way. Nothing is stored, nothing is found, and the entry falls back to the origin on every render.

The name is built by `getFetchUrlPrefix`, which reads the pathname and the search parameters out of the request URL. The pathname stays percent-encoded, but `URLSearchParams` returns decoded keys and values, so a non-ASCII query parameter is the reachable case: it applies to any dynamic route that calls `unstable_cache`, whether or not the route reads `searchParams`, and therefore also to a parameter a caller appends. A callback whose name holds such a character is affected too, though a production build usually renames the binding.

This change encodes the assembled name with `encodeHeaderSafe`. That helper only replaces characters outside the class Node accepts in a header value, so the separating spaces and the URL punctuation are preserved and the name keeps its documented shape. Every name that is representable today is returned unchanged, so this is inert for existing entries. The item name is a label: it is not the cache key, which is derived separately from the callback's key parts and arguments, and the Suspense Cache API neither parses nor matches on it.

The test covers the two parts of the name on separate routes so a failure names the part it comes from, and asserts the constraint rather than either input, since which inputs are live depends on the bundler and on minification. A deployment checks it through the real cache handler implementation, where the failure shows up as an entry that is recomputed on every request.

fixes #76286

### DIFF
--- a/packages/next/src/server/web/spec-extension/unstable-cache.ts
+++ b/packages/next/src/server/web/spec-extension/unstable-cache.ts
@@ -145,8 +145,13 @@ export function unstable_cache<T extends Callback>(
       // U+00FF: the search parameters are decoded, and a JavaScript identifier
       // may hold one. The character class leaves the separating spaces and the
       // URL punctuation untouched, so the shape above is preserved.
+      //
+      // `toWellFormed` replaces lone surrogates, which `cb.name` can hold and
+      // which `encodeURIComponent` rejects. The name identifies the call for
+      // debug metrics, so a replacement character is an acceptable trade for
+      // not failing the render.
       const fetchUrl = encodeHeaderSafe(
-        `unstable_cache ${fetchUrlPrefix} ${cb.name ? ` ${cb.name}` : cacheKey}`
+        `unstable_cache ${fetchUrlPrefix} ${cb.name ? ` ${cb.name}` : cacheKey}`.toWellFormed()
       )
       const fetchIdx =
         (workStore ? workStore.nextFetchId : noStoreFetchIdx) ?? 1
--- a/packages/next/src/server/web/spec-extension/unstable-cache.ts
+++ b/packages/next/src/server/web/spec-extension/unstable-cache.ts
@@ -2,6 +2,7 @@ import type { IncrementalCache } from '../../lib/incremental-cache'
 
 import { CACHE_ONE_YEAR_SECONDS } from '../../../lib/constants'
 import { validateRevalidate, validateTags } from '../../lib/patch-fetch'
+import { encodeHeaderSafe } from '../../lib/encode-header-safe'
 import {
   workAsyncStorage,
   type WorkStore,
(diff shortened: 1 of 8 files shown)


anujbolewar (commented):
Encoding the cache item name before it can be serialized into an HTTP header is the right call, and the char-class approach preserves the URL punctuation while escaping unsafe bytes. The main thing to pin down is the failure mode of the encode helper: if the encoded output can exceed the header limit, does it truncate, throw, or pass through unchanged? That should be an explicit, tested contract rather than implicit. Also consider an e2e that actually exercises a non-ASCII function name, since the added route covers the URL-derived portion only.

packages/next/src/server/web/spec-extension/unstable-cache.ts
```diff
@@ -138,7 +139,15 @@ export function unstable_cache<T extends Callback>(
...
       // $urlWithPath,$sortedQueryStringKeys,$hashOfEveryThingElse
-      const fetchUrl = `unstable_cache ${fetchUrlPrefix} ${cb.name ? ` ${cb.name}` : cacheKey}`
+      //
+      // A cache implementation may serialize this name into an HTTP request
+      // header, so it is encoded here. Both parts can carry a character above
+      // U+00FF: the search parameters are decoded, and a JavaScript identifier
+      // may hold one. The character class leaves the separating spaces and the
+      // URL punctuation untouched, so the shape above is preserved.
+      const fetchUrl = encodeHeaderSafe(
+        `unstable_cache ${fetchUrlPrefix} ${cb.name ? ` ${cb.name}` : cacheKey}`
```
gnoff: very edge case but if cb.name had a lone surrogate like `'\uD800'` then the encodeURIComponent would throw in encodeHeaderSafe.

For this diagnostic data we can just massage it to something like `.toWellFormed()` since this isn't a cache key where byte identical matching would be crucial.

gnoff (approved):

packages/next/src/server/web/spec-extension/unstable-cache.ts
unstubbable: done

PR #96941, opened 2026-08-07
[turbopack] Retain fewer stale cache versions and use a TTL

lukesandberg:
In CI we do not retain old versions of the db at all, but in dev or non-CI settings we would retain up to 2 indefinitely.

Given how large caches can grow, this PR reduces it to no more than 1 old version, and then only for a certain amount of time, currently 3 days.  The idea being to support switching back and forth between feature branches and a next.js upgrade branch, 3 days allows this worflow to survive a weekend.

To support this the `CURRENT` file format is modified.  Instead of just storing a sequence number it now stores a small json object containing the `max_sequence_number` and the mtime.  We store the last_write_time in the file to support the fact that we anticipate `cache` directories to be copied around and the native mtime would record the copy time instead of the commit time.

This should reduce dev time disk usage in the common case.


turbopack/crates/turbo-tasks-backend/src/database/db_versioning.rs
```diff
@@ -112,35 +141,36 @@ pub fn handle_db_versioning(
...
+            // Most recently used first, so the ones we keep are a prefix of this list. Entries
+            // whose age can't be determined sort last and are therefore evicted first.
+            old_dbs
+                .sort_by_cached_key(|entry| time_since_last_used(entry).unwrap_or(Duration::MAX));
+
+            let retained = old_dbs
+                .iter()
+                .take(max_other_db_versions)
+                .take_while(|entry| time_since_last_used(entry).is_some_and(|age| age <= ttl))
+                .count();
```
lukesandberg: does this mean we call time_since_last_used on each CURRENT file twice, we should cache this ourselves

turbopack/crates/turbo-tasks-backend/src/database/db_versioning.rs
```diff
@@ -112,35 +141,36 @@ pub fn handle_db_versioning(
...
+
+            for entry in old_dbs.into_iter().skip(retained) {
```
lukesandberg: i think it would be fine to bake in the assumption that we only retain one old version rather than this generic loop

turbopack/crates/turbo-tasks-backend/src/database/db_versioning.rs
```diff
@@ -153,81 +183,313 @@ pub fn handle_db_versioning(
...
+        Ok(secs) => Duration::from_secs(secs),
+        Err(_) => {
+            println!(
+                "WARNING: Ignoring TURBO_ENGINE_VERSION_TTL={raw:?}, expected a whole number of \
+                 seconds."
+            );
+            DEFAULT_OTHER_DB_VERSION_TTL
+        }
+    }
+}
```
lukesandberg: is there a more structured parse we can use, to support things like `3d` or `72h`

turbopack/crates/turbo-tasks-backend/src/database/db_versioning.rs
```diff
@@ -153,81 +183,313 @@ pub fn handle_db_versioning(
...
+    }
+
+    // Fall back to the most recently modified file in the directory. Not recursive: the
+    // persistence layer keeps its SST/meta/blob files at the top level, so this is enough to tell
+    // a live cache from an abandoned one without walking a multi-GB tree on every startup.
+    read_dir(&dir)
+        .ok()?
+        .filter_map(|e| e.ok()?.metadata().ok()?.modified().ok())
+        .max()
+        .map(age_of)
```
lukesandberg: lets drop this fallback

we can rely on the CURRENT file existing, if it doesn't assume some kind of worst case time (Duration::MAX?)

turbopack/crates/turbo-tasks-backend/src/database/db_versioning.rs
```diff
@@ -153,81 +183,313 @@ pub fn handle_db_versioning(
...
+/// Stamps the version directory at `path` as used now, by setting the mtime of its
+/// [`LAST_USED_FILE`] to the current time.
+///
+/// This is a no-op when the file doesn't exist. Creating it here would be actively harmful: the
+/// persistence layer reads `CURRENT` as a big-endian `u32` sequence number, so an empty file we
+/// fabricated would fail that read and make the directory un-openable. A directory in that state is
+/// either brand new (the persistence layer writes `CURRENT` as it initializes, which stamps it for
+/// us moments later) or not a database, and [`time_since_last_used`] has its own fallback for
+/// judging the age of one that lacks the file.
+fn mark_used(path: &Path) -> Result<()> {
```
lukesandberg: touch_current

would be a better name i think

turbopack/crates/turbo-tasks-backend/src/database/db_versioning.rs
```diff
@@ -74,14 +82,31 @@ pub fn handle_db_versioning(
...
+                // Only try to delete the files if the rename succeeded, it's not safe to delete
+                // contents if we didn't manage to first poison the directory by renaming it.
+                if rename_result.is_ok() {
+                    // It's okay if this fails, as we've already poisoned the directory.
+                    let _ = remove_dir_all(&new_path);
+                }
+            };
+
+            // Of the other versions we keep only the most recently used one, and only if it's
+            // within the TTL.
```
mischnic, line 108: Shouldn't we keep the most recent one regardless of TTL?

lukesandberg: we do.  this is only applied to 'other versions'

the current version is always skipped on line 114

mischnic (approved):

PR #96947, opened 2026-08-07
Allow issue participants to request reopening

marcoshernanz:
## Summary

- Let issue authors and people who commented before closure reopen an issue within 14 days using `Reopen: <reason>`.
- Limit the automation to unlocked issues closed by triage-or-higher human maintainers, with ordered timeline checks, bot filtering, idempotency, and state revalidation.
- Align resolved-issue and contributor guidance with the new reopening window.

## Verification

- `Prettier 3.6.2 --check .github/workflows/issue_reopen.yml .github/comments/resolved.md contributing/repository/triaging.md`
- Mocked workflow harness covering 19 closure, eligibility, boundary, bot, idempotency, and race scenarios
- `autoreview --mode local --stream-engine-output` (Codex and Claude clean)
- Live GitHub API spot-check for ordered same-second `commented` and `closed` timeline events
- Not run: live issue close/reopen smoke test (issue-triggered workflows run only from the default branch)


.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+    issue_number: eventIssue.number,
+  }
+
+  function isAutomationUser(user) {
+    const login = user?.login?.toLowerCase()
+
+    return (
+      !login ||
+      user.type === 'Bot' ||
+      login.endsWith('[bot]') ||
```
bgw: The `[bot]` suffix is only ever added by GitHub when `user.type === 'Bot'`, so this is redundant.

The `-bot` suffix check is valid, some of our (legacy?) bots are not properly registered as apps.

```suggestion
```

.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+    owner,
+    repo,
+    issue_number: eventIssue.number,
+  }
+
+  function isAutomationUser(user) {
+    const login = user?.login?.toLowerCase()
+
+    return (
+      !login ||
```
bgw: When would this happen?

.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+        username: user.login,
+      })
+
+      const eligibleRoles = new Set(['triage', 'write', 'maintain', 'admin'])
+
+      return (
+        data.user?.permissions?.triage === true ||
+        eligibleRoles.has(data.role_name) ||
+        eligibleRoles.has(data.permission)
+      )
```
bgw: I think we can simplify and just look at the `permissions` object. Testing out this API with:

```
gh api /repos/vercel/next.js/collaborators/sokra/permission
```

I think think it gives us what we want.

```suggestion
      return !!data.user?.permissions?.triage
```

.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+  async function getIssue() {
+    const { data } = await github.rest.issues.get(issueParams)
+    return data
+  }
+
+  async function getTimeline() {
+    return github.paginate(github.rest.issues.listEventsForTimeline, {
+      ...issueParams,
+      per_page: 100,
+    })
```
bgw: Is it possible to just get the last 100 events? I think paginating through the whole thing isn't needed and could cause rate limit issues in extreme cases.

.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+
+  if (
+    latestIssue.state !== 'closed' ||
+    latestIssue.locked ||
+    latestIssue.closed_at !== issue.closed_at ||
+    latestIssue.closed_by?.id !== issue.closed_by?.id
+  ) {
+    core.info('Skipping a request because the issue state changed')
+    return
+  }
```
bgw: This assuming the state could change while we're checking it? That seems like a tiny edge case, and this cannot fully protect against TOCTOU either, it just narrows it. This doesn't seem worth it.

.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+
+    const timeline = await getTimeline()
+    const closureIndex = findCurrentClosureIndex(timeline, issue)
+
+    if (closureIndex === -1) {
+      core.info('Skipping a closure missing from the issue timeline')
+      return
+    }
+
+    const marker = `<!-- issue-reopen-window:${timeline[closureIndex].id} -->`
```
bgw: Should we also somehow include a marker comment in `resolved.md`? It doesn't make sense to comment twice when the issue is auto-closed by the triage action?

https://github.com/vercel/next.js/blob/15109c16fa83b3492a2ddc539b3bef56fc3d936d/.github/workflows/triage.yml

.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+      ...issueParams,
+      body,
+    })
+
+    core.info(`Posted reopening instructions on issue #${issue.number}`)
+    return
+  }
+
+  const comment = context.payload.comment
+  const command = comment?.body?.match(/^\s*reopen\s*:\s*(\S[\s\S]*)$/i)
```
bgw:
```suggestion
  const command = comment?.body?.match(/\breopen\s*:\s*\S.*$/si)
```

https://regex101.com/?regex=%5Cbreopen%5Cs*%3A%5Cs*%28%5CS.*%29%24&testString=&flags=is&flavor=javascript&delimiter=%2F

- It's fine if "reopen:" isn't on it's own line, and `\b` is the more ergonomic way of matching word boundaries.
- `[\s\S]` means: "whitespace character or non-whitespace character" which is insane. Just use `.` with the `s` flag
- We don't need to form the group around the reason, because we never read it.

.github/scripts/issue_reopen.js
```diff
@@ -0,0 +1,217 @@
...
+
+    const marker = `<!-- issue-reopen-window:${timeline[closureIndex].id} -->`
+    const markerSearchSince = new Date(
+      Date.parse(issue.closed_at) - 1000
+    ).toISOString()
+    const comments = await github.paginate(github.rest.issues.listComments, {
+      ...issueParams,
+      since: markerSearchSince,
+      per_page: 100,
+    })
```
bgw: Just read the last 100 comments, that's probably sufficient? Don't need to paginate the whole thing.

.github/workflows/issue_reopen.yml
```diff
@@ -0,0 +1,46 @@
...
+          persist-credentials: false
+          sparse-checkout: .github/scripts/issue_reopen.js
+          sparse-checkout-cone-mode: false
+
+      - name: Handle issue closure or reopening request
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const { handleIssueReopening } = require('./.github/scripts/issue_reopen.js')
+            await handleIssueReopening({ core, github, context })
```
bgw: You don't need to use `actions/github-script`, just define a custom action within the repository: https://docs.github.com/en/actions/tutorials/create-actions/create-a-javascript-action#example-using-a-private-action

There are plenty of examples of how to do this here: https://github.com/vercel/next.js/tree/canary/.github/actions

.github/workflows/issue_reopen.yml
```diff
@@ -0,0 +1,46 @@
...
+      contents: read
+      issues: write
+    steps:
+      - name: Checkout trusted workflow code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: ${{ github.sha }}
+          persist-credentials: false
+          sparse-checkout: .github/scripts/issue_reopen.js
+          sparse-checkout-cone-mode: false
```
bgw: You shouldn't need to do this if you define a custom action within the repository instead of using `actions/github-script`. If you define a custom action within the repository, github will automatically download the code's tarball before running it.

bgw (changes requested):

marcoshernanz:
@bgw ): stnemmoc lla desserdda

.github/workflows/issue_reopen.yml
```diff
@@ -0,0 +1,35 @@
...
+          github.event.issue.state == 'closed' &&
+          github.event.comment.user.type == 'User'
+        )
+      )
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Handle issue closure or reopening request
+        uses: vercel/next.js/.github/actions/issue-reopen@canary
```
bgw: This can be a relative path, let's write this as a relative path. We're going to start enforcing pinning of GH actions soonish, and this wouldn't work with that repository setting.

bgw (approved):
Keep a close eye on this after merging, and make sure it behaves as expected! We can disable the job if it misbehaves.

Issue #74343, opened 2024-12-28
Props must be serializable for components in "use client" file (for discussion #46795)


inducingchaos:
### Link to the code that reproduces this issue

https://github.com/inducingchaos/riley-barabash

### To Reproduce

**[IGNORE CODE REPRO - SEE BOTTOM]**

In my experience: create almost any component in a "use client" file that accepts a function, dispatch for state, or anything that is non-serializable as props.

### Current vs. Expected behavior

Current: Typescript (Next.js version) will give a warning (the title) to either a) make props serializable, or b) to postfix the prop name with 'Action'. This has led to an unnecessary amount of renamings simply to avoid this error - where the majority of cases **don't have anything to do with Server Actions at all.**

Expected: No errors when passing props unrelated to SA or the client-server boundary, especially in cases where client components are layered.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: x64
  Version: Darwin Kernel Version 24.0.0: Mon Aug 12 20:54:30 PDT 2024; root:xnu-11215.1.10~2/RELEASE_X86_64
  Available memory (MB): 32768
  Available CPU cores: 16
Binaries:
  Node: 22.9.0
  npm: 10.8.3
  Yarn: N/A
  pnpm: 9.12.0
Relevant Packages:
  next: 15.1.1-canary.16 // There is a newer canary version (15.1.1-canary.22) available, please upgrade! 
  eslint-config-next: 14.2.11
  react: 19.0.0-rc-603e6108-20241029
  react-dom: 19.0.0-rc-603e6108-20241029
  typescript: 5.6.2
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Developer Experience, Linting, Partial Prerendering (PPR)

### Which stage(s) are affected? (Select all that apply)

next dev (local)

### Additional context

Didn't provide a repro because I'm lazy and I think there's enough context within the discussion thread I linked.

See: https://github.com/vercel/next.js/discussions/46795

pawelblaszczyk5:
“use client” should be used only for marking the boundaries between client-server code - similarly to “use server”. Not every client component should be marked with it. Boundaries have additional restrictions, e.g. serializable props you mentioned

inducingchaos:
That's easy to say, and has been said - but it doesn't make a lot of sense in practice (or in this context), nor is it a viable solution to this problem.

[...]

PR #96949, opened 2026-08-07
Allow suppressing TypeScript plugin diagnostics

gnoff:
There are cases where a function prop in a "use client" file is not named Action but it could be reasonable to quiet the linter that complains. See https://github.com/vercel/next.js/issues/74343 for historical context and feedback.

The simplest thing to do is allow these linting rules to be ignored however since the they are implemented in a custom language server they don't automatically respect these suppressions. This PR adds support for the standard ts-expect|ignore directives.

### DIFF
--- a/test/development/typescript-plugin/client-boundary/app/suppressed-props.tsx
+++ b/test/development/typescript-plugin/client-boundary/app/suppressed-props.tsx
@@ -1,0 +1,13 @@
+'use client'
+
+export default function ClientComponent(props: {
+  // @ts-ignore -- This non-serializable prop is intentionally exposed.
+  _ignoredFunction: () => void
+  // @ts-expect-error -- This non-serializable prop is intentionally exposed.
+  _expectedFunction: () => void
+  // @ts-expect-error -- This directive should remain unused.
+  _serializable: string
+  _unsuppressedFunction: () => void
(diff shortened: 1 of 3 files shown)


test/development/typescript-plugin/client-boundary/app/suppressed-props.tsx
```diff
@@ -0,0 +1,13 @@
+'use client'
+
+export default function ClientComponent(props: {
+  // @ts-ignore -- This non-serializable prop is intentionally exposed.
+  _ignoredFunction: () => void
+  // @ts-expect-error -- This non-serializable prop is intentionally exposed.
```
eps1lon, line 6: I don't think people will be able to use this since any `tsc` or standard TS langugage server will fail here about unused `@ts-expect-error`. 

And then there's the the general argument against overwriting language semantics.  Ecosystem interop will be a pain: any existing linters will be confused by these directives because they don't know this is a Next.js LSP specific suppression not a standard TS suppression.

gnoff: Aside from making the lint smarter is there a way to support disabling our custom lints by line and overall that is ergonomic?

PR #96955, opened 2026-08-08
Turbopack: Remove codepath for writing non-directory symbolic links

bgw:
Previously, we had to `stat()` the target of a link during `read_link` to determine if it was a file or a directory to store that in `LinkType`.

This was pretty silly, because we only actually care about this in the write path: Directories on Windows can be written using junction points, which is preferred.

First, I tried splitting the data structures used for the read and write codepaths, so that we'd only pass this in for the write codepath, but then I realized that we never actually want to write file symlinks, and doing so would be a portability hazard for Windows, so this PR just removes support for that instead.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/content.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/content.rs
@@ -10,7 +10,6 @@ use std::{
 
 use anyhow::{Result, bail};
 use bincode::{Decode, Encode};
-use bitflags::bitflags;
 use jsonc_parser::{ParseOptions, parse_to_serde_value};
 use mime::Mime;
 use serde_json::Value;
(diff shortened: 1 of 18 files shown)


turbopack/crates/turbo-tasks-fs/src/content.rs
```diff
@@ -163,43 +162,36 @@ pub(crate) enum FileComparison {
...
+    /// A valid symbolic link target, stored as a unix-style path.
     ///
-    /// If [`LinkType::ABSOLUTE`] is set, `target` is normalized and relative to the *filesystem
-    /// root* (so that absolute system paths never end up in the persistent cache). Otherwise,
-    /// `target` is the raw value read from the link — unnormalized, may contain `..` — and is
-    /// relative to the *directory containing the link*.
+    /// [`LinkTarget::Absolute`] targets are normalized and relative to the *filesystem root* (so
+    /// that absolute system paths never end up in the persistent cache).
+    /// [`LinkTarget::Relative`] targets are raw values read from the link — unnormalized, may
+    /// contain `..` — and are relative to the *directory containing the link*.
```
lukesandberg: nit: this restates `LinkTarget` docs, maybe amend those instead

turbopack/crates/turbo-tasks-fs/src/lib.rs
```diff
@@ -75,9 +78,8 @@ pub trait FileSystem: ValueToString {
     fn read(self: Vc<Self>, fs_path: FileSystemPath) -> Vc<FileContent>;
     /// Reads the target of a symbolic link (or of a junction point on Windows).
     ///
-    /// The base of the returned [`LinkContent::Link`] `target` depends on the link's
-    /// [`LinkType`]: root-relative and normalized for [`LinkType::ABSOLUTE`] links, or the raw
-    /// link-relative on-disk value otherwise.
+    /// The returned [`LinkTarget`] is root-relative and normalized for [`LinkTarget::Absolute`]
+    /// links, or the raw link-relative on-disk value for [`LinkTarget::Relative`] links.
```
lukesandberg: just restating the enum docs?, simplify?

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -681,6 +671,18 @@ async fn realpath_with_links(path: FileSystemPath) -> Result<Vc<RealPathResult>>
     .cell())
 }
 
+/// Resolves a link target using the storage convention described by [`LinkContent::Link`].
+pub async fn resolve_link_target(
+    link_path: &FileSystemPath,
+    target: &LinkTarget,
+) -> Result<FileSystemPath> {
+    let (base, target) = match target {
+        LinkTarget::Absolute(target) => (link_path.root().owned().await?, target),
```
lukesandberg: nit: i think this can be non-async with `FileSystemPath::new_normalized_unchecked(link_path.fs, target)`

lukesandberg (approved):

bgw:
It turns out that while we never create a new non-directory symlink directly, this isn't viable because node-file-trace *does* need to be able to write symlinks pointing to a file if we copy over a dependency that uses a symlink.

PR #96956, opened 2026-08-08
Remove unused `hasWarnings` and `warnings`

martinfrancois:
### Fixing a bug

Nothing reads the two fields `hasWarnings` and `warnings` on `TypeCheckResult`.

`verifyAndRunTypeScript` has four callers:

- `build/type-check.ts` only reads `inputFilesCount`, `totalFilesCount` and `incremental`
- `next-test` only reads `version`
- `setup-dev-bundler` only reads `version`
- `next-typegen` throws the result away

`warnings` is also always empty. It looks for diagnostics with the `Warning` category, but I checked, and TypeScript 6.0.2 has no messages in that category. Deprecation messages use `Suggestion` instead.

So on every build, we filtered and formatted every diagnostic to build a list that was always empty, and that nobody read.

So I'm suggesting removing it instead, as it is clearly not being used. I also checked the GitHub code search across all public repositories and couldn't find anyone using either of them. That aside, it also never really worked correctly, as it would always return `hasWarnings: true` even if `warnings` was empty.

No test added since the fields are being removed. Also since this is such a small change I didn't create an issue to link here, I'd be happy to file one if you prefer me to.

Thanks for the hard work and taking the time for reviewing this PR! 😊

### DIFF
--- a/packages/next/src/lib/typescript/runTypeCheck.ts
+++ b/packages/next/src/lib/typescript/runTypeCheck.ts
@@ -8,8 +8,6 @@ import { CompileError } from '../compile-error'
 import { warn } from '../../build/output/log'
 
 export interface TypeCheckResult {
-  hasWarnings: boolean
-  warnings?: string[]
   inputFilesCount?: number
   totalFilesCount?: number
   incremental: boolean
@@ -108,7 +106,6 @@ export async function runTypeCheck(
 
   if (fileNames.length < 1) {
     return {
-      hasWarnings: false,
       inputFilesCount: 0,
       totalFilesCount: 0,
       incremental: false,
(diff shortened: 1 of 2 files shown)


martinfrancois:
I reran validation at `afe7c8ecd8f2` on September 19. The workspace build, `build-all`, root TypeScript check, and formatting and lint checks for both changed files pass.

All 14 tests across the basic TypeScript build, CLI backend, and API error-handling suites pass in Webpack production mode. The removed fields are still present in current canary, and both affected source files match this PR's base there.

@lukesandberg, can you please rerun CI and review this cleanup of the TypeScript API and CLI results? The old failed run no longer exposes job details, so the local checks are the current evidence.

lukesandberg (approved):

lukesandberg:
@martinfrancois  merging is blocked due to missing commit signatures. can you sign the commits

martinfrancois:
Sure @lukesandberg done, thanks for the quick review!

PR #96964, opened 2026-08-08
docs: add `export const dynamic = 'force-static'` to Route Handlers example on Static Exports

m-kawafuji:
What?
When a static export is enabled, route handlers require export const dynamic = 'force-static'. However, this instruction is missing in the documentation.

Why?
Without export const dynamic = 'force-static', next dev and next build fail when a static export is enabled.

How?
This PR adds the instruction to correct the usage of Route Handlers on static exports.

### DIFF
--- a/docs/01-app/02-guides/static-exports.mdx
+++ b/docs/01-app/02-guides/static-exports.mdx
@@ -232,15 +232,19 @@ export default function Page() {
 
 ### Route Handlers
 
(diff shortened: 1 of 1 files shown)


timneutkens (approved):

timneutkens:
Thank you!

sirgagahandino9-design (commented):
Go

PR #96977, opened 2026-08-08
Preserve the original host on action redirect streams

DavidIlie:
### What

-   Follow-up to #96552.
-   #96552 fixes the first internal self-fetch when a Server Action is
    forwarded between workers.
-   This PR fixes the second internal self-fetch in
    `createRedirectRenderResult`, used to stream app-relative Server
    Action redirects.
-   Adds a distinct `x-action-redirect-forwarded: 1` marker.
-   Restores the original `Host` from `x-forwarded-host` only for the
    exact internal GET + RSC request shape.
-   Tests both direct action redirects and forwarded actions that
    subsequently redirect.

### Why

-   The redirect self-fetch targets Next.js's private/internal origin.
-   Its `Host` becomes something like `localhost:PORT`.
-   Consequently, `headers().get('host')` in the redirect target sees
    the internal host instead of the public tenant host.
-   This can break host-based routing or multi-tenancy.
-   A separate marker is needed because this redirect request does not
    naturally set `x-action-forwarded`.

### Security details

-   The marker alone is not trusted.
-   Restoration requires:
    -   exact marker value `1`;
    -   GET request;
    -   RSC request header;
    -   current `Host` matching the internal forwarding origin;
    -   a valid configured forwarding origin.
-   Negative tests cover forged markers and incorrect request shapes.

### Verification

-   `pnpm build-all`
-   Unit test: 34/34 passed.
-   Action-forward-host e2e: 6/6 passed in:
    -   dev Turbopack
    -   dev webpack
    -   production Turbopack
    -   production webpack
-   Action-forward-loop regression passed.
-   Disallowed-origin security regression passed.
-   Both relative and absolute streamed basePath action redirects
    passed.
-   Next.js types, Prettier, and ESLint passed.


packages/next/src/server/app-render/action-forwarding.test.ts
eps1lon, line 1: These should be e2e tests. If these bypass some outer security layer, it'll just trip up future audits. If they don't bypass any, there's no need to have them as unit tests committed to the repo.

squaredd:
Reporter of #96344 here — thanks for extending this to the redirect stream. I'd only found the action-forwarding self-fetch; `createRedirectRenderResult` going to the private origin as well is the same bug in a second place, and I'd have missed it.

I left the production evidence on #96552 rather than duplicating it here: https://github.com/vercel/next.js/pull/96552#issuecomment-5259682087. Short version — self-hosted `output: standalone`, 9 domains on one server resolving tenant config from `headers().get('host')`, ~1,400 failed actions/day, and the failure is **completely silent**: the non-RSC branch of `createForwardedActionResponse` cancels the body without logging and returns `RenderResult.fromStatic('{}')`, so the client gets **HTTP 200** and nothing shows up in error rates, status codes or RUM. We only found it by grepping pod logs.

Two notes specific to this PR:

- I can confirm the **first** self-fetch from production, but not the second: none of our server actions call `redirect()`, so `createRedirectRenderResult` never runs for us and I have no field data on it. The mechanism is clearly the same one though, and a redirect target reading `headers().get('host')` would break identically in our setup.
- For whoever triages this: this PR and #96552 overlap — as far as I can tell this one contains that fix plus the redirect case. From a self-hosted user's point of view either is fine; what's missing is a maintainer looking at one of them. Neither has a review yet, and it would be a shame if two PRs for the same bug ended up cancelling each other out.

Same offer I made on the other PR: happy to run a canary build against production traffic. We know exactly which actions fail, so we can give a real before/after rather than a synthetic case.

DavidIlie:
> Reporter of #96344 here — thanks for extending this to the redirect stream. I'd only found the action-forwarding self-fetch; `createRedirectRenderResult` going to the private origin as well is the same bug in a second place, and I'd have missed it.
> 
> I left the production evidence on #96552 rather than duplicating it here: [#96552 (comment)](https://github.com/vercel/next.js/pull/96552#issuecomment-5259682087). Short version — self-hosted `output: standalone`, 9 domains on one server resolving tenant config from `headers().get('host')`, ~1,400 failed actions/day, and the failure is **completely silent**: the non-RSC branch of `createForwardedActionResponse` cancels the body without logging and returns `RenderResult.fromStatic('{}')`, so the client gets **HTTP 200** and nothing shows up in error rates, status codes or RUM. We only found it by grepping pod logs.
> 
> Two notes specific to this PR:
> 
>     * I can confirm the **first** self-fetch from production, but not the second: none of our server actions call `redirect()`, so `createRedirectRenderResult` never runs for us and I have no field data on it. The mechanism is clearly the same one though, and a redirect target reading `headers().get('host')` would break identically in our setup.
> 
>     * For whoever triages this: this PR and [Preserve the original host on forwarded Server Action requests #96552](https://github.com/vercel/next.js/pull/96552) overlap — as far as I can tell this one contains that fix plus the redirect case. From a self-hosted user's point of view either is fine; what's missing is a maintainer looking at one of them. Neither has a review yet, and it would be a shame if two PRs for the same bug ended up cancelling each other out.
> 
> 
> Same offer I made on the other PR: happy to run a canary build against production traffic. We know exactly which actions fail, so we can give a real before/after rather than a synthetic case.

I added a focused e2e for that silent branch and a bounded error log containing
the worker status and content type, without logging its body or changing the
existing client response behavior. I put that commit on #96552 because it
belongs to action-worker forwarding; #96977 remains the redirect-streaming
follow-up on top.

One small implementation detail from the local probe: current Node accepts a
`Host` value in the `Headers` object, but fetch dispatch still derives the wire
`Host` from the private URL. So the observed mechanism and fix are the same,
even though describing it as `Headers` rejecting `host` is a little too broad.

PR #96991, opened 2026-08-09
Turbopack: accept a module type argument for import.meta.glob

kelvinampofo:
### What?

Adds support for passing a module type to the TypeScript overloads for `import.meta.glob`:

```ts
const entries = import.meta.glob<Entry>("./entries/*.ts");
```

This works with both lazy and eager imports. If a module type isn’t provided, it defaults to `unknown`, which matches the current behavior.

### Why?

Next.js already supports Vite’s `import.meta.glob` API at runtime, but its TypeScript declarations don’t support [Vite’s module generic](https://github.com/vitejs/vite/blob/57fea001d154e7dd8d5d74d3082731f1dcfd31be/packages/vite/types/importGlob.d.ts#L113-L130).

Because of this, a typed glob works in Vite but fails in Next.js:

```text
TS2558: Expected 0 type arguments, but got 1.
```

Consumers currently have to cast the result or add their own `ImportMeta` declaration to describe the matched modules. Supporting the generic removes the need for those workarounds and makes typed glob imports consistent between Next.js and Vite.

### How?

Each existing overload now accepts a generic module type, `M`, which defaults to `unknown`:

- Lazy imports return `Record<string, () => Promise<M>>`.
- Eager imports return `Record<string, M>`.

Using `unknown` as the default means existing calls without a generic keep the same types as before.

I also updated the existing e2e fixture. It had an old local `ImportMeta.glob` declaration that returned `Record<string, any>`. Removing it revealed the missing generic and allowed me to remove the `as any` casts from the fixture.

The Turbopack documentation has also been updated with examples showing how to use the generic with lazy and eager imports.

### Checklist

- [x] `pnpm --filter=next types` passes
- [x] `pnpm prettier-fix` run
- [x] Documentation follows the docs contribution guide
- [x] E2E test passes


docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -278,6 +278,21 @@ const lazy = import.meta.glob('./dir/*.ts')
 const eager = import.meta.glob('./dir/*.ts', { eager: true })
 ```
 
+Pass the module type to describe what the matched files export:
+
+```ts
```
kelvinampofo, line 283: The [docs guide](https://nextjs.org/docs/community/contribution-guide#language-and-filename) asks for a filename on code block headers. I left it off though, to match the rest of this section, where the blocks show return types rather than a file you would copy into a project.

packages/next/types/global.d.ts
```diff
@@ -130,19 +130,22 @@ interface ImportMeta {
    *
    * // Eager — values are module objects
    * const modules = import.meta.glob('./dir/*.js', { eager: true })
+   *
+   * // The module type can be provided, as in Vite
```
mischnic:
```suggestion
   * // The module type can be provided
```

mischnic (approved):

PR #96994, opened 2026-08-09
render global error for failed cache regeneration

### LINKED ISSUE #96567: global-error.tsx is not rendered when an error occurs during regeneration of a "use cache" route in production mode

https://github.com/amjedomar/next-cache-and-global-error-handler-bug

### To Reproduce

1. Clone the reproduction repository:

```bash
git clone https://github.com/amjedomar/next-cache-and-global-error-handler-bug.git
cd next-cache-and-global-error-handler-bug
```

2. Install the dependencies:

```bash
npm install
```

3. Build and start the application **in production mode**:

```bash
npm run build
npm start
```

4. Open http://localhost:3000/cached. This route uses **`"use cache"`**. The page initially loads successfully because it was generated during the build.

5. Invalidate the cached page by sending a POST request:

```bash
curl -X POST http://localhost:3000/api/invalidate
```

6. Refresh http://localhost:3000/cached.

7. During regeneration, the page code for this route that uses **`"use cache"`** intentionally throws an error, **but `global-error.tsx` is not rendered. Instead, Next.js returns the plain-text response** `Internal Server Error`. ❌

8. Open http://localhost:3000/dynamic for comparison. It throws the same error, but `global-error.tsx` is rendered correctly. ✅

### Current behavior

After building and starting the application **in production mode**, if an error is thrown while regenerating a route that uses **`"use cache"`**, `global-error.tsx` is not rendered. Instead, Next.js returns the plain-text response: `Internal Server Error`.

Image

In contrast, when the same error is thrown from a **dynamic route** in production mode, it is handled correctly by `global-error.tsx` example:

Image

### Expected behavior

After building and starting the application in **production mode**, if an error is thrown while regenerating a route that uses **`"use cache"`**, Next.js should render `global-error.tsx`, just as it does for a **dynamic route**.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Tue Jun  9 22:26:15 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T6031
  Available memory (MB): 36864
  Available CPU cores: 14
Binaries:
  Node: 24.18.1
  npm: 11.16.0
  Yarn: 1.22.22
  pnpm: 10.33.2
Relevant Packages:
  next: 16.3.0-canary.107 // Latest available version is detected (16.3.0-canary.107).
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 5.9.2
Next.js Config:
  output: N/A
```

cacheComponents, Error Handling, Use Cache

next start (local)

### Additional context

You can reproduce the issue by following the instructions in the reproduction repository README: https://github.com/amjedomar/next-cache-and-global-error-handler-bug

Swpn0neel:
## summary

so we were using `use cache` for a static app router, and nextjs was storing the generated page to serve later reqs quickly. if that cache was invalidated, the next request was regenerating the page.

currently, for an uncaught rendering during the regeneration, the error escapes the app rendering system and returns a plain-text internal server error. since the response is no longer a regular nextjs html document, the browser cannot load and display the app’s `global-error.tsx` component.

so, my change detects that rendering failure and renders the current request through an existing dynamic error handling path that gives react the chance to prepare an error response and letting the browser display the app's custom error screen.

## why

a failed page should never be written to the router cache, otherwise nextjs will accidentally serve the error screen as the real cached page to later visitors.

the existing static regeneration process already prevents the failed page from being cached. this change keeps that behavior by only creating a dynamic error response for the visitor whose request triggered the failed regeneration.

the response still uses a 500 status because the page failed to render and it is also marked as a cache miss and receives private no-cache headers, so the error response cannot replace the valid cached page.

and also, the recovery is limited to digested react rendering errors from static routes with a cache key, and it does not change `nofallback` behavior, cache infrastructure errors, or responses that have already started sending data.

## testing

i have added a new production regression test that verifies whether:

- the page is generated successfully during the production build

- its use cache entry can be invalidated

- failed regeneration returns a 500 html response instead of plain text

- the response is marked as a cache miss and cannot be cached

- the browser hydrates and displays the custom global-error.tsx component

and this focused production test passes with both turbopack and webpack.

also, the next.js package type check, focused eslint check, prettier check, and git diff --check also pass.

fixes #96567

### DIFF
--- a/packages/next/src/build/templates/app-page-runtime.ts
+++ b/packages/next/src/build/templates/app-page-runtime.ts
@@ -74,7 +74,10 @@ import {
   NEXT_RESUME_HEADER,
   NEXT_RESUME_STATE_LENGTH_HEADER,
 } from '../../lib/constants' with { 'turbopack-transition': 'next-server-utility' }
-import type { CacheControl } from '../../server/lib/cache-control'
+import {
+  getCacheControlHeader,
+  type CacheControl,
+} from '../../server/lib/cache-control' with { 'turbopack-transition': 'next-server-utility' }
 import { ENCODED_TAGS } from '../../server/stream-utils/encoded-tags' with { 'turbopack-transition': 'next-server-utility' }
 import { sendRenderResult } from '../../server/send-payload' with { 'turbopack-transition': 'next-server-utility' }
 import { NoFallbackError } from '../../shared/lib/no-fallback-error.external' with { 'turbopack-transition': 'next-server-utility' }
@@ -827,9 +830,11 @@ export function createAppPageEntrypoint({
         fallbackRouteParams,
         renderOperation,
         allowEmptyStaticShell,
+        forceDynamicResponse,
       }: {
         span?: Span
         allowEmptyStaticShell?: boolean
+        forceDynamicResponse?: boolean
 
         /**
          * The postponed data for this render. This is only provided when resuming
@@ -872,8 +877,9 @@ export function createAppPageEntrypoint({
             allowEmptyStaticShell,
             serveStreamingMetadata,
             supportsDynamicResponse:
-              renderOperation === 'render' &&
-              (typeof postponed === 'string' || supportsDynamicResponse),
+              forceDynamicResponse ||
+              (renderOperation === 'render' &&
+                (typeof postponed === 'string' || supportsDynamicResponse)),
             buildManifest,
             nextFontManifest,
             reactLoadableManifest,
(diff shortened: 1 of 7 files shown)


Swpn0neel:
@ztanner @eps1lon, could you please review this when you have time? it changes the app-page runtime around static cache regeneration and error handling. the focused production regression passes with both turbopack and webpack. thank you.

PR #96999, opened 2026-08-09
Fix build crash when outputFileTracingIncludes matches a symlink

sokra:
## What?

A glob in `outputFileTracingIncludes` that wholly matches a symlink pointing at a directory made Turbopack builds fail with a fatal internal error instead of completing:

    Error [TurbopackInternalError]: reading file ".../.claude/skills/ink"
    Caused by: Is a directory (os error 21)

Included symlinks are now hashed by their link content (target + link type) instead of the content of whatever they point at, and they stay in the trace as the symlink itself rather than being resolved or expanded. Include entries that have no content to hash (a path that disappeared, a symlink escaping the project root, an unreadable file) are now reported as a build issue and skipped instead of aborting the whole build.

## Why?

`get_glob_includes` already returned symlinks (including ones pointing at a directory) as trace entries, but `NftJsonAsset` hashed them by reading through the link, which fails outright on a directory target. Reading through the link was also the wrong operation in general: `copyTracedFiles` recreates a traced symlink as a symlink (`fs.readlink` + `fs.symlink`), never by copying the resolved file's bytes, so the link's own content is what should be hashed.

## How?

- Added `LinkContent::hash` in `turbo-tasks-fs`, mirroring the existing `FileContent::hash`.
- Added `hash_traced_include` in `next-api`, dispatching by entry type: file → content hash, symlink → link hash, directory → skip (existing behavior, unchanged), anything else/unreadable → emit a new `UnhashableTracedFileIssue` warning and skip.
- Added a Rust unit test covering a plain file, a symlink to a file, a symlink to a directory (the regression), a dangling symlink, a symlink escaping the filesystem root, a directory, and a missing path.
- Extended the `build-trace-extra-entries-turbo` e2e fixture with a symlink-to-directory match and asserted it shows up in the `.nft.json`.

## Test plan

- `cargo test -p next-api` / `-p turbo-tasks-fs`
- `cargo clippy --all-targets` / `cargo fmt --check` on the touched crates
- `test/production/build-trace-extra-entries-turbo` e2e, both Turbopack and webpack modes
-


r34son:
@sokra Appreciate your work! Faced exactly same issue after upgrade to 16.3.0 and migrate to turbopack. Waiting for the fix!

PR #97003, opened 2026-08-10
[CI] Native builder downloads should retry and fail their own layer

gaojude:
The `v16.3.1-canary.10` publish failed because a connection reset while fetching the rustup installer produced a successful layer with no Rust toolchain.

`scripts/native-builder.Dockerfile:96-99`:

```dockerfile
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | \
  sh -s -- -y --default-toolchain "$TOOLCHAIN" --profile minimal && \
rm /tmp/rust-toolchain.toml
```

From run 31341780479:

```
#19 0.258 curl: (35) OpenSSL SSL_connect: Connection reset by peer in connection to sh.rustup.rs:443
#19 DONE 0.3s                  # pipeline status is sh's, and sh on empty stdin exits 0
#20 0.060 /bin/sh: 1: rustup: not found
#20 ERROR: process "/bin/sh -c rustup target add ..." did not complete successfully: exit code: 127
```

`16.3.1-canary.10` never reached npm and `dist-tags.canary` is still `canary.9`, which then failed the same commit's `build-and-test`: create-next-app pins the generated project to `pkg.version` (`packages/create-next-app/templates/index.ts:240`), so npm returned `ETARGET` for `@next/env@16.3.1-canary.10` across the turbopack shards and `test integration windows`.

The installer now downloads to a file and runs as a separate command, with `rustup --version` asserting it installed. The node tarball and cargo-binstall downloads get the same retry loop, neither of which retried before. `--retry-all-errors` needs curl 7.71 and Ubuntu 20.04 ships 7.68, so the retry is a shell loop rather than a curl flag.

Verified by extracting each fetching `RUN` body and executing it under `/bin/sh` with a stub `curl`, reporting exit status and number of `curl` invocations:

```
                  download always fails      transient, succeeds on 3rd try
                  before        after        before        after
rustup installer  exit 0  (1)   exit 1  (5)  exit 0  (1)   exit 0  (3)
node tarball      exit 35 (1)   exit 1  (5)  exit 35 (1)   exit 0  (3)
cargo-binstall    exit 0  (1)   exit 1  (5)  exit 0  (1)   exit 0  (3)
```

The rustup row is the reported bug: a failed download left the layer green. Before, a transient failure was fatal everywhere because nothing retried. The cargo-binstall pipe discards `curl`'s status the same way, though whether `tar` independently rejects the empty input was not confirmed against GNU tar.

### DIFF
--- a/scripts/native-builder.Dockerfile
+++ b/scripts/native-builder.Dockerfile
@@ -45,6 +45,13 @@ RUN HOST_ARCH=$(dpkg --print-architecture) && \
       "deb [arch=${FOREIGN_ARCH}] ${FOREIGN_MIRROR} focal-security main universe" \
       > /etc/apt/sources.list
   
+# Every download below retries with a backoff. Fetches from CI runners fail
+# intermittently at the connection level (the canary.10 build got "OpenSSL
+# SSL_connect: Connection reset by peer" from sh.rustup.rs), and a single lost
+# download costs a whole canary release. curl's own --retry does not cover
+# connection-level errors, and --retry-all-errors needs curl 7.71 while Ubuntu
+# 20.04 ships 7.68, so the retry is an explicit loop.
+
 # Core build tools + GNU cross-compilation sysroots.
 # crossbuild-essential installs headers + libs in the multiarch layout
 # that clang finds via --target. Both archs installed so the image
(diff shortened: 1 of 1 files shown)


scripts/native-builder.Dockerfile
```diff
@@ -45,6 +45,13 @@ RUN HOST_ARCH=$(dpkg --print-architecture) && \
       "deb [arch=${FOREIGN_ARCH}] ${FOREIGN_MIRROR} focal-security main universe" \
       > /etc/apt/sources.list
   
+# Every download below retries with a backoff. Fetches from CI runners fail
+# intermittently at the connection level (the canary.10 build got "OpenSSL
+# SSL_connect: Connection reset by peer" from sh.rustup.rs), and a single lost
+# download costs a whole canary release. curl's own --retry does not cover
+# connection-level errors, and --retry-all-errors needs curl 7.71 while Ubuntu
+# 20.04 ships 7.68, so the retry is an explicit loop.
```
unstubbable, line 53: Should we maybe update the image instead and use that option?

eps1lon: Should be a follow-up since that has a bigger blast radius. But yeah, we should upgrade to 22 or even 24. All our GH hosted runners use ubuntu-latest which is at least 24

bgw: We can only do this prior to Next 17. We build releases with an intentionally old distro to ensure a specific glibc version.

PR #97037, opened 2026-08-10
Prefix `'use cache'` debug logs with the full directive

eps1lon:
With `NEXT_PRIVATE_DEBUG_CACHE` enabled, every log line from the "use cache" wrapper used the same generic `use-cache:` prefix, so output from `"use cache: remote"` or `"use cache: private"` functions was indistinguishable from plain `"use cache"`. Each wrapper line is now prefixed with the full quoted directive, derived from the handler kind: `'use cache'` for the default kind, `'use cache: <kind>'` otherwise.

### DIFF
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -339,8 +339,10 @@ const crossRequestPendingCacheInvocations = new Map<
 
 const isEdgeRuntime = process.env.NEXT_RUNTIME === 'edge'
 
+// The first argument at each call site is the full directive that produced
+// the invocation, e.g. "'use cache'" or "'use cache: remote'".
 const debug = process.env.NEXT_PRIVATE_DEBUG_CACHE
-  ? console.debug.bind(console, 'use-cache:')
+  ? console.debug.bind(console)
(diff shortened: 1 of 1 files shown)


packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -1719,6 +1729,9 @@ export async function cache(
 ) {
   const isPrivate = kind === 'private'
 
+  // The default handler kind corresponds to the bare 'use cache' directive.
+  const logPrefix = kind === 'default' ? `'use cache'` : `'use cache: ${kind}'`
```
unstubbable, line 1733: Did you intentionally drop the trailing `:` that previously had?

unstubbable: I guess the surrounding quotes are also fine.

unstubbable (approved):

packages/next/src/server/use-cache/use-cache-wrapper.ts
eps1lon: Yeah to avoid confusion when `use cache` is used. Looked like the cache kind was accidentally omitted.

PR #60186, opened 2024-01-03
log a dev warning when a missing parallel slot results in a 404

ztanner:
### What & Why?
When visiting a route that attempts to render a slot with no page & no default, the fallback behavior is to trigger a 404. However this can lead to a confusing development experience for complex parallel routing cases as you might not realize a default is missing, or which slot is causing the error.

Previous issues where this caused confusion:
- https://github.com/vercel/next.js/issues/51805
- https://github.com/vercel/next.js/issues/49569

### How?
This is a dev-only modification to track parallel slots that are using the default fallback (aka missing). When the `NotFoundBoundary` is triggered in development mode, this will log a warning about why it 404ed, along with a list of slot(s) that were determined to be missing.

CleanShot 2024-01-03 at 14 34 30@2x

### Future
We should eventually lift this into some sort of dev-only UI to help catch it when not monitoring the browser console (similar to the error overlay). However, this will require some design thought and isn't necessary for the first iteration.

Closes NEXT-1798

Issue #51805, opened 2023-06-26
Nested route unreachable if inside or besides a parallel route


lu-zen:
### Verify canary release

- [X] I verified that the issue exists in the latest Next.js canary release

### Provide environment information

```
Operating System:
      Platform: linux
      Arch: x64
      Version: #1 SMP PREEMPT_DYNAMIC Thu, 30 Mar 2023 14:10:04 +0000
    Binaries:
      Node: 18.16.0
      npm: 9.5.1
      Yarn: 1.22.19
      pnpm: 8.6.4
    Relevant Packages:
      next: 13.4.8-canary.2
      eslint-config-next: N/A
      react: 18.2.0
      react-dom: 18.2.0
      typescript: 5.1.3
    Next.js Config:
      output: N/A
```

### Which area(s) of Next.js are affected? (leave empty if unsure)

App directory (appDir: true), Routing (next/router, next/navigation, next/link)

### Link to the code that reproduces this issue or a replay of the bug

https://github.com/lu-zen/next-parallel-route-bug

### To Reproduce

1. install and start project;
2. navigate to `/nested/misc` and `nested2/misc` paths;

### Describe the Bug

Next returns a 404 error when navigating to a nested route that lives besides or inside a parallel route.

### Expected Behavior

To render the nested route.

lu-zen:
[CodeSandbox](https://codesandbox.io/p/sandbox/green-glitter-2fs2s5) for quick reproduction

ztanner:
I think you might be running into expected behavior when a parallel route reaches an unmatched segment, explained in the docs [here](https://nextjs.org/docs/app/building-your-application/routing/parallel-routes#unmatched-routes). Adding a `default.tsx` file to your parallel route will fallback for nested paths that don't have a corresponding parallel segment. Here's your CodeSandbox forked: https://codesandbox.io/p/sandbox/happy-hoover-5plgwq

lu-zen:
Thanks.

Issue #49569, opened 2023-05-10
[NEXT-1165] [App Router]: Group routes with parallel routes as a page slot, nested route will render NEXT_NOT_FOUND in server side.


Innei:
### Verify canary release

- [X] I verified that the issue exists in the latest Next.js canary release

### Provide environment information

```bash
Operating System:
      Platform: darwin
      Arch: arm64
      Version: Darwin Kernel Version 22.4.0: Mon Mar  6 21:01:02 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8112
    Binaries:
      Node: 18.15.0
      npm: 8.17.0
      Yarn: 1.22.19
      pnpm: 8.1.0
    Relevant packages:
      next: 13.4.1
      eslint-config-next: N/A
      react: 18.2.0
      react-dom: 18.2.0
```

### Which area(s) of Next.js are affected? (leave empty if unsure)

App directory (appDir: true)

### Link to the code that reproduces this issue

https://github.com/Innei/nextjs-group-routes-reproduction

### To Reproduce

1. Start dev
2. Go to home page `/`
3. Click posts, nav to `/posts`
4. Refresh page, got `NEXT_NOT_FOUND`

https://github.com/vercel/next.js/assets/41265413/d1b395f0-01f9-4e84-a045-ce00ef34b974

### Describe the Bug

Render nested route in the server will be got `NEXT_NOT_FOUND`, but it works fine on the client side.

This error will happen in this file structure.

```
├── (root)
│   ├── (site)
│   │   ├── @footer
│   │   ├── @head
│   │   ├── layout.tsx
│   │   ├── page.tsx
│   │   └── posts
│   ├── @headxx
│   │   └── page.tsx
│   ├── error.tsx
│   ├── layout.tsx
│   ├── page.tsx
│   └── providers.tsx
├── favicon.ico
├── globals.css
├── layout.tsx
└── page.tsx
```

### Expected Behavior

No error.

<sub>[NEXT-1165](https://linear.app/vercel/issue/NEXT-1165/[app-router]-group-routes-with-parallel-routes-as-a-page-slot-nested)</sub>

mkarajohn:
I am experiencing this error as well. 

I am using a slot for a modal which uses route interception (pretty much like the example in https://github.com/vercel-labs/nextgram) 

I tried to move the slot inside a `(slots)` group and route interception and the slot itself stops working, with the error

```
Unhandled Runtime Error
[...]

PR #97108, opened 2026-08-10
Prune incomplete parallel route matchers

gnoff:
## Summary

This adds an experimental `strictRouteMatching` flag that leaves a matcher out when its finalized loader tree contains a synthesized default that will always call `notFound()` for a slot actually declared by the owning layout. Adding a `default.tsx` keeps the matcher, and structural router-state branches that are not real slots of that layout do not make it incomplete. The flag defaults to `false` so matcher pruning can roll out independently from the preceding loader-tree correction.

The main goal is to make `children` and named slots behave the same way. If a URL cannot construct a complete route tree on its own, it should be treated as an unmatched URL instead of sometimes showing a slot 404 and sometimes producing a missing page or default error.

## Semantics

For example:

```text
app/split-matcher/
├── layout.tsx
├── foo/page.tsx
├── bar/page.tsx
└── @slot/[...parts]/page.tsx
```

With strict route matching disabled, Next.js emits matchers for `/split-matcher/foo`, `/split-matcher/bar`, and the broad `/split-matcher/[...parts]`. The broad matcher can match any URL, but its `children` branch has neither a matching page nor a default, so it can only construct a permanent 404 tree. With strict route matching enabled, that broad matcher is omitted while `/foo` and `/bar` remain because each combines a real `children` page with the named-slot catchall. Adding `app/split-matcher/default.tsx` deliberately makes the broad matcher complete and keeps it.

The same rule applies without a catchall:

```text
app/disagreeing-slots/
├── layout.tsx
├── @first/foo/page.tsx
└── @second/bar/page.tsx
```

`/foo` is incomplete because `@second` has no matching page or default, and `/bar` is incomplete for the corresponding reason in `@first`, so neither matcher is emitted. A later PR in this stack reports the now-unreachable page files as a project misconfiguration.

A declared `children` route is treated like any named slot:

```text
app/declared-children/
├── layout.tsx
├── page.tsx
└── @panel/
    ├── default.tsx
    └── details/page.tsx
```

`/declared-children` is complete because `children` uses `page.tsx` and `@panel` uses its default. `/declared-children/details` is incomplete because `@panel` matches its page while the declared `children` slot has no matching page or default at `/details`, so that matcher is pruned.

A route composed entirely from declared named slots is also complete:

```text
app/named-only/
├── layout.tsx
├── @left/[...slug]/page.tsx
└── @right/[...slug]/page.tsx
```

The preceding PR's default-on `explicitParallelRouteChildren` behavior means this loader tree contains only `left` and `right`. Strict matching keeps `/named-only/[...slug]` because both declared slots match; it does not invent `children` and then prune the route for failing to satisfy that nonexistent slot.

We got here incrementally. [#47872](https://github.com/vercel/next.js/pull/47872) introduced the 404 fallback for unmatched parallel slots, [#60186](https://github.com/vercel/next.js/pull/60186) added a development warning because this was confusing in practice and linked [#51805](https://github.com/vercel/next.js/issues/51805) and [#49569](https://github.com/vercel/next.js/issues/49569), and [#84702](https://github.com/vercel/next.js/pull/84702) later made a missing default a build error for named slots while leaving `children` on the old fallback for backwards compatibility. Strict matching takes the next step and treats a matcher that can only construct a permanent 404 tree as unmatched.

This changes soft navigations that only worked by preserving a previously active slot even though the URL could not be loaded directly, which is why matcher pruning remains behind its own experimental flag. The interception retention markers backed by `default-null`, including named host slots from the first PR in this stack, are complete route patches and are not pruned. This change however is well motivated because if you did client nav to a route that only matches a named slot and then hard refresh you will end up getting a 404. This is a sign our current semantics are actually broken.

If you want to preserve the perma 404 behavior of the slot-without-default you can just add a default and make it call `notFound()` unconditionally.

## Verification

- `pnpm build-all`
- Focused `normalize-catchall-routes` unit coverage
- Production and development e2e coverage on Turbopack and Webpack
- The pruning e2e matrix with Cache Components enabled


test/e2e/app-dir/interception-dynamic-segment/next.config.js
```diff
@@ -1,6 +1,10 @@
 /**
  * @type {import('next').NextConfig}
  */
-const nextConfig = {}
+const nextConfig = {
+  experimental: {
+    pruneUnmatchedParallelRoutes: true,
```
ztanner: Did you intend to update this test?

test/e2e/app-dir/parallel-routes-pruned-matchers/parallel-routes-pruned-matchers.test.ts
```diff
@@ -0,0 +1,163 @@
+import { nextTestSetup } from 'e2e-utils'
+
+describe('parallel-routes-pruned-matchers', () => {
```
ztanner, line 36: should we assert that a client nav 404s in addition to hard refresh?

ztanner (approved):

test/e2e/app-dir/interception-dynamic-segment/next.config.js
gnoff: yeah because the unflagged behavior (should be) a superset of the flagged behavior so it is asserting that interception routes still work with the default null behavior that exists when parallel slots do not have children during soft nav. Though it does suggest the main test suite actually should get this coverage too

ztanner: Yeah, or maybe we enable it in the existing matrix that we have that turns other flags on? Then coverage is broad across the different parallel route/interception tests. I think that's how we handled the new scroll handler, too.

test/e2e/app-dir/parallel-routes-pruned-matchers/parallel-routes-pruned-matchers.test.ts
gnoff: yeah good call

crates/next-core/src/app_structure.rs
```diff
@@ -1461,15 +1583,34 @@ async fn directory_tree_to_loader_tree_internal(
...
+                subdirectory.modules.default.clone()
+            } else {
+                None
+            };
+
+            // Only emit the issue if this is not the children slot and there's no default
+            // component. The ordinary children route does not require a default.js file.
+            // Also skip validation if the slot is UNDER a catch-all route or if this is a
+            // leaf segment (no child routes).
+            if default.is_none() && key != "children" && !is_inside_catchall && !is_leaf_segment {
```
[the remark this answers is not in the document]
gnoff: :robot: Fixed. I added an interception-host regression where @content has no default but has matching pages for every ordinary child route. Before the fix Turbopack failed the build with the reported @content/default.js error. The interception-host validation now applies the same slot_has_children exception as the earlier parallel-slot validation. The test passes in Turbopack dev and production and in webpack production.

acdlite (approved):

PR #97128, opened 2026-08-10
Fix: Optimistic routing bugs leading to repeated prefetch loops

### LINKED ISSUE #97135: Infinite prefetch loop (livelock, ~600 req/s) when a root parallel slot contains a catch-all and the current page is nested under a sibling dynamic segment

https://github.com/machine23/next-modal-catchall-prefetch-loop

### To Reproduce

1. `npm install && npm run build && npm run start` (production build — the bug never fires in dev, where viewport prefetch is disabled).
2. Open `http://localhost:4611/alice/followers` in a browser.
3. Watch the network tab: an endless stream of `_rsc=` requests carrying `next-router-prefetch: 1`, at hundreds per second, that never stops. (The repro includes `probe.cjs`, a 10-second Playwright counter.)

The repro is 7 tiny files. The shape that matters:

```
app/layout.jsx                      root layout rendering {children}{modal}
app/@modal/default.jsx              returns null
app/@modal/[...catchAll]/page.jsx   returns null   ← the documented "close modal on navigation" pattern
app/page.jsx                        <Link href="/alice/followers">
app/[username]/layout.jsx           async, awaits params, renders a <Link href="/viewer">
app/[username]/page.jsx             async profile page
app/[username]/followers/page.jsx   async, renders ten <Link href={`/user${i}`}>
```

All three ingredients are required:

- a catch-all page inside a **root parallel slot** (`@modal/[...catchAll]` — the canonical modal-close pattern from the parallel-routes docs),
- a **sibling dynamic segment** (`[username]`),
- the current page **nested under** that dynamic segment (`/alice/followers` — one more URL part than the segment consumes). The profile root `/alice` itself does NOT loop, and neither does `/`.

Current (16.3.0 and 16.3.1-canary.10, production build, default config): sitting idle on `/alice/followers` issues ~6,300–6,650 `_rsc` requests per 10 seconds, forever — a prefetch-scheduler livelock. Exactly the first 4 link targets loop (the scheduler's concurrent-request cap) and the rest starve. Measured matrix with the included probe (two runs each):

| next version     | RSC requests in 10 s on `/alice/followers` |
| ---------------- | ------------------------------------------ |
| 16.2.12          | 22 (settles)                               |
| 16.3.0           | 6,284 / 6,502                              |
| 16.3.1-canary.10 | 6,285 / 6,643                              |

Expected: prefetches settle after at most one round per visible link, as on 16.2.12.

`experimental: { optimisticRouting: false }` eliminates the loop completely (each link settles after one `/_tree` + one metadata request), which matches the mechanism below — the regression is 16.3.0 flipping that default to `true`.

```
Operating System: Linux (also reproduced inside node:24-alpine docker)
Node: 24.x
next: 16.3.0 (also 16.3.1-canary.10)
react / react-dom: 19.2.x
Stages affected: next start (production). next dev never triggers it (no viewport prefetch in dev).
```

Navigation, Prefetching (segment cache / optimistic routing)

### Additional context — where the loop comes from (traced through the shipped chunks with instrumentation)

The cycle lives in the Known Routes discovery (`segment-cache/optimistic-routes.js`), default-enabled since 16.3.0:

1. **Trie poisoning at hydration.** `discoverKnownRoutePart()` walks ALL parallel slots of a node against one shared URL-part index, and `discoverDynamicChild()` returns an existing dynamic child **without comparing param name or type**. On `/alice/followers`, the `children` branch creates the trie node for `[username]` and stores the page's pattern under `staticChildren['followers']`; then the `@modal` branch's `[...catchAll]` **reuses the same node** (the mismatch is ignored), sets `nextPartIndex = pathnameParts.length` (a catch-all absorbs all remaining parts), and — since the slot's `__PAGE__` is transparent — terminates AT that node, planting the full **nested followers-page tree as the `[username]` node's direct pattern**.
2. **Synthetic route with a poisoned vary path.** `matchKnownRoutePart('/user3')` consumes `user3` at the dynamic node, runs out of parts, and returns the node's direct pattern: a synthetic fulfilled route whose tree is the *followers* page reified with `username=user3`, metadata vary path `…/followers/__PAGE__/_head`.
3. **Fulfill/re-key mismatch = livelock.** The prefetch task reads the head at the synthetic path → miss → spawns a Pending entry → blocks on it (the #96017 "block until sufficient response" semantics) → issues the metadata-only dynamic request (`next-router-state-tree: ["",{},null,"metadata-only"]` — the request seen looping). The response (the real `/user3` page) fulfills the spawned entry via the constant `/_head` request key, re-pinging the task — and then **re-keys the entry to the response-derived canonical path** `…/$d$username/__PAGE__/_head`, a different cache path. The re-pinged pass calls `readRouteCacheEntry` again; synthetic routes are never stored in the route cache map, so it mints a fresh clone with the same poisoned vary path, misses again (the fulfilled entry now lives under an unreachable key), spawns again, fetches again. The entry is fulfilled-then-moved every round, so it never rejects and no backoff ever applies.

Per-iteration instrumentation logs (read spot → miss → request → fulfill → re-key, repeating) are available on request.

Related, but distinct triggers: #96611, #96627, #97128 — #97128 fixes the same class for proxy rewrites and explicitly defers a general prefetch-retry bound as a TODO; none of them covers the parallel-slot catch-all trigger, which needs no rewrites, no `cacheComponents`, and no special config.

Suggested fix direction: `discoverDynamicChild()` should reject a param name/type mismatch (`'d'` vs `'c'`), and the catch-all discovery branch should not store patterns at a foreign dynamic-child node — either alone breaks the poisoning. A general retry bound (the #97128 TODO) would additionally turn any future bug of this class from a livelock into a bounded burst.

acdlite:
Fixes a few bugs related to optimistic routing.

Originally reported as a next-intl request waterfall in MarkBekooy/prefetching-request-waterfall-bug#1, then extracted into an isolated regression test: A proxy that rewrites every URL to inject a leading path segment (`/one/two` → `/en/one/two`, i.e. i18n with the default locale hidden from the URL) plus a fully dynamic target route like `/[locale]/[...pages]` leads to an infinite prefetch loop that never resolves.

This regression uncovered several oversights in the optimistic routing implementation: when receiving a prefetch response, we did not check whether the response matched the expected result. If the tree mismatched, in some cases the prefetch task would fall into a loop, repeatedly attempting to fulfill the missing data.

Now when this happens, we record on the local route definition that a dynamic rewrite occurred, disabling further attempts to optimistically resolve the route. This is the same strategy we were already using for normal navigation responses, now applied to the prefetch path.

We also received another bug report with a similar root cause: parallel routes with conflicting dynamic params at the same level (`@modal/[...catchAll]` next to `[username]`) cannot be distinguished using the current traversal algorithm, because it assumes that each segment is resolvable independently without inspecting the children or sibling branches. To work around this issue for now, when this scenario is detected, we disable optimistic routing for the conflicting route. We can model this properly in a future PR by having the server send down the "sibling" dynamic route segments, similar to what we do for static siblings already.

Fixes #97135

### DIFF
--- a/packages/next/src/client/components/router-reducer/create-initial-router-state.ts
+++ b/packages/next/src/client/components/router-reducer/create-initial-router-state.ts
@@ -82,7 +82,7 @@ export function createInitialRouterState({
   // trees with InliningHintsStale, which causes the route cache entry to be
   // immediately expired. The next prefetch will re-fetch the tree with
   // correct hints from the /_tree response.
-  const acc = { metadataVaryPath: null }
+  const acc = { metadataVaryPath: null, treeDivergedFromBase: false }
   const initialRouteTree = decodeTransportTreeIntoRouteTree(
     initialTransportData.t,
     // There's no base tree to overlay onto; the initial payload is a full
--- a/packages/next/src/client/components/router-reducer/ppr-navigations.ts
+++ b/packages/next/src/client/components/router-reducer/ppr-navigations.ts
@@ -904,7 +904,7 @@ function reuseActiveSegmentInDefaultSlot(
     reusedRenderedSearch = oldRootRefreshState.renderedSearch
   }
 
-  const acc = { metadataVaryPath: null }
+  const acc = { metadataVaryPath: null, treeDivergedFromBase: false }
   const reusedRouteTree = convertReusedFlightRouterStateToRouteTree(
     parentRouteTree,
     parallelRouteKey,
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -1698,6 +1698,9 @@ function rejectSegmentCacheEntry(
 
 export type RouteTreeAccumulator = {
   metadataVaryPath: PageVaryPath | null
+  // Whether the decoded tree's segment identities diverged from the base
+  // tree it was overlaid onto. See NavigationSeed.treeDivergedFromBase.
+  treeDivergedFromBase: boolean
 }
 
 function convertRootTreePrefetchToRouteTree(
@@ -2201,7 +2204,10 @@ export async function fetchRouteOnCacheMiss(
       //
       // During this traversal, we accumulate additional data into this
       // "accumulator" object.
-      const acc: RouteTreeAccumulator = { metadataVaryPath: null }
+      const acc: RouteTreeAccumulator = {
+        metadataVaryPath: null,
+        treeDivergedFromBase: false,
+      }
       const routeTree = convertRootTreePrefetchToRouteTree(
         serverData,
         renderedPathname,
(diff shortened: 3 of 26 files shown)


cgatian:
We just hit this in production 😨

packages/next/src/client/components/router-reducer/create-initial-router-state.ts
```diff
@@ -83,7 +82,7 @@ export function createInitialRouterState({
   // trees with InliningHintsStale, which causes the route cache entry to be
   // immediately expired. The next prefetch will re-fetch the tree with
   // correct hints from the /_tree response.
-  const acc = { metadataVaryPath: null }
+  const acc = { metadataVaryPath: null, treeDivergedFromBase: false }
```
unstubbable, line 85: nit:
```suggestion
  const acc: RouteTreeAccumulator = {
    metadataVaryPath: null,
    treeDivergedFromBase: false,
  }
```

packages/next/src/client/components/router-reducer/ppr-navigations.ts
```diff
@@ -920,7 +904,7 @@ function reuseActiveSegmentInDefaultSlot(
     reusedRenderedSearch = oldRootRefreshState.renderedSearch
   }
 
-  const acc = { metadataVaryPath: null }
+  const acc = { metadataVaryPath: null, treeDivergedFromBase: false }
```
unstubbable, line 907: nit:
```suggestion
  const acc: RouteTreeAccumulator = {
    metadataVaryPath: null,
    treeDivergedFromBase: false,
  }
```

unstubbable (approved):

PR #97130, opened 2026-08-10
[turbopack] Bail CJS tree-shaking on `var x = module.exports = {}`

sampoder:
This code was causing issues for the CJS tree-shaker:

```javascript
```js
"use strict";
var utils = require('./utils');

var LinkedList = module.exports = {          // ← line 4: local alias for module.exports
    // basic validity tests on a circular linked list a
    valid: function(a) {
        utils.assert(a, "list falsy");
        utils.assert(a._previousSibling, "previous falsy");
        utils.assert(a._nextSibling, "next falsy");
        return true;
    },
    // insert a before b
    insertBefore: function(a, b) {
        utils.assert(LinkedList.valid(a) && LinkedList.valid(b));   // ← line 15: THROWS
        var a_first = a, a_last = a._previousSibling;
        var b_first = b, b_last = b._previousSibling;
        a_first._previousSibling = b_last;
        a_last._nextSibling = b_first;
        b_last._nextSibling = a_first;
        b_first._previousSibling = a_last;
        utils.assert(LinkedList.valid(a) && LinkedList.valid(b));   // line 22
    },
    // replace a single node a with a list b (which could be null)
    replace: function(a, b) {
        utils.assert(LinkedList.valid(a) && (b===null || LinkedList.valid(b)));  // line 26
        if (b!==null) {
            LinkedList.insertBefore(b, a);
        }
        LinkedList.remove(a);
        utils.assert(LinkedList.valid(a) && (b===null || LinkedList.valid(b)));  // line 31
    },
    // remove single node a from its list
    remove: function(a) {
        utils.assert(LinkedList.valid(a));                          // line 35
        var prev = a._previousSibling;
        if (prev === a) { return; }
        var next = a._nextSibling;
        prev._nextSibling = next;
        next._previousSibling = prev;
        a._previousSibling = a._nextSibling = a;
        utils.assert(LinkedList.valid(a));                          // line 42
    }
};
```

Because we were shaking `valid` as it was only used via the `LinkedList` alias. Now we bail if you set up an alias like that.

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
@@ -764,6 +764,21 @@ pub fn as_parent_path_with_in<'a>(
     path.into_boxed_slice()
 }
 
+/// Whether the node at `ast_path` is a whole statement, so its value goes nowhere:
+/// true for `module.exports = {};` and `(module.exports = {});`, false for
+/// `var x = module.exports = {}` or `f(module.exports = {})`.
+fn is_expression_statement(ast_path: &AstNodePath<AstParentNodeRef<'_>>) -> bool {
+    for node_ref in ast_path.iter().rev() {
+        match node_ref {
+            // The `Expr` wrapper of the node itself, and of a parenthesized one.
+            AstParentNodeRef::Expr(..) | AstParentNodeRef::ParenExpr(_, ParenExprField::Expr) => {}
+            AstParentNodeRef::ExprStmt(_, ExprStmtField::Expr) => return true,
+            _ => return false,
+        }
+    }
+    false
+}
+
 /// Returns the [`MemberExpr`] where the current node is the object:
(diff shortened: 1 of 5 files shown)


turbopack/crates/turbopack-tests/tests/execution/webpack/cjs-tree-shaking/bailouts/input/alias-exports-object-literal.js
```diff
@@ -0,0 +1,11 @@
...
+// module reads its own exports back through. Nothing may be dropped, even though
+// `helper` has no importer.
+const Self = (module.exports = {
+  used: function () {
+    return Self.helper()
+  },
+  helper: function () {
+    return 'helper'
+  },
+})
```
mischnic: Why is this a snapshot test? Can't it be an execution test?

sampoder: No real reason, I can make it an execution test

mischnic (approved):

PR #97150, opened 2026-08-11
Scope app-entry export validation to files inside the app directory

### LINKED ISSUE #96859: Turbopack build fails on pages-router files named `sitemap`/`robots`: `"getStaticProps" is not supported in app/` (no app directory)

rodrigo-arias:
Fixes #96859

### What?

Since 16.3.0, builds fail for pages-router files named `sitemap` or `robots` that export `getStaticProps`/`getServerSideProps`:

```
Error: "getStaticProps" is not supported in app/.
```

This is a regression from #94962, which added the metadata conventions (`sitemap`, `robots`, `manifest`, `icon`, …) to the app-entry filename regex in `ReactServerComponentValidator::assert_invalid_api`. The regex only looks at the filename, so a file like `pages/sitemap.js` is now mistaken for an app entry and rejected for exporting `getStaticProps`.

### How?

A file is only treated as an app entry when it's inside `appDir`, reusing the gate `assert_server_filename` already applies to `error.js`. The pages compilation context has no `appDir`, so pages-router files are never validated as app entries.

On the test side:

- The fixtures that exercise the app-entry checks moved under `app-dir/`, since the test harness derives `appDir` from the fixture path. Their contents are unchanged.
- A new fixture (plus a `sitemap.js` fixture glob) asserts that a pages-style `sitemap.js` compiles without errors.
- A new e2e suite, `test/e2e/pages-metadata-filenames`, covers `pages/sitemap.js` with `getStaticProps` and `pages/robots.js` with `getServerSideProps`. It fails without the fix under Turbopack (build and dev) and passes with it under both Turbopack and webpack.

Supersedes #96873 (same approach, closed by its author) and #96967.


rodrigo-arias:
cc @unstubbable — this fixes a 16.3.0 regression from #94962 (`pages/sitemap.js` with `getStaticProps` now fails the build, reported in #96859). The fix reuses the existing `appDir` gate so pages-router files aren't validated as app entries. Happy to adjust if you'd prefer a different approach.

PR #92070, opened 2026-03-29
TURBOPACK: support crsossOriginLoading

fireairforce:
Support: https://nextjs.org/docs/pages/api-reference/config/next-config-js/crossOrigin#options for turbopack.

PR #97164, opened 2026-08-11
Fix unset crossOrigin in Turbopack manifests

timneutkens:
## Summary

Prevent Turbopack from adding `crossorigin=""` to generated script tags when `crossOrigin` is not configured, and add production coverage for static prerendering, dynamic rendering, and `output: 'export'`.

This regressed in #92070 when Turbopack added `crossOriginLoading` support. The client reference manifest changed from an optional cross-origin value to the shared `CrossOrigin` enum. Serde serialized the enum's default `None` variant as the string `"none"`, so React saw a present string and treated it as anonymous CORS, emitting `crossorigin=""`. Webpack continued to omit its `undefined` value.

Skip serializing `CrossOrigin::None` in the Turbopack client reference manifest so an unset option remains absent while preserving explicit anonymous and credentialed modes.

## Verification

- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-turbo test/e2e/app-dir/app-config-crossorigin/index.test.ts`
- `NEXT_TEST_PREFER_OFFLINE=1 pnpm test-start-webpack test/e2e/app-dir/app-config-crossorigin/index.test.ts`


test/e2e/app-dir/app-config-crossorigin/index.test.ts
```diff
@@ -37,3 +55,52 @@ describe('app dir - crossOrigin config', () => {
     })
   })
 })
+
+if (isNextStart) {
+  describe('app dir - unset crossOrigin config', () => {
```
mischnic: This dance of isConfiguredNextStart+early-return and isNextStart seems wrong

timneutkens: :robot: Done -- restructured the file into a single development/production branch so only the relevant nextTestSetup instances are created. I also made the dynamic fixture compatible with Cache Components and exclude the static-export sub-suite only when Cache Components/PPR is forced, since that combination is unsupported. The exact previously failing CI environment now passes.

test/e2e/app-dir/app-config-crossorigin/index.test.ts
```diff
@@ -1,39 +1,108 @@
...
+      }
 
-    // Same origin <script /> and <link /> should not have crossorigin attribute either
-    $('script[src^="/"], link[href^="/"]').each((_, el) => {
-      const crossOrigin = $(el).attr('crossorigin')
-      expect(crossOrigin).toBeUndefined()
+      beforeAll(async () => {
+        await next.build()
+        await next.start()
+      })
```
mischnic: `skipStart` and then a `beforeAll` to `build`+`start` after all?

timneutkens: :robot: Done — removed skipStart and the manual build/start beforeAll from the default-output suite. It now uses nextTestSetup's normal production lifecycle; only the output: export suite remains manual because it starts the custom static output server.

mischnic (approved):

PR #97165, opened 2026-08-11
[PPF] Only track runtime accesses when the promise is used

lubieowoce:
Improves how we track accesses of runtime data in static prerenders (which affects whether a page will use static or runtime shells).
Now, Instead of tracking it when the hanging promise is created, we'll track it when `then/catch/finally` is called on the promise (which we use as a signal for being used).

The promise tracking required some novel work, because:

- Using a proxy seems to cause `Promise.then` to show up in call stacks (we've had issues with proxies throwing off React's IO tracking before, i believe this is a similar problem)
- patching `then/catch/finally` on the promise directly doesn't work, because `then` doesn't get called for native `await`

The only way of doing this that i've found is a custom `Promise` subclass that tracks the access in its `then` implementation. This is what `TrackedPromise` does.

To improve the debuggability, i've added some debug logs enabled by `NEXT_PRIVATE_DEBUG_RUNTIME_DATA=1` to detect when a page is deopted to runtime and what expression caused it.

---

This fixes a false deopt caused by `createServerPathnameForMetadata` - when called on a route with fallback params, it would always track a fallback params access even if no icons/og images that actually use it exist. This meant that we'd never statically optimize any routes that had fallback params. After this PR, we only track a runtime data access if a metadata route that uses that params actually exists, in which case the `pathname` promise gets awaited.
(as a future optimization, we should do the equivalent of vary params tracking on metadata routes to know if the params are used at all. also, the logic in `pathname.ts` is generally in need of cleanup)

### DIFF
--- a/packages/next/src/server/dynamic-rendering-utils.ts
+++ b/packages/next/src/server/dynamic-rendering-utils.ts
@@ -8,7 +8,6 @@ import type {
 } from './app-render/work-unit-async-storage.external'
 import { workUnitAsyncStorage } from './app-render/work-unit-async-storage.external'
 import { getServerReact, getClientReact } from './runtime-reacts.external'
-import { ReflectAdapter } from './web/spec-extension/adapters/reflect'
 
 export function isHangingPromiseRejectionError(
   err: unknown
@@ -112,7 +111,7 @@ export function makeUntrackedHangingPromise<T>(
  * searchParams, and cache entries that are excluded only from static
  * prerenders.
  *
- * Creating one of these during a static prerender records on the prerender
+ * Awaiting one of these during a static prerender records on the prerender
  * store that a runtime prefetch would produce more content than the static
  * response (`runtimeDataAccessed`), which the segment prefetch encoding uses
  * to tell the client whether a runtime prefetch request could be skipped.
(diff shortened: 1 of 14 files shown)


packages/next/src/server/dynamic-rendering-utils.ts
```diff
@@ -202,10 +210,12 @@ export function makeStageHangingPromise<T>(
   expression: string,
   workUnitStore: WorkUnitStore
 ): Promise<T> {
-  trackRuntimeDataAccessed(workUnitStore)
-  return makeHangingPromiseWithError(
-    signal,
-    new HangingPromiseRejectionError(route, expression)
+  return trackPromiseUsed(
+    makeHangingPromiseWithError(
```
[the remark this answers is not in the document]
lubieowoce: no it doesn't

packages/next/src/server/dynamic-rendering-utils.ts
```diff
@@ -389,38 +403,72 @@ export function makeDevtoolsIOAwarePromise<T>(
...
+    Promise.prototype.then.call(tracked, undefined, ignoreReject)
+
+    return tracked
+  }
+
+  then<TResult1, TResult2>(
+    onFulfilled?: (value: T) => TResult1 | PromiseLike<TResult1>,
+    onRejected?: (reason: unknown) => TResult2 | PromiseLike<TResult2>
+  ): Promise<TResult1 | TResult2> {
+    const onUse = this.#onUse
```
unstubbable, line 456: Should we set `this.#onUse` to `null` on first use? Right now we're relying on the fact that the callers pass an idempotent callback as `onUse`, right?

unstubbable (approved):

packages/next/src/server/dynamic-rendering-utils.ts
lubieowoce: hmm, not sure. all our`then`-wrapping proxies fire every time, so i'm inclined to keep this one the same for consistency, doesn't rly cost much

maybe we should just expose something like an `EventTarget` interface and then the consumer can decide if they want `{ once: true }` or not? but that feels a bit heavyweight.
in general i've been thinking about how to unify our tracking code bc it's kind of spread out over a couple different implementations, so i'd like to handle them holistically if possible. not in this PR though

PR #97166, opened 2026-08-11
Restore the live `headers()` view of the incoming request

### LINKED ISSUE #97049: headers() returns stale request headers in Proxy after NextRequest.headers mutation since 16.3.0

https://github.com/r34son/nextjs-proxy-headers-snapshot-regression

### To Reproduce

1. Clone the reproduction:

   ```bash
   git clone https://github.com/r34son/nextjs-proxy-headers-snapshot-regression.git
   cd nextjs-proxy-headers-snapshot-regression
   pnpm install
   ```

2. Start it with Next.js 16.3.0 or the current canary:

   ```bash
   pnpm dev
   ```

3. Open `http://localhost:3000/` or run:

   ```bash
   curl http://localhost:3000/
   ```

The Proxy performs these operations in order:

```ts
const firstHeadersView = await headers()

request.headers.set(
  'x-added-after-headers-call',
  'set-on-next-request',
)

const secondHeadersView = await headers()
```

It then reads the same header from:

- `request.headers` after the mutation;
- the `headers()` view obtained before the mutation;
- a new `headers()` call made after the mutation.

Starting with Next.js 16.3.0, `request.headers` and `headers()` expose inconsistent views of the same incoming request inside Proxy.

After mutating `request.headers`, the value is present on the `NextRequest`:

```json
{
  "valueOnNextRequest": "set-on-next-request"
}
```

However, both the previously obtained `headers()` view and a new `headers()` call made after the mutation return the old value:

```json
{
  "valueBeforeMutation": null,
  "valueOnNextRequest": "set-on-next-request",
  "valueFromFirstViewAfterMutation": null,
  "valueFromSecondViewAfterMutation": null,
  "sameHeadersObject": true
}
```

In Next.js 16.2.4, both `headers()` views observe the mutation:

```json
{
  "valueBeforeMutation": null,
  "valueOnNextRequest": "set-on-next-request",
  "valueFromFirstViewAfterMutation": "set-on-next-request",
  "valueFromSecondViewAfterMutation": "set-on-next-request",
  "sameHeadersObject": true
}
```

The expected behavior is that request-header APIs used during the same Proxy execution provide a consistent and clearly defined view of the incoming request.

In particular, one of the following contracts should apply:

1. A `headers()` call performed after a `NextRequest.headers` mutation observes the current request headers; or
2. `headers()` is explicitly defined as an immutable snapshot captured before Proxy execution, and its use and limitations inside Proxy are clearly documented.

The current behavior is surprising because a new `headers()` call made after the mutation still returns the cached stale snapshot.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: macOS 26.5, Darwin 25.5.0

Binaries:
  Node: 22.18.0
  pnpm: 10.14.0

Relevant Packages:
  next: 16.3.0-preview.9
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.x

Next.js Config:
  output: N/A
```

Headers

next dev (local)

### Regression range

The behavior differs between these versions:

| Next.js version | `request.headers` after mutation | First `headers()` view after mutation | Second `headers()` call after mutation |
| --- | --- | --- | --- |
| 16.2.4 | `set-on-next-request` | `set-on-next-request` | `set-on-next-request` |
| 16.3.0 | `set-on-next-request` | `null` | `null` |
| 16.3.0-preview.9 | `set-on-next-request` | `null` | `null` |

The issue therefore still reproduces on the canary version tested on August 10, 2026.

The behavior change appears to have been introduced by https://github.com/vercel/next.js/pull/95116, which started copying the request headers before creating the sealed userland headers view.

This copy prevents later mutations of `NextRequest.headers` from being visible through `headers()`. Because the `headers()` result is cached, calling `headers()` again after the mutation returns the same stale view.

### Documentation gap

The JSDoc for `headers()` explicitly lists Middleware as a supported execution context. Middleware was renamed to Proxy in Next.js 16, so this communicates that `headers()` is supported inside `proxy.ts`.

The API documentation also describes `headers()` as an API for reading the incoming request headers. It does not explain that, inside Proxy, it may expose a separate snapshot captured before user Proxy code executes.

The Proxy documentation shows how to create modified request headers and forward them using:

```ts
NextResponse.next({
  request: {
    headers: requestHeaders,
  },
})
```

However, it does not define how that header state relates to `headers()` during the current Proxy execution. In particular, the documentation does not state that:

- `NextRequest.headers` and `headers()` can represent different header states;
- the `headers()` value can be captured before Proxy code runs;
- subsequent `headers()` calls return a cached snapshot rather than the current `NextRequest.headers`;
- mutating `NextRequest.headers` will not be reflected by `headers()`.

This makes the current API contract ambiguous. The JSDoc says the API is available in Middleware/Proxy, but the observable behavior differs from what “read the incoming request headers” naturally suggests.

The JSDoc for [`headers()`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/request/headers.ts#L31-L38) explicitly lists Middleware as a supported execution context. Middleware was renamed to Proxy in Next.js 16.

If the snapshot behavior is intentional, the JSDoc and documentation should state it explicitly and recommend the supported way to access or propagate modified headers. Otherwise, `headers()` should remain consistent with the current request-header state, as it was before 16.3.0.

unstubbable:
#94703 and #95116 both changed how internal headers are kept out of userland `headers()`, and between them they left `headers()` detached from the request it describes. #94703 started stripping the dev-only request-id headers alongside the flight headers that had been stripped for much longer. Those deletes ran against the shared request headers, because `HeadersAdapter.from` returns a `Headers` instance unchanged instead of copying it, and removing the request-id headers there broke the dev debug channel when the dev server rendered a redirect target after a server action. #95116 fixed that by copying the headers before stripping them.

Neither change set out to alter whether `headers()` follows the request. The sealed view has read through to the underlying headers for as long as `HeadersAdapter.seal` has existed, and its own unit tests assert that. The copy was only a way to keep the deletes from escaping into `req.headers`, and the detached view was collateral that no test covered.

The result, since 16.3.0, is that a Proxy which writes a header onto `request.headers` and then reads it back through `headers()` gets the value from before the write. Reading the same header directly off `request.headers` returns the new value, so two APIs describing the same request disagree, and a second `headers()` call returns the same stale view. Whether the write is seen at all depends on ordering, because the view is built on first access: a write made before the first `headers()` call is observed, and the identical write made after it is not.

This change drops the copy and stops deleting anything. `HeadersAdapter.seal` now accepts a set of header names to omit from every read operation, and `getHeaders` seals the shared request headers directly with the flight headers and the dev request-id headers hidden. Both earlier intentions survive. The internal headers stay on the request where the framework still reads them, they stay invisible to userland `headers()`, and the sealed view tracks the request again, so the two APIs agree throughout a Proxy run.

Hiding on read is slightly stricter than the delete it replaces. The delete ran once, when the view was created, so an internal header written to the request afterwards would have shown through. It now stays hidden no matter when it is written.

While restructuring `seal` we also corrected the `parent` argument that `forEach` passes to its callback. The native method passes the unsealed target, which hands the callback a mutable handle on the underlying headers and defeats the seal. Both the plain and the hiding handler now pass the sealed proxy instead.

`cookies()` is unaffected and stays a snapshot, because `RequestCookies` parses the `cookie` header when it is constructed. That predates 16.3 and is left alone here.

The fix is based on #97145 by @tachsin. That pull request comes from a fork, and deploy tests do not run for fork pull requests, so it is recreated here on a branch of this repository to get full CI coverage. The commit carries a `Co-authored-by` trailer for them, so their credit is preserved on merge. We also took the liberty of a few improvements over the original: `seal` is restructured into two handlers rather than a guard repeated in each case, `getSetCookie` and `forEach` are handled as described above, and the unit and end-to-end coverage is wider.

fixes #97049
closes #97145

### DIFF
--- a/packages/next/src/server/web/spec-extension/adapters/headers.ts
+++ b/packages/next/src/server/web/spec-extension/adapters/headers.ts
@@ -25,6 +25,102 @@ export type ReadonlyHeaders = Headers & {
   /** @deprecated Method unavailable on `ReadonlyHeaders`. Read more: https://nextjs.org/docs/app/api-reference/functions/headers */
   delete(...args: any[]): void
 }
+
+type ForEachCallback = (value: string, name: string, parent: Headers) => void
+
+/**
+ * Blocks the mutating methods of a sealed `Headers` instance.
+ *
+ * `forEach` is reimplemented so that the callback receives the sealed proxy as
+ * its `parent` argument. The native method passes the unsealed target, which
+ * gives the callback a mutable handle on the underlying headers.
+ */
+const sealHandler: ProxyHandler<ReadonlyHeaders> = {
+  get(target, prop, receiver) {
+    switch (prop) {
+      case 'append':
+      case 'delete':
+      case 'set':
+        return ReadonlyHeadersError.callable
+      case 'forEach':
+        return (callbackfn: ForEachCallback, thisArg?: any): void => {
+          for (const [name, value] of target.entries()) {
+            callbackfn.call(thisArg, value, name, receiver)
+          }
+        }
+      default:
+        return ReflectAdapter.get(target, prop, receiver)
+    }
+  },
+}
+
+/**
+ * Blocks the mutating methods of a sealed `Headers` instance, and omits the
+ * given header names from every read operation.
+ *
+ * The names in `hidden` must be lowercase.
+ */
+function createHidingSealHandler(
+  hidden: ReadonlySet<string>
+): ProxyHandler<ReadonlyHeaders> {
+  const isHidden = (name: string): boolean => hidden.has(name.toLowerCase())
+
+  return {
+    get(target, prop, receiver) {
+      switch (prop) {
+        case 'append':
+        case 'delete':
+        case 'set':
+          return ReadonlyHeadersError.callable
+        case 'get':
+          return (name: string): string | null =>
+            isHidden(name) ? null : target.get(name)
+        case 'has':
+          return (name: string): boolean =>
+            isHidden(name) ? false : target.has(name)
+        case 'getSetCookie':
+          return (): string[] =>
+            isHidden('set-cookie') ? [] : target.getSetCookie()
+        case 'entries':
+        case Symbol.iterator:
+          return function* (): HeadersIterator<[string, string]> {
+            for (const entry of target.entries()) {
+              if (!isHidden(entry[0])) {
+                yield entry
+              }
+            }
+          }
+        case 'keys':
+          return function* (): HeadersIterator<string> {
+            for (const name of target.keys()) {
+              if (!isHidden(name)) {
+                yield name
+              }
+            }
+          }
+        case 'values':
+          return function* (): HeadersIterator<string> {
+            for (const [name, value] of target.entries()) {
+              if (!isHidden(name)) {
+                yield value
+              }
+            }
+          }
+        case 'forEach':
+          return (callbackfn: ForEachCallback, thisArg?: any): void => {
+            for (const [name, value] of target.entries()) {
+              if (!isHidden(name)) {
+                callbackfn.call(thisArg, value, name, receiver)
+              }
+            }
+          }
+        default:
+          return ReflectAdapter.get(target, prop, receiver)
+      }
+    },
+  }
+}
+
 export class HeadersAdapter extends Headers {
   private readonly headers: IncomingHttpHeaders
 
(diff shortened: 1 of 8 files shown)


tachsin:
Thanks for picking this up and getting it full CI coverage. The restructuring into `sealHandler`/`createHidingSealHandler` is cleaner than what I had, and good catch on `getSetCookie` plus the wider test coverage on `forEach`'s `parent` argument. Happy to close #97145 in favor of this one whenever it merges.

packages/next/src/server/web/spec-extension/adapters/headers.ts
```diff
@@ -25,6 +25,102 @@ export type ReadonlyHeaders = Headers & {
...
+
+  return {
+    get(target, prop, receiver) {
+      switch (prop) {
+        case 'append':
+        case 'delete':
+        case 'set':
+          return ReadonlyHeadersError.callable
+        case 'get':
+          return (name: string): string | null =>
```
lubieowoce: would be nice to assert that the types of these methods actually match `Headers.get` etc, because we're reimplementing them, not passing through to the original impl

would also be nice to cache the methods instead of allocating a fresh closure every time a method is accessed
(perhaps we can define them once in an object and just look it up from there instead of doing everything in a big switch)

lubieowoce (approved):
broadly sensible, [we could improve the proxy a bit](https://github.com/vercel/next.js/pull/97166#discussion_r3765653580) if we have time for nits

packages/next/src/server/web/spec-extension/adapters/headers.ts
unstubbable: Yeah, that would be nice. How about 0fabc03?

PR #97203, opened 2026-08-11
[turbopack] Lazily compile dynamic imports in development (client side)

jimmyhmiller:
## Summary

Defer compiling client-side dynamic import targets in Turbopack development until the browser requests their manifest chunk. This avoids compiling untouched dynamic imports while preserving server-side imports, CSS loading, Server Actions, source maps, and Fast Refresh behavior.

One caveat on this is next/dynamic still does some eagerness. I looked a bit at changing things. But it breaks some of the guarantees there and decided to leave it off the table.


turbopack/crates/turbopack-core/src/module_graph/mod.rs
```diff
@@ -795,6 +795,22 @@ impl ModuleGraph {
         Ok(ReadRef::cell(graph))
     }
 
+    /// A graph containing nothing but `entry` and what it references, as an async chunk group
+    /// entry. Used to place modules that are synthesized during chunking and therefore are not
+    /// members of the graph they were synthesized from.
+    #[turbo_tasks::function]
+    pub fn isolated_async_entry(entry: ResolvedVc<Box<dyn Module>>) -> Vc<Self> {
+        Self::from_graphs(
```
mischnic, line 807: This will end up with multiple module graphs though?
1. I think server actions/client references are broken. Because these async graphs aren't scanned
2. This will lead to duplication in module graph size because the async graph doesn't know about modules that are already loaded in the parent and can be skipped

jimmyhmiller: @mischnic 

> I think server actions/client references are broken.

No I already had a test for server actions behind the dynamic import. It passes and I tested it out manually as well. Just pushed a commit that tests client references. No breakage that I can see.

> This will lead to duplication in module graph size because the async graph doesn't know about modules that are already loaded in the parent and can be skipped

That doesn't appear to be the case at all. I added a test for exactly that now. I did find a problem. Because we already do know things in the parent, we were trying to load a module that was already loaded and it caused issues, so pushed the fix for that one.

If I've missed something you meant, or if these tests don't cover your intended scenarios let me know.

mischnic: Yeah, I wasn't entirely correct in the earlier comment. So the real problem is:

In this PR, you have implemented only lazy async chunking. Not lazy module graph discovery (i.e. parsing, transforming, analyzing), etc.
In your old version, you did have both.
So server actions/client references do indeed work fine.
But you leave performance on the table because the unrequested modules are still analyzed.

jimmyhmiller: @mischnic I don't think that's true either. When I very first started doing this I accidentally had that. So i made a test for it to make sure I wasn't making that mistake. 

But you are partially write as well and I didn't quite catch that distinction. So thanks.

If you do import("myfile.js") right now with this change, myfile.js will be parsed and analyzed and all that. But none of its dependencies will. I confirmed this two ways. 1) a very large set of deps that takes a long time 2) a poison pill dep.

If you put the poison pill in the target of the dynamic import, it errors. If you put it in the deps, it does not.

Working now to move it to not parse/analyze the target either. But this is still a pretty big win if your target itself is not a single massive file.

mischnic: Ah, I missed the `LazyCompilationProxyModule`. Alright, that is fine then. I had expected something in the module graph builder

jimmyhmiller: I pushed a change that makes even the initial file not get eager compiled. I need to spend a bit more time with it. (hence the draft) Maybe went overboard on the claude written tests for the scenarios (though it is always nice to have. Check that addition out too.

But glad you agree on the LazyCompilationProxyModule doing what I was thinking it does. I think with this change, we should be fully lazy.

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -1363,6 +1364,30 @@ export async function createHotReloaderTurbopack(
...
+      if (lazyDynamicImports && req.url?.startsWith('/_next/static/chunks/')) {
+        // Turbopack names chunks relative to `/_next`, and so does `AssetMapper`.
+        const chunkPath = decodeURIComponent(
+          req.url.split('?')[0].slice('/_next/'.length)
+        )
+        const keys = assetMapper.getKeysByAsset(chunkPath)
+        try {
+          if (keys.length > 0 && (await project.activateLazyChunk(chunkPath))) {
+            for (const key of keys) {
+              await hotReloader.ensurePage({
```
mischnic, line 1447: This loop is unfortunate, but I guess it's necessary. We don't know which of these pages is actually being requested.

crates/next-core/src/next_config.rs
```diff
@@ -2686,6 +2689,15 @@ impl NextConfig {
         }))
     }
 
+    #[turbo_tasks::function]
+    pub fn turbopack_lazy_dynamic_imports(&self) -> Vc<bool> {
+        Vc::cell(
+            self.experimental
+                .turbopack_lazy_dynamic_imports
+                .unwrap_or(false),
+        )
```
mischnic: Maybe pass the `mode` here directly, then you don't have to do `is_development() && turbopack_lazy_dynamic_imports` everywhere this is called.

turbopack/crates/turbopack-ecmascript/src/references/esm/dynamic.rs
```diff
@@ -79,26 +90,117 @@ impl EsmAsyncAssetReference {
...
+        if self.lazy_compilation && self.resolve_override.is_none() {
+            let request = self.request.await?;
+            // Pattern requests need their resolved result map for runtime dispatch. A constant
+            // request can use one proxy and defer resolution and processing until activation.
+            if let Some(request_string) = &self.request_string {
+                let origin = self.origin.into_trait_ref().await?;
+                let ident = AssetIdent::from_path(origin.origin_path())
+                    .with_layer(origin.asset_context().into_trait_ref().await?.layer())
+                    .with_modifier(
+                        format!("lazy compilation proxy {request:?} at {}", self.import_site)
```
mischnic: Why do you need this `import_site` disambiguation?
If you have multiple dynamic imports to the same module (i.e. same `request` string), then they could just use the same proxy module as well?

turbopack/crates/turbopack-ecmascript/src/manifest/chunk_asset.rs
```diff
@@ -94,14 +95,28 @@ impl ManifestAsyncModule {
                 .cell());
             }
         }
+        // The manifest module is synthesized while chunking, so it is not a member of
+        // `this.module_graph` and has to be placed in a graph of its own.
```
mischnic, line 99: The regular async loaders are also synthesized during chunking, so they are not in the graph either.
Can we do something similar here so that `ModuleGraph::isolated_async_entry` isn't necessary?

turbopack/crates/turbopack-core/src/chunk/chunking_context.rs
```diff
@@ -293,6 +293,16 @@ pub struct ChunkingConfig {
...
 
+/// The owner of a standalone HMR chunk list.
+#[turbo_tasks::task_input]
+#[derive(
+    Eq, PartialEq, Debug, Clone, Copy, Hash, TraceRawVcs, Serialize, Deserialize, Encode, Decode,
+)]
+pub enum HmrChunkListSource {
+    Entry,
+    Dynamic,
+}
```
mischnic: Please add documentation here what this means and what the effect in behavior is.

turbopack/crates/turbopack-ecmascript/src/references/esm/dynamic.rs
```diff
@@ -79,26 +90,117 @@ impl EsmAsyncAssetReference {
...
 #[turbo_tasks::value_impl]
 impl ModuleReference for EsmAsyncAssetReference {
     #[turbo_tasks::function]
     async fn resolve_reference(&self) -> Result<Vc<ModuleResolveResult>> {
-        if let Some(resolved) = &self.resolve_override {
-            return Ok(*ModuleResolveResult::module(*resolved));
+        if self.lazy_compilation && self.resolve_override.is_none() {
+            let request = self.request.await?;
+            // Pattern requests need their resolved result map for runtime dispatch. A constant
+            // request can use one proxy and defer resolution and processing until activation.
```
mischnic, line 126: This function seems overly complicated now.
I think you are trying to defer too much here.

The transforming/parsing of a module only happens when you actually call `module.references()` (among other methods). So the `esm_resolve(request)` call itself only performs resolving. It doesn't parse the module yet (and `LazyCompilationProxyModule::new_resolved` is still able to intercept the `references()` call to prevent parsing and further graph walking).

So I would just drop this first `if` statement and do everything via `LazyCompilationProxyModule::new_resolved` below (in the generic map-over-parseresult)
Then you don't need `request_string` and `import_site` or the `without_lazy_compilation().resolved_cell()` pattern either.

I don't think that performance massively changes if we do a few more resolve calls here.

turbopack/crates/turbopack-ecmascript/src/async_chunk/proxy.rs
```diff
@@ -0,0 +1,402 @@
...
+                Ok(Vc::cell(vec![ResolvedVc::upcast(
+                    reference.to_resolved().await?,
+                )]))
+            }
+        }
+    }
+
+    #[turbo_tasks::function]
+    fn side_effects(self: Vc<Self>) -> Vc<ModuleSideEffects> {
+        ModuleSideEffects::SideEffectFree.cell()
```
mischnic: Is it really side effect free? AFAICT, this module effectively does `module.exports = import("...");` so requiring this module can have side effects?

turbopack/crates/turbopack-ecmascript/src/manifest/loader_module.rs
```diff
@@ -145,21 +145,39 @@ impl EcmascriptChunkPlaceable for ManifestLoaderModule {
...
+                "#,
+                dynamic_id = StringifyModuleId(&dynamic_id),
+            )?;
+
+            return Ok(EcmascriptChunkItemContent {
+                inner_code: code.into(),
+                ..Default::default()
+            }
+            .cell());
+        }
```
mischnic, line 171: Why is this needed? If the module is already available (and thus no chunks need to be loaded), then we can just require the id?

mischnic (commented):
partial review
I'm still not sure convinced by the `proxy.key` -> ManifestAsyncModule::ident setup yet.

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -754,6 +754,23 @@ pub async fn project_update(
...
+    let ctx = &project.turbopack_ctx;
+    ctx.turbo_tasks()
+        .run(async move {
+            Ok(*activate_lazy_chunk_operation(chunk_path)
+                .read_strongly_consistent()
+                .await?)
+        })
+        .await
+        .map_err(|error| napi::Error::from_reason(PrettyPrintError(&error.into()).to_string()))
+}
```
mischnic, line 780: I think this would leak `activate_lazy_chunk_operation` tasks (until we have eviction) for every single chunk request, and be slower than necessary (and this is getting called for every single chunk request).
If we instead do a simple `chunk_path.contains("lazy-compilation")` preflight filter, it should impact most chunks at all. So doing that either on the JS side, or in `project_activate_lazy_chunk` before we go into the `turbo_tasks().run` closure.

turbopack/crates/turbopack-ecmascript/src/async_chunk/proxy.rs
jimmyhmiller: Good catch.

turbopack/crates/turbopack-ecmascript/src/references/esm/dynamic.rs
jimmyhmiller: So I tried this before. And I tried it again. Sadly esm_resolve does parse. Here is the full path I'm seeing empirically on the test I have for this

```
<turbopack_ecmascript::references::esm::dynamic::EsmAsyncAssetReference
    as turbopack_core::reference::ModuleReference>::resolve_reference
-> turbopack_resolve::ecmascript::esm_resolve
-> turbopack_resolve::ecmascript::specific_resolve
-> <turbopack::ModuleAssetContext
       as turbopack_core::context::AssetContext>::resolve_asset
-> <turbopack::ModuleAssetContext
       as turbopack_core::context::AssetContext>::process_resolve_result
-> <turbopack::ModuleAssetContext
       as turbopack_core::context::AssetContext>::process
-> turbopack::ModuleAssetContext::process_with_transition_rules
-> turbopack::ModuleAssetContext::process_default
-> turbopack::process_default
-> turbopack::process_default_internal
-> turbopack::apply_module_type
-> turbopack_ecmascript::module_fragments::part::module::
       EcmascriptModulePartAsset::select_part
-> turbopack_ecmascript::module_fragments::split_module
-> <turbopack_ecmascript::EcmascriptModuleAsset
       as turbopack_ecmascript::EcmascriptParsable>::failsafe_parse
-> turbopack_ecmascript::EcmascriptModuleAsset::parse
```

It ends up parsing the top level module making the top level no longer lazy. The files it references are. But not the top level. 

I've played around a bit with this. But I'm afraid I haven't found anything quite clean. There are so many paths like this were things just want to be eager. 

Maybe we can meet up and chat about this to come up with a good resolution? Not sure how strongly you feel about this. Definitely agree it isn't my favorite chunk of code. Any free time tomorrow?

mischnic: Your trace above is with the module gragments enabled, but the same could indeed be true even with the regular follow reexports mode as well.

The was that is supposed to work though is that `EcmaScriptModulesReferenceSubType::ImportPart(ModulePart::Export(name))` follows through exports.

But that would never have any benefit for `EcmaScriptModulesReferenceSubType::DynamicImport`. So this while thing should be unnecessary.

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -1221,7 +1227,12 @@ impl ChunkingContext for BrowserChunkingContext {
             .emit();
             return Ok(module.as_chunk_item(module_graph, *chunking_context));
         }
-        Ok(if self.await?.manifest_chunks {
+        let use_manifest = self.await?.manifest_chunks
+            && ResolvedVc::try_downcast::<Box<dyn EcmascriptChunkPlaceable>>(
+                module.to_resolved().await?,
+            )
+            .is_some();
+        Ok(if use_manifest {
```
mischnic, line 1237: Why was this necessary?

jimmyhmiller: https://github.com/vercel/next.js/pull/97203/changes/257564af402d9c8416c4797ed653f1a45a9998e2

I did try to pull this off into a separate commit. It was for the css (and other none js files) I found that they were breaking. I can't see this is the most elegant solution in the world... But it did seem to work. I tried to come up with a better predicate to make it clearer. If you have a better idea happy to follow up with it.

mischnic: Can you add a comment that dynamically importing e.g. CSS is the reason for that downcast?

turbopack/crates/turbopack-ecmascript/src/manifest/chunk_asset.rs
jimmyhmiller: I tried playing around with this. Did you mean maybe something like this? https://github.com/vercel/next.js/compare/jimmym/lazy-dynamic-imports-client...jimmym/standalone-manifest-chunk-prototype

turbopack/crates/turbopack-ecmascript/src/manifest/loader_module.rs
```diff
@@ -145,21 +145,39 @@ impl EcmascriptChunkPlaceable for ManifestLoaderModule {
...
+        // We also need the id of the module that we're actually trying to
         // dynamically import.
         let placeable =
             ResolvedVc::try_downcast::<Box<dyn EcmascriptChunkPlaceable>>(manifest.inner)
                 .ok_or_else(|| anyhow!("asset is not placeable in ecmascript chunk"))?;
         let dynamic_id = placeable.chunk_item_id(*manifest.chunking_context).await?;
 
-        // This is the code that will be executed when the dynamic import is reached.
+        if chunks_server_data.is_empty() {
+            writedoc!(
```
mischnic, line 158:
```suggestion
        if chunks_server_data.is_empty() {
        	// No chunks need to be loaded and thus the module is already available,
        	// so we can just require the module.
            writedoc!(
```

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -764,6 +764,27 @@ pub async fn project_update(
...
 
+#[tracing::instrument(level = "info", name = "activate lazy chunk", skip_all)]
+#[napi]
+pub async fn project_activate_lazy_chunk(
+    #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
+    chunk_path: RcStr,
+) -> napi::Result<bool> {
+    if !chunk_path.contains("lazy-compilation") {
+        return Ok(false);
+    }
```
mischnic: nit: use `ACTIVATION_KEY_MARKER` directly

or even better, move this pre-check to the JS side, then we have even less napi FFI operations

mainly because requesting chunks is a very hot operation, particularly if you have browser caching disabled in the devtools

turbopack/crates/turbopack-ecmascript/src/async_chunk/proxy.rs
```diff
@@ -0,0 +1,354 @@
...
+    LazyCompilationState {
+        active: State::new(false),
+    }
+    .cell()
+}
+
+#[turbo_tasks::value]
+#[derive(ValueToString)]
+#[value_to_string("lazy compilation target")]
+enum LazyCompilationTarget {
```
mischnic: This follows the existing convention

```suggestion
enum LazyCompilationReference {
```

turbopack/crates/turbopack-ecmascript/src/async_chunk/proxy.rs
```diff
@@ -0,0 +1,354 @@
...
+    ) -> Result<Vc<EcmascriptChunkItemContent>> {
+        let this = self.await?;
+        if estimated {
+            return Ok(EcmascriptChunkItemContent::default().cell());
+        }
+
+        let state = lazy_compilation_state(this.key.clone()).await?;
+        let inner_code = if state.is_active() {
+            let result = this.target.resolve_reference().await?;
+            let Some(module) = result.first_module().await? else {
```
mischnic: why do we not care about potential other modules in the resolve result here?

turbopack/crates/turbopack-ecmascript/src/async_chunk/proxy.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+}
+
+#[turbo_tasks::value]
+pub struct LazyCompilationProxyModule {
+    target: ResolvedVc<LazyCompilationTarget>,
+    ident: ResolvedVc<AssetIdent>,
+    /// The key that requesting this proxy's manifest chunk activates.
+    pub key: RcStr,
```
mischnic:
```suggestion
    pub(crate) key: RcStr,
```

turbopack/crates/turbopack-ecmascript/src/manifest/chunk_asset.rs
```diff
@@ -141,12 +163,22 @@ fn manifest_chunk_reference_description() -> RcStr {
...
             .ident()
             .owned()
             .await?
-            .with_modifier(manifest_chunk_reference_description())
+            .with_modifier(manifest_chunk_reference_description());
+        // Requesting the manifest chunk of a lazily compiled dynamic import is what activates it,
+        // so the key has to survive into the file name, and the path is the only part of an ident
+        // that appears there literally. It must not move to the proxy's own ident, which also
+        // names chunks that ship with the entrypoint and would activate the import on page load.
+        let Some(proxy) = ResolvedVc::try_downcast_type::<LazyCompilationProxyModule>(self.inner)
```
mischnic, line 176: non-blocking: We try to avoid downcasting in general, but I don't know a better option.

turbopack/crates/turbopack-ecmascript/src/async_chunk/proxy.rs
jimmyhmiller: Changing it to not go through a non-generic method. The way we have the LazyCompilationReference setup, there is only one. If we don't go through the generic resolve reference and just call a method on it (we know the type) we don't have to worry about these guards.

crates/next-napi-bindings/src/next_api/project.rs
jimmyhmiller: Moved it to the JS side

mischnic (approved):

PR #97215, opened 2026-08-12
Fix shared Turbopack runtime initialization race

gnoff:
## Summary

- Allow the shared browser runtime to initialize before any async chunk has created the chunk-loading queue.
- Preserve duplicate-runtime protection when a runtime registry is already installed.
- Update the affected development, production, debug-ID, and worker runtime snapshots.

Shared runtime scripts can execute before other async chunks. Previously, the runtime returned when the chunk-loading global was undefined, leaving later registrations queued and preventing the Next.js client bootstrap and hydration from running.

## Verification

- `cargo test -p turbopack-ecmascript-runtime`
- Targeted Turbopack browser-runtime snapshot fixtures
- Generated development and production runtimes evaluated with no pre-existing queue, including duplicate-runtime protection


dvoytenko (approved):

sampoder (approved):
lgtm! thank you!

PR #97232, opened 2026-08-12
Upgrade web-vitals to v6 and report soft navigations

timneutkens:
## Summary

Upgrades the vendored `web-vitals` dependency to v6.0.1 and enables soft-navigation reporting for CLS, INP, and LCP. This lets `useReportWebVitals` report metrics against the navigation where they occurred, including `navigationURL` for route attribution.

The upgrade also removes the obsolete FID integration and compatibility patch, regenerates the standard and attribution bundles, and updates the public metric types and documentation to match the current upstream API.

Related to #97161.

## Verification

- Added production-mode coverage for initial and soft-navigation Web Vitals reporting with both Turbopack and Webpack.


packages/next/src/client/web-vitals.ts
```diff
@@ -13,10 +12,9 @@ export function useReportWebVitals(
...
   useEffect(() => {
-    onCLS(reportWebVitalsFn)
-    onFID(reportWebVitalsFn)
-    onLCP(reportWebVitalsFn)
-    onINP(reportWebVitalsFn)
+    onCLS(reportWebVitalsFn, { reportSoftNavs: true })
+    onLCP(reportWebVitalsFn, { reportSoftNavs: true })
+    onINP(reportWebVitalsFn, { reportSoftNavs: true })
     onFCP(reportWebVitalsFn)
     onTTFB(reportWebVitalsFn)
```
tunetheweb: These should have `reportSoftNavs: true` too. Though TTFB just returns 0 so if you want to keep just hard nav TTFB, and avoid including these dummy 0 values, then you can.

test/e2e/app-dir/app/useReportWebVitals.test.ts
```diff
@@ -15,30 +16,80 @@ describe('useReportWebVitals hook', () => {
...
 
     // Make sure all registered events in performance-relayer has fired
-    await check(async () => {
-      expect(eventsCount).toBeGreaterThanOrEqual(6)
-      return 'success'
-    }, 'success')
+    await retry(() => {
+      expect(events.length).toBeGreaterThanOrEqual(5)
+    })
+    expect(events.some((event) => event.name === 'FID')).toBe(false)
```
tunetheweb: Should be INP?

packages/next/src/client/web-vitals.ts
timneutkens: :robot: Done -- enabled reportSoftNavs for FCP as well. I kept TTFB limited to hard navigations because web-vitals v6 reports dummy 0 values for soft-navigation TTFB.

test/e2e/app-dir/app/useReportWebVitals.test.ts
timneutkens: :robot: Done -- changed this to positively assert that INP is reported. The fixture now creates an interaction above the INP duration threshold, waits through the next paint, and triggers visibility finalization before polling the reported metrics.

packages/next/src/client/web-vitals.ts
tunetheweb: Cool. I'm not sure we've decided what CrUX is going to do here, but will try to remember to reach out when we do as this decision (while perfectly valid!) may result in a divergence from CrUX TTFB.

timneutkens:
@tunetheweb can you give it another review? Let me know if other changes are needed

packages/next/package.json
```diff
@@ -349,7 +349,7 @@
     "util": "0.12.4",
     "vm-browserify": "1.1.2",
     "watchpack": "2.4.0",
-    "web-vitals": "4.2.1",
+    "web-vitals": "6.0.1",
```
tunetheweb: Note we've done a few bug fixes as people have started using this in anger. We're up to v6.2.0. No API changes so probably worth going straight to that.

tunetheweb (commented):
LGTM apart from the suggestion to jump straight to latest v6 (6.2.0)

packages/next/package.json
timneutkens: :robot: Done — upgraded web-vitals to 6.2.0, refreshed the lockfile, and regenerated both vendored bundles. The focused production Web Vitals test passes with both Turbopack and Webpack.

eps1lon (approved):

PR #97238, opened 2026-08-12
Report Web Vitals for soft navigations

cramforce:
### What?

`useReportWebVitals` now reports the Core Web Vitals (LCP, CLS, INP) for soft navigations — client-side navigations that change the URL — in addition to the initial page load.

- Upgrades the vendored `web-vitals` from v4.2.1 to v6.1.0 (regenerated `src/compiled/web-vitals{,-attribution}`).
- Passes `{ reportSoftNavs: true }` to `onCLS`, `onINP` and `onLCP`.
- Drops `patches/web-vitals@4.2.1.patch`; the `readonly interactionId` fix it carried has landed upstream.
- Removes FID, which no longer exists in `web-vitals` (removed in v5), from `WEB_VITALS`, the `experimental.webVitalsAttribution` schema, and the docs.
- Documents `navigationURL` and the new `soft-navigation` `navigationType`.

### Why?

Core Web Vitals were only measured for full page navigations, so apps that navigate client-side lost measurement after the first page. Chrome 151 shipped the `PerformanceSoftNavigation` and `InteractionContentfulPaint` entries, and `web-vitals` v6 exposes them behind the `reportSoftNavs` option, which yields many more LCP samples and attributes CLS/INP to the route they happened on.

Implements https://github.com/vercel/next.js/discussions/97161.

### How?

`web-vitals` feature detects the Soft Navigations API, so browsers without support keep reporting for the initial page load only — no behavior change there. Where it is supported, metrics for a soft navigation are reported with `navigationType: 'soft-navigation'` and the `navigationURL` they were measured for, which is what consumers should attribute on since metrics are often reported after a later navigation already happened.

FCP and TTFB deliberately keep reporting for the initial page load only: a soft navigation has no server request, so `web-vitals` would report a TTFB of `0` for it.

Notes for reviewers:

- Two behavior changes for existing `useReportWebVitals` consumers: FID is no longer reported at all, and metrics for the initial URL are now finalized when the first soft navigation happens rather than when the page is unloaded. `experimental.webVitalsAttribution: ['FID']` now fails config validation.
- Covered by a new unit test for the registered options and an extended e2e test. The soft-navigation assertions in the e2e test are gated on browser support, because Playwright 1.61 pins Chromium 149 — they will start running once Playwright rolls to 151+.
- #97232 implements the same discussion in parallel; only one of the two should land.

Fixes #97161


notacouch:
https://github.com/vercel/next.js/pull/97232 has been merged, this can be closed out.

PR #97251, opened 2026-08-12
Skip the initial history write on any activation-key mismatch

gaearon:
When the router hydrates, it does a `replaceState` to the current history entry with the tree it just rendered and a marker (`__NA`) saying the entry is ours. Normally the current entry is the one the document loaded on, so that's fine.

#96252 fixed one case where the current entry is *not* the document entry.

With the fix, it *also added a bunch of new tests* for this case, but they seem to be flaky.

It turns out that the fix wasn't *broad enough*, and the new tests were surfacing *an even older* issue.

This PR seems to improve the flakiness by broadening the fix to cover that older issue. Details below.

## What motivated #96252

Suppose you navigate from A to B inside the app, reload B, and press Back before B hydrates. The URL goes back to A (and the browser treats it as `popstate`), but the router isn't running yet, so nothing responds and the page still shows B. Then the router starts, and its first write would put B's tree on A's entry, breaking Back/Forward navigation.

The fix in #96252 was to *skip the history write* in that case, and also to replay `handlePopState` the router missed.

That's two decisions, and #96252 used one condition for both: we're not on the entry the document loaded on, *and* the entry has our marker (`__NA`) on it. However, these two decisions ask different questions:

- Replaying asks whether we can recover the missed Back. Recovering it means restoring the tree the entry was left with, and only our own entries have one, so we *do* need to check them for `__NA`.

- Skipping the write asks whether what we rendered describes the entry we're about to write to. If it isn't the entry we loaded on, our tree doesn't belong on it, no matter who created it. So the `__NA` check **isn't relevant here**.
  - That's what this PR fixes — it removes the `__NA` check here.

This fixes a preexisting issue described below.

## When is this distinction relevant?

For the case #96252 aimed to fix, it is not relevant.

However, it is relevant when something other than the router changes history before hydration.

Some examples:

1. You click an in-page anchor (`/post` → `/post#section`) before hydration.
2. Or a script runs a `pushState` of its own (`/post` → `/post?tp=1`) before hydration.

Either way the current entry is no longer the document's entry, and since the router didn't create it, it has no `__NA`.

To solve the flakes, we need to *broaden the fix* to include these two cases.

Behavior matrix:

| what happens before hydration | 16.3.0 and canary | this PR |
| --- | --- | --- |
| anchor click to `#section` | we overwrite the entry's `null` state with ours | we leave the entry alone |
| script pushes `?tp=1` | we merge our marker and tree into their state | we leave the entry alone |
| script pushes `?tp=1` while the page is hydrating | **`?tp=1` is thrown away** | `?tp=1` survives |

The main observable behavior (predating the Back fix) is the last row (`pushState` query getting thrown away). This is the preexisting bug exposed by the new tests, that we are fixing in this PR.

Starting with #96252, we were also doing an unnecessary no-op `handlePopState` for these cases, because we put `__NA` on the entry with a write and then immediately checked for it. 

This PR actually fixes these cases (by not doing `replaceState`).

As a result:

- We no longer overwrite these entries.
- Our tests, which verify that, no longer have races caused by hydration.

## Testing

The first commit adds the assertions for all three rows. 10 of the 18 tests fail without the second commit and all 18 pass with it.

### DIFF
--- a/packages/next/src/client/components/app-router.tsx
+++ b/packages/next/src/client/components/app-router.tsx
@@ -56,13 +56,10 @@ const globalMutable: {
   pendingMpaPath?: string
 } = {}
 
-// A Back/Forward press before the router's popstate listener exists moves the
-// browser to a different history entry than the one the document was activated
-// on, and the resulting popstate fires with nobody listening. The activation
-// entry is fixed for the document's lifetime and entry keys are stable across
-// replaceState, so until the listener is installed a key mismatch means a
-// traversal went unobserved.
-function hasMissedTraversal(): boolean {
+// The current entry is not the one this document was activated on. Entry keys
+// are stable across replaceState, so until the popstate listener is installed
+// this means a traversal or history write went unobserved.
+function isOnUnobservedEntry(): boolean {
   if (typeof window.navigation === 'undefined') {
     return false
   }
@@ -71,10 +68,7 @@ function hasMissedTraversal(): boolean {
   return (
     activationEntry != null &&
     currentEntry != null &&
-    activationEntry.key !== currentEntry.key &&
-    // Only entries written by the app router can be restored; on any other
-    // entry the traversal is left unhandled, as before.
-    window.history.state?.__NA === true
+    activationEntry.key !== currentEntry.key
   )
 }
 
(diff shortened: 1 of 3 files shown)


gaearon:
I'm gonna rework this a bit.

Issue #123, opened 2016-10-28
Please tone down the error screen


christiannaths:
Beauty is in the eye of the beholder, unfortunately my retinas are burned out...

Seriously though, is it a jerk move to ask for some basic accessibility on the error screen? This is "exceptionally" hard on the eyes (pun intended). After only second or two on this screen, my vision is temporarily discoloured, making it hard to work. Happy to do a PR if I get a thumbs-up from a maintainer.

![screen shot 2016-10-27 at 8 47 04 pm](https://cloud.githubusercontent.com/assets/11584/19793047/e78b1bce-9c87-11e6-9cde-8c053025453b.png)

nodegin:
Just did some modification

![](https://cloud.githubusercontent.com/assets/8536244/19793434/31915b44-9cff-11e6-85bb-6d63b4bb86df.png)

christiannaths:
Wups, just did a PR. Either yours or mine is much better 👍

christiannaths:
Closed mine in favour of yours. Thanks!!!

nkzawa:
👍  I think it's reasonable. 
@rauchg thought ?

rauchg:
Responded in the PR. Very reasonable

PR #97252, opened 2026-08-12
Add a script for adopting fork pull requests

eps1lon:
Fork PRs run without repository secrets, so deploy tests never run on external contributions. The workaround is to re-push the contributor's commits to a branch in `vercel/next.js` and open a replacement PR from there, which is fiddly by hand and carries a real security risk. This adds a `pr-adopt` PNPM script to encode the procedure.

```bash
pnpm pr-adopt <pr-number>            # adopt
node pr-adopt <pr-number> --dry-run  # report without pushing
```

We copy over the title and body as-is and prepend a `Adopts #N. Closes #N.` line so that the underlying PR automatically closes and we create the proper backlinks.

Commits are signed

The description is copied rather than rewritten for two reasons: it is the contributor's own text, and any `Fixes #123` inside it has to survive, since the original PR closes unmerged and so never fires its own.

### The confirmation gate

Adoption runs the contributor's code on a branch where CI has repository secrets, which is exactly what a fork PR is denied. Anything executing during install, build, or test can read and exfiltrate them, so before touching git the script prints the PR link, status, author handle and association, diff size, and every file touched. Confirming requires retyping the author's handle, and a non-interactive stdin is refused rather than silently proceeding.

Two details worth a reviewer's attention:

- **The file list is deliberately unranked.** An earlier revision flagged "high-risk" paths (`.github/`, `package.json`, lockfiles). That was removed: a payload can sit in any fixture or source file, and printing "no high-risk paths touched" manufactures false assurance.
- **The list comes from the paginated REST endpoint.** `gh pr view --json files` silently caps at 100 entries. #96550 has 106 changed files and returns 100, which would present a truncated list as the complete surface. If the count still falls short of `changedFiles` (GitHub stops at 3000), the shortfall is reported rather than hidden.

Draft and closed PRs are both adoptable, since a contributor may still be iterating or may have abandoned an unreviewed change; the status is surfaced rather than enforced. Merged PRs are refused because their commits are already in `canary`.

## Test plan

```console
$ pnpm pr-adopt 96525
Adopting #96525 - [testmode] Fix infinite recursion in testmode passthrough fetch
  Link      https://github.com/vercel/next.js/pull/96525
  Status    OPEN
  Author    @lazerg (CONTRIBUTOR)
  Source    lazerg:fix/issue-96521-testmode-passthrough-recursion
  Diff      2 files, +19 -2, across 1 commit(s)
  Push to   origin (vercel/next.js) as adopt/96525
  Base      canary

  Files touched (2):
    packages/next/src/experimental/testmode/fetch.ts
    test/e2e/testmode/testmode.test.ts

  This grants the contributor code access to secrets.

  A fork PR is denied repository secrets because its code is not trusted.
  Adopting runs that same code on a branch inside vercel/next.js, where CI has
  those secrets. Anything in this diff that executes during install,
  build, or test can read and exfiltrate them. The dangerous line does
  not have to be in the change the PR claims to make, and no file in the
  list above is safe by virtue of where it sits.

  Read the full diff:  gh pr diff 96525 --repo vercel/next.js

  Type the author's handle (lazerg) to confirm you have read
  the full diff and vouch for this code: lazerg
Opening the replacement pull request

Adopted.
  Original  https://github.com/vercel/next.js/pull/96525
  Adopted   https://github.com/vercel/next.js/pull/97257
```


scripts/adopt-pr.js
```diff
@@ -0,0 +1,711 @@
...
+}
+
+async function main() {
+  const { prNumber, dryRun } = parseArgs(process.argv.slice(2))
+  const branch = `adopt/${prNumber}`
+
+  const remote = await detectUpstreamRemote()
+  const pr = await fetchPullRequest(prNumber)
+
+  await preflight(pr, branch, dryRun)
```
ztanner, line 677: I think it'd be worth rejecting if `headRefOid` changes from the time the commands run to when the PR information is fetched, just in case the underlying ref changes and now the contributor is adopting a PR that changed underneath them which could have security implications.  Probably should even confirm with the adopter the sha of what they are fetching before proceeding

scripts/adopt-pr.js
```diff
@@ -0,0 +1,711 @@
...
+  }
+
+  await confirmAdoption(pr, remote, branch)
+
+  const originalBranch = await capture('git', ['branch', '--show-current'])
+
+  console.log(bold(`Checking out #${pr.number} as ${branch}`))
+  // Fetches refs/pull/<n>/head, so the fork does not need to be a remote. The
+  // commits are not rewritten: authorship has to reach the merge intact.
+  await runInherit('gh', [
```
ztanner: `gh pr checkout` can execute contributor-added hooks locally (for example, `.husky/post-checkout`). I think adoption should run with hooks disabled for all Git/rebase subprocesses. It may also be worth using a throwaway repo so the untrusted checkout and any failed rebase state never touch the maintainer’s working copy.

scripts/adopt-pr.js
```diff
@@ -0,0 +1,711 @@
...
+
+  const unsigned = []
+
+  for (const sha of log.split('\n')) {
+    const raw = await capture('git', ['cat-file', 'commit', sha])
+    const lines = raw.split('\n')
+    const headerEnd = lines.indexOf('')
+    const headers = headerEnd === -1 ? lines : lines.slice(0, headerEnd)
+
+    if (!headers.some((line) => line.startsWith('gpgsig'))) {
```
ztanner, line 594: This may be an edge case, but a commit can contain a `gpgsig` header without GitHub considering the signature verified. (ie an unknown key or malformed signature). Should we re-sign unless GitHub reports it as verified?

eps1lon: We can follow-up if that's an issue.

scripts/adopt-pr.js
eps1lon: This is out of scope here. You approve before any of that runs. If we're concerned about this, we should reject the script outright.

eps1lon: I'll see if we can disable the hooks but if that's your threat model, adopting a PR cannot work.

ztanner: I don’t think this makes adoption impossible. The intended trust grant is to CI after review, not to contributor-controlled hooks on the maintainer's machine. Disabling hooks during checkout/rebase/push seems like cheap defense in depth in case review misses something.

scripts/adopt-pr.js
eps1lon: we push the original commit that was approved on now.

scripts/adopt-pr.js
eps1lon: all the commands now run with git hooks disabled.

ztanner (approved):

Issue #96657, opened 2026-08-04
16.3.0: `output: 'standalone'` + adapter crashes `next build` — #93684 stops emitting next-server.js.nft.json but copyTracedFiles still requires it


bryan-ferry:
### Summary

`next build` on **16.3.0 (Turbopack)** crashes with `ENOENT .next/next-server.js.nft.json` whenever **`output: 'standalone'` and a build adapter (`adapterPath` / `NEXT_ADAPTER_PATH`) are both set**.

Cause (analysis below): [#93684](https://github.com/vercel/next.js/pull/93684) made Turbopack skip emitting `next-server.js.nft.json` / `next-minimal-server.js.nft.json` when an adapter is configured ("Adapters don't read these files") — but the `output: 'standalone'` finalize step (`copyTracedFiles`) still reads `next-server.js.nft.json` unconditionally, with no guard. Writer suppressed, reader kept → every standalone+adapter build crashes.

**Real-world impact:** on Vercel, the platform is rolling out adapter-based builds (`@vercel/next` injects `NEXT_ADAPTER_PATH` when the platform-side `NEXT_ENABLE_ADAPTER=1` flag is on). Any app with `output: 'standalone'` in its config (common for teams that also deploy via Docker) fails to deploy on 16.3.0, while 16.2.12 works (the 16.2 branch never received the #93684 gate). We hit this in production the day after 16.3.0's release; nothing in the config changed between the working 16.2.12 deploy and the failing 16.3.0 one.

### Minimal reproduction (4 files, no Vercel code)

```json
// package.json
{
  "name": "repro",
  "private": true,
  "dependencies": { "next": "16.3.0", "react": "19.2.5", "react-dom": "19.2.5" }
}
```

```js
// next.config.js
module.exports = { output: 'standalone' }
```

```js
// pages/index.js
export default function Home() {
  return 'hello'
}
```

```js
// adapter-stub.cjs — minimal adapter, does nothing
module.exports = { name: 'repro-stub', async onBuildComplete() {} }
```

```bash
npm install

# control — passes, and .next/next-server.js.nft.json exists afterwards:
npx next build

# failing — exact same app, adapter set:
rm -rf .next && NEXT_ADAPTER_PATH="$PWD/adapter-stub.cjs" npx next build
```

Output of the failing run:

```
  Running onBuildComplete from repro-stub
> Build error occurred
Error: ENOENT: no such file or directory, open '.../.next/next-server.js.nft.json'
```

[...]

PR #97287, opened 2026-08-13
Emit whole-app server NFTs when `output: 'standalone'` is used with an adapter

### LINKED ISSUE #96646: output: 'standalone' breaks Vercel deployments on Next 16.3 — ENOENT .next/next-server.js.nft.json in onBuildComplete

https://github.com/jonymusky/next163-standalone-repro

### To Reproduce

 To Reproduce:
  1. npx create-next-app -e reproduction-template — the only change is adding output: 'standalone' to next.config.ts (see repo, single commit)
  2. Deploy to Vercel (vercel deploy)
  3. Build compiles and generates static pages fine, then fails at Running onBuildComplete from Vercel:

  > Build error occurred
  Error: ENOENT: no such file or directory, open '/vercel/path0/.next/next-server.js.nft.json'

  Expected: build succeeds (it does on next 16.2.11 with identical config, and locally on 16.3 with next build). Current: deterministic failure on Vercel.
  Reproduced on next@16.3.1-canary.1 (the reproduction template's default) and on 16.3.0; also hit in a production pnpm monorepo across 3 independent apps. Removing
  output: 'standalone' fixes every case immediately.

```bash
next 16.3.1-canary.1 / 16.3.0 (Turbopack), react 19.2, Vercel git + CLI deployments, Node 24.x build container.
```

Output

Vercel (Deployed)

### Additional context

.next/next-server.js.nft.json was emitted in ≤16.2.x with standalone output; 16.3 appears to no longer write it (or writes it elsewhere), and
  Vercel's onBuildComplete still expects it. Workaround we're shipping: output: process.env.VERCEL ? undefined : 'standalone'.

bryan-ferry:
Fixes #96646

### Problem

Since v16.3.0, `next build` crashes for any app that combines a deployment adapter with `output: 'standalone'`:

```
Error: ENOENT: no such file or directory, open '<distDir>/next-server.js.nft.json'
```

#93684 stopped emitting the whole-app server NFTs whenever an adapter is configured ("Adapters don't read these files" — true for adapters, which consume the per-endpoint NFTs in `build-complete.ts`). But the whole-app NFT has a second, adapter-independent consumer: the `output: 'standalone'` finalizer. `copyTracedFiles` reads `distDir/next-server.js.nft.json` unconditionally whenever standalone output is requested — unlike its sibling per-page trace reads, it has no `.catch`. The build runs adapter finalize and standalone finalize in sequence, so with both configured the Rust side skips the writer while the JS side still runs the reader → raw ENOENT.

This breaks every Vercel deployment of a standalone-configured app on 16.3 (the builder injects its adapter via `NEXT_ADAPTER_PATH` under the `NEXT_ENABLE_ADAPTER` rollout), and — per the reports on #96646 — also self-hosted/AWS `cdk-nextjs` users, for whom there is **no config-level escape**: that adapter force-sets `output: 'standalone'` and the construct requires `.next/standalone`, so both sides of the conflict are mandatory.

The combination was never covered by CI: `test/production/next-server-nft/next-server-nft.test.ts` has a `with output:standalone` suite and a `with adapters` suite, but no combined suite — the only state where the writer gate and the reader disagree.

A 2-line stub adapter (`{ name, async onBuildComplete() {} }`) + `output: 'standalone'` + `pages/index.js` reproduces it; the same config passes on 16.2.x. (Originally reported with the full root-cause chain in #96657, which the repro-link bot closed; #96646 is the surviving tracker.)

### Fix

Re-scope the #93684 gate instead of removing it: skip the whole-app server NFTs only when an adapter is configured **and** standalone output is not.

- `is_standalone` was already computed in this exact function two lines below the gate — the change is a reorder plus one condition, no new plumbing.
- Adapter-only builds keep #93684's behavior (no whole-app NFTs).
- Adapter + standalone falls through to the existing emission branches (`has_next_support && !is_standalone` → else emits `Minimal` + `Full`), restoring pre-#93684 behavior for exactly the broken combination.
- Precedent: #94197 already walked this gate's assumption back once when another consumer (cacheHandler tracing) surfaced.

Additionally, `copyTracedFiles` gets the same `.catch` + `Log.warn` treatment its sibling per-page reads already have, so any future writer/reader drift degrades into one actionable warning instead of a raw ENOENT.

#### Why the guard alone would not be a fix

Measured on a production monorepo app (16.3.0, Turbopack, standalone, Vercel adapter injected): with only the JS guard, the build exits 0 but `.next/standalone` is missing **1017 of 2133 files (48%)** versus a no-adapter control — including `next/dist/server/next.js`, `next-server.js` and the rest of the server runtime — and `node server.js` fails immediately with `Cannot find module '.../node_modules/next/dist/server/next.js'`. A silently unbootable standalone directory would trade a loud build failure for a runtime one; restoring emission is the fix, the guard is defense-in-depth.

### Test

Added the missing combined suite (`with adapters and output:standalone`) to `next-server-nft.test.ts`, reusing the existing `my-adapter.mjs` fixture: asserts the build completes, both whole-app NFTs exist and are non-trivial, and the adapter still ran (behavioral assertions only — no new snapshots).

Verified against the minimal repro on 16.3.0: control build passes with the pair present; stub-adapter build crashes with the exact ENOENT; with the gate re-scoped the pair is emitted and the build completes.

### Out of scope (follow-up worthy)

`adapterPath: ''` currently disables the adapter in JS (truthiness gates) while enabling the NFT skip in Rust (`adapter_path.is_some()` → `Some("")`), reproducing this same crash with no adapter at all. Normalizing `''` to `undefined` during config resolution (or `z.string().min(1)`) would make both sides agree; kept out of this PR to stay minimal.


packages/next/src/build/utils.ts
```diff
@@ -1401,7 +1407,13 @@ export async function copyTracedFiles(
...
   }
 
-  await handleTraceFiles(path.join(distDir, 'next-server.js.nft.json'))
+  const nextServerTraceFile = path.join(distDir, 'next-server.js.nft.json')
+  await handleTraceFiles(nextServerTraceFile).catch((err) => {
+    if (err.code !== 'ENOENT') throw err
+    Log.warn(
+      `Failed to copy traced files: ${nextServerTraceFile} was not emitted. The standalone output will be missing the Next.js server's traced dependencies.`
+    )
+  })
```
mischnic: Let's keep the existing strict behavior.
The rust-side fix should prevent this in the first place

bryan-ferry: Done — dropped the guard in f64f2ab, `copyTracedFiles` keeps its strict behavior. The PR is now just the Rust re-scope + the combined `adapters + output:standalone` test suite.

mischnic (approved):

mischnic:
I'm afraid you'll need to rebase this branch using signed commits, otherwise I can't merge this: https://github.com/vercel/next.js/blob/canary/contributing.md#signed-commits

bryan-ferry:
Rebased — the branch is now a single signed commit (cfa6f6c) with the same content: the Rust re-scope + the combined test suite. (CI already passed on this exact content pre-rebase.)

PR #97288, opened 2026-08-13
fix(turbo-persistence): allow 32-bit usize conversion

hardfist:
### What?

Change the pointer-width assertion in `usize_from_u32` from `<` to `<=`, allowing the conversion to compile on 32-bit targets.

### Why?

The surrounding comment states that Turbopack assumes a pointer width of at least 32 bits, but the strict comparison rejects exactly 32-bit targets. Since `u32` and `usize` have the same width on these targets, the conversion remains lossless.

This currently causes `rspack-turbo-persistence` to fail to compile with `error[E0080]` in Rspack's [`i686-pc-windows-msvc` CI job](https://github.com/web-infra-dev/rspack/actions/runs/31665978273/job/94340544786#step:25:446).

### How?

Allow equal source and target bit widths in the compile-time assertion.


bgw (approved):
Thank you!

PR #97309, opened 2026-08-13
[PPF] Instant validation for unstable_navigation()

lubieowoce:
Adds discriminated error messages for `unstable_navigation()` used outside of Suspense.

Previously we only had to discriminate two kinds of errors in each validation flow:
- App Shell -- 1. link data (if it resolves in `Runtime`) or 2. dynamic data (if it resolves in `Dynamic`)
- Static Shell -- 1. runtime data (if it resolves in `Runtime`) or 2. dynamic data (if it resolves in `Dynamic`)

However, with `navigation()` (96908), the App Shell flow needs to distinguish three kinds: 1. link data, 2. navigation (if it resolves in `NavigationRuntime`), and 3. dynamic data. This complicates the validation code a bit, because we can no longer get away with one retry. To accommodate this, i restructured the validation code to no longer recurse as a method of retrying. We just loop over an array that defines what order we should try the stages in + what kind of hole shows up in each stage.

I've also removed `hasAmbiguousErrors` and the associated logic. since we're no longer mixing Static+Runtime segments in a runtime prefetch, this is no longer relevant -- all segments used for validation use one stage, so whether or not the error is ambiguous only depends on the stage.

For ease of reviewing, this is split into two commits:
1. updating the (many) places that need a new `DynamicHoleKind` in our error messages. This one has a lot of pretty mechanical changes, and was also machine-reviewed for consistency already, but the messages themselves are worth looking at (we'll also need to update docs -- leaving that for a follow-up)
2. the actual instant validation changes, i.e. using the new `DynamicHoleKind` and adding the new pass


packages/next/src/server/app-render/instant-validation/instant-validation.tsx
```diff
@@ -983,11 +990,6 @@ export function discoverValidationDepths(loaderTree: LoaderTree): number[] {
  */
 export type ValidationPayloadResult = {
   payload: InitialRSCPayload
-  /** Whether errors from this payload could be ambiguous between runtime
-   * API access (cookies, headers) and uncached IO (connection, fetch).
-   * True when some segments used Static stage. False when all segments
-   * used Runtime stage and errors are definitively from uncached IO. */
-  hasAmbiguousErrors: boolean
```
lubieowoce, line 990: we no longer mix Static and Runtime, so all new-subtree segments are always one stage, and this isn't necessary

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -7355,38 +7362,141 @@ async function validateInstantConfigs(
 
   const { implicitTags, nonce, workStore, isDebugChannelEnabled } = ctx
 
+  type RetryStage = RenderStage.Runtime | RenderStage.NavigationRuntime
+
+  type ValidationSequence = {
+    stageOrder: PrefetchedSegmentStage[]
+    holeResolution: Record<SegmentStage, DynamicHoleKind | null>
+  }
+  const validationSequences = {
```
lubieowoce, line 7381: making it this generic may seem like overkill,  we're gonna need another variant for `StaticShell` soon (when we do `requireStatic`), which is different from `LegacySpeculative`, so it's gonna be useful soon

and making this parametrizable is a step to making it dynamically adjustable to optimize the work. e.g. in theory we could easily detect:
- "we don't have any NavigationRuntime chunks, so anything that's a hole in Runtime has to be Dynamic (it can't be a navigation())
- "we don't have any Runtime or NavigationRuntime chunks, so any hole we see in ShellRuntime has to be Dynamic (i.e. there's nothing to discriminate)

and we can express that by modifying an object like this appropriately before running the discrimination sequence over it

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -7355,38 +7362,141 @@ async function validateInstantConfigs(
...
+        // The narrowed validation found errors to report.
+        return narrowedResult
+      }
+    }
+
+    // If we didn't return some other errors at this point the only thing to return is this validation's result
+    return initialResult
   }
 
-  async function validateAtDepthImpl(
```
lubieowoce, line 7375: i've untangled this a bit. `validateAtDepthImpl` was calling itself recursively and passing a previousBoundaryState to detect that it's a retry, which was kinda confusing to follow.

unpacked into `validateAtDepth` that coordinates runs/retries + `validateOrRetryAtDepth` that runs a single pass

packages/next/src/server/app-render/dev-validation-scheduler.ts
```diff
@@ -79,9 +79,9 @@ export function beginDevValidation(
  * Give incoming requests a chance to enter app rendering and supersede the
  * current validation before another expensive render attempt starts.
  *
- * The regular global `setImmediate` is patched by staged rendering and can run
- * inside the current timer task. The original immediate is required here so we
- * actually pass through the event-loop poll phase where HTTP requests arrive.
+ * @returns Whether validation should continue.
+ * - `true`: validation should continue
+ * - `false` the validation was superseded and should be aborted.
  */
```
lubieowoce, line 85: i swear this is backwards from how i'd write it (if i'm yielding to the request, why am i continuing?). this pattern should be something like
```ts
if (await wasSuperseded()) { return }
```
and this requires doing the opposite of that, which confused me enough that i had to document it

gnoff (approved):

PR #97312, opened 2026-08-13
[backport] Allow literal exports in `'use cache'` files

unstubbable:
Backports #97181 ("Allow literal exports in `'use cache'` files") to `next-16-3`.

A `'use cache'` page or layout that exports a route segment config literal, such as `export const instant = false`, `export const dynamicParams = true`, `export const prefetch = 'partial'`, or `export const maxDuration = 5`, fails the build with "Only async functions are allowed to be exported in a "use cache" file".

Rejecting exported literals is meaningful only in a `'use server'` file, where every export becomes a server reference and a runtime check asserts that each one is a function. A `'use cache'` file wraps only exports that are, or might be, functions, so known non-function values can pass through. The check is now scoped to `in_action_file`.

Cherry-picked from `c0c64a0ab1` with no conflicts. All source files are byte-identical to canary. This touches `crates/next-custom-transforms`, so the release build needs the native binary rebuilt. Behavioral verification is left to CI on this branch; the change carries transform fixtures on both the server and client graph.

### DIFF
--- a/crates/next-custom-transforms/src/transforms/server_actions.rs
+++ b/crates/next-custom-transforms/src/transforms/server_actions.rs
@@ -1908,15 +1908,20 @@ impl<C: Comments> VisitMut for ServerActions<C> {
                                 let mut has_export_needing_wrapper = false;
 
                                 for decl in &var.decls {
-                                    if let Pat::Ident(_) = &decl.name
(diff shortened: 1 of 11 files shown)


acdlite (approved):

PR #97315, opened 2026-08-13
[backport] Keep the dev validation worker alive across HMR updates

unstubbable:
Backports #96988 ("Keep the dev validation worker alive across HMR updates") to `next-16-3`.

Cache Components dev validation reported stack frames that pointed at build output whenever a module had been updated while the dev server ran. The overlay showed a raw `file:` URL, the terminal named the chunk rather than the page, and because the frame never resolved to a source position there was no code frame either, so nothing indicated which line caused the error. This affected both the static shell validation and the instant-navigation validation, since both run on the same worker.

Turbopack's server HMR evaluates an updated module as a script of its own, carrying its source map inline rather than on disk, so only the isolate that ran that `eval` can resolve a frame in it. The validation worker never ran it. The worker now replays the applied updates, the cleared manifest cache entries, and the evicted paths that the dev server reports, in the same order the dev server applied them, which leaves each updated module's inline source map in the worker's own Node.js cache.

Not dropping the worker on every edit also removes a worker respawn and a `loadComponents` call from each validation that follows an edit, which is paid at exactly the moment the user is waiting for the insight. The case in the test suite that covers this went from around 870ms to around 240ms.

This affects Turbopack dev with Cache Components, which is on by default unless `experimental.devValidationWorker` is set to `false`.

Cherry-picked from `f70564f742` with no conflicts. All source files are byte-identical to canary. Behavioral verification is left to CI on this branch; the change carries the `dev-validation-worker-source-maps` suite, whose snapshots turn from a frame naming the chunk into resolved frames.

### DIFF
--- a/bench/dev-validation/README.md
+++ b/bench/dev-validation/README.md
@@ -28,6 +28,17 @@ prints the speedup. Options:
 - `--port=<n>`, `--headless=false`, `--settle-ms=<n>`.
 - `--json-out=<path>` — write the raw stats as JSON.
 
+Set `BENCH_DEV_VALIDATION_INSIGHTS=1` to give every family's leaf page an
+uncached data access, so validation reports one insight per navigation. Without
+it the run only covers the validation renders; with it, it also covers what
+follows an insight — encoding the errors and printing them with a source-mapped
+stack and code frame — which is the part that moves when that printing changes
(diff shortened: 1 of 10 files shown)


acdlite (approved):

PR #97323, opened 2026-08-13
fix(next/image): reject empty image on read/write to disk cache (supersedes #97278)

styfle:
Supersedes #97278 — same change, with the `errors.json` merge conflict resolved correctly so CI passes. If you'd rather keep #97278, cherry-pick the single commit here onto `codex/reject-empty-image-cache-buffers` instead and close this.

The image-optimizer change is unmodified from #97278 (@styfle). The only difference is two digits in `errors.json`.

---

## Why #97278's CI was failing (not a flake)

Four jobs failed — `build`, `build / build`, `build-next / build`, `test devlow package / build` — all with the same final error, and everything downstream was left skipped:

```
[19:25:59] Finished compile in 15.61s
[19:25:59] Starting check_error_codes
check_error_codes FAILED: There are new errors introduced but no corresponding
error codes are found in errors.json file, so make sure you run `pnpm build` or
`pnpm update-error-codes` and then commit the change in errors.json.
 Failed:    next#build
```

`compile` succeeded — only `check_error_codes` failed, which fails `next#build`, which fails every workflow that builds `next`. Deterministic, reproducible, one root cause.

### Root cause

The `Merge branch 'canary'` commit (827c8fc) on that branch left **two `"1470"` keys** in `packages/next/errors.json`:

```json
  "1469": "The payload's transport tree is missing a slot that exists in the loader tree: %s",
  "1470": "`deflateResumeDataCache` should not be called in edge runtime.",
  "1470": "Invariant: image cache entry \"%s\" is empty",
  "1471": "Invariant: cannot write an empty buffer to the image cache"
```

The original commit (bb604fa) correctly claimed 1470/1471 when canary's max code was 1469. Canary then landed its own `1470` (`deflateResumeDataCache`), and the merge kept both lines rather than renumbering.

Duplicate keys are legal JSON but **lossy** — last value wins. Both consumers parse into a map:

- `next-error-code-swc-plugin` → `serde_json` into `FxHashMap<String, String>` (`crates/next-error-code-swc-plugin/src/lib.rs:415`)
- the consolidation step → `JSON.parse` (`packages/next/check-error-codes.js:22`)

so `1470` resolves to `"Invariant: image cache entry ..."` and canary's `deflateResumeDataCache` message **disappears from the map entirely**. During the build the plugin can't find a code for that `InvariantError` (`packages/next/src/server/resume-data-cache/resume-data-cache.ts:207`), spills it to `packages/next/.errors/<hash>.json`, and `check-error-codes.js` exits 1 as soon as that directory exists.

Measured on the #97278 head blob — the file silently loses an entry when parsed:

```
physical keys: 1472  |  parsed keys: 1471
1470 => "Invariant: image cache entry \"%s\" is empty"     # deflateResumeDataCache is gone
```

### Why the merge driver didn't prevent it

The repo registers a merge driver for exactly this case (`packages/next/errors.json merge=errors-json` in `.gitattributes`). Replaying it by hand on the three real blobs resolves cleanly:

```
$ node scripts/merge-errors-json/merge.mjs current.json base.json theirs.json
merge-errors-json: added 1 new message to errors.json     # exit 0, no duplicate keys
```

So the driver wasn't active for that merge — either it was resolved on GitHub's servers (server-side merges don't run local merge drivers) or `scripts/merge-errors-json/install` hadn't been run in that clone.

### The fix

```diff
   "1470": "`deflateResumeDataCache` should not be called in edge runtime.",
-  "1470": "Invariant: image cache entry \"%s\" is empty",
-  "1471": "Invariant: cannot write an empty buffer to the image cache"
+  "1471": "Invariant: image cache entry \"%s\" is empty",
+  "1472": "Invariant: cannot write an empty buffer to the image cache"
```

This deliberately differs from what the merge driver would have produced — the driver keeps ours-side numbering and pushes `deflateResumeDataCache` out to 1472. Error codes are user-facing and `E1470` already shipped on canary meaning `deflateResumeDataCache`, so reassigning it would change a published code's meaning and make the merge back into canary non-additive. Keeping 1470 fixed is also what `pnpm update-error-codes` yields from a clean rebase onto canary, since `nextInitialCode` becomes `max(1470) + 1`.

Verified: 1472 physical keys, 1472 parsed keys (lossless), zero duplicate keys, and all three messages present as map values so every plugin lookup resolves. The 46 duplicate *messages* in the file are pre-existing on canary and unrelated.


styfle:
Closing as redundant — @styfle pushed the same fix directly to #97278 (5f562be, "fix merge conflict"). The resulting tree there is byte-identical to this branch, and all four previously-failing build jobs are now green on #97278:

| job | before | after |
| --- | --- | --- |
| `build` | fail 2m22s | **pass** 3m15s |
| `build / build` | fail 4m34s | **pass** 6m30s |
| `build-next / build` | fail 1m01s | **pass** 1m36s |
| `test devlow package / build` | fail 1m06s | **pass** 1m40s |

The root-cause writeup is preserved in the description above for anyone who hits a duplicate key in `errors.json` again.

PR #97324, opened 2026-08-13
[backport] Fix race when navigating Back before hydration

acdlite:
Backports #96252 ("Fix race when navigating Back before hydration") to `next-16-3`.

Pressing Back (or Forward) before the page finishes hydrating can leave you on the wrong history entry: the browser moves to a different entry than the one the document was activated on, and the `popstate` fires before the router's listener exists, so nobody handles it and you end up viewing content that doesn't match the URL. The fix uses the Navigation API to detect, during hydration, that the current history entry differs from the activation entry, and in that case replays the missed traversal using the same logic we run for `onPopState`.

This PR **relands** #95682, which was reverted in #95853 before 16.3.0 shipped. The original reland was unsafe because, under React Activity, React would sometimes hang while applying the resulting state update. That hang was fixed in [facebook/react#37135](https://github.com/facebook/react/pull/37135) and synced to canary, which is what makes it safe to redo the fix.

**Dependency — do not merge before the React bump.** Because this relands the change that triggered the Activity hang, it is only safe on `next-16-3` if the React fix from facebook/react#37135 is present. The 16.3.1 release will include a React canary bump that contains that fix. This backport MUST land together with, or after, that React canary bump. It must not land before the bump is on the branch — otherwise the Activity hang that caused the original revert can reappear.

- Cherry-picked cleanly from `0fd2e51ed6` (`git cherry-pick -x`); no conflicts, and the resulting patch is byte-identical to the source commit (matching `git patch-id`).
- The only source file changed is `packages/next/src/client/components/app-router.tsx`; everything else is the new `test/e2e/app-dir/back-before-hydration/` fixtures and suite.
- Carries the `back-before-hydration` e2e suite from the original PR.
- Behavioral verification is left to branch CI (full e2e matrix), per branch convention.
- Note: CI on this branch is only meaningful once the React canary bump containing facebook/react#37135 is present. Until then, the Activity hang can surface and CI results here should not be trusted as a green light.

### DIFF
--- a/packages/next/src/client/components/app-router.tsx
+++ b/packages/next/src/client/components/app-router.tsx
@@ -56,6 +56,59 @@ const globalMutable: {
   pendingMpaPath?: string
 } = {}
 
+// A Back/Forward press before the router's popstate listener exists moves the
+// browser to a different history entry than the one the document was activated
+// on, and the resulting popstate fires with nobody listening. The activation
+// entry is fixed for the document's lifetime and entry keys are stable across
+// replaceState, so until the listener is installed a key mismatch means a
+// traversal went unobserved.
+function hasMissedTraversal(): boolean {
+  if (typeof window.navigation === 'undefined') {
+    return false
+  }
+  const activationEntry = window.navigation.activation?.entry
+  const currentEntry = window.navigation.currentEntry
+  return (
+    activationEntry != null &&
+    currentEntry != null &&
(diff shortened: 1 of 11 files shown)


ztanner (approved):

acdlite:
@gaearon says we'll hold off on this one for the time being

PR #97333, opened 2026-08-13
Turbopack: remove stale manifests for deleted routes

### LINKED ISSUE #97035: Turbopack dev keeps deleted App Router manifests, causing optional catch-all routes to 404 until restart

https://github.com/OmniNetOpensource/next-turbopack-stale-app-paths-repro

### To Reproduce

The public repository contains the initial routes, an automated route-swap script, a reset script, and the exact verification commands. I also reproduced it in a production app, but the minimal case does not depend on that codebase.

Create a fresh App Router project with this initial tree:

```text
app/
  layout.js
  [locale]/
    project/
      page.js
      [projectId]/
        home/
          page.js
middleware.js
```

`app/layout.js`:

```jsx
export default function RootLayout({ children }) {
  return <html><body>{children}</body></html>
}
```

`app/[locale]/project/page.js`:

```jsx
export default function ProjectPage() {
  return <main>old localized project route</main>
}
```

`app/[locale]/project/[projectId]/home/page.js`:

```jsx
export default function ProjectHomePage() {
  return <main>old localized home route</main>
}
```

`middleware.js`:

```js
import { NextResponse } from 'next/server'

export function middleware(request) {
  const pathname = request.nextUrl.pathname
  if (pathname.startsWith('/en/') || pathname === '/en') return NextResponse.next()

  const url = request.nextUrl.clone()
  url.pathname = `/en${pathname}`
  return NextResponse.rewrite(url)
}

export const config = {
  matcher: '/((?!_next|.*\\..*).*)',
}
```

Start Turbopack dev and compile both routes:

```bash
pnpm next dev --turbopack
curl -I http://localhost:3000/project
curl -I http://localhost:3000/project/acme/home
```

Both return 200.

Without stopping the dev server:

1. Delete `app/[locale]/project/page.js`.
2. Delete `app/[locale]/project/[projectId]/home/page.js`.
3. Add `app/[locale]/project/[[...slug]]/page.js`:

```jsx
export default function ProjectCatchAllPage() {
  return <main>new localized project catch-all</main>
}
```

Then request:

```bash
curl -I http://localhost:3000/project
curl -I http://localhost:3000/project/acme/home
curl -I http://localhost:3000/project/acme/schedule
curl -I http://localhost:3000/en/project/acme/schedule
```

All four return 404. Restarting the dev server makes all four return 200 without any source changes.

The dev server reports:

```text
unhandledRejection: Error: You cannot define a route with the same specificity as a optional catch-all route
("/[locale]/project" and "/[locale]/project[[...slug]]").
```

After the live file-tree change, `.next/dev/server/app-paths-manifest.json` contains both the new route and deleted routes:

```json
{
  "/[locale]/project/[[...slug]]/page": "app/[locale]/project/[[...slug]]/page.js",
  "/[locale]/project/[projectId]/home/page": "app/[locale]/project/[projectId]/home/page.js",
  "/[locale]/project/page": "app/[locale]/project/page.js",
  "/_not-found/page": "app/_not-found/page.js"
}
```

After restarting dev, it correctly contains only:

```json
{
  "/[locale]/project/[[...slug]]/page": "app/[locale]/project/[[...slug]]/page.js"
}
```

Deleting App Router pages during `next dev --turbopack` should remove their partial manifests and route matchers. The newly added optional catch-all should serve the paths without requiring a dev-server restart.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.0.0: Fri Aug 15 21:20:07 PDT 2025; root:xnu-12377.1.8~8/RELEASE_ARM64_T8142
  Available memory (MB): 24576
  Available CPU cores: 10
Binaries:
  Node: 24.13.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 10.32.1
Relevant Packages:
  next: 16.3.1-canary.10 // Latest available version is detected (16.3.1-canary.10).
  eslint-config-next: N/A
  react: 19.2.7
  react-dom: 19.2.7
  typescript: N/A
Next.js Config:
  output: N/A
```

Also reproduced on macOS arm64 with Node 24.13.0 and pnpm 10.32.1 using:

- Next.js 16.2.12
- Next.js 16.3.0
- Next.js 16.3.1-canary.10 (latest canary at the time of testing)

### Suspected source location

This appears to be a stale per-entry manifest cleanup issue rather than a real route conflict.

`handleEntrypoints` clears and rebuilds `currentEntrypoints`, then calls `handleEntrypointsDevCleanup`:
https://github.com/vercel/next.js/blob/canary/packages/next/src/server/dev/turbopack-utils.ts#L642-L684

The cleanup removes stale asset mappings, subscriptions, and issues, but does not remove the corresponding entry from `TurbopackManifestLoader`:
https://github.com/vercel/next.js/blob/canary/packages/next/src/server/dev/turbopack-utils.ts#L844-L883

`TurbopackManifestLoader` already exposes `delete(key)`, which removes the entry from `appPathsManifests` and the other partial manifest maps:
https://github.com/vercel/next.js/blob/canary/packages/next/src/shared/lib/turbopack/manifest-loader.ts#L231-L239

I could not find a call to `manifestLoader.delete(key)` in the removed-entry cleanup path. The stale `/[locale]/project/page` partial manifest is then merged with the new optional catch-all and triggers the false same-specificity error.

Dynamic Routes, Internationalization (i18n), Middleware, Turbopack

next dev (local)

### Additional context

The top-level dynamic `[locale]` segment is material to this reproduction. A simpler non-localized `app/project/...` case also leaves stale entries in `app-paths-manifest.json`, but its in-memory matcher continued serving the new catch-all in my test instead of returning 404.

This supersedes #97020, which had inline reproduction code but no public repository URL, and #97033, which had the public repository link but was submitted without the repository's `Bug` issue type metadata.

marcoshernanz:
## What?

- remove manifest fragments when Turbopack's development entrypoint cleanup removes a deleted page
- add an end-to-end regression for replacing concrete App Router pages with an optional catch-all without restarting the dev server

## Why?

Turbopack already removes deleted entry keys from its asset mapper, subscriptions, issues, and client state. It did not remove the corresponding records from `TurbopackManifestLoader`, so the next manifest write continued to serialize deleted App Router routes.

When a deleted concrete route is replaced by an optional catch-all, those stale records conflict with the new route at the same specificity. The dev server reports the route conflict and responds with 404 until it is restarted.

This is independently consistent with the cleanup identified in #97194. This draft adds end-to-end coverage that fails on current canary by observing both the 404 and stale manifest records, then passes with the lifecycle fix.

## How?

Pass the existing manifest loader to `handleEntrypointsDevCleanup` and call its existing `delete(key)` method in the same branch that deletes the stale asset mapping. This keeps all per-entry manifest maps synchronized with the authoritative current entrypoint set without special-casing any route shape.

The regression test:

1. compiles two concrete localized routes
2. deletes both pages and adds an optional catch-all
3. waits for three affected paths to return the new catch-all content
4. verifies the development app-paths manifest contains the catch-all and no deleted entries

## Verification

- reproduced the report on `16.3.1-canary.10`: all four reported paths returned 404 after the live route replacement and returned 200 after a restart
- confirmed the regression fails on current canary before the source change with the same-specificity route error and a 404 response
- `pnpm test-dev-turbo test/development/app-dir/hmr-deleted-page/hmr-deleted-page.test.ts`
- `pnpm test-dev-webpack test/development/app-dir/hmr-deleted-page/hmr-deleted-page.test.ts`
- `pnpm test-dev-rspack test/development/app-dir/hmr-deleted-page/hmr-deleted-page.test.ts`
- `CI=1 pnpm build-all`
- `pnpm --filter=next build`
- formatting, ESLint, staged-diff, and signed-commit checks

Fixes #97035


test/development/app-dir/hmr-deleted-page/hmr-deleted-page.test.ts
```diff
@@ -27,4 +27,47 @@ describe('hmr-deleted-page', () => {
...
+        )
+      }
+
+      const appPathsManifest = JSON.parse(
+        await next.readFile('.next/dev/server/app-paths-manifest.json')
+      )
+      expect(appPathsManifest['/[locale]/project/[[...slug]]/page']).toBe(
+        'app/[locale]/project/[[...slug]]/page.js'
+      )
+      expect(appPathsManifest['/[locale]/project/page']).toBeUndefined()
```
lukesandberg, line 77: nit assert these do exist prior to the deletes

lukesandberg (approved):

PR #97335, opened 2026-08-13
Turbopack: report malformed environment variables

### LINKED ISSUE #97265: Turbopack build panic hides NUL-byte environment variable spawn error behind CSS module

https://github.com/abdoufma/turbopack-nul-env-repro

### To Reproduce

This issue requires launching Next.js from a Windows process session where Node observes an inherited environment variable containing a NUL byte.

In the affected session, confirm the environment:

```powershell
node scripts/detect-nul-env.cjs
```

Observed:

```text
Environment variables containing NUL bytes:
- steam_master_ipc_name_override = "Remote\u0000"
```

Then run the Turbopack build directly:

```powershell
node node_modules/next/dist/bin/next build --turbo
```

Current behavior:

Turbopack fails with an internal error that points at an ordinary CSS file:

```text
Failed to write app endpoint /page

Caused by:
- [project]/app/globals.css [app-client] (css)
- creating new process
- spawning node pooled process
- nul byte found in provided data
```

The panic path continues through CSS/PostCSS:

```text
Execution of parse_css failed
Execution of <PostCssTransformedAsset as Asset>::content failed
Execution of PostCssTransformedAsset::process failed
Execution of evaluate_webpack_loader failed
creating new process
spawning node pooled process
nul byte found in provided data
```

Expected behavior:

Turbopack should surface the underlying child-process spawn/environment error with the offending environment variable name, similar to the Webpack build path:

```powershell
node node_modules/next/dist/bin/next build --webpack
```

Webpack reports:

```text
TypeError: The property 'options.env['steam_master_ipc_name_override']' must be a string without null bytes. Received 'Remote\x00'
```

I am not expecting Turbopack to support malformed environment values. The bug is that the diagnostic points at an unrelated CSS module and hides the actionable environment variable name.

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 16255
  Available CPU cores: 12
Binaries:
  Node: 24.16.0
  npm: N/A
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.3.1-canary.12
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

CSS, Error Handling, Turbopack

next build (local)

### Additional context

The same original application builds on another Windows server where the malformed inherited environment variable is not present.

This minimal reproduction uses `next@canary`, locked to `16.3.1-canary.12`.

Fresh verification from this repro on August 12, 2026:

- `node scripts/detect-nul-env.cjs` reports `steam_master_ipc_name_override = "Remote\u0000"`.
- `node node_modules/next/dist/bin/next build --turbo` fails with `TurbopackInternalError` and `nul byte found in provided data` under `app/globals.css` / PostCSS.
- `node node_modules/next/dist/bin/next build --webpack` fails with Node's actionable `ERR_INVALID_ARG_VALUE` and includes `options.env['steam_master_ipc_name_override']`.

The generated Turbopack report URL from the canary CLI included:

```text
Turbopack version: `40503aae`
Next.js version: `0.0.0`
```

The `0.0.0` version in the generated report URL looks suspicious because the CLI header correctly prints:

```text
Next.js 16.3.1-canary.12
```

Related but different reports found before opening this issue:

- https://github.com/vercel/next.js/issues/90860: similar CSS/PostCSS/Turbopack area, but caused by a project-root file named `nul` / Windows device path behavior.
- https://github.com/vercel/next.js/issues/96412: same child-process area, but caused by `STATUS_DLL_INIT_FAILED` rather than a NUL byte in an inherited environment variable.
- https://github.com/vercel/next.js/discussions/89780: same "spawning node pooled process" area, but the error is `program not found`.

marcoshernanz:
## Summary

- validate environment variable names and values before Turbopack spawns a Node.js worker
- identify the malformed variable while redacting values and bytes after a NUL, since internal errors are persisted in panic logs and bug-report URLs
- preserve behavior for valid environments while replacing the opaque `nul byte found in provided data` spawn error with an actionable diagnostic

Fixes #97265

Related: #97281. This version does not include environment values in persisted internal error output.

## Verification

- `cargo test -p turbopack-node --lib process_pool::tests`
- `cargo test -p turbopack-node --lib`
- `cargo clippy -p turbopack-node --lib -- -D warnings`
- `pnpm build-all`
- `pnpm test-dev-turbo test/e2e/app-dir/webpack-loader-errors/webpack-loader-errors.test.ts -t "CLI output"`

The malformed inherited Windows environment cannot be synthesized from JavaScript or PowerShell. I reproduced the exact `std::process::Command::spawn` boundary and error in Rust. A Windows cross-target check was not runnable in the Linux DevBox because the Windows SDK/MSVC linker is unavailable.

### DIFF
--- a/turbopack/crates/turbopack-node/src/process_pool/mod.rs
+++ b/turbopack/crates/turbopack-node/src/process_pool/mod.rs
@@ -299,6 +299,20 @@ impl<R: AsyncRead + Unpin, W: AsyncWrite + Unpin> OutputStreamHandler<R, W> {
     }
 }
 
+fn validate_env_for_child_process(env: &FxHashMap<RcStr, RcStr>) -> Result<()> {
+    for (name, value) in env {
+        if let Some((name_prefix, _)) = name.split_once('\0') {
+            // Bytes after the NUL may contain secrets from a corrupted environment entry.
+            bail!("environment variable name {name_prefix:?} contains a NUL byte");
+        }
+        if value.contains('\0') {
+            // Values may contain secrets, and this error can be persisted in a panic log.
+            bail!("environment variable {name:?} contains a NUL byte in its value");
+        }
+    }
+    Ok(())
+}
+
 impl NodeJsPoolProcess {
     async fn new(
         cwd: &Path,
@@ -334,6 +348,8 @@ impl NodeJsPoolProcess {
             std::env::var("SystemRoot")
                 .expect("the SystemRoot environment variable should always be set"),
         );
+        // `Command::spawn` otherwise reports only that some data contains a NUL byte.
+        validate_env_for_child_process(env)?;
         cmd.envs(env);
         cmd.stderr(Stdio::piped());
         cmd.stdout(Stdio::piped());
@@ -976,5 +992,53 @@ impl Drop for ChildProcessOperation {
                 self.idle_processes.push(process, &ACTIVE_POOLS);
             }
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use rustc_hash::FxHashMap;
+    use turbo_rcstr::RcStr;
+
+    use super::validate_env_for_child_process;
+
+    fn env(entries: &[(&str, &str)]) -> FxHashMap<RcStr, RcStr> {
+        entries
+            .iter()
+            .map(|&(name, value)| (RcStr::from(name), RcStr::from(value)))
+            .collect()
+    }
+
+    #[test]
+    fn accepts_env_without_nul_bytes() {
+        assert!(validate_env_for_child_process(&env(&[("NAME", "value")])).is_ok());
+    }
+
+    #[test]
+    fn reports_prefix_without_exposing_suffix_of_name_containing_nul_byte() {
+        let error = validate_env_for_child_process(&env(&[("BAD\0sensitive-suffix", "value")]))
+            .unwrap_err()
+            .to_string();
+
+        assert_eq!(
+            error,
+            "environment variable name \"BAD\" contains a NUL byte"
+        );
+    }
+
+    #[test]
+    fn reports_name_without_exposing_value_containing_nul_byte() {
+        let error = validate_env_for_child_process(&env(&[(
+            "steam_master_ipc_name_override",
+            "sensitive-prefix\0sensitive-suffix",
+        )]))
+        .unwrap_err()
+        .to_string();
+
+        assert_eq!(
+            error,
+            "environment variable \"steam_master_ipc_name_override\" contains a NUL byte in its \
+             value"
+        );
     }
 }


turbopack/crates/turbopack-node/src/process_pool/mod.rs
```diff
@@ -299,6 +299,20 @@ impl<R: AsyncRead + Unpin, W: AsyncWrite + Unpin> OutputStreamHandler<R, W> {
     }
 }
 
+fn validate_env_for_child_process(env: &FxHashMap<RcStr, RcStr>) -> Result<()> {
+    for (name, value) in env {
+        if let Some((name_prefix, _)) = name.split_once('\0') {
+            // Bytes after the NUL may contain secrets from a corrupted environment entry.
+            bail!("environment variable name {name_prefix:?} contains a NUL byte");
+        }
```
bgw, line 307:
```suggestion
        if name.contains('\0') {
            bail!("environment variable name {name:?} contains a NUL byte");
        }
```

- It's very unlikely that the name would contain secrets in it, even if it did contain a null byte.
- We're using the `Debug` impl (via `:?`) so we don't need to only grab the prefix, it'll just string escape the null byte if we print the whole thing, which probably makes more sense anyways if the null byte is at the beginning or in the middle of the name.

turbopack/crates/turbopack-node/src/process_pool/mod.rs
```diff
@@ -299,6 +299,20 @@ impl<R: AsyncRead + Unpin, W: AsyncWrite + Unpin> OutputStreamHandler<R, W> {
     }
 }
 
+fn validate_env_for_child_process(env: &FxHashMap<RcStr, RcStr>) -> Result<()> {
```
bgw, line 302: This is an improvement, but it would be better still for us to emit an `Issue` (can be a lower severity, maybe a warning), strip out the unsupported environment variable, and continue execution, instead of aborting with a fatal `bail`.

Is there any reason that would be particularly hard to do in `NodeJsPoolProcess::new`? If not, we should do it.

turbopack/crates/turbopack-node/src/process_pool/mod.rs
```diff
@@ -299,6 +299,20 @@ impl<R: AsyncRead + Unpin, W: AsyncWrite + Unpin> OutputStreamHandler<R, W> {
...
+    for (name, value) in env {
+        if let Some((name_prefix, _)) = name.split_once('\0') {
+            // Bytes after the NUL may contain secrets from a corrupted environment entry.
+            bail!("environment variable name {name_prefix:?} contains a NUL byte");
+        }
+        if value.contains('\0') {
+            // Values may contain secrets, and this error can be persisted in a panic log.
```
bgw, line 309: This is fair, values often contain sensitive values, and the value probably isn't all that useful.

turbopack/crates/turbopack-node/src/process_pool/mod.rs
```diff
@@ -334,6 +348,8 @@ impl NodeJsPoolProcess {
             std::env::var("SystemRoot")
                 .expect("the SystemRoot environment variable should always be set"),
         );
+        // `Command::spawn` otherwise reports only that some data contains a NUL byte.
+        validate_env_for_child_process(env)?;
```
bgw, line 352: Make this a doc comment on `validate_env_for_child_process` itself.

```suggestion
        validate_env_for_child_process(env)?;
```

bgw (changes requested):
Follow-up questions:
- Why does `Command::spawn` reject null bytes in environment variable names? It should accept everything that the OS does, and if the OS doesn't support it, then...
- How did the null byte end up in an environment variable if the OS doesn't support it? Did it come from a `.env` file? Should we do the validation there instead? That would let us point to a file in the `Issue`.
- Should we use a similar issue for non-unicode environment variables?

PR #96440, opened 2026-07-31
Turbopack: Improve how DiskWatcher is configured and fix polling watcher bugs

bgw:
- Adds a `DiskWatcherConfig` object to centralize the recursive mode, polling, and invalidation reason options.
- **Polling Fix:** Use the non-recursive mode if the polling watcher is selected, ignoring the OS/platform.
- **Polling Fix:** Treat mtime events as data update events if the polling watcher is selected (this is what https://github.com/vercel/next.js/pull/96288 was doing, but this PR only does it if polling is enabled)
- Only treat the `MetadataKind::Any` event as a data change on macOS.

Fixes #80665
Closes #96288

PR #80687, opened 2025-06-19
docs: watchOptions.pollIntervalMs

icyJoseph:
Closes: https://linear.app/vercel/issue/DOC-4798/document-turbopack-watchoptionspollintervalms
Fixes: #80665

PR #97339, opened 2026-08-13
Turbopack: Fix polling watcher dropping rapid edits

### LINKED ISSUE #96982: Turbopack: `watchOptions.pollIntervalMs` permanently drops file edits on Windows

https://github.com/DenisCDev/turbopack-polling-drop

### To Reproduce

```bash
npm install

# polling watcher
POLL_MS=1000 SESSIONS=5 EDITS=5 npm run probe

# same edit pattern, native watcher, as a control
POLL_MS=0 SESSIONS=5 EDITS=5 npm run probe
```

Each session removes `.next`, boots a fresh `next dev`, rewrites `lib/mod.ts` five times in a row, and records how long until the served page reflects each new value — or `X` if it never did within `WINDOW_MS` (default 8000 ms, eight full poll intervals).

**Current.** With `watchOptions.pollIntervalMs` set, edits are lost silently and permanently. The native watcher, running the identical pattern on the same machine, loses none.

```
watcher=polling 1000ms | 5 sessions x 5 edits | window 8000ms

session  1:  889  763 1028 1044 1053
session  2:  801    X  287 1052 1075
session  3:  691 1021 1037 1029 1060
session  4:  665    X  264 1021 1036
session  5:  739    X  218 1046 1037

detected: 22/25
file mtime advanced on 25/25 writes
```

```
watcher=native | 5 sessions x 5 edits | window 8000ms

session  1:  374   95   75   90   67
session  2:   97   82   74   90   72
session  3:   97   90  397   89  419
session  4:   96   87   78   97   94
session  5:   95  171   75   89   83

detected: 25/25
```

Repeated runs of the polling arm gave 15/25, 20/25 and 22/25. The native arm has not lost an edit in any run I have done.

Two objections answered in the output itself:

- **The write reaches disk.** The probe stats the file around every write; mtime advanced on 25/25, including every lost edit.
- **The loss is permanent, not slow.** With `WINDOW_MS=90000` — ninety poll intervals — a lost edit still never arrives (8/9 detected in that run). Only writing the file again brings the route back.

**Expected.** An edit that is on disk is eventually reflected, however soon it followed the previous one.

**One correlation, offered as an observation.** Losses track the latency of the *preceding* edit. While edits are reflected in ~200–800 ms, faster than the configured 1000 ms interval, the next one tends to be lost; once latency settles at ~1000 ms, losses stop for the rest of the session. I have not identified what detects a change in 200 ms while polling is enabled.

I want to be clear about the limit of this measurement: the probe only observes that the dev server never serves the new value. It cannot tell an event that was never emitted from one that was emitted and not acted on. There is no instrumentation inside Turbopack here.

```
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Home
  Available memory (MB): 16290
  Available CPU cores: 12
Binaries:
  Node: 22.22.0
  npm: 10.9.4
  Yarn: 1.22.22
  pnpm: 10.29.3
Relevant Packages:
  next: 16.3.1-canary.8 // Latest available version is detected (16.3.1-canary.8).
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

Turbopack

next dev (local)

### Additional context

**This is not an old, unfixed bug.** #96440 landed 2026-08-07 22:40 UTC, 44 minutes before `v16.3.1-canary.8` was tagged, so the build measured above already contains it. The polling watcher still drops edits on native Windows with that fix in place.

`watchOptions.pollIntervalMs` is the fallback for environments where native watching does not work — Docker, WSL2, network drives, VMs — which is exactly where users have no alternative. The option is also undocumented today: no JSDoc in `config-shared`, nothing in `next/dist/docs/`, and the docs PR (#80687) has been open since 2025-06-19.

One asymmetry visible from outside Turbopack, which may or may not be related: the dev server logs seven `watch error (...): Io(Os { ..., kind: NotFound })` lines per session in the polling arm and **zero** in the native arm. None of them is the edited file, which exists and is watched. The probe keeps each session's server output under `probe-logs/`.

Why closely spaced writes are not a synthetic case: they are the normal output of format-on-save, `git checkout` and `merge`, codemods, and CLI coding agents that write several files in a burst before requesting the page.

Two caveats. It is intermittent — whether a given session loses anything varies, and short runs can come back clean; every run of 25 observations I attempted reproduced losses. And I have only tested native Windows; I have no macOS or Linux machine to compare against, so I cannot say whether this is Windows-specific.

Happy to run candidate patches or extra instrumentation here — each arm of the repro takes about two minutes.

marcoshernanz:
## What?

Fixes Turbopack polling mode permanently dropping rapid file edits by pinning a `notify` 8.2.0 backport that preserves subsecond file mtime precision.

Adds a deterministic `turbo-tasks-fs` regression test that advances a watched file's mtime twice within one second and verifies the task is invalidated.

Upstream dependency fix: https://github.com/notify-rs/notify/pull/981

## Why?

`watchOptions.pollIntervalMs` selects `notify::PollWatcher`. Released `notify` versions convert `SystemTime` to whole seconds before comparing snapshots. With content hashing disabled, two writes within the same second compare equal, so the second edit emits no event and remains stale until a later edit changes the truncated timestamp.

This is not Windows-only:

- Issue repro on Windows: 22/25 edits detected; all 25 mtimes advanced.
- Unmodified `canary` on Linux: one polling run detected 24/25 edits and never recovered within the 8-second observation window; native control detected 25/25.
- The deterministic regression timed out after 5 seconds against released `notify`.

PR #96440 correctly handles `WriteTime` events once emitted, but cannot recover an event that `PollWatcher` never produces.

The behavior affects the intended Docker/WSL/network-filesystem polling escape hatch. Related reports include #80665, #80687, #71622, and #94146.

## How?

The pinned dependency keeps nanosecond precision for ordinary files without enabling `compare_contents`, which would hash every watched file on every poll. It retains whole-second behavior for directories and content-hashed watchers to avoid additional directory events and preserve existing event classification.

The pin is based on stable `notify` 8.2.0 rather than 8.1.0, so it includes the intervening inotify fixes. A TODO links the upstream PR for removal after release.

## Verification

- Baseline, original reporter repo, current canary, Linux:
  - polling: 24/25, then 25/25
  - native controls: 25/25 and 25/25
- Candidate, untouched reporter probe packaged from this branch:
  - polling: 25/25 and 25/25
  - interleaved native control: 25/25
- `notify` 8.2.0 backport: 10 passed, 1 ignored; 5 doc tests passed
- current upstream `notify`: 82 passed, 2 ignored; hashing/race/serialization integration tests and 8 doc tests passed
- `cargo test -p turbo-tasks-fs`: 119 passed
- `cargo clippy -p turbo-tasks-fs --tests -- -D warnings`
- Full local Next.js package and native binding build
- Independent closeout review: no actionable findings

Fixes #96982

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5196,9 +5196,8 @@ dependencies = [
 
 [[package]]
 name = "notify"
-version = "8.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3163f59cd3fa0e9ef8c32f242966a7b9994fd7378366099593e0e73077cd8c97"
+version = "8.2.0"
+source = "git+https://github.com/marcoshernanz/notify.git?rev=8e42e82e0710643d8f85eba2b2086bf8fb1b8427#8e42e82e0710643d8f85eba2b2086bf8fb1b8427"
 dependencies = [
  "bitflags 2.9.1",
  "fsevent-sys",
@@ -5215,8 +5214,7 @@ dependencies = [
 [[package]]
 name = "notify-types"
 version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5e0826a989adedc2a244799e823aece04662b66609d96af8dff7ac6df9a8925d"
+source = "git+https://github.com/marcoshernanz/notify.git?rev=8e42e82e0710643d8f85eba2b2086bf8fb1b8427#8e42e82e0710643d8f85eba2b2086bf8fb1b8427"
 
 [[package]]
 name = "nu-ansi-term"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -367,6 +367,8 @@ vergen-gitcl = { version = "1.0.8", features = ["cargo"] }
 webbrowser = "1.0.6"
 
 [patch.crates-io]
+# TODO: Remove after https://github.com/notify-rs/notify/pull/981 is released.
+notify = { git = "https://github.com/marcoshernanz/notify.git", rev = "8e42e82e0710643d8f85eba2b2086bf8fb1b8427" }
 bincode = { git = "https://github.com/bgw/bincode.git", branch = "bgw/patches" }
 virtue = { git = "https://github.com/bgw/virtue.git", branch = "bgw/fix-generic-default-parsing" }
 # We have to very frequently bump the swc major version of mdxjs-rs
--- a/turbopack/crates/turbo-tasks-fs/src/watcher/mod.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/watcher/mod.rs
@@ -1053,6 +1053,45 @@ mod tests {
             .unwrap();
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn polling_detects_subsecond_mtime_changes() {
+        let tt = TurboTasks::new(TurboTasksBackend::new(
+            BackendOptions::default(),
+            noop_backing_storage(),
+        ));
+        tt.run_once(async move {
+            let fs = MockFileSystem::new(DiskWatcherConfig {
+                recursive_mode: Some(DiskWatcherRecursiveMode::NonRecursive),
+                poll_interval: Some(Duration::from_millis(100)),
+                report_invalidation_reason: true,
+            });
+            let file_path = fs.root_path.join("file.txt");
+            let initial_mtime = SystemTime::UNIX_EPOCH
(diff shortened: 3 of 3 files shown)


bgw:
I don't think this feature is important enough to justify maintaining a `notify-rs` backport, but we can upgrade the `notify-rs` crate as soon as there's a new version available (and I'm fine upgrading us to an 9.x release candidate).

PR #97345, opened 2026-08-14
Resolve metadata independently across parallel routes

gnoff:
## Motivation

Parallel routes are separate rendering branches, but the existing metadata resolver collects every matched slot into one shared chain. As a result, traversal order can decide which slot supplies the document head, and metadata errors from one slot can interfere with another.

## What this does

This adds a fork-aware resolver behind `experimental.parallelRouteMetadata` while retaining the existing resolver as the default. Each slot resolves metadata and viewport independently, with isolated accumulators, and a single branch is selected for the document head at every fork.

The built-in selector prefers real routes over implicit fallbacks and prefers `children` when it contributes head data. Otherwise it chooses the branch with the deepest head definition, using route depth and lexical slot order only as stable tie-breakers.

Head selection is separate from rendering outcomes. Every slot gets its own `MetadataOutlet`, which replays navigation and regular errors for that slot regardless of which branch supplied the head.

## Review notes

- `generateMetadata`, `generateViewport`, and generated static metadata still start eagerly; a fork does not serialize sibling branches or parent/child generation.
- Branch state is isolated only where the tree forks, preventing metadata mutation from crossing between slots.
- An outlet rejects as soon as either metadata or viewport rejects, but resolves only after both succeed. This lets one failing result reach the slot boundary even if the other never settles.
- Existing navigation convention metadata is preserved for the selected branch. Errors from unselected branches remain observable only through their own rendered outlets.
- The old resolver remains available behind an explicit flag so this implementation can be compared or disabled during rollout.

## Verification

- `pnpm --filter=next types`
- Metadata resolver unit coverage
- Parallel-route metadata, navigation, and not-found coverage in development and production with Turbopack and webpack


packages/next/src/lib/metadata/resolve-metadata-parallel.ts
```diff
@@ -0,0 +1,983 @@
...
+  const shouldResolveMetadata = Boolean(resolutionTarget & METADATA)
+  const shouldResolveViewport = Boolean(resolutionTarget & VIEWPORT)
+  const [moduleResult, staticFilesMetadata] = await Promise.all([
+    errorConvention
+      ? getComponentTypeModule(tree, 'layout').then((mod) => ({
+          mod,
+          modType: errorConvention,
+        }))
+      : getLayoutOrPageModule(tree),
+    shouldResolveMetadata ? resolveStaticMetadata(tree[2], props) : null,
```
ztanner: what happens if metadata resolution throws? ie if I throw in `generateImageMetadata`? It seems like it would reject the entire shared metadata resolution instead of only the owning slot

test/production/next-server-nft/next-server-nft.test.ts
```diff
@@ -297,10 +297,13 @@ async function readNormalizedNFT(next, name) {
            "/node_modules/next/dist/lib/metadata/generate/icon-mark.js",
            "/node_modules/next/dist/lib/metadata/generate/utils.js",
            "/node_modules/next/dist/lib/metadata/get-metadata-route.js",
+           "/node_modules/next/dist/lib/metadata/index.js",
            "/node_modules/next/dist/lib/metadata/is-metadata-route.js",
            "/node_modules/next/dist/lib/metadata/metadata-context.js",
+           "/node_modules/next/dist/lib/metadata/metadata-parallel.js",
```
unstubbable, line 304: Ideally we only pull this rather large module in if the feature flag is enabled (and then don't include the legacy module).

test/e2e/app-dir/metadata-streaming-parallel-routes/next.config.js
```diff
@@ -1,6 +1,10 @@
 /**
  * @type {import('next').NextConfig}
  */
-const nextConfig = {}
+const nextConfig = {
+  experimental: {
+    parallelRouteMetadata: true,
```
unstubbable: We probably want to run this test suite with the flag on and off.

test/e2e/app-dir/parallel-route-not-found/next.config.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @type {import('next').NextConfig}
+ */
+const nextConfig = {
+  experimental: {
+    parallelRouteMetadata: true,
```
unstubbable: We probably want to run this test suite with the flag on and off.

test/e2e/app-dir/metadata-streaming-parallel-routes/app/conditional-slot/layout.tsx
```diff
@@ -0,0 +1,26 @@
...
+    default: 'conditional layout title',
+    template: '[%s]',
+  },
+}
+
+// This route intentionally contains a regular metadata error in an unrendered
+// slot. Keep it request-rendered so the production build can exercise the
+// outlet behavior without treating that intentional error as a prerender
+// failure.
+export const dynamic = 'force-dynamic'
```
unstubbable: Should we use `await connection()` instead to ensure compatibility with Cache Components? By the way, it's a bit concerning that `test/e2e/app-dir/metadata-streaming-parallel-routes` is listed in the `exclude` array of `test/cache-components-tests-manifest.json`.

packages/next/src/lib/metadata/metadata-resolution-primitives.ts
```diff
@@ -235,9 +231,12 @@ export async function mergeMetadata(
     metadataContext: MetadataContext
     buildState: BuildState
     leafSegmentStaticIcons: StaticIcons
-  }
+  },
+  cloneResolvedMetadata = true
```
unstubbable: Can we move this into the existing options param above?

packages/next/src/lib/metadata/metadata-resolution-primitives.ts
```diff
@@ -454,14 +453,19 @@ export async function mergeMetadata(
...
-  const newResolvedViewport = structuredClone(resolvedViewport)
+export function mergeViewport(
+  {
+    resolvedViewport,
+    viewport,
+  }: {
+    resolvedViewport: ResolvedViewport
+    viewport: Viewport | null
+  },
+  cloneResolvedViewport = true
```
unstubbable: Can we move this into the existing options param above?

packages/next/src/lib/metadata/resolve-metadata-parallel.ts
```diff
@@ -0,0 +1,1000 @@
...
+        keyPath: childKeyPath,
+      }),
+    })
+  }
+
+  const childBranches: MetadataBranchAtFork[] = []
+  for (const pendingChildBranch of pendingChildBranches) {
+    childBranches.push({
+      key: pendingChildBranch.key,
+      branch: await pendingChildBranch.branch,
```
unstubbable, line 867: If this throws and further entries in `pendingChildBranches` also have rejected branches, I think those become unhandled rejections if we don't add a `.catch()` to the `branch` before pushing above.

packages/next/src/lib/metadata/resolve-metadata-parallel.ts
gnoff: :robot: Static metadata resolution is branch-local. Its eagerly started promise now has an immediate rejection handler, and the original rejection is consumed only when that branch resolves, so a throwing generateImageMetadata does not reject a sibling branch. The conditional-slot coverage exercises a throwing generateImageMetadata in an unrendered slot while the selected branch still succeeds.

test/e2e/app-dir/metadata-streaming-parallel-routes/next.config.js
gnoff: :robot: Done. The shared streaming assertions now run with parallelRouteMetadata both disabled and enabled; only the parallel-specific fork assertions are limited to the enabled case.

test/e2e/app-dir/parallel-route-not-found/next.config.js
gnoff: :robot: Done. The shared not-found assertions now run with parallelRouteMetadata both disabled and enabled; only the new parallel-routing cases are limited to the enabled case.

test/e2e/app-dir/metadata-streaming-parallel-routes/app/conditional-slot/layout.tsx
gnoff: :robot: Done. The conditional fixture now awaits connection() instead of using force-dynamic, and I verified its unrendered-slot case with Cache Components enabled. The suite manifest exclusion predates this PR, so I have left that broader cleanup out of this change.

packages/next/src/lib/metadata/metadata-resolution-primitives.ts
gnoff: :robot: Done. cloneResolvedMetadata is now an optional field in the existing options object.

packages/next/src/lib/metadata/metadata-resolution-primitives.ts
gnoff: :robot: Done. cloneResolvedViewport is now an optional field in the existing options object.

packages/next/src/lib/metadata/resolve-metadata-parallel.ts
gnoff: :robot: Done. Each eagerly created branch promise receives an immediate rejection handler before it is stored. The original promise is retained for the ordered await, so the error still surfaces through the owning branch without becoming an unhandled sibling rejection first.

test/e2e/app-dir/parallel-route-not-found/parallel-route-not-found.test.ts
```diff
@@ -81,27 +86,27 @@ describe('parallel-route-not-found', () => {
     )
   })
 
-  it('should handle `notFound()` in a slot', async () => {
-    const browser = await next.browser('/not-found-metadata/slot-error')
+  if (parallelRouteMetadata) {
```
unstubbable, line 89: Let's preserve the old assertion as well in an `else` block.

unstubbable (approved):

test/production/next-server-nft/next-server-nft.test.ts
gnoff: :robot: Implemented the selection with a compile-time `__NEXT_PARALLEL_ROUTE_METADATA` define sourced from `experimental.parallelRouteMetadata`. The index keeps the two requires in explicit `if/else` branches, and the emitted-server-bundle test verifies that both Webpack and Turbopack omit the parallel resolver when the flag is disabled and include it when enabled. The package-level NFT still needs to trace both possible configurations. The enabled configuration also still imports `metadata.tsx` for the shared tag builders; excluding that legacy module completely would require extracting those builders into a neutral module, so I am leaving this thread open for that decision.

test/e2e/app-dir/parallel-route-not-found/parallel-route-not-found.test.ts
gnoff: :robot: Done — the disabled-flag branch now retains the existing slot notFound assertion, while the enabled branch keeps the new outlet-specific expectation. The focused Turbopack production suite passes both configurations (9 passed, 4 skipped).

packages/next/src/server/app-render/create-component-tree.tsx
```diff
@@ -630,9 +630,10 @@ async function createComponentTreeInternal(
               missingSlots,
               preloadCallbacks,
               authInterrupts,
-              // `StreamingMetadataOutlet` is used to conditionally throw. In the case of parallel routes we will have more than one page
-              // but we only want to throw on the first one.
-              MetadataOutlet: isChildrenRouteKey ? MetadataOutlet : null,
+              MetadataOutlet:
+                experimental.parallelRouteMetadata || isChildrenRouteKey
```
unstubbable, line 634: If we use `process.env.__NEXT_PARALLEL_ROUTE_METADATA` here as well I think we could remove the whole `renderOpts` plumbing of the flag.

unstubbable (approved):

test/production/next-server-nft/next-server-nft.test.ts
gnoff: :robot: Done — metadata and viewport element rendering now lives in the neutral metadata-elements module. The compile-time switch therefore includes only the selected resolver: the emitted-server-bundle test verifies both directions under Webpack and Turbopack, while the package trace records the shared renderer separately.

packages/next/src/server/app-render/create-component-tree.tsx
gnoff: :robot: Done — create-component-tree now reads __NEXT_PARALLEL_ROUTE_METADATA directly. NextServer and the export worker initialize the runtime value, user and edge bundles receive the existing compile-time define, and the flag has been removed from render options, templates, and app-render types. The metadata matrix passes in Webpack production, Turbopack production, and Turbopack development.

PR #58129, opened 2023-11-07
fix node externals resolving

sokra:
### What?

Various fixes in handling of externals in node.js

* add extensions to builtin externals to allow node.js ESM to work
* improve the auto-externals logic to detect more edge cases
* prepare for `esmExternals` support, but that's blocked by client manifest missing the async flag
* currently only `esmExternals: false` is supported

### Turbopack Changes

* https://github.com/vercel/turbo/pull/6531 
* https://github.com/vercel/turbo/pull/6530

PR #97355, opened 2026-08-14
Respect async field of server actions

mischnic:
Why were we not actually using the `async: boolean` field when loading the module?

### DIFF
--- a/packages/next/src/server/app-render/action-handler.ts
+++ b/packages/next/src/server/app-render/action-handler.ts
@@ -793,6 +793,7 @@ export async function handleAction({
       async (): Promise<HandleActionResult> => {
         // We only use these for fetch actions -- MPA actions handle them inside `decodeAction`.
         let actionModId: string | number | undefined
+        let actionAsync: boolean | undefined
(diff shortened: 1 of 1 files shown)


mischnic:
* **#97355** <a href="https://app.graphite.com/github/pr/vercel/next.js/97355?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/97355?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

unstubbable (commented):
> Why were we not actually using the `async: boolean` field when loading the module?

I don't know. The `await` was added in #58129 when we didn't even have the `async` property. That was only (partially) added later in #71572. An `await` on a non-promise is also fine, so this looks like a micro optimization to me that I think is not even covered by an e2e test. Or do we have a TLA in a server action module somewhere? Falling back to `false` when the async flag is not set in the manifest for some reason also looks slightly risky to me.

PR #97375, opened 2026-08-14
[PPF] unstable_prefetch()

lubieowoce:
Implements `unstable_prefetch()`, which is intended for use in `partialPrefetching`. `await unstable_prefetch()` excludes content from the app shell -- it will only be available when using `prefetch={true}` (speculative prefetch) or during navigations.

As a rule of thumb, `unstable_prefetch()` resolves whenever static `params` would:
- in a static prerender (but not the app shell extracted from it, which is param-less)
- in a runtime prefetch (`prefetch={true}`), but NOT a runtime app shell, which is param-less

Note that `prefetch()` is URL data, so using it in an App Shell without Suspense will trigger an insight.

### Implementation notes

`prefetch()` is treated like URL data, so it resolves in the `PrefetchStatic/PrefetchRuntime` stages added in #96908. Error messages about URL data have not been updated to mention it yet -- we will do that as a follow up, along with docs.

`await prefetch()` does not count as a runtime data access, meaning that it won't affect the static prefetch hint for a route. however `await prefetch(); await cookies()` does deopt the route, because using a speculative runtime prefetch would reveal more content. Note that this may cause us to unnecessarily deopt a shell to runtime even if only the speculative part of the content would be improved by a runtime request; this is not a new issue, but it's something we should optimize.

### DIFF
--- a/test/development/app-dir/cache-components-dev-warmup/dev-warmup.util.ts
+++ b/test/development/app-dir/cache-components-dev-warmup/dev-warmup.util.ts
@@ -343,6 +343,19 @@ export function runDevWarmupTests({
 
           assertLog(
             logs,
+            `after prefetch`,
+            // Same as navigation() below: static prerender timing on initial
+            // load, app-shell timing for a client nav when there's a runtime
+            // prefetch.
+            isInitialLoad
+              ? 'Prerender'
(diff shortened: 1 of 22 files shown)


lubieowoce:
man what the hell happened here, `gh stack` did a silly

PR #97377, opened 2026-08-14
Fix prefetch starvation when a sibling teaches the route pattern first

### LINKED ISSUE #96965: Segment cache: sibling PPR route prefetches are permanently dropped after a same-tick batch (starved task resolves against fallback entry)

https://github.com/hypernomad/next-sibling-segment-prefetch-repro

### To Reproduce

```bash
npm install
npm run build
npx playwright install chromium
npm run repro   # deterministic driver — 4 checks, exits 0 only when control is clean AND the bug reproduces
```

Or manually: `npm run start`, open `http://localhost:3000`, DevTools → Network (filter `products`), click **"1. Prefetch all 5 in the same tick"**.

**Setup:** `/products/[slug]` is a partially-prerendered route (`cacheComponents: true`, `partialPrefetching: true`) with 5 slugs from `generateStaticParams` — content baked into each per-URL shell, plus one trivial dynamic hole (`await connection()`... `return null`).

**Current behavior**

1. Calling `router.prefetch()` for all 5 sibling URLs in the same tick issues network requests for only 4 of them (the task heap is processed most-recent-first with a 4-request bandwidth cap — `hasNetworkBandwidth()` in `segment-cache/scheduler.js`). The first-scheduled sibling (`/products/alpha`) waits for bandwidth.
2. By the time bandwidth frees, the completed siblings' responses have populated a fallback route entry for `/products/[slug]`, and the waiting task resolves against it **without ever fetching alpha's per-URL tree or page segment**.
3. Worse, every later `router.prefetch("/products/alpha")` — seconds or minutes later, including hover-driven Link prefetches — is a **permanent no-op** (zero network) for the lifetime of the page instance.
4. Navigation to the starved URL cannot render slug-specific content from the fallback entry, so it blocks on a full dynamic request — the exact cost prefetch is meant to remove, with no way to warm the URL afterwards.

Control (also asserted by the driver): on a fresh page, a single `router.prefetch("/products/alpha")` correctly fetches tree + page segment.

**Expected behavior**

Each prerendered sibling's per-URL tree/page segment should be fetchable by prefetch — either by letting the queued task run when bandwidth frees instead of resolving against the fallback entry, or at minimum by letting a later explicit `router.prefetch()` of that URL fetch it.

**Real-world impact** (measured on a production marketplace on Vercel, `next@16.3.0`): warming the top-navigation categories can only ever make **one** category commit instantly (~78 ms click→content from the prefetched segment); every starved sibling pays the full dynamic round trip on click (measured 1.3–2.3 s progress bar), and no prefetch strategy can avoid it — batching starves one URL permanently, staggering makes every call after the first response a no-op.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 24.15.0
  npm: 11.12.1
Relevant Packages:
  next: 16.3.1-canary.8 // Latest available version is detected (16.3.1-canary.8).
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

Also reproduced on `next@16.3.0` (production, Vercel).

Navigation, Partial Prerendering (PPR)

next start (local), Vercel (Deployed)

VihaanAgarwal:
Fixes #96965.

Prefetch more sibling URLs of a dynamic route in one tick than the scheduler's request limit, and the excess tasks wait for bandwidth. The first responses teach the optimistic routing trie the route's pattern. When a waiting task finally runs, its cache lookup falls back to `matchKnownRoute` and gets a synthetic entry. The entry is Fulfilled, so the task completes without fetching the URL's own route tree. Later prefetches of that URL hit the same synthetic entry, so the URL can never be warmed for the lifetime of the page. The issue reports this from two production apps: one warm sibling, every other sibling pays the full dynamic request on click.

The fix makes a prefetch task resolve its route against the cache state at schedule time. `schedulePrefetchTask` (and reschedule) records whether the route cache had an entry for the URL, predicted or concrete. If it did, nothing changes: the task may consume a predicted entry, which keeps the intended reuse cases (prefetching a sibling after the pattern is known, or the same pathname with different search params) at zero requests. If it did not, the task committed to a fetch, and a pattern learned in the meantime from a sibling's response must not satisfy it: the task's lookup reads the cache map directly and skips the predicted-entry fallback. A concrete entry written in the meantime still satisfies it.

The new e2e test prefetches five prerendered siblings in one tick and asserts a route tree request fires for each, including the first-scheduled one, which only gets bandwidth after sibling responses have arrived. On canary it fails with exactly `/products/alpha` missing; with the change all five fire.

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -711,18 +711,34 @@ export function readOrCreateRouteCacheEntry(
 ): RouteCacheEntry {
   attachInvalidationListener(task)
 
-  const existingEntry = readRouteCacheEntry(now, key)
-  if (existingEntry !== null) {
-    return existingEntry
-  }
-  // Create a pending entry and add it to the cache.
-  const pendingEntry = createDetachedRouteCacheEntry()
   const varyPath: RouteVaryPath = getRouteVaryPath(
     key.pathname,
     key.search,
     key.nextUrl
   )
   const isRevalidation = false
+  // When the route cache had no entry, not even a predicted one, at the time
+  // the task was scheduled, the task committed to fetching the route tree.
+  // Look up the cache map directly, without the predicted-entry fallback in
+  // readRouteCacheEntry: a pattern learned in the meantime from a sibling's
+  // response contains none of this URL's own data, and consuming it would
+  // complete the task without fetching anything. A concrete entry written in
+  // the meantime still satisfies the task.
+  const existingEntry = task.routeCacheMissedWhenScheduled
+    ? getFromCacheMap(
+        now,
+        getCurrentRouteCacheVersion(),
+        routeCacheMap,
+        varyPath,
+        isRevalidation,
+        false
+      )
+    : readRouteCacheEntry(now, key)
+  if (existingEntry !== null) {
+    return existingEntry
+  }
+  // Create a pending entry and add it to the cache.
+  const pendingEntry = createDetachedRouteCacheEntry()
   setInCacheMap(routeCacheMap, varyPath, pendingEntry, isRevalidation)
   return pendingEntry
 }
--- a/packages/next/src/client/components/segment-cache/scheduler.ts
+++ b/packages/next/src/client/components/segment-cache/scheduler.ts
@@ -168,6 +168,24 @@ export type PrefetchTask = {
   isCanceled: boolean
 
   /**
+   * True if the route cache had no entry for the task's URL, not even a
+   * predicted (matchKnownRoute) entry, at the time the task was scheduled.
+   *
+   * A task resolves its route against the cache state at schedule time. When
+   * this is true, the task committed to fetching the route tree, and a route
+   * pattern learned later from a sibling's response must not satisfy it: a
+   * predicted entry describes the route's structure but contains none of this
+   * URL's own data, so a task that consumed one would complete without
+   * fetching anything. This happens when more siblings are prefetched in one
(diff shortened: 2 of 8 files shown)


bolechen:
Following up as the author of the production reproduction linked from #96965 (paginated reading app, PPR + partialPrefetching).

Just re-verified on **Next.js 16.3.5** (current production dependency), production build (`next build` + `next start`), chromium, 5x serial repeats: the sibling-prefetch-starvation scenario described in this PR's root cause writeup is still **100% reproducible** (5/5 failures) — clicking a next-page `<Link prefetch>` after a same-tick sibling batch still issues a fresh dynamic request instead of hitting the warmed prefetch (`Expected: 2, Received: 4` RSC requests on our e2e assertion). None of the 22 backports that landed between 16.3.1 and 16.3.5 touched this path.

The root cause description here — a starved task falling back to the optimistic-routing trie's synthetic entry and resolving as Fulfilled without ever fetching the URL's own route tree — matches exactly what we traced in our own reproduction (hover-intent prefetch only ever emits the route-tree signal, never the PPRRuntime body, for any sibling beyond the first).

This has been open a month with no maintainer review (`reviewDecision: REVIEW_REQUIRED`, `mergeStateStatus: BLOCKED`). Given it's blocking real production traffic on two independent apps (per #96965's reports), could someone from the Next.js team take a look? Happy to test against a canary build if that helps move this forward.

PR #97381, opened 2026-08-14
fix(postcss): handle variable-assigned default exports in PostCSS config resolution (fix #97292)

### LINKED ISSUE #97292: Turbopack fails to evaluate variable-assigned default exports in postcss.config.mjs

https://github.com/amoh909/nextjs-turbopack-postcss-bug

### To Reproduce

1. Create a `postcss.config.mjs` file at the root of a Next.js project with the following structure:
   ```javascript
   const config = {
     plugins: {
       "@tailwindcss/postcss": {},
     },
   };
   export default config;
   ```
2. Start the local development server (which executes Turbopack).
3. The following trace will be printed upon loading `globals.css`:
   ```text
   Error: Error evaluating Node.js code
   Error: PostCSS config is undefined (make sure to export an function or object from config file)
       [at Module.init (turbopack:///[turbopack-node]/transforms/postcss.ts?config=[project]/postcss.config.mjs:23:11)]

**Current Behavior:**
The Next.js development server fails to compile CSS files (`globals.css` and `page.module.css`), repeatedly printing a low-level compilation error to the terminal terminal: `Error: Error evaluating Node.js code` followed by `Error: PostCSS config is undefined`. The dev server remains alive, but styles do not process because Turbopack fails to read the config object when it is assigned to an intermediate variable prior to being exported.

**Expected Behavior:**
Turbopack's configuration evaluator should successfully parse and resolve the variable reference for the default export (`const config = ...; export default config;`), mirroring standard Node.js module resolution, allowing the CSS assets to compile cleanly without throwing terminal errors.


```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 16109
  Available CPU cores: 16
Binaries:
  Node: 24.12.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 11.14.0
Relevant Packages:
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

CSS

next dev (local)

### Additional context

The configuration file itself is completely valid JavaScript syntax. Running the module directly through the Node.js runtime successfully resolves and extracts the object:

node -e "import('./postcss.config.mjs').then(console.log)"
# Output: [Module: null prototype] { default: { plugins: { '@tailwindcss/postcss': {} } } }

Changing the file structure to use a direct inline object literal expression (`export default { plugins: { ... } };`) completely bypasses the terminal error loop without needing to revert Next.js versions.

SparshGarg999:
Fixes #97292

## Summary
In `packages/next/src/lib/find-config.ts`, when importing configuration modules (such as `postcss.config.mjs`) via ESM `import()`, `mod.default` is extracted unconditionally. When configuration objects are assigned to intermediate variables prior to being exported (`const config = { ... }; export default config;`), module resolution could return `undefined` when `mod.default` is not defined directly on the module namespace object.

This PR updates `findConfig` to safely inspect module default exports, fallbacking appropriately to ensure PostCSS and other ESM configuration files resolve cleanly.

## How was this tested?
- Existing `find-config.test.ts` test suite covers `.mjs` config resolution.


fufuShih:
Sorry to bother you. I have a some problem.

```ts
mod && typeof mod === 'object' && 'default' in mod && mod.default !== undefined
    ? mod.default
    : mod
```

I was wondering whether the `'default' in mod` check might be redundant.

Since if `default` does not exist, `mod.default` would already be `undefined`.

I may be missing something, so I’d appreciate your advice. Thanks.

SparshGarg999:
Good catch @fufuShih! You're completely right — `'default' in mod` is redundant since `mod.default !== undefined` already evaluates to false if the property is absent (or undefined).

I have updated the PR to remove the redundant `'default' in mod` check and formatted it to comply with Prettier's print width guidelines. Thank you for pointing that out!

PR #97383, opened 2026-08-14
[ci] Fix backport canary release dispatch

ztanner:
## Summary

- Use the job-scoped `GITHUB_TOKEN` with explicit `actions: write` permission to dispatch `trigger_release.yml` after a stable backport.
- Avoid requesting Actions write access from the release GitHub App, which currently prevents the dispatch token from being created.
- Preserve the existing forced preminor canary release behavior.

This fixes the failure that left canary semver-behind stable after the `v16.3.1` backport. The evaluation succeeded, but the dispatch failed while creating the App token in the [failing dispatch job](https://github.com/vercel/next.js/actions/runs/31750536217/job/94615208366).

## Verification
https://github.com/vercel/next.js/actions/runs/31830288265/job/94864091005
```
should_dispatch=true
reason=Dispatching canary preminor release because stable release 16.3.1 is ahead of 16.3.1-canary.17
released_version=16.3.1
current_canary_version=16.3.1-canary.17
dispatch_input=false
auto_dispatch_enabled=true
```


eps1lon (approved):
Had to double check because I confused this with `workflows` permissions which the app already has.

PR #97390, opened 2026-08-14
Shard priority to queue to allow more parallelism

jimmyhmiller:
When doing some profiling I found that this was often a bottleneck. There was work to be done, but workers were contending on the lock. I am certain there are other designs but this was the one that came to my mind.

The tradeoff here is of course that things are not perfectly in priority order now. You can have things on a different shard getting worked before the higher priority one. But I didn't see any correctness issue here.

I did a bunch of different things to make sure this was a reasonable model. I tried some microbenchmarks of different designs. Found this one to work well.

I did a loom model of some key features of this and couldn't find any issues.

I did benchmark against v0/chat. My first set of benchmarks had some real nice wins across 6 interleaved runs.

compile: 52.00s -> 39.18s, -24.67% +/- 2.62% (the "compiled in ..." message)
wall: 60.66s -> 43.76s, -27.89% +/- 2.68% (actual wall time to complete build

Since then I've done various runs and never got that level of boost. Usually around 7-10%. But the number changes is typically the baseline. Wall time of the baseline dropped so did compile time. So not sure if I just pulled latest v0/chat that has some other improvements or what.

Regardless it is an improvement. I've got some follow on changes I think could help performance. One unrelated, but one that does some more concurrency. But still making sure those are ready.

I did have codex write up the concrete implementation here. The loom model did help with some of the things like validating that we did need to change the atomics. I do admit this is a bit inelegant. I am not fixated on this exact setup here. Happy to entertain and implement other solutions. Claude Opus has some micro optimization suggestions that mattered in microbenchmarks. But I didn't see any difference in real ones.

I'm sad the 25% isn't replicatable for me on my macbook now. Even with an identical (as far as I remember) setup. But I think it underscores the need for a better benchamrking setup.

I do think this is hopefully still worth it. But happy to iterate on it.

### DIFF
--- a/turbopack/crates/turbo-tasks/src/priority_runner.rs
+++ b/turbopack/crates/turbo-tasks/src/priority_runner.rs
@@ -1,4 +1,5 @@
 use std::{
+    cell::Cell,
     collections::BinaryHeap,
     fmt::Debug,
     future::Future,
@@ -12,6 +13,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use crossbeam_utils::CachePadded;
 use parking_lot::Mutex;
 use pin_project_lite::pin_project;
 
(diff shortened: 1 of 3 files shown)


turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -12,6 +13,7 @@ use std::{
     time::{Duration, Instant},
 };
 
+use crossbeam_utils::CachePadded;
```
jimmyhmiller, line 16: Crossbeam was already included transitively

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -58,6 +60,132 @@ impl<P: Ord, T> PartialOrd for HeapItem<P, T> {
...
+            len: AtomicUsize::new(0),
+        }
+    }
+
+    fn is_empty(&self) -> bool {
+        self.len.load(Ordering::Acquire) == 0
+    }
+
+    fn push(&self, item: HeapItem<P, T>) -> bool {
+        let index = if self.shards.len() == 1 {
```
lukesandberg: 1. do we need this special case?  `n%1 == 0 `for all `n`
2. can we just reject shard_count ==1?

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -58,6 +60,132 @@ impl<P: Ord, T> PartialOrd for HeapItem<P, T> {
...
+        shrink_amortized(&mut shard);
+        self.len.fetch_sub(1, Ordering::AcqRel);
+        Some(item)
+    }
+
+    fn pop_best_of_two(&self, a: usize, b: usize) -> Option<HeapItem<P, T>> {
+        if a == b {
+            return self.pop_from(a, false);
+        }
+        // try_lock never blocks, so acquiring two locks here cannot deadlock.
```
lukesandberg, line 167: true: but i wonder if a consistent order would improve throuighput?

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -58,6 +60,132 @@ impl<P: Ord, T> PartialOrd for HeapItem<P, T> {
...
+        let mut shard_a = self.shards[a].try_lock();
+        let mut shard_b = self.shards[b].try_lock();
+        let take_from_a = match (
+            shard_a.as_deref().and_then(|s| s.peek()),
+            shard_b.as_deref().and_then(|s| s.peek()),
+        ) {
+            (Some(head_a), Some(head_b)) => head_a.priority >= head_b.priority,
+            (Some(_), None) => true,
+            (None, Some(_)) => false,
+            (None, None) => return None,
```
lukesandberg, line 177: i guess this is a little surprising since it conflates empty with contended.

so popular shards are less likely to be drained?  maybe we need something to balance pushes?

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -83,36 +211,58 @@ impl<
 > PriorityRunner<C, T, P, E>
 {
     pub fn new(executor: E) -> Self {
+        let target_workers = tokio::runtime::Handle::current().metrics().num_workers();
+        Self::with_shard_count(executor, (target_workers * 2).next_power_of_two())
```
lukesandberg: so on my macbook that would be 14 cores => 32 shards

using a power of 2 is useful, but i doubt the `% self.shards.len()` can be strength reduced to shifts... probably not a big deal, but if not, then why power of 2?

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -58,6 +60,132 @@ impl<P: Ord, T> PartialOrd for HeapItem<P, T> {
...
+                .into_boxed_slice(),
+            len: AtomicUsize::new(0),
+        }
+    }
+
+    fn is_empty(&self) -> bool {
+        self.len.load(Ordering::Acquire) == 0
+    }
+
+    fn push(&self, item: HeapItem<P, T>) -> bool {
```
lukesandberg, line 114: nit: document the return value

returns true if this push takes the queue from empty-> non-empty?

lukesandberg (commented):
The sharding is a good idea

having each thread push/pull from ever changing shards seems a little surprising, but i guess we get 'probabalistic balancing' which is nice

previously i considered sharding by priority band with the observation that in the cold build case everything has the same priority

the other 'standard' is thread local queues with cross thread stealing on empty, that one optimizes lock contention in the common case.  that would be as simple as assigning threads a single shard index for writing and switch to scanning when empty

turbopack/crates/turbo-tasks/src/priority_runner.rs
```diff
@@ -58,6 +60,132 @@ impl<P: Ord, T> PartialOrd for HeapItem<P, T> {
...
+fn shard_rand() -> u32 {
+    SHARD_RNG.with(|cell| {
+        let mut x = cell.get();
+        x ^= x << 13;
+        x ^= x >> 17;
+        x ^= x << 5;
+        cell.set(x);
+        x
+    })
+}
```
bgw, line 82: Did you get this algorithm from somewhere? Can we link a wikipedia page? Is this meaningfully faster than https://docs.rs/rand/latest/rand/rngs/struct.Xoshiro128PlusPlus.html ? We already pull in `rand` in a bunch of places.

jimmyhmiller:
I changed some tests of the change to have min 2 shards means we are never fully guaranteeing order. And these tests were making sure that everything was always in order. Given the revisit I'm doing for cold builds where apparently everything is always at the same priority, I'm going to leave them for now. Not sure exactly how I'd test that they are in "kind of priority order". If people have ideas I'm open to them.

@lukesandberg 


> previously i considered sharding by priority band with the observation that in the cold build case everything has the same priority

> the other 'standard' is thread local queues with cross thread stealing on empty, that one optimizes lock contention in the common case. that would be as simple as assigning threads a single shard index for writing and switch to scanning when empty

I originally missed this. But the meeting we had was suggesting the same kinds of things. I will explore that.

turbopack/crates/turbo-tasks/src/priority_runner.rs
jimmyhmiller: Sorry, yeah it is just an xorshift taken straight from wikipedia. I think for this simple case it is probably sufficient and a bit faster than your linked on since it is 32bit. Happy to change if you want, but added a comment and link. https://en.wikipedia.org/wiki/Xorshift

turbopack/crates/turbo-tasks/src/priority_runner.rs
jimmyhmiller: added

turbopack/crates/turbo-tasks/src/priority_runner.rs
jimmyhmiller: Fix it so it will be 16 instead of 32. Need to be more thorough on benchmarks to be sure that is good. But I did change to required powers of two with the shift and that made a big enough difference on microbenchmarks I kept it that way.

turbopack/crates/turbo-tasks/src/priority_runner.rs
jimmyhmiller: I will have to think a bit more about this. I think it's a good point. Given that I'm revisiting the sharded stuff I will think about it once I test other things. (You weren't in the meeting, but apparently in initial builds there are no real priorities, so I'm testing some alternatives with that in mind)

turbopack/crates/turbo-tasks/src/priority_runner.rs
jimmyhmiller: In my little benchmarks, it didn't seem to make a difference.

turbopack/crates/turbo-tasks/src/priority_runner.rs
jimmyhmiller: I changed to only powers of two so got rid of this. But I'll be honest I'm not sure how I feel about it. It feels a bit weird to always shard. But as mentioned in another comment, given I am revisiting some things, leaving for now. I don't think it makes or breaks anything.

jimmyhmiller:
Closing as there is not exact clear path right now

PR #97391, opened 2026-08-14
fix(use-cache): avoid retaining composite AbortSignal after prerender…

### LINKED ISSUE #97363: `use cache` wrapper never releases its AbortSignal.any composite, retaining every cached render

https://github.com/devopsaptlife/nextjs-prerender-abort-retention

### To Reproduce

```bash
git clone https://github.com/devopsaptlife/nextjs-prerender-abort-retention
cd nextjs-prerender-abort-retention
npm install
npm run build
node --expose-gc --inspect=127.0.0.1:9231 --require ./preload-any.cjs .next/standalone/server.js &
for i in $(seq 1 40); do curl -s -o /dev/null "http://localhost:3000/p/f-$i"; done
node check-composites.mjs
```

```
composites created: 2   STILL ALIVE after GC: 2
```

Every composite the `use cache` wrapper creates is still alive after three forced full GCs.
`preload-any.cjs` wraps `AbortSignal.any` and holds only `WeakRef`s, so it cannot itself be what
retains them.

Three ingredients are all required, and each one silently hides the bug if missing:
`cacheComponents: true`, a `"use cache"` boundary, and **`generateStaticParams` on the route** —
the last one is what gives the route a fallback-shell prerender, which is the pass that runs with
dynamic-access tracking, which is what makes `dynamicAccessAbortSignal` defined and therefore
makes the wrapper build a composite at all. Without it the wrapper takes the
`: timeoutAbortController.signal` branch and nothing leaks.

**Current.** `dist/server/use-cache/use-cache-wrapper.js` builds one composite per cached
evaluation and never releases it:

```js
const abortSignal = dynamicAccessAbortSignal
  ? AbortSignal.any([dynamicAccessAbortSignal, timeoutAbortController.signal])
  : timeoutAbortController.signal;
const { prelude } = await prerender(…, { signal: abortSignal, … });
clearTimeout(timer);      // the timer is cleaned up — the listener is not
```

Node keeps a composite from `AbortSignal.any()` in its **process-global `gcPersistentSignals`
Set** for as long as any abort listener is attached to it, so that listener can still fire.
`prerender()` attaches one, and on the normal completion path nothing ever aborts — so the
composite is never released, and through its listener closure it retains that whole cached
render: the entry `{html, rscData, segmentData Buffers}`, its streams and its promises.

**Expected.** A finished `use cache` evaluation should be collectable. The wrapper already cleans
up its timer on the same line; the listener needs the same treatment.

**This is not the Node bug.** [nodejs/node#64476](https://github.com/nodejs/node/issues/64476)
covers the case where the composite *aborts* and is fixed in v26.7.0 (no 24.x backport). The case
here — the composite never aborts — is still present on v26.7.0, and upstream Node considers the
pinning deliberate: an observed composite whose source is still alive has to stay reachable so it
can fire later. Their stated resolution is *remove the listener*, which only the code that
attached it can do.

### Why it matters

Production Next 16.3.0, from a full V8 heap snapshot with retainer paths over **strong edges only**
(weak edges retain nothing; including them produces plausible root paths that keep nothing alive):

| scale | measurement |
|---|---|
| production heap snapshot | **1,546** composites pinned in `gcPersistentSignals` vs **1,589** retained cached page values — one per entry |
| a real app on a local bench | **49 of 49** composites alive after three forced full GCs |
| isolated loop | **200,000 of 200,000** retained, exactly linear, zero released by time, timers or repeated GC |

Retainer chain:

```
gcPersistentSignals -> composite AbortSignal -> kEvents -> Listener -> listener closure
  -> Context -> AbortController -> { html, kind, postponed, rscData, segmentData, status }
  -> segmentData Map -> Buffer -> ArrayBuffer
```

Symptom on that deployment: **~+0.45 GB/hour, linear**, to a kernel cgroup OOM every 9–10 hours.

```
Operating System:
  Platform: linux / darwin (both reproduce)
  Arch: x64 / arm64
Binaries:
  Node: 24.19.0 (prod), 24.15.0 (bench)
Relevant Packages:
  next: 16.3.0
  react: 19.2.7
  react-dom: 19.2.7
Next.js Config:
  output: standalone
  cacheComponents: true
  partialPrefetching: true
```

`use cache`, Partial Prerendering (PPR), Runtime

### Additional context

**Eight remedies measured; seven do not work.** Posting these so nobody re-derives them:

| action | result |
|---|---|
| as-is | PINNED |
| abort the per-render controller after the render | PINNED — and **worse**, it pins the source signal too |
| abort the outer (dynamic-access) signal | PINNED |
| `{ once: true }` without an abort | PINNED — nothing fires, so nothing removes it |
| hand-rolled composite instead of `AbortSignal.any` | PINNED — a listener on a long-lived source leaks the same way |
| Node ≥ 26.7.0 | fixes only the composite-aborts case |
| don't build the composite at all | released, but an aborted outer render would no longer stop early |
| **remove the listener when the render finishes** | **released** |

Two more that look equivalent and are not, both measured: `delete signal.stack`-style removal via
`removeEventListener` on **some** listeners is useless (the delete requires the listener count to
reach zero), and retention is **binary** — a `stackTraceLimit` of 10 / 5 / 3 / 1 all retain the
full graph; only 0 releases it.

**A workaround that works**, if it helps shape the real fix — capture what `prerender()` attaches
and detach it where the timer is already cleared:

```js
const detach = [];
if (dynamicAccessAbortSignal) {
  const add = EventTarget.prototype.addEventListener;
  abortSignal.addEventListener = function (...args) {
    if (args[0] === 'abort') detach.push(args);
    return add.apply(this, args);
  };
}
const { prelude } = await prerender(…);
clearTimeout(timer);
delete abortSignal.addEventListener;
for (const args of detach) abortSignal.removeEventListener(args[0], args[1], args[2]);
```

Measured on a real app: composites alive after a forced GC went from **49/49 (100%)** to
**0/17 (0%)**, stable across three crawl rounds, and the `arrayBuffers` ratchet stopped — 16 MB
flat versus 38 → 48 → 54 MB before. Abort behaviour *during* the render is unchanged; the
interceptor only records and forwards. It does not cover the path where `prerender()` throws.

Related: #84648 is the same symptom reported without a diagnosis, and #97351 is the *amplifier* —
the abort-reason `Error` keeps its V8 stack frames, which multiplies the cost of each retained
render by ~1,700×. That one is real but separate: fixing it alone leaves the retention in place,
which I confirmed by shipping it to production and measuring no change in slope.

ekley:
### What?

prerender was receiving a composite AbortSignal that could retain completed cached renders after the render had finished.

This change gives the prerender operation an owned AbortController and forwards aborts from the timeout and dynamic access signals to it, avoiding the need to monkey-patch AbortSignal methods.

### Why?

AbortSignal.any can retain its composite signal while listeners remain attached. Since prerender attaches an abort listener to the signal it receives, the composite signal could keep the completed cached render reachable after prerendering finished, resulting in memory retention.

### How?

Create a dedicated AbortController for the prerender operation and forward aborts from the timeout and dynamic access signals:

```ts
const abortController = new AbortController()
const abortSignal = abortController.signal

const onTimeoutAbort = () => {
  abortController.abort(timeoutAbortController.signal.reason)
}

timeoutAbortController.signal.addEventListener('abort', onTimeoutAbort)
```

The dynamic access abort signal is handled in the same way. Once prerendering finishes, the forwarding listeners are removed in the finally block so the signals can be released cleanly.

This keeps the existing abort behaviour while avoiding monkey-patching and explicitly managing the lifecycle of the signal used by prerender.

### Testing
You can create a `.tgz` file from the Next.js repo by simply going to the `packages/next` directory and running:

```bash
pnpm run build
or
pnpm run build
or
yarn build
```

And once the build has finished, you need to run:

```bash
pnpm pack
```

Then clone this repo: `https://github.com/devopsaptlife/nextjs-prerender-abort-retention`, and from the root directory of the repo, run:

```bash
1. pnpm install
2. pnpm build
3. node --expose-gc --inspect=127.0.0.1:9231 --require ./preload-any.cjs .next/standalone/server.js &
for i in $(seq 1 40); do curl -s -o /dev/null "http://localhost:3000/p/f-$i"; done
4. node check-composites.mjs
```

And you will see:

```bash
composites created: 2   STILL ALIVE after GC: 2 
```
or
```bash
composites created: 1   STILL ALIVE after GC: 1 
```

which means GC doesn't collect the composite cause is still alive

```bash
2. rm -rf .next # delete .next folder cache
2. pnpm remove next
3. pnpm add YOUR_LOCAL_NEXT_JS_FOLDER_PATH/packages/next/next-16.3.1-canary.17.tgz
4. pnpm run build
5. node --expose-gc --inspect=127.0.0.1:9231 --require ./preload-any.cjs .next/standalone/server.js &
for i in $(seq 1 40); do curl -s -o /dev/null "http://localhost:3000/p/f-$i"; done
6. node check-composites.mjs
```

And you should see:

```bash
composites created: 1   STILL ALIVE after GC: 0
```

Fixes [#97363](https://github.com/vercel/next.js/issues/97363)

### DIFF
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -1386,32 +1386,67 @@ async function generateCacheEntryImpl(
       const dynamicAccessAbortSignal =
         dynamicAccessAsyncStorage.getStore()?.abortController.signal
 
-      const abortSignal = dynamicAccessAbortSignal
-        ? AbortSignal.any([
-            dynamicAccessAbortSignal,
-            timeoutAbortController.signal,
-          ])
-        : timeoutAbortController.signal
+      // `prerender` attaches an abort listener to the signal we hand it. Node
+      // pins an `AbortSignal.any` composite in its global `gcPersistentSignals`
+      // set for as long as any listener is attached, so a composite would keep
+      // the whole cached render alive after it finishes. Own the signal instead
+      // and forward the two sources through listeners we can remove once the
+      // render settles.
+      const abortController = new AbortController()
+      const abortSignal = abortController.signal
 
-      const { prelude } = await prerender(
-        resultPromise,
-        clientReferenceManifest.clientModules,
-        {
-          environmentName: 'Cache',
-          filterStackFrame,
-          signal: abortSignal,
-          temporaryReferences,
-          onError(error) {
-            if (abortSignal.aborted && abortSignal.reason === error) {
-              return undefined
-            }
+      const onTimeoutAbort = () => {
+        abortController.abort(timeoutAbortController.signal.reason)
+      }
+      timeoutAbortController.signal.addEventListener('abort', onTimeoutAbort)
 
-            return handleError(error)
-          },
+      let onDynamicAccessAbort: (() => void) | undefined
+      if (dynamicAccessAbortSignal) {
+        onDynamicAccessAbort = () => {
+          abortController.abort(dynamicAccessAbortSignal.reason)
         }
-      )
+        dynamicAccessAbortSignal.addEventListener('abort', onDynamicAccessAbort)
+        // A composite also aborts immediately when a source is already aborted
+        // at creation time.
+        if (dynamicAccessAbortSignal.aborted) {
+          abortController.abort(dynamicAccessAbortSignal.reason)
+        }
+      }
 
-      clearTimeout(timer)
+      let prelude: ReadableStream<Uint8Array>
+      try {
+        prelude = (
+          await prerender(
+            resultPromise,
+            clientReferenceManifest.clientModules,
+            {
+              environmentName: 'Cache',
+              filterStackFrame,
+              signal: abortSignal,
+              temporaryReferences,
+              onError(error) {
+                if (abortSignal.aborted && abortSignal.reason === error) {
+                  return undefined
+                }
+
+                return handleError(error)
+              },
+            }
+          )
+        ).prelude
+      } finally {
+        clearTimeout(timer)
+        timeoutAbortController.signal.removeEventListener(
+          'abort',
+          onTimeoutAbort
+        )
+        if (dynamicAccessAbortSignal && onDynamicAccessAbort) {
+          dynamicAccessAbortSignal.removeEventListener(
+            'abort',
+            onDynamicAccessAbort
+          )
+        }
+      }
 
       if (timeoutAbortController.signal.aborted) {
         // When the timeout is reached we always error the stream. Even for


unstubbable:
superseded by #97476

PR #97393, opened 2026-08-14
Add experimental parameter matching APIs

gnoff:
## Summary

Add per-parameter matching directives behind `experimental.paramMatching`, with Cache Components enabled. `generateStaticParams` remains responsible for concrete build-time prerenders; matching configuration controls how requests with novel parameter values are handled.

This lets an application close a prefix to its build-time values, require blocking generation for the next parameter, and serve immediate fallback UI for another. Unconfigured parameters retain inference from the generated shells, and an unconfigured tail without examples remains dynamic.

## Stack

The prerequisites are ordered bottom to top:

1. #98944: use structural server React keys and remove postponed-state param interpolation; this supersedes the closed #98890 catch-all placeholder change.
2. #98891: generalize the gSP work-unit store to `build-time-generator` and retain the originating function name for diagnostics, without changing existing gSP behavior.
3. #98950: prune seeded RDC entries that depend on unknown roots and suspend fresh root-dependent cache fills independently, including nested caches. This replaces the closed #98892.
4. #98964: keep closed-route admission independent of ISR cache contents. This fixes existing `dynamicParams = false` revalidation and empty-cache behavior without changing manifest or adapter contracts.
5. This PR: add the opt-in matching API and its validation/runtime integration, including the root-fallback query-key correction formerly isolated in #98893.

The prerequisites can land without committing to this API. The closed-parameter prediction fix #98889 and the earlier dev-validation prerequisites #98511 and #98512 have already merged. Opt-in diagnostic logging remains separately committed and can be removed independently.

The prerequisites remain independently reviewable above #98944. This PR preserves its existing history through integration merges; generic generator diagnostics live in #98891, while this layer supplies the matching API's name and integration coverage. The superseded catch-all array conversion is not reintroduced.

## Example

```ts
// next.config.ts
export default {
  cacheComponents: true,
  experimental: { paramMatching: true },
}

// app/[lang]/layout.tsx
export const experimental_paramMatching = {
  lang: 'not-found',
}

// app/[lang]/catalog/[top]/items/[bottom]/page.tsx
export const experimental_paramMatching = {
  top: 'blocking',
  bottom: 'fallback',
}

export function generateStaticParams() {
  return [{ lang: 'en', top: 't1', bottom: 'b1' }]
}
```

The build prerenders `/en/catalog/t1/items/b1`. A novel `lang` returns 404; `/en/catalog/t2/items/b2` blocks on generation; `/en/catalog/t1/items/b2` can receive fallback UI immediately.

The generated form, `export async function experimental_generateParamMatching()`, returns the same partial object and takes no arguments. It can load matching policy from an external configuration system and runs independently of gSP's per-parent-value enumeration. Each unique module is evaluated once per route evaluation, not once globally per build: a shared layout can run again for another page. Repeated occurrences of the same module within one route's parallel tree share the result. It uses the same build-time work-unit context as gSP: public `use cache` helpers work, while request-only/private caching reports the normal unsupported-context error. The context explicitly identifies `experimental_generateParamMatching`, so an invalid `headers()` call names that export rather than gSP. The same applies to private caching, revalidation, and unavailable root parameters; matching generators are not told to obtain concrete params from parent gSP calls.

Ordinary TypeScript object literals work without `as const`. For optional strict mode checking, use `satisfies ParamMatching` imported from `next`. `ParamMatching<'lang' | 'top'>` additionally constrains keys when writing an explicit annotation. Generated types still reject keys outside the module's visible parameters; runtime validation rejects invalid modes and malformed JavaScript or generated exports.

## Layering and validation

Fragments are assignment-merged from layouts toward each page. Descendants replace only supplied keys. A module may configure only parameters at or above its own segment: `[lang]/layout.tsx` cannot configure a later `top` parameter.

The merged explicit policy must follow `not-found → blocking → fallback → dynamic`. Every parameter preceding an explicit `not-found` must also be explicitly `not-found`, directly or through inheritance. Next.js never silently closes an unconfigured prefix or rewrites inherited intent:

```ts
// Ancestor
{ lang: 'blocking' }

// Descendant: invalid after merging
{ category: 'not-found' }

// Descendant: explicitly overrides the affected prefix
{ lang: 'not-found', category: 'not-found' }
```

Omitting `lang` entirely is also invalid in that example. A configuration on the `[lang]` layout can supply it to all descendants.

Parallel branches sharing a matcher must agree and produce a coherent combined policy in URL-parameter order, independent of route groups. Routes sharing a parameter definition must all explicitly configure/inherit its `not-found` policy if any of them does. Explicitly dynamic parameters cannot appear in gSP output or have prerendered parameters beneath them.

Only after explicit configuration is validated are holes filled by inference, constrained by the explicit boundaries.

## Static shells, hints, and development

An inferred empty shell may result in blocking behavior instead of a servable fallback. Explicit `fallback` requires a nonempty shell and cannot silently become blocking. Blocking also validates a reachable useful shell; when no example reaches the boundary, the generic shell provides that validation. `instant = false` opts out of shell validation, not best-effort hint collection.

An explicit fallback moves the required validation to that more generic shell. For example, `top: 'fallback'` requires useful UI with both `top` and `bottom` unknown, even if gSP also supplies `t1/b1`. Once that shell passes, descendant prerenders do not repeat the empty-shell requirement; making more params known does not add a new unknown-data boundary.

Redundant blocking renders are skipped only for API-configured routes when a descendant covers them. Unconfigured routes retain their existing build renders and hint collection. A blocking boundary without an example still gets a best-effort generic render; this does not make that result a servable fallback.

Dev preserves required-or-completed shell selection and adds explicit fallback boundaries to the unknown parameter set. For example, with a `t1/b1` prerender and `top: 'fallback'`, validation treats both parameters as unknown even for `/t1/b1`. Ordinary dev requests remain dynamic renders. Dev supplies exact closed-parameter names from the effective policy, while production persists them in renderer manifest metadata. The shared transport-tree builder marks only those dynamic nodes. For `{ lang: 'not-found', slug: 'fallback' }`, only `[lang]` is marked, even though the generic matcher has `fallback: false`; `[slug]` remains open. Successful document loads, live navigations, and prefetched trees therefore disable optimistic prediction consistently. No dev prerender-manifest entry or new adapter routing directive is needed.

## Deployment and diagnostics

Route admission is checked before consulting ISR. Exact build-generated URLs remain allowed when their cached content expires or disappears; other URLs are admitted by their most specific matcher. This permits runtime generation beneath an allowed closed prefix without admitting an unknown prefix. Cache contents do not grant permission to render, and an application-rendered `notFound()` result does not permanently remove an admitted URL. The prerequisite implements the shared runtime boundary; this layer supplies the equivalent development policy outcome.

Closed-prefix output uses the existing fallback-false adapter contract, with specialized matchers admitting the open suffix beneath each generated prefix. The function output keeps its original source-page entrypoint. Synthetic alias entrypoints and the alternate app-path-routes manifest have been removed; no new adapter/proxy contract is introduced. Local output/launcher checks do not replace a real grouped-function deployment check.

Explicit root fallback matching makes a previously unreachable adapter case possible: a servable fallback whose root parameter is still unknown. For `app/[lang]/layout.tsx` with `lang: 'fallback'`, the fallback's query allowlist must contain `nxtPlang`, matching the routing destination, rather than the logical name `lang`. This correction now lives with the API instead of in #98893. The real `param-matching-root-params` fixture checks the exact route and non-empty key list for HTML, RSC, and segment outputs, and confirms the HTML output has a servable fallback. The synthetic manifest-editing fixture and direct internal-emitter invocation have been removed.

Initially, `output: export` requires every parameter to be explicitly `not-found`. This avoids silently changing matching semantics and leaves room for future client-only fallback support.

Diagnostics print effective matching decisions and deployment patterns only when both `experimental.paramMatching` and `NEXT_PRIVATE_DEBUG_PARAM_MATCHING=1` are enabled.

## Verification and remaining work

- Latest review follow-up is included in `19d60ae82c0` and `9dd593abb6d`. The explicit type-fixture project passes all three checks in both Turbopack dev and production, including rejection of invalid parameter keys. The three build-contract suites pass all 28 production tests; matching unit coverage passes all 98 tests. Repository TypeScript, changed-file ESLint/formatting, and the routing README language check pass. Verification used a full bootstrap with a fresh native Turbopack build. Comments now describe normalized-prefix identity, per-route generator evaluation, and the purpose of build-artifact assertions. Explicit fallback shell validation and removal of synthetic adapter aliases are documented above.

- Root fallback query-key follow-up: the genuine API fixture now requires a servable `/[lang]` fallback, matches its exact routing rule, and requires `['nxtPlang']` on HTML, RSC, and segment outputs. Temporarily removing only the prefix conversion fails that assertion with `['lang']` while the seven behavior tests still pass. Restoring it passes all eight fixture tests; the existing blocking-query fixture also passes (11 production Webpack tests combined). Development Webpack also passes (six tests, two production-only skips). Full bootstrap, repository TypeScript checking, and changed-file lint pass. This checks the real build's adapter contract, not a live deployment; the separate intercepted-parameter (`nxtI`) investigation is deferred.

- Root-cache coverage now uses ordinary builds and requests only. The prerequisite fixture no longer exposes shell-debugging APIs. A real `revalidateTag` request in this layer forces a root-fallback shell to regenerate, and checks that cached root-dependent UI remains deferred while independent cached content is refreshed. Both fixtures pass together in production Webpack (11 tests) and development Webpack (seven tests, four build/ISR-only skips). Removing just the fresh-fill handling, while retaining RDC pruning, makes normal root-fallback requests and the revalidation regression return opaque placeholder content; restoring it makes them pass.

- Generator-name follow-up: all nine new production Webpack diagnostic fixtures pass. The existing matching-generator and gSP-error suites also pass (13 production tests), and the existing matching-generator suite passes in dev (seven tests, one prefetch-only skip). Full bootstrap, repository types, and changed-file lint pass. An initial fixture-only import error was corrected before the successful diagnostic run.

- Restack verification on the combined #98944-based stack: full build and repository type check pass, along with 222 focused unit tests (four snapshots), 43 production Webpack tests, and 26 development Webpack tests (nine snapshots). Existing mode-specific skips are unchanged. That run used the native build available at the time. Subsequent simplification verification rebuilt native Turbopack locally and passed the production file-tracing suite (six tests, five snapshots).

- Node-local hint coverage passes with both bundlers: eight generator-fixture tests in production, and seven in dev with the prefetch-only case skipped. The mixed closed-prefix/open-suffix fixture verifies only `[lang]` is marked, novel slugs still render, novel languages return 404, and the renderer manifest identifies the exact closed subset. With inlining disabled, a prefetched navigation uses no additional requests and still marks only `[lang]`, proving the existing response merge preserves the transport-tree hints. Closed-parameter bits are not duplicated in the build-measured `prefetch-hints.json` tree; `createUniformHintTree` remains unchanged from the stack's base. The legacy fixture covers the inlined path and passes all eight production tests with both bundlers. Both fixtures also pass together in dev (12 passed, four prefetch-dependent cases skipped per bundler). The three unchanged Cache Components prediction regressions passed with both production bundlers in the preceding node-local-hint verification.
- Regression fixtures demonstrate before/after failures for missing closed prefixes, public cached generators, successful dev closure hints, and widened object-literal types. Runtime invalid-mode/key checks and private-cache rejection remain covered.
- The six focused production Webpack suites pass (29 tests), covering generators, types, prefix validation, blocking/partial-prefetch behavior, cached root parameters, and parallel route groups. The main production Turbopack API suite passes (26 tests).
- The main API and existing `cache-components-dev-fallback-validation` suites pass together in Turbopack development (26 tests, 9 snapshots).
- Full bootstrap build and repository TypeScript checks pass. Both touched unit suites pass in the broad unit run; that run has three unrelated macOS `/var` versus `/private/var` path-comparison failures in `runTypeScriptCli.test.ts`.
- The four decoded-value expectations in this PR's new fixture were incorrect. Freshly built canary `34433fd12ee` reproduces encoded server values without matching configuration, and its existing `prerender-encoding` test explicitly expects them. The follow-up corrected those expectations without changing runtime encoding.
- Cache-state integration `398966d753a` incorporates #98964 and adds five API lifecycle cases. Fully closed and partially closed seeded paths, runtime-generated open suffixes, shared dynamic-tail shells, and invalidated unlisted URLs preserve their admission semantics. The routing and runtime suites pass 44 tests with two existing percent-sign URL failures in each production run: fresh-native Turbopack with Partial Prefetching enabled, and Webpack with it disabled. Fully closed-route revalidation and expected-404 logging now pass. Development passes all 25 runnable routing cases (11 production-only skips). The standalone prerequisite passes all 16 new tests with both production bundlers; the existing legacy closed-parameter suite passes eight production tests and five dev tests (three production-only skips). Full bootstrap, repository TypeScript, and changed-file lint/formatting pass. The two encoding failures remain open; this is not a claim that the API is ready to merge.
- Actual grouped-function deployment verification is blocked locally because `VERCEL_ADAPTER_TEST_TOKEN` and `VERCEL_ADAPTER_TEST_TEAM` are not configured.


packages/next/src/build/static-paths/prerender-matcher.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  let previousPhase = -1
+  let previousParamName: string | undefined
+  for (const { paramName } of pathnameSegments) {
+    const mode = policy[paramName]
+    if (!mode) continue
+    const currentPhase = PRERENDER_PARAM_MODES.indexOf(mode)
+    if (currentPhase < previousPhase) {
+      throw new Error(
+        `Invalid parameter matching for "${page}": parameter "${paramName}" uses "${mode}" after parameter "${previousParamName}" uses a later matching phase. Expected parameters to follow not-found, blocking, fallback, then dynamic order.`
```
lubieowoce, line 141: > Expected parameters to follow not-found, blocking, fallback, then dynamic order.

This reads weird, might use some quotes or backticks or something

packages/next/src/build/static-paths/prerender-matcher.ts
```diff
@@ -0,0 +1,210 @@
...
+        }
+      }
+      return value
+    })
+  )
+
+  // Include branches without matching exports: they still inherit their
+  // ancestors' policies. An override in one branch must not erase that intent
+  // from a sibling. Keep module evaluation above separate from tree occurrences
+  // so a shared generator is still called only once.
```
lubieowoce, line 73: This is hard to understand

No idea what "module evaluation" is referring to here, or why a shared generator is relevant

> Keep module evaluation above separate from tree occurrences so a shared generator is still called only once.

i guess this is trying to say "if multiple segments use the same generator because of inheritance, we should still only call it once" or something?

packages/next/src/build/static-paths/prerender-matcher.ts
```diff
@@ -0,0 +1,210 @@
...
+  const treePaths = segments.flatMap((segment) => segment.treePaths)
+  const branchPaths = treePaths.filter(
+    (treePath) =>
+      !treePaths.some(
+        (other) =>
+          other.length > treePath.length && isTreePathPrefix(treePath, other)
+      )
+  )
```
lubieowoce, line 81: what's a branch path? this looks like it's meant to be leaf segments (like page.tsx) but i'm not sure. also why can't we have the segments in a tree instead of juggling prefix paths? this feels like it should be a tree traversal

packages/next/src/server/base-server.ts
```diff
@@ -2778,11 +2804,27 @@ export default abstract class Server<
...
             )
-            addRequestMeta(
-              req,
-              'stagedFallbackParams',
-              getStagedFallbackParams(matchedRoute)
-            )
+            let stagedFallbackParams = getStagedFallbackParams({
+              ...matchedRoute,
+              // Matcher-only entries have no source shell to validate. Stage
+              // the completed shell, retaining any always-dynamic tail.
```
lubieowoce, line 2810: any what

lubieowoce: also what does "stage the completed shell" mean. are we making some params act like they're not fallback? or what

packages/next/src/server/config-shared.ts
```diff
@@ -486,6 +486,12 @@ export function resolveCssChunkingMode(
 }
 
 export interface ExperimentalConfig {
+  /**
+   * Enables per-parameter prerender matching directives in App Router layouts
```
lubieowoce, line 490: "directives" seems a bit strange bc we call `"use client"` etc directives.  normally we call these segment configs i think?

packages/next/src/build/index.ts
```diff
@@ -3894,7 +3905,11 @@ export default async function build(
                   remainingPrerenderableParams:
                     route.remainingPrerenderableParams,
                   throwOnEmptyStaticShell:
-                    prerenderCandidate?.throwOnEmptyStaticShell,
+                    // Matcher-only entries have no source shell to validate.
+                    // Their runtime renders stage the completed shell instead.
```
lubieowoce, line 3909: perhaps i'm missing context but: what's a matcher-only entry? what's a non-matcher-only entry?

also same question about what "stage the complete shell" means as elsewhere. kinda sounds like a robot thought that when we say "staged rendering", we mean it like "staging a play", and is using "stage the params" in the same way (which does not make sense)

packages/next/src/build/static-paths/prerender-matcher.ts
gnoff: :robot: Updated in ac5b30f3118. The diagnostic now says `Expected parameters in this order: "not-found", "blocking", "fallback", then "dynamic".` The unit and fixture assertions use the same wording.

packages/next/src/build/static-paths/prerender-matcher.ts
gnoff: :robot: Clarified in ac5b30f3118. This refers to the loader tree containing multiple occurrences of the same segment module, not to calling a generator for each descendant that inherits its policy. Those occurrences share an AppSegment: we call experimental_generateParamMatching once per module per route and reuse its result at each occurrence. Added a regression that asserts one invocation while keeping the occurrences and their descendants independent.

packages/next/src/build/static-paths/prerender-matcher.ts
gnoff: :robot: Replaced the prefix-path reconstruction with a tree traversal in ac5b30f3118. Collection retains the actual parent/child relationships alongside the existing deduplicated segment list. Each branch inherits its own policy map, and we compare the effective policies at the leaves, including leaves without matching exports. The flat list still drives generateStaticParams unchanged. Existing parallel-inheritance fixtures pass with both bundlers.

packages/next/src/server/base-server.ts
gnoff: :robot: Reworded in ac5b30f3118. Setting this to false allows getStagedFallbackParams to remove the remaining prerenderable parameters from the deferred set, so their values can resolve during the static phase. Parameters that cannot be prerendered remain deferred until the dynamic phase. The comment now describes those phases explicitly instead of saying "stage the completed shell" or "always-dynamic tail"; the rendering logic is unchanged.

packages/next/src/server/config-shared.ts
gnoff: :robot: Changed the description to "parameter matching segment configuration" in ac5b30f3118 and made the Cache Components requirement explicit.

packages/next/src/build/index.ts
gnoff: :robot: There are separate collections for route matching patterns and build-time prerenders. For example, /en/[top] with top configured as blocking still needs a matching entry, but has no build-time fallback output. That is what the old "matcher-only" wording referred to. In 1ff11df7cf3 I replaced this comment with an explanation that false permits remaining prerenderable params to resolve during the static phase, and scoped the metadata to PPR routes. ac5b30f3118 removes the same ambiguous wording from the dev-server comments.

PR #97395, opened 2026-08-15
Turbopack: Split the read and write codepath data structures for symlinks

bgw:
Another attempt at cleanup, similar to https://github.com/vercel/next.js/pull/96955.

We need different information to write symlinks or junction points (most notably, if it's a directory or not) than we get when reading symlinks.

We were trying to use the same data structure for both, but this lead us to always call `get_type()` as part of reading a link, which was just silly.

This also meant that `write_link` had to handle `NotFound` and `Invalid`. `NotFound` is somewhat reasonable (just delete the file), but `Invalid` is awkward.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/content.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/content.rs
@@ -214,14 +214,18 @@ impl LinkTarget {
 pub enum LinkContent {
     /// A valid symbolic link pointing to `target`.
     Link { target: LinkTarget },
-    /// There is no file at this path.
-    NotFound,
-    /// There is no usable symbolic link at this path or some other error occurred.
+    /// The link itself does not exist at the path given to [`FileSystemPath::read_link`].
     ///
-    /// A relative target that steps out of the root and back into it is `Invalid`, since resolving
-    /// it would need the names of the root's own ancestors, which a root-relative path doesn't
-    /// carry.
-    Invalid,
+    /// This says nothing about whether the link's target exists: a dangling link is still
+    /// returned as [`LinkContent::Link`].
+    NotFound,
+    /// The link could not be read.
+    ///
+    /// This includes all I/O errors other than `NotFound`, denied paths, and targets that leave the
+    /// filesystem root. A relative target that steps out of the root and back into it is also
+    /// invalid, since resolving it would need the names of the root's own ancestors, which a
+    /// root-relative path doesn't carry.
+    Invalid { reason: RcStr },
 }
 
 #[turbo_tasks::value_impl]
@@ -246,7 +250,7 @@ impl LinkContent {
                 LinkTarget::Relative { raw, resolved: _ } => SimplifiedLinkContent::Relative(&raw),
             },
             LinkContent::NotFound => SimplifiedLinkContent::NotFound,
-            LinkContent::Invalid => SimplifiedLinkContent::Invalid,
+            LinkContent::Invalid { reason: _ } => SimplifiedLinkContent::Invalid,
         };
         Ok(Vc::cell(RcStr::from(deterministic_hash(
             &salt.await?,
--- a/turbopack/crates/turbo-tasks-fs/src/content.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/content.rs
@@ -224,6 +224,38 @@ pub enum LinkContent {
     Invalid,
 }
 
+#[turbo_tasks::value_impl]
+impl LinkContent {
+    /// Hashes the link itself (its target and type), not the content of whatever the link points
+    /// at. This mirrors [`FileContent::hash`] and is the right content hash for consumers that
+    /// re-create a symlink as a symlink instead of copying the resolved file.
+    #[turbo_tasks::function]
+    pub async fn hash(&self, salt: Vc<RcStr>, algorithm: HashAlgorithm) -> Result<Vc<RcStr>> {
+        #[derive(DeterministicHash)]
+        enum SimplifiedLinkContent<'a> {
+            Absolute(&'a RcStr),
+            Relative(&'a RcStr),
+            NotFound,
+            Invalid, // the actual error message doesn't matter for this API
+        }
+        let simplified = match self {
+            LinkContent::Link { target } => match target {
+                LinkTarget::Absolute { resolved } => {
+                    SimplifiedLinkContent::Absolute(&resolved.path)
+                }
+                LinkTarget::Relative { raw, resolved: _ } => SimplifiedLinkContent::Relative(&raw),
+            },
+            LinkContent::NotFound => SimplifiedLinkContent::NotFound,
+            LinkContent::Invalid => SimplifiedLinkContent::Invalid,
+        };
+        Ok(Vc::cell(RcStr::from(deterministic_hash(
+            &salt.await?,
+            simplified,
+            algorithm,
+        ))))
+    }
+}
+
 /// The target of a symbolic link to create, used by [`WriteLinkContent`].
 ///
 /// Unlike [`LinkTarget`] this carries only the raw path: the write side never needs the target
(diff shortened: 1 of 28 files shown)


turbopack/crates/turbopack-core/src/file_source.rs
```diff
@@ -66,11 +69,31 @@ impl Asset for FileSource {
         let file_type = &*self.path.get_type().await?;
         match file_type {
             FileSystemEntryType::Symlink => match &*self.path.read_link().await? {
-                LinkContent::Link { target, link_type } => Ok(AssetContent::Redirect {
-                    target: target.clone(),
-                    link_type: *link_type,
```
bgw, line 71: This wasn't quite what we want: You can have a junction point that points to another junction point, so we should construct a junction point in that case, even though `get_type` returns `FileSystemEntryType::Symlink`.

The ambiguity of exactly what `link_type` meant made this hard to notice before.

turbopack/crates/turbo-unix-path/src/lib.rs
```diff
@@ -36,10 +36,8 @@ pub fn unix_to_sys(path: &str) -> Cow<'_, str> {
 /// see also [normalize_path] for normalization.
 /// Returns `None` if the joined path would leave the filesystem root.
 pub fn join_path(fs_path: &str, join: &str) -> Option<String> {
-    // Paths that we join are written as source code (eg, `join_path(fs_path, "foo/bar.js")`) and
-    // it's expected that they will never contain a backslash.
     debug_assert!(
-        !join.contains('\\'),
+        !cfg!(windows) || !join.contains('\\'),
         "joined path {join} must not contain a Windows directory '\\', it must be normalized to \
          Unix '/'"
```
bgw, line 42: Backslashes are valid in unix file names on unix platforms.

No production impact here as this assert is stripped out in release builds.

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -899,7 +900,7 @@ impl FileSystem for DiskFileSystem {
         let this = self.await?;
         let inner = &this.inner;
         if inner.is_path_denied(&fs_path) {
-            return Ok(LinkContent::NotFound.cell());
+            return Ok(LinkContent::Invalid.cell());
```
lukesandberg: This should be NotFound for consistency with other cases. we pretend that denied paths don't exist

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -912,55 +913,95 @@ impl FileSystem for DiskFileSystem {
...
         } else {
-            // link-relative, the raw value read from the link, converted to a unix-style format
+            // A Windows target can be partially qualified: `\foo` has a root but no drive and
+            // `C:foo` has a drive but no root, so neither is `is_absolute()`, yet neither is
+            // relative to the link's directory either. Joining one onto that directory would
+            // silently produce a plausible but wrong in-root path.
+            if target_sys_path.has_root()
+                || matches!(
+                    target_sys_path.components().next(),
+                    Some(Component::Prefix(_))
```
lukesandberg: try to handle this as part of the `is_absolute` branch? we should be able to convert and flow in there

turbopack/crates/turbo-tasks-fs/src/content.rs
```diff
@@ -163,47 +163,104 @@ pub(crate) enum FileComparison {
...
-        link_type: LinkType,
-    },
-    // Invalid means the link is invalid it points out of the filesystem root
+    /// A relative target that steps out of the root and back into it is `Invalid`, since resolving
+    /// it would need the names of the root's own ancestors, which a root-relative path doesn't
+    /// carry.
     Invalid,
-    // The target was not found
-    NotFound,
+}
```
bgw, line 261: Be clearer about this:
- Not found is if the symlink itself (not the target) wasn't found.
- Invalid is given due to any other IO error, or if the path leaves the filesystem root.

`RealPathResultError` should:
- Only use `NotFound` for the first read call in the chain.
- Use `Invalid` if `read_link` returns `Invalid`, at any later point in the chain after reading the first symlink.

Fix this messaging in `as_error_message` to be clearer about what's happening.

Add an `RcStr` inside of `Invalid` so that we can give better error messages.

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -1789,14 +1805,201 @@ mod tests {
...
                 }
             );
 
             Ok(())
         }
 
+        /// `read_link` never looks at the target, so a dangling link still reads back as a valid
+        /// [`LinkContent::Link`]. Resolving it is what discovers the target is missing.
+        #[turbo_tasks::function(operation, root)]
+        async fn assert_dangling_symlink_operation(
```
lukesandberg: inline the turbo task into the test body

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -1789,14 +1805,201 @@ mod tests {
...
+
+            tt.stop_and_wait().await;
+        }
+
+        /// A relative target must stay inside the filesystem root at every step, not just at the
+        /// end. Both of these step above the root; one comes back into it and one doesn't, but
+        /// neither can be resolved against a root-relative [`FileSystemPath`], so `read_link`
+        /// rejects both and every [`LinkContent::Link`] stays resolvable by construction.
+        #[turbo_tasks::function(operation, root)]
+        async fn assert_read_escaping_relative_symlink_operation(
```
lukesandberg: ditto on the other ones

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -620,41 +626,45 @@ async fn realpath_with_links(path: FileSystemPath) -> Result<Vc<RealPathResult>>
...
-        ) {
+        let entry_type = *current_path.get_type().await?;
+        if !matches!(entry_type, FileSystemEntryType::Symlink) {
+            // A link we followed points at something that doesn't exist. `read_link` can't detect
+            // this, as it never looks at the target, so a dangling link reads back as valid.
+            //
+            // Only report this for a link we actually followed: resolving a path that simply
+            // doesn't exist is not an error, it just resolves to itself.
+            if followed_link && matches!(entry_type, FileSystemEntryType::NotFound) {
+                error = RealPathResultError::NotFound;
```
lukesandberg: as discussed this should probably be invalid

turbopack/crates/turbopack-core/src/resolve/pattern.rs
```diff
@@ -1795,10 +1795,13 @@ pub async fn read_matches(
...
                                     let fs_path = lookup_dir.join(key)?;
-                                    if let LinkContent::Link { link_type, .. } =
+                                    if let LinkContent::Link { target } =
                                         &*fs_path.read_link().await?
                                     {
-                                        if link_type.contains(LinkType::DIRECTORY) {
+                                        if matches!(
+                                            target.target_type().await?,
+                                            FileSystemEntryType::Directory
+                                        ) {
```
bgw, line 1804: TODO (follow-up PR): this should use realpath to fully resolve the target type

turbopack/crates/turbopack-core/src/file_source.rs
```diff
@@ -66,11 +69,31 @@ impl Asset for FileSource {
...
+                    let write_target = match target {
+                        LinkTarget::Absolute { resolved } => {
+                            WriteLinkTarget::Absolute(resolved.path.clone())
+                        }
+                        LinkTarget::Relative { raw, .. } => WriteLinkTarget::Relative(raw.clone()),
+                    };
+                    let target_fs_path = target.file_system_path();
+                    let write_target_type = match *target_fs_path.get_type().await? {
+                        FileSystemEntryType::Directory => {
+                            WriteLinkTargetType::DirectoryOrJunctionPoint
```
lukesandberg, line 82: i strongly suspect that these Directory cases are errors adn we should bail

modeling a directory as a source doesn't make sense, and we would reject it if it was the direct path

lukesandberg (approved):

turbopack/crates/turbopack-core/src/resolve/pattern.rs
bgw: Looks like @marcoshernanz already started a PR: https://github.com/vercel/next.js/pull/97825

PR #97403, opened 2026-08-15
Enable parallel metadata resolution by default

gnoff:
## Motivation

Parallel routes need deterministic, fork-aware metadata resolution, and the selector API in the preceding PR should work without requiring an experimental opt-in. This PR makes the new resolver the default while retaining a narrow rollback path for applications that encounter regressions.

## What this does

Parallel metadata resolution is now the default. Applications that temporarily need the previous behavior can opt into it with `deprecated.legacyMetadataResolution: true`.

The deprecated option intentionally accepts only `true`: omission selects the current resolver, so accepting `false` would add a redundant state. Configuration normalization derives the existing internal `parallelRouteMetadata` flag from this option without exposing that implementation detail in generated public types or `defaultConfig`.

The affected metadata, viewport, navigation-error, and blocking-metadata tests now exercise the fork-aware resolver by default and the legacy resolver through the deprecated option.

## Review notes

- The fork-aware resolver and `unstable_selectMetadata` / `unstable_selectViewport` APIs live in the preceding stack layers; this PR changes their rollout and configuration surface.
- The legacy escape hatch is deprecated from introduction and can be removed after the new resolver has shipped safely.
- Streamed metadata or viewport failures can occur after the HTTP response commits, so an initial response may remain `200` while the browser renders an error boundary or follows a redirect.

## Verification

- `pnpm build-all`
- `pnpm types:test-lib`
- `HEADLESS=true pnpm test-start-turbo test/e2e/config-schema-check/index.test.ts`
- `HEADLESS=true pnpm test-dev-turbo test/e2e/app-dir/metadata-streaming-parallel-routes/metadata-streaming-parallel-routes.test.ts`


packages/next/src/server/config-shared.ts
```diff
@@ -2314,6 +2333,7 @@ export const defaultConfig = Object.freeze({
     static: process.env.NEXT_STATIC_CACHE_HANDLER_PATH,
   },
   adapterPath: process.env.NEXT_ADAPTER_PATH || undefined,
+  deprecated: {} as DeprecatedConfig,
```
unstubbable, line 2336:
```suggestion
  deprecated: {} satisfies DeprecatedConfig,
```

unstubbable (approved):

PR #97405, opened 2026-08-15
Improve getStaticPaths error message for invalid catch-all array parameters

Rezakarimzadeh98:
## Fixes
Closes #41281

## Summary
Adds explicit type validation for catch-all dynamic route parameters returned from `getStaticPaths`. When a catch-all parameter array item is not a string (e.g., `{ slug: [123] }`), the build now fails with a clear error message instead of causing cryptic failures during SSR rendering.

## Changes
- **packages/next/src/build/static-paths/pages.ts**: Added validation check that for catch-all array parameters, each item must be a string. Throws descriptive error if validation fails.
- **test/integration/prerender-invalid-catchall-params-array-item/**: Added integration test fixture and test case to verify the error message is clear and helpful.

## Testing
✅ Integration test added and passing:
- Test: `test/integration/prerender-invalid-catchall-params-array-item/test/index.test.ts`
- Fixture: Page with catch-all route returning non-string parameter
- Assertion: Build fails with expected error message: "A required parameter (slug) was not provided as an array of strings received number in getStaticPaths for /[...slug]"

## Related
- Issue: #41281
- User provided clear reproduction case and identified exact code location

### DIFF
--- a/packages/next/src/build/static-paths/pages.ts
+++ b/packages/next/src/build/static-paths/pages.ts
@@ -174,6 +174,15 @@ export async function buildPagesStaticPaths({
           )
         }
 
+        if (repeat && Array.isArray(paramValue)) {
+          const invalidItem = paramValue.find((item) => typeof item !== 'string')
+          if (typeof invalidItem !== 'undefined') {
+            throw new Error(
+              `A required parameter (${validParamKey}) was not provided as an array of strings received ${typeof invalidItem} in getStaticPaths for ${page}`
+            )
+          }
+        }
+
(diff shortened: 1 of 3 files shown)


Rezakarimzadeh98:
Thanks for the review. This change fixes #41281 by validating catch-all params returned from getStaticPaths and failing with a clear type-specific error when an item is not a string. The fix is covered by an integration test, and all checks are passing.

Rezakarimzadeh98:
Closing to free the open PR limit so the docs PR can move out of draft.

PR #97412, opened 2026-08-15
fix(codemod): write pnpm overrides to the workspace root

### LINKED ISSUE #98941: `@next/codemod upgrade` creates a nested pnpm-workspace.yaml in pnpm v11 monorepos and breaks install

https://github.com/unrevised6419/next-upgrade-pnpm-monorepo

### To Reproduce

1. Clone the repository (pnpm v11 workspace: `apps/web` depends on `packages/lib` via `workspace:*`, and has `@types/react` / `@types/react-dom` installed)
2. `pnpm install`
3. `cd apps/web`
4. `pnpm dlx @next/codemod@canary upgrade latest` (same result with `pnpm exec next upgrade`)

**Current:** the codemod writes the `@types/react` / `@types/react-dom` overrides to a new `apps/web/pnpm-workspace.yaml`:

```yaml
overrides:
  '@types/react': 19.3.0
  '@types/react-dom': 19.3.0
```

A `pnpm-workspace.yaml` marks a workspace root, so `apps/web` becomes a nested workspace cut off from its siblings, and the codemod's own `pnpm install` fails:

```
[ERR_PNPM_WORKSPACE_PKG_NOT_FOUND] In : "@repro/lib@workspace:*" is in the dependencies but no package named "@repro/lib" is present in the workspace

Packages found in the workspace: @repro/web
Error: Failed to install dependencies
```

`catalog:` specifiers fail the same way, and plain versions of workspace packages (e.g. `0.0.0`) get looked up on the npm registry (`ERR_PNPM_FETCH_404`). The project is left with the stray `pnpm-workspace.yaml` and a partially bumped `package.json`.

**Expected:** the overrides are written to the existing workspace-root `pnpm-workspace.yaml` (found by walking up from `cwd`, the way pnpm locates it), ideally scoped to the upgraded package (`'@repro/web>@types/react'`), and the install succeeds.

Without `@types/react` in the app no overrides are written and the upgrade completes, so this only affects TypeScript apps.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 27.0.0
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 22.23.2
  npm: 10.9.8
  Yarn: N/A
  pnpm: 11.21.0
Relevant Packages:
  next: 16.2.9 (upgrading to 16.3.5)
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
  @next/codemod: 16.4.0-canary.35
```

Not sure

Other (Deployed)

### Additional context

Introduced in #94690, which moved pnpm v11 overrides into `pnpm-workspace.yaml` but writes it relative to `cwd`. Its description notes monorepos were left unchanged; before, a nested `pnpm.overrides` was silently ignored, but a nested `pnpm-workspace.yaml` now breaks installation.

Fix proposed in #97412.

unrevised6419:
### What?

`next upgrade` created a `pnpm-workspace.yaml` in the directory it was run
from. In a pnpm monorepo that is the app package, so the file now lands in the
workspace root instead, and the `@types/react` / `@types/react-dom` overrides
it writes are scoped to the package being upgraded.

### Why?

The presence of `pnpm-workspace.yaml` is what defines a pnpm workspace root.
Writing one next to an app's `package.json` declared a *nested* root: the app
was cut off from the real one and stopped resolving its `workspace:*`
dependencies on sibling packages.

Overrides in a shared root apply to the whole workspace, so hoisting them
without scoping would force the upgraded app's `@types/react` version onto
unrelated packages.

Introduced in #94690.

### How?

- Walk up from the current directory to the nearest `pnpm-workspace.yaml`,
  the way pnpm locates the workspace root, and write there. With no manifest
  anywhere above, the project is standalone and the file is still created in
  place — unchanged for single-package projects.
- Write scoped keys using pnpm's `<package>><dependency>` selector, e.g.
  `'@fixture/web>@types/react': 19.2.18`, so siblings keep their own
  resolution. Falls back to an unqualified key with a warning if the app's
  `package.json` has no `name`.
- Skip the rewrite when every override already holds the value the upgrade
  would set. `js-yaml` cannot round-trip comments, and re-running the upgrade
  should not repeatedly strip them from a file the tool does not own.
- Warn when writing into a shared root, naming the file, the scope, and the
  lost comments.

Verified on a real 13-package pnpm workspace running pnpm v11.21.0, upgrading
its Next.js app from 16.2.9 to 16.3.1: no `pnpm-workspace.yaml` was created
beside the app, the root manifest gained the scoped overrides, `pnpm install`
completed across all 13 projects, and the app's `node_modules` kept its
workspace symlinks.

```
⚠ pnpm only reads overrides from the workspace root, so "@repo/web>@types/react",
  "@repo/web>@types/react-dom" were written to "<root>/pnpm-workspace.yaml".
  They are scoped to "@repo/web" so the rest of the workspace is unaffected.
  Comments in that file were not preserved.
```

Also verified on a throwaway two-package workspace, where a second run with
unchanged versions wrote nothing and printed no warning, and a run after
clearing the `overrides:` block re-emitted it.

Known limitation, not changed here: an actual write still drops comments from
the manifest. Preserving them means swapping `js-yaml` for `yaml`.

### Fixing a bug

- [x] Related issues linked using `fixes #number`
- [x] Tests added — `bin/__testfixtures__/pnpm-v11-workspace-overrides/`,
      following the existing manual fixture convention in that directory:
      `pnpm test:upgrade-fixture bin/__testfixtures__/pnpm-v11-workspace-overrides/apps/web latest`
- [ ] Errors have a helpful link attached — n/a, no new error

Fixes #98941

### DIFF
--- a/packages/next-codemod/bin/__testfixtures__/pnpm-v11-workspace-overrides/README.md
+++ b/packages/next-codemod/bin/__testfixtures__/pnpm-v11-workspace-overrides/README.md
@@ -1,0 +1,61 @@
+Verifies that on pnpm v11+, upgrading an app that lives inside a pnpm
+workspace writes `@types/react` / `@types/react-dom` overrides to the
+workspace root's `pnpm-workspace.yaml` instead of creating a new one next to
+the app's `package.json`.
+
+A `pnpm-workspace.yaml` is what makes a directory a pnpm workspace root, so
+creating one in `apps/web` would declare a nested workspace root: `apps/web`
+would be cut off from the real root and stop resolving `@fixture/ui`
+(`workspace:*`).
+
+The overrides are scoped to the app with pnpm's `<package>><dependency>`
+selector so sibling packages such as `@fixture/ui` keep their own
+`@types/react` resolution.
+See https://pnpm.io/settings/dependency-resolution#overrides.
+
+Run this fixture with pnpm v11+ on PATH, from the app directory:
+
+```sh
+pnpm test:upgrade-fixture bin/__testfixtures__/pnpm-v11-workspace-overrides/apps/web latest
+```
+
+Expected: no `apps/web/pnpm-workspace.yaml` is created, and the root manifest
+gains scoped overrides.
+
+```diff
+diff --git a/packages/next-codemod/bin/__testfixtures__/pnpm-v11-workspace-overrides/apps/web/package.json b/packages/next-codemod/bin/__testfixtures__/pnpm-v11-workspace-overrides/apps/web/package.json
+--- a/packages/next-codemod/bin/__testfixtures__/pnpm-v11-workspace-overrides/apps/web/package.json
++++ b/packages/next-codemod/bin/__testfixtures__/pnpm-v11-workspace-overrides/apps/web/package.json
+@@ -6,9 +6,9 @@
+   "dependencies": {
+     "@fixture/ui": "workspace:*",
+-    "next": "14.3.0-canary.44",
+-    "react": "18.2.0",
+-    "react-dom": "18.2.0",
+-    "@types/react": "^18.2.0",
+-    "@types/react-dom": "^18.2.0"
++    "next": "15.0.4-canary.43",
++    "react": "19.0.0",
++    "react-dom": "19.0.0",
++    "@types/react": "19.0.0",
++    "@types/react-dom": "19.0.0"
+   }
+ }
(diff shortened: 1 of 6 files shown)


unrevised6419:
Question: Should we switch from using `js-yaml` to `yaml` which from what it says it supports comments?

PR #97421, opened 2026-08-15
test: deflake use-cache-size-zero warm reload

gnoff:
## Summary

Keep the first warm reload assertion focused on browser-visible stale-while-revalidate behavior, then poll the route with independent HTTP requests until a later response observes the fresh value. After convergence, perform one final browser reload to confirm the user-visible path also receives an updated value.

This avoids fixed delays, cache debug logging, and repeated browser navigations that can cancel or outpace background regeneration. The convergence poll uses the standard 3-second `retry()` window and returns as soon as freshness is observed; the browser only reloads again after that condition is satisfied.

## Verification

- `__NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true NEXT_TEST_CI=true HEADLESS=true pnpm test-dev-webpack test/development/app-dir/use-cache-size-zero/use-cache-size-zero.test.ts`
- `__NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true NEXT_TEST_CI=true HEADLESS=true pnpm test-dev-turbo test/development/app-dir/use-cache-size-zero/use-cache-size-zero.test.ts`

### DIFF
--- a/test/development/app-dir/use-cache-size-zero/use-cache-size-zero.test.ts
+++ b/test/development/app-dir/use-cache-size-zero/use-cache-size-zero.test.ts
@@ -24,19 +24,12 @@ describe('use-cache-size-zero', () => {
     expect(await browser.elementByCss('p', { waitUntil: false }).text()).toBe(
       'Loading...'
     )
-    await retry(async () => {
-      expect(
-        await browser.elementByCss('p', { waitUntil: false }).text()
-      ).toBeDateString()
-    })
+    // After observing the streamed fallback, wait for the initial document to
+    // finish loading so the cold request is fully settled before reloading.
     const coldValue = await browser
-      .elementByCss('p', { waitUntil: false })
+      .elementByCss('#value', { waitUntil: 'load' })
       .text()
-
-    // Wait for the cache write to settle before reloading. The generated value
-    // can finish streaming before the backing cache has persisted it, which
-    // would race the reload below and turn the expected warm hit into a miss.
-    await waitFor(2000)
+    expect(coldValue).toBeDateString()
 
     // Warm reload: `cacheMaxMemorySize: 0` still caches in development, so the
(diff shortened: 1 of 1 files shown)


acdlite (approved):

test/development/app-dir/use-cache-size-zero/use-cache-size-zero.test.ts
```diff
@@ -40,23 +38,22 @@ describe('use-cache-size-zero', () => {
...
+      await browser.elementByCss('#value', { waitUntil: false }).text()
+    ).toBe(coldValue)
 
-    // That warm reload regenerated a fresh entry in the background, so a later
-    // reload converges to the new value. Read after "load" here (a plain
-    // refresh) since we want the settled value, not the streaming inspection
-    // above.
+    // That warm reload regenerates a fresh entry in the background. Poll with
+    // independent requests so we can observe convergence without navigating
+    // away from (and potentially cancelling) the warm browser request.
```
unstubbable: Wait what? We're not cancelling cache fills, are we?

test/development/app-dir/use-cache-size-zero/use-cache-size-zero.test.ts
```diff
@@ -40,23 +38,22 @@ describe('use-cache-size-zero', () => {
...
     await retry(async () => {
-      await browser.refresh()
-      expect(await browser.elementById('value').text()).not.toBe(coldValue)
+      const $ = await next.render$('/reload')
+      expect($('#value').text()).not.toBe(coldValue)
     })
+
+    // Once independent requests observe convergence, confirm the next browser
+    // reload also exposes a fresh value through the user-visible path.
+    await browser.refresh()
```
unstubbable: How is the fetch before affecting this refresh?

PR #97440, opened 2026-08-16
Keep metadata rendering stable across streaming modes

gnoff:
## Motivation

Streaming and blocking metadata previously used different rendering paths. Metadata placement should depend on when Next.js starts consuming the response, rather than requiring the metadata component itself to render outside Suspense.

## What this does

Selected metadata now always renders through the same hidden Suspense boundary. For blocking responses, an always-present `MetadataBlocker` waits for the same cached resolution before Next.js pulls the response, ensuring metadata remains in the document head.

`MetadataOutlet` remains a single promise. It renders inside Suspense for streaming responses and outside Suspense for blocking responses so navigation and regular errors retain their existing behavior.

HTML-limited bot requests continue to bypass PPR shells. A prerendered shell has already closed its head before dynamic metadata resolves, so resuming that shell cannot provide blocking metadata in the raw head.

## Review notes

- Metadata resolution still starts from a rendered component so the active `workUnitStore` and React cache scope are available.
- Fizz may begin rendering before metadata settles; blocking is enforced before Next.js starts pulling the response.
- The hidden wrapper keeps top-level suspenseful metadata out of the document preamble. Metadata tags hoist out, so the wrapper remains empty.
- The Cache Components blocking and hydration coverage runs with parallel metadata both disabled and enabled.

## Verification

- `pnpm build-all`
- `pnpm test-start-turbo test/e2e/app-dir/metadata-streaming-cache-components/metadata-streaming-cache-components-custom-bots.test.ts`
- `pnpm test-start-webpack test/e2e/app-dir/metadata-streaming-cache-components/metadata-streaming-cache-components-custom-bots.test.ts`


test/e2e/app-dir/metadata-streaming-cache-components/metadata-streaming-cache-components-custom-bots.test.ts
```diff
@@ -56,6 +57,20 @@ describeCacheComponents('metadata streaming with a custom bot list', () => {
     expect($('#dynamic-content').text()).toBe('dynamic content')
   })
 
+  it('should hydrate blocking metadata without a tree mismatch', async () => {
```
unstubbable: Could we run this suite with `experimental.parallelRouteMetadata` both enabled and disabled? Both implementations change here, but the fixture leaves the flag disabled, so the new hydration test only covers the legacy implementation.

The `MyBot` request also bypasses the PPR shell and performs a full dynamic render. That makes this useful blocking-metadata coverage, but it does not exercise the streaming-prerender-to-blocking-resume transition described in the PR. Could we add an explicit regression test for that transition, ideally one that fails before this change?

gnoff: :robot: Updated the Cache Components custom-bot suite to run with parallelRouteMetadata both disabled and enabled.

I also checked the proposed prerender-to-blocking-resume case. Blocking metadata requests intentionally bypass PPR shells: the shell has already closed the head, and allowing the request to resume it caused metadata to be emitted after the head for both implementations. I kept that bypass, clarified the constraint in app-page-runtime.ts and the PR description, renamed the hydration test so it no longer claims to cover an unreachable transition, and simplified MetadataOutlet back to one conditional outlet.

Verified the 15-case suite with both Turbopack and webpack.

unstubbable (approved):

PR #97462, opened 2026-08-17
docs: fix crossOrigin wording

Rezakarimzadeh98:
## What?

Fix the wording in the crossOrigin documentation so the description is clearer and more accurate.

## Why?

The previous phrasing was slightly awkward and could be confusing for readers. This is a documentation-only improvement with no runtime behavior change.

## How?

- adjust the wording around generated <script> tags
- keep the language consistent with the shared app/pages router docs
- scope the change to the relevant documentation page


marcoshernanz (approved):

Rezakarimzadeh98:
Friendly follow-up on this docs-only PR. If there is any wording preference to align with before merge, I can update immediately.

Rezakarimzadeh98:
Still happy to adjust wording if maintainers want a different phrasing — docs-only and CI green.

PR #97463, opened 2026-08-17
[16.3] Turbopack: don't trace embedded WASM loader helpers (#97353)

mischnic:
Backport of https://github.com/vercel/next.js/pull/97353


lukesandberg (approved):

obscura-oss (commented):
## PR Review: #97463 — "[backport] Turbopack: don't trace embedded WASM loader helpers"

**Author:** Niklas Mischkulnig (@mischnic)  
**Type:** Backport of #97353 (already merged)  
**Files changed:** 9 files (8 added, 1 modified)  
**Scope:** Turbopack + test infrastructure

---

### Summary

This PR backports a fix from #97353 that prevents Turbopack's output tracing from failing when a WebAssembly module is referenced through a bundler-friendly import path that includes a generated JS loader referencing an embedded runtime helper (`[turbopack-wasm]/**/loadWasm.ts`). The embedded helper is compiled into the output chunk and has no corresponding disk path, so tracing it would produce a file reference that cannot be resolved.

### Production Code Change (1 file)

```diff
// turbopack/crates/turbopack/src/module_options/mod.rs
+                if is_tracing {
+                    vec![ModuleRuleEffect::ModuleType(ModuleType::Raw)]
+                } else {
+                    vec![ModuleRuleEffect::ModuleType(ModuleType::WebAssembly {
+                        source_ty: WebAssemblySourceType::Binary,
+                    })]
+                },
```

**What it does:** When Turbopack runs in tracing mode (`is_tracing` is true), `.wasm` files are treated as `Raw` modules instead of `WebAssembly` modules. The same treatment is applied to `.wat` files. This prevents the tracing machinery from following the WebAssembly module processing path, which would try to resolve the internal `loadWasm.ts` runtime helper as a file on disk — something that cannot succeed.

**Assessment:** The change is minimal, targeted, and correct. The `is_tracing` flag is an existing boolean in the `ModuleOptions` context, so this is leveraging an established pattern. The rationale is well explained in the accompanying comment in the diff.

### Test Changes (8 files)

A comprehensive new test fixture `turbopack-wasm-output-tracing` is added:

- **`wasm-dep/`** — A simulated npm package (local file dependency) that:
  - Exposes a bundler-friendly `loadWasmModule()` that does `require('./add.wasm')` (static import, never actually called)
  - Exposes a runtime fallback `addOne()` that reads the WASM binary with `fs.readFile` and instantiates it directly
  - Ships a real `add.wasm` binary
- **`app/api/wasm/route.js`** — Route handler that calls the runtime fallback (`addOne`)
- **`turbopack-wasm-output-tracing.test.ts`** — Two test cases:
  1. **Basic smoke test:** Verifies the route builds and serves correctly (returns `{ result: 2 }`)
  2. **Trace verification:** Checks that the output `.nft.json` (node-file-trace) does NOT contain the embedded `loadWasm` helper path, and (under Turbopack) DOES list the actual `.wasm` file

**Assessment:** The test is well-structured and validates exactly the failure mode the code fix addresses. Using a `file:./wasm-dep` dependency is a clean way to simulate a real external package. The test correctly scopes the trace assertion to `isNextStart && !isNextDeploy` since traces only exist locally. The conditional `isTurbopack` check for the WASM file presence is appropriate (the test suite must pass for both webpack and Turbopack).

### Positive Observations

1. **No behavioral change outside tracing mode** — the `if` guard ensures non-tracing builds are unaffected.
2. **Symmetry between .wasm and .wat** — both WebAssembly source types get the same tracing treatment.
3. **Self-contained test** — the fixture includes its own `next.config.js` with `serverExternalPackages: ['wasm-dep']`, which is necessary for the external package to be `require`d on the server side.
4. **Good inline documentation** — the code comment in `mod.rs` clearly explains *why* the change is needed.

### Suggestions / Questions

**Nitpick (minor):** In the test, the `turbopack-wasm` filter could match against a very specific pattern to reduce false positives:

```typescript
// Current:
(file) => file.includes('turbopack-wasm') || file.includes('loadWasm')
// Consider (optional):
(file) => /turbopack-wasm.*loadWasm/.test(file) || file.includes('loadWasm')
```

This is entirely optional — the test as-is is already correct for the failure mode it targets.

### Verdict

**LGTM.** This is a clean, focused backport. The production change is a single `if/else` branch in one file, and the test fixture convincingly reproduces the scenario that previously caused build failures. The original PR (#97353) was already merged, so this change has received prior review and CI validation.

PR #97468, opened 2026-08-17
docs: mention Valibot as validation library option in forms guides

fabian-hiller:
The forms guides recommend validating with a library like Zod. This broadens the wording to also mention Valibot, which covers the same use case with a smaller bundle footprint. The authentication guide already lists Zod or Yup, so Valibot is added to that list as well. The existing Zod examples are unchanged. I'm the author of Valibot.

- Ran `prettier --check` on the touched files with the repo's pinned version.

### DIFF
--- a/docs/01-app/02-guides/forms.mdx
+++ b/docs/01-app/02-guides/forms.mdx
@@ -131,7 +131,7 @@ export async function updateUser(userId, formData) {}
 Forms can be validated on the client or server.
 
 - For **client-side validation**, you can use the HTML attributes like `required` and `type="email"` for basic validation.
-- For **server-side validation**, you can use a library like [zod](https://zod.dev/) to validate the form fields. For example:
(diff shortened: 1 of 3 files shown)


docs/01-app/02-guides/forms.mdx
```diff
@@ -131,7 +131,7 @@ export async function updateUser(userId, formData) {}
 Forms can be validated on the client or server.
 
 - For **client-side validation**, you can use the HTML attributes like `required` and `type="email"` for basic validation.
-- For **server-side validation**, you can use a library like [zod](https://zod.dev/) to validate the form fields. For example:
+- For **server-side validation**, you can use a schema validation library like [Zod](https://zod.dev/) or [Valibot](https://valibot.dev/) to validate the form fields. For example:
```
icyJoseph, line 134: wonder why Yup was left out here initally

icyJoseph (approved):

PR #97469, opened 2026-08-17
[PPF] Fix incorrect eager hint in incremental PPF

lubieowoce:
Fixes a bug that happens incremental in incremental PPF (via `export const prefetch = 'partial'`) where all segments without a config would incorrectly be marked as requiring a speculative prefetch if `nextConfig.partialPrefetching` was unset.
In practice, this means that `<Link href="/foo">` would act like `<Link href="/foo" prefetch={true}>`, so **it'd always send a per-link runtime prefetch instead of only fetching a shell**.

Also updates many tests that were incorrectly asserting that we should be seeing 2 requests (shell + speculative) instead of just 1, for the shell.

## TLDR

`computeSegmentPrefetchHints` was treating all segments without a `prefetch` config as if they were segments where PPF is disabled and setting `SubtreeHasEagerPrefetch` which is the pre-PPF behavior. this made the router think it should be doing a speculative prefetch for every link despite not having `prefetch={true}` or `unstable_eager` anywhere. Fixed by determining PPF-ness once per route and using that instead of the per-segment config.

## The problem, in detail

<details>
<summary>(Collapsed because it's like a whole screen of text)</summary>

Consider a `/posts/[id]` page:

```
app/
  layout.tsx
  posts/[id]/
    page.tsx
```
we have 4 segments here:
```
/ # layout.tsx
/posts
/posts/[id]
/posts/[id]/__PAGE__ # page.tsx
```

Assuming we don't have a global `partialPrefetching` value set:
- if none of the segments have `export const prefetch = 'partial' | 'unstable_eager'`, then the tree should be marked with a `SubtreeHasEagerPrefetch` hint (and NO `SubtreeHasPartialPrefetching`). This is because the legacy prefetch strategies are conceptual eager (i.e. does not use shells).
- if `page.tsx` has `export const prefetch = 'partial'`, any link to that page should use partial prefetching and ONLY fetch the shell, i.e. we only set `SubtreeHasPartialPrefetching` hint.
- if `page.tsx` has `export const prefetch = 'unstable_eager'`, we want the links to fetch the shell AND do speculative prefetch, i.e. we set `SubtreeHasPartialPrefetching` and `SubtreeHasEagerPrefetch`.

however, `computeSegmentPrefetchHints` was handling `prefetch` incorrectly, and if any segment had an `undefined` prefetch config (or had no module at all, like `/posts` in the example), it'd fall into this codepath
https://github.com/vercel/next.js/blob/c713d48763dd9c9d34fe2b99829cf3f1ada78dc5/packages/next/src/server/app-render/create-transport-tree-from-loader-tree.ts#L146-L152
and set the `SubtreeHasEagerPrefetch` hint (which then gets propagated upwards).

essentially, this means that if there was even single segment *without* `export const prefetch = "partial"` (and no global `partialPrefetching` to use as a default value), we'd do a speculative prefetch for the link.
This is visible in the `prefetch-app-shell` tests, which (confusingly) assert that we should be seeing two requests (shell + prefetch) for a link that really should only be fetching a shell.
https://github.com/vercel/next.js/blob/1c50e09d1dc3f5be50c5e2f9b99816ad47e11f05/test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts#L36-L43

</details>

## The solution

the mistake in `computeSegmentPrefetchHints` is that it doesn't consider that partialPrefetching being enabled or not is the property of a whole route tree, and should not be resolved per segment. instead, we should determine it for the whole tree and only do the "not PPF + speculative" combination if PPF is indeed not enabled *anywhere* in the tree.

the new flow implemented here is as follows:
1. first, traverse the loader tree looking for `prefetch` exports. if we only see `prefetch = partial`, use that. if we see prefetch = `unstable_eager` anywhere, use that instead. if we see both, `unstable_eager` wins.
2. if we found no per-segment configs, we fall back to the global `partialPrefetching` config.

notably, this means that:
- `export const prefetch = 'unstable_eager'` overrides `partialPrefetching: true`
- `export const prefetch = 'partial'` overrides `partialPrefetching: 'unstable_eager'`
it makes sense for the segment level config to win, because it's intended for incrementally migrating an app.

after we compute the prefetch strategy for the route, we plumb it through to `computeSegmentPrefetchHints`, and mark the segments appropriately. it still looks at segment-level `prefetch` configs, but only to find `force-disabled` which applies per-segment.

### DIFF
--- a/packages/next/src/server/app-render/create-transport-tree-from-loader-tree.ts
+++ b/packages/next/src/server/app-render/create-transport-tree-from-loader-tree.ts
@@ -15,6 +15,7 @@ import {
 import type { GetDynamicParamFromSegment } from './app-render'
 import { addSearchParamsIfPageSegment } from '../../shared/lib/segment'
 import type { AppSegmentConfig } from '../../build/segment-config/app/app-segment-config'
+import type { ResolvedPartialPrefetching } from './prefetch-config'
 
 export type MissingPrefetchHintPolicy =
   | 'mark-stale'
@@ -61,7 +62,7 @@ export async function computeSegmentPrefetchHints(
   hintTree: PrefetchHints | null,
   prefetchInliningEnabled: boolean,
   missingPrefetchHintPolicy: MissingPrefetchHintPolicy,
-  partialPrefetching: boolean | 'unstable_eager' | undefined,
+  routePartialPrefetching: ResolvedPartialPrefetching,
   // Whether this segment is at or above the root layout (no layout was found
   // above it).
   isRootLayoutOrAbove: boolean
@@ -74,13 +75,15 @@ export async function computeSegmentPrefetchHints(
   // `partialPrefetching` config in next.config.js.
   const mod = layout ? await layout[0]() : page ? await page[0]() : undefined
   const instantConfig = mod ? (mod as AppSegmentConfig).instant : undefined
-  const prefetchConfig =
-    (mod ? (mod as AppSegmentConfig).prefetch : undefined) ??
-    (partialPrefetching === 'unstable_eager'
-      ? 'unstable_eager'
-      : partialPrefetching
-        ? 'partial'
-        : undefined)
+
+  /**
+   * NOTE: Do not use for determining whether Partial Prefetching is on.
+   * we determine that via `routePartialPrefetching`.
+   * */
+  const segmentPrefetchConfig = mod
+    ? (mod as AppSegmentConfig).prefetch
+    : undefined
+
   let prefetchHints = 0
 
   // Union in the precomputed build-time hints (e.g. segment inlining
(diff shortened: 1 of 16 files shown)


packages/next/src/server/app-render/create-transport-tree-from-loader-tree.ts
```diff
@@ -130,25 +133,36 @@ export async function computeSegmentPrefetchHints(
...
-  } else if (prefetchConfig === 'unstable_eager') {
-    // Like 'partial' (uses the PPR fetch strategy) but also marks the segment
-    // as eager, so App Shells keeps prefetching it instead of relying on the
-    // shared app shell.
-    prefetchHints |=
-      PrefetchHint.SubtreeHasPartialPrefetching |
-      PrefetchHint.SubtreeHasEagerPrefetch
-  } else if (prefetchConfig === 'force-disabled') {
-    prefetchHints |= PrefetchHint.PrefetchDisabled
+  switch (routePartialPrefetching) {
```
lubieowoce, line 136: this is the important change, the rest is mostly just plumbing to facilitate it

packages/next/src/server/app-render/prefetch-config.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  return level
+}
+
+enum PartialPrefetchingLevel {
+  Unspecified = 1,
+  Partial = 2,
+  Eager = 3,
+}
```
lubieowoce, line 97: we could in theory just use `undefined | 'partial' | 'unstable_eager'` or something, but i like the explicitness of using an enum. it's also efficient to compare in `moreSpecificLevel`, it's just a `Math.max`.

packages/next/src/server/app-render/instant-validation/instant-config.tsx
```diff
@@ -50,38 +50,6 @@ export function isFrameworkErrorRoute(route: string | undefined): boolean {
   )
 }
 
-/**
- * Matches any `prefetch` config that enables Partial Prefetching for the
- * segment: 'partial' or 'unstable_eager'. A route with Partial Prefetching
- * enabled also runtime-caches its navigations, so this gates the runtime
- * prefetch spawn.
- */
-export async function anySegmentHasPartialPrefetchingEnabled(
```
lubieowoce, line 58: drive-by refactor to make our logic for this consistent.

test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts
```diff
@@ -28,20 +28,13 @@ describe('App Shell prefetching', () => {
...
       await browser
         .elementByCss('input[data-link-accordion="/posts/1"]')
         .click()
-    }, [
-      // Two runtime responses carry the shell text, in order: the Shell
-      // phase's runtime App Shell request, then the Speculative phase's
-      // batched per-link runtime prefetch. The route reads request data,
-      // so its static-attempt hint is unset and the prefetch deopts to
-      // runtime requests — the runtime-completeness contract of Partial
-      // Prefetching routes.
```
lubieowoce: this is an example of the tests asserting on the wrong thing. it is not at all expected that a bare `<Link>` triggers a speculative prefetch

lubieowoce:
Discussed out of band, we're going to just get rid of `unstable_eager` for now, which is the root of the problem, so i'll rework this to do that instead.

PR #97475, opened 2026-08-17
perf(turbopack): skip ignored files during server tracing

jimmyhmiller:
So I was playing around with profiling some very simple next apps when I noticed this. Basically, because of how things were set up we were spending a good chunk of our build time analyzing files we were going to reject. I went around and around trying different combinations to make sure we ignore those files (and only the files we should ignore) to speed up build times. Since it is a constant impact, mostly on small projects, but on some, it can make a big difference. I had some adversarial rounds with agents to see if they can break the config in the ways the old comments hints at why we didn't do that. And this final round finally couldn't get it to break.

For a v0-generated Hello World app, compile time is 28.0% faster (+/-1.7%) and a full clean build is 15.9% faster (+/- 1.2%).


crates/next-api/src/next_server_nft.rs
```diff
@@ -380,78 +452,24 @@ impl ServerNftJsonAsset {
...
+        let server_ignores_glob = next_owned_ignores(&self.ty, has_next_support)
+            .into_iter()
+            .chain(additional_ignores.iter().map(|s| s.as_str()))
+            .chain(if is_standalone {
+                Either::Left(std::iter::empty())
+            } else {
+                Either::Right(
+                    ["**/*/next/dist/server/next.js", "**/*/next/dist/bin/next"].into_iter(),
+                )
+            })
```
mischnic: `is_standalone` is a global setting. So it can be in `next_owned_ignores` as well.
Only `additional_ignores` is indeed page specific.

jimmyhmiller: :robot: Moved the non-standalone next.js and CLI patterns into next_owned_ignores and now pass is_standalone into the shared set used for graph-time pruning. Only the project-provided route excludes remain final-graph-only.

mischnic (approved):

PR #97480, opened 2026-08-18
Store keys in key order in SST blocks that omit hashes

lukesandberg:
## What

Key blocks for short keys (≤ 32 bytes) don't store a per-entry hash, but were still sorted by `(hash, key)`. That forced every binary-search probe to recompute the entry's xxh3 hash just to compare it — roughly ten hashes per lookup. These blocks now store their entries in **key order**, so the search compares key bytes directly and hashes nothing.

All the families `turbo-tasks-backend` produces hit this case, so this is the common path in practice.

## Why this is safe

Hash-based routing is unchanged. Files are still assigned by hash, the index block still routes to a key block by hash, and `min_hash`/`max_hash`, the AMQF, and compaction's coverage model all work exactly as before.

Thus only the order *within* a hash-less block changes.  This adds a small cost to writing, which now must re-sort blocks of keys and a double cost to compaction which now must re-sort blocks in hash order to iterate them, and then sort them back to key order when writing the new file.

## Benchmarks

Measured against `canary`, baseline and comparison run back to back on the same machine.

**Lookups ** All 16 `static_sorted_file_lookup` configurations improved
| entries | hit/uncached | hit/cached | miss/uncached | miss/cached |
| --- | --- | --- | --- | --- |
| 1 Ki | -30.4% | -24.8% | -29.1% | -29.4% |
| 10 Ki | -25.2% | -19.7% | -22.7% | -23.7% |
| 100 Ki | -17.3% | -19.3% | -15.2% | -16.8% |
| 1000 Ki | -7.0% | -25.7% | -7.0% | -10.9% |

**Commits ** Short keys pay for the added per-block sort.

| config | change |
| --- | --- |
| `key_8` 85Ki entries | +3.3% |
| `key_8` 853Ki entries | +3.1% |
| `key_8` 8.33Mi entries | +8.5% |
| `key_32Ki` / `key_4` (large-key configs) | -2.6% to -19.1% |

Six of thirteen write configurations came out as noise (p >= 0.05) and are omitted.

**Compaction** `StaticSortedFileIter` must yield `(hash, key)` order because `MergeIter` merges on it, so it reorders each hash-less block back into hash order.  Plus the additional cost of the commit (above)

| config | change |
| --- | --- |
| 4Mi entries / 8 commits | +6.0% (reproduced at +10.2%) |
| 16Mi entries / 8 commits | +7.5% |
| 4Mi entries / 32 commits | +7.8% |
| 16Mi entries / 32 commits | +5.2% |
| 16Mi entries / 128 commits | -9.5% |


turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
```diff
@@ -883,13 +912,114 @@ impl<E: Entry> StreamingSstWriter<E> {
...
+            key_ranges,
+            key_order,
+            current_pos,
+            occupant,
+            pending_keys,
+            ..
+        } = self;
+        key_order.clear();
+        key_order.extend(0..key_ranges.len() as u32);
+        key_order.sort_by(|&a, &b| {
```
lukesandberg: this can be a sort_by_key i think, would be simpler written as such

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
```diff
@@ -883,13 +912,114 @@ impl<E: Entry> StreamingSstWriter<E> {
...
+            let (b_start, b_end) = key_ranges[b as usize];
+            key_order_bytes[a_start as usize..a_end as usize]
+                .cmp(&key_order_bytes[b_start as usize..b_end as usize])
+        });
+
+        // Apply the permutation to the entries by following its cycles, so each entry is moved at
+        // most once and nothing is cloned. `key_order[i]` is the *original* position of the entry
+        // that belongs at `i`, so applying it needs to track where entries have been moved to:
+        // `current_pos[o]` is where original entry `o` now lives, and `occupant[s]` is which
+        // original entry currently sits in slot `s`.
```
lukesandberg: because the keys are relatively small.... do we really need to be so defensive about the moves?

key blocks are also somewhat bounded, so perhaps storing slices, sorting the slices and then copying into a new buffer would work?

Changing the write_key_to api is also in scope i think we could actually change that to an api that returns a slice which could simplify a lot or possibly a key_sort_token that could be used?

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
lukesandberg: :robot: Done in f35f9ac664 — switched to `sort_by_key` returning the key slice, which also let the separate `key_ranges` and `key_order` scratch vectors collapse into one.

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
lukesandberg: :robot: Partly addressed in f35f9ac664, but I want to flag what didn't work.

**Storing slices, sorting, copying into a new buffer** — I tried exactly this and it's a performance trap, twice over. `pending_keys` is a `VecDeque` and the flushed range is often *not* a suffix: `advance_boundary_to` flushes a prefix while later entries are still queued behind it. So `drain(start..end)` + `push_back` puts entries at the wrong end (tests caught it), and `drain` + `insert(start + offset)` is correct but shifts the untouched tail on every insert. That's O(queue_len) per entry, not O(block_len) — the unit test suite went from 50s to >590s. The block being bounded doesn't save you, because the *queue* isn't.

So the defensiveness isn't about the moves being expensive; it's that the cheap-looking rebuild is quadratic in the wrong variable. The cycle-following swap is now the simple thing that's also correct: it touches only `[start, end)` and never shifts anything outside it. I did keep your `sort_by_key` point and dropped one of the three scratch vectors.

**Changing `write_key_to` to return a slice** — can't be done universally. `Entry` has two implementors: `LookupEntry` holds a contiguous `RcBytes` and could return a slice, but `CollectorEntry<K: StoreKey>` is generic over keys including tuples `(A, B)`, whose bytes are non-contiguous by construction (see `StoreKey for (A, B)` in key.rs). A `key_sort_token` is more promising since it only needs `Ord`, not contiguity — but it'd have to be generic over the key type to avoid allocating, which pushes a new associated type through `Entry`. Happy to do it if you want; it felt out of scope for this PR and I'd rather not expand the diff without your say-so.

Leaving this unresolved since the `write_key_to`/token question is still genuinely open.

lukesandberg: :robot: Done — you were right on both points, and it collapsed much further than I expected.

Verified your claim first: the only production `StoreKey` implementor is `WriteBuffer` (Deref to [u8], always contiguous), and deleting the tuple + `u8` impls broke exactly three sites, all test helpers using `(u8, [u8; 4])` keys. Those now build `Box<[u8]>` via a `tuple_key` helper.

So `StoreKey` gained `as_slice()` and `Entry` gained `key_bytes()`, both with `write_to`/`write_key_to` kept as provided methods (the block builders still use them). The entire reorder is now:

    let entries = &mut self.pending_keys.make_contiguous()[start..end];
    entries.sort_by(|a, b| a.entry.key_bytes().cmp(b.entry.key_bytes()));

That deleted the scratch byte buffer, the `(range, index)` tuple sort, both permutation-tracking vectors, and the cycle-following loop — four writer fields and ~90 lines. Net -60 lines.

`QueryKey` still supports tuples, so lookups keep comparing composite keys without allocating; only *storing* now requires contiguity.

Two things I got wrong while doing this, for the record. I worried `make_contiguous` would be O(queue) per flush and tried to avoid it with a hand-rolled in-place merge across the deque's wrap point — that merge was flatly incorrect (52% failure rate against a brute-force check, which is why I now validate this kind of thing in a scratch script before compiling). A stable insertion-sort version was correct but made the test suite 119s vs 51s. Plain `make_contiguous` is 51s: entries are only appended at the back and drained from the front, so it is a cheap memmove that leaves the deque contiguous for subsequent flushes. The simple version won on both counts.

Also: an intermediate run showed multi_value/vpk_4 writes at +51%, which I chased as a regression. Re-measured after the fix it is +9.0%, in line with the +12.9% from before this refactor — the spike was noise.

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -991,47 +1078,151 @@ struct GetKeyEntryResult<'l> {
...
-    if entry_hash.is_empty() {
-        // No hash stored - compare keys directly instead of recomputing hash
-        query_key.cmp(entry_key) == Ordering::Equal
-    } else {
-        // Hash stored - cheap 8-byte comparison first, then key comparison
-        full_hash.to_be_bytes()[..] == *entry_hash && query_key.cmp(entry_key) == Ordering::Equal
+    match layout {
+        KeyBlockLayout::KeyOnly => {
+            debug_assert!(entry_hash.is_empty(), "KeyOnly entries carry no hash");
+            query_key.cmp(entry_key) == Ordering::Equal
```
lukesandberg: Can we demand QueryKey: Eq?

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -861,87 +939,73 @@ impl StaticSortedFileIter {
...
-                let FixedValueLayout {
+        let block_len = block.len();
+        let Some((layout, fixed)) = KeyBlockLayout::from_block_type(block_type) else {
+            bail!("Invalid key block type: {block_type}");
+        };
+        let hash_len = layout.hash_len();
+
+        let (kind, entries) = if fixed {
+            ensure!(data.len() >= 6, "fixed key block too short");
+            let key_size = data[4] as usize;
```
lukesandberg, line 960: comment

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -767,24 +819,50 @@ pub struct StaticSortedFileIter {
...
+    /// Whether entries carry a hash, and so what order they are stored in.
+    layout: KeyBlockLayout,
     entries: RcBytes,
     /// Number of entries in this key block (max ~819 per 16 KiB block).
     entry_count: u32,
     /// Current position within the key block.
     index: u32,
+    /// Iteration plan for a [`KeyBlockLayout::KeyOnly`] block
+    /// hash to key index sequence
+    hash_order: Vec<(u64, u32)>,
```
lukesandberg: using an Option would be clearer

or like Kind we could introduce an enum mirroring KeyBlockLayout that holds this

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -767,24 +819,50 @@ pub struct StaticSortedFileIter {
...
+}
+
 struct CurrentKeyBlock {
     kind: CurrentKeyBlockKind,
+    /// Whether entries carry a hash, and so what order they are stored in.
+    layout: KeyBlockLayout,
     entries: RcBytes,
     /// Number of entries in this key block (max ~819 per 16 KiB block).
     entry_count: u32,
     /// Current position within the key block.
```
lukesandberg: clarify that it might be the index in hash_order

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
lukesandberg: named struct plz

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -991,47 +1055,67 @@ struct GetKeyEntryResult<'l> {
...
+
+/// Compares a query against an entry, returning the ordering of the query relative to the entry in
+/// the block's own sort order.
+///
+/// Each arm matches the order its layout is stored in, so the binary search in
+/// [`StaticSortedFile::lookup_block_inner`] agrees with the writer:
+///
+/// - [`KeyBlockLayout::HashThenKey`]: compare the stored 8-byte hash first, key only on a tie.
+/// - [`KeyBlockLayout::KeyOnly`]: compare the key directly. This is the point of key ordering —
+///   otherwise the entry's hash would have to be recomputed on every probe.
```
lukesandberg: drop comment: it is just describing the simple logic of the fn

wbinnssmith (approved):

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
lukesandberg: :robot: Done — `hash_order` is now `Option<Vec<HashOrderEntry>>`. `None` for `HashThenKey` (whose entries are already in hash order) instead of the empty-vec sentinel, and the read site in `next_internal` matches on the `Option` rather than re-deriving the case from `layout`.

lukesandberg: :robot: Done — added a named `HashOrderEntry { hash: u64, entry_index: u32 }` struct, so the plan is `Vec<HashOrderEntry>` instead of `Vec<(u64, u32)>`. The destructure in `next_internal` and the sort key in `hash_order_for_block` now read by field name.

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
lukesandberg: :robot: Done — the doc comment now reads: "Current iteration position. Indexes \`hash_order\` when that is present, and the block's entries directly otherwise."

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
lukesandberg: :robot: Done — dropped the per-arm breakdown, keeping just the one-line summary of what the function returns.

PR #97482, opened 2026-08-18
Fix soft navigation leaving previous route metadata in document.head

### LINKED ISSUE #97472: Soft navigation leaves the previous route's metadata mounted in head (duplicate og/twitter tags) on Vercel with 16.3

https://github.com/brookton/next-soft-nav-metadata-repro

### To Reproduce

1. Deploy the linked app to Vercel.
2. As soon as the deployment is READY, open `/` in a browser.
3. Click "Article one" (client-side navigation; the link prefetches).
4. In devtools, run:

```js
[...document.head.querySelectorAll('meta[property="og:type"]')].map((m) => m.content);
```

Buggy result: `["website", "article"]`. Every `og:*`, `twitter:*`, and `description` tag is present twice, previous route's values plus new route's values.

Timing matters: reproduction is reliable in the first minutes after a deployment, while the route's prefetch responses are served cold (`x-vercel-cache: PRERENDER`). Once those entries warm, the same flow is usually clean, so redeploy to re-enter the window. In our production app's CI, where Playwright runs minutes after each preview deployment, the failure reproduced on 100% of attempts including retries.

Current: after a soft navigation, the previous route's entire resolved metadata block stays mounted in `<head>` alongside the new route's block, so `meta[property="og:type"]` resolves to two elements (`website` and `article`). The state persists (verified 8+ seconds after navigation) and never self-corrects.

Expected: the new route's metadata replaces the previous route's, matching what a hard load of the same URL produces (a single correct set).

```bash
Operating System:
  Platform: darwin
  Arch: arm64
Binaries:
  Node: 24.11.1
  npm: 11.6.2
Relevant Packages:
  next: 16.3.1
  react: 19.2.7
  react-dom: 19.2.7
  typescript: 5.9.3
```

Affected: 16.3.0 and 16.3.1. Not affected: 16.2.12 (same app, same fresh-deployment CI window, all browsers pass).

Metadata, Navigation

Vercel (Deployed)

### Additional context

Evidence gathered on a production app (five-browser Playwright CI plus scripted Chromium against the preview deployment):

- Server HTML is always correct. Curling the deployed pages and a local `next build && next start` render shows exactly one metadata set, and hard reloads in the browser are always clean. The defect is purely client-side head reconciliation during soft navigation.
- Blocking prefetch requests in the browser (aborting requests carrying `Next-Router-Prefetch: 1`) makes the identical navigation clean, implicating the prefetch payload path. Artificially delaying prefetch responses does not reproduce it, so it is not a simple in-flight race.
- In the failing runs, the destination route's RSC prefetches were served with `x-vercel-cache: PRERENDER`. We could never reproduce against local `next start`, with cold or warm caches.
- Browser split in the same CI pipeline on the same commit: chrome, ios (iPhone 12 emulation), and android (Pixel 5 emulation) failed on every retry while desktop safari and firefox passed, consistent with per-job prefetch/cache state rather than engine differences.
- 16.3.1's client segment-cache changes relative to 16.3.0 (metadata-only request tree guard, per-navigation cache map) do not fix it.
- Closest existing report appears to be #95347 (metadata boundary resume mismatch), but the signature differs.

Happy to provide Playwright traces (DOM snapshots and network logs) from the failing CI runs.

jasleenkaur-qed42:
## What

After a client soft navigation to an already-prefetched route, the previous route's resolved metadata (`og:*`, `twitter:*`, `description`) could stay mounted in `<head>` next to the new route's tags. Hard loads were always clean. Most visible with cold Vercel `PRERENDER` prefetch responses.

## Why

`Head` uses `useDeferredValue` to show the prefetch/shell head first, then the final head. Hoistable metadata is inserted into `document.head` outside the React tree, so swapping the RSC tree without remounting does not release the old tags.

## How

Key the deferred head by whether the prefetch or final value is showing, so the prefetch → final transition remounts and clears prior hoistables.

## Test plan

- [x] `HEADLESS=true NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/e2e/app-dir/metadata-soft-nav-cache-components/metadata-soft-nav-cache-components.test.ts`
- [ ] Manual: deploy repro from #97472, soft-nav `/` → article while prefetch is cold, assert a single `meta[property="og:type"]`

Fixes #97472

### DIFF
--- a/packages/next/src/client/components/app-router.tsx
+++ b/packages/next/src/client/components/app-router.tsx
@@ -3,6 +3,7 @@ import React, {
   useMemo,
   useInsertionEffect,
   useDeferredValue,
+  Fragment,
 } from 'react'
 import {
   AppRouterContext,
@@ -199,7 +200,16 @@ function Head({
   // We use `useDeferredValue` to handle switching between the prefetched and
   // final values. The second argument is returned on initial render, then it
   // re-renders with the first argument.
-  return useDeferredValue(head, resolvedPrefetchRsc)
+  //
+  // Hoistable metadata (`<meta>`, `<link>`) is inserted into `document.head`
+  // outside the React tree. Swapping the RSC tree without remounting can leave
+  // the previous hoistables mounted alongside the new ones (e.g. both
+  // `og:type=website` from a layout/prefetch shell and `og:type=article` from
+  // the destination). Key by which value is currently shown so the prefetch →
+  // final transition fully releases the old tags. See GitHub #97472.
+  const deferredHead = useDeferredValue(head, resolvedPrefetchRsc)
+  const phase = deferredHead === head ? 'h' : 'p'
+  return <Fragment key={phase}>{deferredHead}</Fragment>
 }
 
 /**
--- a/test/e2e/app-dir/metadata-soft-nav-cache-components/app/layout.tsx
+++ b/test/e2e/app-dir/metadata-soft-nav-cache-components/app/layout.tsx
@@ -6,6 +6,14 @@ export const metadata = {
   // to this default (or retain the previous route's title) — it must never be
   // dropped to an empty string (which makes the browser tab show the URL).
   title: { default: 'Home Default', template: '%s | Site' },
+  metadataBase: new URL('https://example.com'),
+  openGraph: {
+    type: 'website',
+    siteName: 'Site',
+    title: 'Home Default',
+    description: 'Layout open graph defaults',
+    url: '/',
+  },
 }
 
 export default function RootLayout({ children }: { children: ReactNode }) {
--- a/test/e2e/app-dir/metadata-soft-nav-cache-components/app/slow/page.tsx
+++ b/test/e2e/app-dir/metadata-soft-nav-cache-components/app/slow/page.tsx
@@ -6,7 +6,15 @@ import { Suspense } from 'react'
 // part of the prefetchable App Shell and must be fetched on navigation.
 export async function generateMetadata(): Promise<Metadata> {
   await connection()
-  return { title: 'Slow Page' }
+  return {
+    title: 'Slow Page',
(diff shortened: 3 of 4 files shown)


brookton:
Thanks for working on this. The keyed remount looks consistent with the observed stale hoistables. One validation gap: #97472 only reproduced on a fresh Vercel deployment while the destination prefetch response was served with `x-vercel-cache: PRERENDER`; we could not reproduce it with local `next start`.

Before this closes #97472, could you verify the original repro against a fresh deployment of this branch and confirm that the new regression test fails on the base commit without the `Head` change? Ideally, please also confirm that `description`, `og:*`, and `twitter:*` contain only the destination metadata after navigation.

jasleenkaur-qed42:
@brookton On the test: you're right to flag it. I reverted just the Head remount change locally, rebuilt, and reran replaces prefetched open graph tags instead of stacking them against the unpatched code — it still passes. So as written, the test doesn't fail on the base commit, which means it isn't a reliable regression guard yet. I'll work on tightening the repro (likely by making the prefetch/final RSC delivery timing closer to what a cold PRERENDER response looks like) so it actually fails without the fix and passes with it. I'll also expand the assertion to check description and twitter:*, not just og:type.

On the fresh-deployment verification: I don't have a Vercel deployment set up to reproduce the cold PRERENDER cache condition you described, so I can't do that verification myself. If you're able to test this branch against your own deployment (or the original repro environment from 
[vercel/next.js#97472](https://github.com/vercel/next.js/issues/97472)
), I'd appreciate it — otherwise, if you can point me to what specifically differs in that cold-cache path (e.g. is the prefetch RSC payload streamed differently, or is it just a matter of first-request latency), I can try to reproduce the same conditions in the local test harness instead.

jasleenkaur-qed42:
closing

Issue #83799, opened 2025-09-15
Turbopack builds do not correctly process all files (ex: turbopack runtime) through the browserslist system


krzkz94:
### Link to the code that reproduces this issue

https://github.com/krzkz94/turbopack-modern-syntax-browserslist

### To Reproduce

Clone the repository 
https://github.com/krzkz94/turbopack-modern-syntax-browserslist

Install dependencies 
`npm i`

Run the build
`npm run build`

Run the following command in the build folder which will grep all modern syntax
`chmod +x ./check-failures-only.sh && ./check-failures-only.sh`

The above is a simple helper tool to identify transpilation issues, only runs 'grep' against the most common modern syntax I could think of.

According to the default browserslist defined here
https://nextjs.org/docs/architecture/supported-browsers

The syntax below is not being transformed
**Optional Chaining (?.)** 
Introduced in: Chrome 80, Edge 80, Firefox 72, Opera 67, Safari 13.1
**Nullish Coalescing (??)**
Introduced in: Chrome 80, Edge 80, Firefox 72, Opera 67, Safari 13.1
**Nullish Assignment (??=)**
Introduced in: Chrome 85, Edge 85, Firefox 79, Opera 71, Safari 14
**Private Class Fields (#field)**
Introduced in: Chrome 74, Edge 79, Firefox 90, Opera 62, Safari 14.1
**globalThis**
Introduced in: Chrome 71, Edge 79, Firefox 65, Opera 58, Safari 12.1

One thing to note, the turbopack runtime is not transformed either (when it should)

Can a member of the Next/Turbopack team confirm or deny that every dependency (also transient dependencies) should be ran through the browserslist, even if it's not part of transpilePackages? 
The majority of nextjs compatibility issues comes from sub-dependencies (even 4 levels deep), so if you want to make your site compatible with for example smart tv's from 2022 (which run chrome 68), it's basically impossible as your transpilePackages would have to be thousands of lines long. 
I know for webpack at least, transpilePackages ignores sub-dependencies.

### Current vs. Expected behavior

Currently, a lot of modern syntax is not ran through the browserslist compatibility / polyfill / transpilation logic

Expected: All modern syntax that doesn't match the browserslist should be transformed.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
[...]

PR #97484, opened 2026-08-18
fix(turbopack): respect arrow function targets

fireairforce:
## What?

- Emit classic functions from async chunk and manifest loaders when the configured browser target does not support arrow functions.
- Prevent SWC compression from reintroducing arrow functions for those targets across browser, worker, evaluate, runtime, and Node.js chunks.
- Add production regression coverage for generated dynamic-import wrappers and object methods rewritten by the minifier.

## Why?

Turbopack already uses the target's arrow-function support for outer chunk factories, but some generated loaders still hardcoded arrow functions. SWC compression could also turn compatible functions and object methods back into arrow functions even when the output target did not support them.

As a result, production output could contain `ArrowFunctionExpression` nodes that were incompatible with the configured Browserslist target.

## How?

- Query `supports_arrow_functions` from the chunking environment when generating loaders and selecting minifier options.
- Generate equivalent classic-function callbacks for unsupported targets.
- Set SWC's `CompressOptions.arrows` from the same runtime capability.
- Parse emitted static JavaScript in the escheck production test and assert that old-browser output contains no arrow-function AST nodes.

## Tests

- `pnpm build-all`
- `cargo test -p turbopack-ecmascript respects_arrow_function_support`
- `cargo clippy --all-targets -- -D warnings --no-deps`
- `test/production/escheck-output` (4/4 passing)

Related to #83799


test/production/escheck-output/index.test.ts
```diff
@@ -1,11 +1,51 @@
 import { nextTestSetup } from 'e2e-utils'
 import { execSync } from 'child_process'
+import { readFileSync, readdirSync } from 'fs'
+import { join } from 'path'
+import { parse } from 'acorn'
```
mischnic: This fails with
```
next:typescript: 
//:typescript: test/production/escheck-output/index.test.ts(5,23): error TS2307: Cannot find module 'acorn' or its corresponding type declarations.
//:typescript:  ELIFECYCLE  Command failed with exit code 2.
next:typescript:  ELIFECYCLE  Command failed.
```

And doesn't escheck already cover this?

fireairforce: sorry i miss this, updated and fixed.

PR #97499, opened 2026-08-18
docs: clarify root layout creation with Turbopack

DavidIlie:
## Summary

Qualify the installation guide's claim that `next dev` creates a missing root layout.

Webpack development performs that recovery, while the default Turbopack development path does not. The Turbopack reference already documents the limitation, but the installation guide currently makes the promise unconditionally.

State that the automatic creation applies to Webpack and link to the existing Turbopack limitation.

## Verification

### DIFF
--- a/docs/01-app/01-getting-started/01-installation.mdx
+++ b/docs/01-app/01-getting-started/01-installation.mdx
@@ -213,7 +213,7 @@ Both `layout.tsx` and `page.tsx` will be rendered when the user visits the root 
 
 > **Good to know**:
 >
-> - If you forget to create the root layout, Next.js will automatically create this file when running the development server with `next dev`.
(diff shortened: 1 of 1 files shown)


docs/01-app/01-getting-started/01-installation.mdx
```diff
@@ -213,7 +213,7 @@ Both `layout.tsx` and `page.tsx` will be rendered when the user visits the root
 
 > **Good to know**:
 >
-> - If you forget to create the root layout, Next.js will automatically create this file when running the development server with `next dev`.
+> - When using webpack, Next.js will automatically create the root layout if you forget it when running `next dev`. [Turbopack does not support automatic root layout creation](/docs/app/api-reference/turbopack#framework-and-react-features), so you must create the file manually.
```
icyJoseph, line 216: Can we ask the Turbopack team about this once again. I do believe I've asked before and got confirmation on the intention. Since Turbopack is the default, we should lead this with something like:

`You must create a root layout, otherwise... — the webpack bundler auto-generates one for you`

After that I can also update the `/learn` material that states the very same thing.

PR #97503, opened 2026-08-18
[PPF] Do not mark complete shell requests as partial

lubieowoce:
The router code did not respect the isPartial byte for runtime app shells for some reason, potentially leading to unnecessary speculative prefetch requests even if the shell was in fact complete.

The comments say that the byte on those can't be trusted:
https://github.com/vercel/next.js/blob/d00eb6b6c471c9c51ef997f643b049210d9940a5/packages/next/src/client/components/segment-cache/cache.ts#L2694-L2696
 which is not true -- the byte is definitely set to partial when a shell contains holes (due to URL data or link data).
 
 Updates some tests that now no longer trigger a speculative prefetch because the shell is complete.
 (note that our tests use `prefetch = 'partial'` a lot, so some of them were triggering speculative prefetches even when they shouldn't have. see #97469 for more)

### DIFF
--- a/test/e2e/app-dir/segment-cache/prefetch-static-shell/prefetch-static-shell.test.ts
+++ b/test/e2e/app-dir/segment-cache/prefetch-static-shell/prefetch-static-shell.test.ts
@@ -123,11 +123,8 @@ describe('static App Shell prefetch attempt', () => {
         .elementByCss('input[data-link-accordion="/uses-cookies"]')
         .click()
     }, [
-      // The page (the new part) arrives in the runtime shell response.
-      // (The bare cookies() read itself isn't resolved by the runtime
-      // prerender — it stays a hole for the navigation-time dynamic
-      // request — so we only assert on the shell text.)
-      { includes: 'Cookies page shell text', kind: 'runtime' },
+      // Cookies are included in the runtime shell.
+      { includes: 'cookie-content', kind: 'runtime' },
       // No static attempt for the new part: the page content must not
       // arrive in a static per-segment response. (The route tree prefetch
       // is also `kind: 'static'`, but its response doesn't contain rendered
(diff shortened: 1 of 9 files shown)


test/e2e/app-dir/segment-cache/prefetch-static-shell/prefetch-static-shell.test.ts
```diff
@@ -123,11 +123,8 @@ describe('static App Shell prefetch attempt', () => {
         .elementByCss('input[data-link-accordion="/uses-cookies"]')
         .click()
     }, [
-      // The page (the new part) arrives in the runtime shell response.
-      // (The bare cookies() read itself isn't resolved by the runtime
-      // prerender — it stays a hole for the navigation-time dynamic
-      // request — so we only assert on the shell text.)
-      { includes: 'Cookies page shell text', kind: 'runtime' },
```
lubieowoce, line 130: this comment is straight up wrong

test/e2e/app-dir/segment-cache/prefetch-static-shell/prefetch-static-shell.test.ts
```diff
@@ -382,29 +379,15 @@ describe('static App Shell prefetch attempt', () => {
...
         .click()
     }, [
-      // Two runtime responses arrive, in order, and each resolves the
-      // cookies() read (a runtime prefetch renders with the request's
-      // cookies):
-      //
-      // 1. The Shell phase's runtime shell request — the same direct
-      //    runtime shell behavior the hint-unset tests above exercise,
-      //    except that here the session content resolves instead of
-      //    remaining a hole.
```
lubieowoce, line 392: wrong: `/uses-cookies` also resolves session content.

acdlite (approved):

PR #97519, opened 2026-08-18
fix: drop V8 stack frames from prerender interrupt abort reasons

### LINKED ISSUE #97351: Prerender abort reason carries V8 stack frames, so a retained AbortSignal pins the whole render graph (1.7 MB vs 1 KB)

https://github.com/devopsaptlife/nextjs-prerender-abort-retention

### To Reproduce

```bash
git clone https://github.com/devopsaptlife/nextjs-prerender-abort-retention
cd nextjs-prerender-abort-retention
npm install
npm run build
npm run instrument            # marks the two shipped creators, so the path is proven reached
npm start &                   # node --expose-gc .next/standalone/server.js
curl -s localhost:3000/p/a localhost:3000/p/b localhost:3000/p/c > /dev/null
cat hits.log                  # one HIT_DYNAMIC_RENDERING per request
```

`Date.now()` in a server component is synchronous platform IO, which under `cacheComponents`
routes through `node-environment-extensions/io-utils.js` → `io()` →
`abortOnSynchronousPlatformIOAccess` → `abortOnSynchronousDynamicDataAccess` →
`createPrerenderInterruptedError` → `controller.abort(error)`.

Then, for what that error costs while it is reachable:

```bash
node --expose-gc repro.mjs    # standalone, no dependencies
```

```
payload per render: 8 MB x 20 renders = 160 MB

baseline     retained while signals alive:  160 MB (of 160 MB)  -> PINNED
materialize  retained while signals alive:    0 MB (of 160 MB)  -> released
limit0       retained while signals alive:    0 MB (of 160 MB)  -> released
```

**Current:** `createPrerenderInterruptedError` builds a plain `new Error(message)` which becomes
`AbortSignal.reason`. V8 keeps an Error's structured stack trace (its internal `FixedArray` of
`CallSiteInfo`) until `.stack` is *read*, and nothing reads it. Each retained frame pins that
frame's function → its `Context` → the render's whole working set: the cached page value
`{ html, kind, postponed, rscData, segmentData, status }`, its `segmentData` map of per-segment
RSC `Buffer`s, plus the streams and promises. **1,756 KiB retained per reachable signal, versus
1 KiB if the reason carries no frames.**

**Expected:** an abort reason should be cheap to keep. Retaining an `AbortSignal` — which
instrumentation, error reporting and userland code all do routinely — should not transitively
retain an entire rendered page.

**Scope, stated honestly and up front:** this is a *latent amplifier*, not a claim that Next leaks
on its own. I measured that it does not. With the path firing on every request (403 creator hits,
verified by the markers), a clean app accumulated nothing:

| after | rss | heapUsed | arrayBuffers | live Errors | live AbortSignals |
|---|---|---|---|---|---|
| 0 pages | 110 MB | 27 MB | 0 MB | 502 | 3 |
| 80 | 295 MB | 27 MB | 0 MB | 501 | 3 |
| 240 | 308 MB | 26 MB | 0 MB | 501 | 3 |
| 320 | 299 MB | 27 MB | 0 MB | 501 | 3 |

So Next collects these normally. The request is to make the reason cheap so that *when* something
holds a signal, the cost is ~1 KiB rather than ~1.7 MiB.

### Why it matters in practice

On a production 16.3.0 deployment, from a full V8 heap snapshot with retainer paths computed over
**strong edges only** (weak edges retain nothing; including them yields plausible root paths that
keep nothing alive):

```
AbortSignal --property:<symbol kReason>--> Error --property:<symbol>--> array
  --internal:9--> system/CallSiteInfo --hidden[2]--> closure --internal:context-->
  system/Context --> Object{ html, kind, postponed, rscData, segmentData, status }
  --property:segmentData--> Map --> Buffer --> ArrayBuffer --> JSArrayBufferData
```

| metric | value |
|---|---|
| live `AbortSignal`s | **9,189** (3 in the clean app above) |
| live `Error`s still holding frames | 5,214 of 6,448 — 48,640 frames |
| cached page values carrying `segmentData` | 1,589 |
| of those, reachable **only** via a `CallSiteInfo` frame | **227 / 227 sampled** |
| share of all ArrayBuffer bytes under `segmentData` | **53%** |

Only ~270 of those 1,589 entries were inside Next's `memoryCache` LRU — which is correctly capped,
and whose sizing function does account for `segmentData` honestly. The rest were pinned outside
any cache, purely by frames. What keeps 9,189 signals reachable in that app is **not identified**
and may well be our problem, not Next's — but the frames are what turn it into ~2 GB instead of
~9 MB.

```
Operating System:
  Platform: linux
  Arch: x64
Binaries:
  Node: 24.19.0
Relevant Packages:
  next: 16.3.0  (checked byte-identical in 16.3.1-canary.15)
  react: 19.2.7
  react-dom: 19.2.7
Next.js Config:
  output: standalone
  cacheComponents: true
  partialPrefetching: true
```

Partial Prerendering (PPR), Runtime, Dynamic Routes

### Additional context

**Fix options, measured rather than reasoned about:**

- **`error.stack = \`${error.name}: ${error.message}\`` (assignment).** The `stack` *setter*
  overwrites V8's private slot without invoking `prepareStackTrace`. Hook-independent, no global
  state, ~0% CPU, message preserved.
- **`Error.stackTraceLimit = 0` around construction.** ~12× cheaper than today (the capture itself
  dominates), but it mutates a global and `controller.abort(reason)` dispatches listeners
  **synchronously**, so widening that window silently strips stacks from unrelated errors.
- **Reading `.stack` once.** Works *only* because Next installs a `prepareStackTrace` returning a
  string. A hook returning the raw CallSite array, or one that throws, leaves the frames in place —
  measured, 152.6 MB still retained.

**Two traps**, both measured: `delete err.stack` and
`Object.defineProperty(err, 'stack', {value})` do **not** release the frames — they drop the
accessor while the private slot survives. And retention is **binary, not proportional**:
`stackTraceLimit` of 10 / 5 / 3 / 1 all retain the full graph; only 0 releases it.

**Bare aborts retain identically.** Most of `app-render.js`'s ~18 `.abort(...)` sites pass no
reason, so Node synthesizes a `DOMException`, which is `instanceof Error` and carries the same
lazy `stack` accessor — measured 1,756 KiB vs 1,754 KiB for `abort(new Error())`. Nine of those
sites are propagation of the form `x.signal.addEventListener("abort", () => y.abort())`, which
drops the incoming reason and mints a fresh `DOMException`.

**Nothing in production reads these frames**, so removing them looks observably neutral:
`isPrerenderInterruptedError` tests `digest`/`name`/`message`/`instanceof`;
`create-error-handler.js` early-returns on `digest` before its `.stack` reads; `applyOwnerStack` is
`NODE_ENV !== 'production'`-gated. But any replacement reason **must** keep
`name === 'AbortError'` — `pipe-readable.js`'s `isAbortError` tests exactly that, and several call
sites rely on it to avoid logging normal aborts as real errors.

**Where the creator actually ships**, checked against a real `next build` standalone output rather
than assumed, since this decides where a fix has to land:

| file | ships in standalone? | contains the creator? |
|---|---|---|
| `dist/server/app-render/dynamic-rendering.js` | yes | yes |
| `dist/compiled/next-server/app-route-turbo.runtime.prod.js` | yes | yes |
| `dist/compiled/next-server/app-page-turbo.runtime.prod.js` | yes | no — tree-shaken to the checker |
| `dist/compiled/next-server/app-page.runtime.prod.js` | no | — |
| `dist/compiled/next-server/server.runtime.prod.js` | no | — |

Related symptom report: #84648.

shubhxho:
### Fixing a bug

- Related issues linked using `fixes #97351`
- Tests added in `packages/next/src/server/app-render/dynamic-rendering.test.ts`

## What?

A prerender interrupt used to `new Error()` during the render and store that object as `AbortSignal.reason`. V8 CallSites on an Error created on the render stack include the Server Component frame, so a retained signal pins the whole render graph (~1.7MB vs ~1KB).

## Why?

#97351. The first version of this PR assigned `error.stack` to drop frames. That is not the right fix — we cannot expect every Error to limit its frames.

## How?

Create one interrupt Error at module load and reuse it for `controller.abort(...)` and the throw. It is still an `Error` with `digest === NEXT_PRERENDER_INTERRUPTED`, so `isPrerenderInterruptedError` is unchanged. The expression that interrupted the prerender is recorded on `dynamicTracking` (with a stack only when `isDebugDynamicAccesses` is on).

Not in this PR: `io-utils` still aborts with a per-render `syncIOError` (a different diagnostic, already stored on `syncDynamicErrorWithStack`), and the `use cache` `AbortSignal.any` listener leak (#97363).

Fixes #97351

### DIFF
--- a/packages/next/src/server/app-render/dynamic-rendering.test.ts
+++ b/packages/next/src/server/app-render/dynamic-rendering.test.ts
@@ -2,22 +2,45 @@
  * @jest-environment node
  */
 
-import {
-  createPrerenderInterruptedError,
-  isPrerenderInterruptedError,
-} from './dynamic-rendering'
+import { isPrerenderInterruptedError } from './dynamic-rendering'
+import { abortAndThrowOnSynchronousRequestDataAccess } from './dynamic-rendering'
 
-describe('createPrerenderInterruptedError', () => {
-  it('is recognized by isPrerenderInterruptedError and does not keep V8 frames', () => {
-    const error = createPrerenderInterruptedError(
-      'Route / needs to bail out of prerendering at this point because it used Date.now().'
-    )
+function makePrerenderStore() {
+  const controller = new AbortController()
+  return {
+    type: 'prerender',
+    controller,
+    dynamicTracking: {
+      isDebugDynamicAccesses: false,
+      dynamicAccesses: [],
+      syncDynamicErrorWithStack: null,
+      syncDynamicErrorWithStackPostMicrotask: false,
+    },
+    runtimeDataAccessed: { resolve() {} },
+    shouldAttemptStaticPrefetch: null,
+  } as any
+}
 
-    expect(isPrerenderInterruptedError(error)).toBe(true)
-    expect(error.name).toBe('Error')
-    // Assignment materializes the stack as name + message only, dropping the
-    // CallSite frames that would retain the creating closure / render graph.
-    expect(error.stack).toBe(`${error.name}: ${error.message}`)
-    expect(error.stack).not.toMatch(/\n\s+at /)
+describe('prerender interrupt abort reason', () => {
+  it('reuses one Error constructed at module load, not on the render stack', () => {
+    function render() {
+      try {
+        abortAndThrowOnSynchronousRequestDataAccess(
+          '/',
+          'Date.now()',
+          new Error('sync'),
+          makePrerenderStore()
+        )
+      } catch (err) {
+        return err as Error
+      }
+    }
+
+    const first = render()
+    const second = render()
+
+    expect(first).toBe(second)
+    expect(isPrerenderInterruptedError(first)).toBe(true)
+    expect(first.stack).not.toMatch(/at render /)
   })
 })
--- a/packages/next/src/server/app-render/dynamic-rendering.test.ts
+++ b/packages/next/src/server/app-render/dynamic-rendering.test.ts
@@ -1,0 +1,23 @@
+/**
+ * @jest-environment node
+ */
+
+import {
+  createPrerenderInterruptedError,
+  isPrerenderInterruptedError,
+} from './dynamic-rendering'
+
+describe('createPrerenderInterruptedError', () => {
+  it('is recognized by isPrerenderInterruptedError and does not keep V8 frames', () => {
+    const error = createPrerenderInterruptedError(
+      'Route / needs to bail out of prerendering at this point because it used Date.now().'
+    )
+
+    expect(isPrerenderInterruptedError(error)).toBe(true)
+    expect(error.name).toBe('Error')
+    // Assignment materializes the stack as name + message only, dropping the
+    // CallSite frames that would retain the creating closure / render graph.
+    expect(error.stack).toBe(`${error.name}: ${error.message}`)
+    expect(error.stack).not.toMatch(/\n\s+at /)
+  })
+})
--- a/packages/next/src/server/app-render/dynamic-rendering.ts
+++ b/packages/next/src/server/app-render/dynamic-rendering.ts
@@ -292,15 +292,14 @@ export function trackDynamicDataInDynamicRender(workUnitStore: WorkUnitStore) {
 }
 
 function abortOnSynchronousDynamicDataAccess(
-  route: string,
+  _route: string,
   expression: string,
   prerenderStore: PrerenderStoreModern
 ): void {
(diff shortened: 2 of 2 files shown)


eps1lon (changes requested):
We want to fix the root cause here instead. We can't expect every Error to limit its frames.

PR #97522, opened 2026-08-18
Fix retained fetch dedupe response bodies

gaearon:
## Summary

Release response bodies retained by server-side fetch deduplication when the React cache scope completes.

`cloneResponse()` tees each fetched response and keeps one branch for duplicate fetches in the same scope. As the returned branch is read, the retained branch buffers the response body. Previously, those off-heap bytes were only released after garbage collection.

Each new dedupe entry is now registered with `React.cacheSignal()`. Entries in the same scope share one abort listener, which cancels their retained response bodies when the scope finishes. Duplicate fetches continue to replace the retained response without adding another registration.

This carries forward the fix from #94238 by @chanpark-xyz and addresses the fetch-dedupe retention reported in #92287.

## Verification

The exact pushed commit was tested with the public standalone reproduction using 64 concurrent clients and 2 MiB responses. All 3,280 requests completed successfully. `arrayBuffers` remained between 208 MB and 283 MB during the sampled load, then fell to 49 MB after 15 seconds idle.

- `pnpm --filter=next build`
- `pnpm --filter=next types`
- `pnpm exec jest packages/next/src/server/lib/dedupe-fetch.test.ts --runInBand`


packages/next/src/server/lib/dedupe-fetch.ts
```diff
@@ -113,16 +167,19 @@ export function createDedupeFetch(originalFetch: typeof fetch) {
     // We pass the original arguments here in case normalizing the Request
     // doesn't include all the options in this environment.
     const promise = originalFetch(resource, options)
-    const entry: CacheEntry = [cacheKey, promise, null]
+    // React aborts this signal when the cache scope settles, after which its
+    // retained response clones are no longer needed.
+    const entry: CacheEntry = [cacheKey, promise, null, false]
     cacheEntries.push(entry)
+    registerForCleanup(entry, React.cacheSignal() as AbortSignal | null)
```
gaearon, line 174: Note sure why this cast is needed. For some reason, React types say `export interface CacheSignal {}` but docs say it's `AbortSignal`: https://react.dev/reference/react/cacheSignal.

unstubbable (commented):
It's unfortunate that no e2e test caught this, but I'm afraid this would break reading deduped fetches in `after()` because `React.cache` extends into that phase, but the `cacheSignal` is already aborted there. I think we should release at the end of the request instead.

packages/next/src/server/lib/dedupe-fetch.ts
eps1lon: Docs are misleading. It's only an `AbortSignal` in DOM. Not sure why we don't have `react-dom` types available here.

PR #97524, opened 2026-08-18
[PPF] Remove `unstable_eager`

lubieowoce:
The implementation of `unstable_eager` is currently incomplete, causes bugs in the main Partial Prefetching flow (see #97469), and is generally not worth the complexity it adds, so we're removing it. We may add something like this back later if we think it's useful.

Importantly, removing it fixes a bug that caused apps that use incremental opt-in (i.e. no global `partialPrefetching` config and `export const prefetch = 'partial'`) to issue speculative prefetches for all links without `prefetch={true}`, because any segment without an explicit `prefetch = "partial"` was marked as needing eager prefetching, which was then propagated to the whole route. See #97469 for a full explanation.

This fix causes a lot of changes in tests, which were previously incorrectly asserting that we should be seeing two requests. Now that the unnecesary speculative prefetch is gone, most of them only see one.


packages/next/src/client/components/segment-cache/scheduler.ts
```diff
@@ -1172,17 +1169,11 @@ function pingStaticHead(
  * walks prefetch static data and partial entries are acceptable — the
  * dynamic holes are filled by the navigation-time request.
  *
- * Note that on a Partial Prefetching route, non-eager subtrees are still
- * skipped by the Speculative pass of a default (auto) link — eagerness is
- * unaffected by this predicate. But every segment the pass DOES walk (eager
- * segments, and everything on a `prefetch={true}` walk) is held to the
- * runtime-completeness contract. The contract is affordable because most
```
lubieowoce, line 1179: @acdlite i'm having trouble parsing this, "non-eager subtrees are still skipped by the Speculative pass of a default (auto) link" doesn't sound like it was ever true? i'm just gonna remove this section

also not sure about the "The runtime contract is affordable because most routes carry the ShouldAttemptStaticPrefetch hint" bit below. do we actually know this? and even if we did i don't really get why we're justifying the behavior here

packages/next/src/client/components/segment-cache/scheduler.ts
```diff
@@ -2678,24 +2665,21 @@ function doesCurrentSegmentMatchCachedSegment(
...
- * true. However, we also will do a speculative prefetch if the prefetching
- * mode of the segment is set to "unstable_eager".
+ * Decides whether to speculatively prefetch a subtree. Under Partial
+ * Prefetching we only do this if the Link's prefetch prop is set to true —
+ * otherwise the subtree relies on the shell that the Shell phase prefetches.
  */
-export function subtreeHasSpeculativePrefetch(
-  fetchStrategy: FetchStrategy,
-  prefetchHints: number
+export function needsSpeculativePrefetch(
```
lubieowoce, line 2672: this is now not tied to the tree at all, so i renamed it from `subtreeHasSpeculativePrefetch` to `needsSpeculativePrefetch`.

also note that the `(hints & SubtreeHasPartialPrefetching) !== 0` check is only correct **if we look at the root layout's hints** -- if we have sibling slots `@foo` and `@bar` and only `@foo` of them sets `prefetch = 'partial'`, then we'd get `false` in the `@bar` subtree even though the whole route is opted into PPF. so the only place where can relieably know that this hint has been set is the root (because it propagates upwards).

this kinda makes me think that storing`SubtreeHasPartialPrefetching` as a hint on a segment is a bit error-prone, because for our purposes it's really a route-level property

test/e2e/app-dir/segment-cache/prefetch-app-shell/prefetch-app-shell.test.ts
```diff
@@ -28,19 +28,14 @@ describe('App Shell prefetching', () => {
...
         .click()
     }, [
-      // Two runtime responses carry the shell text, in order: the Shell
-      // phase's runtime App Shell request, then the Speculative phase's
-      // batched per-link runtime prefetch. The route reads request data,
-      // so its static-attempt hint is unset and the prefetch deopts to
-      // runtime requests — the runtime-completeness contract of Partial
-      // Prefetching routes.
-      { includes: 'App shell for posts', kind: 'runtime' },
       { includes: 'App shell for posts', kind: 'runtime' },
```
lubieowoce, line 43: this was not correct, it was asserting that we should be seeing a shell and speculative request, which we shouldn't have been. after the fix, we only see one. most of the test changes are fixing these

packages/next/src/client/components/segment-cache/scheduler.ts
acdlite: It's just saying that regardless of the return value of this function, the Speculative phase is skipped by default (when Partial Prefetching is on)

We can delete the second sentence. It's just rationalizing why requiring completeness is feasible. The better explanation is because you only pay for it if you use an API like `cookies` somewhere, i.e. pages that are fully static still require zero runtime requests.

packages/next/src/client/components/segment-cache/scheduler.ts
acdlite: yeah we can remove the subtree bit in a follow up.

acdlite (approved):

PR #97526, opened 2026-08-18
Capture initial URL before hydration

gaearon:
Replaces the detection from https://github.com/vercel/next.js/pull/96252.

Right now we only listen to `popstate` after hydration. But `popstate` can fire before hydration, and if we miss it, we show the wrong page. That's what happened before #95682.

In https://github.com/vercel/next.js/pull/96252, we tried to detect whether we're still on the right page when we hydrate, using the Navigation API. This doesn't work in browsers that don't have the Navigation API. It's also a bit silly. We can know where we started by writing it down early, so then there's no need to guess it after the fact.

This PR captures the initial URL in an inline script that runs early. When the router mounts, if the URL is different from the initially captured one, it means the user went somewhere before we were listening. So we don't write to history yet (that would overwrite the entry they're on now), and we instead handle the change the same way we'd handle a `popstate`.

The inline script prefers `navigation.activation.entry.url` but falls back to `location.href` if there's no Navigation API.

Initializing from the captured URL also fixes hydration mismatches. After a Back press, `usePathname` used to hydrate with the URL the user went back to, but the server HTML was for the page they started on. So this caused a hydration mismatch. You can see this fixing the TODOs in the tests from [#97651](https://github.com/vercel/next.js/pull/97651).

This whole fix relies on `bootstrapScriptContent`, which React emits after the shell. Without the Navigation API, a Back press before that script has run is still not handled. To be more resilient there we'd have to inject the script earlier, into the `<head>`. I don't know how to do this correctly, so I haven't tried.

## Test Plan

Old tests still pass on the new mechanism, and the TODOs from [#97651](https://github.com/vercel/next.js/pull/97651) are flipped. Added regression tests for the error documents (`global-error` and a prerendered 404) so we don't forget `bootstrapScriptContent` in those code paths.

### DIFF
--- a/packages/next/src/client/app-index.tsx
+++ b/packages/next/src/client/app-index.tsx
@@ -19,7 +19,7 @@ import {
   type AppRouterActionQueue,
   createMutableActionQueue,
 } from './components/app-router-instance'
-import AppRouter from './components/app-router'
+import AppRouter, { setInitialUrl } from './components/app-router'
 import type { InitialRSCPayload } from '../shared/lib/app-router-types'
 import { createInitialRouterState } from './components/router-reducer/create-initial-router-state'
 import { MissingSlotContext } from '../shared/lib/app-router-context.shared-runtime'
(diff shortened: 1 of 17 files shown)


95959595t:
maitrongtin78

0384518728

Vào 05:52, T.4, 19 Th8, 2026 dan ***@***.***> đã viết:

PR #97535, opened 2026-08-19
test: run cargo unit tests with miri as well

mischnic:
### What?

Adds a CI job that runs selected Rust unit tests under [Miri](https://github.com/rust-lang/miri) to detect undefined behavior in unsafe code.

The job currently covers the low-level crates that are compatible with Miri:

- `turbo-persistence`
- `turbo-rcstr`
- `turbo-tasks-malloc`

It also makes those crates Miri-compatible by:

- using provenance-free tagged-pointer operations in `turbo-rcstr`;
- disabling mimalloc and native compression under Miri;
- making mmap support a default-enabled Cargo feature that is disabled for Miri and wasm;
- running persistence through file I/O when mmap is disabled;
- skipping only tests measured to exceed the Miri time budget;
- removing leaked test arenas from the analyzer predicate tests.

### Why?

Miri can detect invalid memory access and other undefined behavior that normal Rust tests may not expose. Running it in CI gives low-level unsafe code an additional correctness check.

The job uses an explicit package allowlist because Turbo Tasks builds its generated registry from linker sections, and Miri does not support the linker-defined section symbols required by that registry.

### How?

- Installs the `miri` Rust component in CI and the development container.
- Adds a dedicated `test-cargo-unit-miri` task and reusable workflow configuration.
- Makes `memmap2` optional behind the default `mmap` feature and disables that feature in Miri CI and the wasm dependency graph.
- Uses structural `cfg(miri)` branches for allocator and compression paths Miri cannot execute.
- Re-enables three persistence compaction tests after measuring them successfully under Miri.
- Keeps normal native behavior unchanged and documents measured Miri exclusions at the affected tests.

Verification included focused normal and Miri tests for persistence, rcstr, allocation accounting, compression, and the refactored leak-free predicate cases.


mischnic:
* **#97535** <a href="https://app.graphite.com/github/pr/vercel/next.js/97535?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/97535?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbo-rcstr/src/tagged_value.rs
```diff
@@ -280,40 +323,28 @@ impl TaggedValue {
...
-        #[cfg(not(any(
-            target_pointer_width = "32",
-            target_pointer_width = "16",
-            feature = "atom_size_64",
-            feature = "atom_size_128"
-        )))]
-        {
-            self.value.addr().get()
-        }
->>>>>>> 200e5a17 (tagged_value.rs: no integer-pointer conversion)
```
bgw: Huh?

mischnic: you somehow commented on an old state of this PR.

bgw (approved):

turbopack/crates/turbo-persistence/src/lib.rs
```diff
@@ -45,6 +45,22 @@ pub enum AccessMode {
     File,
 }
 
+impl AccessMode {
+    /// Returns the access mode that can be used by the current execution environment.
+    #[inline]
+    pub(crate) fn effective(self) -> Self {
```
mischnic: Not a fan of `.effective()`. It should just use the correct enum variant in the first place.

turbopack/crates/turbo-rcstr/src/tagged_value.rs
bgw: It's a specific commit. You're right though, it's not in the overall diff. This is getting squashed anyways, so it's cool

.github/workflows/build_reusable.yml
```diff
@@ -36,6 +36,10 @@ on:
         required: false
         description: 'if nextest rust dep is needed'
         type: string
+      needsRustJsInstall:
+        required: false
+        description: 'if Rust JS dependencies should be installed'
+        type: string
```
bgw, line 42: This seems better 👍

turbopack/crates/turbo-persistence/src/lib.rs
sokra: Maybe align with this commit https://github.com/vercel/next.js/commit/c3f89815ff2854a8b9fa849b5b5ba80f57761fad
It does the same for wasm

mischnic: which PR is that included in?

turbopack/crates/turbo-persistence/src/rc_bytes.rs
```diff
@@ -19,8 +19,13 @@ use crate::{
 /// Uses `Rc` for all refcounting, eliminating atomic operations.
 #[derive(Clone)]
 enum Backing {
-    Rc { _backing: Rc<[u8]> },
-    Mmap { _backing: Rc<Mmap> },
+    Rc {
+        _backing: Rc<[u8]>,
+    },
+    #[cfg_attr(miri, allow(dead_code))]
```
sokra, line 25: Let's make the whole `memmap2` dependency disabled in miri mode. It should be conditional in Cargo.toml on `not(miri)`



```suggestion
    #[cfg(not(miri))]
```

turbopack/crates/turbo-persistence/src/tests.rs
sokra, line 1: Why are all these test cases disabled with miri? We want to run them. Can we make that possible?

turbopack/crates/turbo-rcstr/src/tagged_value.rs
```diff
@@ -209,11 +219,42 @@ impl TaggedValue {
...
+        )))]
+        {
+            // An integer → pointer transmute, which const evaluation permits.
+            let value = value.get() as RawTaggedValue;
+            Self {
+                #[allow(clippy::transmute_int_to_non_zero)]
+                value: unsafe {
+                    std::mem::transmute::<RawTaggedValue, RawTaggedNonZeroValue>(value)
+                },
+            }
```
sokra, line 257: Do we need two code paths here? Can we always use the provenance-free path?

turbopack/crates/turbo-rcstr/src/tagged_value.rs
sokra, line 347: Same?

sokra (approved):

PR #97541, opened 2026-08-19
[test] Replace the `turbopack-reports` `sqlite3` dependency with a local addon fixture

eps1lon:
The suite installed `sqlite3` only to get a package that locates its compiled binary through `require('bindings')(...)`, which is the shape turbopack issue 5913 was about. It now carries its own `native-addon` and `bindings` packages instead, both installed as relative `file:` dependencies, with node-gyp compiling the addon during install.

The addon is compiled rather than stubbed because the assertion reads a value off the loaded binary, so a real `process.dlopen()` has to happen. Compiling at install time also keeps the binary matched to whichever Node ABI is running, which a checked-in binary could not do, since a non-context-aware addon cannot use Node-API and is therefore ABI-locked. The page now renders the addon's constant, so a module that resolved to nothing fails the test instead of passing quietly.

Both packages are `file:` rather than `link:` dependencies. A linked package resolves to a path inside the app, which the bundler then treats as app code and tries to bundle, and `bindings` contains a `require` it cannot resolve statically. `serverExternalPackages` is needed because the app router bundles `node_modules` by default, and `sqlite3` only avoided that by being on the built-in list in `server-external-packages.jsonc`.

The fixture packages are registered in `modulePathIgnorePatterns`, following the entries already there. Jest is configured with `throwOnModuleCollision`, so a package name appearing twice outside `node_modules` aborts the whole test run, and the layers above add more copies of `bindings`.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -33,6 +33,8 @@ const customJestConfig = {
     '<rootDir>/e2e/app-dir/self-importing-package/internal-pkg',
     '<rootDir>/e2e/app-dir/self-importing-package-monorepo/internal-pkg',
     '<rootDir>/e2e/app-dir/server-source-maps/fixtures/default/internal-pkg',
+    '<rootDir>/e2e/app-dir/turbopack-reports/bindings',
+    '<rootDir>/e2e/app-dir/turbopack-reports/native-addon',
     '<rootDir>/e2e/transpile-packages-typescript-foreign/pkg',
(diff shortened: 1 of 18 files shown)


mischnic:
Do we really need to compile something manually here? Can't we just use some other npm package with prebuilt binaries like `@swc/core`?

eps1lon:
Yes. Using a 3rd party dependency lead to this whole issue.

mischnic (approved):

PR #97548, opened 2026-08-19
docs: Explicit cache output description

icyJoseph:
Completing the input-output pair for cached functions.

This is addressing user confusion around what happens to a cached function, and functions within the scope.

### DIFF
--- a/docs/01-app/03-api-reference/01-directives/use-cache.mdx
+++ b/docs/01-app/03-api-reference/01-directives/use-cache.mdx
@@ -602,7 +602,7 @@ export async function getTopProducts() {
 }
 ```
 
-Non-function exports pass through unchanged, which is how a `page` or `layout` file keeps its `metadata`, `viewport`, and [route segment config](/docs/app/api-reference/file-conventions/route-segment-config). Framework exports are covered like any other, so [`generateMetadata`](/docs/app/api-reference/functions/generate-metadata) and [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params) must be async in such a file.
+Framework function exports are covered like any other, so [`generateMetadata`](/docs/app/api-reference/functions/generate-metadata) and [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params) must be async in such a file.
 
 > **Good to know**:
 >
--- a/docs/01-app/03-api-reference/01-directives/use-cache.mdx
+++ b/docs/01-app/03-api-reference/01-directives/use-cache.mdx
@@ -45,7 +45,7 @@ const nextConfig = {
 module.exports = nextConfig
 ```
 
-Then, add `use cache` at the function, component, or file level. All functions and components using `use cache` must be async:
+Then, add `use cache` at the function, component, or file level. Cached functions and components must be async:
 
 ```tsx
 // Function level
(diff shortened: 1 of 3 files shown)


docs/01-app/03-api-reference/01-directives/use-cache.mdx
```diff
@@ -45,28 +45,32 @@ const nextConfig = {
 module.exports = nextConfig
 ```
 
-Then, add `use cache` at the file, component, or function level. All functions and components using `use cache` must be async. When used at file level, every exported function becomes a cached function and must also be async:
+Then, add `use cache` at the function, component, or file level. All functions and components using `use cache` must be async:
```
aurorascharff: All functions and components using `use cache` must be async:

"within a use cache scope" maybe since it might be uncler how they "use" it?

docs/01-app/03-api-reference/01-directives/use-cache.mdx
```diff
@@ -102,6 +106,41 @@ In the snippet above, `userId` is captured from the outer scope and `filter` is
...
+}
+
+export async function getOrders(accountId: string) {
+  'use cache'
+  cacheLife('hours')
+
+  return db.orders.findMany({ where: { accountId } })
+}
+
+export async function getOrderTotals(accountId: string) {
```
aurorascharff, line 135: Why did we not cache this one and rather the one that uses it? Maybe we should explain the reason youd want to scope it in this way

aurorascharff: Or state that this is an example created to be able to explain the specifics

docs/01-app/03-api-reference/01-directives/use-cache.mdx
```diff
@@ -102,6 +106,41 @@ In the snippet above, `userId` is captured from the outer scope and `filter` is
...
+  return db.orders.aggregate({ where: { accountId }, _sum: { amount: true } })
+}
+```
+
+When `getOrderSummary` runs, `getOrders` looks up its own storage entry first. Until the entry for that `accountId` revalidates, `db.orders.findMany` doesn't run, and the stored rows become part of `getOrderSummary`'s output. Without a directive, `getOrderTotals` has no entry of its own. Its query runs whenever `getOrderSummary`'s body runs, and on every call from an uncached scope.
```
aurorascharff: Could we use a numered list or something so this is easier to follow

docs/01-app/03-api-reference/01-directives/use-cache.mdx
aurorascharff: would be interesting to see a combination of different cache scopes explained. this is only function level which is also below

docs/01-app/03-api-reference/01-directives/use-cache.mdx
```diff
@@ -352,9 +391,87 @@ Both `cacheLife` and `cacheTag` integrate across client and server caching layer
...
+export async function getMonthlyTotals(accountId: string) {
+  return db.orders.aggregate({ where: { accountId }, _sum: { amount: true } })
+}
+
+export async function getTopProducts() {
+  return db.products.findMany({ orderBy: { sales: 'desc' }, take: 10 })
+}
+```
+
+Non-function exports pass through unchanged, which is how a `page` or `layout` file keeps its `metadata`, `viewport`, and [route segment config](/docs/app/api-reference/file-conventions/route-segment-config). Framework exports are covered like any other, so [`generateMetadata`](/docs/app/api-reference/functions/generate-metadata) and [`generateStaticParams`](/docs/app/api-reference/functions/generate-static-params) must be async in such a file.
```
aurorascharff: Kinda unclear that youre talking about page.tsx level caching here since the example is just a file with data functions

docs/01-app/03-api-reference/01-directives/use-cache.mdx
```diff
@@ -412,69 +529,7 @@ export default async function Page() {
...
-  'use cache'
-
-  const res = await fetch('https://api.example.com/data')
-  const data = await res.json()
-  return data
-}
-```
-
-> **Good to know:** When a cached directive (`use cache`, [`use cache: private`](/docs/app/api-reference/directives/use-cache-private), or [`use cache: remote`](/docs/app/api-reference/directives/use-cache-remote)) is at the top of a file, you can import its exported functions into a Client Component and call them directly; they run on the server and return the result, similar to a [Server Function](/docs/app/glossary#server-function). Prefer calling cached functions on the server and passing results down as props.
+> - When a cached directive (`use cache`, [`use cache: private`](/docs/app/api-reference/directives/use-cache-private), or [`use cache: remote`](/docs/app/api-reference/directives/use-cache-remote)) is at the top of a file, you can import its exported functions into a Client Component and call them directly; they run on the server and return the result, similar to a [Server Function](/docs/app/glossary#server-function). Prefer calling cached functions on the server and passing results down as props.
```
aurorascharff: I did not know this...?? In what situations can it be useful?

icyJoseph: It is useful to know it can happen — 

- https://github.com/vercel/next.js/pull/94301
- https://github.com/vercel/next.js/discussions/94286

icyJoseph: to be clear this was added on this PR — I just forgot to move it with its parent section

aurorascharff (approved):

PR #97551, opened 2026-08-19
docs: generateMetadata values should be serializable with use cache

devjiwonchoi:
### Why?

generateMetadata can be marked with use cache, but Cache Function return values must be serializable. The metadataBase examples use URL objects, so the interaction needs co-located guidance for cached metadata.

### How?

Add a Good to know note that recommends returning metadataBase as a string, such as with url.toString(), and links to the use cache serialization requirements. Runtime behavior is unchanged.

### DIFF
--- a/docs/01-app/03-api-reference/04-functions/generate-metadata.mdx
+++ b/docs/01-app/03-api-reference/04-functions/generate-metadata.mdx
@@ -421,7 +421,7 @@ export const metadata = {
 > **Good to know**:
 >
 > - `metadataBase` is typically set in root `app/layout.js` to apply to URL-based `metadata` fields across all routes.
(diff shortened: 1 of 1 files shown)


docs/01-app/03-api-reference/04-functions/generate-metadata.mdx
```diff
@@ -421,6 +421,7 @@ export const metadata = {
 > **Good to know**:
 >
 > - `metadataBase` is typically set in root `app/layout.js` to apply to URL-based `metadata` fields across all routes.
+> - When `generateMetadata` uses `'use cache'`, its return value must be serializable. `URL` instances are not supported by Cache Functions, so provide `metadataBase` as a string, such as `url.toString()`. See [`use cache` serialization requirements](/docs/app/api-reference/directives/use-cache#serialization).
```
icyJoseph:
```suggestion
> - When `generateMetadata` uses `'use cache'`, its return value must be serializable. `URL` instances are not supported by Cache Functions, so values such as `metadataBase` should be returned as strings (for example, `url.toString()`). See [`use cache` serialization requirements](/docs/app/api-reference/directives/use-cache#serialization).
```

icyJoseph (approved):

icyJoseph (approved):

PR #97559, opened 2026-08-19
docs: add missing cache-components error page

jaineelmodi11:
### What

Adds `errors/cache-components.mdx`. The page is currently missing.

### Why

Route Handlers throw a `DynamicServerError` whose message links to `https://nextjs.org/docs/messages/cache-components`:

```ts
// packages/next/src/server/route-modules/app-route/module.ts#L1442
function createCacheComponentsError(route: string) {
  return new DynamicServerError(
    `Route ${route} couldn't be rendered statically because it used IO that was not cached. See more info here: https://nextjs.org/docs/messages/cache-components`
  )
}
```

That URL currently returns a 404.

The error is reachable from three call sites in the same file (L739, L750, L756) and fires when a `GET` handler awaits uncached IO with `cacheComponents: true`. Anyone who hits it and follows the link lands on a missing page.

### What the page covers

- Why the bail-out happens during prerendering
- Caching the data with `use cache` + `cacheLife`, following the Route Handler pattern already documented in the Cache Components migration guide
- Opting into request-time rendering with `connection()`

### How I found it

I diffed every `/docs/messages/` slug referenced in `packages/next/src` (166 distinct) against the pages in `errors/` (253 files). Six were unmatched; four resolve through redirects and two genuinely 404. The other one is `nested-middleware`, which I'll open as a separate PR.

### DIFF
--- a/errors/cache-components.mdx
+++ b/errors/cache-components.mdx
@@ -1,0 +1,88 @@
+---
+title: Route couldn't be rendered statically because it used uncached IO
+---
+
+## Why This Error Occurred
+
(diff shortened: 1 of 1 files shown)


dianatofficial (commented):

PR #97560, opened 2026-08-19
docs: add missing nested-middleware error page

jaineelmodi11:
### What

Adds `errors/nested-middleware.mdx`. The page is currently missing.

### Why

`NestedMiddlewareError` links to `https://nextjs.org/docs/messages/nested-middleware`:

```ts
// packages/next/src/build/utils.ts#L1534
export class NestedMiddlewareError extends Error {
  constructor(...) {
    super(
      `Nested Middleware is not allowed, found:\n` +
        ...
        `Read More - https://nextjs.org/docs/messages/nested-middleware`
    )
  }
}
```

That URL currently returns a 404.

The error is logged from `setup-dev-bundler.ts#L1020` when a `_middleware.{js,ts}` file is found under `pages` (detected at L810). The file is skipped rather than executed, so the logic in it silently never runs — which makes the missing explanation more costly than usual.

### What the page covers

- That the nested file is skipped, not executed
- Consolidating into a single root file, which in Next.js 16 is `proxy.ts`
- A note that `middleware` is deprecated and renamed to `proxy`, with the `middleware-to-proxy` codemod

### Related

Same audit as the `cache-components` page — these were the only two `/docs/messages/` links in the source without a corresponding page.

### DIFF
--- a/errors/nested-middleware.mdx
+++ b/errors/nested-middleware.mdx
@@ -1,0 +1,60 @@
+---
+title: Nested Middleware is not allowed
+---
+
+## Why This Error Occurred
+
+A file named `_middleware.js` or `_middleware.ts` was found inside the `pages` directory. Nested Middleware was the beta convention in Next.js 12.0 and 12.1, where Middleware ran based on the directory it was declared in. It was replaced in Next.js 12.2 by a single root file.
+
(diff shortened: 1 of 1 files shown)


dianatofficial (commented):

PR #97563, opened 2026-08-19
Add deploy release test skill

timneutkens:
## Summary

- add a repository-local skill for manually triggering `test_e2e_deploy_release.yml`
- require the exact commit preview package to be available before dispatch
- preserve default workflow inputs and verify the resulting branch, SHA, and package URL

### DIFF
--- a/.agents/skills/deploy-release-test/SKILL.md
+++ b/.agents/skills/deploy-release-test/SKILL.md
@@ -2,18 +2,35 @@
 name: deploy-release-test
 description: >
   Validate a commit-specific Next.js preview package and manually trigger the
-  test_e2e_deploy_release.yml GitHub Actions workflow from a pull request's
-  internal vercel/next.js branch. Use when asked to run deploy tests, deployment
-  tests, the deploy release E2E workflow, or a Vercel deployment test for a PR.
-  Covers resolving the latest branch SHA, waiting for vercel-packages, preserving
-  default workflow inputs, dispatching the workflow, and verifying the run.
+  entire Next.js deployment test suite through the test_e2e_deploy_release.yml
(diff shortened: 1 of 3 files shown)


.agents/skills/deploy-release-test/SKILL.md
```diff
@@ -0,0 +1,110 @@
+---
+name: deploy-release-test
+description: >
+  Validate a commit-specific Next.js preview package and manually trigger the
+  test_e2e_deploy_release.yml GitHub Actions workflow from a pull request's
+  internal vercel/next.js branch. Use when asked to run deploy tests, deployment
+  tests, the deploy release E2E workflow, or a Vercel deployment test for a PR.
+  Covers resolving the latest branch SHA, waiting for vercel-packages, preserving
+  default workflow inputs, dispatching the workflow, and verifying the run.
```
ztanner: I think maybe we should qualify this skill and make sure the agent know it's only needed if you want to run the entire Next.js deployment test suite. ie I don't want it to prefer this over the guidance [here](https://github.com/vercel/next.js/blob/canary/contributing/core/testing.md#running-deploy-tests-locally) to sanity check if a change works on specific deployment test(s)

timneutkens: :robot: Done — scoped this skill to explicitly cover only the full deployment test suite. Focused or ambiguous deploy-test requests now point to the documented local NEXT_TEST_VERSION + pnpm test-deploy <path> workflow, and I updated the skill metadata and AGENTS.md summary to match.

dianatofficial (commented):

eps1lon (approved):
nice

PR #97565, opened 2026-08-19
Add experimental.agentMode: agent-aware dev workflow surfaces

gaojude:
Adds experimental `agentMode`: when an AI coding agent is detected driving development, `next dev`, `next build`, and the dev server steer it toward structured tooling (the `next experimental-agent-dev` CLI, `/_next/agent`, MCP) instead of raw HTML fetches and ad-hoc process management. Off by default; activates only when the flag is set and an agent is detected.


dianatofficial (commented):

Issue #95447, opened 2026-07-03
Server Actions: multipart/form-data POST to a non-action route returns 500 in production (next start) but 404 in development (next dev)


rgomezcasas:
### Link to the code that reproduces this issue

https://github.com/rgomezcasas/nextjs-server_action_bug

### To Reproduce

1. Minimal App Router app (`create-next-app`, TypeScript) with **at least one registered Server Action** so it appears in the server reference manifest — e.g. a trivial `app/actions.ts` used by a `<form action={...}>` on the home page.
2. Add a catch-all page that always 404s:
   ```tsx
   // app/[...rest]/page.tsx
   import { notFound } from "next/navigation";
   export default function CatchAll() {
     notFound();
   }
3. Send a multipart/form-data POST to any URL that is not a valid Server Action (no Next-Action header, so it is treated as an MPA action):
printf x > /tmp/f.txt
curl -i -X POST http://localhost:3000/customer/address_file/upload -F 'file=@/tmp/f.txt'
4. Run the request against next dev, then against next build && next start.

### Current vs. Expected behavior

The same request and the same internally-thrown error (Failed to find Server Action…) produce different HTTP responses depending on the mode:

- next dev: 404 with header x-nextjs-action-not-found: 1, body Server action not found.                                                             
- next start (production):  500 Internal Server Error — server logs ⨯ Error: Failed to find Server Action. This request might be from an older or newer deployment.

Expected: a malformed/unknown multipart action POST should be handled gracefully (a 404, matching dev) instead of surfacing as an uncaught 500 in production. Control requests (application/json POST and GET) correctly return a clean 404 in both modes.

Root cause is in packages/next/src/server/app-render/action-handler.ts. For a "multipart POST that is not a fetch action", when areAllActionIdsValid(formData, serverModuleMap) === false the code does a bare throw new Error("Failed to find Server Action…"), guarded by its own comment:

▎ // TODO: This can be from skew or manipulated input. We should handle this case more gracefully but this preserves the prior behavior where decodeAction would throw instead.


Expected: a malformed/unknown multipart action POST should be handled gracefully (a 404, matching dev) instead of surfacing as an uncaught 500 in production. Control requests (application/json POST and GET) correctly return a clean 404 in both modes.

[...]

PR #97568, opened 2026-08-19
fix: serve action-not-found 404 for RSC requests instead of throwing

rogersnm:
### What?

An App Router app that defines no server actions returns HTTP 500 for an unauthenticated `POST` request carrying both a `Next-Action` header and an `RSC: 1` header. It should return 404. The server log shows `InvariantError: Invariant: Expected RSC response, got text/plain. This is a bug in Next.js.`

This PR makes the app page runtime serve the action handler's own deliberate 404 instead of asserting on it, and adds e2e coverage for the request shape that was missing from the existing test.

### Why?

The request shape is trivially reachable: one `POST` and two headers, against any path that resolves to an app page, with no authentication. `RSC: 1` alongside `Next-Action` is what the Next.js client router itself sends on every fetch action, so this is not an exotic request, and internet background scanner traffic hits it. Because the failure surfaces as a 500 rather than a 404, it pollutes error rates and can trip production alerting. In our case it fired a load balancer 5xx alarm.

Reproduced on stock minimal apps on `16.2.12`, `16.3.1` (latest stable), and `16.3.1-canary.24` (latest canary). All three return 500, and the relevant source is unchanged on `canary`.

The mechanism is two halves of Next disagreeing with each other:

1. **The action handler correctly decides to 404.** In `packages/next/src/server/app-render/action-handler.ts`, `handleAction` hits the `if (!hasServerActions())` branch and calls `handleUnrecognizedFetchAction`, which does the right thing: it sets `x-nextjs-action-not-found: 1` (`NEXT_ACTION_NOT_FOUND_HEADER`), sets `content-type: text/plain`, sets `res.statusCode = 404`, and returns `RenderResult.fromStatic('Server action not found.', 'text/plain')`. The accompanying "failed to find server action" warning appears in the logs, confirming this branch runs.

2. **The page runtime then refuses to send that response.** The result reaches `packages/next/src/build/templates/app-page-runtime.ts`. Because the request carried `RSC: 1`, `isRSCRequest` is true, so the runtime checks the result's content type against `RSC_CONTENT_TYPE_HEADER`. It is `text/plain`, which is precisely what step 1 just set, so the guard fires. Which side of the guard the request lands on is decided by the unrelated `cacheComponents` config flag: with `cacheComponents: true` the request gets a clean 404, and with it off (the default) the runtime throws the invariant and the 404 becomes a 500.

Instrumenting the throw site on a real build shows the response is already fully formed and correct at the moment of the throw:

```
{ contentType: 'text/plain', resStatus: 404, actionNotFound: '1', cacheComponents: false, isRSCRequest: true, isDraftMode: false }
```

The 500 even goes out on the wire carrying the 404's own headers (`x-nextjs-action-not-found: 1` and `content-type: text/plain`).

### How?

In `packages/next/src/build/templates/app-page-runtime.ts`, import `NEXT_ACTION_NOT_FOUND_HEADER` and treat the action handler's deliberate 404 as a response to serve rather than an unexpected state to assert on:

```ts
if (cachedData.html.contentType !== RSC_CONTENT_TYPE_HEADER) {
  // An unrecognized action id is answered by the action handler
  // with a plain-text 404 rather than a flight payload, so a
  // non-RSC content type here is a deliberate response and not an
  // unexpected state. Serve it rather than asserting on it.
  const isActionNotFound = Boolean(
    res.getHeader(NEXT_ACTION_NOT_FOUND_HEADER)
  )

  if (nextConfig.cacheComponents || isActionNotFound) {
    res.statusCode = 404
    return sendRenderResult({ ..., result: RenderResult.EMPTY, ... })
  }
  throw new InvariantError(...)
}
```

The change is deliberately narrow. The invariant still fires for any genuinely unexpected content type. The only case newly allowed through is the one Next itself produced and labeled with its own header.

### Testing

- With the fix applied to a real `16.3.1-canary.24` build: the repro request returns 404 with `x-nextjs-action-not-found: 1`, a normal `GET /` still returns 200, an RSC `GET` still returns a `text/x-component` flight payload, and zero invariants appear in the log.
- Setting `cacheComponents: true` on the same app and the same request returns 404 with no invariant, confirming the config flag was the only thing deciding the outcome before this change.
- The existing e2e test `test/e2e/app-dir/no-server-actions/no-server-actions.test.ts` already asserts a 404 plus the `x-nextjs-action-not-found` header for this scenario, but it never sends the `RSC: 1` header, which is why the bug was not caught. This PR adds an `it.each` case covering both a malformed action id and a plausible but missing one, this time with `RSC: '1'` set, asserting a 404 and that the CLI output does not contain "Expected RSC response".

### Repro

1. Create a minimal App Router app with `next@16.3.1`: a root layout and a single `app/page.jsx`. Define no server actions anywhere. Use a default (empty) `next.config.js`.
2. Build and start:

   ```bash
   next build && next start -p 3000
   ```

3. Send the request. Next first responds 307 to `/?_rsc`, its cache-busting redirect, so `-L` matters:

   ```bash
   curl -i -L -X POST http://localhost:3000/ \
     -H 'Next-Action: 000000000000000000000000000000000000000000' \
     -H 'RSC: 1'
   ```

4. Observed: `HTTP/1.1 500 Internal Server Error`, and the server log contains `InvariantError: Invariant: Expected RSC response, got text/plain`.
   Expected: `HTTP/1.1 404 Not Found` with `x-nextjs-action-not-found: 1`.

### Note on the related issue

#95447 ("multipart/form-data POST to a non-action route returns 500 in production but 404 in development") is a different bug with a different trigger and a different throw site. It requires the app to have at least one server action, is triggered by a multipart POST with no `Next-Action` header, and throws a bare `Error("Failed to find Server Action...")` from `action-handler.ts` rather than an `InvariantError` from `app-page-runtime.ts`. Fixing one does not fix the other, and this PR does doesn't close it.

One clarifying data point for that issue: its description notes that with zero server actions the `if (!hasServerActions())` early return diverts to `handleUnrecognizedFetchAction` (the graceful 404 path) in both modes. That holds only when the request omits the `RSC` header. With `RSC: 1`, the graceful 404 is produced and then rejected downstream, which is the case this PR fixes.

Related: #95447


dianatofficial (commented):

PR #97569, opened 2026-08-19
fix(turbopack): support targets without globalThis

fireairforce:
## What?

- Make Turbopack's browser chunks, evaluate chunks, chunk lists, and browser runtime work for targets that do not support `globalThis`.
- Keep the generated output unchanged for modern targets.

## Why?

iOS 12.1 and earlier do not support `globalThis`. Turbopack currently uses it unconditionally in chunk registration wrappers, causing a `ReferenceError` before the runtime or application modules can execute.

## How?

- Detect `globalThis` support from the runtime versions resolved from Browserslist.
- Continue using `globalThis` for supported targets and use `self` to register chunks for older targets.
- Install a guarded `self.globalThis = self` alias when the runtime starts for older targets.
- Add unit coverage for the iOS 12.1/12.2 support boundary.

## Tests

- `cargo fmt --all`
- `cargo test -p turbopack-core detects_global_this_support_for_ios_targets`
- `cargo clippy --all-targets -- -D warnings --no-deps`

Related to #83799


dianatofficial (commented):

PR #97570, opened 2026-08-19
[no-unsafe-inline] Add an experimental flag for React's external browser runtime

eps1lon:
React's Fizz streams its instruction set as inline `<script>` tags, which forces a Content-Security-Policy to allow `unsafe-inline` unless a per-response nonce is used, and a per-response nonce is incompatible with static generation. This change adds `experimental.externalBrowserRuntime`, which switches Fizz to React's `unstable_server-external-runtime` so the instructions are emitted as `<template data-rci>` elements and applied in the browser by a standalone runtime script instead. Since React only ships that runtime on its experimental channel, enabling the flag opts the `app` directory into `react@experimental` alongside `taint`, `blockingSSR`, `transitionIndicator`, and `gestureTransition`.

The visible consequence is one additional static asset under `/_next/static/chunks` and one additional `<script>` tag in the document head. Next.js adds neither of them: React emits the tag itself once `unstable_externalRuntimeSrc` is set, marked `async` and carrying the nonce and integrity. In exchange React stops injecting its inline form-replaying runtime, since the external runtime covers it.

This is App Router only; the Pages Router still inlines React's instructions. It is also one step in a larger effort to let applications drop `unsafe-inline` entirely. The Flight payload and `bootstrapScriptContent` are still emitted as inline scripts, and moving them onto the same mechanism will be gated behind this same feature flag.

### DIFF
--- a/crates/next-api/src/app.rs
+++ b/crates/next-api/src/app.rs
@@ -1463,6 +1463,37 @@ impl AppEndpoint {
             None
         };
 
+        // Like the polyfill, this is a pre-compiled asset distributed as part of next and served
+        // as-is rather than bundled. React only publishes it on the experimental channel, which
+        // `experimental.externalBrowserRuntime` opts into.
+        let external_browser_runtime_output_asset =
+            if matches!(this.ty, AppEndpointType::Page { .. })
+                && *project
+                    .next_config()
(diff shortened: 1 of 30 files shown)


dianatofficial (commented):

PR #97573, opened 2026-08-19
[no-unsafe-inline] Carry the Flight payload as data instead of inline scripts

eps1lon:
Next.js inlines the Flight payload as a stream of `<script>self.__next_f.push(...)</script>` tags, which is the second reason an App Router document needs `unsafe-inline` in its Content-Security-Policy. With `experimental.externalBrowserRuntime` each payload is now written as an inert `<template data-next-flight="...">` carrying the same JSON, and `app-index.tsx` collects those templates and feeds them to the same queue. The payload shape is unchanged, so the consumer that decodes bootstrap, data, form state, and binary segments is untouched.

The collector lives in the client bundle rather than in the shared browser runtime asset that the previous change introduced, which means this adds no additional asset and no additional script tag. That placement is deliberate: the asset is loaded with `async`, `async` scripts do not block `DOMContentLoaded`, and `DOMContentLoaded` is when `app-index.tsx` closes the inline Flight stream. Collecting from a separate async script would therefore race that close and silently drop payloads on a slow or cold cache, surfacing as a hydration failure. This module is the only consumer of the queue, so collecting here removes the race entirely.

Collection happens in two parts, mirroring how React's own external runtime works: one sweep over whatever the parser has already produced, then a `MutationObserver` for payloads that arrive as the rest of the document streams in. A final sweep runs when the stream is about to be closed, because a payload the observer has not delivered yet would otherwise be lost.

The two writers had each grown their own copy of the payload encoding, including the base64 fallback for chunks that are not valid UTF-8. Both now share a single `createFlightMarkup` helper that owns both the inline-script and the template form, so the two wire formats cannot drift apart.

### DIFF
--- a/crates/next-api/src/app.rs
+++ b/crates/next-api/src/app.rs
@@ -1463,6 +1463,37 @@ impl AppEndpoint {
             None
         };
 
+        // Like the polyfill, this is a pre-compiled asset distributed as part of next and served
+        // as-is rather than bundled. React only publishes it on the experimental channel, which
+        // `experimental.externalBrowserRuntime` opts into.
+        let external_browser_runtime_output_asset =
+            if matches!(this.ty, AppEndpointType::Page { .. })
+                && *project
+                    .next_config()
+                    .enable_external_browser_runtime()
+                    .await?
+            {
+                let next_package = get_next_package(project.project_path().owned().await?).await?;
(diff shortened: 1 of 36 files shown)


dianatofficial (commented):

PR #97574, opened 2026-08-19
[no-unsafe-inline] Stop inlining the Turbopack chunk group bootstrap

eps1lon:
With React's instruction set and the Flight payload both carried as data, one inline script was left in a production App Router document: the Turbopack chunk-group bootstrap, which seeds `globalThis[TURBOPACK]` with the page's chunk list so the runtime does not have to fetch it. It is emitted whenever `experimental.turbopackSharedRuntime` is on, which is the default for canary builds, so it affected the default bundler.

Rather than move that payload to a template and apply it from the shared browser runtime asset, this change stops inlining it at all. The browser chunking context gains an `inline_chunk_group_bootstrap` option, and when it is disabled the per-route evaluate chunk is emitted as an ordinary asset instead of having its params hoisted into the HTML. That path is not new: it is exactly what happens when the shared runtime is disabled, which is the case in development. Next.js then produces no `bootstrapScriptContent` for the page, because the existing guard already skips it when the manifest carries no params for the route.

The alternative would have kept the inlining and had the runtime asset push the params instead. That asset is loaded with `async`, so the push would frequently land after the runtime chunk had already resolved the page's chunks, which is the very round trip the inlining exists to avoid. Trading it for one extra request per route is simpler, reuses a well-exercised code path, and does not depend on load order.

The option defaults to inlining, so nothing changes unless `experimental.externalBrowserRuntime` is enabled.

With this the flag delivers what it was named for: a production document contains no inline `<script>` at all, which is now asserted in the test suite. Development is excluded from that assertion, since it still inlines the request id that the dev overlay and HMR socket read, and is not a target for CSP hardening.

### DIFF
--- a/crates/next-api/src/project.rs
+++ b/crates/next-api/src/project.rs
@@ -1643,6 +1643,7 @@ impl Project {
                 .next_config()
                 .turbo_nested_async_chunking(self.next_mode(), true),
             shared_runtime: self.next_config().turbo_shared_runtime(self.next_mode()),
+            inline_chunk_group_bootstrap: self.next_config().turbo_inline_chunk_group_bootstrap(),
             per_page_module_graph: self.per_page_module_graph(),
             debug_ids: self.next_config().turbopack_debug_ids(),
             worker_asset_prefix: self.next_config().turbopack_worker_asset_prefix(),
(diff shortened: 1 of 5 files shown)


dianatofficial (commented):

PR #97576, opened 2026-08-19
fix(next-napi-bindings): detect the target, not the host, in build.rs

sokra:
### What?

`crates/next-napi-bindings/build.rs` now decides whether to emit
`-Wl,--warn-unresolved-symbols` from `CARGO_CFG_TARGET_OS` / `CARGO_CFG_TARGET_FAMILY` instead of
from a `#[cfg(...)]` attribute.

### Why?

`#[cfg(...)]` inside a build script is evaluated for the machine that **runs** the script — the host
— not for the target being compiled. The existing `not(target_arch = "wasm32")` guard was therefore
always true on a Linux host, so cross-compiling to wasm still passed the flag and the link died with:

```
rust-lld: error: unknown argument: -Wl,--warn-unresolved-symbols
```

This is a latent bug for any cross-compilation target, not only wasm.

### How?

Cargo sets `CARGO_CFG_*` for the *target*, so the condition is read from the environment instead of
from a compile-time attribute. Behaviour on native Linux is unchanged.

### DIFF
--- a/crates/next-napi-bindings/build.rs
+++ b/crates/next-napi-bindings/build.rs
@@ -5,7 +5,13 @@ use serde_json::Value;
 fn main() -> anyhow::Result<()> {
     println!("cargo:rerun-if-env-changed=CI");
     println!("cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS");
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_TARGET_FAMILY");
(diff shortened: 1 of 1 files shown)


crates/next-napi-bindings/build.rs
lukesandberg, line 12: lets add a `is_linux_target` here

and does is_macos_target need a !wasm condition as well?

lukesandberg (changes requested):

dianatofficial (commented):

lukesandberg (approved):

PR #97577, opened 2026-08-19
fix(turbo-rcstr): make TaggedValue usable on wasm

sokra:
### What?

Makes `turbo-rcstr`'s tagged value work on wasm — and on any target whose pointer is narrower than the
value — **without changing the representation or the inline capacity on any target**. `RcStr` stays 8
bytes wide everywhere, `MAX_INLINE_LEN` stays 7 everywhere, and `rcstr!` stays a `const`.

### Why?

`TaggedValue` stored a `NonNull<()>` on 64-bit but a bare `NonZeroU64` when
`target_pointer_width = "32"`. `rcstr!` expands to a `const`, so both constructors must be usable
during const evaluation — and with integer storage `new_ptr` needs a pointer→integer cast, which
const-eval forbids. Every `rcstr!` taking the static path failed to compile:

```
error[E0080]: unable to turn pointer into integer
```

Concretely, `cargo check -p turbopack-core --target wasm32-wasip1-threads` reported **4** of these
before this change and **0** after.

### How?

The fix rests on an asymmetry in const evaluation:

| direction | const-eval | why |
|---|---|---|
| pointer → integer | **forbidden** | a pointer is an abstract (allocation, offset) pair; its numeric address doesn't exist until the linker assigns one |
| integer → pointer | allowed | the result carries no provenance, which is fine as long as it is never dereferenced |

64-bit already only needs the legal directions: `new_ptr` is pointer→pointer, `new_tag` is
integer→pointer. So rather than narrow the value on other targets, they now store a struct that keeps
a **real pointer field**:

```rust
#[repr(C, align(8))]
struct RawLittle { ptr: NonNull<()>, pad: [u8; PAD] }
```

`new_ptr` stores the pointer as a pointer; `new_tag` transmutes an integer into the struct. Reading the
value back as an integer (`get_ptr`, `get_value`, `tag_byte`) only happens at **run time**, where
pointer→integer is perfectly legal.

Three details are load-bearing, and each has a static assertion:

- **`align(8)`** — a pointer field alone would align the struct to 4 on wasm32, and reading the value
  as a `u64` then traps with `RuntimeError: operation does not support unaligned accesses`.
- **No padding** — the byte payload is sized to the pointer width, because padding bytes are
  uninitialised and would make the whole-value transmute in `new_tag` invalid.
- **Endian-dependent field order** — the pointer's least-significant byte must land where the value's
  least-significant byte lives (offset 0 little-endian, last byte big-endian), since that is where the
  tag is read from. Both layouts are declared unconditionally so the big-endian assertion still
  compiles (and was verified to fail when deliberately broken) on a little-endian host.

A union satisfies both constructors too, but unions carry no niche in rustc, so `Option<RcStr>` would
grow to 16 bytes. The struct keeps it at 8 — also asserted.

### Testing

- `turbo-rcstr`: **11 passed** on host, **10 passed / 1 ignored** on `wasm32-wasip1-threads`
  (the ignored one needs the static-`RcStr` registry, which depends on the link-section loader hook).
- Round-trips every `rcstr!` length from 0 to 9 on both targets, covering both sides of the inline
  boundary and the static path.
- Each of the new static assertions was verified to **fail** when deliberately broken, including the
  big-endian layout one.
- `MAX_INLINE_LEN == 7` asserted on host and wasm.
- The five `const … = rcstr!(…)` sites in the tree — including two `pub const`s in `turbopack-core` —
  are untouched and still compile as `const`.
- `atom_size_128` still builds; its pre-existing const limitation is unchanged and out of scope.
- `cargo fmt --check`, and clippy `-D warnings` for the host and for `wasm32-wasip1-threads`.

Big-endian and 16-bit are compile-checked and assertion-protected, but not runtime-exercised — no such
target is available here.

### DIFF
--- a/turbopack/crates/turbo-rcstr/src/tagged_value.rs
+++ b/turbopack/crates/turbo-rcstr/src/tagged_value.rs
@@ -1,14 +1,45 @@
-#![allow(clippy::missing_transmute_annotations)]
+//! The tagged value that backs [`crate::RcStr`].
+//!
+//! An `RcStr` is a single 8-byte value (16 with `atom_size_128`) that packs a 2-bit tag with a
+//! payload: a pointer to a static or `Arc`'d string, or up to [`MAX_INLINE_LEN`] bytes of string
+//! data stored inline. The tag lives in the low 2 bits of the value's least significant byte,
+//! which is why pointers must be at least 4-byte aligned.
+//!
+//! # Why the narrow-pointer targets use a struct rather than an integer
+//!
+//! `rcstr!` expands to a `const`, so both constructors below have to be usable during const
+//! evaluation. Const evaluation is asymmetric about pointer/integer conversions:
+//!
+//! * **pointer → integer is forbidden.** During const evaluation a pointer is an abstract
+//!   (allocation, offset) pair; the numeric address does not exist yet because the linker chooses
+//!   it. No spelling avoids this — `transmute`, `as`, `expose_provenance` and `repr(C)` union
+//!   punning are all rejected.
+//! * **integer → pointer is allowed.** The result simply carries no provenance, which is fine so
+//!   long as it is never dereferenced.
+//!
+//! On 64-bit targets the payload is a `NonNull`, so [`TaggedValue::new_ptr`] is a pointer→pointer
+//! cast and [`TaggedValue::new_tag`] is an integer→pointer transmute: both permitted. If narrower
+//! targets stored a bare `u64` instead, `new_ptr` would need the one conversion that is impossible
+//! and every `rcstr!` taking the static path would fail to compile with
+//! `error[E0080]: unable to turn pointer into integer`.
+//!
+//! So those targets keep the value 8 bytes wide — giving the same [`MAX_INLINE_LEN`] as
+//! everywhere else — but hold the address in a real pointer field paired with a byte payload. The
+//! payload is sized to the pointer width so there is no padding: padding bytes would be
+//! uninitialised, which would make the whole-value transmute in `new_tag` invalid.
+//!
+//! The field order follows endianness so that the pointer's least significant byte always lands
+//! where the value's least significant byte lives — offset 0 on little-endian, the last byte on
+//! big-endian. That is where [`TaggedValue::tag_byte`] reads the tag. With a fixed field
+//! order, a big-endian dynamic pointer's tag bits would be read out of the payload (always zero)
+//! and misread as `STATIC_TAG`.
+//!
+//! Reading the value back as an integer (`get_ptr`, `get_value`, `tag_byte`) only ever happens at
+//! run time, where pointer → integer is perfectly legal.
 
 use std::{num::NonZeroU8, os::raw::c_void, ptr::NonNull, slice};
 
 use self::raw_types::*;
-#[cfg(not(any(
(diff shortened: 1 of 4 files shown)


turbopack/crates/turbo-rcstr/src/tagged_value.rs
```diff
@@ -3,8 +3,20 @@
 use std::{num::NonZeroU8, os::raw::c_void, ptr::NonNull, slice};
 
 use self::raw_types::*;
+// On 32-bit targets the tagged value is a `u64`, which means `new_ptr` has to cast the pointer
+// to an integer. That is not permitted in a `const` context, and `rcstr!` expands to a
+// `const`, so every `rcstr!` with a non-inline string would fail to compile with
+// `error[E0080]: unable to turn pointer into integer`.
+//
+// wasm is 32-bit but has no use for the wider tag space, so give it the same pointer-based
+// representation 64-bit targets use. `MAX_INLINE_LEN` drops from 7 to 3 there, so slightly
```
lukesandberg: 'slightly' lol

lukesandberg: i find this comment confusing

if we are picking a u64 representatIon why does max_inlihne_len drop...?

obviosuly if it is a 32bit repr we would have that problem

dianatofficial (commented):

sokra:
Does that work? Is that tested, fleet?

sokra:
I'm a bit confused about this approach. On 32bit platforms, it should use `NonZeroU64`. That means it uses 64bits to represent rcstr on all platforms. This means the inline len is always 7 chars. The representation is the same on all platforms.
Instead figure out how to solve the const constructor on 32bit platforms instead of changing the reprensentation.

sokra:
@lukesandberg Take a look again. This seem to be the right approach now

turbopack/crates/turbo-rcstr/src/tagged_value.rs
```diff
@@ -113,7 +246,26 @@ impl TaggedValue {
 
     #[inline(always)]
     fn get_value(&self) -> RawTaggedValue {
-        unsafe { std::mem::transmute(Some(self.value)) }
+        #[cfg(all(
+            any(target_pointer_width = "32", target_pointer_width = "16"),
```
lukesandberg, line 250: we should drop all the target_pointer_width = 16 stuff?

just add a `compile_error!` in that case we don't need to support it

turbopack/crates/turbo-rcstr/src/tagged_value.rs
```diff
@@ -17,23 +48,85 @@ mod raw_types {
...
+/// The narrow-pointer arm: an 8-byte value built from a real pointer plus a byte payload, so that
+/// both constructors stay const-evaluable. See the module docs.
 #[cfg(all(
-    any(
-        target_pointer_width = "32",
-        target_pointer_width = "16",
-        feature = "atom_size_64"
-    ),
+    any(target_pointer_width = "32", target_pointer_width = "16"),
     not(feature = "atom_size_128")
```
lukesandberg, line 55: is this condition required still? do we need the atom_size_64?

lukesandberg (approved):

PR #97578, opened 2026-08-19
fix(turbo-rcstr): allow the napi feature on wasm targets

sokra:
### What?

Removes `turbo-rcstr`'s hard stop against combining the `napi` feature with wasm targets, and moves
the `napi` dependency out of the non-wasm dependency table.

### Why?

The crate contained:

```rust
#[cfg(all(feature = "napi", target_family = "wasm"))]
compile_error!("The napi feature cannot be enabled for wasm targets");
```

This predates napi v3, which supports `wasm32-wasip1-threads`. With the assertion removed,
`turbo-rcstr/napi` and everything built on it — `turbopack-node/worker_pool`,
`next-api/worker_pool`, `next-napi-bindings` — compiles for wasm.

### How?

Delete the `compile_error!`, drop the now-redundant `cfg` on the `napi_impl` module, and declare
`napi` as a normal optional dependency rather than a non-wasm-only one.

### DIFF
--- a/turbopack/crates/turbo-rcstr/Cargo.toml
+++ b/turbopack/crates/turbo-rcstr/Cargo.toml
@@ -23,8 +23,9 @@ turbo-bincode = { workspace = true }
 turbo-rcstr-macros = { path = "../turbo-rcstr-macros" }
 unty = { workspace = true }
 
-[target.'cfg(not(target_family = "wasm"))'.dependencies]
 napi = { workspace = true, optional = true }
+
+[target.'cfg(not(target_family = "wasm"))'.dependencies]
 scattered-collect = { workspace = true }
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


lukesandberg (approved):

dianatofficial (commented):

turbopack/crates/turbo-rcstr/Cargo.toml
```diff
@@ -23,8 +23,10 @@ turbo-bincode = { workspace = true }
 turbo-rcstr-macros = { path = "../turbo-rcstr-macros" }
 unty = { workspace = true }
 
-[target.'cfg(not(target_family = "wasm"))'.dependencies]
+# napi v3 supports `wasm32-wasip1-threads`, so this is not restricted to native targets.
```
lukesandberg: drop comment

lukesandberg (approved):

PR #97580, opened 2026-08-19
fix(turbo-tasks-fs): create symlinks through the WASI API on wasi

sokra:
### What?

`turbo-tasks-fs` creates symlinks through `std::os::wasi::fs::symlink_path` when targeting wasi.

### Why?

`std::os::unix` does not exist on wasi, so `disk.rs`'s symlink creation failed to compile:

```
error[E0433]: cannot find `unix` in `os`
    --> turbopack/crates/turbo-tasks-fs/src/disk.rs:1375
```

### How?

A `cfg(target_os = "wasi")` arm calls the WASI equivalent, which takes the same `(target, link)`
argument order. That API is still behind the unstable `wasi_ext` feature, enabled for the wasi target
only via `cfg_attr`. The Windows junction-point path and the unix path are unchanged.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/disk.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/disk.rs
@@ -1371,8 +1371,10 @@ impl FileSystem for DiskFileSystem {
                                 })?;
                                 has_old_content = false;
                             }
-                            #[cfg(not(windows))]
+                            #[cfg(all(not(windows), not(target_os = "wasi")))]
                             let io_result = std::os::unix::fs::symlink(&target, &**full_path);
+                            #[cfg(target_os = "wasi")]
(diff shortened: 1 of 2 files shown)


dianatofficial (commented):

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -1371,8 +1371,12 @@ impl FileSystem for DiskFileSystem {
                                 })?;
                                 has_old_content = false;
                             }
-                            #[cfg(not(windows))]
+                            // `std::os::unix` does not exist on wasi; the equivalent lives in
+                            // `std::os::wasi::fs` and takes the same (target, link) argument order.
```
lukesandberg: drop comment, this is obvious from the code

lukesandberg (approved):

PR #97581, opened 2026-08-19
refactor(turbopack-cli-utils): replace crossterm with owo-colors

sokra:
### What?

`turbopack-cli-utils` uses `owo-colors` instead of `crossterm` for issue formatting.

### Why?

Two reasons:

- `crossterm` 0.26 has no wasm backend, and fails to compile for wasm with 8 errors inside the crate
  itself.
- It was pulling its weight for very little: the only uses were `StyledContent` and `Stylize` in
  `issue.rs`, and `owo-colors` — already a dependency of that same file — covers all of them.

### How?

Swap the trait usage over and return `String` from `show_all_message` rather than
`StyledContent<String>`. Rendered output is unchanged.

This also removes `crossterm`, `crossterm_winapi`, `mio 0.8`, `signal-hook` and `signal-hook-mio`
from `Cargo.lock` (and lets cargo drop the `mio 1.2` disambiguation, since only one `mio` remains).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1928,31 +1928,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "22ec99545bb0ed0ea7bb9b8e1e9122ea386ff8a48c0922e43f36d45ab09e0e80"
 
 [[package]]
-name = "crossterm"
-version = "0.26.1"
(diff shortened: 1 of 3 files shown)


dianatofficial (commented):

mischnic (approved):

PR #97583, opened 2026-08-19
fix(turbopack): make the SWC wasm-plugin backend native-only

sokra:
### What?

Makes the SWC wasm-plugin backend native-only, and reports an **error** against `next.config` when
`experimental.swcPlugins` is configured on a platform that cannot execute them.

### Why?

`swc_plugin_backend_wasmtime` cannot be hosted inside wasm: there is no JIT, and `wasi-common` needs
native filesystem APIs. SWC's own wasm32 path skips plugin transforms for the same reason, see
swc-project/swc#3934.

Silently dropping the transform is the worst outcome — the build succeeds and quietly produces
different output than the config asked for. A warning is not much better: it still emits a bundle whose
code was not transformed as configured, which is subtly wrong rather than obviously broken. So this
reports an error instead.

### How?

- `swc_plugin_backend_wasmtime` moves to a non-wasm dependency, and the plugin rule's implementation is
  split into native and wasm arms.
- The wasm arm emits `UnsupportedSwcPluginsIssue` — `IssueStage::Unsupported`, `IssueSeverity::Error`
  — naming the configured plugins and pointing at the resolved `next.config` path (via
  `NextConfig::config_file_path`, as the babel and sass issues do), so the message points at the file
  the user has to edit.

It fires once at config time and only when plugins are actually configured, so projects that don't use
them are unaffected. Native behaviour is unchanged.

### DIFF
--- a/crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
+++ b/crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
@@ -1,36 +1,53 @@
 use anyhow::Result;
-use bincode::{Decode, Encode};
-use serde::{Deserialize, Serialize};
 use turbo_rcstr::RcStr;
-use turbo_tasks::{ResolvedVc, Vc, trace::TraceRawVcs};
+#[cfg(target_family = "wasm")]
+use turbo_rcstr::rcstr;
+#[cfg(not(target_family = "wasm"))]
+use turbo_tasks::ResolvedVc;
+use turbo_tasks::Vc;
 use turbo_tasks_fs::FileSystemPath;
 use turbopack::module_options::ModuleRule;
+#[cfg(not(target_family = "wasm"))]
 use turbopack_ecmascript::{CustomTransformer, TransformPlugin};
 
 use crate::next_config::NextConfig;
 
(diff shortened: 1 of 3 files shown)


crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
```diff
@@ -45,6 +60,19 @@ pub async fn get_swc_ecma_transform_plugin_rule(
...
+/// SWC wasm plugins are not available when Next.js itself is built for wasm: executing them needs
+/// the wasmtime backend, which cannot be hosted inside wasm. SWC skips plugin transforms on wasm32
+/// for the same reason, see <https://github.com/swc-project/swc/issues/3934>.
+#[cfg(target_family = "wasm")]
+pub async fn get_swc_ecma_transform_rule_impl(
+    _project_path: FileSystemPath,
+    _plugin_configs: &[(RcStr, serde_json::Value)],
+    _enable_mdx_rs: bool,
+) -> Result<Option<ModuleRule>> {
+    Ok(None)
```
lukesandberg, line 104: can we report a warning against the next.config? maybe an error even

lukesandberg (approved):

dianatofficial (commented):

PR #97584, opened 2026-08-19
fix(turbopack-node): make process_pool inert on wasm

sokra:
### What?

`turbopack-node`'s `process_pool` feature is inert on wasm, leaving `worker_pool` as the only Node
backend there.

### Why?

The child-process pool needs `tokio::process` and a TCP listener, neither of which exists on wasi:

```
error[E0432]: unresolved import `tokio::process`   # gated #[cfg(not(target_os = "wasi"))] in tokio
error[E0599]: no `TcpListener::bind` on wasi
  --> turbopack/crates/turbopack-node/src/process_pool/mod.rs:315
```

Turning the feature off from the outside is not possible: `process_pool` is a **default** feature of
both `turbopack-node` *and* `next-core`, so `--no-default-features` at the top level does not
suppress it.

`worker_pool` — Node worker threads over napi — is already a first-class alternative selected by
`TurbopackPluginRuntimeStrategy`, so no new mechanism is needed.

### How?

Gate the seven `process_pool` sites on `not(target_family = "wasm")`: the module, the sealed backend
impl, the constructor, the config enum variant, the default-strategy selection, and the `next-api`
import and match arm. Host feature semantics are unchanged.

### DIFF
--- a/crates/next-api/src/project.rs
+++ b/crates/next-api/src/project.rs
@@ -88,7 +88,7 @@ use turbopack_core::{
         VersionState, VersionedContent,
     },
 };
-#[cfg(feature = "process_pool")]
+#[cfg(all(feature = "process_pool", not(target_family = "wasm")))]
 use turbopack_node::child_process_backend;
 use turbopack_node::execution_context::ExecutionContext;
 #[cfg(feature = "worker_pool")]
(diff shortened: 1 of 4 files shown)


lukesandberg (approved):

dianatofficial (commented):

PR #97585, opened 2026-08-19
feat(turbo-tasks-fetch): stub HTTP on wasm targets

sokra:
### What?

Reports HTTP as unsupported on wasm, in the shape of the `reqwest` API `turbo-tasks-fetch` consumes,
so the crate compiles for wasm targets.

### Why?

`reqwest` cannot serve wasm here. Its only wasm backend targets `wasm32-unknown-unknown` and is built
on the browser `fetch` API via `wasm-bindgen`. Under `wasm32-wasip1-threads` that backend is still
selected (it keys off `target_arch = "wasm32"`), and it is both API-incompatible (no
`ClientBuilder::connect_timeout`, no `ClientBuilder::timeout`, no `Error::is_connect`) and — fatally —
**`!Send`**, which `turbo-tasks` requires of every task future. WASI preview1 has no sockets to build a
native client on either.

### How?

This is deliberately **not** a client and does not pretend to be one: building a client fails
immediately with a clear error, so no request is ever attempted, nothing is retried, and no response is
ever produced. The remaining types exist only so the shared fetch code type-checks, and the paths that
could never be reached say so rather than returning plausible dummy values.

The error names the consequence and the alternatives, rather than surfacing as a generic failure:

> HTTP requests are not supported in wasm builds of Next.js: this platform has no HTTP client.
> Features that fetch at build time, such as `next/font/google`, cannot be used here — self-host the
> assets, or use a platform with native Next.js binaries.

### Follow-up

Real support needs the **host** to provide HTTP: for wasm builds of the Next.js bindings that means a
napi callback into the JS `fetch`, kept behind the existing `FetchClientConfig` interface so callers are
unaffected. That is a separate PR by design — a `ThreadsafeFunction` call from a wasi thread cannot be
validated until the wasm bindings can be instantiated, which is blocked on the unreleased emnapi v2.

### DIFF
--- a/turbopack/crates/turbo-tasks-fetch/src/wasm_reqwest.rs
+++ b/turbopack/crates/turbo-tasks-fetch/src/wasm_reqwest.rs
@@ -1,0 +1,183 @@
+//! Reports HTTP as unsupported on wasm, in the shape of the `reqwest` API this crate consumes.
+//!
+//! This is **not** a client and does not try to look like one: building a client fails immediately
+//! with a clear error, so no request is ever attempted, nothing is retried, and no response is ever
+//! produced. The remaining types exist only so that [`crate::client`] type-checks; they are
+//! unreachable.
+//!
+//! `reqwest` itself cannot serve wasm here. Its only wasm backend targets `wasm32-unknown-unknown`
+//! and is built on the browser `fetch` API via `wasm-bindgen`. Under `wasm32-wasip1-threads` that
+//! backend is still selected (it keys off `target_arch = "wasm32"`), and it is both
+//! API-incompatible (no `ClientBuilder::connect_timeout`, no `ClientBuilder::timeout`, no
+//! `Error::is_connect`) and — fatally — `!Send`, which `turbo-tasks` requires of every task future.
+//! WASI preview1 has no sockets to build a native client on either.
+//!
+//! Supporting HTTP on wasm needs the host to provide it: for wasm builds of the Next.js bindings
+//! that means a napi callback into the JS `fetch`, kept behind the existing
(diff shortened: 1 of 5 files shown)


turbopack/crates/turbo-tasks-fetch/src/wasm_reqwest.rs
```diff
@@ -0,0 +1,170 @@
...
+//! provide HTTP on wasm.
+//!
+//! `reqwest` cannot be used on wasm here. Its only wasm backend targets
+//! `wasm32-unknown-unknown` and is built on the browser `fetch` API via `wasm-bindgen`. Under
+//! `wasm32-wasip1-threads` that backend is still selected (it keys off `target_arch = "wasm32"`),
+//! and it is both API-incompatible (no `ClientBuilder::connect_timeout`, no
+//! `ClientBuilder::timeout`, no `Error::is_connect`) and — fatally — `!Send`, which `turbo-tasks`
+//! requires of every task future. WASI preview1 has no sockets to build a native client on either.
+//!
+//! TODO: replace this with a real HTTP client provided by the host — for wasm builds of the
```
lukesandberg: This would be a blocker for full support

how complex would it be to provide a host binding now?

dianatofficial (commented):

turbopack/crates/turbo-tasks-fetch/src/wasm_reqwest.rs
sokra: Implement full support for it. Make it a PR after the testing lands.
Update this PR to make it a proper error

lukesandberg (approved):

PR #97587, opened 2026-08-19
build: add an opt-in patch for swc's wasm32 plugin host

sokra:
### What?

Adds an opt-in mechanism for patching a crates.io dependency — `patches/rust/`, a
`scripts/patch-rust-crate.mjs` helper and a README — plus the first patch, for `swc` 74.0.0.

### Why?

`swc` 74.0.0's `cfg(all(feature = "plugin", target_arch = "wasm32"))` path has apparently never been
compiled by anyone. It has two errors:

- `RustPlugins::apply_inner` returns `n` where the signature is `Result<Program, anyhow::Error>`
- a variable is unused in that arm, and the crate sets `#![deny(unused)]`

Anything enabling `swc_core`'s `__plugin_transform_host` — `turbopack-ecmascript-plugins`, and
therefore `next-core` — consequently cannot build for wasm at all.

Cargo has no support for patch files, and a `[patch.crates-io]` entry in the root `Cargo.toml` would
have to name a directory that does not exist in a fresh checkout, which would break **every** build.

### How?

The patch stays a file. `scripts/patch-rust-crate.mjs` copies the crate out of the cargo registry
into `target/patched-crates/` and applies it; only the builds that need it pass
`--config 'patch.crates-io.swc.path="<dir>"'`. Builds that do not pass the flag resolve the crate
from the registry exactly as before, so nothing else is affected.

Two details worth knowing:

- The script applies the patch **from the repository root with `--directory`**. Run from inside
  `target/`, `git apply` resolves paths against the enclosing repository, prints `Skipped patch` and
  still exits 0 — silently leaving the crate unpatched. The script also asserts the patch applied so
  that cannot pass unnoticed.
- Patch files are exempted from whitespace checks in `.gitattributes`: a unified diff's context line
  for a blank line is a single space, so whitespace tooling flags them and "fixing" them would
  corrupt the patch. The repository's existing pnpm patches have the same property.

The patch is intended to be temporary — delete it once the fixes are released upstream. The generic
script can stay for future patches.

### DIFF
--- a/patches/rust/swc@74.0.0.patch
+++ b/patches/rust/swc@74.0.0.patch
@@ -1,0 +1,33 @@
+diff --git a/src/config/mod.rs b/src/config/mod.rs
+index 3a0a6d6..31ab6cb 100644
+--- a/src/config/mod.rs
++++ b/src/config/mod.rs
+@@ -806,11 +806,15 @@ impl Options {
+ 
+         #[cfg(feature = "plugin")]
+         let plugin_transforms: Box<dyn Pass> = {
++            // Only the embedded (non-wasm32) runtime consumes the metadata context; the wasm32
++            // path below skips plugin transforms entirely.
++            #[cfg(not(target_arch = "wasm32"))]
+             let transform_filename = match base {
+                 FileName::Real(path) => path.as_os_str().to_str().map(String::from),
+                 FileName::Custom(filename) => Some(filename.to_owned()),
+                 _ => None,
+             };
++            #[cfg(not(target_arch = "wasm32"))]
+             let transform_metadata_context = Arc::new(TransformPluginMetadataContext::new(
+                 transform_filename,
+                 self.env_name.to_owned(),
+diff --git a/src/plugin.rs b/src/plugin.rs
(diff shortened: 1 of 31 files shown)


patches/rust/swc@74.0.0.patch
```diff
@@ -0,0 +1,33 @@
+diff --git a/src/config/mod.rs b/src/config/mod.rs
+index 3a0a6d6..31ab6cb 100644
+--- a/src/config/mod.rs
++++ b/src/config/mod.rs
+@@ -806,11 +806,15 @@ impl Options {
+ 
+         #[cfg(feature = "plugin")]
+         let plugin_transforms: Box<dyn Pass> = {
++            // Only the embedded (non-wasm32) runtime consumes the metadata context; the wasm32
```
lukesandberg, line 9: we should just upstream this,  patch is fine to help iterate on the stack, but we shouldn't commit this

lukesandberg (changes requested):

dianatofficial (commented):

sokra:
We no longer need that, do we? If not needed we can remove this PR

PR #97588, opened 2026-08-19
ci: check that Turbopack compiles for wasm

sokra:
### What?

Adds a CI job that checks Turbopack compiles for `wasm32-wasip1-threads`, so wasm portability
regressions are caught rather than rediscovered.

### Why?

Everything under `#[cfg(target_family = "wasm")]` is invisible to host builds **and** to host clippy —
it is only checked when you deliberately build for the target. Four separate defects in this stack were
caught only that way (a wrong build-script condition, two bad imports, and a value that compiled but was
wrong). Without a gate, the next one lands unnoticed.

This also re-enables coverage that had been off for ~2 years: the old `test-next-napi-bindings-wasi` job
was disabled with `if: false` pending napi-rs/napi-rs#2009, which closed in April 2024.

### How?

Modelled on `rust-check` via `build_reusable.yml` (`needsRust`, `skipInstallBuild`, `skipNativeBuild`),
so it does not pay for a JS build. Beyond `rustup target add` it needs two things:

- **a WASI C toolchain**, because `lzzzz` (LZ4, via `turbo-persistence`) and `zstd-sys` have C build
  scripts. The SDK build is selected from `$RUNNER_ARCH` — `build_reusable.yml` defaults to an arm64
  runner, and an x86_64 clang fails there with `Exec format error` — with a pinned sha256 per arch,
  unpacked under `$RUNNER_TEMP` so the workspace stays clean.
- **emnapi**, because `next-napi-bindings`' build script calls `napi_build::setup()`, whose wasi path
  panics without `EMNAPI_LINK_DIR` — so it is required even for `cargo check`. It is installed into a
  scratch directory rather than the root `package.json`, because this job runs with `skipInstallBuild`
  and therefore never runs `pnpm install`.

Two things worth recording, both of which cost a CI round trip to find:

- pnpm is invoked from the repo root with `--dir`, not by `cd`-ing into the scratch directory: corepack
  resolves the pnpm version from the nearest `package.json`, and outside the repo it picks the latest
  pnpm (11.x), which cannot run on the pinned Node 20
  (`ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING`).
- the job is added to `tests-pass`, whose `needs:` list is what actually blocks a PR — a job that runs
  but is absent from that list looks like coverage while blocking nothing.

The `emnapi@2.0.0-alpha.4` pin is deliberate and is the one fragility here: the archive must define
`emnapi_create_env` / `emnapi_delete_env`, which exists only in emnapi v2, still a prerelease. Move to
the stable release once it ships.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -621,26 +621,25 @@ jobs:
       stepName: 'test-next-swc-wasm'
     secrets: inherit
 
-  #[NOTE] currently this only checks building wasi target
   test-next-napi-bindings-wasi:
     name: test next-swc wasi
     needs: ['optimize-ci', 'changes', 'build-next']
-    # TODO: Re-enable this when https://github.com/napi-rs/napi-rs/issues/2009 is addressed.
-    # Specifically, the `platform` value is now `threads` in
-    # https://github.com/napi-rs/napi-rs/blob/e4ad4767efaf093fdff3dc768856f6100a6e3f72/cli/src/api/build.ts#L530
-    if: false
-    # if: ${{ needs.changes.outputs.docs-only == 'false' }}
+    if: ${{ needs.optimize-ci.outputs.skip == 'false' && needs.changes.outputs.docs-only == 'false' }}
 
     permissions:
       contents: read
       id-token: write
     uses: ./.github/workflows/build_reusable.yml
     with:
+      needsRust: 'yes'
+      skipInstallBuild: 'yes'
       skipNativeBuild: 'yes'
       skipNativeInstall: 'yes'
       afterBuild: |
-        rustup target add wasm32-wasip1-threads
-        pnpm dlx turbo@${TURBO_VERSION} run build-native-wasi ${TURBO_ARGS}
+        set -euo pipefail
+
+        source scripts/setup-wasi-env.sh
+        cargo check -p next-napi-bindings --target wasm32-wasip1-threads
       stepName: 'test-next-napi-bindings-wasi'
     secrets: inherit
 
@@ -1209,6 +1208,7 @@ jobs:
       - rust-check
       - rustdoc-check
       - test-next-swc-wasm
+      - test-next-napi-bindings-wasi
       - test-turbopack-dev
       - test-new-tests-dev
       - test-new-tests-start
--- a/contributing/core/building-wasm.md
+++ b/contributing/core/building-wasm.md
@@ -1,0 +1,102 @@
+# Building Turbopack for WebAssembly
+
+Turbopack's napi bindings (`crates/next-napi-bindings`) can be compiled for
+`wasm32-wasip1-threads`. This is the target that would let Turbopack run where no native binding
+exists — unusual CPU architectures and operating systems we do not publish artifacts for.
+
+> This is distinct from the `@next/swc-wasm-*` packages, which are built with `wasm-pack` from
+> `crates/wasm` for `wasm32-unknown-unknown` and contain SWC only, not Turbopack.
+
+## Prerequisites
+
+Beyond the usual Rust toolchain, the target needs a WASI clang and sysroot (several dependencies have
+C build scripts) and emnapi (napi's build script links against it). `scripts/setup-wasi-env.sh`
+provisions both.
+
+## Building on Linux
+
+The setup script supports Linux systems directly. Source it, then build as usual:
+
+```sh
+source scripts/setup-wasi-env.sh
+cargo check -p next-napi-bindings --target wasm32-wasip1-threads
+```
+
+It must be sourced from **Bash** because it uses `BASH_SOURCE`; alternative shells such as zsh should
+invoke Bash explicitly:
+
+```sh
+bash -c 'source scripts/setup-wasi-env.sh && cargo check -p next-napi-bindings --target wasm32-wasip1-threads'
+```
+
+The script exports environment variables into the calling shell, which a subprocess cannot do.
+Running it directly prints an error and does nothing.
+
+The script downloads the wasi-sdk matching your host architecture, verifies it against a pinned
+sha256, installs emnapi, and exports the cross-compilation variables (`WASI_SDK_PATH`,
+`EMNAPI_LINK_DIR`, and the `*_wasm32_wasip1_threads` compiler variables). Downloads are cached under
+`~/.cache/next-wasi-toolchain`, so re-sourcing in a new shell is fast. Set `WASI_SETUP_CACHE_DIR` to
+move the cache.
+
+CI sources the same script, so local and CI builds cannot drift apart.
+
+## Building with Docker
+
+On macOS, Windows, or any other Docker-capable host, use the Linux builder image:
+
+```sh
+docker build -t next-wasi-builder -f scripts/wasi-builder.Dockerfile .
+docker run --rm -it \
+  -v "$PWD:/workspace" \
+  -v next-wasi-cache:/root/.cache/next-wasi-toolchain \
+  -w /workspace \
+  next-wasi-builder
+```
+
+Inside the container, use the same commands as CI:
+
+```sh
+source scripts/setup-wasi-env.sh
+cargo check -p next-napi-bindings --target wasm32-wasip1-threads
+```
+
+The named volume preserves the wasi-sdk and emnapi downloads between runs. Mount a second volume at
+`/workspace/target` if you also want to preserve Rust build artifacts without writing them to the host
+checkout.
+
+## Running tests
+
+Wasm test binaries cannot be run directly: `turbo-tasks` gathers its task registries at link time and
+needs the embedder to supply an `env.read_custom_section` import. `scripts/wasi-test-host/` is a Node
+host that provides it, along with WASI preview1 and thread spawning.
+
+Sourcing the setup script points Cargo's runner at that host, so tests work with the normal command:
+
+```sh
+source scripts/setup-wasi-env.sh
+cargo test -p turbo-tasks --lib --target wasm32-wasip1-threads
+```
+
+Some tests are skipped on wasm; each carries a reason describing the specific platform limitation
+(no unwinding, no mmap, and so on).
+
+### Linting
+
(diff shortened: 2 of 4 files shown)


dianatofficial (commented):

.github/workflows/build_and_test.yml
```diff
@@ -573,23 +573,84 @@ jobs:
...
+        # swc 74.0.0's `cfg(all(feature = "plugin", target_arch = "wasm32"))` path does not compile;
+        # see patches/rust/swc@74.0.0.patch. Applied out-of-tree so that normal builds, which do not
+        # pass the `--config` below, are unaffected.
+        SWC_DIR="$(node ./scripts/patch-rust-crate.mjs swc 74.0.0)"
+
+        cargo check -p next-napi-bindings --target wasm32-wasip1-threads \
+          --config "patch.crates-io.swc.path=\"$SWC_DIR\""
+
+        # Pointing [patch] at a path turns swc into a path dependency, so cargo drops its source and
+        # checksum from the lockfile. Don't leave that behind.
```
sokra: Remove the workarounds for the patch when we drop the swc patch PR

.github/workflows/build_and_test.yml
```diff
@@ -621,26 +621,80 @@ jobs:
...
+  # Guards that Turbopack keeps compiling for wasm, so the wasm fallback Next.js loads when the
+  # native bindings are unavailable can eventually support Turbopack and not just SWC.
+  # See https://github.com/vercel/next.js/issues/96960.
+  #
+  # This is a compile check only. A full `napi build` additionally has to link against emnapi, and
+  # `napi-build` requires emnapi v2 (for the `emnapi_create_env` / `emnapi_delete_env` exports),
+  # which is still unreleased — so linking cannot be gated on yet.
   test-next-napi-bindings-wasi:
-    name: test next-swc wasi
+    name: check next-swc wasi
```
sokra:
```suggestion
    name: test next-swc wasi
```

Keep the original name. We will add testing later

.github/workflows/build_and_test.yml
```diff
@@ -621,26 +621,80 @@ jobs:
       stepName: 'test-next-swc-wasm'
     secrets: inherit
 
-  #[NOTE] currently this only checks building wasi target
+  # Guards that Turbopack keeps compiling for wasm, so the wasm fallback Next.js loads when the
+  # native bindings are unavailable can eventually support Turbopack and not just SWC.
+  # See https://github.com/vercel/next.js/issues/96960.
```
bgw, line 626: Sure this is a case where wasm bindings would be a valid fallback for the user, but I wouldn't note that as a canonical case. Rather, that's actually not a great case for falling back to wasm.

They should fall back to musl (they've got some issue with their package manager preventing it) and/or we should match the glibc version officially supported by our minimum Node.js version.

```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -621,26 +621,80 @@ jobs:
...
+  #
+  # This is a compile check only. A full `napi build` additionally has to link against emnapi, and
+  # `napi-build` requires emnapi v2 (for the `emnapi_create_env` / `emnapi_delete_env` exports),
+  # which is still unreleased — so linking cannot be gated on yet.
```
bgw, line 630: > This is a compile check only

Doesn't `build_and_release` already basically test this for us? We just only run the full platform matrix on `canary` and not on every PR because compiling Turbopack is expensive.

If the goal is to run an actual test suite against the wasm bindings, then doing it in `build_and_test` makes sense, but it sounds like that's not currently the case?

.github/workflows/build_and_test.yml
```diff
@@ -621,26 +621,80 @@ jobs:
...
+        # pnpm is invoked from the repository root with --dir, not by cd'ing into the scratch
+        # directory: corepack resolves the pnpm version from the nearest package.json, and outside
+        # the repo there is none, so it picks the latest pnpm (11.x) which cannot run on the pinned
+        # Node 20 (`ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING`). Staying in the repo keeps the pinned
+        # pnpm from the root package.json's packageManager field.
+        EMNAPI_DIR="$RUNNER_TEMP/emnapi"
+        mkdir -p "$EMNAPI_DIR"
+        printf '{\n  "name": "emnapi-scratch",\n  "private": true\n}\n' > "$EMNAPI_DIR/package.json"
+        pnpm --dir "$EMNAPI_DIR" add emnapi@2.0.0-alpha.4
+        export EMNAPI_LINK_DIR="$EMNAPI_DIR/node_modules/emnapi/lib/wasm32-wasip1-threads"
```
bgw, line 692: We should figure out a way to share this script between `build_and_test` and `build_and_deploy`. Maybe this should live in `scripts/`? Alongside https://github.com/vercel/next.js/blob/69ec884e44d1050013724facfe452248e5b9de55/scripts/build-wasm.cjs ?

How are we expected to build wasm locally outside of CI? Should we do this with a dockerfile, like we do for glibc Linux builds where we have to compile against an old glibc?

bgw (changes requested):

.github/workflows/build_and_test.yml
sokra: If that's already covered by `build_and_release` we can omit it here

scripts/setup-wasi-env.sh
```diff
@@ -0,0 +1,102 @@
...
+  esac
+
+  local sdk_dir="$cache_dir/wasi-sdk-${wasi_sdk_version}-${wasi_sdk_arch}-linux"
+  if [ ! -x "$sdk_dir/bin/clang" ]; then
+    local tar="wasi-sdk-${wasi_sdk_version}-${wasi_sdk_arch}-linux.tar.gz"
+    curl --retry 3 --fail --location --silent --show-error --output "$cache_dir/$tar" \
+      "https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-${wasi_sdk_version%%.*}/${tar}" || return 1
+    echo "${wasi_sdk_sha256}  $cache_dir/${tar}" | sha256sum --check --strict || return 1
+    tar xzf "$cache_dir/$tar" -C "$cache_dir" || return 1
+  fi
```
bgw, line 61: We should either make an effort to support building from other operating systems, or if we only want to support building from Linux hosts, we should provide a dockerfile, like we do with https://github.com/vercel/next.js/blob/canary/scripts/native-builder.Dockerfile

.github/workflows/build_and_test.yml
```diff
@@ -621,26 +621,38 @@ jobs:
...
+      skipInstallBuild: 'yes'
       skipNativeBuild: 'yes'
       skipNativeInstall: 'yes'
       afterBuild: |
-        rustup target add wasm32-wasip1-threads
-        pnpm dlx turbo@${TURBO_VERSION} run build-native-wasi ${TURBO_ARGS}
+        set -euo pipefail
+
+        # Provisions the wasi-sdk, emnapi and the cross-compilation env. Shared with local
+        # builds so the two cannot drift; see contributing/core/building-wasm.md.
```
bgw: Don't need the comment, this is clear from the code
```suggestion
```

.github/workflows/build_and_test.yml
```diff
@@ -621,26 +621,38 @@ jobs:
...
+        source scripts/setup-wasi-env.sh
+
+        # swc's plugin host is gated off for wasm (see turbopack-ecmascript-plugins and
+        # next-napi-bindings), so nothing needs patching here: the wasm32 plugin path that does not
+        # compile is simply never built.
```
bgw:
```suggestion
```

.github/workflows/build_and_test.yml
bgw: Delete the line linking to https://github.com/vercel/next.js/issues/96960

bgw (changes requested):

scripts/wasi-builder.Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    bash \
+    ca-certificates \
+    curl \
+    git \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN corepack enable && \
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal
```
bgw: Whatever version of rust this installs will be wrong because it doesn't have access to the `rust-toolchain.toml`. I think you can pass `--no-update-default-toolchain` here

scripts/wasi-builder.Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Linux environment for building and testing Next.js' wasm32-wasip1-threads bindings from any host
+# that can run Docker.
+#
+# Build from the repository root:
+#   docker build -t next-wasi-builder -f scripts/wasi-builder.Dockerfile .
+#
+# Run with the repository mounted at /workspace:
+#   docker run --rm -it -v "$PWD:/workspace" -w /workspace next-wasi-builder
+
+FROM node:20-bookworm
```
bgw: Node 20 is EOL, but it does look like they published a docker image based on trixie, which is the newest Debian release. Bookworm is pretty old at this point. https://hub.docker.com/layers/library/node/20-trixie/images/sha256-f8ac5234895b491b8162521c1a635809b58e415e44b1767fac7fe82ee3d5455f

```suggestion
FROM node:20-trixie
```

contributing/core/building-wasm.md
```diff
@@ -0,0 +1,95 @@
...
+provisions both.
+
+## Building on Linux
+
+The setup script supports Linux systems directly. Source it, then build as usual:
+
+```sh
+source scripts/setup-wasi-env.sh
+cargo check -p next-napi-bindings --target wasm32-wasip1-threads
+```
```
bgw, line 23: Include a note that this script assumes bash (`BASH_SOURCE`). Many people (myself included) use alternative shells (zsh).

Assuming bash is okay, but we should document.

bgw (approved):

PR #97590, opened 2026-08-19
[ci] Authenticate Turborepo remote caching with OIDC instead of a static PAT

eps1lon:
CI authenticated to Vercel Remote Cache with a long-lived Personal Access Token in the `TURBO_TOKEN` repository secret. That token never expired, is scoped to a team member rather than the team, and is readable by every job that inherits secrets.

Each job now mints its own short-lived, cache-only token instead, using `vercel/setup-turborepo-remote-cache-action` against a Turborepo CLI OIDC policy configured on the Vercel team following https://vercel.com/docs/monorepos/remote-caching/external-ci-cd#openid-connect-oidc

Forks skip the step entirely since they won't have access to repository variables. During outages or any other permission errors, the steps outcome will simply be ignore and we fall back to uncached behavior. 

This could lead to silent regressions or hiding new, incorrect callsites lacking necessary permissions. A Datadog monitor is not as simple as I'd like since DD does not track outcome but conclusion (which is always success for continue-on-error). Adding custom tags via DD CLI feels to heavy. We'll revisit if this becomes a recurring issue.

## Test plan

- Turborepo cache still functions in CI on this branch: https://github.com/vercel/next.js/actions/runs/32314345819/job/96263533767?pr=97590#step:31:60

### DIFF
--- a/.github/workflows/pull_request_stats.yml
+++ b/.github/workflows/pull_request_stats.yml
@@ -17,11 +17,10 @@ env:
   TURBO_VERSION: 2.9.4
   TEST_CONCURRENCY: 6
 
-  TURBO_TEAM: 'vtest314-next-adapter-e2e-tests'
+  TURBO_TEAM: ${{ vars.TURBO_TEAM }}
   # Prefer shared remote cache across runs, but keep local cache enabled so jobs
   # degrade gracefully if the remote cache or token is unavailable.
   TURBO_CACHE: 'local:rw,remote:rw'
-  TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
   NEXT_TELEMETRY_DISABLED: 1
   # Vercel KV Store for test timings
   KV_REST_API_URL: ${{ secrets.KV_REST_API_URL }}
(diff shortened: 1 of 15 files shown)


.github/workflows/pull_request_stats.yml
```diff
@@ -113,13 +118,20 @@ jobs:
...
         if: ${{ steps.docs-change.outputs.DOCS_CHANGE == 'nope' }}
 
+      - name: Set up Turborepo remote cache
+        if: ${{ vars.TURBO_TEAM != '' }}
+        continue-on-error: true
+        uses: vercel/setup-turborepo-remote-cache-action@135046f5efcadb17337d9b29b8d5b4035ae64b10 # v1.0.0
+        with:
+          team: ${{ vars.TURBO_TEAM }}
+
+      # `TURBO_TOKEN` reaches this Docker action through the job environment.
```
eps1lon, line 128: proof: https://github.com/vercel/next.js/actions/runs/32314345874/job/96263676677?pr=97590#step:7:10595

bgw (approved):

PR #97591, opened 2026-08-19
Sweep stale Turbopack output from distDir on dev startup

wbinnssmith:
## Summary

Turbopack dev never removes the content-hashed chunks it supersedes, so `.next/dev` grows without bound. Webpack dev avoids this by wiping `distDir` on every start.

Adds a `maxAgeMs` argument to `recursiveDeleteSyncWithAsyncRetries`: delete files whose mtime is at least that old, remove directories once empty. Turbopack dev passes one week.

## Verification

- `pnpm test-unit test/unit/recursive-delete.test.ts`
- ~Not run:~ dev-server behavior against a real `.next/dev`

### DIFF
--- a/packages/next/src/lib/recursive-delete.ts
+++ b/packages/next/src/lib/recursive-delete.ts
@@ -59,16 +59,22 @@ function unlinkPath(
   }
 }
 
-/** Returns whether `p`'s own mtime is at least `maxAgeMs` in the past. */
-function isStale(p: string, maxAgeMs: number): boolean {
+const FUTURE_MTIME_TOLERANCE_MS = 60 * 60 * 1000
+
+/** Returns whether `p`'s own mtime falls outside the retained time range. */
+function isStale(
+  p: string,
+  staleBefore: number,
+  futureMtimeThreshold: number
+): boolean {
   try {
     // lstat, not stat: a symlink is deleted based on its own age, not its
     // target's.
-    return Date.now() - fs.lstatSync(p).mtimeMs >= maxAgeMs
+    const mtimeMs = fs.lstatSync(p).mtimeMs
+    return mtimeMs <= staleBefore || mtimeMs > futureMtimeThreshold
   } catch (e) {
     if (isError(e) && e.code === 'ENOENT') {
-      // Already gone; let the delete run and no-op on ENOENT.
-      return true
+      return false
     }
     throw e
   }
@@ -95,19 +101,29 @@ export async function recursiveDeleteSyncWithAsyncRetries(
    */
   maxAgeMs?: number
 ): Promise<void> {
-  await deleteContents(dir, exclude, maxAgeMs, '')
+  const now = Date.now()
+  await deleteContents(dir, {
+    exclude,
+    staleBefore: maxAgeMs === undefined ? undefined : now - maxAgeMs,
+    futureMtimeThreshold: now + FUTURE_MTIME_TOLERANCE_MS,
+  })
 }
 
-/**
- * @returns whether anything was kept, so a parent directory knows not to
- * remove itself.
- */
+/** @returns whether anything was kept. */
 async function deleteContents(
   dir: string,
-  exclude: RegExp | undefined,
-  maxAgeMs: number | undefined,
-  /** Relative path to the directory being deleted, used for exclude */
-  previousPath: string
+  {
+    exclude,
+    staleBefore,
+    futureMtimeThreshold,
+    previousPath,
+  }: {
+    exclude: RegExp | undefined
+    staleBefore: number | undefined
+    futureMtimeThreshold: number
+    /** Relative path to the directory being deleted, used for exclude */
+    previousPath?: string
+  }
 ): Promise<boolean> {
   let result
   try {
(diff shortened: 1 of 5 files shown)


packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -159,9 +160,26 @@ const isTestMode = !!(
...
+
 /** Output directory (relative to `distDir`) of server-HMR-managed chunks. */
 const SERVER_HMR_CHUNKS_DIR = join('server', 'chunks')
 
+/**
+ * Directories (relative to `distDir`) holding content-hashed turbopack output,
+ * per the chunking contexts in `next-core`. Entry chunks are deliberately left
+ * out: their paths derive from the route rather than a content hash, so they
+ * are overwritten in place instead of accumulating.
+ */
```
bgw: > Entry chunks are deliberately left
> out: their paths derive from the route rather than a content hash, so they
> are overwritten in place instead of accumulating.

Sure, but is there any harm in deleting them?

You could still end with with entrypoint files left over from stale files.

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -159,9 +160,26 @@ const isTestMode = !!(
 
 const sessionId = Math.floor(Number.MAX_SAFE_INTEGER * Math.random())
 
+/** How long an emitted asset may go unwritten before startup sweeps it. */
+const STALE_OUTPUT_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000 // One week
```
bgw: Make this configurable via an experimental option, configure it in `front` to be way more aggressive (e.g. 2 hours) so that we can get more signal from internal developers if there are consequences of this.

Maybe try temporally setting this value to 0ms in a stacked PR and seeing that none of our e2e tests break.

packages/next/src/lib/recursive-delete.ts
```diff
@@ -59,6 +59,21 @@ function unlinkPath(
   }
 }
 
+/** Returns whether `p`'s own mtime is at least `maxAgeMs` in the past. */
```
bgw: If we get an mtime in the far future (e.g more than an hour in the future), we should also delete it to be safe, assuming the user's system clock was wrong at some point.

bgw (approved):

packages/next/src/lib/recursive-delete.ts
```diff
@@ -59,6 +59,21 @@ function unlinkPath(
...
+function isStale(p: string, maxAgeMs: number): boolean {
+  try {
+    // lstat, not stat: a symlink is deleted based on its own age, not its
+    // target's.
+    return Date.now() - fs.lstatSync(p).mtimeMs >= maxAgeMs
```
lukesandberg: i don't know if it matters but calling date.now on each is kind of rough

instead compute the absolute time in the past that is our threshold and pass that into `isStale` and we can do that once at the beginning of the loop

packages/next/src/lib/recursive-delete.ts
```diff
@@ -59,6 +59,21 @@ function unlinkPath(
...
+  } catch (e) {
+    if (isError(e) && e.code === 'ENOENT') {
+      // Already gone; let the delete run and no-op on ENOENT.
+      return true
```
lukesandberg: or return false because absent files are not stale?

packages/next/src/lib/recursive-delete.ts
```diff
@@ -74,34 +89,64 @@ export async function recursiveDeleteSyncWithAsyncRetries(
...
+      if (exclude?.test(pp)) {
+        keptAnything = true
+        return
+      }
+
+      // Note: readdir does not follow symbolic links, that's good: we want to
+      // delete the links and not the destination.
+      const isDirectory = part.isDirectory()
+      if (isDirectory) {
+        if (await deleteContents(absolutePath, exclude, maxAgeMs, pp)) {
```
lukesandberg: is this function only `async` for recursive reasons?  afaict this is the only await and the only reason `deleteContents` is async is due to the Promise.all?

i think all async and await keywords can just be removed

packages/next/src/server/dev/hot-reloader-turbopack.ts
```diff
@@ -447,6 +465,23 @@ export async function createHotReloaderTurbopack(
...
   }
 
+  // Clean up any old output files from previous runs. This is safe as Turbopack
+  // will restore any missing chunks from persistent cache or recompute them.
+  //
+  // That only holds here, before the project exists: once turbo-tasks has
+  // recorded a write effect for a path, the write dedups on the recorded hash
+  // without checking the file, so a deleted output stays missing for the rest
+  // of the session.
+  await Promise.all(
```
lukesandberg, line 467: another dead await i think

dianatofficial (commented):

packages/next/src/lib/recursive-delete.ts
wbinnssmith: my understanding is that unlinking on Windows can go wrong and we retry a few times asynchronously.

Given this is an optimistic best effort at cleaning things up, It's probably fine to skip awaiting these as long as there aren't any unhandled promise rejections etc

lukesandberg: oic, ouch, i didn't catch that

yeah skipping makes some sense for this case but for the build case it isn't so optional iirc

accumulating all the 'failed delete' files as we go and doing one retry at the end might make sense

but definitely out of scope for this PR

lukesandberg (approved):

PR #97592, opened 2026-08-19
Add next/cache-handlers types entrypoint

marceloboeira:
### What?

Exposes the `cacheHandlers` types (`CacheHandler`, `CacheEntry`) as types-only exports from `next/cache`, so handlers can be checked against the real interface:

```ts
import type { CacheHandler } from 'next/cache'
```

### Why?

Custom cache handler authors currently import `CacheHandler` and `CacheEntry` from `next/dist/server/lib/cache-handlers/types` (an internal path that can move between versions) or hand-copy the interfaces, which drift silently across releases.

### Motivation

I'm working on a custom community cache handler and it'd be great to import the types directly and keep testing against the source of truth, rather than maintaining a hand-copied mirror that has to be re-checked on every Next.js release.

Raised in discussion #96356.

### How?

- Adds a type-only `export type { CacheHandler, CacheEntry }` to `packages/next/cache.d.ts`, re-exported from `./dist/server/lib/cache-handlers/types`. No new subpath or `package.json` `"files"` entries needed since `next/cache` already ships.
- Docs: `cacheHandlers.mdx` now shows the public `next/cache` import instead of the GitHub source links.
- Tests: a `satisfies CacheHandler` / `satisfies CacheEntry` fixture in the `typescript-basic` typechecking suite (runs `tsc` against the installed package), and the `use-cache-custom-handler` e2e fixture's JSDoc now uses the public import.

(Originally proposed as a separate `next/cache-handlers` types-only subpath, following the `next/types` pattern moved the export into `next/cache` per review.)

Related: #96356

closes #97781 (only created to run deploy tests)

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/cacheHandlers.mdx
@@ -71,7 +71,11 @@ Note that `'use cache: private'` does not use cache handlers and cannot be custo
 
 ## API Reference
 
-A cache handler must implement the [`CacheHandler`](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/lib/cache-handlers/types.ts) interface with the following methods:
+A cache handler must implement the `CacheHandler` interface with the following methods. The `CacheHandler` and `CacheEntry` types are exported from `next/cache`:
+
+```ts
+import type { CacheHandler, CacheEntry } from 'next/cache'
+```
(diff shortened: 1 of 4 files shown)


dianatofficial (commented):

unstubbable (changes requested):
Let's export these from `next/cache` instead.

marceloboeira:
@unstubbable done :)

unstubbable (approved):

PR #97595, opened 2026-08-20
Add agent feedback CLI

marcoshernanz:
## Summary

- Add a hidden `next agent-feedback` command that prepares agent-authored feedback for review on nextjs.org.
- Include the installed Next.js version and detected coding agent, then encode the report as a URL fragment.
- Print the review URL and open it in the default browser without submitting anything.

## Verification

- `pnpm jest test/unit/agent-feedback-cli.test.ts --runInBand`
- `pnpm --filter=next types`
- Manually verified the compiled CLI preserves Unicode and emits the expected feedback, version, and agent metadata.

### DIFF
--- a/packages/next/src/bin/next.ts
+++ b/packages/next/src/bin/next.ts
@@ -34,6 +34,7 @@ import type { NextTypegenOptions } from '../cli/next-typegen.js'
 import type { NextPostBuildOptions } from '../cli/next-post-build.js'
 import { ensureProfilesDir } from '../lib/profiles-dir'
 import type { NextRequestInsightsOptions } from '../cli/next-request-insights.js'
+import type { NextAgentFeedbackOptions } from '../cli/next-agent-feedback.js'
 
 /**
  * Create `.next-profiles` (with its `.gitignore`) when profiling/tracing is
@@ -637,6 +638,16 @@ program
     )
   })
   .usage('[directory] [options]')
+
+program
+  .command('agent-feedback', { hidden: true })
+  .description('Open a prepared Next.js agent feedback report for review.')
+  .requiredOption('--feedback <text>', 'The de-identified feedback to review.')
+  .action((options: NextAgentFeedbackOptions) => {
+    return import('../cli/next-agent-feedback.js').then((mod) =>
+      mod.nextAgentFeedback(options)
+    )
+  })
 
 const internal = program
   .command('internal')
--- a/packages/next/src/cli/next-agent-feedback.ts
+++ b/packages/next/src/cli/next-agent-feedback.ts
@@ -1,0 +1,46 @@
+import { spawn } from 'node:child_process'
+
+import { getAgentName } from '../telemetry/agent-name'
+
+export type NextAgentFeedbackOptions = {
+  feedback: string
+}
+
+export function createAgentFeedbackUrl(
+  feedback: string,
+  { nextVersion, agent }: { nextVersion?: string; agent?: string } = {}
+): string | undefined {
+  feedback = feedback.trim()
+  if (!feedback || feedback.length > 2000) {
+    return
+  }
+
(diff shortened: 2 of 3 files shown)


packages/next/src/bin/next.ts
```diff
@@ -638,6 +639,16 @@ program
...
+  .command('agent-feedback', { hidden: true })
+  .description('Open a prepared Next.js agent feedback report for review.')
+  .requiredOption('--feedback <text>', 'The de-identified feedback to review.')
+  .action((options: NextAgentFeedbackOptions) => {
+    return import('../cli/next-agent-feedback.js').then((mod) =>
+      mod.nextAgentFeedback(options)
+    )
+  })
+
 const internal = program
```
bgw, line 652: Can we hide this under `next internal agent-feedback`? Only agents need to know about this, and they'll be informed about it from `AGENTS.md` if the user opts in.

packages/next/src/cli/next-agent-feedback.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  console.log(`Review and submit the report at:\n${url}`)
+
+  const isWindows = process.platform === 'win32'
+  const child = spawn(
+    isWindows ? 'cmd.exe' : process.platform === 'darwin' ? 'open' : 'xdg-open',
+    isWindows ? ['/c', 'start', '', url] : [url],
+    { detached: true, stdio: 'ignore', windowsHide: true }
+  )
+  child.on('error', () => {})
```
bgw, line 44: If this fails, we should log a message, and have the agent ask the user to open the URL

packages/next/src/cli/next-agent-feedback.ts
```diff
@@ -0,0 +1,46 @@
...
+    return
+  }
+
+  console.log(`Review and submit the report at:\n${url}`)
+
+  const isWindows = process.platform === 'win32'
+  const child = spawn(
+    isWindows ? 'cmd.exe' : process.platform === 'darwin' ? 'open' : 'xdg-open',
+    isWindows ? ['/c', 'start', '', url] : [url],
+    { detached: true, stdio: 'ignore', windowsHide: true }
```
bgw, line 42: Maybe we should inherit the stdio instead of ignoring it, there might be useful output.

test/unit/agent-feedback-cli.test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    expect(url.origin + url.pathname).toBe('https://nextjs.org/agent-feedback')
+    expect(
+      JSON.parse(Buffer.from(report!, 'base64url').toString('utf8'))
+    ).toEqual({
+      feedback: 'Unicode works: café ☕',
+      nextVersion: '16.3.1-canary.19',
+      agent: 'Codex',
+    })
+  })
```
bgw, line 21: IMO, this doesn't seem to provide a lot of value as a unit test, it's just asserting a very specific implementation.

packages/next/src/cli/next-agent-feedback.ts
```diff
@@ -0,0 +1,46 @@
...
+  ).toString('base64url')}`
+}
+
+export async function nextAgentFeedback(
+  options: NextAgentFeedbackOptions
+): Promise<void> {
+  const url = createAgentFeedbackUrl(options.feedback, {
+    nextVersion: process.env.__NEXT_VERSION,
+    agent: (await getAgentName()) ?? undefined,
+  })
```
bgw, line 29: IMO, this would all be easier to read if `createAgentFeedbackUrl` was just inlined here (it's a tiny function), and I don't think the unit tests add enough value to justify splitting it off.

Otherwise there's the indirection of what `url === undefined` means.

bgw (approved):

dianatofficial (commented):

PR #97601, opened 2026-08-20
Optimize fallible iterator joins

sokra:
### What?

Adds Criterion micro-benchmarks for `try_join()` and `try_flat_join()` and replaces their `futures::join_all` backend with a specialized iterator join implementation.

At 10,000 immediately-ready futures, median time improves from 1.228 ms to 81.5 µs for `try_join()` and from 1.241 ms to 85.0 µs for `try_flat_join()`.

### Why?

`join_all` switches to `FuturesOrdered` above 30 inputs. Its per-future scheduling machinery is useful for pending futures, but it creates a large performance cliff when every future resolves immediately. The previous implementation was roughly 138× slower than synchronous map-and-collect at 10,000 inputs.

### How?

The new join implementation first polls every future directly using the parent waker. Fully synchronous joins complete without allocating per-future scheduling state.

When futures remain pending, small joins continue direct scanning while large joins lazily install indexed child wakers and poll only futures placed in the wake queue. This retains wake-driven scaling for large pending joins without paying that cost on the immediate path. Completed outputs are collected directly into the final result while preserving input and error order.

The benchmark covers immediate completion, completion on the second poll, and sequential completion across 1–10,000 inputs. Focused tests cover ordering, flattening, error selection, `!Unpin` futures, stale/racing wakes, sequential wake behavior, and cross-thread wakeups.

### Performance

Micro-benchmark medians at 10,000 inputs:

| Case | Before | After |
| --- | ---: | ---: |
| immediate `try_join` | 1.228 ms | 81.5 µs |
| immediate `try_flat_join` | 1.241 ms | 85.0 µs |
| second-poll `try_join` | 1.975 ms | ~472 µs |
| sequential `try_join` | 2.649 ms | ~1.84 ms |

Large-input scalability was also measured against the old `join_all` backend:

| Pattern | Size | Current PR | Old code | Current speedup |
| immediate | 100k | 0.906 ms | 10.372 ms | 11.45× |
| immediate | 1M | 29.117 ms | 168.565 ms | 5.79× |
| second poll | 100k | 4.920 ms | 20.858 ms | 4.24× |
| second poll | 1M | 71.984 ms | 247.227 ms | 3.43× |
| sequential | 100k | 20.290 ms | 28.158 ms | 1.39× |
| sequential | 1M | 217.231 ms | 311.477 ms | 1.43× |

Both implementations completed all 1M-item cases without an OOM, hang, or poll-count failure. The current immediate path has a confirmed working-set cliff: 10× more inputs took 32× longer (independent rerun ~34×), versus 16× on old code. It remains 5.8× faster at 1M and uses less peak RSS, so this appears to be increased cache/allocation cost rather than O(N²) behavior. Sequential N-poll scaling remained close to linear (10.7× for 10× inputs).

Peak RSS at 1M was 55 / 74 / 120 MiB for current immediate / second-poll / sequential, versus 112 / 128 / 127 MiB on old code. These process-level numbers include synthetic benchmark control state.

Two larger clean-build A/B workloads were also measured with release `next-swc` binaries differing only in `join_iter_ext.rs`. Neither produced a reproducible top-level difference:

| Workload | Pairs | Candidate effect | Paired t 95% CI | Signs |
| --- | ---: | ---: | --- | --- |
| `heavy-npm-deps` | 8 | +0.66% slower | [-1.30%, +2.63%] | 5 faster / 3 slower |
| `nested-deps-app-router-many-pages` | 16 | -0.75% faster | [-2.11%, +0.61%] | 9 faster / 7 slower |

The first 8-pair many-pages run suggested -1.92%, but an independent 8-pair confirmation reversed direction (+0.44% slower), so the combined result is reported as **no detected difference**. The observed noise floors were approximately 2.0% and 1.36%, respectively; any complete-build effect is smaller than those runs could resolve.

### Verification

- `cargo test -p turbo-tasks join_iter_ext`
- `cargo clippy -p turbo-tasks --all-targets -- -D warnings`
- `cargo bench -p turbo-tasks --bench mod --no-run`
- `cargo bench -p turbo-tasks --bench mod -- 'try_join|try_flat_join' --test`
- 48 successful clean production Turbopack builds across the two A/B workloads

### DIFF
--- a/turbopack/crates/turbo-tasks/src/join_iter_ext.rs
+++ b/turbopack/crates/turbo-tasks/src/join_iter_ext.rs
@@ -1,13 +1,18 @@
 use std::{
     future::{Future, IntoFuture},
+    mem,
     pin::Pin,
-    task::Poll,
+    sync::{
+        Arc, Mutex, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
+    task::{Context, Poll},
 };
 
 use anyhow::Result;
 use futures::{
-    FutureExt,
     future::{JoinAll, join_all},
+    task::{ArcWake, AtomicWaker, waker_ref},
 };
 use pin_project_lite::pin_project;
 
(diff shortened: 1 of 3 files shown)


turbopack/crates/turbo-tasks/src/join_iter_ext.rs
```diff
@@ -45,6 +50,220 @@ where
...
+                    let mut queue = shared.queue.lock().unwrap();
+                    // Reset `queued` while holding the queue lock. A concurrent wake then either
+                    // observes the old queued item (which this poll handles) or waits for the lock
+                    // and appends a new item; it cannot be lost between those operations.
+                    let queued = mem::take(&mut *queue);
+                    queued
+                        .into_iter()
+                        .filter_map(|task| task.upgrade())
+                        .inspect(|task| task.queued.store(false, Ordering::Release))
+                        .collect::<Vec<_>>()
```
sokra: If you make the queue store Arc instead of Weak we can avoid this additional collect here.

turbopack/crates/turbo-tasks/src/join_iter_ext.rs
```diff
@@ -45,6 +50,220 @@ where
...
+
+                for task in ready {
+                    let mut elem = get_pin_mut(this.elems.as_mut(), task.index);
+                    let elem = elem.as_mut().project();
+                    if elem.output.is_some() {
+                        continue;
+                    }
+                    let waker = waker_ref(&task);
+                    let mut child_cx = Context::from_waker(&waker);
+                    if let Poll::Ready(output) = elem.future.poll(&mut child_cx) {
```
sokra, line 253: Maybe check if `wake_tasks` is still Some for this elem. There might be a race condition where an already completed future ends up in the queue.

sokra: You could also check if `elem.output.is_none()`

bgw:
This scares me. I've written raw `Future` code before, but these sorts of APIs are really hard to get right (especially if you care about performance), and this is a massive amount of code that I'd have to review (this would take me hours to carefully pick through) for what is likely a small top-level win (it's only a problem if we become starved for work).

Are there any other ways we could do this? What about `tokio::task::JoinSet`?

bgw:
If you really do want to go down this path, maybe consider just copying https://github.com/facebook/buck2/blob/a3b535bb144294cbda6b04ae3b58f38c3f503fa2/dice/dice_futures/src/join.rs

At least we can know that one has already been audited and reviewed by somebody.

sokra:
I think it fine to drop the PR if we want to avoid adding complexity of that.

The scaling performance worries me a bit. What would happen if I pass 100k or 1M items to try_join? Would we run into a bad edge case with that? (Fleet could you benchmark for bigger numbers)

sokra:
Ok. Doesn't seem to be too bad. We don't have to merge it

bgw:
> It also depends on Buck2's unpublished mini_vec internals (~2.3k relevant lines).

It would likely be easy to strip that out

> Its `try_join_all` returns the first error in time and cancels the rest, unlike our documented first-error-in-input-order semantics.

Those semantics are likely unimportant.

PR #97606, opened 2026-08-20
fix(turbopack): don't wait forever on a runtime-dependency chunk that is never injected

### LINKED ISSUE #97605: Turbopack: generateComponentChunks can leave a page permanently unhydrated with no errors (runtime-dependency chunk never injected, so registerChunk waits forever)

https://github.com/yunsii/turbopack-generate-component-chunks-repro

### To Reproduce

```bash
git clone https://github.com/yunsii/turbopack-generate-component-chunks-repro
cd turbopack-generate-component-chunks-repro
pnpm install

MERGE=1 pnpm build && pnpm start     # generateComponentChunks: true
# open http://localhost:3999/p0
```

`MERGE=0 pnpm build && pnpm start` is the control — identical source, the flag off.

Serve each build from its own directory (or delete `.next` and restart between runs);
swapping `.next` under a running server produces buildId-mismatch 404s that look like
this bug but are not.

**Current:** with `experimental.turbopackChunking.generateComponentChunks: true`, `/p0`
renders its server HTML completely and then never hydrates. Nothing is interactive.
There are **no console errors or warnings**, no failed requests, and every chunk that is
requested returns 200.

In the DevTools console on `/p0`:

```js
Object.keys(document.getElementById('__next')).some((k) => k.startsWith('__reactFiber'))
// false with MERGE=1, true with MERGE=0
window.next?.router   // undefined with MERGE=1, defined with MERGE=0
```

**Expected:** the page hydrates, as it does with the flag off.

Identical source, only the flag differs (`/p0`):

| | `MERGE=1` | `MERGE=0` |
|---|---|---|
| pages registered via `__NEXT_P` | `['/_app']` | `['/_app', '/p0']` |
| module factories registered | 312 | 312 |
| modules instantiated | 115 | **116** |
| page registration entry module `994547` | factory registered, **never instantiated** | instantiated |
| `__reactFiber` / `window.next.router` | false / undefined | true / defined |
| console errors | 0 | 0 |

`/_app`'s own entry module instantiates in both builds; only the *page* registration
module is skipped, and exactly one module fewer is instantiated.

### Root cause

The blocked step is `await Promise.all(params.otherChunks.map(loadInitialChunk))` in
`registerChunk` — entry modules are instantiated only after it, and it never settles.

The unsettled dependency is a chunk that:

- **exists** in the build output and serves 200
- is **not referenced anywhere in the HTML**
- was **never requested** (0 network requests, no `<script>` element for it)
- has a resolver stuck at `{ loadingStarted: true, resolved: false }`
- would provide module `711138`, which is **already registered** — component-level
  chunking duplicated that module into another chunk that *is* in the HTML

`doLoadChunk` returns early for `SourceType.Runtime`
(`turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts`):

```ts
if (sourceType === SourceType.Runtime) {
  // We don't need to load chunks references from runtime code, as they're already
  // present in the DOM.
  resolver.loadingStarted = true
  if (isCss(chunkUrl)) resolver.resolve()
  return resolver.promise
}
```

`generateComponentChunks` breaks that assumption: a chunk can be listed as a runtime
dependency without ever being injected. Nothing loads it, nothing resolves it, and
`registerChunk` waits on it forever — silently, because a pending promise is not a failure.

This also explains a threshold we can observe within a single build: whether a page breaks
tracks the width of its entry chunk's `otherChunks` list.

| page | `otherChunks` | entry skipped | hydrates |
|---|---|---|---|
| `/p1` | 70 | `598365` | no |
| `/p0` | 69 | `994547` | no |
| `/p3` | 64 | `577091` | no |
| `/p2` | 63 | `119554` | no |
| `/p4` | 50 | — | yes |
| `/p5` | 48 | — | yes |
| `/p6` | 30 | — | yes |

All seven hydrate with the flag off.

The repo includes `debug/audit.mjs`, which reports for every module declared as a chunk
entry whether its factory was registered and whether it was ever instantiated
(`factoryRegistered && !started` is the bug), and `debug/check-hydration.mjs` for the end
symptom. Note that instrumenting the runtime perturbs hydration, so the two must be run
separately.

```
Operating System:
  Platform: linux
  Arch: x64
  Available memory (MB): 44141
  Available CPU cores: 20
Binaries:
  Node: 22.23.1
  npm: 10.9.8
  pnpm: 10.33.0
Relevant Packages:
  next: 16.3.0
  react: 18.2.0
  react-dom: 18.2.0
  typescript: 7.0.2
Next.js Config:
  output: standalone
```

Also reproduces on **`next@16.3.1-canary.25`** with identical numbers
(`registered: ['/_app']`, 312 factories, 115 instantiated). The affected source file is
byte-identical between `v16.3.0` and `canary`.

Turbopack

next build (local), next start

### Additional context

Found in a production app (~1811 module factories, entry chunk `otherChunks` width 78)
with the same signature. It shipped to a test environment and was rolled back; automated
smoke checks did not catch it because they only asserted "no JS errors", which this failure
mode never produces.

A one-line fix is proposed in a PR: let JS runtime dependencies fall through to the DOM
loader, which reuses an existing `<script>` when there is one and injects one otherwise.
Applied to the built output of both 16.3.0 and canary it restores hydration on `/p0` and
`/p1`, leaves `/p4` unchanged, and costs exactly one additional chunk request with no
duplicate requests and no exceptions.

Arguably the deeper fix is on the build side: a chunk that is never injected should not be
listed in another chunk's `otherChunks` — especially when its modules are already provided
by a duplicate chunk that is injected.

yunsii:
### What

`doLoadChunk` returns early for `SourceType.Runtime` JS chunks, on the assumption —
stated in the comment it replaces — that such chunks are "already present in the DOM"
and will resolve their own resolver through `registerChunk`.

When a chunk is listed as a runtime dependency but never injected into the document,
nothing loads it, nothing resolves it, and the
`await Promise.all(params.otherChunks.map(loadInitialChunk))` in `registerChunk` never
settles. Entry modules are instantiated only after that await, so they never run.

For a Pages Router page that means its `__NEXT_P` registration module never executes, the
entrypoint never reaches `routeLoader`, and **the page never hydrates — with no error of
any kind**, because a pending promise is not a failure.

### Why it happens

Observed with `experimental.turbopackChunking.generateComponentChunks: true`. In the
reproduction, the page's entry chunk lists a chunk in `otherChunks` that

- exists in the build output and serves 200,
- is not referenced anywhere in the HTML,
- is never requested (0 network requests, no `<script>` element),
- has a resolver stuck at `{ loadingStarted: true, resolved: false }`,
- and would provide a module that is **already registered**, because component-level
  chunking duplicated that module into another chunk that *is* injected.

So the chunk is redundant and unreferenced, and waiting for it can never end.

### How

Keep the fast path for CSS (unchanged) and the early return for workers (unchanged, so
worker chunks are not re-`importScripts`ed). Let JS in a document fall through to the DOM
loader below, which already reuses an existing `<script src=…>` when there is one — the
normal case, behaviour unchanged — and injects one otherwise.

### Verification

Reproduction: https://github.com/yunsii/turbopack-generate-component-chunks-repro

The runtime JS is embedded in the `next-swc` native binary, so this was verified by
applying the equivalent change to a **built** output (`.next/static/chunks/turbopack-*.js`)
of that reproduction — same artifacts, only this logic differs. The repo contains the
patch script used (`debug/patch-runtime-fix.mjs`) so the check can be repeated.

| page | unpatched | patched |
|---|---|---|
| `/p0` | not hydrated | hydrated |
| `/p1` | not hydrated | hydrated |
| `/p4` (already fine) | hydrated | hydrated (unchanged) |

Verified on both `next@16.3.0` and `next@16.3.1-canary.25`. Cost: exactly **one**
additional chunk request on `/p0` (52 vs 51) — the previously orphaned chunk — with **zero
duplicate requests** and **zero JS exceptions**.

I have not built `next-swc` locally, so I have not run the Rust-side test suites against
this change; happy to follow up if a maintainer points me at the right ones.

### Note on the deeper fix

This makes the runtime resilient, but the chunk graph itself looks inconsistent: a chunk
that is never injected arguably should not appear in another chunk's `otherChunks` at all,
particularly when its modules are already provided by a duplicate chunk that is injected.
That would be a build-side change and I have not attempted it — this PR is the defensive
half, and it removes a failure mode that is completely silent today.

Fixes #97605

### DIFF
--- a/turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
+++ b/turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
@@ -181,21 +181,35 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
     }
 
     if (sourceType === SourceType.Runtime) {
-      // We don't need to load chunks references from runtime code, as they're already
-      // present in the DOM.
       resolver.loadingStarted = true
 
       if (isCss(chunkUrl)) {
         // CSS chunks do not register themselves, and as such must be marked as
         // loaded instantly.
         resolver.resolve()
+        return resolver.promise
       }
 
-      // We need to wait for JS chunks to register themselves within `registerChunk`
-      // before we can start instantiating runtime modules, hence the absence of
-      // `resolver.resolve()` in this branch.
+      // In a worker the runtime's chunks are imported eagerly, so there is nothing
+      // left to do here.
+      if (typeof importScripts === 'function') {
+        return resolver.promise
+      }
 
-      return resolver.promise
+      // In a document, a chunk referenced from runtime code is *usually* already
+      // present as a `<script>` tag and resolves itself through `registerChunk`, so
+      // there is deliberately no `resolver.resolve()` for JS here.
+      //
+      // That is not guaranteed, though: a chunk can be listed as a runtime dependency
+      // without ever being injected into the document -- for example when its modules
+      // were also emitted into another chunk that *is* injected, leaving this one
+      // redundant and unreferenced. Nothing then loads it, nothing resolves it, and
+      // `registerChunk` waits on it forever, so runtime modules are never instantiated
+      // and the page never hydrates -- with no error, because a pending promise is not
+      // a failure.
+      //
+      // Fall through to the DOM loader below instead: it reuses the existing `<script>`
+      // when there is one (unchanged behaviour) and injects one otherwise.
     }
 
     if (typeof importScripts === 'function') {


dianatofficial (commented):

PR #97609, opened 2026-08-20
test: stabilize poisoned proxy error overlay

sokra:
### What?

Stabilizes the RSC poisoned-import error-overlay test for `proxy.js` without weakening or retrying its redbox assertion.

### Why?

An initially broken proxy can emit the expected build error and then force a startup full reload. That reload can clear the overlay before the test begins observing it, causing an intermittent `Expected Redbox but found no visible one` failure even though validation worked.

### How?

The proxy parameter now starts from a valid module and introduces the poisoned import only after the sandbox page has hydrated. This moves the assertion onto the live HMR path and removes the startup-reload race. Middleware and instrumentation retain their initial-compile setup because their error overlays are reliable there and instrumentation does not reliably surface this edit through Turbopack HMR.

### Verification

- `pnpm test-dev-webpack test/development/acceptance-app/rsc-build-errors-poisoned-imports.test.ts`
- `pnpm test-dev-turbo test/development/acceptance-app/rsc-build-errors-poisoned-imports.test.ts`
- Proxy case repeated 5 times with Webpack and 3 times with Turbopack

### DIFF
--- a/test/development/acceptance-app/rsc-build-errors-poisoned-imports.test.ts
+++ b/test/development/acceptance-app/rsc-build-errors-poisoned-imports.test.ts
@@ -94,12 +94,13 @@ runRscBuildErrorsTests(({ next, isTurbopack }) => {
     )
   })
 
-  // The third element defers the poisoned import until after the page has
-  // hydrated. A broken `proxy.js` is compiled while the dev server is starting
-  // up and forces a full reload, which can clear the overlay before
-  // `waitForRedbox` starts observing it. The other two entries surface the
-  // error reliably from the initial compile, and only surface it reliably that
-  // way, so they keep the poisoned import in the initial files.
+  // Proxy runs in the Node.js server compiler, which is also invalidated when
+  // the app route is added on demand. If the initial proxy error reaches the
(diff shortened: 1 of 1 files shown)


test/development/acceptance-app/rsc-build-errors-poisoned-imports.test.ts
```diff
@@ -112,21 +124,29 @@ runRscBuildErrorsTests(({ next, isTurbopack }) => {
...
+
+      const expectPoisonedImportRedbox = async () => {
+        await session.waitForRedbox()
+        expect(await session.getRedboxSource()).toInclude(
+          'You\'re importing a module that depends on `catchError` into a React Server Component module. This API is only available in Client Components. To fix, mark the file (or its parent) with the `"use client"` directive.'
+        )
+      }
+      await expectPoisonedImportRedbox()
+
+      if (isTurbopack && isProxy) {
```
sokra, line 144:
```suggestion
      if (isTurbopack) {
```

can we run that for all?

mischnic (approved):

eps1lon (approved):

PR #97613, opened 2026-08-20
fix: preserve trailing slash during export MPA fallback

### LINKED ISSUE #97608: Static export MPA fallback drops trailing slash after RSC build ID mismatch

https://github.com/QingHeSite/next-trailing-slash-repro

### To Reproduce

2. Run pnpm install.
3. Run pnpm repro:build to generate static exports with different build IDs.
4. Run pnpm repro:serve.
5. Open http://localhost:4173/.
6. Click the link to /target/.
7. Observe that the RSC build ID mismatch triggers a full-page navigation and the browser ends at /target instead of /target/.

Current: When the RSC response has a different build ID, Next.js falls back to a full-page navigation. It converts /target/index.txt to /target, dropping the trailing slash even though trailingSlash: true.
Expected: The fallback navigation should preserve the configured URL format and navigate to /target/. It should not rely on the static server to redirect /target back to /target/.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:26 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T8132
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 20.20.2
  npm: 10.8.2
  Yarn: N/A
  pnpm: 10.33.0
Relevant Packages:
  next: 16.3.1-canary.25
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: export
```

Linking and Navigating, Output, Redirects

### Additional context

_No response_

QingHeSite:
Keeps the trailing slash when a static export falls back to a full page load, so you land on the canonical URL instead of bouncing through a redirect.

Fixes https://github.com/vercel/next.js/issues/97608

### DIFF
--- a/packages/next/src/client/route-params.ts
+++ b/packages/next/src/client/route-params.ts
@@ -11,6 +11,7 @@ import {
   NEXT_RSC_UNION_QUERY,
 } from './components/app-router-headers'
 import { hasBasePath } from './has-base-path'
+import { normalizePathTrailingSlash } from './normalize-trailing-slash'
 import { removeBasePath } from './remove-base-path'
 import type {
   NormalizedPathname,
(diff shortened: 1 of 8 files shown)


gaojude (approved):
CI green across all shards. Simplified the decode to invert the marker unconditionally and delegate trailing-slash policy to normalizePathTrailingSlash, and excluded the new export fixture from the cache-components lane.

PR #97615, opened 2026-08-20
fix(dynamic): respect parent Suspense when no loading is provided

### LINKED ISSUE #73830: dynamic() ignores <Suspense> hierarchy resulting in layout flicker

ahmashadani:
## Summary
Fixes #73830

Implemented with Midfleet infrastructure.

## Details
- lazy-dynamic/loadable.tsx: only wrap with inner Suspense when loading is provided
- loadable.shared-runtime.tsx: null-guard when loading is absent
- unit coverage for parent fallback, custom loading, and ssr:false without loading

## Test plan
- [ ] pnpm test test/unit/next-dynamic.test.tsx
- [ ] pnpm test test/e2e/next-dynamic/next-dynamic.test.ts

### DIFF
--- a/packages/next/src/shared/lib/lazy-dynamic/loadable.tsx
+++ b/packages/next/src/shared/lib/lazy-dynamic/loadable.tsx
@@ -48,8 +48,9 @@ function Loadable(options: LoadableOptions) {
       <Loading isLoading={true} pastDelay={true} error={null} />
     ) : null
 
-    // If it's non-SSR or provided a loading component, wrap it in a suspense boundary
-    const hasSuspenseBoundary = !opts.ssr || !!opts.loading
+    // If provided a loading component, wrap it in a suspense boundary.
+    // Otherwise let a parent suspense boundary handle the fallback.
+    const hasSuspenseBoundary = !!opts.loading
     const Wrap = hasSuspenseBoundary ? Suspense : Fragment
     const wrapProps = hasSuspenseBoundary ? { fallback: fallbackElement } : {}
     const children = opts.ssr ? (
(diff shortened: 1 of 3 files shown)


dianatofficial (commented):

PR #97616, opened 2026-08-20
[ci] Derive the PR Stats baseline from the PR's base ref

eps1lon:
`PR Stats` compared every pull request against a hardcoded base branch, which is `canary`. For a pull request into a release branch that is the wrong baseline for two reasons: the diff is against a different major, and the auto merge of the baseline into the pull request branch cannot succeed. This is especially noticeable on old branches where we fail to produce stats outright because the harness simply can't handle latest `canary`.

We now use the base ref from the PR event falling back to the hardcoded value for local runs.

For a canary pull request `base.ref` equals the old hardcoded value, so behaviour on canary is unchanged. However, stacked PRs now (correctly) diff against their parent frame instead of the trunk of the stack.

### DIFF
--- a/.github/actions/next-stats-action/README.md
+++ b/.github/actions/next-stats-action/README.md
@@ -23,7 +23,9 @@ const StatsConfig = {
   // the command to build the app (app source should be in `.stats-app`)
   appBuildCommand: string,
   appStartCommand: string | undefined,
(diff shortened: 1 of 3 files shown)


bgw (approved):

eps1lon:
Job has been removed

PR #97617, opened 2026-08-20
fix(turbopack): skip symlink cycles in NFT glob walks

### LINKED ISSUE #97550: Turbopack: outputFileTracingIncludes glob walk hits ELOOP on pnpm workspaces with nested packages depending on their ancestor

https://github.com/Stanzilla/turbopack-symlink-cycle-repro

### To Reproduce

1. `pnpm install`
2. `pnpm --filter app exec next build`

The workspace layout that triggers it:

```
pnpm-workspace.yaml         packages: [app, packages/*, packages/umbrella/*]
app/                        Next.js app, output: standalone, depends on `shared`
  next.config.mjs           outputFileTracingRoot: workspace root
                            outputFileTracingIncludes: { '/*': ['included-asset.txt'] }
packages/umbrella/          package `umbrella`
packages/umbrella/shared/   package `shared`, devDependencies: { umbrella: workspace:* }  <-- depends on its physical ancestor
packages/umbrella/leagues/  package `leagues`, dependencies: { shared: workspace:* }
```

Both conditions are necessary:

- at least one `outputFileTracingIncludes` entry (any pattern), and
- a workspace package that is physically nested inside another workspace package and declares that ancestor as a dependency or devDependency.

The nested-ancestor dependency is legal in pnpm workspaces. pnpm plants `packages/umbrella/shared/node_modules/umbrella -> ../..`, which makes the directory tree self-embedding: `.../shared/node_modules/umbrella/leagues/node_modules/shared/node_modules/umbrella/...` repeats forever.

**Current:** the build crashes with a `TurbopackInternalError`:

```
Error [TurbopackInternalError]: Failed to write app endpoint /page

Caused by:
- reading dir ".../app/node_modules/shared/node_modules/umbrella/leagues/node_modules/shared/node_modules/umbrella/leagues/[...repeats...]/node_modules/shared"
- Too many levels of symbolic links (os error 62)

Debug info:
- Execution of *NftJsonAsset::new failed
- Execution of trace_endpoint failed
- Execution of read_glob failed
- Execution of read_glob_inner failed   (repeated ~90 times)
- Execution of read_dir failed
- Execution of <DiskFileSystem as FileSystem>::raw_read_dir failed
- Too many levels of symbolic links (os error 62)
```

(os error 40 on Linux.)

**Expected:** the build completes. The include-glob walk should detect the symlink cycle and skip it, the same way most file walkers do.

### Analysis

Any `outputFileTracingIncludes` entry makes `get_glob_includes` in `crates/next-api/src/nft.rs` call `project_root_path.read_glob(...)`, which walks the entire tracing root because the glob is compiled with `GlobOptions { contains: true, ... }`.

The cycle detection in `resolve_symlink_safely()` in `turbopack/crates/turbo-tasks-fs/src/read_glob.rs` only detects a symlink whose target is an ancestor of the symlink's own virtual path. In this repro the walk enters the cycle through `app/node_modules/shared` (or the hoisted `node_modules/.pnpm/node_modules/shared`), so the virtual path is `app/node_modules/shared/node_modules/umbrella/...` and neither symlink target (`packages/umbrella/shared`, `packages/umbrella`) is ever an ancestor of the virtual path. The check never fires and the walk recurses until the OS returns `ELOOP`.

Two related problems:

1. Multi-symlink cycles evade `resolve_symlink_safely()` entirely (this repro).
2. Even when the ancestor check does fire, it `bail!`s and fails the whole build ("'{...}' is a symlink causes that causes an infinite loop!") instead of skipping the cyclic entry.

`outputFileTracingExcludes` cannot work around this: in `nft_json.rs` the exclude glob filters traced assets in a separate code path and is never applied to the include-glob directory walk.

Real-world impact: `@sentry/nextjs` >= 10.70 unconditionally merges its own entries into `outputFileTracingIncludes` (getsentry/sentry-javascript#23055), so any pnpm monorepo with this (legal) nested-workspace layout now fails to build after a routine Sentry upgrade, with no config escape hatch. We hit this in production on a ~220-package monorepo.

Reproduced on 16.3.0, 16.3.1, and 16.3.1-canary.22. `read_glob.rs` on canary is byte-identical to 16.3.0, so the latest canary is affected as well.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 36864
  Available CPU cores: 12
Binaries:
  Node: 24.19.0
  npm: 11.17.0
  Yarn: N/A
  pnpm: 11.22.0
Relevant Packages:
  next: 16.3.1
  eslint-config-next: N/A
  react: 19.2.3
  react-dom: 19.2.3
  typescript: N/A
Next.js Config:
  output: standalone
```

Turbopack, Output (export/standalone)

next build (local), Other (CI)

### Additional context

A panic log URL is emitted (`https://bugs.nextjs.org/...category=turbopack-error-report...`), but this is a deterministic logic gap rather than a transient panic, hence the full report with a standalone reproduction.

ahmashadani:
## Summary
Fixes #97550

Implemented with Midfleet infrastructure.

## Problem
`outputFileTracingIncludes` triggers `project_root_path.read_glob(...)` with a contains glob. In pnpm monorepos where a nested workspace package depends on its physical ancestor, directory symlinks form a multi-hop cycle. The old check only caught single-hop “target is ancestor of source” loops, so the walk recursed until `ELOOP`.

Even when a cycle was detected, the walk `bail!`ed and failed the whole build instead of skipping the cyclic entry.

## Fix
In `turbopack/crates/turbo-tasks-fs/src/read_glob.rs`:
1. Track **realpath** of directories entered on the current walk; skip when the same real dir is seen again (covers multi-hop cycles).
2. On single-hop ancestor cycles, **skip** the entry instead of failing the walk.
3. Apply the same behavior to `track_glob`.

## Tests
- Updated loop tests to expect skip (not hard fail)
- Added `read_glob_multi_hop_symlink_cycle` reproducing the nested umbrella/shared/leagues layout from the issue

## Test plan
- [ ] `cargo test -p turbo-tasks-fs --lib read_glob`
- [ ] Repro: https://github.com/Stanzilla/turbopack-symlink-cycle-repro with patched Next
- [ ] CI turbopack / next build jobs

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/read_glob.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/read_glob.rs
@@ -1,7 +1,7 @@
-use anyhow::{Result, bail};
+use anyhow::Result;
 use futures::try_join;
 use rustc_hash::FxHashMap;
-use turbo_rcstr::RcStr;
+use turbo_rcstr::{RcStr, rcstr};
 use turbo_tasks::{Completion, ResolvedVc, TryJoinIterExt, Vc, turbobail};
 
 use crate::{
@@ -21,7 +21,7 @@ pub struct ReadGlobResult {
 /// on the order.
 #[turbo_tasks::function(fs)]
 pub async fn read_glob(directory: FileSystemPath, glob: Vc<Glob>) -> Result<Vc<ReadGlobResult>> {
-    read_glob_internal("", directory, glob).await
+    read_glob_internal("", directory, glob, rcstr!("")).await
 }
 
 #[turbo_tasks::function(fs)]
@@ -29,8 +29,11 @@ async fn read_glob_inner(
     prefix: RcStr,
     directory: FileSystemPath,
     glob: Vc<Glob>,
+    /// Real filesystem paths already entered on this walk (newline-separated).
+    /// Used to skip multi-hop symlink cycles (e.g. pnpm nested workspace links).
+    visited_real_dirs: RcStr,
 ) -> Result<Vc<ReadGlobResult>> {
-    read_glob_internal(&prefix, directory, glob).await
+    read_glob_internal(&prefix, directory, glob, visited_real_dirs).await
 }
 
 // The `prefix` represents the relative directory path where symlinks are not resolve.
@@ -38,7 +41,15 @@ async fn read_glob_internal(
     prefix: &str,
     directory: FileSystemPath,
     glob: Vc<Glob>,
+    visited_real_dirs: RcStr,
 ) -> Result<Vc<ReadGlobResult>> {
+    // Skip if we already walked this real directory (multi-hop symlink cycles).
+    let real_dir = directory_real_key(&directory).await?;
+    if visited_contains(&visited_real_dirs, &real_dir) {
+        return Ok(ReadGlobResult::cell(ReadGlobResult::default()));
+    }
+    let next_visited = visited_push(&visited_real_dirs, &real_dir);
+
     let dir = directory.read_dir().await?;
     let mut result = ReadGlobResult::default();
     let glob_value = glob.await?;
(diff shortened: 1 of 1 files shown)


dianatofficial (commented):

PR #97621, opened 2026-08-20
docs: document React browser API

aurorascharff:
### What?

- Documents React Canary's new `browser` API in the SPA guide
- Adds examples for marking a Client Component as browser-only and wrapping it in Suspense

### Why?

The App Router includes React Canary releases, and the `browser` API now provides a Suspense-based option for components that depend on browser APIs such as `window` or `document`.

### How?

The guide calls `browser()` inside `use()` in a Client Component. It then wraps that component in Suspense so prerendering can show the closest fallback before the component renders in the browser.

### Testing

- `prettier --check docs/01-app/02-guides/single-page-applications.mdx`
- `alex docs/01-app/02-guides/single-page-applications.mdx --quiet` (only existing SPA dictionary warnings)

### DIFF
--- a/docs/01-app/02-guides/single-page-applications.mdx
+++ b/docs/01-app/02-guides/single-page-applications.mdx
@@ -218,6 +218,37 @@ This can be useful for third-party libraries that rely on browser APIs like `win
 
 See the [live demo](https://next-spa-patterns.labs.vercel.dev/browser-only) and its [source code](https://github.com/vercel-labs/next-spa-patterns/tree/main/app/browser-only).
 
+The App Router uses React Canary releases, which include the new [`browser`](https://react.dev/reference/react-dom/browser) API. Call `browser` inside `use` in a Client Component that needs browser APIs:
+
+```tsx filename="app/browser-only-editor.tsx"
+'use client'
+
+import { use } from 'react'
+import { browser } from 'react-dom'
+
+export default function BrowserOnlyEditor() {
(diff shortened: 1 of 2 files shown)


dianatofficial (commented):

docs/01-app/02-guides/single-page-applications.mdx
```diff
@@ -216,7 +216,38 @@ const ClientOnlyComponent = dynamic(() => import('./component'), {
...
+export default function Page() {
+  return (
+    <Suspense fallback={<p>Loading editor...</p>}>
+      <BrowserOnlyEditor />
+    </Suspense>
+  )
+}
+```
+
+During prerendering, `use(browser())` suspends to the closest `<Suspense>` boundary. The component then renders normally in the browser.
```
icyJoseph: Should we expand a bit more on why one shouldn't start to add `use client` and `use(browser())` everywhere — this API is for rare cases in which the HTML served cannot match the first frame in the browser, and you compromise by including fallback UI instead. 

So it could potentially impact your SEO — as seen by bots that do not execute JS.

aurorascharff: :robot: Clarified that use(browser()) is for Client Components whose first browser render cannot match the server-rendered HTML. The guide now explains that crawlers without JavaScript see the Suspense fallback, keeps indexable content outside that boundary, and uses the same saved-draft example as the demo.

icyJoseph (approved):

PR #97622, opened 2026-08-20
[PPF] unstable_prefetch()

lubieowoce:
Implements `unstable_prefetch()`, which is intended for use in `partialPrefetching`. `await unstable_prefetch()` excludes content from the app shell -- it will only be available when using `prefetch={true}` (speculative prefetch) or during navigations.

As a rule of thumb, `unstable_prefetch()` resolves whenever static `params` would:
- in a static prerender
  - but NOT the app shell extracted from it, which is param-less
- in a runtime prefetch (`prefetch={true}`)
  - but NOT a runtime app shell, which is param-less

Note that `prefetch()` is URL data, so using it in an App Shell without Suspense will trigger an instant insight.

### Implementation notes

`prefetch()` is treated like URL data, so it resolves in the `PrefetchStatic/PrefetchRuntime` stages added in #96908. The implementation is basically analogous to `unstable_navigation()` except using different stages. i've considered abstracting them into one implementation, but decided against that for now, we can deduplicate later.

Error messages about URL data have not been updated to mention it yet -- we will do that as a follow up, along with docs. 

`await prefetch()` does not count as a runtime data access, meaning that it won't affect the static prefetch hint for a route. however `await prefetch(); await cookies()` does deopt the route, because using a speculative runtime prefetch would reveal more content. Note that this may cause us to unnecessarily deopt a shell to runtime even if only the speculative part of the content would be improved by a runtime request; this is not a new issue, but it's something we should optimize.


packages/next/src/server/request/cache-stages.ts
```diff
@@ -24,7 +25,128 @@ import { RenderStage } from '../app-render/staged-rendering'
...
+    throw new Error(
+      `Route ${workStore.route} used \`unstable_prefetch()\`, which requires Cache Components to be enabled. Learn more: https://nextjs.org/docs/app/api-reference/config/next-config-js/cacheComponents`
+    )
+  }
+
+  if (!isRequestApiAllowedInCurrentPhase(workUnitStore)) {
+    throw new Error(
+      `Route ${workStore.route} used \`unstable_prefetch()\` inside \`after()\` while rendering. The \`unstable_prefetch()\` function is used to indicate the subsequent code must not run in the app shell, but \`after()\` executes after the request, so this function is not allowed in this scope. See more info here: https://nextjs.org/docs/app/api-reference/functions/after`
+    )
+  }
```
lubieowoce, line 45: we don't let you do `after(() => { await cookies() })` because it should create be a dynamic hole in the static prerender but we wouldn't know it because `after` runs after rendering.

now, `navigation()` and `prefetch()` both resolve just fine during static prerenders, so technically we could consider allowing this... but it gets weird in runtime prerenders again. i'm not sure what the semantics of `after(async () => { await prefetch() })` would be in an app shell -- if the page was completely prerendereable as a runtime shell, then that `after` would never run, which feels sketchy. and we wouldn't even find out until the call until after we've already sent the response including the `isPartial` byte indicating that the page is complete, so we don't even get to signal "hey there's this `after` here, do a dynamic request when navigating"

we can revisit this later but my gut says not allowing these is the right call

unstubbable (approved):

PR #97625, opened 2026-08-20
Turbopack: add Module Federation remote consumption

sokra:
## What?

Adds first-class Module Federation configuration and webpack-compatible remote consumption to Turbopack, exposed in Next.js as `experimental.turbopackModuleFederation`.

This layer also fixes script externals to use a dedicated browser script loader. Plain webpack `remoteEntry.js` files do not register as Turbopack chunks, so using the chunk loader caused their load promises to remain pending.

## Why?

Turbopack needs a native container protocol rather than a webpack-style plugin abstraction. Establishing the normalized configuration, import-map integration, and script loader first gives subsequent producer and sharing layers a stable runtime boundary.

## How?

- Normalize and strictly validate webpack-compatible `remotes`, `exposes`, and `shared` forms.
- Map exact and slash-boundary remote imports to generated async proxy modules.
- Load global/script containers, initialize share scopes, call `get()`, execute factories, and expose module namespaces.
- Deduplicate browser script loads and preserve ordered remote fallbacks.
- Add an App Router e2e fixture consuming a real webpack `ModuleFederationPlugin` remote.

## Verification

- `cargo test -p turbopack --features turbopack-node/process_pool module_federation`
- `pnpm --filter=next types`
- `pnpm test-dev-turbo test/e2e/app-dir/turbopack-module-federation-webpack-remote/turbopack-module-federation-webpack-remote.test.ts`
- `pnpm test-start-turbo test/e2e/app-dir/turbopack-module-federation-webpack-remote/turbopack-module-federation-webpack-remote.test.ts`

## Checklist

- [x] e2e tests added
- [x] Errors are actionable

Stack: **1/6**. Followed by #97626.


crates/next-core/src/next_config.rs
```diff
@@ -1333,6 +1333,9 @@ pub struct ExperimentalConfig {
 
     turbopack_chunking: Option<TurbopackChunkingConfig>,
 
+    #[bincode(with = "turbo_bincode::serde_self_describing")]
+    turbopack_module_federation: Option<serde_json::Value>,
```
sokra: Instead of the generic serde_json::Value, use a structured object similar to `CssChunkingConfig` or `TurbopackChunkingConfig`

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -284,6 +285,36 @@ function loadChunkByUrl(
...
+const externalScriptCache = new Map<string, Promise<void>>()
+function loadScriptByUrl(url: string): Promise<void> {
+  let promise = externalScriptCache.get(url)
+  if (promise !== undefined) return promise
+
+  promise = new Promise<void>((resolve, reject) => {
+    if (typeof document === 'undefined') {
+      reject(new Error(`Cannot load external script ${url} without a document`))
+      return
+    }
```
sokra: In WebWorkers it should use `importScripts` to load the script. We also need the test case to cover consumption from WebWorker.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -284,6 +285,36 @@ function loadChunkByUrl(
 }
 browserContextPrototype.L = loadChunkByUrl
 
+const externalScriptCache = new Map<string, Promise<void>>()
+function loadScriptByUrl(url: string): Promise<void> {
```
sokra: There is similar logic in `doLoadChunk`. Make sure that both share the same logic. In case of conflicts the `doLoadChunk` logic should win. `loadScriptByUrl` should also gain the ability to attach to existing script tags like `doLoadChunk`.

turbopack/crates/turbopack-ecmascript-runtime/js/src/nodejs/runtime/runtime-base.ts
```diff
@@ -142,6 +142,21 @@ function loadChunkAsyncByUrl<TModule extends Module>(
...
+    }
+  ).__turbopack_test_load_script__
+  if (loader !== undefined) return loader(url)
+  return Promise.reject(
+    new Error(
+      `External script loading is only supported in browser runtimes: ${url}`
+    )
+  )
+}
+contextPrototype.o = loadScriptByUrl
```
sokra: We don't want this testing code `__turbopack_test_load_script__` in the common runtime.

Also `loadScriptByUrl` doesn't make sense for node.js.
Could module federation use `loadChunkAsyncByUrl` instead, instead of adding a new `loadScriptByUrl` method

turbopack/crates/turbopack/src/module_federation.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub share_key: RcStr,
+    pub share_scope: RcStr,
+    pub version: Option<RcStr>,
+    pub eager: bool,
+    pub singleton: bool,
+    pub strict_version: bool,
+}
+
+impl ModuleFederationConfig {
+    pub fn from_json(json: &str) -> Result<Self> {
```
sokra: Instead of doing this, have a `UnnormalizedModuleFederationConfig` struct that it deserialized by serde in NextConfig

turbopack/crates/turbopack/src/module_federation.rs
```diff
@@ -0,0 +1,612 @@
...
+                );
+            }
+            for external in &remote.external {
+                if external.global.is_empty() {
+                    bail!(
+                        "Module Federation remote '{}' has an empty global name",
+                        remote.request
+                    );
+                }
+                if !(external.url.starts_with("http://") || external.url.starts_with("https://")) {
```
sokra: This validation is too strict. `file://` would also be valid. Also server-relative URLs are valid.
Drop this URL validation.

turbopack/crates/turbopack/src/module_federation.rs
```diff
@@ -0,0 +1,612 @@
...
+    }
+    let name_json = serde_json::to_string(name)?;
+    let source = format!(
+        r#"
+const moduleMap = {{
+  {module_entries}
+}};
+let initializedScope;
+function get(request) {{
+  const loader = moduleMap[request];
```
sokra: Check "hasOwnProperty"

turbopack/crates/turbopack/src/module_federation.rs
```diff
@@ -0,0 +1,612 @@
...
+const failures = [];
+let federatedModule;
+for (const [globalName, url] of candidates) {{
+  try {{
+    await __turbopack_load_script__(url);
+    const container = globalThis[globalName];
+    if (!container) {{
+      throw new Error(`Container global ${{globalName}} is missing after loading ${{url}}`);
+    }}
+    const scopes = globalThis.__turbopack_module_federation_share_scopes__ ||= Object.create(null);
```
sokra: `globalThis.__turbopack_module_federation_share_scopes__`

This should not be a global object. It should be scoped to the runtime and therefore exposed by the runtime under a new property on the `__turbopack_context__`.

turbopack/crates/turbopack/src/module_federation.rs
```diff
@@ -0,0 +1,612 @@
...
+    const scope = scopes[{share_scope}] ||= Object.create(null);
+    {registrations}
+    const initScope = [];
+    await container.init(scope, initScope);
```
sokra: The part the loads and initializes the remote should be in a separate virtual module.

turbopack/crates/turbopack/src/module_federation.rs
```diff
@@ -0,0 +1,612 @@
...
+            remote: remote.clone(),
+            shared: config.shared.clone(),
+            host_name: config.name.clone().unwrap_or_else(|| "host".into()),
+        }
+        .resolved_cell();
+        let mapping = ImportMapping::Dynamic(ResolvedVc::upcast(replacer)).resolved_cell();
+        import_map.insert_exact_alias(remote.request.clone(), mapping);
+        import_map.insert_wildcard_alias(RcStr::from(format!("{}/", remote.request)), mapping);
+    }
+}
```
sokra: shared modules can also be used from the host. In this case it need to load and initialized all remotes that share that module, as they might provide a higher version of the shared module, which would be used in that case.

So shared modules also need to register import mappings and have a wrapper module which takes care of loading (potentially falling back to the host import)

turbopack/crates/turbopack/src/module_federation/config.rs
```diff
@@ -0,0 +1,612 @@
...
+    pub request: RcStr,
+    pub import: Option<RcStr>,
+    pub package_name: Option<RcStr>,
+    pub required_version: Option<RcStr>,
+    pub share_key: RcStr,
+    pub share_scope: RcStr,
+    pub version: Option<RcStr>,
+    pub eager: bool,
+    pub singleton: bool,
+    pub strict_version: bool,
```
sokra, line 297: Some config keys are not yet supported in this PR and support will only be added in a later PR. In these cases, the config should not contain these keys, so providing them will not compile (and silently ignore these unsupported keys). Any unsupported feature should fail to compile.

test/e2e/app-dir/turbopack-module-federation-webpack-remote/turbopack-module-federation-webpack-remote.test.ts
```diff
@@ -62,16 +69,20 @@ describeTurbopack('turbopack module federation with a webpack remote', () => {
...
     const remotePort = await findPort()
     const remoteOutput = join(next.testDir, 'remote-dist')
-    await buildRemote(join(next.testDir, 'remote'), remoteOutput)
+    const remoteContext = join(next.testDir, 'remote')
+    await buildRemote(remoteContext, join(remoteOutput, 'browser'))
+    await buildRemote(remoteContext, join(remoteOutput, 'worker'), true)
     remoteServer = await startStaticServer(remoteOutput, undefined, remotePort)
-    process.env.MF_REMOTE_URL = `http://localhost:${remotePort}/remoteEntry.js`
+    process.env.MF_REMOTE_ORIGIN = `http://localhost:${remotePort}`
+    process.env.MF_REMOTE_URL = '/webpack-remote/remoteEntry.js'
```
sokra: Use MF urls from a different host. Starting with `http://localhost:${remotePort}` and not `/webpack-remote`

test/e2e/app-dir/turbopack-module-federation-webpack-remote/next.config.js
```diff
@@ -5,17 +5,29 @@ const nextConfig = {
...
+      {
+        source: '/webpack-remote/:path*',
+        destination: `${process.env.MF_REMOTE_ORIGIN}/browser/:path*`,
+      },
+      {
+        source: '/webpack-worker-remote/:path*',
+        destination: `${process.env.MF_REMOTE_ORIGIN}/worker/:path*`,
+      },
+    ]
+  },
```
sokra: Avoid these rewrites. Remotes should be consumed from a different origin

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -198,16 +226,33 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
       return resolver.promise
     }
 
-    if (typeof importScripts === 'function') {
-      // We're in a web worker
+    if (typeof document === 'undefined') {
+      // We're in a web worker. Classic workers support importScripts; module workers use a
+      // native dynamic import for external scripts because importScripts is unavailable there.
```
sokra: `typeof importScripts === 'function'` detection for web workers should be used. Module workers should not be supported here.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -252,11 +297,18 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
           document.head.appendChild(createLink())
         }
       } else if (isJs(chunkUrl)) {
-        const previousScripts = document.querySelectorAll(
-          `script[src="${chunkUrl}"],script[src^="${chunkUrl}?"],script[src="${decodedChunkUrl}"],script[src^="${decodedChunkUrl}?"]`
-        )
+        const previousScripts = getExistingScripts(chunkUrl)
         if (previousScripts.length > 0) {
           for (const script of Array.from(previousScripts)) {
+            if (resolveOnLoad) {
```
sokra: I think the `if (resolveOnLoad)` need to be removed here, as the same script might be requested with `resolveOnLoad` later and `registrationRequired` would be set later.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/dom/runtime-backend-dom.ts
```diff
@@ -165,17 +177,33 @@ const chunkResolvers: Map<ChunkUrl, ChunkResolver> = new Map()
...
    * has been loaded.
    */
-  function doLoadChunk(sourceType: SourceType, chunkUrl: ChunkUrl) {
-    const resolver = getOrCreateResolver(chunkUrl)
+  function doLoadChunk(
+    sourceType: SourceType,
+    chunkUrl: ChunkUrl,
+    resolveOnLoad = false
+  ) {
+    const resolver = getOrCreateResolver(chunkUrl, resolveOnLoad)
```
sokra: I think it would make sense to have 2 promises in the resolver, one for resolving on load and one for resolving when the chunk has fully loaded. Promise should be lazily created.

turbopack/crates/turbopack/src/module_federation/config.rs
sokra, line 1: This file is big. Split it into multiple files in a `module_federation` folder.

test/e2e/app-dir/turbopack-module-federation-webpack-remote/app/remote-message.tsx
sokra, line 1: We also want to test consuming a "shared" from the host. So directly from here.
And we want to test consuming another "shared" which is in a higher version in the remote. The remote version should be loaded in this case.

turbopack/crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -190,14 +190,14 @@ impl CachedExternalModule {
                     // First load the URL
```
sokra: Before loading the URL, check the global. Maybe it's already available. In this case you don't need to load the script at all. It's probably already loaded via some alternative way.

turbopack/crates/turbopack-tests/tests/execution/turbopack/runtime/module-federation-state/input/index.js
```diff
@@ -0,0 +1,51 @@
...
+  runtime.c(chunkPath)
+  return runtime.m(moduleId).exports
+}
+
+it('isolates module federation state between runtimes', () => {
+  const repoRoot = path.resolve(process.cwd(), '../../../../../../../..')
+  const runtimePath = path.join(
+    repoRoot,
+    'turbopack/crates/turbopack-tests/tests/snapshot/runtime/default_build_runtime/output/[turbopack]_runtime.js'
+  )
```
bgw, line 23: Can we not use `__dirname` here?

turbopack/crates/turbopack/src/module_federation/container.rs
```diff
@@ -0,0 +1,93 @@
...
+    let mut module_entries = Vec::new();
+    for expose in &config.exposes {
+        let imports = expose
+            .imports
+            .iter()
+            .map(|request| format!("import({})", serde_json::to_string(request).unwrap()))
+            .collect::<Vec<_>>()
+            .join(", ");
+        module_entries.push(format!(
+            "{}: () => Promise.all([{}]).then((modules) => () => modules[modules.length - 1])",
```
bgw, line 27: Use `StringifyJs`?

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-types.d.ts
```diff
@@ -143,6 +152,7 @@ interface TurbopackBaseContext<M> {
   m: Module
   c: ModuleCache<M>
   M: ModuleFactories
+  S: ModuleFederationRuntimeState
```
sokra: Use `R` instead of `S`

turbopack/crates/turbopack/src/module_federation/container.rs
```diff
@@ -0,0 +1,93 @@
...
+  versions_{index}[{version}] ||= {{
+    get: () => import({import}).then((module) => () => module),
+    from: {name},
+    eager: {eager}
+  }};"#,
+            index = registrations.len(),
+            key = serde_json::to_string(&shared.share_key)?,
+            version = serde_json::to_string(version)?,
+            import = serde_json::to_string(import)?,
+            name = serde_json::to_string(name)?,
```
sokra: StringifyJs

turbopack/crates/turbopack/src/module_federation/remote.rs
```diff
@@ -0,0 +1,266 @@
...
+    versions_{index}[{version}] ||= {{
+      get: () => import({import}).then((module) => () => module),
+      from: {host_name},
+      eager: {eager}
+    }};"#,
+            index = registrations.len(),
+            key = serde_json::to_string(&shared.share_key)?,
+            version = serde_json::to_string(version)?,
+            import = serde_json::to_string(import)?,
+            host_name = serde_json::to_string(host_name)?,
```
sokra: StringifyJs

turbopack/crates/turbopack/src/module_federation/remote.rs
```diff
@@ -0,0 +1,266 @@
...
+        let candidates = serde_json::to_string(
+            &this
+                .remote
+                .external
+                .iter()
+                .map(|external| (&*external.global, &*external.url))
+                .collect::<Vec<_>>(),
+        )?;
+        let remote_key = serde_json::to_string(&this.remote.request)?;
+        let share_scope = serde_json::to_string(&this.remote.share_scope)?;
```
sokra: StringifyJs

ScriptedAlchemy:
@sokra Can you please consider implementing the v2 runtime like we have done in rspack? Most users no longer use webpacks implementation. Rspack still maintains support for v1 under a different named export. The implementation differences on the plugin side should be minimal. 
We can accommodate anything needed for turbo within the runtime. Most of my ecosystem will not work with v1 implementation from webpack.

ScriptedAlchemy:
I'm happy to send a pr on top of this one. We can support both. If you've added the underlaying mechanics for v1 then it I'd have what I need to send pr. Would also be a great opportunity for me to improve rspacks implementation and the v2 runtime to bring it up to your standards if it falls short. 🙏💕 I'm willing to do just about anything possible to not fragment the ecosystem. We have finally got import maps, vite, rollup, webpack and rspack, metro all with implementations of v2.

Nsttt:
Very happy to see this moving after so long :)

VugarAhmadov:
We need ModuleFederation v2 support, please 🙏

gioboa (commented):
It’s great to see progress on this highly requested feature. 👏

dmchoi77:
I've been waiting for Turbopack to support Module Federation v2.

HashemKhalifa:
Great work, but why not MF V2?

jame-earnin:
The only blocker for us to use Turbopack is only the Module Federation, and why not use v2 (latest) implementation? right?

Jtutne:
Please support MF v2! 🙏

neardark:
Also hoping to have v2 support.

dandigangi:
LFG! Great work.

ron23:
Please support MF v2 as well 🙏

shashank-u03:
Waiting for Turbopack to support Module Federation v2.

PR #97626, opened 2026-08-20
Next.js: emit Turbopack Module Federation containers

sokra:
## What?

Adds project-level Module Federation container emission for Next.js applications using Turbopack.

A configured producer serves its global container from `/_next/static/chunks/<filename>` in development and production, allowing another Next.js/Turbopack app to consume it.

## Why?

A federation container is a project artifact, not a route artifact. Attaching an independent container runtime to route client assets contaminates prerender endpoints and can introduce duplicate React runtimes. Emitting it at the project boundary keeps it out of route manifests and prerender execution.

## How?

- Generate the container source with the App client module context.
- Build it with an isolated browser chunking context and container-specific chunk-loading global.
- Exclude route-scoped module IDs, export usage, unused-reference data, and shared runtime state.
- Add a two-application App Router fixture where one Next.js app exposes and another consumes a client module.

## Verification

- `cargo check -p next-api --features turbopack-node/process_pool`
- `pnpm --filter=next types`
- `pnpm test-dev-turbo test/e2e/app-dir/turbopack-module-federation-next-remote/turbopack-module-federation-next-remote.test.ts`
- `pnpm test-start-turbo test/e2e/app-dir/turbopack-module-federation-next-remote/turbopack-module-federation-next-remote.test.ts`

## Checklist

- [x] e2e tests added
- [x] Errors are actionable

Stack: **2/6**. Depends on #97625; followed by #97627.


dianatofficial (commented):

crates/next-api/src/project.rs
```diff
@@ -2958,8 +2965,7 @@ async fn any_output_changed(
 
 #[turbo_tasks::function(operation, root)]
 fn all_assets_from_entries_operation(
-    operation: OperationVc<OutputAssets>,
+    entries: ResolvedVc<OutputAssets>,
```
sokra: This must stay an operation argument

crates/next-api/src/project.rs
```diff
@@ -2426,7 +2426,14 @@ impl Project {
...
         async move {
-            let all_output_assets = all_assets_from_entries_operation(output_assets);
+            let mut entries = output_assets
+                .connect()
+                .await?
+                .iter()
+                .copied()
+                .collect::<Vec<_>>();
+            entries.extend(module_federation_output_assets(self).await?.iter().copied());
+            let all_output_assets = all_assets_from_entries_operation(ResolvedVc::cell(entries));
```
sokra: These output assets shouldn't be added here. Instead module federation should be a separate global Endpoint, similar to instrumentation, `_error`, `_app`, `not-found`

crates/next-api/src/app.rs
```diff
@@ -100,6 +107,81 @@ pub struct AppProject {
...
+    let client_chunking_context =
+        ResolvedVc::try_downcast_type::<BrowserChunkingContext>(client_chunking_context)
+            .context("expected a browser chunking context")?;
+    // The container is a project-global artifact, not a route: it gets its own runtime namespace
+    // so it can coexist with the host's runtime and with other containers on the same page.
+    let federation_chunking_context = client_chunking_context
+        .await?
+        .clone_builder()
+        .asset_suffix(AssetSuffix::None.resolved_cell())
+        .without_module_id_strategy()
```
sokra: production module ids should also work in the module federation bundle

crates/next-api/src/app.rs
```diff
@@ -100,6 +107,81 @@ pub struct AppProject {
...
+        .export_usage(None)
+        .without_unused_references()
```
sokra: Thee Shaking should also work in the module federation bundle

crates/next-api/src/app.rs
```diff
@@ -100,6 +107,81 @@ pub struct AppProject {
...
+        .shared_runtime(false)
+        .shared_runtime_chunk(false)
+        .chunk_loading_global(format!("TURBOPACK_{name}").into())
+        .single_chunk()
```
sokra: Code Splitting should also work in the module federation bundle

crates/next-api/src/app.rs
```diff
@@ -100,6 +107,81 @@ pub struct AppProject {
...
+        .await?
+        .build();
```
sokra: When creating a new chunking context, you must emit the output files to a separate output directory (not static/chunks) otherwise files will potentially conflict with the chunks of the normal pages. They are slightly different.

Maybe use `static/chunks/mf` instead

sokra: Looks like this is related to the next PR in the stack

crates/next-api/src/app.rs
sokra: Looks like this is related to the next PR in the stack

packages/next/src/server/dev/turbopack-utils.ts
```diff
@@ -648,6 +648,24 @@ export async function handleEntrypoints({
...
+      key,
+      /** includeIssues=*/ false,
+      entrypoints.moduleFederation,
+      () => ({ type: HMR_MESSAGE_SENT_TO_BROWSER.CLIENT_CHANGES }),
+      (error) => ({
+        type: HMR_MESSAGE_SENT_TO_BROWSER.RELOAD_PAGE,
+        data: `error in Module Federation subscription: ${error}`,
+      })
+    )
+  }
```
sokra: Move this down in this function, next to middleware and instrumentation

crates/next-api/src/app.rs
```diff
@@ -100,6 +107,154 @@ pub struct AppProject {
 #[turbo_tasks::value(transparent)]
 pub struct OptionAppProject(Option<ResolvedVc<AppProject>>);
 
+#[turbo_tasks::value]
+pub struct ModuleFederationEndpoint {
```
sokra: Move that into a separate file

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -154,6 +154,11 @@ impl BrowserChunkingContextBuilder {
         self
     }
 
+    pub fn without_module_id_strategy(mut self) -> Self {
+        self.chunking_context.module_id_strategy = None;
+        self
+    }
```
sokra: remove as unused

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -164,6 +169,11 @@ impl BrowserChunkingContextBuilder {
         self
     }
 
+    pub fn without_unused_references(mut self) -> Self {
+        self.chunking_context.unused_references = None;
+        self
+    }
```
sokra: remove as unused

crates/next-api/src/app.rs
```diff
@@ -100,6 +107,154 @@ pub struct AppProject {
...
+            .await?
+            .build();
+        let EntryChunkGroupResult { asset, .. } = *federation_chunking_context
+            .entry_chunk_group(
+                this.project
+                    .node_root()
+                    .owned()
+                    .await?
+                    .join("static/chunks/mf")?
+                    .join(&filename)?,
```
sokra: A file without contenthash must not be in the `static/chunks/` directory since we apply `Cache-Control: immutable` for this directory.
The entry file must be in `static/` directly instead.

PR #97627, opened 2026-08-20
Turbopack: lazy-load Module Federation exposes

sokra:
## What?

Adds lazy exposed chunks for Turbopack Module Federation containers and proves a real webpack host can consume a Turbopack-produced remote.

## Why?

A remote entry should remain a stable bootstrap rather than eagerly containing every exposed module. Lazy chunks must also resolve from the producer's origin when the host runs on another origin.

## How?

- Add a `single_entry_chunk` mode that keeps the synchronous entry in one configured file while preserving async chunk groups.
- Derive the container chunk base from the executing remote entry, including nested filenames.
- Keep each container on a separate chunk-loading namespace.
- Extend the two-app fixture to assert a non-entry chunk is fetched from the producer.
- Build a webpack `ModuleFederationPlugin` host that consumes the same Turbopack remote.

## Verification

- `pnpm test-dev-turbo test/e2e/app-dir/turbopack-module-federation-next-remote/turbopack-module-federation-next-remote.test.ts`
- `pnpm test-start-turbo test/e2e/app-dir/turbopack-module-federation-next-remote/turbopack-module-federation-next-remote.test.ts`
- webpack-remote regression fixture

## Checklist

- [x] e2e tests added
- [x] Cross-origin lazy loading covered

Stack: **3/6**. Depends on #97626; followed by #97628.


dianatofficial (commented):

turbopack/crates/turbopack-browser/src/chunking_context.rs
```diff
@@ -277,6 +297,24 @@ impl BrowserChunkingContextBuilder {
...
+            ecmascript_ty,
+            ChunkingConfig {
+                min_chunk_size: usize::MAX,
+                max_chunk_count_per_group: 1,
+                max_merge_chunk_size: usize::MAX,
+                ..Default::default()
+            },
+        ));
+        Ok(self)
+    }
```
sokra: This is weird. Do not try to force merging chunks via chunking config.

Instead expose `entry_chunk_group` for the browser chunking context, which creates a special chunk, which is the concatenation of all chunks usually needed for this chunk group. In this case also avoid emitting the output assets for the chunk group.

turbopack/crates/turbopack-ecmascript-runtime/js/src/browser/runtime/base/runtime-base.ts
```diff
@@ -37,10 +37,25 @@ interface TurbopackBrowserBaseContext<M> extends TurbopackBaseContext<M> {
...
+    : moduleFederationChunkBaseDepth !== undefined &&
+        typeof document !== 'undefined' &&
+        document.currentScript?.src
+      ? new URL(
+          moduleFederationChunkBaseDepth === 0
+            ? './'
+            : '../'.repeat(moduleFederationChunkBaseDepth),
+          document.currentScript.src
+        ).href
+      : CHUNK_BASE_PATH
```
sokra: Why do we need special logic for module federation here?

PR #97628, opened 2026-08-20
Turbopack: consume Module Federation share scopes

sokra:
## What?

Adds shared-module consumption to Turbopack's webpack-compatible Module Federation share scopes.

## Why?

Remote consumption alone is insufficient for React and other singleton dependencies. Hosts need to select compatible providers, enforce strict versions, and fall back to local modules using the same observable share-scope protocol as webpack containers.

## How?

- Route configured shared imports through generated consumer proxies.
- Select the highest compatible provider by scope and share key.
- Support singleton and strict-version behavior, `import: false`, and local fallback aliases.
- Preserve the existing Turbopack-provider-to-webpack-consumer direction.
- Extend the webpack fixture with remote provider selection, a lower local version, strict mismatch, and pre-init fallback coverage.

## Verification

- `cargo test -p turbopack --features turbopack-node/process_pool module_federation`
- `pnpm test-dev-turbo test/e2e/app-dir/turbopack-module-federation-webpack-remote/turbopack-module-federation-webpack-remote.test.ts`
- `pnpm test-start-turbo test/e2e/app-dir/turbopack-module-federation-webpack-remote/turbopack-module-federation-webpack-remote.test.ts`
- Next-to-Next regression fixture

## Checklist

- [x] e2e tests added
- [x] Strict failure and fallback behavior covered

Stack: **4/6**. Depends on #97627; followed by #97629.


dianatofficial (commented):

PR #97629, opened 2026-08-20
Turbopack: expand Module Federation sharing compatibility

sokra:
## What?

Expands Module Federation sharing compatibility with common semver ranges, inferred requirements, trailing-slash share keys, and eager local providers.

## Why?

Real webpack configurations rely on more than exact versions. They commonly infer package constraints, share package subpaths, and combine local and remote providers under singleton constraints.

## How?

- Add prerelease-aware version ordering and support comparators, caret/tilde, partial and x ranges, unions, and hyphen ranges.
- Infer `requiredVersion` from the project package manifest using webpack's dependency-field precedence.
- Preserve subpaths through trailing-slash consumer and fallback aliases.
- Keep eager local fallback synchronous.
- Add fixture coverage for range selection, strict mismatch, prefix fallback, and eager import behavior.

## Verification

- `pnpm build-all`
- `pnpm --filter=next types`
- webpack federation fixture in development and production

## Checklist

- [x] e2e tests added
- [x] Common sharing compatibility cases covered

Stack: **5/6**. Depends on #97628; followed by #97630.


dianatofficial (commented):

PR #97635, opened 2026-08-20
[backport] [PPF] Do not mark complete shell requests as partial

lubieowoce:
Backports #97503 ("[PPF] Do not mark complete shell requests as partial") to `next-16-3`.

## Adaptation

The cherry-pick did not apply mechanically. `next-16-3` predates the refactor stack that landed on canary between #96406 and #96878, so none of the functions the original patch touches (`writeResponsePayloadsIntoCache`, `isFullResponsePartial`, `shellWasRequested`) exist on this branch. The behavior change is hand-ported instead; the test changes applied as-is.

On canary the shell-tier and full-payload writes are separate calls, so the fix was simply to stop overriding `isFullResponsePartial` at two write sites. On `next-16-3` both writes funnel through a single `isResponsePartial` in `fetchSegmentPrefetchesUsingDynamicRequest`, which unconditionally forces `true` for any `RuntimeShell` request. Copying the canary expression directly would also have un-marked the one case where forcing partial is correct: when the payload fulfilling the spawned entries is an App Shell extracted from a larger response, it is a strict subset and partial by construction regardless of the server marker.

So this adds a `spawnedEntriesGetExtractedShell` flag alongside the existing `staleAtForSpawnedEntries`, set only in that branch, reproducing canary's post-PR behavior exactly:

| case | full payload | shell payload |
| --- | --- | --- |
| no shell extractable | server marker | — |
| shell is the whole response | server marker | — |
| shell is a strict prefix | server marker | hardcoded `true` |

## Verification

- `pnpm test-start-turbo test/e2e/app-dir/segment-cache/` — 40 suites passed, 3 skipped, 202 tests passed, 16 snapshots. Run across the whole directory rather than just the four suites the patch touches, since the hand-port changes cache keying and could regress unrelated prefetch paths.
- Not run: `pnpm build-all` (`@next/swc#build-native-auto` fails locally — the installed `napi` CLI rejects the `--cargo-cwd` flag this branch passes. Unrelated to this change; the change is client-side TypeScript only and was verified against a JS-only `pnpm --filter=next build`.)

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -3165,6 +3165,13 @@ export async function fetchSegmentPrefetchesUsingDynamicRequest(
     // staleAt that corresponds to whatever payload the spawned entries get
     // filled with below.
     let staleAtForSpawnedEntries = staleAt
+    // Whether the payload that fulfills the spawned entries is an App Shell
+    // extracted from a larger response. Such a payload is a strict subset of
+    // the response, so it does not represent the entire UI of the target page
+    // — it's partial by construction, regardless of the server's marker. When
+    // the spawned entries are instead filled with the full response, its
+    // partialness comes from the marker (see `isResponsePartial` below).
+    let spawnedEntriesGetExtractedShell = false
     if (cacheData === null) {
       // No shell can be extracted without cache metadata (only present when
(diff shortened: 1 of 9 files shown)


dianatofficial (commented):

lubieowoce:
@dianatofficial your review was not requested. please don't post comments like these out of the blue

PR #97647, opened 2026-08-20
fix: keep returned inline server actions reachable

### LINKED ISSUE #96508: Inline `"use server"` function fails when returned from file-level `"use server"` function

https://github.com/hi-ogawa/reproductions/tree/main/next-rsc-returned-nested-action

### To Reproduce

The reproduction exercises the natural composition of the two `"use server"` forms. An exported function inherits the file-level directive and returns a nested function with its own inline directive and one closure capture.

```ts
'use server'

export async function outer(prefix: string) {
  async function inner(value: string) {
    'use server'
    return `${prefix}:${value}`
  }

  return inner
}
```

The routes differ only in how the Client Component receives `outer`:

1. `/imported` imports `outer` directly into the Client Component.
2. `/prop` imports `outer` in the Server Component and passes it to the Client Component as a prop.

In both routes:

1. Select **Get inner**. The client calls `outer('captured')`, stores its return value in a ref, and displays `stored:function`.
2. Select **Call inner**. The client calls the retained function with `'client'`.

The issue is that `/prop` case succeeds but `/imported` case fails with a following error:

```
 POST /imported 200 in 8ms (next.js: 1.6ms, application-code: 7ms)
  └─ ƒ outer("captured") in 0ms app/imported/actions.ts
Error: Failed to find Server Action "609f86498f34d90b3fc12f77914ba6f5f15698574f". This request might be from an older or newer deployment.
Read more: https://nextjs.org/docs/messages/failed-to-find-server-action
```

Run the automated reproduction with:

```sh
pnpm install
pnpm test
```

On `/imported`, `outer` succeeds and returns a function to the client, but invoking that returned function fails:

```text
UnrecognizedActionError: Server Action "<inner-action-id>" was not found on the server.
```

On `/prop`, the same sequence succeeds and produces:

```text
result:captured:client
```

I expected `/imported` to produce the same result. The returned value is serialized to the client as a Server Action reference, so its generated action ID should remain resolvable when the client invokes it.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
Binaries:
  Node: 24.18.1
  pnpm: 10.34.5
Relevant Packages:
  next: 16.3.0-canary.106
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Server Actions

### Additional context

The split result was verified in:

- Turbopack development.
- Webpack development.
- Turbopack production followed by `next start`.

[PR #62052](https://github.com/vercel/next.js/pull/62052) added support for nested inline actions inside module-level `"use server"` files. Its example includes an exported action returning nested actions, but its coverage is at the SWC transform layer. The transform works in this reproduction; the missing piece is publication of the generated nested export for the direct Client Component import path.

97Fakhreddine:
## What?

Fixes a Server Actions issue where an inline `"use server"` function returned from a file-level `"use server"` action becomes uncallable when the outer action is imported directly into a Client Component.

The outer action succeeds and the nested action is serialized back to the client, but invoking the returned action previously failed with:

```text
Failed to find Server Action "<action-id>"
```

Fixes #96508

## Why?

The server transform already creates and registers the nested inline Server Action.

However, for the direct Client Component import path, the generated nested action reference was not kept reachable through the action metadata used to build the route's Server Action manifest.

As a result, the client received a valid Server Action reference whose ID could not be resolved when invoked.

## How?

Track inline Server Actions created inside public file-level Server Actions during the client transform.

Their generated action IDs are included in the existing Server Action metadata while remaining associated with the public action that makes them reachable.

For Turbopack, the nested references are included only in the metadata for the corresponding public export so existing Server Action tree-shaking behavior is preserved.

A regression test covers the following flow:

1. A Client Component directly imports `outer`.
2. `outer("captured")` returns a nested inline Server Action.
3. The client stores the returned function.
4. The client later calls it with `"client"`.
5. The result is `"captured:client"`.

## Tests

Validated with the locally built SWC native binding.

- `pnpm build`
- Webpack development regression test
- Turbopack development regression test
- Webpack production regression test
- Turbopack production regression test
- Server Action tree-shaking coverage for Webpack and Turbopack
- `git diff --check`

The regression reproduces the missing Server Action failure without the fix and passes with the fix.


dianatofficial (commented):

PR #97648, opened 2026-08-20
Turbopack: Show last modified file when waiting for the filesystem to settle

bgw:
This is a follow-up to https://github.com/vercel/next.js/pull/96116

If we're waiting a long time for your filesystem to settle, we should show you the last modified path, so at least you can have an idea of why this is happening.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/watcher/mod.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/watcher/mod.rs
@@ -667,8 +667,8 @@ impl DiskWatcher {
                             delay = delay.max(config.extended_batch_delay_duration);
                         }
 
-                        if batch.add_event(event) {
-                            schedule.extend(delay);
+                        if let Some(last_modified_path) = batch.add_event(event) {
+                            schedule.extend(delay, Some(last_modified_path.into()));
                         }
                     }
                     // Error raised by notify watcher itself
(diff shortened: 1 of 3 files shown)


turbopack/crates/turbo-tasks-fs/src/watcher/mod.rs
```diff
@@ -843,8 +840,25 @@ impl BatchedInvalidations {
         }
     }
 
-    fn mark(&mut self, path: Box<Path>, flags: InvalidationFlags) {
-        *self.paths.entry(path).or_insert(InvalidationFlags::empty()) |= flags;
+    /// Sets the `flags` for `path`. Returns the index that was modified.
+    fn mark(&mut self, path: Box<Path>, flags: InvalidationFlags) -> usize {
+        match self.paths.entry(path) {
```
lukesandberg, line 845: TODO about the raw table api top avoid boxing the path in the hit case?

lukesandberg (approved):

turbopack/crates/turbo-tasks-fs/src/watcher/mod.rs
bgw: I'm not even really sure this is worth the complexity unless I try it, so I don't really want to put a prescriptive TODO comment in here.

PR #97650, opened 2026-08-20
Add prelude metadata header to cached prerenders

gnoff:
## Summary

- Add `x-next-prelude-metadata: 0` to non-empty cached Cache Components prerenders whose metadata remains unresolved.
- Keep the header absent when metadata is already resolved or the static shell is empty. Empty shells require a blocking full render regardless, so the marker would not improve the later bot-routing decision.
- Behind `NEXT_PRIVATE_HTML_LIMITED_BOTS_AFTER_CACHE_BYPASS=1`, remove the HTML-limited bot matcher from adapter-bound `experimentalBypassFor` rules while retaining the server-action and multipart bypasses.
- Under the same flag, write `deployment_config_v3.json` after the adapter finishes. Its `after_cache_bypass` rule requires both an HTML-limited bot user agent and a cached `x-next-prelude-metadata: 0` response.
- Keep this as a private Build Output API bridge for deployment testing while the first-class Next.js, adapter, and BOA interface is designed.

## Verification

- `pnpm test-start-turbo test/e2e/app-dir/metadata-streaming-cache-components/metadata-streaming-cache-components-default-bots.test.ts`
- `pnpm test-start-webpack test/e2e/app-dir/metadata-streaming-cache-components/metadata-streaming-cache-components-default-bots.test.ts`
- `pnpm test-start-turbo test/production/app-dir/adapter-prerender-metadata/adapter-prerender-metadata.test.ts`
- `pnpm --filter=next build`
- Changed-file Prettier and ESLint checks

### DIFF
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -62,6 +62,8 @@ import { Bundler } from '../../lib/bundler'
 import { resolveCacheHandlerPathToFilesystem } from '../../lib/format-dynamic-import-path'
 import { InvariantError } from '../../shared/lib/invariant-error'
 import type { __ApiPreviewProps } from '../../server/api-utils'
+import { HTML_LIMITED_BOT_UA_RE_STRING } from '../../shared/lib/router/utils/is-bot'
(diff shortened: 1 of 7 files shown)


dianatofficial (commented):

PR #97651, opened 2026-08-20
Improve "Back before hydration" test suite

gaearon:
Test-only changes.

- Drop the Suspense version of the `back-before-hydration` tests. I added it to cover what turned out to be the Activity/Suspense retry bug (#95848), which React has since fixed.
- Wait for the document to finish parsing before changing history in those tests so the tested window is exactly "inline scripts have run, external scripts have not".
- Render `usePathname()`/`useSearchParams()` from the fixture layout and assert the router's URL in every scenario. One of these is wrong so there's a TODO for the future.
- Add a case where "nothing interesting happens" to ensure Back/Forward still works there.


dianatofficial (commented):

PR #97656, opened 2026-08-21
Don't serve a dev route whose file has been deleted

gaearon:
Since the router tables are rebuilt some time after the filesystem changes, a request that arrives right after a page is deleted may still match the route and render the deleted page. Check that the matched route's file exists before ensuring it; if it is gone, the route is not a match and the request should fall through to the 404.

The new test deletes pages and requests them at varying delays after the deletion. Before this change, requests made immediately after a deletion returned 200 on both bundlers. Local run results:

Screenshot 2026-08-21 at 02 19 54


dianatofficial (commented):

PR #97666, opened 2026-08-21
Migrate some async blocks to async closures

mischnic:
This is everything except `turbo-tasks-[backend]`

Using ast-grep replacement:

- search: `|$$$ARGS| async move {$$$BODY}`
- replace: `async |$$$ARGS| {$$$BODY}`

This also made it obvious in some places that we don't actually need async-await,
e.g. `.map(async |v| v.to_resolved.await)`

### DIFF
--- a/turbopack/crates/turbopack-cli/src/build/mod.rs
+++ b/turbopack/crates/turbopack-cli/src/build/mod.rs
@@ -262,27 +262,20 @@ async fn build_internal(
         source_maps_type,
     );
 
-    let entry_requests = (*entry_requests
-        .into_iter()
-        .map(|r| async move {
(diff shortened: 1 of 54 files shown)


turbopack/crates/turbopack-cli/src/build/mod.rs
```diff
@@ -262,27 +262,20 @@ async fn build_internal(
         source_maps_type,
     );
 
-    let entry_requests = (*entry_requests
-        .into_iter()
-        .map(|r| async move {
```
lukesandberg, line 267: 🤦

lukesandberg (approved):

PR #92781, opened 2026-04-14
Turbopack: refactor export analysis

mischnic:
Remove `ModuleReferencesVisitor` and move everything into the `ImportMap` `Analyzer`

PR #89561, opened 2026-02-05
Turbopack: Replace next.js wrapper modules with boundary markers

mmastrac:
#  What
                                                                                                                                                                                                                                                                              
Erase concrete wrapper module types (NextServerComponentModule, NextServerUtilityModule, NextDynamicEntryModule) from all data structures and consumption sites, replacing them with ResolvedVc<Box<dyn Module>>.

# Why 

The wrapper modules were thin shims, but were causing issues with the module renaming PR.

# How

  - Added inner_module: Option<ResolvedVc<Box<dyn Module>>> to BoundaryInfo so downstream code can access the wrapped module without downcasting
  - Replaced all typed ResolvedVc<NextServer*Module> / ResolvedVc<NextDynamicEntryModule> in ClientReference, ClientReferenceGraphResult, ServerEntries, ClientManifestEntryType, DynamicImportEntriesMapType, DynamicImportedChunks, and DynamicImportEntriesWithImporter
  with ResolvedVc<Box<dyn Module>>
  - Removed all validation downcasts (try_downcast_type + .expect()) at detection sites — boundary_info() is now the sole detection mechanism
  - Replaced .await?.module / .server_path() / .source_path() accessors with boundary_info().inner_module / .ident().path() / .source_path

PR #97672, opened 2026-08-21
Turbopack: mangle exported names for smaller bundle sizes

sokra:
### What?

Adds export-name mangling to Turbopack, behind a new experimental option
`experimental.turbopackMangleExportNames` (default `false`). It is independent of minification:
`--no-mangling` is a minifier flag and does not affect it.

When enabled, each ECMAScript module's *used* export names — including `default` and
`__esModule` — are replaced by short keys in the emitted output, both where the module registers
its exports and where every consumer reads them. Modules whose export names could be observed by
user code keep their original names, decided per module.

This is a reland of #89060 (on top of the already-merged #89406), originally written by Matt
Mastracci, who is credited as a co-author on the commit.

Stacked on top: #97676 flips the default to `true` on canary releases, so Next.js's own CI
exercises the feature broadly before it is considered for stable.

### Why?

Bundle size. A module's export keys exist only to link modules together: the producing module
emits `{ someVeryLongExportName: … }` and every consumer reads `ns["someVeryLongExportName"]`.
Both sides are generated by us, so as long as producer and consumer agree — and the name isn't
observable from user code — the key can be a single character. Long export names are extremely
common in real dependency graphs (icon sets, utility packages, barrel files), and each one is paid
for once in the module that defines it and once per importing module.

### How?

**Ported, not rebased.** `canary` is ~2500 commits past the original stack's base, and the files it
touched were independently rewritten in the meantime (export-analysis refactor #92781, CJS analysis
for scope hoisting #95826, the `module_fragments` subsystem #95978). A probe rebase produced 16
conflicting files on the first commit alone, so the original branches were used as a reference
implementation — for intent, the identifier alphabet, and test coverage — and the feature was
rebuilt on today's infrastructure. #89561 from the original stack (erasing the Next.js wrapper
module types) is deliberately **not** part of this change; it turned out to be unnecessary, because
those modules already declare whole-module export usage and therefore back off on their own.

**The name table** (`references/esm/mangle/table.rs`) hashes each name into a table of all valid JS
identifiers of the smallest length that fits the name set — 15 exports get single-character keys —
and resolves collisions by open addressing. Hashing rather than assigning `a`, `b`, `c`, … is what
keeps names stable: an unrelated edit elsewhere in the module doesn't renumber every other export,
and a collision only perturbs its own cluster. Assignment happens in two passes: every name that is
*already* a valid identifier at the chosen length keeps itself and reserves its bucket first, and
only then is anything hashed — so an export called `a` keeps `a`, and nothing else can be assigned
it. Both passes iterate in sorted order, so the mapping depends only on the set of names.

A module with exactly **one** mangleable export is special-cased to a fixed key, `f`, rather than a
hashed one. `f` is the most common character in JS keywords (`if`, `for`, `function`), and every
single-export module in the graph then emits the same `.f` / `.f()` byte sequences, which gzip's
back-references pick up across the whole bundle — a bigger win than hashing, at the cost of that one
key changing when a second export is added.

A fixed list, `RESERVED_KEYS`, is withheld from every table for two different reasons: JS reserved
words (`if`, `in`, `do`, `for`, `let`, `new`, `try`, `var`) are legal as quoted property keys but a
minifier will not fold `ns["if"]` into the shorter `ns.if`, so handing one out costs bytes instead of
saving them; and `__esModule` is withheld because the runtime's `esm()` helper defines that property
on every module's exports object regardless of what the module itself exports, so an assigned key
landing on it would collide. (`default` needs no such protection — once it is mangled like any other
export, nothing else emits a property under that literal name.)

**One source of truth for the mapping.** `mangled_export_names(module, chunking_context)` is a
turbo-task that both the producing side (`EsmExports::code_generation`) and the consuming side
(`ReferencedAssetIdent::Module`, the single place a cross-module export access is materialized) ask
for the *target* module's map. Neither side computes a table of its own, so they cannot disagree,
and the task derives export usage from the chunking context itself rather than accepting it as an
argument, so a caller can't supply usage from the wrong graph. Re-export chains need no special
handling, because the consumer side already resolves through re-exports to the module that produces
the binding.

The mangling decision itself lives on `EsmExports` as a `mangle_export_names: bool` field, rather
than a separate trait method every module type has to override. A module that derives its exports
from another one (a facade, a locals module, a part, a rename) inherits the flag with the data,
which removed seven hand-written delegations and the possibility of a new wrapper type forgetting
one.

**A mangling decision must not cross module identities.** A few module types hand out *another*
module's exports value as their own (the WASM loader module, the module-fragments side-effects
wrapper, the client-reference proxy). If that borrowed value carried a real mangling decision, the
producing and consuming sides would key their lookups on two different modules and could compute two
different keys for the same export — this actually broke every WASM- and `@vercel/og`-based test
once the default-on layer exercised it in CI. `EcmascriptExports::borrowed()` is the one place this
is handled: it always returns an unmangled view, and every such pass-through site uses it.

**Back-off is per module**, built on the export-usage information that landed after the original PR
(`BindingUsageInfo` / `ModuleExportUsageInfo`) rather than the original's locals/facade-split
heuristic. A module keeps its names when its usage is `All` (a namespace import that couldn't be
lowered, a computed property access, an unresolvable `export *`, or a chunk-group entry — which
covers client references and the Next.js wrapper modules), when it is read through a namespace value
at all, when its exports are dynamic or not statically known ESM, or when names aren't being mangled
in this build.

`__webpack_exports_info__` gains `canMangle` and `mangledName` per export, which is how a running
test can observe the mapping; with the option off it emits exactly what it emitted before.

### Testing

- `turbo-tasks-hash`/table unit tests: encode/decode round-trip, degenerate-name rejection, table
  sizing, the single-export fixed key (including its own reservation), reserved-word withholding
  (including a reserved bucket-count test that stays in sync with the reserved list), the
  preserved-name pass running before any hashing, uniqueness under heavy collision, order
  independence, wrap-around probing, and same-tier stability.
- 13 `turbopack-tests` execution fixtures under `tests/execution/turbopack/exports/mangle-*`,
  several ported from the original PR and from webpack's `test/configCases/mangle`: named imports,
  re-export chains and default exports (including one literally named `__esModule`), escaping
  namespaces (`Object.keys`, `delete ns.missing`, `export * as`, CJS interop), destructuring,
  prototype-shadowing names (`toString`, `$1`, `__1`), a 60-export two-character table, dynamic
  `import()` with `webpackExports` / `turbopackExports`, a CommonJS consumer of an ESM module,
  dynamic re-exports, scope hoisting on and off, and a control with the option off.
- 3 committed snapshot fixtures under `tests/snapshot/mangle-exports`, so the emitted keys, the
  back-off, and the fixed single-export key are visible in review.
- One fixture under `__skipped__`, which the harness asserts *fails*, recording the
  namespace-materialization gap below.
- Full suite: 546 unit + 272 execution + 125 snapshot tests pass, with no snapshot churn across the
  several refactors this PR went through in review.
- Verified against real builds: targeted app-dir, worker, WASM, and `@vercel/og`-based e2e suites
  pass with the option forced on (the failures that remain are external-network tests that fail
  identically with it off), and a small two-page app shrank by 0.76% of total emitted JS / 0.53%
  gzipped.

### Known limitations, each intentional

- **A module read through `import * as ns` is never mangled**, even when every read is statically
  tracked, because the analysis doesn't yet distinguish a lowered named read from a materialized
  namespace object. Namespace imports are common, so this leaves real wins on the table; unlocking
  it is the highest-value follow-up.
- **Escaping namespaces back off entirely.** Webpack instead keeps mangling and materializes a
  namespace object keyed by the original names. That is the `__skipped__` fixture: implementing it
  turns the suite red until the fixture is moved out.
- **CommonJS export mangling is out of scope.** The producing side is reachable, but the consuming
  accesses live in user source and nothing rewrites them today; it needs its own design pass.

Closes PACK-435


turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -37,6 +37,17 @@ pub trait EcmascriptChunkPlaceable: ChunkableModule + Module {
...
     }
 
+    /// Whether the keys this module exports its bindings under may be shortened. See
+    /// [`crate::references::esm::mangle::mangled_export_names`], which is what decides whether
+    /// they actually are.
+    ///
+    /// Defaults to `false`: a module type that doesn't opt in keeps its original export names,
+    /// which is always correct.
+    #[turbo_tasks::function]
+    fn mangle_export_names(self: Vc<Self>) -> Vc<bool> {
```
lukesandberg: does this need to be a new turbo task function? could it be part of EcmascriptExports?

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+use turbopack_core::{
+    chunk::{ChunkingContext, MinifyType},
+    module_graph::binding_usage_info::ModuleExportUsageInfo,
+};
+
+use crate::chunk::{EcmascriptChunkPlaceable, EcmascriptExports};
+
+/// The export name of the default export. Not mangled: the CommonJS interop layer and the
+/// `__esModule` handling in the runtime look it up by name, an embedder's runtime may read
+/// `.default` off a loaded module by string, and it is only seven characters to begin with.
```
lukesandberg: do we think this is addressible if commonjs mangling happens?

turbopack/crates/turbo-tasks-hash/src/basejs.rs
```diff
@@ -0,0 +1,406 @@
...
+///
+/// 1. Every name that is *already* a valid identifier of at most that length keeps itself and
+///    reserves its bucket. This has to happen for **all** names before anything is hashed, or a
+///    hashed name could take a bucket that a later preserved name needs.
+/// 2. The remaining names are hashed into the table, resolving collisions by open addressing.
+///
+/// Both passes iterate in sorted order, so the result depends only on the *set* of names, never on
+/// the order they arrive in.
+pub fn shorten_to_unique_names<'a>(
+    names: impl IntoIterator<Item = &'a str>,
```
lukesandberg: any reason to not just make this use RcStr?

turbopack/crates/turbo-tasks-hash/src/basejs.rs
lukesandberg, line 1: naively i would expect this to be in turbopack-ecmascript

turbopack/crates/turbo-tasks-hash/src/basejs.rs
```diff
@@ -0,0 +1,406 @@
...
+const FIRST_CHARS: &[u8; 54] = b"_$ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+/// Characters that may appear in an identifier after the first character.
+const REST_CHARS: &[u8; 64] = b"_$0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+/// The character that encodes the value zero. Because it is a legal *leading* character, a value
+/// whose encoding would end in it is degenerate (`a` and `a_` would both decode to the same
+/// value), so such encodings are never produced and never accepted.
+const ZERO_CHAR: char = '_';
+/// The name used when only a single name has to be assigned. Any single character would do; `f`
+/// is what the original implementation used, and it reads as "function" in the common case.
+const SINGLE_ITEM_IDENTIFIER: &str = "f";
```
lukesandberg: we shouldn't have a special case for a single name.  it doesn't optimize much and it makes 1-2 export transitions have a larger blast radius

turbopack/crates/turbo-tasks-hash/src/basejs.rs
```diff
@@ -0,0 +1,406 @@
...
+                name.to_string()
+            } else {
+                SINGLE_ITEM_IDENTIFIER.to_string()
+            };
+            return FxHashMap::from_iter([(name, mangled)]);
+        }
+        _ => {}
+    }
+
+    let len = len_for_count(names.len() as u64);
```
lukesandberg: is this accurate, do we need to claim js keywords?

so there are 54 one char idents
but 54 + 64 - 3 one or two char idents  etc

`if` `do` and `in` are probably the most relevant but im guessing sometimes `try` and `new` could be a problem

turbopack/crates/turbo-tasks-hash/src/basejs.rs
```diff
@@ -0,0 +1,406 @@
...
+    unsafe { String::from_utf8_unchecked(result) }
+}
+
+/// Decodes an identifier back to the value [`encode_js_identifier`] would encode.
+///
+/// Returns `None` when the string is not in the image of [`encode_js_identifier`] — it contains a
+/// character outside the alphabet, or it is a degenerate encoding with trailing zeros. This is
+/// what makes the encoding a bijection, which in turn is what lets an existing short name keep
+/// its own name without ever colliding with an assigned one.
+pub fn decode_js_identifier(s: &str) -> Option<u64> {
```
lukesandberg: do all these need to be pub

turbopack/crates/turbo-tasks-hash/src/basejs.rs
```diff
@@ -0,0 +1,406 @@
...
+    while capacity_for_len(len) < count {
+        len += 1;
+    }
+    len
+}
+
+/// Encodes a value to a valid JS identifier. Always returns at least one character, and never
+/// returns an encoding with a trailing [`ZERO_CHAR`] (those values are the degenerate ones, see
+/// [`ZERO_CHAR`]).
+pub fn encode_js_identifier(mut value: u64) -> String {
```
lukesandberg: not pub?

turbopack/crates/turbo-tasks-hash/src/basejs.rs
```diff
@@ -0,0 +1,406 @@
...
+    if value < FIRST_CHARS.len() as u64 {
```
lukesandberg: we really wouldn't want this algorithm to select `default` but it could.... 

(obviously only with a few thousand exports)

so we would need to treat terms like that act like they are already reserved

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    let mangled = shorten_to_unique_names(names.iter().map(|name| name.as_str()));
+
+    let map = names
+        .iter()
+        .map(|name| {
+            let mangled = mangled
+                .get(name.as_str())
+                .expect("every name passed in has a mangled name");
+            ((*name).clone(), RcStr::from(mangled.as_str()))
```
lukesandberg: should call `take` above instead of `get` so we can zero copy pass the String into RcStr

or just have the shorten routine return RcStr which should also be fine

dianatofficial (commented):

turbopack/crates/turbo-tasks-hash/src/basejs.rs
sokra: `f` is the most common char in keywords in Javascript like `if` `for` `function`, so it would gzip very well.
Also choosing always the same char has additional benefits instead of choosing one by hash (common `.f` and `.f()`).

Yes doing that has some drawback for 1 vs more exports changes, but I think the benefit of better compression is bigger here. So add it again please.

turbopack/crates/turbo-tasks-hash/src/basejs.rs
sokra: While they are not problematic for correctness (`ns["if"]` is ok), they are less efficient `ns.it` vs `ns["if"]`. Can we do something to avoid selecting these 2 or 3 char keywords in the hashing? Maybe by threading these values as conflicts and choosing a new value.

sokra: Maybe `RESERVED_KEYS` works here?

turbopack/crates/turbo-tasks-hash/src/basejs.rs
sokra: The PR on top (which allows to mangle `default`), could remove `default` from the `RESERVED_KEYS` list again.

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,565 @@
...
+/// stays as it is, where a non-reserved `ns["ab"]` becomes `ns.ab`.
+///
+/// Only two- and three-character words are listed. No single character is a reserved word, and by
+/// four characters the table holds 200k+ buckets, so losing a name there is irrelevant.
+const UNQUOTABLE_KEYS: &[&str] = &[
+    // 2 characters
+    "do", "if", "in",
+    // 3 characters. `let` counts: the generated output is a module, and modules are always strict.
+    "for", "let", "new", "try", "var",
+];
```
sokra: Merge RESERVED_KEYS into this list and rename UNQUOTABLE_KEYS -> RESERVED_KEYS.

This avoid the `reversed` argument for `shorten_to_unique_names`

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,565 @@
...
+        let capacity = capacity_for_len(len);
+        let withheld_in_table = withheld
+            .clone()
+            .filter(|name| bucket_of(name, len).is_some())
+            .count() as u64;
+        if capacity >= names.len() as u64 + withheld_in_table {
+            break;
+        }
+        len += 1;
+    }
```
sokra, line 239: Instead of computing this, use a fixed number per len since there is no dynamic `reserved` list anymore

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,571 @@
...
+/// length: nothing at one character, the three two-letter words at two, all eight reserved words
+/// from three, and `__esModule` from ten.
+const fn reserved_in_table(len: u32) -> u64 {
+    match len {
+        0 | 1 => 0,
+        2 => 3,
+        3..=9 => 8,
+        _ => 9,
+    }
+}
```
sokra, line 83: There should be a unit test to verify that matches RESERVED_KEYS

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,571 @@
...
+/// table for that length. Saturates instead of overflowing for absurd lengths.
+///
+/// This counts *values*, not strings: [`encode_js_identifier`] never emits a trailing
+/// [`ZERO_CHAR`], so a value whose last digit would be zero encodes to a shorter string instead,
+/// and every value below the returned bound really does fit in `len` characters.
+fn capacity_for_len(len: u32) -> u64 {
+    if len == 0 {
+        return 0;
+    }
+    (FIRST_CHARS.len() as u64).saturating_mul((REST_CHARS.len() as u64).saturating_pow(len - 1))
```
sokra: Precompute that for len <= 10 and store it a const values similar to `reserved_in_table`. Verify correctness with a unit test against the function

docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -422,26 +422,27 @@ Turbopack can be configured via `next.config.js` (or `next.config.ts`) under the
...
+| `turbopackRemoveUnusedExports`                                                                               | Enable removing unused exports.                                                                                                                                                                                                                                       | `false`       | `true`                        |
+| `turbopackMangleExportNames`                                                                                 | Shorten the export names modules expose to each other, to reduce bundle size. Requires minification with mangling; a module whose export names can be observed (an escaping namespace object, a dynamic `import()`, a CommonJS `require()`) keeps its original names. | `false`       | `false`                       |
```
sokra: > Requires minification with mangling

remove this part. Mangling is not required for that.

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+/// Both the producing and the consuming side call this for the same `module`, which is what
+/// guarantees they agree. The mapping covers the module's *used* exports: unused exports are not
+/// emitted at all, so giving them names would only make the remaining names longer.
+///
+/// A module keeps its original names when any of these hold:
+///
+/// 1. mangling is disabled for the module that owns these exports, or names are not being mangled
+///    at all — either we're not minifying, or minification was asked to keep names
+///    (`MinifyType::Minify` with no `MangleType`), in which case mangled export keys would only
+///    make the output harder to read,
```
sokra: Mangling should not depend on minification. It should be possible to enable/disable it independent of it.
There is no requirement on mangling in minification.

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        || !matches!(
+            *chunking_context.minify_type().await?,
+            MinifyType::Minify {
+                mangle: Some(_),
+                ..
+            }
+        )
+    {
+        return Ok(Vc::cell(None));
+    }
```
sokra: remove that check

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+};
+
+use self::table::shorten_to_unique_names;
+use crate::chunk::{EcmascriptChunkPlaceable, EcmascriptExports};
+
+/// A module's export name mapping: original export name -> the key actually used in the output.
+///
+/// It covers every export the module emits, including the ones whose key is unchanged because the
+/// original name was already short enough to keep — which makes the map a complete description of
+/// the module's emitted keys. A name absent from the map keeps its original name.
```
sokra: It would be more memory efficient if we only include names that were changed in the map.

sokra: It will fallback to the original name anyway on missing entries

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,613 @@
...
+
+    for name in RESERVED_KEYS {
+        if let Some(bucket) = bucket_of(name, len) {
+            used.insert(bucket);
+        }
+    }
+
+    // Pass 1: names that are already valid short identifiers keep themselves and claim their
+    // bucket. This must complete before any hashing happens.
+    let mut to_mangle = Vec::new();
```
sokra: Give to_mangle initial capacity of `names` since we expect most names to be mangled

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,613 @@
...
+    for name in names {
+        match bucket_of(name, len) {
+            // A name that is also reserved can't keep itself; it gets a fresh bucket below.
+            Some(bucket) if used.insert(bucket) => {
+                result.insert(name, name.clone());
```
sokra: This is no longer needed when if omit unchanged names from the map

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    let map = names
+        .iter()
+        .map(|name| {
+            let mangled = mangled
+                .get(name)
+                .expect("every name passed in has a mangled name");
+            ((*name).clone(), mangled.clone())
+        })
+        .collect::<Vec<(RcStr, RcStr)>>();
```
sokra: This is wasteful. Make `shorten_to_unique_names` return a FrozenMap

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+/// A module's export name mapping: original export name -> the key actually used in the output.
+///
+/// It covers every export the module emits, including the ones whose key is unchanged because the
+/// original name was already short enough to keep — which makes the map a complete description of
+/// the module's emitted keys. A name absent from the map keeps its original name.
+#[turbo_tasks::value(transparent)]
+pub struct MangledExportNames(pub FrozenMap<RcStr, RcStr>);
+
+#[turbo_tasks::value(transparent)]
+pub struct OptionMangledExportNames(pub Option<ResolvedVc<MangledExportNames>>);
```
sokra: Since an empty MangledExportNames map has the same effect as a None (OptionMangledExportNames), we can optimize that.

Remove OptionMangledExportNames. Instead always return MangledExportNames, but return an empty one for the cases we returned None before

turbopack/crates/turbopack-ecmascript/src/references/exports_info.rs
```diff
@@ -44,18 +44,54 @@ impl ExportsInfoBinding {
...
                 .map(|e| {
                     let used: Expr = export_usage_info.is_export_used(e).into();
+                    if !mangling_enabled {
+                        return PropOrSpread::Prop(Box::new(swc_core::ecma::ast::Prop::KeyValue(
+                            KeyValueProp {
+                                key: PropName::Str(e.as_str().into()),
+                                value: quote!("{ used: $v }" as Box<Expr>, v: Expr = used),
+                            },
+                        )));
+                    }
```
sokra: Remove that !manglined_enabled check. We always want to emit all keys.

turbopack/crates/turbopack-ecmascript/src/references/exports_info.rs
```diff
@@ -44,18 +44,54 @@ impl ExportsInfoBinding {
...
+                    let mangled = mangled_names.as_ref().and_then(|names| names.get(e));
+                    let can_mangle: Expr = mangled.is_some().into();
+                    let mangled_name: Expr =
+                        match mangled {
+                            Some(mangled) => Expr::Lit(mangled.as_str().into()),
+                            None => Expr::Lit(swc_core::ecma::ast::Lit::Null(
+                                swc_core::ecma::ast::Null { span: DUMMY_SP },
+                            )),
```
sokra, line 78: mangledName should be the original name when its not mangled

sokra: Actually I take that back. Keep `null` when not mangled

sokra: Handle the case when it is mangled, but mangled to the same name. In this case it should be a string

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -95,26 +89,54 @@ pub async fn mangled_export_names(
...
+    if !*mangling_eligible_for_module(*module, chunking_context).await? {
+        return Ok(Vc::cell(FrozenMap::default()));
     }
 
+    // `mangling_eligible_for_module` already established that `module`'s exports are
+    // `EcmascriptExports::EsmExports` with `mangle_export_names` set and a known, non-dynamic,
+    // non-empty used-export set, so all of this is safe to recompute (and cheap: every one of
+    // these is itself a memoized turbo-tasks call).
+    let EcmascriptExports::EsmExports(exports) = *module.get_exports().await? else {
+        unreachable!("mangling_eligible_for_module already checked this");
```
sokra: Make this a `bail!` instead as it could happen due to eventual consistency

docs/01-app/03-api-reference/08-turbopack.mdx
sokra, line 1: CI lint complains about this file

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+    // `mangling_eligible_for_module` established that `module`'s exports are
+    // `EcmascriptExports::EsmExports` with `mangle_export_names` set and a known, non-dynamic,
+    // non-empty used-export set, so recomputing all of this is cheap — every one of these is
+    // itself a memoized turbo-tasks call.
+    //
+    // This is a `bail!` rather than an `unreachable!` because the two calls are separate tasks:
+    // under eventual consistency they can observe different revisions of the module's exports, so
+    // the shape can legitimately have changed in between. That is a transient inconsistency, not a
+    // broken invariant, so it must be an error rather than a panic.
+    let EcmascriptExports::EsmExports(exports) = *module.get_exports().await? else {
```
lukesandberg: this is accessing a lot of fields that mangling_eligible_for_module also reads

perhaps they should be the same function, and it returns a Option<MangledExportNames>

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,634 @@
...
+    54,
+    3_456,
+    221_184,
+    14_155_776,
+    905_969_664,
+    57_982_058_496,
+    3_710_851_743_744,
+    237_494_511_599_616,
+    15_199_648_742_375_424,
+    972_777_519_512_027_136,
```
lukesandberg: can we write out the math that produces this?

or really should we bother precomputing this?  the saturating multiplications are not going to be very expensive

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,634 @@
...
+/// Decodes an identifier back to the value [`encode_js_identifier`] would encode.
+///
+/// Returns `None` when the string is not in the image of [`encode_js_identifier`] — it contains a
+/// character outside the alphabet, or it is a degenerate encoding with trailing zeros. This is what
+/// makes the encoding a bijection, which in turn is what lets an existing short name keep its own
+/// name without ever colliding with an assigned one.
+fn decode_js_identifier(s: &str) -> Option<u64> {
+    if s.is_empty() {
+        return None;
+    }
```
lukesandberg: identifiers cannot be empty.... panic?

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,634 @@
...
+
+/// Characters that may start an identifier. Digits are excluded, so this is one character shorter
+/// than [`REST_CHARS`].
+const FIRST_CHARS: &[u8; 54] = b"_$ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+/// Characters that may appear in an identifier after the first character.
+const REST_CHARS: &[u8; 64] = b"_$0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+/// The character that encodes the value zero. Because it is a legal *leading* character, a value
+/// whose encoding would end in it is degenerate (`a` and `a_` would both decode to the same value),
+/// so such encodings are never produced and never accepted.
+const ZERO_CHAR: char = '_';
```
lukesandberg: isn't this down to the encoding scheme?

instead of multiplying the next character by the base, we multiplied the accumulated value by the base, then trailing zeros would be accounted for

turbopack/crates/turbopack-ecmascript/src/references/esm/export.rs
```diff
@@ -533,6 +533,14 @@ pub struct EsmExports {
     pub exports: FrozenMap<RcStr, EsmExport>,
     /// Unexpanded `export * from ...` statements (expanded in `expand_star_exports`)
     pub star_exports: Vec<ResolvedVc<Box<dyn ModuleReference>>>,
+    /// Whether the keys these exports are emitted under may be shortened, i.e. whether the module
+    /// they belong to has export mangling enabled. Carried with the exports so that a module which
+    /// derives its exports from another one (a facade, a locals module, a part, a rename) inherits
+    /// it automatically. See `references::esm::mangle::mangled_export_names`, which is
+    /// what decides whether the keys actually are shortened.
+    ///
+    /// `false` for exports built for a wrapper whose export names have to stay as written.
```
lukesandberg: shorten comment

turbopack/crates/turbopack-ecmascript/src/collect_module.rs
```diff
@@ -170,6 +170,9 @@ impl EcmascriptChunkPlaceable for EcmascriptCollectModuleWithChunkGroup {
                 )]
                 .into(),
                 star_exports: vec![],
+                // Internal bookkeeping module for turbopack's own module-collection tooling, read
+                // by name ("getList") elsewhere in that same tooling — never mangle it.
+                mangle_export_names: false,
```
lukesandberg: this seems like a TODO.... we should be able to consistently mangle both sides

turbopack/crates/turbopack-ecmascript/src/collect_module.rs
lukesandberg: nit: this should be `rcstr!("getList")` not `into`

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
sokra: Its multiply and pow, so I would better save on doing that on the fly. It's check by a test case against the computation. But we can add the formular as comment too

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/mod.rs
sokra: We could go with:

```
#[turbo_tasks::value(transparent)]
pub struct MangledExportNames(pub Option<FrozenMap<RcStr, RcStr>>);
```

lukesandberg (approved):

turbopack/crates/turbopack-ecmascript/src/references/esm/export.rs
```diff
@@ -697,6 +704,16 @@ impl EsmExports {
...
+        // The keys this module's exports are emitted under. Consumers resolve the same map for this
+        // module (see `ReferencedAsset::get_ident_inner`), so both sides always agree.
+        let mangled_names = mangled_export_names(*module, chunking_context).await?;
+        let export_key = |exported: &RcStr| -> RcStr {
+            mangled_names
+                .as_ref()
+                .and_then(|names| names.get(exported))
+                .cloned()
+                .unwrap_or_else(|| exported.clone())
+        };
```
mischnic: A lot of cloning just so that we do `.as_str()` on the return value again.

turbopack/crates/turbopack-ecmascript/src/references/esm/mangle/table.rs
```diff
@@ -0,0 +1,656 @@
...
+//! downstream minifier that folds `ns["name"]` into the shorter `ns.name` typically only does so
+//! for a non-reserved identifier, so a keyword-shaped key never gets that benefit.
+
+use rustc_hash::FxHashSet;
+use turbo_frozenmap::FrozenMap;
+use turbo_rcstr::{RcStr, rcstr};
+use turbo_tasks_hash::hash_xxh3_hash64;
+
+/// Characters that may start an identifier. Digits are excluded, so this is one character shorter
+/// than [`REST_CHARS`].
```
mischnic:
```suggestion
/// Characters that may start an identifier (i.e. no digits).
```
It's 10 characters shorter?

mischnic (approved):

PR #30758, opened 2021-11-01
Don't require source files to be writeable in dev mode

evan-bradley:
Currently, hot reloading is broken when running dev mode in systems
like Bazel, where the source files are marked as read-only. The goal of
the check is merely to see if a file exists, so checking for read
permissions should be sufficient.

Let me know if any additional information is needed. I didn't see any
good spots to add tests, but I'm willing to add them with some
guidance.

## Bug

- [ ] Related issues linked using `fixes #number`
- [ ] Integration tests added
- [ ] Errors have helpful link attached, see `contributing.md`

PR #97673, opened 2026-08-21
test(dev): add dedicated compiler polling coverage

sokra:
### What?

Adds a dedicated development integration test for compiler file watching in polling mode under both webpack and Turbopack.

It also removes obsolete `CHOKIDAR_USEPOLLING` / `CHOKIDAR_INTERVAL` variables from the unrelated `read-only-source-hmr` suite, which remains focused on read-only source semantics and normal watcher behavior.

Stack: **5/5** · Previous: #98737

### Why?

Next.js did not have an e2e/development test dedicated to `watchOptions.pollIntervalMs`. Existing Rust coverage checks watcher modes, while the filesystem-watcher fuzzer is manual rather than part of the normal development-test matrix.

The test also regresses notify-rs/notify#981. Older PollWatcher versions stored ordinary-file mtimes at whole-second precision, so a second edit in the same second compared equal and was permanently dropped. The preceding stack layer upgrades to `notify 9.0.0-rc.5`, which preserves nanosecond precision for regular files without content hashing.

### How?

The dedicated pages-directory fixture configures only `watchOptions.pollIntervalMs`:

- webpack maps it to compiler `watchOptions.poll`;
- Turbopack passes it to notify and selects PollWatcher.

It intentionally does not set `WATCHPACK_POLLING`, create/delete routes, or test the dev-server route watcher, keeping the test focused on compiler-level polling.

The test writes two versions of an already-loaded page with controlled mtimes at +100 ms and +200 ms inside the same future whole second. It waits for the first HMR update before making the second write, requiring two distinct watcher events rather than allowing changes to coalesce.

A negative control with the older npm-provided native binary observes the first update but permanently misses the second. A fresh local native build from the full stack passes both updates. The dedicated test and `read-only-source-hmr` pass under both Turbopack and webpack; Prettier and ESLint also pass.


dianatofficial (commented):

bgw (changes requested):
We should have some test coverage of watch polling if we don't already, but this test should not do it.

This dates back to https://github.com/vercel/next.js/pull/30758/changes

Our CI infra is very different than it used to be, and no other e2e test requires polling to solve watcher reliability issues.

test/development/read-only-source-hmr/test/index.test.ts
```diff
@@ -14,10 +19,13 @@ describe('Read-only source HMR', () => {
     skipStart: true,
     env: {
       __NEXT_TEST_WITH_DEVTOOL: '1',
-      // Events can be finicky in CI. This switches to a more reliable
-      // polling method.
-      CHOKIDAR_USEPOLLING: 'true',
```
sokra, line 19: Remove this broken env vars from the test case and instead add a new test case that tests polling mode specifically

sokra: Also check if there are already tests that test polling mode

bgw (approved):

PR #97675, opened 2026-08-21
Turbopack: mangle the `default` export key too

sokra:
Stacked on #97672 — mangle the `default` export key too.

### What?

Removes `default` from the set of names excluded from mangling. It now gets a short key exactly
like any other export, subject to the same module-level back-off as everything else.

### Why?

The exclusion in #97672 was conservative rather than necessary: the whole turbopack test suite
passes with `default` included, and I couldn't find a case where it mattered — but that hadn't
been run through e2e yet, so it was left out and filed as a follow-up instead of asserted safe.

### How?

The only runtime code that looks `'default'` up by name is CommonJS interop: `esmNamespaceObject`
synthesizing a `default` binding for a raw CommonJS value, and the proxy traps behind
`ensureDynamicExports` (`export *` never re-exports `default`). Neither path ever sees a mangled
module — a CommonJS target, a module with dynamic re-exports, and a module reached through
`require()` all already return `None` from `mangled_export_names` for other reasons. For a module
that reaches the table, every `default` access is generated by us and resolves through the exact
same mapping as any other export.

`__esModule` stays excluded — unlike `default`, it *is* the interop marker and is read by name off
`module.exports`.

Added a dedicated fixture asserting `default` gets a mangled key, that the key doesn't collide
with the module's other exports, and that the value still flows through both `export { default }`
and `export { default as X }`. Reran the full turbopack suite (271 execution + 124 snapshot, no
churn) and the app-dir e2e suites with mangling forced on: 119 passed / 11 failed, the same
external-network failures as the base feature's flag-off control — no new failure from mangling
`default`.


turbopack/crates/turbopack-ecmascript/src/references/esm/mangle.rs
```diff
@@ -23,19 +28,16 @@ use turbopack_core::{
...
 use crate::chunk::{EcmascriptChunkPlaceable, EcmascriptExports};
 
-/// The export name of the default export. Not mangled: the CommonJS interop layer and the
-/// `__esModule` handling in the runtime look it up by name, an embedder's runtime may read
-/// `.default` off a loaded module by string, and it is only seven characters to begin with.
-const DEFAULT_EXPORT: &str = "default";
-/// Never mangled — it is part of the interop contract with CommonJS consumers.
+/// Never mangled — it is part of the interop contract with CommonJS consumers, and the runtime
+/// reads it by name off a module's exports.
 const ES_MODULE_FLAG: &str = "__esModule";
```
lukesandberg: isn't this off base?

this is meaningly in an ESM context, it is just how a commonjs module signals interop with esm, not the other way around

so i think we can mangle it also

dianatofficial (commented):

sokra:
This PR should be merged into #97672

PR #97676, opened 2026-08-21
Turbopack: enable export mangling by default in production builds

sokra:
Stacked on #97672 — enable export mangling by default on canary releases.

### What?

Two defaults change:

- **Next.js builds.** `experimental.turbopackMangleExportNames` is pinned to `false` for stable
  releases and left *unset* on canary, where Turbopack then defaults it from the build mode: on for
  production builds, off in development. Explicitly setting the option wins in either direction, so
  setting it to `true` in development is honoured.
- **Turbopack's own execution test suite.** `TestOptions::mangle_export_names` defaults to `true`,
  so all 279 fixtures exercise mangling instead of only the handful that opt in.

### Why?

The feature is verified by turbopack's own fixtures, targeted e2e suites, and a couple of
bundle-size measurements — a narrow slice of what Next.js's test suites actually cover. Defaulting
it on for canary puts every production-mode e2e and integration test through the mangled code path
for real users of the canary channel, without committing stable users to it yet. Turning it on for
stable is a separate, later decision that can stack on top of this once canary has soaked it.

Broad exposure has already earned its keep several times over. Turning mangling on in CI and in the
fixture suite surfaced bugs that no hand-picked suite had found, including a module-fragments helper
handing out another module's exports value (#97672's `EcmascriptExports::borrowed()`), a
client-reference proxy memoized per wrapper instance rather than per content, and a
code-elimination bug for export-less modules (fixed in the follow-up PR).

### How?

`defaultConfig` cannot see the build mode, so it only expresses the stable/canary split
(`isStableBuild() ? false : undefined`) and `NextConfig::turbopack_mangle_export_names(mode)`
supplies the mode-dependent default on the Turbopack side. This is deliberately *not* the shape of
the neighbouring `turbopackSharedRuntime: !isStableBuild()`: mangling should not apply in
development, and hard-forcing `false` there in Rust — as an earlier revision of this PR did — would
silently ignore a user who asked for it explicitly.

Mangling does **not** depend on minification. An earlier version of this PR gated it on
`minify(mode)`, which was wrong and has been removed; `--no-mangling` is a minifier flag and does
not affect export mangling.

Verified with a real `next build` A/B on a small app: the option left unset mangles (959,174 B total
emitted JS, 293,407 B gzipped) while explicitly setting it to `false` does not (966,519 B / 294,963
B gzipped), confirming both the default and the override. All 279 execution fixtures pass with the
suite default flipped, with no fixture needing an opt-out.

Note this layer does not change the `turbopack-emit-collect` snapshots on its own: making a module
without re-exports split additionally requires the split trigger added in the follow-up PR, so those
snapshots move there.

This is the layer to revert if canary turns up problems specific to running with the option on by
default; #97672 remains useful as an opt-in feature either way.

### DIFF
--- a/crates/next-core/src/next_config.rs
+++ b/crates/next-core/src/next_config.rs
@@ -2570,13 +2570,20 @@ impl NextConfig {
         )
     }
 
+    /// Whether Turbopack should shorten ("mangle") the export names modules expose to each other.
+    ///
+    /// An explicit value always wins, in either direction — setting this to `true` in development
+    /// is honoured. `mode` only supplies the default when the option is unset: on in production
+    /// builds, off in development, where the extra module splitting costs rebuild time and the
+    /// short names make debugging harder for no benefit.
     #[turbo_tasks::function]
-    pub fn turbopack_mangle_export_names(&self) -> Vc<bool> {
-        Vc::cell(
-            self.experimental
-                .turbopack_mangle_export_names
-                .unwrap_or(false),
-        )
+    pub async fn turbopack_mangle_export_names(&self, mode: Vc<NextMode>) -> Result<Vc<bool>> {
+        Ok(Vc::cell(
+            match self.experimental.turbopack_mangle_export_names {
+                Some(explicit) => explicit,
+                None => !mode.await?.is_development(),
+            },
+        ))
     }
 
     #[turbo_tasks::function]
--- a/packages/next/src/server/config-shared.ts
+++ b/packages/next/src/server/config-shared.ts
@@ -2364,6 +2364,10 @@ export const defaultConfig = Object.freeze({
     turbopackInferModuleSideEffects: true,
     turbopackPluginRuntimeStrategy: 'childProcesses',
     turbopackSharedRuntime: !isStableBuild(),
+    // Pinned off for stable releases. Left unset on canary so the Turbopack side picks the
+    // default from the build mode (on for production builds, off in development) — see
+    // `NextConfig::turbopack_mangle_export_names`. An explicit value always wins either way.
+    turbopackMangleExportNames: isStableBuild() ? false : undefined,
   },
   htmlLimitedBots: undefined,
   bundlePagesRouterDependencies: false,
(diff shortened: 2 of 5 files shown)


lukesandberg (changes requested):
given the desire for more rapid releases we should only make this default true in canary releases, then on in stable is a separate small decision (could imediately stack that on top)

dianatofficial (commented):

crates/next-core/src/next_config.rs
```diff
@@ -2572,12 +2573,18 @@ impl NextConfig {
     }
 
     #[turbo_tasks::function]
-    pub fn turbopack_mangle_export_names(&self) -> Vc<bool> {
-        Vc::cell(
-            self.experimental
+    pub async fn turbopack_mangle_export_names(&self, mode: Vc<NextMode>) -> Result<Vc<bool>> {
+        Ok(Vc::cell(match *mode.await? {
+            // Mangling requires minification with mangling enabled, which never happens in
+            // development.
```
sokra: Do not forbid it. The defaults in next.js should handle that.

sokra: Make sure it defaults to true for build and false for dev, but don't enforce it

docs/01-app/03-api-reference/08-turbopack.mdx
```diff
@@ -432,7 +432,7 @@ Additionally, the following experimental options are available under `experiment
...
-| `turbopackMangleExportNames`                                                                                 | Shorten the export names modules expose to each other, to reduce bundle size. Requires minification with mangling; a module whose export names can be observed (an escaping namespace object, a dynamic `import()`, a CommonJS `require()`) keeps its original names. | `false`       | `false`                       |
+| `turbopackMangleExportNames`                                                                                 | Shorten the export names modules expose to each other, to reduce bundle size. Requires minification with mangling; a module whose export names can be observed (an escaping namespace object, a dynamic `import()`, a CommonJS `require()`) keeps its original names. | `false`       | `true` on canary releases     |
```
sokra: > Requires minification with mangling

That's not required only recommended

packages/next/src/server/config-shared.ts
```diff
@@ -2337,6 +2338,7 @@ export const defaultConfig = Object.freeze({
     turbopackInferModuleSideEffects: true,
     turbopackPluginRuntimeStrategy: 'childProcesses',
     turbopackSharedRuntime: !isStableBuild(),
+    turbopackMangleExportNames: !isStableBuild(),
```
sokra:
```suggestion
    turbopackMangleExportNames: isStableBuild() ? undefined : false,
```

sokra:
This should also enable mangling in the turbopack test cases by default

lukesandberg (approved):

PR #97679, opened 2026-08-21
[DO NOT MERGE] Test React browser bailouts across CI

devjiwonchoi:
Stacked on #96844.

### Why?

Run the full CI suite with React browser bailouts enabled before changing the product default.

### How?

Set a CI-only environment override that enables `experimental.reactBrowserBailout` by default while preserving explicit config values, and forward it to deploy tests.


dianatofficial (commented):

PR #97683, opened 2026-08-21
[backport] [PPF] Remove `unstable_eager`

lubieowoce:
Backports #97524 to `next-16-3`.

Stacked on top of #97635 (the #97503 backport), which this depends on.

### Backport notes

Everything applied cleanly except two files, which were hand-ported:

- `create-transport-tree-from-loader-tree.ts` doesn't exist on `next-16-3` — it's the canary rename of `create-flight-router-state-from-loader-tree.ts`. The relevant logic is byte-identical between the two, so the hunks transplanted directly; `partialPrefetching` is narrowed on 3 signatures here vs. 5 on canary.
- The `Boolean(...)` coercion canary added in `create-component-tree.tsx` belongs in `app-render.tsx` here, because `next-16-3` still calls `createFlightRouterStateFromLoaderTree` from `getRSCPayload`. `walk-tree-with-flight-router-state.tsx` needs a second coercion for the same reason.

All 15 test files applied unchanged.

### Note before merging

This removes `unstable_eager` on a stable release line, so `export const prefetch = 'unstable_eager'` and `partialPrefetching: 'unstable_eager'` will start failing validation for anyone who adopted the flag in 16.3.x. That's intentional upstream, but it's a release-policy call worth making consciously.


dianatofficial (commented):

lubieowoce:
@dianatofficial your review was not requested ([again](https://github.com/vercel/next.js/pull/97635#pullrequestreview-4999508586)). this is not helpful to us, please don't do this

(with comments like these there's a good chance they're authored by a machine, but in case this reaches a human: if you continue doing unwanted reviews like these i'm gonna have to ban you from the repo)

PR #97692, opened 2026-08-21
docs: await params in use server inline example

see-stack:
Since Next.js 15, `params` is a Promise. The inline `use server` example on this page still typed `params` as `{ id: string }` and accessed `params.id` synchronously, which is invalid and inconsistent with the async `params` contract documented in the sibling `page.mdx`, `layout.mdx`, and `route.mdx` pages.

- Updated `params` to `Promise<{ id: string }>` and awaited it before accessing `id`.
- Replaced synchronous `params.id` reads with the resolved `id` value.
- Updated both TypeScript and JavaScript examples.
- Adjusted the `highlight` line numbers to account for the added `await` line.

Docs-only change; no behavior change.


dianatofficial (commented):

PR #97693, opened 2026-08-21
docs: note that response headers must be set before calling handle() in a custom server

### LINKED ISSUE #82387: Docs: Breaking changes in res.setHeader() timing behavior with handle(req, res) in custom servers (Next.js 13/14+)

Hi team! 👋 
First off, thank you for the amazing work on Next.js! it’s been a joy to use and continually improves with each release.

While upgrading from Next.js 12 to 14 in a project that uses a custom Fastify server, I noticed a subtle behavior change that wasn’t obvious from the migration guide or release notes:

### ❌ No longer works in Next.js 14 
```ts
// works in Next.js 12
serve.all('/*', async (req, reply) => {
  await handle(req.raw, reply.raw)

  reply.raw.setHeader('Set-Cookie', 'sessionId=abc123; Max-Age=2592000')

  reply.hijack()
})
```

### ✅ Works in Next.js 14
```ts
serve.all('/*', async (req, reply) => {
  reply.raw.setHeader('Set-Cookie', 'sessionId=abc123; Max-Age=2592000')

  await handle(req.raw, reply.raw)

  reply.hijack()
})
```

After some digging, I realized that handle() now likely begins streaming the response immediately, which means the headers must be set beforehand. This makes sense from an implementation perspective, but it might catch developers like us off guard, especially if they’re migrating custom server setups from v12.

Would it be possible to add a note about this to the [Next.js 14 release notes](https://nextjs.org/blog/next-14) or [custom server documentation](https://nextjs.org/docs/pages/building-your-application/routing/custom-server)? Even a short warning that res.setHeader() must come before handle() would go a long way for those using Fastify or Express with Next.js.

Thank you again for your hard work and time 🙏

### Environment

- Next.js: 14.2.14
- Server: Fastify 4.11.0 (custom server)
- Runtime: Node 20.18.0
- Reproducible: Yes, with minimal custom server setup

https://nextjs.org/blog/next-14

icyJoseph:
Fixes: #82387

A custom server that calls `res.setHeader()` after `await handle(req, res)` silently loses the header, because Next.js has already begun sending the response by the time `handle` resolves. Unlike a normal Node `setHeader()` after `headersSent`, this does not throw, so the header just never reaches the client. This adds a note to the custom server guide showing the correct ordering and pointing out that the same applies to framework wrappers such as `reply.raw` in Fastify.

### DIFF
--- a/docs/01-app/02-guides/custom-server.mdx
+++ b/docs/01-app/02-guides/custom-server.mdx
@@ -59,6 +59,16 @@ app.prepare().then(() => {
 })
 ```
 
+> **Set your own response headers before calling `handle`**: Next.js begins sending the response inside `handle(req, res)`, so by the time its promise resolves `res.headersSent` is already `true` and any later `res.setHeader()` call is silently discarded. Set headers you own, such as `Set-Cookie`, before handing the request to Next.js:
+>
+> ```js
+> res.setHeader('Set-Cookie', 'sessionId=abc123; Max-Age=2592000')
+>
+> await handle(req, res)
+> ```
+>
+> The same ordering applies when wrapping Next.js in another framework, for example `reply.raw` in Fastify or `res` in Express.
+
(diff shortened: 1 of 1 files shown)


dianatofficial (commented):

aurorascharff (approved):

PR #97695, opened 2026-08-21
Add Cache Components option to create-next-app

acdlite:
When initializing a new project with create-next-app, the interactive prompt will now ask the user if they want to enable Cache Components.

For now the option defaults to off. We'll flip the default back on once there's a flow that guides users to the right recommendation based on the kind of site they're building. The option can still be enabled interactively, or with the `--cache-components` flag.

If the user chooses to enable Cache Components, the project will also be initialized with Partial Prefetching enabled. There's no separate prompt for Partial Prefetching; the only reason they are separate configs is because Cache Components shipped in 16.0 before Partial Prefetching existed.

This is not a breaking change because it does not change any behavior in Next.js itself, only the create-next-app CLI. It has no impact on any existing Next.js projects.

### DIFF
--- a/packages/create-next-app/create-app.ts
+++ b/packages/create-next-app/create-app.ts
@@ -43,6 +43,7 @@ export async function createApp({
   bundler,
   disableGit,
   reactCompiler,
+  cacheComponents,
   agentsMd,
 }: {
   appPath: string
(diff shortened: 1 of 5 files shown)


dianatofficial (commented):

gnoff (approved):
Let's start with false for now til we figure out the flow of guiding the user on which to choose. we can then omit the flow for 16.4 with our improvements to the static story

PR #96717, opened 2026-08-05
Avoid reporting client-aborted RSC streams as request errors

eejihoon:
### What?

- Pass the response-backed abort signal to native Node RSC renders.
- Apply the same behavior to the Cache Components development cache-bypass path.
- Add regression coverage for client-aborted RSC responses.

### Why?

- A client disconnect currently destroys the destination stream before React receives the expected `ResponseAborted` reason.
- React consequently reports `"The destination stream closed early."` through `onRequestError`.
- Client cancellation is expected request lifecycle behavior, not a server rendering failure.

### How?

- Abort the React render when the underlying Node response closes.
- Verify that cancelling after the first RSC chunk does not invoke `onRequestError`.
- Cover development and production with Turbopack and Webpack.

Fixes #96704

### Validation

- `pnpm test-dev-turbo test/e2e/on-request-error/basic/basic.test.ts`
- `pnpm test-dev-webpack test/e2e/on-request-error/basic/basic.test.ts`
- `pnpm test-start-turbo test/e2e/on-request-error/basic/basic.test.ts`
- `pnpm test-start-webpack test/e2e/on-request-error/basic/basic.test.ts`
- `pnpm --filter=next build`

PR #96715, opened 2026-08-05
Don't report a client-aborted RSC stream as a render error

lazerg:
When a client disconnects mid-stream, `pipeNodeReadableToNodeResponse` destroys the readable React is piping into. React sees its destination close before the render ended and aborts with a generic `Error("The destination stream closed early.")`, which `onError` has no way to distinguish from a real render error, so it ends up in `onRequestError` as a server render error.

The Web streams path doesn't have this problem: cancelling the readable passes the `ResponseAborted` reason straight through to React. This does the same for the Node path by aborting the pipeable with that reason as soon as the passthrough is destroyed early, before React's own close handler runs. Genuine render errors still go through untouched.

Fixes #96704

Issue #96704, opened 2026-08-05
Client-aborted RSC stream is reported to onRequestError as "The destination stream closed early."


eejihoon:
### Link to the code that reproduces this issue

https://github.com/eejihoon/next-rsc-abort-on-request-error-repro

### To Reproduce

1. Clone https://github.com/eejihoon/next-rsc-abort-on-request-error-repro
2. Run `pnpm install`.
3. Run `pnpm build`.
4. Run `pnpm start`.
5. In a second terminal, run `pnpm reproduce`.
6. Observe the output of the Next.js server after the reproduction client reads the first RSC chunk and aborts the request.

### Current vs. Expected behavior

Current behavior

After the client receives the first chunk of a streaming RSC response and cancels the request, Next.js reports the cancellation to `onRequestError` as a render error:

```text
ON_REQUEST_ERROR {"name":"Error","message":"The destination stream closed early."}
```

The reproduction does not throw an application error. It only suspends a Server Component so that the RSC response remains open, reads the first chunk, and then cancels the client request.

Expected behavior

A client disconnect during an RSC response should be treated as an expected response abort. It should not be passed to `onRequestError` as a server render error.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 22.22.3
  npm: 10.9.8
  Yarn: 1.22.22
  pnpm: 10.33.4
Relevant Packages:
  next: 16.3.1-canary.3
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Instrumentation, Loading UI and Streaming, Runtime, Error Handling

### Which stage(s) are affected? (Select all that apply)

next start (local)

### Additional context

This also reproduces with Next.js 16.3.0. It was originally observed in production immediately after upgrading from the 16.2 release line.

A closely matching report, #96465, was automatically closed because it did not include a public minimal reproduction. The linked repository provides that reproduction and verifies the behavior against the current canary release.

Initial investigation (hypothesis)

- `pipeNodeReadableToNodeResponse` calls `readable.destroy()` when the response closes without an abort reason.
- React's pipeable RSC destination close handler wraps that close as the generic `Error("The destination stream closed early.")`.
[...]

PR #97698, opened 2026-08-21
fix: reuse a single drain listener when piping Node streams through gzip

hamidrezahanafi:
## Summary

On Next 16.3+, App Router HTML and RSC responses are piped with `pipeNodeReadableToNodeResponse`. Under gzip backpressure it called `res.once('drain')` on every write that returned `false`, which leaks one listener per backpressured write onto the zlib stream:

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 drain listeners added to [Gzip]. MaxListeners is 10.
```

The `compression` middleware overrides only `res.on`, forwarding it to the zlib stream once that stream exists (or buffering it until then). It leaves `removeListener` pointing at the response, so a `once` listener is added to the stream but never removed from it. `off` cannot reach it either.

`createWriterFromResponse`, directly above in the same file, already uses a single `res.on('drain')` for the response lifetime. This makes the Node-readable path match: pause on `write() === false`, resume from one listener, `off` on `close`.

Resuming on that event is correct for both cases. With compression active the wrapped `res.write` writes into the zlib stream, so `false` means the zlib stream's buffer is full and its `drain` is the right signal — the middleware separately wires the socket's `drain` to `stream.resume()`. With compression inactive the listener stays on the response, where `off` does remove it.

The pause is latched so a `drain` only resumes a readable that this function paused, and a `drain` arriving after teardown is ignored.

### Scope

- Distinct from #96173 / `releaseCompressionStream`, which destroys the zlib stream on **abort**. This warning fires on a **live** stream that is still flowing: slow clients, large streamed HTML/RSC.
- Deliberately leaves the bare `readable.destroy()` in the `close` handler alone. That is the separate abort-reason bug in #96704, with fixes already open in #96715 and #96717.
- Not fixable by bumping the vendored `compression`: as #96173 established, upstream `compression@1.8.1` still has no premature-close handling, and it is a prebuilt bundle here. `setMaxListeners` or `compress: false` would only hide this.
- Mentioned as a `compress: false` workaround in #95130.

## Verification

`packages/next/src/server/pipe-readable.test.ts` drives real backpressure (48 × 64 KiB through a throttled reader) and asserts exactly one `drain` registration on the zlib stream, no `MaxListenersExceededWarning`, and a complete, byte-correct payload after `gunzip`. The payload assertion is what catches a latch that stalls the stream.

Confirmed it fails without the fix. Measured against the real compiled function on Next 16.3, patching only the drain handling:

| | before | after |
| --- | --- | --- |
| `drain` registrations on `[Gzip]` | 48 (one per backpressured write) | **1** |
| `MaxListenersExceededWarning` | 1 | **0** |
| decompressed payload | complete | complete |

At 120 × 64 KiB the before-count scales with writes (120 registrations); the after-count stays at 1. Separately reproduced end to end on a Next 16.3 standalone production build with `compress: true`: slow-consuming a large gzipped page emitted the warning, and after this change it did not, with `Content-Encoding: gzip` still set and valid gzip on the wire.

I could not run `pnpm test` in this checkout, so please treat CI as the source of truth for harness integration. The assertions themselves were validated against the real function as described above.

### DIFF
--- a/packages/next/src/server/lib/release-compression-stream.test.ts
+++ b/packages/next/src/server/lib/release-compression-stream.test.ts
@@ -12,9 +12,11 @@ type StreamRecord = { closed: boolean }
 /** Records every zlib stream the middleware creates, and whether it closed. */
 function trackCompressionStreams(): {
   records: StreamRecord[]
+  streams: Array<zlib.Gzip | zlib.Deflate>
   restore: () => void
 } {
   const records: StreamRecord[] = []
+  const streams: Array<zlib.Gzip | zlib.Deflate> = []
   const originals: Array<[string, unknown]> = []
 
   for (const name of ['createGzip', 'createDeflate'] as const) {
@@ -31,6 +33,7 @@ function trackCompressionStreams(): {
           record.closed = true
         })
         records.push(record)
+        streams.push(stream)
         return stream
       },
     })
@@ -38,6 +41,7 @@ function trackCompressionStreams(): {
 
   return {
     records,
+    streams,
     restore: () => {
       for (const [name, original] of originals) {
         Object.defineProperty(zlib, name, {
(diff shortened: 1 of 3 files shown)


dianatofficial (commented):

gaojude (approved):

PR #97700, opened 2026-08-21
Add no-adhoc-sleep lint rule and migrate ad-hoc sleep promises

wbinnssmith:
Adopts #96841. Closes #96841.

## What?

Adds an autofixable `@next/internal/no-adhoc-sleep` ESLint rule and migrates fixed delays to `wait()` in runtime code and `waitFor()` in tests.

## Why?

Hand-written sleep promises were duplicated throughout the repository. Canonical helpers keep fixed delays consistent and make `retry()` the clear choice for condition polling.

## How?

The rule only rewrites the exact `new Promise((resolve) => setTimeout(resolve, ms))` pattern when behavior is preserved. It skips value-returning callbacks, timers with extra arguments, comments, and shadowed helpers, and inserts the correct import for runtime or test scope. Fixture apps remain excluded because `next-test-utils` is Jest-only.

`waitFor(number)` delegates to `wait()`, leaving one sleep implementation while preserving every migrated delay. Related trace snapshots and timing-sensitive tests were updated for the resulting dependency and CI behavior.

## Verification

- Full bootstrap build, lint, types, and formatting
- Rule, unit, development, and production integration tests


dianatofficial (commented):

PR #97702, opened 2026-08-21
Fix transient empty titles during optimistic navigation

gnoff:
### What?

- Add a head-specific static-prefetch hint on the route root, measured independently from the route body.
- Use that hint to prefetch a complete static head even when an unrelated layout requires runtime data.
- Preserve the current head while the proven-static destination head is pending during an optimistic navigation.
- Add a controlled soft-navigation regression based on the reported reproduction.

### Why?

The head is a route-level pseudo-segment (`/_head`), but its scheduler decision reused the route body `ShouldAttemptStaticPrefetch` hint. An unrelated dynamic layout could therefore force a fully static head onto the runtime path. If navigation raced the pending prefetch, optimistic routing replaced that pending head with an empty string, transiently clearing `document.title`.

This keeps the existing dynamic-metadata behavior. Only a head whose independent build-time probe completed statically gets the new transition behavior; partial heads still use the runtime completion path.

Reproduction: https://github.com/aurorascharff/reproductions/tree/codex/transient-empty-title/cases/metadata-suspense-title-timing

### How?

- Reuse the segment response completeness probe while collecting build-time prefetch hints and store the result on the root hint node.
- Keep a static head outlined when the dynamic route body would prevent its inlined carrier from being requested statically.
- Thread the root hint through optimistic cache-node construction so a pending static head suspends instead of becoming an empty head.

### Verification

- `pnpm --filter=next types`
- Metadata soft-navigation e2e suite in Turbopack and webpack production modes

### DIFF
--- a/packages/next/src/client/components/app-router-state.ts
+++ b/packages/next/src/client/components/app-router-state.ts
@@ -321,6 +321,7 @@ export function navigateToKnownRoute(
   const accumulation: NavigationRequestAccumulation = {
     separateRefreshUrls: null,
     scrollRef: null,
+    shouldWaitForStaticHead: false,
   }
   // We special case navigations to the exact same URL as the current location.
   // It's a common UI pattern for apps to refresh when you click a link to the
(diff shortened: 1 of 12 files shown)


dianatofficial (commented):

PR #97705, opened 2026-08-21
turbo-persistence: give key block searches a dense region of their own

lukesandberg:
### What?

Rearranges the bytes inside a key block so that the bytes a lookup's binary search compares are contiguous, instead of interleaved with key and value payload the search never reads.

| block kind | before | after |
| --- | --- | --- |
| fixed, `HashThenKey` | `[hash\|key\|val]*n` | `[hash]*n` then `[key\|val]*n` |
| fixed, `KeyOnly` | `[key\|val]*n` | `[key]*n` then `[val]*n` |
| variable, `HashThenKey` | `[type\|pos]*4B` then `[hash\|key\|val]…` | `[hash\|type\|pos]*12B` then `[key\|val]…` |

### Why?

The binary search routines are a random access pattern through the hash/key ranges,  we only need to look at value payloads when we find a plausible match.  So by organizing all the 'random access data' in denser memory range we optimize cache/page fault behavior.

`get` perf impacts are small but real.

| shape | in-block search before → after | as % of a ~1070 ns `get()` |
| --- | --- | --- |
| key=8, value=4 inline | 166 → 146 ns | 1.9% |
| key=4, small value ref | 144 → 119 ns | 2.4% |

The reason the impacts are small is mostly because the 'working sets' of the benchmarks are not large enough and the page size of macos is quite large.


dianatofficial (commented):

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
```diff
@@ -576,6 +578,10 @@ pub struct StreamingSstWriter<E: Entry> {
     // Reusable buffer for building key blocks
     key_buffer: Vec<u8>,
 
+    // Reusable buffer for the tail region of a key block (the bytes the binary search never
+    // probes), appended to `key_buffer` when the block is finished.
```
lukesandberg: key_value_buffer?

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
```diff
@@ -636,6 +642,7 @@ impl<E: Entry> StreamingSstWriter<E> {
                 MIN_SMALL_VALUE_BLOCK_SIZE + MAX_SMALL_VALUE_SIZE,
             ),
             key_buffer: Vec::with_capacity(MAX_KEY_BLOCK_SIZE),
+            key_tail_buffer: Vec::with_capacity(MAX_KEY_BLOCK_SIZE),
```
lukesandberg: should these both be half the MAX_KEY_BLOCK_SIZE????

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
```diff
@@ -1242,15 +1267,27 @@ const FIXED_KEY_BLOCK_HEADER_SIZE: usize = 6;
...
+/// `HashThenKey`, the key for `KeyOnly`), and everything else follows in the tail region, addressed
+/// by the same entry index. See the module docs on key block layout.
 struct FixedKeyBlockBuilder<'l> {
+    /// Receives the header and then the search region.
     buffer: &'l mut Vec<u8>,
+    /// Accumulates the tail region, appended to `buffer` by [`Self::finish`].
+    tail: &'l mut Vec<u8>,
     /// Whether each entry writes its own type byte (set for mixed-type blocks).
     per_entry_type: bool,
+    /// Whether the search region holds hashes (`HashThenKey`) rather than keys (`KeyOnly`).
```
lukesandberg: use layout enum?

turbopack/crates/turbo-persistence/README.md
```diff
@@ -208,17 +210,39 @@ during binary search.
...
+about 16 kB to 480 bytes.
+
+Byte count is unchanged — this is a permutation, not a size change, so it does not affect how many
+entries fit in a block. Note the split point follows the *comparison*, not the key/value boundary:
+for a hashed block the key moves into the tail with the value, because the search reads the key only
+when two hashes are equal, which for a 64-bit hash is vanishingly rare.
+
+Entry positions for index `i` are `header_size + i * search_stride` and
+`header_size + entry_count * search_stride + i * tail_stride`, both with no indirection. The writer
+automatically selects fixed-size format when all entries in a block qualify; otherwise falls back to
```
lukesandberg: simplify comment

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
lukesandberg: :robot: Renamed to `key_value_buffer` in 88d797f. The tail holds the key (for `HashThenKey`, where the search region holds hashes) plus the value, so the name now says what is in it rather than where it lands.

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
lukesandberg: :robot: Good catch — yes. `should_flush` caps the *combined* size of a block at `MAX_KEY_BLOCK_SIZE`, and the two-region split is a permutation of those same bytes, so giving each buffer a full `MAX_KEY_BLOCK_SIZE` reserved twice what a block can hold. Both are now `MAX_KEY_BLOCK_SIZE / 2` in 88d797f.

Worth noting the halves are only first-block hints, not a real bound: the split is not 50/50 (a `HashThenKey` search region is 8 bytes/entry while its tail carries key+value), so I also made the per-block reservation exact. `buffer` was already `reserve`d per block; `tail` never was, it just grew by `extend`. It now reserves `entry_count * tail_stride`, so each buffer is sized to the region it actually receives and the initial capacity stops mattering after the first block.

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
lukesandberg: :robot: Done in 88d797f. `new()` already took `layout`, so the `search_region_is_hash: bool` was a lossy re-derivation of it — the struct now just stores the `KeyBlockLayout`. The two `debug_assert!`s became `debug_assert_eq!` against the variant, which also makes a failure print which layout it actually got.

Same change let the stride computation read off the layout directly (`HashThenKey` => hash in the search region, key in the tail; `KeyOnly` => key is the search region) instead of the `hash_len + key_size - search_stride` arithmetic.

turbopack/crates/turbo-persistence/README.md
lukesandberg: :robot: Condensed in 88d797f — four paragraphs down to two. Dropped the sentence that restated the split after describing it, and the standalone "this exists for cache behaviour" framing, folding the rationale into the mechanism. Kept the 16 kB -> 480 bytes example, the comparison-vs-key/value split point, the note that total bytes are unchanged, and the offset arithmetic.

turbopack/crates/turbo-persistence/README.md
```diff
@@ -115,39 +115,40 @@ The hashes are sorted.
 
 - 1 byte block type (1: key block with hash, 2: key block without hash)
 - 3 bytes entry count
-- foreach entry
+- offset table, foreach entry
```
bgw:
```suggestion
- offset table: foreach entry
```

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
bgw: `lookup_variable_key_block`?

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -1146,10 +1172,14 @@ fn entry_val_size(ty: u8) -> Result<usize> {
...
+///
+/// The trailing 4 bytes of every entry pack 1 byte of type into the top of a 3-byte BE offset.
+/// `HashThenKey` entries carry the key's 8-byte hash ahead of that word — see
+/// [`KEY_BLOCK_TABLE_ENTRY_SIZE_WITH_HASH`].
 #[inline(always)]
-fn read_offset_entry(offsets: &[u8], index: usize) -> (u8, usize) {
-    let base = index * 4;
+fn read_offset_entry(offsets: &[u8], index: usize, table_stride: usize) -> (u8, usize) {
+    // The offset word is last, so skip any hash that precedes it.
+    let base = index * table_stride + (table_stride - KEY_BLOCK_TABLE_ENTRY_SIZE_NO_HASH);
```
bgw: Accept `hash_len` as an argument and use it to simplify the `base` calculation slightly here?

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -1218,28 +1248,119 @@ fn fixed_value_layout(block: &[u8], header_type: u8) -> Result<FixedValueLayout>
...
+
+    /// Bytes per entry in the tail region, for sizing that region.
+    pub fn tail_stride(&self) -> usize {
+        self.tail_stride
+    }
+
+    /// Offset of the tail region, relative to the start of the entry data.
+    pub fn tail_start(&self) -> usize {
+        self.tail_start
+    }
```
bgw: Just make these fields pub

bgw (approved):

PR #97711, opened 2026-08-21
Support immutable static assets with `output: 'export'`

mischnic:
There was no real reason why this wasn't done thus far.
`output: 'export'` already uses the adapter anyway.


dianatofficial (commented):

lukesandberg (approved):

PR #96471, opened 2026-08-01
skill: add Partial Prefetching optimizer

aurorascharff:
## Summary

- add a Partial Prefetching optimizer Skill for selected source Links and destinations
- keep prefetch stages and Link policy in the public [Optimizing prefetching guide](https://nextjs.org/docs/app/guides/optimizing-prefetching), and rely on the same static-shell documentation as the Cache Components optimizer for cache placement, Suspense, loading UI, and authorization
- keep the Skill focused on the product decision, production `instant()` loop, differential verification, and mutation freshness when an optimization adds a cache around mutable data
- reference the canonical [`unstable_prefetch()`](https://nextjs.org/docs/app/api-reference/functions/prefetch) and [`unstable_navigation()`](https://nextjs.org/docs/app/api-reference/functions/navigation) docs instead of explaining the APIs again in the Skill
- link cache-writing optimizations to the canonical [Revalidating](https://nextjs.org/docs/app/getting-started/caching-and-revalidating#revalidating) documentation and verify a populated-cache mutation in the Skill and eval
- bundle a self-contained production rig aligned with the related adoption and optimizer Skills, tuned to record each source-link and prefetch-stage contract
- add an eval that compares baseline, the bundled guide, and the local Skill against the same prefetched-UI and cache-freshness goal
- document the optimizer in the AI agents guide

This follows the same guide/Skill split as #97737. Partial Prefetching adoption preserves existing behavior; this optimizer is the later decision about what is worth preparing before a click and what should wait for navigation.

Related: #97712, #97737, #97759, #98122, #98175

## Eval

One run per treatment with Claude Opus 4.8 and Claude Sonnet 4.5, using a fixed Claude Haiku 4.5 judge. The bundled-guide and Skill treatments use a local Next.js tarball containing the new guide and API references.

| Model | Treatment | Criteria | Score change | Overall | Duration | Tokens processed | Total run cost | Cost change |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Opus 4.8 | Baseline | 10/10 (100%) | — | Pass | 18m 42s | 3.39M | $3.46 | — |
| Opus 4.8 | Bundled guide | 9/10 (90%) | -10 pp | Fail | 13m 37s | 3.02M | $2.84 | -18.0% |
| Opus 4.8 | Local Skill | 10/10 (100%) | +10 pp | Pass | 17m 22s | 2.80M | $2.54 | -10.6% |
[...]

PR #97712, opened 2026-08-21
docs(skills): preserve prefetched UI during Partial Prefetching adoption

aurorascharff:
## What

- Documents how legacy `prefetch={true}` behavior changes with Partial Prefetching and how to choose the UI worth preserving.
- Updates the adoption skill to capture a passing flag-off `instant()` baseline, use the same assertions to drive the migration, and retain them as regression coverage.
- Adds a reusable production test rig, keeps additional per-link optimization separate from adoption, and surfaces `prefetch={false}` links that may no longer be needed.
- Applies the shared test-rig fixes discovered during dogfooding to the Cache Components optimizer.

## Why

Legacy `prefetch={true}` could include uncached dynamic content. With Partial Prefetching, it includes the shared App Shell and cached URL-specific content instead. A passing build or clean insight sweep does not prove that the intended prefetched UI survived the migration, so the skill now records that contract before adoption and verifies it unchanged afterward. The guide presents the same test workflow as an optional programmatic check for manual adopters.

## Testing

- Prettier, Alex, and skill frontmatter validation
- Dogfooded on v0 in https://github.com/vercel/v0/pull/28842
- Partial Prefetching adoption eval: 6/6 in https://github.com/vercel/next.js/pull/97759


dianatofficial (commented):

skills/next-partial-prefetching-adoption/SKILL.md
```diff
@@ -126,19 +152,25 @@ Then check in with the user. Speak their language — no insight slugs or step l
...
+The audit marked candidates beyond the already-preserved legacy contract instead of deciding them. Grep for `TODO(per-link-prefetch)` and walk the list with the user in one conversation. The question per route is whether they want the additional URL-dependent content prefetched ahead of the click, or streaming in after navigation is fine. A per-link prefetch costs a server invocation per prefetchable link — the guide's [trade-offs](https://nextjs.org/docs/app/guides/optimizing-prefetching#trade-offs) section is the checklist. Don't make these calls alone.
+
+Where the answer is no, delete the marker and leave the route on the App Shell default. Where the answer is yes, choose one of these paths:
+
+- **Manual:** follow the [Optimizing prefetching guide](https://nextjs.org/docs/app/guides/optimizing-prefetching), confirm the opted-in link against a production run, and delete the marker when the selected result is verified.
```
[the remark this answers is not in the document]
aurorascharff: Fixed in afb7c8c138. The optimizer has not merged into canary yet, so both references now link to the live optimizer PR (#96471) instead of the future skill directory.

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -40,14 +40,16 @@ Adopt Partial Prefetching in this project using the next-partial-prefetching-ado
...
-2. [Audit routes for URL data](#auditing-routes-for-url-data), resolving the [URL data outside of Suspense](/docs/messages/instant-shell-url-data) insight.
+1. Before you enable the global flag, [audit every `<Link prefetch={true}>`](#auditing-link-prefetchtrue-calls) and choose which prefetched UI to preserve. You can use [`instant()` tests](#preserve-existing-prefetched-ui) to add regression coverage. When the adoption is large enough that reviewers need smaller diffs, or adopted routes should reach users before the rest, [adopt incrementally](#adopting-incrementally) with the flag off, guided by the [dynamic data during prefetching](/docs/messages/instant-link-prefetch-partial) insight.
```
icyJoseph: Feels like here we'd need 1.1, 1.2 and 1.3 — the step is getting very overloaded.

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -40,14 +40,16 @@ Adopt Partial Prefetching in this project using the next-partial-prefetching-ado
...
+2. [Enable `partialPrefetching`](#enable-partial-prefetching), then [audit routes for URL data](#auditing-routes-for-url-data), resolving the [URL data outside of Suspense](/docs/messages/instant-shell-url-data) insight.
 3. Optionally, [prefetch URL data](#prefetching-url-data) on the routes where streaming in after navigation isn't enough.
 
 Both insights are development-only and never block the build. They appear in the dev overlay with fix cards that link the docs page for each fix. If a route isn't ready to adopt, export [`instant = false`](/docs/app/api-reference/file-conventions/route-segment-config/instant) from its page or layout to opt the route out of instant-navigation validation, and return to it later.
 
 ## Enable Partial Prefetching
 
+For an existing app, complete the [`prefetch={true}` audit](#auditing-link-prefetchtrue-calls) before enabling Partial Prefetching.
```
icyJoseph:
```suggestion
> **Good to know**: In an existing app, [audit your `<Link prefetch={true}>` calls](#auditing-link-prefetchtrue-calls) first. The flag-off state is what those links deliver today.
```

The previous section is already mentioning this, I think here we can keep the reinforcement but pull it out from the prose

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -87,6 +91,18 @@ Each `<Link prefetch={true}>` used to prefetch its destination's dynamic content
 | [Reads URL data, or has cached content that depends on it](#url-data)             | Keep `prefetch={true}` to resolve the content ahead of the click.         |
 | [Delivers real-time content that must stay fresh per request](#real-time-content) | Remove `prefetch={true}` and let the content stream in.                   |
 
+### Preserve existing prefetched UI
```
icyJoseph: I think we can shrink wrap this section. Right now it reads like a Skill summary, rather than a reference for a reader. 

It is also introduces "preservation test" as a concept. 

The snippet below is just a broad suggestion.

```suggestion
### Preserve existing prefetched UI

An [`instant()`](/docs/app/guides/instant-navigation#prevent-regressions-with-e2e-tests) test locks in what a navigation shows immediately. With `partialPrefetching` off, write a client-navigation test for each audited link that asserts the UI you chose to preserve, and confirm it passes.

Then adopt the destination and rerun the same test unchanged. A failure names UI the App Shell no longer carries, so you can cache the data or subtree that restores it. A passing test stays as regression coverage.
```

I am not sure we need the content in 102 and 104  — 102 could go into the API reference for instant. The content in 104, reads like a Skill instruction/pre-condition, more, idk, suddenly it gets very specific.

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -87,6 +93,10 @@ Each `<Link prefetch={true}>` used to prefetch its destination's dynamic content
 | [Reads URL data, or has cached content that depends on it](#url-data)             | Keep `prefetch={true}` to resolve the content ahead of the click.         |
 | [Delivers real-time content that must stay fresh per request](#real-time-content) | Remove `prefetch={true}` and let the content stream in.                   |
 
+### Verify prefetched UI with tests
```
icyJoseph, line 87: Maybe a nit but the table above is a summary of the sections that follow — should it be placed at the end of this run

After, Real-time content, and before Adopting incrementally

aurorascharff: yeah true!

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -38,16 +38,20 @@ Adopt Partial Prefetching in this project using the next-partial-prefetching-ado
...
+2. [Enable `partialPrefetching`](#enable-partial-prefetching), then [move URL data behind Suspense](#move-url-data-behind-suspense), resolving the [URL data outside of Suspense](/docs/messages/instant-shell-url-data) insight.
 3. Optionally, [prefetch URL data](#prefetching-url-data) on the routes where streaming in after navigation isn't enough.
 
+For larger migrations, you can [adopt incrementally](#adopting-incrementally) with the global flag off. The [dynamic data during prefetching](/docs/messages/instant-link-prefetch-partial) insight identifies each destination that still uses the legacy full prefetch.
+
 Both insights are development-only and never block the build. They appear in the dev overlay with fix cards that link the docs page for each fix. If a route isn't ready to adopt, export [`instant = false`](/docs/app/api-reference/file-conventions/route-segment-config/instant) from its page or layout to opt the route out of instant-navigation validation, and return to it later.
```
icyJoseph, line 48: Item 1 used to end with

> guided by the [dynamic data during prefetching](/docs/messages/instant-link-prefetch-partial) insight.

And item 2 still ends with:

> resolving the [URL data outside of Suspense](/docs/messages/instant-shell-url-data) insight.

And then it was an OK flow with `Both insights` — with this format, the list ends, and we talk about larger migrations, and an insight, and then we follow with 49, as a reader I may not connect to the insight on item 2.

Maybe a nit, but it feels a bit disconnected. I guess, item 2 is doing too much now?

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -38,16 +38,20 @@ Adopt Partial Prefetching in this project using the next-partial-prefetching-ado
...
 
 ## Enable Partial Prefetching
 
+> **Good to know**: In an existing app, review the UI delivered by [`<Link prefetch={true}>`](#migrate-existing-full-prefetches) before enabling Partial Prefetching. This is the legacy full prefetch that the migration may change.
```
icyJoseph: > This is the legacy full prefetch that the migration may change

Should be?

> Those links use the legacy full prefetch, which change their behavior when Partial Prefetching is enabled.

docs/01-app/02-guides/adopting-partial-prefetching.mdx
```diff
@@ -61,31 +65,29 @@ const nextConfig: NextConfig = {
 export default nextConfig
 ```
 
-After enabling the flag, every `<Link>` prefetches its destination's App Shell, and `<Link prefetch={true}>` no longer includes the route's dynamic content. Audit those links next to preserve what they delivered. A new project has no legacy links to audit and is done here.
+After enabling the flag, every `<Link>` prefetches its destination's App Shell. A link with `prefetch={true}` can also resolve cached URL-specific content, but it no longer includes uncached dynamic content from the legacy full prefetch. This migration step doesn't apply to new projects.
```
icyJoseph: > This migration step doesn't apply to new project

Nit, but we haven't mentioned (in prose) migration just yet. And a new project doesn't need to migrate either, just adopt.

> This doesn't apply to new projects.

But it feels incomplete like that too

icyJoseph (approved):

PR #97713, opened 2026-08-21
Turbopack: Use serde_json's new RawValue::from_string_unchecked API in manifest and source map processing

bgw:
This is a new API added in: https://github.com/serde-rs/json/pull/1331

Without this, serde_json re-parses the value we just encoded to verify that it's valid JSON.

### DIFF
--- a/crates/next-core/src/next_manifests/mod.rs
+++ b/crates/next-core/src/next_manifests/mod.rs
@@ -208,8 +208,14 @@ impl Asset for BuildManifest {
             pages_chunk_group_bootstrap_params: self
                 .pages_chunk_group_bootstrap_params
                 .iter()
-                .map(|(k, v)| Ok((k.clone(), RawValue::from_string(v.to_string())?)))
-                .collect::<Result<FxIndexMap<_, _>>>()?,
+                .map(|(k, v)| {
+                    // SAFETY: These values are produced by `serde_json::to_string` in
+                    // `EvaluateChunk::chunk_group_bootstrap_params`.
+                    (k.clone(), unsafe {
+                        RawValue::from_string_unchecked(v.to_string())
+                    })
+                })
+                .collect(),
             chunk_loading_global: self.chunk_loading_global.clone(),
             ..Default::default()
         };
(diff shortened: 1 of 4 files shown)


crates/next-core/src/next_manifests/mod.rs
```diff
@@ -208,8 +208,14 @@ impl Asset for BuildManifest {
             pages_chunk_group_bootstrap_params: self
                 .pages_chunk_group_bootstrap_params
                 .iter()
-                .map(|(k, v)| Ok((k.clone(), RawValue::from_string(v.to_string())?)))
-                .collect::<Result<FxIndexMap<_, _>>>()?,
+                .map(|(k, v)| {
+                    // SAFETY: These values are produced by `serde_json::to_string` in
```
lukesandberg, line 212: a lot of hops between here and there

could we just replace all those RcStrs with RawValue?  i guess not because of turbotasks, but maybe we should introduce a newtype wrapper for a RawValueRcStr?

turbopack/crates/turbopack-core/src/source_map/utils.rs
```diff
@@ -216,11 +216,10 @@ pub async fn resolve_source_map_sources(
...
 fn unencoded_str_to_raw_value(unencoded: &str) -> Box<RawValue> {
-    RawValue::from_string(
-        serde_json::to_string(unencoded)
-            .expect("serialization of a utf-8 string should always succeed"),
-    )
-    .expect("serde_json::to_string should produce valid JSON")
+    let encoded = serde_json::to_string(unencoded)
+        .expect("serialization of a utf-8 string should always succeed");
+    // SAFETY: `encoded` was just produced by `serde_json::to_string`.
+    unsafe { RawValue::from_string_unchecked(encoded) }
```
lukesandberg, line 222: isn't this just `RawValue::to_raw_value(unencoded).expect(....)`?

lukesandberg (approved):

mischnic:
Is the performance benefit worth the unsafe?

dianatofficial (commented):

PR #97714, opened 2026-08-21
Reduce Turbopack cache size with per-family compression

lukesandberg:
## What?

Adds per-family compression configuration to `turbo-persistence` and configures the Turbopack filesystem cache according to each keyspace's access pattern:

| Family | Compression | Reason |
| --- | --- | --- |
| Infra | LZ4 | Preserve the existing fast encode/decode path |
| TaskMeta | LZ4 | Avoid HC encode CPU for a marginal whole-cache size gain |
| TaskCache | LZ4 | Values are inline and no value blocks compressed in measured caches |
| TaskData | zstd level 3 | Prioritize the dominant disk-size opportunity |

Each meta file records the family's compression algorithm as one explicit byte, which is validated against `FamilyConfig` when the database opens. SST block and blob layouts remain unchanged.

## Why?

LZ4 decode performance is important for cache query latency, but using one codec for every keyspace leaves a substantial disk-size opportunity in `TaskData`. The final split keeps Infra, TaskMeta, and TaskCache on the existing ordinary LZ4 path, while `TaskData` spends modestly more CPU to reduce persistent cache size.

LZ4 HC4 was evaluated and removed before finalizing: across three real caches it made TaskMeta compression **4.24× slower** for only **1.51%** total cache-directory savings. Infra and TaskCache made no compression calls. Removing HC4 does not affect reads because HC4 and ordinary LZ4 use the same decoder.

### Real (tiny) Next.js cache A/B test

Final results after removing HC4, using three cold and five warm runs per workload:

| Workload | Cache size | Cold/write | Warm/read |
| --- | ---: | ---:| ---: |
| `test/e2e/filesystem-cache` | **-19.45%** | +2.38% | +3.98% |
| `test/e2e/app-dir/app-rendering` | **-18.87%** | -0.81% | -2.69% |
| `test/e2e/app-dir/client-reference-chunking` | **-19.84%** | -0.80% | -3.70% |
| Combined medians | **-19.41%** (96.79 → 78.00 MiB) | **+0.23%** | **-1.19%** |

The mixed warm result on the smallest fixture and apparent improvements elsewhere are treated as noise; combined measurements show no material write/read regression.

### Temporary crate benchmark instrumentation

To demonstrate the cost of the retained codecs, temporary write/read benchmarks applied each codec to identical synthetic data. The benchmark code is not included in the final patch.

| Codec | Synthetic write | vs LZ4 | Uncached get | Cached get | DB size |
| --- | ---: | ---: | ---: | ---: | ---: |
| LZ4 | 38.173 ms | baseline | 7.3004 µs | 10.735 µs | 21.59 MiB |
| zstd3 | 57.790 ms | **+51.4%** | 6.2460 µs | 9.5323 µs | 21.49 MiB |

LZ4 HC4 was also evaluated during development: the same write benchmark took 256.31 ms (**+571.7%** vs LZ4), which motivated dropping it after the real-cache family breakdown showed little total-size benefit.

### Native binary size

For Linux x64 GNU, using `build-native-release` followed by the release pipeline's `strip -x`:

- base: 103,140,384 bytes
- final PR: 103,663,432 bytes
- delta: **+523,048 bytes (+0.507%)**

Removing HC4 saved **73,248 installed bytes** versus the earlier PR state; replacing bincode framing with the explicit codec byte saved another **18,048 bytes**. Together those review simplifications removed **91,296 installed bytes** (24,420 gzip bytes) from the earlier branch state.

## Vercel Site real application

> These measurements were captured before HC4 was removed. The final TaskData-only configuration should give up a small amount of size reduction and improve write CPU; the final small-app A/B above measures that configuration directly.

### Size

2.9G	canary
2.2G	this PR

a 25% savings!!!

### Performance

Cold build (n=5), warm build (n=3), medians:

| Scenario | Metric | Canary | Compression | Δ |
|---|---|---|---|---|
| **Cold** | wall (s) | 71.28 | 72.86 | +2.2% |
| | user (s) | 555.21 | 570.23 | +2.7% |
| | sys (s) | 94.49 | 95.00 | +0.5% |
| | maxRSS (GB) | 17.95 | 17.98 | +0.2% |
| **Warm** | wall (s) | 15.00 | 15.32 | +2.1% |
| | user (s) | 7.59 | 9.98 | +31% |
| | sys (s) | 12.79 | 12.87 | +0.6% |
| | maxRSS (GB) | 3.78 | 3.86 | +2.0% |

maxRSS converted at 1 GB = 2^30 bytes.

Looking at tracing data i see in a cold build `persist` span went from 5.06s duration using 63s cpu time to 6.04s using 73s cpu time, a ~20% regression.  Which explains the cpu regression and the time progression.

Because the time regression is in the `persist` span during shutdown the added latency is somewhat hidden.

Of course we also see a small wall/user time regression in warm builds due to the extra decompression costs for zstd.

### DIFF
--- a/turbopack/crates/turbo-persistence/Cargo.toml
+++ b/turbopack/crates/turbo-persistence/Cargo.toml
@@ -15,7 +15,6 @@ verbose_log = []
 [dependencies]
 anyhow = { workspace = true }
 auto-hash-map = { workspace = true }
-bincode = { workspace = true }
 bitfield = { workspace = true }
 byteorder = { workspace = true }
 crc32fast = { workspace = true }
--- a/turbopack/crates/turbo-persistence/Cargo.toml
+++ b/turbopack/crates/turbo-persistence/Cargo.toml
@@ -38,6 +38,7 @@ smallvec = { workspace = true }
 thread_local = { workspace = true }
 tracing = { workspace = true }
 xxhash-rust = { workspace = true }
+zstd = { workspace = true }
 
 [dev-dependencies]
 criterion = { workspace = true }
--- a/turbopack/crates/turbo-persistence/Cargo.toml
+++ b/turbopack/crates/turbo-persistence/Cargo.toml
@@ -15,6 +15,7 @@ verbose_log = []
 [dependencies]
 anyhow = { workspace = true }
 auto-hash-map = { workspace = true }
+bincode = { workspace = true }
 bitfield = { workspace = true }
 byteorder = { workspace = true }
 crc32fast = { workspace = true }
--- a/turbopack/crates/turbo-persistence/README.md
+++ b/turbopack/crates/turbo-persistence/README.md
@@ -54,8 +54,10 @@ Small value blocks are emitted once they accumulate at least `MIN_SMALL_VALUE_BL
 A meta file can contain metadata about multiple SST files. The metadata is stored in a single file to avoid having too many small files.
 
 - Header
-  - 4 bytes magic number (0xFE4ADA4A)
+  - 4 bytes magic number (0xFE4ADA4B)
   - 4 bytes key family
+  - 4 bytes compression codec tag (0: LZ4, 1: LZ4 HC, 2: zstd)
+  - 4 bytes signed compression level (0 for LZ4)
   - 4 bytes count of obsolete SST files
   - foreach obsolete SST file
     - 4 bytes sequence number of the obsolete SST file
@@ -88,7 +90,7 @@ The SST file contains only data without any header.
 
 #### Block Compression
 
-Blocks can be stored compressed or uncompressed. The compression algorithm is selected by the block's key-family configuration and is not encoded in the block. Changing a family's configured algorithm requires rewriting that database, consistent with the crate's no-cross-version-compatibility policy.
+Blocks can be stored compressed or uncompressed. The compression algorithm is recorded once in the containing SST's meta file and must match the key-family configuration used to open the database. It is not repeated in every block. Changing a family's configured algorithm requires rewriting that database, consistent with the crate's no-cross-version-compatibility policy.
 
 The 4-byte header distinguishes compressed from uncompressed storage:
 
@@ -227,7 +229,7 @@ The plain value compressed with dynamic compression. Each blob file has an 8-byt
 - 4 bytes: CRC32 checksum of the compressed data (u32 big-endian)
 - remaining bytes: value data compressed with the blob's key-family configuration
 
-The remaining bytes use the blob's key-family compression configuration. As with SST blocks, the algorithm is not encoded in the blob file. The checksum is verified on the compressed data **before** decompression when the blob is read.
+The codec is validated by the family's meta-file marker rather than repeated in each blob. The checksum is verified on the compressed data **before** decompression when the blob is read.
 
 ## Reading
 
--- a/turbopack/crates/turbo-persistence/README.md
+++ b/turbopack/crates/turbo-persistence/README.md
@@ -88,9 +88,11 @@ The SST file contains only data without any header.
 
 #### Block Compression
 
-Blocks can be stored compressed (LZ4) or uncompressed. The 4-byte header distinguishes them:
+Blocks can be stored compressed or uncompressed. The compression algorithm is selected by the block's key-family configuration and is not encoded in the block. Changing a family's configured algorithm requires rewriting that database, consistent with the crate's no-cross-version-compatibility policy.
 
-- **Header > 0**: Block is LZ4 compressed. Header value is the uncompressed length.
+The 4-byte header distinguishes compressed from uncompressed storage:
+
+- **Header > 0**: Block is compressed with the family's configured algorithm. Header value is the uncompressed length.
 - **Header = 0**: Block is stored uncompressed. Actual length is derived from block offsets.
 
 #### Block Checksum
@@ -223,9 +225,9 @@ The plain value compressed with dynamic compression. Each blob file has an 8-byt
 
 - 4 bytes: uncompressed length (u32 big-endian)
 - 4 bytes: CRC32 checksum of the compressed data (u32 big-endian)
-- remaining bytes: LZ4-compressed value data
+- remaining bytes: value data compressed with the blob's key-family configuration
 
-The checksum is verified on the compressed data **before** decompression when the blob is read.
+The remaining bytes use the blob's key-family compression configuration. As with SST blocks, the algorithm is not encoded in the blob file. The checksum is verified on the compressed data **before** decompression when the blob is read.
 
 ## Reading
 
--- a/turbopack/crates/turbo-persistence/README.md
+++ b/turbopack/crates/turbo-persistence/README.md
@@ -54,10 +54,10 @@ Small value blocks are emitted once they accumulate at least `MIN_SMALL_VALUE_BL
 A meta file can contain metadata about multiple SST files. The metadata is stored in a single file to avoid having too many small files.
 
 - Header
-  - 4 bytes magic number (0xFE4ADA4B)
+  - 4 bytes magic number (0xFE4ADA4A)
   - 4 bytes key family
-  - 4 bytes compression preset byte length
-  - `n` bytes bincode-encoded compression preset (standard config; variants in format order: LZ4, LZ4 HC4, zstd3)
+  - 4 bytes compression algorithm byte length
+  - `n` bytes bincode-encoded compression algorithm, which must match the configuration used to open the database
   - 4 bytes count of obsolete SST files
   - foreach obsolete SST file
     - 4 bytes sequence number of the obsolete SST file
@@ -90,7 +90,7 @@ The SST file contains only data without any header.
 
 #### Block Compression
 
-Blocks can be stored compressed or uncompressed. The compression algorithm is recorded once in the containing SST's meta file and must match the key-family configuration used to open the database. It is not repeated in every block. Changing a family's configured algorithm requires rewriting that database, consistent with the crate's no-cross-version-compatibility policy.
+Blocks can be stored compressed or uncompressed. The compression algorithm is specified in the meta file.
 
 The 4-byte header distinguishes compressed from uncompressed storage:
 
@@ -229,7 +229,7 @@ The plain value compressed with dynamic compression. Each blob file has an 8-byt
 - 4 bytes: CRC32 checksum of the compressed data (u32 big-endian)
 - remaining bytes: value data compressed with the blob's key-family configuration
 
-The codec is validated by the family's meta-file marker rather than repeated in each blob. The checksum is verified on the compressed data **before** decompression when the blob is read.
+The checksum is verified on the compressed data **before** decompression when the blob is read.
 
 ## Reading
 
--- a/turbopack/crates/turbo-persistence/README.md
+++ b/turbopack/crates/turbo-persistence/README.md
@@ -56,8 +56,8 @@ A meta file can contain metadata about multiple SST files. The metadata is store
 - Header
   - 4 bytes magic number (0xFE4ADA4B)
   - 4 bytes key family
-  - 4 bytes compression codec tag (0: LZ4, 1: LZ4 HC, 2: zstd)
-  - 4 bytes signed compression level (0 for LZ4)
+  - 4 bytes compression preset byte length
+  - `n` bytes bincode-encoded compression preset (standard config; variants in format order: LZ4, LZ4 HC4, zstd3)
   - 4 bytes count of obsolete SST files
   - foreach obsolete SST file
     - 4 bytes sequence number of the obsolete SST file
--- a/turbopack/crates/turbo-persistence/README.md
+++ b/turbopack/crates/turbo-persistence/README.md
@@ -56,8 +56,7 @@ A meta file can contain metadata about multiple SST files. The metadata is store
 - Header
   - 4 bytes magic number (0xFE4ADA4A)
   - 4 bytes key family
-  - 4 bytes compression algorithm byte length
-  - `n` bytes bincode-encoded compression algorithm, which must match the configuration used to open the database
+  - 1 byte compression algorithm, which must match the configuration used to open the database
   - 4 bytes count of obsolete SST files
   - foreach obsolete SST file
     - 4 bytes sequence number of the obsolete SST file
--- a/turbopack/crates/turbo-persistence/benches/mod.rs
+++ b/turbopack/crates/turbo-persistence/benches/mod.rs
@@ -1204,8 +1204,13 @@ fn bench_static_sorted_file_lookup(c: &mut Criterion) {
             // Create temp directory and write SST file
             let tempdir = tempfile::tempdir().unwrap();
             let sst_path = tempdir.path().join("00000001.sst");
-            let (meta, _file) =
-                write_static_stored_file(&entries, &sst_path, MetaEntryFlags::FRESH).unwrap();
+            let (meta, _file) = write_static_stored_file(
+                &entries,
+                &sst_path,
+                MetaEntryFlags::FRESH,
+                Compression::Lz4,
+            )
+            .unwrap();
 
             // Open the SST file
             let sst_meta = StaticSortedFileMetaData {
--- a/turbopack/crates/turbo-persistence/benches/mod.rs
+++ b/turbopack/crates/turbo-persistence/benches/mod.rs
@@ -622,7 +622,7 @@ fn prefill_multi_value_database(
         family_configs: [FamilyConfig {
             name: "test",
             kind: FamilyKind::MultiValue,
-            compression: Compression::lz4(),
+            compression: Compression::Lz4,
         }],
     };
     let db =
@@ -697,7 +697,7 @@ fn open_multi_value_db(path: &Path) -> TurboPersistence<SerialScheduler, 1> {
         family_configs: [FamilyConfig {
             name: "test",
             kind: FamilyKind::MultiValue,
-            compression: Compression::lz4(),
+            compression: Compression::Lz4,
         }],
     };
     TurboPersistence::<SerialScheduler, 1>::open_with_config(path.to_path_buf(), db_config).unwrap()
@@ -966,7 +966,7 @@ fn bench_write_multi_value(c: &mut Criterion) {
                             family_configs: [FamilyConfig {
                                 name: "test",
                                 kind: FamilyKind::MultiValue,
-                                compression: Compression::lz4(),
+                                compression: Compression::Lz4,
                             }],
                         };
                         let db = TurboPersistence::<SerialScheduler, 1>::open_with_config(
--- a/turbopack/crates/turbo-persistence/benches/mod.rs
+++ b/turbopack/crates/turbo-persistence/benches/mod.rs
@@ -10,9 +10,9 @@ use quick_cache::sync::GuardResult;
 use rand::{RngExt, SeedableRng, rngs::SmallRng, seq::SliceRandom};
 use tempfile::TempDir;
 use turbo_persistence::{
-    ArcBytes, BlockCache, CompactConfig, DbConfig as TpDbConfig, Entry, EntryValue, FamilyConfig,
-    FamilyKind, MetaEntryFlags, SerialScheduler, StaticSortedFile, StaticSortedFileMetaData,
-    TurboPersistence, hash_key, write_static_stored_file,
+    ArcBytes, BlockCache, CompactConfig, Compression, DbConfig as TpDbConfig, Entry, EntryValue,
+    FamilyConfig, FamilyKind, MetaEntryFlags, SerialScheduler, StaticSortedFile,
+    StaticSortedFileMetaData, TurboPersistence, hash_key, write_static_stored_file,
 };
 use turbo_tasks_malloc::TurboMalloc;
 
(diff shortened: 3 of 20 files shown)


turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -194,6 +194,18 @@ fn family_name(family: u32) -> &'static str {
...
+/// Keep this mapping in sync when inspecting a Turbopack cache.
+fn family_compression(family: u32) -> Result<Compression> {
+    match family {
+        0 => Ok(Compression::Lz4Hc(4)), // Infra
+        1 => Ok(Compression::Lz4Hc(4)), // TaskMeta
+        2 => Ok(Compression::Zstd(3)),  // TaskData
+        3 => Ok(Compression::Lz4Hc(4)), // TaskCache
+        _ => bail!("No compression configuration for family {family}"),
+    }
+}
```
lukesandberg: this is reason enough to add a marker to the meta file for compression configuration

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,95 @@
...
+    lz4::{self, decompress},
+    lz4_hc,
+};
+
+thread_local! {
+    /// Zstd decompression contexts are reusable and relatively expensive to create. Reads can run
+    /// concurrently, so keep one context per thread rather than constructing one per block.
+    static ZSTD_DECOMPRESSOR: RefCell<zstd::bulk::Decompressor<'static>> = RefCell::new(
+        zstd::bulk::Decompressor::new().expect("zstd decompressor initialization should succeed")
+    );
```
lukesandberg: should we maintain a pool in the db so that they can be dropped?

are these large?

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,95 @@
...
+/// The algorithm is configuration, not part of the on-disk data. A database must be reopened with
+/// the same per-family compression configuration that was used to write it.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum Compression {
+    /// Fast LZ4 compression using the default acceleration level.
+    #[default]
+    Lz4,
+    /// LZ4 high-compression mode at the given level (3 through 12).
+    Lz4Hc(i32),
+    /// Zstandard compression at the given level.
```
lukesandberg: what is the range of valid levels?

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,95 @@
...
-        "Decompressed length does not match expected length"
+    };
+    let bytes_written = match compression {
+        Compression::Lz4 | Compression::Lz4Hc(_) => {
+            decompress(block, dest).with_context(context)?
+        }
+        Compression::Zstd(_) => ZSTD_DECOMPRESSOR.with_borrow_mut(|decompressor| {
+            decompressor
+                .decompress_to_buffer(block, dest)
+                .with_context(context)
```
lukesandberg: move the with_context calls to be around the match so the closure can be inlined again

dianatofficial (commented):

turbopack/crates/turbo-persistence/benches/mod.rs
```diff
@@ -342,6 +356,65 @@ fn bench_write(c: &mut Criterion) {
     group.finish();
 }
 
+/// Compares the write cost of the supported family compression configurations on identical data.
+fn bench_compression_write(c: &mut Criterion) {
```
lukesandberg: drop this benchmark

turbopack/crates/turbo-persistence/benches/mod.rs
```diff
@@ -485,6 +562,62 @@ fn bench_read_get(c: &mut Criterion) {
     group.finish();
 }
 
+/// Compares cached and uncached reads across family compression configurations.
+fn bench_compression_read(c: &mut Criterion) {
```
lukesandberg: drop this bench

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,203 @@
...
+                    "Invalid zstd compression level {level}; expected {} through {}",
+                    range.start(),
+                    range.end()
+                );
+                Ok(())
+            }
+        }
+    }
+
+    pub(crate) fn to_meta_fields(self) -> Result<(u32, i32)> {
```
lukesandberg: can we just encode to and from bytes?

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,203 @@
...
+        Ok(compression)
+    }
+}
+
+/// Reusable zstd decompression contexts owned by a database.
+///
+/// A context is about 96 KiB with the linked zstd version. The pool grows only to peak concurrent
+/// zstd reads, and all retained contexts are released when the database is dropped.
+#[derive(Default)]
+pub(crate) struct DecompressionPool {
```
lukesandberg: should we use quickcache?  

this is probably fine, certainly simple but lacking a stricter upper bound or expiry is risky, momentary read concurrency can be quite high

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,203 @@
...
+            .as_ref()
+            .expect("decompressor checked out")
+    }
+}
+
+impl DerefMut for PooledZstdDecompressor<'_> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.decompressor
+            .as_mut()
+            .expect("decompressor checked out")
```
lukesandberg: message is odd, shoudln't this be 'used after drop'?

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,203 @@
...
+    dest: &mut [u8],
+    expected_len: u32,
+) -> Result<()> {
     debug_assert!(
         expected_len > 0,
         "decompress_block called with uncompressed_length=0; uncompressed blocks should use \
          zero-copy mmap path"
     );
-    let bytes_written = decompress(block, dest).with_context(|| {
+    compression.validate()?;
```
lukesandberg: this seems like the wrong place to validate... 

Compression should really be validated on construction

turbopack/crates/turbo-persistence/src/compression.rs
bgw: Maybe this should just use `thread_local`? That'll bound us by number of threads and avoid any sort of global lock.

sokra:
Does that have an impact on binary size of the released *.node file, fleet?

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,124 @@
...
+        Self::Lz4
+    }
+
+    pub const fn lz4_hc4() -> Self {
+        Self::Lz4Hc4
+    }
+
+    pub const fn zstd_3() -> Self {
+        Self::Zstd3
+    }
```
lukesandberg: im not sure these accessors really make sense anymore, inline them.

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,124 @@
...
+/// Reusable zstd decompression contexts owned by one database.
+///
+/// A context is about 96 KiB with the linked zstd version. Each participating thread lazily creates
+/// one lock-free context, and dropping this object releases all of them.
+#[derive(Default)]
+pub(crate) struct DecompressionContext {
+    zstd: ThreadLocal<RefCell<zstd::bulk::Decompressor<'static>>>,
+    #[cfg(test)]
+    zstd_contexts_created: std::sync::atomic::AtomicUsize,
+}
```
lukesandberg: instead of the `thread_local` crate we should just use a std `thread_local!` they are a bit cheaper.

the counter can be a static for tests

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -57,8 +178,141 @@ pub fn checksum_block(data: &[u8]) -> u32 {
...
+            Compression::lz4(),
+            Compression::lz4_hc4(),
+            Compression::zstd_3(),
+        ] {
+            let encoded = compression.encode().unwrap();
+            assert_eq!(Compression::decode(&encoded).unwrap(), compression);
+        }
+        assert!(Compression::decode(&[99]).is_err());
+        assert!(Compression::decode(&[0, 0]).is_err());
+    }
```
lukesandberg: drop this test

turbopack/crates/turbo-persistence/src/meta_file.rs
```diff
@@ -50,7 +51,7 @@ impl Display for MetaEntryFlags {
 }
 
 /// Magic number identifying a `.meta` file.
-pub(crate) const META_FILE_MAGIC: u32 = 0xFE4ADA4A;
+pub(crate) const META_FILE_MAGIC: u32 = 0xFE4ADA4B;
```
lukesandberg: we don't need to bump this

turbopack/crates/turbo-persistence/src/meta_file_builder.rs
```diff
@@ -139,3 +145,53 @@ impl<W: Write> Write for CountingWriter<W> {
...
+        for compression in [
+            Compression::lz4(),
+            Compression::lz4_hc4(),
+            Compression::zstd_3(),
+        ] {
+            let (tempdir, _) = write_empty_meta(compression);
+            let meta = MetaFile::open(tempdir.path(), 1).unwrap();
+            assert_eq!(meta.compression(), compression);
+        }
+    }
```
lukesandberg: drop test it is redundant with tests on compression itself

turbopack/crates/turbo-persistence/README.md
```diff
@@ -54,8 +54,10 @@ Small value blocks are emitted once they accumulate at least `MIN_SMALL_VALUE_BL
 A meta file can contain metadata about multiple SST files. The metadata is stored in a single file to avoid having too many small files.
 
 - Header
-  - 4 bytes magic number (0xFE4ADA4A)
+  - 4 bytes magic number (0xFE4ADA4B)
   - 4 bytes key family
+  - 4 bytes compression preset byte length
+  - `n` bytes bincode-encoded compression preset (standard config; variants in format order: LZ4, LZ4 HC4, zstd3)
```
lukesandberg: don't bother recording the actual compression settings here

turbopack/crates/turbo-persistence/README.md
```diff
@@ -88,9 +90,11 @@ The SST file contains only data without any header.
 
 #### Block Compression
 
-Blocks can be stored compressed (LZ4) or uncompressed. The 4-byte header distinguishes them:
+Blocks can be stored compressed or uncompressed. The compression algorithm is recorded once in the containing SST's meta file and must match the key-family configuration used to open the database. It is not repeated in every block. Changing a family's configured algorithm requires rewriting that database, consistent with the crate's no-cross-version-compatibility policy.
```
lukesandberg: simplify this

just say that the compression algorithm is specified in the meta file

turbopack/crates/turbo-persistence/README.md
lukesandberg: mention here that it must be consistent with the configuration used to open the db

turbopack/crates/turbo-persistence/README.md
```diff
@@ -223,9 +227,9 @@ The plain value compressed with dynamic compression. Each blob file has an 8-byt
 
 - 4 bytes: uncompressed length (u32 big-endian)
 - 4 bytes: CRC32 checksum of the compressed data (u32 big-endian)
-- remaining bytes: LZ4-compressed value data
+- remaining bytes: value data compressed with the blob's key-family configuration
 
-The checksum is verified on the compressed data **before** decompression when the blob is read.
+The codec is validated by the family's meta-file marker rather than repeated in each blob. The checksum is verified on the compressed data **before** decompression when the blob is read.
```
lukesandberg: revert changes to this paragraph

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -796,10 +838,19 @@ impl Iterator for StaticSortedFileIter {
...
 
 impl StaticSortedFileIter {
-    /// Opens an SST file for sequential iteration. Uses `MADV_SEQUENTIAL` for
-    /// read-ahead and wraps the mmap in `Rc<Mmap>` directly (no `Arc`),
-    /// eliminating all atomic refcounting during iteration.
+    /// Opens an LZ4-compressed SST file for sequential iteration. Uses `MADV_SEQUENTIAL` for
+    /// read-ahead and wraps the mmap in `Rc<Mmap>` directly (no `Arc`), eliminating all atomic
+    /// refcounting during iteration. Database compaction uses the configuration-aware internal
+    /// constructor instead.
     pub fn open(db_path: &Path, meta: StaticSortedFileMetaData) -> Result<Self> {
```
lukesandberg: we can just add Compression to these APIs, we don't need the backcompat

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,81 @@
...
+            bincode::decode_from_slice(bytes, bincode::config::standard())
+                .context("Failed to decode compression preset")?;
+        ensure!(
+            consumed == bytes.len(),
+            "Compression preset has {} trailing bytes",
+            bytes.len() - consumed
+        );
+        Ok(compression)
+    }
+}
```
bgw: We don't need bincode?

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,81 @@
...
+    #[default]
+    Lz4,
+    /// Zstandard compression at level 3.
+    Zstd3,
+}
+
+impl Compression {
+    pub(crate) fn encode(self) -> Result<Vec<u8>> {
+        bincode::encode_to_vec(self, bincode::config::standard())
+            .context("Failed to encode compression preset")
```
bgw: Get rid of bincode so we can get rid of the error

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,81 @@
...
-    let bytes_written = decompress(block, dest).with_context(|| {
+    let result: Result<usize> = match compression {
+        Compression::Lz4 => decompress(block, dest).map_err(anyhow::Error::from),
+        Compression::Zstd3 => ZSTD_DECOMPRESSOR.with_borrow_mut(|decompressor| {
+            decompressor
+                .decompress_to_buffer(block, dest)
+                .map_err(anyhow::Error::from)
+        }),
+    };
+    let bytes_written = result.with_context(|| {
```
bgw:
```suggestion
    let bytes_written = match compression {
        Compression::Lz4 => decompress(block, dest).map_err(anyhow::Error::from),
        Compression::Zstd3 => ZSTD_DECOMPRESSOR.with_borrow_mut(|decompressor| {
            decompressor
                .decompress_to_buffer(block, dest)
                .map_err(anyhow::Error::from)
        }),
    }.with_context(|| {
```

turbopack/crates/turbo-persistence/src/meta_file.rs
```diff
@@ -264,16 +270,36 @@ pub struct MetaFile {
 }
 
 impl MetaFile {
-    /// Opens a meta file at the given path. Memory maps the entire file and eagerly deserializes
-    /// all AMQF filters as zero-copy [`qfilter::FilterRef`]s that borrow from the mmap.
+    /// Opens a meta file and uses its recorded compression configuration for referenced SSTs.
+    /// Memory maps the entire file and eagerly deserializes all AMQF filters as zero-copy
+    /// [`qfilter::FilterRef`]s that borrow from the mmap. Database opens use the internal
+    /// constructor to also validate the marker against the runtime family configuration.
```
bgw: Revert comment?

turbopack/crates/turbo-persistence/src/meta_file_builder.rs
```diff
@@ -139,3 +145,40 @@ impl<W: Write> Write for CountingWriter<W> {
...
+
+        let mut bytes = original.clone();
+        bytes[12] = 99;
+        std::fs::write(&path, &bytes).unwrap();
+        assert!(MetaFile::open(tempdir.path(), 1).is_err());
+
+        std::fs::write(&path, &original[..12]).unwrap();
+        assert!(MetaFile::open(tempdir.path(), 1).is_err());
+    }
+}
```
bgw: I don't think these tests are worth it

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -175,25 +176,37 @@ impl ValueBlockCache<ArcBytes> for ArcBlockCacheReader<'_> {
...
+            self.compression,
+        )
     }
 }
 
 /// Iteration-path: lightweight single-entry cache for sequential reads.
-impl ValueBlockCache<RcBytes> for &mut Option<(u16, RcBytes)> {
+struct RcBlockCacheReader<'a> {
+    cache: &'a mut Option<(u16, RcBytes)>,
+    compression: Compression,
```
bgw: Maybe do this as an argument to `get_or_read` instead

turbopack/crates/turbo-persistence/src/static_sorted_file.rs
```diff
@@ -175,25 +176,37 @@ impl ValueBlockCache<ArcBytes> for ArcBlockCacheReader<'_> {
...
+            self.cache,
+            self.verified_blocks,
+            self.compression,
+        )
     }
 }
 
 /// Iteration-path: lightweight single-entry cache for sequential reads.
-impl ValueBlockCache<RcBytes> for &mut Option<(u16, RcBytes)> {
+struct RcBlockCacheReader<'a> {
```
bgw, line 192: The named type is nice here

turbopack/crates/turbo-persistence/src/tests.rs
```diff
@@ -1109,16 +1112,24 @@ fn batch_get_across_families() -> Result<()> {
     let tempdir = tempfile::tempdir()?;
     let path = tempdir.path();
 
-    let db = TurboPersistence::<_, 16>::open_with_parallel_scheduler(
+    let mut config = DbConfig::default();
+    config.family_configs[0].compression = Compression::Lz4;
+    config.family_configs[1].compression = Compression::Lz4;
+    config.family_configs[2].compression = Compression::Zstd3;
+    config.family_configs[3].compression = Compression::Lz4;
```
bgw:
```suggestion
    // set zstd on an arbitrary family, lz4 is used by default
    config.family_configs[2].compression = Compression::Zstd3;
```

turbopack/crates/turbo-persistence/Cargo.toml
```diff
@@ -15,6 +15,7 @@ verbose_log = []
 [dependencies]
 anyhow = { workspace = true }
 auto-hash-map = { workspace = true }
+bincode = { workspace = true }
```
bgw: Maybe don't need this if we drop bincode from the Compression enum

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
```diff
@@ -299,19 +300,30 @@ pub struct StaticSortedFileBuilderMeta<'a> {
...
 // to free entry memory as blocks are written.
 pub fn write_static_stored_file<E: Entry>(
     entries: &[E],
     file: &Path,
     flags: MetaEntryFlags,
+) -> Result<(StaticSortedFileBuilderMeta<'static>, File)> {
+    write_static_stored_file_with_compression(entries, file, flags, Compression::Lz4)
+}
+
+pub(crate) fn write_static_stored_file_with_compression<E: Entry>(
```
lukesandberg: just make compression a required parameter of the `write_static_stored_file` test callers can just specify Lz4 themselves

turbopack/crates/turbo-persistence/src/static_sorted_file_builder.rs
```diff
@@ -575,11 +589,22 @@ pub struct StreamingSstWriter<E: Entry> {
...
+    /// Creates a new streaming SST writer using LZ4 compression.
     ///
-    /// `max_entry_count` is used to pre-allocate buffers and estimate block counts.
+    /// `max_entry_count` is used to pre-allocate buffers and estimate block counts. Databases with
+    /// per-family compression use the configuration-aware internal constructor instead.
     pub fn new(file: &Path, flags: MetaEntryFlags, max_entry_count: u64) -> Result<Self> {
+        Self::new_with_compression(file, flags, max_entry_count, Compression::Lz4)
+    }
+
+    pub(crate) fn new_with_compression(
```
lukesandberg: ditto

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -1,25 +1,62 @@
...
+use std::{cell::RefCell, mem::MaybeUninit, rc::Rc, sync::Arc};
 
-use anyhow::{Context, Result};
+use anyhow::{Context, Result, ensure};
 use lzzzz::lz4::{self, decompress};
 
+/// Compression algorithm used for a family's SST blocks and blob values.
+///
+/// The discriminants are stored in meta files, so existing values must not be changed. New
+/// algorithms get a new discriminant.
```
lukesandberg: the db is versioned with the code, so this is not accurate

bgw (approved):

PR #97717, opened 2026-08-21
Turbopack: Improve file existence error handling in realpath_with_links and in module resolution

bgw:
Fixes some regressions from https://github.com/vercel/next.js/pull/97395.

Human summary of changes:

- Plainly return `NotFound` for `realpath_with_link` whenever any step in the resolution algorithm fails, matching the behavior the libc uses.
- When probing in the module resolution algorithm, explicitly check for `NotFound`.
- Fully remove the extremely non-standard and confusing behavior of trying to return the last processed path from `realpath_with_links` if a failure occurs.
- Modify `realpath` to return a `RealPathError` instead of just an `anyhow::Error`, so that callers that care about the actual error kind but not all of the intermediate links can still use it instead of `realpath_with_links`.
- Rename `RealPathResultError` to `RealPathError`, and make it actually implement `Display` so that it's easier to coerce to an `anyhow::Error`. This requires including a tiny bit more data on this object, but in the hot path it's just one extra clone of a `FileSystemPath`.

Human discussion: https://vercel.slack.com/archives/C09R44U5HQW/p1787322213695299

AI slop summary of why #97395 caused a regression:

> The regression came from newly added missing-target detection in b2e835ded6.
>
> Before that commit, realpath_with_links:
>
> 1. Resolved directory symlinks.
> 2. Continued resolving the remaining path beneath the resolved directory.
> 3. If that remaining path did not exist, returned the resolved path successfully.
> 4. Let the caller’s subsequent existence check report that the candidate was absent.
>
> The commit changed read_link so it no longer checked the symlink target’s type. To detect dangling symlinks, realpath_with_links added this condition:
>
> ```
> if (followed_link || !symlinks.is_empty())
>     && entry_type == FileSystemEntryType::NotFound
> {
>     // report a missing symlink target
> }
> ```
>
> The problem is that !symlinks.is_empty() does not mean the missing path is a symlink target. It only means that some earlier path component was a symlink.
>
> For the reported path:
>
> `node_modules/@workflow/core/dist/serialization/package.json`
>
> resolution worked like this:
>
> 1. Resolve node_modules/@workflow/core, a valid directory symlink.
> 2. Continue into dist/serialization, which exists.
> 3. Probe the optional dist/serialization/package.json, which does not exist.
> 4. Because the earlier @workflow/core component was a symlink, symlinks was nonempty.
> 5. The new condition incorrectly classified the missing package.json as a dangling symlink target.
> 6. Turbopack aborted instead of treating package.json as an absent resolution candidate and continuing to serialization.js.
>
> So the regression was not caused by resolving the directory symlink incorrectly. It was caused by using “we traversed any symlink” as evidence that a later missing descendant was the target of a dangling symlink.


test/production/app-dir/symbolic-links/symbolic-links.test.ts
```diff
@@ -16,4 +16,9 @@ describe('symbolic-file-links', () => {
     const html = await next.render('/')
     expect(html).toContain('hello world')
   })
+
+  it('should resolve modules through directory symlinks', async () => {
```
lukesandberg: confused about this test? isn't it the same as the previous one, how is the fixture different

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -445,14 +445,13 @@ impl FileSystemPath {
...
-        match &result.path_result {
-            Ok(path) => Ok(path.clone()),
-            Err(error) => bail!("{}", error.as_error_message(self, result).await?),
-        }
+    /// Returns the realpath to the file, resolving all symlinks.
+    ///
+    /// The outer [`anyhow::Error`] represents an internal error in turbo-tasks. Any other error is
+    /// represented using a structured `RealPathError`.
+    pub async fn realpath(&self) -> Result<Result<FileSystemPath, RealPathError>> {
+        // `FileSystemPath` is cheap to clone
```
lukesandberg: delort comment, not important

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -606,28 +621,30 @@ async fn get_type(path: FileSystemPath) -> Result<Vc<FileSystemEntryType>> {
...
                 path_result: Ok(current_path),
                 symlinks: symlinks.into_iter().collect(),
             }
             .cell());
         }
 
         if !visited.insert(current_path.path.clone()) {
-            error = RealPathResultError::CycleDetected;
+            error_type = RealPathErrorType::CycleDetected {
+                symlinks: Box::new([]),
```
lukesandberg: why isn't this symlinks.into_iter().collect()?

lukesandberg (approved):

dianatofficial (commented):

PR #97718, opened 2026-08-21
fix(telemetry): stop `detached-flush` evaluating `next.config`, and bound its lifetime

### LINKED ISSUE #97716: `detached-flush` never exits, leaking a reparented process per `next dev` shutdown (regression in 16.0.3)

https://github.com/gesposito/nextjs-detached-flush-orphan

### To Reproduce

Scaffolded with `npx create-next-app -e reproduction-template`, so it runs on canary. The only change from that template is a `setInterval` in `next.config.ts`, standing in for any config plugin that leaves a live handle behind during evaluation.

```bash
git clone https://github.com/gesposito/nextjs-detached-flush-orphan
cd nextjs-detached-flush-orphan
npm install
npm run repro
```

`npm run repro` starts `next dev`, issues one request, sends `SIGINT` as Ctrl+C would, then watches the `detached-flush` child the shutdown spawns:

```
detached-flush child was spawned:
  95001     1  94352 /path/to/node .../next/dist/telemetry/detached-flush.js dev ...

LEAKED: 1 process(es) still alive ~13s after the dev server exited.
  pid=95001 ppid=1 rss=92MB
```

It stays there indefinitely. Its parent is gone, so it is reparented to init and nothing supervises it any more — nothing will ever tell it to stop. It survives until it is killed by hand or the machine reboots.

These are not zombies waiting to be cleaned up. They are live processes: the one above holds 92 MB, and the ones on my machine ran at 370 MB and, in one case, 137% CPU.

**The request matters.** With no traffic no telemetry events are recorded, `flushDetached` returns early, and no child spawns at all — so a shorter reproduction appears to pass. Easy to trip over.

`npm run repro:isolate` pins the cause without a dev server: it invokes `detached-flush` directly, twice, changing **only** `next.config.js`.

| `next.config.js` | result | events file consumed |
|---|---|---|
| `module.exports = {}` | exits 0, ~0.1s | yes |
| `setInterval(() => {}, 1000); module.exports = {}` | **never exits** | yes |

**Expected:** the detached flush child submits its telemetry and exits, as it did before 16.0.3.

**Actual:** it submits its telemetry, unlinks the events file, and then never exits — one such process left behind per `next dev` shutdown, each holding several hundred MB, on any project whose dev-phase `next.config` leaves a live handle behind. 63 accumulated on my machine over ~27 hours: **18.9 GB RSS** and ~350% CPU in aggregate, each with an `esbuild --service` grandchild. Individually invisible, only obvious in aggregate.

Note the telemetry work itself completes: the events file is unlinked in both cases above. It is also not a stuck request — `lsof -a -p <pid> -i` on a real orphan returns empty, so it holds zero sockets.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.3.0: Wed Jan 28 20:54:46 PST 2026; root:xnu-12377.91.3~2/RELEASE_ARM64_T6000
  Available memory (MB): 65536
  Available CPU cores: 10
Binaries:
  Node: 25.8.0
  npm: 11.11.0
  Yarn: N/A
  pnpm: 10.32.1
Relevant Packages:
  next: 16.4.0-canary.0 // Latest available version is detected (16.4.0-canary.0).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Mechanism is platform-independent; only tested on macOS.

Runtime

next dev (local)

### Additional context

Re-filing #97715, which the triage bot closed automatically: I had put an explanatory sentence underneath the URL in the reproduction-link field, and the automated check parses that whole field as a single URL. Same report, that field corrected — apologies for the noise.

The area list has no Telemetry option; the affected code is `packages/next/src/telemetry/`, spawned by the dev server on shutdown, so I picked the nearest one. The detached flush is only spawned on dev-server shutdown.

#### Cause

Two ingredients in `packages/next/src/telemetry/detached-flush.ts` (53 lines):

1. **Line 26** — `await loadConfig(PHASE_DEVELOPMENT_SERVER, dir)`, used only to compute `distDir`. This evaluates the user's entire `next.config` *under the dev-server phase* inside a throwaway, detached, unsupervised process, so every dev-only side effect in the config chain fires there.
2. **Lines 49–52 (the regression)** — commit `ad93e452` (#85867, "Fix telemetry event loss on build failures and server shutdown", 2025-11-10) removed `process.exit(0)` after `unlinkSync` and replaced it with *"let Node.js exit naturally after all pending work completes"*. That never happens for a process holding a ref'd libuv handle. And because it is spawned `detached: true`, the parent's exit leaves it reparented to init, still running, with nothing left that would ever shut it down.

`loadConfig` has been in the file since 2023; the `process.exit(0)` was the only thing making it harmless.

Two consequences beyond the orphan itself: any process the config spawns during evaluation (a bundler service, for instance) is left running too, and a config whose promise never settles hangs the flush before it starts.

#### One caveat on the obvious fix

Putting the `process.exit(0)` back would reintroduce what #85867 removed it for: exiting there discards `NEXT_TELEMETRY_DEBUG` output still queued behind a full pipe buffer. Measured with 300 records and a stalled reader, only the first ~64 KB arrives — **117 of 300**. Flagging it only so this regression is not traded for that one.

#### Affected versions

| 16.0.2 | 16.0.3 | 16.1.x | 16.2.x | 16.3.1 | 16.4.0-canary.0 |
|---|---|---|---|---|---|
| safe | **leaks** | leaks | leaks | leaks | **leaks** |

Verified against published tarballs by checking for an exit *after* `unlinkSync` — the ENOENT branch (line 40) retains its own `process.exit(0)`, so a naive grep gives a false negative. The change was cherry-picked into the 16.0.x line, so it is not accurate to say it landed in 16.1.0.

The linked reproduction runs on `16.4.0-canary.0`, which `next info` reports as the latest available, so this is not something already fixed on canary. It also reproduces against a from-source build of canary at `8c7ed006b8`.

#### The telemetry opt-out does not help

`isDisabled` is consulted only in `notify()` and `submitRecord()`. `loadConfig` (line 26) runs 19 lines before `new Telemetry()` (line 45). The opt-out suppresses one HTTP POST; every leaking action happens upstream of it.

Nor does the spawn itself stop: `flushDetached` only checks whether any events were queued, and `record()` queues them regardless of `isDisabled`. On a real project opted out through the documented command (`next telemetry disable`, i.e. the conf store rather than the env var), a `next dev` shutdown still spawns the child. That is how the orphans on my machine accumulated in the first place.

#### Prior reports

- Discussion #92805 — "Accumulating detached-flush.js processes" (2026-04-14). The chosen answer recommends disabling telemetry and `pkill`; neither addresses the cause, and the opt-out demonstrably does not prevent it.
- PR #92927 — "skip flushDetached if telemetry is disabled" (open since 2026-04-17). Same symptom, same trigger (E2E cycling a dev server), same telemetry-disabled observation. It narrows the blast radius but leaves opted-in users leaking.

#### Not the same as #96317 / #96318

#96317 (`_controller` lost through the promise chain) is a genuine defect, but it cannot produce an orphan: `flushDetached` runs only in the **parent**; the child never calls it. On the shutdown path `record(event, deferred=true)` skips `submitRecord`, so no `AbortController` is ever constructed. And `allEvents.push(...)` is unconditional, so a working abort yields a byte-identical spawn. Filed separately to keep the two from being conflated.

#### No test coverage

`detached-flush` has **no test coverage anywhere in the repo** — the only other mention is a filename inside the `next-server-nft` file-trace snapshot, which never executes it. That is plausibly why the removal shipped and survived nine releases.

#### Workaround

`pkill -f 'telemetry/detached-flush.js'` to reclaim, or pin `next@16.0.2`. The telemetry opt-out is not a workaround.

gesposito:
Fixes #97716

### What

`detached-flush` is spawned `detached` with nothing supervising it, so it has to terminate under every outcome. If it does not, the parent's exit leaves it reparented to init, still running, with nothing left that would ever shut it down.

It discovered `distDir` by calling `loadConfig()`, which evaluates the user's `next.config` **in the development-server phase**. Config plugins legitimately start file watchers and bundler services during that evaluation, and those hold ref'd libuv handles, so the process never exited. One orphan leaked per `next dev` shutdown — plus any process the config itself spawned — on every release since **16.0.3**, regardless of the telemetry setting, because `loadConfig()` runs long before any enablement check.

I accumulated 63 of them in ~27 hours: 18.9 GB RSS and ~350% CPU, each holding an `esbuild --service` grandchild.

### The change

`flushDetached` already knows `distDir` — it writes the events file into it one line before spawning — so it now passes it as a **required** argument. `Telemetry#flushDetached` is the only caller, and it resolves this file with a package-relative `require.resolve('./detached-flush')`, so parent and child are always the same build and there is no older invocation to stay compatible with.

That makes the `loadConfig` import unnecessary, and it is gone. No config is evaluated in this process at all — which means no watcher, no bundler service, and no grandchild to leak. The orphan class is structurally unreachable rather than bounded after the fact. An invocation without `distDir` now exits non-zero instead of silently falling back to the path that leaked.

Three supporting details:

- **`loadEnvConfig()` is now called explicitly.** `.env` can set `NEXT_TELEMETRY_DISABLED`, and `Telemetry` reads it in its constructor (`storage.ts:64`, with a comment saying as much), so dropping that side effect silently would have re-enabled telemetry for opted-out users. See the note under *Behaviour change* below for the part this does **not** cover.
- **An unref'd watchdog bounds the one thing left that can hang this process: the telemetry request.** That request is genuinely unbounded — `submitRecord` always passes its own signal (`storage.ts:341`), so the `if (!signal)` fallback that would apply `AbortSignal.timeout(5000)` in `post-telemetry-payload.ts:19` is dead code on every production path. The watchdog is therefore the only bound on the POST. It is **not** armed under `NEXT_TELEMETRY_DEBUG`: that mode is spawned with `spawnSync` and `detached: false`, so it is supervised and cannot be orphaned, and it is the only mode that writes enough to stdio for an exit to truncate it.
- **The events file is unlinked before the request rather than after.** The events are already in memory and nothing re-reads the file, so this costs nothing — and it means a watchdog exit cannot strand a per-pid `_events_<pid>.json`, which nothing would ever collect.

Argument parsing is also positional now rather than `pop()`-based. The old form mis-assigned every argument when the optional events filename was absent; `eventsFile.includes('/')` was papering over that.

### Why not simply restore the `process.exit(0)` removed in ad93e452

That was my first attempt, and it is wrong. #85867's stated reason for removing the line was correct: exiting there discards whatever `NEXT_TELEMETRY_DEBUG` output is still queued behind a full pipe buffer.

| 300 records, stderr to a pipe whose reader stalls | records delivered |
|---|---|
| with `process.exit(0)` | **117 / 300** (the first ~64 KB) |
| without | 300 / 300 |

This is also why the watchdog is not armed in debug mode — an exit from the watchdog truncates exactly the same way. Restoring the line is incomplete besides: it sits *after* the unbounded `await loadConfig()`, so it never runs in the never-settles case, and it does nothing about grandchildren spawned during evaluation. Removing the evaluation removes all of it at once.

### Verification

Built from source at `canary` `8c7ed006b8`.

| case | before | after |
|---|---|---|
| config with a live handle (`setInterval`) | never exits | **exits, 100 ms** |
| async config that never settles | never exits | **never evaluated** |
| grandchild process spawned by config | leaked (survives) | **never spawned** |
| invocation without `distDir` | silently evaluated the config | **rejected, exit 1** |
| pending handle with nothing else to end it | — | **exits, 10.1 s (watchdog)** |
| `NEXT_TELEMETRY_DEBUG`, 300 events, stalled reader | 300/300 | **300/300** |
| `.env` `NEXT_TELEMETRY_DISABLED` | honoured | **honoured** |

Also verified end to end against a real-world config — a Next app whose `next.config.ts` composes three plugin wrappers, one of which runs a bundler build during evaluation. On stock 16.3.1 the child is still alive at 30 s (SIGKILL, exit 137); with this change on the production argv it exits at 0 s with the events consumed.

Then backported to that app's installed 16.3.1 build (both files) and run as its actual dev server rather than a harness: `next dev`, real traffic, `SIGINT`. The flush child spawns and is gone **0.27 s** later, leaving no orphan and no stale `_events_<pid>.json`. The same cycle before the patch left a ~370 MB process at `PPID 1`; four had accumulated in a day. That app is opted out of telemetry via `next telemetry disable`, which is also direct confirmation that the opt-out does not prevent the spawn.

Standalone reproduction of the bug this fixes: https://github.com/gesposito/nextjs-detached-flush-orphan — scaffolded from `create-next-app -e reproduction-template`, and it reproduces on `16.4.0-canary.0`, which `next info` reports as the latest available.

Reproducing by hand: the dev-phase `distDir` is `.next/dev`. Placing the events file in `.next` hits the ENOENT branch, which returns on its own, and the test passes for the wrong reason.

### Test

Adds `test/unit/telemetry-detached-flush.test.ts` — no fixture app, no dev server, no network. `detached-flush` had **no test coverage anywhere in the repo** before this; the only other reference is a filename in the `next-server-nft` file-trace snapshot, which never executes it. That is plausibly why a removed `process.exit(0)` shipped and survived nine releases.

Four cases, each checked against a mutation of the code it guards rather than only against the fix:

| mutation | caught by |
|---|---|
| re-add `process.exit(0)` after the flush | watchdog + truncation tests |
| arm the watchdog unconditionally | truncation test |
| remove the watchdog | watchdog test |
| make `distDir` optional again | contract test |
| throw a different error from the `distDir` guard | contract test |
| evaluate the config despite `distDir` | config test |

The suite that shipped in my first draft of this change passed all six — including the re-added `process.exit(0)`, the regression it most needed to catch. These four fail on each.

### Behaviour change worth flagging

`loadConfig()` had **two** env-mutating side effects, not one. Besides `loadEnvConfig`, it evaluated the config module itself, whose body can mutate `process.env` — directly, or via something like `require('dotenv').config({ path: '../../.env' })`. Replacing it with `loadEnvConfig` covers `<dir>/.env*` only, so a `NEXT_TELEMETRY_DISABLED` set **inside `next.config.js`** no longer reaches this process when it is spawned from the `next dev` CLI parent, which deliberately never loads the config (`next-dev.ts:210-211`).

The documented opt-outs (`next telemetry disable`, the environment variable, `.env`) are unaffected. Closing the config-file case properly means carrying the resolved flag over the existing `start-server.ts` → `next-dev.ts` IPC message that already passes `distDir`; happy to do that here or separately, whichever you prefer.

### Notes for review

- The diff adds one import (`loadEnvConfig` from `@next/env`, already a dependency of this package), removes two (`loadConfig`, `PHASE_DEVELOPMENT_SERVER`), and does not change the file-trace list that `test/production/next-server-nft` snapshots. The argument check's message uses the `Invariant:` prefix that the rest of the repo's internal-contract errors use. Rebased onto canary after #97687, so there is no `errors.json` entry to add.
- The `distDir` guard is unreachable from the only caller, so it is a guard against a future change to `flushDetached` rather than a user-facing diagnostic — and it is worth being precise about where its message would actually appear. The normal path spawns with no `stdio` option, so the child's stderr is a pipe with no reader attached to a parent that is already exiting; the non-zero exit is the observable part, not the text. Under `NEXT_TELEMETRY_DEBUG` (`stdio: 'inherit'`) it is visible. The test asserts on the message rather than only the exit code, so an unrelated crash cannot satisfy it.
- The `test/e2e/telemetry` suite does not exercise this path — all three files are build-oriented (`skipStart: true`, `isNextStart`-gated assertions) and none of them reference the detached flush.
- Two pre-existing defects noticed while working here, both out of scope and neither introduced by this change: `storage.ts:212` is a self-assignment (`;(prom as any)._controller = (prom as any)._controller`), so `flushDetached`'s `item._controller?.abort()` is always a no-op; and `storage.ts:12` imports `AbortController` from the `@edge-runtime` ponyfill, whose signal native `fetch` rejects with `TypeError: Expected signal ... to be an instance of AbortSignal` — so the telemetry POST currently fails before opening a socket on any Node with a modern undici. `post-telemetry-payload.test.ts` misses it because it mocks `global.fetch` and calls the function with no signal, testing the one branch production never takes. Happy to file both.

### Environment

Verified against a from-source build of `canary` `8c7ed006b8` on Node 25.8.0, macOS 26.3.1 arm64. The mechanism is platform-independent, but stdio flush behaviour is not — I have only tested macOS.

### DIFF
--- a/packages/next/src/telemetry/detached-flush.ts
+++ b/packages/next/src/telemetry/detached-flush.ts
@@ -2,52 +2,71 @@ import fs from 'fs'
 import path from 'path'
 import type { TelemetryEvent } from './storage'
 import { Telemetry } from './storage'
-import loadConfig from '../server/config'
+import { loadEnvConfig } from '@next/env'
 import { getProjectDir } from '../lib/get-project-dir'
-import { PHASE_DEVELOPMENT_SERVER } from '../shared/lib/constants'
+
+// This process is spawned `detached` with nothing supervising it, so it has to
+// terminate on its own under every outcome. Nothing here evaluates the user's
+// config, but the telemetry request is unbounded: `submitRecord` always passes
+// its own signal, so the 5s `AbortSignal.timeout` fallback in
+// `postNextTelemetryPayload` never applies. A request that never settles would
+// leave this process running forever, reparented to init. `unref()` stops the
+// timer from holding the process open; on the normal path the work finishes in
+// ~90ms and it never fires.
+//
+// Not armed under `NEXT_TELEMETRY_DEBUG`. That mode is spawned with `spawnSync`
+// and `detached: false`, so it is supervised and cannot be orphaned, and it is
+// the only mode that writes enough to stdio for an exit to truncate it.
+const EXIT_TIMEOUT_MS = 10_000
+if (!process.env.NEXT_TELEMETRY_DEBUG) {
+  setTimeout(() => process.exit(0), EXIT_TIMEOUT_MS).unref()
+}
 
 // this process should be started with following arg order
 // 1. mode e.g. dev, export, start
 // 2. project dir
-// 3. events filename (optional, defaults to _events.json)
+// 3. events filename
+// 4. dist dir (absolute)
 ;(async () => {
-  const args = [...process.argv]
-  const eventsFile = args.pop()
-  let dir = args.pop()
-  const mode = args.pop()
+  const [, , mode, dirArg, eventsFile, distDir] = process.argv
 
-  if (!dir || mode !== 'dev') {
+  if (!dirArg || mode !== 'dev' || !eventsFile || !distDir) {
     throw new Error(
-      `Invalid flags should be run as node detached-flush dev ./path-to/project [eventsFile]`
+      `Invariant: detached-flush must be invoked as: node detached-flush dev <projectDir> <eventsFile> <distDir>`
     )
   }
-  dir = getProjectDir(dir)
+  const dir = getProjectDir(dirArg)
 
-  const config = await loadConfig(PHASE_DEVELOPMENT_SERVER, dir)
-  const distDir = path.join(dir, config.distDir || '.next')
-  // Support both old format (no eventsFile arg) and new format (with eventsFile arg)
-  const eventsPath = path.join(
-    distDir,
-    eventsFile && !eventsFile.includes('/') ? eventsFile : '_events.json'
-  )
+  // `.env` may set `NEXT_TELEMETRY_DISABLED`, and `Telemetry` reads it in its
+  // constructor, so the environment has to be loaded before that runs. Nothing
+  // else in this process loads it.
+  loadEnvConfig(dir, true, { info: () => {}, error: () => {} })
+
+  const eventsPath = path.join(distDir, eventsFile)
 
   let events: TelemetryEvent[]
   try {
     events = JSON.parse(fs.readFileSync(eventsPath, 'utf8'))
   } catch (err: any) {
     if (err.code === 'ENOENT') {
-      // no events to process we can exit now
-      process.exit(0)
+      // no events to process, and nothing holds the event loop open
+      return
     }
     throw err
   }
+
+  // Claim the batch up front rather than after the flush. The events are in
+  // memory and nothing re-reads the file, so unlinking here costs nothing, and
+  // if the watchdog fires mid-request the per-pid `_events_<pid>.json` is not
+  // left behind for good with nothing that would ever collect it.
+  fs.unlinkSync(eventsPath)
 
   const telemetry = new Telemetry({ distDir })
   await telemetry.record(events)
   await telemetry.flush()
 
-  // finished flushing events clean-up
-  fs.unlinkSync(eventsPath)
-  // Don't call process.exit() here - let Node.js exit naturally after
-  // all pending work completes (e.g., setTimeout in debug telemetry)
+  // Deliberately no `process.exit()` here: nothing in this process holds the
+  // event loop open, so it exits on its own once the pending stdio writes
+  // drain. Exiting explicitly would truncate the `NEXT_TELEMETRY_DEBUG` output
+  // when it exceeds the pipe buffer.
 })()
--- a/packages/next/src/telemetry/storage.ts
+++ b/packages/next/src/telemetry/storage.ts
@@ -261,7 +261,17 @@ export class Telemetry {
 
     spawn(
       process.execPath,
-      [require.resolve('./detached-flush'), mode, dir, eventsFile],
+      // `this.distDir` is passed explicitly so the child does not have to
+      // evaluate the user's `next.config` to discover it. That evaluation runs
+      // dev-phase config plugins (watchers, bundler services) inside a
+      // detached, unsupervised process, which is what left orphans behind.
+      [
+        require.resolve('./detached-flush'),
+        mode,
+        dir,
+        eventsFile,
+        this.distDir,
+      ],
       {
         detached: !this.NEXT_TELEMETRY_DEBUG,
         windowsHide: true,
--- a/test/unit/telemetry-detached-flush.test.ts
+++ b/test/unit/telemetry-detached-flush.test.ts
@@ -1,0 +1,239 @@
+import { spawn } from 'child_process'
+import { existsSync, mkdirSync, writeFileSync } from 'fs'
+import path from 'path'
+import { useTempDir } from '../lib/use-temp-dir'
+
+/**
+ * `detached-flush` is spawned with `detached: true` and nothing supervises it,
+ * so it has to terminate under every outcome or it is reparented to init and
+ * leaks: one process per `next dev` shutdown, holding whatever its work left
+ * behind.
+ *
+ * It used to discover `distDir` by calling `loadConfig()`, which evaluates the
+ * user's `next.config` in the development-server phase. Config plugins
+ * legitimately start file watchers and bundler services there, and those keep
+ * the process alive forever. `distDir` is a required argument and no config
+ * is evaluated here at all, with an unref'd watchdog bounding the one thing
+ * left that can hang: an unbounded telemetry request.
+ */
+const FLUSH_SCRIPT = require.resolve('next/dist/telemetry/detached-flush')
+
+const EVENTS_FILE = '_events_test.json'
+
+const RUN_TIMEOUT_MS = 30_000
+
+/** Past the child's 10s watchdog, so a watchdog exit would be observable. */
+const SLOW_READER_MS = 12_000
+
+/** Marker written by the fixture config, so a test can assert it never ran. */
+const CONFIG_MARKER = 'config-was-evaluated'
+
+/**
+ * Fills a directory from `useTempDir` with the smallest project this script
+ * accepts, and returns the dist dir the events file was written to.
+ */
+function createProject(
+  dir: string,
+  nextConfig: string | ((dir: string) => string)
+): string {
+  writeFileSync(
+    path.join(dir, 'next.config.js'),
+    typeof nextConfig === 'function' ? nextConfig(dir) : nextConfig
+  )
+  writeFileSync(
+    path.join(dir, 'package.json'),
+    JSON.stringify({ name: 'detached-flush-fixture', version: '0.0.0' })
+  )
+
+  // The dev-phase distDir is `.next/dev`, not `.next`. Writing the events file
+  // to `.next` makes `detached-flush` take its ENOENT branch, which returns on
+  // its own, so the process would exit for the wrong reason and these tests
+  // would pass against the unfixed code.
+  const distDir = path.join(dir, '.next', 'dev')
+  mkdirSync(distDir, { recursive: true })
+  writeFileSync(
+    path.join(distDir, EVENTS_FILE),
+    JSON.stringify([{ eventName: 'testEvent', payload: {} }])
+  )
+  return distDir
+}
+
+/**
+ * `runNextCommand` is not reusable here: it resolves and spawns `dist/bin/next`
+ * specifically, and these cases invoke the flush script directly, one of them
+ * with a `-r` preload.
+ */
+function runDetachedFlush(
+  dir: string,
+  distDir?: string,
+  { preload }: { preload?: string } = {}
+): Promise<{ code: number | null; timedOut: boolean; stderr: string }> {
+  return new Promise((resolve, reject) => {
+    const args = preload ? ['-r', preload] : []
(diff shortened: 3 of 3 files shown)


marcoshernanz:
Independent verification on Linux with Node 24.14.1, 16 vCPUs / 32 GB, at 0f45a81f0b46ed090b7b07e7784944173ee6065d:

- Baseline isolate: 3/3 configs that create a refed handle failed to exit within 1.5 s, while all event files were consumed.
- This PR: 3/3 refed-handle cases exited; next.config was evaluated 0 times; all event files were consumed.
- Ten normal child runs all exited, with a 146 ms median and 160 ms p90.
- pnpm build-all passed: 18/18 tasks.
- The focused telemetry-detached-flush suite passed: 4/4, including the 10 s request watchdog and the backpressured 300-record debug stream.
- I packed the built Next package into the issue reproduction, drove GET / to 200, sent SIGINT, observed the detached-flush child, and confirmed no orphan survived the approximately 13 s observation window.

I also mutation-tested the tempting process.exit(0) alternative: it delivered only 107/300 debug records under pipe backpressure, so preserving natural exit for the debug path is important.

The explicit distDir plus loadEnvConfig split looks like the right compatibility boundary: it avoids config side effects while preserving documented .env telemetry opt-outs.

marcoshernanz:
One test-only race surfaced when I ran the full pnpm test-unit suite: the slow-reader test registers its child close listener only after the 12 s pause. Under the parallel suite the child can emit close during that pause, so the subsequent promise never resolves and Jest times out at 60 s.

I tested the following shape successfully: create the close promise immediately after spawn with child.once(close, resolve), leave stderr without a data listener during SLOW_READER_MS so the pipe remains paused, attach the data listener after the delay, and then await the already-created promise.

With that change, the full run completed in 23.6 s and the telemetry-detached-flush suite passed. The remaining failures were pre-existing Node-runtime failures plus SWC snapshot mismatches caused by the older binary available in my branch worktree; none touched this PR.

I did not push into the contributor branch, but this small test change should make the regression coverage reliable under full-suite concurrency.

gesposito:
Thanks for running this, the race is real and I've pushed the fix as a separate commit (b6e4be0f2c).

One correction on the mechanism, since it'll matter to whoever reads this later: it isn't parallel-suite timing. The child writes ~164 KB against a 64 KB pipe buffer, so with no reader attached it blocks on write and cannot exit during the 12 s pause, load changes timing, not byte count. The actual defect is that close fires once and the listener was attached after an await, so any early child exit (crash, OOM under a loaded suite, failed spawn) lands in a missed-event window. 
I reproduced it by making the child exit during the pause: the old shape hangs to the Jest timeout, the new one resolves and fails on the record count instead, which is the better part of the fix, since it turns a bare timeout into a real diagnostic.

dianatofficial (commented):

PR #97722, opened 2026-08-21
fix(app-router): segment prefetch config overrides app-level partialPrefetching default

### LINKED ISSUE #97386: Runtime prefetch runs for every dynamic render, and a prefetch export cannot opt a route out

https://github.com/damianfrizzi/next-runtime-prefetch-repro

### To Reproduce

```bash
git clone https://github.com/damianfrizzi/next-runtime-prefetch-repro
cd next-runtime-prefetch-repro
bun install
./run-all.sh
```

The route `app/item/[id]/page.tsx` is a cached shell. It uses `'use cache'` and `cacheLife('max')`.
The page gives the shell one request-time child as a prop. The child reads `searchParams`, so the
route builds as `◐ Partial Prerender`. The shell renders a Client Component with 200 rows, and each
row holds the token `REPRO_MARKER`. The count of that token gives the number of copies of the shell.

`measure.sh` sends a plain document request. The request carries no `RSC` header and no
`Next-Router-Prefetch` header. No `<Link>` is part of the flow.

`run-all.sh` prints:

| # | Configuration | Bytes | Copies in Flight |
| - | --- | ---: | ---: |
| 1 | `partialPrefetching: true` and `prefetch = 'force-disabled'` | 54,251 | 2 |
| 2 | `partialPrefetching: true`, no `prefetch` export | 54,237 | 2 |
| 3 | `partialPrefetching: false` | 30,466 | 1 |

Row 1 and row 2 differ by 14 bytes. Those bytes are the export line itself.

## 1. `prefetch = 'force-disabled'` does not override the application default

The `partialPrefetching` reference says:

> A segment that exports an explicit `prefetch` value overrides the app-level default for that
> route.

I expect row 1 to hold one copy. It holds two. The gate reads the global flag first and stops:

```js
// server/app-render/app-render.js:475 (Flight) and :2003 (HTML)
if (Boolean(renderOpts.partialPrefetching) || await anySegmentHasPartialPrefetchingEnabled(tree)) {
```

`anySegmentHasPartialPrefetchingEnabled` returns `true` for `'partial'` and `'unstable_eager'`. It
cannot express an opt-out. `spawnRuntimePrefetchWithFilledCaches` then prerenders the whole route,
and it does not read the segment configuration.

A route can therefore opt in per segment. A route cannot opt out. The application flag is the only
control, and the adoption guide describes that flag as the end state:

> Once every route in scope has `prefetch = 'partial'`, enable the global flag and remove the
> per-route exports.

## 2. The runtime prefetch runs for each dynamic render, not per link

The runtime prefetching guide says:

> Runtime prefetching is opted into per link with `<Link prefetch={true}>`.

and:

> Generating it costs a server invocation per prefetchable link, so it is opt-in per link.

I expect a plain document request to carry one copy of the shell. It carries two. The source states
the purpose of the second copy: "so the client can cache runtime-prefetchable content during
hydration". A client that does not hydrate pays the same cost.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 14
Binaries:
  Node: 26.5.0
  npm: 11.17.0
Relevant Packages:
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

Partial Prerendering (PPR), cacheComponents, Runtime

next build (local), Vercel (Deployed), next start (local)

### Additional context

A production application shows the cost. The route has the same shape: a cached shell and one
streamed child.

| | `partialPrefetching: true` | `partialPrefetching: false` |
| --- | ---: | ---: |
| document | 452,894 B | 299,762 B |
| wire, brotli | 45.7 KB | 38.7 KB |

The two copies are equal, so brotli compresses them well and the wire cost stays small. The client
parses both copies. On that route, 78 % of the requests carry a declared crawler user agent, and a
crawler never hydrates.

A route that reads `searchParams` therefore pays for a prefetch that most of its traffic discards,
and it has no way to decline. A fix for behaviour 1 also solves behaviour 2 for the routes that need
it: let `prefetch = 'force-disabled'` and `prefetch = 'auto'` gate the spawn, so a route can opt out
while the application flag stays on.

mittalpk:
Fixes #97386

### What was wrong

`anySegmentHasPartialPrefetchingEnabled(tree)` was checking the app-level `partialPrefetching` setting before checking each segment. Because of that, if the app-level setting was enabled, it would immediately return `true` even when a specific page explicitly disabled prefetching.

### What this caused

A page with:

```ts
export const prefetch = 'force-disabled'
```

could still start and ship a runtime prefetch stream when `partialPrefetching` was enabled at the app level. This resulted in an unnecessary duplicate prefetch stream.

### What changed

The function now receives the inherited prefetch configuration as a second argument:

```ts
inheritedPrefetchConfig: Prefetch | undefined
```

For each segment, we resolve the configuration like this:

```ts
resolvedPrefetchConfig = explicitPrefetchConfig ?? inheritedPrefetchConfig
```

That resolved value is then passed down to the segment's children as their inherited default.

Importantly, we only check whether the resolved value is `'partial'` at the **leaf/page segment**, rather than at every level of the tree.

### Why the leaf-only check is important

Previously, the function walked from the root toward the page and returned `true` as soon as it found a segment resolving to `'partial'`.

Root and layout segments are usually not explicitly configured, so they would inherit the app-level `partialPrefetching` setting. That meant they could resolve to `'partial'` and cause the function to return before it ever reached the page's explicit `force-disabled` override.

We verified this with a live reproduction using the reporter's `damianfrizzi/next-runtime-prefetch-repro` repository. The bug still reproduced even after introducing the `??` configuration resolution, which confirmed that checking for `'partial'` at every ancestor was the underlying problem.

### Other call site

The third call site was fixed as well. `getPrefetchingModeForPage`, which is used for the dev-only `SyncIOMode` selection, had the same issue. It wasn't part of the original report, but leaving it inconsistent would have preserved the same bug there.

### Tests

Added `test/unit/any-segment-has-partial-prefetching-enabled.test.ts` with 8 test cases, including a regression test covering the exact reported scenario: a nested layout with an app-level default and a page-level `force-disabled` override.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -1107,9 +1107,16 @@ async function generateStagedDynamicFlightRenderResultNode(
   // fill caches and then spawn a final runtime prerender whose result stream
   // is embedded in the RSC payload. This is gated because it adds extra server
   // processing and increases the response payload size.
+  //
+  // The app-level default is folded into the per-segment resolution inside
+  // anySegmentHasPartialPrefetchingEnabled (rather than OR'd in here), so a
+  // segment's own explicit `prefetch` export can override it either way, not
+  // just turn it on.
   if (
-    Boolean(renderOpts.partialPrefetching) ||
-    (await anySegmentHasPartialPrefetchingEnabled(loaderTree))
+    await anySegmentHasPartialPrefetchingEnabled(
+      loaderTree,
+      renderOpts.partialPrefetching ? 'partial' : undefined
+    )
   ) {
     // Create a mutable cache that gets filled during the dynamic render.
     const prerenderResumeDataCache = createPrerenderResumeDataCache()
@@ -3878,10 +3885,15 @@ async function renderToStream(
         // runtime-prefetchable content during hydration. This is enabled when
         // Partial Prefetching is on for the route, either per segment (a
         // `prefetch` of 'partial') or globally (the
-        // `partialPrefetching` config).
+        // `partialPrefetching` config). The app-level default is folded into
+        // the per-segment resolution inside
+        // anySegmentHasPartialPrefetchingEnabled, so a segment's own explicit
+        // `prefetch` export can override it either way, not just turn it on.
         if (
-          Boolean(renderOpts.partialPrefetching) ||
-          (await anySegmentHasPartialPrefetchingEnabled(tree))
+          await anySegmentHasPartialPrefetchingEnabled(
+            tree,
+            renderOpts.partialPrefetching ? 'partial' : undefined
+          )
         ) {
           const prerenderResumeDataCache = createPrerenderResumeDataCache()
           requestStore.resumeDataCache = prerenderResumeDataCache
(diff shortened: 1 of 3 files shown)


dianatofficial (commented):

PR #97724, opened 2026-08-22
test: preserve server cache after compile error

wbinnssmith:
### What?

Adds a focused Turbopack development regression test covering server HMR recovery from a compile error.

### Why?

A server compile error must not force a full server module reevaluation when HMR recovers. Clearing the server module cache would discard stable state held by unchanged dependencies and undermine server Fast Refresh module preservation.

### How?

The test reuses the existing module-preservation fixture and its unchanged server dependency evaluation timestamp. It introduces a deterministic syntax error, verifies the Turbopack build-error redbox, repairs the page, waits for the repaired server render, and then confirms the dependency still exposes its original timestamp. The patched fixture is restored in a `finally` block so failures cannot leak state into later tests.

### Testing

- `pnpm test-dev-turbo test/development/app-dir/server-hmr/server-hmr.test.ts`
- 17 tests passed

### DIFF
--- a/test/development/app-dir/server-hmr/server-hmr.test.ts
+++ b/test/development/app-dir/server-hmr/server-hmr.test.ts
@@ -48,6 +48,61 @@ describe('server-hmr', () => {
     )
 
     itTurbopackDev(
+      'does not re-evaluate an unmodified module after recovering from a compile error',
+      async () => {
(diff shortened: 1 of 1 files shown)


dianatofficial (commented):

sokra (approved):

Issue #97564, opened 2026-08-19
building route handler adds an extra unneccesary rsc file


davecarlson:
### Link to the code that reproduces this issue

https://github.com/davecarlson/nextjs-route-handler-rsc-routes-repro

## To Reproduce

```
1. Clone https://github.com/davecarlson/nextjs-route-handler-rsc-routes-repro
2. npm install
3. npm run build
4. npm run verify:files
```

### Current vs. Expected behavior

I see an extra file being created, meaning on large projects you can hit the Routes created per Deployment in 2x time.

I expect the number of handlers outputs to match the number of route handlers, and the extra .rsc .rsc.segment to not be created

For example, this gets created: 
    /api/[tenantId].rsc
    /api/[tenantId]
    /api/[tenantId]/items.rsc
    /api/[tenantId]/items
    /api/[tenantId]/items/[itemId].rsc
    /api/[tenantId]/items/[itemId]

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.6.0: Fri Jul 31 19:19:08 PDT 2026; root:xnu-12377.161.14~5/RELEASE_ARM64_T6050
  Available memory (MB): 65536
  Available CPU cores: 18
Binaries:
  Node: 24.18.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.3.1 // Latest available version is detected (16.3.1).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: N/A
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Route Handlers, Dynamic Routes, Output

### Which stage(s) are affected? (Select all that apply)

Build (Local), Vercel (Deployed)

### Additional context

Checked with latest canary build - same issue exists.

I have a vercel build that fails due to hitting the limit : `AhJzCWvPRMin8Lez85Q2CQQcqYn4`

davecarlson:
My claude found this while debugging: 

### Where it comes from

Both sites look to be in `packages/next/src/build/adapter/build-complete.ts`.

**1. Duplicate outputs.** The output's `type` is computed from the `/route` suffix, and then the
branch for exactly that case pushes an `.rsc` twin anyway - five lines apart:

```js
const output = {
    pathname: normalizedPage,
    sourcePage: page,
[...]

PR #97726, opened 2026-08-22
Stop emitting a separate route entry for a dynamic route's RSC form

unstubbable:
A dynamic app page in an app with an app directory needs two routes today. One matches a request for the page, and one matches a request for its `.rsc` payload or a per-segment prefetch. Simplified, with the request pattern on the left and the artifact it resolves to on the right:

```diff
- /de/<slug><.rsc|.segments/*.segment.rsc>    ->  /de/[slug]<matched suffix>
- /de/<slug>                                  ->  /de/[slug]
+ /de/<slug><.rsc|.segments/*.segment.rsc|>   ->  /de/[slug]<matched suffix>
```

One entry now serves both, because the suffix group gains an empty final alternative. The group therefore always matches. It captures an empty string for a request that carries no suffix, and the destination copies whatever it captured, so a plain request resolves to the page and a suffixed request resolves to the artifact it asks for. The parameter is matched lazily, so a request for `/de/other.rsc` still prefers the shorter parameter and the `.rsc` suffix over a parameter that swallows the suffix.

The empty alternative is deliberate, and an optional group would not be equivalent. A consumer of these entries can resolve the placeholders in a destination from the match result rather than from the pattern. A group that does not take part in the match is then absent from that result, and the literal text `$rscSuffix` survives into the destination. A group that always takes part avoids depending on how any one consumer treats an absent key, and the cost of getting it wrong is a 404 on every plain navigation to every dynamic route.

One case keeps two entries. A pages router route with `fallback: false` requires the preview cookies on the request for the page, and not on the request for its `.rsc` payload. An entry carries one set of conditions and cannot express that difference, so the merge applies only when both forms agree on their conditions.

Every app that has an app directory therefore halves the entries for its dynamic routes, whether or not it uses Cache Components. We measured the same in-progress feature branch of the v0 chat app as the previous change in this stack, which enumerates precomputed flag, locale and device permutations in `generateStaticParams` for a top-level dynamic segment and so multiplies every route below it. Every dynamic entry on that branch merges. Building it before and after the change removes 48% of the routes that the previous change left, replaces each merged pair with a single entry, and changes nothing else once the build ID is normalized. Across both changes that branch loses 65% of its routes.

This collapse follows `experimental.collapseAdapterRoutes`, which the previous change in this stack added. A build that sets it to `false` emits a separate entry for each form.

**Verified with a [full deploy test run](https://github.com/vercel/next.js/actions/runs/32540912007).**

### DIFF
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -2122,38 +2122,82 @@ export async function handleBuildComplete({
           route.page
         ) + getDestinationQuery(route.routeKeys)
 
-      // This route serves two kinds of request for the page: a request for the
-      // `.rsc` payload, and a per-segment prefetch request. The suffix group
-      // accepts both forms, and the destination copies the matched suffix, so
-      // each request resolves to the artifact that it asks for.
-      if (appPageKeys && appPageKeys.length > 0) {
+      const hasAppPages = Boolean(appPageKeys && appPageKeys.length > 0)
+
+      const suffixedHas =
+        isFallbackFalse && !pageKeys.includes(route.page)
+          ? fallbackFalseHasCondition
+          : undefined
+      const plainHas = isFallbackFalse ? fallbackFalseHasCondition : undefined
+
+      // A single entry can serve both forms of the request only when both carry
+      // the same conditions. A pages router route with `fallback: false` is the
+      // one case where they differ: it requires the preview cookies on the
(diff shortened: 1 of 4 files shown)


dianatofficial (commented):

mischnic (approved):

davecarlson:
@unstubbable will this fix this for me ?

https://github.com/vercel/next.js/issues/97564

PR #97727, opened 2026-08-22
Fix duplicate RSC adapter outputs for App Route Handlers

### LINKED ISSUE #97564: building route handler adds an extra unneccesary rsc file

marcoshernanz:
## What

- Emit one `APP_ROUTE` adapter output per App Route Handler instead of materializing a second `.rsc` function from the same entrypoint.
- Preserve RSC-header and segment-prefetch navigation by adding lightweight route aliases that point to the ordinary handler output.
- For dynamic handlers, strip the synthetic RSC suffix before forwarding parameters to the handler.
- Keep the existing App Page and hybrid Pages Router RSC behavior unchanged.

## Why

The Vercel adapter rewrites App Router requests carrying the `rsc: 1` header to an `.rsc` pathname before filesystem routing. Route Handlers therefore still need an RSC-shaped route, even though they return ordinary responses rather than an RSC artifact.

The legacy builder satisfied that invariant by creating a second function with an `.rsc` pathname. The Build Output API port retained the same behavior, but adapter consumers such as `@vercel/next` materialize both outputs, so one handler becomes two deployed functions.

This change preserves the routing invariant while routing both pathname forms to one function. Static aliases are ordered before catch-all App Page routes, and dynamic aliases keep route parameters separate from the synthetic `.rsc` or `.segments/...segment.rsc` suffix.

## Verification

- `pnpm build`
- `pnpm test-start test/production/adapter-config/adapter-config.test.ts`
- `pnpm test-start-turbo test/production/adapter-config/adapter-config.test.ts`
- focused Prettier, ESLint, and `git diff --check`
- real `@vercel/next` issue reproduction: 6 Route Handlers produce 6 functions, including dynamic Node, Edge, nested, and route-group handlers
- static navigation reproduction with a catch-all App Page: `/logout.rsc` and segment-prefetch paths resolve to `/logout`
- dynamic route-table probes: `/api/acme.rsc` and segment-prefetch paths capture `tenantId = acme`, not `acme.rsc`
- isolated root-handler regression: under `basePath: /docs`, `/index.rsc` aliases to the single `/docs` output for both Webpack and Turbopack
- shared export-mode adapter fixture under both Webpack and Turbopack

Fixes #97564

### DIFF
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -47,6 +47,7 @@ import {
 } from '../../lib/constants'
 
 import { normalizeLocalePath } from '../../shared/lib/i18n/normalize-locale-path'
+import { isAppRouteRoute } from '../../lib/is-app-route-route'
 import { getStaticMetadataPrerenderPathname } from '../../lib/metadata/get-metadata-route'
 import { isStaticMetadataFile } from '../../lib/metadata/is-metadata-route'
 import { addPathPrefix } from '../../shared/lib/router/utils/add-path-prefix'
@@ -659,6 +660,11 @@ export async function handleBuildComplete({
       prerenders: [],
       staticFiles: [],
     }
+    const appRoutePathnames = new Set(
+      (appPageKeys ?? [])
+        .filter(isAppRouteRoute)
+        .map((appPath) => normalizeAppPath(appPath))
+    )
 
     if (config.output === 'export') {
       // collect export assets and provide as static files
@@ -1212,11 +1218,6 @@ export async function handleBuildComplete({
             outputs.appPages.push(output)
           } else {
             outputs.appRoutes.push(output)
-            outputs.appRoutes.push({
-              ...output,
-              pathname: normalizePagePath(output.pathname) + '.rsc',
-              id: normalizePagePath(output.pathname) + '.rsc',
-            })
           }
         }
       }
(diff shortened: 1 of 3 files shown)


marcoshernanz:
I substantially revised this after validating the existing RSC rewrite invariant. The first version incorrectly removed Route Handler RSC aliases along with the duplicate function output, which could break client navigation to a handler or append `.rsc` to a dynamic parameter.

The new version still emits one function per handler, but preserves `.rsc` and segment-prefetch routing through lightweight aliases to the ordinary handler output. It now covers static and dynamic handlers, catch-all ordering, dynamic parameter capture, Edge and route-group handlers, root handlers with `basePath`, hybrid Pages Router behavior, and export-mode fixtures under both bundlers.

dianatofficial (commented):

PR #97729, opened 2026-08-22
Fix metadata prefetch cache key for search params

### LINKED ISSUE #97657: Metadata from a prefetched <Link> is applied to a different search string on the same route (production only)

https://github.com/lilyreadmoo/nextjs-repro-metadata-searchparams-prefetch

### To Reproduce

`app/search/page.jsx` — `generateMetadata` reads `searchParams`:

```jsx
import Link from 'next/link'

export async function generateMetadata({ searchParams }) {
  const { q } = await searchParams
  return { title: `Results for ${q}` }
}

export default async function SearchPage({ searchParams }) {
  const { q } = await searchParams
  return (
    <div>
      <h1>Results for {q}</h1>
      <nav>
        <Link href="/search?q=alpha">alpha</Link>{' '}
        <Link href="/search?q=beta">beta</Link>{' '}
        <Link href="/search?q=gamma">gamma</Link>
      </nav>
    </div>
  )
}
```

`app/page.jsx` — two links to the same route with different search params:

```jsx
import Link from 'next/link'

export default function Home() {
  return (
    <ul>
      <li><Link href="/search?q=alpha">alpha</Link></li>
      <li><Link href="/search?q=beta">beta</Link></li>
    </ul>
  )
}
```

Steps:

1. `next build && next start` (the bug does not happen in `next dev`, where `<Link>` prefetching is disabled).
2. Open `/`. Both links are in the viewport, so both are prefetched.
3. Click **alpha**.
4. Observe `document.title`.
5. From the search page, click **gamma** (a search string that was never visited).

After step 3:

| | value |
|---|---|
| `location.href` | `/search?q=alpha` ✅ |
| `<h1>` | `Results for alpha` ✅ |
| `document.title` | `Results for beta` ❌ |

Expected: `document.title` is `Results for alpha`.

After step 5, `document.title` is **still** `Results for beta`, while the URL and `<h1>` are `gamma`. Once the `<head>` for the route has been populated from a prefetch, no further searchParams-only navigation on that route updates it. It is not only `document.title`: in a real app we saw the `og:title` meta tag stay on the stale keyword too, so it looks like the whole resolved-metadata subtree is reused. A full document load (reload / new tab / direct URL) always renders the correct metadata, so this is client-side only.

The server is sending the right metadata. Fetching the navigation's RSC payload from the page after step 5:

```js
const r = await fetch('/search?q=gamma&_rsc=probe', { headers: { RSC: '1' } })
const t = await r.text()
t.includes('Results for gamma') // true
t.includes('Results for beta')  // false
```

So the payload for the current URL contains the correct title; the client router does not apply it to the head.

Which wrong title you get depends on which prefetch resolved — with `alpha` and `beta` both in the viewport we consistently get `beta` in step 3, i.e. the last prefetch wins for the whole session.

`<Link prefetch={false}>` on every link to the route makes the behavior correct in all of the steps above, which points at the head/metadata entry written by a prefetch response being keyed by pathname without the search string.

### Verify canary release

- [x] I verified that the issue exists in the latest Next.js canary release (`16.3.1-canary.26`). It also reproduces on `16.3.1` stable, and on `16.3.0` in a real application.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.3.0
  Available memory (MB): 16384
  Available CPU cores: 10
Binaries:
  Node: 22.22.2
  npm: 10.9.7
Relevant Packages:
  next: 16.3.1-canary.26
  react: 19.2.8
  react-dom: 19.2.8
Next.js Config:
  output: N/A
```

The minimal reproduction above uses the default bundler (Turbopack). The same behavior appears in a real application built with `next build --webpack`, with `export const dynamic = 'force-dynamic'` on the search page. `cacheComponents` / `partialPrefetching` are **not** enabled in either case.

Metadata, Navigation, Prefetching (Segment Cache)

next build (local), next start (local), Other (Deployed) — reproduced on AWS Amplify hosting as well

### Additional context

Real-world impact: a site search page reached from a header that always has a few "popular keyword" `<Link>`s in the viewport. Those links are prefetched on every page, so the browser tab title of the search results page is permanently stuck on one keyword for the whole session, no matter which keyword the user actually searches (including navigations made with `router.push` from the search box). SEO is unaffected because crawlers do full document loads.

marcoshernanz:
### What?

Treat a `LoadingBoundary` response as `Full` when writing the response's head into the segment cache, matching the strategy already used when the scheduler creates that head entry.

Adds a production e2e regression test with two concurrent prefetches for the same pathname and different search params.

### Why?

`pingRuntimeHead` creates loading-boundary head entries with the concrete `Full` strategy because a head has no loading boundary. The canonical re-keying added in #96095 later wrote the fulfilled head with raw `LoadingBoundary` semantics, which drops search params from the metadata cache key. Concurrent query-specific metadata responses could therefore replace each other under one fallback key; whichever response settled last supplied the title for both URLs.

Fixes #97657

### How?

Normalize `LoadingBoundary` to `Full` at the head write boundary only. Page segments retain loading-boundary keying, and shell/PPR responses keep their existing strategies.

The test reveals alpha and beta links in one React render, waits for both prefetches, navigates to alpha, and verifies both body and title. It then performs a search-only navigation to an unprefetched gamma URL and verifies that the head is replaced.

### Testing

- Regression boundary: test passes at `e70495c1ee` and fails at #96095 merge commit `dfa7f4f713` without the fix (`alpha` body, `beta` title).
- Test passes at `dfa7f4f713` with the fix.
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-turbo test/e2e/app-dir/segment-cache/metadata-search-params/metadata-search-params.test.ts` (pass, plus 3 consecutive repeat runs)
- `NEXT_SKIP_ISOLATE=1 pnpm test-start-webpack test/e2e/app-dir/segment-cache/metadata-search-params/metadata-search-params.test.ts` (pass)
- Relevant existing Turbopack suites: 16 tests pass across metadata, static shell, metadata soft navigation, and stale search-param coverage.
- Existing search-param suites: 6 tests pass with webpack. Their Turbopack builds are currently blocked on canary by an unrelated Edge Runtime `process.cwd()` error.
- `pnpm --filter next build`
- Targeted ESLint and Prettier checks
- Structured external review: 0 findings

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -3209,10 +3209,16 @@ function writeServerResponseIntoCache(
         ? now + getStaleTimeMs(navigationSeed.headStaleTimeSeconds)
         : staleAt
 
+    // A head has no loading boundary. Match pingRuntimeHead, which spawns
+    // LoadingBoundary head entries using the concrete Full strategy.
+    const headFetchStrategy =
+      fetchStrategy === FetchStrategy.LoadingBoundary
+        ? FetchStrategy.Full
+        : fetchStrategy
     const writtenHeadEntry = writeSegmentDataIntoCache(
       now,
       map,
-      fetchStrategy,
+      headFetchStrategy,
       head,
       // The decode already resolved the head's partiality from the wire
       // form and the response-level value — see the head read in
--- a/test/e2e/app-dir/segment-cache/metadata-search-params/app/layout.tsx
+++ b/test/e2e/app-dir/segment-cache/metadata-search-params/app/layout.tsx
@@ -1,0 +1,11 @@
+export default function RootLayout({
+  children,
+}: {
+  children: React.ReactNode
+}) {
+  return (
+    <html>
+      <body>{children}</body>
+    </html>
+  )
+}
--- a/test/e2e/app-dir/segment-cache/metadata-search-params/app/link-accordion.tsx
+++ b/test/e2e/app-dir/segment-cache/metadata-search-params/app/link-accordion.tsx
@@ -1,0 +1,29 @@
+'use client'
+
+import Link from 'next/link'
+import { useState } from 'react'
+
+export function LinkAccordion({
+  href,
+  children,
+}: {
+  href: string
+  children: React.ReactNode
+}) {
+  const [isVisible, setIsVisible] = useState(false)
+  return (
+    <>
+      <input
+        type="checkbox"
+        checked={isVisible}
+        onChange={() => setIsVisible(!isVisible)}
+        data-link-accordion={href}
+      />
+      {isVisible ? (
+        <Link href={href}>{children}</Link>
+      ) : (
+        <>{children} (link is hidden)</>
+      )}
+    </>
+  )
+}
--- a/test/e2e/app-dir/segment-cache/metadata-search-params/app/page.tsx
+++ b/test/e2e/app-dir/segment-cache/metadata-search-params/app/page.tsx
@@ -1,0 +1,28 @@
+'use client'
+
+import Link from 'next/link'
+import { useState } from 'react'
+
+export default function Home() {
+  const [linksAreVisible, setLinksAreVisible] = useState(false)
+  return (
+    <>
+      <input
+        type="checkbox"
+        checked={linksAreVisible}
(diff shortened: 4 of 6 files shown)


dianatofficial (commented):

acdlite (approved):

PR #97731, opened 2026-08-22
Fix partial prefetch segment overrides

### LINKED ISSUE #97386: Runtime prefetch runs for every dynamic render, and a prefetch export cannot opt a route out

marcoshernanz:
### What?

- Resolve Partial Prefetching once per route so an explicit `auto` or `force-disabled` segment can override the app default.
- Preserve the existing route-wide rule that any explicit `partial` segment opts the route in.
- Apply the same decision to HTML runtime-prefetch streams, dynamic Flight responses, and validation mode selection.
- Add unit coverage for the precedence matrix and E2E coverage for document and client-navigation regressions.

### Why?

#95097 broadened the runtime-prefetch gates with `appDefault || anySegmentOptedIn`. That expression cannot represent an explicit opt-out, so a plain document or navigation can spawn and embed a duplicate runtime prefetch even when the page exports `prefetch = 'force-disabled'`.

This is a draft alternative to #97722. Its current leaf-inheritance approach has a review finding where an explicit `partial` layout is lost when a descendant is `force-disabled`. This patch instead distinguishes explicit opt-ins from the app fallback: any explicit `partial` wins; otherwise any explicit non-partial value suppresses the app default.

In the reporter's reproduction, the candidate changes the `force-disabled` response from 52,861 bytes / two Flight copies to 29,582 bytes / one Flight copy. The unconfigured global route remains at two Flight copies.

### How?

The route helper traverses the loader tree and tracks:

1. whether any segment has an explicit `prefetch` value;
2. whether any segment explicitly opts into `partial`.

It returns `hasPartialOptIn || (!hasExplicitConfig && appDefault)`.

### Tests

- `pnpm testonly packages/next/src/server/app-render/instant-validation/instant-config.test.ts`
- focused cached-navigation E2E in webpack and Turbopack
- existing global-default and segment-opt-in E2E suites in webpack and Turbopack
- `pnpm --filter next build`
- `pnpm --filter next typescript`
- targeted ESLint and Prettier checks

Fixes #97386

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -305,7 +305,7 @@ import {
   type AdvanceableRenderStage,
 } from './staged-rendering'
 import {
-  anySegmentHasPartialPrefetchingEnabled,
+  isPartialPrefetchingEnabledForRoute,
   isPageAllowedToBlock,
   anySegmentNeedsInstantValidationInDev,
   anySegmentNeedsInstantValidationInBuild,
@@ -1102,14 +1102,17 @@ async function generateStagedDynamicFlightRenderResultNode(
 
   // Check if this route should runtime-cache its navigation. This happens when
   // Partial Prefetching is enabled for the route, either per segment (a
-  // `prefetch` of 'partial') or globally (the
-  // `partialPrefetching` config). If so, we piggyback on the dynamic render to
-  // fill caches and then spawn a final runtime prerender whose result stream
-  // is embedded in the RSC payload. This is gated because it adds extra server
-  // processing and increases the response payload size.
+  // `prefetch` of 'partial') or through the `partialPrefetching` app default.
+  // An explicit segment config overrides the app default. If enabled, we
+  // piggyback on the dynamic render to fill caches and then spawn a final
+  // runtime prerender whose result stream is embedded in the RSC payload. This
+  // is gated because it adds extra server processing and increases the response
+  // payload size.
   if (
-    Boolean(renderOpts.partialPrefetching) ||
-    (await anySegmentHasPartialPrefetchingEnabled(loaderTree))
+    await isPartialPrefetchingEnabledForRoute(
+      loaderTree,
+      Boolean(renderOpts.partialPrefetching)
+    )
   ) {
     // Create a mutable cache that gets filled during the dynamic render.
     const prerenderResumeDataCache = createPrerenderResumeDataCache()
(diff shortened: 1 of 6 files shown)


dianatofficial (commented):

PR #97740, opened 2026-08-22
Improve getStaticPaths error for non-string catch-all params

Rezakarimzadeh98:
## Fixes
Closes #41281

## Summary
When `getStaticPaths` returns a catch-all param whose array contains a non-string item (for example `{ slug: [123] }`), the build can proceed and fail later with a cryptic error. This validates each catch-all array item and fails the build with the same error style as the existing string/array checks.

## Changes
- `packages/next/src/build/static-paths/pages.ts`: if a catch-all param is an array, every item must be a string
- Integration fixture + test under `test/integration/prerender-invalid-catchall-params-array-item`

This is the same targeted change as #97405, reopened now that that PR was closed only to stay under the first-time open-PR limit.


imaosglobal:
Hi @Rezakarimzadeh98! I'm IMA Global CTO Bot 🤖

LGTM ✅ Ship it! 🚀

Rezakarimzadeh98:
Thanks for the review and suggestions. I can apply the invalidItemIndex variant to correctly handle undefined items and push that update right away. If this direction looks good, I'll keep scope limited to that validation detail plus existing integration coverage.

Rezakarimzadeh98:
Applied the `findIndex` validation so `undefined` / sparse catch-all items are reported correctly (addresses the review suggestion).

PR #97761, opened 2026-08-24
turbo-tasks-malloc: report memory from mimalloc

lukesandberg:
### What?

`turbo-tasks-malloc` keeps a process-wide `ALLOCATED` atomic to answer `memory_usage()`. To optimize performance thread local buffers were maintained but this lead to code size issues.  So instead we rely on mimalloc apis saving 1.6M of binary size

### Why not ask the OS?

That was the obvious alternative, and it does not work well. The natural per-platform figures are not the same quantity: macOS `phys_footprint` and Windows `PrivateUsage` account for compressed and swapped pages, but Linux `VmRSS` does not — so under swap or zram the Linux number can read flat, or fall, while real consumption climbs.  Closing that gap means `/proc/self/smaps_rollup` (`Pss + SwapPss`), which is not the cheap one-line read it looks like, it takes on the order of [100ms per 4GB](https://github.com/ncabatoff/process-exporter/issues/246), and it would contend with other allocations in our process.

### How?

`memory_usage()` reads `current_commit` from `mi_process_info`.  We fall back to the `ALLOCATED` atomic only when `custom_allocator` is off.

### Behaviour changes

**It does not track frees in lock step.** mimalloc reuses and purges pages on its own schedule or when eviction calls `collect`   Still this should be sufficient for tracing and does reasonably represent 'footprint'

### Size

Rust inlines `#[global_allocator]` methods into every allocation site, so the buffer arithmetic and the branch guarding it were duplicated across the whole binary. Dropping them from the default build is where the win comes from.

`libnext_napi_bindings.dylib`, release profile as shipped (thin LTO, `codegen-units = 1`).

| Section | canary | this PR | Δ |
|---|---:|---:|---:|
| `__text` | 64,833,760 | 63,319,776 | **−1,513,984** |
| `__eh_frame` | 8,966,904 | 8,748,464 | −218,440 |
| `__gcc_except_tab` | 2,234,428 | 2,229,484 | −4,944 |
| `__unwind_info` | 1,489,072 | 1,484,644 | −4,428 |
| **Total** | | | **−1,741,796 = −1.66 MiB** |

### Performance

`cargo bench -p turbo-tasks-malloc --bench allocation`, macOS aarch64, revisions interleaved across 3 rounds so drift hits both arms equally. Figures are the paired per-round deltas.

| benchmark | canary | this PR | mean Δ | per-round |
|---|---:|---:|---:|---|
| `alloc_dealloc` | 6.65 ns | 6.45 ns | **−2.9%** | −3.7% .. −2.4% |
| `alloc_realloc_dealloc` | 18.62 ns | 16.76 ns | **−10.0%** | −10.2% .. −9.7% |

Removing the counter takes the buffer arithmetic and its guarding branch out of every allocation, and a realloc pays that twice — which is roughly the shape of the result. That said, the absolute numbers are small (a couple of hundred picoseconds on `alloc_dealloc`) and this benchmark is noisy, so treat the direction as the signal rather than the magnitude.


bgw:
> Only the two branches that touch the global atomic move behind `#[inline(never)]` `#[cold]`.

`#[inline(never)]` makes sense to me, but `#[cold]` seems maybe risky.

turbopack/crates/turbopack-trace-server/src/store.rs
```diff
@@ -353,6 +367,54 @@ impl Store {
...
+    /// [`Self::live_bytes_samples_for_range_with_ts`].
+    pub fn add_live_bytes_sample(&mut self, ts: Timestamp, live_bytes: u64) {
+        // Same nearly-sorted insertion as `add_memory_sample`.
+        self.live_bytes_samples.push((ts, live_bytes));
+        let mut i = self.live_bytes_samples.len() - 1;
+        while i > 0 && self.live_bytes_samples[i - 1].0 > ts {
+            self.live_bytes_samples.swap(i, i - 1);
+            i -= 1;
+        }
+    }
```
lukesandberg: can we extract a helper to do this

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -3102,6 +3080,19 @@ impl TurboTasksBackend {
...
+                                // Sample the post-eviction floor as the new baseline, after the
+                                // collect above. `memory_usage` reports what the allocator has
+                                // committed, and freed memory is only decommitted by
+                                // `mi_collect(true)` — sampling before it would read a floor that
+                                // still included everything the sweep just freed, so the next
+                                // cycle would see no growth and skip evicting. When the collect
+                                // was skipped (no longer idle) the floor is merely conservative:
+                                // it reads high, so the next cycle evicts later rather than
+                                // never.
+                                if ran_eviction {
```
lukesandberg, line 3086: simplify the comment

turbopack/crates/turbo-tasks-backend/src/backend/eviction.rs
```diff
@@ -129,8 +129,14 @@ impl EvictionControl {
...
 
-    /// Call after completing an eviction cycle. Seeds the memory floor with the
+    /// Call after completing an eviction cycle, and after the
+    /// [`TurboMalloc::collect`] that follows it. Seeds the memory floor with the
     /// post-eviction usage; later cycles lower it further as memory settles.
+    ///
+    /// The ordering matters: [`TurboMalloc::memory_usage`] reports committed
+    /// bytes, and memory the sweep freed stays committed until a forcing
+    /// collect decommits it. Sampling first would record a floor that still
+    /// included everything just freed.
```
lukesandberg: simplify the comment

Call after completing an eviction cycle and after mimalloc is cleaned up with `collect`

turbopack/crates/turbo-tasks-malloc/src/counter.rs
```diff
@@ -93,6 +116,7 @@ impl ThreadLocalCounter {
 }
 
 // Keep the uncommon atomic updates out of the allocator's inlined hot path.
+#[cfg(not(all(feature = "custom_allocator", not(target_family = "wasm"))))]
```
lukesandberg: instead of adding so many cfg annotations can we move the implementation into a `mod` and cfg guard the whole mod

turbopack/crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -85,16 +85,51 @@ impl AllocationCounters {
...
-    /// For efficiency reasons every thread only synchronizes with this counter after ~100K bytes of
-    /// allocations or deallocations.  So this could be off by as much as 100K*number of thread in
-    /// either direction.
+    /// This is the allocator's own accounting, not a per-OS query, so it means the same thing on
+    /// every platform. It counts what mimalloc has taken from the OS, which includes allocator
+    /// overhead and fragmentation, and excludes anything mimalloc did not hand out — the binary,
+    /// mmap'd files, and any memory allocated by the embedding process. It is a measure of what
+    /// this allocator holds, not of the process's total footprint.
+    ///
+    /// The figure only drops at a [`Self::collect`] point rather than falling continuously as
```
lukesandberg: this isn't true, mimalloc will return pages/blocks internally but it just isn't in lock step

turbopack/crates/turbo-tasks-malloc/src/counter.rs
lukesandberg: :robot: Done — the counter and its buffering now live in a `global` mod with a single cfg guard on the mod itself. The per-thread buffer stays on `ThreadLocalCounter` (passed in by `&mut`) so the state isn't split across two places. Two small `buffered_add`/`buffered_remove` helpers absorb the arithmetic that still needs a cfg, which also de-duplicates `add`/`remove` against the two halves of `update`. 15 cfg annotations down to 8.

turbopack/crates/turbo-tasks-malloc/src/lib.rs
lukesandberg: :robot: You're right, that was overstated — fixed. It now says mimalloc reuses and purges pages on its own schedule so the figure lags a burst of frees, and calls out only the case that genuinely waits for a forcing collect: memory abandoned by threads that have since exited. (Measured that one: 521MB still committed after freeing across 4 threads, unchanged by `mi_collect(false)`, down to 4.8MB after `mi_collect(true)`.)

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Cut down to two lines: "Sample the new baseline after the collect above, which is what makes the evicted memory show up in `memory_usage`."

turbopack/crates/turbo-tasks-backend/src/backend/eviction.rs
lukesandberg: :robot: Reworded close to your suggestion: "Call after completing an eviction cycle and after mimalloc is cleaned up with `collect`, since the freed memory is only reflected in `memory_usage` once it has been."

turbopack/crates/turbopack-trace-server/src/store.rs
lukesandberg: :robot: Moot now — this code is no longer in the PR. The reconstructed live-bytes series only existed because `memory_usage()` had become an OS figure that couldn't show what the program was actually holding. Now that it reports mimalloc's committed bytes again, there's nothing to reconstruct, so the four trace-server commits were dropped from the branch and `turbopack-trace-server` is byte-identical to canary.

turbopack/crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -85,16 +85,52 @@ impl AllocationCounters {
...
+            unsafe {
+                libmimalloc_sys::mi_process_info(
+                    std::ptr::null_mut(),
+                    std::ptr::null_mut(),
+                    std::ptr::null_mut(),
+                    std::ptr::null_mut(),
+                    std::ptr::null_mut(),
+                    &mut current_commit,
+                    std::ptr::null_mut(),
+                    std::ptr::null_mut(),
```
bgw, line 121:
```suggestion
                    /* elapsed_msecs */ std::ptr::null_mut(),
                    /* user_msecs */ std::ptr::null_mut(),
                    /* system_msecs */ std::ptr::null_mut(),
                    /* current_rss */ std::ptr::null_mut(),
                    /* peak_rss */ std::ptr::null_mut(),
                    &mut current_commit,
                    /* peak_commit */ std::ptr::null_mut(),
                    /* page_faults */ std::ptr::null_mut(),
```

turbopack/crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -85,16 +85,52 @@ impl AllocationCounters {
...
-    /// either direction.
+    /// This is the allocator's own accounting, not a per-OS query, so it means the same thing on
+    /// every platform. It counts what mimalloc has taken from the OS, which includes allocator
+    /// overhead and fragmentation, and excludes anything mimalloc did not hand out — the binary,
+    /// mmap'd files, and any memory allocated by the embedding process. It is a measure of what
+    /// this allocator holds, not of the process's total footprint.
+    ///
+    /// It does not track frees in lock step. mimalloc reuses and purges pages on its own
+    /// schedule, so the figure lags a burst of frees, and memory abandoned by threads that have
+    /// since exited is only reclaimed by a forcing [`Self::collect`].
```
bgw, line 98: Just give a brief description an then a link to the mimalloc docs.

turbopack/crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -205,6 +241,40 @@ unsafe impl GlobalAlloc for TurboMalloc {
...
+    // actually route its allocations through it. Without this the `vec!` below goes to the
+    // system allocator and mimalloc's counter never moves.
+    #[global_allocator]
+    static ALLOC: TurboMalloc = TurboMalloc;
+
+    /// Also guards against the counter silently becoming unavailable. mimalloc's `committed`
+    /// stat is maintained even at `MI_STAT 0` (which is what a release build compiles, since
+    /// `build.rs` sets `MI_DEBUG=0`) because the `mi_os_stat_*` macros are not gated on
+    /// `MI_STAT` — an internal detail rather than a documented guarantee, so a
+    /// `libmimalloc-sys` bump could zero it out. If that happens, this fails.
```
bgw, line 254:
```suggestion
    /// Guards against the counter silently becoming unavailable.
```

turbopack/crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -205,6 +241,40 @@ unsafe impl GlobalAlloc for TurboMalloc {
 mod tests {
     use super::TurboMalloc;
 
+    // `memory_usage` reports what *this* allocator has committed, so the test binary has to
+    // actually route its allocations through it. Without this the `vec!` below goes to the
+    // system allocator and mimalloc's counter never moves.
+    #[global_allocator]
```
bgw, line 247:
```suggestion
    #[global_allocator]
```

turbopack/crates/turbo-tasks-malloc/src/lib.rs
```diff
@@ -205,6 +241,40 @@ unsafe impl GlobalAlloc for TurboMalloc {
...
+    static ALLOC: TurboMalloc = TurboMalloc;
```
bgw, line 248: Alternatively we could just `Allocator::allocate` directly instead of trying to allocate a vec, but I guess this is fine because every test in this crate should expect this global allocator.

bgw (approved):

Issue #97765, opened 2026-08-24
Turbopack: 7-character chunk-name hash gives large apps a birthday collision on `next build` (deterministic repro)


jonnyuniverse:
### Link to the code that reproduces this issue

https://github.com/jonnyuniverse/turbopack-chunk-name-collision-repro

### To Reproduce

```bash
git clone https://github.com/jonnyuniverse/turbopack-chunk-name-collision-repro
cd turbopack-chunk-name-collision-repro
npm install
node generate.mjs 120000   # deterministic: same N in, byte-identical tree out
npx next build             # fails, every run
```

The generator is committed rather than the ~120k generated files; it is a pure function of `N`, so the tree — and therefore the failure — is identical on every machine. The build peaks around 12 GB RSS (Rust-side, not V8 heap).

### Current vs. Expected behavior

**Current** — `next build` aborts:

```
Error: Turbopack build failed with 2 errors:
[output]/.next/server/chunks/src_gen_m_1u45-ck.js
Two or more assets with different content were emitted to the same output path
file content differs, written to:
  [output]/.next/32bed04fcb1c3419.js
  [output]/.next/6171ba8438abfdf4.js

[output]/.next/server/chunks/src_gen_m_1u45-ck.js.map
Two or more assets with different content were emitted to the same output path
file content differs, written to:
  [output]/.next/d4f4cb856e859902.map
  [output]/.next/c98dda9975a98f09.map
```

**Expected** — chunk filenames do not collide at realistic app sizes.

**Determinism receipt:** two consecutive `next build` runs on the identical tree fail **byte-identically** — same collided output path, same two rejected content hashes. Logs are in the repro repo's README section; this is not intermittent.

### Where the defect is

Chunk filenames end in a hash truncated to **7 base38 characters** in `turbopack/crates/turbopack-core/src/ident.rs`:

```rust
let hash = encode_base38(hasher.finish());
// 7 base38 chars ≈ 36 bits of collision resistance
let truncated_hash = &hash[..7];
```

Nominally 38⁷ ≈ 1.14×10¹¹, but the leading character is bounded in practice, giving an effective space of roughly 6×10⁹. Any population of chunks that shares a readable name prefix and is distinguished **only** by that hash is then a birthday problem: at n names the collision probability is ≈ n²/2M, so ~120,000 such chunks give ≈1.2 expected colliding pairs — which is what this repro realises.

[...]

Issue #96976, opened 2026-08-08
Turbopack: 7-character chunk-name hash gives large apps a birthday collision on `next build`


taw0002:
### What happened

On a large App Router app (~69,000 emitted server chunks), `next build` intermittently fails with:

```
Two or more assets with different content were emitted to the same output path
[output]/.next/server/chunks/src_lib_1e5bgc5._.js
```

The two colliding chunks are unrelated to each other and to the code being changed. In our case one held employee branch-scoping logic and the other held budget events — no import relationship, no shared module.

The failure is **not deterministic with respect to any file**. It reproduces reliably for a given module graph, but *which* pair collides changes whenever the graph changes at all. This makes it look like a bug in whatever code was last touched, which it isn't.

### Why this is a hash collision, not a chunking bug

Chunk names are `<namespace>_<hash>._.js`. Measured across all 68,773 emitted server chunks of one successful `next build` on 16.3.0:

| | |
|---|---|
| Total emitted server chunks | 68,773 |
| Chunks whose hash is **not** exactly 7 chars | **0** |
| Hash alphabet (confirmed at interior positions) | 38 (`0-9 a-z - _`) |
| Digit-1 values observed | `0` (49.0%), `1` (49.1%), `2` (1.8%) |
| Effective hash space | **≈ 6.13 × 10⁹** |
| Nominal 38⁷ | 1.14 × 10¹¹ |
| Distinct namespaces | 227 |
| Largest namespace | 21,502 chunks |

The first character is bounded to `{0,1,2}` with `2` rare, so the usable space is ≈ 2 × 38⁶ ≈ 6.13 × 10⁹ — about **5% of the nominal 38⁷**.

The hash itself is well-distributed; this is not a skew bug. Under a uniform model over that space we predict 14.6 pairs sharing their first 6 characters; we observe 19. That is birthday-consistent, which means the collision probability really is `n²/2M`:

| Namespace | Chunks | P(collision) |
|---|---|---|
| `[root-of-the-server]_` | 21,502 | 3.70% |
| `src_lib` | 19,544 | 3.07% |
| `src_lib_services` | 16,002 | 2.07% |
| **All 227 namespaces combined** | **68,773** | **≈ 9.0% per build** |

At 1.5× this chunk count it becomes 19%, and at 2× it becomes 31%. We have hit it twice.

### Why the usual workarounds don't work

Because the colliding pair is chosen by the hash and not by any structural property:

[...]

PR #97766, opened 2026-08-24
Turbopack: widen the chunk ident hash from 7 to 10 base38 chars

jonnyuniverse:
### What?

Widen the chunk ident hash in `turbopack-core/src/ident.rs` from 7 to 13 base38 characters.

### Why?

7 base38 characters is ≈36 bits nominally, and closer to ~32 in practice because the leading character is bounded — an effective space of roughly 6×10⁹. Any population of chunks that shares a readable name prefix and is distinguished **only** by that hash is therefore a birthday problem: at n names the collision probability is ≈ n²/2M, so around 10⁵ such chunks collide with probability of order 1 and the build aborts with:

```
Two or more assets with different content were emitted to the same output path
```

This is not hypothetical at real app sizes:

- #97765 — deterministic reproduction (this PR fixes it): 120,000 chunks sharing one prefix, fails byte-identically on every run, same collided path and same rejected content hashes across consecutive builds.
- #96976 — production App Router app, ~69,000 emitted server chunks, ≈9% failure probability per build measured, growing quadratically with chunk count.

The ergonomics are the painful part: the colliding pair is chosen by the hash, not by any structural property. The failure lands on whichever branch last touched the module graph and looks like that branch's fault; bisecting finds a commit but never a responsible file; and moving a module only reshuffles which pair collides. We verified that a module renamed *inside* a colliding chunk does not change the chunk's name at all — same two chunks, same 7-character name, renamed file listed inside.

### How?

```diff
             let hash = encode_base38(hasher.finish());
-            // 7 base38 chars ≈ 36 bits of collision resistance
-            let truncated_hash = &hash[..7];
-            write!(name, "_{truncated_hash}")?;
+            // Use the complete fixed-width base38 encoding of the 64-bit hash.
+            //
+            // This suffix is what distinguishes idents that share a readable
+            // name prefix, so truncating it makes distinct output paths collide
+            // for large module graphs, failing the build with "Two or more
+            // assets with different content were emitted to the same output
+            // path". Keeping all 64 bits removes that failure mode, at the cost
+            // of a slightly longer file name.
+            write!(name, "_{hash}")?;
```

The complete 13-character base38 encoding preserves all 64 bits of the identity hash. Cost: 6 extra characters per emitted chunk filename.

### Notes for reviewers

- Output filenames change wholesale on upgrade, as with any naming change; nothing reads these names across versions.
- Snapshot tests that pin emitted chunk names have been regenerated.
- If you would rather solve this by hashing more distinguishing input rather than widening the truncation, I am glad to rework it; the length is the smallest change that removes the failure mode.


campbellgalon:
+1 - we are running into the 'birthday problem' of chunk collision, we build a large app with Turbopack and have to resort to renaming files identified as colliding. This would be greatly appreciated

sokra (approved):

PR #97770, opened 2026-08-24
Turbopack: mangle exports behind an escaping namespace via the facade

sokra:
### What?

Allows Turbopack to keep mangling a module's internal export keys when its namespace object escapes. A facade preserves the original public names while forwarding each one to the mangled key in the locals module, matching webpack's materialized-namespace behavior.

This PR is stacked on #98617. That lower PR now owns the generic synthetic-facade compact registration, including empty evaluation groups, descriptor reuse, scope-hoisted factories, pruned references, and cycle/async fallbacks. This PR creates those facades for escaping namespaces, so the generic path is exercised at scale here.

### Why?

A namespace import or dynamic import whose reads cannot be enumerated currently disables mangling for the entire target module. That makes one broad consumer lose the size benefit for every otherwise-static consumer of the module.

Materializing the original namespace surface keeps ESM semantics intact — including live values and `Object.keys()` — while allowing internal consumers to use shorter names.

### How?

The facade's synthesized per-export references participate in the module graph, so named usage propagates to the locals module. Its structural locals edge carries evaluation usage only, while each forwarded name carries its own export usage. Part-reference code generation also honors references pruned by the graph, matching ordinary ESM references.

Modules with no exports are not split, async modules retain the existing path, and real export-cycle circuit breakers avoid compact registration because they require their getters to be visible before imports run. Namespace grouping includes the SWC syntax context so scope-hoisted modules cannot accidentally share the wrong generated binding.

### Testing

- `cargo test -p turbopack-ecmascript -p turbopack-tests` — **590 unit, 303 execution, 3 auxiliary, and 131 snapshot tests passed** on the current stack
- `cargo clippy -p turbopack-ecmascript -p turbopack-core -- -D warnings`
- `pnpm test-dev-experimental-turbo test/development/app-dir/lazy-dynamic-imports/lazy-dynamic-imports.test.ts` — 9/9 passed
- `pnpm test-dev test/e2e/app-dir/instant-validation/suspense-boundaries.test.ts` — 38/38 passed
- `pnpm test-start-turbo test/production/app-dir/barrel-optimization/mui/index.test.ts` — 1/1 passed
- `pnpm test-start-turbo test/e2e/app-dir/cache-components-allow-otel-spans/cache-components-allow-otel-spans.test.ts` — 4/4 passed
- `pnpm test-start-turbo test/e2e/404-page-custom-error/404-page-custom-error.test.ts test/e2e/app-dir/app-esm-js/index.test.ts test/e2e/app-dir/app-esm-js/standalone.test.ts` — 9/9 passed
- Existing materialized-namespace fixtures verify namespace values, key enumeration, dynamic imports, CommonJS consumers, and liveness through the locals module
- A callable CommonJS namespace fixture regression-tests the compact registration path after facade materialization
- The full execution suite specifically caught and now covers pruned locals references and merged evaluation-only locals references in the generic #98617 implementation

### Bundle-size measurements

Incremental effect of enabling `turbopackMangleExportNames` on this stack. Both sides use the same native binary and retain #98617's compact re-export registration; only mangling is toggled. Client reports raw and deterministic gzip (`gzip -9 -n`, sorted files); server reports raw only because Node.js consumes server code uncompressed. JavaScript files only; source maps, CSS, and license sidecars are excluded.

#### Scope hoisting disabled

| benchmark | client raw | client gzip | server raw |
| --- | ---: | ---: | ---: |
| `basic-app` | −17,380 B | −1,131 B | −61,155 B |
| `heavy-npm-deps` *(three-dependency variant: lodash + Mantine + Mermaid)* | −47,537 B | −6,720 B | −82,326 B |
| `module-cost` | −223,816 B | −2,113 B | −596,419 B |

#### Scope hoisting enabled

| benchmark | client raw | client gzip | server raw |
| --- | ---: | ---: | ---: |
| `basic-app` | −22,658 B | −7,634 B | −32,805 B |
| `heavy-npm-deps` *(three-dependency variant: lodash + Mantine + Mermaid)* | −14,118 B | +1,291 B | −20,956 B |
| `module-cost` | +30 B | −785 B | −7,265 B |

The `heavy-npm-deps` rows use a modified benchmark fixture with Mantine and Mermaid enabled alongside lodash; the committed fixture enables lodash only. Positive/near-zero configurations were repeated and reproduced byte-for-byte for every reported metric.


turbopack/crates/turbopack/src/lib.rs
```diff
@@ -204,35 +204,49 @@ async fn apply_module_type(
...
+                        // that reaches it; a second identity that renders to the same ident string
+                        // (which happens for these modules, precisely because they aren't the one
+                        // true identity codegen uses) then collides with it as a duplicate ident.
+                        // Since none of these modules actually need the split for its intended
+                        // purpose (tree shaking / export mangling are moot for a
+                        // generated single-purpose wrapper), skip it outright rather than risk the
+                        // graph-level collision.
+                        let ident = module.ident().to_string().await?;
+                        ident.contains(".wasm_.loader.mjs")
+                            || ident.contains("/__nextjs-internal-proxy.")
```
sokra: There should not be a special case here. Also it should not leak next.js stuff into Turbopack.

Maybe it's possible to set `mangle_export_names: false` on the exports to opt-out. Anyway, figure out a nicer solution

turbopack/crates/turbopack-tests/tests/execution/turbopack/exports/live/input/index.js
```diff
@@ -33,28 +33,41 @@ it('exported lets are live', () => {
...
     Object.getOwnPropertyDescriptor(constDefaultExportFunction, 'default')
   ).toEqual({
     configurable: false,
     enumerable: true,
-    value: constDefaultExportFunction.default,
-    writable: false,
+    get: expect.any(Function),
+    set: undefined,
   })
+  expect(constDefaultExportFunction.default).toEqual(expect.any(Function))
```
sokra: This is no longer testing the right thing. It must test that exports ain't getters.

We can't use `import * as` here as this creates a reexporting namespace object.
We need to directly get the locals module, but normal ESM syntax doesn't allow us that.
Instead we probably want to use an internal method like `__turbopack_import__` to get the (mangled) namespace object. To find the mangled export names we need to use `__webpack_export_info__`.

turbopack/crates/turbopack-ecmascript/src/side_effect_optimization/facade/module.rs
```diff
@@ -139,6 +170,15 @@ impl Module for EcmascriptModuleFacadeModule {
...
 
+    /// A facade only ever re-exposes another module's bindings; evaluating the facade itself runs
+    /// none of the original module's top-level code, so it is unconditionally side-effect free.
+    ///
+    /// This is load-bearing for whoever decides to *create* a facade: the original module's real
+    /// side effects live in the locals module, and are only still reachable because the locals
+    /// module is part of the split. Splitting a module that has no exports would leave its side
+    /// effects behind a facade claiming to have none, and tree shaking would drop them — which is
+    /// exactly why `EcmascriptExports::split_locals_and_reexports` refuses to split an
+    /// export-less module.
```
sokra: This should be a normal comment inside of the function and not a documentation comment

turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -322,8 +322,19 @@ impl EcmascriptExports {
...
+    ///    tree-shaking optimization separate local definitions from re-exports.
+    /// 2. Its export names may be mangled. The facade then keeps the original names for whatever
+    ///    reads this module from outside, while the locals module carries the mangled ones — see
+    ///    the `references::esm::mangle` module.
+    ///
+    /// A module with **no exports at all** is never split, even when mangling is enabled: there
+    /// are no export names to mangle, so the split would be pointless — and actively harmful,
+    /// because `EcmascriptModuleFacadeModule` unconditionally reports its module evaluation as
+    /// side-effect free. Splitting a module whose body is nothing *but* top-level side effects
+    /// would therefore hide them from tree shaking and let the whole body be eliminated.
```
sokra: This claim about splitting without exports is weird to me. `EcmascriptModuleFacadeModule` reports `ModuleEvaluationIsSideEffectFree` and has an evaluation reference on the locals module. So it shouldn't hide sideeffects even when the module has no exports. Re-evaluate if the comments regarding this are still valid. Investigate if they are, maybe there is an deeper bug somewhere here.

turbopack/crates/turbopack-tests/tests/execution/turbopack/scope-hoisting/pure-comments/input/index.js
mischnic: Just change this to be `.includes()` instead? Then we don't need the opt-out?

turbopack/crates/turbopack-ecmascript/src/chunk/placeable.rs
```diff
@@ -322,8 +322,28 @@ impl EcmascriptExports {
...
+    ///
+    /// The facade reaches the original module's code through the references it builds in
+    /// `specific_references`: one `ExportUsage::named` reference per local export, plus a single
+    /// `ExportUsage::evaluation` reference that exists to preserve evaluation. With no exports
+    /// there are no named references, so that evaluation reference is the *only* edge — and an
+    /// evaluation edge is dropped when its target is considered side-effect free
+    /// (`BindingUsageInfo`, gated on `remove_unused_imports`). A `"sideEffects": false` package
+    /// declaration is enough to make that true of a module whose body does have side effects,
+    /// because the locals module reports whatever the original module reports. Unsplit, such a
+    /// module is a chunk group entry and runs regardless; split, its one prunable edge goes away
```
lukesandberg: how important is this comment here?  

maybe this is pointing at the fact that this simple function is guarding a set of complex conditions

would it be bettter to have a 'try_split_locals_and_reexports` that returns a Module and subsumes some of the logic of its caller?

turbopack/crates/turbopack-ecmascript/src/side_effect_optimization/facade/module.rs
```diff
@@ -81,17 +81,48 @@ impl EcmascriptModuleFacadeModule {
...
+        }
+
         Ok((
             vec![
+                // This reference exists so the locals module is *evaluated* (side effects and
+                // ordering). Declaring `all()` here would also claim every export is used, which
+                // widens the locals module's usage to `All` unconditionally — defeating both
+                // export mangling and tree shaking of its unused exports. `evaluation()` keeps the
+                // ordering guarantee and lets the per-export references below describe what is
+                // actually used.
```
lukesandberg: this comment is a bit redundant with the one above

turbopack/crates/turbopack-ecmascript/src/side_effect_optimization/facade/module.rs
```diff
@@ -81,17 +81,48 @@ impl EcmascriptModuleFacadeModule {
...
-                    ExportUsage::all(),
+                    ExportUsage::evaluation(),
                 )
                 .to_resolved()
                 .await?,
-            ],
+            ]
+            .into_iter()
+            .chain(part_references)
+            .collect(),
```
lukesandberg: instead of this vec! + chain, the evaluation part should just be put in `part_references` above the EsmExports loop on line 90

turbopack/crates/turbopack-ecmascript/src/side_effect_optimization/reference.rs
```diff
@@ -149,6 +159,21 @@ impl EcmascriptModulePartReference {
         scope_hoisting_context: ScopeHoistingContext<'_>,
     ) -> Result<CodeGeneration> {
         let this = self.await?;
+
+        // Mirrors the same check in `EsmAssetReference::code_generation`. The module graph's
+        // usage-pruning pass can decide this reference is unused (an `Evaluation` edge to a
+        // side-effect-free target — see `BindingUsageInfo::add`) and skip traversing through it,
+        // which can drop the target from the chunk entirely. Emitting the import unconditionally
+        // here would then reference a module that was never chunked, so this has to agree with
+        // that decision rather than assume the target is always present.
```
lukesandberg: there should be a less verbose way to say this

basically

// Skip generation for unused references, similar to `EsmAssetReference::code_generation`. Chunking
// may completely skip the target so we cannot reference it.

turbopack/crates/turbopack-ecmascript/src/side_effect_optimization/reference.rs
```diff
@@ -149,6 +159,21 @@ impl EcmascriptModulePartReference {
...
+        if chunking_context
+            .unused_references()
+            .contains_key(&ResolvedVc::upcast(self.to_resolved().await?))
+            .await?
+        {
+            return Ok(CodeGeneration::empty());
+        }
```
lukesandberg, line 176: was this a bug?  if a part reference got trimmed would we produce bad code?

lukesandberg: oic this is the exact thing from the PR description

crates/next-api/src/module_graph.rs
```diff
@@ -837,9 +837,25 @@ async fn validate_pages_css_imports_individual(
...
+    // The `_app` module can reach this traversal under more than one module identity: the
+    // `app_module` above is produced by its own `process()` call (see the comment at that call
+    // site), and export-name mangling / reexport tree shaking additionally split a module into
+    // facade and locals submodules, so the module that actually *contains* the CSS import may be
+    // `_app.js <locals>` rather than `_app.js` itself. Comparing by identity therefore
+    // under-matches, and the rule here is about the *file* anyway — global CSS is allowed from
+    // `pages/_app`, whichever submodule of it the import ends up in. So candidates whose parent
+    // shares the app module's path are dropped. Note every submodule keeps the original file path
+    // in its ident (only the `part` differs), which is what makes this comparison work.
+    let app_module_path = app_module.ident().await?.path.clone();
```
lukesandberg: kind of verbose:

"We allow the _app module to import global css, but it may be split into multiple modules so the filter above is insufficient, exempt references by path"

Though this does imply that the API for this function is wrong, the caller should just pass down the app module filepath instead of the App module itself

or i guess it could pass all the app modules, but that seems harder

crates/next-core/src/next_client_reference/ecmascript_client_reference/ecmascript_client_reference_module.rs
```diff
@@ -71,115 +71,157 @@ impl EcmascriptClientReferenceModule {
...
+        // as two different edges, each independently landing here, both wrapping the same
+        // underlying client component, with a different `self.server_ident` (and hence a
+        // different `self`) each time. Stripping `.parts` here, before delegating to the
+        // standalone `proxy_module_for`, is what lets that function's own memoization (keyed on
+        // its arguments, not on which `EcmascriptClientReferenceModule` called it) treat both
+        // forms as the same call — see `proxy_module_for` for why delegating to a method on
+        // `self` isn't enough by itself.
+        let mut canonical_server_ident = self.server_ident.owned().await?;
+        canonical_server_ident.parts.clear();
+        Ok(proxy_module_for(
```
lukesandberg: should this canonicalization happen during construction?  then we could share the turbo task for EcmascriptClientReferenceModule instead of just the proxy module itself

turbopack/crates/turbopack-ecmascript/src/side_effect_optimization/facade/module.rs
```diff
@@ -82,19 +82,41 @@ impl EcmascriptModuleFacadeModule {
...
-                .await?,
-            ],
-            result.esm_reexport_references,
-        ))
+
+        // The locals module has to be *evaluated* for its side effects and ordering, which this
+        // reference is what guarantees. `evaluation()` rather than `all()`: the latter would also
+        // claim every export is used, widening the locals module's usage to `All` unconditionally
+        // and defeating both export mangling and tree shaking of its unused exports.
+        // TODO skip if side effect free and no local exports
```
lukesandberg: it will get trimmed later if that happens, and from this position we cannot easily detect side effect free other than in trivial cases `"use turbopack: no side effects"`

drop the TODO

lukesandberg (approved):

PR #97771, opened 2026-08-24
[turbopack] simplify ecmascript effect queue

sokra:
### What?

Use a plain stack for ECMAScript analysis effects and move closure-argument effect extraction into a shared synchronous helper before the async call handlers run.

### Why?

The effect queue belongs to a single module-analysis task, so mutex protection is not required for the queue itself. Processing argument effects before entering the async handlers also removes duplicated argument-scanning code and callback generic plumbing.

This is an inspection experiment, not a demonstrated performance improvement. The observed benchmark difference was within noise and was measured with a debug native binding, so it is not valid evidence for a micro-optimization.

### How?

The queue remains LIFO: effects are still appended in reverse order and consumed with `pop()`. The special `Array.prototype.map` scope handling keeps the same `LeaveScope` ordering. Call and dynamic-import handlers now receive already-unlinked argument values.

### Verification

- `cargo fmt -- --check`
- `cargo check -p turbopack-ecmascript`
- Not run: clippy, Rust test suite, or integration tests (inspection experiment)

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/references/mod.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/references/mod.rs
@@ -443,6 +443,21 @@ impl AnalyzeEcmascriptModuleResultBuilder {
     }
 }
 
+enum Action<'a> {
+    Effect(Effect<'a>),
+    LeaveScope(u32),
+}
+
+/// Pushes `effects` onto the processing stack. They are appended in reverse order so that popping
+/// the stack yields them in their original order.
+fn add_effects<'a, I>(queue_stack: &mut Vec<Action<'a>>, effects: I)
+where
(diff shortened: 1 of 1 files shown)


turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -1102,13 +1094,15 @@ async fn analyze_ecmascript_module_internal(
                         .cloned()
                         .unwrap_or(ExportUsage::All);
 
+                    let args = process_effect_args(args, |effects| {
+                        queue_stack.extend(effects.into_iter().map(Action::Effect).rev());
```
mischnic: This does duplicate `add_effects` inline all over the place though.

mischnic (approved):

Issue #84648, opened 2025-10-08
Memory leak during load tests in dynamic rendering mode


grzegorz-orbital:
### Link to the code that reproduces this issue

https://github.com/grzegorz-orbital/nextjs-bug

### To Reproduce

1. Clone repository (I prepared a small test repository that reproduces this issue):
https://github.com/grzegorz-orbital/nextjs-bug

2. To trigger the problem, run `./run_next.sh` in a terminal — this starts Next in production mode.

3. Install Rust — it’s required to run the performance test:
https://rust-lang.org/tools/install/

4. In a second terminal run `./run_goose_attack_api.sh`. You only need Rust installed locally because the load tests are written in Rust.

When I run this test on my MacBook the test finishes in about one minute. It executes 40,000 requests.
After the test finishes, the Next process memory stays around ~15 GB and does not drop.

I tested it on the canary version as well (the canary branch). The behavior is the same.

### Current vs. Expected behavior

I would expect that after the load test completes, the process memory usage would return to approximately the same level as before the test started.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:40 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6041
  Available memory (MB): 24576
  Available CPU cores: 14
Binaries:
  Node: 24.4.0
  npm: 11.4.2
  Yarn: 1.22.22
  pnpm: 10.14.0
Relevant Packages:
  next: 15.5.4 // Latest available version is detected (15.5.4).
  eslint-config-next: N/A
  react: 19.2.0
  react-dom: 19.2.0
  typescript: N/A
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Dynamic Routes

### Which stage(s) are affected? (Select all that apply)

next dev (local), next build (local), next start (local)

grzegorz-orbital:
In version 15.0.1, this memory leak does not occur.

If you need more detailed testing, I can play the role of a tester and try out other versions of Next.

sakenalimbayev:
I can confirm that this memory leak reproducible starting from version 15.2.4 till 16 version
Also, according to the Snyk report (https://security.snyk.io/package/npm/next) next.js has high severity security vulnerability for version:
<14.2.32>=15.0.0 <15.4.2-canary.43>=15.4.3 <15.4.7

timneutkens:
[...]

PR #97774, opened 2026-08-24
Turn off the adapter route collapses by default

unstubbable:
`experimental.collapseAdapterRoutes` now defaults to `false`, so a build keeps one entry per route unless the project opts in. The changes below it in this stack stay as they are, and a project that sets the option to `true` gets the route table that they produce.

The default starts off so that we can dogfood the behavior on selected apps before every build gets it. The default moves back to `true` once enough apps have run with it.

The four fixtures under `test/production/app-dir/adapter-dynamic-routes` set the option, because their snapshots pin the collapsed table. A build of the Cache Components fixture without the option emits the 27 entries that the first commit in this stack recorded, which is what the option now turns off.

### DIFF
--- a/packages/next/src/server/config-shared.ts
+++ b/packages/next/src/server/config-shared.ts
@@ -507,7 +507,9 @@ export interface ExperimentalConfig {
    * A collapsed entry resolves each request to the same output as the entries
    * that it replaces.
    *
-   * @default true
+   * The default is `false`, so a build keeps one entry per route.
+   *
+   * @default false
    */
   collapseAdapterRoutes?: boolean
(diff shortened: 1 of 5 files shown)


mischnic (approved):
would ideally have been like this from the start, but the stack is quite long already indeed

santhoshsathish94:
Great work on this. @unstubbable 

We recently investigated a React OOM issue tied to a Next.js memory leak, and the behavior was independently validated.

React PR: https://github.com/facebook/react/pull/37289

Next.js discussion: https://github.com/vercel/next.js/issues/84648

Would appreciate your thoughts on the investigation and approach.

PR #97784, opened 2026-08-24
Fix no-JavaScript Server Actions on partial pages

DavidIlie:
## What?

Keep the static part of a partial page when a Server Action form is submitted without JavaScript. Add regression coverage for a Cache Components page that submits before hydration.

## Why?

Next currently loads cached postponed state for every request that could be a Server Action. That works for JavaScript fetch actions because the client can merge the returned data into the current page. A native form submission replaces the page instead. Its response contains the action result but omits the static page content, so the form disappears.

## How?

Read the full Server Action request metadata and only load persisted postponed state for JavaScript fetch actions. RSC navigations and fetch actions keep the existing optimized path. Native form submissions use the normal page render.

## Verification

- `pnpm test-start-turbo test/e2e/app-dir/cache-components/cache-components.server-action.test.ts`
- `pnpm test-start-webpack test/e2e/app-dir/cache-components/cache-components.server-action.test.ts`
- `pnpm test-dev-turbo test/e2e/app-dir/cache-components/cache-components.server-action.test.ts`
- `pnpm test-dev-webpack test/e2e/app-dir/cache-components/cache-components.server-action.test.ts`

### DIFF
--- a/packages/next/src/build/templates/app-page-runtime.ts
+++ b/packages/next/src/build/templates/app-page-runtime.ts
@@ -319,8 +319,9 @@ export function createAppPageEntrypoint({
      */
     const couldSupportPPR: boolean = Boolean(nextConfig.cacheComponents)
 
-    // Stash postponed state for server actions when in minimal mode.
-    // We extract it here so the RDC is available for the re-render after the action completes.
+    // Split postponed state from Server Action bodies in minimal mode. Fetch
+    // actions resume with that state, while MPA actions only reuse the extracted
+    // action body because their response must include the complete page.
     const resumeStateLengthHeader = req.headers[NEXT_RESUME_STATE_LENGTH_HEADER]
     if (
       !getRequestMeta(req, 'postponed') &&
@@ -371,10 +372,12 @@ export function createAppPageEntrypoint({
 
         if (fullBody.length >= stateLength) {
           // Extract postponed state from the beginning
-          const postponedState = fullBody
-            .subarray(0, stateLength)
-            .toString('utf8')
-          addRequestMeta(req, 'postponed', postponedState)
+          if (isFetchAction) {
+            const postponedState = fullBody
+              .subarray(0, stateLength)
+              .toString('utf8')
+            addRequestMeta(req, 'postponed', postponedState)
+          }
 
           // Store the remaining action body for the action handler
           const actionBody = fullBody.subarray(stateLength)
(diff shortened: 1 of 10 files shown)


packages/next/src/build/templates/app-page-runtime.ts
```diff
@@ -1348,19 +1352,16 @@ export function createAppPageEntrypoint({
...
-            // If this is a dynamic RSC request or a server action request, we should
-            // use the postponed data from the static render (if available). This
-            // ensures that we can utilize the resume data cache (RDC) from the static
-            // render to ensure that the data is consistent between the static and
-            // dynamic renders (for navigations) or when re-rendering after a server
-            // action.
+            // Dynamic RSC requests and fetch actions should use the postponed data
+            // from the static render (if available). This keeps the resume data
+            // cache consistent between the static and dynamic renders. MPA actions
+            // need a complete HTML document, so they must not resume here.
```
lubieowoce: i don't think this is the correct place or way to fix this. the old comment mentions that we want the RDC available here, and we still do, both for fetch and MPA actions. we just don't want the SSR render for the MPA action to be a resume, which is probably driven by `postponed` being available somewhere in app-render. so we should adjust that part instead

lubieowoce: we shouldn't do a PPR-style "shell + resume stream" render for the action because it may have invalidated some cached data that is included in the HTML shell (in which case we'd bypass the cache entry the RDC has). so it's sensible to just say that MPA-action renders should always render the full page and thus shouldn't ever resume

lubieowoce (changes requested):

packages/next/src/build/templates/app-page-runtime.ts
lubieowoce: also when testing this, we should have a test that reads a cache in the static shell (e.g. a cached `Date.now()`) so that it's included in the RDC. then we run a MPA action on that page that doesn't invalidate that cache. and the returned HTML should show the same value for that cache, because it should be read from the RDC (which i believe this change would break)

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -3343,6 +3348,18 @@ function prepareAppPage(
           renderOpts.experimental.disableResumeDataCacheCompression
         ),
       }
+    } else if (isPossibleMPAAction) {
```
lubieowoce: i don't like that we're still doing this off of `isPossibleMPAAction` which may just be a random `POST`. the problem with `POSTs` is that we don't know if they're really server action requests until `handleAction` inspects them. i think we should only change the SSR behavior if `handleAction` decided that we are in fact handling this as a MPA server action, i.e. after the `handleAction` call, not ahead of time

PR #97789, opened 2026-08-24
test(turbo-tasks): run the wasm tests with a Node wasi host

sokra:
### What?

Adds a Node host so `wasm32-wasip1-threads` test binaries can run, and wires the `turbo-rcstr`, `turbo-tasks`, and `turbo-tasks-backend` library tests into the WASI CI job.

It also provides the WASI allocation symbols expected by the `lz4-sys` dependency introduced by the base PR, allowing those test binaries to link.

### Why?

The Rust test binaries require host functionality that stock WASI runners do not provide. Turbopack stores its link-time task registries in WebAssembly custom sections and imports a host callback to read them. Its multithreaded runtime also requires shared memory and thread spawning through Node Workers.

The base PR replaces `lzzzz` with `lz4`. `lz4-sys 1.11.1` compiles its C sources against custom allocation shims on `wasm32-wasi*`, but excludes the Rust implementations when the target OS is WASI. Without equivalent symbols supplied by the consuming crate, test links fail on unresolved malloc, calloc, and free shims.

Running these suites in CI gives the remaining WASI layers direct regression coverage instead of relying on native-only tests.

### How?

The Node cargo target runner provides WASI preview1, imported shared memory, and the complete two-phase [`link-section` custom-section protocol](https://github.com/mmastrac/linktime/blob/ae29e51d94a955df2442ed6418b8a712c1f2bfb3/link-section/docs/PREAMBLE.md#wasm).

[`@emnapi/wasi-threads`](https://github.com/toyobayashi/emnapi/tree/main/packages/wasi-threads) supplies Worker lifecycle, compiled-module transfer, load/start ordering, and cleanup for the [`wasi.thread-spawn`](https://github.com/WebAssembly/wasi-threads) import that Node's WASI runtime does not provide. The module is compiled once, then the structured-cloneable `WebAssembly.Module` and shared memory are sent to Workers. A small adapter around the package's lower-level `ThreadManager` and `ThreadMessageHandler` APIs preserves Rust's original one-argument ABI: the high-level `WASIThreads` wrapper requires guest `malloc`/`free` exports for that ABI, and Rust test binaries do not export them. Worker identifiers come from a shared atomic counter because Tokio Workers can spawn additional Workers while the main WebAssembly thread is parked.

On WASI, `turbo-persistence` supplies the three allocation symbols required by liblz4 and forwards them to WASI libc, keeping allocation and deallocation within the same allocator. Native builds do not compile these wrappers.

Tests that depend on unavailable platform behavior remain visible as WASI-only ignored tests with explicit reasons. Native runs keep the full test set enabled. A following stacked PR replaces the unsupported `parking_lot` parker and re-enables the corresponding contention tests.

The workflow installs and builds the workspace before running the host because both the package dependency and compiled production support used by a later stack layer must be available.

### Verification

- Node 20 host unit tests: 9/9
- Exact combined WASI CI command: `turbo-rcstr` 12 passed / 1 ignored; `turbo-tasks` 68 / 44; `turbo-tasks-backend` 65 / 35
- Full 111-test `turbo-tasks` binary through the package-managed host: 95 passed / 16 ignored
- No Worker-side `WebAssembly.compile`; the compiled module is transferred by `ThreadManager`
- `pnpm install --lockfile-only --offline` with a frozen lockfile


turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -2107,6 +2107,7 @@ mod cell_data_tracking_tests {
     // `evict_after_snapshot` uses `parallel::for_each`/`map_collect`, which call
     // `block_in_place` internally and require a multi-threaded Tokio runtime.
     #[tokio::test(flavor = "multi_thread")]
+    #[cfg_attr(target_family = "wasm", ignore = "parking_lot cannot block on wasm")]
```
sokra, line 2141: Instead of ignoring, fix or workaround the parking_lot problem

kirbx01:
Your lint / build check is failing because Prettier reports formatting issues in `scripts/wasi-test-host/spawn.mjs`, I think you should try running `pnpm prettier --write scripts/wasi-test-host/spawn.mjs` that should fix it, followed by `pnpm prettier --check scripts/wasi-test-host/spawn.mjs` to verify.

sokra:
Looks like some wasi unit tests are hanging. Increasing the timeout would not help.

.github/workflows/build_and_test.yml
```diff
@@ -638,8 +645,28 @@ jobs:
...
+
+        # Run the crates whose tests can execute on wasm. This needs a host that supplies the
+        # `env.read_custom_section` import turbo-tasks' link-time registries depend on; no stock
+        # runtime provides it. See scripts/wasi-test-host/README.md, which also lists the platform
+        # limitations behind the tests that are `ignore`d on wasm.
+        export CARGO_TARGET_WASM32_WASIP1_THREADS_RUNNER="node $PWD/scripts/wasi-test-host/run.mjs"
+        node --test scripts/wasi-test-host/lib.test.mjs
+        "${cargo_wasi[@]}" test -p turbo-rcstr --lib --target wasm32-wasip1-threads
+        "${cargo_wasi[@]}" test -p turbo-tasks --lib --target wasm32-wasip1-threads
+        "${cargo_wasi[@]}" test -p turbo-tasks-backend --lib --target wasm32-wasip1-threads
```
bgw: It should be possible to run these using a single invocation of `cargo test` or `cargo nextest`. Running 3 different invocations will reduce parallelism and result in a lot of wasted compilation.

scripts/wasi-test-host/lib.mjs
```diff
@@ -0,0 +1,132 @@
...
+ * Host support for running `wasm32-wasip1-threads` test binaries that link `turbo-tasks`.
+ *
+ * Such a binary cannot be run by a stock WASI runtime, because `turbo-tasks` gathers its task
+ * registries with the `link-section` crate, which on wasm stores them in **custom sections** and
+ * requires the embedder to hand them back through an `env.read_custom_section` import. See
+ * `link-section`'s `docs/PREAMBLE.md`; the implementation below is that contract.
+ *
+ * Everything here is deliberately dependency-free (`node:wasi` + `node:worker_threads`), because the
+ * CI job that uses it runs with `skipInstallBuild`, so `node_modules` does not exist there.
+ */
```
bgw, line 12: We'll probably need to pull this support code out of `wasi-test-host` at some point so that Next.js can use to to load the wasm blob. We should make sure that when we do that, we don't duplicate the support code in two places.

scripts/wasi-test-host/spawn.mjs
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Build the `wasi.thread-spawn` implementation for one instance.
+ *
+ * Returns the new thread id, or a negative value if the thread could not be created, as the
+ * wasi-threads proposal requires.
+ *
+ * @param {{ bytes: Uint8Array, memory: WebAssembly.Memory, threadIds: Int32Array, args: string[], cwd: string, onError: (error: Error, threadId: number) => void }} context
+ */
```
bgw, line 56: Add a link to the spec: https://github.com/WebAssembly/wasi-threads

Add a note that while v8 supports the wasm threads, Node.js's WASI implementation doesn't implement the corresponding wasi-threads proposal.

scripts/wasi-test-host/lib.mjs
```diff
@@ -0,0 +1,132 @@
...
+ * - `targetLength` is too small -> return the size needed, copy nothing;
+ * - otherwise -> copy the section into guest memory and return how many bytes were written.
+ *
+ * Returning 0 unconditionally would let a module instantiate with **empty** registries, which fails
+ * later in confusing ways rather than at the point of the mistake, so the real sections are read.
+ *
+ * @param {WebAssembly.Module} module
+ * @param {WebAssembly.Memory} memory
+ * @returns {(namePtr: number, nameLength: number, targetPtr: number, targetLength: number) => number}
+ */
```
bgw, line 117: Add a link to https://github.com/mmastrac/linktime/blob/ae29e51d94a955df2442ed6418b8a712c1f2bfb3/link-section/docs/PREAMBLE.md#wasm

scripts/wasi-test-host/thread.mjs
```diff
@@ -0,0 +1,70 @@
...
+import { WASI } from 'node:wasi'
+import { workerData } from 'node:worker_threads'
+import process from 'node:process'
+
+import { createReadCustomSection } from './lib.mjs'
+import { createThreadSpawn } from './spawn.mjs'
+
+const { bytes, memory, threadIds, threadId, startArg, args, cwd } = workerData
+
+const module = await WebAssembly.compile(bytes)
```
bgw: Does this mean that we're recompiling all of the wasm version of turbopack on every thread creation???

scripts/wasi-test-host/lib.mjs
```diff
@@ -0,0 +1,132 @@
...
+
+/** Parse the limits of an imported memory out of a wasm binary.
+ *
+ * `WebAssembly.Module.imports()` reports that a memory is imported but not its limits, and an
+ * imported memory has to be created with matching `initial` / `maximum` / `shared`, so the binary
+ * has to be read directly.
+ *
+ * @param {Uint8Array} bytes
+ * @returns {{ module: string, name: string, initial: number, maximum: number | undefined, shared: boolean }}
+ */
```
bgw, line 24: This seems so insane to me. I still don't understand why we have to do this. Is there any sort of documentation or example of other people doing this that we can point to?

scripts/wasi-test-host/lib.mjs
```diff
@@ -0,0 +1,132 @@
+/**
+ * Host support for running `wasm32-wasip1-threads` test binaries that link `turbo-tasks`.
+ *
+ * Such a binary cannot be run by a stock WASI runtime, because `turbo-tasks` gathers its task
+ * registries with the `link-section` crate, which on wasm stores them in **custom sections** and
+ * requires the embedder to hand them back through an `env.read_custom_section` import. See
+ * `link-section`'s `docs/PREAMBLE.md`; the implementation below is that contract.
+ *
+ * Everything here is deliberately dependency-free (`node:wasi` + `node:worker_threads`), because the
+ * CI job that uses it runs with `skipInstallBuild`, so `node_modules` does not exist there.
```
bgw: Instead of using `skipInstallBuild`, I think it would be way better to just pull in https://github.com/toyobayashi/emnapi/tree/main/packages/wasi-threads

Saving the `pnpm install` time doesn't seem worth it for this.

bgw (changes requested):

.github/workflows/build_and_test.yml
```diff
@@ -621,6 +621,13 @@ jobs:
       stepName: 'test-next-swc-wasm'
     secrets: inherit
 
+  # Guards that Turbopack keeps compiling for wasm, so the wasm fallback Next.js loads when the
+  # native bindings are unavailable can eventually support Turbopack and not just SWC.
+  # See https://github.com/vercel/next.js/issues/96960.
+  #
+  # This compiles the bindings and runs the wasm test suites. It stops short of a full `napi build`:
+  # linking additionally requires emnapi v2 (for the `emnapi_create_env` / `emnapi_delete_env`
+  # exports), which is still unreleased — so linking cannot be gated on yet.
```
bgw, line 630: We could use the alpha release?

.github/workflows/build_and_test.yml
```diff
@@ -632,14 +639,36 @@ jobs:
...
         set -euo pipefail
 
+        # Provisions the wasi-sdk, emnapi and the cross-compilation env. Shared with local
+        # builds so the two cannot drift; see contributing/core/building-wasm.md.
         source scripts/setup-wasi-env.sh
-        cargo check -p next-napi-bindings --target wasm32-wasip1-threads
+
+        # A distinct always-true target table merges with the repository's existing `cfg(true)`
+        # rustflags (notably `tokio_unstable`), while making warnings fatal for every command here.
+        cargo_wasi=(cargo --config 'target."cfg(all())".rustflags=["-Dwarnings"]')
```
bgw, line 653: - `cfg(all())` is a workaround for the lack of `cfg(true)` in older versions of rust. As this comment notes `.cargo/config.toml` already uses `cfg(true)`, so there's not reason we can't use it here too.
- Why not put this logic in `.cargo/config.toml`? And why not enable `-Dwarnings` for every target?

scripts/wasi-test-host/lib.mjs
bgw: I think a much better solution would be to pass explicit values to the linker via `.cargo/config.toml`, and then just hardcode them here, commenting that the two numbers must match.

https://lld.llvm.org/WebAssembly.html#cmdoption-initial-memory

> Initial size of the linear memory. Default: the sum of stack, static data and heap sizes.

I get that this default is hard to predict, but we just need to pick an initial value that's large enough. Turbopack consumes so much dynamically allocated memory that oversizing the initial memory by 2x isn't a real problem: We'd end up using that eventually anyways.

We can just set the initial memory to 512MB or 1GB, and the maximum to unlimited (already the default in lld) and move on with our lives.

bgw (changes requested):

.cargo/config.toml
```diff
@@ -83,6 +85,15 @@ rustflags = [
 [target.'cfg(all(target_os = "linux", target_env = "musl"))']
 rustflags = ["-Ctarget-feature=-crt-static"]
 
+# WASI threads: make the imported shared-memory contract explicit. The page counts hardcoded in
+# scripts/wasi-test-host and packages/next/src/build/swc/wasi-runtime.ts must match these byte values.
+[target.wasm32-wasip1-threads]
+rustflags = [
+  "-Dlinker-messages",
+  "-Clink-arg=--initial-memory=536870912",
```
bgw, line 94:
```suggestion
  # 512 MiB
  "-Clink-arg=--initial-memory=536870912",
```

.cargo/config.toml
```diff
@@ -83,6 +85,15 @@ rustflags = [
...
+  "-Clink-arg=--max-memory=4294967296",
```
bgw, line 96:
```suggestion
  # 4 GiB
  "-Clink-arg=--max-memory=4294967296",
```

scripts/wasi-test-host/lib.mjs
```diff
@@ -0,0 +1,64 @@
...
+ * [`link-section` wasm preamble](https://github.com/mmastrac/linktime/blob/ae29e51d94a955df2442ed6418b8a712c1f2bfb3/link-section/docs/PREAMBLE.md#wasm);
+ * the implementation below is that contract.
+ *
+ * Worker lifecycle and compiled-module transfer come from `@emnapi/wasi-threads`; this file contains
+ * only the Turbopack-specific custom-section hook and shared-memory contract.
+ */
+
+// These page counts must match --initial-memory / --max-memory in .cargo/config.toml.
+export const WASI_MEMORY_INITIAL_PAGES = 8_192
+export const WASI_MEMORY_MAXIMUM_PAGES = 65_536
```
bgw: Write this as bytes divided by page size? It's confusing because the numbers don't actually match `.cargo/config.toml`.

Also, include 512 MiB / 4 GiB comments like in the `.cargo/config.toml` file.

scripts/wasi-test-host/lib.test.mjs
```diff
@@ -0,0 +1,268 @@
+/**
+ * Tests for the host support functions.
+ *
+ * `read_custom_section`'s not-found and buffer-too-small branches are invisible in a passing test
+ * run — a host that always returned 0 would still let modules instantiate, just with empty
+ * registries — so they are covered directly here.
+ *
+ * Run with: node --test scripts/wasi-test-host/
+ */
```
bgw: @sokra can you review these tests and figure out if any of them are actually useful? This is a lot of test code...

bgw (approved):

scripts/wasi-test-host/lib.test.mjs
sokra: I think we can drop the whole file. I don't think we need tests for a test runner that runs the tests...

scripts/wasi-test-host/lib.mjs
sokra: Also include comments on both sides that these numbers need to match up

PR #97797, opened 2026-08-24
turbo-persistence: assorted read-path cleanups

lukesandberg:
### What?

Three small, localized optimizations to the read path. No behaviour or format changes.

**Avoid contended atomic refcounts on the mmap** 
   - before we would inc/dec the mmap for the index block, the key bloc and then optionally the inline value of the key block
   - Now we front-load the index block 'parsing', borrow the key block (when it doesn't need decompression), and return inline values from key blocks by _value_ instead of reference
   

**Hoist block cache access out of loops.** 
  - This is most impactful on misses, where we were touching the OnceLocks for every MetaFile

**Make scanned data dense.** 
  - Reorganize `MetaEntry` to separate the hash ranges from the rest of the data
  - This improves the cache effectiveness of our hot 'entry scanning' loops since the stride is now 16 bytes instead of 144.

### Performance

Interleaved A/B of this branch against its base (`LARGE_DB=1`, same machine, alternating runs, quiet host). Criterion medians:

| benchmark | base | this branch | delta | rounds won |
| --- | ---: | ---: | ---: | --- |
| `get` miss, 20 commits uncompacted | 600 ns | 568 ns | **−5.4%** | 7/8 |
| `get` hit, 20 commits uncompacted | 1.415 µs | 1.398 µs | −1.2% | 4/4 |
| `get` hit, compacted | 996 ns | 979 ns | −1.7% | 2/4 |
| `get` miss, compacted | 125.0 ns | 124.5 ns | −0.4% | 2/4 |

Note: as single threaded benchmarks we are not really detecting any benefits from the refcount work, under concurrent reads we should expect an impact


turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -2156,14 +2164,17 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
         }
         cells.sort_by_key(|(hash, _, _)| *hash);
         let inner = self.inner.read();
+        // Resolved once rather than per meta file; see the note in `get_impl`.
```
lukesandberg, line 2167: drop comment

turbopack/crates/turbo-persistence/src/meta_file.rs
```diff
@@ -244,6 +239,13 @@ pub struct MetaFile {
...
     family: u32,
+    /// The hash range of each entry, parallel to `entries` and indexed the same way.
+    ///
+    /// Stored apart from [`MetaEntry`] because a lookup that misses reads nothing else: it walks
+    /// every entry comparing the queried hash against these bounds, and only touches the entry
+    /// itself once a range matches. Inline, that scan strided over a 144-byte `MetaEntry` to read
+    /// 16 useful bytes, so 100 entries spanned 225 cache lines instead of 25.
+    hash_ranges: Vec<HashRange>,
     /// The entries of the file. Dropped before `mmap` (field declaration order).
     entries: Vec<MetaEntry>,
```
lukesandberg, line 253: should these be `Vec` or `Box<[T]>`

turbopack/crates/turbo-persistence/src/meta_file.rs
```diff
@@ -395,6 +397,26 @@ impl MetaFile {
...
+    /// The hash range of the entry at `index`.
+    pub fn hash_range(&self, index: u32) -> HashRange {
+        self.hash_ranges[index as usize]
+    }
+
+    /// The key family and hash range of the entry at `index`.
+    pub fn range(&self, index: u32) -> StaticSortedFileRange {
+        let HashRange { min_hash, max_hash } = self.hash_range(index);
+        StaticSortedFileRange {
+            family: self.family,
```
lukesandberg, line 417: do our callers really need this?

turbopack/crates/turbo-persistence/src/db.rs
```diff
@@ -2022,13 +2023,20 @@ impl<S: ParallelScheduler, const FAMILIES: usize> TurboPersistence<S, FAMILIES>
 
         let mut size = 0;
 
+        // Resolved once rather than per meta file: these are `OnceLock::get_or_init`, so calling
+        // them inside the loop repeats an acquire-load and a branch for every meta file scanned.
+        // A lookup that misses walks all of them, which cost ~127ns per lookup at 100 meta files.
+        // They stay lazily initialized because opening a database that is never read should not
+        // allocate the caches at all.
```
bgw: drop comment

turbopack/crates/turbo-persistence/src/meta_file.rs
```diff
@@ -244,6 +242,13 @@ pub struct MetaFile {
     sequence_number: u32,
     /// The key family of the SST files in this meta file.
     family: u32,
+    /// The hash range of each entry, parallel to `entries` and indexed the same way.
+    ///
+    /// Stored apart from [`MetaEntry`] because a lookup that misses reads nothing else: it walks
+    /// every entry comparing the queried hash against these bounds, and only touches the entry
+    /// itself once a range matches. Inline, that scan strided over a 144-byte `MetaEntry` to read
+    /// 16 useful bytes, so 100 entries spanned 225 cache lines instead of 25.
```
bgw:
```suggestion
    /// Stored separately from [`MetaEntry`] so that lookups can operate over a denser data structure that's hotter in cache.
```

bgw (approved):

PR #97799, opened 2026-08-24
feat(turbopack): resolve `/`-rooted imports from the project directory

sokra:
### What?

A request starting with `/` — `import '/content/where'`, `require('/foo.js')` — now resolves from
the **project directory** (the one holding `next.config`), and cannot reach outside it. Previously
it resolved from the root of the filesystem while reporting *"server relative imports are not
implemented yet"*.

This is the base of a stack: the follow-up on top of it moves `import.meta.glob`'s `/`-rooted
patterns onto the same root, so a pattern and a plain import agree on what `/` means.

### Why?

The feature was half-present and mislabelled. The `Request::ServerRelative` arm rewrote `/foo` to
`./foo` and resolved it against the filesystem root, then emitted a "not implemented yet" issue
unless an import-map alias had already matched. Users got a diagnostic telling them to give up on
something that in fact resolved.

Worse, resolving from the *filesystem* root means `/` escapes the project. In a Turbopack execution
test — where the filesystem root is the repository root — `require('/package.json')` resolved this
repository's own `package.json`. In a monorepo, where Turbopack's root is the workspace root and the
app lives in `apps/web`, `/content/x.js` would reach a workspace-level file rather than the app's.

The intended semantics are Vite's, and they were confirmed by measurement rather than from memory.
In a workspace where `content/*.js` exists *both* at the workspace root and under the Vite root, so
the two are distinguishable, Vite 5.4.21:

- resolves `/content/…` from the configured `root` — changing `root` changes which file wins;
- fails outright for a file that exists only above `root`, rather than walking up;
- behaves identically for a plain `import` and for `import.meta.glob`.

### How?

`ResolveOptions` gains `server_relative_root`, the directory a `/`-rooted request resolves from, and
`ResolveOptionsContext` exposes it so embedders can set it. Next.js sets it to the project directory
in the client, server and edge resolve contexts — all of which already had it to hand, so nothing new
is threaded through.

Three deliberate choices:

- **The option is optional and defaults to the previous behaviour.** `turbopack-core` is used well
  beyond Next.js; an embedder that doesn't set it keeps resolving `/` from the filesystem root. This
  is also why the change is provably contained — the whole `turbopack-tests` suite passes with no
  snapshot churn.
- **It is named and documented in filesystem terms**, tied to the existing `Request::ServerRelative`
  vocabulary, rather than introducing a "project root" concept into `turbopack-core`, which must stay
  free of Next.js concepts.
- **There is no fallback.** Resolution happens in the configured root and stops; a request that isn't
  there is a normal module-not-found. A fallback to the filesystem root would reintroduce exactly the
  escape this fixes.

The "not implemented yet" issue is removed. The equivalent issue for Windows-style absolute requests
is untouched, as is the separate tsconfig `extends` handling, which treats a rooted path as absolute
per TypeScript's own rule.

### Testing

The Turbopack execution harness already has two distinct roots — the repository root as the
filesystem root, and each test's own directory — so pointing the new option at the test directory
makes "project directory" and "filesystem root" genuinely different, and the distinction testable
without a Next.js app. Two fixtures: one where a `/`-rooted import and require resolve from the
project directory, and one asserting that `/package.json`, which exists only at the filesystem root,
does *not* resolve. The former emits no issues at all, which is what pins the removal of the
diagnostic.

The end-to-end case uses the existing monorepo fixture, where the project directory really is not the
filesystem root. A file of the same name exists in both `apps/web/content` and the workspace root, so
the rendered value alone identifies which root was used — a test that merely asserted "it renders"
would have passed under the old behaviour too.

### Known gap

TypeScript resolves a leading `/` as an absolute path on disk, so a `/`-rooted import in a `.ts` file
still fails type checking (`TS2307`) even though it bundles and runs. The e2e page is therefore
JavaScript. A tsconfig `paths` mapping works around it for users, but it also makes the request match
the import map instead, so the two resolvers overlap; that deserves its own decision and is tracked
separately rather than papered over here.


turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2021,34 +2023,19 @@ async fn resolve_internal_inline(
                 query,
                 fragment,
             } => {
+                // A `/`-rooted request is resolved from `server_relative_root` when one is
+                // configured, and from the root of the filesystem otherwise. It is not resolved
+                // relative to the importing file, and it does not fall back to a wider directory,
+                // so it can't reach outside of the configured root.
+                let root = match &options_value.server_relative_root {
+                    Some(root) => root.clone(),
+                    None => lookup_path.root().owned().await?,
```
sokra: When server_relative_root isn't set it should result in ResolvingIssue as before

sokra:
How does this change affect server relative paths in CSS? e.g. `url(/favicon.ico)`

turbopack/crates/turbopack-tests/tests/execution/turbopack/resolving/server-relative-import-unsupported/input/index.js
```diff
@@ -0,0 +1,9 @@
+// Without a `server_relative_root` there is nothing to resolve a `/`-rooted
+// request from, so it keeps reporting that it isn't supported and resolves from
+// the root of the filesystem. The value of this test is the `issues/` snapshot.
```
mischnic: It reports an error AND resolve from the root? Shouldn't it do one of them, not both?

mischnic (approved):

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -2030,35 +2030,40 @@ async fn resolve_internal_inline(
...
+                                .to_string(),
+                        ),
+                        source: None,
                     }
-                };
+                    .resolved_cell()
+                    .emit();
 
-                Box::pin(resolve_internal_inline(root, relative, options)).await?
+                    ResolveResult::unresolvable().cell()
```
mischnic, line 2057: Why is there any special handling for `has_alias` at all here? The other branches don't all have that either

PR #97804, opened 2026-08-24
[backport] Fix stale metadata across dynamic route params

DavidIlie:
## Summary

Fixes a 16.3 Cache Components bug where navigating between parameter values could render the new page body while leaving the previous page's title in the document. A hard reload corrected the title because the server response itself was already correct.

The 16.3 client trusts incomplete runtime metadata vary information and can store a completed parameter-dependent head under the route's generic fallback key. This change keeps PPR runtime head entries keyed to their concrete route parameters; App Shell and body reuse are unchanged.

Canary was fixed as part of the larger response transport refactor in #96439. This is the narrow release-compatible fix. Focused canary coverage is in #97803.

## Verification

- Baseline regression reproduced with Turbopack and webpack
- `pnpm build-all`
- `HEADLESS=true pnpm test-start-turbo test/e2e/app-dir/metadata-soft-nav-cache-components/metadata-soft-nav-cache-components.test.ts`
- `HEADLESS=true pnpm test-start-webpack test/e2e/app-dir/metadata-soft-nav-cache-components/metadata-soft-nav-cache-components.test.ts`

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -3738,6 +3738,14 @@ function fulfillEntrySpawnedByRuntimePrefetch(
       fulfilledVaryPath = tree.shellVaryPath
     } else if (
       fetchStrategy !== FetchStrategy.Full &&
+      // The runtime-stage metadata response does not reliably include params
+      // that were accessed while resolving metadata. Keep it keyed to the
+      // concrete URL until that tracking is available, rather than allowing
+      // one dynamic param's complete head to be reused by another.
+      !(
+        fetchStrategy === FetchStrategy.PPRRuntime &&
+        tree.requestKey === HEAD_REQUEST_KEY
+      ) &&
       segmentVaryParams !== null
     ) {
       fulfilledVaryPath = getFulfilledSegmentVaryPath(
--- a/test/e2e/app-dir/metadata-soft-nav-cache-components/app/link-accordion.tsx
+++ b/test/e2e/app-dir/metadata-soft-nav-cache-components/app/link-accordion.tsx
@@ -3,6 +3,8 @@
 import Link, { type LinkProps } from 'next/link'
 import { useState } from 'react'
 
+const productHandles = ['alpha', 'gamma', 'delta', 'epsilon', 'zeta', 'eta']
+
 export function LinkAccordion({
   href,
   children,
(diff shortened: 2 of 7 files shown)


GeronimoSerial:
We hit this in production on 16.3.0 and independently traced it to the same place this PR describes, so here is a data point in case it helps prioritise the backport.

**Symptom**: navigating client-side between two values of the same dynamic segment (`/attorneys/[slug]`), the page body is correct but `document.title`, `og:title` and `<link rel="canonical">` all belong to the *previously visited* param. It is exactly one navigation stale, deterministic, and only a hard reload clears it. `next dev` does not reproduce it; production builds do.

One detail we did not see mentioned: for us `<link rel="canonical">` sometimes gets **dropped entirely** rather than going stale, which makes this an SEO problem rather than a cosmetic title problem.

**Config**: `cacheComponents: true`, `partialPrefetching: true`, `export const instant = false` on the affected route, `generateMetadata` awaiting `params` and then an async fetch.

**What we verified**
- The server is correct at every layer. SSR HTML is right, and replaying the navigation RSC request with the browser's exact `Next-Router-State-Tree` / `Next-Url` headers returns the correct canonical for each slug. The stale value is introduced entirely on the client.
- We independently found the mechanism in `client/components/segment-cache/vary-path.js` before finding this PR: the metadata segment (`HEAD_REQUEST_KEY`) ends up keyed with the non-root param collapsed to `Fallback`, so every slug shares one metadata entry and a read-before-write produces the off-by-one. That matches this PR's "generic fallback key" description.
- Minimal reproduction: a bare `create-next-app` with a single `[slug]` route, `instant = false`, and a `generateMetadata` that only awaits `params` reproduces it with no data layer at all. Adding `"use cache"`, and `"use cache"` wrapped in React's `cache()`, reproduces identically.
- `16.4.0-canary.35` fixes it in that same repro with no code changes, consistent with #96439.
- No published 16.3.x release fixes it. We tested 16.3.0 through 16.3.5.

**Workarounds we tested that do not work**: `experimental.varyParams: true` (already defaults to true), `experimental.optimisticRouting: false`, and `prefetch={false}` on the links. The only app-level lever that fixes it is `partialPrefetching: false`.

Is there anything blocking this backport, or a target release for the 16.3.x line? Moving to canary is not an option for us, and `partialPrefetching: false` is the only thing standing between us and incorrect canonical tags in the meantime. Happy to test a candidate build against our app if that is useful.

PR #97805, opened 2026-08-24
docs: explain origin matching for allowedOrigins and allowedDevOrigins

icyJoseph:
The `allowedDevOrigins`, and `allowedOrigins` options, behave slightly different to each other, the former drops the PORT, and it is not clear where are these picked up from. This PR improves guidance on how to use these two.

### DIFF
--- a/docs/01-app/03-api-reference/05-config/01-next-config-js/allowedDevOrigins.mdx
+++ b/docs/01-app/03-api-reference/05-config/01-next-config-js/allowedDevOrigins.mdx
@@ -16,3 +16,25 @@ module.exports = {
   allowedDevOrigins: ['local-origin.dev', '*.local-origin.dev'],
 }
 ```
+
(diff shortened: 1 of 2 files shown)


docs/01-app/03-api-reference/05-config/01-next-config-js/allowedDevOrigins.mdx
```diff
@@ -16,3 +16,25 @@ module.exports = {
   allowedDevOrigins: ['local-origin.dev', '*.local-origin.dev'],
 }
 ```
+
+Entries are matched against the request's `Origin` header, and only one part of it is used: the [`hostname`](https://developer.mozilla.org/en-US/docs/Web/API/URL/hostname). For a request from `http://local-origin.dev:3000/dashboard?tab=1`, only `local-origin.dev` is compared against your configuration, leaving out the scheme, the port, the path, and the query string. Write your entries that way too, without `https://` and without a port.
```
aurorascharff: er, and only one part of it is used: 

ai language

aurorascharff (approved):

PR #97808, opened 2026-08-24
Upgrade Turbopack to hashbrown 0.15

lukesandberg:
### What?

Upgrade Turbopack's workspace dependencies to DashMap 7.0.0-rc2 and Hashbrown 0.15.4, removing Hashbrown 0.14 from the `turbo-tasks-backend` dependency graph.

### Why?

`turbo-tasks-backend` previously compiled two Hashbrown versions because its custom DashMap helpers shared Hashbrown 0.14's private raw table types. Converging the backend graph on Hashbrown 0.15 removes that duplicate backend codegen path and slightly reduces the shipped native addon's size.

A clean release A/B of the unstripped x86-64 `next-napi-bindings` cdylib measured:

| Revision | `libnext_napi_bindings.so` |
| --- | ---: |
| canary base `80576037` | 147,824,920 bytes |
| PR | 147,818,576 bytes |
| **Delta** | **−6,344 bytes (−0.0043%)** |

The final gain is small because unrelated SWC dependencies still pull DashMap 6 / Hashbrown 0.14 into the same addon; this change removes only backend-specific Hashbrown 0.14 codegen.  Plus... nearly everything is monomorphized anyway

### How?

DashMap 7 replaces its Hashbrown 0.14 raw table with Hashbrown 0.15's public `HashTable` API. The backend helpers now use that public API while retaining their existing shard locking and allocation behavior:

- heterogeneous cache lookup and insertion still reuse a pre-located shard and precomputed hash;
- the entry helper passes Hashbrown's native `hash_table::Entry` into a closure, preserving precise lock-release timing without a custom self-referential entry type or cloned hasher;
- nonblocking removal still avoids lock-order deadlocks;
- two-key mutation uses a release-mode disjointness assertion followed by `get_many_unchecked_mut` to avoid a redundant duplicate-pointer scan;
- snapshot draining still transfers ownership of each shard table and releases task memory incrementally;
- custom mutable guards remain non-`Send`.

The migration also replaces several raw-pointer iteration and removal paths with safe `HashTable` operations. Existing raw-entry users in `auto-hash-map` and `turbopack-trace-server` explicitly enable Hashbrown 0.15's compatibility feature.

### Performance

Performance testing with the overhead.rs benchmark revealed no regression/progression

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,7 +405,7 @@ version = "0.1.0"
 dependencies = [
  "bincode 2.0.1",
  "codspeed-criterion-compat",
- "hashbrown 0.14.5",
+ "hashbrown 0.15.4",
  "rustc-hash 2.1.1",
  "serde",
  "shrink-to-fit",
@@ -2149,6 +2149,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "7.0.0-rc2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e4a1e35a65fe0538a60167f0ada6e195ad5d477f6ddae273943596d4a1a5730b"
+dependencies = [
+ "cfg-if",
+ "crossbeam-utils",
+ "equivalent",
+ "hashbrown 0.15.4",
+ "lock_api",
+ "parking_lot_core",
+]
+
+[[package]]
 name = "data-encoding"
 version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2554,7 +2568,7 @@ checksum = "d4029edd3e734da6fe05b6cd7bd2960760a616bd2ddd0d59a0124746d6272af0"
 dependencies = [
  "cfg-if",
  "libc",
- "redox_syscall",
+ "redox_syscall 0.3.5",
  "windows-sys 0.48.0",
 ]
 
@@ -3172,7 +3186,6 @@ checksum = "e5274423e17b7c9fc20b6e7e208532f9b19825d82dfd615708b70edd83df41f1"
 dependencies = [
  "ahash 0.8.12",
  "allocator-api2",
- "serde",
 ]
 
 [[package]]
@@ -4416,11 +4429,10 @@ checksum = "b4ce301924b7887e9d637144fdade93f9dfff9b60981d4ac161db09720d39aa5"
 
 [[package]]
 name = "lock_api"
-version = "0.4.10"
+version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1cc9717a20b1bb222f333e6a92fd32f7d8a18ddc5a3191a11af45dcbf4dcd16"
+checksum = "224399e74b87b5f3557511d98dff8b14089b3dadafcab6bb93eab67d3aace965"
 dependencies = [
- "autocfg",
  "scopeguard",
  "serde",
 ]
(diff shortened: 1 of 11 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -924,7 +906,7 @@ enum ShardWork {
     /// (modified-only) shard table out of the map. The iterator owns that table and drains it
     /// directly, freeing each task box as it is serialized. No second map lookup, no flag
     /// bookkeeping (the whole map is discarded right after this snapshot).
-    Drain(RawIntoIter<(TaskId, SharedValue<Box<TaskStorage>>)>),
+    Drain(IntoIter<(TaskId, Box<TaskStorage>)>),
```
lukesandberg: nit: qualify this as `hash_table::IntoIter`

turbopack/crates/turbo-tasks-backend/src/utils/dash_map_multi.rs
```diff
@@ -135,49 +133,45 @@ where
...
-
-        // this assertion is needed for memory safety reasons
+        // Equal keys would resolve to a single entry below. `get_many_mut` panics in that case
+        // (which is what keeps this memory-safe), but check up front so the caller gets an error
+        // message that names the actual problem.
         assert!(
-            !std::ptr::eq(bucket1.as_ptr(), bucket2.as_ptr()),
+            key1 != key2,
             "`get_multiple_mut` was called with equal keys, which breaks mutable referencing rules"
         );
```
lukesandberg, line 141: is this actually better than the default error message we get from get_many_mut?

turbopack/crates/turbo-tasks-backend/src/utils/dash_map_multi.rs
```diff
@@ -135,49 +133,45 @@ where
...
+            guard.insert_unique(h1, (key1.clone(), insert_with()), hash_entry);
+        }
+        if guard.find(h2, eq2).is_none() {
+            guard.insert_unique(h2, (key2.clone(), insert_with()), hash_entry);
+        }
+
+        // `get_many_mut` validates that the keys resolve to distinct entries before creating the
+        // mutable references.
+        let [entry1, entry2] =
+            guard.get_many_mut(
```
lukesandberg: isn't this API renamed to get_disjoint_mut?  use that spelling

turbopack/crates/turbo-tasks-backend/src/utils/dash_map_multi.rs
lukesandberg: i suppose it probably is, but since we are checking here.... shouldn't we use get_many_unchecked_mut below?  avoid redundant checks

turbopack/crates/turbo-tasks-backend/src/utils/dash_map_raw_entry.rs
```diff
@@ -47,20 +50,19 @@ pub fn raw_entry_in_shard<'l, K: Eq + Hash, V, S: BuildHasher + Clone>(
...
+            entry: NonNull::from(entry),
+            shard,
+        })
+    } else {
+        RawEntry::Vacant(VacantEntry {
             hash,
-            insert_slot,
-            shard: guard,
-        }),
+            map_hasher: map_hasher.clone(),
```
lukesandberg: is cloning this hasher expensive?

turbopack/crates/turbo-tasks-backend/src/utils/dash_map_raw_entry.rs
```diff
@@ -47,20 +50,19 @@ pub fn raw_entry_in_shard<'l, K: Eq + Hash, V, S: BuildHasher + Clone>(
...
-        guard.find_or_find_insert_slot(hash, |(k, _v)| eq(k), |(k, _v)| map_hasher.hash_one(k));
-    match result {
-        Ok(bucket) => RawEntry::Occupied(OccupiedEntry {
-            bucket,
-            shard: guard,
-        }),
-        Err(insert_slot) => RawEntry::Vacant(VacantEntry {
+) -> RawEntry<'l, K, V, S> {
+    let mut shard = shard.write();
+    if let Some(entry) = shard.find_mut(hash, |(k, _v)| eq(k)) {
```
lukesandberg: could we use the native entry api here? could our caller?

Cargo.lock
```diff
@@ -2148,6 +2148,20 @@ dependencies = [
  "parking_lot_core",
 ]
 
+[[package]]
+name = "dashmap"
```
lukesandberg, line 2152: how many versions of dashmap do we now depend on?

mischnic (approved):

PR #97825, opened 2026-08-24
Fix Turbopack resolution through chained symlinks

### LINKED ISSUE #97786: Turbopack cannot resolve a package through a symlink-to-symlink chain

https://github.com/Innei/turbopack-symlink-repro

### To Reproduce

```bash
pnpm install
cd apps/web
pnpm dev          # Turbopack -> 500, Can't resolve '@scope/ui/hello'
pnpm dev:webpack  # webpack   -> 200
pnpm build        # Turbopack -> Build error: Can't resolve '@scope/ui/hello'
```

The repro is a pnpm workspace in which one member directory is itself a symlink
to sources kept elsewhere in the repo (in the original project, a git submodule):

```
packages/ui -> ../vendor/packages/ui      # the workspace member is a symlink
vendor/packages/ui/                       # the real package, "@scope/ui"
apps/web/                                 # deps: "@scope/ui": "workspace:*"
```

`pnpm install` links the dependency to the *workspace* path, so the consumer's
`node_modules` entry is a symlink that points at another symlink:

```
apps/web/node_modules/@scope/ui -> ../../../../packages/ui   (symlink)
packages/ui                     -> ../vendor/packages/ui     (symlink)
vendor/packages/ui                                           (real directory)
```

Every real path is inside `turbopack.root` and `outputFileTracingRoot`, both
pinned to the repo root in `apps/web/next.config.mjs`.

**Current:** Turbopack cannot resolve the package, in dev and in build:

```
./apps/web/app/page.tsx:1:1
Error: Module not found: Can't resolve '@scope/ui/hello'
> 1 | import { Hello } from '@scope/ui/hello'
```

In the real project this hit every subpath of the package at once — explicit
`exports` entries and wildcard ones alike — so a single package took the whole
dev server down with ~20 distinct "Can't resolve" errors.

**Expected:** the resolution webpack already performs. Node agrees with webpack:

```bash
node -e "console.log(require.resolve('@scope/ui/hello', { paths: [process.cwd()] }))"
# /…/vendor/packages/ui/src/hello.tsx
```

### Isolation

Repointing the consumer link at the **same real directory** with a single hop
makes Turbopack resolve it:

```bash
cd apps/web
rm node_modules/@scope/ui
ln -s ../../../../vendor/packages/ui node_modules/@scope/ui
rm -rf .next
pnpm dev     # -> 200
pnpm build   # -> Compiled successfully
```

| consumer link | Turbopack dev | Turbopack build | webpack dev |
| --- | --- | --- | --- |
| chained (what pnpm creates) | 500, `Can't resolve` | fails, `Can't resolve` | 200 |
| flattened onto the real dir | 200 | Compiled successfully | 200 |

Same target directory, same `package.json`, same `exports`, same config — only
the shape of the link differs. Ruled out along the way, in the original project:
`transpilePackages` membership, the `"./src/*"` wildcard in `exports`, a nested
`pnpm-lock.yaml` in the submodule, a stale `.next`, and the case of the path the
dev server was started from.

```
Operating System:
  Platform: darwin 26.4.1
  Arch: arm64
Binaries:
  Node: 24.18.0
  pnpm: 11.20.0
Relevant Packages:
  next: 16.3.0
  react: 19.2.0
```

Turbopack

marcoshernanz:
## Summary

- classify symlinks by the final target of the full chain with `LinkTarget::resolved_type`
- preserve non-directory behavior for dangling, invalid, and cyclic chains and retain the existing ancestor-loop guard
- fold two-hop directory-symlink coverage into the existing production symbolic-links suite

Fixes #97786.

## Verification

- `CARGO_INCREMENTAL=0 pnpm build-all`
- `cargo test -p turbo-tasks-fs test_link_target_resolved_type_through_chain -- --nocapture`
- `cargo test -p turbo-tasks-fs read_glob_symlinks -- --nocapture`
- `cargo test -p turbopack-core test_read_matches -- --nocapture`
- `cargo clippy -p turbo-tasks-fs -p turbopack-core --all-targets -- -D warnings`
- `pnpm test-start-turbo test/production/app-dir/symbolic-links -t "chained directory symlinks"`
- manually verified `/directory-symlink-chain` under `next dev --turbopack`

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/content.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/content.rs
@@ -20,7 +20,7 @@ use turbo_tasks_hash::{
 };
 
 use crate::{
-    FileSystemEntryType, FileSystemPath,
+    FileSystemEntryType, FileSystemPath, RealPathErrorType,
     json::UnparsableJson,
     retry::retry_blocking,
     rope::{Rope, RopeReader},
@@ -199,6 +199,21 @@ impl LinkTarget {
     /// A dangling link returns [`FileSystemEntryType::NotFound`].
     pub async fn target_type(&self) -> Result<FileSystemEntryType> {
         Ok(*self.file_system_path().get_type().await?)
+    }
+
+    /// The type of the file this link ultimately points at.
+    ///
+    /// This follows a chain of links. A dangling link returns
+    /// [`FileSystemEntryType::NotFound`], while any other unresolvable link returns
+    /// [`FileSystemEntryType::Error`].
+    pub async fn resolved_type(&self) -> Result<FileSystemEntryType> {
+        match self.file_system_path().realpath().await? {
+            Ok(path) => Ok(*path.get_type().await?),
+            Err(error) => Ok(match error.kind() {
+                RealPathErrorType::NotFound => FileSystemEntryType::NotFound,
+                _ => FileSystemEntryType::Error,
+            }),
+        }
     }
 }
 
--- a/turbopack/crates/turbo-tasks-fs/src/disk.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/disk.rs
@@ -1958,6 +1958,67 @@ mod tests {
             tt.stop_and_wait().await;
         }
 
+        #[cfg(unix)]
+        #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+        async fn test_link_target_resolved_type_through_chain() {
+            use std::os::unix::fs::symlink;
+
+            let scratch = tempfile::tempdir().unwrap();
+            let path = scratch.path().to_owned();
+            create_dir_all(path.join("target-dir")).unwrap();
+            File::create_new(path.join("target-file")).unwrap();
+            symlink("target-dir", path.join("dir-inner")).unwrap();
+            symlink("dir-inner", path.join("dir-outer")).unwrap();
+            symlink("target-file", path.join("file-inner")).unwrap();
+            symlink("file-inner", path.join("file-outer")).unwrap();
+            symlink("../outside", path.join("invalid-inner")).unwrap();
+            symlink("invalid-inner", path.join("invalid-outer")).unwrap();
+
+            let root = canonicalize_to_rcstr(&path).unwrap();
+
+            #[turbo_tasks::function(operation, root)]
+            async fn assert_operation(
+                fs: ResolvedVc<DiskFileSystem>,
+                root_path: FileSystemPath,
+            ) -> anyhow::Result<()> {
+                for (input_path, expected_output) in [
+                    ("dir-outer", FileSystemEntryType::Directory),
+                    ("file-outer", FileSystemEntryType::File),
+                    ("invalid-outer", FileSystemEntryType::Error),
+                ] {
+                    let link = fs.read_link(root_path.join(input_path)?).await?;
+                    let LinkContent::Link { target } = &*link else {
+                        anyhow::bail!("expected a valid link, got {link:?}");
+                    };
+                    assert_eq!(target.resolved_type().await?, expected_output);
+                }
+
+                Ok(())
+            }
+
+            let tt = turbo_tasks::TurboTasks::new(TurboTasksBackend::new(
+                BackendOptions::default(),
+                noop_backing_storage(),
+            ));
+
+            tt.run_once(async move {
+                let fs = disk_file_system_operation(root)
+                    .resolve()
+                    .strongly_consistent()
+                    .await?;
+
+                assert_operation(fs, disk_file_system_root(fs))
+                    .read_strongly_consistent()
+                    .await?;
+
+                anyhow::Ok(())
+            })
+            .await
+            .unwrap();
+
+            tt.stop_and_wait().await;
+        }
+
         /// A relative target must stay inside the filesystem root at every step, not just at the
         /// end. Both of these step above the root; one comes back into it and one doesn't, but
         /// neither can be resolved against a root-relative [`FileSystemPath`], so `read_link`
--- a/turbopack/crates/turbo-tasks-fs/src/read_glob.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/read_glob.rs
@@ -89,12 +89,14 @@ async fn read_glob_internal(
                         // Skip links that leave the filesystem root.
                         let link_content = path.read_link().await?;
                         if let LinkContent::Link { target } = &*link_content {
-                            if matches!(
-                                target.resolved_type().await?,
-                                FileSystemEntryType::Directory
-                            ) {
+                            let Ok(realpath) = target.file_system_path().realpath().await? else {
+                                // same as if we got DirectoryEntry::Error in the main loop
+                                continue;
+                            };
+                            if matches!(*realpath.get_type().await?, FileSystemEntryType::Directory)
+                            {
                                 // Reject links that point to an ancestor before recursing.
-                                resolve_symlink_safely(entry.clone()).await?;
+                                check_symlink_directory_recursion(path, &realpath)?;
 
                                 // Add the directory to `results` if it is a whole match of the glob
                                 handle_file(&mut result, &entry_path, segment, entry);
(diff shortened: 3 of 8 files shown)


turbopack/crates/turbo-tasks-fs/src/content.rs
```diff
@@ -200,6 +200,20 @@ impl LinkTarget {
...
+    pub async fn points_to_directory(&self) -> Result<bool> {
+        let result = self.file_system_path().realpath_with_links().await?;
+        let Ok(path) = &result.path_result else {
+            return Ok(false);
+        };
+        Ok(matches!(
+            *path.get_type().await?,
+            FileSystemEntryType::Directory
+        ))
+    }
```
bgw, line 217: This API doesn't need to be directory-specific. We've got `target_type`, why not a `resolved_type` that returns `Result<FileSystemEntryType>`, mirroring the API, but with `realpath` in the middle.

turbopack/crates/turbo-tasks-fs/src/content.rs
```diff
@@ -200,6 +200,20 @@ impl LinkTarget {
     pub async fn target_type(&self) -> Result<FileSystemEntryType> {
         Ok(*self.file_system_path().get_type().await?)
     }
+
+    /// Whether this link ultimately points to a directory.
+    ///
+    /// This follows symlink chains. Invalid, dangling, and cyclic chains are not directories.
+    pub async fn points_to_directory(&self) -> Result<bool> {
+        let result = self.file_system_path().realpath_with_links().await?;
```
bgw: https://github.com/vercel/next.js/pull/97717 (just merged, you'll need to rebase) fixes the `.realpath()` API so that it returns better error types, so that you shouldn't need to use `realpath_with_links` here.

test/e2e/app-dir/symlink-to-symlink-chain/symlink-to-symlink-chain.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    await next.symlink('packages/ui', 'node_modules/@scope/ui')
+    await next.start()
+  })
+
+  it('resolves a package through both links', async () => {
+    const res = await next.fetch('/')
+    expect(res.status).toBe(200)
+    expect(await res.text()).toContain('hello from symlink chain')
+  })
+})
```
bgw: Let's just extend `test/production/app-dir/symbolic-links/symbolic-links.test.ts`.

We don't want huge test suites (limits parallelism), but we don't want a ton of tiny e2e test suites either because the isolated test setup is kinda expensive (a lot of disk IO).

test/e2e/app-dir/symlink-to-symlink-chain/symlink-to-symlink-chain.test.ts
```diff
@@ -0,0 +1,31 @@
+import { mkdir } from 'node:fs/promises'
+import { join } from 'node:path'
+import { nextTestSetup } from 'e2e-utils'
+
+describe('symlink-to-symlink chain', () => {
+  const { next, skipped } = nextTestSetup({
+    files: __dirname,
+    skipStart: true,
+    skipDeployment: true,
```
bgw: We don't really need to run this as a deployment test, but there's no reason this couldn't work as both a `next dev` and `next build`/`next start` test.

test/e2e/app-dir/symlink-to-symlink-chain/symlink-to-symlink-chain.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  beforeAll(async () => {
+    await mkdir(join(next.testDir, 'packages'), { recursive: true })
+    await mkdir(join(next.testDir, 'node_modules/@scope'), {
+      recursive: true,
+    })
+    await next.symlink('vendor/packages/ui', 'packages/ui')
+    await next.symlink('packages/ui', 'node_modules/@scope/ui')
+    await next.start()
```
bgw: This is running only as a dev test (`skipStart: true`), and then running `next.start()`?

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -1964,6 +1964,83 @@ mod tests {
...
+            let scratch = tempfile::tempdir().unwrap();
+            let path = scratch.path().to_owned();
+            create_dir_all(path.join("target-dir")).unwrap();
+            File::create_new(path.join("target-file")).unwrap();
+            symlink("target-dir", path.join("dir-inner")).unwrap();
+            symlink("dir-inner", path.join("dir-outer")).unwrap();
+            symlink("target-file", path.join("file-inner")).unwrap();
+            symlink("file-inner", path.join("file-outer")).unwrap();
+            symlink("cycle-b", path.join("cycle-a")).unwrap();
+            symlink("cycle-a", path.join("cycle-b")).unwrap();
```
bgw: This is excessive, we don't need to test cycle handling, we can assume `realpath` does this correctly.

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -1964,6 +1964,83 @@ mod tests {
...
+                assert!(!file_target.points_to_directory().await?);
+
+                let cycle = fs.read_link(root_path.join("cycle-a")?).await?;
+                let LinkContent::Link {
+                    target: cycle_target,
+                } = &*cycle
+                else {
+                    anyhow::bail!("expected a valid link, got {cycle:?}");
+                };
+                assert!(!cycle_target.points_to_directory().await?);
```
bgw: This could be a loop over an array of `(input_path, expected_output)` tuples.

turbopack/crates/turbo-tasks-fs/src/read_glob.rs
```diff
@@ -89,8 +88,7 @@ async fn read_glob_internal(
                         // Skip links that leave the filesystem root.
                         let link_content = path.read_link().await?;
                         if let LinkContent::Link { target } = &*link_content {
-                            if matches!(target.target_type().await?, FileSystemEntryType::Directory)
-                            {
+                            if target.points_to_directory().await? {
                                 // Ensure that there are no infinite link loops, but don't resolve
                                 resolve_symlink_safely(entry.clone()).await?;
```
bgw: We don't need `resolve_symlink_safely`? We should only need to call realpath once (as part of the directory check).

bgw (changes requested):

test/e2e/app-dir/symlink-to-symlink-chain/symlink-to-symlink-chain.test.ts
marcoshernanz: Removed the deployment-specific setup. The existing production suite covers build/start, and I also manually verified the chained route under `next dev --turbopack`.

test/e2e/app-dir/symlink-to-symlink-chain/symlink-to-symlink-chain.test.ts
marcoshernanz: Folded the regression coverage into `test/production/app-dir/symbolic-links` and removed the standalone e2e fixture.

turbopack/crates/turbo-tasks-fs/src/disk.rs
marcoshernanz: Removed the dedicated cycle fixture and assertion from the new `resolved_type` unit test.

turbopack/crates/turbo-tasks-fs/src/disk.rs
marcoshernanz: Collapsed the directory, file, and indirect-invalid cases into a single `(input_path, expected_output)` loop.

test/e2e/app-dir/symlink-to-symlink-chain/symlink-to-symlink-chain.test.ts
marcoshernanz: Removed the standalone `skipStart` and manual lifecycle setup entirely; the existing symbolic-links suite now owns the case through standard `nextTestSetup`.

turbopack/crates/turbo-tasks-fs/src/read_glob.rs
marcoshernanz: I initially removed this call, but the existing ancestor-loop test then fell through to OS `ELOOP`: realpath of an ancestor target succeeds even though recursive glob traversal would loop. I kept the one-line guard to prevent task fan-out; `resolved_type` still performs the single chain-resolution check used for directory classification.

turbopack/crates/turbo-tasks-fs/src/content.rs
marcoshernanz: Rebased onto current canary with #97717 included and switched the implementation to the improved `.realpath()` API.

turbopack/crates/turbo-tasks-fs/src/content.rs
marcoshernanz: Updated the API to `resolved_type() -> Result<FileSystemEntryType>`, so it is no longer directory-specific.

bgw (approved):

PR #97827, opened 2026-08-24
[backport] Record fallback cache lifetime for blocking PPR routes

acdlite:
Backports #97821 to `next-16-3`.

Note the original PR is still in draft on `canary`; this backport should land after (or together with) it.

A dynamic PPR route without a servable HTML fallback (`fallback: null`) previously recorded no fallbackRevalidate/fallbackExpire in the prerender manifest, so its per-segment prefetch outputs and RSC template shipped with no cache lifetime. Cache layers interpret a missing revalidate as a 1-second lifetime, causing nearly every prefetch of a non-prerendered URL to serve stale, revalidate in the background, and write a new cache entry. The manifest now records the export's collected cache control for every PPR route regardless of fallback mode, and the adapter's segment and RSC template outputs inherit it when there's no HTML fallback output to inherit from.

Cherry-pick of 8b6d0dcf8b, clean with no conflicts. The regression suite (`test/production/app-dir/adapter-prerender-metadata`) passes on this branch.

### DIFF
--- a/packages/next/src/build/adapter/build-complete.ts
+++ b/packages/next/src/build/adapter/build-complete.ts
@@ -1223,6 +1223,14 @@ export async function handleBuildComplete({
         ctx: {
           htmlAllowQuery?: string[]
           dataAllowQuery?: string[]
+          // The cache lifetime of the route's fallback artifacts, from the
+          // prerender manifest. Segment outputs attach this when the route
+          // has no HTML fallback output to inherit it from (a blocking
+          // route): the segment payloads are cached and served even then,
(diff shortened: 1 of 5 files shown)


acdlite:
Closing: #97821 caused a deploy regression (see #97835, which reverts it on canary) — the change stamped collected cache lifetimes onto fallback-file-seeded entries, whose short default lifetime turns out to drive the fallback-upgrade flow. This backport would have shipped the same regression to 16.3. A redesigned fix is staged on `acdlite/fallback-entry-lifetimes`; if it lands and a backport is still warranted, a fresh one should be cut from the final commit.

PR #97829, opened 2026-08-24
Stop printing a stack frame for error message text

unstubbable:
Next.js printed a stack frame that does not exist for an error whose message ends with text that looks like a file location. A logged connection failure read `at <unknown> (Error: connect ECONNREFUSED ::1:45999)` in front of its real frames.

`stacktrace-parser` tries five patterns on every line, and the pattern for JavaScriptCore makes the `<name>@` prefix optional. That pattern `:<line>:<column>`, and a line of an error message can end that way. A JavaScriptCore frame always holds `@`, also at the top level, so the prefix is not optional in a real stack.

The patch for `stacktrace-parser` now requires that prefix. A line of a message no longer becomes a frame, and every JavaScriptCore frame still parses, including one whose path holds a space.

A Turbopack build error puts the file and the position of the error in its message, and the parser read that line as a frame as well. That frame is gone now, and the location stays in the message. The snapshots of the build errors record the change.


packages/next/src/server/patch-error-inspect.ts
```diff
@@ -442,6 +480,8 @@ function parseAndSourceMap(
   // doesn't implement the name computation correctly.
   const errorName = computeErrorName(error)
 
+  unparsedStack = removeErrorHeader(unparsedStack, errorName, error.message)
```
eps1lon: Removing the "header" does not mean you have just the stack. The error message can contain newlines.

packages/next/src/server/patch-error-inspect.ts
```diff
@@ -430,6 +430,44 @@ function getSourcemappedFrameIfPossible(
...
+ * error code to the name of some system errors while it captures the stack, for
+ * example `AggregateError [ECONNREFUSED]`. Node.js then restores the name. The
+ * first line no longer matches, and this function returns the stack unchanged.
+ */
+function removeErrorHeader(
+  stack: string,
+  errorName: string,
+  message: string
+): string {
+  if (!stack.startsWith(errorName + ': ' + message)) {
```
eps1lon: careful here since `errorName` computation is not 100% accurate iirc.

patches/stacktrace-parser@0.1.10.patch
```diff
@@ -29,3 +29,12 @@ index 50c63025c622ebe479e07544bd796c2d3922d4b8..d704100cc59b6c47db5ad67ecd2662c9
  var geckoEvalRe = /(\S+) line (\d+)(?: > eval line \d+)* > eval/i;
  
  function parseGecko(line) {
+@@ -101,7 +101,7 @@ function parseGecko(line) {
+   };
+ }
+ 
+-var javaScriptCoreRe = /^\s*(?:([^@]*)(?:\((.*?)\))?@)?(\S.*?):(\d+)(?::(\d+))?\s*$/i;
++var javaScriptCoreRe = /^\s*([^@]*)(?:\((.*?)\))?@(\S.*?):(\d+)(?::(\d+))?\s*$/i;
```
eps1lon, line 37: Let's file this against upstream as well since we don't have any JSC coverage.

eps1lon (approved):

PR #78936, opened 2025-05-07
Add nonce handling from CSP in pages router

ijjk:
Our docs never called out that the nonce from a `content-security-policy` header would only automatically be set in app router. This adds handling to ensure it is also set for pages router. 

Fixes #67042
Closes NEXT-4104

Issue #67042, opened 2024-06-19
CSP with nonce does not work for page router


noahsmartin:
### Link to the code that reproduces this issue

https://github.com/noahsmartin/NextJSCSP

### To Reproduce

When using the pages router, adding a nonce to the CSP results in the page not loading necessary scripts. With the app router this works fine. With the pages router following the example code in https://nextjs.org/docs/pages/building-your-application/configuring/content-security-policy causes the <script /> tags that next includes on the root page to not have a nonce and not get loaded. 

1. start the app (npm run dev)
2. Observe console logs in the browser

### Current vs. Expected behavior

I expected no errors from failing to load scripts, but multiple scripts errored. You can see the scripts tag do not have a nonce:

```
    <script src="/_next/static/chunks/webpack.js?ts=1718827141851" defer=""></script>
    <script src="/_next/static/chunks/main.js?ts=1718827141851" defer=""></script>
    <script src="/_next/static/chunks/pages/_app.js?ts=1718827141851" defer=""></script>
    <script src="/_next/static/chunks/pages/index.js?ts=1718827141851" defer=""></script>
    <script src="/_next/static/development/_buildManifest.js?ts=1718827141851" defer=""></script>
    <script src="/_next/static/development/_ssgManifest.js?ts=1718827141851" defer=""></script>
```

The errors look like this 

```
[Error] Refused to load http://localhost:3000/_next/static/chunks/webpack.js?ts=1718827141851 because it does not appear in the script-src directive of the Content Security Policy.
[Error] Refused to load http://localhost:3000/_next/static/chunks/main.js?ts=1718827141851 because it does not appear in the script-src directive of the Content Security Policy.
[Error] Refused to load http://localhost:3000/_next/static/chunks/pages/_app.js?ts=1718827141851 because it does not appear in the script-src directive of the Content Security Policy.
[Error] Refused to load http://localhost:3000/_next/static/chunks/pages/index.js?ts=1718827141851 because it does not appear in the script-src directive of the Content Security Policy.
[Error] Refused to load http://localhost:3000/_next/static/development/_buildManifest.js?ts=1718827141851 because it does not appear in the script-src directive of the Content Security Policy.
[Error] Refused to load http://localhost:3000/_next/static/development/_ssgManifest.js?ts=1718827141851 because it does not appear in the script-src directive of the Content Security Policy.
[...]

PR #97851, opened 2026-08-25
fix: apply CSP nonce to Pages Router streamed Suspense scripts

### LINKED ISSUE #96443: Content-Security-Policy nonce is not applied to streamed Suspense boundary scripts in Pages Router SSR

https://github.com/agilmore/nextjs-pages-nonce-reproduction

### To Reproduce

1. Start the application in production mode (npm run start)
2. Visit http://localhost:3000/ and view source
3. You'll see script tags relating to Suspense boundary streaming which do not have the nonce attribute which all other script tags in the output do

CSP nonce attributes are missing from React streamed suspense boundary scripts, the CSP nonce value should be applied to these scripts as well.

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #136-Ubuntu SMP PREEMPT_DYNAMIC Wed Jul  1 21:53:05 UTC 2026
  Available memory (MB): 31779
  Available CPU cores: 16
Binaries:
  Node: 24.18.1
  npm: 11.16.0
  Yarn: 1.22.22
  pnpm: 11.18.0
Relevant Packages:
  next: 16.2.12 // Latest available version is detected (16.2.12).
  eslint-config-next: N/A
  react: 19.2.4
  react-dom: 19.2.4
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Dynamic Routes, Pages Router, Loading UI and Streaming

next start (local)

### Additional context

_No response_

mataspetrikas:
Hi NextJs team,

### What?

I am working on a large pages-router-based NextJS app and while implementing CSP headers, I have discovered an annoying bug. It seems that nonce is not passed to the suspense boundaries when using Pages Router. The App router seems to be unaffected. The bug was already previously reported in teh past eg [#96443](https://github.com/vercel/next.js/issues/96443)

### Why?

We could see that the Pages router correctly gets the nonce from the header and applies it to the script tags it renders. But it doesn't hand it over to React Fizz renderer. So the scripts in boundaries get blocked when enfiorcing the CSP policy. An example of a blocked script would be the Reacts reveal timing script `requestAnimationFrame(function(){$RT=performance.now()});` that gets inlined by React 19.x.

### How?

We tried different approaches to fix that but we either had to rewrite the script tags in the server or monkey-patch nextJS. Using Claude we could narrow down the problem to **one line** in the `render.tsx` - simply by passing the nonce to the Fizz it would fix the issue.

Tests added that should show how the suspend then works after CSP fix - it fails without a fix, correctly verifying the problem.

Pls let me know if you need any chanegs or additional info.

Fixes #96443

### DIFF
--- a/packages/next/src/server/render.tsx
+++ b/packages/next/src/server/render.tsx
@@ -1377,6 +1377,7 @@ export async function renderToHTMLImpl(
       return await renderToInitialFizzStream({
         ReactDOMServer: ReactDOMServerPages,
         element: content,
+        streamOptions: { nonce },
       })
     }
 
--- a/test/e2e/pages-router-nonce-streaming/middleware.js
+++ b/test/e2e/pages-router-nonce-streaming/middleware.js
@@ -1,0 +1,19 @@
+import { NextResponse } from 'next/server'
+
+const nonce = 'test-nonce'
+const csp = `default-src 'self'; script-src 'self' 'nonce-${nonce}'; style-src 'self' 'unsafe-inline'`
+
+export function middleware(request) {
+  // The renderer reads the nonce off the *request* CSP header, while the browser
+  // needs it on the response, so both are set here.
+  const requestHeaders = new Headers(request.headers)
+  requestHeaders.set('Content-Security-Policy', csp)
+
+  const response = NextResponse.next({ request: { headers: requestHeaders } })
+  response.headers.set('Content-Security-Policy', csp)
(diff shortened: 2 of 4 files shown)


mataspetrikas:
A side note, this is related to the problems described in the #67042 which were only partially fixed in #78936
I think once we started using CSP and streaming/Suspense more we started noticing that the nonce was not working there...

PR #97860, opened 2026-08-25
fix(turbo-tasks): let parking_lot block on wasm instead of panicking

sokra:
### What?

Makes `parking_lot` locks actually block on wasm instead of panicking, and removes 33 of the 35
`ignore`s that panic had forced onto the wasm test suites.

### Why?

`parking_lot_core` picks its thread parker in `thread_parker/mod.rs`. The working wasm parker
(`wasm_atomic.rs`, built on `memory_atomic_wait32` / `memory_atomic_notify`) is selected only under

```
all(feature = "nightly", target_family = "wasm", target_feature = "atomics")
```

Otherwise it falls back to `wasm.rs`, whose `park()` is
`panic!("Parking not supported on this platform")`. `wasm32-wasip1-threads` **does** report
`target_feature = "atomics"`, so the only missing piece was the feature — every contended lock
panicked, which is why so many tests were ignored on wasm.

### How?

Two changes, and **neither alone is sufficient**:

- bump `parking_lot_core` 0.9.8 → 0.9.12, because 0.9.8's atomic parker needs `feature(stdsimd)`,
  removed from Rust long ago;
- enable `parking_lot`'s `nightly` feature **for wasm targets only** (it forwards to
  `parking_lot_core/nightly`). 0.9.12 *still* gates the atomic parker behind it, so a bump on its own
  keeps selecting the panicking stub silently.

Verified on the resolved feature graph rather than the manifest — `nightly` is present for
`wasm32-wasip1-threads` and absent for the host.

### Testing

`contended_mutex_blocks_and_hands_over` is built so it cannot pass vacuously: a barrier guarantees the
waiter actually contends, the waiter must block for a measurable interval, and it must observe the
holder's write. **Negative control**: with the feature removed it fails at
`parking_lot_core-0.9.12/src/thread_parker/wasm.rs:26` with *"Parking not supported on this platform"*.

```
turbo-tasks          54 pass / 28 ignored  ->  72 pass / 11 ignored
turbo-tasks-backend  66 pass / 32 ignored  ->  82 pass / 15 ignored
native                     0 ignored, unchanged
```

Two of the old ignores remain, under a different and verified cause: their work completes in ~500 ms
of an 800 ms serial sum, so the parallelism is real, but dropping a multi-thread tokio runtime while
their blocking helpers are still alive deadlocks on wasm. That was previously hidden because
`panic = abort` killed the process at the failing assertion before teardown ran. It is fixed in the
follow-up runtime layer.

Also fixes a latent bug unrelated to wasm: `test_parallel_try_for_each_mut` asserted every element was
incremented, which only holds when the chunk size is 1. Chunking comes from
`available_parallelism() * 4` and each chunk stops at its first error, so with parallelism 1 the chunk
size is 2 and every second item is skipped — it would fail on any single-core machine. It now asserts
the parallelism-independent contract.


sokra:
Looks like the CI job `build-and-test / test next-swc wasi / build` runs into the 30min timeout quite often. I think it needs a larger timeout

turbopack/crates/turbo-tasks/src/scope_bounded.rs
```diff
@@ -281,14 +281,66 @@ mod tests {
...
+    /// (`wasm_atomic.rs`, built on `memory_atomic_wait32` / `memory_atomic_notify`) is chosen
+    /// only under `all(feature = "nightly", target_family = "wasm", target_feature =
+    /// "atomics")`. Without that feature it silently falls back to `wasm.rs`, whose `park()` is
+    /// `panic!("Parking not supported on this platform")` — so this test is the negative control
+    /// for the feature being enabled, and it fails loudly if the stub is ever selected again.
+    ///
+    /// Timing-based on purpose: a lock that returned immediately without blocking would satisfy a
+    /// pure handover assertion, so real elapsed time is what distinguishes blocking from spinning
+    /// past the guard.
+    #[test]
```
bgw, line 297: This is fine, but probably overkill, as there are tons of other tests that will fail if parking_lot is misconfigured. This just reads as a unit test for parking_lot.

turbopack/crates/turbo-tasks/src/parallel.rs
```diff
@@ -339,11 +334,26 @@ mod tests {
...
+        assert_eq!(input.len(), 5);
+        for (i, &value) in input.iter().enumerate() {
+            let original = i as i32 + 1;
+            assert!(
+                value == original || value == original + 10,
+                "input[{i}] = {value}, expected {original} or {}",
+                original + 10
+            );
+        }
+        assert_eq!(input[0], 11);
```
bgw, line 353: Sure, but I think this also really waters down the test, this really only tests that the first iteration does any sort of mutation. The original test was asserting that we try to start multiple jobs in parallel.

What if we just call `available_parallelism()` here, and size the test appropriately from that? On wasm the test might still be pretty useless, but at least it would still be useful on other platforms.

Otherwise, I think we should just delete the test. In this current state it seems pretty useless.

bgw (approved):

PR #97868, opened 2026-08-25
Turbopack: pass comments to SWC plugins when a file has any comments

### LINKED ISSUE #97866: Turbopack never passes comments to SWC wasm plugins unless a file has both leading AND trailing comments

https://github.com/instagibb/turbopack-comments-repro

### To Reproduce

With Turbopack (NextJS default)
1. Run `npm run dev`
2. Browse to http://localhost:3000/
3. The page reads istanbul's `__coverage__` global store and renders the result (see DevTools console if preferred). The [istanbul ignore hint comment](https://github.com/istanbuljs/nyc#parsing-hints-ignoring-lines) in `leading-only.ts` has not been processed and the function was instrumented while `with-trailing.ts` which is identical except for one trailing comment has the hint honoured and there is no instrumenting

For comparison with Webpack
1. Run `npm run dev:webpack`
2. Browse to http://localhost:3000/
3. The output shows that the istanbul ignore hint comment has been honoured in both instances

Or for a fully automatic test
1. Run `npm run check` to build with both Turbopack and Webpack and print the comparison

Current behaviour: Turbopack only passes the comments to SWC plugins when a file contains both leading AND trailing comments. Otherwise the plugin receives an empty comment store, so any comment-driven behaviour silently does nothing (no error)

See: [`swc_ecma_transform_plugins.rs` lines 136–141 on canary](https://github.com/vercel/next.js/blob/canary/turbopack/crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs#L136-L141)

```rust
let should_enable_comments_proxy =
    !ctx.comments.leading.is_empty() && !ctx.comments.trailing.is_empty();

//[TODO]: as same as swc/core does, we should set should_enable_comments_proxy
// depends on the src's comments availability. For now, check naively if leading
// / trailing comments are empty.
```

Expected behaviour: Turbopack should pass comments to SWC plugins when a file contains any comments (leading OR trailing)

```rust
let should_enable_comments_proxy =
    !ctx.comments.leading.is_empty() || !ctx.comments.trailing.is_empty();
```

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 49152
  Available CPU cores: 12
Binaries:
  Node: 24.14.1
  npm: 11.11.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.4.0-canary.6 // Latest available version is detected (16.4.0-canary.6).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Turbopack, SWC

### Additional context

_No response_

instagibb:
### What?

Allow SWC plugins to receive comments when the file has _any_ by updating `should_enable_comments_proxy` check in  [`swc_ecma_transform_plugins.rs` lines 136–141 on canary](https://github.com/vercel/next.js/blob/canary/turbopack/crates/turbopack-ecmascript-plugins/src/transform/swc_ecma_transform_plugins.rs) to use `||` instead of `&&` 

Also cleaned up the `TODO`

### Why?

Turbopack only exposed comments to SWC plugins when a file had _both_ leading AND trailing comments. Comment driven plugins which worked under Webpack, failed silently when using Turbopack (i.e. acted as if the comment was not there).

This was initially discovered with a proprietary plugin but was reproduced with the [swc-plugin-coverage-instrument](https://www.npmjs.com/package/swc-plugin-coverage-instrument) which failed silently (i.e. comment hint ignored) in an identical way to the proprietary one

### How?

Swapped the `&&` to an `||` instead of removing the check altogether which maintains the gate for files which have no comments

Tests added using the [swc-plugin-coverage-instrument](https://www.npmjs.com/package/swc-plugin-coverage-instrument) plugin which fail on unpatched Turbopack but pass with the fix.

Fixes #97866

### DIFF
--- a/test/e2e/app-dir/swc-plugins-comments/app/layout.tsx
+++ b/test/e2e/app-dir/swc-plugins-comments/app/layout.tsx
@@ -1,0 +1,8 @@
+import { ReactNode } from 'react'
+export default function Root({ children }: { children: ReactNode }) {
+  return (
+    <html>
+      <body>{children}</body>
+    </html>
+  )
+}
--- a/test/e2e/app-dir/swc-plugins-comments/app/page.tsx
+++ b/test/e2e/app-dir/swc-plugins-comments/app/page.tsx
@@ -1,0 +1,34 @@
+import { leadingOnly } from '../lib/leading-only'
+import { withTrailing } from '../lib/with-trailing'
+
+const FILES = ['lib/leading-only.js', 'lib/with-trailing.js']
+
+// The plugin stores each file's coverage map on `globalThis.__coverage__`
+// (keyed by path, relative or absolute depending on the bundler). `fnMap` is
+// empty when the ignore hint was honored.
+function getInstrumentedFunctions(file: string): string[] {
+  const coverage: Record<string, any> = (globalThis as any).__coverage__ ?? {}
+  const key = Object.keys(coverage).find((path) => path.endsWith(file))
+  if (!key) return ['<coverage map not found>']
+  return Object.values(coverage[key].fnMap ?? {}).map((fn: any) => fn.name)
+}
+
+export default function Page() {
+  return (
+    <main>
+      <ul>
+        {FILES.map((file) => {
+          const instrumented = getInstrumentedFunctions(file)
+          return (
+            <li key={file} id={file.replace(/[^a-z-]/g, '-')}>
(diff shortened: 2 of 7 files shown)


instagibb:
Hey just wondering if someone can trigger the CI for a first time contributor so I can address any issues?

PR #97880, opened 2026-08-25
Show performance summary with verbose traces

DavidIlie:
## Summary

- show the slowest recorded operation summary only when Verbose traces is enabled
- leave the recorded trace and request data unchanged

The summary is useful while inspecting the full framework trace, but it adds noise to the default request view. This keeps the default panel focused and reveals the derived performance hint with the existing verbose setting.

Stacked on #97796.

## Verification

- `pnpm --filter=next build`
- focused Request Insights E2E with Turbopack and webpack


DavidIlie:
Closing this standalone PR; the change belongs in the existing Request Insights filters/settings layer.

PR #97901, opened 2026-08-25
test(turbo-tasks-backend): run the db_versioning tests on wasm

sokra:
### What?

Makes `turbo-tasks-backend`'s `db_versioning` tests run on `wasm32-wasip1-threads` instead of being
skipped, by routing temporary-directory creation in tests through a small helper that does not depend
on `std::env::temp_dir()`.

Twelve test instances that were previously skipped now execute on wasm. Three related
`kv_backing_storage` instances remain skipped, but for a different — and verified — reason (see below).
Native behaviour is unchanged: the same tests run exactly as before, and nothing is newly skipped there.

The wider WASI suite also exposed an unrelated intermittent stall in
`test_unbounded_busy_queue_does_not_churn_workers`: passing isolated runs finish in about 0.25 seconds,
while repeated runs can stop making progress indefinitely. That stress test is now ignored only on
WASI, with the observed reason recorded; native coverage remains enabled while the underlying
multi-threaded runtime stall is investigated separately.

### Why?

These tests were skipped with the reason `no temp directory on WASI`, which turned out to be only half
true. WASI genuinely has no notion of a system temp directory — `std::env::temp_dir()` is hard-coded to
`unsupported()` in `std/sys/paths/wasi.rs`, so it panics regardless of `TMPDIR` — but the *filesystem
itself works fine*. A WASI guest can use any directory the host preopens for it.

So the blocker was in how the tests asked for a directory, not in the platform. Asking WASI where "temp"
is fails; creating a directory inside a path the host already granted succeeds.

This matters beyond tidiness: skip reasons are the primary record of what does and does not work on
wasm, and an inaccurate one is worse than no reason at all, because it makes a fixable problem look like
a platform limit and quietly discourages anyone from trying.

Fixing it also uncovered the *real* limitation underneath. With temp directories working, the three
`kv_backing_storage` tests get further and hit a genuine platform gap: `memmap2` ships `unix.rs`,
`windows.rs` and a stub for every other target, so `turbo-persistence` cannot open a database on WASI
at all (`Failed to mmap: platform not supported`). Those three are still skipped, now with that cause
recorded instead of the temp-directory one. That is a real, separate piece of work — persistent caching
cannot function on wasm until `turbo-persistence` grows a non-mmap read path — and it is now visible
rather than hidden behind a misleading reason.

### How?

A single test-only helper owns the platform difference, so the ~11 call sites stay identical in shape
and no test carries its own `cfg`. On native it delegates to `tempfile::tempdir()`, preserving today's
behaviour byte for byte, including having `TempDir` own cleanup. On wasm it creates the directory inside
a preopened path instead of consulting the environment. The helper is crate-private and compiled only
for tests, so it adds nothing to the shipped crate, and it returns `tempfile`'s own error type rather
than erasing it.

Two deliberate choices are worth calling out for review:

- **Every remaining skip is justified by an observed failure**, not by inference. Each of the three
  `kv_backing_storage` tests was run individually and traced to `memmap2`'s stub before being skipped
  again; the twelve `db_versioning` ones were confirmed to pass. Nothing is skipped speculatively.
- **Skips use `ignore` with a reason rather than `cfg`**, so they appear in test output as
  `ignored, <reason>` instead of silently vanishing from the run. That is what keeps the count of
  what-doesn't-work-on-wasm honest and reviewable, and it is asserted that the native run still has
  zero ignored tests.

The wasm coupling is documented at the helper: it relies on the host preopening a directory, which the
repository's Node wasi test host does. A different runner must provide the same, or update the helper.


turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -588,10 +589,16 @@ mod tests {
     ///
     /// This is a lower-level test that verifies the database layer correctly handles
     /// the case where multiple task IDs are stored under the same hash key.
+    // These open a real `TurboKeyValueDatabase`, and `turbo-persistence` reads its files via
+    // `memmap2`, which ships a stub returning "platform not supported" for every target that is
+    // neither unix nor windows (`memmap2/src/stub.rs`). Nothing in the test can work around that.
+    #[cfg_attr(
+        target_family = "wasm",
+        ignore = "no mmap on WASI: memmap2 has no implementation for this target"
```
lukesandberg, line 597: shouldn't we enable the FileRead env var here?

ditto for all the tests that trigger persistence

can we enable the backend to use fileread mode when under wasm without the env var so this just works?

turbopack/crates/turbo-tasks-backend/src/utils/test_temp_dir.rs
```diff
@@ -0,0 +1,33 @@
...
+//! only ever want a private scratch directory, not the system one.
+
+use std::io;
+
+use tempfile::TempDir;
+
+/// The pre-opened directory to create temporary directories under on WASI.
+///
+/// The WASI test host (`scripts/wasi-test-host/run.mjs`) pre-opens the process working directory as
+/// `/`, so this resolves to a real, writable directory. Keep it in sync with that host.
```
bgw: That's different than the actual system tempdir though? Can we pass in the actual tempdir path as an environment variable?

It sounds like this is just going to end up throwing a bunch of random crap in your `target` directory.

turbopack/crates/turbo-tasks-backend/src/kv_backing_storage.rs
```diff
@@ -588,10 +589,16 @@ mod tests {
     ///
     /// This is a lower-level test that verifies the database layer correctly handles
     /// the case where multiple task IDs are stored under the same hash key.
+    // These open a real `TurboKeyValueDatabase`, and `turbo-persistence` reads its files via
+    // `memmap2`, which ships a stub returning "platform not supported" for every target that is
+    // neither unix nor windows (`memmap2/src/stub.rs`). Nothing in the test can work around that.
```
bgw: The `ignore` value is sufficient, we don't need a long comment.

```suggestion
```

(but also Luke's comment about using the file read codepath instead of mmap seems better)

turbopack/crates/turbo-tasks/src/scope_unbounded.rs
```diff
@@ -654,7 +654,10 @@ mod tests {
     /// swallowed by the wind-down: the abort's queue-clear races the panic's unwind through
     /// `catch_unwind` -> `on_item_finished`.
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    #[cfg_attr(target_family = "wasm", ignore = "no unwinding on wasm")]
+    #[cfg_attr(
+        target_family = "wasm",
+        ignore = "no unwinding on wasm: std is built panic=abort, so catch_unwind cannot catch"
+    )]
```
bgw: This sounds like it could be a pretty breaking assumption for Turbopack in general?

At the very least it would prevent us from being able to delete the cache directory upon panic.

bgw (approved):

PR #97902, opened 2026-08-25
Guard filesystem reads against unresolved symlinks

sokra:
### What?

Adds debug-only OS realpath validation to successful `DiskFileSystem` file and directory reads. When a successfully canonicalized path differs from the supplied path, the read returns a normal task error naming both paths.

Fixes pattern/glob traversal and NFT tracing so physical filesystem access uses resolved paths while logical paths remain available for user-visible specifiers and complete symlink-chain recovery.

### Why?

Reading through an unresolved symlink parent gives the same filesystem object multiple path identities. That can make Turbo Tasks dependency tracking and invalidation inconsistent and can produce invalid deployment ZIPs when NFT output contains files below unresolved links.

The checks return errors rather than asserting because paths can disagree temporarily under eventual consistency. Propagating a task error avoids panicking a worker thread while still exposing invalid callers during development.

### How?

The validation lives directly in `DiskFileSystem::read` and `DiskFileSystem::raw_read_dir`. It calls the OS canonicalization API inline instead of the Turbo Tasks realpath task, keeping the diagnostic out of the task dependency graph. The guard runs only after the OS read succeeds, so missing/non-directory probes preserve their existing behavior.

`read_matches` resolves each physical directory immediately before enumeration while retaining logical `PatternMatch` paths.

`read_glob` and `track_glob` now resolve their initial directory before enumeration. Symlinks discovered later through wildcard segments are also traversed through resolved targets. `ReadGlobResult` deliberately retains logical paths rooted at the supplied base, allowing consumers to call `realpath_with_links` and recover the complete symlink chain.

Consumers follow that contract explicitly:

- NFT includes expand each logical match with `realpath_with_links`, emit resolved files and every traversed symlink, skip resolved directory targets, and deterministically deduplicate/sort output.
- `import.meta.glob` uses recursive logical keys as the source of user-visible requests, while module resolution follows and tracks symlinks.
- the hash-glob example resolves returned logical paths before reading.

Webpack-loader context dependencies are covered for both `path/to/symlink/inner/path/*` and `path/to/*/inner/path/*`. The loader fixture performs its directory read with Node `fs`, reports the directory using `addContextDependency`, and Turbopack tracks the resolved target.

### Verification

- `cargo fmt -p turbo-tasks-fs -p turbopack-ecmascript -p next-api -- --check`
- `cargo clippy -p turbo-tasks-fs -p turbopack-ecmascript -p next-api --all-targets`
- `cargo test -p turbo-tasks-fs` (128 passed)
- `cargo test -p next-api` (7 passed)
- `cargo check -p turbo-tasks-fs --examples`
- `import.meta.glob` symlink execution fixture (1 passed)
- Nine targeted node-file-trace CI cases with `release-with-assertions` (9 passed)
- `pnpm build-all`
- `webpack-loader-fs` Turbopack dev e2e (1 passed)
- `build-trace-extra-entries-turbo` Turbopack production e2e (1 passed)
- twoslash Turbopack production, normal mode (4 passed)
- twoslash Turbopack production, cache-components mode (4 passed)
- `bench/heavy-npm-deps` Turbopack development smoke test (HTTP 200)

### Notes

The disk guard is cross-platform, while its symlink-parent regression test is Unix-only, matching neighbouring symlink tests. On Windows, OS canonicalization can also normalize casing and 8.3 short names; a debug read using a non-canonical spelling will therefore return the same diagnostic error.


turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -567,8 +579,33 @@ impl fmt::Display for RealPathError {
...
+    {
+        bail!("{operation} called with unresolved path {path:?}; resolve it to {realpath:?} first");
+    }
+    Ok(())
+}
+
+#[cfg(debug_assertions)]
+#[turbo_tasks::function]
+async fn read_file(path: FileSystemPath) -> Result<Vc<FileContent>> {
+    ensure_path_is_realpath("read_file", &path, &path.realpath().await?)?;
```
sokra: Move this logic into DiskFileSystem::read and run the OS version of realpath instead of the turbo_tasks version

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -567,8 +579,33 @@ impl fmt::Display for RealPathError {
...
+
+    Ok(path.fs().read(path))
+}
+
 #[turbo_tasks::function]
 async fn read_dir(path: FileSystemPath) -> Result<Vc<DirectoryContent>> {
+    #[cfg(debug_assertions)]
+    ensure_path_is_realpath("read_dir", &path, &path.realpath().await?)?;
```
sokra: Similar to this. Move it to DiskFileSystem::read_dir

mischnic:
There is a merge conflict already

crates/next-api/src/nft.rs
```diff
@@ -193,19 +193,28 @@ async fn get_glob_includes(
     let glob_result = project_root_path.read_glob(glob).await?;
 
     // Walk the full glob_result using an explicit stack to avoid async recursion overheads.
-    // Use a BTreeSet to get deterministic order (return value of `read_glob` has random order).
-    let mut result = vec![];
+    // Deduplicate symlinks shared by many matches. The return value of `read_glob` has random
+    // order, so the result is sorted below.
+    let mut result = FxIndexSet::default();
```
sokra: FxHashMap would also work

crates/next-api/src/nft.rs
```diff
@@ -564,3 +573,81 @@ impl Issue for ForbiddenTracedFileIssue {
...
+    use turbo_tasks_backend::{BackendOptions, TurboTasksBackend, noop_backing_storage};
+    use turbo_tasks_fs::{
+        DiskFileSystem, FileSystem,
+        glob::{Glob, GlobOptions},
+    };
+
+    use crate::nft::get_glob_includes;
+
+    #[turbo_tasks::function(operation, root)]
+    async fn get_glob_includes_operation(disk_root: RcStr) -> anyhow::Result<()> {
```
sokra:
```suggestion
    async fn assert_glob_includes_operation(disk_root: RcStr) -> anyhow::Result<()> {
```

turbopack/crates/turbo-tasks-fs/src/read_glob.rs
bgw, line 15: Discussed over zoom: It might be better to return `FxHashMap<RcStr, (DirectoryEntry, ResolvedVc<RealPathWithLinksResult>)>`, but that can be a separate PR.

wbinnssmith (approved):

PR #97903, opened 2026-08-25
Separate server HMR entries from VersionedContentMap

wbinnssmith:
## Summary

Move server HMR chunk-list ownership out of `VersionedContentMap` and into a dedicated `ServerHmrEntryMap`.

`VersionedContentMap` no longer scans all emitted assets to discover server HMR chunk lists or filters them for each pull. Instead, endpoints register their chunk entry lists under the route entry key when they are written. Server HMR snapshots then read only the chunk lists owned by that route.

This keeps `VersionedContentMap` focused on asset lookup and emission, while server HMR state is compartmentalized by route. It also removes the project-wide chunk-list scan and the associated path filtering from the update path.

The endpoint and native binding changes pass the route entry key through to the server HMR pull without changing the update wire format.

## Verification

- `cargo check -p next-api -p next-napi-bindings`


crates/next-api/src/project.rs
```diff
@@ -953,6 +958,32 @@ pub struct Project {
     server_hmr: bool,
 }
 
+impl Project {
+    pub async fn register_server_hmr_entry(
```
wbinnssmith: at some point we could unregister if we decide what the rules should be around that

crates/next-api/src/project.rs
```diff
@@ -953,6 +958,32 @@ pub struct Project {
...
+    }
+
+    pub async fn server_hmr_chunk_lists(
+        &self,
+        entry_key: &str,
+    ) -> Result<ReadRef<ServerHmrChunkLists>> {
+        let chunk_lists = self
+            .server_hmr_entry_map
+            .await?
+            .get(entry_key)
```
lukesandberg: why is .get() untracked, should be fine to track?

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -1949,8 +1949,8 @@ impl NapiServerHmrUpdate {
 #[napi]
 pub async fn project_get_server_hmr_update(
     #[napi(ts_arg_type = "{ __napiType: \"Project\" }")] project: &External<ProjectInstance>,
+    entry_key: RcStr,
```
lukesandberg, line 1952: what is 'entry_key' logically?

lukesandberg: maybe introduce a newtype in rust to document it?

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -27,12 +31,53 @@ use turbopack_nodejs::ecmascript::node::entry::chunk_list_content::{
     EcmascriptBuildNodeChunkListContent, compute_update_from_version_operation,
 };
 
+/// Canonical JSON key identifying a Next.js entrypoint by router, side, and page.
```
lukesandberg, line 37: 'side'?

crates/next-napi-bindings/src/next_api/project.rs
```diff
@@ -1857,15 +1857,15 @@ async fn hmr_update_with_issues_operation(
     .cell())
 }
 
-#[turbo_tasks::value(serialization = "skip")]
+#[turbo_tasks::value(serialization = "skip", evict = "never")]
```
lukesandberg, line 1871: nit: drop a short comment on these 'evict=never'

it is due to the ReadRef?

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -345,9 +430,281 @@ mod tests {
...
+    };
+    use crate::{
+        project::{
+            DefineEnv, DraftModeOptions, PartialProjectOptions, Project, ProjectContainer,
+            ProjectOptions, WatchOptions,
+        },
+        route::{EndpointOutput, EndpointOutputPaths},
     };
 
+    #[derive(turbo_tasks::NonLocalValue, turbo_tasks::trace::TraceRawVcs)]
```
lukesandberg, line 444: nit:import

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -45,18 +84,53 @@ impl ServerHmrChunkLists {
...
         &self.0
     }
+}
+
+#[derive(Debug)]
+pub struct ServerHmrEntryMap {
+    entries: State<FxIndexMap<ServerHmrEntryKey, GcRoot<EndpointOutput>>>,
+}
+
+unsafe impl OperationValue for ServerHmrEntryKey {}
```
lukesandberg, line 94: this impl seems out of place (move it up with the struct?)

also i think you should be able to derive this one

crates/next-api/src/aggregate_hmr.rs
```diff
@@ -345,9 +430,281 @@ mod tests {
...
+                        no_mangling: false,
+                        write_routes_hashes_manifest: false,
+                        current_node_js_version: "22.0.0".into(),
+                        debug_build_paths: None,
+                        deferred_entries: None,
+                        is_persistent_caching_enabled: false,
+                        next_version: "test".into(),
+                        server_hmr: true,
+                    },
+                )
```
lukesandberg, line 683: wild that you are bootstrapping  whole turbopack inside a unit test (pretty cool actually)

extract a helper for the container setup and bootstrapping?  it would improve test readability.... if possible

crates/next-api/src/project.rs
```diff
@@ -884,7 +885,7 @@ impl ProjectContainer {
 }
 
 #[derive(Clone)]
-#[turbo_tasks::value]
+#[turbo_tasks::value(serialization = "skip", evict = "never")]
```
lukesandberg, line 888: should we disable persistence for this or just bincode::skip the new field?

it would require it to be initialized on demand of course

PR #97907, opened 2026-08-26
fix(next): canonicalize percent-encoded param before bfcacheId compare

shoemoney:
Fixes a Cache Components bfcache mismatch on search-param-only navigations.

A search-param-only navigation mints a new bfcacheId when the dynamic param is percent-encoded. `matchSegment` in `packages/next/src/client/components/match-segments.ts` compared `existingSegment[1] === segment[1]` directly, so `%2F` versus `%252F`, or decoded versus encoded forms, were treated as different segments.

`matchSegment` now compares both cache keys through `canonicalizeParamCacheKey`, which splits catch-all keys on `/`, canonicalizes each part with `canonicalizeURLPart` from `route-params` (`decodeURIComponent` then `encodeURIComponent`), and rejoins, so `/` separators are preserved (535e17f).

Verified RED to GREEN. Formatting changes are limited to the touched file.

### DIFF
--- a/packages/next/src/client/components/match-segments.ts
+++ b/packages/next/src/client/components/match-segments.ts
@@ -1,6 +1,18 @@
 import type { Segment } from '../../shared/lib/app-router-types'
 import { canonicalizeURLPart } from '../route-params'
 
+// `paramCacheKey` (segment[1]) is a single canonicalized URL part for
+// dynamic params, but for catch-all params it's multiple parts joined by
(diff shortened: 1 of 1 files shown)


shoemoney:
Done, fixed in 535e17f: canonicalize each /-delimited part of the catch-all cache key individually and rejoin, instead of canonicalizing the whole joined string.

PR #97909, opened 2026-08-26
fix(eslint-plugin-next): support pageExtensions in no-html-link-for-pages

Iitian001:
Fixes https://github.com/vercel/next.js/issues/53473

This PR fixes an issue where the `@next/next/no-html-link-for-pages` rule incorrectly warned about custom page extensions not being recognized. 

### Changes
- Updated `parseUrlForPages` and `parseUrlForAppDir` to accept a `pageExtensions` parameter instead of hardcoding `js(x)` and `ts(x)`.
- Fixed a bug where `parseUrlForAppDir` incorrectly recursed into `parseUrlForPages` for app directory subdirectories.
- Updated `no-html-link-for-pages.ts` to read `context.settings.next.pageExtensions` and pass it to the URL parsers.

With this change, users can define their custom `pageExtensions` in their ESLint settings:
```json
{
  "settings": {
    "next": {
      "pageExtensions": ["page.tsx", "page.ts"]
    }
  }
}
```

### DIFF
--- a/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
+++ b/packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts
@@ -107,8 +107,22 @@ export default defineRule({
       return {}
     }
 
-    const pageUrls = cachedGetUrlFromPagesDirectories('/', foundPagesDirs)
-    const appDirUrls = cachedGetUrlFromAppDirectory('/', foundAppDirs)
+    const pageExtensions = context.settings?.next?.pageExtensions || [
+      'tsx',
+      'ts',
+      'jsx',
+      'js',
+    ]
+    const pageUrls = cachedGetUrlFromPagesDirectories(
+      '/',
+      foundPagesDirs,
+      pageExtensions
+    )
+    const appDirUrls = cachedGetUrlFromAppDirectory(
+      '/',
+      foundAppDirs,
+      pageExtensions
+    )
     const allUrlRegex = [...pageUrls, ...appDirUrls]
 
     return {
--- a/packages/eslint-plugin-next/src/utils/url.ts
+++ b/packages/eslint-plugin-next/src/utils/url.ts
@@ -8,25 +8,42 @@ const fsReadDirSyncCache = {}
 /**
  * Recursively parse directory for page URLs.
  */
-function parseUrlForPages(urlprefix: string, directory: string) {
+function parseUrlForPages(
+  urlprefix: string,
+  directory: string,
+  pageExtensions: string[]
+) {
   fsReadDirSyncCache[directory] ??= fs.readdirSync(directory, {
     withFileTypes: true,
   })
   const res = []
+  const extPattern = new RegExp(
+    `\\.(${pageExtensions
+      .map((ext) => ext.replace(/\\./g, '\\.'))
+      .join('|')})$`
+  )
+
   fsReadDirSyncCache[directory].forEach((dirent) => {
-    // TODO: this should account for all page extensions
-    // not just js(x) and ts(x)
-    if (/(\.(j|t)sx?)$/.test(dirent.name)) {
-      if (/^index(\.(j|t)sx?)$/.test(dirent.name)) {
-        res.push(
-          `${urlprefix}${dirent.name.replace(/^index(\.(j|t)sx?)$/, '')}`
-        )
+    if (extPattern.test(dirent.name)) {
+      const indexRegex = new RegExp(
+        `^index\\.(${pageExtensions
+          .map((ext) => ext.replace(/\\./g, '\\.'))
+          .join('|')})$`
+      )
+      if (indexRegex.test(dirent.name)) {
+        res.push(`${urlprefix}${dirent.name.replace(indexRegex, '')}`)
       }
-      res.push(`${urlprefix}${dirent.name.replace(/(\.(j|t)sx?)$/, '')}`)
+      res.push(`${urlprefix}${dirent.name.replace(extPattern, '')}`)
     } else {
       const dirPath = path.join(directory, dirent.name)
       if (dirent.isDirectory() && !dirent.isSymbolicLink()) {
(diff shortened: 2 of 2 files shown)


Abhirup0 (commented):
Thanks for working on this! A few items need to be addressed before this can be merged:

1. **Regex Escaping Bug (`url.ts`)**: `ext.replace(/\\./g, '\\.')` matches a backslash followed by a character, not a literal dot. Dots in extensions like `page.tsx` remain unescaped wildcards in `new RegExp`. Please update to `ext.replace(/\./g, '\\.')` (or escape regex special characters).
2. **Move RegExp Compilation Outside `forEach`**: `indexRegex`, `pageRegex`, and `layoutRegex` are recompiled on every dirent inside the `forEach` loop. They should be created once outside the loop per function call.
3. **Colocated App Router Files**: In `parseUrlForAppDir`, the `!layoutRegex` fallback registers any colocated component (e.g. `app/dashboard/button.tsx`) as a route. In App Router, only `page.*` (and `route.*`) should be mapped to route URLs.
4. **Tests**: Please add test cases in `test/unit/eslint-plugin-next/no-html-link-for-pages.test.ts` covering `settings.next.pageExtensions` with multi-part extensions like `['page.tsx', 'page.ts']` for both `pages/` and `app/`.

Abhirup0 (commented):
Thanks for addressing the initial feedback and adding the unit tests @Iitian001!

Taking a closer look at the latest commit (`98e00b9`), there are a few important edge cases and a bug with static App Router routes that need to be resolved:

---

### 1. Bug: Nested Static Routes in App Router Never Match (Missing `normalizeURL`)

In `getUrlFromAppDirectory`:
```ts
// packages/eslint-plugin-next/src/utils/url.ts
(url) => `^${normalizeAppPath(url)}$`
```

`normalizeAppPath(url)` returns paths without a trailing slash (e.g. `/about` or `/dashboard`). However, in `no-html-link-for-pages.ts`, the lint rule tests against `normalizeURL(hrefPath)`:
```ts
if (foundUrl.test(normalizeURL(hrefPath)))
```
Because `normalizeURL` always enforces a trailing slash (e.g. `/about/`), `/^\/about$/` will never match `/about/`.

The test in the PR passed because `invalidStaticCode` only tests the root homepage (`/`), where `ensureLeadingSlash('')` happens to end with `/`. For any nested static route (e.g. `<a href="/about">`), the rule fails to trigger.

**Fix:** Wrap `normalizeAppPath(url)` in `normalizeURL(normalizeAppPath(url))` just like in `getUrlFromPagesDirectories`:
```ts
export function getUrlFromAppDirectory(
  urlPrefix: string,
  directories: string[],
  pageExtensions: string[] = ['tsx', 'ts', 'jsx', 'js']
) {
  return Array.from(
    new Set(
      directories
        .map((directory) =>
          parseUrlForAppDir(urlPrefix, directory, pageExtensions)
        )
        .flat()
        .map(
          (url) => `^${normalizeURL(normalizeAppPath(url))}$`
        )
    )
  ).map((urlReg) => {
    urlReg = urlReg.replace(/\[.*\]/g, '((?!.+?\\..+?).*?)')
    return new RegExp(urlReg)
  })
}
```

---

### 2. Regex Alternation Order Suffix Masking

In `parseUrlForPages` and `parseUrlForAppDir`:
```ts
const extPattern = new RegExp(
  `\\.(${pageExtensions
    .map((ext) => ext.replace(/\./g, '\\.'))
    .join('|')})$`
)
```
In JavaScript regex, `(a|b)` alternations are evaluated left-to-right. If `pageExtensions` contains single-part extensions before compound extensions (e.g., `['tsx', 'page.tsx']`), a file named `hello.page.tsx` will match `.tsx` first. Replacing `extPattern` then strips only `.tsx`, producing `hello.page` instead of `hello`.

**Fix:** Sort `pageExtensions` by length in descending order before building the regexes:
```ts
const sortedExtensions = [...pageExtensions].sort((a, b) => b.length - a.length)
const extPattern = new RegExp(
  `\\.(${sortedExtensions
    .map((ext) => ext.replace(/\./g, '\\.'))
    .join('|')})$`
)
```

---

### 3. Handle `pageExtensions: []` in ESLint Settings

In `no-html-link-for-pages.ts`:
```ts
const pageExtensions = context.settings?.next?.pageExtensions || [
  'tsx',
  'ts',
  'jsx',
  'js',
]
```
If a user passes `pageExtensions: []`, `[]` is truthy, so the default fallback is not used. An empty array creates regexes like `/\.()$/` and `/^index\.()$/`.

**Fix:** Ensure it is a non-empty array:
```ts
const pageExtensions =
  Array.isArray(context.settings?.next?.pageExtensions) &&
  context.settings.next.pageExtensions.length > 0
    ? context.settings.next.pageExtensions
    : ['tsx', 'ts', 'jsx', 'js']
```

---

### 4. Clean Up `dirent.name.replace(pageRouteRegex, '')`

In `parseUrlForAppDir`:
```ts
if (pageRouteRegex.test(dirent.name)) {
  res.push(`${urlprefix}${dirent.name.replace(pageRouteRegex, '')}`)
}
```
Because `pageRouteRegex` is anchored with `^` and `$`, replacing it in `dirent.name` always replaces the entire filename with `""`. It is literally equivalent to `res.push(urlprefix)`. We can simplify this and also verify `!dirent.isDirectory()`:
```ts
if (!dirent.isDirectory() && pageRouteRegex.test(dirent.name)) {
  res.push(urlprefix)
} else if (dirent.isDirectory() && !dirent.isSymbolicLink()) {
```

---

### 5. Tests for Colocated Non-Page Files & Nested Static Routes

Could you please add:
1. A test verifying that navigating to a colocated non-page file (e.g. `<a href="/list/foo/helper">`) produces no lint errors (`assert.deepEqual(report, [])`). Colocating helper/component files without triggering false positives is the main purpose of configuring `pageExtensions`.
2. A test for a nested static route in `app/` (e.g., `app/about/page.page.tsx` with `<a href="/about">`).

PR #97916, opened 2026-08-26
Update Convex example to App Router

mikecann:
## Why

Updating the convex example to bring it up to date, update dependencies etc.


mikecann:
@ianmacartney this is the replacement for #91286 and is ready for review when you have a chance. Thanks!

ianmacartney:
I wish you luck with a speedy review

examples/convex/package.json
```diff
@@ -1,24 +1,24 @@
 {
   "private": true,
   "scripts": {
-    "dev": "npm-run-all --parallel dev:backend dev:frontend",
+    "dev": "run-p dev:backend dev:frontend",
```
ianmacartney: We could use the --start flag on convex dev to drop the run-p dependency, if that is not too confusing

mikecann: Ah ye good call. I switched it.

ianmacartney:
Nice! I’ll close mine out when I get a chance. We could consider pinging
the vercel team in a private channel if we don’t see movement here

andrewimm (approved):

andrewimm:
Thanks for updating.
The next.js repo requires signed commits (https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)
If you amend your commit with a signature, and force-push it to the PR branch, it can be landed

mikecann:
Okay thanks, should be good now @andrewimm

PR #97928, opened 2026-08-26
fix(root-params): support root param names that are not valid JS identifiers

### LINKED ISSUE #97927: next/root-params typegen emits invalid TypeScript for hyphenated root param segments (e.g. [lang-country]) — build fails to type check

https://github.com/lcordier42/next-root-params-hyphen-repro

### To Reproduce

1. Have a dynamic **root** segment whose name is not a valid JS identifier, e.g. `app/[lang-country]/layout.tsx` (root layout inside the segment).
2. `npm run build` (i.e. `next build`)
3. The build fails during type checking:

```
.next/types/root-params.d.ts(4,23): error TS1005: '(' expected.
.next/types/root-params.d.ts(4,33): error TS1005: ';' expected.
Failed to type check.
```

**Current:** the root params typegen (enabled by default since #93863, shipped in 16.3.0) emits a named function export from the raw param name into `.next/types/root-params.d.ts`:

```ts
declare module 'next/root-params' {
  export function lang-country(): Promise<string>
}
```

`lang-country` is not a valid identifier, so the generated file is a TS syntax error. Since `next-env.d.ts` imports `./.next/types/root-params.d.ts` unconditionally, every type check of the project fails and `next build` cannot succeed. There is no user-side escape hatch (excluding the file in `tsconfig.json` does not help because of the direct import in `next-env.d.ts`).

**Expected:** valid TypeScript regardless of the segment name. Hyphenated dynamic segments are legal everywhere else in the router — the typed-routes generator (`routes.d.ts`) handles the same segment correctly by quoting the key (`{ "lang-country": string }`).

**Root cause:** `generateRootParamsTypes` in `packages/next/src/server/lib/router-utils/root-params-type-utils.ts` interpolates the param name directly into `export function ${paramName}()` without checking that it is a valid identifier. Still present on canary.

**Suggested fix** (this is what we currently apply as a patch): fall back to a string module export name (TS 4.5+) when the param name is not a valid identifier:

```ts
const exports = Array.from(rootParams.entries())
  .sort(([a], [b]) => a.localeCompare(b))
  .map(([paramName, info]) => {
    if (/^[A-Za-z_$][A-Za-z0-9_$]*$/.test(paramName)) {
      return `  export function ${paramName}(): ${getRootParamReturnType(info)}`
    }
    const safe = '_' + paramName.replace(/[^A-Za-z0-9_$]/g, '_')
    return `  function ${safe}(): ${getRootParamReturnType(info)}\n  export { ${safe} as "${paramName}" }`
  })
```

which generates:

```ts
declare module 'next/root-params' {
  function _lang_country(): Promise<string>
  export { _lang_country as "lang-country" }
}
```

Happy to send a PR with this if the approach sounds right.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 32768
  Available CPU cores: 10
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: 1.22.22
  pnpm: 10.27.0
Relevant Packages:
  next: 16.3.3 // Latest available version is detected (16.3.3).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

TypeScript, dynamicIO (in relation to rootParams typegen)

### Additional context

- Affects 16.3.0 through 16.3.3 (the generator shipped with 16.3.0); not present in <= 16.2.x.
- The runtime `await params` behavior is fine — only the generated `.d.ts` is broken.
- Real-world hit: a production monorepo with a `[lang-country]` root segment (values like `fr-fr`, `en-gb`) upgrading from 15.5.x.

lcordier42:
### Fixes #97927

The `next/root-params` codegen interpolates raw param names into named function exports. A dynamic **root** segment whose name is not a valid JS identifier — e.g. `app/[lang-country]/` — therefore generates invalid code in three places:

1. **`root-params.d.ts` typegen** (`root-params-type-utils.ts`): `export function lang-country(): ...` is a TS syntax error. Since `next-env.d.ts` imports the file unconditionally, every type check of the project fails and `next build` cannot succeed. This is the failure reported in #97927 (repro: https://github.com/lcordier42/next-root-params-hyphen-repro).
2. **Webpack loader** (`next-root-params-loader.ts`): same invalid named export in the generated runtime module.
3. **Turbopack virtual module** (`crates/next-core/src/next_root_params/mod.rs`): same.

### Fix

When a param name is not a valid identifier, declare a safely-named function and re-export it under the original param name via a **string module export name** (arbitrary module namespace identifier — ES2022, TS 4.5+):

```ts
declare module 'next/root-params' {
  function _lang_country(): Promise<string>
  export { _lang_country as 'lang-country' }
}
```

Runtime (webpack loader / Turbopack):

```js
function _lang_country() {
  return getRootParam('lang-country')
}
export { _lang_country as 'lang-country' }
```

Consumers access it through the module namespace (a named import is impossible for such names by construction):

```ts
import * as rootParams from 'next/root-params'
const langCountry = await rootParams['lang-country']()
```

Param names that are valid identifiers keep the exact same output as before.

### Tests

New `hyphenated` fixture in `test/e2e/app-dir/app-root-params-getters/`:

- wired into `typecheck.test.ts` (`tsc --noEmit` over the generated `root-params.d.ts`, with `@ts-expect-error` assertions) — this is the case that fails without the fix
- `hyphenated.test.ts`: renders `/en-us` and asserts the param value read via the module namespace in both layout and page — covers the two runtime generators

### Notes

- This mirrors what the typed-routes generator (`routes.d.ts`) already does for the same segments: it quotes the key (`{ "lang-country": string }`). Only the root-params generators missed the non-identifier case.
- Hyphenated dynamic segments are accepted by the router everywhere else, and hit real-world apps with segments like `[lang-country]` holding values such as `fr-fr` / `en-gb`.

### DIFF
--- a/crates/next-core/src/next_root_params/mod.rs
+++ b/crates/next-core/src/next_root_params/mod.rs
@@ -244,7 +244,67 @@ impl ImportMappingReplacement for NextRootParamsMapper {
     }
 }
 
+/// Names that cannot be used as a function declaration name in module code
+/// (always strict): reserved words, strict-mode reserved words, literals, and
+/// the restricted binding names `eval` / `arguments`. They still work as
+/// string module export names (`export { _default as "default" }`).
+const RESERVED_WORDS: &[&str] = &[
+    "break",
+    "case",
+    "catch",
+    "class",
+    "const",
+    "continue",
+    "debugger",
+    "default",
+    "delete",
+    "do",
+    "else",
+    "enum",
+    "export",
+    "extends",
+    "false",
+    "finally",
+    "for",
+    "function",
+    "if",
+    "import",
+    "in",
+    "instanceof",
+    "new",
+    "null",
+    "return",
+    "super",
+    "switch",
+    "this",
+    "throw",
+    "true",
+    "try",
+    "typeof",
+    "var",
+    "void",
+    "while",
+    "with",
+    // strict mode
+    "yield",
+    "let",
+    "static",
+    "await",
+    "implements",
+    "interface",
+    "package",
+    "private",
+    "protected",
+    "public",
+    // restricted binding names in strict mode
+    "eval",
+    "arguments",
+];
+
 fn is_valid_identifier(name: &str) -> bool {
+    if RESERVED_WORDS.contains(&name) {
+        return false;
+    }
     let mut chars = name.chars();
     match chars.next() {
         Some(c) if c.is_ascii_alphabetic() || c == '_' || c == '$' => {}
--- a/crates/next-core/src/next_root_params/mod.rs
+++ b/crates/next-core/src/next_root_params/mod.rs
@@ -149,14 +149,32 @@ impl NextRootParamsMapper {
                     "#,
                 ))
                 .chain(collected_root_params.iter().map(|param_name| {
-                    formatdoc!(
-                        r#"
-                            export function {PARAM_NAME}() {{
-                                return getRootParam('{PARAM_NAME}');
-                            }}
-                        "#,
-                        PARAM_NAME = param_name,
-                    )
+                    if is_valid_identifier(param_name) {
+                        formatdoc!(
+                            r#"
+                                export function {PARAM_NAME}() {{
+                                    return getRootParam('{PARAM_NAME}');
+                                }}
+                            "#,
+                            PARAM_NAME = param_name,
+                        )
+                    } else {
+                        // Param names like `lang-country` are not valid JS
+                        // identifiers, so they cannot be named function
+                        // exports. Declare a safely-named function and
+                        // re-export it under the original param name via a
+                        // string module export name.
+                        formatdoc!(
+                            r#"
+                                function {SAFE_NAME}() {{
+                                    return getRootParam("{PARAM_NAME}");
+                                }}
+                                export {{ {SAFE_NAME} as "{PARAM_NAME}" }};
+                            "#,
+                            SAFE_NAME = safe_identifier(param_name),
+                            PARAM_NAME = param_name,
+                        )
+                    }
                 }))
                 .join("\n")
             };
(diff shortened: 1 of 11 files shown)


Abhirup0 (commented):
Nice work covering all three generators (typegen, Webpack loader, and Turbopack virtual module) alongside e2e typecheck tests.

One issue to address before merging: **JavaScript reserved keywords** (`default`, `class`, `function`, `import`, `delete`, `switch`, etc.) pass the `VALID_IDENTIFIER_REGEX` / `is_valid_identifier` character check, which causes the generator to emit invalid syntax like `export function default()` or `export function class()`.

Reserved words are valid as export specifiers (`export { _class as "class" }`), but cannot be used as function declaration names. 

Could we add a reserved word check to `isValidIdentifier` (TS) and `is_valid_identifier` (Rust) so segments like `app/[default]/` safely route to the string export branch as well?

lcordier42:
@Abhirup0 good catch, thanks — addressed in ab899a2c3f:

- `isValidIdentifier` (TS, shared by the typegen and the webpack loader) and `is_valid_identifier` (Rust, Turbopack virtual module) now also reject reserved words — full list including strict-mode reserved words, literals (`true`/`false`/`null`), and the restricted binding names `eval`/`arguments` — routing them to the string module export branch (`function _default(): ...` + `export { _default as "default" }`).
- The `hyphenated` e2e fixture now has an `app/[lang-country]/[default]/` root layout, so both failure modes (invalid identifier + reserved word) are covered end to end: typecheck via `tsc --noEmit` and runtime rendering through the module namespace (`rootParams.default()`).

lcordier42:
Follow-up in 75affc709c after checking for existing helpers: the Rust side now reuses swc's `Ident::verify_symbol` (already a dependency of next-core, and the same `EsReserved` machinery Turbopack uses elsewhere) instead of a hand-rolled reserved-word list — it covers invalid characters (Unicode-aware), every reserved-word flavor including `eval`/`arguments`, and even supplies the suggested safe symbol, so `RESERVED_WORDS`/`is_valid_identifier`/`safe_identifier` are gone (-101 lines).

The TS helper keeps its explicit list: nothing equivalent is importable from `next/src` without pulling a vendored babel/acorn bundle into the typegen.

PR #97931, opened 2026-08-26
Re-enable AVIF image optimization

eps1lon:
Reverts #97875 and bumps `sharp` from `^0.35.3` to `^0.35.4`, re-enabling AVIF image optimization.

`minimumReleaseAge` interplay (verified with pnpm 10.33.0 and 11.22.0 in a standalone repo with a `file:` dependency declaring sharp as an optional dependency):

- sharp 0.35.4 was published on 2026-08-26 and is younger than this repo's 48h `minimumReleaseAge`.
- pnpm aborts the whole install with `ERR_PNPM_NO_MATURE_MATCHING_VERSION` when no in-range release satisfies the age gate. This applies to `optionalDependencies` (no warn-and-skip) and through `file:` dependencies alike.
- The gate also covers the `@img/sharp-*` binary packages sharp 0.35.4 depends on, so excluding only `sharp@0.35.4` is not enough.
- `pnpm-workspace.yaml` therefore exempts `sharp@0.35.4` and `@img/sharp-*` (pnpm does not allow version qualifiers on name patterns in this setting). Both entries can be dropped once 0.35.4 is older than 48 hours.
- Downstream users who configure their own `minimumReleaseAge` will hit the same hard install failure when installing a release with this bump until 0.35.4 ages out, unless they add the same excludes. Keeping `^0.35.3` is not a safer alternative: age-gated installs would then silently resolve the vulnerable 0.35.3.

Part of https://linear.app/vercel/issue/VOC-34654/

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -38,6 +38,7 @@ minimumReleaseAge: 2880 # 48 hrs
 minimumReleaseAgeExclude:
   # sharp 0.35.4 and its @img/* binaries fix a security vulnerability, so they
   # are exempt from the 48h release age gate.
+  # These exemptions should be removed on 2026-08-28
   - '@img/sharp-darwin-arm64@0.35.4'
   - '@img/sharp-darwin-x64@0.35.4'
   - '@img/sharp-freebsd-wasm32@0.35.4'
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -36,6 +36,34 @@ ignoredBuiltDependencies:
 blockExoticSubdeps: true
 minimumReleaseAge: 2880 # 48 hrs
 minimumReleaseAgeExclude:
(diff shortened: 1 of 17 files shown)


pnpm-workspace.yaml
```diff
@@ -36,6 +36,34 @@ ignoredBuiltDependencies:
 blockExoticSubdeps: true
 minimumReleaseAge: 2880 # 48 hrs
 minimumReleaseAgeExclude:
+  # sharp 0.35.4 and its @img/* binaries fix a security vulnerability, so they
+  # are exempt from the 48h release age gate.
```
styfle, line 40:
```suggestion
  # are exempt from the 48h release age gate.
  # These exemptions should be removed on 2026-08-28
```

styfle (approved):
Looks good, but probably should add a comment in pnpm-workspace.yaml so we remember to remove those exclusions in a couple days

PR #97941, opened 2026-08-26
Fix request-context retention in the default use cache handler

### LINKED ISSUE #97934: Built-in `use cache` handler retains closed ServerResponse objects through cached ReadableStreams

https://github.com/dacgray/next-use-cache-response-retention-repro

### To Reproduce

The reproduction has one PPR route, one `use cache` function, and no fetches,
middleware, observability SDKs, or application dependencies. The route calls
`connection()` below Suspense so every HTTP request invokes the cached function.
Each of 180 stable keys returns a 256 KiB string.

1. Clone the reproduction.
2. Build and run its standalone image:

```sh
docker build -t next-use-cache-response-retention .
docker run --rm --name next-use-cache-response-retention \
  -p 3000:3000 next-use-cache-response-retention
```

3. In another shell, populate 180 keys and replay the same keys twice:

```sh
npm run load
```

The load script waits two seconds and forces three full GCs before every sample.
It tracks each Node `ServerResponse` with a `WeakRef` when the response closes.

Expected: once an HTTP response has closed, the built-in `use cache` handler
should not keep its `ServerResponse` or request AsyncLocalStorage context alive.
Repeated cache hits should not raise the forced-GC heap floor when cache entry
count and serialized bytes are unchanged.

Actual with the built-in handler:

| sample | closed responses still alive | heapUsed | external | arrayBuffers | RSS |
| --- | ---: | ---: | ---: | ---: | ---: |
| baseline | 0 | 23.0 MiB | 3.6 MiB | 0.1 MiB | 80.2 MiB |
| 180 cache fills | 180 | 81.6 MiB | 48.7 MiB | 45.1 MiB | 249.2 MiB |
| same 180 keys, first hit | 360 | 135.1 MiB | 48.7 MiB | 45.1 MiB | 304.9 MiB |
| same 180 keys, second hit | 361 | 135.8 MiB | 48.7 MiB | 45.1 MiB | 310.0 MiB |

The first request leaves one closed response alive per cache key. The first
cache hit leaves a second response alive per key. Further hits stabilize at
roughly two responses per key. New keys grow linearly until LRU eviction: in a
separate run, adding batches of 18 new keys added exactly 18 live responses per
batch; crossing the configured 50 MiB limit released old responses and
stabilized near 200 live responses.

This means the LRU is bounded by serialized cache body size, but the retained
request contexts are not included in that accounting. A small cached value can
therefore pin a much larger render/request graph.

### Caching-preserved control

The reproduction includes a diagnostic handler that implements the same cache
interface but drains entries into detached bytes and creates a fresh
`ReadableStream` for each hit:

```sh
docker run --rm --name next-use-cache-response-retention \
  -p 3000:3000 next-use-cache-response-retention \
  node --expose-gc \
    --require=/app/scripts/detached-cache-handler.cjs \
    --require=/app/scripts/instrument-server-response.cjs \
    server.js
```

Running the identical load gives:

| sample | entries | hits | stored bytes | closed responses still alive | heapUsed |
| --- | ---: | ---: | ---: | ---: | ---: |
| baseline | 0 | 0 | 0 | 0 | 23.0 MiB |
| 180 cache fills | 180 | 0 | 47,191,570 | 1 | 27.5 MiB |
| first replay | 180 | 180 | 47,191,570 | 1 | 28.0 MiB |
| second replay | 180 | 360 | 47,191,570 | 1 | 28.3 MiB |

Caching is therefore still active and the same content remains stored. Only
the live-stream ownership differs.

### Heap retainer evidence

After 18 keys were filled and hit, a forced-GC heap snapshot contained 37 live
`ServerResponse` objects. Eighteen had this strong path:

```text
globalThis[Symbol.for('@next/cache-handlers-map')]
  -> default handler get closure
  -> LRUCache
  -> LRUNode
  -> entry.value
  -> InternalReadableStream
  -> Symbol(nodejs.webstream.isClosedPromise)
  -> Promise
  -> Symbol(kResourceStore)
  -> WorkStore.afterContext
  -> onClose closure
  -> ServerResponse
```

The other 18 went through the compiled `$$RSC_SERVER_CACHE_0` cache graph; one
was a module warm-up response. The detached-byte control eliminates both groups.

The current canary default handler stores a live stream and tees it again on
every hit:

- `get()`: https://github.com/vercel/next.js/blob/d09816fdaf2fef7a23751de75a9dde6bb7c2403b/packages/next/src/server/lib/cache-handlers/default.ts#L134-L148
- `set()`: https://github.com/vercel/next.js/blob/d09816fdaf2fef7a23751de75a9dde6bb7c2403b/packages/next/src/server/lib/cache-handlers/default.ts#L177-L185

The A/B result and retainer path indicate that persisting the live
`ReadableStream`, rather than detached bytes, keeps the request ALS store and
closed response reachable.

### Verify canary release

The latest npm canary at the time of testing is `16.4.0-canary.8`, and it
reproduces in the public standalone Docker image above.

```text
Next.js: 16.4.0-canary.8
React / React DOM: 19.2.6
Node.js: 22.23.2
Platform: linux/arm64 (Docker)
Build: next build --webpack
Output: standalone
cacheComponents: true
cacheMaxMemorySize: 50 * 1024 * 1024
compress: false
```

Runtime, Performance, Cache Components, Output (standalone)

Other (deployed production server)

### Additional context

This may contribute to the production OOM symptoms in #92287, but that report
requires cached internal streamed `fetch()` calls. This reproduction has no
`fetch()` inside the application and isolates the built-in `use cache` handler.

#96369 describes per-render RSC retention with `cacheComponents: false` and a
custom Redis handler; it was auto-closed because it lacked a public reproduction.
The configuration and retainer identified here are different.

dacgray:
Fixes #97934.

The default use cache handler uses live ReadableStream objects. Those streams retain references to the request’s async context and closed HTTP response, causing memory to grow with cached entries and potentially produce OOMs.

The fix converts each stream into a plain Buffer before caching it. Every cache hit then gets a fresh temporary stream created from those bytes. The content remains cached, but completed requests can now be garbage-collected.

---

Basically - the LRU cache stores live ReadableStream objects, which keep completed request contexts and closed responses in memory until that cache is evicted.

### DIFF
--- a/packages/next/src/server/lib/cache-handlers/default.test.ts
+++ b/packages/next/src/server/lib/cache-handlers/default.test.ts
@@ -1,0 +1,96 @@
+/**
+ * @jest-environment node
+ */
+
+import { AsyncLocalStorage } from 'node:async_hooks'
+import { runInNewContext } from 'node:vm'
+import { setFlagsFromString } from 'node:v8'
+import { createDefaultCacheHandler } from './default'
+import type { CacheEntry, CacheHandler } from './types'
+
+setFlagsFromString('--expose-gc')
+const forceGarbageCollection = runInNewContext('gc') as () => void
+
+describe('default use cache handler', () => {
+  it('does not retain the async context that populated an entry', async () => {
+    const handler = createDefaultCacheHandler(1024 * 1024)
+    const requestStoreRef = await runInRequestContext(() =>
+      populateCache(handler)
+    )
+
+    await expectCollected(requestStoreRef)
+
+    expect(await readCache(handler)).toBe('cached value')
+  })
+
+  it('does not retain the async context that read an entry', async () => {
+    const handler = createDefaultCacheHandler(1024 * 1024)
+    await populateCache(handler)
+
+    const requestStoreRef = await runInRequestContext(async () => {
+      expect(await readCache(handler)).toBe('cached value')
+    })
+
+    await expectCollected(requestStoreRef)
+  })
+})
+
+async function runInRequestContext(
+  callback: () => Promise<void>
+): Promise<WeakRef<object>> {
+  const requestStorage = new AsyncLocalStorage<object>()
+  let requestStoreRef: WeakRef<object> | undefined
+
+  await requestStorage.run({ requestId: 'request' }, async () => {
+    const requestStore = requestStorage.getStore()
+
+    if (!requestStore) {
+      throw new Error('Expected a request store')
+    }
+
+    requestStoreRef = new WeakRef(requestStore)
+    await callback()
+  })
+
+  if (!requestStoreRef) {
+    throw new Error('Expected a request store reference')
+  }
+
+  return requestStoreRef
+}
+
+async function populateCache(handler: CacheHandler): Promise<void> {
+  const entry: CacheEntry = {
+    value: new ReadableStream({
+      start(controller) {
+        controller.enqueue(new TextEncoder().encode('cached value'))
+        controller.close()
+      },
+    }),
+    tags: [],
+    stale: 60,
+    timestamp: Date.now(),
+    expire: 120,
+    revalidate: 60,
+  }
+
+  await handler.set('key', Promise.resolve(entry))
+}
+
+async function readCache(handler: CacheHandler): Promise<string> {
+  const entry = await handler.get('key', [])
+  if (!entry) {
+    throw new Error('Expected a cache entry')
+  }
+
+  return new Response(entry.value).text()
+}
+
+async function expectCollected(ref: WeakRef<object>): Promise<void> {
+  for (let attempt = 0; attempt < 10; attempt++) {
+    forceGarbageCollection()
+    await new Promise<void>((resolve) => setImmediate(resolve))
+  }
+
+  expect(ref.deref()).toBeUndefined()
+}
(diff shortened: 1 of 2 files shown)


Abhirup0 (commented):
LGTM. Verified that:
- Converting cached streams to detached `Buffer` instances severs the `AsyncLocalStorage` / `ServerResponse` retainer graph caused by `ReadableStream.prototype.tee()`.
- Stream consumption in `set()` does not interfere with the `use-cache-wrapper` or `tiered-cache-handler` write paths, as both supply independent stream branches.
- LRU sizing via `value.byteLength` maintains accurate memory accounting.
- The `WeakRef` + GC unit test covers both the cache populate and cache hit paths.

unstubbable (approved):
Thanks for the thorough reproduction and the heap retainer path. I verified locally that both new tests fail with `default.ts` reverted on Node 20 and 22, and pass on Node 24 and later, where the retention no longer reproduces. CI runs Node 20.9, so a regression will fail there. A local run on a newer Node will not catch it.

Merging. We'll follow up with a PR documenting the contract in `types.ts`: a handler consumes the entry's value stream exactly once and does not retain it past `set`, whether it buffers the data or pipes it to its store. The same PR will fix the custom cache handler documentation, which still teaches the pattern this removes.

PR #97944, opened 2026-08-26
Turbopack: shorten CSS module class names

mischnic:
The class names were unnecessarily long.

Use the same as the lightningcss default:
`[hash]_[local]` which is
`<hash of the full file path>_<original class name or identifier>`

Keep the previous longer mechanism to aid in debugging in dev


lukesandberg (approved):

lukesandberg:
should we do the 'hash shortening' approach that we do for module ids?

mischnic:
Ideally yes. But LightningCSS's grid transpilation logic places some constraints on the format: https://lightningcss.dev/css-modules.html#custom-naming-patterns

turbopack/crates/turbopack-css/src/process.rs
```diff
@@ -496,14 +499,24 @@ async fn process_content(
...
             CssModuleType::Module => Some(lightningcss::css_modules::Config {
-                pattern: Pattern {
-                    segments: smallvec![
-                        Segment::Name,
-                        Segment::Literal(Cow::Borrowed("__")),
-                        Segment::Hash,
-                        Segment::Literal(Cow::Borrowed("__")),
-                        Segment::Local,
-                    ],
+                pattern: if module_css_debuggable_idents {
```
mischnic: move this inside: `segments: if ...`

turbopack/crates/turbopack/src/module_options/module_options_context.rs
```diff
@@ -317,9 +317,25 @@ pub struct CssOptionsContext {
...
+        Self {
+            enable_raw_css: false,
+            source_maps: Default::default(),
+            module_css_condition: None,
+            lightningcss_features: Default::default(),
+            module_css_debuggable_idents: true,
+            placeholder_for_future_extensions: (),
+        }
+    }
+}
```
mischnic: It should default to false. Then you can use derive again as well

lukesandberg (approved):

PR #97951, opened 2026-08-26
Fail agent runs on unknown config options

devjiwonchoi:
### Why?

Unknown Next.js configuration options currently produce warnings and allow execution to continue. Agents can overlook those warnings and proceed with a misspelled or obsolete configuration.

### How?

When an agent is detected, treat unknown configuration keys as fatal and instruct the agent to inspect the version-matched options in `node_modules/next/dist/`, infer the intended option from repository context, and rerun validation.

Human runs retain the existing warning behavior, and existing option-specific migration errors remain unchanged.

### DIFF
--- a/packages/next/src/server/config.ts
+++ b/packages/next/src/server/config.ts
@@ -57,12 +57,14 @@ import { NextInstanceErrorState } from './mcp/tools/next-instance-error-state'
 import { Bundler } from '../lib/bundler'
 import type { MemoryEvictionMode } from '../build/swc/types'
 import { hrtimeBigIntDurationToString } from '../build/duration-to-string'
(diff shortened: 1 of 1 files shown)


packages/next/src/server/config.ts
```diff
@@ -83,26 +86,41 @@ function normalizeNextConfigZodErrors(
...
     }
 
+    if (
+      isAgent &&
+      issue.code === 'unrecognized_keys' &&
+      !hasSpecificInstruction
+    ) {
+      shouldExit = true
+      message +=
+        '\nInspect the version-matched configuration options in `node_modules/next/dist/`.'
```
devjiwonchoi, line 119: TODO point to NextConfig type

PR #97956, opened 2026-08-26
Turbopack: create worker loaders late so self-spawning workers don't deadlock

### LINKED ISSUE #97399: Turbopack: next build deadlocks at 0% CPU — content-hashed chunk paths + mutually-referencing async chunks form an await cycle

https://github.com/swarupdonepudi/turbopack-chunk-name-cycle

### To Reproduce

About the linked repository (scaffolded from the App Router reproduction template): honest caveat up front — the linked toy app does **not** yet trigger the hang at small scale. The trigger needs the production chunk merger to form a chunk-level reference ring, which we have so far only observed on a large commercial app: ~24k modules, several hundred `import()` call sites. The repository README documents five small-scale shapes that do NOT reproduce and why, which localizes the trigger conditions. Everything below is a source-level root-cause analysis of v16.3.1 with file/line citations, validated empirically on the affected app, including by intervention. We can share capped traces privately and run any diagnostic build you want on the reproducing machine. (A previous filing, #97396 / #97398, was auto-closed by the reproduction-link check because explanatory text sat inside the link section; this issue supersedes both.)

On the affected app (production build, Turbopack, Next 16.3.1):

1. `next build`
2. Compilation proceeds normally for a few minutes, then parks forever at `Creating an optimized production build ...` — 0% CPU across the whole process tree, indefinitely (we capped runs at 25 minutes; users report leaving it for hours). 100% deterministic on this app; also reproduced on 16.1.0.

Current: `next build` deadlocks silently at 0% CPU. Expected: the build completes, or fails with an error naming the circular chunk reference.

### Root cause: content-hashed chunk paths + mutually-referencing async chunks = an await cycle turbo-tasks cannot detect

Three mechanisms combine, each fine in isolation:

**1. A chunk's path awaits its own content.** Production client chunking enables `ContentHashing::Direct` unconditionally ([`next_client/context.rs:645`](https://github.com/vercel/next.js/blob/v16.3.1/crates/next-core/src/next_client/context.rs#L645)). `chunk_path` then derives the file name from a hash of the chunk's own content ([`chunking_context.rs:713-720`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-browser/src/chunking_context.rs#L713-L720)), and `EcmascriptBrowserChunk::path` passes the chunk itself as that asset ([`ecmascript/chunk.rs:172`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-browser/src/ecmascript/chunk.rs#L170-L172)).

**2. A chunk's content awaits other chunks' paths.** Each `import()` compiles to an `AsyncLoaderModule` whose generated code embeds the target chunk group's file names: chunk content/code ([`content.rs:141`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-browser/src/ecmascript/content.rs#L141), [`content.rs:105`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-browser/src/ecmascript/content.rs#L105)) → `item_code_module_ids_and_paths` ([`code_module_ids_and_paths.rs:57`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-ecmascript/src/chunk/code_module_ids_and_paths.rs#L57)) → `module_factory_with_code_generation_issue` → `AsyncLoaderModule::chunk_item_content` with `estimated = false` ([`item.rs:285-286`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-ecmascript/src/chunk/item.rs#L285-L286)) → `chunks_data` ([`async_chunk/module.rs:198`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-ecmascript/src/async_chunk/module.rs#L198)) → `ChunkData::from_assets` → `ChunkData::from_asset`, which awaits `chunk.path()` of the referenced chunks — the closing edge, [`data.rs:65`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-core/src/chunk/data.rs#L65).

**3. turbo-tasks assumes task cycles are impossible.** A task awaiting an in-progress task parks on its `done_event`; there is no cycle detection anywhere in turbo-tasks/turbo-tasks-backend. The assumption is stated at [`backend/mod.rs:1642`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs#L1642): *"it shouldn't be possible to have cycles in tasks"*.

So when the produced chunk graph contains a ring in the relation "chunk X contains an async loader whose target chunk group contains chunk Y" — mutual dynamic imports across merged chunks — every ring member's path transitively awaits itself. Deterministic silent deadlock.

**The existing guard covers sizing but not codegen**: `content_with_async_module_info` documents that `estimated: true` content "should not invoke other chunking operations that would cause cycles" ([`item.rs:248-256`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-ecmascript/src/chunk/item.rs#L248-L256)), and `AsyncLoaderModule` honors it with placeholder content ([`async_chunk/module.rs:171-185`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-ecmascript/src/async_chunk/module.rs#L171-L185)). But `estimated: true` is only passed during chunk-size estimation ([`chunk_type.rs:67`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-ecmascript/src/chunk/chunk_type.rs#L67)); final code generation passes `false`, where the cycle is live.

This also explains why `next dev` is unaffected (no content hashing; ident-based names have no content dependency, [`chunking_context.rs:707-712`](https://github.com/vercel/next.js/blob/v16.3.1/turbopack/crates/turbopack-browser/src/chunking_context.rs#L707-L712)) and why webpack builds the same app fine.

### Empirical validation (instrumented local next-swc builds at v16.3.1, darwin-arm64)

- **Thread inventory at freeze** (macOS `sample` of a `release-with-debug` build): all 16 tokio workers idle in `Context::park_internal`; no thread executes any turbopack code; several worker threads carry `block_in_place<ScopeInner::wait>` blocking-task type names (replacement workers), i.e. the runtime is healthy but has nothing runnable — everything awaits `done_event`s.
- **Execution ledger** (counters at the scheduling edges): `scheduled == start_ok == 6,056,684`, `start_declined == 0`, `completed == 6,056,568` — **exactly 116 executions start and never finish**, identical across independent runs (deterministic, not a race).
- **In-flight registry at freeze**: the 116 are precisely the cycle's cast — `EcmascriptBrowserChunkContent::code`/`::content` (15 each), `ChunkingContext::chunk_path` (15), `ChunkData::from_asset` (12) / `from_assets` (12), `AsyncLoaderModule::chunk_item_content` (12), `module_factory_with_code_generation_issue` (12), `item_code_module_ids_and_paths` (18), plus manifest/emit ancestors. Counts fit rings of ~15 chunks with ~12 async-loader edges.
- **Closure proof**: using the `hanging_detection` cargo feature, the set of tasks whose `done_event` is awaited at maximum backoff at freeze time is exactly the 116 in-flight set — the waiters and the never-finishers are the same set.
- **Intervention proof**: gating out the single `chunk_content_hashing(...)` call at `next_client/context.rs:645` in a local binding makes the identical build complete fully — `scheduled == start_ok == completed == 6,451,147`, zero stuck tasks, compilation green.
- **Ruled out**: #95695 (scope_and_block deadlock fix) was cherry-picked onto v16.3.1 and tested — no effect. PriorityRunner accounting was instrumented — clean (`active=0, queue=0` at freeze). Loader-pool IPC, memory pressure, stale caches, tracing overhead: all ruled out earlier with dedicated runs.

Timeline note: the mechanism is latent in v16.1.0 (same hashing, same `data.rs`), where we also observed hangs. The chunk-merging heuristics changed considerably between 16.1 and 16.3 (e.g. #95102 merging at `overlap == 1`), which plausibly changes whether a given app's graph produces the ring shape.

### Possible fix directions (your call, listed for completeness)

- Cycle-aware chunk naming: derive the hash from content with async-chunk path references excluded/normalized, or fall back to ident-based names for chunks participating in a reference cycle.
- Extend the `estimated`-style placeholder to code generation for cycle members (two-pass naming).
- Await-cycle detection in turbo-tasks that fails the build with a diagnostic instead of hanging silently — valuable regardless of the chunk-naming fix, since it converts any future cycle bug from a silent hang into an actionable error.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.5.0
  Available memory (MB): 49152
  Available CPU cores: 16
Binaries:
  Node: 24.14.0
  npm: 11.9.0
  Yarn: 3.8.7
  pnpm: 9.12.2
Relevant Packages:
  react: 19.2.8
  react-dom: 19.2.8
Next.js Config:
  output: standalone (affected app); N/A (linked repo)
```

Turbopack, Performance

next build (local)

### Additional context

The affected app cannot be shared, but we can: share capped `NEXT_TURBOPACK_TRACING` traces privately (they contain private paths — happy to use the Drive route offered in #92405), run any diagnostic or canary build against the reproducing app, and iterate on a public minimal reproduction with guidance on which merger conditions produce cross-group chunk rings.

sokra:
Fixes #97399

Replaces #97470 (same change, pushed to a branch in this repo so deploy tests run).

## What

Production builds park forever at `Creating an optimized production build ...` at 0% CPU when a web worker spawns another instance of itself — directly, or through a separate module.

## Why

`WorkerLoaderModule::chunk_group` hard-coded `AvailabilityInfo::root()`, so no module was ever "already available" inside a worker chunk group. A worker that spawns itself therefore kept creating a fresh nested chunk group, and because the loader's generated code embeds that chunk group's content-hashed chunk paths, this closed an await ring that turbo-tasks has no cycle detection for:

```
chunk path -> content hash -> chunk content -> worker loader code
-> chunks_data -> chunk_group -> chunk path -> ...
```

`next dev` is unaffected because ident-based chunk names have no content dependency, which is also why webpack builds the same app fine.

## How

Unroll self-referencing workers the same way `import()` cycles already are: build the worker's chunk group with the *enclosing* chunk group's availability info instead of `root()`, so a self-reference sees its own entry as already available and stops recursing.

That requires creating the worker loader once availability is known — during chunking, rather than while building the module graph:

- Move `WorkerType` to `turbopack-core` and add `ChunkingType::Worker`.
- `WorkerAssetReference` resolves to a new `WorkerEntryModule` marker (carrying the inner module, worker type and asset context) instead of building a `WorkerLoaderModule` up front. Tracing-only references keep resolving to the raw module with their previous chunking type.
- `chunk_group_content_operation` collects worker edges into `worker_modules`. This runs before the non-chunkable-node check because the marker is deliberately not chunkable, and reads the module off the edge for the same reason.
- `make_chunk_group` creates the `WorkerLoaderModule` from each marker via a new `ChunkingContext::worker_loader_chunk_item`, passing `new_availability_info`.
- `WorkerLoaderModule` gains an `availability_info` field (excluded from `ident()`, like `AsyncLoaderModule`).

Unlike an async loader, the worker loader must **not** collapse to an empty asset list when its target is available: `parentImport(id)` works for async imports because they share the parent's realm, but a worker gets a fresh realm and still needs its evaluate chunk to instantiate the entry. Module factories are functions and cannot be transferred via `postMessage`, so instead:

- Each browser runtime records the JS chunks whose module factories it installed, exposed as `__turbopack_get_loaded_chunk_paths__`.
- `createWorker` passes those paths in a new params slot, separate from the worker's own chunks.
- The worker bootstrap loads the parent's already-loaded chunks **first** (cheap — the browser has them cached), then the worker's own chunks, then the shared runtime last. `TURBOPACK_NEXT_CHUNK_URLS` stays exactly the reverse of the registration order it is popped in.

Two details worth calling out for review:

- A worker loader has the same module id in every chunk group (its ident deliberately excludes availability info) but carries a different chunk list per group. Loading the worker's own chunks last is what makes a nested worker pick up the correctly-pruned list.
- The `WorkerLoaderModule` is created during chunking, so its `Module::references()` are never traversed. Both of its dependencies — the worker entry and the `createWorker` runtime helper — are therefore declared on the `WorkerEntryModule` marker, and the marker and loader share one memoized function for each so they resolve to identical `Vc`s (and identical module ids).

Content hashing and chunk identity for non-worker modules are unchanged.

## Tests

Adds production regression tests for a worker that spawns itself, both inline and through a separate imported module file — the two repro shapes reported in the issue.

The worker snapshot fixtures are also useful review material: they show the page chunk registering the `createWorker` helper under the same id the loader requires, and the worker-loader chunk being listed in the page evaluate chunk's `otherChunks`.


turbopack/crates/turbopack-core/src/module_graph/chunk_group_info.rs
```diff
@@ -624,9 +624,10 @@ pub async fn compute_chunk_group_info(graph: &ModuleGraph) -> Result<Vc<ChunkGro
                         )),
                         ChunkingType::Isolated {
                             merge_tag: None, ..
-                        } => ChunkGroupInheritance::ChunkGroup(Either::Left(std::iter::once(
-                            ChunkGroupKey::Isolated(node),
-                        ))),
+                        }
+                        | ChunkingType::Worker { .. } => ChunkGroupInheritance::ChunkGroup(
+                            Either::Left(std::iter::once(ChunkGroupKey::Isolated(node))),
```
sokra: ChunkingType::Worker -> ChunkGroupKey::Worker

turbopack/crates/turbopack-ecmascript/js/src/worker/browser/createWorker.ts
```diff
@@ -54,18 +63,49 @@ function createWorker(
...
     )
     .reverse()
+  const preloadUrls = preloadChunkPaths.map((chunkPath) =>
+    __turbopack_chunk_relative_url__(chunkPath, workerBasePath)
+  )
   const params: unknown[] = [
     chunkUrls,
     __turbopack_chunk_asset_suffix__,
     workerBasePath,
+    preloadUrls,
```
sokra: Let's put `preloadUrl` as first item as it's loaded first

mischnic:
There are a lot of failing tests.

PR #97957, opened 2026-08-26
fix(next/image): reject non-2xx internal image responses

### LINKED ISSUE #82357: Unhelpful error message when image loader encounters a redirect response

https://github.com/Erdragh/NextJS-Image-Loader-repro

### To Reproduce

1. Start the application in development mode (`npm run dev`)
2. Open it in a browser
3. Wait for the unoptimized image to load

My current working theory is the following:
1. The NextJS image loader takes the URL and tries to fetch it ([See NextJS code](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/next-server.ts#L788))
2. The NextJS server gets the URL, sees that it does not contain example.com or `localhost` in its `host` header
3. The NextJS server responds with a redirect response, as configured in next.config.ts
4. The image loader tries to get the content type of the image response ([See NextJS code](https://github.com/vercel/next.js/blob/5d899ef5b979ebe025324d16618a6066e7f411af/packages/next/src/server/image-optimizer.ts#L781)), which will be `null`, because the redirect response doesn't contain any info about any image
5. The image loader fails with a log message that contains `null` ([See NextJS code](https://github.com/vercel/next.js/blob/5d899ef5b979ebe025324d16618a6066e7f411af/packages/next/src/server/image-optimizer.ts#L793))

It seems that in 2. the image loader on the server does not set the Host header for local requests, which leads to the redirect happening. I would like some guidance on what to do there but I can disable the redirect in my project for now.

What this issue report is about is the fact that the error doesn't mention the response it got from the server anywhere, which leads to a very unhelpful error message that just says that the image loader received `null`.

If the image loader runs into this problem I would expect an error message more along the lines of
> Image for `.../path/to/image.png` failed, no content type detected from response with status 307

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 24299
  Available CPU cores: 8
Binaries:
  Node: 24.1.0
  npm: 11.3.0
  Yarn: N/A
  pnpm: 8.15.4
Relevant Packages:
  next: 15.4.2-canary.30 // Latest available version is detected (15.4.2-canary.30).
  eslint-config-next: N/A
  react: 19.1.1
  react-dom: 19.1.1
  typescript: 5.9.2
Next.js Config:
  output: N/A
```

Image (next/image)

### Additional context

_No response_

zeeshan56656:
- Fixes https://github.com/vercel/next.js/issues/82357
- Closes https://github.com/vercel/next.js/pull/96985

### What?

`fetchInternalImage()` only rejects the internal response when `statusCode` is falsy, and `MockedResponse` defaults it to `200`, so nothing really gets rejected there. A `307` coming from a `redirects()` entry in `next.config`, or a `404` for a path that isn't there, both count as a successful image fetch.

That body then reaches `detectContentType()`, which returns `null`, and all you get is:

```
The requested resource isn't a valid image for /path/to/image.png received null
```

Nothing in that line points back at the response that actually came in, so a redirect looks exactly like a corrupt image.

### Why?

`fetchExternalImage()` right above already covers this with `if (!res.ok)` and logs `res.status`. The internal path is the only fetch in the file without that check, so the same failure gets reported at the fetch for remote images, and three steps later as "not a valid image" for local ones.

### How?

The guard now rejects anything outside the 2xx range, which is what `res.ok` means, and the log line carries the status next to the href. The thrown `ImageError` message stays generic, same as the external path, so nothing new shows up in the HTTP response body.

Because that throw passes `mocked.res.statusCode` into `ImageError`, a non-2xx internal response now reports its own status instead of always landing on `400`. `ImageError` still maps anything under 400 to `500`, so a redirect answers `500` and a `404` stays `404`. That is what `fetchExternalImage()` has always done, and it lines the two up: a missing local image now returns `404` just like a missing remote one.

Three assertions in `test/e2e/image-optimizer/util.ts` shift because of that:

- `should not forward cookie header`: `/api/conditional-cookie` answers `401` when the cookie is missing, so `400` becomes `401`
- `should error if the image file does not exist`: `/does_not_exist.jpg` returns `404` now, with the internal response message
- `should error if the resource isn't a valid image`: this one requested `/test.txt`, but the fixture in `public/` is `text.txt`, so it was quietly exercising a missing file instead of a non-image file. Pointed at the file that exists, it keeps its original `400` and finally tests what its name says

Two unit tests cover the new guard, one `307` and one `404`. Every other test in that file sets `statusCode = 200`, so they are untouched.

#96985 is open against the same report with a different angle: it threads the status through `ImageUpstream` so the existing "isn't a valid image" log can print it. This one stops the request at the fetch, so the redirect never reaches the optimizer at all. Whichever one fits better, feel free to close mine.

### Verification

- `jest test/unit/image-optimizer/`: 139/139, and the two new tests fail against a build without the guard
- `pnpm test-dev test/e2e/image-optimizer/content-disposition-type.test.ts`: 97/97
- `pnpm test-start test/e2e/image-optimizer/content-disposition-type.test.ts`: 97/97
- `prettier --check` clean on the three changed files

### DIFF
--- a/packages/next/src/server/image-optimizer.ts
+++ b/packages/next/src/server/image-optimizer.ts
@@ -1003,8 +1003,16 @@ export async function fetchInternalImage(
     await handleRequest(mocked.req, mocked.res, parseReqUrl(href))
     await mocked.res.hasStreamed
 
-    if (!mocked.res.statusCode) {
-      Log.error('image response failed for', href, mocked.res.statusCode)
+    if (
+      !mocked.res.statusCode ||
+      mocked.res.statusCode < 200 ||
+      mocked.res.statusCode > 299
+    ) {
+      Log.error(
+        'internal image response failed for',
+        href,
+        mocked.res.statusCode
+      )
       throw new ImageError(
         mocked.res.statusCode,
         '"url" parameter is valid but internal response is invalid'
--- a/test/e2e/image-optimizer/util.ts
+++ b/test/e2e/image-optimizer/util.ts
@@ -419,7 +419,8 @@ export function runTests(ctx: RunTestsCtx) {
     const query = { w: ctx.w, q: ctx.q, url: '/api/conditional-cookie' }
     const opts = { headers: { accept: 'image/webp', cookie: '1' } }
     const res = await next.fetch(`/_next/image?${toQueryString(query)}`, opts)
-    expect(res.status).toBe(400)
+    // the cookie is not forwarded so the internal handler responds 401
+    expect(res.status).toBe(401)
   })
 
   if (ctx.nextConfigImages?.dangerouslyAllowSVG) {
@@ -1615,7 +1616,7 @@ export function runTests(ctx: RunTestsCtx) {
   })
 
   it("should error if the resource isn't a valid image", async () => {
-    const query = { url: '/test.txt', w: ctx.w, q: ctx.q }
+    const query = { url: '/text.txt', w: ctx.w, q: ctx.q }
     const opts = { headers: { accept: 'image/webp' } }
     const res = await next.fetch(`/_next/image?${toQueryString(query)}`, opts)
     expect(res.status).toBe(400)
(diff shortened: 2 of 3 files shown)


styfle (approved):
Looks good, thanks!

I'm surprised we error in the case of redirects but looks like only external images follow redirects, so this should be safe to land.

We can consider adding redirects for internal images in a future PR.

styfle:
Here's a PR to add redirects to local images https://github.com/vercel/next.js/pull/98271

PR #97964, opened 2026-08-26
fix(eslint-plugin): support pageExtensions in no-html-link-for-pages rule

djjrip:
## Bug

The \@next/next/no-html-link-for-pages\ ESLint rule hardcodes file extension matching to only \.js/.jsx/.ts/.tsx\, ignoring the \pageExtensions\ configuration from \
ext.config.js\. This causes the rule to silently miss pages that use custom extensions like \.mdx\ or \.page.tsx\.

Fixes #53473

## Changes

- **\packages/eslint-plugin-next/src/utils/url.ts\**: Added a \uildPageExtRegex\ helper that dynamically constructs file-matching regexes from a \pageExtensions\ array. Updated \parseUrlForPages\, \parseUrlForAppDir\, \getUrlFromPagesDirectories\, and \getUrlFromAppDirectory\ to accept an optional \pageExtensions\ parameter (defaults to \['tsx', 'ts', 'jsx', 'js']\).
- **\packages/eslint-plugin-next/src/rules/no-html-link-for-pages.ts\**: The rule now reads \pageExtensions\ from \
ext.config.js/mjs/ts\ via regex parsing and passes the value through to the URL resolution functions.

## How it works

1. When the ESLint rule initializes, it scans \
ootDirs\ for a \
ext.config.js\ (or \.mjs\/\.ts\ variant)
2. If found, it extracts the \pageExtensions\ array using a regex match
3. The extensions are passed to the URL parsing functions, which use them to build dynamic regexes instead of the previously hardcoded \/(\.(j|t)sx?)$/\
4. If no config is found or \pageExtensions\ is not set, the default extensions are used — maintaining full backward compatibility


djjrip:
Closing - commit accidentally included unrelated file deletions due to a shallow clone checkout issue on Windows. Will resubmit with a clean diff.

PR #97966, opened 2026-08-26
Fix Turbopack HMR recovery after dropped connections

wbinnssmith:
### What?

Track the latest applied Turbopack HMR 'hash' in the browser runtime and include it when subscriptions are restored after reconnecting. Reload the page when the client's hash differs from the server's current hash.

Add an App Router development test that interrupts HMR traffic, misses an update, restores the connection, and verifies the page reloads into the current revision.

### Why?

A browser that temporarily loses its HMR connection can miss updates and remain out of sync after reconnecting.

### How?

Attach the current HMR hash to Turbopack connection and update messages. The runtime records the last hash it processed and sends it with each subscription. The development server compares that value against its current hash and requests a full reload on mismatch.

### DIFF
--- a/packages/next/src/server/dev/hot-reloader-types.ts
+++ b/packages/next/src/server/dev/hot-reloader-types.ts
@@ -65,6 +65,7 @@ export interface ServerErrorMessage {
 export interface TurbopackMessage {
   type: HMR_MESSAGE_SENT_TO_BROWSER.TURBOPACK_MESSAGE
   data: TurbopackUpdate | TurbopackUpdate[]
+  hmrVersion: string
 }
 
 export interface BuildingMessage {
@@ -117,6 +118,7 @@ export interface ReloadPageMessage {
 
 export interface ServerComponentChangesMessage {
   type: HMR_MESSAGE_SENT_TO_BROWSER.SERVER_COMPONENT_CHANGES
+  hmrVersion?: string
 }
 
 /**
@@ -151,7 +153,7 @@ export interface DevPagesManifestUpdateMessage {
 
 export interface TurbopackConnectedMessage {
   type: HMR_MESSAGE_SENT_TO_BROWSER.TURBOPACK_CONNECTED
-  data: { sessionId: number }
+  data: { sessionId: number; hmrVersion: string }
 }
 
 export interface AppIsrManifestMessage {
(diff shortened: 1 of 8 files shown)


packages/next/src/server/dev/hot-reloader-types.ts
```diff
@@ -65,6 +65,7 @@ export interface ServerErrorMessage {
 export interface TurbopackMessage {
   type: HMR_MESSAGE_SENT_TO_BROWSER.TURBOPACK_MESSAGE
   data: TurbopackUpdate | TurbopackUpdate[]
+  hmrHash: string
```
lukesandberg: string|undefined?

test/development/app-hmr/hmr.test.ts
```diff
@@ -11,6 +11,85 @@ describe(`app-dir-hmr`, () => {
...
+
+      const getHmrValue = (browser: Playwright<unknown>) =>
+        browser.eval(`document.querySelector('#hmr-value')?.textContent`)
+
+      try {
+        const browser = await next.browser('/hmr-reconnect', {
+          async beforePageLoad(page) {
+            await page.routeWebSocket(/\/_next\/hmr/, (clientSocket) => {
+              const serverSocket = clientSocket.connectToServer()
+              function connectServerSocket() {
```
lukesandberg: why have this function? just inline it?

test/development/app-hmr/hmr.test.ts
```diff
@@ -11,6 +11,85 @@ describe(`app-dir-hmr`, () => {
   })
 
   describe('filesystem changes', () => {
+    it('replays updates missed while HMR traffic is interrupted', async () => {
```
bgw: Is "replays" the correct term here? We reload the page?

bgw (approved):
Call this `hmrId` instead of `hmrHash` since it's not actually a hash. `id` still implies that it can be stored as a string and that it's a opaque value.

unstubbable (commented):
Hm, did that regress with https://github.com/vercel/next.js/pull/96022? I was under the impression we wouldn't need to send the hash to the client anymore with that approach. 🤔

wbinnssmith:
> Hm, did that regress with #96022? I was under the impression we wouldn't need to send the hash to the client anymore with that approach. 🤔

We're going to need to track and send back the version regardless so we know if we've drifted and need to recover 😕

PR #91335, opened 2026-03-13
fix(build): don't block SSG on telemetry flush, add persistence spans to trace-build

lukesandberg:
### What?

Two fixes for the Turbopack build tracing introduced in #90397:

1. **Don't block SSG on Turbopack shutdown**: `workerMain()` no longer awaits the shutdown promise before returning. Trace event collection is deferred to `waitForShutdown()`, which the parent process awaits *after* SSG completes. This allows static generation and Turbopack persistence/cache-flush to run in parallel.

2. **Add persistence spans to `trace-build` allowlist**: `turbopack-build-events`, `turbopack-persistence`, and `turbopack-compaction` are now included in the `to-json-build.ts` allowlist so they appear in `.next/trace-build`.

### Why?

- The `await shutdownPromise` in `workerMain()` was too eager — it prevented the caller from acknowledging the build as complete and starting SSG until Turbopack persistence finished flushing to disk.
- The persistence/compaction spans emitted by Rust (`turbopack-persistence`, `turbopack-compaction`) were not in the `to-json-build.ts` allowlist, so they were silently filtered out of `.next/trace-build`.

### How?

**`impl.ts` (worker)**:
- Removed `await shutdownPromise` from `workerMain()` — it now returns build results immediately
- `waitForShutdown()` now returns `{ debugTraceEvents }` after awaiting shutdown, so trace events are collected only after all compilation events (including persistence spans) have been processed

**`index.ts` (parent)**:
- Moved `recordTraceEvents(debugTraceEvents)` from the `workerMain` result handler into the `shutdownPromise` `.then()` chain, so events are replayed into the parent reporter after shutdown completes

**`to-json-build.ts`**:
- Added `turbopack-build-events`, `turbopack-persistence`, `turbopack-compaction` to the allowlist

**Test updates**:
- Enabled `turbopackFileSystemCacheForBuild: true` in the trace-build test fixture
- Updated the Turbopack inline snapshot to include `turbopack-build-events`

PR #49092, opened 2023-05-02
[edge] bump @edge-runtime/* package suite

Schniz:
Bumping `@edge-runtime/*` <picture data-single-emoji=":edge-runtime:" title=":edge-runtime:">:edge-runtime:</picture>  packages to their latest beta, as we had some huge improvements and big changes on how it works internally.

* Using the latest `undici` which provides a WebSocket implementation
* Less polyfill usage as we load modules in the Node.js realm, and can reuse modules from Node.js as time goes on <picture data-single-emoji=":just-right2:" title=":just-right2:">:just-right2:</picture> 
* `instanceof` checks now work within the realm <picture data-single-emoji=":mind_blown:" title=":mind_blown:">:mind_blown:</picture>

PR #97981, opened 2026-08-27
fix(telemetry): flush telemetry detached on next build

### LINKED ISSUE #97929: `next build` waits for telemetry submission before it returns

https://github.com/refirst11/nextjs-turbopack-feature-usage-repro

### To Reproduce

```shell
git clone https://github.com/refirst11/nextjs-turbopack-feature-usage-repro
cd nextjs-turbopack-feature-usage-repro
pnpm install
npx next telemetry status   # must report "Enabled"
pnpm benchmark
```

Current:
- packages/next/src/build/index.ts → await telemetry.flush()
- packages/next/src/cli/next-dev.ts and server/lib/start-server.ts → flushDetached
- `flushDetached` is typed `(mode: 'dev', dir: string) => void`, detached-flush.ts throws when `mode !== 'dev'`
- The success or failure of a POST request does not affect the contents of .next.
- [CI run](https://github.com/refirst11/nextjs-turbopack-feature-usage-repro/actions/runs/32959980062)

Difference in total `next build` wall time between telemetry enabled and
`NEXT_TELEMETRY_DISABLED=1`. GitHub-hosted runners, 6 runs per arm.

| Node | 16.3.3    | 16.4.0-canary.8 |
|------|-----------|-----------------|
| 22   | 0.316 s   | 0.181 s         |
| 23   | 0.433 s   | 0.371 s         |
| 24   | 1.085 s   | 1.102 s         |
| 25   | 1.004 s   | 1.017 s         |
| 26   | 1.090 s   | 1.020 s         |

Expected:
`next build` should not wait for telemetry submission to complete, as the dev server already does not.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.5.0: Tue Jun  9 22:26:46 PDT 2026; root:xnu-12377.121.10~1/RELEASE_ARM64_T8103
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 25.8.2
  npm: 11.11.1
  Yarn: N/A
  pnpm: 11.3.0
Relevant Packages:
  next: 16.4.0-canary.8 // Latest available version is detected (16.4.0-canary.8).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Turbopack

next build (local)

### Additional context

Runner:
Local machine and GitHub Actions

Abhirup0:
### What?

Switches `next build` telemetry flushing from blocking synchronous `await telemetry.flush()` to non-blocking `telemetry.flushDetached('build', dir)`.

Fixes #97929.

### Why?

`next build` previously awaited `telemetry.flush()` synchronously on the critical exit path, blocking process completion on outbound HTTP requests to the telemetry endpoint. This added 200ms–1.1s of latency to build times across environments. `next dev` and server shutdown already use `flushDetached` to prevent blocking process exit.

### How?

1. `packages/next/src/telemetry/storage.ts`: update `flushDetached` to accept `mode: 'dev' | 'build'` and clear `this.queue` after draining events.
2. `packages/next/src/telemetry/detached-flush.ts`: accept `build` mode and load config with `PHASE_PRODUCTION_BUILD` when running in build mode.
3. `packages/next/src/build/index.ts`: replace blocking `await telemetry.flush()` with `telemetry.flushDetached('build', dir)`.
4. `packages/next/src/telemetry/detached-flush.test.ts`: add unit tests covering detached telemetry spawning and queue handling for `build` and `dev` modes.

### DIFF
--- a/packages/next/src/build/index.ts
+++ b/packages/next/src/build/index.ts
@@ -1363,7 +1363,6 @@ export default async function build(
         const errorMessage =
           'The "exportPathMap" configuration cannot be used with the "app" directory. Please use generateStaticParams() instead.'
         Log.error(errorMessage)
-        await telemetry.flush()
         throw new Error(errorMessage)
       }
 
@@ -1597,7 +1596,6 @@ export default async function build(
           for (const [pagePath, appPath] of conflictingAppPagePaths) {
             Log.error(`  "${pagePath}" - "${appPath}"`)
           }
-          await telemetry.flush()
           throw new Error(errorMessage)
         }
       }
@@ -4645,9 +4643,9 @@ export default async function build(
           )
       }
 
-      await nextBuildSpan
+      nextBuildSpan
         .traceChild('telemetry-flush')
-        .traceAsyncFn(() => telemetry.flush())
+        .traceFn(() => telemetry.flushDetached('build', dir))
 
       await shutdownPromise
 
@@ -4701,7 +4699,7 @@ export default async function build(
     // Flush telemetry before finishing (waits for async operations like setTimeout in debug mode)
     const telemetry: Telemetry | undefined = traceGlobals.get('telemetry')
     if (telemetry) {
-      await telemetry.flush()
+      telemetry.flushDetached('build', dir)
     }
 
     // Ensure all buffered spans are on disk before `uploadTrace` reads the file.
--- a/packages/next/src/telemetry/detached-flush.test.ts
+++ b/packages/next/src/telemetry/detached-flush.test.ts
@@ -26,10 +26,13 @@ describe('Telemetry flushDetached', () => {
 
   it('spawns detached process with "build" mode when events are queued', async () => {
     const telemetry = new Telemetry({ distDir })
-    telemetry.record({
-      eventName: 'NEXT_TEST_EVENT' as any,
-      fields: { feature: 'test' },
-    })
+    telemetry.record(
+      {
+        eventName: 'NEXT_TEST_EVENT',
+        payload: { feature: 'test' },
+      },
+      true
+    )
 
     telemetry.flushDetached('build', tmpDir)
 
@@ -52,10 +55,13 @@ describe('Telemetry flushDetached', () => {
 
   it('spawns detached process with "dev" mode when events are queued', async () => {
     const telemetry = new Telemetry({ distDir })
-    telemetry.record({
-      eventName: 'NEXT_DEV_EVENT' as any,
-      fields: { feature: 'dev' },
-    })
+    telemetry.record(
+      {
+        eventName: 'NEXT_DEV_EVENT',
+        payload: { feature: 'dev' },
+      },
+      true
+    )
 
     telemetry.flushDetached('dev', tmpDir)
 
@@ -68,16 +74,22 @@ describe('Telemetry flushDetached', () => {
 
   it('generates unique event files for consecutive flushes without collisions', async () => {
     const telemetry = new Telemetry({ distDir })
-    telemetry.record({
-      eventName: 'NEXT_EVENT_1' as any,
-      fields: { step: 1 },
-    })
+    telemetry.record(
+      {
+        eventName: 'NEXT_EVENT_1',
+        payload: { step: 1 },
+      },
+      true
+    )
     telemetry.flushDetached('build', tmpDir)
 
-    telemetry.record({
-      eventName: 'NEXT_EVENT_2' as any,
-      fields: { step: 2 },
-    })
+    telemetry.record(
+      {
+        eventName: 'NEXT_EVENT_2',
+        payload: { step: 2 },
+      },
+      true
+    )
     telemetry.flushDetached('build', tmpDir)
 
     expect(spawnSpy).toHaveBeenCalledTimes(2)
@@ -101,10 +113,13 @@ describe('Telemetry flushDetached', () => {
 
   it('clears queue and does not re-spawn on subsequent flushDetached calls', async () => {
     const telemetry = new Telemetry({ distDir })
-    telemetry.record({
-      eventName: 'NEXT_BUILD_EVENT' as any,
-      fields: { feature: 'build' },
-    })
+    telemetry.record(
+      {
+        eventName: 'NEXT_BUILD_EVENT',
+        payload: { feature: 'build' },
+      },
+      true
+    )
 
     telemetry.flushDetached('build', tmpDir)
     expect(spawnSpy).toHaveBeenCalledTimes(1)
@@ -112,6 +127,23 @@ describe('Telemetry flushDetached', () => {
     // Second call without new events should do nothing
     telemetry.flushDetached('build', tmpDir)
     expect(spawnSpy).toHaveBeenCalledTimes(1)
+  })
+
+  it('aborts in-flight telemetry requests when flushDetached is called', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_IN_FLIGHT_EVENT',
+      payload: { feature: 'in-flight' },
+    })
+
+    telemetry.flushDetached('build', tmpDir)
+
+    expect(spawnSpy).toHaveBeenCalledTimes(1)
+    const eventsFile = path.join(distDir, spawnSpy.mock.calls[0][1][3])
+    expect(fs.existsSync(eventsFile)).toBe(true)
+    const writtenEvents = JSON.parse(fs.readFileSync(eventsFile, 'utf8'))
+    expect(writtenEvents).toHaveLength(1)
+    expect(writtenEvents[0].eventName).toBe('NEXT_IN_FLIGHT_EVENT')
   })
 
   it('does nothing when queue is empty', () => {
--- a/packages/next/src/telemetry/detached-flush.test.ts
+++ b/packages/next/src/telemetry/detached-flush.test.ts
@@ -14,7 +14,7 @@ describe('Telemetry flushDetached', () => {
     distDir = path.join(tmpDir, '.next')
     spawnSpy = jest
       .spyOn(childProcess, 'spawn')
-      .mockImplementation(() => ({} as any))
+      .mockImplementation(() => ({}) as any)
   })
 
   afterEach(() => {
@@ -131,13 +131,19 @@ describe('Telemetry flushDetached', () => {
 
   it('aborts in-flight telemetry requests when flushDetached is called', async () => {
     const telemetry = new Telemetry({ distDir })
-    telemetry.record({
-      eventName: 'NEXT_IN_FLIGHT_EVENT',
-      payload: { feature: 'in-flight' },
-    })
+    const recordPromise = telemetry.record(
+      {
+        eventName: 'NEXT_IN_FLIGHT_EVENT',
+        payload: { feature: 'in-flight' },
+      },
+      true
+    )
+
+    expect((recordPromise as any)._controller.signal.aborted).toBe(false)
 
     telemetry.flushDetached('build', tmpDir)
 
+    expect((recordPromise as any)._controller.signal.aborted).toBe(true)
     expect(spawnSpy).toHaveBeenCalledTimes(1)
     const eventsFile = path.join(distDir, spawnSpy.mock.calls[0][1][3])
     expect(fs.existsSync(eventsFile)).toBe(true)
--- a/packages/next/src/telemetry/detached-flush.test.ts
+++ b/packages/next/src/telemetry/detached-flush.test.ts
@@ -1,0 +1,90 @@
+import fs from 'fs'
+import path from 'path'
+import os from 'os'
+import childProcess from 'child_process'
+import { Telemetry } from './storage'
+
+describe('Telemetry flushDetached', () => {
+  let tmpDir: string
+  let distDir: string
+  let spawnSpy: jest.SpyInstance
+
+  beforeEach(() => {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'next-telemetry-test-'))
+    distDir = path.join(tmpDir, '.next')
+    spawnSpy = jest
+      .spyOn(childProcess, 'spawn')
+      .mockImplementation(() => ({} as any))
+  })
+
+  afterEach(() => {
+    spawnSpy.mockRestore()
+    try {
+      fs.rmSync(tmpDir, { recursive: true, force: true })
+    } catch {}
+  })
+
+  it('spawns detached process with "build" mode when events are queued', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_TEST_EVENT' as any,
+      fields: { feature: 'test' },
+    })
+
+    telemetry.flushDetached('build', tmpDir)
+
+    expect(spawnSpy).toHaveBeenCalledTimes(1)
+    const [execPath, args, options] = spawnSpy.mock.calls[0]
+
+    expect(execPath).toBe(process.execPath)
+    expect(args[0]).toContain('detached-flush')
+    expect(args[1]).toBe('build')
+    expect(args[2]).toBe(tmpDir)
+    expect(args[3]).toMatch(/^_events_\d+\.json$/)
+    expect(options.detached).toBe(true)
+
+    const eventsFile = path.join(distDir, args[3])
+    expect(fs.existsSync(eventsFile)).toBe(true)
+    const writtenEvents = JSON.parse(fs.readFileSync(eventsFile, 'utf8'))
+    expect(writtenEvents).toHaveLength(1)
+    expect(writtenEvents[0].eventName).toBe('NEXT_TEST_EVENT')
+  })
+
+  it('spawns detached process with "dev" mode when events are queued', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_DEV_EVENT' as any,
+      fields: { feature: 'dev' },
+    })
+
+    telemetry.flushDetached('dev', tmpDir)
+
+    expect(spawnSpy).toHaveBeenCalledTimes(1)
+    const [, args] = spawnSpy.mock.calls[0]
+    expect(args[1]).toBe('dev')
+    expect(args[2]).toBe(tmpDir)
+  })
+
+  it('clears queue and does not re-spawn on subsequent flushDetached calls', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_BUILD_EVENT' as any,
+      fields: { feature: 'build' },
+    })
+
+    telemetry.flushDetached('build', tmpDir)
+    expect(spawnSpy).toHaveBeenCalledTimes(1)
+
+    // Second call without new events should do nothing
+    telemetry.flushDetached('build', tmpDir)
+    expect(spawnSpy).toHaveBeenCalledTimes(1)
+  })
+
+  it('does nothing when queue is empty', () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.flushDetached('build', tmpDir)
+
+    expect(spawnSpy).not.toHaveBeenCalled()
+    expect(fs.existsSync(distDir)).toBe(false)
+  })
+})
--- a/packages/next/src/telemetry/detached-flush.test.ts
+++ b/packages/next/src/telemetry/detached-flush.test.ts
@@ -40,7 +40,7 @@ describe('Telemetry flushDetached', () => {
     expect(args[0]).toContain('detached-flush')
     expect(args[1]).toBe('build')
     expect(args[2]).toBe(tmpDir)
-    expect(args[3]).toMatch(/^_events_\d+\.json$/)
+    expect(args[3]).toMatch(/^_events_\d+_[a-f0-9]+\.json$/)
     expect(options.detached).toBe(true)
 
     const eventsFile = path.join(distDir, args[3])
@@ -63,6 +63,40 @@ describe('Telemetry flushDetached', () => {
     const [, args] = spawnSpy.mock.calls[0]
     expect(args[1]).toBe('dev')
     expect(args[2]).toBe(tmpDir)
+    expect(args[3]).toMatch(/^_events_\d+_[a-f0-9]+\.json$/)
+  })
+
+  it('generates unique event files for consecutive flushes without collisions', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_EVENT_1' as any,
+      fields: { step: 1 },
+    })
+    telemetry.flushDetached('build', tmpDir)
+
+    telemetry.record({
+      eventName: 'NEXT_EVENT_2' as any,
+      fields: { step: 2 },
+    })
+    telemetry.flushDetached('build', tmpDir)
+
+    expect(spawnSpy).toHaveBeenCalledTimes(2)
+    const firstFile = spawnSpy.mock.calls[0][1][3]
+    const secondFile = spawnSpy.mock.calls[1][1][3]
+
+    expect(firstFile).not.toBe(secondFile)
+    expect(fs.existsSync(path.join(distDir, firstFile))).toBe(true)
+    expect(fs.existsSync(path.join(distDir, secondFile))).toBe(true)
+
+    const firstEvents = JSON.parse(
+      fs.readFileSync(path.join(distDir, firstFile), 'utf8')
+    )
+    const secondEvents = JSON.parse(
+      fs.readFileSync(path.join(distDir, secondFile), 'utf8')
+    )
+
+    expect(firstEvents[0].eventName).toBe('NEXT_EVENT_1')
+    expect(secondEvents[0].eventName).toBe('NEXT_EVENT_2')
   })
 
   it('clears queue and does not re-spawn on subsequent flushDetached calls', async () => {
--- a/packages/next/src/telemetry/detached-flush.ts
+++ b/packages/next/src/telemetry/detached-flush.ts
@@ -57,7 +57,9 @@ import {
   // finished flushing events clean-up
   try {
     fs.unlinkSync(eventsPath)
-  } catch (_) {}
+  } catch (_) {
+    // the file may already be gone; nothing to clean up
+  }
   // Don't call process.exit() here - let Node.js exit naturally after
   // all pending work completes (e.g., setTimeout in debug telemetry)
 })()
--- a/packages/next/src/telemetry/detached-flush.ts
+++ b/packages/next/src/telemetry/detached-flush.ts
@@ -34,7 +34,9 @@ import {
   // Support both old format (no eventsFile arg) and new format (with eventsFile arg)
   const eventsPath = path.join(
     distDir,
-    eventsFile && !eventsFile.includes('/') ? eventsFile : '_events.json'
+    eventsFile && !eventsFile.includes('/') && !eventsFile.includes('\\')
+      ? eventsFile
+      : '_events.json'
   )
 
   let events: TelemetryEvent[]
@@ -53,7 +55,9 @@ import {
   await telemetry.flush()
 
   // finished flushing events clean-up
-  fs.unlinkSync(eventsPath)
+  try {
+    fs.unlinkSync(eventsPath)
+  } catch (_) {}
   // Don't call process.exit() here - let Node.js exit naturally after
   // all pending work completes (e.g., setTimeout in debug telemetry)
 })()
--- a/packages/next/src/telemetry/detached-flush.ts
+++ b/packages/next/src/telemetry/detached-flush.ts
@@ -31,7 +31,7 @@ import {
     mode === 'build' ? PHASE_PRODUCTION_BUILD : PHASE_DEVELOPMENT_SERVER,
     dir
   )
-  if (hasCustomExportOutput(config)) {
+  if (mode === 'build' && hasCustomExportOutput(config)) {
     config.distDir = '.next'
   }
   const distDir = path.join(dir, config.distDir || '.next')
--- a/packages/next/src/telemetry/detached-flush.ts
+++ b/packages/next/src/telemetry/detached-flush.ts
@@ -4,10 +4,13 @@ import type { TelemetryEvent } from './storage'
 import { Telemetry } from './storage'
 import loadConfig from '../server/config'
 import { getProjectDir } from '../lib/get-project-dir'
-import { PHASE_DEVELOPMENT_SERVER } from '../shared/lib/constants'
+import {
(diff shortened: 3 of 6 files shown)


Abhirup0:
Addressed the concurrency collision in 39c191d by suffixing _events__.json with a unique random hex token per flush invocation. This ensures multiple consecutive or failure-path flushes within the same process write to isolated event files without overwriting or race conditions. Added unit test coverage verifying distinct event file generation across consecutive flushes.

packages/next/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      fs.rmSync(tmpDir, { recursive: true, force: true })
+    } catch {}
+  })
+
+  it('spawns detached process with "build" mode when events are queued', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_TEST_EVENT' as any,
+      fields: { feature: 'test' },
+    })
```
refirst11:
```suggestion
    telemetry.record(
      {
        eventName: 'NEXT_TEST_EVENT',
        payload: { feature: 'test' },
      },
      true
    )
```

Two changes here: `fields` -> `payload` to match `TelemetryEvent`, and `deferred: true` so the test doesn't fire a live telemetry request. The `as any` is no longer needed either — `eventName` is typed as plain `string`.

packages/next/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect(writtenEvents).toHaveLength(1)
+    expect(writtenEvents[0].eventName).toBe('NEXT_TEST_EVENT')
+  })
+
+  it('spawns detached process with "dev" mode when events are queued', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_DEV_EVENT' as any,
+      fields: { feature: 'dev' },
+    })
```
refirst11:
```suggestion
    telemetry.record(
      {
        eventName: 'NEXT_DEV_EVENT',
        payload: { feature: 'dev' },
      },
      true
    )
```

Same here.

packages/next/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect(args[2]).toBe(tmpDir)
+    expect(args[3]).toMatch(/^_events_\d+_[a-f0-9]+\.json$/)
+  })
+
+  it('generates unique event files for consecutive flushes without collisions', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_EVENT_1' as any,
+      fields: { step: 1 },
+    })
```
refirst11:
```suggestion
    telemetry.record(
      {
        eventName: 'NEXT_EVENT_1',
        payload: { step: 1 },
      },
      true
    )
```

Same here.

packages/next/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    telemetry.flushDetached('build', tmpDir)
+
+    telemetry.record({
+      eventName: 'NEXT_EVENT_2' as any,
+      fields: { step: 2 },
+    })
```
refirst11:
```suggestion
    telemetry.record(
      {
        eventName: 'NEXT_EVENT_2',
        payload: { step: 2 },
      },
      true
    )
```

Same here.

refirst11 (commented):
Thanks for picking this up — I filed #97929, and I'm glad to see it moving. Reusing the existing `flushDetached` path instead of adding a new mechanism is exactly the shape I was hoping for.

A few mechanical things in the new test file:

**`fields:` should be `payload:`.** `TelemetryEvent` is `{ eventName: string; payload: object }` (`storage.ts:32`). `packages/next/tsconfig.json` sets `"strict": true`, lists `jest` in `types`, and has no `include`, so `src/**/*.test.ts` is type-checked — this should fail `tsc` on the missing `payload` plus the excess `fields`. It also means the events written to disk carry `payload: undefined`, so the round-trip the tests are meant to cover isn't actually exercised.

**`record()` should pass `deferred: true`.** Without it, each test starts a real `submitRecord()` — an actual POST to `telemetry.nextjs.org` unless `NEXT_TELEMETRY_DISABLED` happens to be set in the environment. It also diverges from the production path: `cli/next-dev.ts:154-163` passes `true` immediately before `flushDetached`, precisely so the request isn't started just to be aborted.

There's a second effect specific to `clears queue and does not re-spawn on subsequent flushDetached calls`: with `deferred: false`, `record()`'s own `.then()` already removes the promise from `this.queue` (`storage.ts:203-209`), so that test can pass even if the new `queue.clear()` were removed. Passing `true` keeps the entry queued and makes the test actually cover the change.

Suggestions inline for all five call sites, plus one nit in `detached-flush.ts`.

Separately, I have measurements from the original issue pointing at a second blocking point this PR doesn't cover yet. I'll write that up as its own comment so it doesn't clutter this thread.

packages/next/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect(firstEvents[0].eventName).toBe('NEXT_EVENT_1')
+    expect(secondEvents[0].eventName).toBe('NEXT_EVENT_2')
+  })
+
+  it('clears queue and does not re-spawn on subsequent flushDetached calls', async () => {
+    const telemetry = new Telemetry({ distDir })
+    telemetry.record({
+      eventName: 'NEXT_BUILD_EVENT' as any,
+      fields: { feature: 'build' },
+    })
```
refirst11:
```suggestion
    telemetry.record(
      {
        eventName: 'NEXT_BUILD_EVENT',
        payload: { feature: 'build' },
      },
      true
    )
```

Same here. This one is also the test that most depends on `deferred: true` — see the review body.

packages/next/src/telemetry/detached-flush.ts
```diff
@@ -47,7 +55,9 @@ import { PHASE_DEVELOPMENT_SERVER } from '../shared/lib/constants'
   await telemetry.flush()
 
   // finished flushing events clean-up
-  fs.unlinkSync(eventsPath)
+  try {
+    fs.unlinkSync(eventsPath)
+  } catch (_) {}
```
refirst11:
```suggestion
  // finished flushing events clean-up
  try {
    fs.unlinkSync(eventsPath)
  } catch (_) {
    // the file may already be gone; nothing to clean up
  }
```

Nit — matches the empty-catch style used in `storage.ts:233-235`.

packages/next/src/build/index.ts
```diff
@@ -4632,9 +4630,9 @@ export default async function build(
           )
       }
 
-      await nextBuildSpan
+      nextBuildSpan
         .traceChild('telemetry-flush')
-        .traceAsyncFn(() => telemetry.flush())
+        .traceFn(() => telemetry.flushDetached('build', dir))
```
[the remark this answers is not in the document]
refirst11: Confirming this one — it reproduces from a read of the code, and I think the root cause is worth naming because it changes what the fix should be.

`flushDetached` is *designed* to cancel the inline request. `storage.ts:180-182` says so explicitly: "if we know we are going to immediately call `flushDetached` we can skip starting the initial `submitRecord` **which will then be cancelled**".

That cancellation is dead code. `storage.ts:231` calls `item._controller?.abort()`, but `_controller` is always `undefined`, because `record()` ends with:

```ts
;(prom as any)._controller = (prom as any)._controller
```

a self-assignment. The real controller is attached inside `submitRecord` (`res._controller = postController`, line 336), but `submitRecord` is `async`, so `res` is wrapped in another promise and `record()` chains off the wrapper — the controller never reaches `prom`.

So the inline POST runs to completion and the detached child submits the same events again. Which events are affected depends on whether their request had already settled, since settled ones are removed from the queue at `storage.ts:203-209` — that's where the non-determinism comes from.

`next dev` doesn't hit this because `cli/next-dev.ts:154-163` passes `deferred: true` immediately before `flushDetached`, so no inline request is ever started. `next build` has no equivalent.

On the fix: adding `deferred: true` at the build's `record()` call sites is a larger surface than it looks — there are 12 in `build/index.ts` alone, plus more feeding the same instance from `build/webpack/plugins/middleware-plugin.ts`, `build/analyze/index.ts`, `build/type-check.ts`, `build/after-production-compile.ts`, and `export/index.ts`. Repairing `_controller` in `record()` covers all of them in one place:

```ts
  record = (_events, deferred?) => {
    const controller = new AbortController()
    const prom = (deferred ? /* … */ : this.submitRecord(_events, controller))
      // …
    ;(prom as any)._controller = controller
```

with `submitRecord(_events, controller?)` using the passed-in controller instead of creating its own.

That still leaves a narrow race — a request already received by the server can't be un-sent — so `deferred: true` is worth adding at the few sites that record immediately before the flush (`build/index.ts:4286-4327` and `4666`). But it isn't needed at all of them.

refirst11:
Following up on the "second blocking point" I mentioned in my review.

**`next build` doesn't call `process.exit()` on success.** `cli/next-build.ts` only exits explicitly on SIGTERM/SIGINT, so the process ends when the event loop drains. That matters here: moving the flush off the `await` stops the build from *measuring* the wait, but anything still holding a handle keeps the process alive regardless. Two things do.

### 1. The build workers still await their own flush

`build/turbopack-build/impl.ts:372-377`:

```ts
  } finally {
    // Always flush telemetry before worker exits (waits for async operations like setTimeout in debug mode)
    await telemetry.flush()
    await saveCpuProfile()
  }
```

The same `await telemetry.flush()` sits on the webpack worker's return path at `build/webpack-build/impl.ts:460`. Each worker constructs its own `Telemetry` (`turbopack-build/impl.ts:352-354`), so the parent's `flushDetached` doesn't cover it, and `workerMain()` doesn't resolve until that flush completes — inside the `run-turbopack` span.

That's the larger half of the cost. Comparing telemetry-enabled against `NEXT_TELEMETRY_DISABLED=1` on the repro from #97929, via the `.next/trace` spans:

| span | enabled | disabled | delta |
|---|---|---|---|
| `next-build` | 4.272s | 2.761s | +1.511s |
| `run-turbopack` | 2.267s | 1.492s | **+0.775s** |
| `telemetry-flush` | 0.493s | 0.000s | +0.493s |

The `run-turbopack` delta isn't build work. Instrumenting inside `turbopackBuild()` showed it returning at 1.781s vs 1.719s worker uptime across the two arms — the ~0.7s appears *after* the build work finishes, on the exit path.

Two caveats on these numbers: they're local, on Node 25, and the CI table in #97929 shows the same comparison is much smaller on Node 22 (0.18–0.32s vs ~1.0s on Node 24+). So please read the *split between the two spans* rather than the absolute values — the point is that roughly half the cost sits in the worker, which this PR doesn't touch.

The substitution is the one you already made:

```ts
  } finally {
    telemetry.flushDetached('build', NextBuildContext.dir!)
    await saveCpuProfile()
  }
```

One thing it needs alongside it: the worker constructs `Telemetry` with a **relative** `distDir` (`impl.ts:352-354` passes `NextBuildContext.config.distDir`; compare `build/index.ts`, which does `path.join(dir, config.distDir)`). `flushDetached` writes the events file to `this.distDir`, while `detached-flush.ts` resolves it as `path.join(dir, config.distDir)`. Under `next build ./some-dir` those two differ, and the child exits 0 on ENOENT — silent loss. So it would want:

```ts
  const telemetry = new Telemetry({
    distDir: path.join(NextBuildContext.dir!, NextBuildContext.config.distDir),
  })
```

(`path` is already imported there.)

Worth noting that your random-suffix fix in 39c191d is what makes two flushes safe here: the worker is created with `enableWorkerThreads: true` (`turbopack-build/index.ts:15`), so it can share `process.pid` with the parent, and the two would otherwise have collided on `_events_<pid>.json`.

### 2. The in-flight request is never aborted

Details in my reply on the bot's `build/index.ts:4635` thread. `flushDetached`'s `item._controller?.abort()` is a no-op because the controller never reaches the queued promise, so the socket stays open and the process can't drain until the request settles. The same dead abort is what produces the double-counting the bot flagged, so one fix in `record()` addresses both.

Happy to open either of these as its own PR if you'd rather keep this one scoped — just say which you'd prefer. My read is that the numbers in #97929 won't move much until the worker flush is included as well.

Abhirup0:
Thanks so much for the deep dive and the span breakdown @refirst11 , that analysis makes the worker exit bottleneck and the controller propagation issue crystal clear.

I went ahead and included both fixes directly in this PR since they share the same root objective and rely on the same detached flushing mechanism.

### Summary of updates in [`1103cb2`](https://github.com/vercel/next.js/commit/1103cb2419):

1. **Worker flush & absolute `distDir`:**
   - Switched both `turbopack-build` and `webpack-build` workers from `await telemetry.flush()` to `telemetry.flushDetached('build', NextBuildContext.dir!)`.
   - Updated worker `Telemetry` construction to use `path.join(dir, config.distDir)` so sub-directory builds (`next build ./custom-dir`) resolve the `.next` output directory correctly in detached workers.

2. **`AbortController` wiring & in-flight cancellation:**
   - Updated `record()` in `storage.ts` to instantiate `controller` upfront, pass `controller.signal` into `submitRecord()`, and attach `_controller` to `prom`.
   - When `flushDetached()` runs, `item._controller?.abort()` now cleanly cancels active `fetch()` requests and frees the Node socket handles.

3. **Test suite & review suggestions:**
   - Fixed test event payload typing (`fields` -> `payload`) and marked existing unit tests with `deferred: true` to prevent unmocked outbound requests.
   - Added a test confirming in-flight queue cancellation on `flushDetached`.
   - Updated the empty-catch comment in `detached-flush.ts` to match repo conventions.

packages/next/src/telemetry/detached-flush.ts
```diff
@@ -4,10 +4,13 @@ import type { TelemetryEvent } from './storage'
 import { Telemetry } from './storage'
 import loadConfig from '../server/config'
 import { getProjectDir } from '../lib/get-project-dir'
-import { PHASE_DEVELOPMENT_SERVER } from '../shared/lib/constants'
+import {
+  PHASE_DEVELOPMENT_SERVER,
+  PHASE_PRODUCTION_BUILD,
+} from '../shared/lib/constants'
```
refirst11, line 11:
```suggestion
import loadConfig from '../server/config'
import { getProjectDir } from '../lib/get-project-dir'
import { hasCustomExportOutput } from '../export/utils'
import {
  PHASE_DEVELOPMENT_SERVER,
  PHASE_PRODUCTION_BUILD,
} from '../shared/lib/constants'
```

`export/utils.ts` is a single pure function with a type-only import, so this doesn't pull anything into the child's startup path.

packages/next/src/telemetry/detached-flush.ts
```diff
@@ -16,19 +19,24 @@ import { PHASE_DEVELOPMENT_SERVER } from '../shared/lib/constants'
...
     )
   }
   dir = getProjectDir(dir)
 
-  const config = await loadConfig(PHASE_DEVELOPMENT_SERVER, dir)
+  const config = await loadConfig(
+    mode === 'build' ? PHASE_PRODUCTION_BUILD : PHASE_DEVELOPMENT_SERVER,
+    dir
+  )
   const distDir = path.join(dir, config.distDir || '.next')
```
refirst11, line 37:
```suggestion
  const config = await loadConfig(
    mode === 'build' ? PHASE_PRODUCTION_BUILD : PHASE_DEVELOPMENT_SERVER,
    dir
  )
  // Mirrors `build/index.ts` and `turbopack-build/impl.ts`: with
  // `output: 'export'` the user-configured distDir is the *out* dir, and the
  // build writes into `.next` instead. Without this the child resolves a
  // different directory than the one `flushDetached` wrote to.
  if (mode === 'build' && hasCustomExportOutput(config)) {
    config.distDir = '.next'
  }
  const distDir = path.join(dir, config.distDir || '.next')
```

See the review body — this is the `output: 'export'` case.

packages/next/src/telemetry/detached-flush.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    telemetry.flushDetached('build', tmpDir)
+
+    expect(spawnSpy).toHaveBeenCalledTimes(1)
+    const eventsFile = path.join(distDir, spawnSpy.mock.calls[0][1][3])
+    expect(fs.existsSync(eventsFile)).toBe(true)
+    const writtenEvents = JSON.parse(fs.readFileSync(eventsFile, 'utf8'))
+    expect(writtenEvents).toHaveLength(1)
+    expect(writtenEvents[0].eventName).toBe('NEXT_IN_FLIGHT_EVENT')
+  })
```
refirst11, line 153:
```suggestion
  it('aborts in-flight telemetry requests when flushDetached is called', () => {
    const telemetry = new Telemetry({ distDir })
    const prom = telemetry.record(
      {
        eventName: 'NEXT_IN_FLIGHT_EVENT',
        payload: { feature: 'in-flight' },
      },
      true
    )
    const controller = (prom as any)._controller
    expect(controller.signal.aborted).toBe(false)

    telemetry.flushDetached('build', tmpDir)

    expect(controller.signal.aborted).toBe(true)
  })
```

This fails against the old self-assignment (`_controller` would be `undefined`) and passes against 1103cb2, which is the property the test is named for. Dropping the spawn/events-file assertions is deliberate — the first test already covers those.

refirst11 (commented):
Thanks for turning all of that around so quickly. I read through 1103cb2 and the `storage.ts` wiring is right — the controller is created in `record()`, threaded into `submitRecord`, and attached to `prom`, so `flushDetached`'s `abort()` now reaches the in-flight request for every caller, not just `build`. Both worker flushes and the absolute `distDir` look correct too.

Two things I found in the new commit.

### 1. `output: 'export'` with a custom `distDir` now drops build telemetry silently

This one is new, because before this PR `next build` never involved the detached child.

`build/index.ts:1171-1186` overrides the config *before* deriving `distDir`:

```ts
      if (hasCustomExportOutput(config)) {
        config.distDir = '.next'
      }
      const distDir = path.join(dir, config.distDir)
      // …
      const telemetry = new Telemetry({ distDir })
```

`turbopack-build/impl.ts:347-349` mirrors it. So with `output: 'export'` and e.g. `distDir: 'dist'`, the parent and both workers write the events file to `<dir>/.next/_events_*.json`.

`detached-flush.ts` does a fresh `loadConfig()` **without** that override, so it resolves `<dir>/dist/` instead, finds nothing, and takes the `ENOENT → process.exit(0)` path. Every build event is dropped, and the events file is orphaned in `.next`.

The inline suggestion mirrors the same override in the child. An alternative would be to pass the already-resolved `distDir` down from `flushDetached` rather than having the child re-derive it — that would also save a `loadConfig()` per flush — but it changes the argv contract, so the mirror looked like the smaller change.

### 2. The new abort test doesn't observe the abort

`aborts in-flight telemetry requests when flushDetached is called` asserts the spawn and the events-file contents, which is what `spawns detached process with "build" mode when events are queued` already covers. Nothing looks at the controller, so the test would still pass if `_controller` reverted to the self-assignment — which is exactly the regression it exists to catch.

It's also the only `record()` left without `deferred: true`, so it starts a live request to `telemetry.nextjs.org`. The controller is created regardless of `deferred`, so asserting `signal.aborted` works with `true` and keeps the test offline.

refirst11:
> **Correction:** the ~500 ms claim below did not survive measurement. See [this comment](https://github.com/vercel/next.js/pull/97981#issuecomment-5500640962) for the numbers.

One more thing, and this one bears on whether the PR reaches its own goal.

On Node 24 and up, the global `fetch` refuses the ponyfilled `AbortSignal` that `telemetry/storage.ts:12` imports and `submitRecord` passes to every submission:

```
TypeError: RequestInit: Expected signal ("AbortSignal {}")
           to be an instance of AbortSignal.
```

It is thrown during `RequestInit` validation, before any network I/O, and `post-telemetry-payload.ts` ends in `.catch(() => {})`, so nothing surfaces. Two consequences.

**No telemetry is delivered on Node 24+.** Not slow — never sent, in `next build` and `next dev` alike, since `detached-flush.ts` takes the same `record()` → `submitRecord()` path. Measured across a stable and a canary release ([CI run](https://github.com/refirst11/nextjs-turbopack-feature-usage-repro/actions/runs/33553358480)):

| next | node | signal accepted | `postNextTelemetryPayload` returns after |
|---|---|---|---|
| 16.3.3 | v22.23.2 / v23.11.1 | yes | — |
| 16.3.3 | v24.19.0 / v25.9.0 / v26.8.1 | **no** | 500.7 / 502.2 / 501.8 ms |
| 16.4.0-canary.8 | v22.23.2 / v23.11.1 | yes | — |
| 16.4.0-canary.8 | v24.19.0 / v25.9.0 / v26.8.1 | **no** | 501.8 / 502.2 / 501.3 ms |

**`abort()` cannot recover the time this PR is trying to save.** When `fetch` fails before it starts, there is no request left to cancel — what remains is `async-retry`'s fixed 500 ms backoff, and an `AbortController` does not cancel a timer. Firing a submission and aborting it immediately, the way `flushDetached` now does:

```js
const controller = new AbortController()
postNextTelemetryPayload({ context: {}, meta: {}, events: [] }, controller.signal)
controller.abort()
// process exited after 508 ms
```

So on Node 24+ the parent process still cannot drain for up to ~500 ms after the build finishes. The worker change does hold — a pending timer no longer delays `workerMain()` from resolving — but the main-process half does not.

The fix looks like one line: use the global `AbortController` instead of the ponyfill in `storage.ts:12`. That file is the only importer of the ponyfill in the repository; the sole other reference is the NFT trace list in `test/production/next-server-nft/next-server-nft.test.ts:168`, which would need updating. It also only ever runs under Node, never in the edge runtime. The reason for it looks like it has simply expired: the polyfill arrived in #44614 (Jan 2023), when `engines.node` was `>=14.6.0` and Node 14 had no global `AbortController`. #49092 later rewrote the import path from `@edge-runtime/primitives/abort-controller` to `@edge-runtime/ponyfill` as part of an `@edge-runtime/*` suite bump, without revisiting whether it was still needed. `engines.node` is now `>=20.9.0`, and the global has existed since Node 15.

Would you be open to including it in this PR? It is the same file you are already changing, and without it the `abort()` you just wired up has nothing to cancel on Node 24+ — which is where the cost this PR targets is largest.

As a side effect, telemetry from Node 24+ environments — which is currently dropped entirely — starts arriving once this is fixed. Worth noting when reading the aggregates around the release.

This also corrects something in #97929: two submission sites × ~500 ms accounts for the ~1s I reported on Node 24+, so most of that number was this bug rather than the cost of awaiting the flush. The honest cost of the `await` is the 0.18–0.32 s I measured on Node 22. That does not change the case for this PR — it removes a real wait on every version — but it does mean the headline number in the issue was mostly measuring something else.

refirst11:
A correction to my previous comment, and the measurements behind it.

I claimed that on Node 24+ the parent process "still cannot drain for up to ~500 ms after the build finishes" even with this PR. **That is wrong.** I had measured `postNextTelemetryPayload` in isolation — a process with nothing else to do exits 508 ms after the aborted submission — and generalised that to a real build without checking. In a build the pending retry overlaps teardown that has to happen anyway (`await shutdownPromise` and the rest), so it does not extend wall time.

Measured properly: four builds of the same project, differing only in which patches are applied to the installed `dist` ([CI run](https://github.com/refirst11/nextjs-turbopack-feature-usage-repro/actions/runs/33559323806), 6 runs per arm, arms interleaved, first sample of each arm discarded). Telemetry cost is `next build` with telemetry enabled minus the same build under `NEXT_TELEMETRY_DISABLED=1`.

| node | released | native AbortController | PR #97981 | both |
|---|---|---|---|---|
| v22.23.2 | 0.464 s | 0.288 s | **0.096 s** | 0.225 s |
| v25.9.0 | 1.031 s | 0.356 s | **0.151 s** | 0.265 s |

**This PR removes most of the cost on its own, on Node 24+ as well** — 85% on Node 25, 79% on Node 22. My earlier comment understated what you had already achieved.

**The two changes are not additive.** On Node 25, 0.675 s and 0.880 s are each large fractions of the 1.031 s there is to remove; they overlap, because they attack the same wait from opposite ends. Fixing the signal makes the POST succeed in one RTT instead of failing and burning `async-retry`'s fixed 500 ms backoff, so the awaited flush becomes cheap. Removing the await makes its duration irrelevant. Either one alone gets most of the way. I would not read anything into the spread among the last three columns — that is within run-to-run variation on a shared runner.

**What does not overlap is delivery.** After this PR, `detached-flush.ts` still builds a `Telemetry`, calls `record()`, and hands a ponyfilled signal to `fetch`, so on Node 24+ nothing is submitted — before or after this PR. That is why I would still like the one-line swap: not for the milliseconds, which you have taken care of, but because telemetry from every Node 24+ user is currently dropped on the floor.

Same caveat as before: this ports your commits onto the compiled `dist` of a released build rather than building the branch, so read it as a measurement of the shape of the change, not of the exact code that will merge.

refirst11:
Heads-up: this now conflicts with canary. #96862 landed about an hour ago and removes the worker thread for Turbopack builds, taking `workerMain`, the worker's own `Telemetry`, and its `await telemetry.flush()` with it.

That makes the `turbopack-build/impl.ts` half of this PR obsolete — the flush I asked you to move is gone, along with the worker that owned it. Everything else still applies: `webpack-build/impl.ts` still has `workerMain` with its own `Telemetry` and an awaited flush at line 462, and nothing changed in `storage.ts`, `detached-flush.ts`, or `build/index.ts`.

It also dates the measurements I posted above. The +0.775 s I attributed to the Turbopack worker's exit path should now be gone on canary independently of this PR. I have not re-measured since #96862.

@lukesandberg — you have context on both sides of this. Besides #96862, you fixed the SSG side of exactly this problem in #91335. Two things here may be worth a look:

1. Whether the remaining `build/index.ts` change is still the shape you want, now that the Turbopack build runs in the main process.

2. Since Node 24, the global `fetch` refuses the ponyfilled `AbortSignal` that `telemetry/storage.ts:12` hands it, before any network I/O, and the error is swallowed — so telemetry from every Node 24+ user is dropped entirely, in `next dev` as well as `next build` ([CI](https://github.com/refirst11/nextjs-turbopack-feature-usage-repro/actions/runs/33553358480)). The polyfill dates to #44614, when `engines.node` was `>=14.6.0`; it is now `>=20.9.0`. Swapping it for the global looks like one line, but that is a maintainer's call, not mine.

refirst11:
I have opened #98165 for the `AbortController` swap, so it does not add scope here — it is two one-line deletions against canary and stands on its own.

Two unrelated notes for this PR:

- It currently conflicts with canary after #96862, as above.
- All three commits here are unsigned (`verification.reason: unsigned`). The PR template warns that this can block merging, and every recently merged PR I looked at has fully verified commits. Editing the files through the GitHub web UI signs the commits automatically, which is how I handled #98165 — might save you a force-push later.

Abhirup0:
Hi @refirst11,

Thank you so much for the thorough review, the benchmark measurements, and the clear breakdown across each scenario! Your analysis on the root causes and edge cases has been super helpful.

Here is what has been updated and pushed:

1. **Custom `distDir` under `output: 'export'`**:
   Imported `hasCustomExportOutput` from `../export/utils` into `detached-flush.ts` and mirrored the `config.distDir = '.next'` override. The detached child now resolves the same `.next` directory that `build/index.ts` writes to.

2. **Abort verification in `detached-flush.test.ts`**:
   Updated `aborts in-flight telemetry requests when flushDetached is called` to pass `deferred: true` (preventing live outbound network requests) and assert `signal.aborted` directly on `(recordPromise as any)._controller` before and after `flushDetached`.

3. **Rebase on `canary` (#96862)**:
   Rebased against latest `canary`. With #96862 removing the Turbopack worker thread, Turbopack now runs directly in the main build process and is handled by the main `build/index.ts` flush; `webpack-build/impl.ts` continues to use `flushDetached` for its worker.

4. **#98165 (`AbortController` ponyfill removal)**:
   Saw your PR #98165 — keeping that as an independent cleanup makes complete sense.

Thanks again for all your time and dedication on this issue!

packages/next/src/telemetry/detached-flush.ts
```diff
@@ -16,19 +20,27 @@ import { PHASE_DEVELOPMENT_SERVER } from '../shared/lib/constants'
...
   dir = getProjectDir(dir)
 
-  const config = await loadConfig(PHASE_DEVELOPMENT_SERVER, dir)
+  const config = await loadConfig(
+    mode === 'build' ? PHASE_PRODUCTION_BUILD : PHASE_DEVELOPMENT_SERVER,
+    dir
+  )
+  if (hasCustomExportOutput(config)) {
+    config.distDir = '.next'
+  }
```
refirst11:
```suggestion
  if (mode === 'build' && hasCustomExportOutput(config)) {
    config.distDir = '.next'
  }
```

Gating on `build` keeps the dev path resolving the same directory its parent writes to.

refirst11 (commented):
The rebase and the abort test both look right to me — the test now fails against the old self-assignment and passes here, which is exactly what it was missing. Updating `TurbopackInternalError.test.ts` for the new `submitRecord` arity was a good catch; I had not thought to look for callers of it.

One thing the export fix brought with it. `hasCustomExportOutput` is now applied for both modes, but only `build` resolves `distDir` that way. In dev, `router-server.ts:199-200` builds its `Telemetry` from the raw `path.join(opts.dir, config.distDir)`, and `cli/next-dev.ts:150-151` uses the distDir it receives over IPC — neither applies the override.

So with `output: 'export'` and, say, `distDir: 'dist'`, `next dev` writes the events file to `<dir>/dist/` while the child now looks in `<dir>/.next/`, hits ENOENT and exits 0. That is the same silent loss as the build case, mirrored onto dev. Suggestion inline.

Separately, and I mention it again only because the rebase would have discarded any signing you had done before it: all four commits are currently unsigned. The template warns this can block merging, and every recently merged PR I looked at has fully verified commits. Editing through the GitHub web UI signs them automatically. Now that the rebase is behind you seems like the moment to sort it, rather than after a reviewer asks.

That suggestion is the last thing I have. From my side this looks complete — thank you for taking every round of this so quickly.

packages/next/src/telemetry/detached-flush.ts
Abhirup0: Good catch on `next dev` mode — updated and pushed! Gating on `mode === 'build'` ensures `dev` continues resolving the raw `distDir` it writes to while `build` gets the `.next` export override. Thanks again for the spot and the signing heads-up!

refirst11:
Not another review point from me — a procedural one. The bot's last pass was against 1103cb2, and the implementation has changed materially since: the rebase dropped the Turbopack worker changes entirely, and the export `distDir` handling and its `build`-mode gate are both new. Worth a fresh look at what is actually there now.

@vercel run a full code review

PR #95310, opened 2026-06-30
Turbopack: expose list of non-inlined env vars

mischnic:
Add a new method to `EcmascriptAnalyzable` to return the list of non-inlined env vars

In this PR, this isn't used anywhere outside of the tests yet

While `cargo bench -p turbopack-ecmascript` does regress:
```
Benchmarking references/packages-bundle.js/full: Warming up for 1.0000 s
references/packages-bundle.js/full
                        time:   [152.99 ms 153.34 ms 153.70 ms]
                        change: [+11.939% +12.678% +13.307%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
Benchmarking references/packages-bundle.js/tracing: Warming up for 1.0000 s
references/packages-bundle.js/tracing
                        time:   [153.07 ms 153.92 ms 155.15 ms]
                        change: [+34.997% +36.345% +37.554%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
```

There is no measurable overall perf impact:
```
commit 7d29e358c8ac383a3019d552fd54f4d05645bf39 (HEAD, tag: v16.4.0-canary.5)
    v16.4.0-canary.5

pnpm next build --experimental-build-mode=compil  400.89s user 51.96s system 747% cpu 1:00.61 total
pnpm next build --experimental-build-mode=compil  394.99s user 54.10s system 757% cpu 59.289 total
pnpm next build --experimental-build-mode=compil  399.87s user 58.71s system 772% cpu 59.398 total
```
vs
```
commit 9335597135118b8332b8e0d5a3bcedb596a3ef07 (HEAD -> mischnic/env-var-references)
    remove runtime_all

pnpm next build --experimental-build-mode=compil  399.83s user 50.09s system 719% cpu 1:02.57 total
pnpm next build --experimental-build-mode=compil  403.57s user 53.79s system 743% cpu 1:01.52 total
pnpm next build --experimental-build-mode=compil  402.23s user 55.05s system 758% cpu 1:00.29 total
pnpm next build --experimental-build-mode=compil  400.17s user 55.15s system 761% cpu 59.817 total
```

PR #97984, opened 2026-08-27
Turbopack: allow `get_definable_name` to return a list

mischnic:
Followup to #95310

Now we can correctly track env vars in cases like these were a `JsValue::Alternative` is involved
```js
// An alternative (process|unknown).env
let p2
if (foo) {
  p2 = foo
} else {
  p2 = process
}
console.log(p2.env.FOO8)
```

The inlining logic behaves the same as before:
- If `get_definable_name` returns exactly one result, we can inline.
- Otherwise, we can't inline (previously this case returned `None`)

A bit of trivia why we need that boolean in the return type in the first place (because conceptually it should be possible to just model it via `JsValue::Alternative` anyway:

> but why is that a problem? if the value is reassigned, then there should just an alternative, then get_definable_name returns multiple values and it should still not be inlining
>
> Because module is a `FreeVar`, not a normal `Variable`.
>
> The assignment is stored separately in `VarGraph`:
>
> ```
> free_var_ids["module"] -> id
> values[id] -> assigned function
> ```
>
> But later reads still evaluate to:
>
> `JsValue::FreeVar("module")`
>
> The linker expands JsValue::Variable, but not JsValue::FreeVar. Therefore there is no JsValue::Alternatives at this call site.
>
> Instead, `get_definable_name()` detects the graph entry and returns:
>
> `[Some((["module", TypeOf], true))]`
>
> That boolean is how reassignment is represented for free variables. Cardinality remains one, so ignoring the boolean at references/mod.rs:4007 incorrectly permits inlining.
>
> Actual `Alternatives` primarily represent multiple values assigned to tracked variables. For free globals, the original ambient value plus reassignment is represented by `FreeVar` plus `potentially_reassigned`.

No perf impact:

```
commit cfc7da3049f91669a1b61a0f7a5edef503cddc53 (HEAD -> canary, origin/canary, origin/HEAD)
393.17s user 31.34s system 722% cpu 58.775 total
391.91s user 35.37s system 746% cpu 57.256 total
389.48s user 30.40s system 770% cpu 54.513 total

commit 38f2708659be31538c951b12ba806ef53710f074 (HEAD -> mischnic/get-definable-name-list)
392.86s user 34.00s system 745% cpu 57.221 total
388.10s user 32.72s system 766% cpu 54.912 total
391.84s user 37.82s system 717% cpu 59.865 total
```

### DIFF
--- a/turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/mod.rs
+++ b/turbopack/crates/turbopack-ecmascript/src/analyzer/jsvalue/mod.rs
@@ -8,7 +8,7 @@ use std::{
 use anyhow::{Context, Result, bail};
 use bumpalo::boxed::Box as BumpBox;
 use num_bigint::BigInt;
-use smallvec::SmallVec;
+use smallvec::{SmallVec, smallvec};
 use swc_core::ecma::{ast::Id, atoms::Atom};
 use turbo_rcstr::{RcStr, rcstr};
 use turbopack_core::compile_time_info::{
@@ -1268,7 +1268,7 @@ impl JsValue<'_> {
     pub fn get_definable_name(
         &self,
         var_graph: Option<&VarGraph<'_>>,
-    ) -> Option<(DefinableNameSegmentRefs<'_>, bool)> {
+    ) -> SmallVec<[(DefinableNameSegmentRefs<'_>, bool); 1]> {
         let mut current = self;
         let mut segments = SmallVec::new();
         let mut potentially_reassigned = false;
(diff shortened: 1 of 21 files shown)


mischnic:
* **#97984** <a href="https://app.graphite.com/github/pr/vercel/next.js/97984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/97984?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#97985** <a href="https://app.graphite.com/github/pr/vercel/next.js/97985?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

wbinnssmith (approved):

Issue #92703, opened 2026-04-12
Malformed Origin header in Next.js Server Actions returns HTTP 500 instead of 400


FJRG2007:
### Link to the code that reproduces this issue

https://github.com/FJRG2007/nextjs-sa-dos-poc-20260413

### To Reproduce

**Prerequisites:** A self-hosted Next.js application (not deployed on Vercel CDN) with at least one Server Action defined anywhere in the project.       
                                                                  
  **Step 1 — Malformed URL (affects all versions including canary and v16.2.3):**                                                                          
                                                                  
  Send the following raw HTTP request to any page of the application:                                                                                      
                                                                  
  POST /any-page HTTP/1.1                                                                                                                                  
  Host:                                                           
  Content-Type: multipart/form-data; boundary=----PoC                                                                                                      
  Origin: http://
                                                                                                                                                           
  ------PoC--                                                     
                                                                                                                                                           
  With curl:                                                      
  ```bash
  curl -X POST https://your-next-app.example.com/ \
    -H "Content-Type: multipart/form-data; boundary=----PoC" \                                                                                             
    -H "Origin: http://" \
    --data-binary $'------PoC--\r\n'       
  ```                                                                                                                
                                                                                                                                                           
  Step 2 — Origin: null (additionally affects all v15.5.x stable releases including v15.5.15):                                                             
                                       
  ```bash                                                                                                                    
  curl -X POST https://your-next-app.example.com/ \                                                                                                        
    -H "Content-Type: multipart/form-data; boundary=----PoC" \                                                                                             
    -H "Origin: null" \
    --data-binary $'------PoC--\r\n'         
  ```                                                                                                              
                                                                                                                                                           
  ▎ Note: No Next-Action header is required. Content-Type: multipart/form-data is sufficient to reach the CSRF validation code path (isMultipartAction =   
  ▎ true). No knowledge of a valid action ID is needed to trigger the crash.

### Current vs. Expected behavior

new URL(originHeader).host throws TypeError: Invalid URL for any Origin value that is not a parseable absolute URL (e.g. http://, ftp://, invalid, or    
  null on v15.5.x). The exception is not caught inside handleAction(). It propagates through renderToHTMLOrFlightImpl → renderToHTMLOrFlight → the
  base-server error handler, which sets res.statusCode = 500 and logs the full stack trace. The HTTP response is **HTTP 500 Internal Server Error**. The Server
   Action is never executed.

Server log output:      
```                                                                                                                              
[...]

Issue #90090, opened 2026-02-17
500 error instead of 404 when answering a POST request with form data


vinassefranche:
### Link to the code that reproduces this issue

https://github.com/vinassefranche/next-js-server-action-issue-with-form-data

### To Reproduce

1. Start the application in dev mode (`pnpm dev`)
2. Run `node send-form-data-request.mjs` 

### Current vs. Expected behavior

I expect a 404 status code being returned instead of a 500 status code

### Provide environment information

```bash
Operating System:
  Platform: linux
  Arch: x64
  Version: #29~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Jun 26 14:16:59 UTC 2
  Available memory (MB): 63932
  Available CPU cores: 22
Binaries:
  Node: 22.22.0
  npm: 10.9.4
  Yarn: 1.22.22
  pnpm: 9.15.5
Relevant Packages:
  next: 16.1.6 // Latest available version is detected (16.1.6).
  eslint-config-next: N/A
  react: 19.2.3
  react-dom: 19.2.3
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Server Actions

### Which stage(s) are affected? (Select all that apply)

next dev (local), next start (local)

### Additional context

- If the server does not have any server action, the request is answered properly with a 404. It seems like the request is handled by this: https://github.com/vercel/next.js/blob/77bb08d129a46eae14cd73c8f6286741153755f3/packages/next/src/server/app-render/action-handler.ts#L602-L605
- When sending a request with the "next-action" header being empty, we get a 404. Run node send-json-request.mjs to see this

- This looks a bit similar to [this issue](https://github.com/vercel/next.js/issues/86945) and I think some comments on it are actually targeted as the issue I'm creating

vinassefranche:
I believe the culprit is [this throw](https://github.com/vercel/next.js/blob/a5f659034626b4af9aae75e09ddb58c05778f684/packages/next/src/server/app-render/action-handler.ts#L775-L781)
When the formData is empty, we should probably handle the error gracefully like it is done [here](https://github.com/vercel/next.js/blob/a5f659034626b4af9aae75e09ddb58c05778f684/packages/next/src/server/app-render/action-handler.ts#L761-L765)

icyJoseph:
Duplicate here right? https://github.com/vercel/next.js/issues/86945

[...]

Issue #86945, opened 2025-12-08
HTTP 500 on malformed request body for server action (JSON parse SyntaxError)


artemzhdev:
### Link to the code that reproduces this issue

https://github.com/artemzhdev/next-js-server-action-500

### To Reproduce

1. Run your application and trigger any server action from the client (browser)
2. Find POST request related to the server action in the NETWORK tab and copy it as curl
3. Slightly modify it to make the request body to contain any mailformed JSON (like "["). You can just replace "  --data-raw '[]'" with " --data-raw '[' -v" at the end.
4. Try to run this request using terminal

### Current vs. Expected behavior

Current behavior: HTTP 500.
Expected behavior: HTTP 400 (Bad Request)

### Provide environment information

```bash
NextJS version: 15.3.6 & 16.0.1.
```

### Which area(s) are affected? (Select all that apply)

Server Actions

### Which stage(s) are affected? (Select all that apply)

next start (local)

### Additional context

My site is being scanned by some vulnerability scanner. It's not an issue. But I use monitoring to track all 5xx errors. Such responses are tracked as high priority bugs.
Unfortunately, this happens for bad requests produced by vulnerability scanners.
My server action code is not even called. And I cannot find a way to catch it in the middleware.

icyJoseph:
Can you post more of the error stack you got? If possible of course. Thanks in advance.

rafael-tanaka:
I had a similar issue. A brand-new app built with the latest versions of Next.js and React (including all vulnerability patches) is still returning 500 errors, along with the well-known `E {"digest"` response body.
Some security consultants claim that the app is still vulnerable because this specific combination of a 500 error and that response body is considered an indicator of the issue. However, when running fix-react2shell-next, it reports that no vulnerable packages were found.

CURL: 
```
curl -i 'https://yourwebsite.com/' \
  -X POST \
  -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.113 Safari/537.36 Assetnote/1.0.0' \
  -H 'Next-Action: x' \
  -H 'X-Nextjs-Request-Id: b5dce965' \
  -H 'Next-Router-State-Tree: %5B%22%22%2C%7B%22children%22%3A%5B%22__PAGE__%22%2C%7B%7D%2Cnull%2Cnull%5D%7D%2Cnull%2Cnull%2Ctrue%5D' \
[...]

PR #97999, opened 2026-08-27
fix(server-actions): respond 4xx instead of 500 to invalid action requests

icyJoseph:
Closes:

- #86945 HTTP 500 on malformed request body for server action (JSON parse SyntaxError) 
- #90090 500 error instead of 404 when answering a POST request with form data 
- #92703 Malformed Origin header in Next.js Server Actions returns HTTP 500 instead of 400 
- #95447 Server Actions: multipart/form-data POST to a non-action route returns 500 in production (next start) but 404 in development (next dev)

## Claude PR description

`serverModuleMap` is not a nullable map. `createServerModuleMap()` in `manifests-singleton.ts` returns a throwing Proxy: an id of the wrong shape throws `getInvalidServerReferenceIdError`, and a well-formed id absent from the build throws `getActionNotFoundError`. For a string id it returns `undefined` in exactly one case, when the id collides with a well-known property name such as `toString`. (It also falls through to `Reflect.get` for non-string keys, which is React's debug serialization probing the map with symbols rather than looking up a server reference.)

Two validation predicates were written against the wrong contract. `isInvalidStringActionDescriptor` (bound-args variant) and `isInvalidActionIdFieldName` (the `$ACTION_ID_<id>` form-field variant, the one the common repro hits) both do `const entry = serverModuleMap[actionId]` and then test `entry == null`. The subscript throws before that test is ever reached, so the throw escaped the validation helper entirely and landed in the generic action-error handler, which treats any error arriving there as a failure inside user action code and answers 500.

Two details worth having:

- That `entry == null` branch is dead code in both predicates. They gate on `mightBeServerReferenceId`, which is `id.length === 42` and nothing else, while the only `undefined`-returning path is the well-known-property list whose longest member (`propertyIsEnumerable`) is 20 characters. The same branch is genuinely live in `getActionModIdOrError`, which has no length gate, and that function's author documented the proxy's behavior correctly. So the correct handling already existed in the file, one function away from the two that got it wrong.
- Two further failure classes threw into the same catch-all: `new URL(origin)` on a malformed `origin` header (#92703), and the React Flight decoder calls (`decodeReply`, `decodeReplyFromBusboy`, `decodeAction`, `decodeFormState`) on a body they can't parse (#86945).

Separately, `getActionNotFoundError` interpolated the client-supplied id straight into its message. The id only has to be 42 characters to pass validation and its bytes are never checked, so it is now escaped with `JSON.stringify` for the same reason the decode-failure message already was.

### DIFF
--- a/packages/next/src/server/app-render/action-handler.ts
+++ b/packages/next/src/server/app-render/action-handler.ts
@@ -58,6 +58,7 @@ import {
   getServerModuleMap,
   getActionNotFoundError,
   getInvalidServerReferenceIdError,
+  isActionNotFoundError,
 } from './manifests-singleton'
 import { isNodeNextRequest, isWebNextRequest } from '../base-http/helpers'
 import { normalizeFilePath } from './segment-explorer-path'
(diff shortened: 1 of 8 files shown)


icyJoseph:
Closing as https://github.com/vercel/next.js/pull/98123 fixed all of these

PR #98003, opened 2026-08-27
Turbopack: Add support for specifying additional roots

bgw:
Full motivation and plan here: https://app.notion.com/p/vercel/Turbopack-pnpm-Global-Virtual-Store-383e06b059c480579403ddfd71cc2d40?source=copy_link

The goal is to allow `DiskFileSystem` to traverse outside of it's own root to other configured `DiskFileSystem`s when following symlinks. We may allow traversal in other situations in the future, but this is limited to symlink resolution for now.

## Global Virtual Store

The motivation for this is to enable [pnpm's Global Virtual Store feature](https://pnpm.io/global-virtual-store) (and there are other package managers doing this, including nub and bun).

We'd expose the ability to manually configure this in `next.config.js`, but we should also auto-configure ourselves for popular package managers (or at least make a best effort to do so, the `PNPM_HOME` semantics can be complicated). The `ignoreIfMissing` option is provided for this situation: We can configure a bunch of roots automatically, and they only actually get set up if they exist, the check for directory existence is cheap.

## NFT changes

This requires a couple extensions to the `*.nft.json` file format: https://github.com/vercel/next.js/pull/98469

## Related Issues

- #93556
- https://github.com/pnpm/pnpm/issues/14972


turbopack/crates/turbopack-core/src/file_source.rs
```diff
@@ -70,13 +70,25 @@ impl Asset for FileSource {
...
+                    if target_fs_path.fs != self.path.fs {
+                        return match *target_fs_path.get_type().await? {
+                            FileSystemEntryType::File => Ok(AssetContent::File(
+                                target_fs_path.read().to_resolved().await?,
+                            )
+                            .cell()),
+                            FileSystemEntryType::Directory => {
+                                bail!("Cannot read directory as file content: {target_fs_path}")
+                            }
+                            _ => bail!("Invalid cross-filesystem symlink target"),
```
bgw, line 83: Probably don't want to bail here

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -680,6 +666,48 @@ impl DiskFileSystem {
...
+        }
+
+        let ancestors: SmallVec<[&Path; 8]> = target_sys_path.ancestors().collect();
+        for prefix in ancestors.into_iter().rev().skip(1) {
+            let Some(prefix_str) = prefix.to_str() else {
+                return Ok(None);
+            };
+            let Some(canonical) = canonicalize_untracked(RcStr::from(prefix_str))
+                .owned()
+                .await?
```
bgw, line 714: Add some comments about how this is similar to `resolve_link_target_ancestry_slow_path`, or figure out a better way to share code between these two places.

Possible future follow-up PR: I'd also like to just optimize this pattern a bit better, but it'll involve hacking around turbo-tasks in some unsafe ways.

crates/next-api/src/project.rs
```diff
@@ -458,25 +457,201 @@ impl ProjectContainer {
...
+) -> Result<Vc<RcStr>> {
+    let container = container.await?;
+    let paths = container.additional_root_paths.get();
+    let path = paths
+        .get(&key)
+        .with_context(|| format!("Unexpected: additional root {key} is missing"))?;
+    Ok(Vc::cell(path.clone()))
+}
+
+#[turbo_tasks::function(operation, session_dependent)]
```
lukesandberg: not session_dependent

crates/next-api/src/nft_json_builder.rs
```diff
@@ -0,0 +1,265 @@
...
+            tuple.serialize_element(&root)?;
+        }
+        tuple.end()
+    }
+}
+
+#[derive(Default, Serialize)]
+#[serde(rename_all = "camelCase")]
+struct NftFileList {
+    files: Vec<RcStr>,
```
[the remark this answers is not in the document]
bgw: We always emit symlinks for Turbopack. We don't always emit it for webpack, which is why the downstream code has to handle for that.

This code only implements `Serialize` and not `Deserialize`, it does not need to be aware that other writers of nft files do not emit this field.

crates/next-api/src/additional_roots.rs
```diff
@@ -0,0 +1,482 @@
...
+
+impl AdditionalRootInvalidName {
+    fn description(&self) -> StyledString {
+        match self {
+            Self::Empty => StyledString::Text(rcstr!("the name must not be empty")),
+            Self::TooLong => {
+                StyledString::Text(rcstr!("the name must be at most 40 ASCII characters"))
+            }
+            Self::DotOrDotDot => StyledString::Text(rcstr!("the name must not be `.` or `..`")),
+            Self::NonAscii => {
```
lukesandberg: A-Za-z0-9_

crates/next-api/src/additional_roots.rs
```diff
@@ -0,0 +1,482 @@
...
+            push_issue(AdditionalRootIssueReason::NameCollision {
+                existing_key: existing_key.clone(),
+            })
+            .await?;
+            continue;
+        }
+        configured_names.insert(folded_name, additional_root.key.clone());
+
+        let configured_path = additional_root.path.clone();
+        let canonical = match tokio::task::spawn_blocking(move || {
```
lukesandberg: why spawn_blocking?

crates/next-api/src/additional_roots.rs
```diff
@@ -0,0 +1,482 @@
...
+            })
+            .await?;
+            continue;
+        }
+        // We're not inside a turbo-task function: Call an operation to create a cell for us. We
+        // pass the `ProjectContainer` and a key, which both have a stable identity, this reduces
+        // invalidations when additional roots are added or removed.
+        let canonical_root = additional_root_path_operation(container, additional_root.key.clone());
+        let operation = disk_file_system_operation(
+            RcStr::from(format!("additional-root-{}", additional_root.key)),
```
bgw:
```suggestion
            RcStr::from(format!("@{}", additional_root.key)),
```

crates/next-api/src/nft_json_builder.rs
```diff
@@ -88,8 +88,8 @@ pub(crate) struct NftJson {
     version: u8,
     #[serde(skip_serializing_if = "Option::is_none")]
     entry_hash: Option<RcStr>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    additional_roots: Option<Vec<NftAdditionalRoot>>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
```
lukesandberg, line 91: move to previous pr?

packages/next/src/server/config.test.ts
```diff
@@ -308,4 +309,26 @@ describe('loadConfig', () => {
...
+        },
+      })
+
+      expect(
+        Object.entries(result.experimental.turbopackAdditionalRoots ?? {})
+      ).toEqual([
+        ['second', { path: resolve('./second'), ignoreIfMissing: true }],
+        ['first', { path: resolve('../first') }],
+      ])
+    })
```
lukesandberg: this test isn't testing anything...

packages/next/src/server/config.ts
```diff
@@ -1174,6 +1174,20 @@ function assignDefaultsAndValidate(
...
 
+  if (result?.experimental?.turbopackAdditionalRoots) {
+    result.experimental.turbopackAdditionalRoots = Object.fromEntries(
+      Object.entries(result.experimental.turbopackAdditionalRoots).map(
+        ([key, root]) => [
+          key,
+          {
+            ...root,
+            path: isAbsolute(root.path) ? root.path : resolve(root.path),
+          },
```
lukesandberg: intentional? shouldn't this happen in rust land?

test/e2e/app-dir/turbopack-additional-roots/turbopack-additional-roots.test.ts
```diff
@@ -0,0 +1,249 @@
...
+        (content) => `${content}\nexport const broken =`,
+        async () => {
+          const { exitCode, cliOutput } = await next.build()
+          expect(exitCode).toBe(1)
+          expect(cliOutput).toContain('Invalid Turbopack additional root')
+        }
+      )
+
+      expect((await next.build()).exitCode).toBe(0)
+      await next.start({ skipBuild: true })
```
lukesandberg: we don't need a broken build?

lukesandberg: we should get the warning either way?

test/e2e/app-dir/turbopack-additional-roots/turbopack-additional-roots.test.ts
```diff
@@ -0,0 +1,249 @@
...
+    })
+
+    it('can rebuild after changing the additional roots config', async () => {
```
lukesandberg: is this test interesting?

test/e2e/app-dir/turbopack-additional-roots/turbopack-additional-roots.test.ts
```diff
@@ -0,0 +1,249 @@
...
+      )
+      const root = nft.additionalRoots[rootIndex]
+
+      expect(rootIndex).toBeGreaterThanOrEqual(0)
+      expect(path.resolve(path.dirname(nftPath), root.path)).toBe(
+        await fs.realpath(externalRoot)
+      )
+      expect(nft.fileHashes).toHaveLength(nft.files.length)
+      expect(root.fileHashes).toHaveLength(root.files.length)
+      expect(root.symlinks).toEqual([])
```
lukesandberg: maybe toMatchInlineSnapshot on symlinks and additioanlRoots?

test/e2e/app-dir/turbopack-additional-roots/turbopack-additional-roots.test.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    it('runs after relocating standalone output away from the source root', async () => {
+      await next.stop()
+      const standaloneDirectory = await fs.mkdtemp(
+        path.join(os.tmpdir(), 'next-additional-roots-')
+      )
+      let server: any
+
+      try {
+        await fs.remove(standaloneDirectory)
```
bgw: Don't need to remove a newly created tempdir

crates/next-api/src/project.rs
```diff
@@ -429,10 +436,20 @@ pub struct Instrumentation {
...
+    project_file_system: OperationVc<DiskFileSystem>,
+    output_file_system: OperationVc<DiskFileSystem>,
+}
+
+#[turbo_tasks::value(serialization = "skip", evict = "never", eq = "manual", cell = "new")]
 pub struct ProjectContainer {
     name: RcStr,
-    options_state: State<Option<ProjectOptions>>,
+    state: State<Option<ProjectContainerState>>,
+    additional_roots: State<FxIndexMap<RcStr, AdditionalDiskFileSystem>>,
```
bgw: Maybe can improve the naming of these two things

crates/next-api/src/project.rs
```diff
@@ -449,15 +466,158 @@ impl ProjectContainer {
...
+async fn prepare_project_container_state(
+    container_vc: ResolvedVc<ProjectContainer>,
+    options: ProjectOptions,
+) -> Result<Vec<ReadRef<PlainIssue>>> {
+    let container = container_vc.await?;
+    // Operations created during initialization may begin running immediately. Keep both states
+    // advisory-locked until each has been populated so those operations cannot observe partial
+    // initialization when dependency tracking (and therefore invalidation) is disabled.
+    let additional_roots_guard = container.additional_roots.advisory_lock().await;
+    let state_guard = container.state.advisory_lock().await;
```
lukesandberg: shold these states be the same? how critical is it to drop the state_guard first

crates/next-api/src/project.rs
```diff
@@ -449,15 +466,158 @@ impl ProjectContainer {
...
+    )
+    .join(&project_path)?
+    .join(&config_file_name)?;
+    let additional_roots = create_additional_root_file_systems(
+        container_vc,
+        configured_additional_roots,
+        &project_root,
+        watcher_config,
+        map,
+        config_path,
```
bgw, line 573: This could take the `project_fs_op` instead of a `config_path`, and that might let us avoid having separate `State` objects for project+output filesystems and the additional roots.

turbopack/crates/turbo-tasks/src/state.rs
```diff
@@ -258,9 +261,25 @@ impl<T> State<T> {
...
+    ///
+    /// This lock is advisory: [`State::get`], [`State::get_untracked`], and mutation methods do not
+    /// acquire it automatically. Every participant in the coordinated operation must explicitly
+    /// acquire it. Callers must also ensure that work awaited while holding this guard cannot
+    /// depend on another participant acquiring the same lock, or they will deadlock.
+    ///
+    /// Normal state access should use the synchronous lock built into those methods. This API is
+    /// intended only for exceptional initialization sequences that must prevent tasks from
+    /// observing partially initialized state.
+    pub async fn advisory_lock(&self) -> tokio::sync::MutexGuard<'_, ()> {
```
lukesandberg: no strong opinion but if there is just one set of operations coordinating in Project::update perhaps this could be a field on Project/ProjectContainer?

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1394,7 +1394,12 @@ async fn find_package(
             ResolveModules::Nested(root, names) => {
                 let mut lookup_path = lookup_path.clone();
                 let mut lookup_path_value = lookup_path.clone();
-                while lookup_path_value.is_inside_ref(root) {
+                let lookup_root = if lookup_path.fs == root.fs {
+                    root.clone()
+                } else {
+                    lookup_path.root().owned().await?
```
lukesandberg: put an example here?

lukesandberg: maybe Nested shouldn't even have a root path, it should just be defined as 'these directories relative to fs roots'

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -1394,7 +1394,12 @@ async fn find_package(
...
+                };
+                while lookup_path_value.is_inside_ref(&lookup_root) {
```
bgw: - Don't allow configuring `ResolvedModule::Nested` with a root, we always just pass in the project fs root today.
- We don't need `lookup_root` + `is_inside_ref`, we just need to check that the `FileSystemPath` is not the root (i.e. the `path` field is not an empty string).

test/e2e/app-dir/turbopack-additional-roots/turbopack-additional-roots.test.ts
bgw: Oh, it does need to do this... Sort of... It's using `mkdtemp` to decide on a unique name in the tempdir, but then it's moving `.next/standalone` to that path.

We can just make `standalone` a subdirectory of the tempdir though.

crates/next-api/src/project.rs
```diff
@@ -713,56 +907,16 @@ impl ProjectContainer {
...
+                "env_diff",
+                define_env_diff_report(&old_define_env, &new_options.define_env).as_str(),
+            );
             this.options_state.set(Some(new_options));
-            let project = project_operation(self)
+
+            project_operation(self)
                 .resolve()
                 .strongly_consistent()
                 .await?;
```
bgw: Not sure that we need to call this anymore.

crates/next-api/src/project.rs
```diff
@@ -617,6 +616,11 @@ async fn prepare_project_container_state(
         .await?
         .invalidate_with_reason(invalidation_reason);
 
+    project_operation(container_vc)
+        .resolve()
+        .strongly_consistent()
+        .await?;
+
```
bgw: Why?

crates/next-api/src/project.rs
```diff
@@ -450,6 +450,9 @@ pub struct ProjectContainer {
     options_state: State<Option<ProjectOptions>>,
     file_systems_state: State<Option<ProjectFileSystemState>>,
     additional_roots_state: State<Vec<(RcStr, AdditionalDiskFileSystem)>>,
+    #[turbo_tasks(debug_ignore, trace_ignore)]
+    #[bincode(skip)]
+    initialization_lock: tokio::sync::Mutex<()>,
```
bgw: Rename to `fs_map_init_lock`

turbopack/crates/turbopack-core/src/resolve/mod.rs
```diff
@@ -3903,7 +3890,7 @@ mod tests {
 
         let extensions = custom_extensions
             .unwrap_or_else(|| vec![rcstr!(".ts"), rcstr!(".js"), rcstr!(".json")]);
-        let mut options_value = node_esm_resolve_options(lookup_path.clone())
```
bgw, line 3901: Why was this passing through `lookup_path` and not an fs root?

test/e2e/app-dir/turbopack-additional-roots/turbopack-additional-roots.test.ts
```diff
@@ -94,29 +94,46 @@ describe('turbopack additional roots', () => {
...
+         },
+       ]
+      `)
+      expect(nft.additionalRoots).toMatchInlineSnapshot(`
+       [
+         {
+           "fileHashes": [
+             "68a5d859aeaab2b7418a1215c0bb0d26",
+             "65e26ffbe84715407dd7bfe827f6444d",
+           ],
```
bgw: Strip the fileHashes before doing the snapshot

turbopack/crates/turbopack-core/src/resolve/mod.rs
bgw: :robot: You're right—the helper's only caller gets lookup_path from fs.root(), so it was already the filesystem root. The other callers likewise passed lookup_path.root() or tsconfig.root(). This change only removes the now-redundant root parameter after ResolveModules::Nested was changed to walk to the FileSystemPath root itself; it does not correct a bad value at this call site.

lukesandberg (approved):

PR #98005, opened 2026-08-28
[PPF] Use RDC in cachedNavigations codepaths

lubieowoce:
In #94194 we accidentally started clobbering the resume data cache in the `cachedNavigations` codepath (i.e. when rendering the embedded runtime prefetch stream).
This assignment used to be non-destructive because `prerenderResumeDataCache` was separate:
https://github.com/vercel/next.js/blob/b0420899b68a4f7506fdbc93ff5acf6edcfa3d3c/packages/next/src/server/app-render/app-render.tsx#L969-L972
But after unifying `prerenderResumeDataCache` and `renderResumeDataCache`, we started essentially dropping the RDC and using an empty one
https://github.com/vercel/next.js/blob/8c661813a1e66cd85b795e3aeae0f04ae3e3201e/packages/next/src/server/app-render/app-render.tsx#L970-L973
This wasn't caught because the tests that checked RDC behavior weren't exercising this codepath (at the time: no `cachedNavigations` + runtime prefetch opt in, and later on no `partialPrefetching` opt in)

This PR fixes this by prefilling the mutable cache from the RDC. I've added `.partial-prefetching` forks to the RDC and use-cache tests to verify, both fail before this fix.

### DIFF
--- a/test/e2e/app-dir/use-cache/use-cache.test.ts
+++ b/test/e2e/app-dir/use-cache/use-cache.test.ts
@@ -18,6 +18,7 @@ const GENERIC_RSC_ERROR =
   'Minified React error #441; visit https://react.dev/errors/441 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.'
 
 const withCacheComponents = process.env.__NEXT_CACHE_COMPONENTS === 'true'
+const partialPrefetching = process.env.__NEXT_PARTIAL_PREFETCHING === 'true'
 
 describe('use-cache', () => {
   const { next, isNextDev, isNextDeploy, isNextStart, skipped } = nextTestSetup(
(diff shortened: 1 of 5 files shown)


test/e2e/app-dir/use-cache/use-cache.test.ts
```diff
@@ -670,43 +671,49 @@ describe('use-cache', () => {
         expect(await browser.elementById('y').text()).toBe('Loading...')
       })
 
-      it('should omit caches with a short stale time from prerendered shells', async () => {
-        // Disable JS as a hack to see what's included in the static shell
-        let browser = await next.browser('/cache-life-short-stale', {
-          disableJavaScript: true,
-        })
+      // TODO(app-shells): fix this test
+      if (!partialPrefetching) {
```
gnoff, line 675: let's use @gate or @force-gate for these?

gnoff:
The deploy failure looks like it exposes a third path that still drops the RDC. Requests with Next-Router-Segment-Prefetch return early through generateRuntimePrefetchResult(), bypassing both locations updated in this PR. That function still calls createPrerenderResumeDataCache() with no source at app-render.tsx:1602.

After revalidation changes the backing cache value, the direct runtime prefetch can therefore render a different value from the page RDC, which matches the failure at resume-data-cache.test.ts:118. Should this path also use createPrerenderResumeDataCache(requestStore.resumeDataCache ?? undefined)?

PR #92615, opened 2026-04-10
[segment cache]: failing test for Activity restoring stale redirects after revalidation

ztanner:
When a server action revalidates a route and immediately redirects to it, the client can revive a previously preserved hidden `Activity` subtree for that route instead of using the fresh post-revalidation state. In the regression test, that meant replaying a stale redirect from an earlier render and bouncing back to the entry page even though the server had already rendered the protected page correctly.

This introduces a failing test for this particular scenario.

Issue #86577, opened 2025-11-27
[cacheComponents] Activity component route preservation causes significant breakage in application logic, UI behavior and E2E tests


SorooshGb:
### Link to the code that reproduces this issue

https://github.com/SorooshGb/next16-activity-breakage-repro

### This reproduction demonstrates **3** issues caused by Activity-component–driven route preservation when `cacheComponents: true` is enabled

1. Clone the reproduction repo linked above.

2. Install dependencies:
   npm install
   npx playwright install chromium

3. Start the dev server:
   npm run dev

4. Test the following behaviors in the browser:

**A. Dropdown stays open**

- On the home page, open the dropdown once. In this reproduction, the dropdown’s default
  close-on-select behavior is intentionally prevented because of application logic requirements.
- Click any dropdown button that navigates to an Item page.
- Navigate back to the home page.  
  → The dropdown is still open because the component never unmounts.

**B. Dialog behavior breaks because mount-time logic never runs again**

Each row in the dropdown contains two navigation buttons:

- **“View Item”** (eye icon) → navigates to `/item/[id]` without search params  
- **“Add Item”** (plus icon) → navigates to `/item/[id]?newEntry=true`

Expected behavior:
- “Add Item” → dialog should open on arrival  
- “View Item” → dialog should remain closed

Actual behavior:

1. Reload the home page to start clean.
2. Open the dropdown and click **“Add Item”** (`?newEntry=true`).  
   → Dialog opens as expected.
3. Use the browser **Back** button to return home.
4. Click **“View Item”** (no `newEntry` param).  
   → The dialog is still open even though the URL does not include the param.

Now demonstrate that once closed, the dialog never opens again:

5. Reload the home page again.
6. Open the dropdown and click **“Add Item”** (`?newEntry=true`), then close the dialog manually.
7. Navigate back to the home page.
8. Open the dropdown and click **“Add Item”** again (`?newEntry=true`).  
   → The dialog does **not** open, even though the URL includes the param.

This shows that mount-time initialization (including logic based on search params) only runs
once, and the dialog’s open state becomes permanently tied to stale state.

**C. E2E test fails due to duplicate DOM**

With the dev server still running, execute:

npm run test:e2e

This test visits `/sign-in` first (which contains Email and Password fields), then clicks the
link to navigate to `/sign-up` (which also contains Email and Password fields).

Because Activity keeps the previous route mounted, the hidden `/sign-in` page remains in the
[...]

PR #98006, opened 2026-08-28
Fix stale redirect marker in hidden Activity causing hard navigation

### LINKED ISSUE #97989: [cacheComponents] stale redirect marker in hidden Activity route causes later Links to full-page navigate

https://github.com/maxesse/next-redirect-marker-repro

### To Reproduce

1. Clone the reproduction and run it with `next build && next start`.
2. Visit the redirect-capable route. It renders a component behind Suspense which calls `redirect()`, causing Next to add `#__next-page-redirect` to the streamed document.
3. Navigate away using `<Link>`. With Cache Components enabled, the previous route is retained in a hidden React Activity.
4. Confirm that `document.getElementById('__next-page-redirect')` still finds the marker in the hidden route.
5. Click another unrelated same-origin `<Link>`.

The final click performs a full document navigation instead of a normal App Router navigation.

The reproduction includes a small `window` sentinel and network checks to make the document navigation easy to see. Without the stale marker, the same Link click is a normal SPA navigation.

I reproduced this on Next 16.3.3 and 16.4.0-canary.8.

### Current behavior

A redirect marker belonging to an inactive route retained in a hidden Activity affects later navigation from the active route.

The cause appears to be the document-wide check in `navigate-reducer`:

```ts
if (document.getElementById('__next-page-redirect')) {
  return completeHardNavigation(state, url, navigateType)
}
```

Because the lookup searches the whole document, it also finds a redirect marker inside an inactive Activity subtree and turns an unrelated same-origin Link click into a full-page navigation.

I originally found this while using `@next/playwright`'s `instant()` helper. In that case the unexpected MPA navigation lands on the PPR shell, which makes the failure initially look like cached content was not prefetched.

### Expected behavior

A redirect marker belonging to an inactive hidden route should not affect navigation from the currently active route.

Once a streamed redirect is no longer relevant, its marker should either be removed/consumed, or the navigation check should only consider a redirect belonging to the active route.

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.6.0: Fri Jul 31 19:17:26 PDT 2026; root:xnu-12377.161.14~5/RELEASE_ARM64_T6041
  Available memory (MB): 36864
  Available CPU cores: 14
Binaries:
  Node: 24.16.0
  npm: 12.0.2
  Yarn: 1.22.22
  pnpm: 11.21.0
Relevant Packages:
  next: 16.4.0-canary.9
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

cacheComponents, Linking and Navigating, Redirects, Loading UI and Streaming

### Additional context

This looks related to #92767 and the Next team regression test in #92615, but I believe it is a separate case.

Those concern stale RedirectBoundary/route state being restored from Activity. This reproduction is specifically about the streamed `__next-page-redirect` DOM marker remaining in an inactive route and later being found by the document-global check in `navigate-reducer`.

It is also adjacent to the broader hidden Activity behavior tracked in #86577.

The real application where I found this uses a request-time authentication/access check behind Suspense, which may call `redirect('/sign-in')`. This is why a streamed redirect can occur during an otherwise valid cached navigation, but the minimal reproduction does not use an auth library.

pranjal2410719:
## What?
 
When `cacheComponents` is enabled, React's `<Activity mode="hidden">` retains previous routes in the DOM with `display: none`. A `<meta id="__next-page-redirect">` marker left inside such a hidden route was found by the document-global `getElementById` check in `navigateReducer`, causing unrelated `<Link>` clicks to trigger a full-page (MPA) navigation instead of normal App Router navigation.
 
The fix replaces the document-global `getElementById` check with a scoped approach: query all `#__next-page-redirect` markers, then walk up each element's ancestors looking for `display: none` or `hidden` — the signature of a hidden Activity root. We start from `parentElement` (not the element itself) because `<meta>` elements are inherently `display: none`.
 
## Why?
 
With `cacheComponents` enabled, navigating away from a route that triggered a streamed `redirect()` leaves the redirect marker in the hidden Activity subtree. A subsequent `<Link>` click on a completely unrelated route finds this stale marker and performs a hard navigation, breaking the expected SPA behavior.
 
## How?
 
- Added `hasActiveRedirectMarker()` helper that queries all markers and checks visibility via ancestor walk
- Added `isInVisibleSubtree()` helper that checks for `display: none` or `hidden` ancestors
- Replaced `document.getElementById('__next-page-redirect')` with `hasActiveRedirectMarker()`
- Added 4 e2e tests verifying the scoped check works correctly for hidden, visible, and mixed marker scenarios
 
Fixes #97989
Related to #92615, #86577

### DIFF
--- a/packages/next/src/client/components/router-reducer/reducers/navigate-reducer.ts
+++ b/packages/next/src/client/components/router-reducer/reducers/navigate-reducer.ts
@@ -11,6 +11,56 @@ import {
 import { getStaleTimeMs } from '../../segment-cache/cache'
 import { FreshnessPolicy } from '../../render-tree'
 
+/**
+ * Check if any `__next-page-redirect` marker exists outside a hidden Activity
+ * subtree. When `cacheComponents` is enabled, React's `<Activity>` retains
+ * hidden routes in the DOM with `display: none`. A stale redirect marker from
+ * such a route must not trigger a hard (MPA) navigation from the active route.
+ *
+ * This replaces the previous document-global `getElementById` check.
+ */
+function hasActiveRedirectMarker(): boolean {
+  const markers = document.querySelectorAll('#__next-page-redirect')
+  for (let i = 0; i < markers.length; i++) {
+    if (isInVisibleSubtree(markers[i])) {
+      return true
+    }
+  }
+  return false
+}
+
+/**
+ * Walk up from an element's parent to determine whether it sits inside a
+ * hidden subtree. React's `<Activity mode="hidden">` applies
+ * `display: none` to its root; other mechanisms may use the `hidden`
+ * attribute.
+ *
+ * We start from `element.parentElement` (not `element` itself) because the
+ * marker is a `<meta>` element, which inherently has `display: none`.
+ */
+function isInVisibleSubtree(element: Element): boolean {
+  // A marker emitted into `<head>` (e.g. a redirect captured before the first
+  // flush) is always active. The browser's UA stylesheet applies
+  // `display: none` to `<head>`, so we must not treat it as a hidden Activity
+  // subtree. Scope the visibility walk to the rendered `<body>` subtree.
+  if (document.head.contains(element)) {
+    return true
+  }
+  let current: Element | null = element.parentElement
+  while (
+    current &&
+    current !== document.documentElement &&
+    current !== document.body
+  ) {
+    const style = window.getComputedStyle(current)
+    if (style.display === 'none' || current.hasAttribute('hidden')) {
+      return false
+    }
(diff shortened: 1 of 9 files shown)


eps1lon (commented):
We recently changed how `<meta>` tags are handled in `Activity`. If these are hoisted, React should remove them now if they're inside a hidden Activity. Can you split the PR apart into characterizing test and the fix so that it's easier to confirm the previous behavior?

PR #98041, opened 2026-08-28
[PPF] Pass in original searchParams for private-cache pages

lubieowoce:
The handing for page components marked with `use cache: private` was assuming that searchParams can never hang in `prerender-runtime`, so it passed the serialized `innerSearchParams` to the page, which we expected to just be a resolved promise. This *used to* be correct when runtime prerenders always had search params available, because the serialized search params were always equivalent to `outerSearchParams`.

However, for `partialPrefetching` we started using `prerender-runtime` for runtime shells, where search params are not available, and if a private cache awaits them, it needs to abort filling and become dynamic, same as when it awaits params. So we need to trigger `dynamicAccessAbortController` when `searchParams` is awaited:

```tsx
export default async function Page({ searchParams }) {
  'use cache: private'
  // this should trigger `dynamicAccessAbortController.abort()`, so we need the original instrumented
  // searchParams object instead of the serialized one
  await searchParams
}
```

This is fixed by using the outer searchParams object in the `isPageSegmentFunction` codepath of `use-cache-wrapper` -- we're preserving the instrumented promise, so the cache prerender will abort as expected.
(before this fix, a shell prefetch of such a page would get a deserialized hanging promise for `searchParams` and hang until it times out)

2 of the 3 newly added tests ("params in a public cache" and "params in a private cache") were already passing, because `params` was already being preserved, but i'm adding them here to prevent regressions. The "search params in a private cache" one was failing due to the cache timing out, and is now passing. I'm not covering "search params in public cache" because that's an error and is already tested elsewhere.

### DIFF
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -2058,18 +2058,16 @@ export async function cache(
     args = [props, ...otherOuterArgs]
 
     fn = {
-      [name]: async (
-        {
-          params: _innerParams,
-          searchParams: innerSearchParams,
-        }: UseCachePageInnerProps,
-        ...otherInnerArgs: unknown[]
-      ) =>
+      [name]: async (_: UseCachePageInnerProps, ...otherInnerArgs: unknown[]) =>
         originalFn.apply(null, [
           {
-            params: outerParams,
+            params: props.params,
             searchParams:
-              innerSearchParams ??
+              // Preserve the original search params, if this cache can access them.
+              // Notably, in a runtime shell private caches can resolve, but search params
+              // will be hanging, and we need to preserve the original proxied promise object
+              // to trigger `dynamicAccessAbortSignal` when they're accessed.
+              props.searchParams ??
               // For public caches, search params are omitted from the cache
               // key (and the serialized args) to avoid mismatches between
               // prerendering and resuming a cached page that does not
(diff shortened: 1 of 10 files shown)


packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -2164,13 +2162,8 @@ export async function cache(
 
   switch (workUnitStore.type) {
     case 'prerender-runtime':
-    // We're currently only using `dynamicAccessAsyncStorage` for params,
-    // which are always available in a runtime prerender, so they will never hang,
```
lubieowoce, line 2168: > so they will never hang, effectively making the tracking below a no-op.

ah yes i remember writing this 😅

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -2164,13 +2162,8 @@ export async function cache(
...
-    // effectively making the tracking below a no-op.
-    // However, a runtime prerender shares a lot of the semantics with a static prerender,
-    // and might need to follow this codepath in the future
-    // if we start using `dynamicAccessAsyncStorage` for other APIs.
-    //
+    // A runtime prerender may be a runtime shell, which does not have acces to
```
```suggestion
    // A runtime prerender may be a runtime shell, which does not have access to
```

unstubbable (approved):

packages/next/src/server/use-cache/use-cache-wrapper.ts
lubieowoce: goddamn i keep forgetting to run a robot over my comments to check for spelling

PR #98055, opened 2026-08-28
durableUseCacheEntries.unstableEnvVars and ignoredEnvVars

mischnic:
```ts
type durableUseCacheEntries = {
	// Env vars that are constantly changing (e.g. the deployment ID).
	// If a cache entry depends on one of these, it will not be reused across deployments at all. 
    // NEXT_DEPLOYMENT_ID is automatically added to this list (and is the default), so it does not need to be specified.
	unstableEnvVars?: string[]
	// Env vars that do not change the cache entry, but might change often (e.g. passwords, LRU
	// cache stale time, OIDC tokens). Note that a token include also a database name, in which
	// case it should not be in this list.
	ignoredEnvVars?: string[]
}
```

### DIFF
--- a/crates/next-api/src/server_actions.rs
+++ b/crates/next-api/src/server_actions.rs
@@ -235,8 +235,10 @@ impl Asset for ServerActionManifestAsset {
         let actions_value = self.actions.await?;
         let async_module_info = self.module_graph.async_module_info();
         let next_config = self.project.next_config();
-        let durable_use_cache_entries = next_config.enable_durable_use_cache_entries();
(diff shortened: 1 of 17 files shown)


mischnic:
* **#98055** <a href="https://app.graphite.com/github/pr/vercel/next.js/98055?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/98055?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

PR #98090, opened 2026-08-31
Use intrusive locks for task storage

lukesandberg:
## Summary

Keep the resident DashMap, but move ordinary task mutation off DashMap shard write guards. Values are stored as `Box<TaskSlot>`, where opaque `TaskSlot(UnsafeCell<TaskStorage>)` owns the protocol-based unsafe access boundary:

- embed a one-byte `parking_lot::RawMutex` in `TaskStorage` without changing its 128-byte layout
- hold a DashMap shard **read** guard for task lifetime, with the intrusive lock providing per-task mutable exclusion
- use DashMap's atomic `RefMut::downgrade` after insertion so caller work never retains the shard write guard
- share one shard read guard for same-shard pair access; order distinct shard guards by shard index and task locks by `TaskId`
- retain shard write guards for structural insertion, removal, drain, eviction, and clear operations
- lock tasks during snapshot keep-mode and GC candidate scans now that shard read guards permit concurrent task mutation
- reduce resident/snapshot map sharding from 16× to **2× workers²**, selected after an A/B of factors 1/2/4/16

This is intentionally the first incremental step of a proposed stack:

1. **this PR:** intrusive locks while retaining DashMap lifetime/structural management
2. evaluate Papaya to eliminate common-path map reader locking
3. evaluate dense Boxcar + Kovan storage after the map comparison

The previous dense prototype is retained by signed tip `fbea6793a61835f8cc9c83d6af67de79385a63ff`, but is no longer in this PR's history.

## Results

Compared with canary `8ea76d64`:

- representative access mix: **98.04% existing hits, 1.96% actual insertions**
- `overhead.rs`: no statistically detected candidate regression
- statistically detected wins for cached different-key workloads and short parallel uncached workloads
- five alternating real-build samples per revision:
  - cold median: **20.59s → 20.41s** (−0.87%)
  - warm median: **19.91s → 20.28s** (+1.86%)
  - peak RSS effectively flat

The real build is neutral within run-to-run variance.

The shard-factor sweep selected **2× workers²**:

- GC: −5.6% at 10k garbage tasks and −2.8% at 200k versus the previous 16× factor
- real build: neutral/slightly faster (20.43s cold / 20.24s warm versus 20.53s / 20.35s)
- factor 2 reduced the shortest parallel-uncached regression seen at factor 1 while preserving different-key improvements

Focused lock-only microbenchmarks expose an important trade-off:

- existing hit: 14.44ns shard-write baseline vs 30.95ns read+intrusive
- insert/remove roundtrip: 46.54ns vs 62.21ns
- independent same-shard tasks with near-zero work: 182µs vs 380µs
- same-shard pair: 14.86ns vs 46.34ns
- 4,096-task serial scan: 4.91µs vs 50.88µs

The extra task lock is not a universal fast-path win: DashMap's shared shard-read count still changes one contended cache line, then the task lock adds a second transition. Integrated useful work can overlap and wins in `overhead.rs`, but **PR #2 is likely required** to remove the shard-reader bottleneck and realize the intended scaling.

## Verification

- `cargo check -p turbo-tasks-backend --all-targets`
- `cargo check -p turbo-tasks-backend --features trace_task_modification`
- `cargo test -p turbo-tasks-macros`
- `pnpm build-all`
- exact Criterion A/B for `overhead.rs` and the new GC benchmark, including the 1×/2×/4×/16× shard-factor sweep
- focused same-task, same-shard, pair-ordering, insertion, structural-writer, and snapshot-lock tests
- five alternating cold/warm `bench/heavy-npm-deps` builds per revision with GNU time/RSS

## Notes

- `with_task` now serializes on the task lock; its two current callers are test-only count hooks. Same-task re-entry is documented as invalid.
- `StorageWriteGuard` uses localized protocol-based unsafe access because it must own the DashMap read guard that keeps the boxed slot alive. An unlocked `TaskSlot` exposes no task reference; lock projection happens through the raw cell pointer; only a lock-owning guard materializes task references; structural shard-write paths use exclusive slot access; unlock precedes map-guard drop. The guard remains `!Send`.
- The required trace-feature check exposed an existing cfg-only undefined `modified` variable in the touched function; this PR computes it once for both tracing and branch selection.

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/benches/storage_locking.rs
+++ b/turbopack/crates/turbo-tasks-backend/benches/storage_locking.rs
@@ -1,6 +1,7 @@
 use std::{cell::UnsafeCell, hash::BuildHasher, hint::black_box, sync::Arc, thread};
 
 use criterion::{BatchSize, Criterion, Throughput};
+use crossbeam_utils::CachePadded;
 use dashmap::DashMap;
 use parking_lot::{RawMutex, lock_api::RawMutex as RawMutexTrait};
 use rustc_hash::FxBuildHasher;
@@ -8,11 +9,9 @@ use rustc_hash::FxBuildHasher;
 const SHARDS: usize = 64;
 const TASKS: u32 = 4096;
 
-#[repr(C)]
 struct Payload {
     lock: RawMutex,
     value: UnsafeCell<u64>,
-    padding: [u64; 14],
 }
 
 impl Payload {
@@ -20,7 +19,6 @@ impl Payload {
         Self {
             lock: <RawMutex as RawMutexTrait>::INIT,
             value: UnsafeCell::new(value),
-            padding: [0; 14],
         }
     }
 
@@ -46,7 +44,8 @@ impl Drop for PayloadGuard<'_> {
     }
 }
 
-type Map = DashMap<u32, Box<Payload>, FxBuildHasher>;
+// Keep independently locked tasks on separate cache lines without hand-written layout padding.
+type Map = DashMap<u32, Box<CachePadded<Payload>>, FxBuildHasher>;
 
 fn empty_map() -> Map {
     DashMap::with_capacity_and_hasher_and_shard_amount(TASKS as usize, FxBuildHasher, SHARDS)
@@ -55,7 +54,7 @@ fn empty_map() -> Map {
 fn populated_map() -> Map {
     let map = empty_map();
     for id in 1..=TASKS {
-        map.insert(id, Box::new(Payload::new(id as u64)));
+        map.insert(id, Box::new(CachePadded::new(Payload::new(id as u64))));
     }
     map
 }
@@ -102,7 +101,7 @@ pub fn storage_locking(criterion: &mut Criterion) {
             key = key.wrapping_add(1);
             let mut task = map
                 .entry(current)
-                .or_insert_with(|| Box::new(Payload::new(0)));
+                .or_insert_with(|| Box::new(CachePadded::new(Payload::new(0))));
             *task.value.get_mut() = black_box(current as u64);
             drop(task);
             map.remove(&current);
@@ -116,7 +115,7 @@ pub fn storage_locking(criterion: &mut Criterion) {
             key = key.wrapping_add(1);
             let task = map
                 .entry(current)
-                .or_insert_with(|| Box::new(Payload::new(0)))
+                .or_insert_with(|| Box::new(CachePadded::new(Payload::new(0))))
                 .downgrade();
             task.with_mut(|value| *value = black_box(current as u64));
             drop(task);
--- a/turbopack/crates/turbo-tasks-backend/benches/storage_locking.rs
+++ b/turbopack/crates/turbo-tasks-backend/benches/storage_locking.rs
@@ -1,0 +1,238 @@
+use std::{cell::UnsafeCell, hash::BuildHasher, hint::black_box, sync::Arc, thread};
+
+use criterion::{BatchSize, Criterion, Throughput};
+use dashmap::DashMap;
+use parking_lot::{RawMutex, lock_api::RawMutex as RawMutexTrait};
+use rustc_hash::FxBuildHasher;
+
+const SHARDS: usize = 64;
+const TASKS: u32 = 4096;
+
+#[repr(C)]
+struct Payload {
+    lock: RawMutex,
+    value: UnsafeCell<u64>,
+    padding: [u64; 14],
+}
+
+impl Payload {
+    fn new(value: u64) -> Self {
+        Self {
+            lock: <RawMutex as RawMutexTrait>::INIT,
+            value: UnsafeCell::new(value),
+            padding: [0; 14],
+        }
+    }
+
+    fn with_mut<R>(&self, f: impl FnOnce(&mut u64) -> R) -> R {
+        self.lock.lock();
+        let guard = PayloadGuard(self);
+        // SAFETY: `guard` owns this payload's lock for the closure's lifetime.
+        let result = f(unsafe { &mut *self.value.get() });
+        drop(guard);
+        result
+    }
+}
+
+// SAFETY: `value` is accessed only while `lock` is held.
+unsafe impl Sync for Payload {}
+
+struct PayloadGuard<'a>(&'a Payload);
+
+impl Drop for PayloadGuard<'_> {
+    fn drop(&mut self) {
+        // SAFETY: This guard is created only after acquiring this payload's lock.
+        unsafe { self.0.lock.unlock() };
+    }
+}
+
+type Map = DashMap<u32, Box<Payload>, FxBuildHasher>;
+
+fn empty_map() -> Map {
+    DashMap::with_capacity_and_hasher_and_shard_amount(TASKS as usize, FxBuildHasher, SHARDS)
+}
+
+fn populated_map() -> Map {
+    let map = empty_map();
+    for id in 1..=TASKS {
+        map.insert(id, Box::new(Payload::new(id as u64)));
+    }
+    map
+}
+
+fn same_shard_keys(map: &Map, count: usize) -> Vec<u32> {
+    let mut keys = Vec::with_capacity(count);
+    let target = map.determine_shard(map.hash_usize(&1));
+    for key in 1.. {
+        if map.determine_shard(map.hash_usize(&key)) == target {
+            keys.push(key);
+            if keys.len() == count {
+                return keys;
+            }
+        }
+    }
+    unreachable!()
+}
+
+pub fn storage_locking(criterion: &mut Criterion) {
+    let map = populated_map();
+    let mut hit = criterion.benchmark_group("storage_locking_hit");
+    hit.throughput(Throughput::Elements(1));
+    hit.bench_function("dashmap_shard_write", |b| {
+        b.iter(|| {
+            let mut task = map.get_mut(black_box(&1)).unwrap();
+            task.value = UnsafeCell::new(black_box(1));
+        })
+    });
+    hit.bench_function("shard_read_intrusive", |b| {
+        b.iter(|| {
+            let task = map.get(black_box(&1)).unwrap();
+            task.with_mut(|value| *value = black_box(1));
+        })
+    });
+    hit.finish();
+
+    let mut insertion = criterion.benchmark_group("storage_locking_insert_roundtrip");
+    insertion.throughput(Throughput::Elements(1));
+    insertion.bench_function("dashmap_shard_write", |b| {
+        let map = empty_map();
+        let mut key = 1_u32;
+        b.iter(|| {
+            let current = key;
+            key = key.wrapping_add(1);
+            let mut task = map
+                .entry(current)
+                .or_insert_with(|| Box::new(Payload::new(0)));
+            *task.value.get_mut() = black_box(current as u64);
+            drop(task);
+            map.remove(&current);
+        })
+    });
+    insertion.bench_function("downgrade_then_intrusive", |b| {
+        let map = empty_map();
+        let mut key = 1_u32;
+        b.iter(|| {
+            let current = key;
+            key = key.wrapping_add(1);
+            let task = map
+                .entry(current)
+                .or_insert_with(|| Box::new(Payload::new(0)))
+                .downgrade();
+            task.with_mut(|value| *value = black_box(current as u64));
+            drop(task);
+            map.remove(&current);
+        })
+    });
+    insertion.finish();
+
+    let mut contention = criterion.benchmark_group("storage_locking_same_shard_parallel");
+    const THREADS: usize = 4;
+    const OPS: usize = 1000;
+    contention.throughput(Throughput::Elements((THREADS * OPS) as u64));
+    let keys = same_shard_keys(&map, THREADS);
+    contention.bench_function("dashmap_shard_write", |b| {
+        b.iter(|| {
+            thread::scope(|scope| {
+                for &key in &keys {
+                    let map = &map;
+                    scope.spawn(move || {
+                        for _ in 0..OPS {
+                            let mut task = map.get_mut(&key).unwrap();
+                            *task.value.get_mut() = black_box(key as u64);
+                        }
+                    });
+                }
+            });
+        })
+    });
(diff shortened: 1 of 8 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,683 @@
...
+const BITMAP_WORD_BITS: usize = u64::BITS as usize;
+pub(crate) const BITMAP_WORDS: usize = CHUNK_SIZE / BITMAP_WORD_BITS;
+
+/// Value stored in an always-initialized intrusive task slot.
+///
+/// # Safety
+///
+/// Implementations must keep the lock at a stable address, initialize it in `EMPTY`, require the
+/// lock for every payload/presence access, and leave the source lock untouched when vacating.
+pub(crate) unsafe trait TaskSlotValue: Sized {
```
lukesandberg: i don't think we really need a trait here? i guess it is really to make testing simpler?

turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,683 @@
...
+impl<T: TaskSlotValue> TaskSlot<T> {
+    pub(crate) const fn empty() -> Self {
+        Self(UnsafeCell::new(T::EMPTY))
+    }
+
+    fn lock(&self) -> TaskSlotGuard<'_, T> {
+        // SAFETY: The value is initialized by `empty` and never moved after its chunk is published.
+        // Calling `lock` only reads/mutates the embedded raw mutex, whose implementation provides
+        // the synchronization for all subsequent accesses through the returned guard.
+        unsafe { &*self.0.get() }.lock();
```
lukesandberg: there are no guards for RawMutex? an existing parking_lot mechanism we can use?

turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,683 @@
...
+            }
+            let word = self.words.get(self.word_index)?;
+            self.word_index += 1;
+            self.bits = word.load(Ordering::Acquire);
+        }
+    }
+}
+
+struct ChunkedVec<T: TaskSlotValue> {
+    chunks: boxcar::Vec<OnceLock<Box<TaskChunk<T>>>>,
```
lukesandberg: why do we need the OnceLocks? to lazily allocate slots?  i guess this lets us skip intermediate allocations if we need task 1 and then task 1M in a warm build scenario

the downside is an additional atomic to traverse on every access


So i guess we should consider an alternative that avoids allocating 'empty' TaskStorage slots

Idea 1 is:
* go back to pointers?   We could have a parallel pool of TaskStorage structs (logically but not literally a Vec<TaskStorage>) then the main TaskChunk just stores `u32` indices into that vec.

If we want to iterate all storage tasks we can just walk that 'pool' vec but it makes removals tricky (so a literal vec is not the right choice since removals invalidate indices

* so alternatively we could use `Box<TaskStorage>` again, but then we have a bootstrapping problem where we would need additional locks or atomics to support 'insert' (arc_swap so threads aggree on whether a slot was initialized... but then we need to do an atomic read on every access)

(i suppose we are really seeing the benefit of dashmap shard locks simplifying these transitions)


idea 2 is just commit to always allocating the range of chunks between `[0, max-restored-id)`

this is a high water mark we need to support anyway, so we aren't in principle raising the RAM ceiling just when we achieve it

This is certainly the simplest thing to try

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -430,108 +371,83 @@ impl Storage {
...
 
     pub fn access_mut(&self, key: TaskId) -> StorageWriteGuard<'_> {
-        let inner = match self.map.entry(key) {
-            dashmap::mapref::entry::Entry::Occupied(e) => e.into_ref(),
-            dashmap::mapref::entry::Entry::Vacant(e) => e.insert(Box::new(TaskStorage::new())),
-        };
+        let inner = self
+            .map
+            .get(key)
+            .unwrap_or_else(|| self.map.get_or_insert(key));
```
lukesandberg: should this just be `self.map.get_or_insert`?

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -430,108 +371,83 @@ impl Storage {
...
+        } else {
+            (key2, key1, true)
+        };
+        let lower = self.access_mut(lower);
+        let upper = self.access_mut(upper);
+        if reversed {
+            (upper, lower)
+        } else {
+            (lower, upper)
+        }
```
lukesandberg: wouldn't 

```
    assert_ne!(key1, key2, "cannot mutably access the same task twice");
    if key1 < key2 {
        let val1 = self.access_mut(key1);
        let val2 = self.access_mut(key2);
        (val1, val2)
    } else {
        let val2 = self.access_mut(key2);
        let val1 = self.access_mut(key1);
        (key1, key2)
    }
```

be a simpler implementation for the function?

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -565,40 +481,29 @@ impl Storage {
             "evict_after_snapshot must not be called during snapshot mode"
         );
 
-        let counts: Vec<EvictionCounts> = parallel::map_collect(self.map.shards(), |shard| {
-            let mut shard = shard.write();
+        let chunks = self.map.chunks();
+        let counts: Vec<EvictionCounts> = parallel::map_collect(&chunks, |&chunk| {
```
lukesandberg: perhaps our new datastructure should offer a 'parallel_collect' api that does this directly

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -565,40 +481,29 @@ impl Storage {
...
-            // task_cache removals that we couldn't perform inline because the target shard
-            // was contended. We defer them until after the map shard lock is released to
-            // avoid a lock cycle with get_or_create_persistent_task, which takes task_cache
-            // before map. Allocated lazily on first conflict.
+            // Removals that would invert task_cache -> task lock ordering are deferred until no
+            // task lock is held.
             let mut deferred_task_cache_removals: Vec<CachedTaskTypeArc> = Vec::new();
-            shard.retain(|(task_id, task)| {
+            for offset in chunk.chunk.probably_occupied_offsets() {
+                let Some(mut task) = chunk.get(offset) else {
```
lukesandberg: why not an iter_mut _style_ api that does essentially this?

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -671,9 +565,21 @@ impl Storage {
     }
 }
 
+/// Exclusive access to one resident task.
+///
+/// `TaskMapGuard` mirrors parking_lot's default non-send guard marker. Keeping this wrapper `!Send`
+/// prevents a task lock from being held across an `.await` in a sendable future.
```
lukesandberg: drop this comment, no need to re-state type constraints on the implementation

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -47,6 +48,48 @@ type AutoSet<K, const I: usize> = auto_hash_map::AutoSet<K, BuildHasherDefault<F
...
+
+impl fmt::Debug for IntrusiveTaskLock {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("IntrusiveTaskLock").finish_non_exhaustive()
+    }
+}
+
+impl ShrinkToFit for IntrusiveTaskLock {
+    fn shrink_to_fit(&mut self) {}
+}
```
lukesandberg, line 163: so the newtype is necessary for these reasons? impl Debug and ShrinkToFit?

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -1752,6 +1918,47 @@ mod tests {
...
+            size_of::<Mutex<Option<TaskStorage>>>(),
+            136,
+            "inline task slot size changed"
+        );
+        assert_eq!(size_of::<RawMutex>(), 1, "parking_lot RawMutex grew");
+        assert_eq!(
+            size_of::<IntrusiveTaskLock>(),
+            1,
+            "intrusive lock wrapper grew"
+        );
```
lukesandberg: drop these asserts, we are no longer using the Mutex approach and we shouldn't assert on RawMutex

Asserting on IntrusiveTaskLock is fine but low value as it is covered by the overall TaskStorage size assertions

turbopack/crates/turbo-tasks-macros/src/derive/task_storage_macro.rs
```diff
@@ -1406,13 +1426,51 @@ fn generate_typed_storage_struct(grouped_fields: &GroupedFields) -> TokenStream
...
+            #[doc = "Returns whether this slot is authoritatively occupied."]
+            pub(crate) fn is_occupied(&self) -> bool {
+                self.occupied
+            }
+
+            #[doc = "Marks a vacant slot occupied. The caller must hold the intrusive lock."]
+            pub(crate) fn occupy(&mut self) {
+                debug_assert!(!self.occupied);
+                self.occupied = true;
+            }
```
lukesandberg: simple methods like this are often implementated 'by hand' in storage_schema.rs

turbopack/crates/turbo-tasks-macros/src/derive/task_storage_macro.rs
```diff
@@ -1406,13 +1426,51 @@ fn generate_typed_storage_struct(grouped_fields: &GroupedFields) -> TokenStream
...
+
+            #[doc = "Drops and resets payload in place while preserving the held mutex."]
+            #[doc = "The caller must hold this task's intrusive lock."]
+            pub(crate) fn vacate_in_place(&mut self) {
+                debug_assert!(self.occupied);
+                #(self.#field_names = Default::default();)*
+                #reset_flags
+                #reset_lazy
+                self.occupied = false;
```
lukesandberg: do we really need both of these?  isn't take_and_vacate sufficient? i suppose it wastes a small amount of work?

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -175,7 +176,8 @@ pub struct Storage {
     /// Indexed by `map.determine_shard(map.hash_usize(&key))` and guaranteed by construction so
     /// that  `shard_modified_counts.len()==map.shards().len()`
     ///
-    /// Should only be modified while holding the corresponding dashmap shard lock.
+    /// Should only be modified while holding either the task's intrusive lock and shard read
+    /// guard, or the corresponding DashMap shard write lock.
```
lukesandberg: im not sure there is a correctness issue here.... do we need this documentation?

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
```diff
@@ -748,7 +762,24 @@ impl Storage {
 
 pub struct StorageWriteGuard<'a> {
     storage: &'a Storage,
-    inner: RefMut<'a, TaskId, Box<TaskStorage>>,
+    inner: Ref<'a, TaskId, Box<TaskStorage>>,
+    task: *mut TaskStorage,
```
lukesandberg: why do we need the task field? access via `inner`?

turbopack/crates/turbo-tasks-macros/src/derive/task_storage_macro.rs
```diff
@@ -1406,6 +1406,8 @@ fn generate_typed_storage_struct(grouped_fields: &GroupedFields) -> TokenStream
             #(#field_defs,)*
             #flags_field
             #lazy_field
+            #[doc = "Per-task lock used with a shared resident-map guard."]
+            pub(crate) lock: IntrusiveTaskLock,
```
lukesandberg: why does this need to be pub?

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
```diff
@@ -47,6 +48,61 @@ type AutoSet<K, const I: usize> = auto_hash_map::AutoSet<K, BuildHasherDefault<F
...
+    /// The current thread must own this lock and perform no protected access after this call.
+    unsafe fn unlock(&self) {
+        // SAFETY: Forwarded from the caller.
+        unsafe { self.0.unlock() };
+    }
+
+    #[cfg(test)]
+    fn is_locked(&self) -> bool {
+        self.0.is_locked()
+    }
```
bgw, line 78: If you really have to implement the wrapper type, just implement `Deref` instead of reimplementing all the methods.

turbopack/crates/turbo-tasks-backend/src/backend/storage_schema.rs
bgw: We shouldn't ever require a no-op `ShrinkToFit` impl though. The derive macro uses specialization hacks to treat types without a `ShrinkToFit` impl as a no-op.

turbopack/crates/turbo-tasks-backend/benches/storage_locking.rs
```diff
@@ -0,0 +1,238 @@
...
+
+const SHARDS: usize = 64;
+const TASKS: u32 = 4096;
+
+#[repr(C)]
+struct Payload {
+    lock: RawMutex,
+    value: UnsafeCell<u64>,
+    padding: [u64; 14],
+}
```
bgw: Not very clear why this is using `repr(C)` + `padding` here? Is this to make sure it's cache-aligned? We should use `CachePadded<_>` for that.

turbopack/crates/turbo-tasks-backend/src/backend/storage.rs
bgw, line 224: We can decrease the number of shards if we have fine-grained locks.

PR #98102, opened 2026-08-31
Fix fallback params in request-specific renders

timneutkens:
## What?

Keep concrete request parameters out of Cache Components fallback placeholder handling. For `/[locale]/[[...filterSlugs]]`, a request to `/en` must expose an omitted `filterSlugs`, rather than an opaque `%%drp:…%%` value that breaks ordinary optional catch-all code.

The fix covers bot and draft-mode cache bypasses, fetch actions, and native multipart form actions. Native form responses render a complete document while retaining the request's resume data cache.

## How?

- Preserve concrete params during request-specific renders, while retaining unresolved params for prerender shells and action-only fallback requests.
- Recognize adapter-injected route placeholders through their provenance. Ordinary query parameters cannot erase pathname params, and literal encoded App Router pathname values are decoded only once. Pages Router captures retain their existing platform decoding.
- Handle adapter invocations of an optional catch-all route template with no captured params, including locale data requests.
- Record resume metadata for legacy deployment fallback shells whose router state still contains unresolved params. Adapter builds and self-hosted builds retain their static shell classifications.

## Testing

Regression coverage exercises query/pathname collisions, literal placeholder-like values at multiple encoding levels, omitted optional catch-alls, locale data requests, bot and draft bypasses, fetch/native actions, and base-path navigations with and without prefetching.

Validated with production Webpack and Turbopack tests, development HTTP tests, normalization unit tests, the cached-navigation suite, and focused Vercel deployments using both the adapter and legacy runtimes. The original interception failure also passed a fresh deployment run.

The latest canary merge includes the upstream PostCSS deployment guard fix (#98710).

### DIFF
--- a/packages/next/src/build/templates/app-page-runtime.ts
+++ b/packages/next/src/build/templates/app-page-runtime.ts
@@ -1490,6 +1490,11 @@ export function createAppPageEntrypoint({
             placeholderFallbackRouteParams != null &&
             placeholderFallbackRouteParams.length > 0
 
+          const isRequestSpecificRender =
+            !forceStaticRender &&
+            !isDebugPrerender &&
+            (supportsDynamicResponse || isPossibleServerAction)
+
           // When route-module.ts resolved partial nxtP* params during
           // background revalidation, filter fallbackRouteParams to only the
           // params that are still unresolved. This lets doRender produce an
@@ -1508,14 +1513,15 @@ export function createAppPageEntrypoint({
             }
           }
 
-          const fallbackRouteParams =
-            // In production or when debugging the static shell for a
-            // non-prerendered URL, use the prerender manifest's fallback route
-            // params which correctly identifies which params are unknown.
-            ((isProduction && getRequestMeta(req, 'renderFallbackShell')) ||
-              hasPlaceholderFallbackRouteParams ||
-              (isDebugStaticShell && !isPrerendered)) &&
-            fallbackRouteParamsForRender
+          const fallbackRouteParams = isRequestSpecificRender
+            ? null
+            : // In production or when debugging the static shell for a
+              // non-prerendered URL, use the prerender manifest's fallback route
+              // params which correctly identifies which params are unknown.
+              ((isProduction && getRequestMeta(req, 'renderFallbackShell')) ||
+                  hasPlaceholderFallbackRouteParams ||
+                  (isDebugStaticShell && !isPrerendered)) &&
+                fallbackRouteParamsForRender
               ? createOpaqueFallbackRouteParams(fallbackRouteParamsForRender)
               : // For intermediate shells where some params are resolved and
                 // others still have placeholders, use the filtered subset so the
@@ -1550,6 +1556,7 @@ export function createAppPageEntrypoint({
           // fallbackRouteParams because that would replace actual param values
           // with opaque placeholders during segment resolution.
           if (
+            (!isRequestSpecificRender || isPossibleServerAction) &&
             (isProduction || isDebugStaticShell) &&
             nextConfig.cacheComponents &&
             !isPrerendered &&
@@ -1563,13 +1570,6 @@ export function createAppPageEntrypoint({
               addRequestMeta(req, 'fallbackParams', fallbackParams)
             }
           }
-
-          // Forced static and debug renders are prerenders even when the
-          // surrounding request otherwise supports dynamic rendering.
-          const isRequestSpecificRender =
-            !forceStaticRender &&
-            !isDebugPrerender &&
-            (supportsDynamicResponse || isPossibleServerAction)
 
           // Perform the render.
           return doRender({
@@ -2101,7 +2101,11 @@ export function createAppPageEntrypoint({
         // would replace actual param values with opaque placeholders during
         // segment resolution; the resolved values are baked into the URL and
         // already interpolated into the postponed state.
-        if (nextConfig.cacheComponents && prerenderInfo?.fallbackRouteParams) {
+        if (
+          nextConfig.cacheComponents &&
+          prerenderInfo?.fallbackRouteParams &&
+          (!supportsDynamicResponse || isPossibleServerAction)
+        ) {
           const fallbackParams = createOpaqueFallbackRouteParams(
             prerenderInfo.fallbackRouteParams
           )
--- a/packages/next/src/build/templates/app-page-runtime.ts
+++ b/packages/next/src/build/templates/app-page-runtime.ts
@@ -1499,6 +1499,8 @@ export function createAppPageEntrypoint({
             placeholderFallbackRouteParams != null &&
             placeholderFallbackRouteParams.length > 0
 
+          // Forced static and debug renders are prerenders even when the
+          // surrounding request otherwise supports dynamic rendering.
           const isRequestSpecificRender =
             !forceStaticRender &&
             !isDebugPrerender &&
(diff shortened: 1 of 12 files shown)


packages/next/src/build/templates/app-page-runtime.ts
```diff
@@ -1550,6 +1556,7 @@ export function createAppPageEntrypoint({
           // fallbackRouteParams because that would replace actual param values
           // with opaque placeholders during segment resolution.
           if (
+            (!isRequestSpecificRender || isPossibleServerAction) &&
```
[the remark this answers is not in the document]
timneutkens: Fixed in 8545b51a782910abeb909b04add9cf2ebad82abc and refined in 765b95f18f40bbbf3a5fbdb0aaa7a8931baf84cf. Cached Navigation RSC responses remain dynamic render operations while retaining fallback params through request metadata for the staged static segment. The fallbackRouteParams render argument remains null, preserving concrete values during segment resolution as the existing comment documents. Verified the three cached-navigation production assertions (including initial HTML), the four bot/draft/Next-Action/multipart regressions, all 11 server-utils tests, and the Next.js package build.

packages/next/src/build/templates/app-page-runtime.ts
```diff
@@ -545,6 +554,17 @@ export function createAppPageEntrypoint({
     const supportsRDCForNavigations =
       isRoutePPREnabled && nextConfig.cacheComponents === true
 
+    // Cached Navigations stage a reusable static segment during production
+    // HTML renders, dynamic RSC renders, and postponed resumes. All of these
+    // paths need prerender fallback params even though the response itself is
+    // dynamic.
+    const isStagedCachedNavigationRender =
```
lubieowoce: idgi, why is the render using stages relevant? is this a shorthand for "we expect this to hit `generateStagedDynamicFlightRenderResultNode` in `app-render`"? if that's the case i'd pick a better name bc this feels confusing

timneutkens: It’s broader than generateStagedDynamicFlightRenderResultNode. Dynamic RSC requests use that function, while production HTML and postponed HTML resumes use the staged branch in renderToStream. The relevant property is that these renders produce a reusable static stage, so fallback params must be supplied through request metadata to defer them there while retaining concrete values for dynamic segment resolution. I agree the current name obscures that; needsFallbackParamsForCachedNavigationStage would describe the behavior better.

timneutkens: :robot: Good call. This is broader than generateStagedDynamicFlightRenderResultNode: dynamic RSC uses that function, while production HTML (including initial HTML and postponed resumes) uses the staged branch in renderToStream. The relevant behavior is that these paths produce a reusable static stage and therefore need fallback params in the RequestStore. Renamed the predicate to needsFallbackParamsForCachedNavigationStage and clarified the comment in 54c3c4a056.

unstubbable (commented):
My clanker reproduced two routing regressions using ordinary HTTP requests to App Route handlers with `cacheComponents: false`. Both pass on `canary`.

1. **An ordinary query parameter can erase a valid pathname parameter.**

   In `packages/next/src/server/route-modules/route-module.ts:956`, `omittedRouteParamFromQuery` treats any deleted route-named query value as grounds to override valid pathname params, without requiring platform provenance.

   For `/shop/[[...slug]]`, both `/shop/books?slug=index` and `/shop/books?slug=` return `params: {}` instead of `{ slug: ['books'] }`.

   Can we restrict this precedence to actual platform route parameters, rather than any deleted query value?

2. **A literal encoded pathname value is mistaken for an omitted optional catch-all.**

   In `packages/next/src/server/server-utils.ts:186`, the encoding alternative in `isProxyOptionalDefaultValue` also applies to pathname matches without `routeParamKeys`.

   For `/[locale]/[[...filterSlugs]]`, requesting `/en/%255B%255B...filterSlugs%255D%255D` returns only `{ locale: 'en' }`. It should preserve `filterSlugs: ['%5B%5B...filterSlugs%5D%5D']` after one pathname decode.

   Can we distinguish actual platform omission sentinels from literal pathname data? Encoding alone does not distinguish them.

timneutkens:
Addressed both points from [this review](https://github.com/vercel/next.js/pull/98102#pullrequestreview-5224081562) in 895e4d8b98:

- Query normalization can override a valid pathname through an omission only for a platform route parameter. `/shop/books?slug=index` and `/shop/books?slug=` both retain `slug: ['books']`.
- Encoding no longer identifies omission sentinels. The adapter's incoming URL metadata distinguishes captured values from placeholders injected while matching a route template. Captured values retain one-decode semantics, including literal bracket values at multiple encoding levels.

Added a `cacheComponents: false` HTTP regression suite. All six cases pass in development, production with both bundlers, and both the adapter and legacy deployment runtimes. The normalization unit suite passes 18 tests.

The deployment follow-up also fixes omitted catch-alls in locale data requests, records resume metadata for static fallback shells with unresolved router params (22920bd335), and moves the existing Turbopack-only PostCSS guard ahead of deployment setup (32b7b012f1). The failing i18n, base-path, and interception suites passed focused deployment reruns; the PostCSS guard and its Turbopack assertions were verified separately.

packages/next/src/server/app-render/types.ts
```diff
@@ -97,6 +97,8 @@ export interface RenderOptsPartial {
   err?: Error | null
   basePath: string
   cacheComponents: boolean
+  /** Legacy deployment builders require a resume for unresolved router params. */
+  isLegacyPrerender?: boolean
```
unstubbable: We already use the term "legacy prerender" as meaning a non-Cache Components prerender. Should we call this `requiresResumeForFallbackParams` instead?

packages/next/src/server/route-modules/route-module.ts
```diff
@@ -909,20 +916,58 @@ export abstract class RouteModule<
     serverUtils.normalizeCdnUrl(req, combinedParamKeys)
     // When Next is not hosted in a single process, upstream proxies will add query values for route params that were used to match the route.
     // Outside of that environment, there is no reason to do any normalization to honor those query values.
+    const capturedRouteParamKeys = new Set<string>()
     if (!routerServerContext?.isWrappedByNextServer) {
-      serverUtils.normalizeQueryParams(query, routeParamKeys)
+      const initialQuery = parseReqUrl(
+        getRequestMeta(req, 'initURL') || ''
```
unstubbable: This looks like a non-sensical fallback. I think we can assign an `initURL` let variable in lines 732 ff. This block already makes sure it's always defined. And then we can use that variable here.

packages/next/src/server/server-utils.ts
```diff
@@ -110,7 +110,8 @@ export function normalizeDynamicRouteParams(
   query: ParsedUrlQuery,
   defaultRouteRegex: ReturnType<typeof getNamedRouteRegex>,
   defaultRouteMatches: ParsedUrlQuery,
-  ignoreMissingOptional: boolean
+  ignoreMissingOptional: boolean,
+  routeParamKeys?: Set<string>
```
unstubbable: The new `routeParamKeys` argument does not mean all platform route parameters. It means the keys eligible for placeholder removal. The caller already distinguishes `routeParamKeys` from `placeholderRouteParamKeys`, but the callee loses that distinction.

I’d use `placeholderRouteParamKeys` consistently. Passing the broader set would undo the literal-value protection.

PR #98123, opened 2026-08-31
Return client errors for unrecognized Server Actions

gnoff:
## Summary

- return `400 Bad Request` when a Server Action reference ID does not match the expected format
- return `409 Conflict` when a well-formed action ID is unavailable in the current deployment, including deployment skew
- apply the same classification to IDs from the `Next-Action` header and multipart MPA forms, including direct and bound actions in Node and Edge runtimes
- return generic text bodies matching the client error category, without exposing action-ID details
- derive marked forwarded-action responses from the original action ID so an unexpected or mixed-version worker response cannot change the client error semantics

Fetch actions previously had dedicated 404 handling, while equivalent MPA form submissions fell through to the generic error path and returned 500. This removes that transport-dependent behavior and distinguishes malformed or scanning traffic from legitimate stale clients without treating either case as a missing page or an internal server failure.

## Verification

- `pnpm test-dev-turbo test/e2e/app-dir/actions-unrecognized/actions-unrecognized.test.ts`
- `pnpm test-dev-turbo test/e2e/app-dir/actions/app-action.test.ts -t 'forward.*action'`
- `pnpm test-start-turbo test/e2e/app-dir/actions-unrecognized/actions-unrecognized.test.ts`
- `pnpm test-dev-webpack test/e2e/app-dir/actions-unrecognized/actions-unrecognized.test.ts`
- `pnpm test-dev-turbo test/e2e/app-dir/no-server-actions/no-server-actions.test.ts`

### DIFF
--- a/test/e2e/app-dir/actions-unrecognized/actions-unrecognized.test.ts
+++ b/test/e2e/app-dir/actions-unrecognized/actions-unrecognized.test.ts
@@ -6,7 +6,7 @@ import { outdent } from 'outdent'
 describe('unrecognized server actions', () => {
   const unrecognizedActionId = '0'.repeat(42)
 
-  const { next, isNextDeploy, isNextDev } = nextTestSetup({
+  const { next, isNextDeploy } = nextTestSetup({
     files: __dirname,
   })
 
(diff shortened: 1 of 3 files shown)


test/e2e/app-dir/actions-unrecognized/actions-unrecognized.test.ts
```diff
@@ -129,6 +132,53 @@ describe('unrecognized server actions', () => {
   describe.each(['nodejs', 'edge'])(
     'should error and log a warning when submitting a server action with an unrecognized ID - %s',
     (runtime) => {
+      if (!isNextDeploy) {
```
eps1lon: What does this produce when deployed? I usually prefer forking on the assertions instead of having to guess what happens when deployed. Runtime logs aren't available but still worth knowing what status code this produces.

gnoff: :robot: Done — the MPA status/body assertions now run in deploy mode, so deployment validates `400` for malformed IDs and `409` for unavailable IDs. Only runtime-log assertions remain local because deploy logs are unavailable.

eps1lon (approved):

PR #98129, opened 2026-09-01
[PPF] Separate tracking of runtime data accesses for shells and prefetches

lubieowoce:
Expands the static tracking from #96095 to track shells and prefetches separately. We split the `ShouldAttemptStaticPrefetch` hint into `ShouldAttemptStaticShell` and `ShouldAttemptStaticPrefetch`, which the client router considers separately depending on whether we need a shell or a prefetch.

The server-side tracking of `needsRuntimeRequest` becomes a bit more subtle.

First, a runtime data access should only affect the shell if we're in the `ShellStatic` render stage. If we're already in the `PrefetchStatic` stage, then that content is not conceptually part of the shell, and should not deopt the shell to runtime, so we only unset `shouldAttemptStaticShell` if we're still in the shell stage.

Second, accessing data that would only resolve in the prefetch (params, search params, short-stale caches) should **not** affect whether the shell can be static. `makeRuntimeDataHangingPromise` has been split into variants that match this:
 - `makeSessionDataHangingPromise` - data that affects the shell (and the prefetch), e.g. `cookies()`
 - `makeURLDataHangingPromise` - data that only affects the prefetch, e.g. `params`
  - `makeUnknownRuntimeDataHangingPromise`, used for private caches -- we omit these during build so we don't know whether they should affect the shell or only the prefetch, so we conservatively affect both (unless a private cache is only used in the prefetch stage)
  
The server-side `needsRuntimeRequest` promise is now resolved such that data that only resolves in the prefetch does not affect the shell. If we're accessing data that only resolves in the prefetch, we will delay the promise's resolution to the `PrefetchStatic` stage, which means that when the client rewinds the static response to a shell, it will see an unresolved promise and consider the response as not having used runtime data. (see `markRuntimeDataAccessWhenStageReached`)


packages/next/src/server/app-render/dynamic-rendering.ts
```diff
@@ -348,15 +345,6 @@ export function abortAndThrowOnSynchronousRequestDataAccess(
...
 ): never {
-  // The synchronously accessed request data would have been available during
-  // a runtime prerender, which would have rendered past this point instead of
-  // aborting — so a runtime prefetch would produce more content than this
-  // render. Record that, same as when request data access creates a hanging
-  // promise (see makeRuntimeHangingPromise). Unlike
-  // `abortOnSynchronousPlatformIOAccess`, which aborts a runtime prerender
-  // all the same and therefore must not record anything.
-  trackRuntimeDataAccessed(prerenderStore, expression)
-
```
lubieowoce, line 359: For pages, this is only used for `draftMode.enable()/disable()`, which would not in fact be available in a runtime prerender, it's just an error in any kind of prerender.

packages/next/src/client/components/segment-cache/scheduler.ts
```diff
@@ -1184,12 +1183,49 @@ function walkRequiresRuntimeCompleteness(
...
+      return false
+  }
+}
+
+function hasHint(
+  allHints: number,
+  /** Accepts either a single hint or a mask created by unioning hints together. */
+  mask: PrefetchHint | (number & {})
+): boolean {
+  return (allHints & mask) !== 0
```
acdlite: AI slop?

lubieowoce: i added this because i don't find `(hints & MyHint) !== 0` very readable as "hint is set", it should get inlined anyway

lubieowoce: ...also i had multiple mistakes where i wrote `... === 0` instead of `.. !== 0` or vice versa (or forgot to negate correctly when negating some other condition), so i kinda just wanted to make my life easier 😅

acdlite (approved):

packages/next/src/client/components/segment-cache/scheduler.ts
acdlite: I find this a bit silly but ok

acdlite: I'll just inline it back the next time I edit the code myself

lubieowoce: ok, removed it bc i'm not that attached to it

PR #98139, opened 2026-09-01
Remove Edge runtime handling in use-cache-wrapper

mischnic:
Cache components are only compatible with Node.js runtime anyway. So this was basically dead code. Removing it doesn't break any existing functionality.

### DIFF
--- a/packages/next/src/server/use-cache/use-cache-wrapper.ts
+++ b/packages/next/src/server/use-cache/use-cache-wrapper.ts
@@ -348,12 +348,6 @@ const crossRequestPendingCacheInvocations = new Map<
   Promise<SharedCacheResult>
 >()
 
-if (process.env.NEXT_RUNTIME === 'edge') {
-  throw new Error(
-    "internal error: 'use cache' is not supported in Edge Runtime"
-  )
-}
-
 // The first argument at each call site is the full directive that produced
 // the invocation, e.g. "'use cache'" or "'use cache: remote'".
 const debug = process.env.NEXT_PRIVATE_DEBUG_CACHE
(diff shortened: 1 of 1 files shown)


packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -348,7 +348,11 @@ const crossRequestPendingCacheInvocations = new Map<
   Promise<SharedCacheResult>
 >()
 
-const isEdgeRuntime = process.env.NEXT_RUNTIME === 'edge'
+if (process.env.NEXT_RUNTIME === 'edge') {
+  throw new Error(
+    "internal error: 'use cache' is not supported in Edge Runtime"
+  )
+}
```
mischnic: Do we want this? Or are we confident that this is never happening anyway?

mischnic:
* **#98140** <a href="https://app.graphite.com/github/pr/vercel/next.js/98140?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#98139** <a href="https://app.graphite.com/github/pr/vercel/next.js/98139?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/98139?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

packages/next/src/server/use-cache/use-cache-wrapper.ts
unstubbable: I don't think we need this. The `runtime` route segment config is prohibited by an SWC transform whenever `cacheComponents` or `useCache` is enabled.

unstubbable (approved):

PR #98140, opened 2026-09-01
Support durable use-cache entries with client components

mischnic:
The current setup is:
1. A use-cache function returns an RSC value that may contain client reference proxies. Flight serializes each reference using the client module `id`, export `name`, client `chunk` list, and `async`. The resulting Flight payload is stored in the cache entry.
2. When decoding the cache entry, the client module `id` is resolved through `rscModuleMapping` to obtain the corresponding RSC module `id` and `export`.

Cache decoding does not need the client chunk list because it does not load client chunks. The outer RSC serialization emits the client reference again using the current client reference manifest, including the current chunks. We can therefore serialize cached references with an empty chunk list.

The client module `id` is also unstable across builds. Instead, cache entries can store the stable client reference name. When decoding, we resolve: client reference name -> current client module ID -> current RSC module ID
This makes cached client references independent of build-specific module IDs and chunk filenames.


mischnic:
* **#98140** <a href="https://app.graphite.com/github/pr/vercel/next.js/98140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/98140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#98139** <a href="https://app.graphite.com/github/pr/vercel/next.js/98139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `canary`


<h2></h2>

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -348,6 +348,85 @@ const crossRequestPendingCacheInvocations = new Map<
...
+        return undefined
+      }
+
+      return {
+        // Cache entries must not depend on build-local module IDs or chunks.
+        // The stable client reference name is resolved against the current
+        // manifest when the entry is read.
+        id: clientReferenceName,
+        name: clientReferenceManifestEntry.name,
+        chunks: ['stub'],
```
unstubbable, line 390: why not an empty array?

packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -348,6 +348,85 @@ const crossRequestPendingCacheInvocations = new Map<
...
+ *  name: `*`,
+ *  chunks: ["stub"],
+ *  async: false
+ * }
+ * ```
+ *
+ * This means that not the unstable client module id is used for the cache entry, but the stable
+ * client reference name.
+ */
+function createClientModulesForCache(
```
unstubbable, line 366: We should be able to avoid constructing this proxy per invocation. I think we could change the behavior of the existing proxy in `createMappingProxy` in `packages/next/src/server/app-render/manifests-singleton.ts` when the prop is `clientModules` or `rscModuleMapping` and the work unit store is a cache store. Although we'd need to wrap the `createFromReadableStream` call in a cache store, which technically should not be a problem, but we don't have one handy there, and it also feels a bit hacky. So maybe only reuse the proxy for `clientModules`, but at least move the cache-specific `rscModuleMapping` proxy to the manifest file to avoid the per-invocation construction?

packages/next/src/server/use-cache/use-cache-wrapper.ts
mischnic: I thought just in case this ends up being read at some point, then at least we't get `Failed to load chunk stub (404)` instead of `Failed to load module 1234`
But I can make it empty if you disagree

unstubbable: Aha, that makes sense to me.

packages/next/src/server/use-cache/use-cache-wrapper.ts
mischnic: I tried to do that. I hope this is what you had in mind.

unstubbable (approved):

PR #98149, opened 2026-09-01
test: support deploy exclusions with force gates

gnoff:
## Summary

- add static gate conditions for the deployment CI variants used by existing test suites
- support `@force-gate` on parameterized `describe.each` suites
- document and test the force-gate infrastructure needed by the migration

This establishes the gate support used by the deployment-exclusion migration PRs above it.

## Verification

- `pnpm typescript`
- `pnpm jest test/unit/gate/pragma-transform.test.ts test/unit/gate/runtime.test.ts --runInBand`

### DIFF
--- a/test/lib/gate/conditions.ts
+++ b/test/lib/gate/conditions.ts
@@ -134,6 +134,21 @@ export const conditions: Record<string, Condition> = {
   ci: staticCondition('running in CI (`NEXT_TEST_CI`)', () =>
     Boolean(process.env.NEXT_TEST_CI)
   ),
+  adapter: staticCondition('running the adapter test variant', () =>
+    Boolean(process.env.NEXT_ENABLE_ADAPTER === '1')
+  ),
+  standaloneOutput: staticCondition(
+    'running the standalone-output test variant (`TEST_OUTPUT_STANDALONE`)',
+    () => process.env.TEST_OUTPUT_STANDALONE === 'true'
+  ),
+  turbopackDev: staticCondition(
+    'the test run uses the Turbopack development build (`TURBOPACK_DEV`)',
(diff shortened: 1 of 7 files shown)


eps1lon (dismissed):

test/lib/gate/conditions.ts
```diff
@@ -134,6 +134,25 @@ export const conditions: Record<string, Condition> = {
   ci: staticCondition('running in CI (`NEXT_TEST_CI`)', () =>
     Boolean(process.env.NEXT_TEST_CI)
   ),
+  adapter: staticCondition('running the adapter test variant', () =>
+    Boolean(process.env.NEXT_ENABLE_ADAPTER === '1')
+  ),
+  cacheComponentsTest: staticCondition(
```
eps1lon: The confusing part is that we already have `cacheComponents` which is a lazy flag based on config evaluation.

gnoff: :robot: `cacheComponents` is the lazy resolved fixture config. This suite needs the runner-level Cache Components shard because the exclusion must be decided before fixture setup: that shard patches files before startup, which deploy cannot do, and the old `skipDeployment` condition used the same `__NEXT_CACHE_COMPONENTS` value. I renamed the static condition to `cacheComponentsShard` and documented the distinction.

eps1lon: We skip hooks as of https://github.com/vercel/next.js/pull/98169 so this is probably based on outdated information. Which suite even?

eps1lon (changes requested):
See https://github.com/vercel/next.js/pull/98149#discussion_r3928709246

eps1lon (approved):

PR #92803, opened 2026-04-14
fix(csp): Provide nonce in `create-component-styles-and-scripts`

maisano-patreon:
### What?

Add `nonce={ctx.nonce}` to `<script>` and `<link>` tags in `createComponentStylesAndScripts`.

### Why?

`createComponentStylesAndScripts` creates `<script>` and `<link>` tags for boundary components (`loading`, `error`, `not-found`, etc.) without providing the CSP nonce. The sibling function `getLayerAssets`, which seemingly does the same thing for layout/page chunks, already includes it.

In my app, this causes the loading boundary's script tag to be emitted without a nonce attribute, which breaks nonce-based CSP. Of dozens of chunks loaded, the only one without a nonce is the loading chunk (e.g. `loading_tsx_0fhk6ua._.js`).

### How?

Pass `nonce={ctx.nonce}` in the props, matching what `getLayerAssets` already does.

I was unable to create a minimal reproduction in the test suite, but I've confirmed that this patch resolves the CSP violation in my application.

My application is quite large – it uses Next 16.2.0 w/ turbopack. I spent some time trying to create a minimal repro w/ both turbopack and webpack but was unable to. Patching the compiled code in my repo, I was able to verify this fix.

I will happily write tests in order to fix this if someone more familiar might know how to hit this failure mode.

PR #98152, opened 2026-09-01
Add CSP nonce to script tags of loading and template files

controversial:
> `createComponentStylesAndScripts` renders the script tags for a segment's loading, error, and templatefound files, but unlike getLayerAssets it did not pass `ctx.nonce`. When such a file has a chunk that no layout or client reference has already loaded, the tag is served without a nonce and a nonce-based Content-Security-Policy blocks it.

### What?

Patches a bug we encountered while using [nonces](https://nextjs.org/docs/app/guides/content-security-policy#nonces) for a CSP with `strict-dynamic`, as in the docs example. In Turbopack, when a `loading.tsx` (or `error` or `template`) file has a chunk that isn’t already loaded by a layout or client component on the page, Next renders a `<script>` tag for that chunk with no `nonce`. (the same chunk script would’ve gotten a `nonce` if it were emitted for a layout or page).

Supersedes #92803; this is the same fix but with a new test included. 

### Why?

While building out our `loading.tsx` pages, we encountered browsers blocking a script and reporting CSP violations, because chunk scripts did not properly carry a `nonce`.

### How?

One-line change to bring `createComponentStylesAndScripts` to parity with `getLayerAssets`.

### DIFF
--- a/test/e2e/app-dir/csp-nonce-segment-scripts/csp-nonce-segment-scripts.test.ts
+++ b/test/e2e/app-dir/csp-nonce-segment-scripts/csp-nonce-segment-scripts.test.ts
@@ -1,4 +1,5 @@
 import { nextTestSetup } from 'e2e-utils'
+import { retry } from 'next-test-utils'
 
 describe('csp-nonce-segment-scripts', () => {
   const { next } = nextTestSetup({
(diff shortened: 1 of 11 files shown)


test/e2e/app-dir/csp-nonce-segment-scripts/csp-nonce-segment-scripts.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+describe('csp-nonce-segment-scripts', () => {
+  const { next } = nextTestSetup({
+    files: __dirname,
+  })
+
+  // The loading and template files import a client component that the page
+  // also imports, so their chunk is not already loaded by a layout or by a
+  // client reference, and it gets a script tag of its own.
+  it('should add the nonce to the script tags of loading and template files', async () => {
```
lukesandberg, line 12: this test should also assert that the button works (the client code executes)

controversial: Thanks @lukesandberg; I pushed another test!

lukesandberg (approved):

lukesandberg (approved):

PR #98154, opened 2026-09-01
test: migrate caching deploy exclusions to force gates

gnoff:
## Summary

- migrate Cache Components, PPR, prefetching, prerendering, revalidation, and SSG deployment exclusions to `@force-gate`
- remove the corresponding `skipDeployment` options and `skipped` control flow while preserving each suite's rationale

This keeps caching-related exclusions together so their deployed semantics can be reviewed by the same owners.

## Verification

- verified on the combined top-of-stack tree with `pnpm typescript`
- compared ordinary-mode collection before and after: all 4,670 existing test names matched
- collected all 510 affected files in deploy mode: 4,578 skipped tests, zero failures

### DIFF
--- a/test/e2e/cache-handlers-upstream-wiring/index.test.ts
+++ b/test/e2e/cache-handlers-upstream-wiring/index.test.ts
@@ -3,15 +3,13 @@ import { retry } from 'next-test-utils'
 import { join } from 'path'
 
 describe('cache-handlers-upstream-wiring', () => {
+  // TODO(deploy-test-completion): Re-enable this suite in deploy mode.
+  // It likely asserts local CLI or runtime output that deploy tests do not expose.
+  // @force-gate !deploy
   describe('pages router non-edge', () => {
-    const { next, skipped, isNextDev } = nextTestSetup({
+    const { next, isNextDev } = nextTestSetup({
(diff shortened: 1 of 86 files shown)


test/e2e/cache-handlers-upstream-wiring/index.test.ts
```diff
@@ -98,40 +94,36 @@ describe('cache-handlers-upstream-wiring', () => {
...
+    })
 
-      let outputIndex = 0
+    let outputIndex = 0
 
-      beforeEach(() => {
-        outputIndex = next.cliOutput.length
-      })
+    beforeEach(() => {
+      outputIndex = next.cliOutput.length
```
eps1lon, line 109: Test failures are originating from here because `.skip` still runs test hooks. The problem is that lazy gates are  not skipping the describe block because they need the test setup. So we need to transform the Jest hooks to be noops if the gate evaluates to false.

eps1lon: Fixed in https://github.com/vercel/next.js/pull/98169 (which moves this test suite to `@force-gate !cacheComponents` to confirm).

eps1lon (approved):
needs a rebase

PR #98161, opened 2026-09-01
test: remove nextTestSetup deploy exclusion API

gnoff:
## Summary

- remove `skipDeployment` and `skipped` from `nextTestSetup` after every caller has migrated to `@force-gate`
- update the test review guidance to identify new deployment exclusions by their gate directive

This final layer makes `@force-gate` the only suite-scoped deployment exclusion mechanism.

## Verification

- verified on the combined top-of-stack tree with `pnpm typescript`
- verified the feature-oriented stack has the same Git tree as the previously tested migration


eps1lon (approved):
🎉

PR #98164, opened 2026-09-02
docs: local development: Rewrite docker section, add Windows Dev Drive and WSL2 instructions

bgw:
Follow-up to https://vercel.slack.com/archives/C046HAU4H7F/p1787344658655399

Summary:
- Move system configuration changes into it's own section, separate from application and project configuration changes.
- The Docker section was very light on details and didn't do a good job of explaining why this problem happens.
- Also suggest https://docs.docker.com/desktop/features/synchronized-file-sharing/ for Docker and explicitly discourage `watchOptions.pollIntervalMs` as a workaround.
- Suggest Windows Dev Drive (https://learn.microsoft.com/en-us/windows/dev-drive/). I don't have experience with it, but Microsoft recommends it and claims it improves filesystem performance.
- Recommend trying WSL 2.

### DIFF
--- a/docs/01-app/02-guides/local-development.mdx
+++ b/docs/01-app/02-guides/local-development.mdx
@@ -12,7 +12,7 @@ The development process with `next dev` is different than `next build` and `next
 
 `next dev` compiles routes in your application as you open or navigate to them. This enables you to start the dev server without waiting for every route in your application to compile, which is both faster and uses less memory. Running a production build applies other optimizations, like minifying files and creating content hashes, which are not needed for local development.
 
-## Improving local dev performance
+## System configuration
 
 ### 1. Check your computer's antivirus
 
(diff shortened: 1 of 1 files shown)


docs/01-app/02-guides/local-development.mdx
```diff
@@ -48,7 +48,35 @@ On macOS, you can disable [Gatekeeper](https://support.apple.com/guide/security/
...
+
+Windows 11's Dev Drive feature uses an alternative filesystem to improve performance for common development workflows. [Microsoft has a guide explaining how to configure it](https://learn.microsoft.com/en-us/windows/dev-drive/).
+
+Alternatively, Windows provides an easy way to set up a Linux virtual machine using [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/). You may see better performance by using WSL 2.
+
+When using WSL 2, store your project directory inside the virtual machine's filesystem (e.g. `/home/$USER/project`) and not on a Windows drive (e.g. `C:\` or `/mnt/c`). If you do not do this, you will experience worse performance and file updates may not be reflected properly in development.
+
+## Improving application development performance
+
+### 1. Update Next.js and use Turbopack
```
lukesandberg, line 81: are the numbers important ? can we get auto numbering working/

docs/01-app/02-guides/local-development.mdx
lukesandberg: dev drive is a better solution here, so mention... link below?

lukesandberg (approved):

docs/01-app/02-guides/local-development.mdx
bgw: It's just sorta an order of priority? IDK, I was mostly preserving it from the other section.

This isn't an actual `<ol>` element, so markdown's auto-numbering doesn't apply here.

lukesandberg: maybe there is an mdx'ism for this? no big deal ofc

bgw: Yeah, I don't think there is.

docs/01-app/02-guides/local-development.mdx
```diff
@@ -48,7 +50,35 @@ On macOS, you can disable [Gatekeeper](https://support.apple.com/guide/security/
...
+
+For the best development experience:
+
+- Run Next.js without Docker (`npm run dev` or `pnpm dev`) during development.
+- Reserve Docker for production deployments and testing production builds.
+- If you must use Docker for development, consider running Docker natively on a Linux machine or a manually configured virtual machine. If using a manually configured virtual machine, move the source code into the VM and do not share it from the host.
+- If you're using Docker Desktop, evaluate using [synchronized file shares](https://docs.docker.com/desktop/features/synchronized-file-sharing/). Synchronized file shares still add latency compared to running Next.js directly on the host, but can improve file-watching reliability without resorting to polling.
+- Avoid using `watchOptions.pollIntervalMs` in `next.config.js` to work around Fast Refresh issues. Only use it as a last resort. Polling adds significant latency and can drastically increase CPU and I/O usage.
```
icyJoseph, line 67: Should this PR be merged, https://github.com/vercel/next.js/pull/80687/changes and x-references here? I understand we kind of want to discourage usage though — not all of the config options need a reference, specially if its there to serve what feels like an edge case in dev environments

icyJoseph (approved):

docs/01-app/02-guides/local-development.mdx
bgw: I think we can have the documentation if we are very clear on that documentation page why people should never use it. We've fixed most of the compatibility issues with Turbopack for that feature.

The only outstanding issue I know of is caused by https://github.com/notify-rs/notify/pull/981, but we're waiting on a new release of notify-rs.

PR #98182, opened 2026-09-02
Pages Router: key the client router filter prefetch marker by the `as` path

### LINKED ISSUE #98180: Pages Router: `prefetch()` checks the client router filter against `as` but stores the `__appRouter` marker under the `href` pathname

https://github.com/Stanzilla/next-pages-router-prefetch-bloom-filter-repro

### To Reproduce

The repository README has the full walkthrough. Short version:

1. `pnpm install && pnpm build && pnpm start` (prefetching is a no-op in `next dev`, so a production build is required).
2. Open `http://localhost:3000/players/alice` (a `pages/players/[name].tsx` route with `getServerSideProps`).
3. In the console run `window.beforeNav = 1`.
4. Hover the link **Room 00000000-0000-4000-8000-000000000c4c**. Do not click it. The link is a "route as modal" link (see `examples/with-route-as-modal`): `href={{ pathname: router.pathname, query: { ...router.query, modal: id } }}` and `as="/room/<id>/scoreboard"`, `shallow`.
5. Click the tab **stats**, a plain `<Link href="/players/alice?tab=stats" shallow>`.

The `as` path in step 4 is a false positive of the static client router Bloom filter that `next build` generates from the four App Router paths in the project (`/api/health`, `/api/version`, `/api/status`, `/_not-found`). `scripts/find-collision.mjs` reads the filter out of the built router chunk and re-derives the colliding id, so the collision is deterministic for this repository.

**Current**

- After step 4, `window.next.router.components['/players/[name]']` is `{ __appRouter: true }`. The cached route info of the page the user is on has been replaced by the marker.
- After step 5 the page shows "Application error: a client-side exception has occurred" and the console logs `TypeError: Cannot destructure property 'title' of 'n' as it is undefined.` (the custom `_app` reads `pageProps.title`). `window.beforeNav` is still `1`, so this is a client crash, not a navigation.
- On the static route `/` (same links, `pages/index.tsx`) the symptom is different: hovering the link then clicking tab **b** performs a full page reload to `/?tab=b` (`window.beforeNav` becomes `undefined`).
- `prefetch={false}` on the link does not help. It only disables the viewport prefetch, `onMouseEnter` and `onTouchStart` still call `router.prefetch()`.

**Expected**

The shallow tab navigation stays a client-side navigation and the page keeps rendering. A Bloom filter match on the `as` path `/room/.../scoreboard` must not change how the pages route `/players/[name]` behaves.

### Root cause

`Router.prefetch()` in `packages/next/src/shared/lib/router/router.ts` evaluates the filter against the `as` path but stores the marker under the `href` pathname:

```ts
const urlPathname = parsed.pathname            // pathname of `href`
...
if (await this._bfl(asPath, resolvedAs, options.locale, true)) {
  this.components[urlPathname] = { __appRouter: true } as any
}
```

`Router.change()` then looks the marker up with the `href` pathname as well (`this.components[pathname]?.__appRouter`, after `removeBasePath`/`removeTrailingSlash`). For a static route both keys are `/`, so the guard fires and hard-navigates a shallow query change. For a dynamic route the marker is stored under the pattern (`/players/[name]`, because `href.pathname` is `router.pathname`) but looked up with the concrete path (`/players/alice`), so the guard misses and the shallow `getRouteInfo()` early return hands the marker to `_app` as route info (tracked separately in https://github.com/vercel/next.js/issues/98181).

The filter is on by default when an `app/` directory exists and has a 0.01 % false positive rate per checked path. Every hover or touch on a link re-checks a path, so on a large pages-router site with many `href !== as` links this shows up as sporadic, hard to reproduce crashes tied to specific ids.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 36864
  Available CPU cores: 12
Binaries:
  Node: 24.19.0
  npm: 11.17.0
  Yarn: N/A
  pnpm: 11.24.0
Relevant Packages:
  next: 16.4.0-canary.13
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Also reproduced with `next@16.3.3` and `next@16.3.4`. The relevant code is unchanged on `canary` (538f3f6).

### Additional context

A fix that keys the marker by the normalised `as` pathname in both `prefetch()` and `change()` is proposed in a linked pull request.

Disclosure: the investigation, the reproduction repository and this report were prepared with AI assistance (Claude Code) on behalf of the reporter.

Stanzilla:
### What?

`Router.prefetch()` in the Pages Router evaluates the client router filter (`_bfl`) against the `as` path, but stored the resulting `{ __appRouter: true }` marker in `router.components` under the `href` pathname. `Router.change()` looked the marker up with the `href` pathname as well. This PR keys the marker by the normalized `as` pathname and effective locale, preserves existing cache entries, and checks for markers before and after route resolution.

### Why?

`href` and `as` are the same URL for most links, so the mismatch was invisible. They differ for the "route as modal" pattern (`examples/with-route-as-modal`): `href` stays on the current route (`router.pathname` + query) and `as` shows the pretty URL. When the filter matches `as`, including a false positive, the marker could replace the cached route info of the page the user is currently on:

- On a static route (`/`), the `change()` guard fired for later shallow navigation and hard-navigated.
- On a dynamic route, the marker was stored under the pattern (`/players/[name]`) but looked up with the concrete path (`/players/alice`). The guard could miss it, and shallow navigation could render without page props, causing missing content or an application error.

A prefetch must also preserve already-loaded route info when its key matches the current route. This can happen when the page was reached through a rewrite and its canonical URL is prefetched. Hash-only navigation renders that cached entry directly, so preserving it lets the normal client-side hash update retain the page props and update `router.asPath`.

Fixes #98180

### How?

- `getAppRouterMarkerKey(router, as, locale)` parses the pathname, normalizes the trailing slash, and preserves the effective locale. It expects a path without `basePath` and does not strip that prefix again. An explicit locale prefix takes precedence over the supplied locale.
- The helper returns `null` for non-local URLs, leaving their existing navigation handling unchanged.
- `prefetch()` stores the marker under that key only when the corresponding cache entry is absent. It does not replace loaded route info.
- `change()` checks both the normalized `as` key and the `href` route. It checks the resolved route again after config rewrites, before `getRouteInfo()` reads the cache.
- Non-shallow navigations continue to consult the client router filter directly. Hash-only navigation retains its normal client-side path.

The complementary change in #98650 (adopts #98187) adds defensive checks when `getRouteInfo()` reads marker entries.

### Tests

The e2e suites run in production (`next start`) and deploy modes. `router.prefetch()` is a no-op in development.

`test/e2e/app-dir/pages-prefetch-as-app-route` covers:

- Hard navigation to an App Router destination after prefetching a link whose `href` and `as` differ.
- Shallow navigation on static and dynamic Pages Router routes without a reload or lost server-provided props.
- Hard navigation when the `href` route holds a marker and `as` differs, or a config rewrite resolves to a marked route.
- Client-side hash navigation after an awaited prefetch of the current route's canonical URL, with preserved props and an updated `router.asPath`.

`test/e2e/app-dir/pages-prefetch-as-app-route-base-path` covers an internal `/docs` route with `basePath: '/docs'`, including navigation to `/docs/docs` and an unaffected shallow update on the index page.

`test/e2e/app-dir/pages-prefetch-as-app-route-i18n` covers a French-only redirect prefetch without forcing an unrelated English navigation to reload. Marker-dependent tests wait for the specific key they need, rather than any marker.

The fix was also verified at runtime against the reproduction in https://github.com/Stanzilla/next-pages-router-prefetch-bloom-filter-repro by patching the compiled `router.js` and re-running the scenario. The e2e suite was not run locally.

Disclosure: this change and its description were prepared with AI assistance (Claude Code) on behalf of the author.


unstubbable (changes requested):
I think this needs changes before merging:

- The `as` lookup can miss a marker that `getRouteInfo()` still reads as component data. With `pages/modal.js` and `app/modal/[id]/page.js`, prefetching `/modal` and then navigating with `href="/modal"` and `as="/pretty"` now throws because the cached marker has no `Component`. Previously this hard-navigated.
- The helper strips `basePath` from paths that already exclude it. With `basePath: '/docs'`, prefetching the internal route `/docs` stores the marker under `/`, replaces the index page’s cache entry, and breaks subsequent shallow navigation.
- The marker loses the effective locale. A prefetch that matches a French-only redirect at `/fr/legacy` now marks `/legacy` and forces an unrelated English navigation to reload.
- The dynamic regression test does not catch the lost page props. Its assertions still pass without `slug`; only the next tab link becomes `/blog/undefined?tab=b`. It should assert that the server-provided `slug` survives.

Can we address these remaining cache collisions and path/locale issues, and add regression coverage for these cases?

Stanzilla:
Thanks for the review. Commit https://github.com/vercel/next.js/pull/98182/commits/8b30738dccbe1481131933db0c206fc6b88a2553 addresses each point:

1. **Marker under the href route.** `change()` now checks the marker under the `as` key and under the href route key. It hard navigates when either one is set. With `pages/modal.js` next to `app/modal/[id]/page.js`, a prefetch of `/modal` followed by `router.push('/modal', '/pretty')` hard navigates again instead of reading the marker as component data. New test: "should hard navigate when the href route was flagged by the prefetch and `as` differs".

2. **basePath.** The helper no longer strips the `basePath`. Both call sites already hold the path without it: `next/link` passes it that way to `prefetch()`, and `change()` computes `cleanedAs` after `removeBasePath`. New fixture `pages-prefetch-as-app-route-base-path` with `basePath: '/docs'` and `app/docs/page.js`: after the prefetch, shallow navigation on the index page stays client side.

3. **Locale.** The key now carries the effective locale, in the same way `_bfl` evaluates `addLocale(as, locale)`: `options.locale || this.locale` in `prefetch()` and `nextState.locale` in `change()`. A locale prefix that is already part of the path wins, detected with `normalizeLocalePath`. New fixture `pages-prefetch-as-app-route-i18n`: a French prefetch that matches the `/fr/legacy` redirect marks `/fr/legacy`, and the English navigation to `/legacy` stays client side.

4. **Lost page props.** `pages/blog/[slug].js` renders the `slug` prop. The dynamic test asserts that it is still `first` after the shallow navigation and that the tab link stays `/blog/first?tab=b`.

The draft #98187 additionally makes `getRouteInfo()` ignore markers, which covers the case where `route` changes after the guard (middleware rewrites).

I cannot run the e2e suites locally, so CI on this fork is the verification once the workflows are approved.

---
Written by an agent (Claude Code, claude-fable-5-1).

unstubbable (commented):
The original examples look addressed. Could you also add coverage for config rewrites and hash-only navigation after prefetching?

The marker guard runs before config rewrites resolve the destination, while hash-only navigation bypasses both that guard and `getRouteInfo()`. I’d like to make sure markers cannot reach rendering through either path. The hash-only case would not be covered by #98187’s guards.

Also, `hoverLink()` currently waits for any marker, but the index now has two marker-producing links. Can we make the completion check specific to the prefetch the test depends on?

Stanzilla:
Both paths are now covered in a54bb450.

**Config rewrites.** `change()` runs the marker guard again on the route that `resolveRewrites()` returns, directly before `getRouteInfo()`. The new fixture button pushes `/pretty` after a hover prefetch of `/modal` marked `/modal`. Without the second guard the navigation passes the first guard under `/pretty`, then `getRouteInfo()` reads the marker under `/modal` and the page crashes with `Cannot read properties of undefined (reading 'getInitialProps')`. With the guard the router hard navigates to `/pretty`.

**Hash-only navigation.** The hash shortcut is skipped when `this.components[route]` holds a marker, so the request falls through to the regular guard and hard navigates. For the test I needed a marker under the current route through public APIs. `_bfl()` skips a path that equals the current `asPath`, so a prefetch of the current URL never writes one. The fixture loads `pages/modal.js` at `/pretty` through the config rewrite: the route is `/modal`, the `asPath` is `/pretty`. A hover prefetch of the canonical `/modal` link then passes the current-path check and stores the marker under the route of the current page. A hash link click after that rendered the marker and dropped the props from `getServerSideProps`. The test asserts that the prop text stays rendered after the hash change.

**Completion check.** `hoverLink()` in this suite, and the equivalent waits in the basePath and i18n suites, now take the expected marker key (`/dashboard`, `/modal`, `/docs`, `/fr/legacy`) and wait for `window.next.router.components[key].__appRouter`.

Verification: I replayed both scenarios in a standalone app on `next@16.4.0-canary.28` in a browser, first with the unmodified router and then with the two guards patched into `dist/shared/lib/router/router.js`. Unpatched: the rewrite push shows the client-side exception page, the hash click keeps the URL client side but renders an empty prop. Patched: the rewrite push hard navigates to `/pretty`, the hash click keeps the rendered prop and moves to `#section`. I did not run the e2e suite locally.

---
Written by an agent (Claude Code, claude-fable-5-1).

packages/next/src/shared/lib/router/router.ts
```diff
@@ -1379,7 +1428,16 @@ export default class Router implements BaseRouter {
...
-    if (!isQueryUpdating && this.onlyAHashChange(cleanedAs) && !localeChange) {
+    // A hash-only change re-renders the cached route info of the current
+    // route. When a prefetch replaced that entry with the app router marker,
+    // take the full navigation path below so the marker guard hard navigates
+    // instead of rendering the marker.
+    if (
+      !isQueryUpdating &&
+      this.onlyAHashChange(cleanedAs) &&
+      !localeChange &&
+      !(this.components[nextState.route] as any)?.__appRouter
```
unstubbable: Does falling through actually cause a reload for a hash-only change? `handleHardNavigation()` assigns `window.location.href`, which normally performs a same-document navigation when only the fragment differs.

The prop assertion could then pass simply because React does not render again, while we have skipped the code that updates `router.asPath`.

Can the test verify the intended behavior here? If this should reload, assert a document navigation. If it should stay client-side, assert that `router.asPath` updates alongside `location.hash`.

Stanzilla: You are right, it did not reload. `location.href` with a hash-only difference is a same-document navigation. The page stayed, `router.asPath` stayed `/pretty`, and the prop assertion passed only because React did not render again. I confirmed this in a browser before the change.

29fe1167 replaces the fall-through with a fix in `prefetch()`: the marker is only stored when `this.components[key]` holds no entry. A loaded pages route in the cache is ground truth for that key, so the prefetch must not replace it. This covers the current page loaded through a rewrite whose canonical URL is prefetched, and a prefetch that finishes after a navigation stored the same route. The hash shortcut is back to its original form and always finds the real route info of the current route. Non-shallow navigations consult the client router filter themselves and still hard navigate. The guard after `resolveRewrites()` stays.

The test now asserts the client-side behavior. The page loads at `/pretty`, a button awaits `router.prefetch('/modal')` and shows `done`, because no marker is observable anymore. After the hash link click the test expects `window.beforeNav` to survive, `router.asPath` to become `/pretty#section` next to `location.hash`, and the prop from `getServerSideProps` to stay rendered.

Verification in a standalone app on `next@16.4.0-canary.28` with the compiled router patched. With the unconditional marker write the hash click keeps the URL client side but renders an empty prop. With the conditional write the prop stays, `router.asPath` is `/pretty#section`, `beforeNav` is still 1, and the page scrolls to the section. The rewrite push still hard navigates. I did not run the e2e suite locally.

CI on a54bb450: the failed `test turbopack production (7/7)` job is `test/production/deterministic-build/deployment-id.test.ts`. Its two builds exit with `middleware.ts(3,44): error TS1470` (`import.meta` in CommonJS output) on all three attempts, which is not related to this change.

---
Written by an agent (Claude Code, claude-fable-5-1).

unstubbable (approved):

PR #98196, opened 2026-09-02
turbo-tasks: use Hash for persistent task keys

lukesandberg:
## Summary

Use each task input's existing `Hash` implementation for persistent task-cache keys instead of specializing its bincode encoder. `Xxh3Hash64Hasher` implements `std::hash::Hasher` with deterministic little-endian typed writes.

`DynTaskInputs::hash_value(&mut dyn Hasher)` is the single erased argument-hash functor shared by both cache layers:

- in-memory hashing prefixes the native-function pointer and receiver;
- persistence hashing prefixes the stable registry function ID and receiver.

The registered function already determines the argument tuple type, so the shared functor omits general `DynHash`'s redundant `TypeId`. This avoids both the short-input overhead of `DynHash` and the code-size cost of monomorphizing `Hash::<Xxh3Hash64Hasher>` per argument tuple.

Persistent inputs already exclude unordered collections and process-dependent values because those break deterministic serialization. The `TaskInput` docs now state the same requirement for `Hash`. Full task-type equality still verifies every persistent hash-bucket candidate.

## Measurements

Identical `cargo build -p next-napi-bindings --release` builds:

| Artifact | Bincode baseline | Shared erased `Hash` | Delta |
|---|---:|---:|---:|
| Raw `.so` | 148,125,576 B | 147,268,816 B | **-856,760 B (-0.58%)** |
| Stripped | 102,361,720 B | 101,896,504 B | **-465,216 B (-0.45%)** |
| Stripped + gzip | 34,433,707 B | 34,324,960 B | **-108,747 B (-0.32%)** |

Committed Criterion benchmark estimates:

| Input | Bincode | Shared erased `Hash` | Delta |
|---|---:|---:|---:|
| String tuple/vector | 29.76 ns | 32.24 ns | +8% |
| Mixed strings/`Option<u32>` | 33.55 ns | 38.80 ns | +16% |
| 128-byte `RcStr` | 33.52 ns | 22.67 ns | -32% |
| String + 256 `u32`s | 382.14 ns | 106.75 ns | -72% |

`samply` profiles (about 10,000 samples/path) identified `DynHash`'s `TypeId` and typed writes as the earlier 28-45% short-string regression. The shared functor removes those frames. `perf` hardware counters are unavailable in this VM (`<not supported>`). These are isolated hash-kernel results, not end-to-end build percentages; persistent hashing occurs only after an in-memory miss and before a database read.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings -A deprecated`
- `cargo test -p turbo-tasks-hash -p turbo-tasks -p turbo-tasks-backend`
- `cargo check -p turbo-tasks-backend --benches`
- `cargo bench -p turbo-tasks-backend --bench mod -- task-input-hash`

### DIFF
--- a/crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
+++ b/crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
@@ -40,6 +40,13 @@ mod native_types {
     // Manual impl because `serde_json::Value` doesn't implement `TaskInput`, but `JsonValue` can
     // never contain any `Vc` types.
     impl turbo_tasks::TaskInput for JsonValue {
+        fn persistence_hash(&self, state: &mut dyn turbo_tasks::DeterministicHasher) {
+            turbo_tasks::TaskInput::persistence_hash(
+                &serde_json::to_string(&self.0).expect("JSON serialization should never fail"),
+                state,
+            );
+        }
+
         fn is_transient(&self) -> bool {
             false
         }
--- a/crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
+++ b/crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
@@ -40,7 +40,7 @@ mod native_types {
     // Manual impl because `serde_json::Value` doesn't implement `TaskInput`, but `JsonValue` can
     // never contain any `Vc` types.
     impl turbo_tasks::TaskInput for JsonValue {
-        fn persistence_hash(&self, state: &mut dyn turbo_tasks::DeterministicHasher) {
+        fn persistence_hash<H: turbo_tasks::DeterministicHasher>(&self, state: &mut H) {
             turbo_tasks::TaskInput::persistence_hash(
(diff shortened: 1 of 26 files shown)


mischnic:
But we already have `trait DeterministicHash` as well?

lukesandberg:
> But we already have `trait DeterministicHash` as well?

`DeterministicHash` explicitly says: 'no VCs' and here we need vcs

crates/next-core/src/next_shared/transforms/swc_ecma_transform_plugins.rs
```diff
@@ -40,6 +40,13 @@ mod native_types {
     // Manual impl because `serde_json::Value` doesn't implement `TaskInput`, but `JsonValue` can
     // never contain any `Vc` types.
     impl turbo_tasks::TaskInput for JsonValue {
+        fn persistence_hash<H: turbo_tasks::DeterministicHasher>(&self, state: &mut H) {
+            turbo_tasks::TaskInput::persistence_hash(
```
lukesandberg: in some of these cases it would be better to delegate to the underlying Hash impl if we could somehow assert that it is reliable

PR #98217, opened 2026-09-03
Prerender each declared static variant combination

unstubbable:
A page that exports `unstable_generateStaticVariants` declares the static variant combinations it wants prerendered. Until now the build read those declarations, validated them, and then discarded them. This change creates a prerender candidate for every combination and serves a request from the available output for the combination it resolved.

The build creates a prerender candidate for every route a page produces and every combination. For a dynamic page, this covers each row of `generateStaticParams` and each fallback shell. For a page with no dynamic segments, it covers the page itself. Each successful candidate writes its artifact under a path that carries the hash of the combination, for example `/__variants/1u0zqp3/blog/my-post`.

The hash sits in the path because that is how a platform addresses a cached artifact today. It serves such a path straight from its filesystem before any function runs, and rebuilds the same path to revalidate the artifact, with no proxy in the loop. A platform that could pick a variant-specific prerender from request headers alone would need no prefix, so this shape is the one the current constraints leave rather than the one to keep.

A proxy resolves the variants of a request, finds an available prerender combination they match, and rewrites to the prefixed path. Routing removes the prefix again before it matches a route, and puts the hash in `nxtV`, because route matching runs on route patterns and no prefixed path matches one. The origin recovers the combination from that hash rather than from the resolved values, because a request does not always arrive with them. A platform rebuilds the request from the artifact when it fills or revalidates a prerender, and the path is all it has to work from.

The render then divides the values into two tiers. The matched combination assigns the static tier, and the artifact is keyed on those values, so a render may put them in its output. Everything else the request resolved is the runtime tier, which no artifact key mentions, so a read of one waits until request time. A request whose combination has no available prerender output is not rewritten to a prefixed path, even if the page declared that combination. Its values remain in the runtime tier. Where the route is partially prerendered, the request is served from the one prerender that omits every variant. That shared prerender keeps the request from taking a cache entry of its own, however many values it happens to carry.

Two manifests expose the available combinations in the forms their consumers need. A proxy has a pathname and nothing else, so `variants-manifest.json` is keyed by pathname for a static route and holds a route regex for a dynamic one. It retains an empty entry when a declaring route produced no output, so a less-specific route cannot supply another page's combinations. The origin knows which route it renders, so the prerender manifest contains the non-empty groups keyed by page.

Variants need a `proxy.ts` file, which always runs on the Node.js runtime, so `experimental.variants` together with a `middleware.ts` is now a build error. Serving a combination from a deployment needs routing rules that a later change adds.

---
Extracted from #96832


packages/next/src/build/index.ts
```diff
@@ -3004,12 +3051,26 @@ export default async function build(
         path.join(distDir, SERVER_DIRECTORY, MIDDLEWARE_MANIFEST)
       )
 
+      // This file is written whenever the flag is on, even when no page
+      // declared anything. A proxy then reads a file that is either present and
+      // current, or absent because the feature is off, and never has to tell a
+      // stale file from a missing one.
+      if (config.experimental.variants) {
+        await writeManifest(
+          path.join(distDir, SERVER_DIRECTORY, VARIANTS_MANIFEST),
```
mischnic: I think you need to add this to the proxy NFT file, otherwise it won't be deployed into the function?

unstubbable: Yes, deployment support will be the next PR in this stack (WIP). The file is already added in the [monolithic draft PR](https://github.com/vercel/next.js/pull/96832) that does support deployments. (Although the implementation is not quite correct, I'll switch to `pushAsset` in this stack.)

PR #91416, opened 2026-03-15
fix(hmr): reconnect WebSocket after sleep/wake

timneutkens:
## Summary

When a computer sleeps and wakes, the `/_next/hmr` WebSocket connection can silently become stale. The current code relies entirely on `onclose`/`onerror` events, but:

- `onclose` may fire late or not at all after sleep/wake — the TCP connection is dead but the browser hasn't detected it yet
- `setTimeout` timers freeze during sleep, so queued retries can fire at once after wake, rapidly exhausting the 12-reconnect limit and triggering a spurious full-page reload

This PR adds `visibilitychange` and `online` event listeners to both the App Router and Pages Router HMR WebSocket clients. When the page becomes visible again or the browser regains connectivity, the WebSocket state is checked and a reconnection is triggered immediately if it's no longer open. The reconnection counter is reset since sleep-induced disconnections are not indicative of the dev server being down.

## How it works

- **`visibilitychange`**: When the document becomes `visible`, check `webSocket.readyState`. If not `OPEN`, reset the reconnection counter, clear any pending retry timer, and call `init()` to reconnect immediately.
- **`online`**: Same logic — when the browser fires the `online` event after regaining network connectivity.

This matches the approach recommended by the Socket.IO team and other WebSocket libraries for handling sleep/wake reconnection.

## Test plan

- Manual: open a Next.js dev app, put the computer to sleep (or simulate by disabling/re-enabling the network interface), verify the WebSocket reconnects and logs `[HMR] connected` when the tab is re-focused
- Verified: `pnpm --filter=next types` passes with no errors

PR #98229, opened 2026-09-03
fix(hmr): only reconnect a closed WebSocket on visibilitychange/online

### LINKED ISSUE #98212: Dev HMR client reconnects forever after prerender activation (visibilitychange while socket is CONNECTING)
### Where I found this

Chat app on a custom server, `next()` plus `ws`, started with `npm run dev`. Cold tab on `localhost:8080`, chat never connects. Server log is nothing but `/_next/hmr` upgrades, one a second. No call to our ticket endpoint, so the client code never ran. Refresh: works. Close the tab, open a new one, same URL: broken again. Console shows `WebSocket is closed before the connection is established`, then `[HMR] connected` 245 times. Chrome had "Preload pages" on. Prod build never does this.

https://github.com/hairywelly/next-hmr-visibility-repro

### To Reproduce

1. `npm install && npm run dev`
2. In Chrome open http://localhost:3000. The page has a Speculation Rules `prerender` for `/target`.
3. Wait a second, click the link to `/target` (or open `/?auto`, which navigates after 2.5s).
4. Open the console on `/target`.

Control: open http://localhost:3000/target directly.

Current: `WebSocket connection to 'ws://localhost:3000/_next/hmr?id=…' failed: WebSocket is closed before the connection is established`, then `[HMR] connected` again every second, indefinitely. `performance.getEntriesByType('navigation')[0].activationStart` is > 0 on the broken page and 0 on the control.

Expected: one HMR connection, as on the control page.

If the killed first socket had already reached the server it also takes the React debug channel with it (`debug-channel.ts` hands it out once per request id), and the page never hydrates. That is the case above: the server logged two `/_next/hmr` upgrades before the loop started.

Measured with headless Chrome over CDP, 16.3.3 and 16.4.0-canary.15: prerendered load opens 8 new HMR sockets in 8s; direct load opens 0.

### Cause

`packages/next/src/client/dev/hot-reloader/app/web-socket.ts` (since #91416):

- `handleVisibilityChange` calls `init()` when `readyState !== OPEN`. CONNECTING is not OPEN, so activation of a prerendered document fires it mid-handshake.
- `init()` closes the previous socket without clearing its `onclose`/`onerror`. That socket's `handleDisconnect` schedules `init()` again in 1s, which closes the new socket, and so on. `reconnections` resets on every `onopen`, so `WEB_SOCKET_MAX_RECONNECTIONS` never trips.

Fix that works locally:

```diff
   function init() {
     if (webSocket) {
+      webSocket.onclose = null
+      webSocket.onerror = null
       webSocket.close()
     }
...
   function handleVisibilityChange() {
-    if (document.visibilityState === 'visible' && webSocket.readyState !== WebSocket.OPEN) {
+    if (document.visibilityState === 'visible' && webSocket.readyState === WebSocket.CLOSED) {
```

Happy to open a PR.

### Verify canary release

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.6.0: Sat Jul 11 15:27:04 PDT 2026; root:xnu-12377.161.13~4/RELEASE_ARM64_T6050
  Available memory (MB): 49152
  Available CPU cores: 18
Binaries:
  Node: 24.16.0
  npm: 11.16.0
  Yarn: N/A
  pnpm: N/A
Relevant Packages:
  next: 16.3.3 // There is a newer version (16.3.4) available, upgrade recommended! 
  eslint-config-next: N/A
  react: 19.2.4
  react-dom: 19.2.4
  typescript: N/A
Next.js Config:
  output: N/A
```

Developer Experience, Turbopack

next dev (local)

marcoshernanz:
### What?

The `visibilitychange`/`online` handlers in the dev HMR WebSocket clients (App Router and Pages Router) now only reconnect when the socket is `CLOSED`, instead of whenever it is not `OPEN`.

### Why?

#91416 added those handlers so a stale socket reconnects after sleep/wake. But they also fired while the socket was still `CONNECTING`, which `init()` then closed. Closing a connecting socket fires its `onerror`/`onclose` → `handleDisconnect`, which schedules another `init()` one second later, which closes the new (healthy) socket, whose `handleDisconnect` schedules another `init()`… `reconnections` is reset on every `onopen`, so `WEB_SOCKET_MAX_RECONNECTIONS` never trips and the loop never ends.

This is not a rare race: Chrome defers the WebSocket handshake of a prerendered document (Speculation Rules `prerender`, or the "Preload pages" omnibox setting) until activation, and activation fires `visibilitychange`. So every prerendered `next dev` page hit this deterministically. Measured with real prerender activation over CDP (canary `65c832c`): `activationStart` 3330 ms, then a new `/_next/hmr` socket every second (7 sockets / 6× `[HMR] connected` in 6 s). If the first socket had already reached the server it also consumed the React debug channel for the request id (`debug-channel.ts` hands it out once), and the page never hydrated. Same trigger, reproduced synthetically (`visibilitychange` dispatched while `readyState === CONNECTING`): App Router 8 sockets in 6 s and no hydration; Pages Router 3 sockets, two `WebSocket is closed before the connection is established` errors and a 1 s delayed connection.

### How?

`readyState !== WebSocket.OPEN` → `readyState === WebSocket.CLOSED` in both handlers of both clients.

Only `handleDisconnect` and `init()` ever call `close()`, and `readyState` becomes `CLOSED` in the same task that fires `error`/`close`. So whenever a handler observes `CLOSED`, `handleDisconnect` has already run: the socket's handlers are detached and only a retry timer is pending. `init()` from the handler then cancels that timer and reconnects immediately, which is exactly the sleep/wake behaviour #91416 wanted (verified: `online`/`visibilitychange` while `CLOSED` reconnects in ~30 ms instead of waiting for the timer, then stays stable). A `CONNECTING` socket is left alone and either opens or fails into the normal retry path. This restores the pre-#91416 invariant that `init()` only ever runs with no socket or a closed, detached one.

With the fix, the same real-prerender run opens exactly one socket (created during prerendering, opened right after activation) and hydrates.

Regression tests dispatch `visibilitychange` right after the HMR socket is constructed (while it is `CONNECTING`) and assert that exactly one `/_next/hmr` WebSocket is created. Both fail on canary (App Router: 14 sockets, Pages Router: 3) and pass with the fix, in Turbopack and webpack dev modes.

Root-cause analysis and the proposed condition come from @hairywelly's report in #98212.

Fixes #98212

### DIFF
--- a/packages/next/src/client/dev/hot-reloader/app/web-socket.ts
+++ b/packages/next/src/client/dev/hot-reloader/app/web-socket.ts
@@ -142,10 +142,15 @@ export function createWebSocket(
     return newWebSocket
   }
 
+  // Reconnect when the page becomes visible or the browser comes back online
+  // (e.g. after sleep/wake) and the socket is CLOSED. A socket that is still
+  // CONNECTING must be left alone: closing it would fire its disconnect
+  // handler, which schedules a reconnect that closes the new socket, and so on
+  // (e.g. when a prerendered page is activated while the handshake is pending).
   function handleVisibilityChange() {
     if (
       document.visibilityState === 'visible' &&
-      webSocket.readyState !== WebSocket.OPEN
+      webSocket.readyState === WebSocket.CLOSED
     ) {
       reconnections = 0
       clearTimeout(timer)
@@ -154,7 +159,7 @@ export function createWebSocket(
   }
 
   function handleOnlineEvent() {
-    if (webSocket.readyState !== WebSocket.OPEN) {
+    if (webSocket.readyState === WebSocket.CLOSED) {
       reconnections = 0
       clearTimeout(timer)
       init()
--- a/packages/next/src/client/dev/hot-reloader/pages/websocket.ts
+++ b/packages/next/src/client/dev/hot-reloader/pages/websocket.ts
@@ -95,10 +95,15 @@ export function connectHMR(options: { path: string; assetPrefix: string }) {
     source.onmessage = handleMessage
   }
 
+  // Reconnect when the page becomes visible or the browser comes back online
+  // (e.g. after sleep/wake) and the socket is CLOSED. A socket that is still
+  // CONNECTING must be left alone: closing it would fire its disconnect
+  // handler, which schedules a reconnect that closes the new socket, and so on
+  // (e.g. when a prerendered page is activated while the handshake is pending).
   function handleVisibilityChange() {
     if (
       document.visibilityState === 'visible' &&
-      source.readyState !== WebSocket.OPEN
+      source.readyState === WebSocket.CLOSED
     ) {
       reconnections = 0
       clearTimeout(timer)
@@ -107,7 +112,7 @@ export function connectHMR(options: { path: string; assetPrefix: string }) {
   }
 
   function handleOnlineEvent() {
-    if (source.readyState !== WebSocket.OPEN) {
+    if (source.readyState === WebSocket.CLOSED) {
       reconnections = 0
       clearTimeout(timer)
       init()
--- a/test/development/app-hmr/websocket-reconnect.test.ts
+++ b/test/development/app-hmr/websocket-reconnect.test.ts
@@ -1,0 +1,50 @@
+import { FileRef, nextTestSetup } from 'e2e-utils'
+import { retry, waitFor } from 'next-test-utils'
+import path from 'path'
+
+describe('app-dir HMR WebSocket reconnect', () => {
+  const { next } = nextTestSetup({
+    files: new FileRef(path.join(__dirname, 'fixtures', 'default-template')),
+  })
+
+  it('should not replace the socket when the page becomes visible while it is connecting', async () => {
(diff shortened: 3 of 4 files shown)


marcoshernanz:
CI note for `3c44a5d`: lint, types, and all "new and changed tests" flake/deploy jobs pass. The two red test jobs are unrelated to this change and also fail or flake on canary at the same base commit:

- `test turbopack production (7/7)`: `test/production/app-dir/app-fetch-build-cache` › "should not use stale data if present" — identical failure in canary run [33785606413](https://github.com/vercel/next.js/actions/runs/33785606413) for the base commit `65c832c`.
- `test dev (5/10)`: `test/e2e/app-dir/actions/app-action-node-middleware.test.ts` › "should support updating the action" — passed only on retry 2/2, 2/2 and 1/2 in the last three canary runs (33785606413, 33785194051, 33777688010); here it exhausted its retries.

bgw (changes requested):
This has the exact same test twice

PR #98231, opened 2026-09-03
fix(after): make cookies() in after() reflect Route Handler mutations

### LINKED ISSUE #98222: cookies() inside after() returns pre-mutation values in a Route Handler

https://github.com/Nixxx19/nextjs-after-cookies-repro

### To Reproduce

`./repro.sh` in the linked repo. It builds a one-route app and runs it with `next start`.

The route handler reads `cookies()`, sets `session=NEW`, reads it again, and then reads it once more inside `after()`:

```js
export async function GET() {
  const store = await cookies()
  const inbound = store.get('session')?.value ?? '<none>'

  store.set('session', 'NEW')
  const inHandler = (await cookies()).get('session')?.value ?? '<none>'

  after(async () => {
    const inAfter = (await cookies()).get('session')?.value ?? '<none>'
    console.log(`AFTER_READ inbound=${inbound} inHandler=${inHandler} inAfter=${inAfter}`)
  })

  return Response.json({ inbound, inHandler })
}
```

Send a request carrying `Cookie: session=OLD`.

Current:

```
response headers: set-cookie: session=NEW; Path=/
response body:    {"inbound":"OLD","inHandler":"NEW"}
AFTER_READ inbound=OLD inHandler=NEW inAfter=OLD
```

`inAfter` is `OLD`. The handler set the cookie, its own read returns `NEW`, and the response carries `Set-Cookie: session=NEW`, but the read inside `after()` sees the inbound jar.

Expected: `inAfter` is `NEW`, matching what the handler set and what the response sends.

This is the pattern the docs recommend. `after.mdx`, under "In Route Handlers and Server Functions", says you can call `cookies` and `headers` directly inside the `after` callback and that it "is useful for logging activity after a mutation", with an example that performs a mutation and then logs `(await cookies()).get('session-id')` from inside `after()`. That log records the pre-mutation value.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 24576
  Available CPU cores: 10
Binaries:
  Node: 25.9.0
  npm: 11.12.1
  pnpm: 10.33.2
Relevant Packages:
  next: 16.4.0-canary.15
  react: 19.2.8
  react-dom: 19.2.8
```

Runtime, Route Handlers

next start (local)

### Additional context

`after()` sets `workUnitStore.phase = 'after'` in `server/after/after-context.ts`. `cookies()` only returns the mutable jar while `areCookiesMutableInCurrentPhase()` holds, and that is `requestStore.phase === 'action'` (`server/web/spec-extension/adapters/request-cookies.ts:220-222`), so once the phase flips it falls back to the immutable inbound snapshot `workUnitStore.cookies`.

That snapshot is only reconciled by `synchronizeMutableCookies` (`server/async-storage/request-store.ts`), and its only caller is the Server Action path at `server/app-render/action-handler.ts:1439`. Route Handlers have no equivalent, so nothing reconciles the jar before `after()` runs.

Verified on 16.4.0-canary.15. I have not tried to fix it; resyncing before the phase flips, or calling `synchronizeMutableCookies` on the route-handler finalize path, look like the two obvious places, but which one is right is your call.

marcoshernanz:
### What?

`cookies()` called inside `after()` in a Route Handler returned the inbound request cookies, ignoring any `cookies().set()` / `.delete()` the handler performed. Server Actions already returned the post-mutation values in the same scenario.

```ts
export async function GET() {
  const store = await cookies()
  store.set('session', 'NEW')

  after(async () => {
    // before: 'OLD' (inbound cookie), even though the response sends `Set-Cookie: session=NEW`
    // after:  'NEW'
    console.log((await cookies()).get('session')?.value)
  })

  return Response.json({ ok: true })
}
```

### Why?

The `after()` docs recommend reading `cookies()` inside the callback in Route Handlers "for logging activity after a mutation", but the callback saw pre-mutation values.

While a Route Handler runs, `cookies()` returns `requestStore.userspaceMutableCookies`, which reflects mutations. `after()` callbacks run once the response is done, when the request store is in the `'after'` phase and `cookies()` returns the immutable snapshot `requestStore.cookies`, which is still the inbound jar.

Server Actions don't hit this because `executeActionAndPrepareForRender` calls `synchronizeMutableCookies()` once the action body has finished, so the snapshot reflects the mutations. Route Handlers never did that.

### How?

Call the same `synchronizeMutableCookies()` in `AppRouteRouteModule.renderToResponse` once the handler has run (in a `finally`, so it also covers handlers that end with `redirect()` / `notFound()`). `requestStore.mutableCookies` is also what `finalizeResponse` and `handleHandlerError` use to emit `Set-Cookie`, so `after()` now sees exactly what the response sent.

Added e2e tests for a Route Handler (fails before this change) and a Server Action (control, passes before and after).

Fixes #98222

### DIFF
--- a/packages/next/src/server/route-modules/app-route/module.ts
+++ b/packages/next/src/server/route-modules/app-route/module.ts
@@ -12,7 +12,10 @@ import {
   type RouteModuleHandleContext,
   type RouteModuleOptions,
 } from '../route-module'
-import { createRequestStoreForAPI } from '../../async-storage/request-store'
+import {
+  createRequestStoreForAPI,
+  synchronizeMutableCookies,
+} from '../../async-storage/request-store'
 import {
   createPrerenderWorkStore,
   createWorkStore,
@@ -839,6 +842,13 @@ export class AppRouteRouteModule extends RouteModule<
         requestStore,
         context
       )
+    } finally {
+      // `after()` callbacks run once the response is done, when `cookies()`
+      // returns the immutable `cookies` snapshot instead of the mutable
+      // cookies. Synchronize them so that the callbacks see any cookies the
+      // handler set, like `executeActionAndPrepareForRender` does for Server
+      // Actions.
+      synchronizeMutableCookies(requestStore)
     }
 
     return this.finalizeResponse(
--- a/test/e2e/app-dir/next-after-app-api-usage/app/cookies-mutation/route-handler/route.js
+++ b/test/e2e/app-dir/next-after-app-api-usage/app/cookies-mutation/route-handler/route.js
@@ -1,0 +1,16 @@
+import { cookies } from 'next/headers'
+import { after } from 'next/server'
+
+export async function GET() {
+  const path = '/cookies-mutation/route-handler'
+
+  const cookieStore = await cookies()
+  cookieStore.set('test-cookie', 'mutated')
+
+  after(async () => {
+    const value = (await cookies()).get('test-cookie')?.value
+    console.log(`[${path}] cookies() in after(): test-cookie=${value}`)
+  })
+
+  return new Response()
+}
--- a/test/e2e/app-dir/next-after-app-api-usage/app/cookies-mutation/server-action/page.js
+++ b/test/e2e/app-dir/next-after-app-api-usage/app/cookies-mutation/server-action/page.js
@@ -1,0 +1,23 @@
+import { cookies } from 'next/headers'
+import { after } from 'next/server'
+
+export default async function Page() {
+  return (
+    <form
+      action={async () => {
+        'use server'
+        const path = '/cookies-mutation/server-action'
+
+        const cookieStore = await cookies()
+        cookieStore.set('test-cookie', 'mutated')
+
+        after(async () => {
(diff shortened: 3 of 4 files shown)


marcoshernanz:
CI note: the remaining `test turbopack production (7/7)` failure is `test/production/app-dir/app-fetch-build-cache › should not use stale data if present` (60s timeout, 3/3 attempts). It fails identically on canary at this PR's base commit 65c832c8b8 ([run 33785606413](https://github.com/vercel/next.js/actions/runs/33785606413/job/100766452520)), and the suite doesn't touch `after()` or cookies, so it's unrelated to this change. The `test dev (10/10)` HMR failure from the first pass (`app-action-node-middleware › should support updating the action`) passed on re-run.

marcoshernanz:
Reworked the fix to be smaller and to mirror the Server Action code path exactly: the sync now lives in `AppRouteRouteModule.renderToResponse` (a `finally` around the handler run) instead of `AfterContext`'s close listener. Net change in `packages/next`: +12/−1 in `app-route/module.ts`, half of which is a comment. Same tests.

PR #98232, opened 2026-09-03
Fix scroll to top after a redirect during client navigation

### LINKED ISSUE #98172: Scroll to top does not work through redirects in proxy

https://github.com/icyJoseph/scroll-proxy-redirect

### To Reproduce

1. Clone the reproduction, `pnpm install`, `pnpm exec playwright install chromium`.
2. `pnpm run build && pnpm run start` (production build — this does not reproduce under `next dev`).
3. Open `http://localhost:3000/en` and scroll to the bottom of the page (it is 3000px tall).
4. Click the link **"Go to /about (relies on proxy 308 -> /en/about)"**. `proxy.ts` responds with a 308 redirect to `/en/about`.
5. Observe the scroll position on `/en/about`: the page keeps the previous scroll offset (`window.scrollY` is ~2540), so the `<h1>` is off-screen.
6. Go back to `/en`, scroll to the bottom again, and click **"Go to /en/about (no proxy redirect)"** instead. This navigation lands at `window.scrollY === 0`, as expected.

`pnpm test` automates steps 3-6 as two Playwright tests (the redirected case and the non-redirected control).

**Expected:** a `<Link>` navigation to a different route scrolls the destination to the top, whether or not `proxy.ts` redirected the request.

**Current:** when the navigation is 308-redirected by `proxy.ts`, the scroll offset of the origin page is preserved on the destination page. Navigating to the exact same final URL with a `<Link>` that is not redirected resets scroll to 0, so the redirect is what makes the difference.

Test output from the reproduction (production build):

```
[proxy-redirect] scrollY after navigation: 2540
    Error: Expected target page to scroll to top; got scrollY=2540
[no-redirect]    scrollY after navigation: 0

  1 failed
  1 passed
```

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.6.0: Fri Jul 31 19:17:26 PDT 2026; root:xnu-12377.161.14~5/RELEASE_ARM64_T6041
  Available memory (MB): 24576
  Available CPU cores: 14
Binaries:
  Node: 22.18.0
  npm: 10.9.3
  Yarn: N/A
  pnpm: 10.34.1
Relevant Packages:
  next: 16.4.0-canary.14 // No Next.js version data was found.
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Linking and Navigating, Middleware, Redirects

### Additional context

- Reproduces on `16.4.0-canary.14` and on `16.3.4` stable, with the same dev/prod split: `next dev` passes both tests, `next build` + `next start` fails the redirected navigation.
- Not related to `experimental.appNewScrollHandler`: the failure is identical with the flag set to `true` and to `false`, so the reproduction leaves it unset.
- Minor, possibly unrelated observation: `proxy.ts` sets `status: 308`, but the `next dev` server log reports the redirected RSC request as `307`.
- Not yet bisected to a specific canary release.

marcoshernanz:
When a client-side navigation is redirected by the proxy to a route that optimistic routing didn't predict (e.g. `/about` → `/en/about` under a locale-prefix proxy), the destination page rendered and the URL was corrected, but the page kept the previous scroll position instead of scrolling to the top. Navigating to the same destination without a redirect scrolled correctly.

Since #95207, a redirect discovered during a navigation triggers a `RedirectRetry`, which re-resolves the route and reuses the data already received. That retry was dispatched with `FreshnessPolicy.HistoryTraversal`. But `HistoryTraversal` is the back/forward policy: it restores a previous state rather than rendering a new one, so it deliberately skips assigning a `ScrollRef` to new segments (`accumulateScrollRef`), reads the BFCache regardless of staleness (`createCacheNodeForSegment`), and skips default-slot reuse. When the redirect changes the tree shape, the retry creates the new leaf segments without a `ScrollRef`, so nothing scrolls. It also means that if the destination was visited earlier in the session, the retry renders the stale BFCache entry instead of the fresh response it just received.

The retry now uses `FreshnessPolicy.Default`, the same policy a seeded redirect from a Server Action uses (`server-action-reducer.ts`). `Default` still reuses the seed data instead of re-fetching (segments with seed data are created with `needsDynamicRequest: false`), and additionally scrolls new segments into view and respects BFCache staleness like any other forward navigation. `ServerPatchAction.freshnessPolicy` is narrowed to `RefreshAll | Default` accordingly.

Adds an end-to-end test that scrolls to the bottom of a tall page, navigates through a proxy redirect, and asserts the destination is scrolled to the top (with a non-redirected control). The test fails on `canary` and passes with this change; the `redirect-rewrite-dynamic` test from #95207 continues to pass.

Fixes #98172

### DIFF
--- a/packages/next/src/client/components/render-tree.ts
+++ b/packages/next/src/client/components/render-tree.ts
@@ -1599,7 +1599,15 @@ async function finishNavigationTask(
       // The route matched, but the request was redirected, so we committed the
       // wrong canonical URL. Re-resolve the route to invalidate the now-stale
       // route cache and correct the URL — but reuse the data we already received
-      // (HistoryTraversal) instead of re-fetching it. See issue #95195.
+      // instead of re-fetching it.
+      //
+      // This is a forward navigation to the redirect destination that happens
+      // to be seeded with the response, so it uses the Default policy (like a
+      // redirect from a Server Action): segments that changed are created from
+      // the seed and scrolled into view like any other navigation. It must not
+      // use HistoryTraversal, which suppresses the scroll and reads the BFCache
+      // regardless of staleness, because back/forward restores a previous
+      // state rather than rendering a new one. See issues #95195 and #98172.
       const isHardRetry = false
       const primaryRequestResult = await primaryRequestPromise
       dispatchRetryDueToTreeMismatch(
@@ -1610,7 +1618,7 @@ async function finishNavigationTask(
         task.route,
         routeCacheEntry,
         navigateType,
-        FreshnessPolicy.HistoryTraversal
+        FreshnessPolicy.Default
       )
       return
     }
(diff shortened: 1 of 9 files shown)


gilangabdian:
Really nice investigation and fix! I was reading through the change and was wondering if there are any edge cases or scenarios that might still be missed by the current regression test.

For example, since `FreshnessPolicy.Default` also changes how BFCache is considered during the retry, do you think there are any cases involving a previously visited destination or other redirect/rewrite combinations that could behave differently?

Just curious if there are any scenarios you considered but decided not to cover in this PR. Thanks!

PR #98233, opened 2026-09-03
Key the Pages Router app route prefetch marker by the `as` path

### LINKED ISSUE #98180: Pages Router: `prefetch()` checks the client router filter against `as` but stores the `__appRouter` marker under the `href` pathname

### LINKED ISSUE #98181: Pages Router: shallow `getRouteInfo()` returns the `__appRouter` prefetch marker as route info and crashes `_app`

https://github.com/Stanzilla/next-pages-router-prefetch-bloom-filter-repro

### To Reproduce

The marker can be planted directly, so this does not depend on the Bloom filter and reproduces in `next dev` as well:

1. Open any pages-router page with a dynamic route in the repository, for example `http://localhost:3000/players/alice`.
2. In the console run:

```js
window.beforeNav = 1
window.next.router.components['/players/[name]'] = { __appRouter: true }
window.next.router.push('/players/[name]?tab=stats', '/players/alice?tab=stats', { shallow: true })
```

The same state is reached organically through `router.prefetch()` when the client router filter matches the `as` path of a link whose `href` is the current route pattern (see https://github.com/vercel/next.js/issues/98180 for the end-to-end reproduction: hover the **Room 00000000-0000-4000-8000-000000000c4c** link, then click a tab).

**Current**

The page crashes. In production the UI shows "Application error: a client-side exception has occurred" and the console logs `TypeError: Cannot destructure property 'title' of 'n' as it is undefined.` (`_app` destructures `pageProps`). With the default `_app` the error is `Element type is invalid` because `Component` is `undefined`. `window.beforeNav` stays `1`, so no navigation happened.

**Expected**

The router must never hand `{ __appRouter: true }` to the render path. For a shallow navigation it should either fetch the route info again or hard-navigate, but not render the marker.

### Root cause

`Router.getRouteInfo()` in `packages/next/src/shared/lib/router/router.ts` returns whatever is cached for the current route on a shallow navigation without checking that it is route info:

```ts
let existingInfo: PrivateRouteInfo | undefined = this.components[route]
if (routeProps.shallow && existingInfo && this.route === route) {
  return existingInfo
}
```

`Router.prefetch()` stores `{ __appRouter: true }` in the same `this.components` map when the client router filter matches. The `__appRouter` guard in `Router.change()` only catches the case where the marker sits under the concrete `href` pathname, and `_bfl` is skipped for shallow navigations, so a marker that ends up under the route pattern (https://github.com/vercel/next.js/issues/98180 explains how) flows straight through `getRouteInfo()` into `_app` with no `Component` and no `props`.

The same unguarded lookup exists in the rewrite branch ("Check again the cache with the new destination").

```
Operating System:
  Platform: darwin
  Arch: arm64
  Available memory (MB): 36864
  Available CPU cores: 12
Binaries:
  Node: 24.19.0
  npm: 11.17.0
  Yarn: N/A
  pnpm: 11.24.0
Relevant Packages:
  next: 16.4.0-canary.13
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Also reproduced with `next@16.3.3` and `next@16.3.4`. The relevant code is unchanged on `canary` (538f3f6).

### Additional context

A fix that makes `getRouteInfo()` treat an `__appRouter` marker as a cache miss is proposed in a linked pull request. It is independent of the fix for the marker key so that a stale marker can never crash the page even if another code path writes one.

Disclosure: the investigation, the reproduction repository and this report were prepared with AI assistance (Claude Code) on behalf of the reporter.

marcoshernanz:
### What?

`Router.prefetch()` in the Pages Router checks the client router filter against a link's `as` path and, on a match, records `{ __appRouter: true }` in `router.components` so that clicking the link hard-navigates right away. That marker was stored under the `href` pathname and looked up by the `href` pathname in `Router.change()`. This PR keys it by the normalized `as` pathname in both places, and makes `Router.getRouteInfo()` treat a marker as a cache miss so it can never be handed to `_app` as route info.

### Why?

`href` and `as` are the same URL for most links, so the mismatch was invisible. They differ for "route as modal" links (`href={{ pathname: router.pathname, query: { ...router.query, modal: id } }}` with `as="/pretty/url"`, see `examples/with-route-as-modal`). When the filter matches `as`, or hits one of its false positives (0.01% per checked path by default; every hover/touch checks a path), the marker replaced the cached route info of the page the user is currently on:

- on a static route the `change()` guard then fired for every later navigation to that route, so a shallow query change hard-navigated;
- on a dynamic route the marker sat under the pattern (`/blog/[slug]`) but was looked up by the concrete `href` (`/blog/first`), so the guard missed and the shallow `getRouteInfo()` early return handed the marker to `_app`: `Component` and `props` are `undefined`, which renders the page without its props or crashes a custom `_app` that reads `pageProps`. Every object-form `<Link href={{ pathname: '/blog/[slug]', ... }}>` through that pattern also hard-navigated.

Fixes #98180
Fixes #98181

This consolidates the fixes proposed by @Stanzilla in #98182 and #98187 (thanks for the detailed report and reproduction) into one change with a single fixture in the existing `pages-to-app-routing` suite; Benjamin is credited as co-author.

### How?

- `getAppRouterMarkerKey(router, as)` strips `basePath` and the locale prefix from `as`, takes the pathname and removes the trailing slash. It returns `null` for non-local URLs (`mailto:` etc.), which can never be an App Router path. This mirrors the normalisation `change()` already applies to `href` before the lookup.
- `prefetch()` writes the marker under `getAppRouterMarkerKey(this, asPath)`; `change()` reads it with `getAppRouterMarkerKey(this, cleanedAs)`. For `href === as` links nothing changes.
- `getRouteInfo()` resets `existingInfo` when it carries `__appRouter` (both the initial lookup and the re-check after a middleware rewrite), so the shallow early return does not fire and `cachedRouteInfo` is not the marker; the route is loaded again instead.

### Tests

New `prefetch-marker` fixture in `test/e2e/app-dir/pages-to-app-routing` (pages `/` and `/blog/[slug]` with a shallow tab link and a route-as-modal link whose `as` is `app/dashboard`):

- planting the marker under the current dynamic route and doing a shallow `router.push()` re-renders with the new query and keeps the page's props (dev + prod);
- after the route-as-modal link has been prefetched (prod only, prefetch is a no-op in dev): clicking it still hard-navigates to the app route; a shallow navigation on the static route stays client-side; a shallow navigation on the dynamic route stays client-side and keeps its props; an object-form navigation to another slug stays client-side.

All four navigation tests fail on `canary` (hard navigation, or `Pages Blog:` rendered without `slug`); the marker still hard-navigates the app-route link before and after.

Also verified on the same build: full `pages-to-app-routing` (prod + dev, incl. the `basePath` fixture), `app-dir/app` "should successfully detect app route during prefetch" (the original #47761 test), `invalid-href` (prod + dev, covers external `as`), `client-shallow-routing`, `middleware-shallow-link`, `dynamic-route-interpolation`, `basepath` (35), `i18n-support` (103), plus `basePath: '/base'` and `i18n` variants of the new fixture.

x-ref: #47761 (introduced the marker), #47949

### DIFF
--- a/packages/next/src/shared/lib/router/router.ts
+++ b/packages/next/src/shared/lib/router/router.ts
@@ -171,6 +171,34 @@ function resolveDynamicRoute(pathname: string, pages: string[]) {
   return removeTrailingSlash(pathname)
 }
 
+/**
+ * Key under which `prefetch()` stores the `{ __appRouter: true }` marker in
+ * `router.components` when the client router filter matches an `as` path, and
+ * under which `change()` looks it up again.
+ *
+ * The filter is evaluated against the `as` path, so the marker has to be keyed
+ * by the `as` pathname as well. `href` and `as` can differ (e.g. `href` is the
+ * current route with an extra query param and `as` is a different URL), and
+ * keying the marker by the `href` pathname would replace the cached route info
+ * of a pages route the filter never matched.
+ *
+ * Returns `null` for non-local URLs (e.g. `mailto:`), which can never be an
+ * App Router path.
+ */
+function getAppRouterMarkerKey(router: NextRouter, as: string): string | null {
+  if (!isLocalURL(as)) {
+    return null
+  }
+
+  let { pathname } = parseRelativeUrl(hasBasePath(as) ? removeBasePath(as) : as)
+
+  if (process.env.__NEXT_I18N_SUPPORT) {
+    pathname = normalizeLocalePath(pathname, router.locales).pathname
+  }
+
+  return removeTrailingSlash(pathname)
+}
+
 function getMiddlewareData<T extends FetchDataOutput>(
   source: string,
   response: Response,
@@ -1446,9 +1474,13 @@ export default class Router implements BaseRouter {
     let route = removeTrailingSlash(pathname)
     const parsedAsPathname = as.startsWith('/') && parseRelativeUrl(as).pathname
 
-    // if we detected the path as app route during prefetching
+    // if we detected the `as` path as app route during prefetching
     // trigger hard navigation
-    if ((this.components[pathname] as any)?.__appRouter) {
+    const appRouterMarkerKey = getAppRouterMarkerKey(this, cleanedAs)
+    if (
+      appRouterMarkerKey !== null &&
+      (this.components[appRouterMarkerKey] as any)?.__appRouter
+    ) {
       handleHardNavigation({ url: as, router: this })
       return new Promise(() => {})
     }
@@ -2073,6 +2105,15 @@ export default class Router implements BaseRouter {
 
     try {
       let existingInfo: PrivateRouteInfo | undefined = this.components[route]
+
+      // `prefetch()` stores an `__appRouter` marker in `this.components` when
+      // the client router filter matches a path. It is not route info (it has
+      // no `Component` and no `props`), so treat it as a cache miss and load
+      // the route again instead of rendering it.
+      if ((existingInfo as any)?.__appRouter) {
+        existingInfo = undefined
+      }
+
       if (routeProps.shallow && existingInfo && this.route === route) {
         return existingInfo
       }
@@ -2173,6 +2214,9 @@ export default class Router implements BaseRouter {
 
           // Check again the cache with the new destination.
           existingInfo = this.components[route]
+          if ((existingInfo as any)?.__appRouter) {
+            existingInfo = undefined
+          }
           if (
             routeProps.shallow &&
             existingInfo &&
@@ -2416,7 +2460,6 @@ export default class Router implements BaseRouter {
       return
     }
     let parsed = parseRelativeUrl(url)
-    const urlPathname = parsed.pathname
 
     let { pathname, query } = parsed
     const originalPathname = pathname
@@ -2553,8 +2596,12 @@ export default class Router implements BaseRouter {
 
     const route = removeTrailingSlash(pathname)
 
-    if (await this._bfl(asPath, resolvedAs, options.locale, true)) {
-      this.components[urlPathname] = { __appRouter: true } as any
+    const appRouterMarkerKey = getAppRouterMarkerKey(this, asPath)
+    if (
+      appRouterMarkerKey !== null &&
+      (await this._bfl(asPath, resolvedAs, options.locale, true))
+    ) {
+      this.components[appRouterMarkerKey] = { __appRouter: true } as any
     }
 
     await Promise.all([
--- a/test/e2e/app-dir/pages-to-app-routing/fixtures/prefetch-marker/app/dashboard/page.tsx
+++ b/test/e2e/app-dir/pages-to-app-routing/fixtures/prefetch-marker/app/dashboard/page.tsx
@@ -1,0 +1,3 @@
+export default function Page() {
+  return <h1 id="page-title">App Dashboard</h1>
+}
--- a/test/e2e/app-dir/pages-to-app-routing/fixtures/prefetch-marker/app/layout.tsx
+++ b/test/e2e/app-dir/pages-to-app-routing/fixtures/prefetch-marker/app/layout.tsx
@@ -1,0 +1,7 @@
+export default function Root({ children }: { children: React.ReactNode }) {
+  return (
+    <html>
+      <body>{children}</body>
+    </html>
+  )
+}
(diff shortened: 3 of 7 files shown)


marcoshernanz:
Closing this draft: the reporter's #98182 and #98187 already cover the same change.

PR #98234, opened 2026-09-03
Do not error when a cached page forwards its props to a child component

### LINKED ISSUE #98173: Incorrect error message when awaiting params underneath 'use cache'

https://github.com/mdj-uk/next-bad-error-message/tree/37cf318e579a31ec0dca77e209ac89b5b8502d70

### To Reproduce

1. start app in dev
2. navigate to dynamic route b

You'll see error message "Route /[slug] used `searchParams` inside "use cache"...." even though `searchParams` are not used

There's an error due to calling `await params` in a 'use cache' context. The error should refer to `params`. But instead it refers to `searchParams`.

In stable (16.3.4) the full error is

> Route /[slug] used `searchParams` inside "use cache". Accessing dynamic request data inside a cache scope is not supported. If you need some search params inside a cached function await `searchParams` outside of the cached function and pass only the required search params as arguments to the cached function. See more info here: [https://nextjs.org/docs/messages/next-request-in-use-cache](https://nextjs.org/docs/messages/next-request-in-use-cache)

In canary ( 16.4.0-canary.13), error message is slightly different: 

> Route "/[slug]": `searchParams` can't be read inside `"use cache"`. Await it outside the cached function and pass what you need as an argument. Learn more: [https://nextjs.org/docs/messages/next-request-in-use-cache](https://nextjs.org/docs/messages/next-request-in-use-cache)

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.2.0: Wed Nov 15 21:53:18 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6000
  Available memory (MB): 16384
  Available CPU cores: 8
Binaries:
  Node: 26.8.1
  npm: 8.15.0
  Yarn: N/A
  pnpm: 12.1.0
Relevant Packages:
  next: 16.3.4 // Latest available version is detected (16.3.4).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

Use Cache

next dev (local)

### Additional context

_No response_

marcoshernanz:
### What?

A cached page that forwards its props to a child Server Component reports `` `searchParams` can't be read inside `"use cache"` `` even though it never reads a search param. This shows a redbox in `next dev` and fails `next build --debug-prerender`.

```tsx
export default async function Page(props: PageProps<'/[slug]'>) {
  'use cache'

  return <PageInner {...props} />
}

async function PageInner(props: PageProps<'/[slug]'>) {
  const { slug } = await props.params

  return <h1>Page {slug}</h1>
}
```

Fixes #98173

### Why?

`searchParams` can't be read inside a public `"use cache"` scope, so we pass in a promise that errors. That promise threw when its `then` property was *read*, which means merely touching the promise was enough to trigger the error.

React's dev-only debug info serializer walks the props of every Server Component it renders and probes each prop for thenability by reading `.then`:

```
at throwForSearchParamsAccessInUseCache
at Object.get                     <- the erroring searchParams proxy
at renderDebugModel
at stringify
at emitOutlinedDebugModelChunk
at outlineDebugModel
at outlineComponentInfo
at renderFunctionComponent
```

React catches the error, but by then we have already recorded it on `workStore.invalidDynamicUsageError` — which we need so that a `searchParams` access swallowed by a `try`/`catch` is still reported (#77838) — and `collectResult` fails the cache render with it.

This regressed in #82578, which replaced the hanging `searchParams` promise for public caches with the erroring one. It only affects `next dev` and `next build --debug-prerender`, because debug info is not serialized in a production build.

### How?

The guard moves from the promise to the object that the promise resolves to. Awaiting the promise is now side-effect free, and consuming the search params is what errors. React's probe only reads well-known properties (`then`, `$$typeof`, `toJSON`, `@@iterator`), so it no longer trips the guard.

React does enumerate the resolved object when it serializes it as a component prop, so the `ownKeys` trap throws without recording the error. That keeps `{ ...(await searchParams) }` erroring in user code — the thrown error propagates through the render — while the serializer's own enumeration stays silent. Without this split, guarding `ownKeys` reintroduces the exact same false positive.

Everything that errored before still errors, with unchanged messages:

| | before | after |
| --- | --- | --- |
| `(await searchParams).foo` | errors | errors |
| `searchParams.foo` (no `await`) | errors | errors |
| `{ ...(await searchParams) }`, `Object.keys(await searchParams)` | errors | errors |
| `'foo' in (await searchParams)` | errors | errors |
| forwarding the promise into a nested cached function | errors | errors |
| passing the promise to a Client Component | errors | errors |
| **forwarding page props to a child Server Component** | **errors** | **no error** |

The only user-visible difference for the cases that still error is that the reported source location is now the read itself rather than the `await`, which the updated snapshots reflect:

```diff
- "source": "app/search-params-used/page.tsx (8:17) @ Page
+ "source": "app/search-params-used/page.tsx (8:38) @ Page
  >  8 |   const param = (await searchParams).foo
-      |                 ^",
+      |                                      ^",
```

### Verified

- `test/e2e/app-dir/use-cache-search-params` passes in `test-dev-turbo`, `test-dev-webpack` and `test-start-turbo`, with and without `cacheComponents`. Two tests are added: one for the forwarded-props case, one pinning that enumeration still errors.
- `test/e2e/app-dir/cache-components-errors` and `test/e2e/app-dir/use-cache` (228 tests) pass with `cacheComponents`, as do `use-cache-hanging`, `use-cache-hanging-inputs` and `dynamic-data`.
- On the reproduction from #98173, `next dev`, `next build` and `next build --debug-prerender` are clean for `/a`, and `/b` now reports the `params` error the issue author expected instead of the misleading `searchParams` one.

### DIFF
--- a/packages/next/src/server/request/search-params.ts
+++ b/packages/next/src/server/request/search-params.ts
@@ -42,6 +42,7 @@ import {
 import {
   throwWithStaticGenerationBailoutErrorWithDynamicError,
   throwForSearchParamsAccessInUseCache,
+  throwForSearchParamsEnumerationInUseCache,
 } from './utils'
 
 export type SearchParams = { [key: string]: string | string[] | undefined }
@@ -522,7 +523,34 @@ export function makeErroringSearchParamsForUseCache(): Promise<SearchParams> {
     return cachedSearchParams
   }
 
-  const promise = Promise.resolve({})
+  // The error is triggered when a search param is read from the resolved
+  // object, and not when the promise itself is awaited. Merely touching the
+  // promise must stay side-effect free, because React's dev-only debug info
+  // serializer probes every prop of a rendered Server Component for
+  // thenability by reading `.then`. Otherwise a cached page that only forwards
+  // its props to a child Server Component would error without reading any
+  // search param.
+  const underlyingSearchParams = new Proxy({} as SearchParams, {
+    get: function get(target, prop, receiver) {
+      if (typeof prop === 'string' && !wellKnownProperties.has(prop)) {
+        throwForSearchParamsAccessInUseCache(workStore, get)
+      }
+
+      return ReflectAdapter.get(target, prop, receiver)
+    },
+    has: function has(target, prop) {
+      if (typeof prop === 'string' && !wellKnownProperties.has(prop)) {
+        throwForSearchParamsAccessInUseCache(workStore, has)
+      }
+
+      return Reflect.has(target, prop)
+    },
+    ownKeys: function ownKeys() {
+      throwForSearchParamsEnumerationInUseCache(workStore, ownKeys)
+    },
+  })
+
+  const promise = Promise.resolve(underlyingSearchParams)
 
   const proxiedPromise = new Proxy(promise, {
     get: function get(target, prop, receiver) {
@@ -534,10 +562,8 @@ export function makeErroringSearchParamsForUseCache(): Promise<SearchParams> {
         return ReflectAdapter.get(target, prop, receiver)
       }
 
-      if (
-        typeof prop === 'string' &&
-        (prop === 'then' || !wellKnownProperties.has(prop))
-      ) {
+      if (typeof prop === 'string' && !wellKnownProperties.has(prop)) {
+        // Sync access of a search param on the unawaited promise.
         throwForSearchParamsAccessInUseCache(workStore, get)
       }
 
--- a/packages/next/src/server/request/utils.ts
+++ b/packages/next/src/server/request/utils.ts
@@ -26,6 +26,25 @@ export function throwForSearchParamsAccessInUseCache(
   throw error
 }
 
+/**
+ * Same as `throwForSearchParamsAccessInUseCache`, except that the error is not
+ * recorded as an invalid dynamic usage. This is used for enumerating the search
+ * params, which React's dev-only debug info serializer does itself when it
+ * serializes them as a component prop. Recording it would report an error for a
+ * cached page that never read a search param. An enumeration in user code is
+ * still surfaced, because the thrown error propagates through the render.
+ */
+export function throwForSearchParamsEnumerationInUseCache(
+  workStore: WorkStore,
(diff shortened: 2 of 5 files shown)


marcoshernanz:
Closing this. The diagnosis above is accurate and reproducible, but the implementation isn't the right shape to land.

The fix relies on the erroring `searchParams` object exposing only properties that React's dev debug serializer treats as safe, which is an empirical property of the current serializer rather than a structural guarantee. It also needed a second, non-recording thrower for `ownKeys` (because the serializer enumerates the object itself), which creates an inconsistency I don't think should be decided in a drive-by fix: a `searchParams` read swallowed by `try`/`catch` is still reported, but a swallowed `{ ...(await searchParams) }` is not.

The more robust fix is for the guard to not record errors that React's own introspection caused, rather than narrowing what counts as an access. That needs a signal from React that debug-info serialization is in progress, so it belongs with whoever owns this guard — particularly given the plan to stop erroring on `searchParams` in `"use cache"` and to track params via Server Queries.

Leaving the root cause here for whoever picks it up:

- `makeErroringSearchParamsForUseCache` throws when the promise's `then` property is *read*.
- React's dev-only debug serializer probes every Server Component prop for thenability: `renderFunctionComponent` → `outlineComponentInfo` → `outlineDebugModel` → `emitOutlinedDebugModelChunk` → `stringify` → `renderDebugModel` → `typeof value.then === 'function'`.
- React catches the throw, but `throwForSearchParamsAccessInUseCache` has already set `workStore.invalidDynamicUsageError`, and `collectResult` then fails the cache render.
- So a cached page that only forwards its props to a child Server Component errors without reading any search param, in `next dev` and in `next build --debug-prerender`. A normal production build is unaffected, since debug info isn't serialized.
- Regressed in #82578, which replaced the hanging `searchParams` promise for public caches with the erroring one.

Issue #98111, opened 2026-08-31
Turbopack panics when a Git worktree symlinks node_modules to the primary checkout


Innei:
### Link to the code that reproduces this issue

https://github.com/Innei/next-turbopack-external-node-modules-worktree-repro

### To Reproduce

The reproduction is based on the official `reproduction-template` and uses the
latest available canary release.

1. Clone the reproduction repository.
2. Install dependencies with `pnpm install`.
3. Run `./reproduce.sh`.

The script creates a sibling Git worktree, symlinks the worktree's `node_modules`
to the dependency installation in the primary checkout, and starts `next dev` from
the worktree.

The same application starts normally when `next dev` is run directly from the
primary checkout.

### Current vs. Expected behavior

**Current behavior**

Turbopack prints its ready message and then exits with an internal panic before the
application can be served:

```text
Symlink node_modules could not be resolved: the symlink target leaves the filesystem root or its parent directory could not be resolved
```

**Expected behavior**

`next dev` should resolve the accessible external `node_modules` symlink and run the
application from the Git worktree. If external dependency roots cannot be supported
by default, Next.js should provide an explicit opt-in configuration for them and
report an actionable configuration error instead of an internal Turbopack panic.

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 25.4.0
  Available memory (MB): 131072
  Available CPU cores: 16
Binaries:
  Node: 24.18.0
  npm: 11.7.0
  Yarn: 1.22.22
  pnpm: 11.24.0
Relevant Packages:
  next: 16.4.0-canary.12
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

### Which area(s) are affected?

Turbopack, Module Resolution

### Which stage(s) are affected?

next dev (local)

### Additional context

AI-assisted development increasingly runs several independent coding sessions in
parallel, with each session isolated in its own Git worktree. Reinstalling a large
dependency graph for every short-lived worktree adds significant setup time and
filesystem overhead, so sharing the primary checkout's existing `node_modules`
through a symlink is a practical worktree strategy.

This makes external dependency symlinks especially relevant for multi-session agent
workflows, where worktrees are created and discarded frequently. Supporting this
layout, or providing a narrowly scoped allowlist for external dependency roots,
would make Turbopack usable in those workflows without requiring a separate full
installation for every session.

[...]

PR #98242, opened 2026-09-03
Turbopack: don't crash when resolving `next/package.json` errors

marcoshernanz:
### What?

When `next/package.json` can't be resolved because resolution *errors* (as opposed to the package simply not being found), Turbopack crashed with a fatal `TurbopackInternalError` asking the user to report a bug:

```
FATAL: An unexpected Turbopack error occurred. A panic log has been written to …
Error [TurbopackInternalError]: Symlink node_modules could not be resolved: the symlink target leaves the filesystem root or its parent directory could not be resolved
```

This happens e.g. when `node_modules` is a symlink whose target is outside of the workspace root — a git worktree that shares the `node_modules` of the primary checkout ([#98111](https://github.com/vercel/next.js/issues/98111)) — since Turbopack intentionally doesn't follow symlinks that leave the root.

This PR reports that error through the existing recoverable `MissingNextFolderIssue` instead, including the underlying error, so `next dev` stays up and `next build` fails with the actionable message from #93877:

```
Error: Could not find the Next.js package (next/package.json)
Resolved from: /…/worktree/app
Filesystem root used for resolution: /…/worktree
Resolution error: Symlink node_modules could not be resolved: the symlink target leaves the filesystem root or its parent directory could not be resolved
…
Possible causes:
  - …
  - node_modules/next was removed, renamed, or has a broken symlink.
  - The workspace root is incorrect — see turbopack.root in the Next.js config docs for how to configure it.
  …
https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack#root-directory
```

This does **not** make symlinks outside of the root work; that's tracked in #93556 and being implemented in #98003. It only turns the internal-error crash into the existing user-facing error.

### Why?

`try_get_next_package` only handled the "not found" case (`Ok` with no source). A resolution `Err` was propagated with `?` and no issue was emitted, so `strongly_consistent_catch_collectables` at the napi boundary had nothing to downgrade the failure to and it surfaced as an internal error — the exact failure mode #93877 fixed for the not-found case. Every other resolution in Turbopack already reports errors this way (`handle_resolve_error` → `ResolvingIssue` with the error message).

### How?

- `try_get_next_package`: match on `result.await`; on `Err`, emit `MissingNextFolderIssue` and return `None`, exactly like the not-found path. The success and not-found paths are unchanged.
- `MissingNextFolderIssue`: new `error_message: Option<RcStr>` (formatted with `PrettyPrintError`, like `ResolvingIssue`), rendered as a `Resolution error:` line. `None` for the not-found case, so that output is unchanged.
- New test `test/development/app-dir/node-modules-symlink-outside-root` (moves the installed `node_modules` out of the project and symlinks it back). Fails on `canary` with the FATAL above, passes with this change. `test/development/app-dir/concurrent-install` still passes.

Related: #98111, #93556, #98003

### DIFF
--- a/crates/next-core/src/next_import_map.rs
+++ b/crates/next-core/src/next_import_map.rs
@@ -6,7 +6,7 @@ use either::Either;
 use next_taskless::{EDGE_NODE_EXTERNALS, NODE_EXTERNALS};
 use rustc_hash::FxHashMap;
 use turbo_rcstr::{RcStr, rcstr};
-use turbo_tasks::{FxIndexMap, ResolvedVc, Vc, fxindexmap};
+use turbo_tasks::{FxIndexMap, PrettyPrintError, ResolvedVc, Vc, fxindexmap};
 use turbo_tasks_fs::{
     FileContent, FileSystem, FileSystemPath,
     glob::{Glob, GlobOptions},
@@ -1311,6 +1311,9 @@ pub async fn get_next_package(context_directory: FileSystemPath) -> Result<FileS
 struct MissingNextFolderIssue {
     path: FileSystemPath,
     root: FileSystemPath,
+    /// Set when resolving `next/package.json` failed with an error, rather than the package not
+    /// being found.
+    error_message: Option<RcStr>,
 }
 
 #[async_trait]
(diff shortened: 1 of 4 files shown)


marcoshernanz:
Closing: the underlying limitation is tracked in #93556 / #98003, and this per-call-site error reclassification isn't clearly the right layer for it (the same gap exists for other direct `resolve` consumers). Keeping the investigation notes locally.

Issue #64760, opened 2024-04-19
Unhandled Runtime Error：ChunkLoadError: Loading chunk app/layout failed.


zhangwei900808:
### Link to the code that reproduces this issue

https://github.com/zhangwei900808/next-issue-demo

### To Reproduce

1、clone project
2、pnpm install
3、pnpm run dev
4、open the page and you can see chrome console output: Uncaught SyntaxError: Invalid or unexpected token
5、same minutes you can see error:Uncaught ChunkLoadError: Loading chunk app/layout failed.
<img width="998" alt="image" src="https://github.com/vercel/next.js/assets/4385154/37de7c45-c9dc-442d-9360-e7c69d66f949">

### Current vs. Expected behavior

1、First time open the page is error but you refresh page is work!
2、I expected is work all the time 

### Provide environment information

```bash
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 23.4.0: Fri Mar 15 00:10:42 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6000
  Available memory (MB): 65536
  Available CPU cores: 10
Binaries:
  Node: 20.11.1
  npm: 10.2.4
  Yarn: N/A
  pnpm: 7.29.1
Relevant Packages:
  next: 14.2.2 // Latest available version is detected (14.2.2).
  eslint-config-next: N/A
  react: 18.2.0
  react-dom: 18.2.0
  typescript: N/A
Next.js Config:
  output: N/A
```

### Which area(s) are affected? (Select all that apply)

Not sure

### Which stage(s) are affected? (Select all that apply)

next dev (local)

zhangwei900808:
same guys has same problem?

kilimanjjjaro:
Same here!

mihal277:
I'm also experiening this issue on next.js version 14.2.3.

reynaldomarchell:
Are you have solved this problem? I just got the same problem here...

SeongYeongJu:
I'm having the same issue.

wadenarel:
![image](https://github.com/vercel/next.js/assets/139941269/9796d88b-2f17-4aa7-bfa5-4dca483b466a)

petejodo:
can also confirm that it seems to only occur when `export const runtime = "edge"` is in `app/layout.ts`. Removing that removed that error

and-sm:
I have the same error, occurs only in dev mode.

[...]

PR #98244, opened 2026-09-03
Fix spurious webpack rebuild on cold start when Tailwind content directories don't exist

### LINKED ISSUE #66526: [NEXT-3982] ChunkLoadError: Loading chunk app/layout failed. (timeout: http://localhost:3000/_next/static/chunks/app/layout.js)

https://github.com/Intellyse/next-js-bug

### To Reproduce

1. clone project
2. npm install
3. npm run dev
4. open the page and you can see chrome console output: Uncaught SyntaxError: Invalid or unexpected token
5. same minutes you can see error:Uncaught ChunkLoadError: Loading chunk app/layout failed.
image

1. First time open the page is error but you refresh page is work!
2. I expected is work all the time

```bash
Operating System:
  Platform: wls ubutnu
  Arch: arm64
  Version: wls version 2
Binaries:
  Node: 20.9.0
  npm: 10.1.0
Relevant Packages:
  next: 14.2.3 // I tried also last canary version (15.0.0-canary.10).
  eslint-config-next: 14.2.3
  react: 18.2.0
  react-dom: 18.0.0
  typescript: 5.4.5
Next.js Config:
  output: N/A
```

Webpack

next dev (local)

### Additional context

The error seems happen randomly, and we have this issue also in production. It seems related with something with webpack

```
__webpack_require__.f.j = function (chunkId, promises) {
  // JSONP chunk loading for javascript
  var installedChunkData = __webpack_require__.o(installedChunks, chunkId)
    ? installedChunks[chunkId]
    : undefined;
  if (installedChunkData !== 0) {
    // 0 means "already installed".

    // a Promise means "currently loading".
    if (installedChunkData) {
      promises.push(installedChunkData[2]);
    } else {
      if ("webpack" != chunkId) {
        // setup Promise in chunk cache
        var promise = new Promise(function (resolve, reject) {
          installedChunkData = installedChunks[chunkId] = [resolve, reject];
        });
        promises.push((installedChunkData[2] = promise));

        // start chunk loading
        var url = __webpack_require__.p + __webpack_require__.u(chunkId);
        // create error before stack unwound to get useful stacktrace later
        var error = new Error();
        var loadingEnded = function (event) {
          if (__webpack_require__.o(installedChunks, chunkId)) {
            installedChunkData = installedChunks[chunkId];
            if (installedChunkData !== 0) installedChunks[chunkId] = undefined;
            if (installedChunkData) {
              var errorType = event && (event.type === "load" ? "missing" : event.type);
              var realSrc = event && event.target && event.target.src;
              error.message =
                "Loading chunk " + chunkId + " failed.\n(" + errorType + ": " + realSrc + ")";
              error.name = "ChunkLoadError";
              error.type = errorType;
              error.request = realSrc;
              installedChunkData[1](error);
            }
          }
        };
        __webpack_require__.l(url, loadingEnded, "chunk-" + chunkId, chunkId);
      } else installedChunks[chunkId] = 0;
    }
  }
};

```
the error rise at line: var error = new Error();

marcoshernanz:
### What?

When a PostCSS plugin reports a directory that doesn't exist as a `dir-dependency` / `context-dependency`, the postcss-loader now registers it with `addMissingDependency` instead of `addContextDependency`. Existing directories are still registered as context dependencies.

Tailwind CSS v3 reports a `dir-dependency` for the base directory of every `content` glob, including ones that don't exist. The `create-next-app` Tailwind template shipped `./pages/**/*`, `./components/**/*` and `./app/**/*` (or the `./src/...` variants), so most App Router projects created with it register one or two non-existent directories on every cold start.

### Why?

webpack's watcher (watchpack) reports a *context* dependency that doesn't exist as **removed** as soon as it attaches to it, i.e. right after the first compilation that registered it. That forces a second, spurious client compilation ~20ms after `ensurePage` has already resolved, so the HTML is sent while webpack is rebuilding. ~300ms later the rebuild re-emits `static/chunks/app/layout.js` (the chunk that contains the global CSS) while the browser is downloading it. Dev chunks are served straight from disk, so the browser gets a truncated file: either `ERR_CONTENT_LENGTH_MISMATCH`, or, when `stat` observes the file mid-write, a complete-looking response with truncated JS that throws `SyntaxError: Invalid or unexpected token` and ends 120s later in `ChunkLoadError: Loading chunk app/layout failed. (timeout: …/app/layout.js)`. A reload works because the watcher state is settled by then.

Both reproductions attached to #66526 and #64760 have `content: ['./src/pages/**/*…', …]` without a `src/pages` directory, plus a large client component tree in the root layout (Clerk / antd), which makes the download slow enough to overlap with the rewrite. Timeline traced on the #64760 reproduction (`next@14.2.2`, cold `next dev`, `fs.writeFile` and webpack hooks instrumented):

```
[+12502ms] WEBPACK[client] done                                      ← ensurePage('/') resolves, render starts
[+12524ms] WEBPACK[client] invalid
[+12525ms] WEBPACK[client] watchRun modified=[] removed=["…/src/pages"]  ← phantom removal
[+13005ms] HTTP  fin GET /  200                                      ← HTML sent
[+13046ms] HTTP  req GET /_next/static/chunks/app/layout.js          ← browser starts downloading (9 MB)
[+13291ms] writeFile START static/chunks/app/layout.js               ← rebuild rewrites it → truncated response
```

watchpack already special-cases *files* listed as missing (`if (!missing.has(file)) onRemove(...)`) but not directories, so registering the directory as a missing dependency is the intended way to express "watch this path for creation".

The same sequence reproduces on 13.5.1, 13.5.2, 14.2.x and 16.3.4 (`--webpack`), so this is not a regression from #55562. Turbopack is unaffected.

### How?

Only the non-existent branch changes. A missing dependency still invalidates the compilation when the directory is created: creating `src/pages/Extra.tsx` with a new Tailwind class triggers a rebuild and the class shows up in the CSS, and later edits inside the new directory are picked up too (verified identical to the behaviour before this change).

With the change, the cold start of the reproduction runs a single client compilation (no `removed=[…]` `watchRun`) and `app/layout.js` is written once, before the HTML is sent: 0/3 truncated responses vs 2/2 before. `test/development/basic/tailwind-jit`, `test/e2e/app-dir/tailwind-css`, `test/e2e/app-dir/no-double-tailwind-execution` and `test/e2e/postcss-config-cjs` pass in webpack dev mode. The added unit test fails without the loader change.

Scope: this removes the trigger the tracked reproductions hit (and a wasted full client recompile on every cold start for these projects). It does not make chunk serving robust against any rebuild that lands while a chunk is being downloaded, e.g. an HMR edit during a slow download; that would need atomic emits or gating static chunk requests on the compiler being idle.

Fixes #66526
Closes NEXT-3982

### DIFF
--- a/packages/next/src/build/webpack/loaders/postcss-loader/src/index.ts
+++ b/packages/next/src/build/webpack/loaders/postcss-loader/src/index.ts
@@ -1,6 +1,27 @@
+import { existsSync } from 'fs'
 import Warning from './Warning'
 import SyntaxError from './Error'
 import { normalizeSourceMap, normalizeSourceMapAfterPostcss } from './utils'
+
+/**
+ * Registers a directory reported by a PostCSS plugin as a dependency of the
+ * module.
+ *
+ * A directory that doesn't exist (e.g. a Tailwind CSS `content` glob like
+ * `./pages/**` in a project without a `pages` directory) is registered as a
+ * missing dependency instead of a context dependency. Webpack still
+ * invalidates the module once the directory is created, but the watcher no
+ * longer reports the non-existent directory as removed right after the first
+ * compilation, which triggered a spurious rebuild that rewrote the emitted
+ * chunks while the browser could still be downloading them.
+ */
+function addContextDependency(loaderContext: any, directory: string): void {
+  if (existsSync(directory)) {
+    loaderContext.addContextDependency(directory)
+  } else {
+    loaderContext.addMissingDependency(directory)
+  }
+}
 
 /**
  * **PostCSS Loader**
@@ -94,10 +115,10 @@ export default async function loader(
             this.addMissingDependency(message.file)
             break
           case 'context-dependency':
-            this.addContextDependency(message.file)
+            addContextDependency(this, message.file)
             break
           case 'dir-dependency':
-            this.addContextDependency(message.dir)
+            addContextDependency(this, message.dir)
             break
           case 'asset':
             if (message.content && message.file) {
--- a/test/unit/postcss-loader.test.ts
+++ b/test/unit/postcss-loader.test.ts
@@ -1,0 +1,74 @@
+/* eslint-env jest */
+import os from 'os'
+import path from 'path'
+import postcss from 'postcss'
+import { Span } from 'next/dist/trace'
+import loader from 'next/dist/build/webpack/loaders/postcss-loader/src'
+
+const existingDir = path.resolve(os.tmpdir())
+const missingDir = path.join(
+  existingDir,
+  `next-postcss-loader-missing-${process.pid}-${Date.now()}`
+)
+
+function runLoader(messages: Array<Record<string, unknown>>) {
+  const loaderContext = {
+    resourcePath: path.join(existingDir, 'styles.css'),
+    context: existingDir,
+    sourceMap: false,
+    emitWarning: jest.fn(),
+    emitFile: jest.fn(),
+    addDependency: jest.fn(),
+    addBuildDependency: jest.fn(),
+    addMissingDependency: jest.fn(),
+    addContextDependency: jest.fn(),
+    currentTraceSpan: new Span({ name: 'test' }),
+    getOptions: () => ({
+      postcss: async () => ({
+        postcssWithPlugins: postcss([
+          {
+            postcssPlugin: 'test-messages',
+            Once(_root, { result }) {
+              result.messages.push(...(messages as any[]))
+            },
+          },
+        ]),
+      }),
+    }),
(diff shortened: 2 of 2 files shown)


packages/next/src/build/webpack/loaders/postcss-loader/src/index.ts
```diff
@@ -94,10 +115,10 @@ export default async function loader(
             this.addMissingDependency(message.file)
             break
```
bgw, line 116: Why are we not receiving a `missing-dependency` message here? I think that's the real issue.

PR #98247, opened 2026-09-03
test(turbopack): strongly read test operations

lukesandberg:
## Summary

Replace top-level consistency suppression in straightforward backend tests with local operation roots that are read strongly consistently. Inner operation functions keep each migration scoped to its test.

## Verification

- `cargo test -p turbo-tasks-backend --test filter_unused_args --test panics --test task_statistics`

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/tests/task_statistics.rs
+++ b/turbopack/crates/turbo-tasks-backend/tests/task_statistics.rs
@@ -6,7 +6,7 @@ use std::future::IntoFuture;
 
 use anyhow::Result;
 use serde_json::json;
-use turbo_tasks::{Vc, unmark_top_level_task_may_leak_eventually_consistent_state};
+use turbo_tasks::Vc;
 use turbo_tasks_testing::{Registration, register, run_without_cache_check};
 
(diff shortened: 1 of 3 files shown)


turbopack/crates/turbo-tasks-backend/tests/task_statistics.rs
```diff
@@ -97,36 +112,42 @@ async fn test_vc_receiving_task() -> Result<()> {
...
+                    "<task_statistics::WrappedU64 as dyn task_statistics::Doublable>::double": {
+                        "cache_miss": 10,
+                        "cache_hit": 15,
+                    },
+                    "<task_statistics::WrappedU64 as dyn task_statistics::Doublable>::double_vc": {
+                        "cache_miss": 10,
+                        "cache_hit": 15,
+                    },
+                })
+            );
```
bgw: Move this assertion after the operation has been read strongly consistently?

Also maybe this test file shouldn't use a strongly consistent read... It might make the number of executions less predictable? This test is a bit of a hack.

bgw (approved):

PR #98248, opened 2026-09-03
test(turbopack): add strong read operation roots

lukesandberg:
## Summary

Replace top-level consistency suppression in read-ref and trait-ref cell tests with explicit operation roots. The operations separately cover counter creation, counter reads, and trait upcasts while preserving the tests’ warm-cache and snapshot behavior.

## Verification

- `cargo test -p turbo-tasks-backend --test read_ref_cell --test trait_ref_cell`

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/tests/read_ref_cell.rs
+++ b/turbopack/crates/turbo-tasks-backend/tests/read_ref_cell.rs
@@ -5,49 +5,74 @@
 use std::{collections::HashSet, mem::take, sync::Mutex};
 
 use anyhow::Result;
-use turbo_tasks::{
-    Invalidator, ReadRef, Vc, get_invalidator,
-    unmark_top_level_task_may_leak_eventually_consistent_state, with_turbo_tasks,
-};
+use turbo_tasks::{Invalidator, ReadRef, ResolvedVc, Vc, get_invalidator, with_turbo_tasks};
 use turbo_tasks_testing::{Registration, register, run_once};
 
 static REGISTRATION: Registration = register!();
 
+#[turbo_tasks::function(operation, root)]
+fn create_counter_operation(nonce: u32) -> Vc<Counter> {
+    let _ = nonce;
+    Counter::cell(Counter {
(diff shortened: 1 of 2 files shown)


turbopack/crates/turbo-tasks-backend/tests/read_ref_cell.rs
lukesandberg, line 1: to reduce the diff you can define these new operations inline inside of run_once

that should simplify review

turbopack/crates/turbo-tasks-backend/tests/trait_ref_cell.rs
lukesandberg, line 1: ditto

turbopack/crates/turbo-tasks-backend/tests/read_ref_cell.rs
```diff
@@ -5,49 +5,74 @@
...
+
+            let counter_value = counter.get_value();
+
+            assert_eq!(*counter.get_value().strongly_consistent().await?, 0);
+            assert_eq!(*counter_value.strongly_consistent().await?, 0);
+
+            read_counter_operation(counter)
+                .read_strongly_consistent()
+                .await?
+                .incr();
```
sokra: I don't see the benefit of doing `read_counter_operation` here. This operations does nothing. It takes a resolved vc and returns it. There is no point in doing a strongly consistent read on something that has no dependencies. You can just do `counter.await?.incr()`.

sokra: Is this wrapping needed to remove `unmark_top_level_task_may_leak_eventually_consistent_state`?

lukesandberg: yeah exactly. we need an operation to do a strongly consistent read and all top level reads must be strongly consistent.

it would be better here to actually wrap the whole test body in one operation instead of a bunch of little ones

bgw (approved):

PR #98249, opened 2026-09-03
[PPF] Instant validation for requireStatic = "shell" | "prefetch"

lubieowoce:
Adjusts Instant Validation internals to validate the `ShellStatic` stage instead of `ShellRuntime` when `requireStatic >= "shell"`. If that's the case, the client will do a static request (and extract the shell from it), so the app shell cannot access session data (unlike the default, where we use a runtime shell).

Note that in many cases, this is technically more restrictive than needed. For `requireStatic >= "shell"`, we'll do a static prefetch that goes up to the `Static` stage which resolves more things, like static `params`, `prefetch()` ( in the `PrefetchStatic` stage) and `navigation()` (in the `NavigationStatic` stage). However:
1. The client expects to be able to rewind the static prefetch into a shell, so we should be validating that the shell is actually useful. 
2. For `requireStatic = "shell"`, we might still do a *runtime* prefetch (with `<Link prefetch={true}>`), which will not have `navigation()`, and validating `ShellStatic` catches that.

---


test/development/app-dir/cache-components-dev-streaming/cache-components-dev-streaming.test.ts
lubieowoce: neither of the changed tests here work like they should despite `revealStage` being `ShellStatic` and `ShellRuntime` respectively. it seems like the `_revealAfter` mechanism has no effect here, and the promise is already resolved by the time we get to the client.

i think this is because `_revealAfter` can prevent us from revealing content too early, but it doesn't actually delay the later stages from streaming in. in both cases, session data and link data show up in a task or two, so the client gets both, and has no need to show the fallback. i don't consider this a blocker for this PR but we might want to look into this separately at some point

packages/next/src/server/app-render/instant-validation/instant-validation.tsx
```diff
@@ -206,7 +208,16 @@ export async function collectStagedSegmentData(
 
   let partialStages: SegmentStage[]
   switch (prefetchKind) {
-    case ValidationPrefetchKind.Shell: {
+    case ValidationPrefetchKind.StaticAppShell: {
+      partialStages = [
+        RenderStage.ShellStatic,
+        RenderStage.Static,
+        RenderStage.NavigationStatic, // TODO(cache-stages): only if needed
```
unstubbable: @lubieowoce, could you clarify the intent behind putting `Static` before `NavigationStatic` here? I expected `Static` to already include navigation-stage data, which seems to prevent the retry from identifying `unstable_navigation()` specifically. Replacing `Static` with `PrefetchStatic` in both stage lists would change the IV messages in four dev snapshots from “URL data” to “unstable_navigation()”. Is the generic “URL data” message intentional for `ensureStatic="shell"` and `"prefetch"`?

test/development/app-dir/cache-components-dev-streaming/cache-components-dev-streaming.test.ts
unstubbable: Yeah, `_revealAfter` is not about delaying later stages, it's only about preventing flashing fallbacks from the earlier stages.

packages/next/src/server/app-render/instant-validation/instant-validation.tsx
lubieowoce: oh i must've not had my coffee when i wrote that, damn

lubieowoce: yeah this order makes no sense, it's gotta be monotonically increasing

test/e2e/app-dir/instant-validation/head-and-reporting.util.ts
```diff
@@ -5,9 +5,31 @@ import {
...
+import { ErrorSnapshot, RedboxSnapshot } from '../../../lib/add-redbox-matchers'
 
 const partialPrefetching = !!process.env.__NEXT_PARTIAL_PREFETCHING
 
+const INSTANT_INSIGHT_PATTERNS = {
+  urlData: /Next\.js encountered URL data/,
+  navigation: /Next\.js encountered `?unstable_navigation\(\)`?/,
+  runtimeData: /Next\.js encountered runtime data/,
+  uncachedData: /Next\.js encountered uncached data/,
+}
```
lubieowoce: after committing some embarrasingly mismatched snapshots where Instant Validation reported one thing in dev and another in build, i decided to make the new `ensureStatic` tests explicitly state which message they're expecting to see, and check that it's the same across dev/build. really, we should make the whole test suite do this, because it's too easy to commit a bad snapshot, but i'm gonna leave that for a follow up

unstubbable (approved):

PR #98252, opened 2026-09-04
fix(typegen): merge cacheLife types into next/cache instead of shadowing it

### LINKED ISSUE #98251: Every next/cache export is offered twice in autocomplete after next build

https://github.com/m-elagamy/nextjs-cache-autocomplete-duplication

### To Reproduce

Follow-up to #77823, which was closed. The closing analysis was right about the part I got
wrong: the .next size growth is ordinary build cache (.next/cache), not duplicated application
code, and no server bundle contains duplicated next/cache code. I withdraw that part entirely.

The closing comment also said:

> With the reported Next.js and TypeScript versions, the TypeScript server exposes only one
> completion for each next/cache export before and after the build.

and asked for a new issue if the duplicate completions could be reproduced. This is that
reproduction, on Next.js 16.3.4 / TypeScript 5.9.3.

Steps:

1. git clone https://github.com/m-elagamy/nextjs-cache-autocomplete-duplication.git
2. cd nextjs-cache-autocomplete-duplication
3. pnpm install
4. pnpm run build
5. node repro-autocomplete.js

`repro-autocomplete.js` needs no IDE. It drives the same TypeScript LanguageService API that
tsserver — and therefore VS Code IntelliSense — uses, against the project's own tsconfig.json
and node_modules. It requests completions on a bare identifier with
`includeCompletionsForModuleExports: true` (what the editor sends), resolves each entry to the
import it would insert, and runs twice: once with `.next/types/cache-life.d.ts` in the program
and once without, so the cause is isolated rather than just observed.

It exits 1 when duplicates are present.

To see it in an editor instead: open any file in the project after building and type
`revalidatePath`. Two identical suggestions appear, both labelled `Add import from "next/cache"`.

Expected: each next/cache export is offered once in autocomplete.

Actual: after a build, every export is offered twice, and both entries are labelled identically
as `Add import from "next/cache"`. There is no way to tell them apart in the completion list.

Output of node repro-autocomplete.js:

  export              before build   after build
  ------------------------------------------------
  revalidatePath      1             2   <-- DUPLICATED
  revalidateTag       1             2   <-- DUPLICATED
  updateTag           1             2   <-- DUPLICATED
  refresh             1             2   <-- DUPLICATED
  unstable_cache      1             2   <-- DUPLICATED
  unstable_noStore    1             2   <-- DUPLICATED
  cacheTag            1             2   <-- DUPLICATED
  cacheLife           2             3   <-- DUPLICATED
  unstable_cacheTag   1             2   <-- DUPLICATED
  unstable_cacheLife  1             2   <-- DUPLICATED

  FAIL: 10/10 next/cache exports gained a duplicate completion after the build.

Two notes:

- The reproduction does not import next/cache anywhere. These are auto-import suggestions, so
  the duplication appears in any file in the project.
- cacheLife starts at 2 even before the build. That is a separate, pre-existing issue: the
  internal path next/dist/server/use-cache/cache-life also exports cacheLife, so the editor
  offers a deep internal import alongside the public one. Happy to split that out.

```bash
Operating System:
  Platform: win32
  Arch: x64
  Version: Windows 11 Pro
  Available memory (MB): 32603
  Available CPU cores: 12
Binaries:
  Node: 24.12.0
  npm: 11.6.2
  Yarn: N/A
  pnpm: 11.1.3
Relevant Packages:
  next: 16.3.4 // Latest available version is detected (16.3.4).
  eslint-config-next: N/A
  react: 19.2.8
  react-dom: 19.2.8
  typescript: 5.9.3
Next.js Config:
  output: N/A
```

TypeScript

### Additional context

Root cause

.next/types/cache-life.d.ts has no top-level import or export, so it is a script, not a module.
That makes its `declare module 'next/cache' { ... }` a standalone ambient module declaration
rather than a module augmentation — it shadows node_modules/next/cache.d.ts instead of merging
with it, which is why it has to re-export every other next/cache member.

Ambient declarations win module resolution, so imports resolve to the generated file only:

  before build:  "next/cache" -> node_modules/next/cache.d.ts
  after build:   "next/cache" -> .next/types/cache-life.d.ts   (package types shadowed)

But TypeScript's auto-import index still indexes node_modules/next/cache.d.ts as a module
exporting those names. Both modules are therefore offered under the specifier next/cache,
producing two identical entries per export.

The file is written by next build, next dev and next typegen, so this is not build-specific:

- next/dist/build/index.js  (distDir/types/cache-life.d.ts)
- next/dist/server/lib/router-utils/setup-dev-bundler.js
- next/dist/cli/next-typegen.js

One measured caveat, in case it saves someone time: turning the generated file into a module
augmentation does remove the duplication, but on its own it also silently drops the profile-name
validation — the catch-all `export function cacheLife(profile: string): void` in next/cache.d.ts
merges back in and any string is accepted again. So the shadowing appears to be load-bearing
rather than accidental.

Happy to help with a fix if that would be useful.

itsybitsci:
### What?

After `next build`, `next dev`, or `next typegen`, every `next/cache` export is offered twice in editor autocomplete. Both entries are labelled identically as `Add import from "next/cache"`, so there is no way to tell them apart. It affects any file in the project, since these are auto-import suggestions.
### Why?
`.next/types/cache-life.d.ts` has no top-level import or export. That makes it a script rather than a module, which in
turn makes its `declare module 'next/cache'` a standalone ambient module declaration instead of a moduleaugmentation. So it shadows the package's own types rather than merging with them, and that is why the generated file
has to redeclare every unrelated `next/cache` export just to keep them resolvable.
Ambient declarations win module resolution, so after a build imports resolve to the generated file. But TypeScript'sauto-import index still indexes `node_modules/next/cache.d.ts` as a module exporting those same names. Both are then
offered under the `next/cache` specifier, producing two identical completions per export.
### How?
Emit a top-level `export {}` in the generated file. That makes it a module, so the `declare module 'next/cache'` blockbelow becomes an augmentation and merges into the package's types. The redeclarations of `revalidatePath`,
`revalidateTag`, `unstable_cache`, `cacheTag` and the rest were only needed while the file was shadowing, so they areremoved.

Merging also brings back the package's catch-all `cacheLife(profile: string)` overload from `next/cache.d.ts`, which the shadowing file used to hide. On its own that would make `cacheLife('bogus')` compile, where today it is a type error (thanks to @m-elagamy for measuring this).

To keep that check, `next/cache` now declares an empty `CacheLifeProfiles` interface, and the generated file augments it with one key per configured profile alongside the per-profile overloads. The catch-all's parameter is `keyof CacheLifeProfiles extends never ? string : never`, so it accepts any string until the types have been generated and nothing once they have. Configured names, the built-in profiles and the object form still resolve to their own overloads, with their documentation. This is the same shape as the draft in vercel-labs/next.js#199.

Tests: `test/e2e/cache-life-typecheck` builds an app with a custom `blog` profile, waits for `cache-life.d.ts` to be written, and runs `tsc --noEmit` against it. The validation file calls `cacheLife` with the custom profile, a built-in profile, the object form and the `unstable_` alias, and has `@ts-expect-error` on two unknown names, so it fails if either the generated overloads or the unknown-name check stop working. A shadowing file would also fail it, since the catch-all and the object form would no longer be visible. The existing inline snapshots in `cache-life-type-utils.test.ts` cover the generated file's shape. The LanguageService unit test from the first revision is removed per review.

Fixes #98251

### DIFF
--- a/test/e2e/cache-life-typecheck/typecheck-validation.ts
+++ b/test/e2e/cache-life-typecheck/typecheck-validation.ts
@@ -1,0 +1,13 @@
+import { cacheLife, unstable_cacheLife } from 'next/cache'
+
+export function validate() {
+  cacheLife('blog')
+  cacheLife('minutes')
+  cacheLife({ stale: 30, revalidate: 60, expire: 120 })
+  unstable_cacheLife('blog')
+
+  // @ts-expect-error unknown profile
+  cacheLife('bogus')
+  // @ts-expect-error unknown profile
+  unstable_cacheLife('bogus')
+}
--- a/packages/next/cache.d.ts
+++ b/packages/next/cache.d.ts
@@ -20,6 +20,8 @@ import { cacheTag } from 'next/dist/server/use-cache/cache-tag'
 
 export { cacheTag }
 
+export interface CacheLifeProfiles {}
+
 /**
  * Cache this `"use cache"` for a timespan defined by the `"default"` profile.
  * ```
@@ -128,7 +130,9 @@ export function cacheLife(profile: 'max'): void
  *
  * You can define custom profiles in `next.config.ts`.
  */
-export function cacheLife(profile: string): void
+export function cacheLife(
+  profile: keyof CacheLifeProfiles extends never ? string : never
+): void
 
 /**
  * Cache this `"use cache"` using a custom timespan.
--- a/packages/next/src/server/lib/router-utils/cache-life-type-utils.test.ts
+++ b/packages/next/src/server/lib/router-utils/cache-life-type-utils.test.ts
@@ -20,19 +20,13 @@ describe('cache-life-type-utils', () => {
     expect(output).toMatchInlineSnapshot(`
      "// Type definitions for Next.js cacheLife configs
 
-     declare module 'next/cache' {
-       export { unstable_cache } from 'next/dist/server/web/spec-extension/unstable-cache'
-       export {
-         updateTag,
-         revalidateTag,
-         revalidatePath,
-         refresh,
-       } from 'next/dist/server/web/spec-extension/revalidate'
-       export { unstable_noStore } from 'next/dist/server/web/spec-extension/unstable-no-store'
-       export { io } from 'next/dist/server/request/io'
-       export { unstable_navigation } from 'next/dist/server/request/cache-stages'
-       export { unstable_prefetch } from 'next/dist/server/request/cache-stages'
+     export {}
 
+     declare module 'next/cache' {
+       export interface CacheLifeProfiles {
+         "default": true
+         "hours": true
+       }
        
          /**
           * Cache this \`"use cache"\` for a timespan defined by the \`"default"\` profile.
@@ -62,39 +56,6 @@ describe('cache-life-type-utils', () => {
           */
          export function cacheLife(profile: "hours"): void
          
-         /**
-          * Cache this \`"use cache"\` using a custom timespan.
-          * \`\`\`
-          *   stale: ... // seconds
-          *   revalidate: ... // seconds
-          *   expire: ... // seconds
-          * \`\`\`
-          *
-          * This is similar to Cache-Control: max-age=\`stale\`,s-max-age=\`revalidate\`,stale-while-revalidate=\`expire-revalidate\`
-          *
-          * If a value is left out, the lowest of other cacheLife() calls or the default, is used instead.
-          */
-         export function cacheLife(profile: {
-           /**
-            * This cache may be stale on clients for ... seconds before checking with the server.
-            */
-           stale?: number,
-           /**
-            * If the server receives a new request after ... seconds, start revalidating new values in the background.
-            */
-           revalidate?: number,
-           /**
-            * If this entry has no traffic for ... seconds it will expire. The next request will recompute it.
-            */
-           expire?: number
-         }): void
-       
-
-       import { cacheTag } from 'next/dist/server/use-cache/cache-tag'
-       export { cacheTag }
-
-       export const unstable_cacheTag: typeof cacheTag
-       export const unstable_cacheLife: typeof cacheLife
      }
      "
     `)
@@ -159,10 +120,7 @@ describe('cache-life-type-utils', () => {
           *   expire:     1814400 seconds (3 weeks)
           *   stale:      2592000 seconds (1 month)
           *   revalidate: 5184000 seconds (2 months)
-          *   expire:     10368000 seconds (4 months)
-          *   stale: ... // seconds
-          *   revalidate: ... // seconds
-          *   expire: ... // seconds"
+          *   expire:     10368000 seconds (4 months)"
     `)
   })
 
@@ -180,19 +138,12 @@ describe('cache-life-type-utils', () => {
     expect(output).toMatchInlineSnapshot(`
      "// Type definitions for Next.js cacheLife configs
 
-     declare module 'next/cache' {
-       export { unstable_cache } from 'next/dist/server/web/spec-extension/unstable-cache'
-       export {
-         updateTag,
-         revalidateTag,
-         revalidatePath,
-         refresh,
-       } from 'next/dist/server/web/spec-extension/revalidate'
-       export { unstable_noStore } from 'next/dist/server/web/spec-extension/unstable-no-store'
-       export { io } from 'next/dist/server/request/io'
-       export { unstable_navigation } from 'next/dist/server/request/cache-stages'
-       export { unstable_prefetch } from 'next/dist/server/request/cache-stages'
+     export {}
 
+     declare module 'next/cache' {
+       export interface CacheLifeProfiles {
+         "partial": true
+       }
        
          /**
           * Cache this \`"use cache"\` for a timespan defined by the \`"partial"\` profile.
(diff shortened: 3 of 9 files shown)


test/unit/cache-life-types.test.ts
eps1lon, line 1: This is a bit too close to implementation details. We should have existing tests performing actual type-checking. Those would suffice.

m-elagamy:
Thanks for picking this up so quickly.

One thing worth checking before this lands. Making the generated declaration an augmentation does remove the duplicate completions, but packages/next/cache.d.ts still declares the catch-all:

export function cacheLife(profile: string): void

While the generated file shadows next/cache, that overload is out of the picture and an unknown profile name is an error. Once it merges instead, the overload merges back in and any string is accepted again — so cacheLife("mintues") would compile silently, where today it errors.

Measured on next@16.3.4 / typescript@5.9.3, driving the LanguageService directly and swapping only the generated file and the package types in memory:

variant                       revalidateTag  "minutes"   "bogus-profile"
-------------------------------------------------------------------------
ambient (Next.js today)       2 (dup)        accepted    rejected
augmentation (this PR)        1              accepted    ACCEPTED   <-- regression
augmentation + no catch-all   1              accepted    rejected

Script here if useful: https://github.com/m-elagamy/nextjs-cache-autocomplete-duplication/blob/master/verify-fix-tradeoff.js

So the shadowing looks load-bearing rather than accidental, which is probably why it was written that way — it is what makes profile names checkable at all.

Dropping the catch-all overload restores the check, but then cacheLife("minutes") errors in a project that has never run dev / build / typegen. The draft in vercel-labs/next.js#199 avoids that by declaring an augmentable CacheLifeProfiles interface in the package and resolving the catch-all parameter to never once profiles are declared, which keeps both properties at once.

Happy to verify whichever direction you go with.

test/unit/cache-life-types.test.ts
itsybitsci: Agreed, removed. Replaced with `test/e2e/cache-life-typecheck`, which runs `tsc --noEmit` on an app with a custom profile after typegen, with `@ts-expect-error` on unknown names.

itsybitsci:
> Thanks for picking this up so quickly.
> 
> One thing worth checking before this lands. Making the generated declaration an augmentation does remove the duplicate completions, but packages/next/cache.d.ts still declares the catch-all:
> 
> export function cacheLife(profile: string): void
> 
> While the generated file shadows next/cache, that overload is out of the picture and an unknown profile name is an error. Once it merges instead, the overload merges back in and any string is accepted again — so cacheLife("mintues") would compile silently, where today it errors.
> 
> Measured on next@16.3.4 / typescript@5.9.3, driving the LanguageService directly and swapping only the generated file and the package types in memory:
> 
> ## variant revalidateTag "minutes" "bogus-profile"
> ambient (Next.js today) 2 (dup) accepted rejected augmentation (this PR) 1 accepted ACCEPTED <-- regression augmentation + no catch-all 1 accepted rejected
> 
> Script here if useful: https://github.com/m-elagamy/nextjs-cache-autocomplete-duplication/blob/master/verify-fix-tradeoff.js
> 
> So the shadowing looks load-bearing rather than accidental, which is probably why it was written that way — it is what makes profile names checkable at all.
> 
> Dropping the catch-all overload restores the check, but then cacheLife("minutes") errors in a project that has never run dev / build / typegen. The draft in [vercel-labs#199](https://github.com/vercel-labs/next.js/pull/199) avoids that by declaring an augmentable CacheLifeProfiles interface in the package and resolving the catch-all parameter to never once profiles are declared, which keeps both properties at once.
> 
> Happy to verify whichever direction you go with.

Thanks, that's a real regression and your script made it easy to confirm. I went with the interface approach from the vercel-labs draft: `next/cache` now exports an empty `CacheLifeProfiles` interface, the generated file augments it with the configured names, and the catch-all's parameter is `keyof CacheLifeProfiles extends never ? string : never`.

With generated types, `cacheLife('bogus')` is an error again while configured names, built-ins and the object form still work. Without generated types any string is still accepted, so custom names don't break before the first build.

One correction on your table's last row: dropping the catch-all wouldn't make `cacheLife('minutes')` fail, since the built-in profiles have their own overloads in the package; only custom names would.

The e2e test now guards both directions with `@ts-expect-error` lines.

m-elagamy:
You're right, thanks for the correction — I re-checked and next/cache.d.ts declares
all seven built-ins as their own overloads, so dropping the catch-all only ever put
custom profile names at risk, not the built-ins. My table had that right; the sentence
under it did not.

The conditional parameter type is a better answer than either direction I measured,
since it keeps custom names working before the first typegen too.

Happy to run the harness against your branch if a second confirmation is useful.

itsybitsci:
> You're right, thanks for the correction — I re-checked and next/cache.d.ts declares all seven built-ins as their own overloads, so dropping the catch-all only ever put custom profile names at risk, not the built-ins. My table had that right; the sentence under it did not.
> 
> The conditional parameter type is a better answer than either direction I measured, since it keeps custom names working before the first typegen too.
> 
> Happy to run the harness against your branch if a second confirmation is useful.

Thanks for re-checking. A second confirmation would be great, yes. The branch is [`fix/cache-life-types-augmentation`](https://github.com/itsybitsci/next.js/tree/fix/cache-life-types-augmentation) on my fork, head commit 2196b65. The cases worth covering are the ones from your table: built-in names, a custom profile from `next.config`, and a bogus name, ideally both before the first typegen run and after. If anything comes out different from what `test/e2e/cache-life-typecheck` asserts, I'd like to know.

m-elagamy:
Verified against 2196b65 using the PR’s actual type generator and cache.d.ts, with Next.js 16.3.4 and TypeScript 5.9.3. All 65 assertions passed.
- All seven built-ins, the custom blog profile, and the object form work before and after type generation, through both cacheLife and unstable_cacheLife.
- Unknown names (bogus, mintues) are accepted before type generation and rejected afterward.
- All ten checked cache exports produce exactly one next/cache auto-import suggestion in both cases.
- The PR’s typecheck validation fixture also passes.
This verification used TypeScript LanguageService with in-memory overlays, rather than a full build of the fork. The results confirm the intended type-checking and autocomplete behavior.

test/e2e/cache-life-typecheck/typecheck-validation.ts
```diff
@@ -0,0 +1,13 @@
+import { cacheLife, unstable_cacheLife } from 'next/cache'
+
+export function validate() {
+  cacheLife('blog')
+  cacheLife('minutes')
+  cacheLife({ stale: 30, revalidate: 60, expire: 120 })
+  unstable_cacheLife('blog')
+
+  // @ts-expect-error unknown profile
+  cacheLife('bogus')
```
eps1lon, line 10: we don't error on unknown profiles if I remember correctly. Can you check if we really don't have existing tests (even if manual) for configured cache profiles?

itsybitsci: No existing test type-checks configured profiles. The closest are the generator unit test (asserts on generated text, doesn't run `tsc`) and `use-cache-invalid-cache-life` (runtime validation of config values). Hence the new `test/e2e/cache-life-typecheck`.

Unknown profiles do error today, in both places:

- **Types:** the generated `cache-life.d.ts` on canary is a bare `declare module 'next/cache'`, an ambient declaration that shadows the package's `cache.d.ts`, so after typegen `cacheLife('bogus')` fails to compile. 
- **Runtime:** `cacheLife()` throws `Unknown cacheLife() profile "bogus" is not configured in next.config.js`.

That shadowing is also what caused the duplicate auto-imports. The conditional parameter type keeps rejecting unknown names while merging instead of shadowing; before the first typegen any string is still accepted, as today.

If you'd rather not enforce profile names at the type level, I can drop the conditional type and `CacheLifeProfiles` and make it a plain augmentation. Happy either way.

eps1lon: Makes sense. Thanks for the clarification.

eps1lon (approved):

PR #92135, opened 2026-03-31
Restart the `next dev` server if `.next` is deleted

bgw:
This PR is AI generated.

If this directory is deleted while the dev server is running, we can't reasonably recover, so we should restart the process.

Addresses https://vercel.slack.com/archives/C03KAR5DCKC/p1774757692640999

PR #98253, opened 2026-09-04
Dev server: More aggresively scold users for deleting .next when we detect it while dev is running

bgw:
Feedback: https://vercel.slack.com/archives/C03KAR5DCKC/p1788491653257349?thread_ts=1788490509.032699&cid=C03KAR5DCKC

Message was added in https://github.com/vercel/next.js/pull/92135

We can't detect this in most other cases where users (or agents tbh) delete `.next`, but in this case we can.

### DIFF
--- a/packages/next/src/server/lib/start-server.ts
+++ b/packages/next/src/server/lib/start-server.ts
@@ -600,6 +600,8 @@ export async function startServer(
       if (dirWatchPaths.includes(removedPath)) {
         Log.error(
           `The directory at "${removedPath}" was deleted.\n\n` +
+            'Deleting this directory removes caches and will cause slower ' +
+            'performance.\n\n' +
             'Deleting this directory while Next.js is running can lead to ' +
             'undefined behavior. Restarting the server to recover...'
         )


packages/next/src/server/lib/start-server.ts
```diff
@@ -600,6 +600,8 @@ export async function startServer(
       if (dirWatchPaths.includes(removedPath)) {
         Log.error(
           `The directory at "${removedPath}" was deleted.\n\n` +
+            'Deleting this directory removes caches and will cause slower ' +
+            'performance.\n\n' +
             'Deleting this directory while Next.js is running can lead to ' +
             'undefined behavior. Restarting the server to recover...'
         )
```
lukesandberg, line 607: nit: i think these clauses should be combined?

Deleting this directory while Next.js is running can lead to undefined behavior in this session and slow down the next session because caches have been removed.

lukesandberg (approved):

packages/next/src/server/lib/start-server.ts
bgw: I think it's a little confusing because we'd follow that by a sentence saying "Restarting the server to recover" which might make it sound like that would solve the performance issue.

lukesandberg (approved):

PR #98255, opened 2026-09-04
fix: track dynamic accesses in final runtime prerenders

lubieowoce:
The fix implemented in #95665 (`params/searchParams` not hanging correctly in Runtime App Shells during prospective prerenders) was incomplete. We started using `makeHanging[Search]Params` in the prospective prerender (if it was an app shell and wouldn't resolve them), but not during the final prerender. This meant that, when encoding the cache key, we would only fire `dynamicAccessAbortSignal` in the prospective prerender:

https://github.com/vercel/next.js/blob/090f1b72753b2d0094235d27d46ffa0da5fc7d35/packages/next/src/server/use-cache/use-cache-wrapper.ts#L2160-L2174

...but the final prerender wasn't using `makeHanging[Search]Params`, so it would not hit this early exit. The hanging input abort signal would still correctly abort the key encoding (because it accounts for `stageController.finalStage`) so the key would still be partial, as expected. We would continue on, and attempt to look the key up in the Resume Data Cache. But the prospective prerender exited early and never put anything under that key, so we would incorrectly end up here, and log:

https://github.com/vercel/next.js/blob/090f1b72753b2d0094235d27d46ffa0da5fc7d35/packages/next/src/server/use-cache/use-cache-wrapper.ts#L2730-L2734

This is fixed by consistently using `makeHanging[Search]Params` if the runtime prerender won't reach the stage where they resolve, i.e. if it's an App Shell. This means we'll trigger `dynamicAccessAbortSignal` while encoding the cache key in *both* renders, so we'll do the early exit consistently and avoid the log.

(This feels a bit unsatisfying, because ideally this would all just work based on stages, which is what I idealistically thought when writing the code originally. But unfortunately our usage of `dynamicAccessAbortSignal` for `params/searchParams` requires these special cases to work consistently. Maybe this can be improved in the future.)

### Tests

This scenario is covered by the existing `prefetch-runtime` tests -- i discovered the bug by running them and seeing the `Unexpected cache miss after cache warming phase` warning. We now assert that the warning is not logged. The assertion is negative and would also pass if the error message changed, so i've exported the message as a constant to ensure `use-cache-wrapper` and the tests stay in sync.

### DIFF
--- a/packages/next/src/server/request/search-params.ts
+++ b/packages/next/src/server/request/search-params.ts
@@ -244,8 +244,6 @@ function createRuntimePrerenderSearchParams(
       ? createVaryingSearchParams(varyParamsAccumulator, underlyingSearchParams)
       : underlyingSearchParams
 
-  const result = makeUntrackedSearchParams(userspaceSearchParams)
-
   const searchParamsStage = RENDER_STAGES_BY_DATA_KIND.runtimeLinkData
 
   const { stagedRendering } = workUnitStore
@@ -255,20 +253,23 @@ function createRuntimePrerenderSearchParams(
     if (workUnitStore.finalStage < searchParamsStage) {
       return makeHangingSearchParams(workStore, workUnitStore)
     } else {
-      return result
+      return makeUntrackedSearchParams(userspaceSearchParams)
     }
   }
 
-  // Unlike `createRuntimePrerenderParams`, which uses `delayUntilStage`, we
-  // resolve with `waitForStage(...).then(...)` here. Switching search params to
-  // `delayUntilStage` drops the source code frame from the instant-validation
-  // "URL data outside of Suspense" error when a page awaits `searchParams` at
-  // the top level (params, read via a nested component, is unaffected). See the
-  // `missing suspense around search params` cases in the instant-validation
-  // `suspense-boundaries` tests. The underlying reason in React's async I/O
-  // await tracking isn't understood yet. TODO: align search params with params
-  // on `delayUntilStage` once resolved.
-  return stagedRendering.waitForStage(searchParamsStage).then(() => result)
+  // If search params don't resolve in this prerender, caches need to treat them as a hanging input.
+  if (
+    stagedRendering.finalStage &&
+    stagedRendering.finalStage < searchParamsStage
+  ) {
+    return makeHangingSearchParams(workStore, workUnitStore)
+  } else {
+    return stagedRendering.delayUntilStage(
+      searchParamsStage,
+      'searchParams',
+      userspaceSearchParams
+    )
+  }
 }
 
 function createRenderSearchParams(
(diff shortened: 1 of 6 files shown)


packages/next/src/server/request/search-params.ts
```diff
@@ -255,20 +253,23 @@ function createRuntimePrerenderSearchParams(
...
 
-  // Unlike `createRuntimePrerenderParams`, which uses `delayUntilStage`, we
-  // resolve with `waitForStage(...).then(...)` here. Switching search params to
-  // `delayUntilStage` drops the source code frame from the instant-validation
-  // "URL data outside of Suspense" error when a page awaits `searchParams` at
-  // the top level (params, read via a nested component, is unaffected). See the
-  // `missing suspense around search params` cases in the instant-validation
-  // `suspense-boundaries` tests. The underlying reason in React's async I/O
-  // await tracking isn't understood yet. TODO: align search params with params
-  // on `delayUntilStage` once resolved.
```
lubieowoce, line 270: unless i'm really missing something, this comment about instant validation stacks seems like a bad copy-paste or robot hallucination (added [here](https://github.com/vercel/next.js/pull/95665/changes/0d0e5d098b0658fe9bccd19da2c901430a214710)). the explanation makes no sense to me, because `createRuntimePrerenderSearchParams` is not used in any codepath related to instant validation, which uses a `request` store.

unstubbable: This is indeed very weird. The test failure is [here](https://github.com/vercel/next.js/actions/runs/29060812030/job/86262597133), but when I check out commit 083b104613b today, I can't reproduce it locally anymore with `pnpm test-dev-experimental-turbo test/e2e/app-dir/instant-validation/suspense-boundaries.partial-prefetching.test.ts -t "static prefetch - missing suspense around search params"`. That command was still in my shell history, so it must have failed back then. Not sure what my agent did differently to come to this conclusion, but it did fix the CI test failure. Maybe some kind of test interference with a different test that uses runtime prefetching or cached navs, and that poisoned the IO tracking somehow? 🤷

test/e2e/app-dir/segment-cache/prefetch-runtime/prefetch-runtime.test.ts
```diff
@@ -29,6 +30,16 @@ describe('runtime prefetching', () => {
     currentCliOutputIndex = next.cliOutput.length
   }
 
+  // We never expect to see this logged here.
+  afterEach(() => {
+    'Unexpected cache miss after cache warming phase during prerendering.'
```
unstubbable: Did you leave this intentionally here? Is this an expression statement or a directive or both?

unstubbable (approved):

test/e2e/app-dir/segment-cache/prefetch-runtime/prefetch-runtime.test.ts
lubieowoce: yeah this is a new directive we're working on, `"unexpected cache"`. sorry for leaking, i'll get rid of it quickly before anyone notices

packages/next/src/server/request/search-params.ts
lubieowoce: > Maybe some kind of test interference with a different test that uses runtime prefetching or cached navs, and that poisoned the IO tracking somehow? 

that's a terrifying possibility, but these are dev tests and i don't think we do `prerender-runtime` anywhere during dev, except maybe for navs inspector?

PR #97867, opened 2026-08-25
Turbopack: improve Pat::Assign modelling in analyzer

mischnic:
The value of the argument is actually `Argument | Default value`. Previously, the `take_pat_value` would take it and the recursive `visit_pat` would not see it anymore.

PR #97697, opened 2026-08-21
Turbopack: correctly trace through TypeScript `__importStar`

mischnic:
It was just deopting too early, we don't even need to special case this

No measurable perf impact
before
```
commit 7d29e358c8ac383a3019d552fd54f4d05645bf39 (HEAD, tag: v16.4.0-canary.5)
    v16.4.0-canary.5

pnpm next build --experimental-build-mode=compil  400.89s user 51.96s system 747% cpu 1:00.61 total
pnpm next build --experimental-build-mode=compil  394.99s user 54.10s system 757% cpu 59.289 total
pnpm next build --experimental-build-mode=compil  399.87s user 58.71s system 772% cpu 59.398 total
```
after:
```
pnpm next build --experimental-build-mode=compil  396.48s user 58.23s system 760% cpu 59.821 total
pnpm next build --experimental-build-mode=compil  403.07s user 54.17s system 778% cpu 58.716 total
pnpm next build --experimental-build-mode=compil  398.27s user 57.03s system 738% cpu 1:01.61 total
```

PR #98258, opened 2026-09-04
Turbopack: fix require handling with alternative

mischnic:
Regression from 
- https://github.com/vercel/next.js/pull/97867 (which uncovered this bug in this particular case with the default parameter assignment)
and
- https://github.com/vercel/next.js/pull/97697 (because previously `unknown | WellKnownFunction` was just unknown. Now it stays `unknown | WellKnownFunction`)

Fixes
```
./node_modules/.pnpm/@vue+compiler-sfc@3.5.26/node_modules/@vue/compiler-sfc/dist/compiler-sfc.esm-browser.js:45756:20
Error: Module not found: Can't resolve 'less'
  45754 | );
  45755 | const less = (source, map, options, load = require) => {
> 45756 |   const nodeLess = load("less");
        |                    ^^^^^^^^^^^^
```

because basically
```js
const load = globalThis.foobar || require
const x = load("less");
```
would get replaced to
```js
const load = globalThis.foobar || require
const x = __turbopack_context__.require(191912); 
```
which would be wrong if `globalThis.foobar` isn't actually `require` at runtime.

Instead, if we are inside an alternative, don't emit a codegen replacement but only add reference


mischnic:
* **#98258** <a href="https://app.graphite.com/github/pr/vercel/next.js/98258?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/98258?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

turbopack/crates/turbopack-ecmascript/src/code_gen.rs
```diff
@@ -267,6 +267,7 @@ impl CodeGens {
 }
 
 pub trait IntoCodeGenReference {
+    fn into_reference(self) -> ResolvedVc<Box<dyn ModuleReference>>;
```
sokra, line 270: This looks like you actually want this:


```suggestion
pub trait CodeGenReference : ModuleReference {
```

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
sokra, line 1736: This has a lot of arguments. Adding a boolean argument could be confusing. Maybe you want to use an enum instead `WellKnownFunctionCallContext::Alternative | Primary`

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -269,10 +269,19 @@ impl AnalyzeEcmascriptModuleResultBuilder {
...
-        self.references.insert(reference);
-        self.add_code_gen(code_gen);
+    pub fn add_reference_code_gen<R: IntoCodeGenReference>(
+        &mut self,
+        reference: R,
+        path: AstPath,
+        skip_code_gen: bool,
+    ) {
+        if skip_code_gen {
+            self.references.insert(reference.into_reference());
```
sokra: This case should probably be an error when not in traced mode.

turbopack/crates/turbopack-tests/tests/execution/turbopack/basic/maybe-require/input/index.js
```diff
@@ -0,0 +1,7 @@
+globalThis.foobar = (v) => `this is ${v}`
+const load = globalThis.foobar || require
```
sokra, line 2: When it's `const load = require || globalThis.foobar` this will crash at runtime. I think we should error in this case even when the module is valid. So only traced mode will allow that. Bundled mode will always emit an unanalysible issue (which will be a warning in node_modules).

turbopack/crates/turbopack-ecmascript/src/code_gen.rs
mischnic: Conceptually yes.
But `resolved_cell()` is not a member of `trait ModuleReference`, so I cannot call `.resolved_cell()` for a generic `impl IntoCodeGenReference`.

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
mischnic: Done

turbopack/crates/turbopack-ecmascript/src/code_gen.rs
sokra: Ok I see... IntoCodeGenReference is not a turbo_tasks::value_trait, but just a normal trait...

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -1919,10 +1957,19 @@ where
         JsValue::explain_args(args, 10, 2)
     }
 
-    // Compute error mode from in_try and attributes.optional
+    if link_context == ValueLinkContext::InAlternative && !analysis.analyze_mode.is_tracing_assets()
+    {
+        // We are in an alternative (can't do any replacement anyway) and are not tracing assets, so
+        // we can skip further processing.
+        return Ok(());
```
sokra, line 1964: Not a fan of silently ignoring a user problem here, but I accept that for now.
We maybe want to improve that in the future...

sokra (approved):

PR #98262, opened 2026-09-04
Set tracer library version

lukahartwig:
### What?

Sets the correct Next.js library version for Otel traces.

### Why?

The tracer instance was configured with a hard coded "0.0.1" library version. This is confusing since it makes it look like trace spans were produced by `next@0.0.1`. Instead I'm setting the library version to the current Next.js version.

### How?

I'm using the `__NEXT_VERSION` env var, which gets replaced by the package.json version during build.


lukahartwig:
Flaky test should be fixed by https://github.com/vercel/next.js/pull/98024

EDIT: fixed

PR #98263, opened 2026-09-04
test: add failing nested scroll restoration case

aurorascharff:
## Summary

Adds an `it.failing` regression test for a Cache Components scroll-restoration bug. A route-owned nested scroller restores on the first Back navigation, but after Forward and Back to the same retained route its `scrollTop` resets to `0`.

The fixture lives in the existing back/forward cache suite. Once the framework behavior is fixed, the witness test can be changed from `it.failing` to `it`.

Reproduction: https://cache-components-nested-scroll-rest.vercel.app

## Verification

- `HEADLESS=true pnpm test-dev-turbo test/e2e/app-dir/back-forward-cache/back-forward-cache.test.ts`
- Confirmed the test fails with expected `1000` and received `0` when temporarily run without `it.failing`

### DIFF
--- a/test/e2e/app-dir/back-forward-cache/app/nested-scroll-restoration/detail/page.tsx
+++ b/test/e2e/app-dir/back-forward-cache/app/nested-scroll-restoration/detail/page.tsx
@@ -1,0 +1,3 @@
+export default function Page() {
+  return <h2 id="nested-scroll-detail">Detail</h2>
+}
--- a/test/e2e/app-dir/back-forward-cache/app/nested-scroll-restoration/page.tsx
+++ b/test/e2e/app-dir/back-forward-cache/app/nested-scroll-restoration/page.tsx
@@ -1,0 +1,19 @@
+import Link from 'next/link'
+
+export default function Page() {
+  return (
+    <div
+      id="nested-scroll-container"
+      style={{ height: 300, overflowY: 'auto' }}
+    >
+      <Link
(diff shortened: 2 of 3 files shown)


cheri-cherto (commented):
**Conclusion: approve** — submitting as a comment because I lack write access to request a formal review state on this repo.

Reviewed the diff (test-only change; no prior reviews to reconcile — this is the first review).

The new fixture and test follow existing conventions in this suite: `nextTestSetup({ files: __dirname })`, a real fixture directory instead of inline `files`, and `it.failing` as an unconditional witness for a known bug — consistent with other unconditional `it.failing` usages in this repo (e.g. `special-project-paths.test.ts`, `cache-components-dev-streaming.test.ts`). This differs from the `@gate`/`@force-gate` pattern (matrix-dimension differences); `it.failing` is the right tool here since the bug isn't tied to a flag/mode axis, and Jest will flag the suite the day it starts passing, so it can't silently rot.

The failing CI job (`app-action-node-middleware.test.ts` HMR test) is unrelated to this diff — no touched files intersect with actions/middleware — and looks like pre-existing flakiness, not something introduced here.

No blocking issues. Non-blocking nit: the PR description says "Once the framework behavior is fixed, the witness test can be changed from `it.failing` to `it`" but doesn't reference a tracking issue for the underlying bug — worth linking one if it exists, purely for future discoverability.

PR #98266, opened 2026-09-04
fix: size default error page icon via inline style

mitul-s:
### What

Keep the default error page warning icon at 32px when application styles target `svg`.

The SVG dimensions were set only as presentation attributes, allowing app CSS to override them. This moves the dimensions into `errorStyles.icon` and adds an e2e regression test.


test/e2e/app-dir/default-error-page-ui/default-error-page-ui.test.ts
```diff
@@ -105,6 +105,29 @@ describe('app dir - default error page UI', () => {
...
   })
 
+  it('should keep the warning icon at 32px when app CSS targets svg elements', async () => {
+    const browser = await next.browser('/trigger-error')
+
+    await browser.elementByCss('#trigger-error').click()
+
+    if (isNextDev) {
+      return
+    }
```
bgw, line 115:
```suggestion
  (isNextDev ? it.skip : it)('should keep the warning icon at 32px when app CSS targets svg elements', async () => {
    const browser = await next.browser('/trigger-error')

    await browser.elementByCss('#trigger-error').click()
```

bgw (approved):

PR #98274, opened 2026-09-04
docs: clarify Cache Components build output

aurorascharff:
## Summary

- Clarify that Cache Components uses PPR for every App Router page, including pages whose build-time HTML is empty.
- Explain that the current tree-view symbols describe output rather than a different rendering model: an empty PPR HTML output can show `ƒ` at a fixed pathname or `◐` on a dynamic-path fallback row.
- Clarify that Cache Components pages do not use `●`; generated paths from `generateStaticParams` show `○` or `◐` based on their HTML output.
- Clarify that `instant = false` opts out of instant-navigation validation and, when it is the highest config, static-shell validation without changing the PPR rendering model.
- Keep the guide's existing structure and examples while correcting factual wording about TypeScript checking, `generateStaticParams`, Partial Prefetching, cache lifetime output, and build debugging.
- Remove signposting, filler, and unverified claims in the sections being corrected.

## Verification

- `pnpm prettier --with-node-modules --ignore-path .prettierignore --check docs/01-app/02-guides/building.mdx`
- `npx eslint --config eslint.config.mjs docs/01-app/02-guides/building.mdx`
- `git diff --check`
- Verified with `cacheComponents: true` that a fixed page covered by `instant = false` prints `ƒ`, writes 0-byte HTML, and remains `PARTIALLY_STATIC` in the prerender manifest. Removing the opt-out produces the expected prerender-blocking error.
- Verified that a Cache Components route using `generateStaticParams` prints an unmarked route group, a `◐` fallback, and a `○` generated pathname without any `●` row.
- Verified `instant = false` with synchronous runtime values and `--debug-prerender` source frames against a minimal Cache Components app using the local Next.js package and webpack.
- Not run: `pnpm lint` (docs-only change; focused formatting, ESLint, and runtime checks passed).

### DIFF
--- a/docs/01-app/02-guides/ai-agents.mdx
+++ b/docs/01-app/02-guides/ai-agents.mdx
@@ -148,7 +148,7 @@ Learn more: https://nextjs.org/docs/messages/blocking-prerender-dynamic
     ...
 ```
 
-In `next dev` the stack frame resolves to your source, in both the dev overlay and the terminal. A production build minifies server code, so when the build error alone isn't enough, [`next build --debug-prerender`](/docs/app/guides/building#debug-prerender-errors) turns on server source maps and continues past the first failure.
+`next dev` shows all validation errors with source-mapped stacks in the overlay and terminal. Production builds minify server code and stop after a route fails. Use [`next build --debug-prerender`](/docs/app/guides/building#debug-prerender-errors) to enable server source maps and continue checking other routes.
 
 The `Learn more` link resolves to a per-error page under [`/docs/messages`](https://nextjs.org/docs/messages/blocking-prerender-dynamic) written for agents to read. Each page follows the same shape, with the canonical patterns for every fix, the trade-offs against the other fixes, and the gotchas an agent is likely to miss on a first attempt. The [Instant navigation guide](/docs/app/guides/instant-navigation#ai-workflow) walks through the full loop an agent runs on these errors, from reading the insight to writing an `instant()` test that fails before the fix and passes once it's in place.
 
--- a/docs/01-app/02-guides/ai-agents.mdx
+++ b/docs/01-app/02-guides/ai-agents.mdx
@@ -148,7 +148,7 @@ Learn more: https://nextjs.org/docs/messages/blocking-prerender-dynamic
     ...
 ```
 
-In `next dev` the stack frame resolves to your source, in both the dev overlay and the terminal. A production build minifies server code, so when the build error alone isn't enough, [`next build --debug-prerender`](/docs/app/guides/building#debugging-build-errors) turns on server source maps and continues past the first failure.
+In `next dev` the stack frame resolves to your source, in both the dev overlay and the terminal. A production build minifies server code, so when the build error alone isn't enough, [`next build --debug-prerender`](/docs/app/guides/building#debug-prerender-errors) turns on server source maps and continues past the first failure.
 
 The `Learn more` link resolves to a per-error page under [`/docs/messages`](https://nextjs.org/docs/messages/blocking-prerender-dynamic) written for agents to read. Each page follows the same shape, with the canonical patterns for every fix, the trade-offs against the other fixes, and the gotchas an agent is likely to miss on a first attempt. The [Instant navigation guide](/docs/app/guides/instant-navigation#ai-workflow) walks through the full loop an agent runs on these errors, from reading the insight to writing an `instant()` test that fails before the fix and passes once it's in place.
 
--- a/docs/01-app/02-guides/building.mdx
+++ b/docs/01-app/02-guides/building.mdx
@@ -55,16 +55,18 @@ bun run build
 
 After a successful build, Next.js prints a route table with a symbol next to each route, showing how it is served:
 
-| Symbol | Name              | Behavior                                                                     |
-| ------ | ----------------- | ---------------------------------------------------------------------------- |
-| `○`    | Static            | Fully prerendered at build time. Served without server rendering.            |
-| `◐`    | Partial Prerender | Static shell served immediately. Dynamic content streams in at request time. |
-| `●`    | SSG               | Prerendered static HTML (from `generateStaticParams` or `getStaticProps`).   |
-| `ƒ`    | Dynamic           | Server-rendered on demand for each request.                                  |
+| Symbol | Name              | Behavior                                                                   |
+| ------ | ----------------- | -------------------------------------------------------------------------- |
+| `○`    | Static            | Fully prerendered at build time. Served without server rendering.          |
+| `◐`    | Partial Prerender | Prerenders what it can at build time. Remaining content renders on demand. |
+| `●`    | SSG               | Prerendered static HTML (from `generateStaticParams` or `getStaticProps`). |
+| `ƒ`    | Dynamic           | Cannot use PPR. Server-rendered on demand for each request.                |
 
 `○` Static, `●` SSG, and `ƒ` Dynamic are the modes Next.js applications have always rendered in, including on the Pages Router: each route is either prerendered to static HTML or server-rendered on demand for each request.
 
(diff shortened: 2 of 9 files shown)


docs/01-app/02-guides/ai-agents.mdx
```diff
@@ -148,7 +148,7 @@ Learn more: https://nextjs.org/docs/messages/blocking-prerender-dynamic
     ...
 ```
 
-In `next dev` the stack frame resolves to your source, in both the dev overlay and the terminal. A production build minifies server code, so when the build error alone isn't enough, [`next build --debug-prerender`](/docs/app/guides/building#debugging-build-errors) turns on server source maps and continues past the first failure.
+In `next dev` the stack frame resolves to your source, in both the dev overlay and the terminal. A production build minifies server code, so when the build error alone isn't enough, [`next build --debug-prerender`](/docs/app/guides/building#debug-prerender-errors) turns on server source maps and continues past the first failure.
```
icyJoseph:
```suggestion
In `next dev`, stack frames resolve to your source in both the dev overlay and terminal.

Production builds minify server code, so stack traces no longer point to the source you need to inspect and fix. To trace a prerendering failure back to your code, run [`[next build --debug-prerender](https://chatgpt.com/docs/app/guides/building#debug-prerender-errors)`](/docs/app/guides/building#debug-prerender-errors). It enables server source maps and continues past the first failure.
```

That being said, do we mention in this document, that the build only prints the very first failure it encounters? cuz otherwise the last sentence, before and now, is a bit out of context

aurorascharff: I think at some point I was thinking about this, so that's a good catch. We definitely want to say and I thought I had it written at some point, but we want to say that this is showcasing the build output for the sake of the guide. However, the development workflow is better, for example, because it can get all validation errors at once, and you don't have to keep on finding them one by one by running the build.

docs/01-app/02-guides/building.mdx
```diff
@@ -83,7 +120,7 @@ app/
         └── page.tsx
 ```
 
-We'll work in `app/products/[id]/page.tsx`:
+The page uses the ID to fetch the product:
```
icyJoseph: Why `ID`, and not `id` here?

docs/01-app/02-guides/building.mdx
```diff
@@ -147,15 +180,15 @@ Error: Route "/products/[id]": Next.js encountered uncached or runtime data duri
...
+The detailed stack identifies the [`params`](/docs/app/api-reference/file-conventions/page#params-optional) read as the first blocked access.
 
-Running the route in `next dev` shows the full error immediately, with the component and line already resolved. See [Ensuring instant navigations](/docs/app/guides/instant-navigation) for working with these errors in depth.
+This guide uses production builds to show how each change affects the route table.
 
-With the cause located, let's apply each fix and see how the build output changes.
+During development, `next dev` shows all validation errors in the dev overlay and terminal with stack frames that resolve to your source. It reruns validation as you edit. See [Ensuring instant navigations](/docs/app/guides/instant-navigation) for the development workflow.
 
-### A streaming route
+### Stream request-time work
```
icyJoseph: Past this point in some areas, we have dropped the guide narrative altogether, and they read more like a Skill than a guide.

For example:

> Choose the `[stream]` option by adding a [`loading.js`](/docs/app/api-reference/file-conventions/loading) file:

Could be 

> To `[stream]` request-time work, add a [`loading.js`](/docs/app/api-reference/file-conventions/loading) file:

Also applies to the headings. Saying "request-time work" is also very open ended.

docs/01-app/02-guides/building.mdx
```diff
@@ -330,15 +359,23 @@ Route (app)           Revalidate  Expire
...
+The default [`use cache` profile](/docs/app/api-reference/functions/cacheLife#preset-cache-profiles) revalidates after 15 minutes. The route's default [`expireTime`](/docs/app/api-reference/config/next-config-js/expireTime) is one year, so the table shows `15m` and `1y`. You can configure both values in `next.config.ts`.
 
-### A blocking route
+### Allow the route to block
 
-The error's last suggestion, `[block]`, is the alternative to the fixes above. Say we had removed them, and only set `instant = false` on the original page. This doesn't change how the route renders. It only opts out of validation to purposely allow a blocking route:
+Static-shell validation helps protect instant navigation performance. Choose `[block]` when a route is intentionally allowed to wait for request-time work, such as before it redirects, or when you want to defer its refactor while [adopting Cache Components incrementally](/docs/app/guides/migrating-to-cache-components#adopting-incrementally).
```
icyJoseph: Here's another one, I almost would want to stay closer to what's on the red diff. 

This first sentence is a fact, but doesn't contribute to the whole? or at least not when it is the entry to the section. 

I am not following the `such as before it redirects` either, maybe some content is hidden in the PR diff review.

You know what, line 368 is a better entry I feel like, it is the specific case, and then we can follow with 366-modified, to be the generic cases or intention behind allowing to block

docs/01-app/02-guides/building.mdx
```diff
@@ -358,24 +401,45 @@ export default async function Page({ params }) {
...
+  ├ ƒ /products/2
+  └ ƒ /products/3
+
+○  (Static)   prerendered as static content
+ƒ  (Dynamic)  server-rendered on demand
+```
+
+Here, the indented `/products/[id]` row is the fallback for IDs not returned by `generateStaticParams`. Its prerender is empty, so it uses `ƒ`.
+
+Without `generateStaticParams`, the build prints only the fallback row:
```
icyJoseph: Is this the right output for the same case, with the only change being `gsP`'s presence? I think we discussed this internally? It just reads like a contradiction

- Its prerender is empty, so it uses `ƒ`.
- The fallback uses `◐` because it represents the PPR shell for the dynamic route pattern, even when its HTML is empty.

is there an additional reason why it is like this?

aurorascharff: Without GSP it cant determine the output of the route i suppose. I removed that part as its not useful

docs/01-app/02-guides/building.mdx
```diff
@@ -330,54 +361,77 @@ Route (app)           Revalidate  Expire
...
+  }
   const product = await res.json()
   return <div>{product.name}</div>
 }
 ```
 
-The build passes, and the output looks the same as the streaming route. Dynamic segments show a `◐` fallback row even when nothing meaningful is prerendered:
+With the `loading` boundary removed, no fallback can render while the page awaits `params` and the product lookup. Removing `use cache` means the lookup also runs for each request. The server waits for both before returning the product page or a `404`. During a client navigation, the destination has no fallback to show until this work completes.
+
+[`instant`](/docs/app/api-reference/file-conventions/route-segment-config/instant) is scoped by the route tree. Because this page is the only segment that sets it, `instant = false` opts the product route out of instant-navigation validation and lets the build accept an empty static shell. Setting `false` on a parent layout would extend the static-shell opt-out to its descendants, so place it as low as possible. PPR remains enabled because the flag changes validation, not the rendering model.
```
icyJoseph: > Setting `false` on a parent layout would extend the static-shell opt-out

It would only turn off validation. If it can be pre-rendered into a static shell it will.

aurorascharff: ugh i was reading this locally earlier but apparently my brain is not on

icyJoseph: > and lets the build accept an empty static shell

Nuanced because if someone put a Suspense in a shared layout, then that'd be the static shell. We need to somehow carve that in. Does the printout icon change if that's the case though? Then maybe that's a good indicator?

docs/01-app/02-guides/building.mdx
```diff
@@ -53,26 +49,65 @@ yarn build
...
+The option disables server minification, enables server source maps, and prevents one failed route from stopping the rest of the build.
 
-The symbol reflects what the route does at prerender time, not which validation configs (such as [`instant`](/docs/app/api-reference/file-conventions/route-segment-config/instant)) it exports. Without Cache Components, or on the Pages Router, prerendered pages can also show `●` (SSG).
+> **Warning**: Do not deploy builds produced with `--debug-prerender`. These builds include debugging options and are not optimized for production.
+
+### Build specific routes
+
+Use [`--debug-build-paths`](/docs/app/api-reference/cli/next#building-specific-routes) to compile and prerender selected routes while debugging a large application. The option accepts comma-separated paths and glob patterns. Prefix a pattern with `!` to exclude it. You can also combine it with `--debug-prerender`.
+
+Only matching routes appear in the route table. When you select an App Router route, Next.js also includes the built-in `/_not-found` route.
```
icyJoseph: This is one good piece of info that should also be in the CLI reference, could we add it on this PR? https://nextjs.org/docs/app/api-reference/cli/next#building-specific-routes

icyJoseph (approved):

PR #98278, opened 2026-09-04
[PPF] Fix missing runtime-follow up if hints are stale

lubieowoce:
if a page didn't use runtime data during build but started doing so after the revalidation (i.e. the hints say its statically-prefetchable but now it's not), we're supposed to be doing a follow-up runtime request based on the `needsRuntimeRequest` promise (introduced in #96095), but this was not actually happening.

the bug goes as follows:
1. if we get an entry from a static prefetch, but `needsRuntimeRequest` is true, we save it as `PPR` (or `StaticShell` if it's a shell rewound from the static response). otherwise, we'd have saved it as `PPRRuntime` (or `RuntimeShell` for shells) to indicate that it satisfies those requests.
2. `wouldRuntimeRequestProvideMore` compares the runtime strategy for the current static walk to the entry's strategy to check if it should do a runtime request. this boils down to a comparison of `RuntimeShell > PPR` for a shell request
3. however, in the `FetchStrategy` enum, `RuntimeShell = 2` and `PPR = 3`, so `RuntimeShell > PPR` is false, which makes `wouldRuntimeRequestProvideMore` return `false`. so, if we have a `PPR` entry, **we'll never do a follow-up runtime shell request**. 

This is fixed by swapping the order of the two values in the enum.

Now, if we end up with a `PPR` response where `needsRuntimeRequest` is true due to runtime data, that indicates that we *should've done a runtime request in the first place*, we just didn't know we should do so because the prefetch hint was stale, so the new ordering achieves the correct behavior, i.e. `pingSegmentBundle` decides that it needs a runtime shell request, which makes it match the intended PPF prefetching behavior.

The enum ordering is a bit of a lie, because there is no single correct ordering -- a `PPR` entry *might actually provide more content* than a `ShellRuntime` entry if the route uses static params. In practice this is resolved by the vary path (shell responses are keyed with fallback params, and a PPR entry with static params would have the actual param values, and the param-ful entry would end up being preferred) so the strategy order wouldn't matter.

Now that `RuntimeShell` beats `PPR`, i had to disable the strategy upgrading (`StaticShell -> RuntimeShell` and `PPR -> PPRRuntime`) on routes that don't use Partial Prefetching -- otherwise,  a rewound`ShellRuntime` entry from a different param was winning over `PPR` in `pingSegmentBundle`, so it was preventing issuing new static requests even if we needed to do them. However, without PPF, the runtime tiers are not meaningful anyway, so this should be fine.

### Other bugs

This fix also revealed a bug where `FetchStrategy.Full` requests were writing a shell as `RuntimeShell` even though the byte offset only marks a *static* shell. I've refactored the logic for picking the strategies to fix this and make it more explicit. 

### Edge cases

there is one (somewhat involved) edge case to this that i know of.
if the post-revalidation static response has:

1. a shell that didn't access runtime data (i.e. `needsRuntimeRequest = false` in the shell stage)
2. a prefetch that did access runtime data (i.e. `needsRuntimeRequest = true` in the static stage)

then we'll:

1. write the shell as `ShellRuntime` (because `needsRuntimeRequest = false`, and a runtime shell would not in fact provide more data)
2. write the prefetch as `PPR` (and not `PPRRuntime`, because `needsRuntimeRequest = true`, and a runtime prefetch *would* provide more data)

with the new ordering, this means we'll end up preferring the `ShellRuntime` entry over the `PPR` entry even though the shell is a strict subset of it and would be missing static params and content gated behind `prefetch()` or `navigation()`. (note that this only happens if they share the same vary path)

this can be observed as follows:
1. when a `<Link prefetch={true}>` to such a page is revealed, we won't request a runtime shell (due to already having a `ShellRuntime` entry), but we will request a runtime prefetch (because `PPR` is not enough). so far this is correct
2. if the navigation happens **before** this runtime prefetch completes, we will display what we already have. but we'll prefer the rewound `ShellRuntime` shell entry over than the `PPR` entry it originated from because of the enum ordering. so we may end up not showing content despite having it in the cache

i'm not sure how to resolve this yet, but i think the current `recordedFetchStrategy` trick can't handle this, and we need a more involved mechanism that can detect when the `PPR` entry is actually better. i don't consider this a blocker for this fix though

### DIFF
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -3791,10 +3791,10 @@ function addSegmentPathToUrlInOutputExportMode(
  * - `StaticShell` provides the shell-stage variant extracted from a static response —
  *   param-dependent content reduced to pending fallbacks, and never any content that
  *   depends on session data (cookies, headers)
+ * - `PPR` can provide static shells for each segment, including prerendered param-dependent
+ *   content at concrete paths (excluding dynamic data)
  * - `RuntimeShell` provides the shell stage rendered by a runtime request, which can
  *   additionally include shell-stage content that depends on session data
- * - `PPR` can provide shells for each segment (even for segments that use dynamic data),
- *   including prerendered param-dependent content at concrete paths
  * - `PPRRuntime` can additionally include content that uses searchParams, params, or cookies
  * - `Full` includes all the content, even if it uses dynamic data
  *
--- a/packages/next/src/client/components/segment-cache/cache.ts
+++ b/packages/next/src/client/components/segment-cache/cache.ts
@@ -2818,15 +2818,13 @@ function writeResponsePayloadsIntoCache(
   // writeServerResponseIntoCache.
   map: CacheMap<SegmentCacheEntry>
 ): Array<FulfilledSegmentCacheEntry> | null {
-  const isStaticResponse =
-    fetchStrategy === FetchStrategy.PPR ||
-    fetchStrategy === FetchStrategy.StaticShell
-  const shellWasRequested =
-    fetchStrategy === FetchStrategy.StaticShell ||
-    fetchStrategy === FetchStrategy.RuntimeShell
-
   let fulfilledEntries: Array<FulfilledSegmentCacheEntry> | null
-  if (shellPayload === null) {
+  if (
+    shellPayload === null ||
+    // `FetchStrategy.LoadingBoundary` is not used in Cache Components,
+    // and we only recover shells in Cache Components.
+    fetchStrategy === FetchStrategy.LoadingBoundary
+  ) {
     if (fetchStrategy === FetchStrategy.StaticShell) {
       // A static shell was requested but the response carries no shell (its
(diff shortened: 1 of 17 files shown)


packages/next/src/client/components/segment-cache/cache.ts
```diff
@@ -2960,13 +3020,14 @@ function writeResponsePayloadsIntoCache(
...
       metadataVaryPath,
-      shellWasRequested ? null : spawnedEntries,
+      wasShellRequested ? null : spawnedEntries,
       null,
       map
     )
+
     const shellFulfilledEntries = writeServerResponseIntoCache(
       now,
-      isStaticResponse ? FetchStrategy.StaticShell : FetchStrategy.RuntimeShell,
```
lubieowoce, line 2969: this was buggy for `Full` responses (for which `isStaticResponse` was `false`), which actually contain a `StaticShell` but were keyed at `RuntimeShell`. this was not noticeable because used to `PPR` beat `RuntimeShell`, so the end behavior was the same as if it was `StaticShell`. it became visible now that `RuntimeShell` beats `PPR`.

acdlite (approved):

PR #98281, opened 2026-09-04
Add offline zstd dictionary tooling and train a taskdata dictionary

lukesandberg:
## What?

Adds static zstd dictionary support to `turbo-persistence` and enables a 64 KiB dictionary for the TaskData family.

| Family | Compression |
| --- | --- |
| Infra | LZ4 |
| TaskMeta | LZ4 |
| TaskCache | LZ4 |
| TaskData | zstd3 + static dictionary |

Also adds `zstd_dictionary`, a small offline tool that trains a replacement dictionary from copied cache directories and prints an LZ4/zstd3/new-dictionary comparison.

## Why?

TaskData dominates persistent cache size and contains recurring serialized structures derived from JS/CSS and module graphs. A static dictionary gives zstd useful matches immediately.

Corrected held-out evaluation groups small values into production-like 8–12 KiB compression units:

| Metric | zstd3 | Dictionary | Delta |
| --- | ---: | ---: | ---: |
| Raw compressed bytes | 52,387,060 | 44,924,356 | **-14.25%** |
| Median encode diagnostic | 772.23 ms | 822.26 ms | **+6.48%** |
| Median decode diagnostic | 225.30 ms | 192.72 ms | **-14.46%** |

The raw byte savings equal 9.65% of the copied holdout cache-directory size. Timings are single-process diagnostics; real Next.js builds are the final validation.

A separate TaskMeta experiment did **not** justify replacing LZ4: dictionary-zstd was 26.55% smaller, but 4.01× slower to encode and 4.61× slower to decode.

## How?

- `CompressionConfig` supports LZ4, zstd3, and zstd3 with a static dictionary.
- Meta files record the zstd dictionary ID and validate it at open time.
- Writers reuse prepared dictionary compressors; readers reuse a thread-local decompressor.
- `sst_inspect` accepts multiple source dictionaries keyed by dictionary ID.
- The checked-in dictionary and replacement command live in `src/database/taskdata-dictionary/`.

```sh
cargo run -p turbo-persistence --release --bin zstd_dictionary -- \
  --family 2 \
  --source-dictionary turbopack/crates/turbo-tasks-backend/src/database/taskdata-dictionary/taskdata.zdict \
  --output /tmp/taskdata.zdict \
  path/to/cache-a path/to/cache-b
```

The command compares against its training inputs, so its output is diagnostic rather than held-out evidence.

## Verification

```sh
cargo test -p turbo-persistence -- --test-threads=1
cargo test -p turbo-persistence --all-targets --no-run
cargo test -p turbo-tasks-backend
cargo fmt --all -- --check
cargo clippy -p turbo-persistence -p turbo-tasks-backend --all-targets -- -D warnings
```

The production corpus used 194 `test/production` files, yielding 204 training and 38 held-out TaskData caches. The checked-in artifact is 65,536 bytes with zstd dictionary ID `1166300072`.


turbopack/crates/turbo-persistence/src/offline.rs
```diff
@@ -0,0 +1,448 @@
...
+    static_sorted_file::{
+        BLOB_VALUE_REF_SIZE, BLOCK_TYPE_FIXED_KEY_NO_HASH, BLOCK_TYPE_FIXED_KEY_WITH_HASH,
+        BLOCK_TYPE_INDEX, BLOCK_TYPE_KEY_NO_HASH, BLOCK_TYPE_KEY_WITH_HASH,
+        FIXED_KEY_BLOCK_MIXED_VALUE_TYPE, KEY_BLOCK_ENTRY_TYPE_BLOB,
+        KEY_BLOCK_ENTRY_TYPE_INLINE_MIN, KEY_BLOCK_ENTRY_TYPE_KEY_DELETED,
+        KEY_BLOCK_ENTRY_TYPE_KEY_VALUE_DELETED_MIN, KEY_BLOCK_ENTRY_TYPE_MEDIUM,
+        KEY_BLOCK_ENTRY_TYPE_SMALL, KEY_DELETED_REF_SIZE, MEDIUM_VALUE_REF_SIZE,
+        SMALL_VALUE_REF_SIZE,
+    },
+};
```
lukesandberg: can some of these APIs become `pub(crate)` instead of `pub` now that this library has been moved into the crate proper?

turbopack/crates/turbo-persistence/README.md
```diff
@@ -362,6 +362,36 @@ Configuration options for compactions are:
...
+
+The no-dictionary zstd level 3 baseline is always included during evaluation. Family 2 (TaskData) is
+selected by default; `--family <id>` overrides it. The tool follows `CURRENT`, deletion files, and
+meta-file supersession, verifies checksums, and models the same 12.5% minimum-savings threshold used
+by the writer. Index blocks and ineligible key blocks remain unchanged in modeled SST sizes.
+
+Only SST blocks are used. External blob references are counted and reported, but `.blob` payloads
+are not read. Timing fields are single-pass diagnostics; use the byte/count fields for repeatable
+comparisons of a copied cache snapshot.
+
```
lukesandberg, line 396: this is a bit of a crate layering inversion

'taskdata' is a turbo-tasks-backend concept

so i think we should just call this `zstd_dictionary` or something like that and then the documentation for training the zstd dictionary for taskdata should be in the turbo-tasks-backend crate

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+    let command = match args.next().as_deref().and_then(|value| value.to_str()) {
+        Some("train") => Command::Train,
+        Some("evaluate") => Command::Evaluate,
+        Some("--help" | "-h") => {
+            print_help();
+            std::process::exit(0);
+        }
+        Some(value) => bail!("Expected `train` or `evaluate`, got {value}"),
+        None => bail!("A `train` or `evaluate` subcommand is required"),
+    };
```
lukesandberg: we should use a normal argument parsing library like clap. i think we already have a dependency on it

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+            .with_context(|| format!("Failed to read dictionary {}", path.display()))?;
+        ensure!(
+            !dictionary.is_empty(),
+            "Dictionary {} is empty",
+            path.display()
+        );
+        let name = path
+            .file_name()
+            .and_then(|value| value.to_str())
+            .unwrap_or("dictionary")
```
lukesandberg: just reject non-utf8 names

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+            compressor,
+            decompressor,
+            setup_ns,
+        });
+    }
+    Ok(candidates)
+}
+
+fn production_stored_size(original_len: usize, compressed_len: usize) -> (usize, bool) {
+    if compressed_len < original_len - original_len / 8 {
```
lukesandberg: document the 12.5% constraint here.  ideally we could reference some logic from the normal db

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+        (compressed_len, true)
+    } else {
+        (original_len, false)
+    }
+}
+
+fn evaluate_block(
```
lukesandberg: short descriptive comment

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+        Some("--help" | "-h") => {
+            print_help();
+            std::process::exit(0);
+        }
+        Some(value) => bail!("Expected `train` or `evaluate`, got {value}"),
+        None => bail!("A `train` or `evaluate` subcommand is required"),
+    };
+    let mut options = Options {
+        command,
+        family: 2,
```
lukesandberg: this shoudl be required

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+                    "--family",
+                )?;
+            }
+            Some("--max-dictionary-size") => {
+                options.max_dictionary_size = parse_number(
+                    args.next()
+                        .context("--max-dictionary-size requires an integer")?,
+                    "--max-dictionary-size",
+                )?;
+            }
```
lukesandberg: this doesn't need to be configurable, it can be a const

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+            Some("--max-samples") => {
+                options.max_samples = parse_number(
+                    args.next().context("--max-samples requires an integer")?,
+                    "--max-samples",
+                )?;
+            }
```
lukesandberg: ditto use a const

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+    println!(
+        "Usage:\n  taskdata_dictionary train --output <PATH> [OPTIONS] <CACHE_DIRECTORY>...\n  \
+         taskdata_dictionary evaluate [OPTIONS] <CACHE_DIRECTORY>...\n\nTrain or evaluate zstd \
+         dictionaries using active TaskData SST blocks.\n\nShared options:\n      --family <ID>      \
+         Family ID (default: 2 / TaskData)\n      --json <PATH>      Write a JSON report\n\nTrain \
+         options:\n  -o, --output <PATH>    Dictionary output path\n      --max-dictionary-size \
+         <BYTES>  Maximum size (default: 65536)\n      --max-samples <COUNT>          Sample \
+         cap (default: 10000)\n      --force            Replace an existing output\n\nEvaluate \
+         options:\n  -d, --dictionary <PATH>  Candidate dictionary (repeatable)\n\n  -h, --help     \
+         Show this help"
```
lukesandberg: use a raw string?  or ideally clap would handle this

turbopack/crates/turbo-persistence/src/bin/taskdata_dictionary.rs
```diff
@@ -0,0 +1,1183 @@
...
+                    "--max-dictionary-size",
+                )?;
+            }
+            Some("--max-samples") => {
+                options.max_samples = parse_number(
+                    args.next().context("--max-samples requires an integer")?,
+                    "--max-samples",
+                )?;
+            }
+            Some("--force") => options.force = true,
```
lukesandberg: i don't think we need this, force should just be the behavior by default. or we could reject present outputs by default but i don't think we need to be flexible

imho we should just stomp on preexisting outputs this behavior isn't surprising

turbopack/crates/turbo-persistence/src/offline.rs
```diff
@@ -0,0 +1,448 @@
...
+                "Mixed key block has an invalid length"
+            );
+            Ok((0..count)
+                .map(|index| block[7 + index * stride + hash_len + key_size])
+                .collect())
+        }
+    }
+}
+
+fn entry_value_size(entry_type: u8) -> Result<usize> {
```
lukesandberg: this api already exists in StaticSortedFile

can we just reuse, make some of those apis pub(crate)?

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,992 @@
...
+    fn open(path: PathBuf, family: u32) -> Result<Self> {
+        let mut families = collect_sst_info(&path)
+            .with_context(|| format!("Failed to inspect cache {}", path.display()))?;
+        let mut ssts = families.remove(&family).with_context(|| {
+            format!(
+                "Cache {} has no active SSTs for family {family}",
+                path.display()
+            )
+        })?;
+        ssts.sort_by_key(|sst| sst.sequence_number);
```
lukesandberg, line 241: why sort? leave a comment

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,992 @@
...
+        let recorded_codecs = ssts
+            .iter()
+            .map(|sst| format!("{:?}", sst.compression))
```
lukesandberg: we should `ensure!` that everything is using Zstd compression

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,992 @@
...
+                &self.path,
+                StaticSortedFileMetaData {
+                    sequence_number: sst.sequence_number,
+                    block_count: sst.block_count,
+                },
+                sst.compression,
+            )
+            .with_context(|| format!("Failed to open {:08}.sst", sst.sequence_number))?,
+        );
+        self.current_compression = Some(sst.compression);
```
lukesandberg: why do we need to store this... isn't ijt always equal to `self.current.map(|c|c.compression)`

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,992 @@
...
+            pending: ssts.into(),
+            current: None,
+            current_compression: None,
+            seen_blobs: HashSet::new(),
+            recorded_codecs,
+            duplicate_blob_references: 0,
+        })
+    }
+
+    fn open_next_sst(&mut self) -> Result<bool> {
```
lukesandberg, line 263: instead of reading each sst sequentially we want to interleave all the elements from each cache directory. so we sample each repo.

itertools.interleave() might be usefu?

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,992 @@
...
+                    self.duplicate_blob_references += 1;
+                    return Ok(None);
+                }
+                let value = read_blob(&self.path, sequence_number, compression)?;
+                Ok(Some(Sample {
+                    kind: SampleKind::Blob,
+                    data: value,
+                }))
+            }
+            IterValue::KeyDeleted | IterValue::KeyValueDeleted { .. } | IterValue::Slice { .. } => {
```
lukesandberg, line 340: i guess we skip inline values because they are stored in key blocks?

leave a comment about that

turbopack/crates/turbo-tasks-backend/README.md
```diff
@@ -0,0 +1,24 @@
+# turbo-tasks-backend
+
+## Training a TaskData zstd dictionary
+
+TaskData is persistence family `2`. After producing copied Turbopack cache database directories,
+train and evaluate a dictionary without rebuilding the applications:
+
+```sh
+cargo run -p turbo-persistence --bin zstd_dictionary -- train \
```
lukesandberg:
```suggestion
cargo run -p turbo-persistence --release --bin zstd_dictionary -- train \
```

turbopack/crates/turbo-tasks-backend/README.md
```diff
@@ -0,0 +1,24 @@
...
+  --family 2 --output taskdata.zdict \
+  path/to/database-a path/to/database-b
+
+cargo run -p turbo-persistence --bin zstd_dictionary -- evaluate \
```
lukesandberg:
```suggestion
cargo run -p turbo-persistence --release --bin zstd_dictionary -- evaluate \
```

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1000 @@
...
+                path.display(),
+                temporary.display()
+            )
+        })?;
+        Ok(())
+    })();
+    if result.is_err() {
+        let _ = fs::remove_file(&temporary);
+    }
+    result
```
lukesandberg: we don't have corruption concerns for a one-off tool like this, a plain write_all will be fine

if we crash halfway through and produce a corrupted output, that is fine

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1000 @@
...
+    paths: &[PathBuf],
+    family: u32,
+    byte_budget: usize,
+) -> Result<TrainingSelection> {
+    let mut paths = paths.to_vec();
+    paths.sort();
+    let mut iterators = paths
+        .into_iter()
+        .map(|path| CacheSampleIter::open(path, family))
+        .collect::<Result<Vec<_>>>()?;
```
lukesandberg, line 598: is itertools::interleave feasible here

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1000 @@
...
+        self.blob.merge(&other.blob);
+    }
+
+    fn total(&self) -> Metric {
+        Metric {
+            count: self.slice.count + self.medium.count + self.blob.count,
+            bytes: self.slice.bytes + self.medium.bytes + self.blob.bytes,
+        }
+    }
+}
```
lukesandberg: i don't think we really care about this

if we are interested in cache layout we can use sst inspect but im not sure the training reports need to care at all, it is just excess information

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1000 @@
...
+            .then(|| self.raw_compressed_bytes as f64 / self.input_bytes as f64);
+    }
+}
+
+#[derive(Default, Clone, Serialize)]
+struct CompressionByKind {
+    slice: CompressionMetric,
+    medium: CompressionMetric,
+    blob: CompressionMetric,
+}
```
lukesandberg: again we don't care about the per-kind breakdown

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1028 @@
...
+    Slice,
+    Medium,
+    Blob,
+}
+
+struct Sample {
+    kind: SampleKind,
+    data: Arc<[u8]>,
+}
+
```
lukesandberg: why do we care about SAmpleKind at all?

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1028 @@
...
+    decode_ns: u64,
+}
+
+impl CompressionMetric {
+    fn merge(&mut self, other: &Self) {
+        self.input_bytes += other.input_bytes;
+        self.raw_compressed_bytes += other.raw_compressed_bytes;
+        self.estimated_stored_bytes += other.estimated_stored_bytes;
+        self.encode_ns += other.encode_ns;
+        self.decode_ns += other.decode_ns;
```
lukesandberg: can we use durations here for more type safety

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -35,12 +88,25 @@ fn decompress_block(
...
-                .decompress_to_buffer(block, dest)
-                .map_err(anyhow::Error::from)
-        }),
+        CompressionConfig::Lz4 => decompress(block, dest).map_err(anyhow::Error::from),
+        CompressionConfig::Zstd3 | CompressionConfig::Zstd3WithDictionary(_) => ZSTD_DECOMPRESSOR
+            .with_borrow_mut(|state| {
+                let dictionary = compression.dictionary().unwrap_or_default();
+                let key = compression
+                    .dictionary()
+                    .map(|dictionary| dictionary.as_ptr() as usize);
```
lukesandberg: why coerce to an int instead of just storing the static dictionary reference?

turbopack/crates/turbo-persistence/src/compression.rs
```diff
@@ -144,10 +214,41 @@ impl Compressor {
 mod tests {
     use super::*;
 
+    #[test]
+    fn dictionary_debug_does_not_include_bytes() {
+        let dictionary = Box::leak(vec![42; 64 * 1024].into_boxed_slice());
+        let debug = format!("{:?}", CompressionConfig::Zstd3WithDictionary(dictionary));
+        assert!(debug.starts_with("Zstd3WithDictionary"));
+        assert!(debug.len() < 100);
+    }
```
lukesandberg: good to know i guess, but kind of a pointless test? the bytes will be statically included in the binary so they are not secret

turbopack/crates/turbo-persistence/src/offline.rs
```diff
@@ -0,0 +1,194 @@
...
+    verify_checksum(
+        reader,
+        expected_checksum,
+        &format!("blob file {}", path.display()),
+    )?;
+    ensure!(
+        uncompressed_length > 0,
+        "Blob file {} has an invalid uncompressed length of zero",
+        path.display()
+    );
```
lukesandberg, line 137: do we not support blobs with no compression like other blocks?

turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -870,11 +798,18 @@ fn main() -> Result<()> {
     // Parse arguments
     let mut db_path: Option<PathBuf> = None;
     let mut verbose = false;
+    let mut source_dictionary: Option<PathBuf> = None;
 
     let mut i = 1;
     while i < args.len() {
         match args[i].as_str() {
```
lukesandberg: sst_inspect should adopt clap as well now that it is gaining more features

we should also support multiple dictionaries using identifiers just like production. this will allow us to adopt zstd for other families in the future

turbopack/crates/turbo-persistence/README.md
```diff
@@ -54,9 +54,10 @@ Small value blocks are emitted once they accumulate at least `MIN_SMALL_VALUE_BL
 A meta file can contain metadata about multiple SST files. The metadata is stored in a single file to avoid having too many small files.
 
 - Header
-  - 4 bytes magic number (0xFE4ADA4A)
+  - 4 bytes magic number (0xFE4ADA4B)
```
lukesandberg: version changes don't need to change the magic number.

this is a mechanism for the future when/if we add cross version compatibility

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+            .to_str()
+            .context("Dictionary filename must be UTF-8")?
+            .to_owned()
+    };
+    Ok(DictionaryInfo {
+        name,
+        path: path.map(Path::to_path_buf),
+        bytes: dictionary.len(),
+        dictionary_id: zstd::zstd_safe::get_dict_id_from_dict(dictionary).map(|id| id.get()),
+        xxh3_64: (!baseline).then(|| format!("{:016x}", xxh3_64(dictionary))),
```
lukesandberg: why do we care about the zstd id and hash?

do we need these things?

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+fn evaluate_cache(
+    path: &Path,
+    family: u32,
+    source_dictionary: Option<&'static [u8]>,
+    candidates: &mut [Candidate],
+) -> Result<CacheReport> {
+    let mut iter = CacheSampleIter::open(path.to_path_buf(), family, source_dictionary)?;
+    let mut samples = Metric::default();
+    let mut results = empty_results(candidates);
+    while let Some(sample) = iter.next_sample()? {
```
lukesandberg, line 535: ok, i think this is a key assymetry

for training we don't care about block boundaries, but for evaluation we do

so ideally we would use the original blocks, but really we can approximate here, the CacheSampleIter could group the values into blocks

blobs and medium values would be their own samples

small values would be accumulated into 16K blocks

or something like that

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+    }
+    finalize_results(&mut combined_candidates);
+    EvaluationReport {
+        schema_version: SCHEMA_VERSION,
+        family,
+        timing_note: "Single-pass wall-clock diagnostics; byte/count fields are the comparison \
+                      contract.",
+        threshold_note: "Estimated stored bytes apply the 12.5% rule per logical value and \
+                         exclude fixed container headers; they are a proxy for grouped \
+                         small-value blocks.",
```
lukesandberg: so we really need these notes in the output

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+        Ok(())
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn rejects_non_utf8_dictionary_name() {
+        use std::os::unix::ffi::OsStringExt;
+
+        let path = PathBuf::from(OsString::from_vec(vec![0xff]));
+        assert!(super::dictionary_info(Some(&path), b"data", false).is_err());
```
lukesandberg: drop test

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+        };
+        let error = train(&source, &output_dir.path().join("dictionary.zdict"))
+            .err()
+            .expect("one small cache should not train a 64 KiB dictionary");
+        assert!(format!("{error:#}").contains("Failed to train a 65536-byte dictionary"));
+        Ok(())
+    }
+
+    #[test]
+    fn trains_replaces_output_and_evaluates() -> Result<()> {
```
lukesandberg, line 937: drop test

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+        let cache =
+            make_cache_with_config(2, CompressionConfig::Zstd3WithDictionary(dictionary), 20)?;
+        assert!(CacheSampleIter::open(cache.path().to_path_buf(), 2, None).is_err());
+        let mut iter = CacheSampleIter::open(cache.path().to_path_buf(), 2, Some(dictionary))?;
+        assert!(iter.next_sample()?.is_some());
+        Ok(())
+    }
+
+    #[test]
+    fn insufficient_samples_fail_with_context() -> Result<()> {
```
lukesandberg: drop test

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+            b"incompressible-medium".to_vec(),
+            incompressible.into(),
+        )?;
+        db.commit_write_batch(batch)?;
+        db.shutdown()?;
+        Ok(tempdir)
+    }
+
+    #[test]
+    fn clap_requires_family_output_and_cache() {
```
lukesandberg: drop test

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1009 @@
...
+                "tool", "train", "--family", "2", "--output", "dict", "cache"
+            ])
+            .is_ok()
+        );
+        assert_eq!(DICTIONARY_SIZE, 64 * 1024);
+        assert_eq!(SAMPLE_BYTE_BUDGET, 64 * 1024 * 1000);
+    }
+
+    #[test]
+    fn threshold_proxy_is_strict() {
```
lukesandberg: drop test

turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -864,75 +793,61 @@ fn print_family_summary(family: u32, sst_count: usize, stats: &SstStats) {
     println!();
 }
 
+fn entry_type_help() -> String {
```
lukesandberg: drop this help text, it isn't useful and users already know it

turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -864,75 +793,61 @@ fn print_family_summary(family: u32, sst_count: usize, stats: &SstStats) {
...
 
     if !db_path.is_dir() {
         bail!("Not a directory: {}", db_path.display());
     }
 
+    let mut source_dictionaries = BTreeMap::new();
+    for path in source_dictionary {
+        let bytes =
+            fs_err::read(&path).with_context(|| format!("Failed to read {}", path.display()))?;
+        let dictionary = Box::leak(bytes.into_boxed_slice()) as &'static [u8];
```
lukesandberg, line 821: justify the leak

turbopack/crates/turbo-persistence/src/bin/sst_inspect.rs
```diff
@@ -980,3 +916,38 @@ fn main() -> Result<()> {
...
+            "first.zdict",
+            "--source-dictionary",
+            "second.zdict",
+            "database",
+        ])
+        .unwrap();
+        assert_eq!(cli.source_dictionary.len(), 2);
+        assert_eq!(cli.db_path, PathBuf::from("database"));
+    }
+}
```
lukesandberg: drop these tests they are useless

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1088 @@
...
+        #[command(flatten)]
+        source: Source,
+        /// Candidate dictionary. May be supplied more than once.
+        #[arg(short, long)]
+        dictionary: Vec<PathBuf>,
+        /// Optional JSON evaluation report.
+        #[arg(long)]
+        json: Option<PathBuf>,
+    },
+}
```
lukesandberg: i wonder if we need two commands

Train could evaluate against the provided sources,  we generally have much more data than we need and this would be simpler

the real evaluation is an actual next.js build

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1088 @@
...
+            source,
+            small_block: Vec::with_capacity(MIN_SMALL_VALUE_BLOCK_SIZE),
+            small_block_sst: None,
+        }
+    }
+
+    fn next_sample(&mut self) -> Result<Option<Sample>> {
+        loop {
+            match self.source.next_sample()? {
+                Some(sample) => match sample.small_value_sst {
```
lukesandberg: instead of this field we can just use the existing constants for the MAX_SMALL_VALUE size to detect if the samble should be emitted or accumulated

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1088 @@
...
+            }
+        }
+    }
+}
+
+/// Groups logical small values into production-like SST-local compression units for evaluation.
+struct EvaluationSampleIter {
+    source: CacheSampleIter,
+    small_block: Vec<u8>,
+    small_block_sst: Option<u32>,
```
lukesandberg: i don't think we need to worry about crossing SSTs in our accumulation, value sizes should work

turbopack/crates/turbo-persistence/src/bin/zstd_dictionary.rs
```diff
@@ -0,0 +1,1088 @@
...
+    let mut combined_candidates = empty_results(candidates);
+    for cache in &caches {
+        combined_samples.merge(&cache.samples);
+        for (combined, current) in combined_candidates.iter_mut().zip(&cache.candidates) {
+            combined.combined.merge(&current.combined);
+        }
+    }
+    finalize_results(&mut combined_candidates);
+    EvaluationReport {
+        schema_version: SCHEMA_VERSION,
```
lukesandberg: we don't care about schema version or structured output a simple textual report is fine

turbopack/crates/turbo-tasks-backend/README.md
lukesandberg, line 1: delete this file

move the information into the taskdata-dictionary subdir

turbopack/crates/turbo-persistence/README.md
```diff
@@ -362,6 +363,37 @@ Configuration options for compactions are:
...
+LZ4 or plain zstd without extra options. Pass `--source-dictionary <path>` when any input SST records
+a nonzero dictionary ID; it is ignored for LZ4 and plain-zstd SSTs. The tool follows `CURRENT`,
+deletion files, and meta-file supersession, and uses `StaticSortedFileIter` to read slice, medium, and
+blob values. Checksums, dictionary IDs, and decompressed lengths are verified.
+
+Evaluation groups small logical values into SST-local 8–12 KiB units, while medium values and blobs
+remain independent. The 12.5% minimum-savings rule is applied per approximated unit, so this remains
+comparative rather than exact SST-size modeling. Estimated stored bytes exclude fixed container
+headers. Timing fields are single-pass diagnostics; use byte/count fields for repeatable comparisons
+of one copied cache snapshot.
```
lukesandberg: way too much discussion

just say, it produces a new dictionary

turbopack/crates/turbo-tasks-backend/src/database/taskdata-dictionary.md
```diff
@@ -0,0 +1,43 @@
...
+The copied holdout cache directories occupied 77,370,864 bytes. The raw compressed-byte delta is
+7,462,704 bytes, or 9.65% of that directory total; this is a directional total-cache estimate, not
+an exact rewritten-cache measurement.
+
+The original 38-cache evaluation treated every logical value as a compression unit and overstated
+the benefit, so those numbers are intentionally not retained here. The corrected result meets the
+accepted ≥2% size / ≤10% encode-regression / no-decode-regression gate.
+
+Timing is machine-specific single-process diagnostic data. The stable byte result is the primary
+receipt.
```
lukesandberg: we don't need all this, this information will be tracked by PR descriptions

just describe how to run the commands

turbopack/crates/turbo-tasks-backend/scripts/train-taskdata-dictionary.sh
lukesandberg, line 1: lets drop this script from the commit the real dictionary will be trained on another corpus

PR #98291, opened 2026-09-05
docs: add a section on measuring memory at runtime

xabierlameiro:
### Improving Documentation

- [x] Ran `prettier` with the repo config on the changed file; no changes.
- [x] Read the Docs Contribution Guide.

## What this adds

The memory guide covers `next dev` and `next build`: dependency count, Webpack
options, `--experimental-debug-memory-usage`, heap profiles and snapshots. It
says nothing about measuring a production server that grows over hours, which
is the case most memory issues in the tracker describe.

This adds a short section, "Measure memory at runtime", after "Analyze a
snapshot of the heap". It lists the four things that make a runtime measurement
reliable (sample from a warm process, force a GC before every sample, watch
`rss` and `arrayBuffers` as well as `heapUsed`, measure one route per fresh
process) and explains what a snapshot diff between an early and a late sample
gives you.

## Why the method matters

Forcing a GC before each sample is what separates retention from garbage that
hasn't been collected yet, and it's also what makes some real leaks look fine
if you only watch `heapUsed`: part of the retention lives outside the V8 heap.
Without warm-up, the first requests to a route (compilation, caches) read as
growth. Without one route per process, growth can't be attributed. These are
the mistakes I made before I got a reliable measurement, and none of them are
written down in the guide today.

## External tools

The last paragraph names memlab and next-leak as tools that automate this
loop. They're external, the same way the Chrome DevTools link already in this
guide is. I wrote next-leak; if naming it is a problem, I'm happy to drop that
paragraph, the method stands on its own.


nvtestflt03-hash:
Great work on this PR! 🎉 The four rules for reliable runtime memory measurement are exactly what's been missing from the docs. Forcing GC before each sample and watching `rss` + `arrayBuffers` alongside `heapUsed` are insights that will save so many developers hours of debugging. Really appreciate you taking the time to document the mistakes you learned from — that's what makes great documentation. Keep it up! 💪🚀

PR #98293, opened 2026-09-05
Use sparse pointer pages for task storage

lukesandberg:
### What?

This draft replaces the resident task `FxDashMap` with sparse, directly indexed task storage:

```text
TaskId
  → permanent Boxcar directory entry per 128 IDs
  → optional 128-pointer page
  → optional Box<TaskStorage> per resident task
```

It is a sibling experiment to #98277 and remains based on the intrusive-lock foundation in #98090. Persistent and transient IDs use separate directory spaces; `task_cache` remains a DashMap.

The earlier Kovan/inline-chunk implementation on this branch has been removed. Pointer lifetime is now governed by the existing operation/snapshot coordinator, generalized as an always-enabled storage epoch.

### Why?

Inline dense chunks made memory depend on **resident density** within every touched TaskId range. Warm filesystem-cache restores and eviction can leave live tasks uniformly sparse across a monotonically increasing ID high-water mark, where almost no inline chunk becomes empty.

Pointer pages retain direct shift/mask lookup and page-parallel global work while allocating the 128-byte task payload only when resident. Boxing is acceptable here because the existing DashMap already boxes every resident task.

The epoch removes per-lookup Kovan pins, Kovan node overhead, its x86 CPU floor, and idle-worker reclamation delays. It also lets an exclusive phase synchronously null task pointers, drop task boxes, and reclaim empty pointer pages.

#### Pointer pages

- The permanent Boxcar layer has one 8-byte page pointer per 128 historical IDs.
- Each live page has 128 atomic task pointers, occupied/modified counts, and a 16-byte occupancy bitmap.
- `null` is authoritative absence; no `present` byte or post-lock identity validation is needed because deletion only occurs under coordinator exclusion.
- Common lookup is direct index arithmetic, acquire pointer loads, and the task's intrusive lock.
- Eviction scans and partially evicts concurrently, records full-removal candidates, then enters a short exclusive phase to recheck candidates, null pointers, drop boxes, and reclaim empty pages.
- Shutdown drain similarly removes pointers only under exclusion.

At an 8M TaskId high-water mark with every task evicted, the measured permanent directory is **2.26 MB** and has zero live pointer pages, versus 64 MB for a flat pointer-per-ID directory or about 1 GB for inline TaskStorage slots.

#### Lifetime-bound task guards

Every pointer dereference requires an opaque storage-access token from a live operation admission or exclusive phase. `ExecuteContext` now uses a GAT:

```rust
type TaskGuardImpl<'ctx>: TaskGuard + 'ctx
where
    Self: 'ctx;
```

`task`, `task_pair`, and task-iteration callbacks return guards borrowing `&'ctx ExecuteContext`. `operation_suspend_point` still requires `&mut self`, so ordinary Rust code cannot suspend, move, or drop the context while a task guard is live.

This removes the previous runtime `ExecuteContext::drop` panic/abort backstop. Forgetting a task guard remains unsupported: it leaks the intrusive lock, causing later admitted work to block before an exclusion barrier can reclaim the pointer.

The same-context nested-lock check remains as a borrowed `Cell<u8>`; `task_pair` is the supported two-lock path. Guards are compile-time asserted `!Send`, contexts `Send + !Sync`, and parallel child contexts each own an independent counter while sharing the parent's operation admission through scope join.

Two guarded flows are explicitly phased:

- parent-task mutation produces an owned `ConnectChildrenPlan` before child/queue work mutably uses the context;
- task-output reads drop all task guards before queue execution or mutable error conversion.

#### Pointer-page architecture vs the preceding Kovan implementation

| Workload | Pointer pages |
|---|---:|
| Cold `bench/heavy-npm-deps` median | **−4.9%** |
| Warm build median | **−1.9%** |
| GC, 10k / 50k / 200k | neutral to modestly faster |
| Raw mutable lookup | **28.1 ns → 13.6 ns** |
| Build 8,192 resident tasks | **503 µs → 378 µs** |
| Dense 100k memory | **138.2 B/task** (DashMap: 151.6 B/task) |

#### Lifetime follow-up vs the first pointer-page head

The rejected release-mode task-counter progression measured:

| Counter bookkeeping pair | Time |
|---|---:|
| Atomic RMW | 11.52 ns |
| Atomic load/store | 0.32 ns |
| Selected borrowed `Cell` | 1.10 ns |

The `Cell` version also removes one `Arc` allocation per context and an Arc refcount clone/drop per task guard.

Cached `overhead.rs` paths improved reproducibly in both benchmark orders (approximately 26–60%, depending on size). Uncached microbenchmarks were order-sensitive between runs, so no win is claimed there. GC remained statistically neutral.

Five alternating real-build pairs measured cold **21.77 s → 20.07 s** and warm **21.80 s → 20.07 s**, with effectively flat RSS. The ~8% magnitude is larger than this mechanism plausibly explains alone; treat the favorable direction as evidence, not that exact magnitude.

### Memory behavior

- Dense 100k: **138.2 B/task** versus DashMap **151.6 B/task**.
- Sparse residents pay one 8-byte slot within a touched 1 KiB pointer page plus one task box; fully empty pages are reclaimed.
- A high-water directory still scales with `max_task_id / 128`. TaskId reuse remains necessary to bound that small permanent ratchet, but is no longer needed to reclaim task payload/page memory.
- The occupancy bitmap costs 0.125 bytes per task capacity and retains the previously measured 41% / 85% / 97% subsequent-scan wins at 50% / 10% / 1% occupancy.

### Verification

- `cargo check -p turbo-tasks-backend --all-targets`
- `cargo check -p turbo-tasks-backend --features trace_task_modification`
- `cargo test -p turbo-tasks-macros`
- `cargo test -p turbo-tasks-backend` — 112 unit tests plus all integration/doc targets
- `pnpm build-all`
- exact `overhead.rs` and gated `gc.rs` Criterion A/Bs
- five alternating cold/warm real-build pairs
- focused page boundary, high-ID, namespace, concurrent publication, exclusion, removal/recreation, snapshot drain, and eviction-recheck tests

`build-all` needed one environmental retry after native Rust compilation exhausted the 32 GB sandbox; the identical command passed after clearing generated targets.

This should remain a draft while reviewers evaluate the epoch/token coverage, lifetime-bound guard API, two-phase eviction barrier, and permanent high-water directory tradeoff.


turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,1116 @@
...
+    pub(crate) modified_count: AtomicU64,
+    occupied_count: AtomicUsize,
+    probably_occupied: [AtomicU64; BITMAP_WORDS],
+    slots: Box<[TaskSlot<T>; CHUNK_SIZE]>,
+}
+
+impl<T: TaskSlotValue> TaskChunk<T> {
+    fn new() -> Self {
+        Self {
+            modified_count: AtomicU64::new(0),
```
lukesandberg: given that there are only 128 slots AtomicU64 for modified and AtomicUsize for occupied_count are overkill

turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,1116 @@
...
+            occupied_count: AtomicUsize::new(0),
+            // Start conservatively set so chunk construction and dense first insertion need no
+            // atomic read-modify-write per slot. The first scan locks/rechecks vacant slots and
+            // clears their stale hints; subsequent sparse scans skip them.
+            probably_occupied: [const { AtomicU64::new(u64::MAX) }; BITMAP_WORDS],
```
lukesandberg, line 137: i think we already have a FixedBitmap dep, but i guess it isn't atomic

i also wonder about starting with 'probably_occupied = true` for all slots

or maybe we just don't need these optimizations.  `occupied_count` allows us to skip vacant blocks and perhaps that is enough.

worth revisiting this optimization

turbopack/crates/turbo-tasks-macros/src/derive/task_storage_macro.rs
```diff
@@ -1406,15 +1426,40 @@ fn generate_typed_storage_struct(grouped_fields: &GroupedFields) -> TokenStream
...
+            #[doc = "Intrusive lock protecting this stable task slot"]
+            pub(crate) lock: IntrusiveTaskLock,
+            #[doc = "Authoritative slot presence, accessed only while `lock` is held"]
+            pub(crate) occupied: bool,
         }
 
         #[automatically_derived]
         impl TaskStorage {
+            #[doc = "Constructs fresh task payload with unlocked, vacant slot metadata."]
             pub fn new() -> Self {
```
lukesandberg: make const?

turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,1116 @@
...
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn remove_discard(&self, key: TaskId) -> bool {
+        let Some(guard) = self.get(key) else {
+            return false;
+        };
+        guard.vacate();
+        true
+    }
```
lukesandberg: if these are only needed for tests, i suggest we drop them

turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,1116 @@
...
+            kovan::flush();
+        }
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn get(&self, offset: usize) -> Option<TaskMapGuard<'a, T>> {
+        // Persistent TaskId zero is reserved and can only appear as a conservative bitmap hint.
+        if !self.transient && self.base_id + offset == 0 {
+            return None;
+        }
```
lukesandberg: do we really need this logic here? isn't reserving 0 the callers concern?

turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,1116 @@
...
+            self.task_id(offset),
+            protection,
+            self.entry,
+            offset,
+            self.len,
+            true,
+        )
+    }
+
+    pub(crate) fn swap_modified_count(&self) -> Option<u64> {
```
lukesandberg: take_modified_count is a more intuitive name

turbopack/crates/turbo-tasks-backend/src/backend/dense_task_map.rs
```diff
@@ -0,0 +1,1116 @@
...
+            self.word_index += 1;
+            self.bits = word.load(Ordering::Acquire);
+        }
+    }
+}
+
+struct ChunkDirectoryEntry<T: TaskSlotValue + Send + 'static> {
+    chunk: AtomOption<TaskChunk<T>>,
+    transition: Mutex<()>,
+    retire_candidate: AtomicBool,
```
lukesandberg, line 205: this protocol is complex

i think we can take a simpler approach

the only time we might clear out a chunk is during eviction.  there we process chunks at a time and each thread is processing a single chunk.

There can be concurrent writes which is the concern, but perhaps a higher level primitive is in order... we could open chunks for 'draining' and then exclude lookups while that is in flight.  this could be done using a single AtomicBool 'exclusive' flag and callers could spin upon seeing it.  eviction should be able to move through chunks quickly so this shouldn't add much latency, and it could enable a simple 'take' approach for retiring.

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -175,56 +178,132 @@ pub trait ChildExecuteContext<'e>: Send + Sized {
...
+    fn assert_clear(&self) {
+        assert!(
+            self.is_clear(),
+            "operation reached a suspend point while a task guard was alive"
+        );
+    }
+}
+
+#[cfg(test)]
+mod task_lock_counter_tests {
```
lukesandberg, line 236: tests should be at the bottom of the file

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -234,16 +313,40 @@ pub struct ExecuteContextImpl<'e> {
...
+            // Keep the coordinator admission permanently live rather than allowing an exclusive
+            // phase to free a task pointer still held by an escaped guard. This is a protocol
+            // violation, so fail loudly after preserving memory safety even if panic is caught.
+            std::mem::forget(guard);
+            panic!("execute context dropped while a task guard was alive");
+        }
+        // Child and GC contexts borrow rather than own their admission/phase, so they cannot leak
+        // it to preserve an escaped pointer. Abort rather than permit an unwind catcher to release
+        // the borrowed epoch while a task guard remains live.
+        std::process::abort();
```
lukesandberg, line 333: this needs to print an error

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -234,16 +313,40 @@ pub struct ExecuteContextImpl<'e> {
...
+            return;
+        }
+        if let ExecutePhase::Normal { guard, .. } = &mut self.phase
+            && let Some(guard) = guard.take()
+        {
+            // Keep the coordinator admission permanently live rather than allowing an exclusive
+            // phase to free a task pointer still held by an escaped guard. This is a protocol
+            // violation, so fail loudly after preserving memory safety even if panic is caught.
+            std::mem::forget(guard);
+            panic!("execute context dropped while a task guard was alive");
```
lukesandberg, line 328: i suppose this drop impl is the biggest thing to advocate for using lifetimes to prevent executioncontext from being dropped too early

turbopack/crates/turbo-tasks-backend/src/backend/task_page_map.rs
```diff
@@ -0,0 +1,845 @@
...
+        } else {
+            (1024 * 1024) / PAGE_SIZE
+        };
+        Self {
+            persistent: PagedVec::with_page_capacity(persistent_page_capacity),
+            transient: PagedVec::with_page_capacity(1),
+        }
+    }
+
+    fn namespace_and_index(&self, key: TaskId) -> (&PagedVec<T>, usize) {
```
lukesandberg, line 368: id_space?

turbopack/crates/turbo-tasks-backend/src/backend/task_page_map.rs
```diff
@@ -0,0 +1,845 @@
...
+        self.persistent.clear(access);
+        self.transient.clear(access);
+    }
+
+    pub(crate) fn reclaim_empty_pages(&self, access: StorageAccessToken<'_>) -> usize {
+        access.assert_exclusive();
+        self.persistent.reclaim_empty_pages(access) + self.transient.reclaim_empty_pages(access)
+    }
+
+    #[allow(dead_code)]
```
lukesandberg, line 470: cfg(test?

PR #98312, opened 2026-09-07
turbo-tasks-backend: fix strongly consistent read hanging on a canceled task

lukesandberg:
### What

Fixes a shutdown hang in `next build`: a strongly consistent read could wait forever on a task that shutdown had canceled, which in turn blocked `stop_and_wait` and hung the build until the test timeout.

Also makes the `hanging_detection` feature usable — it panics due to a `data` access check today.

### Why the read hangs

1. Shutdown sets `stopped`. A task dispatched after that point is canceled before it runs, so it has no output.
2. Its parent completes and connects its children. `process_new_children` uses `!child.has_output()` as the test for "not computed yet" and marks such a child `InitialDirty` so it gets scheduled.
3. `make_task_dirty` promotes the canceled task's `SessionDependent` to plain `Dirty` and clears `current_session_clean`. That is correct for a live task that must re-run, and wrong for one that never will.
4. `AggregatedDataUpdate::from_task` then snapshots `dirty_state() == (true, false)` and contributes a dirty container that is not session-clean.
5. The parent is itself clean, so its own `update_dirty_state(None -> None)` early-returns and never evaluates `all_clean_event`. Its dirty-container count never drains.
6. The strongly consistent reader waits on that event forever, holding a foreground job that `stop_and_wait` blocks on.

`has_output()` is a proxy for "needs computing", and a canceled task breaks that assumption.

### The fix

Record a terminal error output when a task execution is canceled. 

Make hanging_detection more reliable

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/Cargo.toml
+++ b/turbopack/crates/turbo-tasks-backend/Cargo.toml
@@ -22,8 +22,6 @@ verify_serialization = []
 verify_aggregation_graph = []
 verify_immutable = []
 verify_determinism = ["turbo-tasks/verify_determinism"]
-# Reports an EventListener that is not notified within its timeout, printing the description and
-# note that `listen_with_note` call sites already carry. Diagnostic only.
 hanging_detection = ["turbo-tasks/hanging_detection"]
 task_dirty_cause = ["turbo-tasks/task_dirty_cause"]
 trace_aggregation_update_stats = []
--- a/turbopack/crates/turbo-tasks-backend/Cargo.toml
+++ b/turbopack/crates/turbo-tasks-backend/Cargo.toml
@@ -22,6 +22,9 @@ verify_serialization = []
 verify_aggregation_graph = []
 verify_immutable = []
 verify_determinism = ["turbo-tasks/verify_determinism"]
+# Reports an EventListener that is not notified within its timeout, printing the description and
+# note that `listen_with_note` call sites already carry. Diagnostic only.
+hanging_detection = ["turbo-tasks/hanging_detection"]
 task_dirty_cause = ["turbo-tasks/task_dirty_cause"]
 trace_aggregation_update_stats = []
 trace_aggregation_update_queue = []
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -621,23 +621,6 @@ impl TurboTasksBackend {
                 );
             }
 
-            // A canceled task never runs again, so it can never become clean and nothing will
-            // ever fire its `all_clean_event`. Subscribing below would park this reader forever:
-            // `task_execution_canceled` notifies the waiters that already exist, but a reader
-            // arriving *after* the cancel re-creates the activeness state and waits on an event
-            // with no remaining notifier.
-            //
-            // Only the `Canceled` arm of `check_in_progress` is taken here, deliberately. Its
-            // `Scheduled`/`InProgress` arms return as soon as the task itself is running, which
-            // is right for the weakly consistent path below but would short-circuit the
-            // dirty-container wait that makes *this* read strongly consistent.
-            if matches!(task.get_in_progress(), Some(InProgressState::Canceled)) {
-                let description = task.get_task_description();
-                drop(task);
-                drop(reader_task);
-                anyhow::bail!("{description} was canceled");
-            }
-
             let is_dirty = task.is_dirty();
 
             // Check the dirty count of the root node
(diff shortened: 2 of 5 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -1987,6 +2004,29 @@ impl TurboTasksBackend {
...
+        // will never run again this session, and being marked plain `Dirty` also clears the
+        // session-clean flag set below, so it stays a dirty container of its parent forever. The
+        // parent then never reaches a clean state, and a strongly consistent reader waiting on
+        // the parent's `all_clean_event` waits forever — taking `stop_and_wait` with it, since
+        // the reader holds a foreground job.
+        //
+        // Giving the task an output makes it terminal for this session, so nothing re-dirties it
+        // and the aggregate settles. The `SessionDependent` marking below still ensures the next
+        // session re-executes it rather than trusting this error.
+        if task.get_output().is_none() {
```
lukesandberg: comment is too verbose, and i wonder if the `is_none` check is correct? why not just always set the output?

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -621,6 +621,23 @@ impl TurboTasksBackend {
...
+            // ever fire its `all_clean_event`. Subscribing below would park this reader forever:
+            // `task_execution_canceled` notifies the waiters that already exist, but a reader
+            // arriving *after* the cancel re-creates the activeness state and waits on an event
+            // with no remaining notifier.
+            //
+            // Only the `Canceled` arm of `check_in_progress` is taken here, deliberately. Its
+            // `Scheduled`/`InProgress` arms return as soon as the task itself is running, which
+            // is right for the weakly consistent path below but would short-circuit the
+            // dirty-container wait that makes *this* read strongly consistent.
+            if matches!(task.get_in_progress(), Some(InProgressState::Canceled)) {
```
lukesandberg: this is not strictly required to fix a hang, right?

so should we drop it?

At the very least the comment is too verbose

turbopack/crates/turbo-tasks-backend/tests/shutdown_cancellation.rs
```diff
@@ -0,0 +1,104 @@
...
+//!
+//! Turbo-tasks does not cancel a task that is already executing; cancellation applies to a task
+//! that is *dispatched* after `stopped` is set. So a task launched during shutdown is canceled
+//! before it runs, and a canceled task never runs again — it can never become clean, so nothing
+//! will ever fire its `all_clean_event`. A strongly consistent reader waiting on that event would
+//! wait forever, and `stop_and_wait` would in turn block on the foreground job the reader holds.
+//!
+//! Only `next build` reaches this in production (the dev server uses a short-timeout
+//! `project_on_exit`), which is why it went untested.
+
```
lukesandberg:
```suggestion
//! Regression test for a bug where shutdown would cause a strongly consistent read to hang
//! if a task in scope was cancelled.

```

turbopack/crates/turbo-tasks-backend/tests/shutdown_cancellation.rs
```diff
@@ -0,0 +1,104 @@
...
+    // surface that rather than hang.
+    Ok(Vc::cell(*launched_during_shutdown().await?))
+}
+
+/// A strongly consistent read whose task launches a child during shutdown.
+///
+/// The read is expected to fail — the child is canceled. What must not happen is
+/// `stop_and_wait` blocking forever on the reader's foreground job.
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn shutdown_unblocks_read_of_canceled_child() {
```
lukesandberg, line 47:
```suggestion
#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
async fn shutdown_unblocks_read_of_canceled_child() {
```

turbopack/crates/turbo-tasks-backend/tests/shutdown_cancellation.rs
```diff
@@ -0,0 +1,104 @@
...
+            // the property under test.
+            let _ = pauses_then_launches_a_child()
+                .read_strongly_consistent()
+                .await;
+            anyhow::Ok(())
+        })
+        .await;
+    });
+
+    for _ in 0..500 {
```
lukesandberg: 500 seems a little high?

instead of polling is there a `Barrier` we could use for proper synchronization

turbopack/crates/turbo-tasks-backend/Cargo.toml
```diff
@@ -22,6 +22,9 @@ verify_serialization = []
 verify_aggregation_graph = []
 verify_immutable = []
 verify_determinism = ["turbo-tasks/verify_determinism"]
+# Reports an EventListener that is not notified within its timeout, printing the description and
+# note that `listen_with_note` call sites already carry. Diagnostic only.
+hanging_detection = ["turbo-tasks/hanging_detection"]
```
lukesandberg:
```suggestion
hanging_detection = ["turbo-tasks/hanging_detection"]
```

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Both addressed: comment trimmed to four lines, and the `is_none` check dropped in favour of always setting the output.

On whether it was correct — I don't think the guard was protecting anything real. `task_execution_canceled` runs on a task that was dispatched but never executed, so it has no output to clobber; and if it somehow did have one, that output came from a previous execution whose result we're about to mark `SessionDependent` dirty anyway, so preserving it buys nothing. Unconditional is simpler and 245/245 backend tests pass with it.

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
lukesandberg: :robot: Correct, it isn't required — I verified that by disabling each change in turn: with the output fix in place the test passes whether or not this guard exists. Dropped it.

It was speculative defense in a hot read path, which isn't worth the cost or the extra branch to reason about. Removing it also let me revert the `reader_description` hoist, so `try_read_task_output` is now untouched by this PR.

turbopack/crates/turbo-tasks-backend/tests/shutdown_cancellation.rs
lukesandberg: :robot: Applied — module doc replaced with your two-line version.

turbopack/crates/turbo-tasks-backend/tests/shutdown_cancellation.rs
lukesandberg: :robot: Applied.

turbopack/crates/turbo-tasks-backend/Cargo.toml
lukesandberg: :robot: Applied — comment dropped.

turbopack/crates/turbo-tasks-backend/tests/shutdown_cancellation.rs
lukesandberg: :robot: Switched to two `tokio::sync::Barrier`s and removed the poll loop entirely — the 500 iterations are gone.

Two barriers rather than one because the test needs to shut down *between* them: the task waits on PARKED, the test joins it and starts `stop_and_wait`, then RELEASED lets the task proceed so its child is dispatched after `stopped` is set. That's what makes the child get cancelled rather than run.

One `sleep(50ms)` remains between the two, to let shutdown actually set `stopped` before the task resumes. I couldn't find a synchronisation point that observes that transition from outside the backend — if there's one I've missed, happy to use it. The sleep only affects whether the test exercises the cancellation path, not whether it can produce a false pass: the PARKED barrier already guarantees the task is parked, and if the child were dispatched too early it would simply run normally and the test would pass trivially rather than hang.

sokra (approved):

PR #98331, opened 2026-09-07
[ci] Align build-and-deploy triggers with build-and-test

eps1lon:
Change detection (`git diff HEAD~1`) only covers the whole change set when the checkout is the PR merge commit. `build-and-test` runs on `pull_request` while `build-and-deploy` ran on `push` for upstream PR branches, so a multi-commit push whose last commit was docs-only made `build-and-deploy` skip the preview build while deploy tests in `build-and-test` waited up to 30 minutes for a preview tarball that never got published.

`build-and-deploy` now triggers on the same events as `build-and-test` (`push` to the default branch and `pull_request`), while still accepting release tags and workflow dispatch. The `deploy-target` fork guard is removed since upstream PRs are no longer covered by a push run, and `create-release-branch.js` now rewrites the workflow's branch list alongside the existing `refs/heads/canary` replacement so release branches keep their staging deploys.

Preview tarball creation now uses the PR head sha instead of falling back to the merge-commit sha on `pull_request` `opened` events (`github.event.after` only exists on `synchronize`), matching the sha that `upload_preview_tarballs.yml` publishes under and that tests poll for. This fallback was a pre-existing bug that only affected fork PRs; it would have started affecting every upstream PR's first run once previews moved to the `pull_request` event.

## test plan

- [determine deploy target detects two file changes even the last commit only changed one file](https://github.com/vercel/next.js/actions/runs/34203720949/job/101988232190?pr=98331#step:6:65)

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -3,12 +3,6 @@ name: build-and-deploy
 
 on:
   push:
-    # Same branch trigger as build_and_test.yml so that change detection
-    # (`git diff HEAD~1`) evaluates the same commits in both workflows.
-    # PR branches are covered by the `pull_request` event below, where the
-    # checkout is the merge commit and HEAD~1 covers the whole PR. On branch
-    # pushes HEAD~1 would only cover the last commit of the push.
-    # create-release-branch.js rewrites this list for release branches.
     branches: ['canary']
     # Run on release tags so publishing is driven by the tag push rather than
     # the release-commit branch push (see the deploy-target step).
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -3,9 +3,13 @@ name: build-and-deploy
 
 on:
   push:
-    # Don't run when graphite base branches are pushed
-    branches-ignore:
-      - 'graphite-base/**'
+    # Same branch trigger as build_and_test.yml so that change detection
+    # (`git diff HEAD~1`) evaluates the same commits in both workflows.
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_deploy.yml
```diff
@@ -3,9 +3,13 @@ name: build-and-deploy
...
   push:
-    # Don't run when graphite base branches are pushed
-    branches-ignore:
-      - 'graphite-base/**'
+    # Same branch trigger as build_and_test.yml so that change detection
+    # (`git diff HEAD~1`) evaluates the same commits in both workflows.
+    # PR branches are covered by the `pull_request` event below, where the
+    # checkout is the merge commit and HEAD~1 covers the whole PR. On branch
+    # pushes HEAD~1 would only cover the last commit of the push.
+    # create-release-branch.js rewrites this list for release branches.
```
eps1lon:
```suggestion
```

.github/workflows/build_and_deploy.yml
icyJoseph, line 180: Should this be updated too though? 

```txt
node scripts/set-preview-version.js "${{ github.event.pull_request.head.sha || github.sha }}"
```

I guess it is only needed for tracing purposes or correctness?

icyJoseph (approved):

.github/workflows/build_and_deploy.yml
eps1lon: Yeah it's just "informational" and a pre-existing bug that doesn't need fixing here. But I'll follow-up to align the two once this has soaked for a bit.

PR #98339, opened 2026-09-08
[PPF] Model caches with `stale < MIN_SHELL_STALE` as prefetch-only data

lubieowoce:
Caches that have `stale < MIN_SHELL_STALE` were using `makeStageHangingPromise` in prerenders, which (in theory) was meant to track a runtime data access during a static prerender, which in turn would make the page use runtime shells. This is semantically incorrect, because both static and runtime shells would exclude the cache, and both static and runtime prefetches would include it, so it does not actually signal runtimeness, only prefetchness. The only thing we should be doing for these caches is excluding the content, same as `unstable_prefetch()`. 

This PR replaces `makeStageHangingPromise` with `makePrefetchHangingPromise`, which is not tracked as runtime data at all. We also use it for `unstable_prefetch()` to signal the correspondence.

Note that in practice, the usage of `makeStageHangingPromise` in `use-cache-wrapper` was *technically* fine, because its runtime data tracking was effectively dead code. We only returned it if `stagedRendering.finalStage < stage`, but for static prerenders `stage = staticLinkDataStage = PrefetchStatic` and `finalStage = Static`, so the condition was never actually true, and we never actually tracked it as a runtime data access. However semantically it should still not be tracked as runtime data.
(The condition *can* be true in a runtime prerender that only prerenders the shell. We don't do any tracking in runtime prerenders, so it was also a no-op).

(this was extracted from #98129, where we separate runtime data access tracking into session data and URL data. i couldn't figure which one `makeStageHangingPromise` should be tracked as, but then i realized it shouldn't be tracked at all)


packages/next/src/server/use-cache/use-cache-wrapper.ts
```diff
@@ -2555,35 +2539,67 @@ export async function cache(
...
+                  // Final prerender (with staging).
+                  // This entry cannot be part of the shell, so we delay it to the prefetch.
+                  let prefetchStage: AdvanceableRenderStage
+                  if (prerenderStore.type === 'prerender') {
+                    prefetchStage = RENDER_STAGES_BY_DATA_KIND.staticLinkData
+                  } else {
+                    prefetchStage = RENDER_STAGES_BY_DATA_KIND.runtimeLinkData
+                  }
+                  if (
+                    // If the render ends before the post-shell stage (e.g. a
```
```suggestion
                    // If the render ends before the prefetch stage (e.g. a
```

unstubbable (approved):

PR #98341, opened 2026-09-08
Turbo Tasks: abort unneeded in-flight work

lukesandberg:
### What?

Abort in-flight Turbo Tasks work when its result is no longer needed:

- invalidated executions stop before doing work that will immediately be discarded
- executions that lose activeness are interrupted recursively with the inactive subtree
- GC can opportunistically interrupt executions that would otherwise be collectible

Aborted tasks remain recoverable and execute again if they become live. The task is kept or made dirty, and ordinary stale/active scheduling performs a clean execution.

The abort lifecycle is also auditable through Turbo Tasks traces and the existing opt-in per-function task statistics.

### Why?

Turbo Tasks executes work asynchronously and sometimes speculatively. Previously, a task always ran to completion after starting, even when invalidation made its result stale or its final active parent disconnected. Besides wasting resources, this delayed the useful replacement execution.

Runtime reporting distinguishes an abort request from work that was actually interrupted, requests that raced successful completion, and requests skipped because a function is not cancellation-safe.

### How?

Each in-progress cancelable native Turbo Task stores a futures abort handle, while the executor polls its future through an abortable wrapper. Invalidation and liveness transitions signal that handle. The backend then performs one of two recovery paths:

- stale work reuses the existing stale-priority scheduling path
- disconnected work becomes dirty without being scheduled until it is live again

Successful completion disarms its abort state before ordinary completion bookkeeping. The abort path reverses speculative child activeness, wakes cell waiters, and rechecks liveness to handle reconnection racing with interruption. Transient root and one-shot futures are not aborted because they cannot necessarily be recreated.

Turbo Task functions are abortable by default. `#[turbo_tasks::function(non_cancelable)]` opts out concrete cancellation-unsafe functions. The Node evaluator entry points use this because dropping them after the process pool produced a result can lose protocol progress.

Abort telemetry uses first-trigger-wins attribution (`invalidation`, `inactive`, or `gc`) while independently preserving whether later liveness changes made the task unneeded. `turbo_tasks::function` spans record task ID, cancelability, execution outcome, and abort trigger. Lifecycle events report requested, observed, raced-completion, and skipped outcomes. The existing opt-in `TaskStatisticsApi` exposes corresponding per-function counters without adding always-on output.

### Risks

Cancellation is equivalent to dropping a future and can violate application-level invariants around `tokio::select!`, channel protocols, partial I/O, and multi-await state transitions. The known Node evaluator boundary is opted out. A broader failpoint-based audit of Node pools and filesystem tasks remains follow-up work.

Telemetry is high-volume when full Turbo Tasks tracing is enabled, but it uses the existing opt-in tracing/statistics facilities and adds no always-on logging.

### Verification

- deterministic invalidation, inactivity, completion-race, and non-cancelable abort tests
- first-trigger-wins and GC/inactivity state tests
- existing per-function task-statistics JSON tests plus new abort counter assertions
- shutdown, dirty-state, graph, and GC regression targets
- `cargo test -p turbo-tasks --lib`
- feature-enabled `cargo check`
- `cargo fmt --all -- --check` and `git diff --check`
- `pnpm build-all`
- filesystem-cache HMR integration test
- manual binary trace capture/query verifying requested → observed/raced reconciliation and skipped non-cancelable records

### DIFF
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -2111,7 +2111,7 @@ impl TurboTasksBackend {
             return None;
         };
         let InProgressState::InProgress(in_progress) = in_progress else {
-            // Shutdown cancellation or re-scheduling won the race. Preserve that state.
+            // Another state transition won the race. Preserve that state.
             let old = task.set_in_progress(in_progress);
             debug_assert!(old.is_none(), "InProgress already exists");
             return None;
@@ -2154,21 +2154,12 @@ impl TurboTasksBackend {
             decrease_active_counts_of_new_children(new_children, &mut ctx);
             return Some(priority);
         }
-        let immutable = task.immutable();
         drop(task);
 
         decrease_active_counts_of_new_children(new_children, &mut ctx);
 
-        if immutable {
-            // An immutable result never needs to be re-executed. This can happen when losing
-            // activeness races with the final completion bookkeeping.
-            done_event.notify(usize::MAX);
-            drop(in_progress_cells);
-            return None;
-        }
-
-        // Unlike shutdown cancellation, an unneeded abortion is recoverable: discard the current
-        // execution state and make the task dirty without scheduling it while it is disconnected.
+        // Discard the aborted execution state and make the task dirty without scheduling it while
+        // it is disconnected.
         let mut queue = AggregationUpdateQueue::new();
         let mut task = ctx.task(task_id, TaskDataCategory::All);
         make_task_dirty_internal(
@@ -2257,14 +2248,14 @@ impl TurboTasksBackend {
                     _ => None,
                 };
             }
-            let (abort_handle, registration) = AbortHandle::new_pair();
-            abort_registration = registration;
-            let abort_handle = match &task_type {
+            let (abort_handle, registration) = match &task_type {
                 TaskType::Cached(task_type) if task_type.native_fn.is_cancelable => {
-                    Some(abort_handle)
+                    let (abort_handle, registration) = AbortHandle::new_pair();
+                    (Some(abort_handle), Some(registration))
                 }
-                _ => None,
+                _ => (None, None),
             };
+            abort_registration = registration;
             let old = task.set_in_progress(InProgressState::InProgress(Box::new(
                 InProgressStateInner {
                     stale: false,
@@ -2519,6 +2510,12 @@ impl TurboTasksBackend {
         has_invalidator: bool,
     ) -> Result<TaskExecutionCompletePrepareResult, TaskPriority> {
         let mut task = ctx.task(task_id, TaskDataCategory::All);
+        if let Some(InProgressState::InProgress(in_progress)) = task.get_in_progress_mut() {
+            // The task future completed without observing a concurrent abort request. From this
+            // point on, use the ordinary completion bookkeeping: stale tasks re-execute and
+            // inactive tasks finalize into a state a later GC pass can collect.
+            in_progress.disarm_abort();
+        }
         let is_recomputation = task.is_dirty().is_none();
         // Without dependency tracking, the SessionDependent dirty state is never read (no session
         // restore), so skip the work
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -2255,7 +2255,12 @@ impl TurboTasksBackend {
             }
             let (abort_handle, registration) = AbortHandle::new_pair();
             abort_registration = registration;
-            let abort_handle = matches!(&task_type, TaskType::Cached(_)).then_some(abort_handle);
+            let abort_handle = match &task_type {
+                TaskType::Cached(task_type) if task_type.native_fn.is_cancelable => {
+                    Some(abort_handle)
+                }
+                _ => None,
+            };
             let old = task.set_in_progress(InProgressState::InProgress(Box::new(
                 InProgressStateInner {
                     stale: false,
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -2103,8 +2103,15 @@ impl TurboTasksBackend {
     ) -> Option<TaskPriority> {
         let mut ctx = self.execute_context(turbo_tasks);
         let mut task = ctx.task(task_id, TaskDataCategory::All);
-        let Some(InProgressState::InProgress(in_progress)) = task.take_in_progress() else {
-            panic!("Task execution aborted, but task is not in progress: {task:#?}");
+        let Some(in_progress) = task.take_in_progress() else {
+            // Completion or another abort callback won the race.
+            return None;
+        };
+        let InProgressState::InProgress(in_progress) = in_progress else {
+            // Shutdown cancellation or re-scheduling won the race. Preserve that state.
+            let old = task.set_in_progress(in_progress);
+            debug_assert!(old.is_none(), "InProgress already exists");
+            return None;
         };
         let InProgressStateInner {
             stale,
@@ -2144,9 +2151,18 @@ impl TurboTasksBackend {
             decrease_active_counts_of_new_children(new_children, &mut ctx);
             return Some(priority);
         }
+        let immutable = task.immutable();
         drop(task);
 
         decrease_active_counts_of_new_children(new_children, &mut ctx);
+
+        if immutable {
+            // An immutable result never needs to be re-executed. This can happen when losing
+            // activeness races with the final completion bookkeeping.
+            done_event.notify(usize::MAX);
+            drop(in_progress_cells);
+            return None;
+        }
 
         // Unlike shutdown cancellation, an unneeded abortion is recoverable: discard the current
         // execution state and make the task dirty without scheduling it while it is disconnected.
@@ -2154,10 +2170,13 @@ impl TurboTasksBackend {
         let mut task = ctx.task(task_id, TaskDataCategory::All);
         make_task_dirty_internal(
             &mut task,
-            true,
+            // The aborted execution's in-progress state was already taken above, so there is
+            // nothing left to mark stale.
+            /* make_stale */
             false,
+            /* schedule_when_active */ false,
             #[cfg(feature = "task_dirty_cause")]
-            TaskDirtyCause::ExecutionAborted,
+            TaskDirtyCause::BecameInactive,
             &mut queue,
             &mut ctx,
         );
@@ -2167,6 +2186,15 @@ impl TurboTasksBackend {
         // A connection may race with dropping the aborted future. Re-check liveness after the
         // dirty transition so a revived task is not left dirty but unscheduled.
         let mut task = ctx.task(task_id, TaskDataCategory::All);
+        if task.get_in_progress().is_some() {
+            // A reader/connection already scheduled (or started) the dirty task while the abort
+            // callback propagated its dirty transition. Keep that execution and only release
+            // listeners attached to the aborted execution.
+            done_event.notify(usize::MAX);
+            drop(task);
+            drop(in_progress_cells);
+            return None;
+        }
         let should_schedule = if self.should_track_activeness() {
             task.has_activeness()
         } else {
@@ -2836,7 +2864,7 @@ impl TurboTasksBackend {
             make_task_dirty_internal(
                 &mut dependent,
                 make_stale,
-                true,
+                /* schedule_when_active */ true,
                 #[cfg(feature = "task_dirty_cause")]
                 cause.clone(),
                 queue,
--- a/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
+++ b/turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
@@ -18,7 +18,10 @@ use std::{
     hash::BuildHasherDefault,
     mem::take,
     pin::Pin,
-    sync::{Arc, LazyLock},
+    sync::{
+        Arc, LazyLock,
+        atomic::{AtomicBool, Ordering},
+    },
     time::SystemTime,
 };
 
@@ -70,9 +73,9 @@ use crate::{
         operation::{
             AggregationUpdateJob, AggregationUpdateQueue, ChildExecuteContext,
             CleanupOldEdgesOperation, ConnectChildOperation, ExecuteContext, ExecuteContextImpl,
-            LeafDistanceUpdateQueue, Operation, OutdatedEdge, TaskGuard, TaskType, TaskTypeRef,
-            capture_all_outgoing_edges, connect_children, get_aggregation_number, get_uppers,
-            make_task_dirty_internal, prepare_new_children,
+            LeafDistanceUpdateQueue, MakeTaskDirtyOptions, Operation, OutdatedEdge, TaskGuard,
+            TaskType, TaskTypeRef, capture_all_outgoing_edges, connect_children,
+            get_aggregation_number, get_uppers, make_task_dirty_internal, prepare_new_children,
         },
         snapshot_coordinator::{OperationGuard, SnapshotCoordinator},
         storage::Storage,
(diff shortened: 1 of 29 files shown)


turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_children.rs
```diff
@@ -82,6 +82,7 @@ pub fn connect_children(
                     make_task_dirty_internal(
                         &mut child,
                         false,
+                        true,
```
lukesandberg: add /* schedule_when_active */ here and elsewhere

turbopack/crates/turbo-tasks/src/task_dirty_cause.rs
```diff
@@ -25,6 +25,8 @@ pub enum TaskDirtyCause {
     /// hard-delete: its edges were scrubbed, so it must re-execute to rebuild them.
     Resurrected,
     Unknown,
+    /// An in-flight execution was aborted because no active work needed it anymore.
+    ExecutionAborted,
```
lukesandberg: there are 2 kinds of aborts

RacingInvalidation
BecameInactive

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -79,7 +80,9 @@ impl ConnectChildOperation {
             let Some(InProgressState::InProgress(InProgressStateInner { new_children, .. })) =
                 parent_task.get_in_progress()
             else {
-                panic!("Task is not in progress while calling another task: {parent_task:?}");
+                // The parent execution was aborted before this connect operation acquired its
+                // first guard. Ignore the child call from the dropped execution.
+                return;
```
lukesandberg, line 91: can we check that this is true? 

skipping a dirty call might have odd consequences if it isn't

turbopack/crates/turbo-tasks-backend/src/backend/operation/connect_child.rs
```diff
@@ -166,7 +169,18 @@ impl ConnectChildOperation {
             let Some(InProgressState::InProgress(InProgressStateInner { new_children, .. })) =
                 parent_task.get_in_progress_mut()
             else {
-                panic!("Task is not in progress while calling another task: {parent_task:?}");
+                // Abortion can race while the aggregation update above temporarily releases the
```
lukesandberg, line 178: same comment, can we check that the parent was aborted?

turbopack/crates/turbo-tasks-backend/src/backend/operation/invalidate.rs
```diff
@@ -98,17 +98,22 @@ pub fn make_task_dirty(
...
-/// Requires the guard to be allocated with [TaskDataCategory::All]
+/// Requires the guard to be allocated with [TaskDataCategory::All].
+///
+/// `schedule_when_active` is false when an unneeded execution is aborted: the dirty transition
+/// must propagate, but the disconnected task must remain unscheduled until it becomes live again.
 pub fn make_task_dirty_internal<'e, E: ExecuteContext<'e>>(
     task: &mut E::TaskGuardImpl,
     make_stale: bool,
+    schedule_when_active: bool,
     #[cfg(feature = "task_dirty_cause")] cause: TaskDirtyCause,
```
lukesandberg: these 3 fields should probably be wrapped in an options struct to increase call site legibility

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -1174,10 +1174,14 @@ impl<'e> ExecuteContext<'e> for ExecuteContextImpl<'e> {
...
-            && task.is_gc_collectible()
-        {
-            collector(task.id());
+        if let ExecutePhase::Gc(collector) = self.phase {
+            if task.is_gc_collectible() {
+                collector(task.id());
+            } else if task.is_gc_collectible_ignoring_in_progress()
+                && let Some(InProgressState::InProgress(in_progress)) = task.get_in_progress()
+            {
+                in_progress.abort_unneeded();
```
lukesandberg: leave a comment that this should be pretty rare

also...  shouldn't we also call `collect()` on such a task?  technically it is in progress so ideally it would settle, but during GC that isn't really possible hmmm

we are just creating some garbage that won't be collected (in this cycle at least)

lukesandberg: is it possible that we would reject this later?

e.g. this would be a task with a parent_count==0 but not collectible.  We would detect this as a root which would preserve collectibility, but wouldn't we also reject it? (as an unexpected kind of root?)

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2099,128 @@ impl TurboTasksBackend {
...
+            decrease_active_counts_of_new_children(new_children, &mut ctx);
+            return Some(priority);
+        }
+        let immutable = task.immutable();
+        drop(task);
+
+        decrease_active_counts_of_new_children(new_children, &mut ctx);
+
+        if immutable {
+            // An immutable result never needs to be re-executed. This can happen when losing
```
lukesandberg: the race is

we abort
the task completes without observing it
?

i think the better solution there would be for aborts to be cleared from inprogressstate if the task completes successfully anyway

in that case we can rely on the normal bookkkeeping to tear things down (e.g. re-execute as stale or just finalize a task that will ultimately be collectible)

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2099,128 @@ impl TurboTasksBackend {
...
+            // activeness races with the final completion bookkeeping.
+            done_event.notify(usize::MAX);
+            drop(in_progress_cells);
+            return None;
+        }
+
+        // Unlike shutdown cancellation, an unneeded abortion is recoverable: discard the current
+        // execution state and make the task dirty without scheduling it while it is disconnected.
```
lukesandberg: drop the comparison to shutdown

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2114,13 +2257,23 @@ impl TurboTasksBackend {
...
                 };
             }
+            let (abort_handle, registration) = AbortHandle::new_pair();
+            abort_registration = registration;
+            let abort_handle = match &task_type {
+                TaskType::Cached(task_type) if task_type.native_fn.is_cancelable => {
+                    Some(abort_handle)
+                }
+                _ => None,
+            };
```
lukesandberg: shouldn't we only compute a handle for abortable tasks

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -1540,35 +1544,46 @@ impl<B: Backend> Executor<TurboTasks<B>, ScheduledTask, TaskPriority> for TurboT
                             InlineExecutionSpanSlot::set(&span);
 
                             async {
-                                let result = CaptureFuture::new(future).await;
+                                let result =
+                                    Abortable::new(CaptureFuture::new(future), abort_registration)
```
lukesandberg: would there be any benefit to making abort a fature of CaptureFuture? or does monomorphization mean there is no real overhead to the separation of concerns?

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -1540,35 +1544,46 @@ impl<B: Backend> Executor<TurboTasks<B>, ScheduledTask, TaskPriority> for TurboT
...
+                                        .await;
 
-                                // wait for all spawned local tasks using `local` to finish
+                                // Wait for all spawned local tasks using `local` to finish. The
+                                // main task future has already been dropped on abort, so local work
+                                // can no longer be added while this drains.
```
lukesandberg:
```suggestion
                                // Wait for all spawned local tasks using `local` to finish. 
```

no need to reference abort here

turbopack/crates/turbo-tasks/src/manager.rs
```diff
@@ -1540,35 +1544,46 @@ impl<B: Backend> Executor<TurboTasks<B>, ScheduledTask, TaskPriority> for TurboT
...
+                                                // This is safe because we waited for all local
+                                                // tasks to complete above.
+                                                raw_vc
+                                                    .to_non_local_unchecked_sync(&*this)
+                                                    .map_err(|err| err.into())
+                                            }
+                                            Ok(Err(err)) => Err(err.into()),
+                                            Err(err) => {
+                                                Err(TurboTasksExecutionError::Panic(Arc::new(err)))
+                                            }
```
lukesandberg: This is a lot of layers of `Result` wrapping

perhaps this would be the benefit of making Abort a feature of CaptureFuture, we could use a custom enum

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,137 @@ impl TurboTasksBackend {
...
+        turbo_tasks: &TurboTasks<TurboTasksBackend>,
+    ) -> Option<TaskPriority> {
+        let mut ctx = self.execute_context(turbo_tasks);
+        let mut task = ctx.task(task_id, TaskDataCategory::All);
+        let TaskTypeRef::Cached(task_type) = task.get_task_type() else {
+            unreachable!("only cached native tasks have abort handles")
+        };
+        let native_fn = task_type.native_fn;
+        let Some(in_progress) = task.take_in_progress() else {
+            // Completion or another abort callback won the race.
```
lukesandberg: is this actually possible?

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,137 @@ impl TurboTasksBackend {
...
+            return None;
+        };
+        let InProgressState::InProgress(in_progress) = in_progress else {
+            // Another state transition won the race. Preserve that state.
```
lukesandberg: ditto

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,137 @@ impl TurboTasksBackend {
...
+                reason: TaskExecutionReason::Stale,
+            });
+            debug_assert!(old.is_none(), "InProgress already exists");
+            drop(task);
+            decrease_active_counts_of_new_children(new_children, &mut ctx);
+            return Some(priority);
+        }
+        drop(task);
+
+        decrease_active_counts_of_new_children(new_children, &mut ctx);
```
lukesandberg: do we need to drop and reaquire?

can we append the decrement task to the queue and execute below on 2194?

turbopack/crates/turbo-tasks-backend/src/backend/operation/aggregation_update.rs
```diff
@@ -3188,6 +3191,16 @@ impl AggregationUpdateQueue {
         if is_empty {
             task.take_activeness();
         }
+        if is_zero && let Some(InProgressState::InProgress(in_progress)) = task.get_in_progress() {
```
lukesandberg, line 3195:
```suggestion
        // If activeness dropped to zero, no caller cares anymore, abort it if in flight.
        if is_zero && let Some(InProgressState::InProgress(in_progress)) = task.get_in_progress() {
```

turbopack/crates/turbo-tasks-backend/src/backend/operation/update_collectible.rs
```diff
@@ -12,20 +12,43 @@ use crate::{
...
+    /// Reverses a collectible update made by an aborted execution. Unlike a user-level `unemit`,
+    /// rollback may remove a collectible from a non-root emitter: `root` is required to *read* or
+    /// explicitly remove collectibles, but any task may emit them.
+    pub(in crate::backend) fn run_rollback(
+        task_id: TaskId,
+        collectible: CollectibleRef,
+        count: i32,
+        ctx: impl ExecuteContext<'_>,
+    ) {
+        Self::run_internal(task_id, collectible, count, true, ctx);
```
lukesandberg: use an inline `/* rollback= */` comment on this boolean, here and in `run`

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -1199,10 +1210,86 @@ impl<'e> ExecuteContext<'e> for ExecuteContextImpl<'e> {
...
-            && task.is_gc_collectible()
-        {
-            collector(task.id());
+        if let ExecutePhase::Gc(collector) = self.phase {
+            if task.is_gc_collectible() {
+                collector(task.id());
+            } else if task.is_gc_collectible_ignoring_in_progress()
+                && let Some(InProgressState::InProgress(in_progress)) = task.get_in_progress()
+            {
+                // This should be rare: activeness normally aborts disconnected work before GC
```
lukesandberg: at this point the gc phase is holding a global lock.... so the task in question couldn't complete even if it did get scheduled.  does this open up an opportunity for more synchronous teardown?

also, i do wonder if this is really possible.

for a task to become collectible its activeness must have already dropped or be in the middle of dropping? so perhaps we shouldn't do anything because we know that activeness will drop as soon as GC releases the lock

so it does seem like, to the extent this is possible it isn't really useful to abort here

turbopack/crates/turbo-tasks-backend/src/backend/operation/mod.rs
```diff
@@ -1199,10 +1210,86 @@ impl<'e> ExecuteContext<'e> for ExecuteContextImpl<'e> {
...
+            AbortRequestOutcome::RacedCompletion => tracing::event!(
+                name: "turbo_tasks::abort_raced_completion",
+                target: "turbo_tasks::abort",
+                tracing::Level::TRACE,
+                event = "raced_completion",
+                task_id = %task_id,
+                function = native_fn.name(),
+                trigger = reason.as_str(),
+            ),
+            AbortRequestOutcome::Accepted | AbortRequestOutcome::Duplicate => {}
```
lukesandberg: these tracing events are not useful, we don't capture and log or visualize them

what would be useful would be to tag the span for the executing task as 'aborted' (much like the inline execution tags)

if that is feasible, we should do it, but either way drop the the `tracing::event!` calls

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,167 @@ impl TurboTasksBackend {
...
+        Span::current().record("abort_trigger", abort_reason.as_str());
+        tracing::event!(
+            name: "turbo_tasks::abort_observed",
+            target: "turbo_tasks::abort",
+            tracing::Level::TRACE,
+            event = "observed",
+            task_id = %task_id,
+            function = native_fn.name(),
+            trigger = abort_reason.as_str(),
+        );
```
lukesandberg: drop the event!

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,167 @@ impl TurboTasksBackend {
...
+        let InProgressStateInner {
+            stale,
+            done_event,
+            mut new_children,
+            collectible_deltas,
+            ..
+        } = *in_progress;
+
+        // Only children that were not already connected received a speculative active-count
+        // increment during this execution.
```
lukesandberg: add `See ConnectChildOperation::run`

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,167 @@ impl TurboTasksBackend {
...
+        }
+
+        let in_progress_cells = if aborted_as_unneeded {
+            task.take_in_progress_cells()
+        } else {
+            None
+        };
+        if let Some(cells) = &in_progress_cells {
+            for state in cells.values() {
+                state.event.notify(usize::MAX);
```
lukesandberg, line 2113: document our expectation here

any task waiting on one of our inprogress cells will not be notified and then read either a stale value or a missing value

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,167 @@ impl TurboTasksBackend {
...
+            }
+        }
+
+        if !aborted_as_unneeded {
```
lukesandberg: aborted_as_unneeded is ambiguous, abort always means we don't need the execution

the name would be better as 'aborted_stale' vs 'aborted_dead' or something like that

turbopack/crates/turbo-tasks-backend/src/backend/mod.rs
```diff
@@ -2079,6 +2096,167 @@ impl TurboTasksBackend {
...
+            debug_assert!(stale, "only stale or unneeded executions may be aborted");
+            let priority = compute_stale_priority(&task);
+            let old = task.set_in_progress(InProgressState::Scheduled {
+                done_event,
+                reason: TaskExecutionReason::Stale,
+            });
+            debug_assert!(old.is_none(), "InProgress already exists");
+            drop(task);
+            self.rollback_aborted_collectibles(task_id, collectible_deltas, turbo_tasks);
+            decrease_active_counts_of_new_children(new_children, &mut ctx);
```
lukesandberg: nit: just inline this function? it is trivial... and it clearly is paralel to the subsequent code.... could we merge paths somehow?

PR #98347, opened 2026-09-08
[scripts] Move scripts (and benchmarks) off of `node-fetch`

eps1lon:
We're using Node.js versions with a built-in `fetch` implementation in all of these scripts.


.github/workflows/test_e2e_project_reset_cron.yml
```diff
@@ -35,13 +35,5 @@ jobs:
...
           package-manager-cache: false
 
-      - name: Setup pnpm
-        run: |
-          npm i -g corepack@0.31
-          corepack enable
-
-      - name: Install dependencies
-        run: pnpm install
-
```
eps1lon, line 45: `scripts/reset-project.mjs` has no more dependencies so we can skip the install.

mischnic (approved):

PR #98351, opened 2026-09-08
Add codegen-only constant export inlining to Turbopack

sokra:
### What?

Adds the experimental `turbopackInlineConstantExports` optimization for Turbopack. It inlines short, identity-free constant exports during code generation while keeping the existing analyzer-aware cross-module constants feature independent at the consumer.

The change covers named imports, static namespace members, and default expressions. Identity-bearing and unsupported values remain runtime bindings. The test matrix includes module omission, side-effect preservation, feature interactions, escaping namespaces, and both ordinary and mutual import cycles.

It also unifies exporter-side constant analysis so the analyzer-aware and codegen-only consumers share one cached export summary.

### Why?

Turbopack's existing constants feature feeds imported values into compile-time analysis, which can remove dependencies from statically-false branches. Some optimizations need Webpack-style semantics instead: substitute the value in generated output and remove an unnecessary binding edge, without changing dependency analysis in the importing module.

Keeping the consumer paths separate allows side-effect-free constant exporters to leave the bundle while preserving dependencies discovered before code generation and retaining module evaluation whenever side effects require it.

The original codegen-only implementation performed a second exporter analysis. For strict `'use turbopack: constants'` modules, that could emit duplicate or spurious `NonConstantIssue` diagnostics. Sharing the exporter summary avoids duplicate analysis and diagnostics while allowing eligible default expressions to participate consistently.

### How?

The new consumer path runs as a fallback from `Effect::ImportedBinding`, after the existing analyzer-aware constants path. It reads an eligible primitive from the shared exporter summary and emits a marked literal code-generation replacement; it does not create a binding reference or expose the value to the consumer linker. Independent evaluation references continue to govern exporter side effects.

Both consumers now query one cached exporter-analysis task. That task includes named bindings and default expressions, and strict-mode nonconstant diagnostics are emitted once while building the shared summary rather than once per consumer path.

### Verification

- `NODE_PATH=/vercel/sandbox/test-deps/node_modules cargo test -p turbopack-tests --test execution codegen_inline_exports -- --nocapture`
- `cargo test -p turbopack-tests --test snapshot comptime -- --nocapture`
- `cargo check -p turbopack-ecmascript -p turbopack -p next-core -p turbopack-tests --tests`


turbopack/crates/turbopack-ecmascript/src/references/constant_value.rs
```diff
@@ -26,20 +26,34 @@ use crate::{
 pub struct ConstantValueCodeGen {
     value: CompileTimeDefineValue,
     path: AstPath,
+    add_compile_time_marker: bool,
```
sokra: Remove that. We always want to add a marker when we did modifications to the code. Minificiation will remove it, so that's not a problem in production.

turbopack/crates/turbopack-ecmascript/src/references/constant_value.rs
```diff
@@ -37,12 +38,40 @@ impl ConstantValueCodeGen {
...
+                self.path,
+                visit_mut_expr,
+                |expr: &mut Expr| {
+                    *expr = value_to_expr(&value);
+                }
+            ));
+        }
 
-        Ok(CodeGeneration::visitors(vec![visitor]))
+        Ok(CodeGeneration::visitors(visitors))
```
sokra, line 74: Is that an preexisting bug that can also happen with the opt-in constant inlining?

sokra: In `cross_module_constants`?

PR #98353, opened 2026-09-08
[test] Run deploy tests against local JavaScript changes by default

unstubbable:
Deploy tests now use the checkout's built JavaScript packages when `NEXT_TEST_VERSION` is unset. The harness reuses the isolated-test packer, includes the tarballs in the deployment source, and installs them through relative `file:` references.

The previous published-build workflow remains available through `NEXT_TEST_VERSION`. For example, `NEXT_TEST_VERSION=canary` skips local packing and tests the published canary. Exact versions and CI preview tarball URLs remain supported. Existing CI deploy jobs already set `NEXT_TEST_VERSION`, so their package selection does not change. The harness skips local package preparation when it reuses an existing deployment or invokes a custom deploy script, preserving those workflows.

Staged monorepo peer dependencies use matching local versions so npm accepts prerelease builds without conflicting direct-package overrides. Fixture overrides retain their nested rules, including qualified identity overrides with or without an explicit `.` entry. Configuration updates preserve linked source files.

Deployments of local JavaScript packages use published SWC binaries matching the local Next.js version. They do not include locally built native binaries because those may target a different platform.

**Verification**

```
VERCEL_TURBOPACK_TEST_TEAM=vtest314-next-turbo-e2e-tests \
  pnpm test-deploy-turbo test/e2e/app-dir/third-parties/basic.test.ts
```

```
VERCEL_TEST_TEAM=vtest314-next-turbo-e2e-tests \
  pnpm test-deploy-webpack test/e2e/app-dir/third-parties/basic.test.ts
```

Both deploy modes passed. Additional temporary checks, removed afterward, verified that:

- Deployments install the local JavaScript packages and use matching local versions for monorepo peers.
- Qualified npm overrides preserve a nested `sharp@0.34.5` pin, both with and without an explicit `.` entry. The check resolved `sharp` from Next's own dependency context.
- Archive preparation leaves checkout manifests and tarballs unchanged. Repacked archives preserve other package files, permissions, links, and external peer declarations.

### DIFF
--- a/test/lib/next-modes/next-deploy.ts
+++ b/test/lib/next-modes/next-deploy.ts
@@ -3,12 +3,15 @@ import path from 'path'
 import dns from 'dns'
 import execa from 'execa'
 import fs from 'fs-extra'
+import { load, dump } from 'js-yaml'
+import * as tar from 'next/dist/compiled/tar'
 import { NextInstance, type NextInstanceOpts } from './base'
 import * as projectEnv from '../../../scripts/reset-project.mjs'
 import { Span } from 'next/dist/trace'
 import { setTimeout } from 'timers/promises'
 import { FileRef } from '../e2e-utils'
 import { PROXY_HOST_MAP_ENV_KEY } from '../browsers/launch'
+import { packPackages } from '../create-next-install'
 
 export class NextDeployInstance extends NextInstance {
   private _cliOutput: string
@@ -327,6 +330,10 @@ export class NextDeployInstance extends NextInstance {
       return
     }
 
+    if (!process.env.NEXT_TEST_VERSION) {
+      await this.prepareLocalPackages(parentSpan)
+    }
+
     // Original Vercel CLI deployment logic
     // ensure Vercel CLI is installed
     try {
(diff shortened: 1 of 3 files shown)


test/lib/next-modes/next-deploy.ts
```diff
@@ -531,6 +538,252 @@ export class NextDeployInstance extends NextInstance {
...
+   * This method stages local JavaScript tarballs and rewrites the fixture's
+   * dependencies and overrides to relative `file:` references. It adds
+   * published SWC dependencies because locally built native binaries may target
+   * a different platform than the remote build.
+   */
+  private async prepareLocalPackages(parentSpan: Span): Promise<void> {
+    const packagePaths = process.env.NEXT_TEST_PKG_PATHS
+      ? new Map<string, string>(JSON.parse(process.env.NEXT_TEST_PKG_PATHS))
+      : await packPackages(parentSpan)
+    for (const name of ['next', '@next/env']) {
```
eps1lon, line 569: This should go through the dependencies of this test instead e.g. when testing `@next/third-parties`. `next` would need an explicit check because we implicitly add it as a dependency.

eps1lon (approved):

PR #98355, opened 2026-09-08
refactor(turbo-rcstr): drop unsupported 16-byte pointer width

sokra:
## What?

Removes turbo-rcstr's 16-byte/128-bit pointer representation and the unused `atom_size_64` feature plumbing. Unsupported pointer widths now fail with an error that names the supported 32-bit and 64-bit widths.

## Why?

This follows up on review feedback from #97577. The previous negative cfg made unknown pointer widths fall through to the 64-bit representation, while the 16-bit implementation and `atom_size_64` features had no users. Making both supported widths explicit keeps representation assumptions visible and produces a useful failure for unsupported targets.

## How?

The narrow tagged-value layout is now specific to 32-bit targets, and the pointer-sized layout is specific to 64-bit targets. `atom_size_128` remains unchanged. The dead feature declarations are removed from turbo-rcstr and its proc-macro crate.

## Verification

- `cargo test -p turbo-rcstr --lib`
- `cargo test -p turbo-rcstr --lib --target wasm32-wasip1-threads`
- `cargo check -p next-napi-bindings`
- `cargo check -p next-napi-bindings --target wasm32-wasip1-threads`
- `cargo clippy -p turbo-rcstr -p turbo-rcstr-macros -- -D warnings`
- `cargo clippy -p turbo-rcstr -p turbo-rcstr-macros --target wasm32-wasip1-threads -- -D warnings`


lukesandberg (approved):

bgw (approved):
RIP my future 128-bit CPU architecture

PR #98361, opened 2026-09-08
Adds .resolve to loaders api

jimmyhmiller:
This fully fixes handlebars loader in my tests and a number of other loaders depend on this as well. The implementation for getResolve is a bit different in turbopack, but this should be semantically the same.


test/e2e/app-dir/turbopack-loader-resolve/turbopack-loader-resolve.test.ts
```diff
@@ -0,0 +1,12 @@
+import { nextTestSetup } from 'e2e-utils'
+
+describe('turbopack-loader-resolve', () => {
```
mischnic: Can we call this `test/e2e/app-dir/webpack-loader-resolve` instead

The test isn't Turbopack specific. It's just the Webpack loader API

mischnic (approved):

PR #98372, opened 2026-09-08
[turbopack] Defer lazy import canonicalization

jimmyhmiller:
Previously I had a bunch of special cases that felt rather ad-hoc. Part of this was in attempt to isolate some changes. Here I think I have a solution that is a bit less ad-hoc. It did require refactoring some code into this "canonicalization" setup. So that when a lazy proxy is forced, it can be transformed into exactly the code it would have been had their been no lazy proxy (modulo a proxy of course).

This let me handle more cases I wasn't handling before. I've added more tests. Trying out v0/chat. I've looked at visualizations of graphs in various scenarios and this matches my intuition of how the feature ought to work.

One thing that isn't as clear to me is the exact boundaries of where code belongs. I sadly had to move things between crates for dependency sake. I think I got an okay clean setup. But not 100% sure it is right.

I think the end result is way less ad-hoc. I know Niklas didn't want this new LazyDynamic enum here. But after trying some variations and running into problems with duplication, this refactoring out of the "canonicalization" came to me and I think the end result is decently clean.

### DIFF
--- a/turbopack/crates/turbopack/src/lib.rs
+++ b/turbopack/crates/turbopack/src/lib.rs
@@ -50,16 +50,10 @@ use turbopack_css::{CssModule, EcmascriptCssModule};
 use turbopack_ecmascript::{
     AnalyzeMode, EcmascriptInputTransforms, EcmascriptModuleAsset, EcmascriptModuleAssetType,
     EcmascriptOptions,
-    chunk::EcmascriptChunkPlaceable,
-    module_fragments::part::module::EcmascriptModulePartAsset,
-    references::{
-        FollowExportsResult,
-        external_module::{CachedExternalModule, CachedExternalTracingMode, CachedExternalType},
-        follow_reexports,
-    },
-    rename::module::EcmascriptModuleRenameModule,
-    side_effect_optimization::{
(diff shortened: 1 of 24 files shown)


turbopack/crates/turbopack/src/lib.rs
```diff
@@ -200,46 +198,32 @@ async fn apply_module_type(
...
+
+                if is_lazy_dynamic_import {
+                    let target = LazyCompilationTarget::deferred(*module, canonicalization)
+                        .to_resolved()
+                        .await?;
+                    ResolvedVc::upcast(
+                        LazyCompilationProxyModule::new(*target)
+                            .to_resolved()
+                            .await?,
+                    )
```
mischnic, line 219: Can't this just be `LazyCompilationProxyModule::new_deferred(module, canonicalization)`
and `LazyCompilationProxyModule::new_direct(module)`
and `LazyCompilationTarget` just stays an internal implementation detail of that module?

jimmyhmiller: I did a stack fold with gh stacks. I thought it would do something smart with this PR. But it seems to have just put the changes over on the other PR. I will close this and address this there.

PR #98380, opened 2026-09-08
Added notice for root-params under Route Groups

PacifiK2460:
Added a warning for root-params used inside Route Groups. When a root parameter exists within a Route Group, the compiler is unable to generate and export the corresponding parameters. As a result, the application cannot consume those parameters correctly and an unclear compiler error is displayed.


icyJoseph:
I do believe this is a bug, tracked here: https://github.com/vercel/next.js/issues/96756

PR #98382, opened 2026-09-08
Turbopack: fix output tracing include glob traversal

sokra:
### What?

Make Turbopack's `outputFileTracingIncludes` file patterns root-relative and targeted during output tracing. Reject unanchored partial globs when code attempts directory-prefix matching. Add production coverage using a real pnpm-installed WASM package, plus unit coverage for anchored directory pruning and include-pattern expansion.

### Why?

Turbopack previously treated include file patterns as unanchored partial matches. The directory-prefix matcher could therefore accept every descendant after matching the first literal segment: an exact WASM include under `node_modules/.pnpm` caused 2,093 `read_glob_inner` executions instead of traversing only its 5 directory prefixes.

Unanchored patterns also cannot soundly answer whether a directory may contain a future match, so allowing `contains: true` with directory-prefix matching risks both over-traversal and missed files. Include behavior also differed from webpack and the documented project-root-relative semantics.

### How?

Build anchored alternatives for each include pattern: one for the direct match and one for descendants of a matched directory. Anchoring lets `read_glob` prune unrelated directories, while the recursive alternative preserves existing behavior for matched directories and directory symlinks, including their resolved target contents.

`can_match_in_directory` now fails immediately for `contains: true`, while direct partial matching remains supported. Route-key and exclusion matching continue to use direct partial matches and are unchanged.

The production fixture resolves `lightningcss-wasm/lightningcss_node.wasm` from a real `lightningcss-wasm@1.28.2` dependency, avoiding assumptions about pnpm's store naming or patch hashes.

### Verification

- `CARGO_INCREMENTAL=0 pnpm build-all`
- `cargo fmt -p turbo-tasks-fs -p next-api -- --check`
- `cargo test -p turbo-tasks-fs glob` (84 passed)
- `cargo test -p next-api` (17 passed)
- `pnpm test-start-turbo test/production/app-dir/output-file-tracing-includes-read-glob/output-file-tracing-includes-read-glob.test.ts test/production/build-trace-extra-entries-turbo/build-trace-extra-entries-turbo.test.ts test/production/build-trace-extra-entries-monorepo/build-trace-extra-entries-monorepo.test.ts` (3 passed)
- `pnpm test-start-webpack test/production/app-dir/output-file-tracing-includes-read-glob/output-file-tracing-includes-read-glob.test.ts`
- The shared fixture asserts that both bundlers include a root-relative exact match and exclude a nested file with the same suffix
- Measured the exact reproduction at 5 include-specific `read_glob_inner` executions, down from 2,093

### DIFF
--- a/crates/next-api/src/nft.rs
+++ b/crates/next-api/src/nft.rs
@@ -181,18 +181,30 @@ pub async fn trace_endpoint(
 /// a directory (including a directory symlink) matched by the user pattern without making the
 /// original pattern an unanchored partial match.
 fn include_glob_pattern(globs: &[&str]) -> RcStr {
-    let mut alternatives = Vec::with_capacity(globs.len() * 2);
-    for glob in globs {
-        alternatives.push((*glob).to_owned());
+    if globs.contains(&"**") {
+        return rcstr!("**");
+    }
+    if let [glob] = globs
+        && glob.ends_with("/**")
+    {
+        return RcStr::from(*glob);
+    }
+
+    let mut pattern = String::new();
+    pattern.push('{');
+    for (index, glob) in globs.iter().enumerate() {
+        if index > 0 {
+            pattern.push(',');
+        }
+        pattern.push_str(glob);
         if *glob != "**" && !glob.ends_with("/**") {
-            alternatives.push(format!("{glob}/**"));
+            pattern.push(',');
+            pattern.push_str(glob);
+            pattern.push_str("/**");
         }
     }
-    if alternatives.len() == 1 {
-        alternatives.pop().unwrap().into()
-    } else {
-        format!("{{{}}}", alternatives.join(",")).into()
-    }
+    pattern.push('}');
+    pattern.into()
 }
 
 /// Apply outputFileTracingIncludes patterns to find additional files
@@ -642,13 +654,16 @@ mod include_glob_tests {
 
     #[test]
     fn include_glob_supports_braces_and_multiple_patterns() {
-        let glob = glob(&["assets/{one,two}", "config/*.json"]);
+        let combined = glob(&["assets/{one,two}", "config/*.json"]);
 
-        assert!(glob.matches("assets/one"));
-        assert!(glob.matches("assets/two/nested.txt"));
-        assert!(glob.matches("config/runtime.json"));
-        assert!(!glob.matches("config/nested/runtime.json"));
-        assert!(!glob.matches("assets/three"));
+        assert!(combined.matches("assets/one"));
+        assert!(combined.matches("assets/two/nested.txt"));
+        assert!(combined.matches("config/runtime.json"));
+        assert!(!combined.matches("config/nested/runtime.json"));
+        assert!(!combined.matches("assets/three"));
+
+        let match_all = glob(&["config/*.json", "**"]);
+        assert!(match_all.matches("any/deeply/nested/file"));
     }
 }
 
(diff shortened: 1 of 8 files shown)


test/production/app-dir/output-file-tracing-includes-read-glob/app/next.config.ts
```diff
@@ -0,0 +1,11 @@
+import type { NextConfig } from 'next'
+
+const nextConfig: NextConfig = {
+  outputFileTracingIncludes: {
+    '/*': [
+      '../node_modules/.pnpm/lightningcss-wasm@1.28.2_patch_hash=d2d6e72f68b8ae36c669d00a13784346180c35c614b37a4412d651117c518ecf/node_modules/lightningcss-wasm/lightningcss_node.wasm',
```
sokra: Avoid this very specific path and use `require.resolve("lightningcss-wasm/lightningcss_node.wasm")` to get it. Same in the test case

crates/next-api/src/nft.rs
```diff
@@ -183,6 +177,24 @@ pub async fn trace_endpoint(
     .await
 }
 
+/// Build root-relative include alternatives. The recursive alternative preserves the contents of
+/// a directory (including a directory symlink) matched by the user pattern without making the
+/// original pattern an unanchored partial match.
+fn include_glob_pattern(globs: &[&str]) -> RcStr {
+    let mut alternatives = Vec::with_capacity(globs.len() * 2);
```
sokra: Instead of creating a vec and joining it, write to a string directly. That avoids copying the bytes twices.

turbopack/crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -92,6 +95,10 @@ impl Glob {
     // Returns true if the glob might match a filename underneath this `path` where the
     // path represents a directory.
     pub fn can_match_in_directory(&self, path: &str) -> bool {
+        assert!(
+            !self.opts.contains,
+            "Glob::can_match_in_directory cannot be used when GlobOptions::contains is true"
+        );
```
bgw: Should it just always return `true`? I don't see why it needs to panic.

turbopack/crates/turbo-tasks-fs/src/glob.rs
```diff
@@ -77,7 +77,10 @@ pub struct GlobOptions {
     /// Allows glob to match any part of the given string(s).
     /// NOTE: this means that a pattern like `node_modules/package_name` with `contains:true` will
     /// match `foo_node_modules/package_name_bar` If you want to match a _directory_ named
-    /// `node_modules/package_name` you should use `**/node_modules/package_name/**`
+    /// `node_modules/package_name` you should use `**/node_modules/package_name/**`.
+    ///
+    /// A partial match cannot safely determine whether a directory might contain a match, so this
+    /// option cannot be used with [`Glob::can_match_in_directory`].
     pub contains: bool,
```
bgw, line 84: Do we have any remaining uses of `contains`? If not, maybe we should just delete it.

crates/next-api/src/nft.rs
```diff
@@ -183,6 +177,36 @@ pub async fn trace_endpoint(
...
+        pattern.push_str(glob);
+        if *glob != "**" && !glob.ends_with("/**") {
+            pattern.push(',');
+            pattern.push_str(glob);
+            pattern.push_str("/**");
+        }
+    }
+    pattern.push('}');
+    pattern.into()
+}
```
bgw, line 204: This seems like it may not be very generalized. Where do we get the globs from? Can we be sure that they fit a specific format?

bgw (approved):

crates/next-api/src/nft.rs
sokra: They are from user config, so we don't have a lot of control over them.

mischnic:
> Include behavior also differed from webpack and the documented project-root-relative semantics.

I don't see any tests for that though?

PR #98397, opened 2026-09-09
Fix `experimental.inlineCss` duplicating the stylesheet into every RSC and segment payload

### LINKED ISSUE #95141: experimental.inlineCss duplicates the full CSS bundle into every RSC/segment payload (standalone build-output bloat)

https://github.com/mikaelh/next-inlinecss-rsc-dup-repro

### To Reproduce

1. `npm install` (pulls `next@canary`) then `npm run build` on the linked repo. It is a minimal App Router app: `output: "standalone"`, `experimental.inlineCss: true`, a 155 KB global stylesheet imported in the root layout, and one `[slug]` route with `generateStaticParams` → 60 tiny static pages.
2. Inspect `.next/standalone/.next/server/app`, e.g. `grep -rl '\.u1000{' .next/standalone/.next/server/app | wc -l` and count occurrences per file.

**Current** (verified on `next@16.3.0-canary.66`): with `inlineCss`, the whole stylesheet is inlined into each page's HTML `<style>` **and also serialized into the page's RSC payloads** — the inline flight inside the `.html`, the `page.rsc`, and each `.segments/*.segment.rsc`.

Observed in the repro for one trivial static page (`a1.html`, **411 KB** for "tiny content"): 1 `<style>` block, but the CSS marker rule appears **3×** (`<style>` + flight). Across the build the 155 KB CSS is embedded in **62 `.html`**, **186 `.rsc`** (of which **124** are `*.segments/*.segment.rsc`) → `.next/server/app` = **69 MB** for 60 pages of "tiny content". So each prerendered page stores the stylesheet ~3–4× (HTML `<style>` + HTML flight + `page.rsc` + each `segment.rsc`); build output scales with `CSS_size × pages × ~4`.

Amplifier: moving the `@import` from the root layout into a nested layout (route group) makes the per-segment copies multiply (~2× → ~5× per page in a larger app), so per-route CSS scoping is counter-productive under `inlineCss`. On a real content site (~700 prerendered pages, 172 KB CSS) the duplicated CSS is ~205 MB of a 453 MB `server/app`; moving CSS to a nested layout took `server/app` from 453 MB to 868 MB.

**Expected:** inlined CSS referenced/deduplicated once per prerendered page rather than copied into every RSC/segment payload. The segment-cache / client-navigation RSC payloads don't need the full stylesheet embedded — on client navigation the browser already has the styles applied. Keep inlining into the initial HTML (the FCP/LCP win) without duplicating across the RSC.

```
Operating System: Linux
Binaries:
  Node: 24.x
Relevant Packages:
  next: 16.3.0-canary.66
  react: 19.2.0
  react-dom: 19.2.0
Next.js Config:
  output: "standalone"
  experimental.inlineCss: true
```

Output (Standalone), Partial Prerendering (PPR)

next build (local)

### Additional context

`inlineCss` is otherwise a real FCP/LCP win, so disabling it is an unwanted trade-off — the request is to keep the initial-HTML inline but stop duplicating the bundle across the RSC/segment payloads. Measured via `du` + counting the inlined CSS chunk across `.html` / `.rsc` / `.segments/*.segment.rsc`. (Supersedes #95140, which was auto-closed for a parsing issue — the reproduction link is now in the correct field.)

martinfrancois:
Fixes #95141

**TL;DR** `experimental.inlineCss` serialised the whole stylesheet into the RSC payload once per segment. This keeps the `<style>` in the HTML and replaces the text in the payload with a reference. The issue's reproduction goes from 68 MB to 11 MB of `server/app`; a production home document from 898 KB to 650 KB. No change to what the browser gets styled with.

### What?

`renderCssResource` emitted `<style>{text}</style>` per segment that referenced a stylesheet, so Flight wrote the text into the flight data inlined in the HTML, into `page.rsc`, and into every `.segments/*.segment.rsc` file. It now emits a client component reference, `InlinedStylesheet`, carrying only `href`, `precedence`, `nonce` and `crossOrigin`. The server resolves the text and renders the hoisted `<style>` once per document; the client adopts that element on hydration, or fetches the file through a `<link>` when the stylesheet is not in the document yet.

### Why?

Every `inlineCss` page ships its stylesheet three or more times, and the count grows with segment depth, so the apps the flag targets pay the most: [a Cache Components storefront measured 4 copies per document](https://github.com/vercel/next.js/issues/95141#issuecomment-4870375528). The bytes are parsed on every load, inflate standalone output, and get cached by CDNs.

[Issue reproduction](https://github.com/mikaelh/next-inlinecss-rsc-dup-repro), 60 static pages, 155 KB stylesheet, `output: "standalone"`:

| | before | after |
|---|---:|---:|
| `server/app` | 68 MB | 11 MB |
| copies of the stylesheet in `server/app` | 503 in 251 files | 69 in 65 files (the `<style>` of each HTML file) |
| `a1.html` | 411 KB, 3 copies | 141 KB, 1 copy |
| `a1.rsc` | 2 copies | 0 |

www.fmartin.ch 117 KB stylesheet, served documents:

| | before | after |
|---|---:|---:|
| home | 898 KB, 3 copies | 650 KB, 1 copy |
| events | 715 KB, 3 copies | 471 KB, 1 copy |
| stylesheet requests on a fresh load | 0 | 0 |
| styled on load, on client navigation, on the 404 page | yes | yes |

### How?

| where `InlinedStylesheet` renders | it renders | why that is correct |
|---|---|---|
| server (Fizz) | `<style href precedence>{text}</style>`, text from a per-render lookup over the client reference manifest, provided by the SSR root through `InlinedCssContext` | Fizz hoists by href and emits one `<style>` per document, in whichever branch it reaches first, so not-found and error branches stay styled |
| client, stylesheet in the document | empty `<style href precedence />` | hydration adopts the existing element by `data-href` and never reconciles a resource's children |
| client, stylesheet not in the document | `<link rel="stylesheet" precedence>` | the file is fetched, the path RSC-request navigations always took |

Unchanged: RSC requests for client navigations (already `<link>`), dev mode (never inlines), the pages router, and how CSS is collected per segment.

Trade-off: a prefetched segment for a nested layout no longer embeds that layout's stylesheet, so the first client navigation into it loads the stylesheet file instead (one cached request, and React holds the navigation until it has loaded, so nothing paints unstyled). This is the path live RSC navigations already take; the root stylesheet is never refetched.

### Reading order

1. `packages/next/src/server/app-render/render-css-resource.tsx`: the switch from `<style>` to the reference, plus `createInlinedCssLookup` and the href helper both sides share.
2. `packages/next/src/client/components/inlined-stylesheet.tsx`: the three cases above, with the hydration invariant in the docblock.
3. `packages/next/src/server/app-render/app-render.tsx`: `lookupInlinedCss` on the render context, the provider in `App` and `ErrorApp`, and the prop threaded to their call sites like `nonce` and `images` already are.
4. `packages/next/src/shared/lib/inlined-css-context.shared-runtime.ts` and the two vendored-context files: the same shape as `HooksClientContext`.
5. `entry-base.ts` exports the component next to `HTTPAccessFallbackBoundary`; `get-asset-query-string.ts` only narrows a parameter type.

### Tests

`test/e2e/app-dir/app-inline-css`, extended:
- stylesheet once in the HTML and absent from the served prefetch payload and `index.rsc` (fails on the previous renderer)
- hydration adds no stylesheet link; not-found page styled; route stylesheet applied on client navigation; edge-runtime page

`test/e2e/app-dir/app-inline-css-segment-cache`, new, with `cacheComponents`:
- no stylesheet text in any `.rsc` or `.segments/*.segment.rsc` file, nor in the served `/_tree`, `/_full` and page-segment prefetch payloads
- PPR shell with a resumed `cookies()` hole is styled
- nested layout stylesheet applies on client navigation and appears once on a direct load

Both suites pass in webpack and Turbopack production modes:

```
pnpm test-start test/e2e/app-dir/app-inline-css test/e2e/app-dir/app-inline-css-segment-cache
pnpm test-start-turbo test/e2e/app-dir/app-inline-css test/e2e/app-dir/app-inline-css-segment-cache
```

6 of the 11 tests fail on the previous renderer; the other five guard the paths the new mechanism depends on.

To reproduce the numbers: build https://github.com/mikaelh/next-inlinecss-rsc-dup-repro with this branch and run `grep -ro '\.u1000{' .next/standalone/.next/server/app | wc -l`.


martinfrancois:
I resolved the conflicts and pushed `64919b718527`, preserving the upstream metadata entry point and the stylesheet component export. The JavaScript workspace build, including type generation, passes.

The new workflows are waiting for maintainer approval. @gaojude, given your CSS-inlining work in #73182, would you review this and approve those runs? CI still needs to validate the native build and Turbopack behavior.

PR #98408, opened 2026-09-09
Turbopack: forward module graph, entrypoint write, and emit spans to the JS trace

mischnic:
Bildschirmfoto 2026-09-09 um 16 32 57

## Summary

Adds three Turbopack build-phase spans to the JS-side trace (`.next/trace`, `.next/trace-build`, and trace uploads), forwarded from Rust via `turbo_tasks.send_compilation_event` like the existing `turbopack-persistence` span:

- `turbopack-module-graph` — whole-app module graph computation (`whole_app_module_graph_operation`)
- `turbopack-write-entrypoints` — computing all output assets for the entrypoints write phase (`all_entrypoints_write_to_disk_operation`)
- `turbopack-emit` — writing all output assets to disk (`emit_all_output_assets_once_with_issues_operation`)


lukesandberg (approved):
in the case of invalidations these might all log multiple times which is fine and would be a useful signal

PR #98420, opened 2026-09-09
fix(next-rspack): stabilize rebuilds and React 18 builds

### LINKED ISSUE #98213: [next-rspack] `next dev` aborts with a Rust panic when the error overlay or browser-log forwarding source-maps a frame during a rebuild

https://github.com/Dwlad90/next-rspack-stale-compilation-repro

### To Reproduce

```bash
npm install
npm run generate   # 5000 tiny modules so a rebuild takes long enough to be hit
npm run repro      # boots `next dev -p 3456`, edits a module, posts stack frames, repeats
```

The driver posts to `POST /__nextjs_original-stack-frames` (what the error overlay does) while
a rebuild is in flight. No browser is needed. Result on my machine:

Two runs from a clean `.next`: `REPRODUCED after 1 trial(s)` both times, `attempted to read from stolen value: ...BuildModuleGraphArtifact`. Control with the stack-frame posting switched off (`POSTERS=0`): green through 8 trials with the same edits, so the source-map request is load-bearing, not the edit alone.

Without the driver: `next dev`, open the page in a browser whose console shows an error (any
`console.error` in a client component), then append a comment to an imported file every
second. Either the overlay's stack-frame request or the forwarded console error makes the
server source-map a stack during the rebuild.

**Current:** the dev server dies and has to be restarted.

```
Panic occurred at runtime. Please file an issue on GitHub with the panic info and backtrace below: https://github.com/web-infra-dev/rspack/issues

thread '<unnamed>' panicked at /Users/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rspack_core-0.100.4/src/utils/steal_cell.rs:60:26:
attempted to read from stolen value: rspack_core::artifacts::module_ids_artifact::ModuleIdsArtifact
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The stolen value is sometimes `BuildModuleGraphArtifact` instead.

**Expected:** a stack frame that cannot be mapped during a rebuild comes back unmapped. The dev
server keeps running.

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin 25.6.0
Binaries:
  Node: 24.11.0
  npm: 11.x
  pnpm: 11.x
Relevant Packages:
  next: 16.3.3 (also 16.3.4 and 16.4.0-canary.15 pin the same @next/rspack-core 1.0.3)
  next-rspack: 16.3.3
  @next/rspack-core: 1.0.3 (@next/rspack-binding 1.0.3, rspack_core 0.100.4)
  react: 19.2.0
  react-dom: 19.2.0
  typescript: 7.0.2
Next.js Config:
  output: N/A
```

Developer Experience, Rspack

next dev (local)

#### Cause

`packages/next/src/server/dev/middleware-webpack.ts` reads `Stats.compilation` from request
handlers, outside the compile lifecycle:

- `getOverlayMiddleware` → `POST /__nextjs_original-stack-frames` → `getOriginalStackFrames`
- `browser-logs/receive-logs.ts` → `getSourceMappedStackFrames` → `mapFramesUsingBundler` →
  `getOriginalStackFrames`

Both reach `getSourceMapFromCompilation` → `getModuleById`, which does `compilation.modules`
and then `chunkGraph.getModuleId(module)` for every module until one matches.

Under webpack that is fine: a kept `Stats.compilation` is a JavaScript snapshot. Under rspack it
is not. The binding hands JS a raw pointer to `Compiler.compilation`, and `Compiler::rebuild`
replaces that struct in place (`std::mem::replace`), so the kept object aliases the compilation
being rebuilt on tokio threads. `compilation.modules` dereferences rspack's
`StealCell<BuildModuleGraphArtifact>` (stolen during make) and `getModuleId` dereferences
`StealCell<ModuleIdsArtifact>` (stolen during the module-ids hooks). Whichever pass the rebuild
is in is the artifact that panics, and the binary aborts.

The rspack side is filed separately (web-infra-dev/rspack#15451): the stale handle should fail
closed. This issue is about the host side, which is independently fixable and would also stop
the abort today.

#### Evidence

I wrapped every native binding accessor from JS and logged any call made while a compiler was
between `watchRun` and `done` with no compile hook on the stack. The last line before every
panic was `getModuleId` or `compilation.modules` from `getModuleById`, via one of the two routes
above. In one window between a client `thisCompilation` and the panic there were 1,096,130
`getModuleId` calls from that scan. Making the stale `Compilation` object throw kept the same
loop green through 41 trials; the baseline panicked at trials 1, 6 and 18.

#### Two smaller problems on the same path

1. `logging.browserToTerminal: false` does not stop the forwarder from source-mapping. `handleLog`
   still processes `formatted-error` / `any-logged-error` entries whenever the file logger is
   enabled, and the file logger is enabled by `experimental.mcpServer`, which defaults to `true`.
   So a project that has switched browser logs off still runs this code on every browser error.
2. `getModuleById` is a linear scan with one native call per module per frame. With a few
   thousand modules and a 40-frame stack that is well over 100k native calls for one request.

#### Fix that works for us

Skip the lookup while the owning compiler is rebuilding. The scan is synchronous and rebuilds
only start from JavaScript, so the check is not racy:

```ts
function getModuleById(id: string | undefined, compilation: webpack.Compilation) {
  // Under rspack a compilation kept from an earlier build aliases the one being rebuilt.
  if (compilation.compiler?.watching?.running === true) return undefined
  const { chunkGraph, modules } = compilation
  return [...modules].find((module) => chunkGraph.getModuleId(module) === id)
}
```

Both rspack's and webpack's `Watching` expose `running`. We carry this as a `pnpm patch` and
the panic has not recurred; a frame requested mid-rebuild comes back unmapped and maps on the
next request.

`packages/next/src/server/dev/middleware-webpack.ts` on `canary` (2026-09-03) still has the
unguarded `getModuleById`.

sokra:
### What?

Stabilize the Next.js–Rspack integration in the two shared failure paths recurring across the `v16.4.0-canary.11` through `.22` release runs:

- avoid reading Rspack compilation artifacts while a dev rebuild is running or after compilation has failed;
- allow React 18 Pages Router edge builds to tree-shake unused React 19-only cache code without Rspack rejecting the export first;
- cover the dev-server liveness behavior after middleware error frames are mapped.

### Why?

A cross-check of all 27 workflow attempts found 58 native Rspack panics—54 stolen `BuildModuleGraphArtifact`, two stolen `ModuleIdsArtifact`, and two missing code-generation entries. These were distinct from the npm package-availability failures and the connection-refused/uninitialized-instance errors that followed a compiler crash.

The dev panics happen because Rspack reuses its native compilation allocation during rebuilds. Next.js retains the previous `Stats.compilation` for source mapping, so an overlay or browser-log request can otherwise read an artifact currently owned by a compilation pass. Failed compilations can similarly expose no code-generation result. In both cases an unavailable source map should degrade to an unmapped frame, not terminate the dev server.

Separately, every release attempt contained the same React 18 production failure: `next/cache` pulls in `dedupe-fetch` for a Pages Router edge build, and Rspack validates the unused `React.cache` member before tree shaking. Keeping that lookup dynamic matches webpack's successful handling of the same supported React 18 build.

### How?

The source-map lookup now checks Rspack's watcher/compilation state before traversing the module graph. This keeps webpack behavior unchanged and prevents request-time code from racing Rspack's compilation lifecycle.

Rspack's parser configuration now treats missing imported exports as warnings, matching webpack's behavior while preserving Rspack's stricter handling for re-export conflicts. Shared runtime code remains unchanged, so webpack and Turbopack are unaffected.

The regression exercises real middleware error-overlay mapping and then verifies the dev server remains responsive. With the guard removed, the retained test deterministically panics on the stolen build-module-graph artifact.

### Verification

- `NEXT_TEST_REACT_VERSION=18.3.1 __NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true NEXT_RSPACK=1 NEXT_TEST_USE_RSPACK=1 pnpm test-start-rspack test/e2e/app-dir/app-static/app-static-custom-handler.test.ts`
- `NEXT_TEST_REACT_VERSION=18.3.1 __NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true TURBOPACK_DEV=1 NEXT_RSPACK=1 NEXT_TEST_USE_RSPACK=1 pnpm test-dev-rspack test/development/app-dir/server-navigation-error/server-navigation-error.test.ts`
- `NEXT_TEST_REACT_VERSION=18.3.1 __NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true TURBOPACK_DEV=1 NEXT_RSPACK=1 NEXT_TEST_USE_RSPACK=1 pnpm test-dev-rspack test/development/acceptance-app/rsc-build-errors.test.ts test/development/pages-dir/client-navigation/rendering.test.ts`
- `NEXT_TEST_REACT_VERSION=18.3.1 __NEXT_EXPERIMENTAL_STRICT_ROUTE_TYPES=true pnpm test-dev-webpack test/development/app-dir/server-navigation-error/server-navigation-error.test.ts`
- `pnpm --filter=next types`

Fixes #98213

### DIFF
--- a/packages/next/src/server/lib/dedupe-fetch.ts
+++ b/packages/next/src/server/lib/dedupe-fetch.ts
@@ -5,6 +5,11 @@ import * as React from 'react'
 import { cloneResponse } from './clone-response'
 import { InvariantError } from '../../shared/lib/invariant-error'
 
+// Keep this lookup dynamic. Rspack validates namespace-member exports before
+// tree shaking, so a React 18 Pages Router edge bundle that only re-exports
+// another `next/cache` API would otherwise fail on this unused React 19 API.
+const ReactCache = Reflect.get(React, 'cache') as typeof React.cache
+
 const simpleCacheKey = '["GET",[],null,"follow",null,null,null,null]' // generateCacheKey(new Request('https://blank'));
 
 // Headers that should not affect deduplication
@@ -42,7 +47,7 @@ type CacheEntry = [
 ]
 
 export function createDedupeFetch(originalFetch: typeof fetch) {
-  const getCacheEntries = React.cache(
+  const getCacheEntries = ReactCache(
     // eslint-disable-next-line @typescript-eslint/no-unused-vars -- url is the cache key
     (url: string): CacheEntry[] => []
   )
--- a/packages/next/src/server/lib/dedupe-fetch.ts
+++ b/packages/next/src/server/lib/dedupe-fetch.ts
@@ -5,11 +5,6 @@ import * as React from 'react'
 import { cloneResponse } from './clone-response'
 import { InvariantError } from '../../shared/lib/invariant-error'
 
-// Keep this lookup dynamic. Rspack validates namespace-member exports before
-// tree shaking, so a React 18 Pages Router edge bundle that only re-exports
-// another `next/cache` API would otherwise fail on this unused React 19 API.
-const ReactCache = Reflect.get(React, 'cache') as typeof React.cache
-
 const simpleCacheKey = '["GET",[],null,"follow",null,null,null,null]' // generateCacheKey(new Request('https://blank'));
 
 // Headers that should not affect deduplication
@@ -47,7 +42,7 @@ type CacheEntry = [
 ]
 
 export function createDedupeFetch(originalFetch: typeof fetch) {
-  const getCacheEntries = ReactCache(
+  const getCacheEntries = React.cache(
     // eslint-disable-next-line @typescript-eslint/no-unused-vars -- url is the cache key
     (url: string): CacheEntry[] => []
   )
--- a/packages/next/src/build/webpack-config.ts
+++ b/packages/next/src/build/webpack-config.ts
@@ -2331,6 +2331,9 @@ export default async function getBaseWebpackConfig(
   webpack5Config.module!.parser = {
     javascript: {
       url: 'relative',
+      // Rspack defaults missing imported exports to errors, unlike webpack.
+      // Preserve compatibility for imports removed by tree shaking.
+      ...(isRspack ? { importExportsPresence: 'warn' } : undefined),
     },
   }
   webpack5Config.module!.generator = {
--- a/packages/next/src/server/dev/middleware-webpack.ts
+++ b/packages/next/src/server/dev/middleware-webpack.ts
@@ -69,6 +69,18 @@ function getModuleById(
   id: string | undefined,
   compilation: webpack.Compilation
 ) {
+  // Rspack reuses its native Compilation allocation between rebuilds, so a
+  // Stats object retained from the previous build aliases the build in
+  // progress. Reading its module graph while Rspack has stolen the graph
+  // artifact aborts the process instead of throwing a JavaScript error. A
+  // failed compilation can likewise have no code generation results to read.
+  if (
+    process.env.NEXT_RSPACK &&
+    (compilation.compiler.watching?.running || compilation.errors.length > 0)
+  ) {
+    return undefined
+  }
+
   const { chunkGraph, modules } = compilation
 
   return [...modules].find((module) => chunkGraph.getModuleId(module) === id)
--- a/test/development/app-dir/server-navigation-error/server-navigation-error.test.ts
+++ b/test/development/app-dir/server-navigation-error/server-navigation-error.test.ts
@@ -64,6 +64,14 @@ describe('server-navigation-error', () => {
       `)
     })
 
+    it('keeps the dev server running after mapping middleware errors', async () => {
+      const browser = await next.browser('/middleware/redirect')
+      // Reloading asks the overlay to map the error while Rspack is rebuilding.
+      await browser.refresh()
+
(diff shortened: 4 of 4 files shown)


packages/next/src/server/lib/dedupe-fetch.ts
```diff
@@ -5,6 +5,11 @@ import * as React from 'react'
 import { cloneResponse } from './clone-response'
 import { InvariantError } from '../../shared/lib/invariant-error'
 
+// Keep this lookup dynamic. Rspack validates namespace-member exports before
+// tree shaking, so a React 18 Pages Router edge bundle that only re-exports
+// another `next/cache` API would otherwise fail on this unused React 19 API.
+const ReactCache = Reflect.get(React, 'cache') as typeof React.cache
```
sokra: Solve that in a different way. Avoid making changes that affect all bundlers

sokra:
How to run the Rspack test suite?

PR #98421, opened 2026-09-09
Optimize strict module factory chunks

sokra:
### What?

Reduce Turbopack ECMAScript chunk size by emitting strict module factories in a shared strict-mode scope instead of repeating `"use strict"` in every factory.

The chunk format now chooses among the existing flat representation, a mixed representation with flat non-strict factories plus a nested strict array, and an all-strict representation that wraps the complete chunk in a strict IIFE. Each factory is generated through the existing single code path; the production minimizer removes its redundant directive when the factory is created inside a strict wrapper. Browser and Node.js emitters share the mode selection and serialization implementation, while the runtime remains compatible with the existing flat format.

### Why?

ECMAScript modules are always strict, so large chunks currently repeat the same directive across many factories. Moving strictness to the scope where those factories are created removes redundant bytes without copying arrays or changing strict/sloppy execution semantics.

For `bench/basic-app`, this reduces total raw JavaScript by 16,831 bytes (0.161%). Partitioning factories changes compression behavior, resulting in a 1,286-byte (0.051%) gzip increase across all JavaScript; format selection therefore remains based on emitted raw bytes.

### How?

- Determine each chunk item's strictness from the already-collected chunk metadata.
- Generate every factory once with its normal directive, then rely on the minimizer to remove redundant directives inside strict wrappers.
- Compare the expected minified directive bytes removed with the exact wrapper bytes added and keep small chunks in the existing flat format.
- Keep non-strict factories flat in mixed chunks and append strict factories as an array produced by a strict IIFE.
- Wrap the complete registration/export in a strict IIFE when every factory is strict, keeping those factories flat and avoiding a nested array entirely.
- Use the shorter arrow IIFE when the target supports it, with a function IIFE fallback for older targets.
- Preserve source-map sections, scope-hoisted module IDs, factory naming, and ordering within each partition.
- Add focused strict, sloppy, mixed, all-strict, old-target, and below-threshold coverage and update affected Turbopack snapshots.

### Verification

- `cargo nextest run -p turbopack-tests -E 'test(snapshot)'`
- `cargo clippy -p turbopack-ecmascript -p turbopack-browser -p turbopack-nodejs`
- `pnpm build-all`
- `pnpm --dir turbopack/crates/turbopack-ecmascript-runtime/js check:nodejs`
- `bench/basic-app` production Turbopack build, repeated to confirm deterministic size totals


turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs
```diff
@@ -0,0 +1,79 @@
...
+pub fn write_module_factories(
+    code: &mut CodeBuilder,
+    chunk_items: &[ReadRef<CodeModuleIdsAndPaths>],
+    group_strict_factories: bool,
+) -> Result<()> {
+    if group_strict_factories {
+        *code += STRICT_FACTORY_GROUP_PREFIX;
+        write_factories(code, chunk_items, ModuleFactoryMode::is_strict)?;
+        *code += STRICT_FACTORY_GROUP_SEPARATOR;
+        write_factories(code, chunk_items, |mode| !mode.is_strict())?;
```
sokra: When there are no non-strict chunk items this will emit an empty array. This could be avoided.

turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs
```diff
@@ -0,0 +1,79 @@
...
+
+/// Strict-mode directive emitted by a module factory. It is omitted from factories that a chunk
+/// creates inside its strict factory group.
+pub const STRICT_MODE_DIRECTIVE: &str = "\"use strict\";\n\n";
+/// Start of the strict array's IIFE.
+const STRICT_FACTORY_GROUP_PREFIX: &str = "\n(function(){\"use strict\";return[";
+/// Separates the strict array returned by the IIFE from the non-strict array.
+const STRICT_FACTORY_GROUP_SEPARATOR: &str = "\n]})(),[";
+/// Closes the non-strict factory array.
+const STRICT_FACTORY_GROUP_SUFFIX: &str = "\n]";
```
sokra:
```suggestion
const STRICT_FACTORY_GROUP_SEPARATOR: &str = "\n]})(),";
/// Closes the non-strict factory array.
const STRICT_FACTORY_GROUP_SUFFIX: &str = "\n";
```

Emit these items flat and not wrapped in an array

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-types.d.ts
```diff
@@ -81,11 +81,18 @@ type ModuleCache<M> = Record<ModuleId, M>
...
  */
-type CompressedModuleFactories = Array<ModuleId | Function>
+type FlatCompressedModuleFactories = Array<ModuleId | Function>
+/**
+ * Strict and non-strict factories can be supplied as separate arrays. Factories
+ * in the first array are created by a strict-mode IIFE in the emitted chunk.
+ */
+type CompressedModuleFactories =
+  | FlatCompressedModuleFactories
+  | [FlatCompressedModuleFactories, FlatCompressedModuleFactories]
```
sokra:
```suggestion
  | Array<ModuleId | Function | FlatCompressedModuleFactories>
```

turbopack/crates/turbopack-browser/src/ecmascript/content.rs
```diff
@@ -102,21 +105,13 @@ impl EcmascriptBrowserChunkContent {
...
-        });
-        for item in &chunk_items {
-            for (id, item_code, _) in &**item {
-                write!(code, "\n{}, ", StringifyJs(id))?;
-                code.push_code(item_code);
-                write!(code, ",")?;
-            }
-        }
+        let group_strict_factories =
+            should_group_strict_factories(content.strict_module_factory_count().await?);
```
sokra: group_strict_factories can be more cheaply determined from `chunk_items` which doesn't require async logic

turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs
sokra: Swap order. Put non-strict module factories first, then the strict IIFE. That makes it easier

turbopack/crates/turbopack-tests/tests/snapshot/basic-tree-shake/export-named/output/1jsg_tests_snapshot_basic-tree-shake_export-named_input_0xj7cshvyczs_._.js
sokra, line 1: When there are *only* strict module factories, the chunk format can become even more efficient. You can remove the IIFE to top level:


```
(function(){"use strict";(globalThis["TURBOPACK"] || (globalThis["TURBOPACK"] = [])).push(["output/1jsg_tests_snapshot_basic-tree-shake_export-named_input_0xj7cshvyczs_._.js",...
...

File: turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs [hunk 5]
```diff
@@ -0,0 +1,74 @@
...
+use crate::{
+    chunk::{CodeModuleIdsAndPaths, ModuleFactoryMode},
+    utils::StringifyJs,
+};
+
+/// Strict-mode directive emitted by a module factory. It is omitted from factories that a chunk
+/// creates inside its strict factory group.
+pub const STRICT_MODE_DIRECTIVE: &str = "\"use strict\";\n\n";
+/// Start of the strict array's IIFE.
+const STRICT_FACTORY_GROUP_PREFIX: &str = "\n(function(){\"use strict\";return[";
```
sokra: Instead of `function()` you can also use an arrow function which is a bit shorter

turbopack/crates/turbopack-ecmascript/src/chunk/item.rs
```diff
@@ -124,7 +127,7 @@ impl EcmascriptChunkItemContent {
 }
 
 impl EcmascriptChunkItemContent {
-    async fn module_factory(&self) -> Result<ResolvedVc<PersistedCode>> {
+    async fn module_factory(&self, omit_use_strict: bool) -> Result<ResolvedVc<PersistedCode>> {
```
sokra: Remove the `omit_use_strict` argument and always generate `"use strict"` when needed.
When wrapped in a IIFE with `"use strict"` the minimizer will remove the duplicate `"use strict"`. This avoids calling `module_factory` twice with different arguments.

turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs
```diff
@@ -0,0 +1,126 @@
...
+        StrictFactoryMode::AllStrict => {
+            all_strict_prefix(supports_arrow_functions).len() + ALL_STRICT_SUFFIX.len()
+        }
+        StrictFactoryMode::Flat => unreachable!(),
+    };
+    if strict * STRICT_MODE_DIRECTIVE.len() > wrapper_len {
+        mode
+    } else {
+        StrictFactoryMode::Flat
+    }
```
sokra: Derive the number of strict modules for it to be worth and use a constant `const STRICT_THRESHOLD: usize = X`

Validate the value against the equation in a unit test.

You might use two different constant if `supports_arrow_functions` makes a difference. If not, remove the argument.

```
    if strict > STRICT_THRESHOLD {
        mode
    } else {
        StrictFactoryMode::Flat
    }
```

turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs
```diff
@@ -0,0 +1,158 @@
...
+const ALL_STRICT_FUNCTION_PREFIX: &str = "(function(){\"use strict\";";
+const ALL_STRICT_SUFFIX: &str = "})()";
+
+/// Number of strict factories a chunk needs before hoisting the directive is worth its wrapper.
+/// The minimizer removes the factory directives inside a strict wrapper, so grouping pays off once
+/// the removed directives outweigh the wrapper that replaces them. Both thresholds hold for the
+/// arrow and the function wrapper alike; `strict_thresholds_match_wrapper_sizes` keeps them honest
+/// if a wrapper string changes.
+const MIXED_STRICT_THRESHOLD: usize = 2;
+const ALL_STRICT_THRESHOLD: usize = 1;
```
sokra: It's a bit consuming as all string is not worth it for `1` (`ALL_STRICT_THRESHOLD`) chunk item. It's using `>` so that's ok, but to make it less consuming change it to:


```suggestion
const MIXED_STRICT_THRESHOLD: usize = 3;
const ALL_STRICT_THRESHOLD: usize = 2;
```

And `if strict > threshold` -> `if strict >= threshold`

turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs
```diff
@@ -0,0 +1,158 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Length of the `"use strict";\n\n` directive that
+    /// `EcmascriptChunkItemContent::module_factory` writes into every strict factory, and that
+    /// the minimizer removes again inside a strict wrapper. Kept here because only these
+    /// assertions need it.
+    const STRICT_MODE_DIRECTIVE_LEN: usize = "\"use strict\";\n\n".len();
```
sokra:
```suggestion
    /// Length of the `"use strict";` directive that
    /// `EcmascriptChunkItemContent::module_factory` writes into every strict factory, and that
    /// the minimizer removes again inside a strict wrapper. Kept here because only these
    /// assertions need it.
    const STRICT_MODE_DIRECTIVE_LEN: usize = "\"use strict\";".len();
```

new lines are not kept during minimification

turbopack/crates/turbopack-ecmascript/src/chunk/factory_group.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    /// `threshold` must be the largest strict-factory count that is *not* worth grouping: at the
+    /// threshold the directives it would remove do not outweigh `wrapper_len`, one above it they
+    /// do.
+    fn assert_threshold(threshold: usize, wrapper_len: usize) {
+        assert!(threshold * STRICT_MODE_DIRECTIVE_LEN <= wrapper_len);
+        assert!((threshold + 1) * STRICT_MODE_DIRECTIVE_LEN > wrapper_len);
+    }
```
sokra: Pass the wrapper as string. Remove all whitespace from the wrapper before measuring

turbopack/crates/turbopack-ecmascript-runtime/js/src/shared/runtime/runtime-utils.ts
```diff
@@ -578,29 +578,27 @@ function getChunkPath(chunkData: ChunkData): ChunkPath {
...
 function installCompressedModuleFactories(
   chunkModules: CompressedModuleFactories,
   offset: number,
   moduleFactories: ModuleFactories,
   newModuleId?: (id: ModuleId) => void
 ) {
+  const strictFactories = chunkModules[chunkModules.length - 1]
+  const length = Array.isArray(strictFactories)
+    ? chunkModules.length - 1
+    : chunkModules.length
```
sokra: This code can be simplified and shortened when the strictFactories always appear as first element.

You also need to update `write_module_factories` to always emit it at the start

sokra:
Run a few more measurements for other apps in the bench folder, e.g. heavy npm deps

turbopack/crates/turbopack-tests/tests/snapshot/basic-tree-shake/dynamic-import/output/1jsg_tests_snapshot_basic-tree-shake_dynamic-import_input_lib_1b34ju5a6z6v3.js
```diff
@@ -1,4 +1,4 @@
-(globalThis["TURBOPACK"] || (globalThis["TURBOPACK"] = [])).push(["output/1jsg_tests_snapshot_basic-tree-shake_dynamic-import_input_lib_1b34ju5a6z6v3.js",
+(()=>{"use strict";(globalThis["TURBOPACK"] || (globalThis["TURBOPACK"] = [])).push(["output/1jsg_tests_snapshot_basic-tree-shake_dynamic-import_input_lib_1b34ju5a6z6v3.js",
 "[project]/turbopack/crates/turbopack-tests/tests/snapshot/basic-tree-shake/dynamic-import/input/lib.js [test] (ecmascript)", ((__turbopack_context__) => {
 "use strict";
```
mischnic, line 3: Why is there still a `use strict` per module here then?

turbopack/crates/turbopack-ecmascript/src/chunk/code_module_ids_and_paths.rs
```diff
@@ -13,8 +13,37 @@ use crate::chunk::{
...
+        .content_with_async_module_info(async_module_info, false)
+        .await?
+        .await?
+        .options
+        .strict;
+    let code = item.chunk_item.code(async_module_info);
+    Ok((code.await?, ModuleFactoryMode(strict)))
+}
+
+pub type CodeModuleIdAndPath = (ModuleId, ReadRef<Code>, RcStr, ModuleFactoryMode);
```
mischnic: Can we finally make this a struct

turbopack/crates/turbopack-ecmascript/src/chunk/code_module_ids_and_paths.rs
```diff
@@ -13,8 +13,37 @@ use crate::chunk::{
...
+        .chunk_item
+        .into_trait_ref()
+        .await?
+        .content_with_async_module_info(async_module_info, false)
+        .await?
+        .await?
+        .options
+        .strict;
+    let code = item.chunk_item.code(async_module_info);
+    Ok((code.await?, ModuleFactoryMode(strict)))
```
mischnic: How slow is this? Should we just make `.code()` return `strict` as well instead?

mischnic (approved):

PR #98422, opened 2026-09-09
Typecheck the Turbopack ECMAScript runtime in CI

sokra:
### What?

Integrate the Turbopack ECMAScript runtime package checks into the repository TypeScript task graph and resolve the latent type errors those checks expose.

### Why?

The runtime package already had checks for its environment-specific TypeScript projects, but CI never executed them. This allowed its public runtime context types to diverge from their implementations, including the `EsmExport` binding representation.

### How?

The package now participates in the existing Turbo `typescript` task used by CI. Its package-local Turbo configuration builds workspace dependencies first, ensuring the React Refresh type output is available on clean checkouts without adding build dependencies to every repository typecheck.

The runtime declarations now reflect the actual flat ESM binding format and HMR behavior. Browser development references and return types also mirror the runtime files and concatenation order used by the Rust runtime builder. Turbopack snapshots that embed these runtime sources and source maps are regenerated to match.

### Verification

- `cargo nextest run -p turbopack-tests -E 'test(snapshot)'` (125 passed)
- `pnpm --dir turbopack/crates/turbopack-ecmascript-runtime/js check`
- `pnpm lint-typescript --filter=@vercel/turbopack-ecmascript-runtime --log-order=stream --force` after removing `packages/react-refresh-utils/dist`
- Focused Prettier and ESLint checks
- `git diff --check`


mischnic:
the cargo unit tests are failing

mischnic (approved):

PR #98425, opened 2026-09-09
Use pnpm for isolated test installs

sokra:
### What?

Migrate the five local isolated test installs that explicitly used npm to pnpm. The Nx fixture now has pnpm workspace metadata, while filesystem-layout-sensitive fixtures use pnpm's hoisted linker with copied package files.

### Why?

The npm-based Nx install bypassed the repository's centralized supply-chain protections and could select a package immediately after publication, including temporarily incomplete multi-package releases. Using pnpm makes isolated installs inherit the repository's `minimumReleaseAge`, exclusions, and exotic-subdependency policy.

The other npm installs depended on npm-style real package directories. On Node versions affected by nodejs/node#65113, hoisted/copy mode preserves that layout without leaving these fixtures outside the shared pnpm security configuration; fixed Node releases use normal pnpm linking.

### How?

- Use normal pnpm workspace resolution for the Nx fixture.
- Use `node-linker=hoisted` and `package-import-method=copy` for filesystem tests only on affected Node releases; Node 24.21+ and 26.8+ use normal linking. Node 20 CI keeps the workaround because no fixed Node 20 release exists.
- Validate local `@next/env` tarballs through the lockfile when hoisted installs do not expose pnpm's virtual-store path marker.
- Keep the deployment-environment npm install unchanged.

### Verification

- `pnpm build-all`
- `pnpm types`
- A 9-version throwaway assertion verified the affected/fixed Node release matrix
- `pnpm test-dev-turbo test/e2e/app-dir/nx-handling/nx-handling.test.ts test/e2e/handle-non-hoisted-swc-helpers/index.test.ts test/e2e/filesystem-cache/filesystem-cache.test.ts test/e2e/filesystem-cache/warm-restart-task-stats.test.ts test/e2e/filesystem-cache/evict-after-snapshot.test.ts` — all 25 tests passed after installing the sandbox's missing Playwright browser
- Production Turbopack: Nx, non-hoisted SWC helper, build-cache-default, and warm restart passed (9/9)
- `filesystem-cache.test.ts` production baseline: 15/17 passed; the same two cache-growth bounds fail under both the unchanged npm fixture and the pnpm fixture at nearly identical percentages, so they are pre-existing sandbox-specific failures
- Generated-layout inspection: no package symlinks outside expected `.bin` command shims; package files are copied; `node_modules/.pnpm` is metadata-only


test/e2e/filesystem-cache/filesystem-cache.test.ts
```diff
@@ -52,19 +52,24 @@ for (const cacheEnabled of [false, true]) {
...
         },
       },
-      // We need to use npm here as pnpms symlinks trigger a weird bug (kernel bug?)
-      installCommand: 'npm i',
+      installCommand: 'pnpm install',
       // Next is always started with caching, but this can disable it for the followup restarts
-      buildCommand: `npm run build`,
-      startCommand: isNextDev ? 'npm run dev' : 'npm run start',
+      buildCommand: `pnpm run build`,
+      startCommand: isNextDev ? 'pnpm run dev' : 'pnpm run start',
```
sokra: Avoid passing these commands when they are equal to the default value

test/e2e/app-dir/nx-handling/package.json
```diff
@@ -28,5 +28,5 @@
   "workspaces": [
     "apps/*"
   ],
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  "packageManager": "pnpm@10.33.0"
```
sokra: Should this be specified at all? What happens when we omit it?

test/e2e/filesystem-cache/warm-restart-task-stats.test.ts
```diff
@@ -30,35 +30,32 @@ const STATS_RELATIVE_PATH = '.next/warm-restart-task-stats.json'
...
     files: __dirname,
-    packageJson: {
-      packageManager: 'npm@10.9.2',
-      scripts: {
-        build: `${env} next build`,
-        dev: `${env} next dev`,
-        start: 'next start',
-      },
+    overrideFiles: {
+      '.npmrc': `# The default pnpm symlinks trigger a kernel bug in this test. Use an
```
sokra: We assumed it's a "kernel bug", but in fact it was a node.js bug: https://github.com/nodejs/node/pull/65113

It is fixed in https://github.com/nodejs/node/pull/65873 resp. https://github.com/nodejs/node/pull/65113

Check if our CI uses at least this version (upgrade minor version if neccessary). Only apply the workaround on older node.js major versions.

This comment applies to multiple changes in this PR that have the same code comment

test/e2e/app-dir/nx-handling/nx-handling.test.ts
```diff
@@ -4,14 +4,12 @@ describe('nx-handling', () => {
   const { next } = nextTestSetup({
     skipDeployment: true,
     files: __dirname,
-    installCommand: 'npm i',
-    buildCommand: 'npm run build',
-    startCommand: isNextDev ? 'npm run dev' : 'npm run start',
+    buildCommand: 'pnpm run build',
+    startCommand: isNextDev ? 'pnpm run dev' : 'pnpm run start',
```
sokra, line 8: This looks like it's the default and not needed?

test/unit/pnpm-symlink-workaround.test.ts
eps1lon, line 1: can be deleted.

test/lib/pnpm-symlink-workaround.ts
```diff
@@ -0,0 +1,21 @@
+const PNPM_HOISTED_CONFIG = `# Work around the Node.js realpath bug fixed in 24.21.0 and 26.8.0.
+# https://github.com/nodejs/node/pull/65113
+node-linker=hoisted
+package-import-method=copy
+`
+
+export function hasNodeRealpathFix(nodeVersion: string): boolean {
```
eps1lon: no need to export. only used internally.

test/lib/pnpm-symlink-workaround.ts
```diff
@@ -0,0 +1,21 @@
...
+  const [major, minor] = nodeVersion.split('.').map(Number)
+  return (
+    (major === 24 && minor >= 21) || major > 26 || (major === 26 && minor >= 8)
+  )
+}
+
+export function getPnpmSymlinkWorkaround(
```
eps1lon: Name should be descriptive. We already reference "realpath fix" which seems like a good candidate.

test/lib/create-next-install.js
```diff
@@ -377,7 +420,18 @@ async function createNextInstall({
...
+            !envDir.includes('@next+env@file') &&
+            !(
+              envTarballPath &&
+              (await lockfileResolvesLocalTarball(
+                installDir,
+                '@next/env',
+                envTarballPath
+              ))
+            )
+          ) {
```
eps1lon, line 435: Does this run on every test?

eps1lon (dismissed):

wbinnssmith (approved):

test/lib/pnpm-realpath-workaround.ts
```diff
@@ -0,0 +1,21 @@
+const PNPM_HOISTED_CONFIG = `# Work around the Node.js realpath bug fixed in 24.21.0 and 26.8.0.
+# https://github.com/nodejs/node/pull/65113
+node-linker=hoisted
+package-import-method=copy
```
bgw, line 4: `package-import-method=copy` shouldn't be needed to work around `realpath` bugs, because the alternatives it would use are a reflink or a hardlink, neither of which impact `realpath`.

PR #98426, opened 2026-09-09
Fix startTime and shutdown delivery of forwarded Turbopack trace spans

mischnic:
before:
Bildschirmfoto 2026-09-09 um 14 33 59

after:
Bildschirmfoto 2026-09-09 um 16 32 57

## Summary

Two fixes for Turbopack compilation events forwarded to the JS trace (`.next/trace`). The span-forwarding work stacks on top of this PR.

- **`Span.manualTraceChild` serialized the event recording time as `startTime`** instead of the supplied epoch start, so forwarded spans rendered shifted forward by their own duration and overflowed their parent span. `startTime` now reflects the supplied start.
- **Compilation events were lost at build shutdown**: `CompilationEventQueue` delivered events via an untracked `tokio::spawn`, and the build teardown aborted the events subscription right after `project.shutdown()` resolved, so shutdown-time events (e.g. `turbopack-persistence` on cold builds, where it is the only persistence event since builds use `ReadWriteOnShutdown`) were discarded whenever delivery lost the race. The queue now tracks in-flight deliveries and flushes/closes subscriber channels at the end of `stop_and_wait`, and the build waits for the subscription to drain before aborting it.

## Verification

- New `trace.test.ts` unit test: a `manualTraceChild` span started in the past keeps its own start time (fails before the fix, passes after).
- New `message_queue::test_flush_and_close` unit test: events are delivered before the channel closes; post-close subscriptions replay history and close.
- `cargo check` / `cargo test -p turbo-tasks --lib message_queue`, root `pnpm typescript`.
- The stacked PR on top verifies end-to-end that `turbopack-persistence` is now emitted on every cold build.


turbopack/crates/turbo-tasks/src/message_queue.rs
```diff
@@ -59,58 +89,86 @@ impl CompilationEventQueue {
...
+        self.closed.store(true, Ordering::Release);
+        loop {
+            // Create the listener before checking the counter so no completion is missed.
+            let listener = self.deliveries_idle.listen();
+            if self.pending_deliveries.load(Ordering::Acquire) == 0 {
+                break;
+            }
+            listener.await;
+        }
+        // Dropping the senders closes the subscriber channels once they are drained.
```
lukesandberg, line 147: should we drop the channels first? instead of last? that way each channel closes as soon as it is drained

turbopack/crates/turbo-tasks/src/message_queue.rs
```diff
@@ -59,58 +89,86 @@ impl CompilationEventQueue {
...
+                        all_channel.remove(*ix);
+                    }
                 }
             }
 
-            for ix in removal_indices.iter().rev() {
-                all_channel.remove(*ix);
-            }
+            pending_deliveries.fetch_sub(1, Ordering::AcqRel);
+            deliveries_idle.notify(usize::MAX);
```
lukesandberg, line 127: notifying after every sub is odd, shouldn't it just be on the subs that drop us to zero?

turbopack/crates/turbo-tasks/src/message_queue.rs
lukesandberg: should we be using a tokio condition variable and mutext instead of an atomic and an event?

lukesandberg (approved):

PR #98447, opened 2026-09-09
Add in warning for workers around the fsevents bug

jimmyhmiller:
We removed this because we fixed the original crashes. This crash does not happen out of the box. But it isn't hard to trigger with plugins. And we'd rather not have people experience this without being explicit.

### DIFF
--- a/packages/next/src/server/config.ts
+++ b/packages/next/src/server/config.ts
@@ -1711,7 +1711,11 @@ function assignDefaultsAndValidate(
           Log.warn(
             `\`experimental.turbopackPluginRuntimeStrategy = ` +
               `'forceWorkerThreads'\` is enabled, bypassing protection ` +
-              `against a known potential crash in Node.js ${affectedNodeRange}.`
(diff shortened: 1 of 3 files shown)


packages/next/src/server/config.ts
```diff
@@ -1690,6 +1691,45 @@ function assignDefaultsAndValidate(
...
+              `against a known potential crash in Node.js ${affectedNodeRange}.`
+          )
+        } else {
+          Log.warn(
+            `\`experimental.turbopackPluginRuntimeStrategy = ` +
+              `'workerThreads'\` is disabled on Node.js ${nodeVersion}.\n` +
+              `A Node.js worker-thread teardown bug can abort the process ` +
+              `when a native addon, such as fsevents, has a live Node-API ` +
+              `threadsafe function as a worker exits.\n` +
+              `See https://github.com/nodejs/node/issues/65100.\n` +
```
lukesandberg, line 1727: nit: reference the bug in the other warning as well?

bgw (approved):
Looks familiar...

PR #98469, opened 2026-09-10
Turbopack: Add symlinks and additional roots to NFT metadata

bgw:
Implements the following extension:

```typescript
export interface NftFileList {
  // File paths relative to the directory containing the `.nft.json` file, or
  // the current root (if inside of `NftAdditionalRoot`).
  //
  // When using webpack, these paths may exist outside of the tracing root. The
  // [`@vercel/next` package ignores these paths][vc-next].
  //
  // When using Turbopack, these paths are all guaranteed to exist within the
  // `turbopack.root` specified or inferred from the `next.config.js` file.
  //
  // [vc-next]: https://github.com/vercel/vercel/blob/%40vercel/next%404.20.5/packages/next/src/server-build.ts#L1022-L1026
  files: string[]
  // Turbopack extension: A parallel array to `files` (same indices and length)
  // with file content hashes. For symlinks, the hash of the path of the target
  // is stored.
  fileHashes?: string[]
  // Turbopack extension: Explicit symlink mapping information.
  //
  // If included, it's safe to assume that if a file is not in `symlinks` that
  // it is not a symlink. If this is an empty array, there are no symlinks.
  //
  // If omitted, the processor of the nft file must call `read_link` on every
  // file to determine if it is a symlink and determine the target path.
  //
  // This field is always included if `NftJson` includes `additionalRoots`, and
  // it is always included on `NftAdditionalRoot`.
  symlinks?: NftSymlink[]
}

export interface NftJson extends NftFileList {
  version: 1
  // Turbopack extension: A hash of the entrypoint that refers to these traced
  // files. This hash only depends on the content of the entrypoint file, and
  // not all of its traced dependencies.
  entryHash?: string
  // Turbopack extension: Paths stored with different base paths, typically
  // outside of the tracing root.
  additionalRoots?: NftAdditionalRoot[]
}

// Turbopack extension: A collection of paths stored with a different base path.
export interface NftAdditionalRoot extends NftFileList {
  // Stable unique identifier provided in the `next.config.js`. This can be used
  // to generate the output path where these files are copied to (e.g.
  // `.additionalRoots/$[name}`).
  //
  // This is guaranteed to use a character set that is valid on most
  // filesystems, and the identifiers are guaranteed to not have overlaps on
  // case-insensitive filesystems.
  name: string
  // A source path on the build machine that the paths in `files` are relative
  // to. The final build output directory should not depend on this path.
  absolutePath: string
  // Always specified on NftAdditionalRoot.
  symlinks: NftSymlink[]
}

// Turbopack extension: Information on a symlink, including which additional
// root it maps to. Symlinks that do not cross root boundaries (the common case)
// omit the index into `additionalRoots`.
//
// It is often complicated to transform raw symlink targets to root-relative
// paths, and including this information here ensures that the NFT reader gets
// the same result that Turbopack's tracing system expects.
//
// Because the link target type is unspecified, on Windows the reader needs to
// call `stat` to determine if a link target is a directory or file.
export type NftSymlink =
  | [
      // Index of `files` that refers to a symlink.
      number,
      // The target path of the link. In `NftJson`, this path is relative to the
      // directory containing the `.nft.json` file. In `NftAdditionalRoot`, this
      // is relative to the current root.
      string,
    ]
  | [
      // Index of `files` that refers to a symlink.
      number,
      // The target path of the link relative to the specified root.
      string,
      // An index into `additionalRoots`, -1 if the target path is relative to
      // the `.nft.json` file's directory,
      //
      // If the symlink target is relative to the same root as the symlink
      // itself (the "current root"), this field is omitted.
      number,
    ]
```

Nothing currently populates `additionalRoots`. https://github.com/vercel/next.js/pull/98003 will do it.


packages/next/src/build/utils.ts
```diff
@@ -1273,47 +1274,59 @@ export async function copyTracedFiles(
...
+              )
             } catch (err: any) {
-              // Windows doesn't support creating symlinks without elevated privileges, unless
-              // "Developer Mode" is turned on. If we failed to create a symlink due to EPERM, try
-              // creating a junction point instead.
-              //
-              // Ideally we'd just preserve the input file type (junction point or symlink), but
-              // there's no API in node.js to differentiate between a junction point and a symlink,
-              // so we just try making a symlink first. Symlinks are preferred because they support
-              // relative paths and non-directory (file) targets.
```
bgw, line 1309: Don't delete this comment

packages/next/src/build/utils.ts
```diff
@@ -1323,6 +1336,31 @@ export async function copyTracedFiles(
...
+                  try {
+                    await fs.symlink(symlink, fileOutputPath, 'junction')
+                  } catch (junctionErr: any) {
+                    if (junctionErr.code !== 'EEXIST') {
+                      throw junctionErr
+                    }
+                  }
+                } else if (err.code !== 'EEXIST') {
+                  throw err
+                }
```
bgw, line 1359: This whole block of code is duplicated from the branch above. Dedupe it with a helper function.

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -721,30 +746,56 @@ pub(crate) fn format_absolute_fs_path(path: &Path, name: &str, root_path: &Path)
...
+    ///
+    /// `map` provides other configured filesystems used to resolve symlink targets that leave
+    /// this filesystem's root.
+    pub fn new_with_file_system_map(
+        name: RcStr,
+        root: Vc<RcStr>,
+        denied_paths: Vec<RcStr>,
+        watcher_config: DiskWatcherConfig,
+        map: OperationVc<DiskFileSystemMap>,
     ) -> Vc<Self> {
```
bgw, line 799: Merge these constructors, only expose `new_with_options` that takes a `map` argument.

crates/next-api/src/nft_json_builder.rs
```diff
@@ -0,0 +1,248 @@
...
+    symlinks: Vec<NftSymlink>,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "camelCase")]
+struct NftAdditionalRoot {
+    #[serde(flatten)]
+    file_list: NftFileList,
+    name: RcStr,
+    absolute_path: RcStr,
```
bgw, line 78: Make this path relative to the `*.nft.json` file.

crates/next-api/src/nft_json_builder.rs
```diff
@@ -0,0 +1,248 @@
...
+            Self::AdditionalRoot { root_index, path } => (Some(*root_index), path),
+        }
+    }
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct AssetReference {
+    location: AssetLocation,
+    hash: RcStr,
+    target: Option<AssetLocation>,
```
lukesandberg, line 30:
```suggestion
    // Present for symlinks
    target: Option<AssetLocation>,
```

lukesandberg: symlink_target might be a better name

turbopack/crates/turbo-tasks-fs/src/content.rs
```diff
@@ -276,15 +279,13 @@ impl LinkContent {
...
-    Clone, Debug, Hash, PartialEq, Eq, TraceRawVcs, NonLocalValue, DeterministicHash, Encode, Decode,
-)]
+#[derive(Clone, Debug, Hash, PartialEq, Eq, TraceRawVcs, NonLocalValue, Encode, Decode)]
 pub enum WriteLinkTarget {
-    /// Normalized and relative to the *filesystem root*.
-    Absolute(RcStr),
+    /// Normalized and relative to `root`.
+    Absolute {
+        root: ResolvedVc<DiskFileSystem>,
+        path: RcStr,
```
bgw, line 287: If it's not important that this is a `DiskFileSystem` specifically, make this a `FileSystemPath`.

crates/next-api/src/nft_json_builder.rs
```diff
@@ -0,0 +1,248 @@
...
+}
+
+struct NftSymlink {
+    file_index: usize,
+    target: RcStr,
+    root: Option<isize>,
+}
+
+impl Serialize for NftSymlink {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
```
bgw, line 50:
```suggestion
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
```

crates/next-api/src/nft_json_builder.rs
```diff
@@ -0,0 +1,248 @@
...
+impl NftJsonBuilder {
+    pub async fn new(project: ResolvedVc<Project>, nft_path: &FileSystemPath) -> Result<Self> {
+        let mut root_configs = FxHashMap::default();
+
+        // Files not listed under `additionalRoots` have paths relative to the nft.json file, which
+        // lives in the output filesystem. We need to remap that back to the project filesystem. We
+        // can assume that the project directory is a parent of the output directory.
+        let project_root = project.project_fs().root().owned().await?;
+        let output_base = nft_path.parent();
+        let project_base = project_root.join(&output_base.path)?;
```
lukesandberg, line 110: project_output_base?

bgw, line 110: Use turbo-unix-path directly so that we don't end up enforcing that the output directory lives under the project directory.

crates/next-api/src/nft_json.rs
```diff
@@ -171,15 +142,18 @@ impl Asset for NftJsonAsset {
...
                 )
                 .map(async |referenced| {
-                    let (referenced_chunk_path, hash) = match referenced {
-                        AssetOrModule::Asset(v) => (
-                            Either::Left(v.path().await?),
-                            Either::Left(
-                                v.content()
+                    let (referenced_chunk_path, hash, content) = match referenced {
+                        AssetOrModule::Asset(v) => {
+                            let content = v.content();
```
bgw, line 147: resolve this vc earlier

packages/next/src/build/nft.ts
```diff
@@ -0,0 +1,229 @@
...
+      `path ${JSON.stringify(relativePath)} escapes additional root ${root.name}`
+    )
+  }
+  return {
+    source,
+    destination: path.join('nextAdditionalRoots', root.name, relativePath),
+  }
+}
+
+export function mapNftFileEntries(
```
lukesandberg, line 166: clarify in docs that the symlinks data in the return is only accurate if they are in the actual nft file

packages/next/src/build/nft.ts
```diff
@@ -0,0 +1,229 @@
...
+       * the symlink itself.
+       */
+      number,
+    ]
+
+export interface MappedNftFileEntry {
+  source: string
+  destination: string
+  hash?: string
+  symlinkTarget?: string
```
bgw, line 115: Add a comment that there are two different reasons this could be undefined: Because we know it's not a symlink (if we have the symlink info) or because we don't know if it's a symlink or not.

lukesandberg, line 115: comment

packages/next/src/build/nft.ts
```diff
@@ -0,0 +1,229 @@
...
+
+function mapBasePath(
+  traceFileDirectory: string,
+  baseRoot: string,
+  relativePath: string
+): { source: string; destination: string } {
+  const source = path.resolve(traceFileDirectory, relativePath)
+  const destination = relativePathIfInside(baseRoot, source)
+  if (destination === undefined) {
+    invalid(`path ${JSON.stringify(relativePath)} escapes the base root`)
```
lukesandberg, line 145: is this back compat

packages/next/src/build/nft.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
```
bgw, line 146: Probably better to warn and just out output the file than to throw in the base path case. Fine to throw in the additional root case.

packages/next/src/build/nft.ts
```diff
@@ -0,0 +1,229 @@
...
+          : mapAdditionalRootPath(roots[currentRootIndex], file)
+
+      let symlinkTarget: string | undefined
+      if (nextSymlink?.[0] === fileIndex) {
+        const [, target, rootIndex] = nextSymlink
+        nextSymlink = symlinks[++symlinkCursor]
+        const targetRootIndex = rootIndex ?? currentRootIndex
+        symlinkTarget =
+          targetRootIndex === -1
+            ? mapBasePath(traceFileDirectory, baseRoot, target).destination
```
lukesandberg, line 199: it would be fine to reject invalid paths from this callsite even if it isn't from the other callsite

packages/next/src/build/adapter/build-complete.ts
```diff
@@ -2738,26 +2739,15 @@ async function loadNFT(
...
-  const traceFileDir = path.dirname(traceFilePath)
-  for (let i = 0; i < files.length; i++) {
-    const relativeFile = files[i]
-    const contentHash = fileHashes?.[i]
-    const tracedFilePath = path.join(traceFileDir, relativeFile)
-    const fileOutputPath = path.relative(repoRoot, tracedFilePath)
-    assets[fileOutputPath] = tracedFilePath
-    if (contentHash) {
-      assetsHashes[fileOutputPath] = contentHash
+  for (const entry of mapNftFileEntries(nft, traceFilePath, repoRoot)) {
```
lukesandberg, line 2744: maybe a comment that this callsite doesn't care about symlink data?

packages/next/src/build/utils.ts
```diff
@@ -1273,47 +1274,59 @@ export async function copyTracedFiles(
...
+        const fileOutputPath = resolveNftOutputPath(
           outputPath,
-          path.relative(tracingRoot, tracedFilePath)
+          entry.destination
         )
 
         if (!copiedFiles.has(fileOutputPath)) {
           copiedFiles.add(fileOutputPath)
 
           await fs.mkdir(path.dirname(fileOutputPath), { recursive: true })
```
lukesandberg, line 1294: TODO: adopt synchronous file io

implement in rust????

packages/next/src/build/utils.ts
```diff
@@ -1273,47 +1274,59 @@ export async function copyTracedFiles(
...
+              }
+            }
 
-          if (symlink) {
             try {
-              await fs.symlink(symlink, fileOutputPath)
+              await fs.symlink(
+                symlink,
+                fileOutputPath,
+                isDirectory ? 'dir' : 'file'
```
lukesandberg, line 1318: please write an essay on how windows is dumb and we happen to rely on writing symlinks to files/dirs that don't quite exist yet maybe

packages/next/src/build/utils.ts
```diff
@@ -1323,6 +1336,31 @@ export async function copyTracedFiles(
                 throw err
               }
             }
+          } else if (traceData.symlinks === undefined) {
+            const symlink = await fs.readlink(tracedFilePath).catch(() => null)
+            if (symlink) {
+              try {
+                await fs.symlink(symlink, fileOutputPath)
```
bgw, line 1343: This should also compute file vs directory?

packages/next/src/build/utils.ts
lukesandberg: helper function to get symlinks data makes sense

packages/next/src/cli/internal/static-routes-info.ts
```diff
@@ -347,32 +348,27 @@ function collectServerEntryFiles(
...
 
-  const nft = readJsonFile<{ files: string[] }>(
-    path.join(distDir, entryRel + '.nft.json')
-  )
-  if (!nft?.files) return
-
-  for (const relPath of nft.files) {
-    // Resolve relative to the entry's dir. If the normalized result stays
-    // inside distDir it's a server chunk; if it leaves distDir it's an
-    // unbundled trace dep (e.g. ../../../node_modules/...).
```
bgw, line 364: Don't delete thes comments

turbopack/crates/turbo-tasks-fs/src/content.rs
```diff
@@ -259,9 +264,7 @@ impl LinkContent {
         }
         let simplified = match self {
             LinkContent::Link { target } => match target {
-                LinkTarget::Absolute { resolved } => {
-                    SimplifiedLinkContent::Absolute(&resolved.path)
-                }
+                LinkTarget::Absolute { raw, .. } => SimplifiedLinkContent::Absolute(raw),
```
bgw, line 267: This should use the `DiskFileSystem`'s name + `resolved.path`, not the raw.

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -974,6 +1028,9 @@ impl FileSystem for DiskFileSystem {
                     .resolve_link_target_ancestry_slow_path(self, &target_sys_path)
                     .await?;
             }
+            if target_fs_path.is_none() {
+                target_fs_path = this.lookup_in_file_system_map(&target_sys_path).await?;
```
lukesandberg, line 1032: fast path for an empty?

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -650,51 +686,40 @@ impl DiskFileSystem {
...
-        // The whole path was consumed without reaching the filesystem root.
-        Ok(None)
+        visit_canonicalized_ancestry(target_sys_path, |canonical, rest| {
+            if let Some(path) = map.lookup(&canonical.join(rest)) {
+                ControlFlow::Break(Some(path))
+            } else {
+                ControlFlow::Continue(())
+            }
+        })
+        .await
```
bgw, line 722: Add a special-case for if the filesystem map is empty

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -1314,13 +1385,46 @@ impl FileSystem for DiskFileSystem {
...
+                }
+            }
+        };
+
+        #[derive(DeterministicHash)]
+        struct ResolvedWriteLinkContent<'a> {
+            target: &'a [u8],
+            target_type: &'a WriteLinkTargetType,
+        }
+        let content_hash = hash_xxh3_hash128(&ResolvedWriteLinkContent {
```
lukesandberg, line 1417: can't we just use WriteLinkTarget::hash?

turbopack/crates/turbo-tasks-fs/src/disk.rs
```diff
@@ -1314,13 +1385,46 @@ impl FileSystem for DiskFileSystem {
...
+            target: target_path.as_os_str().as_encoded_bytes(),
+            target_type: &target.target_type,
+        });
 
         #[turbo_tasks::value(eq = "manual", cell = "new")]
         struct WriteLinkEffect {
             full_path: Arc<PathBuf>,
             fs: ResolvedVc<DiskFileSystem>,
-            target: ResolvedVc<WriteLinkContent>,
```
bgw, line 1323: WriteLinkEffect could potentially still take a WriteLinkContent here, the only thing that cannot take this is the captured effect.

turbopack/crates/turbo-tasks-fs/src/fs_map.rs
```diff
@@ -0,0 +1,96 @@
...
+use turbo_unix_path::sys_to_unix;
+
+use crate::{DiskFileSystem, FileSystemPath};
+
+/// An ordered set of canonical system roots and their owning filesystems.
+///
+/// The roots must not overlap: no root may be an ancestor of another root. [`Self::lookup`]
+/// relies on this invariant when selecting the nearest preceding root in path order.
+#[turbo_tasks::value(shared)]
+pub struct DiskFileSystemMap(BTreeMap<PathBuf, ResolvedVc<DiskFileSystem>>);
```
lukesandberg, line 18: frozenmap?

bgw, line 18: Potential follow-up: Explore if I can use a frozen map for this. Instead of the cursor API, I could operate on indicies.

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -960,170 +931,4 @@ mod tests {
         .await
         .unwrap()
     }
-
-    mod hash_file {
```
bgw, line 964: Is it worth preserving these tests on the other hash implementation?

turbopack/crates/turbopack-core/src/file_source.rs
```diff
@@ -70,13 +70,21 @@ impl Asset for FileSource {
         match file_type {
             FileSystemEntryType::Symlink => match &*self.path.read_link().await? {
                 LinkContent::Link { target } => {
+                    let target_fs_path = target.file_system_path();
                     let write_target = match target {
-                        LinkTarget::Absolute { resolved } => {
-                            WriteLinkTarget::Absolute(resolved.path.clone())
+                        LinkTarget::Relative { raw, resolved } if resolved.fs == self.path.fs => {
+                            WriteLinkTarget::Relative(raw.clone())
```
lukesandberg, line 76: is it important for WriteLinkTarget to draw this particular distinction (writing a symlink to the current filesystem or a different file system?)

turbopack/crates/turbopack-core/src/file_source.rs
```diff
@@ -94,7 +102,10 @@ impl Asset for FileSource {
                     })
                     .cell())
                 }
-                _ => bail!("Invalid symlink"),
+                LinkContent::NotFound => {
+                    Ok(AssetContent::File(FileContent::NotFound.resolved_cell()).cell())
```
lukesandberg, line 106:
```suggestion
                   // should never happen, we already know the symlink exists but there is a TOCTOU issue
                    Ok(AssetContent::File(FileContent::NotFound.resolved_cell()).cell())
```

turbopack/crates/turbopack-ecmascript/src/references/external_module.rs
```diff
@@ -497,22 +497,33 @@ impl Asset for ExternalsSymlinkAsset {
...
-            .parent()
-            .get_relative_path_to(&*this.chunking_context.output_root().await?)
-            .context("path must be inside output root")?;
-
-        let target = format!(
-            "{path_to_output_root}/{output_root_to_project_root}/{project_root_to_target}",
-        )
-        .into();
+        let project_root = this.chunking_context.root_path().await?;
+        let target = if this.target.fs == project_root.fs {
```
lukesandberg, line 501: this is an odd concern for the caller to maintain

WriteLinkTarget should just wrap a FileSystemPath (or be a FileSystemPath), and write_link should produce an appropriate encoding

packages/next/src/build/utils.ts
bgw: We can't implement in rust until we've worked out all the kinks with the wasm build because this codepath can be used with webpack without a native turbopack binary.

Synchronous file IO can be a follow-up in a different PR.

packages/next/src/build/nft.ts
```diff
@@ -0,0 +1,216 @@
...
+  )
+}
+
+function mapBasePath(
+  traceFileDirectory: string,
+  baseRoot: string,
+  relativePath: string
+): { source: string; destination: string } {
+  const source = path.resolve(traceFileDirectory, relativePath)
+  if (!isInside(baseRoot, source)) {
```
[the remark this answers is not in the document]
bgw: :robot: Done -- standalone copying now skips base-root escapes instead of throwing, emits one warning with the total omitted count and up to five random examples, and the webpack production test verifies the build completes.

packages/next/src/build/nft.ts
bgw: :robot: Done -- compatibility is preserved per callsite: adapter mapping can return destinations outside baseRoot, while standalone copying handles those escapes by omitting them and warning.

packages/next/src/build/nft.ts
bgw: :robot: Done -- base-root escapes no longer throw. Standalone output omits them and emits an aggregate warning; paths escaping an additional root still throw as invalid metadata.

packages/next/src/build/nft.ts
bgw: :robot: Done -- validation is now callsite-specific: adapters allow mappings outside the base root, standalone opts into skip-and-warn behavior, and additional-root paths remain strict.

turbopack/crates/turbo-tasks-fs/src/fs_map.rs
bgw: Will follow up in a separate PR.

crates/next-api/src/nft_json_builder.rs
bgw: :robot: Done -- root classification now uses turbo-unix-path directly, avoiding the FileSystemPath join that required output to live under the project filesystem root.

crates/next-api/src/nft_json_builder.rs
bgw: :robot: Done -- additional-root paths are now serialized relative to the .nft.json directory, and mapping resolves them from the trace file directory.

turbopack/crates/turbo-tasks-fs/src/path.rs
bgw: Seems like it's probably not worth it, too boilerplate heavy, not enough value.

packages/next/src/build/utils.ts
```diff
@@ -1513,6 +1517,30 @@ startServer({
...
+      .join('\n')
+    const warning = [
+      `${count} traced files were not included in the standalone output`,
+      'because their paths are outside of `outputFileTracingRoot`.',
+      'Random sample:',
+      sampleOutput,
+      'Set `outputFileTracingRoot` to a common parent directory',
+      'to include these files.',
+    ].join('\n')
+    Log.warn(warning)
```
bgw, line 1542: Stop random sampling, show the first 100

crates/next-api/src/nft_json_builder.rs
lukesandberg, line 137: select_location?

to_location?

turbopack/crates/turbopack-core/src/rebase.rs
```diff
@@ -77,7 +77,23 @@ impl Asset for RebasedAsset {
...
         if let Some(source) = *self.module.source().await? {
-            Ok(source.content())
+            let source_content = source.content();
+            let AssetContent::Redirect(redirect) = &*source_content.await? else {
+                return Ok(source_content);
+            };
+            // treat symlinks targets as relative to the symlink. When we rebase the symlink, we
+            // should also rebase its target path
+            let redirect = WriteLinkContent {
+                target: FileSystemPath::rebase(
```
lukesandberg, line 87: rebase might need some logic for changing roots

lukesandberg (approved):

turbopack/crates/turbopack-core/src/rebase.rs
bgw: Oh, actually it looks like it's fine. The returned `FileSystemPath` always uses the `new_base`'s `ResolvedVc<Box<dyn FileSystem>>`.

PR #98483, opened 2026-09-10
fix: separate prefetch requests in response cache

RainYr1021:
### Fixing a bug
I found that prefetch requests and normal requests may be batched together when they use the same cache key.

I changed the batching key to include isPrefetch so that the two types of requests are handled separately.

I ran the unit tests and E2E tests, and they both passed.


RainYr1021:
Hi, just checking whether this PR needs any changes from me. The unit and E2E tests pass locally. Thanks for taking a look!

PR #98495, opened 2026-09-10
test(turbo-tasks): run bounded scope parallelism on wasm

sokra:
## What

Run `test_scope_runs_in_parallel` on wasm instead of ignoring it. The wasm wrapper owns its Tokio runtime and uses `shutdown_background()`, and the timing bound allows Node Worker startup while still requiring completion faster than the 800ms serial workload. Native keeps the existing `#[tokio::test]` path.

## Why

The ignore attributed this to runtime teardown. Reproducing it showed the actual failure was the native-oriented `< 400ms` assertion: wasm completes in about 560ms, which is still genuinely parallel.

## Verification

- wasm: 1 passed in 0.56s
- native: 1 passed in 0.20s
- `cargo fmt --check -p turbo-tasks`

This is stacked on #98356.


sack1580-oss:
bc1qjg85am5rguwygufepkact7xg3yjd03ct0jnhzu

turbopack/crates/turbo-tasks/src/scope_bounded.rs
```diff
@@ -359,14 +359,25 @@ mod tests {
...
-        target_family = "wasm",
-        ignore = "timing bound is too strict for Node Worker startup"
-    )]
     async fn test_scope_runs_in_parallel() {
+        test_scope_runs_in_parallel_impl().await;
+    }
+
+    #[cfg(target_family = "wasm")]
+    #[test]
+    fn test_scope_runs_in_parallel() {
```
lukesandberg, line 370: can't this just be the way the test works in both cases?  this style works just as well for non-wasm

PR #98497, opened 2026-09-10
fix(turbopack): add `get_relative_request_to` for paths used as module requests

sokra:
### What?

Adds `get_relative_request_to`, next to `get_relative_path_to`, for callers that need a relative
path they can hand to the resolver as a module request. It returns `./c`, `../c` or `.`, where
`get_relative_path_to` returns `c`, `../c` or `.`.

Fixes `import.meta.glob`'s `base`-relative keys, which currently come back without the leading `./`:
`{ base: '../outside' }` produces `one.js` where Vite (and our own docs, and the existing execution
fixture) expect `./one.js`. That fixture is failing on `canary` today; it passes here unchanged.

### Why?

The `./` prefix is not decoration — it is what makes a string a *relative* request. Without it, `c`
resolves as the package `c` rather than the file next to `from`.

`get_relative_path_to` stopped emitting the prefix for a path below `from`, and every caller that
builds a request had to add it back by hand. Eight of them do, each with the same three lines. That
works, but it puts the requirement in the call sites instead of the API, and nothing tells a new
caller that it exists — so the ninth request-building call site, `import.meta.glob`'s `base`-relative
key, was written without it and silently produced the wrong keys. The two changes were authored
independently and each was green on its own; only together do they produce the bug, which is why CI
caught it on `canary` rather than on either pull request.

Rather than add a tenth hand-written prefix, this makes "give me a request" something you can ask
for. The failure mode it removes is the silent one: a caller that forgets is no longer possible,
because the choice is now the function you call.

### How?

`get_relative_request_to` shares its traversal with `get_relative_path_to` through one private
helper, so the two can't drift; the only difference is whether a result that would be bare gets the
prefix. `get_relative_path_to` keeps its signature and behaviour exactly, including the fast path
that returns `target` by reference when `from` is empty — the request form can't borrow there, and
the unit tests assert that difference so the fast path isn't reused by mistake. `FileSystemPath`
gains the matching method.

The eight hand-written prefixes now call the new function and drop their snippets. Deciding which
call sites to convert needed care, because the callers of `get_relative_path_to` genuinely want
different things, and the ones that want a plain path deliberately keep it:

- source-map paths reported back to the client,
- the Next.js template expander, whose invariant check expects the bare form,
- the paths written into NFT files, which are data rather than requests.

Two locals that hold requests rather than paths are renamed accordingly, and one leftover
`strip_prefix("./")` is removed — it had become a no-op undoing a prefix that is no longer produced.

### Testing

The `import-meta-glob-relative` execution fixture already asserts the correct keys and currently
fails, so the fix is verified by that fixture passing **unmodified** — its expectations were right
all along.

The unit test that covered these cases before the prefix went away is restored, now run against both
functions and against `FileSystemPath`, which had been left with no coverage of either method. Its
original assertions turn out to describe the request form exactly, so they pin that
`get_relative_request_to` reproduces the earlier contract rather than merely resembling it.

Beyond that: the full `turbopack-tests` suite, and the `next-server-nft` and `next-taskless` tests,
which are what would break if a call site that wants a plain path had been converted by mistake.

### DIFF
--- a/turbopack/crates/turbo-tasks-fs/src/path.rs
+++ b/turbopack/crates/turbo-tasks-fs/src/path.rs
@@ -12,7 +12,9 @@ use turbo_tasks::{
     turbobail, turbofmt,
 };
 use turbo_tasks_hash::HashAlgorithm;
-use turbo_unix_path::{get_parent_path, get_relative_path_to, join_path, normalize_path};
+use turbo_unix_path::{
+    get_parent_path, get_relative_path_to, get_relative_request_to, join_path, normalize_path,
+};
 
 use crate::{
     DirectoryContent, DirectoryEntry, FileContent, FileJsonContent, FileMeta, FileSystem,
@@ -97,8 +99,13 @@ impl FileSystemPath {
         }
     }
 
-    /// Returns a unix-style path of `other` relative to `self`. Supports traversing upwards (`../`)
-    /// within the filesystem.
+    /// Returns a unix-style path of `other` relative to `self`, as a plain path: `dir/file.js`,
+    /// `../file.js`, or `.`. Supports traversing upwards (`../`) within the filesystem.
+    ///
+    /// Returns [`None`] when the two are on different filesystems.
+    ///
+    /// The result is not prefixed with `./`, so it is a path and not a module request. Use
+    /// [`FileSystemPath::get_relative_request_to`] to build an import specifier.
     pub fn get_relative_path_to(&self, other: &FileSystemPath) -> Option<RcStr> {
         if self.fs != other.fs {
             return None;
@@ -106,6 +113,25 @@ impl FileSystemPath {
 
         Some(match get_relative_path_to(&self.path, &other.path) {
             Cow::Borrowed(path) if std::ptr::eq(path, other.path.as_str()) => other.path.clone(),
+            Cow::Borrowed(path) => path.into(),
+            Cow::Owned(path) => path.into(),
+        })
+    }
+
+    /// Returns a unix-style path of `other` relative to `self`, as an explicitly relative module
+    /// request: `./dir/file.js`, `../file.js`, or `.`. Supports traversing upwards (`../`) within
+    /// the filesystem.
+    ///
+    /// Returns [`None`] when the two are on different filesystems.
+    ///
+    /// The `./` prefix is what makes the result a relative request rather than a reference to a
+    /// package of that name. Use [`FileSystemPath::get_relative_path_to`] for a plain path.
+    pub fn get_relative_request_to(&self, other: &FileSystemPath) -> Option<RcStr> {
+        if self.fs != other.fs {
+            return None;
+        }
+
+        Some(match get_relative_request_to(&self.path, &other.path) {
             Cow::Borrowed(path) => path.into(),
             Cow::Owned(path) => path.into(),
         })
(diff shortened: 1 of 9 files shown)


mischnic (approved):

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -111,6 +118,25 @@ impl FileSystemPath {
...
+    /// package of that name. Use [`FileSystemPath::get_relative_path_to`] for a plain path.
+    pub fn get_relative_request_to(&self, other: &FileSystemPath) -> Option<RcStr> {
+        if self.fs != other.fs {
+            return None;
+        }
+
+        Some(match get_relative_request_to(&self.path, &other.path) {
+            Cow::Borrowed(path) => path.into(),
+            Cow::Owned(path) => path.into(),
+        })
```
bgw, line 137: There's an impl for `From<Cow<'_, str>> for RcStr`: https://turbopack-rust-docs.vercel.sh/rustdoc/turbo_rcstr/struct.RcStr.html#impl-From%3CCow%3C'_,+str%3E%3E-for-RcStr

That impl should have a fast-path for converting from `Cow::Owned`, but that logic should live there, not here.

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -768,6 +794,102 @@ mod tests {
...
+            assert_eq!(from_path.get_relative_path_to(&on_other_fs), None);
+
+            anyhow::Ok(())
+        })
+        .await
+        .unwrap();
+    }
+
+    /// The cases this covers are the ones `get_relative_path_to` was asserted against before it
+    /// stopped prefixing `./`, so they pin that the request form still produces them.
```
bgw, line 851: Claudism: This is a comment about how the code used to look.

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -768,6 +794,102 @@ mod tests {
...
+        target: &str,
+    ) -> (FileSystemPath, FileSystemPath) {
+        (
+            FileSystemPath::new_normalized_unchecked(fs, from.into()),
+            FileSystemPath::new_normalized_unchecked(fs, target.into()),
+        )
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_get_relative_path_to() {
```
bgw, line 810: IMO, this test is overkill because it's 90% the same as the test in `turbopack/crates/turbo-unix-path/src/lib.rs`. There's some small value in checking that the cross-filesystem bailout works, but it's a very simple check that we're not likely to get wrong. This just feels like 60 LOC to test a 3 LOC branch.

turbopack/crates/turbo-tasks-fs/src/path.rs
```diff
@@ -768,6 +794,102 @@ mod tests {
...
+            anyhow::Ok(())
+        })
+        .await
+        .unwrap();
+    }
+
+    /// The cases this covers are the ones `get_relative_path_to` was asserted against before it
+    /// stopped prefixing `./`, so they pin that the request form still produces them.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_get_relative_request_to() {
```
bgw, line 853: I think this test is also too duplicative of the one in `turbopack/crates/turbo-unix-path/src/lib.rs`

bgw (approved):
Thanks, this is a bit better than the solution I had of updating all the callsites to prefix `./`. Left some non-blocking comments.

Issue #98498, opened 2026-09-10
Streamed metadata wrapper is a <div> as the first child of <body>, so any third-party <div> prepended before hydration breaks hydration and gets deleted


nickkosmides:
### Link to the code that reproduces this issue

Reproduces on any App Router site with streaming metadata (the default). Minimal repro: a root layout whose `<body>` starts with a non-`<div>` (e.g. a skip link `<a>`), plus one inline script that prepends a `<div>` to `<body>` before hydration:

```tsx
// app/page.tsx
export default function Page() {
  return (
    <>
      <script
        dangerouslySetInnerHTML={{
          __html: `var el=document.createElement('div');el.id='third-party';el.textContent='x';document.body.insertBefore(el,document.body.firstChild);`,
        }}
      />
      <p>page</p>
    </>
  )
}
```

I can turn this into a repository if that helps; the patch PR linked below also adds it as an e2e test under `test/e2e/app-dir/metadata-streaming`.

### To Reproduce

1. `next build && next start` (or `next dev`) with the page above.
2. Open it. The inline script runs during parsing and puts `<div id="third-party">` in front of everything in `<body>`, the way consent managers (InMobi/Quantcast Choice: `#qc-cmp2-container`), chat widgets and some extensions do when their script finishes before hydration.
3. Console: `Hydration failed because the server rendered HTML didn't match the client` (React #418 in production). `#third-party` is gone from the DOM afterwards.

Change the prepended element to an `<aside>` and nothing happens: no error, node kept.

### Current vs. Expected behavior

**Current:** with streaming metadata, `MetadataWrapper` renders

```tsx
<div hidden>
  <MetadataBoundary>
    <Suspense name="Next.Metadata">…</Suspense>
  </MetadataBoundary>
</div>
```

and because it is rendered above the root layout, React places it as the **first child of `<body>`** on every page (`next/dist/export/worker.js` forces `serveStreamingMetadata: true`, so it is in every prerendered document too; `htmlLimitedBots` only affects request-time renders).

[...]

PR #98499, opened 2026-09-10
Render the streamed-metadata wrapper as a custom element, not a <div>

### LINKED ISSUE #98500: Streamed metadata wrapper is a <div> as the first child of <body>, so a third-party <div> prepended before hydration breaks hydration and gets deleted

https://github.com/nickkosmides/next-metadata-wrapper-prepend-repro

### To Reproduce

1. `npm install && npm run build && npm run start` (verified against `next@canary`), open http://localhost:3000 with the console open.
2. `app/page.tsx` has an inline script that prepends `<div id="third-party">` to `<body>` during parsing — before hydration — the way consent managers (InMobi/Quantcast Choice: `#qc-cmp2-container`), chat widgets and some extensions do when their script finishes before React does. The root layout's own first body child is an `<a>` skip link, not a `<div>`.
3. Console: `Minified React error #418` (hydration mismatch). `#third-party` is gone from the DOM afterwards.
4. Change `'div'` to `'aside'` in `app/page.tsx`: no error, node kept.

(An earlier copy of this report, #98498, was auto-closed for lacking a repository link; this one supersedes it.)

**Current:** with streaming metadata (the default), `MetadataWrapper` in `packages/next/src/lib/metadata/metadata.tsx` renders

```tsx
<div hidden>
  <MetadataBoundary>
    <Suspense name="Next.Metadata">…</Suspense>
  </MetadataBoundary>
</div>
```

above the root layout, so React places it as the **first child of `<body>`** on every page. It is in every prerendered document too: `next/dist/export/worker.js` forces `serveStreamingMetadata: true`, and `htmlLimitedBots` only affects request-time renders.

React 19 hydrates `<body>` children by tag name and skips a foreign node only when its tag differs (`canHydrateInstance`, `inRootOrSingleton`). So any third party that prepends a `<div>` to `<body>` before hydration finishes has that `<div>` claimed for Next's wrapper; the wrapper's expected child (a Suspense comment) is not found inside it; hydration fails at the root; React regenerates the document with `clearContainerSparingly`, which removes every element in `<body>` that is not a script/style/stylesheet link — the third party's DOM is deleted.

An application cannot avoid this: the wrapper precedes anything the root layout renders, so an app whose own first body child is not a `<div>` is exposed purely because of the framework's wrapper. React's source describes the tag-name skip as intentional and a prepended `<div>` as "an edge case"; Next's wrapper turns that edge case into the default for every App Router site.

**Expected:** a third-party element prepended to `<body>` is skipped by hydration, as React already does for any tag other than the one Next renders first. React's requirement (#77620) is a *host element* around the top-level Suspense boundary, not a `<div>`; a custom element such as `<next-metadata hidden>` satisfies it, renders and hydrates identically, and is a tag nothing prepends. Next already uses `<next-route-announcer>` the same way. Patch: #98499.

```
Next.js: canary (16.4.0-canary.25) in the repro; 16.3.4 in production
React: 19.2.8
Node: 24.15.0, macOS 15 / Linux (production)
Browsers: Chromium 141, Safari 18, in-app browsers
```

Metadata, Hydration

next build (local), next start (local), Vercel/other hosting (production)

### Additional context

Seen on a production site (Next 16.3.4) with a consented InMobi CMP: the CMP's corner "privacy" tab disappeared intermittently, and every such load carried a #418. Controlled Playwright run against that site: foreign `<div>` prepended before hydration → #418 on 2/2 loads and the node deleted; foreign `<aside>` in the same place → 0/2, node kept. Hydration finished at ~500–650 ms and the CMP's container arrived at ~900–1250 ms on a fast desktop, so the race is lost on slower devices or warm CMP caches — which is why it looks random in the wild.

nickkosmides:
### What?

Render the streamed-metadata wrapper as a custom element, `<next-metadata hidden>`, instead of `<div hidden>`, in both `metadata.tsx` and `metadata-parallel.tsx`. Adds an e2e test that prepends a third-party `<div>` to `<body>` before hydration and asserts that hydration succeeds and the node survives.

### Why?

With streaming metadata (the default, and forced for every prerendered page by the export worker), `MetadataWrapper` is rendered above the root layout, so its `<div hidden>` is the **first child of `<body>`** on every App Router page.

React 19 hydrates `<body>` children by tag name and skips a foreign node only when its tag differs (`canHydrateInstance`, `inRootOrSingleton`). So a third-party script that prepends its own `<div>` to `<body>` before hydration finishes — consent managers do exactly this (InMobi/Quantcast Choice prepends `#qc-cmp2-container`), as do chat widgets and some extensions — gets that `<div>` claimed for Next's wrapper. The wrapper's expected child (a Suspense comment) is not found inside it, hydration fails at the root (#418), and React regenerates the document with `clearContainerSparingly`, which deletes every element in `<body>` it does not own — the third party's UI included.

An application cannot work around this: the wrapper precedes anything the root layout renders, so an app whose own first body child is an `<a>` or a `<header>` is exposed purely because of the framework's wrapper. React's own source describes the tag-name skip as intentional and a prepended `<div>` as "an edge case"; a `<div>` wrapper in first position turns that edge case into the default for every site.

React's requirement here (#77620) is a *host element* around the top-level Suspense boundary, not a `<div>` specifically. A custom element satisfies it, renders identically (`hidden=""`, `display: none` from the UA stylesheet), hydrates identically, and is a tag nothing else prepends. Next already uses `<next-route-announcer>` in the same spirit.

Measured on a production site (Next 16.3.4, React 19.2.8) with a consented CMP: a foreign `<div>` prepended before hydration → #418 on 2/2 loads and the node deleted; a foreign `<aside>` in the same place → no error, node kept. Verified in jsdom that `renderToString` emits `<next-metadata hidden=""><!--$--><!--/$--></next-metadata>` and that `hydrateRoot` with a prepended foreign `<div>` reports no recoverable error and keeps the node.

Fixes #98500 (supersedes #98498, which the bot closed for lacking a repository link; reproduction: https://github.com/nickkosmides/next-metadata-wrapper-prepend-repro)

### How?

- `HIDDEN_METADATA_WRAPPER_TAG = 'next-metadata'` in `lib/metadata/constants.ts`, used via `React.createElement` in both wrappers (no JSX intrinsic-element typing needed).
- `test/e2e/app-dir/metadata-streaming/app/third-party-prepend/page.tsx` prepends a `<div>` during parsing; the new case in `metadata-streaming.test.ts` asserts the wrapper is not a `<div>`, that `#third-party` is still in the DOM after hydration, and that no hydration error was logged.

Note: I authored this against a sparse checkout and could not run the monorepo's test suite locally; the patch is a tag-name change plus one test, and I will address CI feedback promptly.

### DIFF
--- a/packages/next/src/lib/metadata/constants.ts
+++ b/packages/next/src/lib/metadata/constants.ts
@@ -13,3 +13,14 @@ export const ViewportMetaKeys: { [k in keyof ViewportLayout]: string } = {
 } as const
 
 export const IconKeys: (keyof Icons)[] = ['icon', 'shortcut', 'apple', 'other']
+
+/**
+ * The element streamed metadata is rendered inside. React needs a host
+ * element around a top-level Suspense boundary, and this one becomes the
+ * first child of `<body>`. It is a custom element rather than a `<div>`
+ * because React hydrates `<body>` children by tag name: a third-party script
+ * that prepends its own `<div>` to `<body>` before hydration would otherwise
+ * be claimed for this wrapper, hydration would fail, and the client-side
+ * regeneration would delete the third party's DOM.
+ */
+export const HIDDEN_METADATA_WRAPPER_TAG = 'next-metadata'
--- a/packages/next/src/lib/metadata/metadata-parallel.tsx
+++ b/packages/next/src/lib/metadata/metadata-parallel.tsx
@@ -5,6 +5,7 @@ import type { LoaderTree } from '../../server/lib/app-dir-module'
 import type { SearchParams } from '../../server/request/search-params'
 import { createServerSearchParamsForMetadata } from '../../server/request/search-params'
 import { createServerPathnameForMetadata } from '../../server/request/pathname'
+import { HIDDEN_METADATA_WRAPPER_TAG } from './constants'
 import type { MetadataErrorType } from './metadata-resolution-primitives'
 import {
   resolveMetadataForBranch,
--- a/packages/next/src/lib/metadata/metadata-parallel.tsx
+++ b/packages/next/src/lib/metadata/metadata-parallel.tsx
@@ -126,14 +126,15 @@ export function createMetadataComponents({
         </MetadataBoundary>
       )
     }
-    return (
-      <div hidden>
-        <MetadataBoundary>
-          <Suspense name="Next.Metadata">
-            <Metadata />
-          </Suspense>
-        </MetadataBoundary>
-      </div>
+    // Not a <div>: see the same spot in ./metadata.tsx.
+    return React.createElement(
+      HIDDEN_METADATA_WRAPPER_TAG,
+      { hidden: true },
+      <MetadataBoundary>
+        <Suspense name="Next.Metadata">
+          <Metadata />
+        </Suspense>
+      </MetadataBoundary>
     )
   }
 
--- a/packages/next/src/lib/metadata/metadata.tsx
+++ b/packages/next/src/lib/metadata/metadata.tsx
@@ -10,6 +10,7 @@ import {
   resolveViewport,
 } from './resolve-metadata'
 import { isHTTPAccessFallbackError } from '../../client/components/http-access-fallback/http-access-fallback'
+import { HIDDEN_METADATA_WRAPPER_TAG } from './constants'
 import type { MetadataContext } from './types/resolvers'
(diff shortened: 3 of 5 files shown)


nickkosmides:
Thanks — the missing import in `metadata-parallel.tsx` is fixed in aaabebc.

PR #98504, opened 2026-09-10
Durable use cache: optimize env var existence checks

mischnic:
We now separately track code that accesses non-inlined env vars at runtime in two categories
- (existing) actual reads, the full value is accessed
- (new) only unset/falsy/truthy is read at runtime

For the second case, we only need to invalidate use-cache functions when the given env var transitioned betwen unset/falsy/truthy. But transitioning between two different truthy values doesn't caused invalidation.


mischnic:
* **#98558** <a href="https://app.graphite.com/github/pr/vercel/next.js/98558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#98504** <a href="https://app.graphite.com/github/pr/vercel/next.js/98504?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/next.js/98504?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `canary`


<h2></h2>

test/production/app-dir/use-cache-code-hash/use-cache-code-hash.test.ts
```diff
@@ -23,12 +30,13 @@ async function getCodeHashes(
...
           codeHash: worker?.durability?.codeHash,
-          runtimeEnvVars: worker?.durability?.runtimeEnvVars,
+          runtimeEnvVarsRead: worker?.durability?.runtimeEnvVarsRead,
+          runtimeEnvVarsExistence: worker?.durability?.runtimeEnvVarsExistence,
         })
       }
     }
   }
-
+  hashes.sort((a, b) => a.page.localeCompare(b.page))
```
lukesandberg, line 39: why this change? just more determinsim?

mischnic: it should already have been deterministic in the technical sense
this mostly makes it easier for a human to find the page you're interested in

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
```diff
@@ -1791,6 +1837,10 @@ impl VisitAstPath for Analyzer<'_, '_> {
                 prop: prop_value,
                 ast_path: as_parent_path_in(self.arena, ast_path),
                 span: member_expr.span(),
+                in_truthiness_context: is_in_boolean_context(
```
lukesandberg, line 1840: we could defer this test to effect processing, this would be slightly more efficient since we would mostly not need to walk the AST path

turbopack/crates/turbopack-ecmascript/src/references/mod.rs
```diff
@@ -256,7 +257,8 @@ impl AnalyzeEcmascriptModuleResultBuilder {
             successful: false,
             source_map: None,
             cjs_static_exports: None,
-            env_var_info_runtime: Default::default(),
+            env_var_info_runtime_read: Default::default(),
+            env_var_info_runtime_existence: Default::default(),
```
lukesandberg: should this just be a map<string, AccessMode> instead of 2 sets

lukesandberg (approved):

turbopack/crates/turbopack-ecmascript/src/analyzer/graph/visitor.rs
mischnic: no, we can't. I first wanted to do it like this.

in the effect processing stage, we have `[AstParentKind]`, which is only the type of the AST node, not the actual AST node. So here, we can only tell if this is `anyFn(process.env.FOO)` and not if it's actually `Boolean(process.env.FOO)`

PR #98507, opened 2026-09-10
[ci] Remove Turbopack test manifest pipeline

eps1lon:
The manifest is no longer considered for skipping test since https://github.com/vercel/next.js/pull/81170. It also didn't update anymore automatically so I removed updating areweturboyet.com which was broken for a while now anyway when we switched away from static tokens.


.github/workflows/turbopack-nextjs-build-integration-tests.yml
```diff
@@ -1,27 +0,0 @@
...
-    - cron: '0 6 * * *'
-  workflow_dispatch: {}
-
-jobs:
-  test-dev:
-    name: Next.js integration tests
-    permissions:
-      contents: read
-      id-token: write
-    uses: ./.github/workflows/integration_tests_reusable.yml
```
eps1lon, line 14: Rspack test manifest still uses this workflow.

.github/workflows/turbopack-nextjs-build-integration-tests.yml
```diff
@@ -1,27 +0,0 @@
-name: Turbopack Next.js production integration tests
```
mischnic, line 1: Are we actually running all of these tests in one of the other jobs? 
We still have `TURBOPACK_BUILD` and `TURBOPACK_DEV` in a few test files?

bgw (approved):
areweturboyet.com is 100% dead. I'd like to keep maintaining arewerspackyet.com, but I think realistically I don't have the time to do that, so I guess this is okay.

.github/workflows/turbopack-nextjs-build-integration-tests.yml
eps1lon: This is the same as the turbopack-* shards:
- https://github.com/vercel/next.js/blob/ec62d6a48fde98ccbda4eb884cdbc58cfe585b52/.github/workflows/build_and_test.yml#L490-L491
- https://github.com/vercel/next.js/blob/ec62d6a48fde98ccbda4eb884cdbc58cfe585b52/.github/workflows/build_and_test.yml#L445-L446

eps1lon:
> areweturboyet.com is 100% dead. I'd like to keep maintaining arewerspackyet.com, but I think realistically I don't have the time to do that, so I guess this is okay.

Any removal for rspack related things would be unintentional. Can you point out how this affects rspack?

PR #98511, opened 2026-09-10
Prevent synchronous IO from entering Cached Navigations

unstubbable:
Cached Navigations incorrectly reused uncached timestamps as static page content. A repeat visit showed the previous request's `Date.now()` result while the new response was still pending. This affected both dynamic RSC responses and the RSC payload embedded in the initial HTML.

The production renderer left synchronous IO untracked because it assumed the build had already checked it. A private cache exposed the gap: the static prerender suspended at the cache call and never reached the `Date.now()` after it. A production request resolved the private cache during the static stage and included the uncached timestamp in the reusable response.

Production staged renders now end the static stage when synchronous request-time IO occurs. They mark the interrupted static and shell prefixes as empty, while the rest of the response renders normally. The client skips those prefixes instead of waiting for a Flight root that the interrupted stage may not have emitted.

The regression tests cover dynamic RSC navigation, initial HTML hydration, and full prefetching with a paramless private-cache fixture.

### DIFF
--- a/packages/next/src/server/app-render/app-render.tsx
+++ b/packages/next/src/server/app-render/app-render.tsx
@@ -1071,13 +1071,14 @@ async function generateStagedDynamicFlightRenderResultNode(
 
   const selectStaleTime = createSelectStaleTime(experimental)
   const staleTimeIterable = new StaleTimeIterable()
+  const prefetchMode = await getPrefetchingModeForPage(renderOpts, loaderTree)
 
   const stageController = new StagedRenderingController({
     abortSignal: null,
     abandonController: null,
     // Synchronous request-time data ends the static stage before its result can
     // enter the Cached Navigation.
-    syncIO: SyncIOMode.AllowedInRuntimeOrDynamic,
+    syncIO: getSyncIOMode(prefetchMode),
     finalStage: null,
   })
 
@@ -1110,10 +1111,7 @@ async function generateStagedDynamicFlightRenderResultNode(
   // fill caches and then spawn a final runtime prerender whose result stream
   // is embedded in the RSC payload. This is gated because it adds extra server
   // processing and increases the response payload size.
-  if (
-    Boolean(renderOpts.partialPrefetching) ||
-    (await anySegmentHasPartialPrefetchingEnabled(loaderTree))
-  ) {
+  if (prefetchMode === PrefetchingMode.Partial) {
     // Create a mutable cache that gets filled during the dynamic render.
     const prerenderResumeDataCache = createPrerenderResumeDataCache()
     requestStore.resumeDataCache = prerenderResumeDataCache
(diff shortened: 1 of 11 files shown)


gnoff (approved):

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -1178,8 +1177,18 @@ async function generateStagedDynamicFlightRenderResultNode(
...
       void countShellAndStaticStageBytes(staticStream, stageController).then(
         (byteLengths) => {
-          staticStageByteLengthDeferred.resolve(byteLengths[RenderStage.Static])
-          shellByteLengthDeferred.resolve(byteLengths[RenderStage.ShellStatic])
+          // A sync interruption can precede the Flight root row. Do not expose
+          // that incomplete prefix as a cacheable payload.
+          staticStageByteLengthDeferred.resolve(
+            stageController.getSyncInterruptReason()
+              ? 0
+              : byteLengths[RenderStage.Static]
```
lubieowoce: why is this needed? StagedRenderingController already gets out of `Static` when we hit sync IO:
https://github.com/vercel/next.js/blob/afb0dbac6a0daeeca12d72759e51bd3493e16a0d/packages/next/src/server/app-render/staged-rendering.ts#L176-L180

so any content emitted after twouldn't make it into the cached nav anyway

and just because it was aborted prematurely doesn't mean it'll be useless, because it may have been in a suspense boundary. sure, we'd error for this in build, but having some content seems better than none at all?

lubieowoce: oh did this hang in `decodeBufferedStage` because the root is blocked? in that case we should probably just reject the decode after `setTimeout(_, 0)`, that's not the only scenario that can block the root. e.g. we can have

```ts
export default async function RootLayout({ children }) {
  // not blocking during build, blocking after revalidation
  if (await cacheThatBecomesTrueAfterRevalidation()) {
    await dynamicData() 
  }
  // ...
}
```

which (off the top of my head, untested) should be fixed by something like this:

```ts
export function decodeBufferedStage<T>(
  buffer: Uint8Array,
  headers: RequestHeaders | undefined
): Promise<T> {
  const stream = new ReadableStream<Uint8Array>({
    start(controller) {
      controller.enqueue(buffer)
      controller.close()
    },
  })
  return new Promise<T>((resolve, reject) => {
    // The stream is buffered, so the decode should yield a root microtaskily.
    // If it doesn't, then the root must be blocked.
    const timeout = setTimeout(() => {
      reject(new Error('Root is blocked.'))
      stream.cancel()
    });

    createFromNextReadableStream<T>(stream, headers, {
      allowPartialStream: true,
    }).then((root) => {
      clearTimeout(timeout)
      resolve(root);
    }, reject);
  });
}
```
(we may also just want to return a hanging promise, but if we know it's useless bc it blocks, then i guess there's no point)

packages/next/src/server/app-render/app-render.tsx
```diff
@@ -1075,10 +1075,9 @@ async function generateStagedDynamicFlightRenderResultNode(
   const stageController = new StagedRenderingController({
     abortSignal: null,
     abandonController: null,
-    // TODO(cached-navs): this assumes that we checked during build that there's no sync IO.
-    // but it can happen e.g. after a revalidation or conditionally for a param that wasn't prerendered.
-    // we should change this to track sync IO, log an error and advance to dynamic.
-    syncIO: SyncIOMode.Untracked,
+    // Synchronous request-time data ends the static stage before its result can
+    // enter the Cached Navigation.
+    syncIO: SyncIOMode.AllowedInRuntimeOrDynamic,
```
lubieowoce: may want to use `getSyncIOMode(prefetchMode)` for consistency. in PPF we only allow sync IO in the dynamic stage so if we hit some at any point we should go straight to dynamic

...not that it matters very much, because we don't care about recovering stages past Static here anyway, but at least then it won't look like an exception when it really doesn't need to be

lubieowoce (approved):

PR #98512, opened 2026-09-10
Align fallback parameter staging with shell validation

unstubbable:
`next dev` reported missing Suspense boundaries in layouts that the build accepted. For `/[top]/items/[bottom]`, `generateStaticParams` returned `[{ top: 't1' }]`, and the page wrapped its `bottom` access in Suspense. A request for `/t2/items/b2` still reported the layout's access to `top` as an error. Development treated both parameters as unresolved because the requested `top` value was not generated, although the required static shell only needed to defer `bottom`.

Production Cached Navigations used the same overly broad parameter set and omitted eligible static content from repeat visits. Resumes also reconstructed that set from the original build manifest, even after an on-demand prerender had produced a more complete shell.

This replaces the alternative proposed in #98460. That proposal fixes the development error by selecting a separate fallback parameter set for validation while keeping response staging unchanged. The two sets are not intended to differ for the same shell target. Correcting only validation would preserve the incorrect staging decision and leave production Cached Navigations without the eligible static content.

Staging and static-shell validation now use one `stagedFallbackParams` set for each selected shell target. Required partial shells retain their unresolved parameters, even when a later request can complete them. Prerenders record their parameter set in postponed state, and resumes use that recorded set rather than reconstructing it from the generic source.

Dynamic RSC requests now read and revalidate the completed-shell cache key, so they find partial artifacts that the fully resolved pathname lookup missed. Request metadata and `RequestStore` both expose the set as `stagedFallbackParams`. Action-only fallback detection checks actual unresolved parameters instead of treating deferred values as missing.

### DIFF
--- a/packages/next/src/build/index.ts
+++ b/packages/next/src/build/index.ts
@@ -338,6 +338,11 @@ export interface DynamicPrerenderManifestRoute
   remainingPrerenderableParams?: readonly FallbackRouteParam[]
 
   /**
+   * Whether this candidate must produce a nonempty static shell.
+   */
+  throwOnEmptyStaticShell?: boolean
+
+  /**
    * When defined, it describes the revalidation configuration for the fallback
    * route.
    */
@@ -3888,6 +3893,8 @@ export default async function build(
                   experimentalPPR: isRoutePPREnabled,
                   remainingPrerenderableParams:
                     route.remainingPrerenderableParams,
+                  throwOnEmptyStaticShell:
+                    prerenderCandidate?.throwOnEmptyStaticShell,
                   renderingMode: isAppPPREnabled
                     ? isRoutePPREnabled
                       ? RenderingMode.PARTIALLY_STATIC
--- a/packages/next/src/build/templates/app-page-runtime.ts
+++ b/packages/next/src/build/templates/app-page-runtime.ts
@@ -34,6 +34,7 @@ import { UNDERSCORE_NOT_FOUND_ROUTE } from '../../shared/lib/entry-constants' wi
 import {
   getFallbackRouteParams,
   getPlaceholderFallbackRouteParams,
+  getStagedFallbackParams,
   buildDynamicSegmentPlaceholder,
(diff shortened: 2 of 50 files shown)


gnoff:
Flagging an inherited integration issue with the completed-shell reuse here. The problematic reset already exists in the parent revision, so this is **not a regression introduced by this PR**.

With `cacheComponents`, `partialPrefetching`, and `experimental.cachedNavigations` enabled, the new lookup correctly loads the completed shell and its Resume Data Cache (RDC). But [`generateStagedDynamicFlightRenderResultNode` then replaces that RDC with an empty mutable cache](https://github.com/vercel/next.js/blob/e61c2dd0fd344c3f007327f15f5181058e451e7b/packages/next/src/server/app-render/app-render.tsx#L1113-L1119):

```ts
const prerenderResumeDataCache = createPrerenderResumeDataCache()
requestStore.resumeDataCache = prerenderResumeDataCache
```

The [HTML resume path has the same reset](https://github.com/vercel/next.js/blob/e61c2dd0fd344c3f007327f15f5181058e451e7b/packages/next/src/server/app-render/app-render.tsx#L3901-L3906).

Reproduced against `e61c2dd0fd3` in production with `/items/[top]/[bottom]`, where `generateStaticParams` supplies only `top`, and the top layout renders a random token from a `use cache` function:

1. Request `/items/novel/FIRST` and wait for the upgraded `/items/novel/[bottom]` shell to be written.
2. Request `/items/novel/SECOND` as RSC.
3. The response has a different token, even though the original token is present in the completed shell's saved RDC. Cache diagnostics confirm a hit for `/items/novel/[bottom]`, followed by an RDC miss for the same cached function/key.

The reproduction uses a custom cache handler that evicts entries after writing, so ordinary cache hits cannot hide the lost RDC. With only `cachedNavigations` disabled, the same probe passes, including stale-shell revalidation and checks that `bottom` is not baked into the shared artifact.

gnoff (approved):

Issue #97802, opened 2026-08-24
Turbopack build saturates 4 GB and stalls or OOMs


shunkakinoki:
### Link to the code that reproduces this issue

https://github.com/shunkakinoki/next-turbopack-build-memory-reproduction

### To Reproduce

1. Clone the reproduction repository.
2. Run `bun install --frozen-lockfile --minimum-release-age=0`.
3. Run `bun run reproduce -- canary`.

The command generates 100 App Router routes and 12,000 deterministic client modules (79.8 MiB of TS/TSX), then runs `next build` in a Linux container constrained to 2 CPUs and 4 GB of memory. It samples cgroup memory once per second and bounds the build to 180 seconds.

The current canary result is captured in [this native Linux x64 workflow run](https://github.com/shunkakinoki/next-turbopack-build-memory-reproduction/actions/runs/32740714530). The same fixture on 16.3.2 is captured in [this run](https://github.com/shunkakinoki/next-turbopack-build-memory-reproduction/actions/runs/32740825895).

A smaller control completes:

```bash
REPRO_COMPONENTS_PER_ROUTE=80 bun run reproduce -- 16.3.2
```

### Current vs. Expected behavior

Current behavior:

- Next.js 16.3.2 remains in `Creating an optimized production build ...`, reaches the 4,096 MiB cgroup ceiling, records 206,580 at-limit events, and is killed by the kernel with exit 137.
- Next.js 16.4.0-canary.4 remains in the same compilation phase for the full 180-second bound, reaches 4,096 MiB, and records 260,056 at-limit events before the harness stops it with exit 124.
- The 8,000-module control completes on the same runner and limits in 65 seconds, although it also peaks at 4,095 MiB.

Expected behavior:

`next build` should complete without retaining the entire 4 GB working set, or fail promptly with an actionable memory error. It should not remain in compilation while continuously pinned at the cgroup ceiling.

### Provide environment information

```text
Operating System:
  Platform: linux
  Arch: x64
  Kernel: 6.17.0-1022-azure

Container limits:
  CPU quota: 2
  Memory: 4 GB
  Memory + swap: 4 GB total (swap cannot extend the limit)

Runner host reported by next info:
  Available CPU cores: 4
  Available memory: 15990 MB

Binaries:
  Bun: 1.4.0
  Node value reported by next info under Bun: 26.3.0

Relevant packages:
  next: 16.4.0-canary.4
  react: 19.2.8
  react-dom: 19.2.8
[...]

PR #98522, opened 2026-09-10
fix(turbopack): intern analyzer AST paths to cut compile-phase memory

gabriel-laet:
### Fixing a bug

- Related issues: #97802 (compile-phase OOM during `Creating an optimized production build`; this PR was verified on a reduced fan-out, not the original 100×12k / 4 GB harness)
- Tests added: `cargo test -p turbopack-ecmascript` on the original tree (563 passing, including unchanged analyzer graph/effect snapshots and 5 new `ast_path` tests). `cargo check -p turbopack-ecmascript` on current `canary`. Did not run clippy, `turbopack-tests`, or e2e locally (CI will).
- Signed commits: SSH-signed. If GitHub shows unverified, the signing key still needs to be added on the account (`admin:ssh_signing_key`).

## What?

Turbopack's ECMAScript analyzer stored a full `Vec<AstParentKind>` from the module root to every effect and every retained `CodeGen` entry.

In a left-nested `a + b + c + …` chain the n-th term is n levels deep, so those buffers are **O(M²) per large module**. The whole app graph is analyzed before any codegen, so hundreds of such modules at once pin tens of GB during `next build --turbopack` and never print `Compiled successfully`. `.next` stays ~1 MB.

This is **not** N copies of the module graph. The graph was already shared. The quadratic term is the AST **path** storage.

`AstPath` is now a persistent linked list with prefix sharing. An interner on the analyzer (DFS visit) allocates only the suffix that changed. Visitors still get the same `Vec<AstParentKind>` (transient `to_vec()`). Also keys `AssetContext::resolve_asset` by directory instead of the importing file (fewer tasks; not the memory win).

## Why?

Related to #97802. Repro: https://github.com/gabriel-laet/nimbus-endpoint-fanout (defaults: 180 `route.ts` + 180 client pages, 1200 cyclic schema + 1200 cyclic kit). Linux 8 vCPU, `MemoryMax=16G`, Next 16.4.0-canary.15, `turbopackFileSystemCacheForBuild: false`:

| Build | `generate:single` | `generate:records` |
|---|---|---|
| canary.15 published native | Compiled 11s, 1.31 GB RSS | **SIGKILL 137** after 74s, 16.8 GB RSS, never `Compiled`, `.next` 184 KB |
| this patch (native overlay) | Compiled 6s, 1.15 GB | **Compiled 39s**, **4.06 GB RSS**, `.next` 128 MB |

`turbopackFileSystemCacheForBuild` was already off. Failure is JS compilation, not SSG or persistence.

## How?

- New `turbopack-ecmascript/src/ast_path.rs`: prefix-sharing `AstPath` + `AstPathInterner`. Effects, early-return bookkeeping, and retained `CodeGen`s hold `AstPath`. Node drop is iterative (paths can be thousands deep).
- Analyzer walks DFS; consecutive effects share almost the whole prefix. Chain case goes from O(M²) to O(M) path entries.
- Codegen visitors still receive `Vec<AstParentKind>`.
- `resolve_asset` keyed by directory (26% fewer tasks on a small fan-out; little RAM).

## Caveats

- **Public API:** `AstPath` is no longer a slice; `resolve_asset` takes a lookup directory. Out-of-repo callers need updates.
- **Persistent cache:** `AstPath` encodes as `bincode::serde::Compat<Vec<AstParentKind>>`, intended to match the old `#[bincode(with_serde)]` newtype. Cache compatibility was **not** verified; a cache version bump may be needed.
- Prefix sharing depends on DFS order for **memory only**, never correctness. Interner is `RefCell` on `Analyzer`.
- With `enable_types` (turbopack-cli only) the types branch uses a `dir/_` placeholder origin, which can change those error locations.
- After this, fan-out of large modules is still **linear** in (endpoints × module size). The 180×1200 case still needs ~4 GB on Linux.
- Does not address the `import()` fan-out (N server entries × D dynamic imports); that is #98529. `pnpm generate:async` is unchanged by this PR: 2.7 → 2.9 min at the 2400 size, SIGKILL 137 at the default 6000 size either way.
- Did not run the issue author's original 100 routes × 12k client modules / 2 CPU / 4 GB harness. Did not bisect the 16.3.0 regression.

## Do not merge as

An interning / turbo-tasks “N copies of the kernel” fix. The kernel was already interned.


bgw:
Thank you for the attempted contribution, but I must close this PR for a few reasons:
- Yes, this can be quadratic, but only with deeply nested expressions (like your example case of a binary operation). These are relatively rare, but I admit that they do happen.
- Most `AstPath`s are transient, and so if a path does use a lot of memory but is transient, we don't care that much (at least from a memory usage perspective). `AstPath`s only get persisted in rare cases (e.g. they're a path that references an import).

You need to satisfy both of these bullet points for this to become a significant source of memory usage.

Interning is only potentially useful once we've determined that an `AstPath` must be persisted outside of the lifetime of the `Analyzer` instance, but this PR eagerly interns everything.

This problematic because the interning operation is rather expensive. So this is making a uncommon code pattern that exists in your synthetic benchmark use less memory at the cost of making common code patterns much slower.

---

**Additionally:** Our PR template and `AGENTS.md` have instructions to tell agents not to write entire PR descriptions, issue descriptions, or comments for external contributors.

You're free (and even encouraged!) to use an agent to assist you in writing code, investigating issues, collecting data, and in understanding the changes needed. But you must write your own description as evidence that you understand your own PR. You've clearly worked around these restrictions (here, in https://github.com/vercel/next.js/pull/98529, and on the associated issue). Please don't do that.

We receive a huge number of pull requests. We have these rules for external contributors because it takes us a huge amount of time and effort to sort through pull requests and issues, and all we're asking is that you write your own descriptions and comments by hand.

gabriel-laet:
@bgw sorry for any inconvenience. just wanted to help creating the scenario to reproduce.

all evidence is cloned from a production private repo (therefore that's why it's AI generated to simulate as close as possible). 

we just can't use turbopack anymore unfortunately and we're trying to find a fix for it and thought this patch could be useful

bgw:
I understand. It can be useful to have AI generate reproductions, but there's also a lot of risk in cases like this what it will find some way to "cheat" and create an unrealistic pathological benchmark.

While we doubt it would fix your issue, we do think that interning could be somewhat useful here, but it would need to be a very different implementation that only interns the persisted data structures, possibly in a way that avoids `Arc`. Inspired by your PR, @lukesandberg was doing a small bit of investigation with an agent to see if that's easy to do.

bgw:
It's pretty early (and may not merge), but keep an eye on https://github.com/vercel/next.js/pull/98581 for where we're going with this.

PR #98529, opened 2026-09-10
fix(turbopack): share async chunk groups across server entries

gabriel-laet:
### Fixing a bug

- Related issues: #97802 (compile-phase hang during `Creating an optimized production build`). This PR addresses a **second** root cause from that thread, independent of the analyzer AST-path intern in #98522. Not `fixes #97802` — the issue author’s original 100×12k / 4 GB harness was not run.
- Tests added: none new. `cargo test --release -p turbopack-core -p turbopack-nodejs` on the PR head (`de8141d2`): 154 passing (turbopack-nodejs has no unit tests; turbopack-core doctests only build under CI's `release-with-assertions` profile, pre-existing). `cargo check -p next-core` on current `canary` after cherry-pick. Did not run clippy, `turbopack-tests`, or e2e locally.
- Signed commits: SSH-signed. If GitHub shows unverified, the signing key still needs to be added as a signing key on the account.

## What?

On a production **server** build, the async chunk group for a dynamically imported module was computed **once per referencing chunk group**, keyed by that group’s available modules.

Every App Router `route.ts` / `page.tsx` is its own server entry chunk group. If N entries can reach D `import()` targets, Turbopack instantiates **N × D** async chunk groups (each with its own items, chunks, and output assets) instead of one group per dynamic target.

This PR records which chunk groups reference each group and, when the chunking context opts in, gives async loaders an availability that **only depends on the module graph**: modules/batches reachable from the async entry and present in **every** referencing group. All parents then share one loader and one async chunk group.

Enabled for **Node.js server (RSC/SSR) production** contexts only. Browser chunking is unchanged.

A second commit bounds `make_production_chunks` pair evaluation and stops allocating a RoaringBitmap intersection for every candidate pair (priority-route check uses `intersection_len` on a precomputed subset).

## Why?

Related to #97802. Independent of #98522:

- #98522 is **analysis**: O(M²) `AstPath` buffers on deep `a + b + …` chains. Repro: https://github.com/gabriel-laet/nimbus-endpoint-fanout `pnpm generate:records`. It does not change the `import()` fan-out below.
- This PR is **chunking**: N server entries × D dynamic imports. Repro: same repo, `pnpm generate:async` (default 180 routes × 6000 `import()`s = 1.08M async chunk groups without sharing; `NIMBUS_ASYNC=2400` for the smaller size).

Linux 8 vCPU, `MemoryMax=16G`, Next 16.4.0-canary.15, `turbopackFileSystemCacheForBuild: false`, RSS = `/usr/bin/time -v` max RSS:

| Build | `generate:async` 180 × 6000 | `NIMBUS_ASYNC=2400` 180 × 2400 | `generate:async-single` 1 × 6000 | `generate:records` 180 × 1200 |
|---|---|---|---|---|
| canary.15 published native | **SIGKILL 137** after 7.5 min, 16.8 GB, `.next` 132 KB / 11 files | Compiled 2.7 min, 9.8 GB | Compiled 17s, 3.2 GB | SIGKILL 137 (see #98522) |
| canary.15 + #98522 | **SIGKILL 137** after 8.1 min, 16.8 GB | Compiled 2.9 min, 9.7 GB | not run | Compiled 39s, 4.06 GB |
| + this PR (on top of #98522) | **Compiled 77s**, 3.9 GB, `.next` 206 MB / 14,121 files | Compiled 31s, 2.2 GB | Compiled 19s, 3.0 GB | Compiled 46s, 4.07 GB |

On a 32 GiB machine the stock 180 × 6000 build does finish: 4.2 min at 19.4 GiB peak, vs 31s at 4.0 GiB with this PR. Sharing async groups is the webpack-style intersection semantics for async chunks.

## How?

- `ChunkGroupInfo`: track referencing chunk groups.
- `NodeJsChunkingContext`: `shared_async_chunk_groups` option.
- `crates/next-core` production Node server contexts: opt in.
- Async loader availability = graph-derived intersection of parent availabilities.
- Modules only available in some parents go in the async chunk; runtime skips already-registered factories (`shared/runtime/runtime-utils.ts`) — output-size tradeoff, not correctness.
- `make_production_chunks`: no per-pair `a & b` allocation for the priority-route check; cap pair evaluations per group.

## Caveats

- **Server-only.** Client/browser contexts keep per-parent async variants.
- **Output size:** modules that are not in every parent’s availability set are duplicated into the shared async chunk. Runtime ignores duplicate factories.
- **API:** `ChunkingContext::is_shared_async_chunk_groups_enabled` (default `false`) and the `NodeJsChunkingContext` builder's `shared_async_chunk_groups`.
- Does not replace #98522. Land intern first or in parallel; this does not intern AST paths.
- Did not run the issue author’s original 4 GB harness.


crates/next-core/src/next_server/context.rs
```diff
@@ -1132,7 +1132,11 @@ pub async fn get_server_chunking_context_with_client_assets(
                     ..Default::default()
                 },
             )
-            .module_merging(*scope_hoisting.await?);
+            .module_merging(*scope_hoisting.await?)
+            // Server bundles don't pay for duplicated modules the way browser bundles do, and a
+            // per-referencing-chunk-group async chunk group multiplies the chunking work by the
+            // number of entries.
+            .shared_async_chunk_groups(true);
```
bgw, line 1139: This is a pretty flawed assertion. If the application is deployed to a serverless platform, cold starts matter, and therefore the per-lambda bundle size matters.

It's true that it does not matter as much as it does in the browser, but changing this flag could have significant changes that would need to be more carefully evaluated.

PR #93521, opened 2026-05-06
bundle-analyzer: add compare view with treemap diff

wbinnssmith:
Adds a compare mode to the bundle analyzer. When the user picks a historical snapshot from the new baseline picker in the top bar, the UI switches into compare mode: routes are diff'd by name and size, and the per-route view shows a `DiffTreemap` that colors each source tile by its change status (added/removed/grew/shrank/unchanged).

The compare sidebar shows the import chain for the selected tile on either the baseline ("A") or current ("B") side. Route totals are fetched in parallel so the route picker can sort by largest impact and classify routes as `changed`/`identical` rather than just name-presence.

PR #98530, opened 2026-09-10
bundle-analyzer: move comparison state into URL

wbinnssmith:
## Summary

Move bundle comparison into a dedicated `/compare` route and persist the selected baseline, comparison snapshot, route, and view in URL search parameters. This makes comparison views linkable and separates the App Router route entries from the shared client implementation.

Use root-relative data URLs and proxy historical snapshots during development so analyzer data continues to load from nested routes.

Follow-up to review feedback on #93521.

## Verification


apps/bundle-analyzer/next.config.mjs
```diff
@@ -4,6 +4,10 @@ const developmentRewrites = () => {
       source: '/data/:path*',
       destination: 'http://localhost:4000/data/:path*',
     },
+    {
+      source: '/history/:path*',
+      destination: 'http://localhost:4000/history/:path*',
+    },
```
lukesandberg: why are these needed?

delort?

lukesandberg (approved):

PR #98531, opened 2026-09-10
Add weights for parallel metadata selection

gnoff:
## Motivation

Parallel routes can contribute competing metadata and viewport values, but the default heuristic cannot express an intentional preference or fallback.

## What this does

Adds independent weight fields to Metadata and Viewport. At each parallel-route fork, the highest-weight branches form the candidate set and the existing default heuristic breaks ties. Weights inherit through a branch, and an explicit zero resets selection to the default weight.

Weight is selection policy rather than metadata output. SelectedMetadata and the new SelectedViewport type explicitly exclude it, and metadata element rendering consumes only these selected output types. Metadata and viewport weights are tracked independently. If a layer errors before returning a value, its branch retains the inherited weight while the error still propagates through that slot outlet.

## Verification

- pnpm --filter=next types
- HEADLESS=true pnpm test-start-turbo test/e2e/app-dir/metadata-streaming-parallel-routes/metadata-streaming-parallel-routes.test.ts


packages/next/src/lib/metadata/resolve-metadata-parallel.ts
```diff
@@ -748,13 +827,16 @@ async function walkMetadataTree(
       context.pathname,
       context.metadataContext
     )
+    metadataWeight = getAccumulatorWeight(metadata, metadataWeight)
```
unstubbable, line 830: Could deriving the weight from the full accumulator make selection wait for a losing slot’s `generateImageMetadata()`, even when neither branch exports a weight? That seems like it could delay the selected metadata in streaming mode and block shell flushing in blocking mode. Could we resolve static/inherited weights independently of file-based metadata?

PR #98535, opened 2026-09-10
bundle-analyzer: suspend while loading analyzer data

wbinnssmith:
## Summary

Use SWR Suspense for required analyzer datasets and move loading behavior into explicit Suspense and error boundaries.

Split selected-route and valid-comparison subtrees so resolved route and module data has non-null contracts. Keep optional history, metadata, and one-sided comparison data nullable where absence is part of the UI state.

Stacked on #98530.

## Verification

- `pnpm --filter=@next/bundle-analyzer-ui exec tsc --noEmit`
- focused ESLint for changed files


apps/bundle-analyzer/components/ui/skeleton.tsx
```diff
@@ -31,3 +31,28 @@ export function TreemapSkeleton() {
...
+
+export function TableSkeleton() {
+  return (
+    <div className="flex h-full w-full flex-col p-4" aria-hidden="true">
+      <div className="flex h-9 items-center gap-4 border-b border-border px-3">
+        <Skeleton className="h-3 w-1/2" />
+        <Skeleton className="h-3 w-24" />
+        <Skeleton className="h-3 w-24" />
+      </div>
+      {Array.from({ length: 12 }, (_, index) => (
```
lukesandberg:
```suggestion
      </div>
      { /* Chosen by a fair (d20) dice roll. */}
      {Array.from({ length: 12 }, (_, index) => (
```

lukesandberg: should it be 15? either way add a comment

apps/bundle-analyzer/components/analyzer.tsx
```diff
@@ -13,33 +22,110 @@ import { TopBar, Environment, CompareView } from '@/components/top-bar'
...
+}: {
+  children: ReactNode
+  defaultView: CompareView
+}) {
+  const [mounted, setMounted] = useState(false)
+
+  useEffect(() => setMounted(true), [])
+
+  if (!mounted) return <AnalyzerFallback view={defaultView} />
+
```
lukesandberg:
```suggestion
```

i feel like you don't need this at all

but i am only a lowly compiler engineer

apps/bundle-analyzer/lib/analyzer-data.ts
```diff
@@ -20,6 +20,23 @@ export function useHistoryIndex() {
...
+  options: SWRConfiguration<Data> = {}
+): Data {
+  const { data } = useSWR<Data>(key, fetcher, {
+    ...options,
+    suspense: true,
+  })
+
+  if (data === undefined) {
+    throw new Error(`SWR did not resolve data for ${key}`)
+  }
```
lukesandberg: dead code

if this is dead, then this abstraction isn't helping

lukesandberg (approved):

PR #98536, opened 2026-09-10
bundle-analyzer: exclude traces from output types

wbinnssmith:
## Summary

Exclude traced project files from emitted output-type classification in the bundle analyzer.

Traced TypeScript sources are synthetic provenance entries, but were classified as assets because their filenames do not use an emitted JavaScript extension. This could make JavaScript sources appear red in the treemap even when their emitted chunks are JavaScript.

## Verification


apps/bundle-analyzer/lib/analyze-data.ts
```diff
@@ -464,6 +464,7 @@ export class AnalyzeData {
       } else if (outputFile.filename.startsWith('[project]/')) {
         traced = true
```
lukesandberg, line 465: why does startsWith [project] mean traced? or server for that matter?

lukesandberg: is this a bundle-analyzer convention?

wbinnssmith: `[project]` is part of the input filesystem, and it wouldn't be considered an outputFile without having been traced, I believe

sokra (approved):

PR #98539, opened 2026-09-10
Virtualize bundle analyzer source table

wbinnssmith:
Virtualize source rows with `react-virtuoso` to avoid mounting the full table at once.

Preserve sorting, filtering, grouped package expansion, sticky headers, totals, selection, and semantic table markup. Render an additional viewport around the visible range to avoid blank content during fast scrolling.


apps/bundle-analyzer/components/ui/skeleton.tsx
```diff
@@ -56,3 +58,54 @@ export function TableSkeleton() {
     </div>
   )
 }
+
```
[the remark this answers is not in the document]
wbinnssmith: :robot: CompareView is a string enum with table and treemap values, which is assignable to this literal union. The project type-check does not report an error here.

apps/bundle-analyzer/components/diff-table.tsx
```diff
@@ -425,7 +426,75 @@ function buildRenderItems<Row extends DiffRow>(
...
+    return {
+      name: item.packageName,
+      a,
+      b,
+      delta: b - a,
+      identical: a === b,
+    }
+  }
+
+  return [...items].sort((left, right) => {
```
lukesandberg: can we use `toSorted()`

apps/bundle-analyzer/components/diff-table.tsx
```diff
@@ -425,7 +426,75 @@ function buildRenderItems<Row extends DiffRow>(
...
+
+    if (column === 'name') {
+      const comparison = a.name.localeCompare(b.name)
+      if (comparison !== 0) return comparison * sign
+    } else {
+      const aValue = a[column]
+      const bValue = b[column]
+      const comparison =
+        column === 'delta' && direction === 'desc'
+          ? Math.abs(bValue) - Math.abs(aValue)
```
lukesandberg: im not sure the magnitude comparison is correct

i think descending deltas would have all the positive changes at the top and the negative changes at the bottom

at the very least `desc` only is odd

lukesandberg (approved):

PR #98542, opened 2026-09-11
Link bundle analyzer empty states to source mode

wbinnssmith:
Offer a direct environment switch when the active Client or Server filter has no rows but the other environment has sources matching the selected file types.

Keep the generic empty state when neither environment has matching sources.


apps/bundle-analyzer/components/analyzer.tsx
```diff
@@ -680,6 +713,63 @@ function SingleAnalyzerContent({
...
+  for (let index = 0; index < data.sourceCount(); index++) {
+    const flags = data.getSourceFlags(index)
+    const hasEnvironment =
+      (environment === Environment.Client && flags.client) ||
+      (environment === Environment.Server && flags.server)
+    const hasType =
+      (typeFilter.includes('js') && flags.js) ||
+      (typeFilter.includes('css') && flags.css) ||
+      (typeFilter.includes('json') && flags.json) ||
+      (typeFilter.includes('asset') && flags.asset)
```
lukesandberg: the cute way to do this would be `typeFilter.any(f=> flags[f])`

apps/bundle-analyzer/components/analyzer.tsx
```diff
@@ -680,6 +713,63 @@ function SingleAnalyzerContent({
...
+export function AlternateEnvironmentEmptyState({
+  environment,
+}: {
+  environment: Environment
+}) {
+  const pathname = usePathname()
+  const searchParams = useSearchParams()
+  const currentLabel = environment === Environment.Client ? 'client' : 'server'
+  const alternateLabel =
+    environment === Environment.Client ? 'server' : 'client'
```
lukesandberg: `const [currentLabel, alternateLabel] = environment === Environment.Client ? ['client', 'server'] : ['server', 'client']`

is maybe better?

lukesandberg (approved):

